PR #666, opened 2019-05-31
Add ConfigureModuleItemSchema

seanh:
This schema will be used to protect the `/module_item_configurations` route. When an unconfigured assignment is launched the content item form that's rendered gets submitted to `/module_item_configurations` so that the chosen document URL can be saved in the DB. This schema will be used to validate those form submissions.

### DIFF
--- a/lms/validation/_module_item_configuration.py
+++ b/lms/validation/_module_item_configuration.py
@@ -1,0 +1,26 @@
+"""Validation for the configure_module_item() view."""
+
+import marshmallow
+from webargs import fields
+
+
+__all__ = ["ConfigureModuleItemSchema"]
+
+
+class ConfigureModuleItemSchema(marshmallow.Schema):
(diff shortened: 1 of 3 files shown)


lms/validation/_module_item_configuration.py
```diff
@@ -0,0 +1,26 @@
...
+    class Meta:
+        """Marshmallow options for this schema."""
+
+        # Silence a strict=False deprecation warning from marshmallow.
+        # TODO: Remove this once we've upgraded to marshmallow 3.
+        strict = True
+
+    def __init__(self, request):
+        super().__init__()
+        self._request = request
```
seanh: The whole `Meta` and `__init__` is duplicated on every schema class and should be moved into a base class but so far I've been too lazy. We should make `request` optional too as this schema doesn't actually need it.

Issue #352, opened 2018-11-22
Entering an invalid LMS domain breaks Google Picker (blank white box)


seanh:
A user can enter an invalid URL into the <samp>LMS Domain</samp> field at <https://lms.hypothes.is/welcome> and it will go into the database and be used, and then cause problems later on. For example it causes the Google Picker dialog to break, showing an empty white box (and error messages in the JavaScript console).

See:

https://hypothes-is.slack.com/archives/C4K6M7P5E/p1542899352076800

klemay:
I answered three tickets today where users had entered an invalid URL in the LMS Domain field— it's a common issue. The help text ("ex. yourschool.instructure.com") doesn't seem to be sufficiently guiding users. And maybe there's a term other than "LMS Domain" that we should be using...? People enter things like `Marian University` or the http version of their domain when https is production.

Additionally, users don't seem to understand that credentials generated for `yourschool.instructure.com` won't work for `yourschool-test.instructure.com` and vice versa. 

It's unfortunate that this mistake is irreversible too (meaning, users need to un-install, generate new credentials, and re-install to fix it). 

We could probably make a small change to the form at https://lms.hypothes.is/welcome to communicate:
- This field is necessary for the app to work properly
- The value of this field should match the URL where app launches are coming from (so different credentials are required for test and production domains)

seanh:
If possible (if there's a suitable Canvas API endpoint) I think some JavaScript on the `/welcome` page should use the given LMS URL, developer key, and developer secret to make a test request to the Canvas API and only allow the form to be submitted if the test request succeeds. Related: https://github.com/hypothesis/lms/issues/592

mkdir-washington-edu:
User suggestion:
"Perhaps changing where it says "LMS Domain" to add a bolded note that this URL will be used for authorization and must exactly match the domain the LMS is located?"

PR #678, opened 2019-06-07
Fix the new Google Picker in non-Canvas LMS's

seanh:
## Make the Google deployment settings required

There are only two deployments of this app - lms.hypothes.is and qa-lms.hypothes.is - and both of those have the Google Drive-related environment variables set. This app isn't intended to be used by third parties. So there's no need for the Google Drive-related settings to be optional, and the code can be simpler if it assumes that these settings are always present.

In development environments, if the developer wants to test the app without going to the trouble of creating a Google Drive client, they can just set these envvars to placeholder values as the README already says to do. Everything will work fine except that clicking on the **Select PDF from Google Drive** won't work.

We could make this even easier for developers by providing placeholder values in `development.ini` but that isn't done for any of the other settings yet.

## Fix the new Google Drive support in non-Canvas LMS's

In LMS's that don't support content-item selection, so they use the unconfigured assignment launch view to configure an assignment's document, pass the necessary config settings for Google Drive support to
the frontend so that the Google Drive button works.

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -135,10 +135,24 @@ class BasicLTILaunchViews:
         we'll save it in our DB. Subsequent launches of the same assignment
         will then be DB-configured launches rather than unconfigured.
         """
+        oauth_consumer_key = self.request.lti_user.oauth_consumer_key
+
+        lms_url = self.request.params.get("custom_canvas_api_domain")
+
+        if not lms_url:
+            lms_url = self.request.find_service(name="ai_getter").lms_url(
+                oauth_consumer_key
+            )
+
         return {
             "content_item_return_url": self.request.route_url(
                 "module_item_configurations"
             ),
+            "google_client_id": self.request.registry.settings["google_client_id"],
+            "google_developer_key": self.request.registry.settings[
+                "google_developer_key"
+            ],
+            "lms_url": lms_url,
             "form_fields": {
                 "authorization": BearerTokenSchema(self.request).authorization_param(
                     self.request.lti_user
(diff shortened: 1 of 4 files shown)


seanh:
@robertknight This means that the frontend can assume that the Google settings are always present. The logic for not showing the button if they're not present can be removed. The backend would crash on startup if these settings were not present, and the rest of the backend code assumes they're set.

lms/views/basic_lti_launch.py
```diff
@@ -135,10 +135,24 @@ def unconfigured_basic_lti_launch(self):
...
         will then be DB-configured launches rather than unconfigured.
         """
+        oauth_consumer_key = self.request.lti_user.oauth_consumer_key
+
+        lms_url = self.request.params.get("custom_canvas_api_domain")
+
+        if not lms_url:
+            lms_url = self.request.find_service(name="ai_getter").lms_url(
+                oauth_consumer_key
+            )
```
seanh: This is the same as what the legacy views do -- they use the `custom_canvas_api_domain` launch param or, if that param is missing (e.g. if we're in a non-Canvas LMS) they fall back on the application instance's `lms_url` which is an almost completely unvalidated thing that the user typed in to the `/welcome` form. (The lack of validation of that form field is a problem in all sorts of ways and needs to be fixed later.)

seanh: https://github.com/hypothesis/lms/issues/352

lms/views/basic_lti_launch.py
```diff
@@ -135,10 +135,24 @@ def unconfigured_basic_lti_launch(self):
...
+
         return {
             "content_item_return_url": self.request.route_url(
                 "module_item_configurations"
             ),
+            "google_client_id": self.request.registry.settings["google_client_id"],
+            "google_developer_key": self.request.registry.settings[
+                "google_developer_key"
+            ],
```
seanh: As you can see it's a lot simpler if the backend code, templates and frontend code can all just assume that these settings exist. Less tests to write too

PR #685, opened 2019-06-11
Move lms/config/resources.py into the right place

seanh:
For some reason the app's Pyramid resources were put in a `resources.py` module in `lms/config/` even though they have nothing to do with config.

Extract these into a `lms/resources/` package.

Also renamed `Root` to `DefaultResource` and `LTILaunch` to `LTILaunchResource`. This makes it more obvious what these classes are and makes them easier to grep for.

Also add tests for `DefaultResource`.

### DIFF
--- a/lms/resources/_default.py
+++ b/lms/resources/_default.py
@@ -1,0 +1,15 @@
+"""Default traversal resources."""
+from pyramid.security import Allow
+
+
+__all__ = ["DefaultResource"]
+
+
+class DefaultResource:
+    """The application'a default root resource."""
+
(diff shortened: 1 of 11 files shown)


lms/resources/_default.py
```diff
@@ -0,0 +1,15 @@
+"""Default traversal resources."""
+from pyramid.security import Allow
+
+
+__all__ = ["DefaultResource"]
+
+
+class DefaultResource:
+    """The application'a default root resource."""
```
seanh:
```suggestion
    """The application's default root resource."""
```

PR #686, opened 2019-06-12
Reduce the number of parallel build tasks

robertknight:
Each build task spins up its own Docker container and, for the backend tasks, each container also installs build dependencies (build-base, python3-dev etc.). It appears that this has significant overhead and on our current Jenkins setup, builds are prone to failing as a result. Specifically, we're seeing quite a lot of timeouts when Jenkins is invoking `docker run` to run backend tests and when trying to kill containers after a successful test run. See my notes and linked issues [in this Slack message](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1560333606006300).

Reduce the contention by combining the backend
lint/docstrings/formatting tasks into one, and the frontend lint + tests
into one.

~~I also added config to suppress some lint failures seen in a fresh build of `make backend-lint` with no pre-existing tox env for "py36-lint".~~ (_seanh addressed this in a separate PR_)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,26 +27,12 @@ node {
 
     // Run each of the stages in parallel.
     parallel failFast: true,
-    "Formatting": {
-        stage("Formatting") {
-            testApp(image: img, runArgs: runArgs) {
-                installDeps()
-                run("make checkformatting")
-            }
-        }
-    },
-    "Docstrings": {
-        stage("Docstrings") {
-            testApp(image: img, runArgs: runArgs) {
-                installDeps()
-                run("make checkdocstrings")
-            }
-        }
-    },
     "Backend lint": {
         stage("Backend lint") {
             testApp(image: img, runArgs: runArgs) {
                 installDeps()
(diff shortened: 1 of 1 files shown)


robertknight:
Build times in a quiet period with no contention:

Screenshot 2019-06-12 13 27 46

In contrast with the corresponding steps of recent master builds, the build actually completed quicker overall due to less time spent in the backend-tests tasks:

Screenshot 2019-06-12 13 31 27

robertknight:
Closing this for now as I want to focus on other things for a bit. I will re-open if/when this becomes a problem for me again.

robertknight:
Well, that didn't last long, I've just experienced another set of build failures due to flakiness with the CI system. I think there may also be a general issue with Docker if the daemon has been continuously running for a long-enough period of time.

robertknight:
I'm going to go ahead and merge this because flaky CI builds are blocking me from merging other LMS PRs, and the most useful thing I can think to do is merge this and see if it improves the situation.

PR #688, opened 2019-06-12
New content item selection view

seanh:
~~Depends on https://github.com/hypothesis/lms/pull/687~~

This replaces the content item selection view (used when creating assignments, in Canvas only) with a new-style view when the new_oauth feature flag is on. Similar to how the assignment launch views have been replaced. To test it you just need to create assignments in Canvas, with the feature flag on and off, and also make sure that the created assignments can be launched.


lms/views/content_item_selection.py
```diff
@@ -1,90 +1,98 @@
...
+Canvas LMS's Content Item docs are also useful:
+
+https://canvas.instructure.com/doc/api/file.content_item.html
+
+.. note::
+
+   The views in this module are currently used only when the new_oauth feature
+   flag is on. They replace legacy views that're still used when the feature
+   flag is off.
+"""
```
robertknight: Great docs :)

lms/views/content_item_selection.py
```diff
@@ -1,90 +1,98 @@
...
-        ]
-
-    custom_lms_url = None
-    if "custom_canvas_api_domain" in lti_params:
-        custom_lms_url = lti_params["custom_canvas_api_domain"]
-
-    params = {
-        "content_item_return_url": content_item_return_url,
+def content_item_selection(request):
+    template_variables = {
```
robertknight: This name strikes me as slightly verbose, albeit clear. Is there a convention for naming the variable containing the template context in other views or h that can be used?

lms/views/content_item_selection.py
```diff
@@ -1,90 +1,98 @@
...
 
-    if canvas_files_available(request, params=lti_params):
-        params["course_id"] = lti_params["custom_canvas_course_id"]
-    return params
+    # Pass the URL of the LMS that is launching us to our JavaScript code.
+    # When we're being launched in an iframe within the LMS our JavaScript
+    # needs to pass this URL (which is the URL of the top-most page) to Google
+    # Picker, otherwise Picker refuses to launch inside an iframe.
+    if "custom_canvas_api_domain" in request.params:
+        template_variables["lms_url"] = request.params["custom_canvas_api_domain"]
```
robertknight: We should rename this to something more specific soon, in case someone else ends up trying to use it for a different purpose, eg. `top_frame_origin` or something.

I also wonder whether we might be able to get this value from somewhere else, such as the HTTP_REFERER header, rather than rely on the user to specify it correctly.

lms/views/legacy_content_item_selection.py
```diff
@@ -0,0 +1,90 @@
...
+        "oauth_version": lti_params["oauth_version"],
+        "oauth_nonce": lti_params["oauth_nonce"],
+        "oauth_consumer_key": lti_params["oauth_consumer_key"],
+        "oauth_signature_method": lti_params["oauth_signature_method"],
+        "oauth_signature": lti_params["oauth_signature"],
+        "jwt_token": jwt,
+    }
+    # These fields appear in blackboard launches, but not in canvas
+    # launches
+    if "resource_link_id" in lti_params:
```
robertknight: What is going to happen with the Blackboard-specific logic here in future?

tests/lms/views/content_item_selection_test.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+
+from lms.views.content_item_selection import content_item_selection
+
+
+class TestContentItemSelection:
+    def test_it_passes_the_content_item_return_url_to_the_template(
+        self, pyramid_request
+    ):
+        template_variables = content_item_selection(pyramid_request)
```
robertknight: Again, this name `template_variables` feels a bit verbose to me for something so common. Is there an existing convention we could use? In `TestCanvasFileBasicLTILaunch` and a few others the name `data` is used. In a few places in h the name `result` is. used.

seanh:
Rebased

lms/views/content_item_selection.py
seanh: Not that I'm aware of. `template_data` is the best alternative I can think of. In the Pyramid docs I think this is variously referred to as "template variables" or "template data". It's hard to find examples because in most cases it's just `return { ... }` and docs will say things like "returns a dict". The legacy lms code called it `params` which I thought was unhelpful

lms/views/content_item_selection.py
seanh: Yeah. This whole variable is dodgy to be honest. Its value is either the unvalidated LMS domain entered by a user into https://lms.hypothes.is/welcome, or its the value of the `custom_canvas_api_domain` launch param which is the Canvas API's domain which is not quite the same thing as the domain of the top-level frame. In practice it seems to mostly work. If there's a better way to get the right value that would be good

lms/views/legacy_content_item_selection.py
seanh: I think that comment is misleading. These params aren't just for Blackboard, they're for all LMS's that don't use content item selection, which is all but Canvas. I think Blackboard was just the first non-content item selection LMS that support was added for and so various code comments refer to "Blackboard".

I think what has happened here is that this legacy view code was used for both LMS's that support content item selection (Canvas) and for doing unconfigured assignment launches in LMS's that don't support it. With the feature-flagged code we've already moved those unconfigured assignment launches, which includes Blackboard, to the new `unconfigured_basic_lti_launch()` view. So if Blackboard is gonna break, we've already broken it!

I don't think it is broken though because the new `unconfigured_basic_lti_launch()` does retain support for these "Blackboard-specific" params.

Our Blackboard test site was down. It's back up now, but it doesn't work with dev environments, so I'll have to deploy to QA to test it

tests/lms/views/content_item_selection_test.py
seanh: I think I prefer something more verbose like `template_variables` or `template_data`. The verboseness is much less of a crime in my mind than variable names like `data`, `result`, etc that tell you nothing about the variable, could equally apply to any variable in the code. You might as well just call it `variable` :)

Perhaps it could be shortened to `template_vars` (although that's the same length as `template_data`) or `tmpl_vars` or something. I've never been much of a fan of those sorts of shortenings though as they just introduce confusion about what "tmpl" is short for.

There might be a better solution but it's not coming to me right now

seanh:
Rebased

PR #692, opened 2019-06-13
Fix `make lint` by pinning Prospector

seanh:
Pin Prospector to the latest version that works.

Pinning it in `tox.ini` like this won't enable it to be automatically updated by Dependabot. For the work needed to enable Dependabot to handle this see https://github.com/hypothesis/lms/pull/691


seanh:
Gonna self-merge this as it's blocking everything

PR #693, opened 2019-06-13
Move RPC server and client configs into base template

seanh:
We're going to be needing the `js-rpc-server-config`,
`js-hypothesis-config` in pages that don't use the
`basic_lti_launch.html.jinja2` template or a sub-template thereof. But
currently these config objects are rendered by
`basic_lti_launch.html.jinja2` so are only available in pages based on
that template.

Move the `<script class="js-rpc-server-config">` and
`<script class="js-hypothesis-config">` objects out of the
`basic_lti_launch.html.jinja2` template and into the `base.html.jinja2`
template. Also move the `<script>` tags for the postMessage JSON RPC
server JavaScript with them.

This puts these JSON objects in the same place in the templates as the
existing `<script class="js-config">` object.

Rather than the inclusion of these objects in the page being conditional
on whether the view's template was `basic_lti_launch.html.jinja2` or a
sub-template thereof, their inclusion is now conditional on whether
`context.rpc_server_config` and `context.hypothesis_config` are defined,
using `{% if %}` statements in `base.html.jinja2`.

### DIFF
--- a/lms/templates/base.html.jinja2
+++ b/lms/templates/base.html.jinja2
@@ -32,6 +32,17 @@
       {% block content %}{% endblock %}
     </div>
 
-    {% block scripts %}{% endblock %}
+    {% block scripts %}
+      {% if context.rpc_server_config %}
+        <script type="application/json" class="js-rpc-server-config">{{ context.rpc_server_config|tojson }}</script>
+        {% for url in asset_urls("postmessage_json_rpc_server_js") %}
+          <script src="{{ url }}"></script>
+        {% endfor %}
+      {% endif %}
+
+      {% if context.hypothesis_config %}
+        <script type="application/json" class="js-hypothesis-config">{{ context.hypothesis_config|tojson }}</script>
+      {% endif %}
+    {% endblock %}
   </body>
 </html>
--- a/lms/templates/basic_lti_launch/basic_lti_launch.html.jinja2
+++ b/lms/templates/basic_lti_launch/basic_lti_launch.html.jinja2
@@ -3,13 +3,3 @@
 {% block content %}
   <iframe width="100%" height="100%" src="{{ via_url }}"></iframe>
 {% endblock %}
-
-{% block scripts %}
-  <script type="application/json" class="js-rpc-server-config">{{ context.rpc_server_config|tojson }}</script>
-  <script type="application/json" class="js-hypothesis-config">{{ context.hypothesis_config|tojson }}</script>
-
-  {% for url in asset_urls("postmessage_json_rpc_server_js") %}
-    <script src="{{ url }}"></script>
-  {% endfor %}
-{% endblock %}
-


lms/templates/base.html.jinja2
```diff
@@ -32,7 +32,19 @@
...
+
+      {% if context.rpc_server_config %}
+        <script type="application/json" class="js-rpc-server-config">{{ context.rpc_server_config|tojson }}</script>
+        {% for url in asset_urls("postmessage_json_rpc_server_js") %}
+          <script src="{{ url }}"></script>
+        {% endfor %}
+      {% endif %}
+
+      {% if context.hypothesis_config %}
+        <script type="application/json" class="js-hypothesis-config">{{ context.hypothesis_config|tojson }}</script>
```
robertknight: Hmm. The new file picker frontend uses the same `js-hypothesis-config` class for its config block which is confusing. Probably better to:

1) Use a more explicit name for the Hypothesis client config (`js-hypothesis-client-config`?)
2) Change the LMS frontend to use a different name for its config blog (`js-config`)?

lms/templates/base.html.jinja2
```diff
@@ -32,7 +32,19 @@
       {% block content %}{% endblock %}
     </div>
 
-    <script type="application/json" class="js-config">{{ js_config|tojson }}</script>
-    {% block scripts %}{% endblock %}
+    {% block scripts %}
+      <script type="application/json" class="js-config">{{ js_config|tojson }}</script>
```
robertknight: I couldn't find any other references to `js-config` in the codebase so I think this is unused. The LMS frontend JS is using `js-hypothesis-config`, which is confusing and conflicts with the config used for the client config.

lms/templates/base.html.jinja2
seanh: Yep, the plan is to change the new file picker to read js-config, yes. Note that js-hypothesis-config is _already_ used by the backend and the JavaScript RPC server to contain the Hypothesis client config, so the file picker templates that're rendering a second `js-hypothesis-config` are in conflict with that. That presumably would break the RPC server -> client integration but as it happens there are no pages that contain both the file picker and the client together, so nothing was broken.

Regarding the naming of the config objects, I was thinking in future to merge them into a single js-config object like this:

```
<script type="application/json" class="js-hypothesis-config">
    {
        // Top-level JS settings (e.g. for the file picker) go here.
        ...
        "rpc_server": { ... },
        "hypothesis_client": { ... },
</script>
```

The file-picker's settings could potentially go under their own `file_picker` key as well.

lms/templates/base.html.jinja2
seanh: Yeap ... That's going to be fixed but is unrelated to what's in this PR.

seanh:
Rebased

PR #694, opened 2019-06-13
Add LTILaunchResource.js_config

seanh:
During the user provisioning project we added two JavaScript config objects -- `<script class="js-rpc-server-config">` to hold the config settings for our postMessage-JSON-RPC server, and `<script class="js-hypothesis-config">` to hold the config settings that the RPC server would pass over postMessage to the Hypothesis client (for example the grant token).

The code that generates these config objects is in the Pyramid context object `LTILaunchResource`, specifically the `LTILaunchResource.rpc_server_config` and `LTILaunchResource.hypothesis_config` properties. They get rendered into the templates like this: `{{ context.rpc_server_config|tojson }}`.

Later during the Canvas Files / OAuth rewrite project we added a _third_ JavaScript config object to contain the config settings for _the LMS app's own JavaScript code_ (other than the RPC server): `<script class="js-config">`. Unfortunately when I added this I forgot about `LTILaunchResource.rpc_server_config` and `LTILaunchResource.hypothesis_config` and instead implemented `js_config` as a global, top-level variable added to the template environment by a Pyramid subscriber in `lms/subscribers/`.

This PR corrects that by moving `js_config` to `LTILaunchResource.js_config` and changing the templates to read it from there.

Note that apart from `content_item_selection.html.jinja2` and `new_content_item_selection.html.jinja2` accessing `context.js_config.authorization_param` directly, the new `context.js_config` isn't otherwise used yet. Future PRs will change `base.html.jinja2` to render the entire `context.js_config` the same as it renders `context.rpc_server_config` and `context._hypothesis_config`, rather than individual templates rendering their own `class="js-config"` object.

It's probably a good idea to merge these three config objects into a single one with different sections, but that's beyond the scope of this PR.


seanh:
Rebased

lms/resources/_lti_launch.py
```diff
@@ -154,6 +157,29 @@ def h_userid(self):
...
+        This is a mutable config dict. It can be accessed, for example by
+        views, as ``request.context.js_config``, and they can mutate it or add
+        their own view-specific config settings. The modified config object
+        will then be passed to the JavaScript code in the response page.
+
+        :rtype: dict
+        """
+        if self._js_config is None:
+            # Initialize self._js_config for the first time.
+            self._js_config = {"urls": {}}
```
robertknight: What is "urls" going to be used for?

seanh:
Rebased

lms/resources/_lti_launch.py
seanh: This was intended to hold all the URLs that the JS needs to do its job. Instead of passing parameters like the course_id, file_id and authentication param, the server would put the file list, via url, and popup window URLs with the params already in them. So JS doesn't need to construct URLs. So far though we've forgotten to use it! But I intend to start doing so

Given that our proxy API already has to accept the JWT auth param as a query param (because of the popup window) we _could_ also put it as a query param rather than a header in the XHR API requests as well, meaning the server could render the URLs with the param already on them, then the JS wouldn't have to worry about sending auth headers either

PR #696, opened 2019-06-14
Rename the JSON config block that the LMS frontend app reads from

robertknight:
This PR fixes a couple of things that were confusing about the JSON config that the LMS frontend / file picker app reads.

- The base template defined a "js-config" block, but this was not actually used anywhere
- The file picker app read its config from a "js-hypothesis-config" block, which was defined in each of the views that rendered the app. The "js-hypothesis-config" name is overloaded as elsewhere in the app it is used for configuration for the _Hypothesis client_

### DIFF
--- a/lms/templates/content_item_selection.html.jinja2
+++ b/lms/templates/content_item_selection.html.jinja2
@@ -15,7 +15,7 @@
     <script async defer src="{{ url }}"></script>
   {% endfor %}
 
-  <script class="js-hypothesis-config" type="text/json">
+  <script class="js-config" type="text/json">
   {
     "authToken": "{{ js_config.authorization_param }}",
     "authUrl": "{{ request.route_url('canvas_api.authorize') }}",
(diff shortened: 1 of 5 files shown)


lms/templates/content_item_selection.html.jinja2
```diff
@@ -15,7 +15,7 @@
     <script async defer src="{{ url }}"></script>
   {% endfor %}
 
-  <script class="js-hypothesis-config" type="text/json">
+  <script class="js-config" type="text/json">
   {
     "authToken": "{{ js_config.authorization_param }}",
```
seanh: It's a little confusing that this now `class="js-config"` object actually depends on one of the settings from the `js_config` template variable (but ignores the rest). The `js_config` variable's purpose is to be rendered by the base template as a  `class="js-config"` object, which _this_ `class="js-config"` would conflict with :) But I can sort that out in following PRs

robertknight: > But I can sort that out in following PRs

Yes, that was my intention. I forgot to call that out in the description.

PR #695, opened 2019-06-13
Rename authorization param

seanh:
Rename `authorization_param` in the `js-config` object to `authToken`. The new file picker JavaScript already reads `authToken` not `authorization_param`, and the naming style matches that of other JavaScript config settings.

Depends on https://github.com/hypothesis/lms/pull/694

PR #697, opened 2019-06-14
Read the js_config object from the Pyramid context

seanh:
Depends on https://github.com/hypothesis/lms/pull/694, https://github.com/hypothesis/lms/pull/693 and https://github.com/hypothesis/lms/pull/695.

Read the JavaScript config object `<script type="application/json" class="js-config">` from the Pyramid context object (`LTILaunchResource.js_config`) instead of rendering it directly in templates.
    
This makes the generation and rendering of js-config consistent with that of js-rpc-server-config and js-hypothesis-config, and means that the js-config object is generated in unit-tested Python code rather than in untested template code. js-config can be modified on a per-view basis by having the Python view mutate `request.context.js_config`.
    
1. Have `base.html.jinja2` render `context.js_config` into the page, as it already does with `context.rpc_server_config` and `context.hypothesis_config`
    
2. `unconfigured_basic_lti_launch.html.jinja2` and `content_item_selection.html.jinja2` no longer need to render their own `js-config` objects directly
    
3. The `unconfigured_basic_lti_launch()` and `content_item_selection()` views now mutate `request.context.js_config`, adding the view-specific JavaScript file picker settings that those views need (these view-specific settings were previously rendered by the view's templates, `unconfigured_basic_lti_launch.html.jinja2` and `content_item_selection.html.jinja2`)

### DIFF
--- a/lms/templates/base.html.jinja2
+++ b/lms/templates/base.html.jinja2
@@ -33,6 +33,10 @@
     </div>
 
     {% block scripts %}
+      {% if context.js_config %}
+        <script type="application/json" class="js-config">{{ context.js_config|tojson }}</script>
+      {% endif %}
+
       {% if context.rpc_server_config %}
         <script type="application/json" class="js-rpc-server-config">{{ context.rpc_server_config|tojson }}</script>
         {% for url in asset_urls("postmessage_json_rpc_server_js") %}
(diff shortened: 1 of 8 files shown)


seanh:
The `unconfigured_basic_lti_launch()` and `content_item_selection()` views are pretty similar in that both just mutate `context.js_config` to add the settings needed by the file picker. But most of the file picker settings actually differ between these views, there is very little actual duplication here, and I think the simplicity of keeping them as two separate views is worth more. (See the less from AJ's previous implementation of these as one view: much more complicated code, inadvertently submitting all the context item selection-unique params to the module item configuration view and vice-versa, posting our auth JWTs to LMS's, etc).

Having said that future PRs _will_ deduplicate some of the code in these two views by moving it into the context and helpers, and will also merge their two templates into one.

seanh:
Rebased

Issue #622, opened 2019-05-17
Use the new popup-based Canvas API authorization when launching Canvas Files assignments


seanh:
The new Canvas Files and OAuth 2.0 frontend and backend needs to be used when launching assignments, as well as when creating them.

1. User launches a Canvas Files assignment

2. Server renders JavaScript code, authentication JWT and any other necessary config into HTML page

3. Server does not render any Via iframe in the initial page load. Without JavaScript the user would see only a blank white page.

   If the assignment is an HTML or PDF public URL or a Google Drive assignment then the server *will* prerender the Via iframe into the initial response, so there's nothing for the JavaScript code to do. But for Canvas Files assignments the server will never render the iframe.

4. JavaScript code runs on page load and notices that there's no Via iframe, so it sends a JWT-authenticated XHR request to the `/via_url` proxy API.

5. If the API responds with an error:

   1. The same steps below apply whether the error was because we don't have an access token for the user or because of some other reason, the frontend should behave the same in all cases.

   2. The frontend code shows an authorization dialog:

      ```
      You need to authorize Hypothesis to access this assignment:

                                       <button>Authorize</button>
      ```

      For bonus points if the `/via_url` request failure was for some reason other than that we simply don't have an access token for this user, then we could add some information about the failure into the dialog:

       ```
       Authorization failed: <reason>. Try again:

                       <button>Authorize</button>
       ```

   3. When the user clicks the <kbd>Authorize</kbd> button the popup window opens and goes to the Canvas authorize page.

   4. When the popup window closes for any reason the frontend code tries the Via URL proxy API request again.

   5. If the request succeeds this time continue to step 6 below. If the request failed to back to ii above.

6. If the API responds successfully with the Via URL then use it to inject Via iframe into DOM

seanh:
@robertknight FYI I've tweaked the steps above from what I first described. I think the server should make the decision about whether the Via `<iframe>` can be rendered in the initial page load, or whether we instead need to render a message asking the user to click <kbd>Authorize</kbd> and open the popup window. JavaScript code is only involved in this second case, and the JavaScript code doesn't need to decide whether it should get involved or not -- the server will either include the authorization message and JavaScript code in the page or not, as appropriate.

This actually looks pretty simple now

seanh:
I'll prioritise the backend parts of this first next week to unblock the frontend

seanh:
Added a checklist above

seanh:
[...]

PR #706, opened 2019-06-18
Don't fetch Canvas file URLs on launch

seanh:
Fix `canvas_file_basic_lti_launch()` to not fetch the Canvas file's
public URL from the Canvas API as part of the initial response to the
launch request. Instead add the proxy API URL that will return the
Canvas file's URL to the `"urls"` in the JavaScript config object.

The idea is that JavaScript code will make an XHR request to the
`via_url` proxy API and use the Canvas file URL from the response to
inject the Via iframe into the DOM.

This means that the server's response to the initial launch request can
be fast as it doesn't need to talk the the Canvas API, and all requests
to the Canvas API are kept within the `/api/canvas/` proxy API views.

For more reasoning see
https://github.com/hypothesis/lms/issues/622#issuecomment-499909135

### DIFF
--- a/lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
+++ b/lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
@@ -1,17 +1,1 @@
-{% extends "lms:templates/basic_lti_launch/basic_lti_launch.html.jinja2" %}
-
-{% block content %}
-  {% if via_url %}
-    {{ super() }}
-  {% else %}
-    <main class="modal-content">
-      <h1>Authorize Hypothesis</h1>
-
-      <p class="modal-text">
-        Hypothesis needs your authorization to access this assignment.
-      </p>
-
-      <button class="btn" type="submit">Authorize</button>
-  </main>
-  {% endif %}
-{% endblock %}
+{% extends "lms:templates/base.html.jinja2" %}
--- a/lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
+++ b/lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
@@ -1,1 +1,0 @@
-{% extends "lms:templates/base.html.jinja2" %}
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -49,10 +49,7 @@ class BasicLTILaunchViews:
         self.context = context
         self.request = request
 
-    @view_config(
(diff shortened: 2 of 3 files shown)


lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
```diff
@@ -1,17 +1 @@
...
-      <h1>Authorize Hypothesis</h1>
-
-      <p class="modal-text">
-        Hypothesis needs your authorization to access this assignment.
-      </p>
-
-      <button class="btn" type="submit">Authorize</button>
-  </main>
-  {% endif %}
-{% endblock %}
```
seanh: If the frontend receives an error from the `via_url` API then it will have to show a dialog with an <kbd>Authorize</btn> like this one. (After the user closes the authorization window the frontend will re-try the `via_url` request and if the request fails again the frontend will return to this dialog with the <kbd>Authorize</btn> button again.)

lms/templates/basic_lti_launch/canvas_file_basic_lti_launch.html.jinja2
```diff
@@ -1,17 +1 @@
...
+{% extends "lms:templates/base.html.jinja2" %}
```
seanh: Given that this template now just extends the base template and adds nothing, it can probably be deleted...

lms/views/basic_lti_launch.py
```diff
@@ -65,15 +64,17 @@ def canvas_file_basic_lti_launch(self):
...
         file_id = self.request.params["file_id"]
 
-        try:
-            document_url = canvas_api_client.public_url(file_id)
-        except CanvasAPIError:
-            return {}
+        self.context.js_config["urls"].update(
+            {
+                "via_url": self.request.route_url(
+                    "canvas_api.files.via_url", file_id=file_id
```
seanh: So rather than reading the `file_id` from `js-config` and constructing the URL to the `via_url` API, the frontend just reads the entire URL from `js-config`. We should fix it to do this for the `list_files` API and for the popup window's URL (complete with `authorization` query param) as well.

The backend _could_ even add the `authorization` query param to the `via_url` here, and then the frontend would not need to read `authToken` from the config and add it as a header.

robertknight: From the front-end's point of view, the most important thing is that API calls behave consistently. For that reason, either the auth should always be baked into the URL or never. I think my preference is to stick with the Authorization header for general consistency across projects making authenticated calls to some kind of API.

PR #710, opened 2019-06-19
Don't subclass HTTPError in services

seanh:
Change `lms.services.HAPIError`, `HAPINotFoundError` and
`CanvasAPIError` into simple `Exception` subclasses rather than
subclassing `HTTPInternalServerError`.

The reason for subclassing `HTTPInternalServerError` was that views that
call services that raise these errors could just call them blindly, not
catching the services exceptions that might be raised, and since the
services exceptions were `HTTPError`s they would cause the correct 500
Internal Server Error responses to be sent.

While this does simplify view code and tests by removing
exception-wrapping from views, it causes some problems:

We want services exceptions to all be subclasses of
`lms.services.ServiceError`. This is so that some code calling a
services method can just catch `ServiceError` and this will catch any
error deliberately raised by services code (but won't catch unexpected
exceptions, those will be uncaught and cause a crash, which is what we
want for an unexpected exception). Since specific subclasses are used an
`except` that only wants to handle a specific kind of service error can
just catch a specific `ServiceError` subclass or classes (and can
optionally be followed by a catch-all `except ServiceError`).

If service classes are going to be `ServiceError` subclasses and also
`HTTPInternalServerError` subclasses, then we get into multiple
inheritance and cooperating classes, some of them third-party classes,
and things become tricky and tightly coupled to Pyramid.

Also, being simple `Exception` classes makes services errors simpler and
makes it easier to have those errors extend and reuse each other, by not
bringing in the complications of Pyramid's `HTTPError`s and their fields
and behaviours.

While having the views catch `ServiceError`s and re-raise them as
`HTTPError`s is more view code, it also makes it explicit in the views
that they may raise a 500, rather than this being an implicit
consequence of calling services code.

Finally, HTTP considerations and Pyramid belong in the views layer, not
the services layer.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -1,6 +1,3 @@
-from pyramid.httpexceptions import HTTPInternalServerError
-
-
 class ServiceError(Exception):
     """Base class for all :mod:`lms.services` exceptions."""
 
@@ -27,32 +24,25 @@ class LTIOAuthError(LTILaunchVerificationError):
     """Raised when OAuth signature verification of a launch request fails."""
 
 
-class HAPIError(HTTPInternalServerError):  # pylint: disable=too-many-ancestors
+class HAPIError(ServiceError):
     """
     A problem with an h API request.
 
     This exception class is raised whenever an h API request times out or when
     an unsuccessful, invalid or unexpected response is received from the h API.
 
-    An HAPIError is a 500 Internal Server Error response (HTTPInternalServerError
-    subclass) rather than, say, a 502 Bad Gateway because Cloudflare intercepts
-    gateway error responses and replaces our error page with its own error
-    page, and we don't want that.
-
-    Any arguments or keyword arguments other than ``response`` are passed
-    through to HTTPInternalServerError.
-
     :param response: The response from the HTTP request to the h API
     :type response: requests.Response
     """
 
-    def __init__(self, *args, response=None, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, explanation=None, response=None):
+        super().__init__()
+        self.explanation = explanation
         self.response = response
 
     def __str__(self):
         if self.response is None:
-            return super().__str__()
+            return self.explanation
 
         # Log the details of the h API response. This goes to both Sentry and
         # the application's logs. It's helpful for debugging to know how h
@@ -65,11 +55,11 @@ class HAPIError(HTTPInternalServerError):  # pylint: disable=too-many-ancestors
         return " ".join([part for part in parts if part])
 
 
-class HAPINotFoundError(HAPIError):  # pylint: disable=too-many-ancestors
+class HAPINotFoundError(HAPIError):
     """A 404 error from an API request."""
 
 
-class CanvasAPIError(HTTPInternalServerError):  # pylint: disable=too-many-ancestors
+class CanvasAPIError(ServiceError):
     """
     A problem with a Canvas API request.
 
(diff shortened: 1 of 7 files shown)


lms/views/decorators/h_api.py
```diff
@@ -40,12 +40,16 @@ def wrapper(context, request):
...
+            try:
+                hapi_svc.patch(
+                    f"users/{context.h_username}",
+                    {"display_name": context.h_display_name},
+                )
+            except HAPINotFoundError:
+                # Call the h API to create the user in h if it doesn't exist already.
+                hapi_svc.post("users", user_data)
+        except HAPIError as err:
+            raise HTTPInternalServerError(explanation=err.explanation) from err
```
seanh: The nested `try`'s here are unfortunate, but the solution is to move the "Try PATCH and if it 404s try POST" (a.k.a upsert) logic into the service rather than having it here in the view, and that's outside of the scope of this PR

lms/views/decorators/legacy_h_api.py
```diff
@@ -62,16 +62,20 @@ def wrapper(request, jwt, context=None):
...
+            try:
+                hapi_svc.patch(
+                    f"users/{context.h_username}",
+                    {"display_name": context.h_display_name},
+                )
+            except HAPINotFoundError:
+                # Call the h API to create the user in h if it doesn't exist already.
+                hapi_svc.post("users", user_data)
+        except HAPIError as err:
+            raise HTTPInternalServerError(explanation=err.explanation) from err
```
seanh: I'm having to make the same changes twice because we have `h_api` view decorators and `legacy_h_api` ones. The legacy ones will be deleted soon

lms/views/api/canvas/files.py
```diff
@@ -43,13 +45,22 @@ def __init__(self, request):
     @view_config(request_method="GET", route_name="canvas_api.courses.files.list")
     def list_files(self):
         """Return the list of files in the given course."""
-        return self.canvas_api_client.list_files(self.request.matchdict["course_id"])
+        try:
+            return self.canvas_api_client.list_files(
+                self.request.matchdict["course_id"]
+            )
+        except CanvasAPIError as err:
+            raise HTTPInternalServerError(explanation=err.explanation) from err
```
robertknight: I had a query about whether this would produce a different result than not catching the exception at all. @seanh explained [on Slack](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1561036091010000) that the difference is that the `explanation` will be presented to the user, whereas exception info for an uncaught exception is translated into a generic message for the user, in case it contains information that shouldn't be exposed.

lms/services/exceptions.py
```diff
@@ -27,32 +24,25 @@ class LTIOAuthError(LTILaunchVerificationError):
...
+        super().__init__()
+        self.explanation = explanation
         self.response = response
 
     def __str__(self):
         if self.response is None:
-            return super().__str__()
+            return self.explanation
 
         # Log the details of the h API response. This goes to both Sentry and
```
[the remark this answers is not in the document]
seanh: The exceptions get stringified and sent to both Sentry and Papertrail, so adding info into the string representation of an exception is what adds that info to the logs. The actual sending to Sentry and Papertrail is done by those services integrations, not by our own code (for example Sentry automatically reports any non-HTTP exceptions that get raised).

If the exception were caught by the view and either swallowed completely, or wrapped in something other than an `HTTPInternalServerError` (for example wrapped in an `HTTPBadRequest`) that would prevent it from being sent to Sentry, but it would still get logged to Papertrail.

PR #711, opened 2019-06-20
Deduplicate exception classes

seanh:
Depends on https://github.com/hypothesis/lms/pull/710.

Extract a base class `ExternalRequestError` to remove duplicate code
from `HAPIError` and `CanvasAPIError`.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -1,6 +1,3 @@
-from pyramid.httpexceptions import HTTPInternalServerError
-
-
 class ServiceError(Exception):
     """Base class for all :mod:`lms.services` exceptions."""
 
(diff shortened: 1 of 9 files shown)


lms/services/exceptions.py
```diff
@@ -27,36 +24,26 @@ class LTIOAuthError(LTILaunchVerificationError):
...
 
     def __str__(self):
         if self.response is None:
-            return super().__str__()
+            return self.explanation
 
-        # Log the details of the h API response. This goes to both Sentry and
-        # the application's logs. It's helpful for debugging to know how h
-        # responded.
+        # Log the details of the response. This goes to both Sentry and the
```
robertknight: This code doesn't log the exception details directly. Does it mean that this method is called when an uncaught exception is logged, in order to get the representation to log?

PR #713, opened 2019-06-20
Oauth2 redirect error handling

seanh:
Depends on <https://github.com/hypothesis/lms/pull/710> and <https://github.com/hypothesis/lms/pull/711>.

When we send the access token request to Canvas, catch any error
responses or network timeouts etc and show the user an error page with a
"Try again" button.

`CanvasAPIClient.get_token()` now raises `CanvasAPIServerError` if
there's a problem with the access token request.

The `oauth2_redirect()` view that calls `CanvasAPIClient.get_token()`
catches the `CanvasAPIServerError` and wraps it in an
`HTTPInternalServerError`.

This causes Pyramid to call the new `oauth2_redirect_error()` exception
view which renders a template with a "Try again" button that links back
to the original `authorize()` view to restart the flow.

The user is put into an infinite loop, within the popup window, of authorizing with Canvas and then seeing an error message from us with a <kbd>Try again</kbd> button, until the authorization succeeds or the user gives up.

Peek 2019-06-18 16-33

### DIFF
--- a/lms/services/canvas_api.py
+++ b/lms/services/canvas_api.py
@@ -6,6 +6,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from lms.models import OAuth2Token
 from lms.services import CanvasAPIError
+from lms.services.exceptions import CanvasAPIServerError
 from lms.services._helpers import CanvasAPIHelper
 
 
@@ -27,15 +28,20 @@ class CanvasAPIClient:
         Get an access token for the current LTI user.
 
         Posts to the Canvas API to get the access token and returns it.
+
+        :raise lms.services.CanvasAPIServerError: if the Canvas API request
+            fails for any reason
         """
         access_token_request = self._helper.access_token_request(authorization_code)
-        access_token_response = requests.Session().send(access_token_request)
 
-        # TODO: Handle access token error responses.
-        # These aren't documented in the Canvas docs as far as I can see.
-        # They are documented in the OAuth 2 spec but I don't yet know if Canvas's
-        # error responses follow this:
-        # https://tools.ietf.org/html/rfc6749#section-5.2
+        try:
+            access_token_response = requests.Session().send(access_token_request)
+            access_token_response.raise_for_status()
+        except RequestException as err:
+            raise CanvasAPIServerError(
+                explanation="Authorizing with Canvas failed",
+                response=getattr(err, "response", None),
+            ) from err
 
         # TODO: Validate access_token_response.
 
--- a/lms/services/canvas_api.py
+++ b/lms/services/canvas_api.py
@@ -29,6 +29,10 @@ class CanvasAPIClient:
 
         Posts to the Canvas API to get the access token and returns it.
 
+        :arg authorization_code: The Canvas API OAuth 2.0 authorization code to
+            exchange for an access token
+        :type authorization_code: str
+
         :raise lms.services.CanvasAPIServerError: if the Canvas API request
             fails for any reason
         """
--- a/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
+++ b/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
@@ -1,0 +1,33 @@
+{% extends "lms:templates/base.html.jinja2" %}
+
+{% block content %}
+  <div class="Dialog__background" style="z-index: 1;"></div>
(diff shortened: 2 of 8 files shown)


seanh:
The easiest way to test his is to hack [`get_token()`](https://github.com/hypothesis/lms/blob/dadf4d9098075bc2249583e6a38d9faadb0c494d/lms/services/canvas_api.py#L25-L47) or [`access_token_request()`](https://github.com/hypothesis/lms/blob/dadf4d9098075bc2249583e6a38d9faadb0c494d/lms/services/_helpers/canvas_api.py#L50-L76) to send invalid requests to Canvas, or to send requests to a URL that will timeout, etc, which should trigger the error page.

seanh:
Not shown in the GIF above but I added some support links to the error page. The code for these will move into a shared location in future when other views use them too. The email link doesn't pre-fill a bunch of info in the body like the client's email link does -- we'd have to decide what info would be useful for the lms app and then make it available to the template, and that info doesn't work if they click on the open a support ticket link instead anyway. Otherwise the text is very similar to that in the client and the links are the same.

Screenshot from 2019-06-20 12-07-43

seanh:
Rebased

lms/views/api/canvas/authorize.py
```diff
@@ -1,52 +1,75 @@
...
 
-    return {}
+        authorization_param = (
+            BearerTokenSchema(self.request).authorization_param(self.request.lti_user),
+        )
+        return {
+            "authorize_url": self.request.route_url(
+                "canvas_api.authorize", _query=[("authorization", authorization_param)]
+            )
+        }
```
seanh: Usually exception views go in `lms/views/error.py` or `lms/views/api/error.py` because they're usually generic, but in this case we have a specific error view for the popup window (an HTML page with an error message and <kbd>Cancel>/kbd> and <kbd>Try again </kbd> buttons with specific actions). It's literally an error view for just one view, and shares view defaults with that view. So I've put it in the class with the view that it belongs to

lms/services/canvas_api.py
```diff
@@ -27,15 +28,24 @@ def get_token(self, authorization_code):
...
-        # error responses follow this:
-        # https://tools.ietf.org/html/rfc6749#section-5.2
+        try:
+            access_token_response = requests.Session().send(access_token_request)
+            access_token_response.raise_for_status()
+        except RequestException as err:
+            raise CanvasAPIServerError(
+                explanation="Authorizing with Canvas failed",
+                response=getattr(err, "response", None),
+            ) from err
```
seanh: This catches all kinds of requests exceptions (network timeouts and things where we don't even get a response, as well as non-2xx responses, etc). It's the same technique we figured out for when the LMS app calls the h API

lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
```diff
@@ -0,0 +1,38 @@
+{% extends "lms:templates/base.html.jinja2" %}
+
+{% block content %}
+  <div class="Dialog__background" style="z-index: 1;"></div>
```
robertknight: Hmm ... there is a hazard here that it might not be obvious to someone modifying `Dialog.js` and `_Dialog.scss` that styles are referenced outside of those two files. Re-using classes between server and client-rendered HTML is something we'll want to do though.

For the moment, I think it would help simply to update the documentation [here](https://github.com/hypothesis/lms/blob/e740307635ad79d5ae725d6c0c7a18ef7dfb8ddc/lms/static/styles/file-picker-app.scss#L13) to indicate that these SASS components are used by the frontend and server-rendered HTML.

seanh: Done

seanh:
One weird thing that I noticed retrospectively about this PR is that if you open `http://localhost:8001/canvas_oauth_callback` directly in a browser tab you'll get the authorization failed try again page:

Screenshot from 2019-06-25 13-02-57

The <kbd>Cancel</kbd> button won't work, and the <kbd>Try again</kbd> button will navigate you to an authorization error page.

I'm not sure there's much we can, or need to, do about this. The endpoint simply isn't meant to be visited directly and there's nothing we could do to e.g. make it a 404 that wouldn't potentially break the error message UX in some real fault condition.

PR #716, opened 2019-06-21
Enable validating requests-library responses using our validation schemas

seanh:
Depends on <https://github.com/hypothesis/lms/pull/715>.

In various places we use the `requests` library to send requests to the Canvas API. `requests` will raise if we don't get a response (network error or timeout) or if the response is non-2xx. But after that we still need to validate that:

1. We can successfully parse the response body as JSON
2. The JSON contains all the required params that we need from the response
3. Each param has a valid value

Unfortunately `lms.validation` assumes that our validation schemas are all for validating Pyramid requests. So refactor `lms.validation` a little to also enable schemas that validate requests-library response objects:

1. Add `_BaseSchema` as the base class for _all_ validation schemas, just to remove some boilerplate that we were putting in every schema

2. Add `PyramidRequestSchema(_BaseSchema)` as the base class for all schemas for validating Pyramid request objects, and change all our existing schemas to inherit from it

3. Add `RequestsResponseSchema(_BaseSchema)` as the base class for future schemas that validate requests response objects, with a very similar API to that of `PyramidRequestSchema`

### DIFF
--- a/lms/validation/__init__.py
+++ b/lms/validation/__init__.py
@@ -1,19 +1,22 @@
 """
-Schemas for parsing and validating requests.
+Schemas for parsing and validating things.
 
-This package contains schemas that views can use to validate and parse
-requests. The idea is that the request is validated before the view is called.
-If validation fails an error response is sent back and the view is never
-called. If validation succeeds the parsed and validated parameters are made
-available to the view as ``request.parsed_params``. The view's own code can
-assume that all the parsed params are valid and that all required params are
-present.
+This package contains schemas for parsing and validating things like Pyramid
+requests, or :mod:`requests`-library responses.
 
-Usage::
+Validating Pyramid Requests
+---------------------------
 
-    from lms.validation import FOO_SCHEMA
+When validating Pyramid requests the idea is that the request is validated
+before the view is called.  If validation fails an error response is sent back
+and the view is never called. If validation succeeds the parsed and validated
+parameters are made available to the view as ``request.parsed_params``. The
+view's own code can assume that all the parsed params are valid and that all
+required params are present. Example::
 
-    @view_config(..., schema=FOO_SCHEMA)
+    from lms.validation import FooSchema
+
+    @view_config(..., schema=FooSchema)
     def foo_view(request):
         validated_arg_1 = request.parsed_params["validated_arg_1"]
         validated_arg_2 = request.parsed_params["validated_arg_2"]
@@ -23,6 +26,22 @@ Note that we're using our own view deriver (the ``schema`` argument to
 ``view_config``) to integrate our schemas and views, rather than using webargs's
 ``@use_args()`` / ``@use_kwargs()`` decorators. For the reasons for this see
 commit message f61a5ff3cae6b983e24db809d8e4b4933aca1e92.
+
+Validating requests-Library Responses
+-------------------------------------
+
+To validate a :mod:`requests`-library response you pass the response object to a
+suitable validation schema's ``__init__()`` method and then call the schema's
+``parse()`` method::
+
+    import requests
+
+    response = requests.get(...)
+
+    try:
+        parsed_params = BarSchema(response).parse()
+    except lms.validation.ValidationError as err:
+        ...
 """
 from lms.validation._exceptions import (
     ValidationError,
--- a/lms/validation/__init__.py
+++ b/lms/validation/__init__.py
@@ -24,8 +24,6 @@ Note that we're using our own view deriver (the ``schema`` argument to
 ``@use_args()`` / ``@use_kwargs()`` decorators. For the reasons for this see
 commit message f61a5ff3cae6b983e24db809d8e4b4933aca1e92.
 """
-from webargs import pyramidparser
-
 from lms.validation._exceptions import (
     ValidationError,
     ExpiredSessionTokenError,
@@ -42,7 +40,6 @@ from lms.validation._module_item_configuration import ConfigureModuleItemSchema
 
 
 __all__ = (
-    "parser",
     "CanvasOAuthCallbackSchema",
     "BearerTokenSchema",
     "LaunchParamsSchema",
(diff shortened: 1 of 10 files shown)


lms/validation/__init__.py
```diff
@@ -58,14 +54,6 @@ def foo_view(request):
 )
 
 
-parser = pyramidparser.PyramidParser()
-
-
-@parser.error_handler
-def _handle_error(error, _req, _schema, _status_code, _headers):
-    raise ValidationError(messages=error.messages) from error
```
seanh: All this `parser` and `_handle_error()` stuff has been moved into `PyramidRequestSchema` which is the only place where it's needed, since it's specific to validating Pyramid requests

lms/validation/__init__.py
```diff
@@ -24,8 +24,6 @@ def foo_view(request):
 ``@use_args()`` / ``@use_kwargs()`` decorators. For the reasons for this see
```
seanh: I need to update the package's docstring

lms/validation/_bearer_token.py
```diff
@@ -62,19 +62,9 @@ class BearerTokenSchema(marshmallow.Schema):
...
-    class Meta:
-        """Marshmallow options for this schema."""
-
-        # Silence a strict=False deprecation warning from marshmallow.
-        # TODO: Remove this once we've upgraded to marshmallow 3.
-        strict = True
-
     def __init__(self, request):
-        super().__init__()
-        self._request = request
```
seanh: We're always accessing the request `self.context["request"]` now, instead of a mix of that and `self._request`. (I'm not sure why `self.context["foo"]` is used instead of `self.foo`, it's a marshmallow pattern.)

lms/validation/_bearer_token.py
```diff
@@ -109,14 +107,10 @@ def lti_user(self):
...
-                self,
-                self.context["request"],
-                locations=["headers", "querystring", "form"],
-            )
-        except HTTPUnprocessableEntity as error:
+            return self.parse(locations=["headers", "querystring", "form"])
+        except ValidationError as error:
             try:
-                authorization_error_message = error.json["authorization"][0]
+                authorization_error_message = error.messages["authorization"][0]
```
seanh: This change is a little tricky to understand. There's a `_handle_error()` hook method below that causes validation to raise `lms.validation.ValidationError` instead of the webargs Pyramid integration's default behaviour of raising a Pyramid `HTTPUnprocessableEntity`. Previously `_bearer_token.py` and `_lti_launch.py` imported `webargs.pyramidparser.parser` themselves and used _their own_ copy of `parser` to do the parsing. Our `_handle_error` hook wasn't attached to that parser so it raised the default `HTTPUnprocessableEntity` and we had to catch and wrap that here. Now they call `self.parse()` instead so they do use our one, customised copy of the parser (the one from `PyramidRequestSchema`), so `parse()` raises `ValidationError` not `HTTPUnprocessableEntity`.

lms/validation/_helpers/_base.py
```diff
@@ -25,3 +31,20 @@ def __init__(self, request):
...
+        error.
+
+        :raise lms.validation.ValidationError: if the request isn't valid
+        """
+        return _PYRAMID_PARSER.parse(self, self.context["request"], *args, **kwargs)
+
+
+@_PYRAMID_PARSER.error_handler
+def _handle_error(error, _req, _schema, _status_code, _headers):
+    raise ValidationError(messages=error.messages) from error
```
seanh: I'd like to try moving `_handle_error` and `_PYRAMID_PARSER` into `PyramidRequestSchema` itself and see if that works

lms/validation/_launch_params.py
```diff
@@ -53,10 +50,7 @@ def lti_user(self):
 
         :rtype: LTIUser
         """
-        try:
-            kwargs = parser.parse(self, self.context["request"], locations=["form"])
-        except HTTPUnprocessableEntity as err:
-            raise ValidationError(err.json) from err
+        kwargs = self.parse(locations=["form"])
```
seanh: Same thing here: `self.parse()` already raises `ValidationError` not `HTTPUnprocessableEntity`, so it's not necessary to wrap the exception here

tests/lms/validation/__init___test.py
```diff
@@ -5,39 +5,12 @@
...
-                mock.sentinel.status_code,
-                mock.sentinel.headers,
-            )
-
-        # It exposes the webargs exception's messages as .messages.
-        assert exc_info.value.messages == webargs_exception.messages
-
-        # It raises the new exception from the original exception, so that the
-        # original exception isn't lost.
-        assert exc_info.value.__cause__ == webargs_exception
```
seanh: I should re-add this test on the new `_handle_error()`

lms/validation/_helpers/_base.py
```diff
@@ -45,6 +42,40 @@ def parse(self, *args, **kwargs):
...
+
+        :raise lms.validation.ValidationError: if the response isn't valid
+        """
+        response = self.context["response"]
+
+        try:
+            response_params = response.json()
+        except ValueError as err:
+            raise ValidationError(
+                {"_schema": "response doesn't have a valid JSON body"}
```
seanh: The `"_schema"` key is marshmallow's standard way of handling an error that doesn't apply to any one particular field

robertknight: My reading of the `ValidationError` API docs is that `_schema` is the default field, so `ValidationError(<string message>)` would be an equivalent way of writing this?

lms/validation/_helpers/_base.py
seanh: Done

lms/validation/_helpers/_base.py
seanh: Ah, except this is an `lms.validation.ValidationError` not a `marshmallow.ValidationError`. I've moved the JSON parsing into a marshmallow pre-load method so that it can make use of this `marshmallow.ValidationError` feature

PR #717, opened 2019-06-21
Break up TestCanvasAPIClient

seanh:
This test class was getting too long, and this will make adding more tests easier.

### DIFF
--- a/tests/lms/services/canvas_api_test.py
+++ b/tests/lms/services/canvas_api_test.py
@@ -14,8 +14,10 @@ from lms.services import CanvasAPIError, CanvasAPIServerError
 from lms.services.canvas_api import CanvasAPIClient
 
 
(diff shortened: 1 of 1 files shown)


tests/lms/services/canvas_api_test.py
```diff
@@ -14,8 +14,10 @@
 from lms.services.canvas_api import CanvasAPIClient
 
 
-class TestCanvasAPIClient:
-    def test_get_token_sends_an_access_token_request(
+class TestGetToken:
+    """Unit tests for CanvasAPIClient.get_token()."""
```
robertknight: I wonder if we ought to include the class name in the test class name as well to avoid needing these comments (eg. `TestCanvasAPIClient_GetToken`). That might be a bit redundant though if the module being tested only has one class.

PR #723, opened 2019-06-25
Three small validation and error pages fixes

seanh:


lms/validation/_helpers/_base.py
```diff
@@ -22,6 +22,23 @@ class Meta:
...
+
+    Subclasses can override this to control where parameters are searched for::
+
+        class MySchema(PyramidRequestSchema):
+            locations = ["form"]
+
+            ...
+
+    For the list of available locations see:
+    https://webargs.readthedocs.io/en/latest/quickstart.html#request-locations
```
robertknight: Can you think of a scenario where we might want to use the same schema in different contexts but search in different locations?

seanh: I'm not sure whether that'll come up or not. Currently you would have to subclass the schema and override its `locations` attr. If the need comes up in future we can devise a way to specify locations in view config. The standard webargs view decorator (which we aren't using because it's a bad fit for Pyramid) lets you do just that: https://webargs.readthedocs.io/en/latest/framework_support.html#id6

PR #726, opened 2019-06-28
Canvas API backend error handling 2/6: Add validated_response() helper method

seanh:
Depends on <https://github.com/hypothesis/lms/pull/729> for CI to pass.

Add a `validated_response()` helper that takes a `requests.PreparedRequest` and an optional validation schema, sends the request, and validates the response. It'll either return a `requests.Response` with the parsed and validated params as `response.parsed_params`, or if anything goes wrong with the request or validation it'll raise `CanvasAPIAccessTokenError` or `CanvasAPIServerError`.

Change the existing `get_token()` method to use the new helper for the access token request to the Canvas API.

In the future validation and error handling will be added to the other Canvas API requests, using the same helper.

This isn't really a Canvas-specific helper, it's really just a helper for working with `requests`, and some duplication could be removed by also using the helper when calling the h API. In the interests of time and scope I'm only using it for Canvas right now.

### DIFF
--- a/tests/lms/services/_helpers/canvas_api_test.py
+++ b/tests/lms/services/_helpers/canvas_api_test.py
@@ -158,7 +158,7 @@ def raise_from(request):
 
     request.addfinalizer(finalizer)
 
-    # Replace CanvasAPIError.raise_from() with a mocl.
+    # Replace CanvasAPIError.raise_from() with a mock.
     # This is done manually because using the normal mock.patch() or even
     # mock.patch.object() to patch this static method didn't seem to work.
     CanvasAPIError.raise_from = mock.create_autospec(CanvasAPIError.raise_from)
--- a/tests/lms/services/_helpers/canvas_api_test.py
+++ b/tests/lms/services/_helpers/canvas_api_test.py
@@ -1,6 +1,19 @@
+from unittest import mock
+
 import pytest
+from requests import ConnectionError
+from requests import HTTPError
+from requests import ReadTimeout
+from requests import TooManyRedirects
 
 from lms.services._helpers.canvas_api import CanvasAPIHelper
+from lms.services.exceptions import (
+    CanvasAPIError,
+    CanvasAPIAccessTokenError,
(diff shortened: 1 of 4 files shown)


tests/lms/services/_helpers/canvas_api_test.py
```diff
@@ -48,17 +55,118 @@ def test_public_url_request(self, ai_getter, route_url):
...
+def raise_from(request):
+    # Always replace the original CanvasAPIError.raise_from() after each test.
+    original = CanvasAPIError.raise_from
+
+    def finalizer():
+        CanvasAPIError.raise_from = original
+
+    request.addfinalizer(finalizer)
+
+    # Replace CanvasAPIError.raise_from() with a mocl.
```
robertknight: s/mocl/mock

tests/lms/services/_helpers/canvas_api_test.py
```diff
@@ -48,17 +55,118 @@ def test_public_url_request(self, ai_getter, route_url):
...
+    # This is done manually because using the normal mock.patch() or even
+    # mock.patch.object() to patch this static method didn't seem to work.
```
robertknight: Thanks for the clarification. Out of interest what was the error/issue?

seanh:
Yes, I agree that there's a lot of logic here that's just generic to doing API requests and could even be reused for the h API requests. So maybe an `ExternalRequestHelper` and `ExternalRequestError`s, rather than Canvas-specific stuff. That'd be a refactor to do in future,

tests/lms/services/_helpers/canvas_api_test.py
seanh: It appeared that mock replaced the static method with a non-callable mock. I didn't investigate for long, to be honest

tests/lms/services/_helpers/canvas_api_test.py
seanh: Fixed

seanh:
Rebased

PR #728, opened 2019-06-28
Canvas API backend error handling 4/6: Add schemas for list_files and public_url responses

seanh:
Add schemas for parsing and validating the Canvas API's list_files and public_url responses. Depends on <https://github.com/hypothesis/lms/pull/727> for the tests to pass.

### DIFF
--- a/tests/lms/validation/_canvas_test.py
+++ b/tests/lms/validation/_canvas_test.py
@@ -1,0 +1,272 @@
+import json
+from unittest import mock
+
+import pytest
+import requests
+
+from lms.validation import (
+    CanvasListFilesResponseSchema,
+    CanvasPublicURLResponseSchema,
+    ValidationError,
+)
+
+
+class TestCanvasListFilesResponseSchema:
+    def test_it_returns_the_list_of_files(self, list_files_response):
+        parsed_params = CanvasListFilesResponseSchema(list_files_response).parse()
+
+        assert parsed_params == [
+            {
+                "display_name": "TEST FILE 1",
+                "id": 188,
+                "updated_at": "2019-05-08T15:22:31Z",
+            },
+            {
+                "display_name": "TEST FILE 2",
+                "id": 181,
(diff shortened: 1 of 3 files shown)


seanh:
I pushed two commits to deduplicate a couple of things:

1. Combined `TestCanvasListFilesResponseSchema`'s three tests for missing required fields into a single parametrized test
2. Extracted tests for when the JSON is a completely invalid format (not even an object) and when it doesn't even parse as JSON into an abstract base class that's shared by both schema test classes. Also extracted instantiation of the schema objects themselves into a `schema` fixture to make this possible.

Beyond that I think the tests are largely similar-looking but not quite the same, and very simple, such that the complexity of deduplicating the code would outweigh the benefits.

For example `CanvasPublicURLResponseSchema` has tests for when a required field (`public_url`) is missing from the top-level JSON object, whereas `CanvasListFilesResponseSchema` has tests for when a required field is missing from _one of the objects_ in the list of objects that it parses. On the surface these seem similar but in fact both the test set up (deleting a required field) and the expected error message format are different.

tests/lms/validation/_canvas_test.py
```diff
@@ -0,0 +1,246 @@
...
+import requests
+
+from lms.validation import (
+    CanvasListFilesResponseSchema,
+    CanvasPublicURLResponseSchema,
+    ValidationError,
+)
+
+
+class CommonResponseSchemaTests(metaclass=ABCMeta):
```
seanh: Making this an ABC means that pytest doesn't try to run this test class directly

robertknight: How does that work? Does pytest have specific logic to recognize that a class uses `ABCMeta` or does it happen via some other mechanism?

seanh: Ah no, I was mistaken, it's because the class name doesn't match the test discovery class name pattern (doesn't start with `Test`). I removed the ABC stuff cos it's actually probably more confusing than semantic

PR #729, opened 2019-06-28
Canvas API backend error handling 1/6: Add CanvasAPIError factory

seanh:
Given an exception caused by an attempted Canvas API request, there's
quite a lot of logic about how we wrap that exception in the appropriate
lms.services.CanvasAPIError subclass:

* The exception may be a requests exception representing a 401
  Unauthorized response from Canvas, in which case we wrap it in a
  CanvasAPIAccessTokenError.

* Any other exception (a requests exception representing a 400 Bad
  Request response from Canvas, or a 500 Server Error from Canvas, or a
  network error, or a validation error when validating Canvas's
  response, ...) gets wrapped in a CanvasAPIServerError.

* In cases where a response was actually received from Canvas it's set
  as the exception's `response` attribute, otherwise `response` is
  `None`

* The exception has a `details` attribute containing a string
  representation of the wrapped exception and the validation errors
  and/or response that caused it

Move all this exception-creating logic into a new
CanvasAPIError.raise_from() factory.

The CanvasAPIError's raised here will eventually bubble up to error
views which will turn them into appropriate JSON responses for our
frontend code to receive.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -28,8 +28,14 @@ class ExternalRequestError(ServiceError):
     """
     A problem with a network request to an external service.
 
-    :param response: The response from the HTTP request to the h API
-    :type response: requests.Response
+    :arg explanation: A short error message for displaying to the user
+    :type explanation: str
+
+    :arg response: The external service's response to our HTTP request, if any
+    :type response: requests.Response or ``None``
+
+    :arg details: Additional details about what went wrong, for debugging
+    :type details: JSON-serializable dict or ``None``
     """
 
     def __init__(self, explanation=None, response=None, details=None):
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -32,10 +32,11 @@ class ExternalRequestError(ServiceError):
     :type response: requests.Response
     """
 
-    def __init__(self, explanation=None, response=None):
+    def __init__(self, explanation=None, response=None, details=None):
         super().__init__()
         self.explanation = explanation
         self.response = response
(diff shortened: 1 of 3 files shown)


lms/services/exceptions.py
```diff
@@ -32,10 +32,11 @@ class ExternalRequestError(ServiceError):
     :type response: requests.Response
```
robertknight: The docstrings here have gotten out of sync with the actual params.

tests/lms/services/exceptions_test.py
```diff
@@ -54,8 +62,143 @@ def test_when_theres_a_response_it_uses_it_in_str(
...
+            priority=1,
+            body=json.dumps({"foo": "bar"}),
+            **kwargs
+        )
+
+        response = requests.get("https://example.com")
+
+        try:
+            response.raise_for_status()
+        except requests.RequestException as err:
```
robertknight: What are the limitations of constructing the `RequestException` manually vs dynamically generating them as you are here?

seanh: :man_shrugging: I _think_ you could construct a `requests.RequestException` with all the necessary fields manually. `requests` does have some fields that it refuses to let you set, so we might run into that (but you can get around that by mocking). But I think it's just nicer to use the real `requests` to create the exception (using `httpretty` to trick `requests` into doing that) because then we're testing in better integration with `request`.

lms/services/exceptions.py
seanh: Thanks!

PR #739, opened 2019-07-01
Hide "Select PDF from Canvas" button if Canvas files is not available

### LINKED ISSUE #624: Canvas files details and niceties
This is a card for adding all the details and polish to the new Canvas files and OAuth 2.0 support. All of these need to be handled both when creating assignments and when launching assignments:

- [x] Don't show the Canvas Files option if this application instance has no Canvas developer key or if the LMS isn't even Canvas

- [x] Don't open the authorization popup window if we already have a non-expired access token

- [x] Don't open the authorization popup window even if our access token is expired, if we can use the refresh token to get a new access token (https://github.com/hypothesis/lms/pull/753)

- [ ] Support scoped dev keys: https://github.com/hypothesis/lms/issues/256

- [x] Long filenames wrapping issue: https://github.com/hypothesis/lms/issues/618

- [x] Mouse pointer doesn't change on hover over buttons:

   Screenshot from 2019-05-17 17-38-47

robertknight:
- Add `enableLmsFilePicker` config option to file picker JS app and set
   it in the `content_item_selection` view

   It would have been possible to just test for the presence of the existing
   `courseId` config field which is only set when the Canvas files API is
   available. However I opted to use a separate, more explicit flag instead, as
   it is possible that a field like "courseId" might be set for other
   reasons in future, and that other LMSes may use different conditions
   to enable or disable the LMS file picker.

 - Change `new_content_item_selection` template to use the `js-config`
   JSON script tag rendering provided by the base template instead of
   duplicating it. All the necessary config was already present in that
   tag.

Resolves one of the tasks in #624

### DIFF
--- a/lms/static/scripts/file_picker_v2/components/FilePickerApp.js
+++ b/lms/static/scripts/file_picker_v2/components/FilePickerApp.js
@@ -32,6 +32,7 @@ export default function FilePickerApp({
     authToken,
     authUrl,
     courseId,
+    enableLmsFilePicker = false,
     formAction,
     formFields,
     googleClientId,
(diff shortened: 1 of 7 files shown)


robertknight:
Hmm ... there is an unrelated test failing that I need to debug.

robertknight:
@seanh - Something that caught me out here is that we have two different views that render the file picker and they both configure `context.js_config` separately. That makes it easy to make a change in one place and forget to make any necessary changes in the other. We could address this with a code comment, but that doesn't seem particularly robust.

PR #741, opened 2019-07-02
Catch and display errors reported by LMS backend API calls

robertknight:
This PR implements handling of different types of error that can happen when fetching files from the LMS backend. There are three possibilities for the kind of error that may happen:

1. The server responds with a generic 400 error but no details
2. The server responds with a 4xx error and specific details
3. Some other error happens,  such as a network failure

In the first case, the frontend assumes an authorization error and displays the same authorization message and button text as when the dialog is shown for the first time. In the other cases,  the error details are displayed inline:

lms-file-picker-try-again

Clicking the "send us an email" link will open an email to support@hypothes.is with the details pre-filled. Clicking the "Try again" button will re-authorize and re-fetch files. In other words, exactly the same steps as when clicking the "Authorize" button when the dialog first appears. Note that in this PR, the authorization form is always displayed when the dialog is first shown. Fixing that will happen in a follow-up PR.

The new `ErrorDisplay` component used to display errors inline has also been added to the error dialog which is used when requests to the Google Drive Picker fail:

google-picker-error-dialog

I also fixed a bug where clicking "X" in this dialog did nothing.

See commit messages for implementation notes.

### DIFF
--- a/lms/static/scripts/file_picker_v2/components/ErrorDisplay.js
+++ b/lms/static/scripts/file_picker_v2/components/ErrorDisplay.js
@@ -1,0 +1,81 @@
+import propTypes from 'prop-types';
+import { createElement } from 'preact';
+
+function emailLink({ address, subject = '', body = '' }) {
+  return `mailto:${address}?subject=${encodeURIComponent(
+    subject
+  )}&body=${encodeURIComponent(body)}`;
+}
+
+/**
+ * Displays details of an error, such as a failed API call and provide the user
+ * with information on how to get help with it.
+ */
+export default function ErrorDisplay({ message, error }) {
+  let details = '';
+  try {
+    if (error.details) {
+      details = JSON.stringify(error.details, null, 2 /* indent */);
+    }
+  } catch (e) {
+    // Ignored
+  }
+
+  const supportLink = emailLink({
+    address: 'support@hypothes.is',
+    subject: 'Hypothesis LMS support',
+    body: `
+
+Error message: ${error.message}
+
+Technical details:
+
+${details}
+    `,
+  });
+
+  return (
+    // nb. Wrapper `<div>` here exists to apply block layout to contents.
+    <div>
+      <p>
+        {message}: <i>{error.message}</i>
+      </p>
+      <p>
+        If the problem persists{' '}
+        <a href={supportLink} target="_blank" rel="noopener noreferrer">
+          send us an email
+        </a>{' '}
+        or <a href="https://web.hypothes.is/get-help/">open a support ticket</a>
+        . You can also visit our{' '}
+        <a
+          href="https://web.hypothes.is/help/"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          help documents
+        </a>
+        .
+      </p>
+      {!!details && (
+        <details>
+          <pre className="ErrorDisplay__details">{details}</pre>
+        </details>
+      )}
+    </div>
+  );
+}
+
+ErrorDisplay.propTypes = {
+  /**
+   * A short message explaining that a problem happened.
+   */
+  message: propTypes.string.isRequired,
+
+  /**
+   * An `Error`-like object with details of the problem.
+   *
+   * This is assumed to have a string `message` property and may have a
+   * JSON-serializable `details` property.
+   */
+  error: propTypes.object.isRequired,
+};
--- a/lms/static/scripts/file_picker_v2/components/LMSFilePicker.js
+++ b/lms/static/scripts/file_picker_v2/components/LMSFilePicker.js
@@ -2,11 +2,13 @@ import { createElement } from 'preact';
 import { useCallback, useEffect, useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
+import { ApiError, listFiles } from '../utils/api';
+
 import AuthWindow from '../utils/AuthWindow';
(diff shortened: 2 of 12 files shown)


lms/static/scripts/file_picker_v2/components/ErrorDisplay.js
```diff
@@ -0,0 +1,81 @@
...
+ * with information on how to get help with it.
+ */
+export default function ErrorDisplay({ message, error }) {
+  let details = '';
+  try {
+    if (error.details) {
+      details = JSON.stringify(error.details, null, 2 /* indent */);
+    }
+  } catch (e) {
+    // Ignored
```
robertknight: AFAIK the only way `JSON.stringify` can fail is if the input object contains circular references, which should never happen here. Since this error comes from an arbitrary source, I've added the `try ... catch` just to be defensive.

lms/static/scripts/file_picker_v2/components/LMSFilePicker.js
```diff
@@ -25,19 +27,20 @@ export default function LMSFilePicker({
...
+    files: null,
 
-  /**
-   * `true` if we are waiting for the user to authorize the app's access
-   * to files in the LMS.
-   */
-  const [isAuthorizing, setAuthorizing] = useState(!isAuthorized);
+    // Fetch error details. Set when state is "error".
+    error: null,
+  });
```
robertknight: The purpose of refactoring several `useState` calls into one here is to represent the state of the dialog in such a way that it cannot be in an invalid combination of states (eg. "fetching" and "authorizing" at the same time, or "authorizing" but with a non-empty file list)

seanh:
Gonna upload some testing notes...

seanh:
# User clicks <kbd>Cancel</kbd> instead of <kbd>Authorize</kbd> :white_check_mark: 

This one is a little confusing, especially in the cropped GIF, because when the user clicks <kbd>Cancel</kbd> in Canvas in the popup window they're first taken to an error dialog within the popup window. If they click <kbd>Cancel</kbd> again in that dialog then the popup window closes and they get back to the original dialog.

auth-cancelled

This could be handled better if no error message were shown in the popup window, and instead the popup window closed itself and any error message was shown back in the original frame. This would require postMessage communication between the popup window and the original frame if an error message is to be passed from the popup window to the original frame in order to explain to the user that authorization failed.

That's beyond the scope of this PR and probably too edge-casey to consider doing at this point. As far as this PR goes I think the behaviour in this case is correct.

seanh:
### User closes popup window without clicking either <kbd>Cancel</kbd> or <kbd>Authorize</kbd> :white_check_mark: 

closedpopup

### UI flicker

After the user closes the popup window you can see a slight "flicker" of the UI back in the original frame as it attempts the list_files request, gets a failure, and goes back to the _Allow file access_ dialog. This flicker will probably be more noticeable on production where the list_files request will be slower. Here it is when I've added an exaggerated three second delay to the list_files response so that you can see what's actually happening rather than a flicker:

Peek 2019-07-04 11-11

I think this flicker could be avoided by making the loading spinner, which shows while the list_files request is in flight, be part of the _Allow file access_ dialog rather than part of the _Select a file_ dialog, and don't show the _Select a file_ dialog until the list of files has been successfully received.

This is a detail and I don't think is required for releasing the feature.

### "You closed the popup" error message

As with when the user clicks <kbd>Cancel</kbd> instead of <kbd>Authorize</kbd> in the popup window (see above), when the user closes the popup window without clicking either  <kbd>Cancel</kbd> or <kbd>Authorize</kbd> it would be nice to show a "Authorization popup window was closed" error message to explain to the user why they're being asked to authorize again.

This would require some sort of communication between the popup window and the original frame, and I think it implies an "Authorization succeeded" message from popup to original frame, since I'm not sure that the popup would be able to send an "The user is closing me without clicking <kbd>Cancel</kbd> or <kbd>Authorize</kbd> " message.

Beyond the scope of this PR, and I don't think required to release the feature.

seanh:
I don't think this is important or a blocker, but it's possible for the server's error details to be too tall to fit:

Screenshot from 2019-07-04 11-24-54

seanh:
# Non-access token error from our server :white_check_mark: 

Something not related to the access token went wrong with our server's request to the Canvas API -- error response or invalid response from Canvas, network error so that no response was received, etc. See https://github.com/hypothesis/lms/pull/731 for ways to produce several such errors in dev.

error

The same UI flicker as mentioned above is there again.

seanh:
# Access token error from server :white_check_mark: 

We don't have an access token for this user, or our access token is expired:

Peek 2019-07-04 11-32

seanh:
Non-blocking suggestion: change the dialog title from _Select a file_ to _Fetching files changed_, when an error message and <kbd>Try again</kbd> button are being shown in this dialog.

Screenshot from 2019-07-04 11-35-43

seanh:
# Error message with no details :white_check_mark: 

Although it never actually happens currently, theoretically our server could send a 400 response with an `error_message` but no `details` (`details` would be `null`). The frontend seems to handle this fine:

Screenshot from 2019-07-04 11-38-17

seanh:
The support and email links work fine :white_check_mark: 

When there's an `error_message` but no `details` then the email body says "Technical details:" with nothing following it. But this is an edge case and doesn't currently ever happen.

PR #749, opened 2019-07-04
Get rid of save_token()

seanh:
Depends on https://github.com/hypothesis/lms/pull/747 for CI to pass.

Simplify CanvasAPIClient by getting rid of the separate save_token()
method, and just having get_token() save the token. So instead of:

    token = canvas_api_client.get_token()
    canvas_api_client.save_token(token)

It's now just:

    canvas_api_client.get_token()


seanh:
Rebased

PR #752, opened 2019-07-04
Add get_refreshed_token()

seanh:
Depends on https://github.com/hypothesis/lms/pull/748 and https://github.com/hypothesis/lms/pull/751 for CI to pass

### DIFF
--- a/lms/services/canvas_api.py
+++ b/lms/services/canvas_api.py
@@ -7,6 +7,7 @@ from lms.services import CanvasAPIAccessTokenError
 from lms.services._helpers import CanvasAPIHelper
 from lms.validation import (
     CanvasAccessTokenResponseSchema,
+    CanvasRefreshTokenResponseSchema,
     CanvasListFilesResponseSchema,
     CanvasPublicURLResponseSchema,
 )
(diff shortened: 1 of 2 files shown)


seanh:
Rebased

lms/services/canvas_api.py
```diff
@@ -50,6 +51,22 @@ def get_token(self, authorization_code):
...
+        response = self._helper.validated_response(
+            self._helper.refresh_token_request(refresh_token),
+            CanvasRefreshTokenResponseSchema,
+        )
+
+        new_access_token = response.parsed_params["access_token"]
+
+        self._save(
+            new_access_token,
+            response.parsed_params.get("refresh_token", refresh_token),
```
seanh: If no new refresh token is given in the refresh token response, then we keep the original refresh token in the DB with the new access token. This is the case with Canvas: they want you to keep the same refresh token and keep on using it

lms/services/canvas_api.py
```diff
@@ -50,6 +51,22 @@ def get_token(self, authorization_code):
...
+            response.parsed_params.get("expires_in"),
+        )
+
+        return new_access_token
```
seanh: Code that calls this is going to need the new access token right away so return it rather than making the caller have to query the DB for it

PR #753, opened 2019-07-04
Automatically refresh and retry if list_files or public_url fails

seanh:
Depends on https://github.com/hypothesis/lms/pull/752 for CI to pass.

### DIFF
--- a/lms/services/canvas_api.py
+++ b/lms/services/canvas_api.py
@@ -85,10 +85,13 @@ class CanvasAPIClient:
 
         :rtype: list(dict)
         """
-        return self._helper.validated_response(
-            self._helper.list_files_request(self._oauth2_token.access_token, course_id),
+        oauth2_token = self._oauth2_token
+
+        return self.send_with_refresh_and_retry(
+            self._helper.list_files_request(oauth2_token.access_token, course_id),
             CanvasListFilesResponseSchema,
-        ).parsed_params
+            oauth2_token.refresh_token,
+        )
 
     def public_url(self, file_id):
         """
@@ -108,10 +111,25 @@ class CanvasAPIClient:
 
         :rtype: str
         """
-        return self._helper.validated_response(
-            self._helper.public_url_request(self._oauth2_token.access_token, file_id),
+        oauth2_token = self._oauth2_token
+
+        return self.send_with_refresh_and_retry(
+            self._helper.public_url_request(oauth2_token.access_token, file_id),
             CanvasPublicURLResponseSchema,
-        ).parsed_params["public_url"]
+            oauth2_token.refresh_token,
+        )["public_url"]
+
+    def send_with_refresh_and_retry(self, request, schema, refresh_token):
+        try:
+            return self._helper.validated_response(request, schema).parsed_params
+        except CanvasAPIAccessTokenError:
+            if not refresh_token:
+                raise
+
+            new_access_token = self.get_refreshed_token(refresh_token)
+            return self._helper.validated_response(
+                request, schema, new_access_token
+            ).parsed_params
 
     def _save(self, access_token, refresh_token, expires_in):
         """
--- a/tests/lms/services/canvas_api_test.py
+++ b/tests/lms/services/canvas_api_test.py
@@ -337,6 +337,151 @@ class TestPublicURL:
             canvas_api_client.public_url("test_file_id")
 
 
+class TestSendWithRefreshAndRetry:
+    def test_it_sends_the_request_and_returns_the_parsed_params(
+        self,
+        canvas_api_client,
+        canvas_api_helper,
+        prepared_request,
+        Schema,
+        refresh_token,
+    ):
+        parsed_params = canvas_api_client.send_with_refresh_and_retry(
+            prepared_request, Schema, refresh_token
+        )
+
+        canvas_api_helper.validated_response.assert_called_once_with(
+            prepared_request, Schema
+        )
+        assert (
+            parsed_params
+            == canvas_api_helper.validated_response.return_value.parsed_params
+        )
+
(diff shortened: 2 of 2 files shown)


lms/services/canvas_api.py
```diff
@@ -68,10 +68,13 @@ def list_files(self, course_id):
...
-        return self._helper.validated_response(
-            self._helper.list_files_request(self._oauth2_token.access_token, course_id),
+        oauth2_token = self._oauth2_token
+
+        return self.send_with_refresh_and_retry(
+            self._helper.list_files_request(oauth2_token.access_token, course_id),
             CanvasListFilesResponseSchema,
-        ).parsed_params
+            oauth2_token.refresh_token,
+        )
```
seanh: This is just the exact same request and response, but now sent via the new `send_with_refresh_and_retry()`

lms/services/canvas_api.py
```diff
@@ -91,10 +94,25 @@ def public_url(self, file_id):
...
-        return self._helper.validated_response(
-            self._helper.public_url_request(self._oauth2_token.access_token, file_id),
+        oauth2_token = self._oauth2_token
+
+        return self.send_with_refresh_and_retry(
+            self._helper.public_url_request(oauth2_token.access_token, file_id),
             CanvasPublicURLResponseSchema,
-        ).parsed_params["public_url"]
+            oauth2_token.refresh_token,
+        )["public_url"]
```
seanh: This is just the exact same request and response, but now sent via the new `send_with_refresh_and_retry()`

lms/services/canvas_api.py
```diff
@@ -91,10 +94,25 @@ def public_url(self, file_id):
...
+        try:
+            return self._helper.validated_response(request, schema).parsed_params
+        except CanvasAPIAccessTokenError:
+            if not refresh_token:
+                raise
+
+            new_access_token = self.get_refreshed_token(refresh_token)
+            return self._helper.validated_response(
+                request, schema, new_access_token
+            ).parsed_params
```
seanh: So the logic is:

1. If the request succeeds the first time then send back the response
2. If the request fails with a non-access token error (some `CanvasAPIError` that is not a `CanvasAPIAccessTokenError` then _don't_ refresh, send back an error response
3. If the request fails for access token-related reasons then attempt a refresh request
4. If the refresh request fails for any reason then send back an error response with the details of the failed refresh request on the `error_message` and `details`
5. If the refresh request succeeds then re-try the original request but with the new access token, and either send back a successful response or an error response with the `error_message` and `details` based on the second failed request

Only one refresh and retry is ever attempted per proxy API call.

lms/services/canvas_api.py
```diff
@@ -91,10 +94,25 @@ def public_url(self, file_id):
...
+        )["public_url"]
+
+    def send_with_refresh_and_retry(self, request, schema, refresh_token):
+        try:
+            return self._helper.validated_response(request, schema).parsed_params
+        except CanvasAPIAccessTokenError:
+            if not refresh_token:
+                raise
+
+            new_access_token = self.get_refreshed_token(refresh_token)
```
seanh: `get_refreshed_token()` saves the new access token to the DB for next time, as well as returning it for right now.

lms/services/canvas_api.py
```diff
@@ -91,10 +94,25 @@ def public_url(self, file_id):
...
+            return self._helper.validated_response(
+                request, schema, new_access_token
```
seanh: re-send the same request but with the new access token

lms/services/canvas_api.py
```diff
@@ -68,10 +68,13 @@ def list_files(self, course_id):
 
         :rtype: list(dict)
         """
-        return self._helper.validated_response(
-            self._helper.list_files_request(self._oauth2_token.access_token, course_id),
+        oauth2_token = self._oauth2_token
```
robertknight: If the refresh token used is always `self._oauth2_token` then that lookup could be pushed down into `send_with_refresh_and_retry`.

lms/services/canvas_api.py
```diff
@@ -91,10 +94,25 @@ def public_url(self, file_id):
...
+        return self.send_with_refresh_and_retry(
+            self._helper.public_url_request(oauth2_token.access_token, file_id),
             CanvasPublicURLResponseSchema,
-        ).parsed_params["public_url"]
+            oauth2_token.refresh_token,
+        )["public_url"]
+
+    def send_with_refresh_and_retry(self, request, schema, refresh_token):
+        try:
+            return self._helper.validated_response(request, schema).parsed_params
```
robertknight: Is the `expires_in` checked before the request is actually sent? If not, we might be able to save a few hundred ms by not even sending the initial request if we know the token has definitely expired.

seanh:
Rebased

lms/services/canvas_api.py
seanh: Except that would cause two DB queries to happen instead of one. This could be optimised in a nicer way but for now this is how we avoid querying for the OAuth2Token twice

lms/services/canvas_api.py
seanh: Nope, although we record `expires_in` (and also the `received_at` needed to interpret it) we never currently make use of it. I agree that this would be a simple and worthwhile optimisation to add because the common case of an expired access token (Canvas access tokens last for one hour) results in quite a few network hops:

1. The XHR request from the browser to our server and back
2. Our server's first attempt to send the Canvas API request and wait for the response
3. Refresh request and response
4. Second try at the original request and response

So the top-level XHR request here could be quite slow to respond, especially since some Canvas API's are pretty slow, and eliminating one of the network round-trips here should be worthwhile.

PR #756, opened 2019-07-05
Improve aesthetics and accessibility of file list

### LINKED ISSUE #742: Difficult to tell which file is selected, in Canvas file picker
The grey background colour used to highlight the selected file isn't distinct enough from the two different greys used for non-selected files. For example in the screenshots below, which file is selected?

Screenshot from 2019-07-02 12-08-54
Screenshot from 2019-07-02 12-08-40
Screenshot from 2019-07-02 12-08-24
Screenshot from 2019-07-02 12-08-04

I think the solution is to use a bright "action colour" instead of a shade of grey. I attempted a fix using red. I think using the same colour as the <kbd>Select</kbd> button's hover colour makes sense, but unfortunately I'm not sure the black PDF icon looks great against this red:

Screenshot from 2019-07-02 12-24-59

Maybe the icon just needs to be tweaked.

Red is an odd colour to be used for positive actions like _Select_. Red normally means warning or error. But it's our brand colour and I think we use it like this in lots of places in multiple apps.

### LINKED ISSUE #743: Icons and text not vertically aligned in Canvas file picker
The vertical alignment of these PDF icons is off. There's more empty space below than above the icon, and the icon's middle isn't vertically centered with the text label's middle. The filename text is also not vertically centred (more space below than above) and then the date seems top-aligned which differs from the filename (it looks like they're all top-aligned but the tallness of the icon pushes down the filename text making the alignment of all three elements look different):

Screenshot from 2019-07-02 12-16-59

Here I've fixed it by adding `vertical-align: middle` to the icon, filename and date. I think middle-aligned looks better anyway and this keeps them all aligned with each other and above- and below-gaps equal:

Screenshot from 2019-07-02 12-17-55

robertknight:
This PR implements several improvements to the aesthetics and accessibility of the file list:

**General aesthetics and UX:**

- Use a much stronger visual highlight for selected rows
- Remove borders between cells. These were inherited from global styles in `lms.css`
- Remove the focus outline from the table, but only when not using keyboard input (see [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) notes)
- Vertically center file names and labels

**Accessibility:**

- Add accessible labels to the dialog's "X" button and the whole table itself
- Add an the accessible label for the contents of the name column, so that screen readers just read the filename and don't try to describe the icon
- Identify the table as a [grid](https://www.w3.org/TR/wai-aria-practices/#grid) for a11y purposes. A "grid" is interactive whereas a plain table is typically for static data
- Sync keyboard focus with selected item in the table, so that screen readers will read out the selected row when navigating up and down

**File list with no row selected:**

file-list-no-selection

**File list with top row selected:**

file-list-first-row-selected

**File list with middle row selected:**

file-list-middle-row-selected

Fixes #742 
Fixes #743

### DIFF
--- a/lms/static/styles/mixins/focus.scss
+++ b/lms/static/styles/mixins/focus.scss
@@ -1,0 +1,23 @@
+/**
+ * Display an outline on an element only when it has keyboard focus.
+ *
+ * This requires the browser to support the :focus-visible pseudo-selector [1]
+ * or for the JS polyfill [2] to be loaded.
+ *
+ * [1] https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
+ * [2] https://github.com/WICG/focus-visible
+ */
+@mixin outline-on-keyboard-focus {
+  // Selector for browsers using JS polyfill, which adds the "focus-visible"
+  // class to a keyboard-focused element.
+  &:focus:not(.focus-visible) {
+    outline: none;
+  }
+
+  // Selector for browsers with native :focus-visible support.
+  // (Do not combine with above, as an unsupported pseudo-class disables the
+  // entire selector)
+  &:focus:not(:focus-visible) {
+    outline: none;
+  }
+}
--- a/lms/static/scripts/file_picker_v2/components/Dialog.js
+++ b/lms/static/scripts/file_picker_v2/components/Dialog.js
@@ -89,7 +89,11 @@ export default function Dialog({
             {title}
             <span className="u-stretch" />
             {onCancel && (
-              <button className="Dialog__cancel-btn" onClick={onCancel}>
+              <button
+                aria-label="Close"
+                className="Dialog__cancel-btn"
+                onClick={onCancel}
+              >
                 ✕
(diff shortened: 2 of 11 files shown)


lms/static/styles/mixins/focus.scss
```diff
@@ -0,0 +1,23 @@
+/**
```
robertknight: This file was copied from a file of the same name in the client repo. In future I think we'll want to extract this into a shared package.

seanh:
What do you think about also making the text in the selected row bold, to match the button?

Screenshot from 2019-07-05 10-55-22

robertknight:
> What do you think about also making the text in the selected row bold, to match the button?

Hmm, I gave it a try locally but the change in font weight makes the text jump around a little bit when it changes from selected/unselected, more so if the display name is long enough to be elided. It also no longer matches the table headings in that case.

seanh:
White icon fix works for me

PR #758, opened 2019-07-05
Pass correct `origin` value to Google Picker in new tabs

### LINKED ISSUE #421: Google Picker not working for assignments launched in a new window
**Note: this issue shows that the Google Picker is broken in Moodle for one install configuration. Choosing to load assignments in an embedded iframe does not cause Google Picker to break, and that is what I've advised users to do in the meantime.**

When you install the Hypothesis app for Moodle, whether for an individual course or at the institutional level, you have the option to have Hypothesis-enabled assignments launch in the same window or in a new window:

pasted image 0

Choosing the "new window" option is preferable for many instructors, in order to maximize the space given to the original text and annotation sidebar.  

However, choosing this option now causes the Google Picker to break. Note that an instructor can still copy-and-paste a public URL during the assignment creation process and it'll work just fine - it's only the Google Picker that is broken. 

I have edited our Moodle app so that a dev can easily reproduce this:

## Steps to reproduce

1. Log in to our Moodle instance (note that you can access our Moodle instance using [credentials in this doc](https://docs.google.com/document/d/1p7XiZclKWDlkOxlcuHXdkrsGtemGeR5gQiJ4qUq7o00/edit#heading=h.xcv7oqstt3uq)). 
2. Choose any course; you're welcome to use the one called [Katelyn's testing environment](https://hypothesisuniversity.moodlecloud.com/course/view.php?id=8). 
3. Turn editing on and create a new assignment (you might find [these instructions](https://docs.google.com/document/d/17poBMaXGEp93MdL1MvAcS2uCYKbobh027kfQa_qVZ_s/edit#heading=h.wq28g2mnwz80) helpful))
4. After you click "Save and Display," click on your new assignment. You'll be prompted to enter a public URL or use the Google Picker. Select **Use Google Picker**.
5. Authorize a Google account

### Expected behavior
A list of your Google Drive files shows 

### Actual behavior
A blank window:

screen shot 2019-01-16 at 4 24 28 pm

The browser console shows the following errors:
screen shot 2019-01-16 at 4 24 05 pm

robertknight:
When the file picker form is rendered into a top-level tab instead of an
iframe, pass the URL of the current page instead of the LMS URL as the
"origin" parameter that is forwarded to the Google Picker API.

(_Note that testing this with a unit test is fiddly because `window` and `window.top` are "unforgeable" APIs. I could add some indirection here to solve that, but it seems a bit pointless for something so simple_)

Fixes #421

### DIFF
--- a/lms/static/scripts/file_picker_v2/components/FilePickerApp.js
+++ b/lms/static/scripts/file_picker_v2/components/FilePickerApp.js
@@ -59,7 +59,12 @@ export default function FilePickerApp({
     return new GooglePickerClient({
       developerKey: googleDeveloperKey,
       clientId: googleClientId,
-      origin: lmsUrl,
+
+      // If the form is being displayed inside an iframe, then the backend
+      // must provide the URL of the top-level frame to us so we can pass it
+      // to the Google Picker API. Otherwise we can use the URL of the current
+      // tab.
+      origin: window === window.top ? window.location.href : lmsUrl,
     });
   }, [googleDeveloperKey, googleClientId, lmsUrl]);
 


robertknight:
I tested this locally in Moodle. It worked as expected, after I ensured that cached JS resources were cleared.

PR #764, opened 2019-07-05
Work around bug in PyLTI's handling of URL params with percent encoded characters

### LINKED ISSUE #689: New document selection frontend fails when document URL has non-ASCII chars
There's an issue with the new file picker that, when a public HTML URL assignment is created, if the URL contains non-ASCII characters then launching the assignment fails with _You're not authorized to view this page_. Even if you turn the feature flag off launching the assignment will still fail if the flag was on when the assignment was created. But if the flag was off during assignment creation then it works.

There's a difference in escaping of the URL that might be suspect. With an assignment created with the feature flag off the LTI launch URL (copied from dev tools) is `https://lms.hypothes.is/lti_launches?url=https://en.wikipedia.org/wiki/G%C3%B6reme_National_Park`. With the flag on it's` https://lms.hypothes.is/lti_launches?url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FG%25C3%25B6reme_National_Park`

robertknight:
Fix an OAuth 1.0 signature verification failure when handling an LTI
launch request where:

 - There is a "url" query param (aka. the launch request is "URL configured")
 - The "url" query param, _after decoding_, contains percent-encoded
   query parameters.

For example, given an LTI launch request with the following URL:

  "https://lms.hypothes.is/lti_launches?url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FG%25C3%25B6reme_National_park"

The LMS would generate an OAuth signature base string using the "url"
parameter "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark".

When our lms app received the request, it would parse the incoming
request so `request.params["url"]` was "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark".

However when PyLTI generates the signature base string for comparison,
it _decoded_ this value and used:

  "https://en.wikipedia.org/wiki/Göreme_National_Park"

The result is that the signatures did not match and the request aborted
with a 403 HTTP status.

Since this problem only affects parameters that PyLTI extracts from the
URL and not those passed to `verify_request_common` in the `parameters`
kwarg, and query params are present in both, the workaround here is to
exclude the query string + fragment from the value passed to
`verify_request_common` as the `url` argument.

Fixes #689

### DIFF
--- a/lms/services/launch_verifier.py
+++ b/lms/services/launch_verifier.py
@@ -63,7 +63,16 @@ class LaunchVerifier:
         try:
             valid = pylti.common.verify_request_common(
                 consumers,
-                self._request.url,
+                # We pass only the host + path of the URL (`request.path_url`)
+                # to `verify_request_common` in the `url` arg because:
+                #
+                # - Query params are also passed in `request.params` below
+                # - If query params are passed in both the URL and the `parameters`
+                #   dict, then the value from the URL is used
+                # - If values in the query param contain percent-encoded characters,
+                #   these are incorrectly _decoded_ in the result, whereas they
+                #   are handled correctly if passed in the `parameters` arg.
+                self._request.path_url,
                 self._request.method,
                 dict(self._request.headers),
                 dict(self._request.params),
--- a/tests/lms/services/launch_verifier_test.py
+++ b/tests/lms/services/launch_verifier_test.py
@@ -1,5 +1,6 @@
 import time
 from unittest import mock
+from urllib.parse import urlencode
 
 import oauthlib.oauth1
 import oauthlib.common
@@ -138,13 +139,40 @@ class TestVerifyLaunchRequest:
 
         assert pylti.common.verify_request_common.call_count == 1
 
+    # See https://github.com/hypothesis/lms/issues/689
+    def test_it_verifies_urls_with_percent_encoded_chars_in_params(
+        self, pyramid_request
+    ):
+        # Add a "url" query parameter where the value, after decoding the query
+        # string, contains percent-encoded chars.
+        params = {"url": "https://en.wikipedia.org/wiki/G%C3%B6reme_National_Park"}
+        pyramid_request.POST.update(params)
+
+        # Sign the pyramid_request using oauthlib.
+        sign(pyramid_request)
+
+        # Update `pyramid_request.url` to include the query string.
+        # We do this after signing because oauthlib will, correctly,
+        # include both the "url" param from `pyramid_request.params` and the "url"
+        # param from the query string. PyLTI however will only retain one copy
+        # when generating the signature for verification.
+        pyramid_request.url += "?" + urlencode(params)
+
+        launch_verifier = LaunchVerifier(mock.sentinel.context, pyramid_request)
+        launch_verifier.verify()
+
     @pytest.fixture
     def launch_verifier(self, pyramid_request):
         return LaunchVerifier(mock.sentinel.context, pyramid_request)
 
     @pytest.fixture
     def pyramid_request(self, pyramid_request):
-        pyramid_request.url = "http://example.com/"
+        # `pyramid.testing.DummyRequest` sets `url` and `path_url` attrs without
+        # a path by default. oauthlib however normalizes the URL to include a
+        # path when it generates the signature base string, whereas PyLTI does
+        # not. Add an empty path to both to match a real application.
+        pyramid_request.url += "/"
+        pyramid_request.path_url += "/"
 
         # Add the OAuth 1 params (version, nonce, timestamp, ...)
         oauthlib_client = oauthlib.oauth1.Client(


robertknight:
To reproduce the issue directly with pylti, run the following with Python 3.6:

```py
import time
from urllib.parse import urlencode

import pylti.common


consumers = {"TEST_CONSUMER": {"secret": "TEST_SECRET"}}
params = {
    "oauth_consumer_key": "TEST_CONSUMER",
    "oauth_signature_method": "HMAC-SHA1",
    "oauth_timestamp": int(time.time()),
    "oauth_nonce": "foo",
    "oauth_signature": "DUMMY_SIGNATURE",
}
query = urlencode({"url": "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"})
url = f"https://example.com/lti_launch?{query}"
pylti.common.verify_request_common(
    consumers=consumers, url=url, method="POST", params=params, headers={}
)
```

This will fail with the error:

```
oauth2.Error: Invalid signature. Expected signature base string: b'POST&https%3A%2F%2Fexample.com%2Flti_launch&oauth_consumer_key%3DTEST_CONSUMER%26oauth_nonce%3Dfoo%26oauth_si
gnature_method%3DHMAC-SHA1%26oauth_timestamp%3D1562586532%26url%3Dhttps%253A%252F%252Fen.wikipedia.org%252Fwiki%252FG%25C3%25B6reme_National_ark
```

If you extract the part of the message starting at "url%3D", and decode it, you'll get the query string param:

`"url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FG%C3%B6reme_National_ark"`

Decode this again to get the URL value and you'll get:

`"url=https://en.wikipedia.org/wiki/Göreme_National_ark"`

Which doesn't match the original "url" param from the query string above. The "ö" character should still be percent-encoded.

If on the other hand the `url` value is passed in the params:

```py
import time

import pylti.common


consumers = {"TEST_CONSUMER": {"secret": "TEST_SECRET"}}
params = {
    "oauth_consumer_key": "TEST_CONSUMER",
    "oauth_signature_method": "HMAC-SHA1",
    "oauth_timestamp": int(time.time()),
    "oauth_nonce": "foo",
    "oauth_signature": "DUMMY_SIGNATURE",
}
params.update({"url": "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"})
url = f"https://example.com/lti_launch"
pylti.common.verify_request_common(
    consumers=consumers, url=url, method="POST", params=params, headers={}
)
```

Then the error produced is:

```
oauth2.Error: Invalid signature. Expected signature base string: b'POST&https%3A%2F%2Fexample.com%2Flti_launch&oauth_consumer_key%3DTEST_CONSUMER%26oauth_nonce%3Dfoo%26oauth_si
gnature_method%3DHMAC-SHA1%26oauth_timestamp%3D1562587085%26url%3Dhttps%253A%252F%252Fen.wikipedia.org%252Fwiki%252FG%2525C3%2525B6reme_National_ark'
```

And if you extract the part from "url%3D" and double-decode it as before, you get the correct output:

```
"url=https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"
```

robertknight:
To reproduce the issue directly with pylti, run the following with Python 3.6:

```py
import time
from urllib.parse import urlencode

import pylti.common


consumers = {"TEST_CONSUMER": {"secret": "TEST_SECRET"}}
params = {
    "oauth_consumer_key": "TEST_CONSUMER",
    "oauth_signature_method": "HMAC-SHA1",
    "oauth_timestamp": int(time.time()),
    "oauth_nonce": "foo",
    "oauth_signature": "DUMMY_SIGNATURE",
}
query = urlencode({"url": "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"})
url = f"https://example.com/lti_launch?{query}"
pylti.common.verify_request_common(
    consumers=consumers, url=url, method="POST", params=params, headers={}
)
```

This will fail with the error:

```
oauth2.Error: Invalid signature. Expected signature base string: b'POST&https%3A%2F%2Fexample.com%2Flti_launch&oauth_consumer_key%3DTEST_CONSUMER%26oauth_nonce%3Dfoo%26oauth_si
gnature_method%3DHMAC-SHA1%26oauth_timestamp%3D1562586532%26url%3Dhttps%253A%252F%252Fen.wikipedia.org%252Fwiki%252FG%25C3%25B6reme_National_ark
```

If you extract the part of the message starting at "url%3D", and decode it, you'll get the query string param:

`"url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FG%C3%B6reme_National_ark"`

Decode this again to get the URL value and you'll get:

`"url=https://en.wikipedia.org/wiki/Göreme_National_ark"`

Which doesn't match the original "url" param from the query string above. The "ö" character should still be percent-encoded.

If on the other hand the `url` value is passed in the params:

```py
import time

import pylti.common


consumers = {"TEST_CONSUMER": {"secret": "TEST_SECRET"}}
params = {
    "oauth_consumer_key": "TEST_CONSUMER",
    "oauth_signature_method": "HMAC-SHA1",
    "oauth_timestamp": int(time.time()),
    "oauth_nonce": "foo",
    "oauth_signature": "DUMMY_SIGNATURE",
}
params.update({"url": "https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"})
url = f"https://example.com/lti_launch"
pylti.common.verify_request_common(
    consumers=consumers, url=url, method="POST", params=params, headers={}
)
```

Then the error produced is:

```
oauth2.Error: Invalid signature. Expected signature base string: b'POST&https%3A%2F%2Fexample.com%2Flti_launch&oauth_consumer_key%3DTEST_CONSUMER%26oauth_nonce%3Dfoo%26oauth_si
gnature_method%3DHMAC-SHA1%26oauth_timestamp%3D1562587085%26url%3Dhttps%253A%252F%252Fen.wikipedia.org%252Fwiki%252FG%2525C3%2525B6reme_National_ark'
```

And if you extract the part from "url%3D" and double-decode it as before, you get the correct output:

```
"url=https://en.wikipedia.org/wiki/G%C3%B6reme_National_ark"
```

robertknight:
PyLTI currently depends on an [unmaintained OAuth 1.0 Python library](https://github.com/joestump/python-oauth2/issues/223) (see also: https://github.com/mitodl/pylti/issues/78).

Since the only thing we use PyLTI for is OAuth 1.0 signature verification, I think we should consider replacing PyLTI with oauthlib directly, which we already use in our tests. oauthlib also supports OAuth 2.0 so this may come in useful for supporting [LTI v1.3](https://www.imsglobal.org/why-platforms-and-tools-should-adopt-lti-13) in future.

seanh:
I'd be happy to get rid of PyLTI. We're gonna end up needing to do so one day anyway, since it's unmaintained. We just need to get a simple implementation of OAuth 1 request signature verification working using oauthlib or another maintained library. Probably wait a couple of weeks before shipping it, just to control the amount of changes we're shipping at once

robertknight:
> Probably wait a couple of weeks before shipping it, just to control the amount of changes we're shipping at once

I agree that replacing the OAuth signature verification implementation is something we should do after everything behind the `new_oauth` flag is shipped and the dust has settled.

This PR is intended to be a very small diff that avoids a regression in certain cases, which we can potentially ship sooner.

seanh:
Rebased

PR #781, opened 2019-07-09
Open support ticket link in new tab

seanh:


lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -41,11 +41,14 @@ ${details}
         {message}: <i>{error.message}</i>
       </p>
       <p>
-        If the problem persists{' '}
-        <a href={supportLink} target="_blank" rel="noopener noreferrer">
-          send us an email
-        </a>{' '}
-        or <a href="https://web.hypothes.is/get-help/">open a support ticket</a>
+        If the problem persists <a href={supportLink}>send us an email</a> or{' '}
```
robertknight: In other apps I think we have generally not used `target="_blank"` for email links, but I think there was an issue if that link happens to be handled by a browser-based email client (eg. if you configure Chrome to open mailto: links with Gmail for example).

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -41,11 +41,14 @@ ${details}
...
+        <a
+          href="https://web.hypothes.is/get-help/"
+          target="_blank"
+          rel="noopener noreferrer"
```
robertknight: The "noreferrer" attribute exists as a [fallback security feature](https://mathiasbynens.github.io/rel-noopener/) for browsers that don't support "noopener". It is possible it might prove an annoyance for support trying to figure out where their traffic is coming from though, so we might want to reconsider at some point.

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
seanh: Yep, changed it to open all email links in new tabs instead, which is needed for webmail clients

PR #784, opened 2019-07-09
Even out file picker form's vertical rhythm

seanh:
Before:

Screenshot from 2019-07-09 18-02-37

After:

Screenshot from 2019-07-09 18-00-17


robertknight:
On my system the buttons are slightly misaligned on this branch. I'm just figuring out why:

Button alignment

seanh:
You need https://github.com/hypothesis/lms/pull/783

robertknight:
Ah, thanks. The too-close spacing between text and form controls here also affects the URL picker dialog. I think we should seek to standardise that in future. This is a clear improvement in this context though.

seanh:
For the URL picker I abandoned https://github.com/hypothesis/lms/pull/785

PR #785, opened 2019-07-09
Fix vertical rhythm of URL form

seanh:
Before:

Screenshot from 2019-07-09 18-11-18

After:

Screenshot from 2019-07-09 18-12-11

It seems like we might want a separate vertical rhythm CSS fle in which elements like titles, paragraphs, divs that contain buttons, forms, etc all have their top and bottom margins set to the same values (or multiples thereof). So far I'm just making individual fixes to components that I find problems with.


seanh:
Bah, I'm not sure that this is actually an improvement

PR #786, opened 2019-07-09
Fix ragged right edges of document source buttons

seanh:
Before:

Screenshot from 2019-07-09 18-44-47

After:

Screenshot from 2019-07-09 18-44-13


seanh:
It _does_ seem to work in Firefox for me, but I've changed it to the `inline-flex` version anyway, if that has older browser support.

PR #787, opened 2019-07-09
Update to Alpine Linux v3.10.0 

dmfine:
while keeping python3.6 from previous Alpine release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,12 @@ RUN yarn install --frozen-lockfile
 RUN yarn build
 
 # Stage 2: Build the rest of the app using build output from Stage 1.
-FROM alpine:3.9.4
+FROM alpine:3.10.0
 MAINTAINER Hypothes.is Project and contributors
 
 # Install system build and runtime dependencies.
-RUN apk add ca-certificates python3 libpq collectd collectd-disk supervisor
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,11 +10,12 @@ RUN yarn install --frozen-lockfile
...
 
 # Stage 2: Build the rest of the app using build output from Stage 1.
-FROM alpine:3.9.4
+FROM alpine:3.10.0
 MAINTAINER Hypothes.is Project and contributors
 
 # Install system build and runtime dependencies.
-RUN apk add ca-certificates python3 libpq collectd collectd-disk supervisor
+RUN apk add ca-certificates libpq collectd collectd-disk supervisor \
+  && apk add --repository http://dl-cdn.alpinelinux.org/alpine/v3.9/main/ 'python3<3.7'
```
hmstepanek: This feels dangerous to me.  It looks like this is needed because alpine 3.10 points to python3.7 and doesn't have python3.6 so we have to grab python3.6 from an older version of alpine but I wonder if that has any other consequences like some sys calls might not be available or the byte code may not be correct? This might be a good opportunity to use the python3.6 alpine base image instead. I'd like to get @seanh  's thoughts on this.

dmfine:
Going to implement this with a prebuilt image instead

PR #788, opened 2019-07-10
Don't disable all feature flags in cookie

seanh:
The way the feature flags mechanism works is that it consults a list of
feature flag "providers" in this order:

- Config file
- Environment variable
- Cookie
- Query string

When queried about a particular feature flag each provider can return
None, True or False, and the *last* provider to return either True or
False is used. So for example the feature flags cookie can return True
to enable a feature that is disabled in the config file, because the
cookie is consulted after the config file.

Unfortunately the cookie provider was _always_ returning True or False,
rather than None, even for flags that the cookie has nothing to say
about because they aren't listed in FEATURE_FLAGS_ALLOWED_IN_COOKIE.
This resulted in the cookie always having the last say (other than the
query string) and always disabling any flags that had been enabled in
the config file or envvars.

Fix the cookie provider to return None rather than False if a cookie
isn't listed in FEATURE_FLAGS_ALLOWED_IN_COOKIE, so that it doesn't
override earlier providers' values for these flags.

### DIFF
--- a/lms/extensions/feature_flags/_helpers.py
+++ b/lms/extensions/feature_flags/_helpers.py
@@ -25,7 +25,7 @@ class FeatureFlagsCookieHelper:
         self._jwt_cookie_helper.set(response, flags)
 
     def get(self, flag):
-        return self.get_all().get(flag, False)
+        return self.get_all().get(flag, None)
 
     def get_all(self):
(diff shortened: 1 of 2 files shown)


seanh:
Longer-term fix for the wider problem with the ff cookie: https://github.com/hypothesis/lms/issues/790

PR #817, opened 2019-07-16
Clean up application_instance.py

seanh:


lms/models/application_instance.py
```diff
@@ -49,36 +39,47 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    @classmethod
+    def build_from_lms_url(  # pylint:disable=too-many-arguments
+        cls, lms_url, email, developer_key, developer_secret, encryption_key=None
+    ):
```
robertknight: I'm not familiar with idioms in Python regarding static class methods vs standalone factories for creating class instances but this seems reasonable.

lms/models/application_instance.py
```diff
@@ -49,36 +39,47 @@ class ApplicationInstance(BASE):
...
+
+
+def _build_shared_secret():
     """Generate a shared secret."""
     return secrets.token_hex(32)
 
 
-def build_unique_key():
+def _build_unique_key():
     """Use the key base to generate lms key."""
```
robertknight: This docstring doesn't mean a lot to me. What is the "lms key"? What is the "key base"? I think it would be clearer just to inline this code in the one place where it is used.

lms/models/application_instance.py
```diff
@@ -49,36 +39,47 @@ class ApplicationInstance(BASE):
...
-def build_shared_secret():
+
+def _build_aes_iv():
+    """Build a 16 byte initialization vector."""
+    return Random.new().read(AES.block_size)
+
+
+def _build_shared_secret():
     """Generate a shared secret."""
     return secrets.token_hex(32)
```
robertknight: I don't see any value in this helper. Consider inlining it where it is used.

lms/models/application_instance.py
```diff
@@ -49,36 +39,47 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    @classmethod
+    def build_from_lms_url(  # pylint:disable=too-many-arguments
```
robertknight: `build_from_lms_url` is a bit misleading because there are several non-optional parameters here. Perhaps this ought to be named `build` or `build_from_config`?

lms/models/application_instance.py
seanh: I'm not sure that there's one agreed-upon approach, people use both. Personally I think code like this:

```python
from lms.models import build_from_lms_url

ai = build_from_lms_url(...)
```

Is a lot clearer now that its been rewritten like this:

```python
from lms.models import ApplicationInstance

ai = ApplicationInstance.build_from_lms_url(...)
```

You can actually tell what type of object is being created.

Plus I just like grouping and namespacing things -- the `lms.models` package now exports a single name `ApplicationInstance` instead of exporting both `ApplicationInstance` and `build_from_lms_url` as top-level names.

Of course our pattern at Hypothesis is to have thin models, and not put methods like these in the model code at all, but instead have an `ApplicationInstanceBuilder` service. Because that scales better (a single model class can have many associated services, rather than the one model class growing too big); a service can handle stuff involving coordinating multiple model classes; keeping the model objects (which tend to pop up all over the codebase) thin and simple prevents them from coupling our code together too much; and finally model classes are already complicated enough just dealing with the SQLAlchemy field types and constraints and optimisations etc.

But, in this case I haven't bothered to extract this out into a service yet.

seanh:
Rebased

PR #819, opened 2019-07-16
Specify public API of oauth2_token.py

seanh:


lms/models/oauth2_token.py
```diff
@@ -6,6 +6,9 @@
 from lms.db import BASE
 
 
+__all__ = ["OAuth2Token"]
```
robertknight: What are modern best practices wrt. when to use `__all__` and when to avoid it?

seanh: Every module "should" have an `__all__` declaring its public interface, even if it's an empty list: https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces

Of the three ways that Python has for declaring a public interface (docstrings, leading underscores, and `__all__`) I think `__all__` is the most explicit and unambiguous. But we should still use leading underscores as well (rather than always having to look up and check the `__all__`), the two complement each other -- `__all__` providing the explicit single source of truth and being a nice summary in one place at the top of the file, underscores providing hints to the reader throughout the code.

Some of the benefits of `__all__`:

Although a leading underscore usually means private, you can actually add something that does have a leading underscore to the module's public interface by adding it to the `__all__`. Documenting something (e.g. in a module, class or function's docstring) is supposed to mean it's part of the public interface too, even if it has a leading underscore. You might want to do this if:

* Something was originally private, but got used anyway, so now you're making it officially part of the interface, but don't want to rename it
* You have some other reason to name something with a leading underscore, other than indicating that it's private. For example I've seen this done in the standard library to avoid name conflicts

Conversely something that _does not_ have a leading underscore can be made _not_ part of the interface by not putting it into the `__all__`. For example sometimes a module import's a name from another module and makes that name part of its own public interface by adding it to `__all__`, while other imported names are not added to the interface.

We frequently do this in package's `__init__.py` files to make the public interface of the package be a flat list of names, and the internal structuring of the package into different files and directories an internal details. For example here's `lms/validation/__init__.py`:

```python
from lms.validation._exceptions import (
    ValidationError,
    ExpiredSessionTokenError,
    MissingSessionTokenError,
    InvalidSessionTokenError,
    MissingStateParamError,
    ExpiredStateParamError,
    InvalidStateParamError,
)
from lms.validation._oauth import (
    CanvasOAuthCallbackSchema,
    CanvasAccessTokenResponseSchema,
    CanvasRefreshTokenResponseSchema,
)
from lms.validation._launch_params import LaunchParamsSchema
from lms.validation._bearer_token import BearerTokenSchema
from lms.validation._module_item_configuration import ConfigureModuleItemSchema
from lms.validation._canvas import (
    CanvasListFilesResponseSchema,
    CanvasPublicURLResponseSchema,
)


__all__ = (
    "CanvasOAuthCallbackSchema",
    "CanvasAccessTokenResponseSchema",
    "CanvasRefreshTokenResponseSchema",
    "BearerTokenSchema",
    "LaunchParamsSchema",
    "ConfigureModuleItemSchema",
    "CanvasListFilesResponseSchema",
    "CanvasPublicURLResponseSchema",
    "ValidationError",
    "ExpiredSessionTokenError",
    "MissingSessionTokenError",
    "InvalidSessionTokenError",
    "MissingStateParamError",
    "InvalidStateParamError",
    "ExpiredStateParamError",
)
```

PR #824, opened 2019-07-16
Remove support for reading launch params from DB

seanh:
This is no longer used now that the new_oauth feature flag is turned on for everyone.

### DIFF
--- a/lms/resources/_lti_launch.py
+++ b/lms/resources/_lti_launch.py
@@ -7,7 +7,6 @@ import jwt
 from pyramid.httpexceptions import HTTPBadRequest
 from pyramid.security import Allow
 
-from lms.util import lti_params_for
 from lms.validation import BearerTokenSchema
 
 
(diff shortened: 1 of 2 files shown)


seanh:
Previously when the app received a launch request (either to create or view an assignment), if it was a Canvas assignment (or if you were creating a new assignment and has the Canvas files option available), then our server would return a redirect, redirecting the iframe to Canvas's authorization page. When the user clicked <kbd>Authorize</kbd> and Canvas then redirected the iframe back to our redirect_uri, we would _then_ launch the assignment or show the file picker UI, in response to this OAuth redirect request rather than in response to the original launch request.

This meant that the code for showing assignments or the file picker could not assume that `request.params` held the LTI launch params, as it might hold OAuth redirect params instead, in which case the launch params had to instead be retrieved from the DB where they had been previously stashed.

Now that we have a JavaScript SPA that does authorization in a popup window this whole nightmare is over and we no longer sometimes render the file picker or assignment view at `https://lms.hypothes.is/canvas_oauth_callback` rather than at `/lti_launches`

PR #826, opened 2019-07-16
Move via_url() to view-specific helpers

seanh:
Tests will be failing because legacy views need to be deleted


seanh:
Moved into https://github.com/hypothesis/lms/pull/828

PR #827, opened 2019-07-16
Add migration to delete no-longer-used DB tables

seanh:


robertknight:
Rebased

robertknight:
I think this is good to deploy now @seanh ?

PR #828, opened 2019-07-16
Delete unused legacy views and their associated code

seanh:
What this PR does:

* Deletes a bunch of legacy views that're no longer used, now that the new_oauth feature flag is on for everyone on QA and production: `views/canvas_proxy.py`, `views/decorators/legacy_h_api.py`, `views/legacy_content_item_selection.py`, `views/lti_launches.py`, `views/module_item_configurations.py` and `views/oauth.py ` are now gone.

Deleting these views is the key that unlocks a waterfall of enabling us to delete a lot of other stuff that is no longer used because it was only being called by these views, or by other stuff that was being called by the views and is now deleted. All of these are now unused and have been deleted:

* The templates the belonged to the deleted views.
* The legacy JavaScript code and its associated assets etc, that was used by those templates
* All of `lms/util/` except for the `via_url()` helper function which has been moved into `lms/views/helpers/`
* The `OauthState`, `Token` and `User` models and their associated helper functions (the DB migrations to delete the tables are in https://github.com/hypothesis/lms/pull/827)
* All of `lms/exceptions.py`
* A bunch of global  text fixtures from `conftest.py`
* All of `tests/lms/factories.py`


lms/assets.ini
```diff
@@ -1,14 +1,8 @@
...
 
-file_picker_js =
-  scripts/file_picker.bundle.js
-
 frontend_apps_js =
   scripts/frontend_apps.bundle.js
 
-content_item_selection_js =
-  scripts/content_item_selection.bundle.js
-
```
robertknight: There are references to these bundles in `gulpfile.js` which should also be removed.

lms/assets.ini
```diff
@@ -1,14 +1,8 @@
 [bundles]
 
-file_picker_js =
-  scripts/file_picker.bundle.js
```
robertknight: When these dependencies are removed, the following frontend dependencies can also be removed:

- lodash (replaced by JS standard lib APIs)
- jquery (replaced by Preact + DOM APIs)
- superagent (replaced by whatwg-fetch)

I can do that in a follow up PR as some small changes will also be needed in the gulpfile

lms/assets.ini
seanh: Done, thanks

lms/assets.ini
seanh: Cool, thanks.

Now that I think of it we'll probably have some unused backend deps now too

seanh:
Rebased, and deleted some unused test fixtures and factories

seanh:
@robertknight I've deleted that gulpfile stuff, and also some unused test fixtures and factories, this is ready for review now

PR #831, opened 2019-07-17
Delete a useless assertion

seanh:


tests/lms/services/exceptions_test.py
```diff
@@ -200,5 +200,3 @@ def _requests_exception(**kwargs):
             response.raise_for_status()
         except requests.RequestException as err:
             return err
-
-        assert False, "We should never get here"
```
robertknight: It looks like the original purpose of this was to provide a more obvious error if for some reason the code above failed to generate the expected exception. Is that not useful now?

seanh: The line is never reached, so this just messes up the coverage report!

robertknight: Can't you tell the code coverage generator to ignore that line?

seanh: Ah, yes let's do that instead. Changed

PR #848, opened 2019-07-25
Use docker for dev

hmstepanek:
- Add a dockerfile for starting up lms with development settings
- Add a make docker-dev command that builds the lms docker image from
local code as opposed to git HEAD.
- Add the python code as a volume for local development so that the code can be rebuilt without restarting/building the container.
- Add a make run-docker-test command for testing python code with docker.

To run lms dev using docker:
1. Configure environment variables.
2. Run make services
3. Run make docker-dev
4. Run make docker-run-dev.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -1,0 +1,57 @@
+# Stage 1: Build frontend assets.
+FROM node:12.6-alpine as frontend-build
+
+ENV NODE_ENV production
+COPY .babelrc gulpfile.js package.json yarn.lock ./
+COPY scripts/gulp ./scripts/gulp
+COPY lms/static ./lms/static
+RUN yarn install --frozen-lockfile
+
+RUN yarn build
+
+# Stage 2: Build the rest of the app using build output from Stage 1.
+FROM python:3.6.9-alpine3.10
+MAINTAINER Hypothes.is Project and contributors
+
+# Install system build and runtime dependencies.
+RUN apk add libpq collectd collectd-disk supervisor
+
+# Create the lms user, group, home directory and package directory.
+RUN addgroup -S lms \
+  && adduser -S -G lms -h /var/lib/lms lms
+WORKDIR /var/lib/lms
+
+# Copy minimal data to allow installation of dependencies.
+COPY requirements.txt ./
+
+# Install build deps, build, and then clean up.
+RUN apk add --virtual build-deps \
+    build-base \
+    postgresql-dev \
+    python3-dev \
+  && pip3 install --no-cache-dir -U pip \
+  && pip3 install --no-cache-dir -r requirements.txt \
+  && pip3 install --no-cache-dir pip-tools pyramid_debugtoolbar \
+  pyramid_ipython coverage pytest factory-boy httpretty codecov \
+  prospector==1.1.6.2 black sphinx sphinx-autobuild \
+  && apk del build-deps
+
+# Copy collectd config
+COPY conf/collectd.conf /etc/collectd/collectd.conf
+RUN mkdir /etc/collectd/collectd.conf.d \
+  && chown lms:lms /etc/collectd/collectd.conf.d
+
+# collectd always tries to write to this immediately after enabling the logfile plugin.
+# Even though we later configure it to write to stdout. So we do have to make sure it's
+# writeable.
+RUN touch /var/log/collectd.log && chown lms:lms /var/log/collectd.log
+
+# Copy frontend assets.
+COPY --from=frontend-build /build build
+
+# Copy config files.
+COPY ./conf ./conf
+
+EXPOSE 8001
+USER lms
+CMD ["bin/init-env", "supervisord", "-c", "conf/supervisord.conf"]
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,11 @@ help:
 	@echo "make checkdocstrings   Crash if building the docstrings fails"
 	@echo "make pip-compile       Compile requirements.in to requirements.txt"
 	@echo "make docker            Make the app's Docker image"
+	@echo "make docker-dev        Make the app's Docker image applying the dev configuration"
 	@echo "make run-docker        Run the app's Docker image locally"
+	@echo "make run-docker-dev    Run the app's Docker image locally and interactively"
+	@echo "make run-docker-test   Run the unit tests via the app's Docker image locally"
+	@echo "make test-db           Make a test database
 	@echo "make clean             Delete development artefacts (cached files, "
 	@echo "                       dependencies, etc)"
 
@@ -37,7 +41,7 @@ shell:
 
 .PHONY: sql
 sql:
-	docker-compose exec postgres psql --pset expanded=auto -U postgres
+	docker-compose exec lmspostgres psql --pset expanded=auto -U postgres
 
 .PHONY: lint
 lint: backend-lint frontend-lint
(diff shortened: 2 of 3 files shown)


conf/supervisord-dev.conf
```diff
@@ -0,0 +1,27 @@
+[supervisord]
+nodaemon=true
+environment=PYTHONUNBUFFERED="1"
+logfile=/dev/null
+logfile_maxbytes=0
+
+[program:web]
+command=newrelic-admin run-program gunicorn --paste conf/development.ini
```
hmstepanek: This is really the only difference between prod and dev. Maybe this could be specified via an environment variable?

Makefile
```diff
@@ -85,7 +90,9 @@ run-docker:
 	docker run \
 		--net lms_default \
 		-e DATABASE_URL=postgresql://postgres@postgres/postgres \
-		-e FEATURE_FLAGS_COOKIE_SECRET \
+		-e GOOGLE_CLIENT_ID \
```
hmstepanek: These were just missing from this command.

Makefile
```diff
@@ -77,6 +78,10 @@ pip-compile:
 docker:
 	git archive --format=tar.gz HEAD | docker build -t hypothesis/lms:$(DOCKER_TAG) -
 
+.PHONY: docker
+docker-dev:
```
hmstepanek: This points to the local code as opposed to pointing to git HEAD. Pointing at git HEAD would require the code to be committed before testing which of course is a pain when developing code locally so here we don't require that.

Dockerfile-dev
```diff
@@ -0,0 +1,55 @@
...
+COPY requirements.txt ./
+
+# Install build deps, build, and then clean up.
+RUN apk add --virtual build-deps \
+    build-base \
+    postgresql-dev \
+    python3-dev \
+  && pip3 install --no-cache-dir -U pip \
+  && pip3 install --no-cache-dir -r requirements.txt \
+  && pip3 install --no-cache-dir pip-tools pyramid_debugtoolbar pyramid_ipython \
```
hmstepanek: This is the only difference between dev and prod. We could simplify this by just doing FROM our custom image and then install these after maybe but then the COPY .. would happen before the install so I think it's prob simpler and more efficient to just have 2 Dockerfiles.

Makefile
```diff
@@ -85,7 +91,35 @@ run-docker:
...
+		-e JWT_SECRET \
+		-e LMS_SECRET \
+		-e RPC_ALLOWED_ORIGINS \
+		-e VIA_URL \
+		-e SESSION_COOKIE_SECRET \
+		-e OAUTH2_STATE_SECRET \
+		-p 8001:8001 \
+		hypothesis/lms:$(DOCKER_TAG)
+.PHONY: run-docker-dev
+run-docker-dev:
```
hmstepanek: In order to use docker for dev with all other services running also in dev they all need to be on the same network (I chose to use h_default which is used in the h repo as the network for everything).

Dockerfile-dev
```diff
@@ -0,0 +1,57 @@
...
+COPY requirements.txt ./
+
+# Install build deps, build, and then clean up.
+RUN apk add --virtual build-deps \
+    build-base \
+    postgresql-dev \
+    python3-dev \
+  && pip3 install --no-cache-dir -U pip \
+  && pip3 install --no-cache-dir -r requirements.txt \
+  && pip3 install --no-cache-dir pip-tools pyramid_debugtoolbar \
```
hmstepanek: These are dev dependencies.

Makefile
```diff
@@ -102,6 +111,60 @@ run-docker:
...
+		-e H_JWT_CLIENT_SECRET \
+		-e JWT_SECRET \
+		-e LMS_SECRET \
+		-e RPC_ALLOWED_ORIGINS \
+		-e VIA_URL \
+		-e SESSION_COOKIE_SECRET \
+		-e OAUTH2_STATE_SECRET \
+		-p 8001:8001 \
+		hypothesis/lms:$(DOCKER_TAG) \
+	    -c "newrelic-admin run-program gunicorn --paste conf/development.ini"
```
hmstepanek: Override the default command with the command we use in make dev.

Makefile
```diff
@@ -102,6 +111,60 @@ run-docker:
 		-p 8001:8001 \
 		hypothesis/lms:$(DOCKER_TAG)
 
+.PHONY: run-docker-dev
+run-docker-dev:
+	docker run \
+		--net h_default \
+		-it \
```
hmstepanek: -it makes the docker run interact with stdin and out enabling the use of pdb.

Makefile
```diff
@@ -102,6 +111,60 @@ run-docker:
...
+		-e H_JWT_CLIENT_SECRET \
+		-e JWT_SECRET \
+		-e LMS_SECRET \
+		-e RPC_ALLOWED_ORIGINS \
+		-e VIA_URL \
+		-e SESSION_COOKIE_SECRET \
+		-e OAUTH2_STATE_SECRET \
+		-p 8001:8001 \
+		hypothesis/lms:$(DOCKER_TAG) \
+	    -c "coverage run --source lms,tests/lms -m pytest -v -Werror -p no:cacheprovider"
```
hmstepanek: This is currently broken cause imp is deprecated. It's a python3 issue.

Makefile
```diff
@@ -102,6 +113,72 @@ run-docker:
 		-p 8001:8001 \
 		hypothesis/lms:$(DOCKER_TAG)
 
+.PHONY: run-docker-dev
+run-docker-dev:
+	docker run \
+		--name lms \
+		--net h_default \
+		-v /Users/hmstepanek/Documents/work/dev/lms/lms:/var/lib/lms/lms \
```
hmstepanek: Mount the python code as a volume so changes get rebuilt without re-building the docker container.

Makefile
```diff
@@ -102,6 +113,72 @@ run-docker:
...
+		-e RPC_ALLOWED_ORIGINS \
+		-e VIA_URL=https://example.com/ \
+		-e SESSION_COOKIE_SECRET \
+		-e OAUTH2_STATE_SECRET \
+		-p 8001:8001 \
+		hypothesis/lms:$(DOCKER_TAG) \
+	    -c "pytest -v -p no:cacheprovider tests/lms"
+
+.PHONY: test-db
+test-db:
```
hmstepanek: Add the command for building the test database. I wonder if this just shouldn't be part of the test command.

docker-compose.yml
```diff
@@ -1,6 +1,10 @@
 version: '3'
 services:
-  postgres:
+  lmspostgres:
```
hmstepanek: Name it something different than the h postgres container or the names will conflict and docker dns will randomly chose one.

docker-compose.yml
```diff
@@ -1,6 +1,10 @@
...
     image: postgres:9.4-alpine
     ports:
       - '127.0.0.1:5433:5432'
+networks:
```
hmstepanek: Specify the docker sub-network the container will be available on.

seanh:
Hi Hannah,

Let's hold off on further integrating Docker into our development environment for now, and reconsider later.

There's a few things in progress with the dev env currently. We've just added pyenv support to h and are in the process of automating the installation and upgrading of the required Python versions and tox using pyenv, and of extending pyenv to the rest of our apps. I'm preparing everything to try to make Jon's onboarding as a backend developer in October as smooth as possible. We're preparing to be able to move all our apps to Python 3.7, and then 3.8, and also to change patch versions of Python and keep everyone's dev env running the right patch version, with minimal or zero disruption to developers. For a number of reasons we want the dev env to be able to support multiple versions of Python at once and running the tests in multiple versions, for example because we want to start extracting parts of our app into standalone dependencies and we'll need those to work with multiple versions. I want to make progress on better automating the creation of synchronized dev data in the h and lms DBs, and setting synchronized envvars. I want to remove some dev-prod differences around error handling and debugging. Etc etc.

That's quite a lot of stuff and I think a parallel Docker-based way of running your dev env is going to have difficulty keeping up, and maintaining both will create more work for us (when people have problems with their dev set up I have to help them fix it).

I think next year might be a good time to revisit this, once all the above developments are delivered and we've learned from onboarding a fresh backend developer. We can then consider any remaining issues with the dev set up, and whether Docker might have a role in solving them

hmstepanek:
Hi @seanh Let's talk about this when I get back! I know there are a lot of changes being made right now and that you don't want to have to support multiple setups which I totally understand. On the other-hand, I also much prefer this setup to the existing one and while I'm fine with it not being the recommended setup, I would find it helpful if it was part of master so that I don't have to keep working off of a branch. I realize this is a lot in one PR though so I'm happy to break it up into smaller pieces. I've been kinda holding off/making slight tweaks locally to it overtime to improve upon it so it's gotten a bit larger than it was initially although going into it I expected that would be the case. I think most of these changes are beneficial to the existing docker run/make setup for dev which while we rarely use, are still make commands and I think are vast improvements over what's there currently. Perhaps we can hold off on closing until we've had a chance to discuss?

seanh:
Let's wait until the features I described above are delivered, and the dust has settled, and then reconsider. I don't want to have multiple competing dev setups in git.

I think a better way to approach this would be incrementally. Identify a particular problem or pain point with our current setup, and fix it in a small and simple way using Docker. Rather than starting from scratch

Issue #806, opened 2019-07-12
Add support for Canvas API scoped developer keys


seanh:
## Problem

When creating or editing a developer key in Canvas there's an **Enforce Scopes** option that you can use to limit what "scopes" a client using this developer key has access to (basically which API endpoints the client can use).

Our app currently doesn't declare to Canvas what scopes it needs, which means that Canvas treats us as a client that requires every scope, so either an unscoped dev key or one with every scope checked will work. But if the user creates a dev key with any scopes unchecked then, when we open the Canvas authorization popup window, the user sees this error message from Canvas:

![error-when-scope-is-wrong](https://user-images.githubusercontent.com/22498/61131925-3f9e8c80-a4b2-11e9-87d0-a2cafa9d218a.png)

## We don't want to encourage scoped dev keys

We absolutely **do not want to encourage** people to use scoped developer keys with our app. Using a scoped developer key means that users might deny us a scope that, while we don't need it now, we might need that additional scope in the future, and when we add code that uses that scope our users will be faced with the unfriendly error message from Canvas shown above.

Canvas does not provide any way for our app to request additional scopes, it just runs users into the above error message.

Until we've found a way to implement good error handling for when the developer key is missing a scope that we need, we shouldn't encourage scoped developer keys.

Having said that, some users insist on using scoped keys, so our app should support them.

## Creating a scoped developer key

To create a scoped developer key you need to:

1. Login to Canvas as an admin
2. Go to **Admin -> University of Hypothesis -> Developer keys**
3. Either **create a new developer key** or **edit an existing one**
4. Turn on the **Enforce Scopes** option
5. Check the scopes that you want to allow
6. Click **Save**

Any application installs that use this developer key will now be scoped.

![Peek 2019-07-12 14-29](https://user-images.githubusercontent.com/22498/61131730-e20a4000-a4b1-11e9-889e-ce770be5762a.gif)

## Scopes our app needs

Our app currently uses two Canvas API endpoints:

1. [List a course's files](https://canvas.instructure.com/doc/api/files.html#method.files.api_index), which requires the `"url:GET|/api/v1/courses/:course_id/files"` scope
2. [Get a download URL for a file](https://canvas.instructure.com/doc/api/files.html#method.files.public_url), requires the `"url:GET|/api/v1/files/:id/public_url"`

## How to add scopes support to our app

[...]

PR #854, opened 2019-07-30
Include scope in canvas auth request

hmstepanek:
This is a partial fix for https://github.com/hypothesis/lms/issues/806.

### DIFF
--- a/lms/views/api/canvas/authorize.py
+++ b/lms/views/api/canvas/authorize.py
@@ -28,6 +28,7 @@ class CanvasAPIAuthorizeViews:
                         "response_type": "code",
                         "redirect_uri": self.request.route_url("canvas_oauth_callback"),
                         "state": CanvasOAuthCallbackSchema(self.request).state_param(),
+                        "scope": "url:GET|/api/v1/courses/:course_id/files url:GET|/api/v1/files/:id/public_url",
                     }
                 ),
                 "",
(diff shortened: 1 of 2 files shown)


seanh:
This looks good to me and seems to work locally. The only thing I want to think about is whether we should add any more scopes. You suggested a couple (https://github.com/hypothesis/lms/issues/806#issuecomment-516091760). I want to have a look over them...

hmstepanek:
Yes, I can do that. I think merging this with just these two scopes is good for now as well since as you said it’s hard to anticipate what scopes we will need in the future.

lms/views/api/canvas/authorize.py
```diff
@@ -28,6 +28,7 @@ def authorize(self):
                         "response_type": "code",
                         "redirect_uri": self.request.route_url("canvas_oauth_callback"),
                         "state": CanvasOAuthCallbackSchema(self.request).state_param(),
+                        "scope": "url:GET|/api/v1/courses/:course_id/files url:GET|/api/v1/files/:id/public_url",
```
robertknight: If we end up with many scopes in future I think it would be good to note what the various scopes are for.

robertknight:
> I think merging this with just these two scopes is good for now as well since as you said it’s hard to anticipate what scopes we will need in the future.

Yes, I mainly wanted to flag up that when we do add new scopes, we'll need to consider existing scoped dev keys that exist.

In this first PR, we're not adding any new scopes since all authorizations currently implicitly require all scopes.

PR #856, opened 2019-08-02
Use frontend SPA for all LTI launches

robertknight:
In preparation for adding logic to the frontend app which will need to
be triggered for all launch types, not just Canvas PDFs, make all LTI
launches use the `BasicLtiLaunchApp` frontend mini-app.

 - Rename the `via_url` frontend config param to `via_url_callback` to
   indicate that it is a URL that the frontend will need to invoke in
   order to get the Via URL, not the Via URL itself

 - Add a new `via_url` frontend config param which is the actual Via
   URL, and is configured by the backend for launches where it can be
   determined in the initial LTI launch

 - Make the frontend render the `via_url` URL in an iframe if provided,
   or fetch the `via_url_callback` URL otherwise

 - Move the content of `canvas_file_basic_lti_launch.html.jinja2` into
   the `basic_lti_launch.html.jinja2` template to make the frontend app
   load for all launches


robertknight:
Gah, forgot to run `make lint`.

PR #857, opened 2019-08-02
SpeedGrader integration V1 (integration branch)

robertknight:
This is an integration branch that has the initial version of Canvas SpeedGrader support, for testing the feature end-to-end as it is being worked on.


robertknight:
Separate PRs have now been opened for the individual pieces.

PR #865, opened 2019-08-05
SpeedGrader 2/n - Add an API endpoint for frontend to trigger submitting of grading LTI launch URL to Canvas

robertknight:
**Depends on https://github.com/hypothesis/lms/pull/864**

Add an API endpoint to record a grading LTI launch URL in SpeedGrader

Add an API endpoint that the frontend will use to record a
grading-oriented LTI launch URL in Canvas's SpeedGrader for use when a teacher
grades the current user's annotations on an assignment.

This operation will be triggered by the frontend rather than by the
backend during the /lti_launches request, where all the necessary
information is already available, because it involves making (possibly
slow) calls to the external LMS. Instead these potentially slow
operations can happen concurrently with the assignment loading in the
user's browser.

 - Add `/api/lti/submissions` endpoint to record SpeedGrader Launch URL
 - Add schema to validate JSON body of requests to this endpoint

### DIFF
--- a/lms/validation/_api.py
+++ b/lms/validation/_api.py
@@ -1,0 +1,40 @@
+"""Schema for JSON APIs exposed to the frontend."""
+
+from webargs import fields
+
+from lms.validation._helpers import PyramidRequestSchema
+
+
+__all__ = ["ApiRecordSubmissionSchema"]
+
+
+class ApiRecordSubmissionSchema(PyramidRequestSchema):
+    """Schema for validating requests from the frontend to record submissions."""
+
+    locations = ["json"]
+
+    document_url = fields.Str()
+    """
+    URL of the document for this assignment.
+    """
+
+    canvas_file_id = fields.Str()
+    """
+    ID of the Canvas file for this assignment.
+    """
+
+    h_username = fields.Str(required=True)
+    """
+    h username generated for the active user.
+    """
+
+    lis_outcome_service_url = fields.Str(required=True)
+    """
+    URL provided by the LMS to submit grades or other results to.
+    """
+
+    lis_result_sourcedid = fields.Str(required=True)
+    """
+    Opaque identifier provided by the LMS to identify a submission. This
+    typically encodes the assignment context and LMS user.
+    """
--- a/lms/validation/_api.py
+++ b/lms/validation/_api.py
@@ -5,33 +5,25 @@ from webargs import fields
 from lms.validation._helpers import PyramidRequestSchema
 
 
-__all__ = ["ApiRecordSubmissionSchema"]
+__all__ = ["APIRecordSubmissionSchema"]
 
 
-class ApiRecordSubmissionSchema(PyramidRequestSchema):
+class APIRecordSubmissionSchema(PyramidRequestSchema):
     """Schema for validating requests from the frontend to record submissions."""
 
     locations = ["json"]
 
     document_url = fields.Str()
-    """
-    URL of the document for this assignment.
-    """
+    """URL of the document for this assignment."""
 
     canvas_file_id = fields.Str()
-    """
-    ID of the Canvas file for this assignment.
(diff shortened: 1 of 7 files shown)


lms/views/api/lti.py
```diff
@@ -0,0 +1,75 @@
...
+    Currently this view only supports SpeedGrader-based grading in Canvas by
+    submitting an LTI Launch URL to Canvas. In future it will need to persist the
+    metadata to facilitate grading in other LMSes via a different UI.
+
+    This work _could_ be done by the backend during an LTI launch, but as it
+    involves potentially slow requests to the external LMS, it is triggered
+    asynchronously by the frontend while the assignment is loading.
+
+    [1] https://canvas.instructure.com/doc/api/file.assignment_tools.html
+    """
```
seanh: :+1:

lms/views/api/lti.py
```diff
@@ -0,0 +1,75 @@
...
+
+    shared_secret = request.find_service(name="ai_getter").shared_secret(
+        lti_user.oauth_consumer_key
+    )
+    outcome_request_params = LTIOutcomesRequestParams(
+        consumer_key=lti_user.oauth_consumer_key,
+        shared_secret=shared_secret,
+        lis_outcome_service_url=parsed_params["lis_outcome_service_url"],
+        lis_result_sourcedid=parsed_params["lis_result_sourcedid"],
+    )
```
seanh: **Code design:** I think there's a factory function for constructing an `LTIOutcomesRequestParams` here, which could move a bunch of code out of this view. The suggestion below presents a challenge to the pattern that our value classes have no methods, so we might need to think about that (maybe the factory function belongs elsewhere).

```python
class LTIOutcomesRequestParams:
    ...

    @classmethod
    def from_request(cls, request):
        ...
        return cls(...)
```

tests/lms/validation/_api_test.py
```diff
@@ -0,0 +1,50 @@
+import json
+
+import pytest
+
+from lms.validation import ValidationError
+from lms.validation._api import ApiRecordSubmissionSchema
+
+
+class TestApiRecordSubmissionSchema:
```
seanh: **Style:** `TestAPIRecordSubmissionSchema`

tests/lms/views/api/lti_test.py
```diff
@@ -0,0 +1,111 @@
...
+    return svc
+
+
+@pytest.fixture(autouse=True)
+def ai_getter(pyramid_config):
+    svc = mock.create_autospec(ApplicationInstanceGetter, instance=True, spec_set=True)
+
+    def shared_secret(consumer_key):
+        if consumer_key != "TEST_OAUTH_CONSUMER_KEY":
+            raise Exception("Incorrect consumer key")
```
seanh: This line of code is never used...

lms/validation/_api.py
```diff
@@ -0,0 +1,40 @@
...
+
+    canvas_file_id = fields.Str()
+    """
+    ID of the Canvas file for this assignment.
+    """
+
+    h_username = fields.Str(required=True)
+    """
+    h username generated for the active user.
+    """
```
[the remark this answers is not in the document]
robertknight: > Or alternatively the h_username property could be moved out of LTILaunchResource and into some other class, since it's now also needed outside of LTI launches.

I think that would be better, as it seems to me that things could be added to `LTILaunchResource` in future which are appropriate for LTI launches but not for other contexts.

robertknight:
Issues to follow up later:

- [ ] Derive `h_username` from `request.lti_user` in the API call rather than taking it as a parameter supplied by the frontend code
- [ ] Extract SpeedGrader URL construction into a service or helper (see [comment](https://github.com/hypothesis/lms/pull/865#discussion_r311088485))

PR #866, opened 2019-08-06
Add fallback error handler for `/api/*` routes

robertknight:
Add a fallback error handler for `/api/*` routes so that unhandled
exceptions result in JSON rather than HTML responses, to be consistent
with the frontend's expectations.

### DIFF
--- a/tests/lms/views/api/error_test.py
+++ b/tests/lms/views/api/error_test.py
@@ -48,7 +48,7 @@ class TestForbidden:
         assert result["message"] == "You're not authorized to view this page"
 
 
-class TestApiError:
+class TestAPIError:
     def test_it(self, pyramid_request):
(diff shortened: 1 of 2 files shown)


tests/lms/views/api/error_test.py
```diff
@@ -46,3 +46,13 @@ def test_it_shows_a_generic_error_message_to_the_user(self, pyramid_request):
         result = error.forbidden(pyramid_request)
 
         assert result["message"] == "You're not authorized to view this page"
+
+
+class TestApiError:
```
seanh:
```suggestion
class TestAPIError:
```

robertknight:
I'm pretty sure renaming isn't going to fail CI, so I'll merge now. YOLO.

PR #867, opened 2019-08-06
SpeedGrader 3/n - Call API view to record a submission when a student starts an assignment

robertknight:
**Depends on https://github.com/hypothesis/lms/pull/865**

Call `record_submission` API from frontend during LTI launch

When an assignment is launched by a student/learner, call the
`record_submission` API view provided by the backend to register a
grading LTI launch URL for the current (user, assignment) in Canvas.

This behavior is currently only enabled when the student has the "speedgrader"
feature flag enabled and when the LMS is Canvas.

 - Support making POST requests to API with JSON bodies from frontend

 - Pass params for call to `record_submission` API call to frontend in
   JS config under `submissionParams` key

 - If `submissionParams` key is present in JS config, call
   `record_submission` API from frontend during an LTI launch. This
   happens after the content URL is fetched, but concurrently with the
   content iframe loading.

 - If the frontend call to record a submission fails, replace the iframe
   content with an error, to prevent confusion where the student does
   the work but SpeedGrader does not show a submission.

   The initial error view is currently very basic, it just asks the user
   to reload the page.

 - Tweak the wording of the error display to make it more obvious that
   it comes from Hypothesis and also to make it more generic, as there
   are now different error conditions that can cause it to be displayed.

---

**Testing**

1. Create an assignment using Hypothesis when logged into Canvas as a teacher
2. Launch the assignment when logged in as a teacher - no call to `/api/lti/submissions` should be made
3. Launch the assignment when logged in as a student and the "speedgrader" flag off - no call to /api/lti/submissions` should be made
4. Launch the assignment when logged in as a student and the "speedgrader" flag is on - a call to `POST /api/lti/submissions` should be made with `h_username`, `lis_result_sourcedid` and `lis_outcome_service_url` params in the JSON request
5. Modify the `record_submission` API view to make it throw an exception (`raise Exception("failed")`) will do. When the assignment is launched, it may show the iframe briefly but will then show an error dialog. This is not a very polished error in this PR.
6. Repeat steps (1) and (3) in a non-Canvas LMS, no call to `/api/lti/submissions` should be made

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -47,6 +47,27 @@ class BasicLTILaunchViews:
         # Configure the front-end mini-app to run.
         self.context.js_config.update({"mode": "basic-lti-launch"})
 
+        # Add config used by frontend to call `record_submission` API.
+        params = self.request.params
+        if (
+            # The outcome reporting params are typically only available when
+            # students (not teachers) launch an assignment.
+            params.get("lis_result_sourcedid")
+            and params.get("lis_outcome_service_url")
+            # This feature is initially restricted to Canvas.
+            and params.get("tool_consumer_info_product_family_code") == "canvas"
+            and request.feature("speedgrader")
+        ):
+            self.context.js_config["submissionParams"] = {}
+
+            self._set_submission_param("h_username", context.h_username)
+            self._set_submission_param(
+                "lis_result_sourcedid", params.get("lis_result_sourcedid")
+            )
+            self._set_submission_param(
+                "lis_outcome_service_url", params.get("lis_outcome_service_url")
+            )
+
     @view_config(canvas_file=True)
     def canvas_file_basic_lti_launch(self):
         """
@@ -80,6 +101,8 @@ class BasicLTILaunchViews:
                 )
             }
         )
+
+        self._set_submission_param("canvas_file_id", file_id)
 
         return {}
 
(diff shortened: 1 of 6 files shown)


robertknight:
WIP - This is currently failing if a teacher launches an assignment with the "speedgrader" flag on.

robertknight:
> WIP - This is currently failing if a teacher launches an assignment with the "speedgrader" flag on.

This is resolved, there is a separate issue whereby the PDF fails to load in SpeedGrader if it comes from Google Drive. I'll file that separately.

lms/views/basic_lti_launch.py
```diff
@@ -47,6 +47,27 @@ def __init__(self, context, request):
...
+        ):
+            self.context.js_config["submissionParams"] = {}
+
+            self._set_submission_param("h_username", context.h_username)
+            self._set_submission_param(
+                "lis_result_sourcedid", params.get("lis_result_sourcedid")
+            )
+            self._set_submission_param(
+                "lis_outcome_service_url", params.get("lis_outcome_service_url")
+            )
```
seanh: These 20 lines of code feel to me like a candidate for extraction into a separate method with its own unit tests

robertknight: I'd entertain suggestions of what that might look like, but I'm honestly not sure that I see the value here.

PR #868, opened 2019-08-06
SpeedGrader 4/n - Basic implementation of focusing on a user in SpeedGrader

robertknight:
**Depends on https://github.com/hypothesis/lms/pull/865 and https://github.com/hypothesis/lms/pull/867**

This PR implements focusing on the target student in SpeedGrader in the most basic possible way, by pre-filling the search bar with `user:<username>` where `<username>` is the auto-generated hex digit string that serves as the username for provisioned users in the LMS app.

- Make `context.hypothesis_config` mutable in LTI launch views, so that custom configuration can be passed to the client in grading launches
- Respond to the `focused_user` LTI launch param added in https://github.com/hypothesis/lms/pull/865 by adding `{"query":"user:<username>"}` to the Hypothesis client's configuration

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -68,6 +68,12 @@ class BasicLTILaunchViews:
                 "lis_outcome_service_url", params.get("lis_outcome_service_url")
             )
 
+        # If the launch has been configured to focus on the annotations from
+        # a particular user, translate that into Hypothesis client configuration.
+        focused_user = self.request.params.get("focused_user")
+        if focused_user:
+            self.context.hypothesis_config.update({"query": f"user:{focused_user}"})
+
     @view_config(canvas_file=True)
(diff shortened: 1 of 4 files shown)


robertknight:
There is a known issue at present where the SpeedGrader view does not work if the PDF comes from Google Drive. I will file this as a separate issue.

lms/views/basic_lti_launch.py
```diff
@@ -68,6 +68,12 @@ def __init__(self, context, request):
                 "lis_outcome_service_url", params.get("lis_outcome_service_url")
             )
 
+        # If the launch has been configured to focus on the annotations from
+        # a particular user, pass
```
seanh: Comment got cut short?

robertknight: Agh yes. Will fix.

Issue #796, opened 2019-07-11
Don't report CanvasAPIAccessTokenErrors to Sentry


seanh:
This error is raised when we don't yet have an access token for the user in our DB, or our access token doesn't work. Its expected behaviour and causes the frontend to show the user a dialog with a button for getting a new access token from Canvas. So this should not be reported to Sentry.

https://sentry.io/organizations/hypothesis/issues/1099418910/?referrer=github_plugin

```
NoResultFound: No row was found for one()
  File "lms/services/canvas_api.py", line 167, in _oauth2_token
    .filter_by(consumer_key=self._consumer_key, user_id=self._user_id)
  File "sqlalchemy/orm/query.py", line 3289, in one
    raise orm_exc.NoResultFound("No row was found for one()")

CanvasAPIAccessTokenError: We don't have a Canvas API access token for this user
(8 additional frame(s) were not displayed)
...
  File "newrelic/hooks/framework_pyramid.py", line 140, in _wrapper
    return wrapper(context, request)
  File "pyramid/viewderivers.py", line 116, in _class_requestonly_view
    response = getattr(inst, attr)()
  File "lms/views/api/canvas/files.py", line 62, in via_url
    self.request.matchdict["file_id"]
  File "lms/services/canvas_api.py", line 114, in public_url
    oauth2_token = self._oauth2_token
  File "lms/services/canvas_api.py", line 174, in _oauth2_token
    ) from err



'https://lms.hypothes.is/api/canvas/files/827960/via_url'

ENVIRONMENT

{'HTTP_ACCEPT': '*/*',
 'HTTP_ACCEPT_ENCODING': 'gzip',
 'HTTP_AUTHORIZATION': 'Bearer '
                       '***',
 'HTTP_CDN_LOOP': 'cloudflare',...
```

PR #873, opened 2019-08-07
Don't report CanvasAPIAccessTokenError to Sentry

hmstepanek:
Filter out CanvasAPIAccessTokenError's from getting reported to Sentry. This is a fix for https://github.com/hypothesis/lms/issues/796.

This filter setup was taken from the [h repo](https://github.com/hypothesis/h/tree/master/h/sentry).


seanh:
I see why these are getting reported to Sentry:

1. The `CanvasAPIClient` service's `_oauth2_token()` helper raises `CanvasAPIAccessTokenError`:
 https://github.com/hypothesis/lms/blob/35221aaba58cf7dc8215421e180270bf108e8ce6/lms/services/canvas_api.py#L171-L174

2. `_oauth2_token()` was called by `list_files()` (still in the service) (though of course it's called by other places in the service too):

   https://github.com/hypothesis/lms/blob/35221aaba58cf7dc8215421e180270bf108e8ce6/lms/services/canvas_api.py#L88

3. `CanvasAPIClient.list_files()` was called by the `/list_files` API's view, which does not catch the error:

   https://github.com/hypothesis/lms/blob/cada2bb3601990f92310dcd22e91a08f7a0862a6/lms/views/api/canvas/files.py#L51

4. So at this point we have an uncaught exception being raised by a view. This is where (I assume) Sentry's Pyramid integration plugs in. Sentry's Pyramid integration doesn't report [`HTTPError`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html)s (or subclasses thereof) but `CanvasAPIAccessTokenError` is not an `HTTPError`. The inheritance goes `CanvasAPIAccessTokenError` -> `ExternalRequestError` -> `ServiceError` -> `Exception`: https://github.com/hypothesis/lms/blob/master/lms/services/exceptions.py

5. The uncaught error is later caught by an error view. This is what catches it and turns it into a nice JSON API response for our frontend code, rather than our app crashing with a 500 Server Error. But I'm guessing that by the time we get to this error view, Sentry's Pyramid integration has already decided to report the exception to Sentry:  


https://github.com/hypothesis/lms/blob/cada2bb3601990f92310dcd22e91a08f7a0862a6/lms/views/api/error.py#L15-L23

So possible solutions to stop this from being reported:

1. Make `CanvasAPIAccessTokenError` be a subclass of one of Pyramid's `HTTPError`s.

   I don't think I like this solution. It feels like mixing view-layer concerns (what type of HTTP status code should be returned) in with the services layer. And also I think subclassing and therefore having to collaborate with Pyramid's `HTTPError` complicates our services exceptions.

2. Make the `list_files()` API view catch the `CanvasAPIAccessTokenError` and raise the appropriate `HTTPError` subclass instead (with exception chaining).

   I don't think I like this either. I like the fact that views don't necessarily have to catch service exceptions, if all they're going to do is re-raise them. It keeps the view code thinner (especially when there are multiple views that might raise services exceptions). And then the exception is handled separately in the error views.

3. Introduce a new layer, as you've done in this PR and as we do in h, that filters out certain errors from Sentry.

   I think this is my preferred solution. It feels like a lot of code just to filter out this one exception, compared so solutions 1 and 2. But it's a nice separation of concerns: it separates the decision of what not to send to Sentry into a separate layer where it doesn't compromise either our services or our views. And of course it's consistent with h.

Issue #869, opened 2019-08-06
Make sure SpeedGrader view works when PDF comes from Google Drive or PDF URL has query params


robertknight:
Original context: https://github.com/hypothesis/lms/issues/656#issuecomment-518586861

Make sure the SpeedGrader view shows the document correctly if an assignment's PDF comes from Google Drive.

In testing locally I found an issue where the PDF loads in the normal LTI launch but not when using the grading view that was submitted to Canvas. I notice that the `via_url` config param passed into the page is different in the two contexts. In the normal LTI launch it is not URL encoded, just concatenated onto the end of the Via URL, whereas in the grading view, the URL is encoded and then concatenated onto the end of the Via URL.

PR #874, opened 2019-08-08
SpeedGrader 5/n - Workaround SpeedGrader launch URL encoding bug

robertknight:
**Depends on https://github.com/hypothesis/lms/pull/865 and https://github.com/hypothesis/lms/pull/867**

Query params in LTI launch URLs submitted to Canvas for use in SpeedGrader get
doubly-encoded so they are still percent-encoded after Pyramid/WSGI have parsed
the request. Implement a workaround by decoding the query param a second time in
the LTI Launch view and signpost the potential issue when generating the
SpeedGrader launch URL.

More generic solutions are possible, but they could have unexpected interactions
with different parts of the system. Therefore this PR just handles the one query
param where this is a problem.

Fixes https://github.com/hypothesis/lms/issues/869

### DIFF
--- a/lms/validation/_launch_params.py
+++ b/lms/validation/_launch_params.py
@@ -1,4 +1,6 @@
 """Schema for validating LTI launch params."""
+from urllib.parse import unquote
+
 import marshmallow
 
 from lms.services import LTILaunchVerificationError
@@ -6,7 +8,7 @@ from lms.validation._helpers import PyramidRequestSchema
 from lms.values import LTIUser
 
 
-__all__ = ("LaunchParamsSchema",)
+__all__ = ("LaunchParamsSchema", "URLConfiguredLaunchParamsSchema")
 
 
 class LaunchParamsSchema(PyramidRequestSchema):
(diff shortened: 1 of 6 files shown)


lms/validation/_launch_params.py
```diff
@@ -68,3 +70,28 @@ def _verify_oauth_1(self, _data):
...
+    Schema for an "URL-configured" Basic LTI Launch.
+
+    An URL-configured launch is one where the content URL is provided by a "url"
+    launch param.
+    """
+
+    url = marshmallow.fields.Str(required=True)
+
+    @marshmallow.post_load
+    def _decode_url(self, _data):  # pylint:disable=no-self-use
```
seanh: I do wonder about globally disabling the no-self-use rule

lms/views/api/lti.py
```diff
@@ -55,6 +55,14 @@ def record_submission(request):
...
     current_score = lti_outcomes_client.read_result(outcome_request_params)
     if current_score is None:
+        # **WARNING**
+        #
+        # Canvas has a bug with handling of percent-encoded characters in the
+        # the SpeedGrader launch URL. Code that responds to the launch will
+        # need to handle this for fields that may contain such chars (eg.
+        # the "url" field).
+        #
+        # See https://github.com/instructure/canvas-lms/issues/1486
```
seanh: Cool, thanks for reporting this

lms/validation/_launch_params.py
robertknight: Hmm ... I _do_ think it provides useful feedback generally, in that I think it is preferable to use a plain function if a method doesn't need `this`, or it may indicate a bug if `self` is not used. The exception is for functions that need to be a method for use with some external library or if you're overriding an inherited method and just want to return a constant.

PR #888, opened 2019-08-13
Use Hypothesis client's "focus mode" to focus on a user in SpeedGrader

robertknight:
~~**Depends on https://github.com/hypothesis/h/pull/5672**~~

In Canvas's SpeedGrader view, configure the Hypothesis client to focus
on a specific user using the "focus" config [1]. This requires making a
call to the Hypothesis API to look up the user's display name first.

This change depends on [2] being deployed first to implement the
`GET /api/users/{userid}` endpoint in h.

[1] https://github.com/hypothesis/client/pull/1291
[2] https://github.com/hypothesis/h/pull/5672

----

SpeedGrader now looks like this:

Screenshot 2019-08-13 10 37 29

If there is a problem fetching the user's display name for any reason, we fall back to a placeholder since the experience is otherwise usable:

Screenshot 2019-08-13 10 38 53


lms/views/basic_lti_launch.py
```diff
@@ -278,3 +279,23 @@ def _set_submission_param(self, name, value):
 
         if "submissionParams" in self.context.js_config:
             self.context.js_config["submissionParams"][name] = value
+
+    def _set_focused_user(self, username):
+        """Configure the Hypothesis client to focus on a particular user."""
+
```
robertknight: I'm guessing you'll want this logic to be moved to a service @seanh? We could either move just the logic that calls the API, or go further and move the whole "focus" config generation into a service. Do you have a preference?

seanh: I think the logic "When it's a SpeedGrader view, then add the "focus" settings to the client config" does belong in the views. So far, although the base client config is generated in `LTILaunchResource`, per-view config customisations have been done in the views.

But I think the logic of how to generate an h userid, and how to generate a client `"focus"` config dict, belongs elsewhere. As does calling the h API.

## Generating the h userid of the focused user

`LTILaunchResource` currently contains a handful of methods for generating properties of Hypothesis users, including the `userid`. But it's hardcoded for generating properties about the currently authenticated user. I think it might be time we extracted a reusable `HUser` out of `LTILaunchResource`:

```python
class HUser:
    username: str
    userid: str
    display_name: str
    provider: str
    provider_unique_id: str
```

`LTILaunchResource` already contains the code necessary to generate the values for all five params from the LTI launch params, in order to create an `HUser` for the authenticated user.

During SpeedGrader launches, to generate an `HUser` for the focused user:

1. `username` is our custom `focused_user` launch param
2. `userid` can be generated from `username` and the `authority` deployment setting
3. `display_name` will have to be retrieved by making a request to the h API. In order to generate a display name from the LTI params like `LTILaunchResource` does we would need the three `lis_person_name_*` params and we don't have those, for the focused user, during a SpeedGrader launch (the launch params will refer to the teacher instead).
4. `provider` is the `tool_consumer_instance_guid` launch param, same as for the authenticated user's `provider`
5. `provider_unique_id` will have to be `None`. Which should be fine as we don't need it. But the inconsistency irks. It can't be generated because it requires the `user_id` LTI param and during a SpeedGrader launch we don't have that for the focused user. And we can't get it from the h API along with `display_name` either, because h's user API doesn't return the `provider_unique_id`.

   We could change the value of the `focused_user` param that we submit to SpeedGrader to be the LTI `user_id` instead of the h `userid`. We'd then be able to generate both the `provider_unique_id` and the `username` from that `focused_user` param (for the `username` we'd also need the `tool_consumer_instance_guid` launch param, which we do have in SpeedGrader launches)

## Calling the h API

I wonder if we need a high-level function or method the current low-level client (the `"hapi"` service) and handle things like generating the URL for the user API, error handling, and default values, keeping these details out of the view.

So far we've been able to keep low-level h API details out of the views by using view decorators: https://github.com/hypothesis/lms/blob/master/lms/views/decorators/h_api.py We have high-level view decorators `upsert_h_user`, `upsert_course_group`, and `add_user_to_group`.

You could add a view decorator that checks for the `focused_user` param, makes the request to the h user API, and adds the display name (or maybe the entire `HUser` object for the focused user) to the `context`.

Or you could add a high-level h API service that wraps the low-level one, or just add a high-level method to the existing low-level service, and have the view call the service.

seanh: Regarding where the code for calling the h API and inserting the `"focus"` dict into the config would go: I think it would be most in harmony with the current design if this were implemented as a view decorator. Maybe called `focus_user`. Add it to the class's list of view decorators. This decorator would:

1. Do nothing if there's no `"focused_user"` request param

2. Call the h API to get the display name

   Using the existing low-level h API client, and doing error handling, as the existing view decorators do. This is also where defaulting to `"(Couldn't fetch student name)"` would go if you want that.

3. Create an `HUser` for the focused user

4. Generate the `"focus"` dict with the `"username"` and `"displayName"` settings in it, and insert the dict into `context.hypothesis_config`.

robertknight: What are the guidelines for using a service vs a decorator to share functionality between views or extract functionality out of views in the lms app? In h the distinction seems mainly to do with separating concerns that are specific to a particular endpoint (eg. a service call to create an entity in the DB) vs. broader lower-layer concerns applicable to multiple views (but not used outside of views) like logging, or adding CORS headers or configuring how certain categories of response are handled (eg. mapping 404 to 403).

robertknight: Thanks for the feedback. I'll take a look at this tomorrow, going with the proposal to create an `HUser` struct and view decorator. I'm not completely convinced about using decorators for this but it is in keeping with the way that the rest of this view works.

seanh: No guidelines. The reason those parts were implemented as view decorators was because, when implementing provisioning last year, we didn't want to touch the horrible view code, and view decorators were easy to write separately and then apply to multiple views.

It does turn out to be handy though that you can add the view decorators once in `@view_defaults` but then still turn them off for individual views in `@view_config`.

robertknight:
I have refactored the code to move the call to the h API service into a new high-level h API client, `HypothesisOperationsService` (name="h_ops") which returns `HUser` instances rather than a dict.

On reflection I decided to keep the focused-user generation code in the launch views because the details of how to fallback when the user cannot be found are context-specific (a dummy user is OK here) and it isn't clear to me that would always be appropriate.

lms/services/h_operations.py
```diff
@@ -0,0 +1,33 @@
...
+
+        try:
+            user_info = self.hapi_svc.get(path=f"users/{userid}").json()
+            return HUser(
+                authority=authority,
+                username=username,
+                display_name=user_info["display_name"],
+            )
+        except HAPIError as err:
+            raise err
```
robertknight: At present there is no wrapping of the exception type from the lower-level code here. This just makes it explicit that we're choosing to let these errors propagate. I'm not sure there is a whole lot of value in wrapping the exception at this level. Thoughts?

lms/values.py
```diff
@@ -16,3 +16,34 @@ class LTIUser(NamedTuple):
...
+    display_name: str
+    """The display name for the user, generated from the LMS user's display name."""
+
+    username: str
+    """
+    Generated username for h user.
+
+    This is derived from the LMS userid.
+    """
+
```
robertknight: I haven't added the `provider` and `provider_unique_id` fields yet because they are not used in this PR.

lms/services/h_operations.py
```diff
@@ -0,0 +1,33 @@
...
+    on top of that client.
+    """
+
+    def __init__(self, _context, request):
+        self.hapi_svc = request.find_service(name="hapi")
+        self.request = request
+
+    def fetch_user(self, username):
+        authority = self.request.registry.settings["h_authority"]
+        userid = f"acct:{username}@{authority}"
```
seanh: Is there some way to avoid duplicating this string formatting? It's now in three places on this branch. If we change `HUser` to make `display_name` optional then you can do:

```python
>>> HUser("seanh", "hypothes.is").userid
'acct:seanh@hypothes.is'
```

Here's the diff to enable this (I had to move `display_name` to the bottom because optional args must come after all required ones; I also swapped the order of username and authority so that they make sense when used as positional args):

```diff
diff --git a/lms/values.py b/lms/values.py
index 64572dd1..c12c1761 100644
--- a/lms/values.py
+++ b/lms/values.py
@@ -27,6 +27,13 @@ class HUser(NamedTuple):
     usernames.
     """
 
+    username: str
+    """
+    Generated username for h user.
+
+    This is derived from the LMS userid.
+    """
+
     authority: str
     """
     The authority which the user belongs to.
@@ -34,16 +41,9 @@ class HUser(NamedTuple):
     This should always match ``settings["h_authority"]``
     """
 
-    display_name: str
+    display_name: str = ""
     """The display name for the user, generated from the LMS user's display name."""
 
-    username: str
-    """
-    Generated username for h user.
-
-    This is derived from the LMS userid.
-    """
-
     @property
     def userid(self):
         return f"acct:{self.username}@{self.authority}"
```

lms/services/h_operations.py
seanh: `HAPIError` is already a services error (its full name is `lms.services.HAPIError`) so I agree there's no need to wrap it in a different exception class.

I think it's good to be explicit about re-raising in the code here. I'd also add a docstring, and document this in the docstring with a `:raise`. The reason is that what exceptions it throws and when/why is an important part of the interface of a service, and is something that someone trying to write a view that uses the service needs to know (and I wouldn't want them to have to dig in to the lower-level `hapi_svc` to spot it).

On the other hand, when _views_ call services that might raise exceptions, I have _not_ had the views explicitly catch and re-raise exceptions like this. They just let them fly. (And the exceptions are later caught be error views.) Given that we don't write code that calls our view code, I think in the case of views it's nicer to keep the code cleaner.

lms/values.py
```diff
@@ -16,3 +16,34 @@ class LTIUser(NamedTuple):
...
+    @property
+    def userid(self):
+        return f"acct:{self.username}@{self.authority}"
```
seanh: :+1:

tests/lms/views/basic_lti_launch_test.py
```diff
@@ -69,14 +72,43 @@ def test_it_doesnt_add_report_submission_config_if_speedgrader_flag_off(
...
+        assert context.hypothesis_config["focus"] == {
+            "user": {
+                "username": "user123",
+                "displayName": "(Couldn't fetch student name)",
+            }
+        }
+
+    @pytest.fixture
+    def hops_service(self, pyramid_config):
+        svc = mock.create_autospec(HypothesisOperationsService, instance=True)
```
seanh:
```suggestion
        svc = mock.create_autospec(HypothesisOperationsService, instance=True, spec_set=True)
```

tests/lms/services/h_operations_test.py
```diff
@@ -0,0 +1,40 @@
...
+    def user_response(self):
+        return {"display_name": "Jim Smith"}
+
+    @pytest.fixture
+    def hops_service(self, hapi_service, pyramid_request):
+        return HypothesisOperationsService({}, pyramid_request)
+
+    @pytest.fixture
+    def hapi_service(self, pyramid_config):
+        svc = mock.create_autospec(HypothesisAPIService, instance=True)
```
seanh:
```suggestion
        svc = mock.create_autospec(HypothesisAPIService, instance=True, spec_set=True)
```

robertknight:
I have renamed `HypothesisOperationsService` to `HAPIClient` per one of the suggestions above and will follow up with a rename of the low-level client to `HAPIRequests` in a follow-up PR. I've also addressed the other items.

PR #905, opened 2019-08-15
Combine `LTILaunchResource` h user properties into `h_user`

robertknight:
Combine `LTILaunchResource.h_{username, display_name, userid}`
properties into a single `h_user` property which returns an `HUser`.

This removes the duplication of logic for generating the userid from the
username and authority between `LTILaunchResource` and `HUser`. In
future it will also make it possible to add other convenience properties
to `HUser`.

This is a follow-up to [1]

[1] https://github.com/hypothesis/lms/pull/888#pullrequestreview-274807714

### DIFF
--- a/tests/lms/resources/_lti_launch_test.py
+++ b/tests/lms/resources/_lti_launch_test.py
@@ -143,10 +143,10 @@ class TestLTILaunchResource:
     def test_h_display_name(
         self, request_params, expected_display_name, pyramid_request
     ):
-        pyramid_request.params = request_params
+        pyramid_request.params.update(request_params)
 
         assert (
-            resources.LTILaunchResource(pyramid_request).h_display_name
+            resources.LTILaunchResource(pyramid_request).h_user.display_name
             == expected_display_name
         )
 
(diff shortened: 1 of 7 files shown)


tests/lms/resources/_lti_launch_test.py
```diff
@@ -269,45 +269,26 @@ def test_h_username_raises_if_tool_consumer_instance_guid_is_missing(
...
             "tool_consumer_instance_guid": "VCSy*G1u3:canvas-lms",
             "user_id": "4533***70d9",
         }
 
-        userid = resources.LTILaunchResource(pyramid_request).h_userid
+        userid = resources.LTILaunchResource(pyramid_request).h_user.userid
 
         assert userid == "acct:2569ad7b99f316ecc7dfee5c0c801c@TEST_AUTHORITY"
 
-    def test_h_userid_raises_if_tool_consumer_instance_guid_is_missing(
```
robertknight: After combining the properties, these tests become unnecessary because they are already covered by the `h_user` tests.

PR #925, opened 2019-08-20
Run frontend tests in headless Chrome

robertknight:
Following on from https://github.com/hypothesis/client/pull/1279, this PR converts the LMS frontend tests to run in a modern browser (headless Chrome) rather than PhantomJS, providing a nicer development experience and a test environment that more closely matches end users' browser.

There are several pieces to this:

- Improve the reliability of the existing tests under Chrome (first two commits)
- Switch tests to run under Chrome instead of PhantomJS locally and on CI (third commit). This uses a setup very similar to the [recent client PR](https://github.com/hypothesis/client/pull/1279)
- Remove polyfills that were only used when running tests and are not needed in modern browsers (last commit)

One change I have made from the setup in the client repo is that I haven't added the Puppeteer package (a pre-built version of Chrome installed via node). In Jenkins, we have to use Chromium installed via Alpine anyway and for local development, I think it is reasonable to assume that every developer working on the lms app will have a reasonably current version of Chrome or Chromium installed. The downside is that different users might have slightly different versions. My instinct is that probably won't be a problem and it simplifies our dependencies. We can always revisit this if it proves to be the wrong choice.

---

- [x] Make sure the tests run in Jenkins
- [x] Fix remaining reliability issues with tests in Chrome. They currently fail intermittently on my local system


package.json
```diff
@@ -50,14 +51,13 @@
     "through2": "^3.0.1",
     "uglifyify": "^5.0.1",
     "vinyl": "^2.2.0",
-    "watchify": "^3.11.1",
-    "whatwg-fetch": "^3.0.0"
+    "watchify": "^3.11.1"
   },
   "devDependencies": {
+    "@types/enzyme": "^3.10.3",
```
robertknight: This is not essential, but it provides a direct way to specify the argument/return types of the `waitFor*` helpers added in this PR expect as arguments and it provides intellisense for devs who have it set up in their editors (VS Code out of the box, ale + tsserver in Vim).

Dockerfile.frontend-test
```diff
@@ -0,0 +1,17 @@
+# nb. We use Alpine as a base image and then install Node + Yarn separately
```
robertknight: I don't think we have a convention for naming auxiliary Dockerfiles in our repos. I'm happy to take suggestions. The reason I'm not just using `alpine:3.10` and installing the dependencies within the image in Jenkins on every run is because this approach makes the build a bit faster / lighter on Jenkins resources by benefiting from Docker layer caching.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -236,7 +213,7 @@ describe('BasicLtiLaunchApp', () => {
     fakeConfig.submissionParams = undefined;
 
     renderLtiLaunchApp();
-    await nextTick();
+    await new Promise(resolve => setTimeout(resolve, 0));
```
robertknight: This code does the same as the previous `await nextTick()`, but I removed that helper since it was only used in one place.

Dockerfile.frontend-test
seanh: I did notice that this is `frontend-test` whereas the make command is `frontend-tests`

robertknight:
> Since it's only used on Jenkins, can we move the new Dockerfile into a Jenkins/ directory to clarify that it's not used in local development?

Done. I renamed the file to have `.dockerfile` extension as this makes it easier for tools to determine the file type. It looks like supported conventions by IDEs are to either name the file `Dockerfile`, in which case you need to use a dir per image, or give it a `.dockerfile` extension.

> By the I think there is always an env.JENKINS_URL set on Jenkins which could be used, rather than RUNNING_IN_DOCKER, and I believe this gets passed in to containers running in Jenkins by default. 

I believe you're right. I added a specific env var to make it marginally easier to test locally. An alternative approach would be to just use `chromium-browser` if that binary exists in `PATH`.

Issue #767, opened 2019-07-05
Provide a helpful error message and details when the LMS URL is wrong


seanh:
When you register at https://lms.hypothes.is/welcome you have to enter your **LMS Domain**. This domain is used as the domain for our server to send requests to, and also to direct the authorization popup window to, when using the Canvas API. If you enter this domain wrongly (e.g. just enter example.com) then when you try to create a Canvas file assignment you get stuck in an _Allow file access_ loop:

![Peek 2019-07-05 16-35](https://user-images.githubusercontent.com/22498/60732809-eaf39280-9f42-11e9-8ca4-9d24d1987a37.gif)

We may not be able to do much about the wrong page showing up in the popup window, but we should at least be able to provide a better error message when the list_files request fails.

Related: [Validate LMS URL's on the /welcome page](https://github.com/hypothesis/lms/issues/592)

hmstepanek:
This happens when it is the first time the user has tried to access their files (as shown above in the gif). The authorize message appears because the user does not have an oauth2 token yet and so it fails out before a request is even sent to canvas for the files. It fails on a self._oauth2_token attribute doesn't exist on [this line]( https://github.com/hypothesis/lms/blob/74eacedb5d03931d80fe267413426c851a5eb1e1/lms/services/canvas_api.py#L88).

hmstepanek:
I started a slack thread on this [here](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1565124134028700). I'm not sure where to go from here so I'll await feedback from the team.

robertknight:
Fixed in https://github.com/hypothesis/lms/pull/941

PR #941, opened 2019-08-27
Provide user with error details indicating an incorrect canvas url

hmstepanek:
This is a fix for https://github.com/hypothesis/lms/issues/767. This is dependent on https://github.com/hypothesis/lms/pull/939.

To test:
1. Create an lms instance in the lms database that points to https://example.com rather than the correct canvas url. 
2. Try to create a new assignment by selecting a file from canvas.
3. It will prompt you to "authorize Hypothesis to access your files in Canvas". Click authorize.
4. A pop up will appear that opens example.com. Close this.
5. You should see:
image

### DIFF
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
@@ -1,4 +1,4 @@
-import { createElement } from 'preact';
+import { Fragment, createElement } from 'preact';
 import { useCallback, useEffect, useRef, useState } from 'preact/hooks';
 import propTypes from 'prop-types';
 
@@ -34,11 +34,15 @@ export default function LMSFilePicker({
   authUrl,
   courseId,
   lmsName,
+  registeredLmsUrl,
   onCancel,
   onSelectFile,
 }) {
   // The main state of the dialog and associated data.
   const [dialogState, setDialogState] = useState(INITIAL_DIALOG_STATE);
+
+  // Authorization attempt was made. Set after state transitions to "authorizing".
+  const [authorizationAttempted, setAuthorizationAttempted] = useState(false);
 
   // The file within `files` which is currently selected.
   const [selectedFile, selectFile] = useState(null);
@@ -69,6 +73,7 @@ export default function LMSFilePicker({
   // fetch files.
   const authorizeAndFetchFiles = useCallback(async () => {
     setDialogState({ ...INITIAL_DIALOG_STATE, state: 'authorizing' });
+    setAuthorizationAttempted(true);
 
     if (authWindow.current) {
       authWindow.current.focus();
@@ -120,7 +125,12 @@ export default function LMSFilePicker({
           <Button
             key="showAuthWindow"
             onClick={authorizeAndFetchFiles}
-            label={dialogState.state === 'error' ? 'Try again' : 'Authorize'}
+            label={
+              dialogState.state === 'error' ||
+              (dialogState.state === 'authorizing' && authorizationAttempted)
+                ? 'Try again'
+                : 'Authorize'
+            }
           />
         ) : (
           <Button
@@ -138,7 +148,16 @@ export default function LMSFilePicker({
           error={dialogState.error}
         />
       )}
-      {dialogState.state === 'authorizing' && (
+      {dialogState.state === 'authorizing' && authorizationAttempted && (
+        <ErrorDisplay
+          message=<Fragment>
+            <a>{`Failed to authorize with the ${lmsName} instance at `}</a>
+            <a href={`${registeredLmsUrl}`}>{`${registeredLmsUrl}`}</a>
+          </Fragment>
+          error={new Error('')}
+        />
+      )}
+      {dialogState.state === 'authorizing' && !authorizationAttempted && (
         <p>
           To select a file, you must authorize Hypothesis to access your files
           in {lmsName}.
@@ -178,6 +197,11 @@ LMSFilePicker.propTypes = {
    * The name of the LMS to display in API controls, eg. "Canvas".
    */
   lmsName: propTypes.string.isRequired,
+
+  /**
+   * The url of the LMS, eg. "https://foobar.instructure.com".
+   */
+  registeredLmsUrl: propTypes.string.isRequired,
 
   /** Callback invoked if the user cancels file selection. */
   onCancel: propTypes.func.isRequired,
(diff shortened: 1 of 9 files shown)


seanh:
Couple of bits of feedback here:

1. I don't think the <samp>Details</samp> twisty saying `registeredLmsUrl": "https://example.com"` is needed. That's not really an error message and doesn't provide anything not present in the <samp>Authorization failed for Canvas https://example.com</samp> above it.

2. I think we could tweak the <samp>Authorization failed for Canvas https://example.com</samp> to be a bit clearer. Maybe something along the lines of:

   > **Allow file access**
   > 
   > Failed to authorize with the Canvas instance at **https://example.com**.

lms/views/basic_lti_launch.py
```diff
@@ -210,6 +210,9 @@ def unconfigured_basic_lti_launch(self):
                     "google_developer_key"
                 ],
                 "lmsUrl": self.context.lms_url,
+                "registeredLmsUrl": self.request.find_service(name="ai_getter").lms_url(
+                    oauth_consumer_key
+                ),
```
seanh: Isn't `registeredLmsUrl` the same value as the `lmsUrl` variable above it?

lms/views/content_item_selection.py
```diff
@@ -80,6 +80,9 @@ def content_item_selection(context, request):
             # needs to pass this URL (which is the URL of the top-most page) to Google
             # Picker, otherwise Picker refuses to launch inside an iframe.
             "lmsUrl": context.lms_url,
+            "registeredLmsUrl": request.find_service(name="ai_getter").lms_url(
+                request.lti_user.oauth_consumer_key
```
seanh: Isn't `registeredLmsUrl` the same value as the `lmsUrl` variable above it?

lms/views/basic_lti_launch.py
hmstepanek: No, lmsUrl is the lmsUrl from the client's point of view (the lms url you currently have open in the browser), registeredLmsUrl is the url from the database that was provided when registering a new lms application with Hypothesis.

hmstepanek:
I see your point about the details repeating the information. I do feel there is value in providing the support contact information though so I have made an empty error (without any details or error message) to pass into the ErrorDisplay). 

 > If the problem persists send us an email or open a support ticket. You can also visit our help documents.

An alternative to this approach is to recreate the contact support message inside the LMSFilePicker but I think it's probably better to keep that error message in one place and just re-use it. Another option still would be to extract it out into a common place somewhere but I think it's ok for now. Let me know if you think otherwise.

lms/views/basic_lti_launch.py
robertknight: To expand on Hannah's comment above: When the user clicks the "Authorize" button in the client, that opens the `canvas_api.authorize` route's URL in a popup window. That URL in turn redirects to the Canvas instance's OAuth 2 authorization endpoint. The `canvas_api.authorize` route uses the application instance's LMS URL from our database. The `lmsUrl` variable here on the other hand is the `custom_canvas_api_domain` variable from the LTI launch, if present, with a fallback to the application instance's LMS URL from the database. The `lmsUrl` frontend config key is currently used by the Google Picker, which needs to know the origin of the top-level frame in the window, if the LTI launch happens inside an iframe.

If `lmsUrl` and `registeredLmsUrl` are different here, that probably does indicate a misconfiguration of the LTI installation though, one that ideally we would have caught at the point when the LTI installation is registered.

seanh: It's called `lms_url` in `ApplicationInstance` but has been renamed to `registeredLmsUrl` here (understandably, since something else called `lmsUrl` is already here).

Suggest maybe naming it `applicationInstanceLmsUrl` instead?

Something that avoids introducing a second name for the same thing ("application instance" vs "registration" / "registered"). "Register" / "registration" is not currently used in the code to refer to this concept. (It does appear in a UI string: "Register here..." but that's all.) 

The existing `lmsUrl` name (the `custom_canvas_api_domain` variable from the LTI launch, if present, with a fallback to the application instance's LMS URL from the database) might not be a great name itself either. There might be a better name for that, that better captures what it is or what it's used for and distinguishes it from `applicationInstanceLmsUrl`.

Anyway, I'd like us to consider the naming here, as that's important for code clarity, and having both `lmsUrl` and `registeredLmsUrl` in the same place isn't super clear.

seanh: One approach might be to name them `applicationInstanceLmsUrl` (or just `lmsUrl`) and `customCanvasApiDomain` and allow `customCanvasApiDomain`  to be null rather than falling back. Then put the  `customCanvasApiDomain || applicationInstanceLmsUrl` logic in the JavaScript code where `customCanvasApiDomain` is used instead. Not sure if that will work out well or not, but it does avoid the naming problem or the introduction of confusing names

hmstepanek: > . One approach might be to name them applicationInstanceLmsUrl (or just lmsUrl) and customCanvasApiDomain and allow customCanvasApiDomain to be null rather than falling back. Then put the customCanvasApiDomain || applicationInstanceLmsUrl logic in the JavaScript code where customCanvasApiDomain is used instead. Not sure if that will work out well or not, but it does avoid the naming problem or the introduction of confusing names.

I generally like this approach although it requires a bit of refactoring.

hmstepanek:
@robertknight or @seanh can one of you review this again? thx!

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -138,7 +148,16 @@ export default function LMSFilePicker({
           error={dialogState.error}
         />
       )}
-      {dialogState.state === 'authorizing' && (
+      {dialogState.state === 'authorizing' && authorizationAttempted && (
+        <ErrorDisplay
+          message=<Fragment>
```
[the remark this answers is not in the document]
hmstepanek: I noticed chrome parsed it a bit funny syntax-wise but since the linter was ok I didn't read to much into it. Thanks for the pointer. Seems to be happy with that change.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -82,6 +82,73 @@ describe('LMSFilePicker', () => {
     assert.isTrue(wrapper.exists('FakeButton[label="Authorize"]'));
   });
 
+  
+  it('xx shows the try again prompt with error details after a failed authorization attempt', async () => {
```
hmstepanek: Having trouble with this test.

hmstepanek: 1. try to access a file and get an ApiError not authorized which prompts to authorize
2. click authorize
3. close the authorization window <- how to do this??
4. the original error should have changed to show the invalid domain after the failed auth attempt.

PR #947, opened 2019-08-29
WIP - Fix flakey dialog test

robertknight:
Add some debugging output to help track down the cause of flakey focusing tests for `Dialog`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,31 +27,32 @@ node {
 
     // Run each of the stages in parallel.
     parallel failFast: true,
-    "Backend lint": {
-        stage("Backend lint") {
-            testApp(image: img, runArgs: runArgs) {
-                installDeps()
-                run("make checkformatting")
-                run("make checkdocstrings")
-                run("make backend-lint")
-            }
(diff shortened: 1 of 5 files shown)


robertknight:
After some experimentation locally, it looks like the problem is not what I initially thought. When the test fails, the dialog is rendered, and then the assert to check that a particular element was focused is run, but the effect that actually calls `element.focus()` is not run first.

This looks related to https://github.com/preactjs/preact/issues/1794. A fix has been merged upstream but not released yet.

robertknight:
In conversation with the Preact devs, a new v10.x release is planned for early next week. Once that lands, all we should need to do is upgrade our dependencies.

One thing that can be done in the meantime is that I can verify that building Preact from master and linking it locally into the lms app using `npm link` or `yalc` does resolve the issue.

robertknight:
Quick update on this - it is still blocked pending a new Preact release. I might put a short-term workaround in place if necessary.

robertknight:
This should be fixed by https://github.com/hypothesis/lms/pull/975.

PR #991, opened 2019-09-18
Add a LMS grader UI component to the top of the page

LMS007:
This adds the basic component for a LMS Grader and a away to drop the sidebar into a user focused mode. Switching between users will re-render the sidebar with a different user. Future versions will use postmessage and more elegantly update the sidebar without costly re-renders.

This does not add any support for submitting a grade at this time, more so the UX for grade submission is not designed, but It's assumed it would fit  inside the top affixed LMSGrader component.

2 flags are required in LMS to get the pdf into grader mode. 
```
export FEATURE_FLAGS_ALLOWED_IN_COOKIE="blackboard_grading"
export FEATURE_FLAG_LMS_GRADER="TRUE"
```

Additionally we need to hack up the list of students. The simple way to do that is to do that is drop this code in frontend-apps/index.js right before the render call.

```
config.grading.students = [
  {
    username: 'user1',
    displayName: 'User 1'
  },
  {
    username: 'user2',
    displayName: 'User 2'
  },
  {
    username: 'user3',
    displayName: 'User 3'
  }
]
```

Screen Shot 2019-09-18 at 1 29 55 PM

note: _The UX for the drop down is just left as default.  I suspect we may want to use a different element structure so we can more easily style this but that is for future ux discussion_


lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,98 @@
...
+
+  // Initially set the focused student sidebar config
+  setFocusedStudentConfig(students[currentStudent]);
+
+  /**
+   * Main handler function to perform any state updates
+   * required change a focused user.
+   */
+  const onSelectStudent = studentIndex => {
+    setFocusedStudentConfig(students[studentIndex]);
```
robertknight: If updating the Hypothesis client config in the page was implemented as an effect with a dependency on `students[currentStudent]` then it would be automatically run whenever that changed, so there wouldn't be a need to call it here.

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,98 @@
...
+        </li>
+      </ul>
+      {children}
+    </header>
+  );
+}
+
+LMSGrader.propTypes = {
+  // Callback to alert the parent component that a change has occurred and re-rendering may be needed.
+  onChangeUser: propTypes.func.isRequired,
```
robertknight: `onChangeUser` is may be a little ambiguous. `onChangeSelectedUser` perhaps?

lms/static/scripts/frontend_apps/components/StudentsSelector.js
```diff
@@ -0,0 +1,92 @@
...
+      >
+        <img src="/static/images/arrow-left.svg" />
+      </button>
+      <div className="StudentsSelector__student">{studentList()}</div>
+      <button
+        aria-label="next student"
+        disabled={!hasNextStudent}
+        onClick={onNextStudent}
+      >
+        <img src="static/images/arrow-right.svg" />
```
robertknight: This `src` uses a relative path (`static/...`) whereas the one above uses an absolute path (`/static/...`), which I think is what you intended to do here.

lms/static/scripts/frontend_apps/components/StudentsSelector.js
```diff
@@ -0,0 +1,92 @@
...
+  const onPrevStudent = () => {
+    if (selectedStudentIndex > 0) {
+      onSelectStudent(selectedStudentIndex - 1);
+    }
+  };
+
+  /**
+   * Build the <select> list from the current array of students
+   */
+  const studentList = () => {
```
robertknight: Since this function has no arguments and is used unconditionally, it could just be a value rather than a function.

lms/static/scripts/frontend_apps/components/StudentsSelector.js
```diff
@@ -0,0 +1,92 @@
...
+    const options = students.map((student, i) => {
+      return (
```
robertknight: AFAIK the `=> { return <JSX expression> }` can be simplified slightly to just `=> (<JSX expression>)`. We don't have any hard rules about this, I just find it marginally easier to read with  fewer tokens.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -224,4 +232,33 @@ describe('BasicLtiLaunchApp', () => {
...
+    beforeEach(() => {
+      fakeConfig.lmsGrader = true;
+      fakeConfig.grading = {
+        students: [{ username: 'user1' }, { username: 'user2' }],
+      };
+    });
+
+    it('renders the LMSGrader component', () => {
+      const wrapper = renderLtiLaunchApp();
+      const LMSGrader = wrapper.find('.LMSGrader');
```
robertknight: I would suggest to mock the `LMSGrader` component here as we do for `Dialog`, `ErrorDisplay` etc. above. Otherwise the side-effects/implementation details of that component end up complicating these tests. For example, here you've had to mock `document.querySelector` even though the `BasicLtiLaunchApp` component doesn't touch that as its a concern of the `LMSGrader` component.

lms/static/styles/components/_LMSGrader.scss
```diff
@@ -0,0 +1,35 @@
...
+    margin-left: 20px;
+  }
+  ul {
+    background-color: $grey-7;
+    padding: 0;
+    margin: 0;
+    list-style: none;
+    display: flex;
+    
+    .LMSGrader__course-assignment {
```
robertknight: Nesting a class selector within a class selector generates a descendant selector (`.LMSGrader .LMSGrader__course-assignment`) which has a higher specificity that can complicate reasoning about how styles are combined. Usually we would either put each element declaration at the top level, or use:

```scss
.Component {
  &__element {
    // Styles
  }
}
```

Which will only generate a `Component__element` class rather than `.Component .Component__element`.

lms/static/styles/components/_LMSGrader.scss
```diff
@@ -0,0 +1,35 @@
+.LMSGrader {
+  font-size: 18px;
+  height: 100%;
+  width: 100%;
+  position: fixed;
+  color: white;
+  h2, h3 {
+    margin-left: 20px;
+  }
+  ul {
```
robertknight: This will generate a `.LMSGrader ul` selector which will apply to any list rendered anywhere within the grader, including its child components. This means that a change here could unexpectedly affect a list rendered by some other component. Our convention is to use class selectors only in most cases to avoid these hazards.

lms/static/styles/components/_StudentsSelector.scss
```diff
@@ -0,0 +1,28 @@
...
+    }
+    &:disabled {
+      &:hover {
+        background-color: $grey-6;
+      }
+      cursor: default;
+      opacity: 0.5;
+    }
+  }
+  .StudentsSelector__student {
```
robertknight: See notes about descendant selectors in the `LMSGrader` component.

lms/static/styles/components/_LMSGrader.scss
LMS007: oops, this was an oversite! I forgot it was a root class like this

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,98 @@
...
+
+export default function LMSGrader({ children, onChangeUser }) {
+  const { grading } = useContext(Config);
+  const students = grading.students ? grading.students : [];
+
+  // Initial current student is the first in the list
+  const [currentStudent, setCurrentStudent] = useState(0);
+
+  // Reference to the sidebar config DOM node
+  const sidebarConfigEl = document.querySelector('.js-hypothesis-config');
```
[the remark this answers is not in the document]
LMS007: Thanks for the suggestion, that makes a lot of sense and should make this quite a bit better.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -179,9 +191,30 @@ export default function BasicLtiLaunchApp() {
...
+
+    if (lmsGrader) {
+      // Use the LMS Grader
+      return (
+        <LMSGrader
+          onChangeUser={username => {
+            setFocusedUser(username);
+          }}
+        >
+          {iFrame}
```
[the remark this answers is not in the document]
LMS007: I gave this a try and I'd rather leave it nested. For one thing, when its nested, the sidebar reacts to the grader's block content and automatically gets placed below it because the <ul> is its sibling. If we move the iframe outside of this component it renders under the grader and is hidden.

Screen Shot 2019-09-20 at 5 24 24 PM


Id rather not do a fixed padding or anything. But I think the dOM didn't make sense as it was. It makes it more of a symbolic dom element now, but has good readability.

Screen Shot 2019-09-20 at 5 23 25 PM

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,93 @@
...
+      .first()
+      .simulate('click');
+    assert.isTrue(onChange.calledWith(0));
+  });
+
+  it('clicking the previous button when there are no previous students should not trigger the callback', () => {
+    const onChange = sinon.spy();
+    const wrapper = renderSelector({ onSelectStudent: onChange });
+    wrapper
+      .find('button')
```
LMS007: I realize this is a deep mount, It just seemed a bit easier to trigger the event this way. If there is another simple way to do this, I'd like to change this to a shallow mount and not rely on the Student Selector. 

but having said that, these are pretty strongly coupled in any case..

lyzadanger:
Hi, @LMS007 !  A couple of high-level thoughts before a deeper review:

* What do you think of having an initial grader state that does not focus a user by default? I think it could potentially be jarring for every teacher launch of an assignment to always focus on the first (arbitrary) student in the list. It seems like it might be nice to have a launch state that is not user-focused until the instructor selects a student from the list (or starts paging through students with the arrow controls).
* Maybe take a minute or two to do some reflection about what component "owns" the change/selection of a focused user. There are several components that juggle some callbacks and state management, and it might avoid future confusion to make sure we have clarity on this now.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -179,9 +184,31 @@ export default function BasicLtiLaunchApp() {
...
+        className="js-via-iframe"
+        src={ltiLaunchState.contentUrl}
+      />
     );
+
+    if (lmsGrader) {
+      // Use the LMS Grader
+      return (
+        <LMSGrader
+          onChangeSelectedUser={username => {
```
LMS007: So this callback is only intended to alert this component to rebuild the iframe.  This is a change that is propagated from 2 components down, but I don't think it's enough of a use case to include a redux store as that would be more complex than a prop callback. We may rethink this when we get the submit feature working

lms/static/styles/components/_BasicLtiLaunchApp.scss
```diff
@@ -12,3 +12,8 @@
   left: calc(50% - #{$spinner-size / 2});
   top: 100px;
 }
+
+.js-via-iframe {
+  /* Fix to allow the sidebar to scroll to the bottom while we have fixed grader header. */
+  padding-bottom: 96px;
```
LMS007: A hack to get the sidebar to be able to scroll to its bottom. The grader is pushing it down and since the sidebar is in an iframe and we use height: 100% it gets cut off. This is just a simple fix, but it's not a great solution because its a calculated padding based on the height of the grader itself. I need to come up with a better css fix.

You'll only see this issue if you have enough annotations to fill up the entire height. Then add a new page note and you won't be able to see the submit button w/o this css fix.

lms/static/images/arrow-left.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 16 16" aria-hidden="true" focusable="false" class="Icon Icon--arrow-left"><g fill-rule="evenodd"><rect fill="none" stroke="none" x="0" y="0" width="24" height="24"></rect><path fill="none" stroke="white" stroke-linecap="square" stroke-linejoin="miter" stroke-width="2" d="M7 12L3 8l4-4M4 8h9-9z"></path></g></svg>
```
LMS007: This is from the style guide 
https://hypothesis-pattern-library.netlify.com/

**I did change the color

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -179,9 +184,31 @@ export default function BasicLtiLaunchApp() {
   }, [authToken, authUrl, fetchContentUrl, lmsName]);
 
   if (ltiLaunchState.state === 'fetched-url') {
-    return (
-      <iframe width="100%" height="100%" src={ltiLaunchState.contentUrl} />
+    const iFrame = (
+      <iframe
+        key={sideBarKey}
```
LMS007: note: will replace the `key` re-rendering with postMessage later on

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,85 @@
...
+    setCurrentStudent(studentIndex);
+    onChangeSelectedUser(students[studentIndex].userid);
+  };
+
+  return (
+    <div className="LMSGrader">
+      <header>
+        <ul className="LMSGrader_flex-container">
+          <li className="LMSGrader__course-assignment">
+            <h2>Course Name</h2>
```
lyzadanger: I wonder if we might pull these two elements for now (course, assignment name)—we haven't actually talked about how we'll populate them (it shouldn't be too tricky, but I'd hate to hold up rolling this out because we're hung up on displaying these values)...thoughts?

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,85 @@
...
+  const onSelectStudent = studentIndex => {
+    setCurrentStudent(studentIndex);
+    onChangeSelectedUser(students[studentIndex].userid);
+  };
+
+  return (
+    <div className="LMSGrader">
+      <header>
+        <ul className="LMSGrader_flex-container">
+          <li className="LMSGrader__course-assignment">
```
lyzadanger: This classname feels perhaps a little too specific?

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,97 @@
...
+        value={i}
+      >
+        {student.displayName}
+      </option>
+    ));
+
+    if (selectedStudentIndex < 0) {
+      // add the default option (no selection)
+      options.push(
+        <option hidden disabled selected>
```
lyzadanger: The way these attrs work together make it impossible to get out of user-focused mode once you're in it, unfortunately. It also makes it confusing when first opening the `select` because it looks like the first student is selected:

image

Every BlackBoard instructor launching an assignment is going to get chucked into this mode, so I think we need to have an escape hatch back to viewing everyone, perhaps a default option like "All Students"?

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -0,0 +1,73 @@
...
+      />
+    );
+  };
+
+  it('does not have a focused user by default', () => {
+    renderGrader();
+    sinon.assert.notCalled(fakeUpdateClientConfig);
+  });
+
+  it('clicking the next button changes to the first focused user', () => {
```
lyzadanger: I'm a little confused by this test as it seems to be testing behavior in `StudentSelector`...can you help me grok?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  it('calls the onSelectStudent callback when the select list is changed', () => {
+    const onChange = sinon.spy();
+    const wrapper = renderSelector({ onSelectStudent: onChange });
+    wrapper.find('select').simulate('change');
+    assert.isTrue(onChange.called);
+  });
+
+  it('calls the callback with the next student index', () => {
```
lyzadanger: Should this test name reflect that there is a click involved here? And/or which button it's clicking on?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,96 @@
...
+    const onChange = sinon.spy();
+    const wrapper = renderSelector({ onSelectStudent: onChange });
+    wrapper
+      .find('button')
+      .last()
+      .simulate('click');
+    assert.isTrue(onChange.calledWith(1));
+  });
+
+  it('clicking the previous button calls the callback with the next student index ', () => {
```
lyzadanger: Maybe use this test description as a template for the preceding test (which tests the "next" button)?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
lyzadanger: > I'd like to change this to a shallow mount and not rely on the Student Selector.

But we're _in_ `StudentSelector`'s tests, right? Not quite following...

lms/static/styles/components/_BasicLtiLaunchApp.scss
lyzadanger: Definitely want to get eyes other than mine on this one...gives me pause.

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
LMS007: aws, this comment was put on the wrong file, it was supposed to be on LMSGrader-test and I have since refactored that to not need the real child component.

lms/static/images/arrow-left.svg
LMS007: https://github.com/hypothesis/lms/issues/1014

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -45,6 +46,8 @@ export default function BasicLtiLaunchApp() {
     authToken,
     authUrl,
     lmsName,
+    grading,
```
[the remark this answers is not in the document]
LMS007: Let me know, we can try to nest it under a single parent and check for existence

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -179,9 +184,31 @@ export default function BasicLtiLaunchApp() {
...
 
   if (ltiLaunchState.state === 'fetched-url') {
-    return (
-      <iframe width="100%" height="100%" src={ltiLaunchState.contentUrl} />
+    const iFrame = (
+      <iframe
+        key={sideBarKey}
+        width="100%"
+        height="100%"
+        className="js-via-iframe"
```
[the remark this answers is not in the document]
LMS007: Just gives us a sass style to adjust the padding so the sidebar dose not go over the lms grader, nothing else.

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,85 @@
+import { createElement } from 'preact';
+import propTypes from 'prop-types';
+import { useEffect, useState } from 'preact/hooks';
+
+import StudentSelector from './StudentSelector';
+import updateClientConfig from '../utils/update-client-config';
+
+/**
+ * The LMS Grader menu that is fixed at the top of the page. This menu shows which assignment is currently
```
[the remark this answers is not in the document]
LMS007: More like a tool bar

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,85 @@
...
+   */
+  const onSelectStudent = studentIndex => {
+    setCurrentStudent(studentIndex);
+    onChangeSelectedUser(students[studentIndex].userid);
+  };
+
+  return (
+    <div className="LMSGrader">
+      <header>
+        <ul className="LMSGrader_flex-container">
```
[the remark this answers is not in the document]
LMS007: That is incorrect and I'm going to change this name to make it have more meaning

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,85 @@
...
+  const renderStudentCount = () => {
+    if (currentStudent >= 0) {
+      return <span>{`${currentStudent + 1}/${students.length}`}</span>;
+    } else {
+      return <span>{`${students.length} students`}</span>;
+    }
+  };
+
+  /**
+   * Callback to set the current selected student and fires a callback
```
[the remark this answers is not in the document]
LMS007: Thinking about this a bit more, I think I'm gonna move that second call to the useEffect above, that works a bit nicer. I'll fix the comment too

lms/static/scripts/frontend_apps/components/StudentSelector.js
LMS007: Perfect

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -224,4 +240,31 @@ describe('BasicLtiLaunchApp', () => {
 
     assert.notCalled(fakeApiCall);
   });
+
+  context('when lmsGrader is mode is true', () => {
```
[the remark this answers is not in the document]
LMS007: but I gotta fix my grammar!

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -224,4 +240,31 @@ describe('BasicLtiLaunchApp', () => {
...
+      };
+    });
+
+    it('renders the LMSGrader component', () => {
+      const wrapper = renderLtiLaunchApp();
+      const LMSGrader = wrapper.find('FakeLMSGrader');
+      assert.isTrue(LMSGrader.exists());
+    });
+
+    it('creates an iframe key equal to the focused username', () => {
```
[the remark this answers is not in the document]
LMS007: well the sidebar uses `username` and we need to translate from userid to username

We could change the client to user userid down the road?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
LMS007: Its not the click so much, but rather testing that the onSelectStudent function does what its supposed to do. I'll reword this

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: Yep, its fixed, mostly just needs a new comment.

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
LMS007: awe, I messed that up!

lms/static/scripts/frontend_apps/components/LMSGrader.js
lyzadanger: Thoughts on this one? I notice that presently, there's no padding on the left here, either:

image

I think it would be ideal just to not render these in this round of work—we can add back when we have values to show?

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -0,0 +1,90 @@
...
+   * Callback to set the current selected student.
+   */
+  const onSelectStudent = studentIndex => {
+    setCurrentStudentIndex(studentIndex);
+  };
+
+  return (
+    <div className="LMSGrader">
+      <header>
+        <ul className="LMSGrader__grading-components">
```
lyzadanger: Updated CSS class names looking good.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A student navigation tool which shows a student selection list and a previous and next button.
+ */
+
+export default function StudentSelector({
+  onSelectStudent,
+  selectedStudentIndex,
+  students,
+}) {
+  // Disable the next button if at the end of the list. The length is equal to
```
lyzadanger: I like the current behavior of being able to page through all students as well as being able to click back into "All Students." These comments here are helpful. Looking at the slightly fiddly arithmetic (which warrants the comments), I'm wondering if the logic here isn't about having next or previous _students_, but having next or previous _views_. i.e. this is more about paging forward and back overall. Something as simple as renaming these values might bring a little more clarity, e.g. `hasNextView`, `hasPrevView`. Or maybe that's _too_ generic. I'll leave it up to you!

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+  const hasPrevStudent = selectedStudentIndex >= 0;
+
+  /**
+   * Select the next student in the list.
+   */
+  const onNextStudent = () => {
+    onSelectStudent(selectedStudentIndex + 1);
+  };
+  /**
+   * Select the previous student in the list.
```
lyzadanger: Do we need some sort of commenting here about how, presumably, when we get to student index `0`, this will be an "all students" view?

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  /**
+   * Build the <select> list from the current array of students.
+   */
+  const buildStudentList = () => {
+    const options = students.map((student, i) => (
+      <option
+        key={`student-${i}`}
+        selected={selectedStudentIndex === i}
```
lyzadanger: `selected` is a Boolean attribute. Its presence indicates that the `option` is selected. It shouldn't take a value (its absence represents `false`). (I may have missed this on my last review—apologies!)

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+        value={i}
+      >
+        {student.displayName}
+      </option>
+    ));
+    options.unshift(
+      <option
+        key={'all-students'}
+        selected={selectedStudentIndex === -1}
```
lyzadanger: Ditto on `selected` here.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+        value={-1}
+      >
+        All Students
+      </option>
+    );
+
+    return (
```
lyzadanger: This function might feel a little cleaner if its responsibility were to return an Array of `option` elements, and let the main render function build the `select`.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+      >
+        {options}
+      </select>
+    );
+  };
+
+  return (
+    <div className="StudentSelector">
+      <button
+        aria-label="previous student"
```
lyzadanger: Nit: This label is inaccurate when at student index `0`.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+        disabled={!hasPrevStudent}
```
lyzadanger: `disabled` is also a Boolean attribute.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+   * Select the previous student in the list.
+   */
+  const onPrevStudent = () => {
+    onSelectStudent(selectedStudentIndex - 1);
+  };
+
+  /**
+   * Build the <select> list from the current array of students.
+   */
+  const buildStudentList = () => {
```
lyzadanger: Should this be an IIFE instead of a function invoked later?

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -0,0 +1,95 @@
...
+        aria-label="previous student"
+        disabled={!hasPrevStudent}
+        onClick={onPrevStudent}
+      >
+        <img src="/static/images/arrow-left.svg" />
+      </button>
+      <div className="StudentsSelector__student">{buildStudentList()}</div>
+      <button
+        aria-label="next student"
+        disabled={!hasNextStudent}
```
lyzadanger: Boolean attr again here.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -0,0 +1,78 @@
...
+    renderGrader();
+    sinon.assert.calledWith(
+      fakeUpdateClientConfig,
+      sinon.match({
+        focus: {},
+      })
+    );
+  });
+
+  it('changes the sidebar config to focus to the specified user when onSelectStudent is called with a valid user index', () => {
```
lyzadanger: What happens if `onSelectStudent` is called with an invalid user index?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,90 @@
...
+      />
+    );
+  };
+
+  it('shall have "All Students" as the default option', () => {
+    const wrapper = renderSelector({ selectedStudentIndex: -1 });
+    assert.equal(wrapper.find('select [selected=true]').text(), 'All Students');
+  });
+
+  it('sets the selected option should to the second user', () => {
```
lyzadanger: I can't make sense of this test description :)

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,90 @@
...
+    const wrapper = renderSelector({ selectedStudentIndex: 1 });
+    assert.equal(wrapper.find('select [selected=true]').text(), 'User 2');
+  });
+
+  it('calls the onSelectStudent callback when the select list is changed', () => {
```
lyzadanger: Oh! This test description threw me for a minute. Perhaps more accurate/clear to say something like `'...when the select input value changes'` or something?

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
```diff
@@ -0,0 +1,90 @@
...
+    const onChange = sinon.spy();
+    const wrapper = renderSelector({ onSelectStudent: onChange });
+    wrapper
+      .find('button')
+      .last()
+      .simulate('click');
+    assert.isTrue(onChange.calledWith(1));
+  });
+
+  it('calls the callback with the next student index when clicking the previous button', () => {
```
lyzadanger: I think you mean `previous student index` here?

lms/static/scripts/frontend_apps/components/StudentSelector.js
LMS007: I like this, since the first slot is not not really a student, this makes more sense.

lms/static/scripts/frontend_apps/components/StudentSelector.js
LMS007: yes

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: That bit of logic is handled in the `useEffect` on the state var change, and handled in the first test in this file. I can call out a unique test for this, but it won't test any more lines of actau code but may make logical sense when reading the tests.

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
LMS007: cleaned it up

lms/static/scripts/frontend_apps/components/test/StudentSelector-test.js
LMS007: yes, fixed!

PR #1018, opened 2019-10-03
Add `make devdata`

seanh:
Depends on https://github.com/hypothesis/devdata/pull/2

This adds a `make devdata` command that:

1. Clones the https://github.com/hypothesis/devdata/ repo to a temporary directory. (This will fail if you don't have your git ssh keys set up and access to the private repo.)

2. Reads an `lms/devdata.json` file from the private repo, and creates or updates application instances in the DB according to the contents of the JSON file.

3. Copies an `lms/devdata.env` file from the private repo to a git-ignored location in your h work tree. This file is read by tox and the envvars it contains are set in the tox envs.

This gets lms working with your local dev instances of h, via and the client (pending corresponding h, via and client PRs being merged) and also gets the Google Drive integration working, and gets lms working with various "localhost (make devdata)" application instances and assignments that have been created in https://hypothesis.instructure.com/

It doesn't get Blackboard or other LMS's working yet


lms/config/__init__.py
```diff
@@ -14,11 +14,11 @@ def configure(settings):
...
-        "via_url": sg.get("VIA_URL", required=True),
-        "jwt_secret": sg.get("JWT_SECRET", required=True),
-        "google_client_id": sg.get("GOOGLE_CLIENT_ID", required=True),
-        "google_developer_key": sg.get("GOOGLE_DEVELOPER_KEY", required=True),
-        "google_app_id": sg.get("GOOGLE_APP_ID", required=True),
+        "via_url": sg.get("VIA_URL"),
+        "jwt_secret": sg.get("JWT_SECRET"),
+        "google_client_id": sg.get("GOOGLE_CLIENT_ID"),
+        "google_developer_key": sg.get("GOOGLE_DEVELOPER_KEY"),
+        "google_app_id": sg.get("GOOGLE_APP_ID"),
```
seanh: The `required=True` makes it impossible to run the Pyramid app at all (e.g. to run a command line script to set those envvars) without those envvars already being set. I've just removed the concept of a required envvar that crashes on startup unless set. It's unnecessary and has only gotten in the way of people setting up their envs manually, and now gets in the way of scripting

conf/development.ini
```diff
@@ -24,6 +24,13 @@ feature_flags_cookie_secret = "notasecret"
 # The secret string that's used to sign the OAuth2 state param.
 oauth2_state_secret = "notasecret"
 
+via_url = http://localhost:9080
+h_authority = lms.hypothes.is
+h_api_url_public = http://localhost:5000/api/
+h_api_url_private = http://localhost:5000/api/
+rpc_allowed_origins = http://localhost:5000
+sentry_environment = dev
```
seanh: These aren't sensitive at all so should just have sensible dev defaults in the config file

setup.py
```diff
@@ -21,4 +21,8 @@
     packages=find_packages(),
     include_package_data=True,
     zip_safe=False,
+    entry_points="""\
+    [console_scripts]
+    devdata = lms.scripts:devdata
+    """,
```
seanh: This is the standard way of adding command line scripts to a Pyramid app, the same way as builtin scripts like pshell etc work. The `bin/hypothesis` thing we have in h is... :man_shrugging:

README.markdown
```diff
@@ -185,11 +101,6 @@ You should now be able to run the tests successfully by typing:
 
     make test
 
-### Further steps
-
-* [Getting the Google Drive integration working locally](docs/google-drive.md)
-* [Installing your local lms app in Canvas](docs/canvas.md)
```
robertknight: Are you sure you want to get rid of *all* of this documentation? Even if some of it is no longer mandatory to know when getting the app ready for the first time, it looks like some of the information in there could still be useful down the line.

lms/scripts/devdata.py
```diff
@@ -0,0 +1,111 @@
...
+from pyramid.paster import bootstrap
+
+import lms
+from lms import models
+
+
+__all__ = ["devdata"]
+
+
+class DevDataFactory:
```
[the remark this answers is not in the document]
seanh: Yes. This is a problem that we have with a lot of tooling that's duplicated between repos

lms/config/__init__.py
seanh: I'm just removing it because it's more trouble than it's worth. With the new `development.ini` defaults on this branch, plus `make devdata`, it's indeed not needed in dev if you have access to devdata. I don't think this app is of much interest outside of our team, so not so worried about people who don't have devdata (they will now be able to launch the app with required settings not set, and it will crash when it tries to use them). In a QA or prod deploy it will be possible to launch the app with required settings missing and it'll crash when it tries to use them, but only two QA/prod deploys exist and both have all the settings already. In both of those cases we do have manual setup docs that tell you to set up all these settings (those docs do seem unlikely to be maintained, though). So it's just not providing much benefit and it gets in the way.

README.markdown
seanh: I've put most of this back, re-phrased as "How to set up manually" if you can't use devdata for some reason (e.g. doing a prod setup). Seems unlikely to be maintained unfortunately

PR #1019, opened 2019-10-03
Automate installing docker-compose and creating the lms_test DB

seanh:
Just copying two things that we've already had in h for a while, over to lms: install docker-compose automatically in tox so that developers don't have to go hunting through the Docker docs trying to find out how to install it on their system. And create the test DB automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,9 @@ help:
 	@echo "                       dependencies, etc)"
 
 .PHONY: services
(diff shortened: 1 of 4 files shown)


seanh:
I'm gonna go ahead and merge this as it's needed for Monday and is just a copy-paste from h. If there are any problems with it we'll find them on Monday

PR #1050, opened 2019-10-15
SvgIcon component

LMS007:
- Copy svg icon component from the client
- One key difference is, unlike the client, this does not use the `name` prop to reference pre-imported svg. Instead the svg icon is to be imported by the caller and passed to the srec prop.
- Added SvgIcon to StudentSelector
- Added `stringify` to the browserify transform config to import svg files as strings.


yarn.lock
```diff
@@ -126,16 +126,16 @@
     "@babel/types" "^7.0.0"
 
 "@babel/helper-module-transforms@^7.1.0", "@babel/helper-module-transforms@^7.4.4":
-  version "7.4.4"
-  resolved "https://registry.yarnpkg.com/@babel/helper-module-transforms/-/helper-module-transforms-7.4.4.tgz#96115ea42a2f139e619e98ed46df6019b94414b8"
-  integrity sha512-3Z1yp8TVQf+B4ynN7WoHPKS8EkdTbgAEy0nU0rs/1Kw4pDgmvYH3rz3aI11KgxKCba2cn7N+tqzV1mY2HMN96w==
+  version "7.5.5"
```
robertknight: Hmm, it seems surprising to me that the `yarn.lock` file would have changed so much from adding the `stringify` dependency. I'm guessing that at some point you removed and re-generated the `yarn.lock` file while trying to resolve the Jenkins issue? Can you reset the `yarn.lock` file to the version from the commit where you branched from master (`git checkout <commit> yarn.lock`) and then re-run `yarn install` which should then only apply the changes related to the `stringify` dependency.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -72,15 +74,21 @@ export default function StudentSelector({
         disabled={!hasPrevView}
         onClick={onPrevView}
       >
-        <img src="/static/images/arrow-left.svg" />
+        <SvgIcon
+          src={require('../../../images/arrow-left.svg')}
```
robertknight: The suggestion I mentioned on the call about handling the SVG markup here was to provide a way for `SvgIcon` to verify that the markup comes from a trusted source which is known not to include any inline event/styles etc. Unlike `` the browser will apply them in this context.

It looks like the `stringify` transform doesn't mark the string in any way to indicate that it came from the JS bundle, so I was thinking along the lines of a helper function to tag the markup. For example, something like:

```js

/**
 * Wrap `markup` to indicate that it came from a trusted source
 * (ie. not unsanitized user input)
 *
 * Returns an object with a `trustedHTML` property whose value is `markup`.
 */
function trustMarkup(markup) {
  return { trustedHTML: markup };
}

// in SvgIcon.js ...

function SvgIcon({ src }) {
  const markup = {
    __html: src.trustedHTML,
  };
}

// in a module that uses SvgIcon ...

<SvgIcon
  src={trustMarkup(require('path/to/icon.svg'))}
/>
```

LMS007:
May be helpful for someone to double check the yarn file. I tried rebuilding it but it still looks suspect.

`yard add stringify`

lms/static/scripts/frontend_apps/utils/test/trusted-test.js
```diff
@@ -0,0 +1,12 @@
+import { trustMarkup } from '../trusted';
+
+describe('trustMarkup', () => {
+  it('returns the provided object wrapped in a key called `trustedHTML`', async () => {
```
robertknight: The `async` isn't needed here.

LMS007: copy paste error, good catch!

PR #1056, opened 2019-10-16
Use pyramid_exclog in dev

seanh:
Use pyramid_exclog to print exceptions and related information to the
terminal in dev, not just in production.

Also removed pyramid_debugtoolbar in dev, use pyramid_exclog's terminal
printing rather than pyramid_debugtoolbar's HTML interface for
debugging.

Also remove the logic that prevented error views from being registered
(and therefore nice error pages from being shown) in dev.

This makes dev the same as production when an error happens:

1. The proper error page will be rendered in the browser
2. The traceback and additional info will be printed to the terminal (in
   production this same output should appear in Papertrail)

### DIFF
--- a/lms/app.py
+++ b/lms/app.py
@@ -10,6 +10,15 @@ def configure_jinja2_assets(config):
 def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config = configure(settings=settings)
 
+    # Make sure that pyramid_exclog's tween runs under pyramid_tm's tween so
+    # that pyramid_exclog doesn't re-open the DB session after pyramid_tm has
+    # already closed it.
+    config.add_tween(
+        "pyramid_exclog.exclog_tween_factory", under="pyramid_tm.tm_tween_factory"
+    )
+    config.add_settings({"exclog.extra_info": True})
+    config.include("pyramid_exclog")
+
     config.include("pyramid_jinja2")
     config.include("pyramid_services")
     config.include("pyramid_tm")
--- a/lms/views/error.py
+++ b/lms/views/error.py
@@ -1,8 +1,5 @@
-import os
-
 from pyramid import httpexceptions
 from pyramid import i18n
-from pyramid.settings import asbool
 from pyramid.view import forbidden_view_config, notfound_view_config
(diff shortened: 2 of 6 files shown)


lms/app.py
```diff
@@ -10,6 +10,9 @@ def configure_jinja2_assets(config):
 def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config = configure(settings=settings)
 
+    config.add_settings({"exclog.extra_info": True})
+    config.include("pyramid_exclog")
+
```
jon-betts: Should this also contain the tween ordering to ensure that exclog is under pyramid_tm?

lms/views/error.py
```diff
@@ -73,16 +70,6 @@ def error(request):
...
-    debug = asbool(os.environ.get("DEBUG") or config.registry.settings.get("debug"))
-    if debug:
-        # Don't register the error pages in development environments.  Let
-        # pyramid_debugtoolbar show the traceback in the browser and terminal
-        # instead.
-        # If you want to test the error pages in your dev env you can set the
-        # environment variable DEBUG to false:
-        #     export DEBUG=false
-        return
-
```
jon-betts: :+1: Random different behavior in dev makes testing a nightmare

lms/app.py
seanh: Hmm, yes probably

lms/views/error.py
seanh: Yeah this has been a continual PITA

lms/app.py
seanh: Done. ptweens:

```
"pyramid.tweens" config value NOT set (implicitly ordered tweens used)

Implicit Tween Chain

Position    Name                                                             
--------    ----                                                             
-           INGRESS                                                          
0           pyramid_tm.tm_tween_factory                                      
1           pyramid.tweens.excview_tween_factory                             
2           pyramid_exclog.exclog_tween_factory                              
-           MAIN                                                             
___________________________________ summary ____________________________________
  py36-dev: commands succeeded
  congratulations :)
```

PR #1068, opened 2019-10-24
Add isort to `make format`

seanh:


tests/lms/conftest.py
```diff
@@ -1,17 +1,17 @@
...
+
 import jwt
+import pytest
 import sqlalchemy
-from sqlalchemy.orm import sessionmaker
 from pyramid import testing
 from pyramid.request import apply_request_extensions
+from sqlalchemy.orm import sessionmaker
 
+import httpretty
```
robertknight: `httpretty` may need to be added to the list of known third-party libs

.isort.cfg
```diff
@@ -0,0 +1,13 @@
...
+[settings]
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=88
+known_third_party=dateutil,factory,faker,hypothesis,pytest,slugify,webtest,zope
+# isort crashes on app.py with "file contains syntax errors" even though the
+# file is fine. Just skip it.
+skip=lms/app.py
```
robertknight: Any idea what triggers the issue in `lms/app.py`?

Issue #1030, opened 2019-10-08
Change lms to use via2.hypothes.is 


hmstepanek:
# Problem:
Now that we have a via replacement for lms pdf we need to change the link in production to use the new service.

# Release Plan:
- [x] Release the use of the new service url under a feature flag in lms
- [x] Release to prod by the end of the sprint

hmstepanek:
Feature flag info for lms:
Here's the docstring for how to read a feature flag in lms: https://github.com/hypothesis/lms/blob/6c60cc6026da46db4874da2bfc5881d07c94390a/lms/extensions/feature_flags/__init__.py#L14-L28
This is where the via_url (which comes from the VIA_URL envvar) is read: https://github.com/hypothesis/lms/blob/51db0594e07b172455b5c7cd54fc43f5064d9972/lms/views/helpers/_via.py#L47
So there you could check request.feature("via2") and read either the "via_url" or the "via2_url" depending on the feature flag.
If you also add your new feature flag's name to the FEATURE_FLAGS_ALLOWED_IN_COOKIE envvar (https://github.com/hypothesis/lms/blob/6c60cc6026da46db4874da2bfc5881d07c94390a/lms/extensions/feature_flags/__init__.py#L110-L113) then it will appear on the https://lms.hypothes.is/flags page for toggling on and off

PR #1069, opened 2019-10-25
Switch to via2 permanently & remove feature flag

hmstepanek:
This is the final part of https://github.com/hypothesis/lms/issues/1030.


robertknight:
Are there any issues arising from our security discussion yesterday which need to be addressed before we can make this the default?

hmstepanek:
I'm implementing the ip blocking for proxying certain internal urls now (aws metadata + internal). I'm going to premise this with I'm not a security expert but I think the risk is mainly that someone gets information/secrets out of the server by proxying particular internal secret urls not that they inject malicious malware into our server via proxying urls so the security lockdown we need to implement should not be a blocker for making the switch to the new service permanent. Unless of course they are able to get a secret our of the server and use that to log into the server and inject malware. As far as I know that's not possible though.

seanh:
@hmstepanek Needs rebase. Ping me after this is rebased and I'll test it---I just want to make double-sure that assignments of all types are working across all LMS's, and that grading is still working as well

PR #1082, opened 2019-10-30
Remove U+FEFF BOM from lms/app.py

seanh:
Remove a bite order mark from `app.py`. This shouldn't be here and I'm hoping it was the cause of occasional isort crashes on app.py that I've been seeing


seanh:
Merged into https://github.com/hypothesis/lms/pull/1083

Issue #918, opened 2019-08-16
Make it possible to find out which school/LMS instance an LMS-created group is associated with


judell:
We can query the group table for groups created in the lms authority, but can't know the lms domain.

Says @jeremydean:

> i have no idea what school they're at and just google around or use my own rolodex 
> to figure it out

Adding the lms domain we got from https://lms.hypothes.is/welcome to the group table would solve the problem.

klemay:
## User story:

As a Hypothesis success/edu team member, I want to check on the engagement of our pilot schools (and those not currently in a pilot) with the LMS app. I want to see:

- How many groups were created at the school
- How many annotations are in those groups
- Which groups have a lot of annotations, which groups barely have any, etc

## The current state of things
At present, I can query Metabase for private groups created in the lms.hypothes.is authority. The query returns:
- Group ID
- Group name
- Creation date
- Creator ID
- Creator Name
- Creator Username
- Users in group
- Annotations in group

In order to know which school each group belongs to, I have to Google the `Creator name`, sometimes adding the `Group name` to narrow it down. Often, I'm not able to figure out which school the group creator is from at all. This makes it impossible to follow up on leads.

## The request
Add the LMS domain for LMS groups to Metabase.

judell:
> Add the LMS domain for LMS groups to Metabase.

I would say, rather, as in the original statement:

> Adding the lms domain we got from https://lms.hypothes.is/welcome to the group table 

i.e. to the group table in postgres, so we can see it through the lens of metabase.

robertknight:
I discussed this issue with Lyza and Sean yesterday. This is not purely an operations task as we may need to modify the LMS app to store additional metadata with the groups it creates. It is possible to link a group with an LMS domain given both the h and lms databases but it involves quite a few joins at present:

1. Look up the `user_id` of at least one member in the group in the `user_group` table in h
2. Look up the `provider` associated with that `user_id` (there should be at least one) in the `user_identity` table.
3. Look up the application instance in the `application_instances` lms table, using the `provider` as a `consumer_key` 
4. Finally, get the `lms_url` and `requesters_email` attributes of that the application instance

[...]

PR #1088, opened 2019-10-30
Record group info on LTI launch

seanh:
Depends on this DB migration: https://github.com/hypothesis/lms/pull/1087

The issue:

* This PR fixes <https://github.com/hypothesis/lms/issues/918>, which asks for a way to associate an LMS-created group found in the h DB with school/institution or LMS that the group belongs to

Testing:

* To test this PR you need to launch your local instance of the LMS app from an LMS such as Canvas. You can use the various pre-created "localhost" assignments in "Test Course 1" in https://hypothesis.instructure.com/courses/83 to do this. See https://github.com/hypothesis/lms/wiki/Testing-the-LTI-App-in-Canvas

Background:

* Whenever a teacher or student launches an assignment using our LMS app, the LMS app creates a group in h for the LMS course that the assignment belongs to. The h group has the same name as the LMS course
* There's a 1:1 mapping between LMS courses and h groups
* The LMS app uses [h's groups API](https://h.readthedocs.io/en/latest/api-reference/#tag/groups) to create the group
* One of the parameters that h's groups API supports is `groupid`. This is unique identifier for a group that is provided by the caller (in this case: the LMS app). The LMS app generates a unique ID based on the LMS course and other information it receives from the LMS, and passes it to the h API as the groupid parameter.
* An h groupid looks like this: `"group:<AUTHORITY_PROVIDED_ID>@<AUTHORITY>"`. In the case of the LMS app `<AUTHORITY>` is `lms.hypothes.is`. The `<AUTHORITY_PROVIDED_ID>` is unique within a given authority, and appending `@<AUTHORITY>` makes groupids globally unique
* The h DB actually [stores the `<AUTHORITY_PROVIDED_ID>` part only](https://github.com/hypothesis/h/blob/cfd3d2af55714273bf8984045f842c887ad2097b/h/models/group.py#L79-L82) in the groups table, _not_ the full `"group:<AUTHORITY_PROVIDED_ID>@<AUTHORITY>"`" groupid (the full groupid can be generated from the `authority` and `authority_provided_id` columns of h's `group` table
* The LMS app does not keep any record of what groups it has and has not created in h, or what the groupids of those groups are. It is always able to generate the groupid for the current course on every single launch, based on the LTI launch parameters that the LMS passes to us. And it always calls the h API to upsert a group with that ID, on every single launch. This is to avoid nasty synchronization issues between the h and LMS DBs. Groups and their IDs are stored in the h DB only. (Also, it _is_ necessary to update the group on every launch, not just to create it once, because the course might have been changed in the LMS. For example if the teacher changes the course's title, we update the name of the h group.)

What this PR does:

* This PR adds a new `group_info` table to the LMS DB which maps the groupids that the LMS app has generated to a bunch of information about the school/institution/LMS that the groupid was generated for
* On every LTI launch the LMS app upserts a row into the `group_info` table for this launch's groupid
* This `group_info` is for analytics purposes only and isn't ever read by the app except to upsert into it. It remains the case that, as far as the LMS app is concerned, the h DB is the only place where groups and their IDs exist, and the only source of truth about whether a given group has already been created or not

IntegrityErrors:

* If two concurrent requests try to create the same `group_info` row at the same time then one of them will fail with an IntegrityError. If this happens the failed request will be retried, and on the second attempt it will update the row rather than creating it, and will succeed

For example this diff will change the code to simulate conflicting concurrent requests:

```diff
diff --git a/lms/services/group_info_upsert.py b/lms/services/group_info_upsert.py
index 6a13feb8..70653c67 100644
--- a/lms/services/group_info_upsert.py
+++ b/lms/services/group_info_upsert.py
@@ -4,6 +4,9 @@ from lms.models import GroupInfo
 __all__ = ["GroupInfoUpsert"]
 
 
+FIRST_TIME = True
+
+
 class GroupInfoUpsert:
     """
     A callable that upserts :class:`lms.models.GroupInfo` records.
@@ -34,6 +37,21 @@ class GroupInfoUpsert:
         )
 
         if not group_info:
+            global FIRST_TIME
+
+            if FIRST_TIME:
+                FIRST_TIME = False
+                # A concurrent request adds a conflicting GroupInfo.
+                # This will cause the current request to raise an
+                # IntegrityError when pyramid_tm tries to commit the DB
+                # session.
+                self._db.add(
+                    GroupInfo(
+                        authority_provided_id=authority_provided_id,
+                        consumer_key=consumer_key,
+                    )
+                )
+
             group_info = GroupInfo(
                 authority_provided_id=authority_provided_id, consumer_key=consumer_key
             )
```

You'll then need to make sure there is no existing `group_info` row in your DB (since the IntegrityError cannot happen if the row already exists):

```
$ make sql
postgres=# delete from group_info;
```

Then launch one of the localhost Canvas assignments. The launch should succeed, and the group_info will have been created in the DB. The IntegrityError will be printed to the terminal but will be filtered out from Sentry reporting since it's retryable.

### Example SQL query

All LMS-created h groups in the h DB have a unique `authority_provided_id` that is provided to the h API by the LMS app when it creates or updates the group. To retrieve the `authority_provided_id`'s run this query **against the h DB**:

```sql
SELECT authority_provided_id FROM public.group WHERE authority = 'lms.hypothes.is' LIMIT 5;
```

Given an `authority_provided_id` from h you can run this query **against the lms DB** to get more info about the group:

```sql
SELECT * FROM group_info, application_instances
  WHERE group_info.authority_provided_id = '<AUTHORITY_PROVIDED_ID>'
  AND group_info.consumer_key = application_instances.consumer_key;
```

Output:

```
id                                     | 2
authority_provided_id                  | <AUTHORITY_PROVIDED_ID>
consumer_key                           | Hypothesis***
context_id                             | ***
context_title                          | Test Course 1
context_label                          | Test Course 1
tool_consumer_info_product_family_code | canvas
tool_consumer_info_version             | cloud
tool_consumer_instance_name            | University of Hypothesis
tool_consumer_instance_description     | 
tool_consumer_instance_url             | 
tool_consumer_instance_contact_email   | ***
tool_consumer_instance_guid            | ***:canvas-lms
custom_canvas_api_domain               | hypothesis.instructure.com
custom_canvas_course_id                | 83
id                                     | 1
consumer_key                           | Hypothesis***
shared_secret                          | ***
lms_url                                | https://hypothesis.instructure.com
requesters_email                       | ***
created                                | 2019-10-05 12:01:39.773757
developer_key                          | 
developer_secret                       | 
aes_cipher_iv                          | 
provisioning                           | t
```

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -38,6 +38,11 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    #: A list of all the GroupInfo's for this application instance.
+    group_infos = sa.orm.relationship(
+        "GroupInfo", back_populates="application_instance"
+    )
+
     @classmethod
     def build_from_lms_url(  # pylint:disable=too-many-arguments
         cls, lms_url, email, developer_key, developer_secret, encryption_key=None
--- a/lms/models/group_info.py
+++ b/lms/models/group_info.py
@@ -1,0 +1,90 @@
+import sqlalchemy as sa
+
+from lms.db import BASE
+
+__all__ = ["GroupInfo"]
+
+
+class GroupInfo(BASE):
+    """
+    Some info about an LMS group that was created in h.
+
+    This info is stored purely for metrics/analytics purposes and shouldn't be
+    used for application logic. The app should treat the h API as the
+    "single-source of truth" about what h groups exist and what their IDs and
+    other properties are.
+    """
+
+    __tablename__ = "group_info"
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    #: The authority_provided_id of the group in h.
+    #:
+    #: This corresponds to the ID part of the groupid that's used in h's groups
+    #: API. For example if the groupid is "group:SOME_ID@lms.hypothesis.is"
+    #: then the authority_provided_id is the "SOME_ID" part without the leading
+    #: "group:" or the trailing "@lms.hypothes.is".
+    #:
+    #: This also corresponds to the group.authority_provided_id column in h's
+    #: DB.
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False, unique=True)
+
+    #: The LTI consumer_key (oauth_consumer_key) of the application instance
+    #: that this access token belongs to.
+    consumer_key = sa.Column(
+        sa.String(),
+        sa.ForeignKey("application_instances.consumer_key", ondelete="cascade"),
+        nullable=False,
+    )
+
+    #: The ApplicationInstance that this group belongs to.
+    application_instance = sa.orm.relationship(
+        "ApplicationInstance", back_populates="group_infos"
+    )
+
+    #: The value of the context_id param this group was last launched with.
+    context_id = sa.Column(sa.UnicodeText())
+
+    #: The value of the context_title param this group was last launched with.
+    context_title = sa.Column(sa.UnicodeText())
+
+    #: The value of the context_label param this group was last launched with.
+    context_label = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_info_product_family_code param this group was last launched with.
+    tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_info_version param this group was last launched with.
+    tool_consumer_info_version = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_instance_name param this group was last launched with.
+    tool_consumer_instance_name = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_instance_description param this group was last launched with.
+    tool_consumer_instance_description = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_instance_url param this group was last launched with.
+    tool_consumer_instance_url = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_instance_contact_email param this group was last launched with.
+    tool_consumer_instance_contact_email = sa.Column(sa.UnicodeText())
+
+    #: The value of the tool_consumer_instance_guid param this group was last launched with.
+    tool_consumer_instance_guid = sa.Column(sa.UnicodeText())
+
+    #: The value of the custom_canvas_api_domain param this group was last launched with.
+    custom_canvas_api_domain = sa.Column(sa.UnicodeText())
+
+    #: The value of the custom_canvas_course_id param this group was last launched with.
+    custom_canvas_course_id = sa.Column(sa.UnicodeText())
+
+    def __str__(self):
+        return self.__repr__(_class_name="GroupInfo")
+
+    def __repr__(self, _class_name="lms.models.GroupInfo"):
+        return (
+            f"<{_class_name}"
+            f" authority_provided_id:'{self.authority_provided_id}'"
+            f" consumer_key:'{self.consumer_key}'>"
+        )
--- a/lms/resources/_lti_launch.py
+++ b/lms/resources/_lti_launch.py
@@ -73,6 +73,32 @@ class LTILaunchResource:
         return display_name
 
     @property
+    def h_authority_provided_id(self):
+        """
+        Return a unique h authority_provided_id for the request's group.
+
+        The authority_provided_id is deterministic and is unique to the LTI
+        course. Calling this function again with params representing the same
+        LTI course will always return the same authority_provided_id. Calling
+        this function with different params will always return a different
+        authority_provided_id.
+
+        :raise HTTPBadRequest: if an LTI param needed for generating the
+            authority_provided_id is missing
+        """
+        # Generate the authority_provided_id based on the LTI
+        # tool_consumer_instance_guid and context_id parameters.
+        # These are "recommended" LTI parameters (according to the spec) that in
+        # practice are provided by all of the major LMS's.
+        # tool_consumer_instance_guid uniquely identifies an instance of an LMS,
+        # and context_id uniquely identifies a course within an LMS. Together they
+        # globally uniquely identify a course.
+        hash_object = hashlib.sha1()
+        hash_object.update(self._get_param("tool_consumer_instance_guid").encode())
+        hash_object.update(self._get_param("context_id").encode())
+        return hash_object.hexdigest()
+
+    @property
     def h_groupid(self):
         """
         Return a unique h groupid for the current request.
@@ -84,20 +110,10 @@ class LTILaunchResource:
         return the same groupid. Calling this function with different params will
         always return a different groupid.
 
-        :raise HTTPBadRequest: if an LTI param needed for generating the group
-          name is missing
+        :raise HTTPBadRequest: if an LTI param needed for generating the
+            groupid is missing
         """
-        # Generate the groupid based on the LTI tool_consumer_instance_guid and
-        # context_id parameters.
-        # These are "recommended" LTI parameters (according to the spec) that in
-        # practice are provided by all of the major LMS's.
-        # tool_consumer_instance_guid uniquely identifies an instance of an LMS,
-        # and context_id uniquely identifies a course within an LMS. Together they
-        # globally uniquely identify a course.
-        hash_object = hashlib.sha1()
-        hash_object.update(self._get_param("tool_consumer_instance_guid").encode())
-        hash_object.update(self._get_param("context_id").encode())
-        return f"group:{hash_object.hexdigest()}@{self._authority}"
+        return f"group:{self.h_authority_provided_id}@{self._authority}"
 
     @property
     def h_group_name(self):
--- a/lms/services/group_info_upsert.py
+++ b/lms/services/group_info_upsert.py
@@ -1,0 +1,45 @@
+"""A service that upserts :class:`lms.models.GroupInfo` records."""
+from lms.models import GroupInfo
+
+__all__ = ["GroupInfoUpsert"]
+
+
+class GroupInfoUpsert:
+    """
+    A callable that upserts :class:`lms.models.GroupInfo` records.
+
+    Usage:
+
+        group_info_upsert = request.find_service(name="group_info_upsert")
+        group_info_upsert(authority_provided_id, consumer_key, **kwargs)
+    """
+
+    def __init__(self, _context, request):
+        self._db = request.db
+
+    def __call__(self, authority_provided_id, consumer_key, **kwargs):
+        """
+        Upsert a row into the group_info DB table.
+
+        Find the existing group_info with the given ``authority_provided_id``
+        and update it with the given ``consumer_key`` and ``kwargs``.
+
+        If no group_info with the given ``authority_provided_id`` exists then
+        create one.
+        """
+        group_info = (
+            self._db.query(GroupInfo)
+            .filter_by(authority_provided_id=authority_provided_id)
+            .one_or_none()
+        )
+
+        if not group_info:
+            group_info = GroupInfo(
+                authority_provided_id=authority_provided_id, consumer_key=consumer_key
+            )
+            self._db.add(group_info)
+
+        group_info.consumer_key = consumer_key
+
+        for name, value in kwargs.items():
+            setattr(group_info, name, value)
--- a/lms/views/decorators/h_api.py
+++ b/lms/views/decorators/h_api.py
@@ -5,8 +5,13 @@ import functools
 from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
 
 from lms.services import HAPIError, HAPINotFoundError
+from lms.views.decorators._helpers import upsert_group_info
 
-__all__ = ["upsert_h_user", "upsert_course_group", "add_user_to_group"]
+__all__ = [
+    "upsert_h_user",
+    "upsert_course_group",
+    "add_user_to_group",
+]
 
 
 def upsert_h_user(wrapped):
@@ -72,7 +77,7 @@ def upsert_course_group(wrapped):
     """
 
     @functools.wraps(wrapped)
-    def wrapper(context, request):
+    def wrapper(context, request):  # pylint:disable=too-many-branches
         if not context.provisioning_enabled:
             return wrapped(context, request)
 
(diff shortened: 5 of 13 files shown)


lms/models/application_instance.py
```diff
@@ -38,6 +38,11 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    #: A list of all the GroupInfo's for this application instance.
+    group_infos = sa.orm.relationship(
+        "GroupInfo", back_populates="application_instance"
+    )
```
seanh: We're not actually making use of this but this just means that, given an `ApplicationInstance` model object instance, `my_application_instance.group_infos` will be a list of all its related `GroupInfo`s.

lms/resources/_lti_launch.py
```diff
@@ -97,7 +96,24 @@ def h_groupid(self):
         hash_object = hashlib.sha1()
         hash_object.update(self._get_param("tool_consumer_instance_guid").encode())
         hash_object.update(self._get_param("context_id").encode())
-        return f"group:{hash_object.hexdigest()}@{self._authority}"
+        return hash_object.hexdigest()
```
seanh: I split the existing `h_groupid` property into two separate properties `h_authority_provided_id` and `h_groupid`. See commit messages

lms/services/group_info_upsert.py
```diff
@@ -0,0 +1,45 @@
...
+        if not group_info:
+            group_info = GroupInfo(
+                authority_provided_id=authority_provided_id, consumer_key=consumer_key
+            )
+            self._db.add(group_info)
+
+        group_info.consumer_key = consumer_key
+
+        for name, value in kwargs.items():
+            setattr(group_info, name, value)
```
seanh: This method is the one that can cause `IntegrityError`s. It won't itself raise an `IntegrityError`, but it can modify the DB session in such a way that `IntegrityError` will be raised later. It's kind of nice that code that does this doesn't need to worry about it. For example it doesn't need to itself flush the DB and catch the `IntegrityError`. Just let it happen, and the request will be retried.

If the `IntegrityError` that's being triggered happens 3 times in a row, i.e. it happens on each retry, for example because it's a deterministic rather than transient error, then the app will crash and send an error response to the client and a crash report to Sentry, and we have a bug.

lms/views/decorators/h_api.py
```diff
@@ -106,6 +111,41 @@ def wrapper(context, request):
...
+                    "tool_consumer_info_version",
+                    "tool_consumer_instance_name",
+                    "tool_consumer_instance_description",
+                    "tool_consumer_instance_url",
+                    "tool_consumer_instance_contact_email",
+                    "tool_consumer_instance_guid",
+                    "custom_canvas_api_domain",
+                    "custom_canvas_course_id",
+                ]
+            },
```
seanh: This is a dict comprehension that pulls a bunch of params out of the request (or `None` for each param that doesn't exist). This sort of thing belongs in the `lms/validation/` package (and written as a Marshmallow schema) but in this case I judged it not worth the trouble

lms/models/group_info.py
```diff
@@ -0,0 +1,90 @@
...
+    consumer_key = sa.Column(
+        sa.String(),
+        sa.ForeignKey("application_instances.consumer_key", ondelete="cascade"),
+        nullable=False,
+    )
+
+    #: The ApplicationInstance that this group belongs to.
+    application_instance = sa.orm.relationship(
+        "ApplicationInstance", back_populates="group_infos"
+    )
```
seanh: Again, we're not actually using this, but this just means that `my_group_info.application_instance` is the related `ApplicationInstance` object

lms/views/decorators/h_api.py
```diff
@@ -106,6 +111,41 @@ def wrapper(context, request):
     return wrapper
 
 
+def upsert_group_info(wrapped):
+    """Create or update the :class:`lms.models.GroupInfo` for this launch."""
+
+    @functools.wraps(wrapped)
+    def wrapper(context, request):
+        if not context.provisioning_enabled:
+            return wrapped(context, request)
```
seanh: `provisioning_enabled` is a feature flag that's scheduled to be removed soon, so we should soon be able to remove all these `provisioning_enabled`-based `if`'s. For now they're still getting peppered around though

lms/views/basic_lti_launch.py
```diff
@@ -24,6 +24,7 @@
     add_user_to_group,
     report_lti_launch,
     upsert_course_group,
+    upsert_group_info,
```
jon-betts: It seems like we want this everywhere we use `upsert_course_group`. Is the freedom we give ourselves by having them independent just the freedom to make a mistake by including one and not other other?

lms/views/decorators/h_api.py
jon-betts: Truth be told, we likely don't care much about the exact contents here. Only if something causes an exception. In this case the worse that should happen is you get None if they changed the names.

lms/views/decorators/h_api.py
jon-betts: Why do we want to disable this if provisioning is enabled?

lms/views/decorators/h_api.py
```diff
@@ -106,6 +111,41 @@ def wrapper(context, request):
     return wrapper
 
 
+def upsert_group_info(wrapped):
```
jon-betts: As per previous comment, would this be better as a private function in the `upsert_course_group` decorator? It would simplify the diff quite a bit too.

tests/lms/models/group_info_test.py
```diff
@@ -0,0 +1,82 @@
...
+from sqlalchemy.exc import IntegrityError
+
+from lms.models import ApplicationInstance, GroupInfo
+
+
+class TestGroupInfo:
+    def test_persist_and_retrieve(self, application_instance, db_session, group_info):
+        db_session.add(group_info)
+
+        retrieved_group_info = db_session.query(GroupInfo).one()
```
jon-betts: In this case is this getting persisted to a test db, or just in RAM?

lms/models/group_info.py
jon-betts: Ha! On first reading of this in the tests I thought this was an instance of the `application` model. It's a kind of terrible name.

tests/lms/models/group_info_test.py
```diff
@@ -0,0 +1,82 @@
...
+        assert retrieved_group_info.application_instance == application_instance
+
+    def test_authority_provided_id_cant_be_None(
+        self, application_instance, db_session, group_info
+    ):
+        group_info.authority_provided_id = None
+        db_session.add(group_info)
+
+        with pytest.raises(
+            IntegrityError, match='"authority_provided_id" violates not-null constrain'
```
jon-betts:
```suggestion
            IntegrityError, match='"authority_provided_id" violates not-null constraint'
```

tests/lms/models/group_info_test.py
```diff
@@ -0,0 +1,82 @@
...
+            db_session.flush()
+
+    def test_application_instance_cant_be_None(
+        self, application_instance, db_session, group_info
+    ):
+        group_info.application_instance = None
+        db_session.add(group_info)
+
+        with pytest.raises(
+            IntegrityError, match='"consumer_key" violates not-null constrain'
```
jon-betts:
```suggestion
            IntegrityError, match='"consumer_key" violates not-null constraint'
```

tests/lms/models/group_info_test.py
```diff
@@ -0,0 +1,82 @@
...
+            str(group_info)
+            == "<GroupInfo authority_provided_id:'test_authority_provided_id' consumer_key:'test_consumer_key'>"
+        )
+
+    def test___repr__(self, db_session, group_info):
+        db_session.add(group_info)
+        db_session.flush()
+        assert (
+            repr(group_info)
+            == "<lms.models.GroupInfo authority_provided_id:'test_authority_provided_id' consumer_key:'test_consumer_key'>"
```
jon-betts: Should we be asserting so much about the `__str__` and `__repr__` methods? Seems like it might be better to assert they have some things in them we care about without being quite so prescriptive. Could be a nice use of the string matcher.

Is this to hit coverage?

tests/lms/models/group_info_test.py
```diff
@@ -0,0 +1,82 @@
...
+            retrieved_group_info.authority_provided_id == "test_authority_provided_id"
+        )
+        assert retrieved_group_info.consumer_key == "test_consumer_key"
+
+    def test_application_instance_relation(
+        self, application_instance, db_session, group_info
+    ):
+        db_session.add(group_info)
+
+        retrieved_group_info = db_session.query(GroupInfo).one()
```
jon-betts: I don't know enough about SQLAlchemy to know if this is making a full round trip to the DB or just in memory. If it's not I think we'd definitely want to make sure it was. Does that mean a flush before the query or do I just not get this?

tests/lms/services/group_info_upsert_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+class TestGroupInfoUpsert:
+    def test_it_adds_a_new_GroupInfo_if_none_exists(
+        self, db_session, group_info_upsert
+    ):
+        group_info_upsert(
+            "TEST_AUTHORITY_PROVIDED_ID",
+            "TEST_CONSUMER_KEY",
+            context_title="TEST_TITLE",
+            context_label="TEST_LABEL",
```
jon-betts: I was thinking that this might be a good place for sentinels, but sentinels are not comparable with their stringification, and probably not serialisable to the DB anyway.

I wonder if we could fixture out settings here so we have:

```
@pytest.fixture etc..
settings = {
    authority_provided_id="TEST_AUTHORITY_PROVIDED_ID",
   ...
}
...
group_info_upsert(**settings)
...
assert group_info.consumer_key == settings.consumer_key
```
Rather than having the string repetition

tests/lms/services/group_info_upsert_test.py
```diff
@@ -0,0 +1,100 @@
...
+            lms_url="TEST_LMS_URL",
+            shared_secret="TEST_SHARED_SECRET",
+            requesters_email="TEST_EMAIL",
+        )
+        pyramid_request.db.add(application_instance)
+        return application_instance
+
+    @pytest.fixture(autouse=True)
+    def group_infos(self, application_instance, pyramid_request):
+        """Add some "noise" group infos."""
```
jon-betts: This is a super nice idea. Stops your tests from passing because you just retrieve *anything* that was lying around.

lms/resources/_lti_launch.py
jon-betts: The diff is very confusing, but the code is actually very clear

lms/views/decorators/h_api.py
seanh: We're actually disabling it if provisioning is _not_ enabled. When provisioning is disabled the entire groups integration is disabled (no group even gets created in h)

tests/lms/models/group_info_test.py
seanh: It actually gets sent to Postgres: `db_session.add(...)` adds it to the sqlalchemy session (which doesn't necessarily send it to Postgres yet) but then the following `db_session.query(...)` will cause an sqlalchemy _autoflush_ to happen (all queries cause sqlalchemy to flush all SQL statements to Postgres). Although the SQL statements do get sent to Postgres (and Postgres would crash if there was anything wrong with them) we never do ask Postgres to _commit_ the transaction though

tests/lms/models/group_info_test.py
seanh: The query will cause the SQL to be autoflushed to Postgres. One way to test this is to turn on debug logging of the SQL statements that get sent:

```diff
diff --git a/tests/lms/conftest.py b/tests/lms/conftest.py
index 943baa34..6cc66fb3 100644
--- a/tests/lms/conftest.py
+++ b/tests/lms/conftest.py
@@ -28,7 +28,7 @@ SESSION = sessionmaker()
 
 @pytest.fixture(scope="session")
 def db_engine():
-    engine = sqlalchemy.create_engine(TEST_DATABASE_URL)
+    engine = sqlalchemy.create_engine(TEST_DATABASE_URL, echo=True)
     db.init(engine)
     return engine
 ```

Then tell pytest not to swallow the stdout/logging from tests that pass:

```terminal
$ export PYTEST_ADDOPTS="-s"
```

Then run just the one test:

```terminal
$ tox tests/lms/services/group_info_upsert_test.py::TestGroupInfoUpsert::test_it_adds_a_new_GroupInfo_if_none_exists
```

You should see a lot of SQL statements printed, including one `INSERT into application_instances ...` (the `application_instance` fixture) and four `INSERT into group_info`'s (the three from the `group_infos` noise fixture and the fourth from the call to the upsert service in the test body itself).

If you set `engine="debug"` instead of `True` it'll also print out the results of SQL queries and you can see the `'TEST_AUTHORITY_PROVIDED_ID'` etc coming back out of the DB.

tests/lms/models/group_info_test.py
seanh: It's just so that `__str__()` and `__repr__()` are in fact tested. It think it probably tests exactly the right amount to be honest: we do actually care that it prints the correct `authority_provided_id` ('test_authority_provided_id') for example, not just that it prints any string

lms/views/basic_lti_launch.py
seanh: I think the place to move this might be into a new `lms/views/decorators/_helpers.py` module (a module containing private helpers for the view decorators). The reason is that `upsert_group_info()` would then be part of the public API of this `_helpers.py`, which means that there is a place to put unit tests that call `upsert_group_info()` directly (in `tests/lms/views/decorators/_helpers_test.py`) like its current unit tests do. Rather than having to test it indirectly via the `upsert_course_group()` decorator that uses it (then `upsert_course_group()`'s tests get too long). This also means that `upsert_group_info()` is potentially reusable, though I don't know that we ever will reuse it.

```python
from lms.views.decorators._helpers import upsert_group_info

def upsert_course_group(wrapped):
    @functools.wraps(wrapped)
    def wrapper(context, request):
        ...
        upsert_group_info(...)
        ...
```

This might seem like too deep a hierarchy (not only do views have decorators but the view decorators now have helpers). On the other hand it kind of makes sense: the view decorators themselves are Pyramid integration stuff, so it makes sense that they should do little else than call helper functions (or services or whatever, some non-Pyramid-connected code).

judell:
Thanks @seanh. The tables should land in the analytics server tomorrow, at which point we should be off to the races.

PR #1095, opened 2019-11-01
Testing async tests

LMS007:
I have created a somewhat straightforward component that does some updating in a asynchronous way. This was loosely modeled after the fetch grade updated I recently made and noticed some trouble testing that.

I tried to create some tests that are able to determine all code paths for this component but most seem to produce a result that is not intuitive. It would be helpful to discover which practices to avoid or use for such tests to clear up confusion.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/AsyncTestComponent.js
+++ b/lms/static/scripts/frontend_apps/components/AsyncTestComponent.js
@@ -1,0 +1,35 @@
+import { createElement } from 'preact';
+import propTypes from 'prop-types';
+import { useState, useEffect } from 'preact/hooks';
+import asyncMethod from './AsyncTestComponentMethod';
+
+export default function AsyncTestComponent({ firstValue, secondValue }) {
+  const [stateValue, setStateValue] = useState('');
+
+  useEffect(() => {
+    const asyncSetValues = async () => {
+      setStateValue('initial_value');
+      try {
+        if (firstValue) {
+          const asyncValue = await asyncMethod();
+          setStateValue(firstValue + asyncValue);
+        }
+      } catch (e) {
+        setStateValue(e.name);
+      }
+
+      if (secondValue) {
+        setStateValue(secondValue);
+      }
+    };
+
+    asyncSetValues();
+  }, [firstValue, secondValue]);
+
+  return <div>{stateValue}</div>;
+}
+
+AsyncTestComponent.propTypes = {
+  firstValue: propTypes.string,
+  secondValue: propTypes.string,
+};
--- a/lms/static/scripts/frontend_apps/components/AsyncTestComponentMethod.js
+++ b/lms/static/scripts/frontend_apps/components/AsyncTestComponentMethod.js
@@ -1,0 +1,5 @@
+export default function asyncMethod() {
+  return new Promise(resolve => {
+    resolve('dummy');
+  });
+}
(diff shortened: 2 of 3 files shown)


robertknight:
Thanks for creating this example. I think I get the gist of the problems you were encountering. Below are my thoughts after some reflection. Perhaps you can have a look through and see if they could be applied to the grading components as a concrete use case?

---

As I see it, tests for a component that trigger asynchronous effects as part of their rendering can ultimately be distilled as:

1. Render the component in some initial state
2. Wait for the component to reach an "idle" state where nothing further will happen without external input.
3. Make assertions about the output of the component at that point

(1) and (3) are something that all our tests do already. The trickier part is (2) because there is no universal way to see from outside whether a component is "busy" or not.

In order to write this kind of test, you therefore need to somehow expose this idle/not-idle state. This state doesn't have to be visible to the user, but a test needs to be able to observe it. I can suggest a few ways of doing this, I'm sure you can think of others. 

1. The UI might naturally expose this state already as a "loading" indicator. Using this as the advantage that it is how a user would manually test the component, and is therefore directly relevant to whether the component is working as expected or not.

2. If there is no natural indicator, you could include something hidden (from the user) in the rendered output. For example:

```js
// Helper that produces no user-visible output.
const Hidden = () => null;

function Widget() {
  // Helper state that indicates whether the component is
  // "busy" or not.
  const [active, setActive] = useState(false);
  
  useEffect(() => {
    const fetchData = async () => {
      setActive(true);
      
      // Async stuff.
      
      setActive(false);
    };
  }, []);
  
  return (
    <div>
      Stuff the user sees
      <Hidden isActive={active}/> 
    </div>
  )
}
```

3. A different mechanism would be to pass a callback to the component which is invoked at various points to indicate the hidden state:

```js
function YourComponent({ onIdle }) {
  useEffect(() => {
    const fetchData = async () => {
      const data = await fetchDataFromApi(...);
      await doOtherStuff();
      
      // Signal that we are done.
      onIdle(true);
    };
    
    // Kick off activity.
    fetchData();
  }, [...]);
}
```

An advantage of (3) is that you can make a promise in a test which resolves when the callback is invoked.

The hidden state here can be a simple boolean, a string, or something richer.

Once the state is somehow exposed, a test can then wait on it. If the hidden state was just included in the output as above, a crude method of waiting would be an async `while` loop:

```js
const delay = ms => new Promise(resolve => setTimeout(resolve, ms));

// Render the widget and kick off activity.
const wrapper = renderWidget(...);

// Wait for widget to become idle
while (true) {
  await delay(1);
  
  // Update wrapper to reflect current output of widget.
  wrapper.update();
  
  if (!wrapper.find('Hidden').prop('isActive')) {
    break;
  }
}

// Check that output is as expected
assert.equal(wrapper.text(), "Expected output")
```

If using a callback to allow the component to report its state:

```js
let onIdle
const idle = new Promise(resolve => onIdle = resolve);

const widget = renderWidget({ onIdle });

await idle;

assert.equal(widget.text(), ...);
```

However the state is exposed, I think it should be treated as part of the "public API" of the component so you can change/refactor implementation details without needing to update the tests.

Perhaps with more experience we can extract some more concrete conventions about how to do this and perhaps some helpers.

robertknight:
One addendum to the above, in some cases it may help to extract logic from the effect into a helper (as a hook, or just an async function) which can be more easily tested/understood on its own. It depends on whether there is a lot of interplay between the stuff that happens in the effect and the state of the component.

seanh:
Is this PR intended to be merged at some point or should we close it, @LMS007 ?

PR #1127, opened 2019-11-06
Replace shallow rendering with mocking of imported components

robertknight:
Change UI tests to follow the pattern introduced in https://github.com/hypothesis/client/pull/1470 of using `mount` rather than `shallow` Enzyme rendering for all tests, but with imported components mocked using a `mockImportedComponents` helper.

This makes the tests more consistent, and removes some manual boilerplate for mocking components.

Some of the `SubmitGradeForm` tests broke due to waiting for async tasks to complete in a brittle way (see [related discussion in issue #1095](https://github.com/hypothesis/lms/pull/1095)). This commit resolves that by making the tests wait for the same loading indicators that the user sees in order to determine whether network requests have completed or not. This makes use of a `waitFor` utility that I had introduced previously but forgotten about when looking at #1095.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
@@ -1,7 +1,9 @@
-import { Fragment, createElement } from 'preact';
-import { shallow } from 'enzyme';
+import { createElement } from 'preact';
+import { mount } from 'enzyme';
 
 import SubmitGradeForm, { $imports } from '../SubmitGradeForm';
+import mockImportedComponents from './mock-imported-components';
+import { waitFor } from './util';
 
 describe('SubmitGradeForm', () => {
   const fakeStudent = {
@@ -11,22 +13,7 @@ describe('SubmitGradeForm', () => {
     LISOutcomeServiceUrl: '',
   };
   const renderForm = (props = {}) => {
-    return shallow(<SubmitGradeForm student={fakeStudent} {...props} />);
-  };
-
-  // eslint-disable-next-line react/prop-types
-  const FakeErrorDialog = ({ children }) => {
-    return <Fragment>{children}</Fragment>;
-  };
-
-  // eslint-disable-next-line react/prop-types
-  const FakeSpinner = ({ children }) => {
-    return <Fragment>{children}</Fragment>;
-  };
-
-  // eslint-disable-next-line react/prop-types
-  const FakeValidationMessage = ({ children }) => {
-    return <Fragment>{children}</Fragment>;
+    return mount(<SubmitGradeForm student={fakeStudent} {...props} />);
   };
 
   const fakeSubmitGrade = sinon.stub().resolves({});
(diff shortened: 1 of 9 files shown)


lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
```diff
@@ -11,39 +13,30 @@ describe('SubmitGradeForm', () => {
...
-    return <Fragment>{children}</Fragment>;
+    return mount(<SubmitGradeForm student={fakeStudent} {...props} />);
   };
 
   const fakeSubmitGrade = sinon.stub().resolves({});
   const fakeFetchGrade = sinon.stub().resolves({ currentScore: 1 });
   const fakeValidateGrade = sinon.stub();
   const fakeFormatToNumber = sinon.stub();
 
+  const isFetchingGrade = wrapper => {
```
LMS007: I like this

lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
```diff
@@ -174,31 +167,19 @@ describe('SubmitGradeForm', () => {
       assert.isTrue(
         wrapper
           .find('.SubmitGradeForm__grade-wrapper')
-          .find(FakeSpinner)
-          .props('classNames')
-          .className.includes('is-active')
-      );
-    });
-
-    it('removes the `active` class on the <Spinner> component after the grade has fetched', async () => {
```
LMS007: Did you feel this was redundant?

lms/static/scripts/frontend_apps/components/test/mock-imported-components.js
```diff
@@ -0,0 +1,65 @@
+/**
```
LMS007: Wondering if `test/util.js` and this file could be combined? They all sorta feel like important test utils.

lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
robertknight: I should have added a note. This wasn't redundant before but it is with the other changes in this PR because the presence of the `is-active` class is used by `isFetchingGrade`.

lms/static/scripts/frontend_apps/components/test/mock-imported-components.js
robertknight: I wondered about this myself. I put this code in a separate file so the imports would be consistent between tests in the lms and client apps. Lyza asked about having a clearer separation between tests and utilities used by tests. A possible structure might be something like:

```
components/test/util/
  wait.js
  mock-imported-components.js
```

Perhaps this is mature enough to move into a separate package which the client and lms apps could import (as `@hypothesis/test-utils` or something).

Thoughts?

robertknight:
I'm going to merge this to unblock any future developments but I note the issue here and in the client tests about possibly moving `mock-imported-components.js` to a separate location or maybe even a helper package.

PR #1129, opened 2019-11-07
Upgrade Jenkins to Postgres 11

robertknight:
In preparation for switching our production RDS database to Postgres 11,
change Jenkins CI to run tests in PG 11.

See also https://github.com/hypothesis/h/pull/5824. Unlike h, we don't have Travis so this change means there will be no CI tests running in PG 9.4 any more once this lands. I personally think that's not likely to be an issue given the relatively low volume of transactions, data and schema simplicity in the lms app (ie. no exotic PG-specific features are used).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ node {
     "Backend tests": {
         stage("Backend tests") {
             // Run the Postgres test DB in a Docker container.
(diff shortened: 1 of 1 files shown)


hmstepanek:
Should we also upgrade the docker compose file to point at the new version of postgres?

seanh:
Yeah we'll have to remember to upgrade it in docker-compose too. Which I guess will delete everyone's dev dbs

PR #1134, opened 2019-11-07
Moving the unit-tests to make space for functional ones

jon-betts:
This follows the style we are adopting in the libraries

This is in preparation for adding tests for the LTI certification and to prevent the merge requests being hideous.


seanh:
The nice thing about our Jenkins set up here is that it actually runs `make test` so if `make test` is working locally it should be working on CI

Issue #1132, opened 2019-11-07
Create a hello world functional test as a basis for LTI certification tests


jon-betts:
This ticket is to setup a 'hello' world functional test that sets the ground work for tests to come.

# Background

The LTI certification tests are a completely manually driven process. To have any chance of being sure we pass them we will need to automate where it's possible to do so.

This means:

 * We can be sure we don't have regressions
 * We can re-test at very low cost
 * Ultimately that we can move faster without worrying about having broken everything

# Acceptance critera

 * [x] We have at least one functional test which calls the LMS app
 * [x] This test passes in arguments in a reusable way
 * [x] This test makes some assertion about the page it receives (it can fail)
 * [x] The tests run in CI
 * [x] If the tests fail, the build in CI fails

jon-betts:
A merge request for this at: https://github.com/hypothesis/lms/pull/1135/files

jon-betts:
This was hella painful as the CI part was horrible. After removing Codecov (potentially spuriously) this was failing because make was trying build the javascript assets again in a docker container that didn't have the relevant tools.

This is going now and just requires a deploy.

jon-betts:
Done and deployed.

PR #1135, opened 2019-11-07
Adding some hello-world functional tests

jon-betts:
* Also changed the asset cache file to be CWD agnostic


jon-betts:
Addresses ticket: https://github.com/hypothesis/lms/issues/1132

robertknight:
Great to see that you're working on this. There is a lot of complexity in the interactions of different pieces of the lms app so I definitely think we're going to need some level of functional and e2e tests.

jon-betts:
The intention for these ones is to form a basis for doing some tests which mirror the LTI certification stuff to take out some of the guess work.

Also hah! Immediately hoisted by my own petard :face_with_head_bandage: :gun: Code coverage has dropped so the build won't pass... fair enough.

lms/assets.py
```diff
@@ -24,12 +25,22 @@ def __init__(self, path, loader, auto_reload=False):
...
         self._auto_reload = auto_reload
 
+    @classmethod
+    def _find_file(cls, path):
+        path = os.path.abspath(os.path.join(resource_filename("lms", "."), "..", path))
+
+        if not os.path.isfile(path):
+            raise FileNotFoundError(f"Expected to find a file at '{path}'")
+
+        return path
```
jon-betts: This was added to ensure we can run the tests from any directory, rather than just the root

tests/unit/lms/assets_test.py
```diff
@@ -0,0 +1,97 @@
+import os
+from unittest import mock
+
+import pytest
+
+from lms.assets import _CachedFile
+
+
+class Test_CachedFile:
```
jon-betts: These tests have been added:

a) Because there weren't any
b) Because I modified the file and dropped the coverage failing the build

jon-betts:
Ok so this was failing because of codecov, which I didn't touch but have now removed. Now it's failing because `yarn` is missing, which I also didn't touch??

https://jenkins.hypothes.is/job/lms/job/PR-1135/6/console

```
installing javascript dependencies
yarn install
make: yarn: Command not found
```

jon-betts:
So the issue here is that `make` thinks the assets need to be built again. This is already done in the Dockerfile, but because the special cheat file we add to let `make` know the node modules are up to date isn't there it builds them again.

I've attemped to put this into the Dockerfile, but we could also try changing the Jenkins file... as I've just seen this build failed too.

tests/unit/conftest.py
```diff
@@ -1,86 +1,15 @@
...
-    patcher = mock.patch(target, **options)
-    obj = patcher.start()
-    request.addfinalizer(patcher.stop)
-    return obj
-
-
-@pytest.fixture
-def patch(request):
-    return functools.partial(autopatcher, request)
+from tests.conftests import *
```
seanh: Is this necessary or does pytest find the higher up `conftest.py` itself anyway?

jon-betts: It would if you ran it from that directory, but if you run it from inside `tests/unit` or `tests/functional` as we do, then `pytest` won't recurse outside of the CWD.

PR #1138, opened 2019-11-08
Removing codecov and replacing it with a coverage ratchet value

jon-betts:
`codecov` was apparently getting routinely ignored, and failing so instead the idea is to:

 * Introduce a fixed percentage coverage value
 * Any time the total coverage falls below this the build will fail
 * Every so often we will review the value and increase it to the current level (never decrease)

The idea is that for any particular project new patches must match the general level of quality of the project from a coverage point of view. Of course beating this is encouraged, and over time if we only ever revise the number upwards we should move toward 100% coverage without having a massive cliff edge of work to achieve it.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,11 +5,6 @@ source =
     lms
     tests/unit
 
-[paths]
-source =
-    src/
-    */site-packages/
-
 [report]
 show_missing = True
 precision = 2
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,16 @@
+[run]
+branch = True
+parallel = True
+source =
+    lms
+    tests/unit
+
+[paths]
+source =
+    src/
+    */site-packages/
+
+[report]
+show_missing = True
+precision = 2
+fail_under = 96.11
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .pydeps
-.coverage*
+.coverage
+.coverage.*
 .tox
 .cache
 *.pyc
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ node {
             try {
                 testApp(image: img, runArgs: "${runArgs} -e TEST_DATABASE_URL=${databaseUrl(postgresContainer)} -e CODECOV_TOKEN=${credentials('LMS_CODECOV_TOKEN')}") {
                     installDeps()
-                    run("make backend-tests coverage codecov")
+                    run("make backend-tests coverage")
                 }
             } finally {
                 postgresContainer.stop()
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,7 +43,7 @@ node {
             postgresContainer = docker.image("postgres:11.5").run("-P -e POSTGRES_DB=lmstest")
 
             try {
-                testApp(image: img, runArgs: "${runArgs} -e TEST_DATABASE_URL=${databaseUrl(postgresContainer)} -e CODECOV_TOKEN=${credentials('LMS_CODECOV_TOKEN')}") {
+                testApp(image: img, runArgs: "${runArgs} -e TEST_DATABASE_URL=${databaseUrl(postgresContainer)}") {
                     installDeps()
                     run("make backend-tests coverage")
                 }
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ help:
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
 	@echo "make test              Run the unit tests"
 	@echo "make coverage          Print the unit test coverage report"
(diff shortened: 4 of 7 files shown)


robertknight:
> Every so often we will review the value and increase it to the current level (never decrease)

How will you ensure that this actually happens? A useful thing that codecov provides from my point of view, at least in the client repo where I do much of my work, is a "soft ratchet" where each PR includes an indicator as to how it affects the overall coverage relative to an automatically evolving baseline (the master branch). At the same time, since it doesn't prevent merging, developers are able to use their judgement to ignore occassional drops if they agree with the reviewer that there is a good reason for this.

jon-betts:
I think the approach is that if there's a good reason to not have coverage on a specific line or file then this should be covered by `# pragma: no cover` comments, along with an explanation. Which is nice because it should jump out in review and prompt discussion.

This way we can continue to allow any purposeful reduction in true coverage, but never allow any accidental reduction. It also becomes an act of commission rather than omission, which tips the default behavior in the direction of good practice.

As for ensuring this happens, I think we just do it as and when we remember and spot coverage has increased. It's probably something we're going to be thinking about more during our local testing as we know it won't pass CI. Any time we increase global coverage is a good time.

If we are worried people won't do this, then... that's exactly the situation we are in now, so I don't really see we have much to lose. If anyone ever does it, it's a bonus; if it never happens again we're still better off than we are now.

robertknight:
> I think the approach is that if there's a good reason to not have coverage on a specific line or file then this should be covered by # pragma: no cover comments, along with an explanation. 

That seems reasonable.

> As for ensuring this happens, I think we just do it as and when we remember and spot coverage has increased. 

Now what about when new developers come along who don't have this institutional knowledge and you are busy with some other project and no longer keeping as close an eye on the lms app 😉?

The main risk I would be concerned about here is that someone forgets a test for some important branch of the code and this goes unnoticed by the reviewer. If using a fixed coverage constant, this could happen as long as the overall project's coverage is above the threshold. If someone has forgotten to update the value in a while, you could end up building up a buffer (say, threshold of 90% and previous devs have gotten it up to 93%) that masks such mistakes.

seanh:
We're just bringing this app in line with how our libraries (h-cookiecutter-pypackage, h-pyramid-sentry and h-matchers) already work for coverage. We'll be doing this for h and bouncer too.

Increasing the `--fail-under` value over time can be automated using a GitHub Actions "cron" job.

The fact is that Codecov's soft ratchet is completely ineffective --- almost all of the developers on our team ignore it all the time. Including me. Codecov does not in practice help us to catch missing tests for important branches. And Codecov also produces various annoyances such as crashing on QA (for example Codecov is broken on Jenkins for LMS right now, which is blocking LMS work), posting annoying comments on PRs, using colour coding that's illegible to me, etc.

.coveragerc
```diff
@@ -0,0 +1,16 @@
+[run]
+branch = True
+parallel = True
+source =
+    lms
+    tests/unit
+
+[paths]
+source =
+    src/
```
seanh: Is `src/` wrong here? No `src` dir in lms

robertknight:
OK, well if the existing tools aren't working for the people doing most of the work then this makes sense. I just tend to raise my eyebrows when people put processes in place that rely on them remembering to do something at some future point.

jon-betts:
I think the main thing to remember is while everything you say is true, it's also true _right now_ but worse. Even unattended this situation prevents the code base from getting any worse, which presently it can. 

Opposing the introduction of seatbelts because we worry some people might not put them on doesn't really make us any safer overall.

Also this is mostly to catch _me_ out on my lazy days!

seanh:
@jon-betts We can also remove the `-e CODECOV_TOKEN` line from `Jenkinsfile`

.coveragerc
jon-betts: This block is saying that `src/` is equivalent to `*/site-packages/` which allows coverage to match when it's installed as a module.

So this is both wrong and redundant in the case of an app.

seanh:
I also wonder whether the `coverage run` command in `tox.ini` still needs its command line arguments (`--source lms,tests/unit`) now that they're in `.coveragerc`?

jon-betts:
Don't think so. I'll scoop those bits out.

jon-betts:
I've fixed that stuff up hopefully, so we should see if it still passes CI.

seanh:
Actually, thinking about it, I don't think it'd be worthwhile to write a GitHub Actions cron job to automate increasing the fail-under value. We should instead spend the time getting our projects to 100% coverage, which they're very close to anyway, making liberal use of nocover comments as needed. Once fail-under is 100 you won't be able to miss any coverage without explicitly marking it as nocover.

Anyway, that's something for later

PR #1155, opened 2019-11-11
LTI Certification 1.1 compliance - Part 1.2 - OLD

jon-betts:
For: https://github.com/hypothesis/lms/issues/1112

This is to present a nice message when certain parameters are missing during an LTI launch.


tests/unit/lms/views/decorators/lis_result_sourcedid_test.py
```diff
@@ -46,8 +46,9 @@ def test_it_raises_if_validation_raises_other_than_ValidationError(
 
         with pytest.raises(TypeError, match="foo"):
             upsert_lis_result_sourcedid(context, pyramid_request)
-            lis_result_sourcedid_svc.upsert.assert_not_called()
-            wrapped.assert_called_once_with(context, pyramid_request)
```
jon-betts: Wat? These lines could never be reached as the first raises an Exception. I've moved them down and guessed at what they were supposed to do.

tests/unit/lms/views/decorators/lis_result_sourcedid_test.py
```diff
@@ -134,12 +135,11 @@ def test_it_raises_if_upsert_service_raises(
             lis_result_sourcedid_value
         )
         lis_result_sourcedid_svc.upsert.side_effect = TypeError("service raised")
-
         with pytest.raises(TypeError, match="service raised"):
             upsert_lis_result_sourcedid(context, pyramid_request)
 
-            lis_result_sourcedid_svc.upsert.assert_not_called()
-            wrapped.assert_not_called()
```
jon-betts: Same again here.

tests/unit/lms/views/error_test.py
```diff
@@ -1,167 +1,155 @@
 from unittest import mock
```
jon-betts: This file looks like total spaghetti in the diff, just open the new one and it's quite obvious what's been done

tests/unit/lms/views/error_test.py
```diff
@@ -1,167 +1,155 @@
...
-                renderer="lms:templates/error.html.jinja2",
-            ),
-            mock.call(
-                error.validation_error,
-                context=ValidationError,
-                renderer="lms:templates/validation_error.html.jinja2",
-            ),
-        ]
+        assert (
+            pyramid_config.add_exception_view.call_args_list
```
jon-betts: Does the order matter? Or were we just matching the order because it was easy?

jon-betts:
I don't think this is done at all, but it's probably time for a review before carrying on.

jon-betts:
This is sort of done. I'd like to review the diff as it was pulled apart and rebased a lot before merging. This needs to be looked at after: https://github.com/hypothesis/lms/pull/1161

jon-betts:
I've squished this out into: https://github.com/hypothesis/lms/pull/1174

Issue #1136, opened 2019-11-07
"Hypothesis login lost" error when grading


seanh:
1. As a teacher, open up the Blackboard grading UI (i.e. open an assignment that has grading/evaluation labelled)
2. Wait several minutes (exact time unknown)
3. Continue trying to grade, e.g. try to submit a grade, or change to another student
4. You will see a "Hypothesis login lost" error message and will have to reload the entire browser tab to fix it

Relates to 
https://github.com/hypothesis/lms/issues/1199

![Screenshot from 2019-11-07 18-58-32](https://user-images.githubusercontent.com/22498/68421112-85840c80-0195-11ea-879c-ba3846cc28e7.png)

LMS007:
This fix will be to move to post message and not kill off the iframe over and over.

PR #1157, opened 2019-11-11
Add client rpc call to postMessage to sidebar

LMS007:
- Change onChangeSelectedUser handler in BasicLtiLaunchApp to send rpc call instead of changing the key on the iframe. Full page reloads no longer occur.
- Add client/random.js (similar to client file)
- Add client/client.js (similar to client postmessage-json-rpc.js file)

Fixes https://github.com/hypothesis/lms/issues/1136


lms/static/scripts/postmessage_json_rpc/server/server.js
```diff
@@ -63,6 +63,12 @@ export default class Server {
     if (!this._isJSONRPCRequest(event)) {
       return;
     }
+    // Save the last reference used to the sidebar window so we can
+    // send messages back at a later time in the lms app.
+    window._sidebarWindow = {
```
LMS007: This is not my first choice with where to store and look up the frame for the sidebar. I ran into some difficulty because the rpc server is simply injected into the code via a <script> tag and executed at that time ahead of the lms app. I attempted to manipulate the import into some sort of singleton class, but could not get it to only instantiate once and remain compatible in its original capacity with the <script> tag

Ideally, I wanted to store this sidebar frame in the server instance (e.g. `this.sidebarWindow`) and have a getter that could later be used in `BasicLtiLaunchApp.js` if we could import `postmessage_json_rpc/server` so we don't have to pollute the global scope like this. But it didn't work -- I'm open to suggestions?

lms/static/scripts/postmessage_json_rpc/client/test/random-test.js
```diff
@@ -0,0 +1,13 @@
+import { generateHexString } from '../random';
```
LMS007: Copied from the client

lms/static/scripts/postmessage_json_rpc/client/random.js
```diff
@@ -0,0 +1,23 @@
+/* global Uint8Array */
```
LMS007: copied from the client

lms/static/scripts/postmessage_json_rpc/client/client.js
```diff
@@ -0,0 +1,98 @@
+import { generateHexString } from './random';
```
LMS007: copied from the client and modified to use async, we should copy this back to the client or fold this into a shared util package.

LMS007:
@lyzadanger 
Decided against a separate bundle (for now), posted a few reasons in slack, we can discuss tomorrow

https://hypothes-is.slack.com/archives/C4K6M7P5E/p1573683986049900

lms/static/scripts/postmessage_json_rpc/client/test/client-test.js
```diff
@@ -0,0 +1,159 @@
+import EventEmitter from 'tiny-emitter';
```
lyzadanger: I totally see that you brought this over from the client implementation, but I'm curious about this dependency. It seems like it exists to allow these tests to run in non-Node envs that don't have `events.EventEmitter`? But then, I'm like, when do we run these tests not in a Node environment/where we wouldn't have access to the `events` core module?

lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -1,6 +1,29 @@
...
+ */
+function startRpcServer() {
+  server = new Server();
+  server.register('requestConfig', requestConfig);
+}
+
+/**
+ * @typedef {Object} SidebarFrame
+ * @prop {Object} frame - The reference to the sidebar window
+ * @prop {string} origin - The sidebar window's origin uri
```
lyzadanger: Omit `uri` here

lms/static/scripts/postmessage_json_rpc/server/server-test.js
```diff
@@ -128,7 +128,7 @@ describe('postmessage_json_rpc/server#Server', () => {
       return assertThatTheServerDidntRespond();
     });
 
-    it("responds with an error if there's no request identifier", () => {
+    it("xx responds with an error if there's no request identifier", () => {
```
lyzadanger: `xx`?

lms/static/scripts/postmessage_json_rpc/server/server.js
```diff
@@ -63,6 +63,12 @@ export default class Server {
     if (!this._isJSONRPCRequest(event)) {
       return;
     }
+    // Save the last reference used to the sidebar window so we can
+    // send messages back at a later time in the lms app.
+    this._sidebarWindow = {
```
lyzadanger: I didn't notice this until now, but it could be confusing that this reference name has `Window` in it, but this is not a `Window`. Not a big deal, just threw me a little.

lms/templates/base.html.jinja2
```diff
@@ -40,9 +40,6 @@
 
       {% if context.rpc_server_config is defined %}
         <script type="application/json" class="js-rpc-server-config">{{ context.rpc_server_config|tojson }}</script>
-        {% for url in asset_urls("postmessage_json_rpc_server_js") %}
```
lyzadanger: We're confident that removing this doesn't break other parts of the app, e.g. content selection?

lms/static/scripts/postmessage_json_rpc/client/client.js
```diff
@@ -0,0 +1,98 @@
...
+    jsonrpc: '2.0',
+    method,
+    params,
+    id,
+  };
+
+  try {
+    frame.postMessage(request, origin);
+  } catch (err) {
+    return Promise.reject(err);
```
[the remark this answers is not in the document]
LMS007: So it shows up in the console as an error. Not sure any of that should be user facing -- as an improvement we could add in a toast style notification to the UI.  I could also just wrap the `rcpCall` in `BasicLtiLaucnApp` in a try/catch and then formally print errors to `console.error`. That way it won't show up as an unresolved promise, but either way it does get to the console.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -197,17 +195,26 @@ export default function BasicLtiLaunchApp() {
...
      */
-    const changeSelectedUserKey = userid => {
-      setSidebarKey(userid);
+    const changeSelectedUser = user => {
+      const sidebar = getSidebarWindow();
+
+      if (sidebar) {
+        rpcCall(sidebar.frame, sidebar.origin, 'changeFocusModeUser', [
+          {
+            username: user.userid,
```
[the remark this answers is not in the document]
LMS007: yes

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -197,17 +195,26 @@ export default function BasicLtiLaunchApp() {
...
      * Callback when the selected student changes. This function
-     * changes the key on the iframe so it gets rebuilt.
+     * makes an rpc call to change to the focused user of the sidebar.
      */
-    const changeSelectedUserKey = userid => {
-      setSidebarKey(userid);
+    const changeSelectedUser = user => {
+      const sidebar = getSidebarWindow();
+
+      if (sidebar) {
```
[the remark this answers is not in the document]
LMS007: I'll add a comment about this, but LMSGrader called this right away because of line 43 in LMSGrader

`onChangeSelectedUser({}); // any non-real userid will clear out a previously focused user`

This is necessary to clear out the user, but its also the same code path for when there is no valid student (e.g. All Students). What the means is onChangeSelectedUser is triggered right away because `useEffect` always fire once at the start. 

I think all of that is okay, but where it breaks down is that the sidebar is usually not ready at the outside so we don't kick off an RPC message. It only works because there is no focused user at the start and we don't need to clear out the user since its not there already.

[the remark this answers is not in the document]
LMS007: An improvement here could be that getSidebarWindow() could be async, what do you think?

lms/static/scripts/postmessage_json_rpc/client/test/client-test.js
LMS007: cant answer that, so lets discuss next week, good q.

lms/static/scripts/postmessage_json_rpc/server/server-test.js
LMS007: thats my magic sauce!

`test-watch --grep xx`

I will delete.

lms/static/scripts/postmessage_json_rpc/server/server.js
LMS007: ya, i cant think of a better name atm. hmm..

lms/templates/base.html.jinja2
LMS007: not 100% sure, but seems to work

👀

lms/static/scripts/postmessage_json_rpc/client/client.js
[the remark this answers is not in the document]
robertknight: Since this is now an async function, you can remove the `try..catch` block entirely as async/await automatically converts exceptions to rejected promises.

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -36,11 +36,11 @@ export default function LMSGrader({
         },
       });
       // let the parent component know the index changed
-      onChangeSelectedUser(students[currentStudentIndex].userid);
+      onChangeSelectedUser(students[currentStudentIndex]);
     } else {
       // clear focused user
       removeClientConfig(['focus']);
```
robertknight: Is there any need to update the client config here still? When trying to resolve an issue locally to do with the username => userid change I got confused because this component is now updating the config in two different ways, but AFAIK only the `onChangeSelectedUser` change actually makes a difference.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -7,16 +7,24 @@ import {
...
+import AuthWindow from '../utils/AuthWindow';
 import Dialog from './Dialog';
 import Button from './Button';
 import ErrorDisplay from './ErrorDisplay';
 import LMSGrader from './LMSGrader';
 import Spinner from './Spinner';
 
+/**
+ * @typedef User
+ * @property {string} userid - Unique user's id
```
robertknight: The `username` => `userid` change here breaks the focused mode switching in the client, presumably until https://github.com/hypothesis/client/pull/1496 lands?

I suggest we change either the client or the server so that we can deploy one part of this without breaking the other.

For example, if you change https://github.com/hypothesis/client/pull/1496/ to accept either `userid` or `username` temporarily, then we can deploy the client change first without breaking Speed Grader or the current Blackboard/Moodle grading implementation, and then separately deploy a change to the LMS app.

Also, I found that the logic which [generates the static config for SpeedGrader](https://github.com/hypothesis/lms/blob/2c888895c922bd999028149e2298c653b3d340ef/lms/views/basic_lti_launch.py#L306) hasn't been updated yet to use `userid`. I presume that will break SpeedGrader?

lms/static/scripts/frontend_apps/components/LMSGrader.js
robertknight: Oh ... I guess this will still matter if the sidebar has not yet loaded? In that case I think it is a bit confusing that the logic of deciding _how_ to handle a changed user is partly handled in the parent component (to make an RPC call) but partly handled in this component (to update the static config). I suggest that both should be handled in the same place.

lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -1,6 +1,37 @@
...
+ *
+ * @returns {SidebarFrame}
+ */
+function getSidebarWindow() {
+  if(server._sidebarWindow) {
+    return Promise.resolve(server._sidebarWindow);
+  }
+  else {
+    return new Promise((resolve, reject) => {
+      _sidebarWindowLoaded
```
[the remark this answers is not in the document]
LMS007: I think I mistakenly pushed this part of the code, but its clearly not complete -- my fault.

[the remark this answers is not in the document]
LMS007: Yep, this is exactly what I needed, just didn't flush it out for that last commit. Thanks Rob!

lms/static/scripts/frontend_apps/components/LMSGrader.js
LMS007: ^ Makes sense, I'll move it to the parent

lms/static/scripts/postmessage_json_rpc/client/test/client-test.js
```diff
@@ -0,0 +1,159 @@
...
+          error: {
+            message: 'Something went wrong',
+          },
+        },
+      });
+
+      try {
+        await call;
+      } catch (e) {
+        assert.equal(e.message, 'Something went wrong');
```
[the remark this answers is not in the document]
LMS007: good catch!

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: Will fix the client to handle both, then rip it out after we update lms, great idea Rob.

LMS007:
Pushed a fixup _(note: will need to squash all this when merged)_

- Delete update-client-config.js, now only use rpc to set focus mode
- Move rpc call to set focus mode from BasicLtiLaunchApp into LMSGrader
- Fix getSidebarWindow to correctly return a promise that resolves to the sidebar window frame
- Clean up postmessage_json_rpc/../client.js try/catch block
- Various test improvements / and needed refactoring


The major change with this last commit is I ripped out the `update-client-config.js` file and this no longer edits that confg in the DOM . Instead everything is done via RPC request and I think that's all we need now as close as I can tell.

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -19,30 +18,34 @@ export default function LMSGrader({
...
 }) {
   // No initial current student selected
   const [currentStudentIndex, setCurrentStudentIndex] = useState(-1);
 
+  /**
+   * Makes an RPC call to the sidebar to change to the focused user.
+   *
+   * @param {User} user - The user to focus on in the sidebar
+   */
+  const changeFocusedUser = async user => {
```
LMS007: This is much cleaned up now, all the logic for setting the focus to the sidebar is done here. We no longer need to tamper with the config in the DOM

lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -19,30 +18,34 @@ export default function LMSGrader({
...
-        },
-      });
-      // let the parent component know the index changed
-      onChangeSelectedUser(students[currentStudentIndex].userid);
+      changeFocusedUser(students[currentStudentIndex]);
     } else {
-      // clear focused user
-      removeClientConfig(['focus']);
-      onChangeSelectedUser('0'); // any non-real userid will work
+      changeFocusedUser({}); // any non-real userid will clear out a previously focused user
```
LMS007: Note, no need to keep track of a cancel var since we're not setting state as like was done here
https://github.com/hypothesis/lms/blob/master/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js#L25

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -12,15 +12,18 @@ describe('LMSGrader', () => {
       LISOutcomeServiceUrl: '',
     },
     {
-      userid: 'student1',
+      userid: 'student2',
```
lyzadanger: I'm wondering if it might be a good idea to use `userid`s that look like actual `userid`s to avoid further confusing future devs.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -84,61 +90,73 @@ describe('LMSGrader', () => {
     );
   });
 
-  it('passes a default value of "0" to onChangeSelectedUser when no a student is selected', () => {
+  it('does not set a focus user by default', async () => {
```
lyzadanger: This test name is much more friendly

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -84,61 +90,73 @@ describe('LMSGrader', () => {
...
+            username: undefined,
+            displayName: undefined,
+          },
+        ]
+      )
+    );
   });
-
-  it('passes the unique userid to onChangeSelectedUser when a student is selected', () => {
+  it('calls rpcCall to focus on that user when onSelectStudent is called with a valid student index', async () => {
```
lyzadanger: Whoa, this is quite a mouthful. Possible to rename more clearly?

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -84,61 +90,73 @@ describe('LMSGrader', () => {
...
 
-  it('changes the sidebar config to focus to the specified user when onSelectStudent is called with a valid user index', () => {
+  it('does not set a focus user when the user index is invalid', async () => {
     const wrapper = renderGrader();
     act(() => {
       wrapper
         .find(FakeStudentSelector)
         .props()
-        .onSelectStudent(0); // initial index is -1
+        .onSelectStudent(0); // invalid choice
```
lyzadanger: I'm a little confused as to why this _wasn't_ an invalid choice in the previous test?

lms/static/scripts/postmessage_json_rpc/server/server.js
```diff
@@ -33,6 +33,11 @@ export default class Server {
 
     // The methods that can be called remotely via this server.
     this._registeredMethods = {};
+
+    // Public field which resolves when sidebar is loaded.
```
lyzadanger: Not sure this comment is necessary

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: yep, good idea

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: lol, you should have seen the other version i was going for, it was longer!  Basically this just needs to change, "it changes a user" I 'll make it better

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: ahh, this should be -1

I think this needs to be `fakeRpcCall.secondCall` need to double check other tests here

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -76,7 +77,7 @@ describe('LMSGrader', () => {
     );
   });
 
-  it('set the selected student count to "Student 2 of 2" when the index changers to 1', () => {
+  it('set the selected student count to "Student 2 of 2" when the index changers to 1', async () => {
```
LMS007: @lyzadanger 

Sinon drives me nuts sometimes!

Since this technically calls a promise (look at line 86), we seem to need to make the function `async` even though I don't need to `await` on that promise or even care about it returning because the stuff I'm testing here is not dependent on that to resolve, but if I don't make the test `async`, the spy `fakeRpcCall` seems to not work as expected in future tests. Specifically the  `restHistory` does not reset that spy's history and I observed that the call count was growing in spite of calling `restHistory`

I think what was happening is the first test didn't finish until after the second test started? Anyway so you'll notice if you remove `async` from this test, the next test fails because the callcount is 3 instead of 1 and `calledOnceWithExactly` is never true.

robertknight: > I think what was happening is the first test didn't finish until after the second test started? 

That seems possible. My immediate hypothesis about what is going on is that it is to do with the fact that async functions always return a Promise:

Making this function async will cause it to return a Promise, even though there is no explicit return in the function. The promise will "immediately" resolve to `undefined`. Since the result is a Promise, mocha will wait for it to resolve before continuing to the next test. It does this waiting by queuing a microtask to run once all the promises resolve. This microtask will run _after_ any other tasks in the queue.

If the function returns synchronously on the other hand, the next test will run immediately without the microtask queue being flushed.

If this test results in some microtasks being scheduled but not yet executed by the time the test returns, the tasks could end up running at some later point and manipulating shared state in those other tasks.

An alternative fix is to change the usage of sinon so that each test gets a fresh stub, rather than sharing the same stub instance between tests and using `resetHistory`. This change would ensure that any code which happens to run after the test returns won't end up affecting future tests.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -29,15 +32,19 @@ describe('LMSGrader', () => {
...
+      },
+      '../../postmessage_json_rpc/server': {
+        getSidebarWindow: fakeSidebarWindow,
+      },
     });
   });
 
   afterEach(() => {
+    fakeRpcCall.resetHistory();
+    fakeOnChange.resetHistory();
```
robertknight: One reason we usually create new stubs in `beforeEach` rather than having a single shared stub that is reset between tests is that this creates a hazard where a test can trigger some activity (eg. in a timeout or Promise) that ends up affecting a later test due to sharing the same state. If each test gets a fresh stub instance, this can't happen.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
```diff
@@ -84,61 +91,73 @@ describe('LMSGrader', () => {
...
+            username: undefined,
+            displayName: undefined,
+          },
+        ]
+      )
+    );
   });
-
-  it('passes the unique userid to onChangeSelectedUser when a student is selected', () => {
+  it('sets the focused user if a valid index one is selected', async () => {
```
robertknight: "valid index one"? Did you mean "sets the focused user when a valid index is passed"?

robertknight:
I found a small issue in testing this where the "spinner" can be left showing even though no grade is being loaded when switching to the "All students" view.

**Steps to reproduce**

1. Load up an assignment with the grading view visible
2. Switch to the "All students" view
3. Switch to a specific student and then quickly switch back to "All students" before the grade has finished loading

**Expected result**

No spinner in the grade input

**Actual result**

A spinner is displayed indefinitely in the grade input

spinner-state

No console error is logged.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: yes, i think thats the issue I ran into, so I'll go ahead and make sure each stub is brand new as that seems to be a much better standard. I sorta think things like `resetHistory` lead to less than ideal patterns and perhaps that function should just not be part of sinon. The API is quiet huge and hard to lock in best practices with since there are so many ways to do things.

lms/static/scripts/frontend_apps/components/test/LMSGrader-test.js
LMS007: Thanks Rob. That makes sense, I think as best practice, we always re-create the stubs (if they change) but always the spies each test.

PR #1158, opened 2019-11-12
Remove blackboard_grading feature flag

seanh:
This feature flag is turned on for everyone in QA and prod now so it's no longer needed

### DIFF
--- a/tests/unit/lms/views/api/lti_test.py
+++ b/tests/unit/lms/views/api/lti_test.py
@@ -12,14 +12,15 @@ from lms.views.api.lti import LTIOutcomesViews
 
 class TestRecordCanvasSpeedgraderSubmission:
     def test_it_passes_correct_params_to_read_current_score(
-        self, pyramid_request, lti_outcomes_client
+        self, ai_getter, pyramid_request, lti_outcomes_client
     ):
         LTIOutcomesViews(pyramid_request).record_canvas_speedgrader_submission()
 
+        ai_getter.shared_secret.assert_called_once_with("TEST_OAUTH_CONSUMER_KEY")
         lti_outcomes_client.read_result.assert_called_once_with(
             LTIOutcomesRequestParams(
                 consumer_key="TEST_OAUTH_CONSUMER_KEY",
-                shared_secret="oauth-secret",
+                shared_secret=ai_getter.shared_secret.return_value,
                 lis_outcome_service_url="https://hypothesis.shinylms.com/outcomes",
                 lis_result_sourcedid="modelstudent-assignment1",
             )
@@ -43,6 +44,7 @@ class TestRecordCanvasSpeedgraderSubmission:
     )
     def test_it_records_result_if_no_score_exists(
         self,
+        ai_getter,
         pyramid_request,
         lti_outcomes_client,
         document_url,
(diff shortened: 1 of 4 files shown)


tests/unit/lms/views/api/lti_test.py
```diff
@@ -12,14 +12,15 @@
...
+        self, ai_getter, pyramid_request, lti_outcomes_client
     ):
         LTIOutcomesViews(pyramid_request).record_canvas_speedgrader_submission()
 
+        ai_getter.shared_secret.assert_called_once_with("TEST_OAUTH_CONSUMER_KEY")
         lti_outcomes_client.read_result.assert_called_once_with(
             LTIOutcomesRequestParams(
                 consumer_key="TEST_OAUTH_CONSUMER_KEY",
-                shared_secret="oauth-secret",
+                shared_secret=ai_getter.shared_secret.return_value,
```
seanh: There's no need to set `ai_getter.shared_secret.return_value` to `"oauth-secret"`. Mock methods already have return values (each mock method returns another unique mock object) and they can just be accessed as `.return_value` (and nothing in these tests cares that the return value is actually a string)

tests/unit/lms/views/api/lti_test.py
```diff
@@ -122,13 +128,14 @@ def pyramid_request(self, pyramid_request):
 
 
 class TestRecordResult:
-    def test_it_records_result(self, pyramid_request, lti_outcomes_client):
+    def test_it_records_result(self, ai_getter, pyramid_request, lti_outcomes_client):
 
         LTIOutcomesViews(pyramid_request).record_result()
 
+        ai_getter.shared_secret.assert_called_once_with("TEST_OAUTH_CONSUMER_KEY")
```
seanh: This assertion was previously hidden in the body of the `ai_getter` fixture (see below) so you couldn't see, just from looking the `TestRecordResult` tests themselves, whether this was actually being tested. You would have to read the code body of the `ai_getter` fixture itself. And if we had a habit of putting our assertions inside fixtures then you would have to read the code of every one of this test's fixtures, and every fixture of every test, to know what is being tested. And the more complex fixtures would tangle all the different tests together. Just put your assertions in your tests

tests/unit/lms/views/api/lti_test.py
```diff
@@ -158,12 +165,5 @@ def lti_outcomes_client(pyramid_config):
 @pytest.fixture(autouse=True)
 def ai_getter(pyramid_config):
     svc = mock.create_autospec(ApplicationInstanceGetter, instance=True, spec_set=True)
-
-    def shared_secret(consumer_key):
-        if consumer_key != "TEST_OAUTH_CONSUMER_KEY":
-            raise Exception("Incorrect consumer key")
-        return "oauth-secret"
-
-    svc.shared_secret.side_effect = shared_secret
```
seanh: The `raise Exception("Incorrect consumer key")` (apparently used instead of an assertion) was never being hit which was reducing our coverage score. I've replaced this with explicit `assert_called_once_with(...)`'s in the relevant tests, which makes this fixture simpler, increases coverage/means we don't have dead test fixture code, and makes the tests easy to understand and less entangled with each other

tests/unit/lms/views/api/lti_test.py
jon-betts: This is quite nice. I've not done it before, but good to know

tests/unit/lms/views/api/lti_test.py
jon-betts: Principal of least surprise is a good thing

PR #1165, opened 2019-11-13
Using error view decorators and fixing html Validation errors

jon-betts:
Previously we would return JSON for ValidationErrors regardless. We now specify what should happen if the user requests "text/html".

As a part of this I ran a cleaner over the imports in the tests as I remembered I'd left an un-used import. It cleaned out quite a few others too.

### DIFF
--- a/lms/app.py
+++ b/lms/app.py
@@ -83,7 +83,6 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.include("lms.routes")
     config.include("lms.assets")
     config.include("lms.views")
-    config.include("lms.views.error")
     config.include("lms.services")
     config.include("lms.validation")
     config.include("lms.tweens")
--- a/lms/views/error.py
+++ b/lms/views/error.py
@@ -1,5 +1,6 @@
 import h_pyramid_sentry
 from pyramid import httpexceptions, i18n
+from pyramid.config import not_
 from pyramid.view import (
     exception_view_config,
     forbidden_view_config,
@@ -50,7 +51,7 @@ def http_server_error(exc, request):
 
 @exception_view_config(
     context=ValidationError,
-    accept="text/html",
+    path_info=not_("^/api/.*"),
     renderer="lms:templates/validation_error.html.jinja2",
 )
 def validation_error(exc, request):
--- a/lms/views/error.py
+++ b/lms/views/error.py
@@ -20,13 +20,13 @@ def _http_error(exc, request):
 
 
 @notfound_view_config(renderer=DEFAULT_RENDERER)
-def notfound(request):
+def notfound(_, request):
     request.response.status_int = 404
     return {"message": _("Page not found")}
 
 
 @forbidden_view_config(renderer=DEFAULT_RENDERER)
-def forbidden(request):
+def forbidden(_, request):
     request.response.status_int = 403
     return {"message": _("You're not authorized to view this page")}
 
@@ -60,7 +60,7 @@ def validation_error(exc, request):
 
 
 @exception_view_config(context=Exception, renderer=DEFAULT_RENDERER)
-def error(request):
+def error(_, request):
     """
     Handle an unexpected exception.
 
--- a/lms/views/error.py
+++ b/lms/views/error.py
@@ -1,22 +1,16 @@
 import h_pyramid_sentry
 from pyramid import httpexceptions, i18n
-from pyramid.view import forbidden_view_config, notfound_view_config
+from pyramid.view import (
+    exception_view_config,
+    forbidden_view_config,
+    notfound_view_config,
+)
 
 from lms.validation import ValidationError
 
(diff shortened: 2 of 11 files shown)


tests/unit/lms/views/error_test.py
```diff
@@ -103,45 +103,6 @@ class TestError(ExceptionViewTest):
     }
 
 
-@pytest.mark.usefixtures("pyramid_config")
```
jon-betts: We don't need to test this any more. We would just be testing that Pyramid decorators work

lms/app.py
```diff
@@ -83,7 +83,6 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.include("lms.routes")
     config.include("lms.assets")
     config.include("lms.views")
-    config.include("lms.views.error")
```
jon-betts: Adding the decorators means this isn't required anymore

lms/views/error.py
```diff
@@ -1,48 +1,66 @@
...
+@notfound_view_config(renderer=DEFAULT_RENDERER)
+def notfound(_exc, request):
     request.response.status_int = 404
     return {"message": _("Page not found")}
 
 
-@forbidden_view_config(renderer="lms:templates/error.html.jinja2")
-def forbidden(request):
+@forbidden_view_config(renderer=DEFAULT_RENDERER)
+def forbidden(_exc, request):
```
jon-betts: The `_exc` param doesn't do anything here, but when debugging it contains the exception and therefore stack trace of what went wrong, which is extremely helpful.

tests/unit/lms/authentication/__init___test.py
```diff
@@ -1,5 +1,4 @@
 import pytest
-from pyramid import testing
```
jon-betts: I ran an import cleaner on the tests to get rid of some I knew I'd left in, but it picked up loads more besides.

seanh:
How to test calling the API from a terminal:

You need to get a bearer token to use in your `Authorization` header, like so:

```
$ make shell
>>> from lms.validation.authentication import BearerTokenSchema
>>> from lms.values import LTIUser
>>> BearerTokenSchema(request).authorization_param(LTIUser(user_id='f82...a00', oauth_consumer_key='Hypothesis5ab...dde', roles='Instructor'))
'Bearer eyJ...XeU'
```

(I used a real `user_id` and `oauth_consumer_key` from our Canvas test site.)

You can now use this bearer token string to make all kinds of valid and invalid, authenticated and unauthenticated requests to the API and see the responses in the terminal. For example using [HTTPie](https://httpie.org/):

```terminal
$ http GET 'http://localhost:8001/api/lti/result' Authorization:'Bearer eyJ...XeU'
```

seanh:
I did a bunch of testing locally:

* Visit <http://localhost:8001/foo>: HTML error page and 404 status
* Tested in Potemkin:
  * Launch with no consumer key, shared secret, or oauth sig: 403 and HTML error
    page
  * With key and secret but no sig: same
  * With a ValidationError (I had to hack the lms code to actually raise
    ValidationError in response to an LTI launch): 422 and HTML validation error
    page
* Tested with HTTPie:
  * Attempt to call a `/api/*` URL that doesn't exist: 404 with JSON body
  * Call a correct API URL with no authentication: 403 with JSON body
  * Call a correct API URL with correct authentication: JSON response as expected

Our Canvas files support relies on our JSON API and it particularly relies on getting correct error responses from our JSON API so I also tested that in Canvas. When trying to create a Canvas files assignment, when the frontend tries to get the list of files:

1. If we have no access token for the user. You can create this situation by running `make sql` then doing `DELETE FROM oauth2_token;`. When you click <kbd>Select PDF from Canvas</kbd> it will ask you to authorize and then it will work.

2. If you have already authorized then it has the access token in the DB and getting the list of files will just work.

3. If our access token is invalid (e.g. expired) then it will also "just work" because it'll automatically use the refresh token to get a new access token. To simulate this situation: `UPDATE oauth2_token SET access_token = 'foo';`

4. If both the access token and the refresh token fail to work it will show you an error message and ask you to re-authorize (and then it will work). To simulate: `UPDATE oauth2_token SET access_token = 'foo'; UPDATE oauth2_token SET refresh_token = 'foo'; `.

There are various other situations. For example the Canvas API fails to respond or sends an error or invalid response (that is not to do with our access/refresh token). But I thought the above was enough testing here

lms/views/error.py
```diff
@@ -1,48 +1,66 @@
...
+)
 def http_server_error(exc, request):
     """Handle an HTTP 5xx (server error) exception."""
     h_pyramid_sentry.report_exception()
     return _http_error(exc, request)
 
 
+@exception_view_config(
+    context=ValidationError,
+    accept="text/html",
```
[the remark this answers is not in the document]
seanh: (I ended up pushing a commit that changed this to `path_info` because I found it wasn't working as desired)

PR #1171, opened 2019-11-14
Consolidate hash code and speed up unit tests (by 20% locally)

jon-betts:
* Removed repeated (and different code) implementing password hashing
 * The effected tests now run around 10x faster without real loss of coverage
 * As a result the whole test suite is now 20% faster (on my machine)
 * Added doc strings

The times for running `make backend-test` according to `pytest` (not including `tox` stuff which adds ~10s):

 * Before 56.57 seconds
 * After 45.96 seconds

A similar (10-15 second) improvement can be seen in Jenkins, but its far less dramatic as the tests take ages anyway.


jon-betts:
Thanks for the review!

I think there's a good argument to reduce them everywhere. Most sources site 100,000 repetitions or lower.

Issue #932, opened 2019-08-22
Run IMS Certification Tests for LTI 1.1


klemay:
Part of https://github.com/hypothesis/lms/issues/705, not blocked by https://github.com/hypothesis/lms/issues/876 but it may not make sense to run the tests until #876 is implemented. 

The IMS website is so confusing/complicated that they created a video for members to teach them how to navigate it... I'll try to point y'all to where you need to go. 

To run the test:

1. Go to https://www.imsglobal.org/user and log in (credentials in Passpack)
2. Go to https://www.imsglobal.org/lti-v1-certification-suite (if you skip step 1 you will be redirected to the login page and likely won't be able to navigate your way here - so please log in first!)
3. Click on "Process for Tool Providers"
![screenshot-www imsglobal org-2019 08 22-15-32-55](https://user-images.githubusercontent.com/12087705/63544092-43322400-c4f2-11e9-9eef-214e6b513b97.png)
4. You'll be directed to a "wizard" that will walk you through the testing process

The testing process is divided into the following stages:
- Overview
- Configuration (basically you'll create a key and secret to use with their testing suite)
- Tests
- Results (this will tell us if we need to make any changes to get certified)
- Submit (and then we wait for IMS to get back to us)

The certification tests are divided into the following sections:

- Tool registration (5 tests)
- Invalid Launch Requests (9 tests)
- Valid Launch Requests (4 tests)
- Types of User Roles (8 tests)
- Levels of User Information (7 tests)
- Deep Linking Message (2 tests)
- Outcomes 1.0 service (4 tests)
- Tool Consumer Profile (1 test)
- Tool Settings service (4 tests)
- Names and Roles Service (4 tests)

seanh:
I've run the test suite:

## Summary

If I'm reading the certification suite tool correctly, there appear to be several separate certifications here. Some of the certifications apply to our app, and some don't. And there are different tests for each certification:

* 1.0
* 1.1
* 1.2 (doesn't apply to us)
* Deep Linking (doesn't seem to apply to us / I don't understand the tests, see below for details)
* Outcomes-1 (doesn't apply to us until we have released grading for all LMS's, but looks as if it _may_ be required for 1.1 certification, unclear, see below for details)
* Tool Settings (service) (doesn't apply to us)
* Names and Roles (service) (doesn't apply to us)

In order to pass all the tests that the test suite claims "apply to" LTI 1.0 and 1.1 we need to:

[...]

PR #1174, opened 2019-11-14
LTI Certification 1.1 compliance - Part 1.2

jon-betts:
For: https://github.com/hypothesis/lms/issues/1112

This is to present a nice message when certain parameters are missing during an LTI launch.


tests/conftest.py
```diff
@@ -58,41 +57,6 @@ def db_engine():
     return engine
```
jon-betts: The fixtures for the unit and func tests have to be different. We can't roll back the transaction everytime, because we need things to _really_ reach the DB in the func tests.

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -0,0 +1,99 @@
...
+
+import oauthlib.common
+import oauthlib.oauth1
+import pytest
+from h_matchers import Any
+
+from lms.models import ApplicationInstance
+from tests.functional.base_class import TestBaseClass
+
+
```
jon-betts: If there are any smallish changes that need to be made here, you might want to hold off until the next ticket, as it rewrites a decent portion here.

tests/unit/lms/resources/_lti_launch_test.py
```diff
@@ -153,7 +153,8 @@ def test_h_display_name(
     def test_h_authority_provided_id_raises_if_theres_no_tool_consumer_instance_guid(
```
jon-betts: We are now picky about values we weren't before. All of these tests we starting with empty params and adding a few things. Now we start with default params and remove them or change them as required.

I think this actually reads better too, as quite often the test is something like "when_X_missing" and then never mentions "X". Now a test like that will remove "X" as it's first order of business.

tests/unit/lms/validation/_lti_launch_params_test.py
```diff
@@ -62,7 +83,10 @@ def set_url_(url):
...
 
-    @pytest.fixture
-    def pyramid_request(self, pyramid_request):
-        pyramid_request.content_type = "application/x-www-form-urlencoded"
-        return pyramid_request
+
+@pytest.fixture
+def pyramid_request(pyramid_request):
+    pyramid_request.content_type = "application/x-www-form-urlencoded"
+    pyramid_request.params.update({"resource_link_id": "DUMMY-LINK"})
```
jon-betts: Due to the addition of this requirement, we now need this for both schema to work

tests/unit/lms/views/error_test.py
```diff
@@ -78,6 +78,15 @@ class TestHTTPServerError(ExceptionViewTest):
     expected_result = {"message": exception.args[0]}
```
jon-betts: Oops, looks like a test got lost in the mix somewhere

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -0,0 +1,99 @@
...
+        lti_params.pop("resource_link_id")
+        lti_params.pop("launch_presentation_return_url")
+
+        result = self.lti_launch(app, lti_params, status=422)
+
+        self.assert_response_is_html(result)
+        assert "resource_link_id" in result
+
+    # ---------------------------------------------------------------------- #
+    # Assertions
```
seanh: I think we can do without these divider comments. We always put all the helper methods, fixtures etc at the bottom of a class or module. And I think it's pretty clear that if something doesn't begin with `def test_` it's not a test, if it has `@pytest.fixture` it's a fixture, if it begins with `def assert_` it's for doing asserts, if none of the above it's just a helper

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @classmethod
+    def assert_response_is_html(cls, response):
+        assert response.headers["Content-Type"] == Any.string.matching("^text/html")
+        assert response.html
```
seanh: This seems like a case for a matcher. With the old-style matchers I would just create a matcher class called `NonEmptyHTMLResponse` or something. Not sure what exactly the interface would be with the new matchers but I'm sure it could be done.

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -0,0 +1,99 @@
...
+            params=params,
+            headers={
+                "Accept": "text/html",
+                "Content-Type": "application/x-www-form-urlencoded",
+            },
+            status=status,
+        )
+
+    @classmethod
+    def _oauth_sign_params(cls, url, params):
```
seanh:
```suggestion
    def oauth_sign_params(cls, url, params):
```

Test classes don't have interfaces :man_shrugging:

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -0,0 +1,99 @@
...
+    OAUTH_CLIENT = oauthlib.oauth1.Client(OAUTH_CONSUMER_KEY, SHARED_SECRET)
+
+    def test_a_good_request_loads_fine(self, app, lti_params):
+        result = self.lti_launch(app, lti_params, status=200)
+
+        self.assert_response_is_html(result)
+
+        # Check a random parameter to see it's passed to the body
+        assert "tool_consumer_instance_guid" in result.text
+        assert lti_params["tool_consumer_instance_guid"] in result.text
```
seanh: This can be done in a follow-up PR but I think it might be nice to keep `tests/functional/lti_certification/` to strictly test cases that come directly from the certification tool. We can add whatever tests we want outside of this directory. Section 2 of the tests actually does contain a bunch of happy path tests (see https://github.com/hypothesis/lms/issues/932#issuecomment-550422989) so this test could become `test_2_1_launch_as_an_instructor()` and others

tests/functional/lti_certification/test_lti_certification_1_1.py
jon-betts: Might be tricky depending on the HTML response object chosen. I don't think there would be a general solution.

tests/functional/lti_certification/test_lti_certification_1_1.py
jon-betts: I was thinking of this to add some of the throw random crap at it type tests

tests/functional/lti_certification/test_lti_certification_1_1.py
jon-betts: I'll pluck them out in the next one, just helps my eyes keep things straight

Issue #1111, opened 2019-11-06
LTI Certification Test 1.1: Return user to tool consumer with error message when no resource_link_id


seanh:
Sentry crash: https://sentry.io/organizations/hypothesis/issues/1307186234/

```
context_id=con-182
context_label=SI182
context_title=Design of Personal Environments
context_type=CourseSection
custom_context_memberships_url=https://apps.imsglobal.org/lti/cert/tp/tp_membership.php/context/con-182/membership?b64=a2puNjk3b3E5YTQ3Z28wZDRnbW5xYzZyYjU%3D
custom_context_setting_url=https://apps.imsglobal.org/lti/cert/tp/tp_settings.php/lis/CourseSection/con-182/bindings/ims/cert/custom?b64=a2puNjk3b3E5YTQ3Z28wZDRnbW5xYzZyYjU%3D
custom_link_setting_url=$LtiLink.custom.url
custom_system_setting_url=https://apps.imsglobal.org/lti/cert/tp/tp_settings.php/ToolProxy/Hypothesis1b40eafba184a131307049e01e9c147d/custom?b64=a2puNjk3b3E5YTQ3Z28wZDRnbW5xYzZyYjU%3D
custom_tc_profile_url=https://apps.imsglobal.org/lti/cert/tp/tp_tcprofile.php?b64=a2puNjk3b3E5YTQ3Z28wZDRnbW5xYzZyYjU%3D
launch_presentation_document_target=iframe
launch_presentation_locale=en_US
launch_presentation_return_url=https://apps.imsglobal.org/lti/cert/tp/tp_return.php/basic-lti-launch-request
lis_course_section_sourcedid=id-182
lis_person_contact_email_primary=jane@school.edu
lis_person_name_family=Lastname
lis_person_name_full=Jane Q. Lastname
lis_person_name_given=Jane
lis_person_sourcedid=school.edu:jane
lti_message_type=basic-lti-launch-request
lti_version=LTI-1p0
oauth_callback=about:blank
oauth_consumer_key=Hypothesis1b40eafba184a131307049e01e9c147d
oauth_nonce=4d352ca11be546e8eba87e806bad5747
oauth_signature=zGSw/n81lXXFmMwepN9rE5NWHYQ=
oauth_signature_method=HMAC-SHA1
oauth_timestamp=1573044336
oauth_version=1.0
resource_link_title=Link 1234
roles=Instructor
tool_consumer_info_product_family_code=imsglc
[...]

PR #1176, opened 2019-11-14
LTI Certification 1.1 compliance - Part 1.1

jon-betts:
For ticket: https://github.com/hypothesis/lms/issues/1111

 * Introduced a new exception LTIToolRedirect
 * We raise this in the validator when things have gone wrong that
   should be reported to the user in the original tool
 * Added tests to cover

### DIFF
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -1,9 +1,14 @@
 from urllib.parse import unquote
 
-from marshmallow import fields, post_load
+from marshmallow import Schema, ValidationError, fields, post_load
 
 from lms.validation._exceptions import LTIToolRedirect
 from lms.validation._helpers import PyramidRequestSchema
+
+
+class LaunchPresentationReturnURLSchema(Schema):
+    """Schema containing only validation for the return URL."""
+    launch_presentation_return_url = fields.URL()
 
 
 class LaunchParamsSchema(PyramidRequestSchema):
@@ -15,29 +20,31 @@ class LaunchParamsSchema(PyramidRequestSchema):
     """
 
     resource_link_id = fields.Str(required=True)
-    launch_presentation_return_url = fields.URL()
+    launch_presentation_return_url = fields.Str()
 
-    # If we have an error in one of these fields we should redirect to the LTI
-    # app if possible
+    # If we have an error in one of these fields we should redirect back to
+    # the calling LMS if possible
     lti_redirect_fields = {"resource_link_id"}
 
     locations = ["form"]
 
-    # The parent params are all specified with typing stuff and are pretty
-    # crazy, so we'll disable the warnings to get slightly easier to understand
-    # params
-    def handle_error(self, error, data, **kwargs):  # pylint: disable=arguments-differ
-
+    def handle_error(self, error, data, *, many, **kwargs):
         messages = error.messages
-        valid_data = error.valid_data
 
-        return_url = valid_data.get("launch_presentation_return_url")
+        try:
+            # Extract the launch_presentation_return_url and check it's a real
+            # URL
+            return_url = LaunchPresentationReturnURLSchema().load(data).get(
+                "launch_presentation_return_url")
+        except ValidationError:
+            return_url = None
+
         if return_url:
             reportable_fields = set(messages.keys()) & self.lti_redirect_fields
             if reportable_fields:
                 raise LTIToolRedirect(return_url, messages)
 
-        super().handle_error(error, data, **kwargs)
+        super().handle_error(error, data, many=many, **kwargs)
 
 
 class LaunchParamsURLConfiguredSchema(LaunchParamsSchema):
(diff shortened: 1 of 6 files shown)


lms/validation/_lti_launch_params.py
```diff
@@ -14,9 +15,26 @@ class LaunchParamsSchema(PyramidRequestSchema):
     """
 
     resource_link_id = fields.Str(required=True)
+    launch_presentation_return_url = fields.URL()
+
+    # If we have an error in one of these fields we should redirect to the LTI
+    # app if possible
+    lti_redirect_fields = {"resource_link_id"}
```
jon-betts: This is a bit daft now, but we know more are coming

lms/validation/_lti_launch_params.py
```diff
@@ -14,9 +15,26 @@ class LaunchParamsSchema(PyramidRequestSchema):
...
 
     locations = ["form"]
 
+    def handle_error(self, error, data, **kwargs):
+        messages = error.messages
+        valid_data = error.valid_data
```
jon-betts: This is how you get at the previously validated fields

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -43,6 +54,15 @@ def assert_response_is_html(cls, response):
...
         assert response.html
 
+    def assert_redirected_to_tool_with_message(
+        self, response, lti_params, message=Any.string()
+    ):
+        expected_url = lti_params["launch_presentation_return_url"]
+        url = urlparse(response.headers["Location"])
+
+        assert url._replace(query=None).geturl() == expected_url
+        assert parse_qs(url.query) == Any.dict.containing({"lti_msg": [message]})
```
jon-betts: Compare the URLs without query params first then check the dict contains the bit we care about. Other stuff is allowed too and will be preserved (see the LTIToolRedirect tests for that)

jon-betts:
There's some lint stuff to do, but this is working in the LTI cert tool:

lti_cert_result

lms/validation/_lti_launch_params.py
```diff
@@ -14,9 +15,30 @@ class LaunchParamsSchema(PyramidRequestSchema):
...
+
+    # If we have an error in one of these fields we should redirect to the LTI
+    # app if possible
+    lti_redirect_fields = {"resource_link_id"}
 
     locations = ["form"]
 
+    # The parent params are all specified with typing stuff and are pretty
+    # crazy, so we'll disable the warnings to get slightly easier to understand
+    # params
```
seanh: This comment confused me for a while. I wasn't sure what the "parent" being referred to was and didn't realise that this comment had anything to do with/was explaining the reason for the `# pylint` comment below.

You might want to update the comment: I don't think the triggering of the pylint warning has anything to do with typing. And I don't think marshmallow's params are particularly crazy here, if you remove the typing they can be written as just `(self, error, data, *, many, **kwargs)` and that won't trigger a pylint warning.

The problem that's triggering the pylint warning seems to be that you have to keep the `many` split out, you can't merge it into `kwargs`, which may be an issue with pylint:

* * *

You can avoid the pylint warning without using any typing, like this:

```python
def handle_error(self, error, data, *, _many, **kwargs):
```

But, I think the fact that your current signature `(self, error, data, **kwargs)` triggers the pylint warning may be a bug in pylint. I may be wrong, but aren't `**kwargs` keyword-only arguments? So just `**kwargs` is the same as `*, _many, **kwargs)`. Could be wrong. But I think pylint might be wrong here.

jon-betts: I don't really know, what `*` is doing here. You can't have `_many` because that is different too. However you can have:
```
def handle_error(self, error, data, *, many, **kwargs)
```
As long as you pass many onto your call to super it seems happy.

lms/validation/_lti_launch_params.py
```diff
@@ -14,9 +15,30 @@ class LaunchParamsSchema(PyramidRequestSchema):
     """
 
     resource_link_id = fields.Str(required=True)
+    launch_presentation_return_url = fields.URL()
```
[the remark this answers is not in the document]
jon-betts: We now treat this as a string and use a smaller private schema to validate the URL on it's own.

lms/validation/_lti_launch_params.py
```diff
@@ -14,9 +15,30 @@ class LaunchParamsSchema(PyramidRequestSchema):
...
+
+    # If we have an error in one of these fields we should redirect to the LTI
+    # app if possible
```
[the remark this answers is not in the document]
jon-betts: Yeah. I'm not sure what the spec actually says. I was going by the tool description. I thought a minimalist approach was best, but I suspect that's where this is going.

lms/validation/_lti_launch_params.py
seanh: `_many` is ok, pylint will allow that.

The `*` enforces that `many` / `_many` can _only_ be used as a keyword argument, never as a positional argument

PR #1177, opened 2019-11-14
Don't show our grading UI in Canvas

seanh:
Oops


jon-betts:
Bombs away. This needs some unit-tests to prevent regression here. This is being done out of order due to the urgency.

PR #1179, opened 2019-11-14
Adding testing for LTI 1.1 Certification - Parts 1.8 & 1.9

jon-betts:
From https://www.imsglobal.org/specs/ltiv1p1/implementation-guide (search "Basic Launch Data") the only acceptable value is "basic-lti-launch-request".

Addresses:

 * https://github.com/hypothesis/lms/issues/1118 - LTI message type invalid
 * https://github.com/hypothesis/lms/issues/1119 - LTI message type missing


jon-betts:
I've tested this in the tool and it seems to pass.

PR #1181, opened 2019-11-14
LTI 1.1 Certification - Parts 4.5 and 4.6

jon-betts:
This is about running the tool without providing context information we have and addresses: https://github.com/hypothesis/lms/pull/1181


jon-betts:
I've tested this in the tool and it looks good in the tests there (we get a message, even if it's quite unfriendly)

PR #1182, opened 2019-11-15
Adding BDD tests for LTI certification sections 1-4

jon-betts:
At the moment this is just a dump of all the parameters and details from the LTI cert tools

 * https://www.imsglobal.org/lti-v1-certification-suite


jon-betts:
I don't think there's anything in here we need any more.

PR #1184, opened 2019-11-15
Adding the schema to the resource instead of the view

jon-betts:
This is a hacky fix to get us going for now


seanh:
Rebased...

PR #1190, opened 2019-11-18
Removing view decorators with services and inlining

jon-betts:
## What this fixes

 * Validation is now called before code that requires validated parameters
 * Decorators and other "helper" style code has been moved into services and views
 * Views which were previously in the `BasicLTILaunchView` class, but behaved differently have now been split

## Rationale

The application order is something like:

 * resources
 * decorators <<< code of interest here
 * schema
 * view

We are running into trouble, because our schema is the first line of defence against dodgy requests. The issue is that the decorators are using functions in the resource which access values which have not been checked yet.

The plan is move the code in the decorators into the view so that it is executed after schema validation like so:

 * resources
 * schema
 * view <<< code of interest here

It will still call code in the resources potentially, but by this time the schema will have had a chance to kick in and check for errors.

## Major changes

 * `lms/views/decorators/h_api.py` -> `lms/services/lti_hypothesis_bridge.py` class `LTIHypothesisBridge`
 * Moved `lms.views.decorators._helpers` one function as a private function of  `LTIHypothesisBridge`
 * Views now directly call the service
 * In-lined other decorators into the views that called them as class methods
 * Split `lms/views/basic_lti_launch.py` into:
  * `lms/views/lti/__init__.py` - A base class
  * `lms/views/lti/basic_launch.py` - All launch end-points
  * `lms/views/lti/configure_module.py` - The configure module item end-point
 * Merged, split and generally tried to move around tests to match
 * Extended functional tests to cover configured and un-configured end-points

## Review notes

 * The `LTIHypothesisBridge` class isn't a service in the Pyramid sense of the word. It's just a class with class methods.
 * There are a few places where "canvas" is mentioned. Not sure if this is right in an "everyone has grading" world
 * There are a few placeholdery names in there as I'm not quite sure how to describe what is being done. Better suggestions are very welcome.
 * As many files have been split and merged, git's tracking of when files have been "renamed" is pretty suspect. You may find it easier to just review the files a-new.
 * I've split off `ConfigureModuleItemView` as a child of `LTIViewBaseClass`, but I don't know if it actually really needs all the setup that `LTIViewBaseClass` does? There's a lot of conditional stuff in there, it could be that none of it actually kicks in in the case of `ConfigureModuleItemView`.
 * `@view_defaults` decorators in a class hierarchy don't appear to inherit from each other as I'd expect. If there's a better way of doing this there is some repetition in those decorators.


lms/services/lti_hypothesis_bridge.py
```diff
@@ -107,36 +113,29 @@ def wrapper(context, request):  # pylint:disable=too-many-branches
...
-        except HAPIError as err:
-            raise HTTPInternalServerError(explanation=err.explanation) from err
-
-        return wrapped(context, request)
-
-    return wrapper
+            cls._upsert_group_info(context, request)
+
+    @classmethod
+    def _upsert_group_info(cls, context, request):
```
jon-betts: This is the function that was originally in `lms/views/decorators/_helpers.py`. It's only called here and is really part of `add_user_to_group()`.

lms/views/lti/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
+
+        # Configure the front-end mini-app to run.
+        self.context.js_config.update({"mode": "basic-lti-launch"})
+
+        # Add config used by frontend to call `record_submission` API.
+        params = self.request.params
+        if (
```
jon-betts: What is this `if` actually testing as a description? It would be nice to tuck this into a named private function

lms/views/lti/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+            # The outcome reporting params are typically only available when
+            # students (not teachers) launch an assignment.
+            params.get("lis_result_sourcedid")
+            and params.get("lis_outcome_service_url")
+            # This feature is initially restricted to Canvas.
+            and params.get("tool_consumer_info_product_family_code") == "canvas"
```
jon-betts: "canvas" - Is this right? Should this be on for all LMS's now?

lms/views/lti/basic_launch.py
```diff
@@ -8,81 +8,34 @@
 
 The spec requires Basic LTI Launch requests to have an ``lti_message_type``
 parameter with the value ``basic-lti-launch-request`` to distinguish them
-from other types of launch request (other "message types") but our code
-doesn't actually require basic launch requests to have this parameter.
```
jon-betts: We do check now

lms/views/lti/basic_launch.py
```diff
@@ -8,81 +8,34 @@
...
-        upsert_lis_result_sourcedid,
-    ],
     permission="launch_lti_assignment",
     renderer="lms:templates/basic_lti_launch/basic_lti_launch.html.jinja2",
     request_method="POST",
     route_name="lti_launches",
     schema=LaunchParamsSchema,
 )
-class BasicLTILaunchViews:
-    def __init__(self, context, request):
```
jon-betts: All gone to the base class

lms/views/lti/basic_launch.py
```diff
@@ -243,65 +201,55 @@ def unconfigured_basic_lti_launch_not_authorized(self):
...
-        This happens after an unconfigured assignment launch. We show the user
-        our document selection form instead of launching the assignment, and
-        when the user chooses a document and submits the form this is the view
-        that receives that form submission.
-
-        We save the chosen document in the DB so that subsequent launches of
-        this same assignment will be DB-configured rather than unconfigured.
-        And we also send back the assignment launch page, passing the chosen
-        URL to Via, as the direct response to the content item form submission.
+        TODO! - What is this really up to? All sorts. Think of a better name
```
jon-betts: I'm not sure how to summarise what this is doing, part of it is reporting to H, part is just local stuff. Suggestions very welcome? Maybe two functions?

lms/views/lti/basic_launch.py
```diff
@@ -243,65 +201,55 @@ def unconfigured_basic_lti_launch_not_authorized(self):
...
-            self.context.js_config["submissionParams"][name] = value
-
-    def _set_focused_user(self, username):
-        """Configure the Hypothesis client to focus on a particular user."""
-
-        h_api_client = self.request.find_service(name="h_api_client")
+        if (
+            request.lti_user.is_instructor
+            or request.params.get("tool_consumer_info_product_family_code") == "canvas"
+        ):
```
jon-betts: Another "canvas" instance here. Is this right?

tests/conftest.py
```diff
@@ -44,7 +44,7 @@
     "h_jwt_client_secret": "TEST_JWT_CLIENT_SECRET",
     "h_authority": "TEST_AUTHORITY",
     "h_api_url_public": "https://example.com/api/",
-    "h_api_url_private": "https://private.com/api/",
+    "h_api_url_private": "https://example.com/private/api/",
```
jon-betts: Yeah... uhm I checked what `private.com` is. I would suggest not doing that in public.

tests/functional/lti_certification/test_lti_certification_1_1.py
```diff
@@ -16,15 +18,12 @@ class TestLTICertification(TestBaseClass):
...
 
+    @pytest.mark.usefixtures("http_intercept")
     def test_a_good_request_loads_fine(self, app, lti_params_1x):
         result = self.lti_launch(app, lti_params_1x, status=200)
 
         self.assert_response_is_html(result)
 
-        # Check a random parameter to see it's passed to the body
-        assert "tool_consumer_instance_guid" in result.text
-        assert lti_params_1x["tool_consumer_instance_guid"] in result.text
```
jon-betts: Some of the output doesn't do this, and frankly we don't care. 200 and HTML is good enough

tests/unit/lms/views/content_item_selection_test.py
```diff
@@ -118,3 +118,8 @@ def helpers(patch):
     helpers = patch("lms.views.content_item_selection.helpers")
     helpers.canvas_files_available.return_value = True
     return helpers
+
+
+@pytest.fixture(autouse=True)
+def LTIHypothesisBridge(patch):
+    return patch("lms.views.content_item_selection.LTIHypothesisBridge")
```
jon-betts: Simple classes with class methods as "services" are super easy to mock out

tests/unit/lms/views/lti/__init___test.py
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+    @pytest.mark.parametrize(
+        "key,value",
+        [
+            ("lis_result_sourcedid", None),
+            ("lis_outcome_service_url", None),
+            ("tool_consumer_info_product_family_code", None),
+            # Anything non canvas should do here
+            ("tool_consumer_info_product_family_code", "whiteboard"),
```
jon-betts: Again... is this right. Does "canvas" behave differently than everything else?

jon-betts:
So this is probably never getting merged as it's such a _phatt boi_, so I'm going to try and split more minimal parts off.

Shame, because I think it's beautiful.

PR #1191, opened 2019-11-19
First stab at splitting out LTI-H service

jon-betts:
Also adding func-tests to cover the LTI failures we get in the spec test


seanh:
I've squashed the commits and wrote a nicer commit message

PR #1193, opened 2019-11-20
Moving all H API functionality to lms.services.h_api.HAPI

jon-betts:
This is collecting from LTIHService and HAPIRequests

Once again this has eliminated code, so our coverage appears to have gone down, despite being fine.


lms/services/h_api.py
```diff
@@ -0,0 +1,177 @@
...
+        Add the user as a member of the group.
+
+        :param h_user: An `HUser` to associate with the group
+        :param group_id: The id of the group
+        """
+        self._api_request("POST", f"groups/{group_id}/members/{h_user.userid}")
+
+    def _api_request(
+        self, method: str, path: str, data: dict = None, headers: dict = None
+    ):
```
jon-betts: All options that were not in use in this class have been removed, simplifying the function. Type declarations have moved to the signature to clean up the docs. I also removed some redundancy in the description describing the `:raise:` portion again.

lms/services/lti_h.py
```diff
@@ -92,38 +76,38 @@ def upsert_course_group(self):
...
-        )
-
         try:
             try:
-                hapi_svc.patch(
-                    f"groups/{self._context.h_groupid}",
-                    {"name": self._context.h_group_name},
+                self.h_api.update_group(
+                    group_id=self._context.h_groupid,
+                    group_name=self._context.h_group_name,
```
jon-betts: I think this is much clearer about our intent

tests/unit/lms/services/lti_h_test.py
```diff
@@ -23,8 +21,7 @@ def test_it_does_nothing_if_the_feature_isnt_enabled(
 
         lti_h_svc.upsert_course_group()
 
-        hapi_svc.patch.assert_not_called()
-        hapi_svc.put.assert_not_called()
+        h_api.update_group.assert_not_called()
```
jon-betts: These tests make much more sense now, rather than making generic assertions about patch and put

tests/unit/lms/views/index_test.py
```diff
@@ -0,0 +1,8 @@
+from lms.views.index import index
```
jon-betts: This is only here to pass the coverage ratchet. As this PR eliminates well tested code, the overall coverage percentage has decreased.

lms/services/lti_h.py
```diff
@@ -89,40 +84,28 @@ def upsert_course_group(self):
...
+            h_user=self._context.h_user,
+        )
 
-        hapi_svc = self.h_api_requests_service
+        self._upsert_group_info()
 
-        is_instructor = any(
-            role in self._request.lti_user.roles.lower()
-            for role in ("administrator", "instructor", "teachingassisstant")
-        )
```
jon-betts: This whole thing was already on the `lti_user`

lms/services/lti_h.py
```diff
@@ -1,17 +1,39 @@
+from functools import wraps
+
 from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
 
 from lms.services import HAPIError, HAPINotFoundError
 
 
+def lti_h_action(function):
+    """Disable an action if provisioning is not enabled and catch HAPIError."""
```
jon-betts: This decorator simplifies the code a lot by removing the repeated disabling and catching. Doesn't have a very informative name though. `@disable_if_not_provisioned_and_catch_hapi_errors` seems a bit much

lms/services/lti_h.py
```diff
@@ -1,17 +1,39 @@
+from functools import wraps
```
jon-betts: This whole file looks horrible in the diff, but if you read the plain code it's miles more clear

lms/services/h_api.py
```diff
@@ -0,0 +1,180 @@
...
+        :return: the response from the h API
+        :rtype: requests.Response
+        """
+        headers = headers or {}
+        headers["Hypothesis-Application"] = "lms"
+
+        request_args = {"headers": headers}
+
+        if data is not None:
+            request_args["data"] = json.dumps(data, separators=(",", ":"))
```
jon-betts: The separators here compactify the JSON. Not much but not really any point in wasting the space in the request

tests/unit/lms/views/index_test.py
```diff
@@ -0,0 +1,8 @@
...
+
+
+class TestIndexView:
+    def test_it_does_nowt(self, pyramid_request):
+        template_params = index(pyramid_request)
+
+        assert template_params == {}
```
seanh: If you wanted, this could just be:

```python
    def test_it_does_nowt(self, pyramid_request):
        assert index(pyramid_request) == {}
```

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -102,9 +102,9 @@ def test_it_uses_placeholder_display_name_for_focused_user_if_api_call_fails(
         }
 
     @pytest.fixture
-    def h_api_client(self, pyramid_config):
-        svc = mock.create_autospec(HAPIClient, instance=True)
-        pyramid_config.register_service(svc, name="h_api_client")
+    def h_api(self, pyramid_config):
+        svc = mock.create_autospec(HAPI, instance=True)
```
seanh:
```suggestion
        svc = mock.create_autospec(HAPI, instance=True, spec_set=True)
```

Not sure why the `spec_set=True` was missing here already but you always want it unless you know you don't

tests/unit/lms/services/lti_h_test.py
```diff
@@ -281,21 +243,27 @@ def lti_h_svc(pyramid_request):
...
+def h_user():
+    return HUser(
+        authority="TEST_AUTHORITY",
+        username="test_username",
+        display_name="test_display_name",
+    )
+
+
+@pytest.fixture
+def context(h_user):
```
seanh: Composing fixtures is great :+1:

lms/services/lti_h.py
jon-betts: It would still be nice to catch the exception here, as that's universal too. Although that would fix the naming issue for the decorator.

The NoOpService would work, but I wonder if we can just not call it in the first place? It's a kind of obscure way to hide the if. I can see this being one of the things that might have us scratching our heads 2 years down the line if it's still there.

tests/unit/lms/services/h_api_test.py
```diff
@@ -0,0 +1,218 @@
...
+        return HAPI(sentinel.context, p_request)
+
+
+@pytest.fixture(autouse=True)
+def requests(patch):
+    return patch("lms.services.h_api.requests")
+
+
+@pytest.fixture
+def p_request():
```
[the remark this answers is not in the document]
jon-betts: It's because it literally wasn't working with this change. I had no idea why. It wasn't cascading as I would expect. I can give it another shot.

tests/unit/lms/views/index_test.py
jon-betts: Yep, but in general in all of the other tests we tend to assign to a variable and then test that variable. I assumed this was for debugging purposes.

jon-betts:
## Re: typing

With regards to the type annotations, I only put them there because the original doc string specified the type of every part of the signature. I don't really think that's necessary at all.

We don't have to go about saying every argument is a string etc. If we feel the need to do that, then we probably shouldn't be using Python. 

So I think I'll just scrub them out.

## Re: params as types

To get linting to shut up you have to mention every parameter. Often there isn't much to say. I don't actually think bloating out the descriptions with lots of types in general makes them nicer to read. It just ends up a TL;DR block of text so sometimes it's a bit naffer. We could _not_ mention the type at all, but I'm not sure that's better?

## Re: HAPI vs HAPIClient

It seems quite clear to me from context that we are consuming rather than providing the API in this case. If API isn't doing the work of telling us that then is `HClient` a better name. Presumably it's inherent here that we are doing so via an API rather than ???.

## Re: splitting HAPI

I think if you're going to do the effort of making something re-usable and actually re-using it, then this is probably a good idea. I tried to de "H" if it, and make it more general in a prelude to perhaps creating an HTTPClient base class or something similar. But I think we should wait until we actually want to do that.

At the moment, I think it's less clear to have the code split across two files. 150 lines of code (a good 50 of which are comments and doc strings) just isn't a lot of code.

Our tests use plenty of mocking fanciness all over the place, so I don't think this is any more or less "fancy" than other mock we do. I suspect it's just less familiar.

jon-betts:
## Re: raise

Using `:raise:` versus `:raises:` seems fine. 

I don't see the point in repeating in each and every doc string that the method can raise a particular error if we mention it at the class level. It just drops the signal to noise ratio. I think general behavior can and should be described at the class level (for example if every method has other expected conditions or behavior common to all). Otherwise I think we invite copy pasting a long description into each function.

jon-betts:
## Re: arg vs param

`:param:` seems to be the most common approach mentioned in online sources. It's the PyCharm and sphinx default I think. So I'd prefer to stick with it if possible. It also means I don't have to try and fight pycharm to insert it.

seanh:
Was just testing this in Canvas: it seems to be failing with a "Could not find requested resource" error whenever a student tries to launch an assignment. Something no longer catching `HAPINotFound`, maybe?

lms/services/lti_h.py
```diff
@@ -1,128 +1,114 @@
...
+    def wrapper(self, *args, **kwargs):
+        # Pylint doesn't understand that self is "us" in this context
+        if not self._context.provisioning_enabled:  # pylint: disable=protected-access
+            return None
+
+        try:
+            return function(self, *args, **kwargs)
+
+        except HAPIError as err:
+            raise HTTPInternalServerError(explanation=err.explanation) from err
```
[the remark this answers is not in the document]
jon-betts: Well, that already does wrapping to produce the HAPIError. If all we do with that is catch that here to wrap it again, we could could out some middle man and not bother.

seanh:
Looks like an indentation error in `_upsert_h_user()`. It tries to create the group _if updating the group succeeds_. Should be if updating fails:

```diff
diff --git a/lms/services/lti_h.py b/lms/services/lti_h.py
index 3dfe0ac2..49e73fda 100644
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -107,8 +107,8 @@ class LTIHService:
             if not self._request.lti_user.is_instructor:
                 raise HTTPBadRequest("Instructor must launch assignment first.")
 
-        # Try to create the group with the current instructor as its creator.
-        self.h_api.create_group(group_id=group_id, group_name=group_name, h_user=h_user)
+            # Try to create the group with the current instructor as its creator.
+            self.h_api.create_group(group_id=group_id, group_name=group_name, h_user=h_user)
 
     def _upsert_group_info(self):
         """Create or update the GroupInfo for the given request."""
```

lms/services/h_api.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Add the user as a member of the group.
+
+        :param h_user: An `HUser` to associate with the group
+        :param group_id: The id of the group
+        """
+        self._api_request("POST", f"groups/{group_id}/members/{h_user.userid}")
+
+    def _api_request(
+        self, method: str, path: str, data: dict = None, headers: dict = None
```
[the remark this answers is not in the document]
jon-betts: As mentioned below, it's because the doc string was totally over the top before. I'm just going to remove this. We don't need to mention every type of every string in our app in doc strings.

seanh:
Other than that one issue it tests good :+1: Should make sure we have a failing unit test for the above case though

PR #1196, opened 2019-11-21
Move group info logic out of LTI-H to Group Info service

jon-betts:
## What has changed

 * The `GroupInfoUpsert` service is now called `GroupInfoService`
 * It now has a single function instead of being callable, so it resembles our other services
 * `LTIHService` is no longer responsible for picking which `GroupInfo` fields to update
 * The list of fields to update is read from the `GroupInfo` model class

## How is this better?

 * LTI-H is easier to read as it doesn't have a random bunch of GroupInfo stuff in it
 * The model class is now in charge of which fields are metadata or not
   * As a bonus if you add a field to that class it will all just continue to work
   * Contrast this from before where you would have to find a random helper in views and update that too
 * The testing for each service becomes easier as the one below becomes genuinely separated (we no longer have to check that GroupInfo models are inserted in our LTI-H tests etc.)

## Hopefully it's easier to understand now

Initially to understand how we performed GroupInfo upserts you would have to look:

 * In a view and notice a decorator
 * Follow that decorator to see it called a helper and in what circumstances
 * This helper contained part of the code and then handed off to GroupInfoUpsert
 * Read GroupInfoUpsert and remember what you'd seen in the helper

The same journey now is:

 * See the view directly call LTI-H service
 * See LTI-H service call GroupUpsert service (where it all is now pretty much)

This has been branched off of https://github.com/hypothesis/lms/pull/1193 and should be rebased and merged after that.

# How the code has migrated over the commits

This code has taken quite an interesting journey over the refactors:

 * It started off in a view helper that was called by a view decorator
 * It was then in-lined into the newly created LTI-H service
 * As that service was separated into H-API and LTI-H this stuck out like a sore thumb
 * Then it was moved into the GroupInfo service finally to come to rest 50% there and 50% in the model class


lms/db/__init__.py
```diff
@@ -3,21 +3,42 @@
...
 
 log = logging.getLogger(__name__)
 
 
+class BaseClass:
+    """Functions common to all SQL Alchemy models."""
+
+    @classmethod
+    def iter_properties(cls):
+        """Iterate over all declared SQL Alchemy properties."""
```
seanh:
```suggestion
        """Iterate over all declared SQLAlchemy properties."""
```

lms/db/__init__.py
```diff
@@ -3,21 +3,42 @@
...
+        return inspect(cls).iterate_properties
+
+    @classmethod
+    def iter_columns(cls):
+        """Iterate over all declared SQL Alchemy columns."""
```
seanh:
```suggestion
        """Iterate over all declared SQLAlchemy columns."""
```

lms/models/group_info.py
```diff
@@ -88,3 +90,9 @@ def __repr__(self, _class_name="lms.models.GroupInfo"):
             f" authority_provided_id:'{self.authority_provided_id}'"
             f" consumer_key:'{self.consumer_key}'>"
         )
+
+    @classmethod
+    def metadata_fields(cls):
+        for column in cls.iter_columns():
+            if column.key not in cls.NON_METADATA_FIELDS:
+                yield column.key
```
seanh: I'm having trouble figure out what a "metadata" field is vs a "non-metadata" field, semantically. Might need another name? Or a docstring?

lms/services/group_info.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    A callable that upserts :class:`lms.models.GroupInfoService` records.
+
+    Usage:
+
+        group_info = request.find_service(name="group_info")
+        group_info.upsert(authority_provided_id, consumer_key, request.params)
+    """
+
+    VALID_PARAMS = list(GroupInfo.metadata_fields())
```
seanh: Yeah, this is nice :+1:

lms/models/group_info.py
seanh: I think the difference is that the non-metadata fields are the ones that identify the row (id, authority_provided_id, and consumer_key) whereas the metadata fields are the ones that _don't_ identify the row and that might actually be changed on an update, and come from `request.params`. I can't think of a better name for it but maybe just an explanatory docstring would help

seanh:
FYI the reason it was called "group info _upsert_ service" rather than just "group info service" is that you might want more than one service per model. It doesn't really matter with GroupInfo (I don't think you will need more than one). But in h for example, we started off writing things like AnnotationService, UserService, and GroupService. But there turns out to be quite a lot of different things that you want to do with groups, for example. Or different ways you want to retrieve groups. So a single GroupService would become too big. Hence a lot of small GroupSomethingServices.

That's one of the advantages of a "database services" layer above the models: there can be only one models.Group. But there can be lots of services.GroupSomethingService's.

In the case of a new model and service like GroupInfo and GroupInfoUpsert, calling it "upsert" is just starting off how you intend to finish and setting a good example. Proactively making lots of small classes rather than making big classes and then only splitting them up once they become unweildy (you will end up with lots of classes that are either too big and overdue for splitting up; or are quite big but just not quite big enough for anybody to have bothered refactoring yet; rather than having lots of nice small classes). Avoiding making it read as if the pattern is one model = one service.

lms/models/group_info.py
jon-betts: This might be named backwards, they are really the data fields. The `metadata` is presumably `id` etc.

jon-betts:
The main reason for the change with regards to 'GroupInfoUpsert' was the fact the class ended up as a callable, which is fine, but didn't fit with the pattern of other classes.

I think the 'metadata_fields' are named backwards. I'll take a look at that.

lms/services/group_info.py
```diff
@@ -0,0 +1,48 @@
+"""A service that upserts :class:`lms.models.GroupInfoService` records."""
+
+from lms.models import GroupInfo
+
+__all__ = ["GroupInfoService"]
+
+
+class GroupInfoService:
+    """
+    A callable that upserts :class:`lms.models.GroupInfoService` records.
```
seanh:
```suggestion
    A service that upserts :class:`lms.models.GroupInfoService` records.
```

lms/services/group_info.py
```diff
@@ -0,0 +1,48 @@
...
+    SKIP_COLUMNS = {"authority_provided_id", "id"}
+
+    def __init__(self, _context, request):
+        self._db = request.db
+
+    def upsert(self, authority_provided_id, consumer_key, params):
+        """
+        Upsert a row into the group_info DB table.
+
+        Finds a `GroupInfoService` row if available, and creates it if not, setting
```
seanh:
```suggestion
        Finds a ``GroupInfoService`` row if available, and creates it if not, setting
```

lms/services/group_info.py
```diff
@@ -0,0 +1,48 @@
...
+        the `consumer_key` and relevant values from `params`.
```
seanh:
```suggestion
        the ``consumer_key`` and relevant values from ``params``.
```

tests/unit/lms/services/lti_h_test.py
```diff
@@ -127,24 +113,13 @@ def test_it_doesnt_upsert_GroupInfo_into_the_db_if_creating_the_group_fails(
...
-            tool_consumer_instance_name="test_tool_consumer_instance_name",
-            tool_consumer_instance_description="test_tool_consumer_instance_description",
-            tool_consumer_instance_url="test_tool_consumer_instance_url",
-            tool_consumer_instance_contact_email="test_tool_consumer_instance_contact_email",
-            tool_consumer_instance_guid="test_tool_consumer_instance_guid",
-            custom_canvas_api_domain="test_custom_canvas_api_domain",
-            custom_canvas_course_id="test_custom_canvas_course_id",
-        )
+        return {
+            field.key: "TEST_" + field.key.upper() for field in GroupInfo.iter_columns()
```
seanh:
```suggestion
            field.key: f"TEST_{field.key.upper()}" for field in GroupInfo.iter_columns()
```

tests/unit/lms/services/group_info_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @pytest.fixture
+    def group_info_svc(self, pyramid_config, pyramid_request):
+        """The group_info_svc service that is being tested."""
+        return GroupInfoService(mock.sentinel.context, pyramid_request)
+
+    @pytest.fixture
+    def params(self):
+        return {
+            field.key: "TEST_" + field.key.upper() for field in GroupInfo.iter_columns()
```
seanh:
```suggestion
            field.key: f"TEST_{field.key.upper()}" for field in GroupInfo.iter_columns()
```

lms/services/group_info.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    A callable that upserts :class:`lms.models.GroupInfoService` records.
+
+    Usage:
+
+        group_info = request.find_service(name="group_info")
+        group_info.upsert(authority_provided_id, consumer_key, request.params)
+    """
+
+    # Things we don't try and update on GroupInfo objects from params
```
seanh:
```suggestion
    # GroupInfo columns that upsert() *doesn't* update, even if they're in the given params.
```

This might make more sense just inlined into `upsert()` as a local variable or even just a literal value

tests/unit/lms/db/__init___test.py
```diff
@@ -0,0 +1,61 @@
...
+        with pytest.raises(ValueError):
+            ModelClass().update_from_dict({}, skip_keys=["a"])
+
+    @pytest.fixture
+    def model(self):
+        model = ModelClass()
+        model.id = 1234
+        model.column = "original_value"
+
+        return model
```
seanh: This method can be shortened to:

```python
        return ModelClass(id=1234, column="original_value")
```

tests/unit/lms/services/group_info_test.py
```diff
@@ -0,0 +1,124 @@
...
+                authority_provided_id="IGNORE ME 2",
+                something_unrelated="IGNORED ME 3",
+            ),
+        )
+
+        group_info = self._get_inserted_group_info(db_session)
+
+        assert group_info.authority_provided_id == self.AUTHORITY
+
+    def _get_inserted_group_info(self, db_session):
```
seanh: I don't think leading `_`'s are needed in test class helper methods

tests/unit/lms/services/group_info_test.py
```diff
@@ -0,0 +1,124 @@
...
+            params=dict(params, context_title="NEW_TITLE"),
+        )
+
+        group_info = self._get_inserted_group_info(db_session)
+
+        assert group_info.consumer_key == self.CONSUMER_KEY
+        assert group_info.context_label == params["context_label"]
+        assert group_info.context_title == "NEW_TITLE"
+
+    def test_it_ignores_non_metadata_params(self, db_session, group_info_svc, params):
```
seanh:
```suggestion
    def test_it_ignores_authority_provided_id(self, db_session, group_info_svc, params):
```

This doesn't actually test that `id` is ignored

tests/unit/lms/services/group_info_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+import pytest
+
+from lms.models import ApplicationInstance, GroupInfo
+from lms.services.group_info import GroupInfoService
+
+
+class TestGroupInfoUpsert:
+    AUTHORITY = "TEST_AUTHORITY_PROVIDED_ID"
+    CONSUMER_KEY = "TEST_CONSUMER_KEY"
```
seanh: I'm not sure these add much beyond just using literal `"TEST_AUTHORITY_PROVIDED_ID"` wherever needed.

`AUTHORITY` should at least be renamed since it's not an authority (a word that has a specific meaning in Hypothesis) but an authority_provided_id.

tests/unit/lms/db/__init___test.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_we_can_specify_keys_to_skip(self, model):
+        model.update_from_dict(
+            {"id": 4321, "column": "new_value"}, skip_keys={"column"}
+        )
+
+        assert model.id == 4321
+        assert model.column == "original_value"
+
+    def test_we_fail_to_update_when_skip_keys_is_not_a_set(self):
+        with pytest.raises(ValueError):
```
seanh:
```suggestion
        with pytest.raises(
            ValueError,
            match="Expected a set of keys to skip but found '<class 'list'>'",
        ):
```

tests/unit/lms/db/__init___test.py
```diff
@@ -0,0 +1,61 @@
...
+class TestBase:
+    def test_we_can_get_properties(self):
+        keys = {prop.key for prop in ModelClass.iter_properties()}
+
+        assert {"id", "relationship", "column"} == keys
+
+    def test_we_can_get_columns(self):
+        keys = {prop.key for prop in ModelClass.iter_columns()}
+
+        assert {"id", "column"} == keys
```
seanh:
```suggestion
        assert keys == {"id", "column"}
```

tests/unit/lms/db/__init___test.py
```diff
@@ -0,0 +1,61 @@
...
+    id = sa.Column(sa.Integer, primary_key=True)
+    column = sa.Column(sa.Integer, sa.ForeignKey("child.id"))
+    relationship = sa.orm.relationship("Child")
+
+
+class TestBase:
+    def test_we_can_get_properties(self):
+        keys = {prop.key for prop in ModelClass.iter_properties()}
+
+        assert {"id", "relationship", "column"} == keys
```
seanh:
```suggestion
        assert keys == {"id", "relationship", "column"}
```

seanh:
I fixed the CI fail and squashed the commits and added a commit message. I think it's worthwhile to compose commit messages that explain what's being changed and why. Writing a commit message also revealed that there were a few separate changes being done in one commit here:

```
Factor out the _group_info_upsert() helper

Previously the _group_info_upsert() helper was a private helper in the
lti_h service that was responsible for extracting the whitelisted params
(the ones that correspond to columns on the group_info DB table) from
the request, and passing them in to the GroupInfoUpsert service.

Factor this private helper away by:

* Adding a new BaseClass that all SQLAlchemy models inherit from, and
  that provides:

  * iter_properties() and iter_columns() classmethods.
    iter_columns() returns an iterator over all of the model's SQLAlchemy
    columns.

  * An update_from_dict() method that updates the columns of a model
    object from a dict of params.

    Any key in the given dict whose name does not match one of the
    models columns is ignored.

* Having GroupInfoUpsert (now renamed to GroupInfoService) call this new
  update_from_dict() method passing in request.params directly.

  Even though request.params might contain other keys (that don't
  correspond to any of the model's columns) those will be ignored by
  update_from_dict().

  GroupInfoService also removes a couple of specific keys
  ("authority_provided_id" and "id") from the fields to be updated, even
  though those do correspond to columns. We don't want to update those
  columns.

This means that a whitelist of the fields to be updated is no longer
needed. The list is computed dynamically by getting the list of the
model's column names, then subtracting "authority_provided_id" and "id".

As a result, _group_info_upsert() is no longer needed, since that
whitelisting was all it provided,

In addition to the above refactoring, this commit also:

* Renames GroupInfoUpsert to GroupInfoService
* Renames GroupInfoUpsert/GroupInfoService's __call__ method to upsert()
* Refactors an includeme() test and some other tests
```

Also pushed some commits to do some tidy up, and made some further suggestions.

seanh:
It's late and I'm probably talking nonsense but I can't help thinking there's a generic crupdater service possible:

```python
crupdate.upsert(
    GroupInfo,
    filter_by=dict(
        authority_provided_id=self._context.h_authority_provided_id,
        consumer_key=self._request.lti_user.oauth_consumer_key,
    ),
    skip_columns=["authority_provided_id", "id"],
    request.params
)
```

```python
class Crupdate:
    def create(self, model_class, data):
        model_obj = model_class(**data)
        self.db.add(model_obj)
        return model_obj

    def get(self, model_class, filter_by):
        return db.query(model_class).filter_by(**filter_by).one_or_none()

    def upsert(self, model_class, filter_by, data, skip_columns=None):
        model_obj = self.get(model_class, filter_by)

        if not model_obj:
            model_obj = self.create(model_class, filter_by)

        self.update(model_obj, data, skip_columns)

    def update(model_obj, data, skip_columns=None):
        for key in self.iter_columns(model_obj, skip_columns):
            if key in data:
                setattr(model_obj, key, data[key])

    def iter_columns(self, model, skip_columns):
        """Return all of model's column names that aren't in skip_columns."""
        if skip_columns is None:
            skip_columns = # Compute the primary key columns of model_class.
        ...
```

tests/unit/lms/services/group_info_test.py
jon-betts: The intention is to: 

 * Prevent typos
 * Indicate to the reader that this is _the_ value, rather than just _a_ value when compared with
 * Generally it's a good idea in classes not to have repeated literal values. As in code, so in tests etc.

tests/unit/lms/services/group_info_test.py
jon-betts: I wonder if this is the correct way to fix this. The original name of the function states the intention of the test. If that isn't implemented well, then the test should probably change.

tests/unit/lms/db/__init___test.py
jon-betts: Spot the boy who hasn't done SQLAlchemy in the last 5 years.

tests/unit/lms/db/__init___test.py
jon-betts: Not sure what was going on there. There's another one below it too.

tests/unit/lms/db/__init___test.py
jon-betts: As a general rule, never test the string of an exception unless you are writing tests for the exception. Otherwise you get into the fun world of changing the wording of an exception message (which should by definition have no impact on the app, if it does you have other issues) and having all your tests break.

If you aren't getting enough information from the type of the exception then you probably introduce a subclass. In this case we probably should be raising a `TypeError` though: https://docs.python.org/3/library/exceptions.html#TypeError

> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError

I don't think you need a custom exception for every different example of I screwed up the arguments to a function though. The traceback should be enough to determine the origin, and this should really _never_ happen outside of development.

I will change this round to `TypeError` though, as that's more appropriate.

tests/unit/lms/services/group_info_test.py
jon-betts: I don't think I necessarily phrased that very well. What I mostly mean here, is that the name of a test function indicates what it is trying to prove, not necessarily how it does it.

For example `test_it_makes_it_upper_case()` will test some specific examples of upper case, but not all of them. So changing it to `test_it_makes_these_specific_strings_upper_case()` is more accurate description of the _implementation_, but a less accurate description of the _intention_. It also becomes very hard to think of how to write that test better if it just describes itself (it's kind of circular).

In most cases it's not really possible to test all possible inputs, so most tests necessarily pick some special values or circumstances to stand in for the general case. That's not actually true here as it's trivial to cover both options, so I'll add that in.

tests/unit/lms/services/group_info_test.py
jon-betts: Fixed

tests/unit/lms/db/__init___test.py
seanh: I think it's probably a good idea to raise custom exception types (see [Modules should never raise core Python exceptions](https://utcc.utoronto.ca/~cks/space/blog/python/NeverRaiseCoreExceptions) for a good post on this) or, in cases where you don't feel it's worthwhile to create your own exception class, then do test the string. That's been my practice so far.

The problem is just that because a built-in exception class is used any mistake in the test code within the `with` or anywhere in `BaseClass.__init__()` or `BaseClass.update_from_dict()` (or anything called by those) could raise `TypeError` and cause this test to pass when it shouldn't. The test just asserts that something raised `TypeError`. It doesn't actually test that the `raise` you think was hit was in fact hit.

We have actually run into this issue a number of times before: the code has a bug, but the tests were all passing, because the test was catching the wrong exception. When it goes wrong it leads to bugs being deployed and they can be difficult to debug. 

On the other hand, what's the disadvantage of testing the string here? If you edit the string that `BaseClass` raises (seems unlikely you'd ever edit that string) then the test will start failing and you'll have to edit the string in the test as well---that doesn't seem difficult or time consuming to me, compared to potentially deploying a bug or unexpected behavior.

tests/unit/lms/services/group_info_test.py
seanh: Alright, I get can get on board with this one. As as a general pattern, maybe all test classes should start declaring their CONSTANTS as class attrs.

I do wonder about putting these at the bottom of the class though (maybe above the class's fixtures and helper methods) instead of at the top. I know the top of a module or class is the normal place to put a constant. But we usually put all our test fixtures and helper methods at the bottom of the class just to get them out of the way, so you don't have to scroll past a bunch of noise and boilerplate before getting to the tests. After all I shouldn't usually need to know what the string value of `CONSUMER_KEY` is.

```diff
diff --git a/tests/unit/lms/services/group_info_test.py b/tests/unit/lms/services/group_info_test.py
index 8a944f08..b2194b09 100644
--- a/tests/unit/lms/services/group_info_test.py
+++ b/tests/unit/lms/services/group_info_test.py
@@ -7,9 +7,6 @@ from lms.services.group_info import GroupInfoService
 
 
 class TestGroupInfoUpsert:
-    AUTHORITY = "TEST_AUTHORITY_PROVIDED_ID"
-    CONSUMER_KEY = "TEST_CONSUMER_KEY"
-
     def test_it_adds_a_new_GroupInfo_if_none_exists(
         self, db_session, group_info_svc, params
     ):
@@ -75,6 +72,9 @@ class TestGroupInfoUpsert:
             .one()
         )
 
+    AUTHORITY = "TEST_AUTHORITY_PROVIDED_ID"
+    CONSUMER_KEY = "TEST_CONSUMER_KEY"
+
     @pytest.fixture
     def group_info_svc(self, pyramid_config, pyramid_request):
         """The group_info_svc service that is being tested."""
```

tests/unit/lms/services/group_info_test.py
seanh: Yes, that makes sense. Also realising this test was a perfect `@pytest.mark.parametrize` opportunity

PR #1198, opened 2019-11-21
Inlining _upsert_lis_result_sourceid decorator into the view

jon-betts:
This follows on from previous work and has the same separated unit tests which may not be the way to go.

Should be merged after: https://github.com/hypothesis/lms/pull/1197


jon-betts:
Killing this off for: https://github.com/hypothesis/lms/pull/1201

Issue #1220, opened 2019-11-27
Add a gherkin based test set to cover sections 1-4 of the LTI 1.1 spec


jon-betts:
This ticket is being created backwards, as most of this is already done.

## Requirements:

* [x] - We have test coverage of sections 1 - 4
* [x] - The build fails if the tests don't pass - https://github.com/hypothesis/lms/pull/1273
* [x] - The tests test configured assignments - https://github.com/hypothesis/lms/pull/1274
   * Extra credit: also fun with un-configured - Can't see a good way to do this, so going to skip
* [x] - Tests we have duplicated from the functional tests are removed - https://github.com/hypothesis/lms/pull/1275

## Implementation plan:

Split into multiple PRs with:

 * [x] All the fixture `.ini` files and the bin script that generates the `average.ini`'s from them
   *  https://github.com/hypothesis/lms/pull/1231
* [x] The Gherkin scenarios for all the tests, but without any contents yet
    * https://github.com/hypothesis/lms/pull/1233
* [x] Implement the 1.1 test in plain Behave (using step contexts; but not yet HOG)
    * This PR introduces the "step contexts" concept and base class. 
    * Only contains the step contexts needed for 1.1
    * https://github.com/hypothesis/lms/pull/1234
* [x] Higher Order Gerkinify-y the 1.1 test
    * This introduces the HOG concept and adds the HOG code itself. Only contains the feature steps needed for 1.1
   * https://github.com/hypothesis/lms/pull/1235
* [x] Multiple: The rest of the tests 
    * e.g. one PR per section of tests, adding the only what is needed for that section
   * [x] Section 1 - https://github.com/hypothesis/lms/pull/1236
   * [x] Section 2 - https://github.com/hypothesis/lms/pull/1237
   * [x] Section 3 - https://github.com/hypothesis/lms/pull/1238
   * [x] Section 4 - https://github.com/hypothesis/lms/pull/1239

jon-betts:
A PR adding the tests is here: https://github.com/hypothesis/lms/pull/1182

jon-betts:
Tickets have been added to address the outstanding bits, but they need a review etc.

jon-betts:
Finally, I think we're done!

PR #1233, opened 2019-12-02
Part 2: Adding the scenario skeletons from the LTI cert 1.1 test

jon-betts:
* This includes tests 1.x - 4.x + 6.x
* All were taken directly from the HTML of the certification test


jon-betts:
This is part 2 of a four-parter, so you're going to see some things in odd places. The parent is: #1220

jon-betts:
I'm not going to deploy, as there's no benefit, and it introduces a (small) risk.

Issue #1242, opened 2019-12-05
Move the LIS value object and validator into the service


jon-betts:
After the previous refactor, it's now become obvious that:

 * The validator and value are only used in the service
 * The value doesn't do anything a dict wouldn't do better
 * We can in-line the validator and remove the value
 * The tests from the previous PR already cover the validator's behavior

jon-betts:
This is done now. Sean finished up my PRs over the weekend.

PR #1247, opened 2019-12-05
Removing the LIS value object and in-lining the validator

jon-betts:
* We now have a small vanilla marshmallow schema in the services
 * We don't use an object at all
 * The validator tests are already covered in the service tests

Also:

 * Fixed the __all__ in values
 * The previous tests were quite funny if you read them

For ticket: https://github.com/hypothesis/lms/issues/1242

This should be rebased after: https://github.com/hypothesis/lms/pull/1201


jon-betts:
Fun fact, at the time of writing this refactor removes over twice as much code as it adds! It all does the same stuff + extra test coverage, and it's less than 50% of the size.

seanh:
I'll leave this until after https://github.com/hypothesis/lms/pull/1201 is merged, to review. It'll be easier then

jon-betts:
We had some chat which I'll try and summarise:

## Two ideas of validators in one directory
  * I think there are two concepts squished into one:
    * Schema as a layer that guards the views
    * Schema that are used in random places to parse things out of `dict`s
    * The layer ones should probably remain where they are for now (move near the views one day?)
    * The other ones should probably go live with the places they are used

## Re-use of code

 * It's probably fine to have base classes in the validator name space, as it's the right name, even if they are used outside of the guarding context
    * We are likely missing a more generic base-class for generic use
    * The semantics of wrapping `ValidationError` might not make sense out side of a view guard
    * Non-views using schema should catch errors and handle them themselves...
    * ... unless the `ValidationError` concept _does_ make sense and you want things caught by the error views. Go you!

## Recording decisions

 * We don't have great places to record these decisions, but a nice suggestion might be a `README.md` in the validator directory:
   * You're quite likely to spot it (it'll stand out from the code)
   * It's where developers look (not in a random wiki or otherwise hidden away)
   * It can be updated in PRs as decisions change
   * It can detail what should go there, and _what should not_

# How to finish this PR

We've not discussed the exact solution in detail yet, but I'll have a quick stab:

 * Add a README.md to the validators directory explaining the difference
 * Make the base-class externally available for nice defaults (like EXCLUDE)
 * Use that base class in the service

jon-betts:
This has been split up into three different pull requests:
* Removing an unneccessary value object from the LIS service #1268
* Preparing the validators to be used outside of the directory #1269
* Inlining the LIS schema into the service where it is used #1270

PR #1249, opened 2019-12-06
Fixing unbound response bug in error handling in LTI Outcomes service

jon-betts:
We referred to a variable which can be undeclared by the time we get to it.

This fixes: https://github.com/hypothesis/lms/issues/1106


jon-betts:
I don't know. Might be tricky in such a dynamic language. It certainly doesn't right now?

robertknight:
Oh, of course, this would require reasoning about whether the right-hand side of the assignment expression can throw or not, which is going to be very hard to do accurately.

PR #1250, opened 2019-12-06
A small refactor to attach code to the classes they are used in

jon-betts:
This is to clear up some homeless code that was floating about in the files next to classes where they are used.

Adding class methods leaves the code nomadic if required, but also implies stronger scoping that the file level.


jon-betts:
Squished the original fix in the rebase. Thank goodness for tests.

Issue #1128, opened 2019-11-06
Instructors unable to annotate the first time a Hypothesis assignment is launched in Blackboard and Moodle


klemay:
## Overview
The very first time an instructor creates a Hypothesis assignment in a Blackboard course site, they are unable to annotate until they refresh the page or close and re-open the assignment. It’s unclear whether this is specific to our Blackboard instance (does it happen for other Bb institutions?), or whether it’s specific to our app (does the same problem appear for other LTI apps?). 

Note that we are able to reproduce this with and without the grading feature flag, so this is likely not related to the grading portion of our app. 

## Steps to reproduce
1. In our Blackboard instance, [create a new course site](https://help.blackboard.com/Learn/Administrator/Hosting/Course_Management/Creating_Courses) and [enroll yourself as an instructor](https://help.blackboard.com/Learn/Administrator/Hosting/Course_Management/Managing_Enrollments)
2. [Create a Hypothesis-enabled assignment](https://web.hypothes.is/help/creating-hypothesis-enabled-readings-in-blackboard/) (with or without evaluation enabled)
Once you have successfully selected an HTML page or PDF, the assignment will launch

## Expected behavior
The selected document loads with the Hypothesis sidebar on top of it. A group has been created for the course, and as a teacher I can annotate the document. 

## Actual behavior
The group name does not appear as expected. Instead we see “...” in place of the group name. Selecting text and clicking “Annotate” creates a card in the sidebar with no place to type an annotation:

![pasted image 0](https://user-images.githubusercontent.com/12087705/68335190-006ff900-00aa-11ea-9021-634fa7bda8ed.png)

Reloading the browser tab or closing and re-opening the assignment will fix this. Furthermore, you will never run into this problem again in this course site— it seems to only happen with the very first assignment / with group creation.

## Additional details
We have reproduced this with all of the following conditions:
- With and without the grading feature flag enabled
- With a course created through the Course Wizard and through a course created manually
- With HTML pages and with Google Drive PDFs
- With the Hypothesis assignment being the first assignment in a course, and with a non-Hypothesis assignment being the first assignment created in the course.

Again, it’s unclear whether this is unique to our Blackboard instance (as it’s not connected to an SIS system and not a true representation of a working Bb environment like what you’d see at a school). 

In the meantime, we are updating our docs at https://web.hypothes.is/help/creating-hypothesis-enabled-readings-in-blackboard/ to suggest that teachers refresh the page or relaunch the assignment as the final step of assignment creation.

[...]

PR #1252, opened 2019-12-06
Fix no grading on teacher's first launch of an assignment

jon-betts:
Currently this is trying to fix the issue where grading doesn't launch, but it looks like there might be others too.

This is for: https://github.com/hypothesis/lms/issues/1128

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -221,30 +221,14 @@ class BasicLTILaunchViews:
         we'll save it in our DB. Subsequent launches of the same assignment
         will then be DB-configured launches rather than unconfigured.
         """
-        oauth_consumer_key = self.request.lti_user.oauth_consumer_key
 
-        lti_params = {
-            # TODO - This isn't really an LTI param
-            "authorization": BearerTokenSchema(
-                self.request).authorization_param(
-                self.request.lti_user
-            ),
-            # TODO - is this just read from a param in the first place?
-            "oauth_consumer_key": oauth_consumer_key,
-            "user_id": self.request.lti_user.user_id,
-        }
 
-        # Directly persist certain fields
-        for param in [
-                    "context_title",
-                    "resource_link_title",
-                    "lis_outcome_service_url",
-                    "resource_link_id",
-                    "tool_consumer_instance_guid",
-                    "context_id",
-                ]:
-            if param in self.request.params:
-                lti_params[param] = self.request.params[param]
+        params = self._extract_lti_params(self.request)
+
+        # Copy the Authorization header as a parameter
+        params["authorization"] = BearerTokenSchema(self.request).authorization_param(
+            self.request.lti_user
+        )
 
         # Add the config needed by the JavaScript document selection code.
         self.context.js_config.update(
@@ -255,7 +239,7 @@ class BasicLTILaunchViews:
                 # (currently only Canvas supports this).
                 "enableLmsFilePicker": False,
                 "formAction": self.request.route_url("module_item_configurations"),
-                "formFields": lti_params,
+                "formFields": params,
                 "googleClientId": self.request.registry.settings["google_client_id"],
                 "googleDeveloperKey": self.request.registry.settings[
                     "google_developer_key"
@@ -266,6 +250,19 @@ class BasicLTILaunchViews:
         )
 
         return {}
+
+    @classmethod
+    def _extract_lti_params(cls, request):
+        """Copy all of the LTI params from a request minus OAuth 1 params."""
+
+        # Exclude OAuth 1 variable signing fields just so the client-side
+        # doesn't see these and think they are already done.
+
+        return {
+            param: value
+            for param, value in request.params.items()
+            if param not in ["oauth_nonce", "oauth_timestamp", "oauth_signature"]
+        }
 
     # pylint:disable=no-self-use
     @view_config(
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -222,7 +222,6 @@ class BasicLTILaunchViews:
         will then be DB-configured launches rather than unconfigured.
         """
 
-
         params = self._extract_lti_params(self.request)
 
         # Copy the Authorization header as a parameter
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -254,8 +254,9 @@ class BasicLTILaunchViews:
     def _extract_lti_params(cls, request):
         """Copy all of the LTI params from a request minus OAuth 1 params."""
 
-        # Exclude OAuth 1 variable signing fields just so the client-side
-        # doesn't see these and think they are already done.
+        # Exclude OAuth 1 variable signing fields as they should not be
+        # re-used. If this request needs to be signed, it needs to be signed
+        # again.
 
         return {
             param: value
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -223,6 +223,29 @@ class BasicLTILaunchViews:
         """
         oauth_consumer_key = self.request.lti_user.oauth_consumer_key
 
+        lti_params = {
+            # TODO - This isn't really an LTI param
+            "authorization": BearerTokenSchema(
+                self.request).authorization_param(
+                self.request.lti_user
+            ),
+            # TODO - is this just read from a param in the first place?
+            "oauth_consumer_key": oauth_consumer_key,
+            "user_id": self.request.lti_user.user_id,
+        }
+
+        # Directly persist certain fields
+        for param in [
+                    "context_title",
+                    "resource_link_title",
+                    "lis_outcome_service_url",
+                    "resource_link_id",
+                    "tool_consumer_instance_guid",
+                    "context_id",
+                ]:
+            if param in self.request.params:
+                lti_params[param] = self.request.params[param]
+
         # Add the config needed by the JavaScript document selection code.
         self.context.js_config.update(
             {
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -278,45 +279,52 @@ def pyramid_request(self, pyramid_request):
...
+    @pytest.fixture
+    def bearer_token_schema(self, BearerTokenSchema):
+        return BearerTokenSchema.return_value
+
     @pytest.fixture
     def pyramid_request(self, context, pyramid_request):
         pyramid_request.params = {
+            "user_id": "TEST_USER_ID",
             "resource_link_id": "TEST_RESOURCE_LINK_ID",
+            "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
```
jon-betts: This is interesting. Our tests here didn't really setup a request properly. These parameters were not passed because we were plucking them out of the request object `LTIUser` etc.

jon-betts:
I think I've got this right?

Basically we now pass through *everything* we get bar some signed OAuth1 fields. This means we should translate the exact launch request as the configured one we got when un-configured.

I suspect this will fix a few other bugs we haven't found yet as well.

jon-betts:
This needs some tests around the fact that module item configuration now calls the sync to h stuff.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -289,47 +290,96 @@ def pyramid_request(self, pyramid_request):
         return pyramid_request
 
 
+class TestConfigureModuleItem(ConfiguredLaunch):
+    def make_request(self, context, pyramid_request):
+        BasicLTILaunchViews(context, pyramid_request).configure_module_item()
```
jon-betts: The real diff here is the addition of the `ConfiguredLaunch` base class. The rest is a result of moving it to be above the un-configured launches.

I've also used the `make_request()` function as in other classes to highlight the similarity between them.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -289,47 +290,96 @@ def pyramid_request(self, pyramid_request):
...
         pyramid_request.registry.settings[
             "google_developer_key"
         ] = "TEST_GOOGLE_DEVELOPER_KEY"
 
         BasicLTILaunchViews(context, pyramid_request).unconfigured_basic_lti_launch()
 
-        BearerTokenSchema.assert_called_once_with(pyramid_request)
-        bearer_token_schema.authorization_param.assert_called_once_with(
-            pyramid_request.lti_user
-        )
```
jon-betts: We don't care about the internal implementation of this. It output is correct, we don't care how it happened.

White box testing like this can increase the cost refactoring later.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -289,47 +290,96 @@ def pyramid_request(self, pyramid_request):
...
-            },
+            "formFields": Any.dict(),
             "googleClientId": "TEST_GOOGLE_CLIENT_ID",
             "googleDeveloperKey": "TEST_GOOGLE_DEVELOPER_KEY",
             "customCanvasApiDomain": context.custom_canvas_api_domain,
             "lmsUrl": context.lms_url,
             "urls": {},
         }
 
+        # Test param pass through
```
jon-betts: I've pulled off the params to make it easier to see what's failing. This way we get separate assertion failures for the root elements and the params.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -289,47 +290,96 @@ def pyramid_request(self, pyramid_request):
...
+                "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
+                "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
+                "user_id": "TEST_USER_ID",
+                "context_id": "TEST_CONTEXT_ID",
+                "some_random_rubbish": "SOME_RANDOM_RUBBISH",
+            }
+        )
+
+    @pytest.fixture
+    def bearer_token_schema(self, BearerTokenSchema):
```
jon-betts: Private to this test method

tests/unit/lms/views/basic_lti_launch_test.py
jon-betts: Righto, I'll take another look at this. I still suspect this is a mocking issue as we don't seem to have the ability to add input that can generate a reliable output.

lms/views/basic_lti_launch.py
```diff
@@ -255,6 +250,19 @@ def unconfigured_basic_lti_launch(self):
...
+    def _extract_lti_params(cls, request):
+        """Copy all of the LTI params from a request minus OAuth 1 params."""
+
+        # Exclude OAuth 1 variable signing fields just so the client-side
+        # doesn't see these and think they are already done.
+
+        return {
+            param: value
+            for param, value in request.params.items()
+            if param not in ["oauth_nonce", "oauth_timestamp", "oauth_signature"]
```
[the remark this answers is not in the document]
jon-betts: All the other things are invariant over signing and should be kept. These are the only things that change as a result of signing.

tests/unit/lms/views/basic_lti_launch_test.py
jon-betts: My mistake here was assuming that the test as written actually check outputs that would matter if you broke the code. We don't really check it because of heavy mocking.

To be clear I wasn't attempting to fix this problem by just removing tests, I assumed (incorrectly) that the assertions here were unnecessary because the final assertion about the output would actually test the class.

So I think what I've said is it's `unconfigued_basic_lti_launch()`'s job to put a valid JWT token in the output. Whether it does this by calling `BearerTokenSchema` or not is an implementation detail. We provide inputs, run the code and check the outputs.

This also has the benefit of being able to read the tests and actually see what the end-point is supposed to do. I don't really believe in the unit-tests as documentation, but it's a nice direction to head for.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -156,11 +158,40 @@ def test_it_does_not_call_list_result_upsert_if_canvas(
...
+        pyramid_config.register_service(lti_h_service, name="lti_h")
+        return lti_h_service
+
+    @pytest.fixture
+    def LtiLaunches(self, patch):
+        return patch("lms.views.basic_lti_launch.LtiLaunches")
+
+    @pytest.fixture(autouse=True)
+    def ModuleItemConfiguration(self, patch):
+        return patch("lms.views.basic_lti_launch.ModuleItemConfiguration")
```
jon-betts: All these fellas were hanging out at the bottom of the file, but they only apply to these tests.

tests/unit/lms/views/basic_lti_launch_test.py
jon-betts: I've moved this and the others down. I didn't do this in the parent class as it's not really running tests, and you need to know front and center that you are expected to implement that function.

tests/unit/lms/views/basic_lti_launch_test.py
jon-betts: Actually not required at all

lms/views/basic_lti_launch.py
```diff
@@ -255,6 +250,19 @@ def unconfigured_basic_lti_launch(self):
 
         return {}
 
+    @classmethod
+    def _extract_lti_params(cls, request):
+        """Copy all of the LTI params from a request minus OAuth 1 params."""
+
+        # Exclude OAuth 1 variable signing fields just so the client-side
+        # doesn't see these and think they are already done.
```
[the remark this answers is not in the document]
jon-betts: I've updated this with something more informative. Basically we shouldn't be reusing signed stuff, even if it works.

tests/unit/lms/views/basic_lti_launch_test.py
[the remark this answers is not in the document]
jon-betts: I've updated it to be more informative.

tests/unit/lms/views/basic_lti_launch_test.py
seanh: :+1:

tests/unit/lms/views/basic_lti_launch_test.py
seanh: My thoughts on isolated testing (note: "isolated" not mocked, mocks are just one possible end to the means that is isolation) here: https://seanh.cc/post/when-to-use-mocks/#isolated-tests-vs-integrated-tests. Or at least those are the thoughts from 2017 that drove the approach to unit testing that we currently have. I wouldn't say I believe much in tests as documentation either, but I do believe in unit tests as drivers of better code design (at least if you pay attention to what your unit tests are telling you, which isn't easy), as a practical way to get high coverage, to help with refactoring and change management, etc.

But whether and how that applies to views in particular I think is open to discussion.

I think what you're talking about here is integrated tests. Not necessarily full end-to-end tests (or what we call "functional" tests in Hypothesis) but tests where at least the view is tested in integration with `BearerTokenSchema` (in this case) rather than tested strictly in isolation from all its first-party dependencies (all first-party dependencies mocked out).

The current pattern for `tests/unit/` tests in Hypothesis apps is that those are isolated unit tests not integrated tests.

We did talk about adding a `tests/integrated/` and views might be a good kind of thing to add such integrated tests for.

But we also said that only `tests/unit/` would count towards coverage so you would still require isolated unit tests for the views as well to get 100% coverage.

Unless of course we come up with a different idea...

lms/views/basic_lti_launch.py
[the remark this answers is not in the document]
seanh: :+1:

lms/views/basic_lti_launch.py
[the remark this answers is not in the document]
seanh: :+1:

PR #1262, opened 2019-12-11
Proof of concept for Moodle scale (and other) handling

jon-betts:
This is proof of concept work for the Moodle grading spike: https://github.com/hypothesis/lms/issues/1265

## How does this work?

 * We have a polymorphic tool_consumer service
 * This is swapped out to be the tool specific one on a per request basis
 * This has some generic things like "want grading?"
 * This is then free to have tool specific logic to decide
 * We then return a GradingType object to the client based on one of these
 * This includes things like:
   * Type: loat, enum
   * Min and max values
   * Exact values in the case of enum (used for Moodle scales)
 * The interface currently shows the type only to prove it got there

## Caveats:

 * This doesn't make the UI do anything useful with it
 * This doesn't account for how we will translate scores back and forward
 * No tests, comments, docstrings etc
 * Debugging code everywhere
 * Pulls in an env TOKEN which I don't know how to handle

### DIFF
--- a/lms/logic/grading/grading_types.py
+++ b/lms/logic/grading/grading_types.py
@@ -3,23 +3,20 @@ from typing import NamedTuple
 
 class FloatGrading(NamedTuple):
     min: float = 0
-    max: float = 1
-    min_increment: float = 0
-    type = "float"
+    max: float = 10
 
+    def as_dict(self):
+        data = self._asdict()
+        data['type'] = 'float'
 
-class IntegerGrading(NamedTuple):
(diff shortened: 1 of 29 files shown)


jon-betts:
This ticket is closed because we aren't going to use this code, but...

# :warning: **DON'T DELETE THE BRANCH !** :warning:

We will pick bits out of this code in the future.

jon-betts:
I re-opened this as something/one (me?) closed the branch here by mistake

PR #1269, opened 2019-12-13
Preparing the validators to be used outside of the directory

jon-betts:
* Making the baseclass public
 * Making the exceptions public
 * Adding a README.md to explain the rules of the road
 * Fixing a million imports

Split off from: https://github.com/hypothesis/lms/pull/1247


seanh:
I hope you don't mind but I took the liberty of rewriting `validation/README.md`. Nothing about the code design decision has changed. Just changed how it's documented:

* Add a very short summary under the top-level title:
  "Guard schemas for views."

* Also added a summary list of the three different types of things that
  this package contains, at the top of the "What is in this package?"
  section. Just making it even more explicit what this section of the
  README is doing (enumerating the types of things the package
  contains).  For the same reason added numbers to the subheadings

* Separated "what this package contains" into three things rather than
  two: 1. Guard schemas; 2. Exceptions raised by guard schemas; 3. Base
  classes and other helpers for schemas

* Don't define schemas as things that "define a set of parameters which
  are required to call a particular view". Although schemas can contain
  required parameters, they can also contain only optional parameters,
  and they can do much more than defining the required parameters: they
  actually parse, deserialize and validate the parameters and the
  request as whole. The parsed params (`request.parsed_params`) might
  actually be quite different from the raw `request.params` if parsing
  and deserialization has happened. There are many ways that a request
  might have all the required parameters but still fail validation. I've
  defined this instead as "parse and validate Pyramid requests".

* Don't document how to use guard schemas in `@view_config`'s in this
  file. That's already documented in the `__init__.py`'s docstring, so
  just link to that instead. Let's keep the `README.md` for documenting
  code architecture / design from the point of view of someone who's
  thinking of changing the code in `lms/validation/`, and use docstrings
  for documenting the code's public interface from the user's point of
  view.

* When talking about "What shouldn't be in this package" talk about more
  than schemas that are "just used to pick things out of dicts". That is
  one type of schema but non-guard schemas can parse requests more
  generally, which is much more than picking things from dicts: parsing
  query params, form data, JSON, possibly combinations thereof, raising
  `ValidationError`'s, doing some deserialization/parsing/processing of
  params rather than just picking them, etc etc.

  Also added mention of response schemas.

  Also expanded on the reasoning for why these schemas should not live
  in this package, but may nonetheless use helpers from this package

* Be consistent about calling them "validation schemas" or just
  "schemas", not "validators". This isn't because "validation schemas"
  is somehow better than "validators". It's just because being
  consistent is clearer.

* Say "views" not "view classes". Not all views go in classes.

* Fix references to `lms.validators`: the package is called
  `lms.validation`

* There's no such thing as "error views" in Pyramid, they're called
  "exception views"

* Refer to "this package" rather than "this directory", since it's
  actually a Python package

* Various minor typos, e.g. change "view directory" (in reference to an
  `@view_config(...)`) to "view configuration"

seanh:
Rebased

lms/validation/__init__.py
```diff
@@ -64,6 +69,9 @@ def foo_view(request):
     "APIRecordSpeedgraderSchema",
     "APIReadResultSchema",
     "APIRecordResultSchema",
+    "PlainSchema",
+    "PyramidRequestSchema",
+    "RequestsResponseSchema",
```
seanh: `lms/validation/__init__.py`'s `__all__` explicitly lists the names that make up `lms.validation`'s public interface. So here, I'm explicitly adding the base schemas to the public interface.

PR #1270, opened 2019-12-13
Inlining the LIS schema into the service where it is used

jon-betts:
This is a part of: https://github.com/hypothesis/lms/pull/1247 in an attempt to make it easier to review.

This should be re-based etc after:

 * https://github.com/hypothesis/lms/pull/1269
 * https://github.com/hypothesis/lms/pull/1268


seanh:
I've pushed a commit to change this to use `PyramidRequestSchema` rather than `PlainSchema`:

```diff
-from marshmallow import ValidationError, fields
+from marshmallow import fields
 
 from lms.models import LISResultSourcedId
-from lms.validation import PlainSchema
+from lms.validation import PyramidRequestSchema, ValidationError
 
 __all__ = ["LISResultSourcedIdService"]
 
@@ -10,7 +10,8 @@ class LISResultSourcedIdService:
     """Methods for interacting with LISResultSourcedId records."""
 
     # The relevant parameters from a request
-    class _ParamsSchema(PlainSchema):
+    class _ParamsSchema(PyramidRequestSchema):
+        locations = ["form"]
         lis_result_sourcedid = fields.Str(required=True)
         lis_outcome_service_url = fields.Str(required=True)
         context_id = fields.Str(required=True)
@@ -66,8 +67,7 @@ class LISResultSourcedIdService:
         :type lti_user: :class:`lms.values.LTIUser`
         """
         try:
-            params = self._ParamsSchema().load(request.POST)
-
+            params = self._ParamsSchema(request).parse()
         except ValidationError:
             # We're missing something we need in the request.
             # This can happen if the user is not a student, or if the needed
```

While this doesn't have _much_ benefit in this particular case I don't think it has any downsides either (`_ParamsSchema` is no worse for using  `PyramidRequestSchema` rather than `PlainSchema`) and it does have some minor benefits:

* `PyramidRequestSchema` already knows how to parse Pyramid/WebOb request objects (via `webargs.PyramidParser` which `PyramidRequestSchema` is based on) so the code that calls `_ParamsSchema` no longer needs to. Granted, there is not much knowledge to know here. But this

  ```python
  params = self._ParamsSchema().load(request.POST)
  ```

  is coupled to WebOb in a small way by knowing that `request.POST` is what contains the params in the case of a form post. It's also kind of difficult to read, since WebOb's `request.POST` contains the form field params only and only in the case of an HTTP POST form submission, but it's not obvious from the name `POST` that this really does mean form fields only and not a JSON body or something else. Finally, it's duplicating this knowledge about `request.POST` which `PyramidRequestSchema` already has.

  I think this:

  ```python
  params = self._ParamsSchema(request).parse()
  ```

  is a little nicer and also allows all request-parsing validators to have a consistent interface.

* It allows the schema itself to declare what part(s) of the request it reads from rather than the caller having to do so, and in a way that's more readable (actually says "form") and more decoupled from WebOb:

  ```python
  class _ParamsSchema(PyramidRequestSchema):
      locations = ["form"]
      ...
  ```

* I think what we're really doing here is parsing a request, not parsing a dict.

  There are two steps to parsing a request:

  1. Decide what part(s) of the request to read from (`request.POST`)
  2. Parse and validate the params from that part of the request.

  These both come under "validation" in my mind which means we're actually validating `request` not just `request.POST`.

  If you agree with this definition then `params = self._ParamsSchema().load(request.POST)` actually splits the "validation" 50/50 between `_ParamsSchema` itself and the code that passes `request.POST` to it.

  I think it might be a tiny bit nicer to have all the "validation" done inside the validation schema. And `PyramidRequestSchema` is after all the "base class for schemas that validate Pyramid requests".

* We might run into problems because `request.POST` is not actually a dict but a `NestedMultiDict` or, in the case of non-form posts, a `NoVars` object? I think we might have been fine anyway but this pushes the responsibility for knowing these sorts of things about WebOb onto `webargs` :man_shrugging:

PR #1273, opened 2019-12-16
Trying to make the Jenkins build fail when the gherkin tests fail

jon-betts:
Is this all that is required? This is for https://github.com/hypothesis/lms/issues/1220


jon-betts:
Looks like this is working well now!

PR #1276, opened 2019-12-16
Don't list all students if grading is disabled

jon-betts:
* Previously if grading is disabled, we still run the whole block
 * This means we look up all students for every call
 * This will be generating lots of unnecessary DB calls

This has two commits:

 * The first is a straight indent of the code to show the tests passing
 * The second is a refactor for an early return and to create the config separately from lookup

### DIFF
--- a/lms/views/helpers/frontend_app.py
+++ b/lms/views/helpers/frontend_app.py
@@ -41,7 +41,7 @@ def configure_grading(request, js_config):
     }
 
 
-def _find_students_for_assignment(context_id, resource_link_id, request):
+def _students_for_assignment(context_id, resource_link_id, request):
     service = request.find_service(name="lis_result_sourcedid")
 
     for student in service.fetch_students_by_assignment(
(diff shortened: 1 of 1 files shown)


lms/views/helpers/frontend_app.py
```diff
@@ -12,42 +12,48 @@ def configure_grading(request, js_config):
...
                 "LISResultSourcedId": student.lis_result_sourcedid,
                 "LISOutcomeServiceUrl": student.lis_outcome_service_url,
             }
-        )
+            for student, h_user in _find_students_for_assignment(
+                context_id=request.params.get("context_id"),
+                resource_link_id=request.params.get("resource_link_id"),
+                request=request,
+            )
+        ],
```
seanh: :+1: I like the use of a list comprehension in a dict literal to simplify the code here

lms/views/helpers/frontend_app.py
```diff
@@ -12,42 +12,48 @@ def configure_grading(request, js_config):
...
+    }
+
+
+def _find_students_for_assignment(context_id, resource_link_id, request):
```
seanh:
```suggestion
def _students_for_assignment(context_id, resource_link_id, request):
```

jon-betts:
This looks good in Moodle and Canvas. Speedgrader and our grading our still working fine.

Issue #1243, opened 2019-12-05
BasicLTILaunchViews __init__ lis code should move to the lis service [DEAD]


jon-betts:
Currently there is some pretty inscruitable code in the __init__ method of BasicLTILaunchViews which is accessing LIS related params. I believe it's trying to replicate the validation of the LIS params which is now done in the LIS service.

As we have this code in one place I think we can replace this with a 'has_lis_blah' type function instead of just having un-documented param checks in an if statement.

jon-betts:
Work for this starting in: https://github.com/hypothesis/lms/pull/1277

jon-betts:
I think this probably is a bit premature for a few reasons:

 * There are related 'buddy' values that are always moving around together
 * This makes me think we are missing some abstractions over them
 * I've seen very similar code to this elsewhere
 * I don't want to make things worse because I don't understand

There are two halves to the PR that was made:

 * Some good and safe things about naming canvas specific things canvas
 * Some more dodgy behavior changing stuff related to this

I'm therefore going to separate the PR in two, keep the good bit and then delete this ticket.

PR #1277, opened 2019-12-16
Concentrating logic about LIS parameters in the LIS service [DEAD]

jon-betts:
* Removed a few places which were interpretting these themselves
 * This also means it's interpretted the same way in all places
 * Doesn't have any new test coverage yet

For https://github.com/hypothesis/lms/issues/1243

### DIFF
--- a/lms/services/lis_result_sourcedid.py
+++ b/lms/services/lis_result_sourcedid.py
@@ -54,6 +54,32 @@ class LISResultSourcedIdService:
             .all()
         )
 
+    def is_assignment_gradable(self, request):
+        """
+        Does this request have the parameters required to enable grading?
+
+        :param request: A pyramid request
+        :rtype: bool
+        """
+
+        return self.get_grading_parameters(request) is not None
+
+    def get_grading_parameters(self, request):
+        """
+        Get the parameters for grading, if present.
+
+        :param request: A pyramid request
+        :return: A dict of parsed parameters for grading or None
+        """
+
+        try:
+            return self._ParamsSchema(request).parse()
+        except ValidationError:
+            # We're missing something we need in the request.
+            # This can happen if the user is not a student, or if the needed
+            # LIS data is not present on the request.
+            return
+
     def upsert_from_request(self, request, h_user, lti_user):
         """
         Update or create a record based on the LTI params found in the request.
@@ -67,12 +93,9 @@ class LISResultSourcedIdService:
         :arg lti_user: The LTI-provided user that this record is associated with
         :type lti_user: :class:`lms.values.LTIUser`
         """
-        try:
-            parsed_params = self._ParamsSchema(request).parse()
-        except ValidationError:
-            # We're missing something we need in the request.
-            # This can happen if the user is not a student, or if the needed
-            # LIS data is not present on the request.
+
+        parsed_params = self.get_grading_parameters(request)
+        if parsed_params is None:
             return
 
         lis_result_sourcedid = self._find_or_create(
--- a/lms/services/lis_result_sourcedid.py
+++ b/lms/services/lis_result_sourcedid.py
@@ -56,7 +56,7 @@ class LISResultSourcedIdService:
 
     def is_assignment_gradable(self, request):
         """
-        Does this request have the parameters required to enable grading?
+        Determine if the request has the parameters required to enable grading.
 
         :param request: A pyramid request
         :rtype: bool
@@ -74,11 +74,12 @@ class LISResultSourcedIdService:
 
         try:
             return self._ParamsSchema(request).parse()
+
         except ValidationError:
             # We're missing something we need in the request.
             # This can happen if the user is not a student, or if the needed
             # LIS data is not present on the request.
-            return
+            return None
 
     def upsert_from_request(self, request, h_user, lti_user):
         """
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -36,6 +36,7 @@ class BasicLTILaunchViews:
     def __init__(self, context, request):
         self.context = context
         self.request = request
+        self.lis_result_service = request.find_service(name="lis_result_sourcedid")
 
         self.context.js_config.update(
             {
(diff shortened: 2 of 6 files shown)


jon-betts:
This is failing on coverage as expected.

lms/views/basic_lti_launch.py
```diff
@@ -36,36 +36,41 @@ class BasicLTILaunchViews:
...
-            )
+        self._add_canvas_submission_params()
 
         # If the launch has been configured to focus on the annotations from
         # a particular user, translate that into Hypothesis client configuration.
         focused_user = self.request.params.get("focused_user")
         if focused_user:
             self._set_focused_user(focused_user)
 
+    def _add_canvas_submission_params(self):
```
seanh: Glad we've got the word "Canvas" in here

seanh:
```suggestion
    def add_canvas_submission_params(self):
```

I don't think these need to be `_`'d since views classes don't really have public interfaces that need to be expressed (we don't write code that imports and calls view classes). Some of the other existing methods in this class don't have a leading `_`: `sync_lti_data_to_h()`, `store_lti_data()`. Unfortunately some of the other methods _do_ have `_`'s though so it's inconsistent.

We _should_ move helper methods like this to the bottom of the class, beneath the views, though. (Again some of the existing methods are grouped at the bottom of the class but some aren't.)

lms/views/basic_lti_launch.py
```diff
@@ -36,36 +36,41 @@ class BasicLTILaunchViews:
...
+        students (not teachers) launch an assignment.
+        """
+
+        if not self._is_launched_by_canvas():
+            return
+
+        lis_params = self.lis_result_service.get_grading_parameters(self.request)
+
+        if not lis_params:
+            return
```
seanh: I guess this will now return if `context_id` and/or `resource_link_id` is missing (whereas previously the `js_config` below would have been set as long as `lis_result_sourcedid` and `lis_outcome_service_url` were present). In practice I don't think this makes a difference -- `context_id` and `resource_link_id` _are_ present in Canvas launches.

lms/services/lis_result_sourcedid.py
```diff
@@ -54,6 +54,33 @@ def fetch_students_by_assignment(
...
+    def get_grading_parameters(self, request):
+        """
+        Get the parameters for grading, if present.
+
+        :param request: A pyramid request
+        :return: A dict of parsed parameters for grading or None
+        """
+
+        try:
+            return self._ParamsSchema(request).parse()
```
seanh: May want to rename `_ParamsSchema` to `_GradingParamsSchema`

lms/views/helpers/frontend_app.py
```diff
@@ -12,9 +12,12 @@ def configure_grading(request, js_config):
     Note that this is entirely distinct from Canvas Speedgrader, which provides
     its own UI.
     """
+
+    lis_result_sourcedid_svc = request.find_service(name="lis_result_sourcedid")
+
     if (
         request.lti_user.is_instructor
-        and _is_assignment_gradable(request)
+        and lis_result_sourcedid_svc.is_assignment_gradable(request)
```
seanh: Again there's a functional change here -- previously only `lis_outcome_service_url` is required but now an assignment will be considered ungradeable if any of `lis_outcome_service_url`, `lis_result_sourcedid`, `context_id` or `resource_link_id` are missing. Again, AFAIK it doesn't make a difference -- at least in the cases of Blackboard, Moodle and Brightspace they _do_ send all these params.

lms/services/lis_result_sourcedid.py
```diff
@@ -54,6 +54,33 @@ def fetch_students_by_assignment(
...
 
+    def is_assignment_gradable(self, request):
+        """
+        Determine if the request has the parameters required to enable grading.
+
+        :param request: A pyramid request
+        :rtype: bool
+        """
+
+        return self.get_grading_parameters(request) is not None
```
seanh: You could take this one step further and remove the `request` argument from `is_assignment_gradeable()` and `get_grading_parameters()`, since service factories receive the request from pyramid_services. The two methods could then become `@property`'s which would simplify the integration code (the views). Instead of:

```python
if lis_result_sourcedid_svc.is_assignment_gradable(self.request):
    ...
```

it would be just:

```python
if lis_result_sourcedid_svc.is_assignment_gradable:
    ...
```

It would also simplify the integration tests (view tests) as there's no need to test `lis_result_sourcedid_svc.is_assignment_gradable.assert_called_once_with(pyramid_request)` since it's no longer the view's responsibility to correctly pass the request in. Tests would just set `lis_result_sourcedid_svc.is_assignment_gradable` to `True` and to `False` (instead of setting `lis_result_sourcedid_svc.is_assignment_gradable.return_value` as they should currently) and assert that the views do the right thing for the `True` and `False` cases.

lms/views/basic_lti_launch.py
```diff
@@ -36,36 +36,41 @@ class BasicLTILaunchViews:
...
-            self._set_submission_param(
-                "lis_outcome_service_url", params.get("lis_outcome_service_url")
-            )
+        self._add_canvas_submission_params()
 
         # If the launch has been configured to focus on the annotations from
         # a particular user, translate that into Hypothesis client configuration.
         focused_user = self.request.params.get("focused_user")
         if focused_user:
             self._set_focused_user(focused_user)
```
jon-betts: This is canvas too, and this is a good opportunity to say so

jon-betts:
This is a copy paste comment from the ticket, but it covers it well, so I thought I'd put it here too:

I think this probably is a bit premature for a few reasons:

 * There are related 'buddy' values that are always moving around together
 * This makes me think we are missing some abstractions over them
 * I've seen very similar code to this elsewhere
 * I don't want to make things worse because I don't understand

There are two halves to the PR that was made:

 * Some good and safe things about naming canvas specific things canvas
 * Some more dodgy behavior changing stuff related to this

I'm therefore going to separate the PR in two, keep the good bit and then delete this ticket.

jon-betts:
Moved the bits we like out to:

 * https://github.com/hypothesis/lms/pull/1277

PR #1281, opened 2019-12-17
Marking Canvas LTI Launch View specific behavior and moving into sub-functions

jon-betts:
Moving some functions in the view `__init__` which are canvas specific into smaller sub-functions . 

This is a bit easier to read, but mostly it's to highlight the fact that they aren't general behavior that matters in anything other than a Canvas context. If that changes, we can change them.

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -319,7 +319,7 @@ class BasicLTILaunchViews:
         )
 
     def initialise_canvas_submission_params(self):
-        """ Add config used by UI to call Canvas `record_submission` API."""
+        """Add config used by UI to call Canvas `record_submission` API."""
         lis_result_sourcedid = self.request.params.get("lis_result_sourcedid")
         lis_outcome_service_url = self.request.params.get("lis_outcome_service_url")
 
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -52,31 +52,54 @@ class BasicLTILaunchViews:
             }
         )
 
-        # Add config used by frontend to call `record_submission` API.
-        params = self.request.params
-        if (
-            # The outcome reporting params are typically only available when
-            # students (not teachers) launch an assignment.
-            params.get("lis_result_sourcedid")
-            and params.get("lis_outcome_service_url")
-            # This feature is initially restricted to Canvas.
-            and self._is_launched_by_canvas()
-        ):
-            self.context.js_config["submissionParams"] = {}
+        if self.is_launched_by_canvas():
+            self.add_canvas_submission_params()
+            self.set_canvas_focused_user()
 
-            self._set_submission_param("h_username", context.h_user.username)
-            self._set_submission_param(
-                "lis_result_sourcedid", params.get("lis_result_sourcedid")
-            )
-            self._set_submission_param(
-                "lis_outcome_service_url", params.get("lis_outcome_service_url")
-            )
+    def add_canvas_submission_params(self):
+        """
+        Add config used by frontend to call Canvas `record_submission` API.
+
+        The outcome reporting params are typically only available when
+        students (not teachers) launch an assignment.
+        """
+
+        lis_result_sourcedid = self.request.params.get("lis_result_sourcedid")
+        lis_outcome_service_url = self.request.params.get("lis_outcome_service_url")
+
+        if lis_result_sourcedid and lis_outcome_service_url:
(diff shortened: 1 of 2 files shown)


lms/views/basic_lti_launch.py
```diff
@@ -335,27 +310,64 @@ def _set_via_url(self, document_url):
         self.context.js_config["urls"].update(
             {"via_url": via_url(self.request, document_url)}
         )
-        self._set_submission_param("document_url", document_url)
+        self.set_canvas_submission_param("document_url", document_url)
+
+    # ---------------------------------------------------------------------- #
+    # Canvas specific functions
```
seanh: I'd vote we delete these kind of comments -- it's clear from the `canvas` in the names of all the methods below that they're Canvas-specific, and a comment like this is just begging for something not Canvas-specific to be added beneath it and unwittingly turn the comment into a lie.

lms/views/basic_lti_launch.py
```diff
@@ -335,27 +310,64 @@ def _set_via_url(self, document_url):
...
+            self.request.params.get("tool_consumer_info_product_family_code")
+            == "canvas"
+        )
+
+    def initialise_canvas_submission_params(self):
+        """
+        Add config used by frontend to call Canvas `record_submission` API.
+
+        The outcome reporting params are typically only available when
+        students (not teachers) launch an assignment.
```
seanh: This seems like a code comment rather than a docstring--it doesn't seem to be describing what the method does but rather a detail of how it works. Making it a code comment would also allow it to be placed near the `lis_result_sourcedid` and `lis_outcome_service_url` params that it refers to, currently it might not be clear what "outcome reporting params" refers to. It'd be a bit more verbose but I think more clear if the code were written like this:

```python
    def initialise_canvas_submission_params(self):
        """Add config used by frontend to call Canvas `record_submission` API."""

        lis_result_sourcedid = self.request.params.get("lis_result_sourcedid")
        lis_outcome_service_url = self.request.params.get("lis_outcome_service_url")

        # Don't try to submit assignments to Canvas when the assignment is
        # launched by a teacher, or when the submission-related params are
        # missing for some other reason.
        if (not lis_result_sourcedid) or (not lis_outcome_service_url):
            return

        self.context.js_config["submissionParams"] = {
            "h_username": self.context.h_user.username,
            "lis_result_sourcedid": lis_result_sourcedid,
        }
```

lms/views/basic_lti_launch.py
```diff
@@ -335,27 +310,64 @@ def _set_via_url(self, document_url):
...
             self.context.js_config["submissionParams"][name] = value
 
-    def _set_focused_user(self, username):
+    def set_canvas_focused_user(self):
         """Configure the Hypothesis client to focus on a particular user."""
 
+        # If the launch has been configured to focus on the annotations from
+        # a particular user, translate that into Hypothesis client configuration.
+
+        # This parameter is only passed as a part of Canvas Speedgrader config
```
seanh:
```suggestion
        # This parameter is only passed as a part of Canvas SpeedGrader config
```

lms/views/basic_lti_launch.py
```diff
@@ -335,27 +310,64 @@ def _set_via_url(self, document_url):
...
         try:
-            display_name = h_api.get_user(username).display_name
+            display_name = h_api.get_user(focused_user).display_name
         except HAPIError:
             # If we couldn't fetch the student's name for any reason, fall back
             # to a placeholder rather than giving up entirely, since the rest
             # of the experience can still work.
             display_name = "(Couldn't fetch student name)"
 
+        # TODO! - Could/should this be replaced with a LISSourcedId lookup?
```
seanh:
```suggestion
        # TODO! - Could/should this be replaced with a LISResultSourcedId lookup?
```

I think the problem, unfortunately, is that we don't save `LISResultSourcedId`'s to the DB for Canvas. We could start doing so but we'd have to maintain the API lookup as a fallback for preexisting assignments. Other than that it feels like we probably should do it

lms/views/basic_lti_launch.py
jon-betts: I guess I was trying to prevent that from happening by putting a big marquee, but I'll kick it out.

jon-betts: Gone

lms/views/basic_lti_launch.py
jon-betts: It was taken from a comment before. I've used your comment, but I don't think the early return is required for a single statement.

PR #1283, opened 2019-12-17
Fix floating point precision with grade

LMS007:
I noticed this bug when submitted a grade of 3.3. When you multiple 0.33 by 10 in js, you end up with a floating point precision problem that looks like this 3.3000000000000003. This fix converts all grades to a string, checks the least significate digits and rounds to that precision. 

To replicate this bug, just submit a grade of 3.3, and refresh and you'll see 3.3000000000000003


LMS007:
My one concern here is this new scaleGrade method feels like a utility method, but I hesitate to create a single util new file to drop this into. One idea would be to put it into grader-service.js since its feels like it relates to those fetch/get utils.

LMS007:
Moved to the validation util, felt like it could be stuck in there since those functions relate to grading formatting.

PR #1284, opened 2019-12-18
Changing the LTI outcomes to use dicts not templates [DEAD]

jon-betts:
What has happened?

 * We swap Jinja2 templates for dicts
 * We swap XPATH for dict key lookups

Desired benefits

 * This allows us to create a data structure and then modify it
 * This allows us to separate out canvas specific behavior
 * In future we may be able to tuck this away elsewhere
 * XPATH with namespaces is gross

Other benefits / changes:

 * We now lookup from the XML root instead of searching
 * Tests now more localised and easy to follow
 * Failure to lookup a dict key gives feedback about which one
    * XPATH just fails
 * ~This adds a general purpose XML manipulation library which can be moved out of the code (and as of writing, and no tests, doesn't greatly increase the number of lines).~ Used https://github.com/martinblech/xmltodict instead


seanh:
Can this be split up to make it easier to review? It's quite hard to review large diffs, especially when it also includes unrelated changes. At a glance it looks like it might be split up into:

1. Changing the names of existing parameters
2. Editing existing docstrings
3. Moving existing helpers to the bottom of the class
4. Moving `_get_oauth_client()` into a separate method
5. Replacing `LTI_OUTCOME_REQUEST_TEMPLATE` with the new approach
6. Replacing `REPLACE_RESULT_REQUEST_TEMPLATE` with the new approach
7. Replacing `READ_RESULT_REQUEST_TEMPLATE` with the new approach
8. Replacing the parsing of the response to extract the score, with the new approach

The first 4 look like they could be separate, trivial PRs that'd be very quick and easy to review, and getting them out of the way would make the diffs of the following PRs much cleaner.

jon-betts:
It can definitely be broken up, but I don't think I understand how I'd do it in the parts you've specified, as I don't think some of those happened as you described.

How about this?

 * Change all POX body generation to new method
 * Change remaining XML generation to new method
 * Change tests to generate code using new method
 * Change XML parsing to dict based
 * Change tests to parse responses using new method

I think most of the other changes you identified apart from some doc-string reformatting are a necessary consequence of the above, rather than just shuffling things about. There aren't really any cases where things that are at the top of the file go to the bottom without becoming something radically different.

jon-betts:
I've done all the XML generation in one request, because I think it's not too big, and otherwise we have a horrible mash of Templates and dicts all at once in https://github.com/hypothesis/lms/pull/1287

jon-betts:
Split off the parsing into the next one: https://github.com/hypothesis/lms/pull/1288

And comment changes into: https://github.com/hypothesis/lms/pull/1288

PR #1287, opened 2019-12-19
LTI Outcomes: Python dicts for XML generation and separate Canvas specific stuff

jon-betts:
From https://github.com/hypothesis/lms/pull/1284

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -75,7 +75,7 @@ class LTIOutcomesClient:
             return None
 
     def record_result(  # pylint:disable=no-self-use
-        self, outcomes_request_params, score=None, **kwargs,
+        self, outcomes_request_params, score=None, **canvas_extras,
     ):
         """
         Record a score or grading view launch URL for an assignment in the LMS.
@@ -110,7 +110,7 @@ class LTIOutcomesClient:
             }
 
         # Canvas specific adaptations
-        self._canvas_request_modification(request, **kwargs)
+        self._canvas_request_modification(request, **canvas_extras)
 
         self._send_request(
             outcomes_request_params, request_body={"replaceResultRequest": request}
@@ -120,6 +120,8 @@ class LTIOutcomesClient:
     def _canvas_request_modification(
         cls, request, lti_launch_url=None, submitted_at=None, **_
     ):
+        # For details of Canvas extensions see:
+        # https://erau.instructure.com/doc/api/file.assignment_tools.html
 
         if lti_launch_url:
             request["resultRecord"].setdefault("result", {})["resultData"] = {
@@ -147,7 +149,7 @@ class LTIOutcomesClient:
         :rtype: ElementTree.Element
         """
 
-        xml_body = xmltodict.unparse(cls._pox_wrapper(request_body))
+        xml_body = xmltodict.unparse(cls._pox_envelope(request_body))
 
         # Sign request using OAuth 1.0.
         oauth_client = OAuth1(
@@ -196,22 +198,19 @@ class LTIOutcomesClient:
 
         return xml
 
-    @classmethod
-    def _pox_wrapper(cls, body):
+    @staticmethod
+    def _pox_envelope(body):
         """
         Return ``body`` wrapped in an imsx_POXEnvelopeRequest envelope.
-        
-        Return an xmltodict-format XML dict - an XML document represented as a dict,
-        suitable for converting into an XML string by passing it to
+
+        Return an xmltodict-format XML dict - an XML document represented as
+        a dict, suitable for converting into an XML string by passing it to
         ``xmltodict.unparse()``.
-        
-        The returned dict renders to an ``<imsx_POXEnvelopeRequest>`` document with
-        the given ``body`` as the contents of the ``<imsx_POXBody>`` tag.
+
+        The returned dict renders to an ``<imsx_POXEnvelopeRequest>`` document
+        with the given ``body`` as the contents of the ``<imsx_POXBody>`` tag.
 
         :rtype: dict
-
-        See https://www.imsglobal.org/specs/ltiomv1p0/specification for
-        specs for all messages.
         """
         return {
             "imsx_POXEnvelopeRequest": {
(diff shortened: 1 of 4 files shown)


lms/services/lti_outcomes.py
```diff
@@ -176,16 +137,17 @@ def find_element(cls, xml_element, path):
         return xml_element.find(f".//{xpath}")
 
     @classmethod
-    def _send_request(cls, outcomes_request_params, pox_body):
+    def _send_request(cls, outcomes_request_params, request_body):
```
jon-betts: `pox_body` is a bad name because there are two POX bodies. The one we send and the one we receive.

lms/services/lti_outcomes.py
```diff
@@ -118,11 +50,19 @@ def read_result(self, outcomes_request_params):  # pylint:disable=no-self-use
 
         :return: The last-submitted score or `None` if no score has been submitted.
         """
-        body = READ_RESULT_REQUEST_TEMPLATE.render(
-            lis_result_sourcedid=outcomes_request_params.lis_result_sourcedid
-        )
```
seanh: So previously both `read_result()` and `record_result()` had to render their respective templates themselves and pass the rendered body (an XML string) into `_send_request()`. Now they can both just pass in a dict literal :+1:

lms/services/lti_outcomes.py
```diff
@@ -118,11 +50,19 @@ def read_result(self, outcomes_request_params):  # pylint:disable=no-self-use
...
+            outcomes_request_params,
+            request_body={
+                "readResultRequest": {
+                    "resultRecord": {
+                        "sourcedGUID": {
+                            "sourcedId": outcomes_request_params.lis_result_sourcedid
+                        }
+                    }
+                }
+            },
```
seanh: A lot cleaner than the previous `READ_RESULT_REQUEST_TEMPLATE`, and also inlined in the place where it's used :+1:

lms/services/lti_outcomes.py
```diff
@@ -135,11 +75,7 @@ def read_result(self, outcomes_request_params):  # pylint:disable=no-self-use
             return None
 
     def record_result(  # pylint:disable=no-self-use
-        self,
-        outcomes_request_params,
-        score=None,
-        lti_launch_url=None,
-        submitted_at=None,
+        self, outcomes_request_params, score=None, **kwargs,
```
seanh: Ok, so `outcomes_request_params` and `score` are still here but `lti_launch_url` and `submitted_at` have become `**kwargs`... They pop out again as separate `lti_launch_url` and `submitted_at` params in `_canvas_request_modification()`. So this is done to decouple `record_result()` from `_canvas_request_modification()`, to decouple it from Canvas specifics.

I wonder if `**kwargs` could be named something like `**canvas_kwargs`

lms/services/lti_outcomes.py
```diff
@@ -159,14 +95,39 @@ def record_result(  # pylint:disable=no-self-use
...
-            lti_launch_url=lti_launch_url,
-            submitted_at=submitted_at,
+
+        request = {
+            "resultRecord": {
+                "sourcedGUID": {
+                    "sourcedId": outcomes_request_params.lis_result_sourcedid
+                }
+            }
+        }
```
seanh: A lot cleaner than the previous `REPLACE_RESULT_REQUEST_TEMPLATE`, inlined where it's used, and contains only the unconditional parts on their own :+1:

lms/services/lti_outcomes.py
```diff
@@ -159,14 +95,39 @@ def record_result(  # pylint:disable=no-self-use
...
+
+        if score:
+            request["resultRecord"]["result"] = {
+                "resultScore": {"language": "en", "textString": score}
+            }
```
seanh: Since the XML is now data, we can conditionally mutate it :+1:

lms/services/lti_outcomes.py
```diff
@@ -159,14 +95,39 @@ def record_result(  # pylint:disable=no-self-use
...
+
+        # Canvas specific adaptations
+        self._canvas_request_modification(request, **kwargs)
```
seanh: Canvas-specific adaptations shuffled away into a separate helper :+1:

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -3,53 +3,18 @@
...
-          <language>en</language>
-          {% if not exclude_score %}
-          <textString>{{ score }}</textString>
-          {% endif %}
-        </resultScore>
-      </result>
-    </readResultResponse>
-  </imsx_POXBody>
-</imsx_POXEnvelopeResponse>"""
-)
```
seanh: Even more inline XML removed :)

lms/services/lti_outcomes.py
```diff
@@ -233,3 +195,24 @@ def _send_request(cls, outcomes_request_params, pox_body):
             raise LTIOutcomesAPIError("LTI outcome request failed")
 
         return xml
+
+    @classmethod
+    def _pox_wrapper(cls, body):
```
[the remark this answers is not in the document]
jon-betts: Yes it would

lms/services/lti_outcomes.py
jon-betts: This is kind of phase 1 to pushing the caller outside of this class entirely. So the next phase is to remove the kwargs entirely and have a call-back passed in.

lms/services/lti_outcomes.py
```diff
@@ -233,3 +195,24 @@ def _send_request(cls, outcomes_request_params, pox_body):
...
+        """
+        Envelope for LTI Outcome Service requests sent to the LMS.
```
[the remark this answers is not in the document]
jon-betts: Yep, this is better. I was just took what was there before.

lms/services/lti_outcomes.py
```diff
@@ -233,3 +195,24 @@ def _send_request(cls, outcomes_request_params, pox_body):
             raise LTIOutcomesAPIError("LTI outcome request failed")
 
         return xml
+
+    @classmethod
```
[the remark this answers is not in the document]
jon-betts: I thought for some reason that if you made it static you had to call it by ClassName rather than just self or cls, but that doesn't seem to be the case. Changed that now.

jon-betts:
You trust I tested this incorrectly. But it's done now and looks ok as far as I can tell in Canvas and Moodle.

PR #1288, opened 2019-12-19
LTI Outcomes: Use python dicts to navigate XML

jon-betts:
Child of: https://github.com/hypothesis/lms/pull/1284

This helps in a few ways:

 * Some general de-maddening of the tests
 * There was some horribly grouped stuff that was test specific
 * When we fail on a key lookup we now know which key was missing
 * Generally a bit easier on the eyes than XPATH

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -1,5 +1,5 @@
 from typing import NamedTuple
-from xml.etree import ElementTree
+from xml.parsers.expat import ExpatError
 
 import requests
 import xmltodict
@@ -39,8 +39,6 @@ class LTIOutcomesClient:
     See https://www.imsglobal.org/specs/ltiomv1p0/specification.
     """
 
-    XML_NS = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0"
-
     def __init__(self, _context, request):
         pass
 
@@ -65,13 +63,10 @@ class LTIOutcomesClient:
         )
 
         try:
-            score = self.find_element(
-                result, ["readResultResponse", "result", "resultScore", "textString"]
+            return float(
+                result["readResultResponse"]["result"]["resultScore"]["textString"]
             )
-            if score is None:
-                return None
-            return float(score.text)
-        except (ValueError, TypeError):
+        except (KeyError, ValueError):
             return None
 
     def record_result(  # pylint:disable=no-self-use
@@ -132,21 +127,14 @@ class LTIOutcomesClient:
             request["submissionDetails"] = {"submittedAt": submitted_at.isoformat()}
 
     @classmethod
-    def find_element(cls, xml_element, path):
-        """Extract element from LTI Outcomes Management XML response."""
-        xpath = "/".join([f"{{{cls.XML_NS}}}{name}" for name in path])
-
-        return xml_element.find(f".//{xpath}")
-
-    @classmethod
     def _send_request(cls, outcomes_request_params, request_body):
         """
         Send a signed request to an LMS's Outcome Management Service endpoint.
 
-        :arg request_body: The content to send as the body element of the
+        :arg request_body: The content to send as the POX body element of the
                            request
-        :return: Parsed XML response
-        :rtype: ElementTree.Element
+        :return: The returned POX body element
+        :rtype: dict
         """
 
         xml_body = xmltodict.unparse(cls._pox_envelope(request_body))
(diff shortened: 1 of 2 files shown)


jon-betts:
Should be merged after: https://github.com/hypothesis/lms/pull/1287

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -214,63 +166,50 @@ def test_it_gracefully_handles_RequestException(
...
-        """Return parsed XML of last request."""
-
-        def xml():
-            request = httpretty.last_request()
-            return ElementTree.fromstring(request.body)
-
-        return xml
-
-    @pytest.fixture
-    def record_result_request_fields(self, request_xml):
```
jon-betts: This function is very bad for understanding the tests. Each of the things here pretty much only relate to one test but are grouped together here.

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -300,23 +239,12 @@ def requests(self, patch):
         return patch("lms.services.lti_outcomes.requests")
 
     @classmethod
-    def element_text(cls, xml, path):
-        element = LTIOutcomesClient.find_element(xml, path)
```
jon-betts: Uses the code we are testing to run the tests. This seems like something to avoid.

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -300,23 +239,12 @@ def requests(self, patch):
...
-                    "imsx_POXHeader",
-                    "imsx_POXRequestHeaderInfo",
-                    "imsx_messageIdentifier",
-                ],
-            )
-            == "999999123"
-        )
+        body = cls.sent_body()
+
+        header = body["imsx_POXEnvelopeRequest"]["imsx_POXHeader"]
```
jon-betts: Although it's not obvious from the code above, the previous code would have match "imsx_POXHeader" anywhere in the XML. We now specifically navigate from root.

seanh:
Let's try this: it's not a technique that I've tried on GitHub before but since this PR depends on <https://github.com/hypothesis/lms/pull/1287>, which is not merged yet, I've changed this PR's base branch from `master` to <https://github.com/hypothesis/lms/pull/1287>'s branch. So the diff on this PR now shows only the changes that belong to _this_ PR, and it can be more easily reviewed.

We may not merge this PR into <https://github.com/hypothesis/lms/pull/1287>. Maybe we'll merge <https://github.com/hypothesis/lms/pull/1287> into master first and then rebase this PR against master. But it can at least be reviewed easily against <https://github.com/hypothesis/lms/pull/1287>

lms/services/lti_outcomes.py
```diff
@@ -65,13 +63,10 @@ def read_result(self, outcomes_request_params):  # pylint:disable=no-self-use
...
-            score = self.find_element(
-                result, ["readResultResponse", "result", "resultScore", "textString"]
+            return float(
+                result["readResultResponse"]["result"]["resultScore"]["textString"]
             )
-            if score is None:
-                return None
-            return float(score.text)
-        except (ValueError, TypeError):
+        except (KeyError, ValueError):
```
seanh: Ok, so even though we're now using dicts to _render_ XML, we were still using XPath to _parse_ XML and pull certain parts out of it. Now, we use xmltodict to parse the XML into a dict and then just pull things out of the dict.

lms/services/lti_outcomes.py
```diff
@@ -129,22 +124,15 @@ def _canvas_request_modification(
         if submitted_at:
             request["submissionDetails"] = {"submittedAt": submitted_at.isoformat()}
 
-    @classmethod
-    def find_element(cls, xml_element, path):
-        """Extract element from LTI Outcomes Management XML response."""
-        xpath = "/".join([f"{{{cls.XML_NS}}}{name}" for name in path])
-
-        return xml_element.find(f".//{xpath}")
```
seanh: Yeah. Pretty good to see this gone. WTF is `{{{`

tests/unit/lms/services/lti_outcomes_test.py
seanh: Yeah, that sounds terrible, and now how fixtures should be used

lms/services/lti_outcomes.py
jon-betts: The inner one is a string format, then I think the next is an escape.

ElementTree munges the namespace into the tag name like:

 * `{namespace}tag`

Although the kicker here is you can call:

```
xml_element.find('//ns:tag/ns:another_tag', {'ns': "http:...." })
```

jon-betts:
That re-pointing to get the diff worked fine, although it confused the hell out of me!

I've tested this against Moodle (which actually cares about the value we read) and it's working fine.

PR #1292, opened 2019-12-19
A separate OAuth1 service to cut down on passing parameters about

jon-betts:
* The new OAuth1 service now encapsulates the secret and oauth key
 * This pulls it out of the objects which everyone was passing about
 * This is part of the work to eliminate the LTIOutcomesRequestParams object

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -4,7 +4,6 @@ from xml.parsers.expat import ExpatError
 import requests
 import xmltodict
 from requests import RequestException
-from requests_oauthlib import OAuth1
 
 from lms.services.exceptions import LTIOutcomesAPIError
 
@@ -13,12 +12,6 @@ __all__ = ["LTIOutcomesClient", "LTIOutcomesRequestParams"]
 
 class LTIOutcomesRequestParams(NamedTuple):
     """Common parameters used by all LTI Outcomes Management requests."""
-
-    consumer_key: str
-    """OAuth 1.0 consumer key used to sign the request."""
-
-    shared_secret: str
-    """OAuth 1.0 shared secret used to sign the request."""
 
     lis_outcome_service_url: str
     """URL to submit requests to, provided by the LMS during an LTI launch."""
@@ -40,7 +33,7 @@ class LTIOutcomesClient:
     """
 
     def __init__(self, _context, request):
-        pass
+        self.oauth1_service = request.find_service(name="oauth1")
 
     def read_result(self, outcomes_request_params):  # pylint:disable=no-self-use
         """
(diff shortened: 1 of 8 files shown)


lms/services/lti_outcomes.py
```diff
@@ -82,12 +81,6 @@
 class LTIOutcomesRequestParams(NamedTuple):
     """Common parameters used by all LTI Outcomes Management requests."""
 
-    consumer_key: str
-    """OAuth 1.0 consumer key used to sign the request."""
-
-    shared_secret: str
-    """OAuth 1.0 shared secret used to sign the request."""
```
seanh: These aren't really params for the LTI outcomes request, but rather things that are needed in order to sign the request, so removing them from here makes sense :+1:

lms/services/lti_outcomes.py
```diff
@@ -110,7 +103,7 @@ class LTIOutcomesClient:
     XML_NS = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0"
 
     def __init__(self, _context, request):
-        pass
+        self.oath1_service = request.find_service(name="oauth1")
```
seanh:
```suggestion
        self.oauth1_service = request.find_service(name="oauth1")
```

lms/services/lti_outcomes.py
```diff
@@ -187,24 +179,13 @@ def _send_request(cls, outcomes_request_params, pox_body):
...
-        )
-
         try:
             response = None  # Bind the variable so we can refer to it in the catch
             response = requests.post(
                 url=outcomes_request_params.lis_outcome_service_url,
                 data=xml_body,
                 headers={"Content-Type": "application/xml"},
-                auth=oauth_client,
+                auth=self.oath1_service.get_client(),
```
seanh:
```suggestion
                auth=self.oauth1_service.get_client(),
```

tests/unit/lms/services/oauth1_test.py
```diff
@@ -0,0 +1,53 @@
...
+        assert "oauth_signature=" in auth_header
+
+    @pytest.fixture
+    def service(self, context, pyramid_request):
+        return OAuth1Service(context, pyramid_request)
+
+    @pytest.fixture
+    def context(self):
+        # We don't use context, so it doesn't matter what it is
+        return mock.sentinel.context
```
seanh: Nice use of fixture composition too :+1:

lms/services/lti_outcomes.py
jon-betts: I have no idea why I do that, but I do it every time...

jon-betts:
I've tested this locally so I'll add it to the LMS log jam and test it there too.

PR #1295, opened 2019-12-20
Renaming LISResultSourcedId to GroupInfo

jon-betts:
`LISResultSourcedId` is both confusing and misleading. The object is actually a compound of various pieces of data of which `lis_result_sourcedid` is only one. It's also completely hostile to understanding what the hell it is.

I've tried to fix everywhere referring to this but it's likely I've missed something. The LTI field is still called `lis_result_sourcedid` so that's still all over our code.

### DIFF
--- a/tests/unit/lms/models/grading_info_test.py
+++ b/tests/unit/lms/models/grading_info_test.py
@@ -1,15 +1,15 @@
 import pytest
 import sqlalchemy.exc
 
-from lms.models import ApplicationInstance, LISResultSourcedId
+from lms.models import ApplicationInstance, GradingInfo
 
 
-class TestLISResultSourcedId:
+class TestGradingInfo:
     def test_it_persists_and_returns_attrs(
-        self, application_instance, db_session, lis_result_sourcedid
+        self, application_instance, db_session, grading_info
     ):
-        db_session.add(lis_result_sourcedid)
-        lrs = db_session.query(LISResultSourcedId).one()
+        db_session.add(grading_info)
+        lrs = db_session.query(GradingInfo).one()
 
         assert lrs.lis_result_sourcedid == "result_sourcedid"
         assert lrs.lis_outcome_service_url == "https://somewhere.else"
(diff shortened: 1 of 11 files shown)


tests/unit/lms/services/grading_info_test.py
```diff
@@ -3,47 +3,42 @@
...
         )
 
-        assert len(students) == 2
+        assert len(grading_infos) == 2
 
     def test_it_returns_empty_list_if_no_matching_records(
         self, svc, lti_params, lti_user
     ):
-        students = svc.fetch_students_by_assignment(
+        grading_infos = svc.get_by_assignment(
```
jon-betts: Not super grammatical, but what are you going to do.

tests/unit/lms/views/helpers/frontend_app_test.py
```diff
@@ -84,36 +80,24 @@ def test_it_sets_course_and_assignment_names(self, grading_request):
...
-    return [lis_result_1, lis_result_2, lis_result_3]
+def grading_infos():
+    return [
+        GradingInfo(
+            h_username=f"username_{index}",
+            h_display_name=f"Student {index}",
+            lis_result_sourcedid=f"lis_result_sourcedid_{index}",
+            lis_outcome_service_url="http://example.com/service_url",
+        )
+        for index in range(3)
```
jon-betts: The exact content of these records is irrelevant as long as they are different

tests/unit/lms/models/grading_info_test.py
```diff
@@ -70,29 +70,29 @@ def application_instance(self, db_session):
...
-            oauth_consumer_key=application_instance.consumer_key,
-            user_id="339483948",
-            context_id="random context",
-            resource_link_id="random resource link id",
-            tool_consumer_info_product_family_code="MyFakeLTITool",
-            h_username="ltiuser1",
-            h_display_name="My Fake LTI User",
-        )
+    def grading_info(self, grading_info_params):
+        return GradingInfo(**grading_info_params)
```
jon-betts: Squished some repetition while I was here

PR #1297, opened 2019-12-20
Fixing a bug where we fail upon getting None as a grade

jon-betts:
Fixing a bug I introduced (but haven't released) detected during testing in QA. This is happens if we get None as a grade back.

The error was in breaking existing test coverage and not covering None as a case.


jon-betts:
Sean has eyeballed this on a call with me.

PR #1298, opened 2019-12-20
LTI Outcomes: Removing the value object and replacing with a single value

jon-betts:
Generally the service URL doesn't change for an installation. Therefore
the only variable variable here is 'lis_result_sourcedid'.

This simplifies the callers a lot, which still making it obvious where
the critical part of information is coming from.

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -1,4 +1,3 @@
-from typing import NamedTuple
 from xml.parsers.expat import ExpatError
 
 import requests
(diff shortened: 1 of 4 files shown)


jon-betts:
Looks like I've left a comment on a random ticket / not saved it so long story short:

I agree `lis_outcome_service_url` shouldn't be on ApplicationInstance

Explaining the thought process between keeping the two separate:

* Both are always passed in, the question is just whether it's via constructor or method
* Therefore, the choice of one vs. the other is mostly about signalling to ourselves and making the code easier to read
* Whilst both can change, one is much more changeable than the other
* A common way to deal with this with a client object is to have `HAPI(hostname=blah).get_annotation(annotation_id)`
* In this case the service URL looks more like a constructor than a method parameter to me

PR #1341, opened 2020-01-15
WIP: Add canvasapi command-line script

seanh:
This adds a canvasapi command-line script (call it like this:
`tox -qe dev --run-command 'canvasapi conf/development.ini'`) that
allows you to call our Canvas API client (the one from
`lms/services/canvas_api.py`) from the command line in order to easily
experiment with the Canvas API.

This script turned out to be unnecessary: it turns out that a Canvas user
can generate a non-expiring access token for their account that can be
easily passed to curl, httpie, etc. So you can easily experiment with the
Canvas API using those tools. It isn't necessary to have a CLI for the
LMS app's Canvas API client. See:
https://community.canvaslms.com/docs/DOC-16005-42121018197

Nonetheless, I'm committing this work on a branch for posterity.

The script will use an existing, working Canvas API access token from
the LMS app's DB, if there is one. If there's no access token in the DB,
or if its expired (and can't be refreshed) then the script will run the
LMS app locally on a host and port and print out a URL for you to open
in a browser in order to get a new access token. After you click the
[Authorize] button in Canvas in the browser window, the script will
automatically continue using the freshly-authorized access token.

The script requires a Canvas LTI user_id and oauth_consumer_key which
are hard-coded to `***` values, so it won't work unless you fix these
`***` values in the code.

The script is also hard-coded to only call Canvas's list_files API for a
hard-coded course ID.

When the script prints out a URL for you to open in order to authorize
us with Canvas and get a new access token, it assumes that you will log
in to Canvas (or are already logged in to Canvas) as the Canvas user
corresponding to the hard-coded LTI user_id and oauth_consumer_key.

### DIFF
--- a/lms/scripts/__init__.py
+++ b/lms/scripts/__init__.py
@@ -1,3 +1,4 @@
+from lms.scripts.canvasapi import canvasapi
 from lms.scripts.devdata import devdata
 
-__all__ = ["devdata"]
+__all__ = ["canvasapi", "devdata"]
--- a/lms/scripts/canvasapi.py
+++ b/lms/scripts/canvasapi.py
@@ -1,0 +1,154 @@
+import configparser
+import json
+import sys
+from wsgiref.simple_server import make_server
+
+from pygments import highlight
+from pygments.lexers import JsonLexer
+from pygments.formatters import TerminalFormatter
+from pyramid.paster import bootstrap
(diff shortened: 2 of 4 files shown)


seanh:
Closing this because it turned out to be unnecessary

Issue #1334, opened 2020-01-14
Prove that we can connect to Blackboard Classic API via OAuth2


jon-betts:
As a part of #1322 (Work out what is required to support Blackboard local file integration) the first order of business should be to work out how to authenticate.

Blackboard lists "Three Legged Oauth (3LO)" which it's assumed just means the grant flow, but it's not totally clear.

This ticket should therefore either:

Produce some prototype code to demonstrate:

 * [x] Connecting via OAuth2
 * [x] Calling at least one API end-point which requires authentication successfully
 * :no_entry_sign: The code does not need to be production quality

... or conclude that we need to do something else:
  * STOP!
  * Probably time to re-plan

Details of premilinary research can be found at: https://github.com/hypothesis/lms/issues/1319

jon-betts:
I've created some code which proves this can be done. I don't know how standard this is, but I've given it a shot.

jon-betts:
This has been done, but not really reviewed. We'll have to do that later.

PR #1347, opened 2020-01-16
OAuth2 demonstration for Blackboard Classic

jon-betts:
This is for ticket: https://github.com/hypothesis/lms/issues/1334

This is demonstrating:

 * Generating an authentication code URL
 * Getting OAuth2 tokens from codes returned from that URL
 * Refreshing tokens using refresh tokens
 * A way to separate retrying from calling withing a 'retry_session' context manager (inside things retry, outside they don't)
 * A call which requires auth and one that doesn't

# :warning: Don't merge or close this! :warning: 

I intent to keep this alive as a separate branch and merge other parts of the proof of concept into this branch.


jon-betts:
There's no further work going on in this. So it's basically done. I think if I close the ticket though there are bots which close the PR which I don't want so I'll leave it open.

jon-betts:
I've spoken with Sean and this PR is probably way to rough to review as it stands, so I'll need to get it to a better place before it can be worked over.

Issue #1335, opened 2020-01-14
Prove we can list and retrieve Blackboard Classic attachments


jon-betts:
Once we have proven we can authenticate (https://github.com/hypothesis/lms/issues/1334) we should attempt to list and retrieve attachments.

The output of this ticket should be a working prototype which:

 * [x] Demonstrates listing files
 * [x] Demonstrates getting a particular file
 * [x] Determines if we can use the URL to annotate a file directly (or if we need to do something else):
   * We can get a temporary link which a user can follow
   * This doesn't last very long, so we need to get this repeatedly
 * :no_entry_sign: Does not need to be production quality
 * :no_entry_sign: Should not be integrated into the LMS product (a demo script is fine).

Background research at: https://github.com/hypothesis/lms/issues/1319

jon-betts:
Some extra details about potential alternatives and some Classic vs Ultra info in:

https://community.blackboard.com/discuss/viewtopic/162/2280?post_id=7917#p7917

jon-betts:
This has been done, but not really reviewed. We'll have to do that later.

PR #1350, opened 2020-01-20
Blackboard Classic file tree building

jon-betts:
For https://github.com/hypothesis/lms/issues/1335

Significant changes to how this is working:

 * Merged the OAuth client and manager
 * Found out we have recursive methods on the course and content folder
 * Added a separate tree building class (not connected to the blackboard code, so re-usable?)
 * Tried to prune out some calls and code we aren't actually using
 * More besides I'm sure


seanh:
Any links (e.g. to docs for Blackboard API endpoints) or documentation (e.g. PR description explaining and motivating the main changes, PR diff comments, ...). It's quite large and I'm not really sure what to make of it. There's quite a lot here including, at a glance, a Blackboard Classic API client with tree building and quite a lot of other stuff, a generic HTTP JSON API client, a generic OAuth 2 client, and a new top-level logic package. I know this isn't meant for merging all as one but it's a bit overwhelming to even give feedback on.

At a glance I'd probably caution against rewriting the world all at once, when it comes to things like a generic API client and generic OAuth 2 client, especially since I'm simultaneously working on our OAuth 2 and our Canvas API and h API code

jon-betts:
With regards to re-writing the world etc, unless there is a different client I can use, I do need to write one. So re-usable or not is the only question. I don't think it really increases the complexity to make it re-usable etc. Anyway this wasn't really intended to be in a fit state to review necessarily, but it does work.

I think we've agreed that I'll just work on this until it's good enough to demo, and then try and work out a sane way to break it up for review.

Issue #1377, opened 2020-01-27
Enable eslint-plugin-jsx-a11y and fix existing issues


robertknight:
As part of our general effort to improve accessibility and achieve WCAG compliance, introduce static linting to check for a11y errors in markup.

- [ ] Enable the jsx-a11y ESLint plugin for the lms frontend. Any existing failures should be flagged with a comment (`FIXME-A11Y`) and then temporarily suppressed
- [ ] Fix errors in `BasicLtiLaunchApp.js`
- [ ] Fix errors in `Dialog.js`
- [ ] Fix errors in `FileList.js`
- [ ] Fix errors in `LMSFilePicker.js`
- [ ] Fix errors in `StudentSelector.js`
- [ ] Fix errors in `ValidationMessage.js`

See also: The [corresponding issue](https://github.com/hypothesis/client/issues/1727) for the Hypothesis client.

klemay:
@robertknight since all of these items are checked off, can we close this issue?

PR #1378, opened 2020-01-27
Enable jsx-a11y ESLint plugin

robertknight:
This enables the [jsx-a11y ES Lint plugin](https://github.com/evcohen/eslint-plugin-jsx-a11y) which can
statically detect a subset of common accessibility errors.

Flag the current set of known errors with a `FIXME-A11Y` comment and
suppress them temporarily, so that we can fix them incrementally while
guarding against new errors.

Part of #1377. Corresponding [client PR](https://github.com/hypothesis/client/pull/1728).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "hypothesis"
+    "hypothesis",
+    "plugin:jsx-a11y/recommended"
   ],
   "rules": {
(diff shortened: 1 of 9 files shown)


robertknight:
The corresponding PR for the client was merged (https://github.com/hypothesis/client/pull/1728). As this does exactly the same for the LMS app I'm going to go ahead and merge it.

PR #1395, opened 2020-01-30
Add an initial set of frontend a11y tests using axe-core

robertknight:
This adds an initial set of automated runtime accessibility tests to the LMS frontend JS apps using axe-core, following the approach of https://github.com/hypothesis/client/pull/1729.

Several tests have been skipped with `it.skip` due to issues that were found. These will need to be fixed incrementally.

This initial set of tests is not exhaustive in that it doesn't cover all component states, but it provides a starting point.


lms/static/scripts/test-util/accessibility.js
```diff
@@ -0,0 +1,87 @@
+import { run } from 'axe-core';
```
robertknight: This module was copied from the client repo. I think we are or soon will be at a point where it could make sense to extract a `@hypothesis/ui-test-utils` package out of the `test-util` directory here / in the client and share this code.

LMS007: There are a number of util like things that would be nice to share....rpc files come to mind. I wonder if we can think broadly with any shared code changes?

Issue #1386, opened 2020-01-29
Refactor: Merge the three JavaScript config objects into one


seanh:
The LMS app currently renders three separate JavaScript config objects into the page:

1. The [`js-lms-config` object](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/templates/base.html.jinja2#L38) contains config settings for the LMS app's own Preact app. It's generated by [the `js_config` property](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/resources/_lti_launch.py#L174-L195) in Python (note: inconsistent naming)

2. The [`js-rpc-server-config` object](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/templates/base.html.jinja2#L42) contains config for [the LMS app's standalone JSON RPC client and server](https://github.com/hypothesis/lms/tree/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/static/scripts/postmessage_json_rpc) (which are separate from the Preact app). It's generated by [the `rpc_server_config` property](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/resources/_lti_launch.py#L250-L254)

3. The [`js-hypothesis-config` object](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/templates/base.html.jinja2#L47) contains config for the Hypothesis client. The JSON RPC server passes this config to the client. It's generated by [the `hypothesis_config` property](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/resources/_lti_launch.py#L197-L248)

It's confusing to have three separate config objects and, since we're going to be merging the JSON RPC server code into the Preact app anyway (see https://github.com/hypothesis/lms/issues/1387), we may as well merge the three configs into one js-config object (with sub-objects for the app, RPC, and client configs).

PR #1436, opened 2020-02-05
Move rpc_server_config into js_config

seanh:
Part of https://github.com/hypothesis/lms/issues/1386.

We want to simplify our JavaScript config by merging multiple separate
JavaScript config objects into one nested object.

How it currently works (before this commit)

The backend currently has an `LTILaunchResource.rpc_server_config`
property:
https://github.com/hypothesis/lms/blob/1f10d599e5e2502a4bfbde023da35e852729788e/lms/resources/_lti_launch.py#L250-L254

The `rpc_server_config` property contains the config settings for the
JavaScript JSON-RPC server:
https://github.com/hypothesis/lms/blob/master/lms/static/scripts/postmessage_json_rpc/server/server.js

It gets rendered into the HTML as `"js-rpc-server-config"`:
https://github.com/hypothesis/lms/blob/1f10d599e5e2502a4bfbde023da35e852729788e/lms/templates/base.html.jinja2#L41-L43

The JSON-RPC server then reads this config from the HTML:
https://github.com/hypothesis/lms/blob/1f10d599e5e2502a4bfbde023da35e852729788e/lms/static/scripts/postmessage_json_rpc/server/server.js#L23

What this commit changes
------------------------

This commit merges `LTILaunchResource.rpc_server_config` into the
`LTILaunchResource.js_config` property, which is the "general" config
object for the frontend code.

The settings that were previously rpc_server_config are now an
`"rpc_server"` sub-dict within the dict that `js_config` returns.

The `"js-rpc-server-config"` object is no longer rendered into the HTML,
since the `"rpc_server"` config is now part of the `"js-lms-config"` that
already gets rendered:
https://github.com/hypothesis/lms/blob/1f10d599e5e2502a4bfbde023da35e852729788e/lms/templates/base.html.jinja2#L36-L39

Finally, the JSON-RPC server itself (`server.js`) is changed to read its
config from `js-lms-config["rpc_server"]` rather than from the old
`js-rpc-server-config`.

And a couple of tests need updating.

How to test this
----------------

Launch any assignment in Canvas and test that you're logged into your
Canvas account in the Hypothesis client. If the JSON-RPC server's config
wasn't working, you wouldn't be.

### DIFF
--- a/lms/resources/_lti_launch.py
+++ b/lms/resources/_lti_launch.py
@@ -185,7 +185,16 @@ class LTILaunchResource:
         """
         if self._js_config is None:
             # Initialize self._js_config for the first time.
-            self._js_config = {"urls": {}}
+            self._js_config = {
+                # The config object for the postMessage-JSON-RPC server.
+                "rpcServer": {
+                    "allowedOrigins": self._request.registry.settings[
+                        "rpc_allowed_origins"
+                    ],
+                },
+                # URLs for the frontend to use (e.g. API endpoints for it to call).
+                "urls": {},
+            }
 
             if self._request.lti_user:
                 self._js_config["authToken"] = BearerTokenSchema(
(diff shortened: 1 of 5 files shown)


seanh:
@jon-betts Fixed

PR #1437, opened 2020-02-05
Add a convenient `make sure` command

seanh:
Run the code formatters, linters, tests, functests, bddtests, and
coverage, backend and frontend, in a single `make all` command.

Handy for running locally before you push a pull request, so that you
don't have to type:

    make checkformatting lint test functests bdd coverage

You can _almost_ run it all faster in parallel with:

    make -j `nproc` all

... but sadly this fails because the unit tests, functional tests and
bddtests all share the same test DB. That should be easy enough to fix in
future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ help:
 	@echo "make bddtests          Run the gherkin tests"
 	@echo "make docstrings        View all the docstrings locally as HTML"
 	@echo "make checkdocstrings   Crash if building the docstrings fails"
+	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
 	@echo "make pip-compile       Compile requirements.in to requirements.txt"
 	@echo "make upgrade-package   Upgrade the version of a package in requirements.txt."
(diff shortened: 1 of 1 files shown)


seanh:
Renamed it to `make sure`.

I think I want to keep `checkformatting` rather than `format` -- if the formatting is wrong you want the `make sure` command to stop and exit with an unsuccessful status. Not to keep going (running the tests etc) and exit successfully, having silently edited your working tree. I think that might lead to people thinking all is well and pushing their commits as a PR, not realising that `make format` had created uncommited changes. I do realise that it's annoying to have to subsequently run `make format` yourself though, but :man_shrugging: 

Despite the timing I want to keep the order of tests (unit tests, then functests, then bddtests) the same because I think they're in order of most-likely-to-fail and most-likely-to-produce-the-best-debugging-info. Also most-likely-to-contain-new-or-changed-tests. We should get the tests, functests and bddtests on separate DBs, then we'll be able to run them all in parallel anyway.

PR #1439, opened 2020-02-06
Ignore npm package-lock.json file

LMS007:
LMS uses yarn to manage the lockfile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ secret_env.sh
 .autoenv
 docker.sh
 .devdata.env
+
+# lms uses Yarn rather than npm to manage the lockfile.
+package-lock.json


robertknight:
You shouldn't have this file in your repository if you're using the `make` commands to install dependencies or `yarn install` directly. If you have run `npm install` by mistake, I would recommend to just remove the file, otherwise there is a risk of JS tools seeing it and trying to do something with it. The `.gitignore` file is really for files which we'd expect to appear in the local repository while you're working but shouldn't be committed. For example, the `build/` directory containing compiled JS/CSS assets.

LMS007:
understood, I was confused that we have this in the client ignore list. Is there a difference?

Issue #1441, opened 2020-02-07
LMS front-end calls to the back-end should specify "application/json"


jon-betts:
After a new change to webargs, if you don't send "application/json" then it will silently drop all content. This eventually leads to a 422 (unprocessable entity) response from us on grading submission calls.

This change adds a Content-Type header if a body is set.

jon-betts:
PR for this here: https://github.com/hypothesis/lms/pull/1444

PR #1444, opened 2020-02-07
Specify the content type when sending JSON payloads

jon-betts:
This is a good idea, but mostly motivated by webargs no-longer accepting JSON content without it.

This is for: https://github.com/hypothesis/lms/issues/1441

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/api.js
+++ b/lms/static/scripts/frontend_apps/utils/api.js
@@ -58,7 +58,7 @@ async function apiCall({ path, authToken, data }) {
   const result = await fetch(path, {
     method: data === undefined ? 'GET' : 'POST',
     body,
-    headers: headers,
+    headers,
   });
   const resultJson = await result.json();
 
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/utils/api.js
```diff
@@ -47,16 +47,18 @@ export class ApiError extends Error {
...
+    headers['Content-Type'] = 'application/json; charset=UTF-8';
   }
 
   const result = await fetch(path, {
     method: data === undefined ? 'GET' : 'POST',
     body,
-    headers: {
-      Authorization: authToken,
-    },
+    headers: headers,
```
robertknight:
```suggestion
    headers,
```

You can shorten eg. ` { propertyA: propertyA, propertyB: propertyB }` to `{ propertyA, propertyB }` and we usually do.

jon-betts: Cool. I've added that. I'll wait for the tests to pass and merge I think.

PR #1445, opened 2020-02-07
Add front-end formatting to the format command

jon-betts:


jon-betts:
Sean have this a thumbs up in slack

PR #1446, opened 2020-02-07
Separate JSON schemas into their own class and check content type

jon-betts:
* Raise an exception when JSON payloads have wrong content type
 * Added a content_type to the base pyramid request object
 * Lots of changes to validation/_api_test.py to use a new fixture
   * This fixture is a JSON request which...
   * Sets the content type
   * Encodes the JSON content
 * Added the content type and some variations to the base tests

----

# To review

* Easiest is probably to login to canvas as a student
* Open an assigment and pop it into a new window
* Use the dev tools to look at network requests
* Filter by 'submissions' and copy the contents as a curl request
* Modify the content type to something other than application/json

### DIFF
--- a/lms/validation/_base.py
+++ b/lms/validation/_base.py
@@ -101,7 +101,8 @@ class JSONPyramidRequestSchema(PyramidRequestSchema):
 
     @pre_load
     def check_content_type(self, data, **_):
-        """Check the request has content type 'application/json'.
+        """
+        Check the request has content type 'application/json'.
 
         :raise HTTPUnsupportedMediaType: If the content type is wrong
         """
--- a/lms/validation/_base.py
+++ b/lms/validation/_base.py
@@ -1,5 +1,7 @@
 """Base classes for validation schemas."""
 import marshmallow
+from marshmallow import pre_load
+from pyramid.httpexceptions import HTTPUnsupportedMediaType
 from webargs import pyramidparser
 
 from lms.validation._exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


lms/validation/_base.py
```diff
@@ -89,6 +91,31 @@ def _handle_error(error, _req, _schema, _status_code, _headers):
...
+        :raise HTTPUnsupportedMediaType: If the content type is wrong
+        """
+        content_type = self.context["request"].content_type
+
+        if content_type != "application/json":
+            raise HTTPUnsupportedMediaType(
+                f"Unexpected content type. Expected 'application/json' but found '{content_type}'",
+            )
+
+        return data
```
seanh: :+1: This worked out pretty nicely

tests/unit/lms/validation/_api_test.py
```diff
@@ -144,3 +134,18 @@ def all_fields(self):
             "lis_result_sourcedid": "modelstudent-assignment1",
             "score": 0.5,
         }
+
+
+@pytest.fixture
+def json_request(pyramid_request):
```
seanh: Changing the name of the fixture is a good idea, makes the tests more readable

seanh: This would be confusing if a test used both `json_request` and `pyramid_request` at once, since `json_request` mutates `pyramid_request`. And is the same object. It could be fixed. But no one does use them both though

tests/unit/lms/validation/_api_test.py
```diff
@@ -144,3 +134,18 @@ def all_fields(self):
...
+    def _make_json_request(data, exclude=None):
+        pyramid_request.content_type = "application/json"
+
+        if exclude:
+            for field in exclude:
+                data.pop(field, None)
+
+        pyramid_request.body = json.dumps(data)
+        return pyramid_request
```
seanh: I think this function returns the same object instance each time you call it

seanh:
I gave this a fairly thorough test locally just in case and it all seems to work

tests/unit/lms/validation/_api_test.py
seanh: The fix would be to add a global `make_pyramid_request` fixture that returns a new Pyramid request every time its called. Then change the global `pyramid_request` fixture to depend on `make_pyramid_request`. Then this `json_request` fixture would depend on `make_pyramid_request` as well, not on `pyramid_request`

tests/unit/lms/validation/_api_test.py
seanh: (The fix would be as suggested above)

jon-betts: Right, so if any single test called it twice, it would be mutating the same instance. We just happen to be ok here because there's only ever one request.

I'm going to assume this is ok for now.

PR #1447, opened 2020-02-07
Specify query string as the place to look for the Canvas OAuth schema

jon-betts:
This is to head off problems with defaults changing in webargs.


jon-betts:
I've tested loading a canvas files assigment in dev:

https://hypothesis.instructure.com/courses/83/assignments/463

Seems to work fine?

PR #1451, opened 2020-02-10
Remove Via 3 and Via 2 feature flags and settings

seanh:
It now just uses whatever version of Via is in the `via_url` setting all the time. Fixes https://github.com/hypothesis/via3/issues/80


seanh:
Gonna close this for now as I think we're gonna be switching the LMS app back to legacy Via, and we'll still want to be able to test Via 3 behind a feature flag

PR #1455, opened 2020-02-10
Move hypothesis_config into js_config

seanh:
Part of https://github.com/hypothesis/lms/issues/1386.

We want to simplify our JavaScript config by merging multiple separate
JavaScript config objects into one nested object, so move the separate
`LTILaunchResource.hypothesis_config` property into
`LTILaunchResource.js_config` as a `"hypothesisClient"` sub-dict.

How it currently works (before this commit)

The backend currently has an `LTILaunchResource.hypothesis_config`
property:
https://github.com/hypothesis/lms/blob/0abc66323919997cd76d5109021e9eafafe5c336/lms/resources/_lti_launch.py#L206-L257

The `hypothesis_config` property contains the config settings that the
JSON-RPC server will pass to the Hypothesis client.

It gets rendered into the HTML as `"js-hypothesis-config"`:
https://github.com/hypothesis/lms/blob/0abc66323919997cd76d5109021e9eafafe5c336/lms/templates/base.html.jinja2#L41-L44

The JSON-RPC server then reads this config from the HTML:
https://github.com/hypothesis/lms/blob/0abc66323919997cd76d5109021e9eafafe5c336/lms/static/scripts/postmessage_json_rpc/server/methods.js#L13-L14

What this commit changes
------------------------

This commit moves `LTILaunchResource.hypothesis_config` into the
`LTILaunchResource.js_config` property, which is the "general" config
object for the frontend code.

The settings that were previously the separate `hypothesis_config` property are
now a `"hypothesisClient"` sub-dict within the dict that `js_config` returns.

The `"js-hypothesis-config"` object is no longer rendered into the HTML,
since the `"hypothesisClient"` config is now part of the
`"js-config"` that already gets rendered:
https://github.com/hypothesis/lms/blob/1f10d599e5e2502a4bfbde023da35e852729788e/lms/templates/base.html.jinja2#L36-L39

Finally, the JSON-RPC server itself (`methods.js`) is changed to read
the client's config from `js-config["hypothesisClient"]` rather than
from the old `js-hypothesis-config`.

And a couple of tests need updating.

How to test this
----------------

Launch any assignment in Canvas and test that you're logged into your
Canvas account in the Hypothesis client. If the JSON-RPC server's config
wasn't working, you wouldn't be.

### DIFF
--- a/lms/resources/_lti_launch.py
+++ b/lms/resources/_lti_launch.py
@@ -192,7 +192,7 @@ class LTILaunchResource:
                     ],
                 },
                 # The config object for the Hypothesis client server.
-                "hypothesisClient": self._hypothesis_config,
+                "hypothesisClient": self._get_hypothesis_config(),
                 # URLs for the frontend to use (e.g. API endpoints for it to call).
                 "urls": {},
             }
@@ -204,8 +204,7 @@ class LTILaunchResource:
 
         return self._js_config
 
-    @property
-    def _hypothesis_config(self):
+    def _get_hypothesis_config(self):
         """
         Return the Hypothesis client config object for the current request.
 
--- a/lms/resources/_lti_launch.py
+++ b/lms/resources/_lti_launch.py
@@ -32,7 +32,6 @@ class LTILaunchResource:
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
         self._ai_getter = self._request.find_service(name="ai_getter")
-        self._hypothesis_config = None
         self._js_config = None
 
     @property
@@ -192,6 +191,8 @@ class LTILaunchResource:
                         "rpc_allowed_origins"
                     ],
                 },
+                # The config object for the Hypothesis client server.
+                "hypothesisClient": self._hypothesis_config,
                 # URLs for the frontend to use (e.g. API endpoints for it to call).
                 "urls": {},
             }
@@ -204,7 +205,7 @@ class LTILaunchResource:
         return self._js_config
 
     @property
-    def hypothesis_config(self):
+    def _hypothesis_config(self):
         """
         Return the Hypothesis client config object for the current request.
 
@@ -213,11 +214,6 @@ class LTILaunchResource:
         request, such as an authorization grant token for the client to use to
         log in to the Hypothesis account corresponding to the LTI user that the
         request comes from.
-
-        This is a mutable config dict. It can be accessed, for example by
-        views, as ``request.context.hypothesis_config``, and they can mutate it or add
-        their own view-specific config settings. The modified config object
-        will then be passed to the Hypothesis client.
 
         See: https://h.readthedocs.io/projects/client/en/latest/publishers/config/#configuring-the-client-using-json
 
@@ -241,20 +237,17 @@ class LTILaunchResource:
             }
             return jwt.encode(claims, client_secret, algorithm="HS256")
 
-        if self._hypothesis_config is None:
-            self._hypothesis_config = {
-                "services": [
-                    {
-                        "apiUrl": api_url,
-                        "authority": self._authority,
-                        "enableShareLinks": False,
-                        "grantToken": grant_token().decode("utf-8"),
-                        "groups": [self.h_groupid],
-                    }
-                ]
-            }
-
-        return self._hypothesis_config
+        return {
+            "services": [
+                {
+                    "apiUrl": api_url,
+                    "authority": self._authority,
+                    "enableShareLinks": False,
+                    "grantToken": grant_token().decode("utf-8"),
+                    "groups": [self.h_groupid],
+                }
+            ]
+        }
 
     @property
     def provisioning_enabled(self):
--- a/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
@@ -51,10 +51,10 @@ describe('BasicLtiLaunchApp', () => {
       },
     });
 
-    // fake js-hypothesis-config
+    // fake js-config
     fakeHypothesisConfig = sinon.stub(document, 'querySelector');
     fakeHypothesisConfig
-      .withArgs('.js-hypothesis-config')
+      .withArgs('.js-config')
       .returns({ text: JSON.stringify({}) });
   });
 
--- a/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
@@ -6,8 +6,10 @@ describe('postmessage_json_rpc/methods#requestConfig', () => {
   beforeEach('inject the client config into the document', () => {
     configEl = document.createElement('script');
     configEl.setAttribute('type', 'application/json');
-    configEl.classList.add('js-hypothesis-config');
-    configEl.textContent = JSON.stringify({ foo: 'bar' });
+    configEl.classList.add('js-config');
+    configEl.textContent = JSON.stringify({
+      hypothesisClient: { foo: 'bar' },
+    });
     document.body.appendChild(configEl);
   });
 
--- a/lms/static/scripts/postmessage_json_rpc/server/methods.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods.js
@@ -10,6 +10,7 @@
  * Return a Hypothesis client config object for the current LTI request.
  */
 export function requestConfig() {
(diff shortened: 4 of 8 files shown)


lms/resources/_lti_launch.py
```diff
@@ -32,7 +32,6 @@ def __init__(self, request):
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
         self._ai_getter = self._request.find_service(name="ai_getter")
-        self._hypothesis_config = None
```
seanh: This was here because `LTILaunchResource.hypothesis_config` was mutable, and views could mutate it. It's now a sub-dict of `LTILaunchResource.js_config` which is also mutatable by views, so this is no longer needed

lms/resources/_lti_launch.py
```diff
@@ -183,6 +182,45 @@ def js_config(self):
 
         :rtype: dict
         """
+
+        def hypothesis_client_config():
```
seanh: I left this as an internal function, within the `LTILaunchResource.js_config()` method. Later, when all this config stuff moves into a separate service, we'll probably want to move each block of config into a separate method (perhaps private helper methods) and have one master method that stitches them all together. Just to avoid having one really big method. But for now this will do.

jon-betts: As far as I can tell this function doesn't form a closure around anything and so doesn't _need_ to be inline. What's the motivation for an inline function, rather than a private function?

jon-betts: As far as I can tell this function doesn't form a closure around anything and so doesn't need to be inline. What's the motivation for an inline function, rather than a private function?

lms/views/basic_lti_launch.py
```diff
@@ -362,6 +362,6 @@ def set_canvas_focused_user(self):
             display_name = "(Couldn't fetch student name)"
 
         # TODO! - Could/should this be replaced with a GradingInfo lookup?
-        self.context.hypothesis_config.update(
+        self.context.js_config["hypothesisClient"].update(
```
jon-betts: Reaching inside this data structure feels like a signal this should be wrapped up in the service when that comes along.

tests/unit/lms/resources/_lti_launch_test.py
```diff
@@ -343,18 +343,16 @@ def test_views_can_mutate_js_config(self, lti_launch):
...
 
         with pytest.raises(
             HTTPBadRequest,
             match='Required parameter "oauth_consumer_key" missing from LTI params',
         ):
-            LTILaunchResource(pyramid_request).hypothesis_config
+            LTILaunchResource(pyramid_request).js_config
 
-    def test_hypothesis_config_raises_if_theres_no_tool_consumer_instance_guid(
+    def test_js_config_raises_if_theres_no_tool_consumer_instance_guid(
```
jon-betts: Shorter is betterer! :+1:

tests/unit/lms/resources/_lti_launch_test.py
```diff
@@ -366,29 +364,35 @@ def test_hypothesis_config_raises_if_theres_no_tool_consumer_instance_guid(
             HTTPBadRequest,
             match='Required parameter "tool_consumer_instance_guid" missing from LTI params',
         ):
-            LTILaunchResource(pyramid_request).hypothesis_config
+            LTILaunchResource(pyramid_request).js_config
 
-    def test_hypothesis_config_contains_one_service_config(self, lti_launch):
-        assert len(lti_launch.hypothesis_config["services"]) == 1
+    def test_js_config_contains_one_service_config_for_the_client(self, lti_launch):
+        assert len(lti_launch.js_config["hypothesisClient"]["services"]) == 1
```
jon-betts: Are we getting to the point here were a separate test class to hold "hypothesisConfig" related tests might be nice? 

* It allows you to collapse some if you aren't interested in them 
* Facilitates shorter names for everything (In "TestHypothesisConfig" we can use "test_it_blah")
* Could then some helper methods, or a 'client_config' fixture etc?
* Stages us nicely for the future if these get plucked out of the class to a service

lms/resources/_lti_launch.py
```diff
@@ -203,59 +243,6 @@ def js_config(self):
 
         return self._js_config
 
-    @property
-    def hypothesis_config(self):
```
jon-betts: As per the comment above, if this was mostly just changed to a function `_hypothesis_config()` I think it would be easier to read:

 * This diff (not trivial to see if this function has changed)
 * The function above (as you don't have to read lots of function it might not even call)

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -51,10 +51,10 @@ describe('BasicLtiLaunchApp', () => {
       },
     });
 
-    // fake js-hypothesis-config
+    // fake js-lms-config
     fakeHypothesisConfig = sinon.stub(document, 'querySelector');
     fakeHypothesisConfig
-      .withArgs('.js-hypothesis-config')
+      .withArgs('.js-lms-config')
```
jon-betts: One day maybe just 'lms-config'? Although, I do love a bit of titular redundancy

lms/views/basic_lti_launch.py
seanh: Yep. I think as we discussed, the aim will be for nothing to be mutating the config directly other than the service itself. And then views or whatever will call the service to have the config mutated for them

tests/unit/lms/resources/_lti_launch_test.py
seanh: I'm probably gonna skip this one for now because almost the very next thing to do is cut all this code and its related tests, and move it all into a service. If the tests are still too long at that point then I might split them into separate test classes

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
seanh: Yep. Next thing to do is to rename this to `js-config` to match with the name of the `LTILaunchResource.js_config` property. Note that the `js-` prefix is a pattern used to denote CSS classes that exist for the purpose of enabling JavaScript to find things rather than for styling (see <https://github.com/hypothesis/frontend-toolkit/blob/master/docs/css-style-guide.md#classes-used-in-code>) so that won't be changing

lms/resources/_lti_launch.py
seanh: Yep. Sometimes if a function only exists to be a private helper of one other function it can be readable to make it an inner function just for that reason, even if no closure is used. In retrospect though this function is a bit long to be an inner function. Moved it back out and made it a private method instead

lms/resources/_lti_launch.py
```diff
@@ -214,11 +215,6 @@ def hypothesis_config(self):
         log in to the Hypothesis account corresponding to the LTI user that the
         request comes from.
 
-        This is a mutable config dict. It can be accessed, for example by
-        views, as ``request.context.hypothesis_config``, and they can mutate it or add
-        their own view-specific config settings. The modified config object
-        will then be passed to the Hypothesis client.
```
seanh: Removed this chunk of docstring as it's no longer true

seanh:
@jon-betts Ready for re-review

lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -10,6 +10,7 @@
  * Return a Hypothesis client config object for the current LTI request.
  */
 export function requestConfig() {
-  const configEl = document.querySelector('.js-hypothesis-config');
-  return JSON.parse(configEl.textContent);
+  const configEl = document.querySelector('.js-lms-config');
```
LMS007: I'm assuming once you get the other PR merged this is `.js-config`

lms/static/scripts/postmessage_json_rpc/server/methods-test.js
```diff
@@ -6,8 +6,10 @@ describe('postmessage_json_rpc/methods#requestConfig', () => {
   beforeEach('inject the client config into the document', () => {
     configEl = document.createElement('script');
     configEl.setAttribute('type', 'application/json');
-    configEl.classList.add('js-hypothesis-config');
-    configEl.textContent = JSON.stringify({ foo: 'bar' });
+    configEl.classList.add('js-lms-config');
```
LMS007: 'js-config'

seanh:
Rebased after merging https://github.com/hypothesis/lms/pull/1457

PR #1458, opened 2020-02-12
Use separate DBs for tests, functests and bddtests

seanh:
Rather than all sharing a single lms-test DB have the tests, functests and
bddtests use separate lms-test, lms-functests and lms-bddtests DBs.

<del>I was noticing a problem that, when the three different tests are run together
in a single `make` command, something seemed to be sometimes leaving test data
behind in the DB that would cause later test runs to fail until the DB was
manually cleared. Was happening to me locally quite often. I didn't figure out
why. But this change seems to get rid of the problem.</del> -- this doesn't seem to fix that, I think the db_session fixture in the functests may be wrong

This also means that the tests, functests and bddtests can be run in parallel,
for example:

```terminal
$ make -j `nproc` sure
```

This takes the time to run `make sure` on my machine down from 1 min 30s to 42s.

### DIFF
--- a/bin/create-db
+++ b/bin/create-db
@@ -4,5 +4,5 @@
 #
 # Usage:
 #
-#     create-testdb <DB_NAME>
+#     create-db <DB_NAME>
 make services args="exec postgres psql -U postgres -c 'CREATE DATABASE $1;'" > /dev/null 2>&1 || true
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,22 +8,14 @@ from sqlalchemy.orm import sessionmaker
 
 from lms import db
 
-__all__ = [
-    "patch",
-    "db_engine",
-    "SESSION",
-    "TEST_DATABASE_URL",
-    "TEST_SETTINGS",
-]
-
 SESSION = sessionmaker()
-TEST_DATABASE_URL = os.environ.get(
-    "TEST_DATABASE_URL", "postgresql://postgres@localhost:5433/lms_test"
-)
 
-# Settings that will end up in pyramid_request.registry.settings.
+
+def get_test_database_url(default):
+    return os.environ.get("TEST_DATABASE_URL", default)
+
+
 TEST_SETTINGS = {
-    "sqlalchemy.url": TEST_DATABASE_URL,
     "via_url": "http://TEST_VIA_SERVER.is/",
     "via3_url": "http://TEST_VIA3_SERVER.is/",
     "jwt_secret": "test_secret",
@@ -48,12 +40,13 @@ TEST_SETTINGS = {
     "h_api_url_private": "https://example.com/private/api/",
     "rpc_allowed_origins": ["http://localhost:5000"],
     "oauth2_state_secret": "test_oauth2_state_secret",
+    "session_cookie_secret": "notasecret",
 }
 
 
 @pytest.fixture(scope="session")
 def db_engine():
-    engine = sqlalchemy.create_engine(TEST_DATABASE_URL)
+    engine = sqlalchemy.create_engine(TEST_SETTINGS["sqlalchemy.url"])
     db.init(engine)
     return engine
 
--- a/tox.ini
+++ b/tox.ini
@@ -70,9 +70,9 @@ whitelist_externals =
     tests,functests,bddtests: sh
 commands =
     dev: {posargs:pserve conf/development.ini --reload}
-    tests: sh bin/create-testdb lms_test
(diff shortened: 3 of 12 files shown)


bin/create-bddtestsdb
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+#
+# Create the "lms_bddtests" database in Postgres, if it doesn't exist already.
+make services args='exec postgres psql -U postgres -c "CREATE DATABASE lms_bddtests;"' > /dev/null 2>&1 || true
```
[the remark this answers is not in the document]
seanh: Done

seanh:
Fixed Jenkins.

Also changed `make test` to run `coverage report` and removed the separate `make coverage` command as done in Via 3: you can't run `make test` and `make coverage` in parallel since coverage depends on tests.

I think <code>make -j `nproc` sure</code> does actually make it clear enough when something (e.g. coverage) fails: the `make` command exits with status 2 and it prints the error from what failed at the end. For example here's it with coverage hacked to fail:

```terminal
$ make -j `nproc` sure
... (lots of output)
ERROR: InvocationError for command /home/seanh/Projects/lms/.tox/tests/bin/coverage report (exited with code 2)
2! $ 
```

It prints the output from the failed `coverage report` (the table of coverage percentages) right before the error message too.

One weird thing I'm seeing is that if the test DBs don't exist yet and you run <code>make -j `nproc` sure</code> the tests fail because the DBs don't exist, even though they should have been created by tox running `create-db`. If you do the same thing but without the <code>-j `nproc`</code> then it does work. I'm not sure why that's happening. But <code>-j `nproc`</code> does work fine as long as the DBs already exist.

I'm still seeing a problem that some of the tests, I think the functests, are sometimes leaving the DB unclean and causing future test runs to fail. This doesn't seem to have fixed that. I think the DB session fixture in the functests may be wrong.

seanh:
This separate PR to quieten the test and functest output will help with `make sure`'s gargantuan output: https://github.com/hypothesis/lms/pull/1460 There doesn't seem to be any way to stop behave from printing out screens and screens though

tests/conftest.py
```diff
@@ -17,9 +17,7 @@
 ]
 
 SESSION = sessionmaker()
-TEST_DATABASE_URL = os.environ.get(
-    "TEST_DATABASE_URL", "postgresql://postgres@localhost:5433/lms_test"
-)
+TEST_DATABASE_URL = os.environ["TEST_DATABASE_URL"]
```
jon-betts: I get why this is done, but this is going to make it a collosal pain for me to run any tests as they no longer work without specifying the env var.

This is fine if you run the tests through "make test" at all times, but if you use an IDE like me, or you run the tests using pytest on the command line this is very unhelpful.

Having a different defaulting `TEST_DATABASE_URL` in the `functional/conftest.py`, `unit/conftest.py` and `bdd/environment.py` is easy enough, but we need to stop directly reading this variable into `TEST_SETTINGS`.

A pytest fixture won't work here as the BDD stuff isn't pytest based. I think we can do this by changing this to `get_test_settings(db_url)` or something similar. It's only used in three places from what I can see, so that shouldn't be too hard?

I've also spotted that two of the three places is adding "session_cookie_secret" to the settings, so maybe we should chuck them in the main config and be done with it.

bin/create-db
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+#
+# Create the given database in Postgres, if it doesn't exist already.
+#
+# Usage:
+#
+#     create-db <DB_NAME>
+make services args="exec postgres psql -U postgres -c 'CREATE DATABASE $1;'" > /dev/null 2>&1 || true
```
jon-betts: Much better I think :+1:

tox.ini
```diff
@@ -62,13 +62,20 @@ deps =
 setenv =
```
jon-betts: `coverage` is still present in the dependencies section.

seanh:
>  Is there a way to make make run in parallel by default? For example can make call make?

I don't know of a way to configure it to use parallel by default but make calling make might work. I think it's probably best not to bother though because, as you've noted, parallel can be confusing. I think it might still be useful but happy to let the user have to opt-in to it

seanh:
> It's still kind of a nightmare to read. You'd basically have to run the thing that failed again to see what actually happened, but this is still valuable, as we have a potentially quicker way of checking.

Right, it's only really good for a yes/no answer and then if the answer is no you'll likely want to use a less confusing command to get to the bottom of why

PR #1459, opened 2020-02-12
Quieten pytest's output in `make test`

seanh:
When the tests pass less output is nicer. Do the same for `make functests` also.


seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/1460

PR #1462, opened 2020-02-12
Use legacy Via in dev by default

seanh:
Change the dev env back to legacy Via instead of Via 3 by default.

You can still visit http://localhost:8001/flags and turn on the
use_via3_url feature flag to have it use Via 3 in dev instead.

When it's time to change the dev env backt o Via 3 by default again we
can simply revert this commit.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -15,13 +15,12 @@ There are three presentations for developers that describe what the Hypothesis L
 
 ### You will need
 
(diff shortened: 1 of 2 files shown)


seanh:
The feature flags are extremely easy to use, you just go to http://localhost:8001/flags and check or uncheck a checkbox

Issue #1463, opened 2020-02-13
LMS frontend tests are printing ~100 lines of errors and warnings


seanh:
```
$ make frontend-tests
[11:29:57] Using gulpfile ~/Projects/lms/gulpfile.js
[11:29:57] Starting 'test'...
[11:29:57] Starting 'build-css'...
[11:29:57] Finished 'build-css' after 234 ms
[11:29:57] Starting '<anonymous>'...

START:
[32m13 02 2020 11:30:05.285:INFO [framework.browserify]: [39mbundle built
[32m13 02 2020 11:30:05.301:INFO [karma-server]: [39mKarma v4.4.1 server started at http://0.0.0.0:9876/
[32m13 02 2020 11:30:05.301:INFO [launcher]: [39mLaunching browsers ChromeHeadless_Custom with concurrency unlimited
[32m13 02 2020 11:30:05.304:INFO [launcher]: [39mStarting browser ChromeHeadless
[32m13 02 2020 11:30:05.578:INFO [HeadlessChrome 79.0.3945 (Linux 0.0.0)]: [39mConnected on socket FiWghkxElTQSOOmQAAAA with id 65862597
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog />`, expected an array.'
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog />`, expected an array.'
ERROR: 'Failed Dialog type: Invalid Dialog `buttons[0]` of type `object` supplied to `<Dialog buttons="[object Object],[object Object]" />`, expected a single ReactElement.'
ERROR: 'Failed Dialog type: Invalid Dialog `buttons[0]` of type `object` supplied to `<Dialog buttons="[object Object],[object Object]" />`, expected a single ReactElement.'
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog initialFocus="[object Object]" />`, expected an array.'
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog initialFocus="[object Object]" />`, expected an array.'
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog title="Test dialog" />`, expected an array.'
ERROR: 'Failed Dialog type: Invalid Dialog `children` of type `object` supplied to `<Dialog title="Test dialog" />`, expected an array.'
[...]

PR #1466, opened 2020-02-14
Fix most prop types warnings from LMS frontend tests

robertknight:
This PR fixes most of the warnings mentioned in https://github.com/hypothesis/lms/issues/1463.

The issue was that runtime validation of component `propTypes` schemas was not actually working until the recent Preact v10.3 upgrade. When that upgrade happened, it ended up surfacing existing issues.

This PR fixes all of the easier propTypes issues. See individual commit messages for details. There will be another PR coming that resolves issues with props that use `propTypes.element` or `propTypes.node` in their schema


seanh:
Works for me, seems to get rid of a lot of the warnings

PR #1467, opened 2020-02-14
Patch `propTypes.{element, node}` to work with Preact elements

robertknight:
This is the second part of https://github.com/hypothesis/lms/issues/1463. It fixes the remaining propTypes-related warnings when running lms frontend tests.

Several components make use of `propTypes.{element, node}` to validate that a prop contains a JSX element or a value that can appear in a JSX expression (string, boolean, null).

This fails validation checks because `propTypes.{element, node}` only work with React elements but not Preact ones. The Preact docs don't specify a preferred way to resolve this so I'm asking the community.

In the meantime, this resolves the related warnings by adding two custom prop type checkers which do the same thing and patches `propTypes.{element, node}` to use them.

---

- [x] Check with Preact devs / community whether there is a better solution (_Not that I found so far. That might change later_)
- [x] Add tests for our custom checkers, if we still end up using them

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/prop-types.js
+++ b/lms/static/scripts/frontend_apps/utils/prop-types.js
@@ -1,0 +1,93 @@
+import { isValidElement } from 'preact';
+
+/**
+ * Function that checks the prop `name` in `props` has an expected type or
+ * throws otherwise.
+ *
+ * @typedef {(props: Object, name: string, component: string) => any} CheckPropFunction
+ */
+
+/**
+ * Take a prop type checker function and return a checker where the prop is
+ * optional. The optional checker has an `isRequired` property that requires the
+ * prop.
+ *
+ * This matches how the standard prop types work (eg. `propTypes.object` is
+ * optional, `propTypes.object.isRequired` is required).
+ *
+ * @param {CheckPropFunction} checkPropType
+ * @return {CheckPropFunction}
+ */
+function checkOptionalProp(checkPropType) {
+  const check = (props, propName, componentName) => {
+    if (!(propName in props)) {
+      return null;
+    }
+    return checkPropType(props, propName, componentName);
+  };
+  check.isRequired = checkPropType;
+  return check;
+}
+
+/**
+ * Check that a prop is a Preact element (ie. the result of a call to `createElement`
+ * or a JSX element (`<div .../>`, `<Widget .../>`).
+ *
+ * This is the same as `propTypes.element`, but for Preact elements.
+ *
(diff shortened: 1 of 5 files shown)


robertknight:
On second thoughts, the monkey-patching might not be a good idea. It means that the custom implementations need to match every detail of the originals to avoid being super-confusing. It probably would be better for components just to import these instead of the standard ones if they need them. The code could be modified easily enough if the components were ever to be used with React for any reason.

lms/static/scripts/frontend_apps/utils/prop-types.js
```diff
@@ -0,0 +1,75 @@
+import { isValidElement } from 'preact';
```
LMS007: I'm wondering if this file be in frontend-toolkit repo?

robertknight: Yes, I think it would make sense to move it to that repo in future. I'm putting it here to start with to validate it in the context of a concrete use case.

robertknight:
After looking around and not finding a better solution I have updated this PR with tests. Longer term I think this is best solved upstream (eg. in `preact/debug`). This approach solves the problem for the moment. If the same issue crops up in the client repo, which it probably will at some point, it would be a good idea to extract this into the frontend-toolkit repo.

lms/static/scripts/frontend_apps/utils/prop-types.js
LMS007: roger that

lms/static/scripts/frontend_apps/utils/prop-types.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  return new Error(
+    `Expected prop "${propName}" supplied to ${componentName} to be a renderable value (string, boolean, element etc.)`
+  );
+}
+
+export const node = checkOptionalProp(checkNodeProp);
+
+/**
+ * Patch `propTypes.{element, node}` to work with Preact (rather than React)
```
LMS007: is node vs element interchangeable I'm not clear what the difference actually is?

robertknight: - `element` - The return value from a call to `createElement(...)`, ie. a JSX `<...>` expression (eg. `<div/>`, `<MyComponent/>`)
- `node` - Any value which can appear in the middle of a JSX expression. This includes elements but also strings, booleans and `null` (eg. `<div>Foo bar {someString} {someBool && }</div>`)

Does that clarify things?

PR #1470, opened 2020-02-15
Move JavaScript config into a service

seanh:
Having written this I'm beginning to doubt that JS config really belongs in a service. The problems with the base template (described below) really imply that this belongs in the context because if you put it in the context then everything just works.

It could be split out from the main `LTILaunchResource` class into a separate `JSConfig` class, but still kept in the context resources rather than in the services. `LTILaunchResource` would just delegate the `js_config` property to a collaborator class:

```python
class LTILaunchResource:

    ...

    @property
    @functools.lru_cache()
    def js_config(self):
        return JSConfig(self, self._request).config

    ...
```

Here's a much simpler PR that implements that approach: https://github.com/hypothesis/lms/pull/1471

Anyway, here's the notes from moving it into a service:

* * *

Move the JavaScript config dict from `LTILaunchResource.js_config` into
a new `JSConfig` service:
`request.find_service(name="js_config").config`.

This is pretty much just a straight copy-paste of of
`LTILaunchResource.js_config` over to `JSConfig.config`, with some minor
changes.

`JSConfig` service only available during LTI launch requests
------------------------------------------------------------

The new `JSConfig` service is only available during LTI launch requests
(requests where the context is `LTILaunchResource`).

This is achieved by using the `context=LTILaunchResource` argument to
pyramid_service's `register_service_factory()`. For requests where the
context it not `LTILaunchResource`,
`request.find_service(name="js_config")` will fail.

This is necessary because `JSConfig` assumes that the given `context` is
an `LTILaunchResource`. It reads various `LTILaunchResource` properties
so it will crash otherwise.

`request.js_config`
-------------------

The `request.find_service(name="js_config").config` dict is also
accessible as `request.js_config` because this was the only way I could
get the templates to work:

On master the `base.html.jinja2` template, which is the base template
for all pages, does this:

```jinja2
{% if context.js_config is defined %}
  <script type="application/json" class="js-config">{{ context.js_config|tojson }}</script>
{% endif %}
```

This works because:

1. If `context` is an `LTILaunchResource` then `context.js_config` is
   defined

2. If `context` is `DefaultResource` (which is the case for all the
   non-LTI launch pages) then `context.js_config` is not defined and it
   won't attempt to render the JS config

3. If `context` is an exception (which is the case for the error and
   validation error pages) then `context.js_config` is not defined

When the JS config has been moved into a service this
`{% if context.js_config is defined %}` no longer works. The template
can do `{{ request.find_service(name="js_config").config|tojson }}` but
this creates two crashes:

1. For non-`LTILaunchResource` pages the `find_service()' lookup fails
   and crashes.

2. For error and validation error pages, if the original request that
   failed and triggered error view rendering was an LTI launch request,
   then the `find_service()` will succeed but calling `JSConfig.config`
   might crash.

   For example maybe the request errorred because one of the LTI launch
   params needed for rendering the JS config, such as the `"user_id"`
   param, was missing.

   Since the app is now trying to render the JS config _again_ as part
   of the error page, rendering the error page also crashes!

   (One confusing thing about this case is that `context` in the error
   view template is the exception that was raised, but if that template
   then calls the `JSConfig` service the `JSConfig` service's
   `self._context` is still the original `LTILaunchResource`.)

To get around this I had `JSConfig` add a `request.js_config` so that
the base template can use an `is defined` check to handle both cases.

The alternative would be to refactor the templates so that error templates
and non-LTI launch templates don't contain the JS config in the first
place.

Code design improvements
------------------------

The new `JSConfig` class is slightly nicer than the old code because it
splits the four sub-parts of the initial JS config (authToken,
hypothesisClient, rpcServer and urls) into four separate private helper
properties. The tests are also separated into four separate test
classes, which makes the test code a bit nicer.

Future
------

For future commits the plan is to find places where the
`JSConfig.config` dict is being mutated and centralise those in one
place in the `JSConfig` service itself and remove its mutable dict
feature.


seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/1471/

PR #1471, opened 2020-02-15
Move JS config into collaborator

seanh:
This just moves `LTILaunchResource.js_config` into a separate helper class that `LTILaunchResource` calls, without changing the public interface yet. User code still access the config dict directly as `context.js_config` (singleton) for now.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -1,0 +1,100 @@
+import datetime
+import urllib
+
+import jwt
+
+from lms.validation.authentication import BearerTokenSchema
+
+
+class JSConfig:
+    """The config for the app's JavaScript code."""
+
+    def __init__(self, context, request):
+        self._context = context
+        self._request = request
+
+    @property
+    def config(self):
+        """
+        Return the configuration for the app's JavaScript code.
+
+        This is a mutable config dict. It can be accessed, for example by
+        views, and they can mutate it or add their own view-specific config
+        settings. The modified config object will then be passed to the
+        JavaScript code in the response page.
+
+        :rtype: dict
+        """
+        return {
+            "authToken": self._auth_token,
+            "hypothesisClient": self._hypothesis_config,
+            "rpcServer": self._rpc_server_config,
+            "urls": self._urls,
+        }
+
+    @property
+    def _auth_token(self):
+        """Return the authToken setting."""
+        if not self._request.lti_user:
+            return None
+
+        return BearerTokenSchema(self._request).authorization_param(
+            self._request.lti_user
+        )
+
+    @property
+    def _hypothesis_config(self):
+        """
+        Return the Hypothesis client config object for the current request.
+
+        See: https://h.readthedocs.io/projects/client/en/latest/publishers/config/#configuring-the-client-using-json
+
+        """
+        if not self._context.provisioning_enabled:
+            return {}
+
+        client_id = self._request.registry.settings["h_jwt_client_id"]
+        client_secret = self._request.registry.settings["h_jwt_client_secret"]
+        api_url = self._request.registry.settings["h_api_url_public"]
+        audience = urllib.parse.urlparse(api_url).hostname
+
+        def grant_token():
+            now = datetime.datetime.utcnow()
+            claims = {
+                "aud": audience,
+                "iss": client_id,
+                "sub": self._context.h_user.userid,
+                "nbf": now,
+                "exp": now + datetime.timedelta(minutes=5),
+            }
+            return jwt.encode(claims, client_secret, algorithm="HS256")
+
+        return {
+            "services": [
+                {
+                    "apiUrl": api_url,
+                    "authority": self._request.registry.settings["h_authority"],
+                    "enableShareLinks": False,
+                    "grantToken": grant_token().decode("utf-8"),
+                    "groups": [self._context.h_groupid],
+                }
+            ]
+        }
+
+    @property
+    def _rpc_server_config(self):
+        """Return the config for the postMessage-JSON-RPC server."""
+        return {
+            "allowedOrigins": self._request.registry.settings["rpc_allowed_origins"],
+        }
+
+    @property
+    def _urls(self):
+        """
+        Return a dict of URLs for the frontend to use.
+
+        For example: API endpoints for the frontend to call would go in
+        here.
+
+        """
+        return {}
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -1,5 +1,5 @@
 import datetime
-import urllib
+from urllib.parse import urlparse
 
 import jwt
 
@@ -53,21 +53,7 @@ class JSConfig:
         if not self._context.provisioning_enabled:
             return {}
 
-        client_id = self._request.registry.settings["h_jwt_client_id"]
-        client_secret = self._request.registry.settings["h_jwt_client_secret"]
         api_url = self._request.registry.settings["h_api_url_public"]
-        audience = urllib.parse.urlparse(api_url).hostname
-
-        def grant_token():
-            now = datetime.datetime.utcnow()
-            claims = {
-                "aud": audience,
-                "iss": client_id,
-                "sub": self._context.h_user.userid,
-                "nbf": now,
-                "exp": now + datetime.timedelta(minutes=5),
-            }
-            return jwt.encode(claims, client_secret, algorithm="HS256")
 
         return {
             "services": [
@@ -75,11 +61,29 @@ class JSConfig:
                     "apiUrl": api_url,
                     "authority": self._request.registry.settings["h_authority"],
                     "enableShareLinks": False,
-                    "grantToken": grant_token().decode("utf-8"),
+                    "grantToken": self._grant_token(api_url),
                     "groups": [self._context.h_groupid],
                 }
             ]
         }
+
+    def _grant_token(self, api_url):
+        """Return an OAuth 2 the client can use to log in to h."""
+        now = datetime.datetime.utcnow()
+
+        claims = {
+            "aud": urlparse(api_url).hostname,
+            "iss": self._request.registry.settings["h_jwt_client_id"],
+            "sub": self._context.h_user.userid,
+            "nbf": now,
+            "exp": now + datetime.timedelta(minutes=5),
+        }
+
+        return jwt.encode(
+            claims,
+            self._request.registry.settings["h_jwt_client_secret"],
+            algorithm="HS256",
+        ).decode("utf-8")
 
     @property
     def _rpc_server_config(self):
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -1,13 +1,11 @@
 """Traversal resources for LTI launch views."""
-import datetime
+import functools
 import hashlib
-import urllib
 
-import jwt
 from pyramid.httpexceptions import HTTPBadRequest
 from pyramid.security import Allow
 
-from lms.validation.authentication import BearerTokenSchema
+from lms.resources._js_config import JSConfig
 from lms.values import HUser
 
 
@@ -30,7 +28,6 @@ class LTILaunchResource:
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
         self._ai_getter = self._request.find_service(name="ai_getter")
-        self._js_config = None
 
     @property
     def h_user(self):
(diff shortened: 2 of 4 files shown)


lms/resources/lti_launch.py
```diff
@@ -162,82 +159,9 @@ def h_provider_unique_id(self):
...
-                {
-                    "apiUrl": api_url,
-                    "authority": self._authority,
-                    "enableShareLinks": False,
-                    "grantToken": grant_token().decode("utf-8"),
-                    "groups": [self.h_groupid],
-                }
-            ]
-        }
+        return JSConfig(self, self._request).config
```
seanh: This is going to have to change (to just `return JSConfig(self, self._request)`) if user code is going to be calling `JSConfig` methods in future, but I'll leave that until I get to it

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+        self, context, context_property, pyramid_request
+    ):
+        # Make reading context.<context_property> raise HTTPBadRequest.
+        setattr(
+            type(context),
+            context_property,
+            mock.PropertyMock(side_effect=HTTPBadRequest("example error message")),
+        )
+
+        with pytest.raises(HTTPBadRequest, match="example error message"):
```
jon-betts: Generally I think we want to steer clear of checking error text. If `HTTPBadRequest` is something we're worried about the code raising for different reasons we should probably use an error we don't think it can. Maybe `OSError`, or `SyntaxError` or something?

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+@pytest.fixture(autouse=True)
+def BearerTokenSchema(patch):
+    return patch("lms.resources._js_config.BearerTokenSchema")
+
+
+@pytest.fixture
+def bearer_token_schema(BearerTokenSchema):
+    return BearerTokenSchema.return_value
```
jon-betts: Both of the bearer token fixtures here are only needed in `TestJSConfigAuthToken`. Moving them there would make that class easier to understand and prevent the large number of ambient fixtures we often accumulate at the bottom of test classes.

I've seen large collections of fixtures at the bottom of the file make it a pain to refactor when you need to split up tests.

lms/resources/_js_config.py
```diff
@@ -0,0 +1,100 @@
+import datetime
+import urllib
+
+import jwt
+
+from lms.validation.authentication import BearerTokenSchema
+
+
+class JSConfig:
```
jon-betts: Is this `JSConfig`, or is it `LTILaunchJSConfig` etc.?

lms/resources/lti_launch.py
jon-betts: Goodbye code! This class is much easier to grok now.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+        self, authToken, bearer_token_schema, BearerTokenSchema, pyramid_request
+    ):
+        BearerTokenSchema.assert_called_once_with(pyramid_request)
+        bearer_token_schema.authorization_param.assert_called_once_with(
+            pyramid_request.lti_user
+        )
+        assert authToken == bearer_token_schema.authorization_param.return_value
+
+    def test_it_is_None_for_non_lti_users(
+        self, authToken, BearerTokenSchema, context, pyramid_request
```
jon-betts: We don't use the `authToken` fixture here

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+    ):
+        pyramid_request.lti_user = None
+
+        assert JSConfig(context, pyramid_request).config["authToken"] is None
+
+    @pytest.fixture
+    def authToken(self, config):
+        return config["authToken"]
```
jon-betts: I think this might be better inline. We only use it once, and it adds a layer of indirection about what the test is asserting.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+            key="TEST_JWT_CLIENT_SECRET",
+            audience="example.com",
+        )
+        after = int(datetime.datetime.now().timestamp())
+        assert claims["iss"] == "TEST_JWT_CLIENT_ID"
+        assert claims["sub"] == "acct:example_username@TEST_AUTHORITY"
+        assert before <= claims["nbf"] <= after
+        assert claims["exp"] > before
+
+    def test_it_includes_the_group(self, pyramid_request, config):
```
jon-betts: `pyramid_request` is not used here

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+        assert JSConfig(context, pyramid_request).config["hypothesisClient"] == {}
+
+    def test_it_is_mutable(self, config):
+        config.update({"a_key": "a_value"})
+
+        assert config["a_key"] == "a_value"
+
+    @pytest.fixture
+    def config(self, config):
+        return config["hypothesisClient"]
```
jon-betts: This fixture makes sense to be as it cuts down the repetition a lot.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+class TestJSConfigRPCServer:
+    """Unit tests for the "rpcServer" sub-dict of JSConfig."""
+
+    def test_it(self, config):
+        assert config == {"allowedOrigins": ["http://localhost:5000"]}
+
+    @pytest.fixture
+    def config(self, config):
+        return config["rpcServer"]
```
jon-betts: I'm not sure this fixture makes sense to me though. I think this is only included here because it was useful in the class above. I think this might be premature optimisation.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+class TestJSConfigURLs:
+    """Unit tests for the "urls" sub-dict of JSConfig."""
+
+    def test_it(self, config):
+        assert config == {}
+
+    @pytest.fixture
+    def config(self, config):
+        return config["urls"]
```
jon-betts: Similar to the situation above.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -303,130 +303,6 @@ def test_h_user_userid(self, pyramid_request):
...
-    ):
-        ai_getter.provisioning_enabled.return_value = False
-
-        assert lti_launch.js_config["hypothesisClient"] == {}
-
-    def test_views_can_mutate_js_config_hypothesisClient_config(self, lti_launch):
-        lti_launch.js_config["hypothesisClient"].update({"a_key": "a_value"})
-
-        assert lti_launch.js_config["hypothesisClient"]["a_key"] == "a_value"
-
```
jon-betts: Get outa here! :) The remaining tests are still pretty chunky, but you're in with a chance now

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -504,11 +388,6 @@ def pyramid_request(self, pyramid_request):
...
-@pytest.fixture(autouse=True)
-def BearerTokenSchema(patch):
-    return patch("lms.resources.lti_launch.BearerTokenSchema")
-
-
 @pytest.fixture
-def bearer_token_schema(BearerTokenSchema):
-    return BearerTokenSchema.return_value
+def JSConfig(patch):
+    return patch("lms.resources.lti_launch.JSConfig")
```
jon-betts: This is the only fixture not attached to the class. I'd tend towards localism if we can.

lms/resources/lti_launch.py
```diff
@@ -162,82 +159,9 @@ def h_provider_unique_id(self):
         return self._get_param("user_id")
 
     @property
+    @functools.lru_cache()
```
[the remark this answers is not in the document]
seanh: Sadly cached_property is a Python 3.8 thing and we're still on 3.6

lms/resources/_js_config.py
```diff
@@ -0,0 +1,100 @@
...
+
+        This is a mutable config dict. It can be accessed, for example by
+        views, and they can mutate it or add their own view-specific config
+        settings. The modified config object will then be passed to the
+        JavaScript code in the response page.
+
+        :rtype: dict
+        """
+        return {
+            "authToken": self._auth_token,
```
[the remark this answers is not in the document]
seanh: Yep, I've tested and it's fine as far as I can tell. Well spotted

lms/resources/_js_config.py
```diff
@@ -0,0 +1,100 @@
...
+    @property
+    def _urls(self):
+        """
+        Return a dict of URLs for the frontend to use.
+
+        For example: API endpoints for the frontend to call would go in
+        here.
+
+        """
+        return {}
```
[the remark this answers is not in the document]
seanh: It provides a convenient hook for the docstring explaining what `"urls"` is meant to be for, though :man_shrugging:

lms/resources/_js_config.py
```diff
@@ -0,0 +1,100 @@
...
+        See: https://h.readthedocs.io/projects/client/en/latest/publishers/config/#configuring-the-client-using-json
+
+        """
+        if not self._context.provisioning_enabled:
+            return {}
+
+        client_id = self._request.registry.settings["h_jwt_client_id"]
+        client_secret = self._request.registry.settings["h_jwt_client_secret"]
+        api_url = self._request.registry.settings["h_api_url_public"]
+        audience = urllib.parse.urlparse(api_url).hostname
```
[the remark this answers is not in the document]
seanh: Done

lms/resources/_js_config.py
```diff
@@ -0,0 +1,100 @@
...
+
+        def grant_token():
```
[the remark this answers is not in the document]
seanh: Moved it into a separate private function. Warning though: you're gonna be seeing more stuff like this in following PRs as I'm mostly just moving stuff around not rewriting it

tests/unit/lms/resources/_js_config_test.py
seanh: It might as well be consistent with the rest of the test classes in the file, and if we ever send more settings to the rpcServer and need to add more tests to this class we'll be starting off how we mean to continue

tests/unit/lms/resources/_js_config_test.py
seanh: I'm just gonna leave this hear as their are gonna be multiple tests in this class that all need `config['urls']` imminently

tests/unit/lms/resources/lti_launch_test.py
seanh: This is patching `lti_launch.py`'s module-level `from lms.resources._js_config import JSConfig`, so we want the fixture to be a module-level fixture so that it applies to all of `lti_launch_test.py`. Of course `lti_launch_test.py` only contains one test class so it doesn't actually make a difference in this case yet. But I've seen it happen too many times that a test module contains multiple test classes and some of them are unwittingly using the real version of a dependency instead of the patched one, leading to confusing test bugs, of both the failing test and the nobody-noticed-this-test-was-passing-when-it-should-have-been-failing varieties. Usually happens when someone comes along and adds a new test class. So as a general rule just always put import patch fixture at module level.

It should also have `autouse=True` though, I'll fix that

lms/resources/_js_config.py
seanh: I think just `JSConfig` for now as it's the one and only JavaScript config object (which happens only to be used in LTI launches currently). If we ever need JS config for non-LTI launch requests, I don't know yet whether that would be handled by this class or a completely different class

tests/unit/lms/resources/_js_config_test.py
seanh: Just following my pattern again of always making import patches module-level fixtures. I'm gonna stick to that one as I've seen it go confusingly wrong too many times otherwise. In all other cases I'll put fixtures locally in test classes unless they're actually used by more than one class. The `bearer_token_schema` one could move but it's very related to `BearerTokenSchema`

lms/resources/lti_launch.py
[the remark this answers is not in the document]
jon-betts: Gaddam!

lms/resources/_js_config.py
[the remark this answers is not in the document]
jon-betts: Yeah I suppose so. Otherwise it's just "have a dict... you're welcome"

lms/resources/_js_config.py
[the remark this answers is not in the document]
jon-betts: Yeah, I think this is a continual problem. Especially if you indent the code or something, it's not obvious what you wrote vs. moved.

It's a good time to review this stuff I think, but if it's slowing things down feel free to ignore.

lms/resources/_js_config.py
```diff
@@ -53,34 +53,38 @@ def _hypothesis_config(self):
...
+            "nbf": now,
+            "exp": now + datetime.timedelta(minutes=5),
+        }
+
+        return jwt.encode(
+            claims,
+            self._request.registry.settings["h_jwt_client_secret"],
+            algorithm="HS256",
+        ).decode("utf-8")
+
```
jon-betts: Bangin'

tests/unit/lms/resources/_js_config_test.py
jon-betts: If we're keeping this for future tests + consistency, then this function should probably be called `config()` as it is in the other classes.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.fixture(autouse=True)
+def BearerTokenSchema(patch):
+    return patch("lms.resources._js_config.BearerTokenSchema")
+
+
+@pytest.fixture
+def bearer_token_schema(BearerTokenSchema):
+    return BearerTokenSchema.return_value
```
jon-betts: I can understand the desire to autouse `BearerTokenSchema` but I don't really get why `bearer_token_schema` is here specifically / exists at all.

seanh: Fixtures like this `bearer_token_schema` one are nowt but a convenience, to avoid tests having to say `BearerTokenSchema.return_value` all the time.

It avoids tests having to repeatedly say things like `BearerTokenSchema.return_value.authorization_param.assert_called_once_with(...)` or
`BearerTokenSchema.return_value.authorization_param.return_value`. Or tests having to assign their own local variable `bearer_token_schema = BearerTokenSchema.return_value`.

It originated as an aid to understanding as we had developers on the team who found `.return_value` difficult to understand, especially when there were multiple chained `.return_value`'s.

But I think it also just makes the test code nicer and more readable as `bearer_token_schema` is shorter than `BearerTokenSchema.return_value`.

It's used consistently throughout our tests: whenever there's a `FooBar` fixture that will be a mocked class, and the `foo_bar` fixture below it will be the mocked object instance of that class.

The only reason `bearer_token_schema` is located at module level is to keep it right nextdoor to its close cousin `BearerTokenSchema`.

PR #1475, opened 2020-02-18
Delete test DB tables before running tests

seanh:
On master the BDD tests pollutes the test DB such that future runs of the unit
tests fail:

```terminal
$ make backend-tests
(the tests pass)
$ make bddtests
(the tests pass)
$ make backend-tests
(a bunch of assertions fail because there's unexpected data in the test DB)
```

The only way to fix this once its happened is to manually delete your test DB.

This is arguably a problem with the BDD tests: they don't seem to clean up after
themselves, whereas the functests and unit tests do.

But as a fix I've changed the functests and unit tests to delete all the test DB
tables at the beginning of a test run. Ensuring that the test DB is in a clean
state before you start a test run seems more robust than trying to always leave
the DB clean _after_ a test run no matter what happened.

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -81,8 +81,19 @@ BASE = declarative_base(
 SESSION = sessionmaker()
 
 
-def init(engine):
-    """Initialise the database tables."""
+def init(engine, drop=False):
+    """
+    Create all the database tables if they don't already exist.
+
+    If `drop=True` is given then delete all existing tables first and then
+    re-create them. Tests use this. If `drop=False` existing tables won't be
+    touched.
+
+    :param engine: the sqlalchemy Engine object
+    :param drop: whether or not to delete existing tables
+    """
+    if drop:
+        BASE.metadata.drop_all(engine)
     BASE.metadata.create_all(engine)
 
 
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -11,15 +11,11 @@ TEST_SETTINGS["sqlalchemy.url"] = get_test_database_url(
 )
 
 
-@pytest.fixture(scope="session", autouse=True)
-def auto_use_dn(db_engine):
-    # TODO - Is this right?
-    return db_engine
-
-
-def _clean_database(engine):
+@pytest.fixture(autouse=True)
+def clean_database(db_engine):
+    """Delete any data added by the previous test."""
     tables = reversed(db.BASE.metadata.sorted_tables)
-    with contextlib.closing(engine.connect()) as conn:
+    with contextlib.closing(db_engine.connect()) as conn:
         tx = conn.begin()
         tnames = ", ".join('"' + t.name + '"' for t in tables)
         conn.execute("TRUNCATE {};".format(tnames))
@@ -35,7 +31,6 @@ def pyramid_app():
 
 @pytest.fixture
 def app(pyramid_app, db_engine):
-    _clean_database(db_engine)
     db.init(db_engine)
 
     return TestApp(pyramid_app)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,7 +47,13 @@ TEST_SETTINGS = {
 @pytest.fixture(scope="session")
 def db_engine():
     engine = sqlalchemy.create_engine(TEST_SETTINGS["sqlalchemy.url"])
-    db.init(engine)
+
+    # Delete all database tables and re-initialize the database schema based on
+    # the current models. Doing this at the beginning of each test run ensures
+    # that any schema changes made to the models since the last test run will
+    # be applied to the test DB schema before running the tests again.
+    db.init(engine, drop=True)
+
     return engine
 
 


lms/db/__init__.py
```diff
@@ -81,8 +81,10 @@ def update_from_dict(self, data, skip_keys=None):
 SESSION = sessionmaker()
 
 
-def init(engine):
+def init(engine, drop=False):
     """Initialise the database tables."""
```
jon-betts: Doc string for the param here? Seems like an obvious, but important one.

jon-betts:
Oh I forgot to say, I couldn't see why we didn't have a drop in test coverage until I remembered, the tests always call this!

Would be a pointless pain to test.

lms/db/__init__.py
seanh: Added one

tests/functional/conftest.py
```diff
@@ -35,7 +26,6 @@ def pyramid_app():
 
 @pytest.fixture
 def app(pyramid_app, db_engine):
-    _clean_database(db_engine)
     db.init(db_engine)
```
jon-betts: Does this need to be like this?

```suggestion
    db.init(db_engine, drop=True)
```

As the `db_engine()` fixture is scoped to the session, but `app()` isn't and needs to be cleared between runs (as we potentially setup a bunch of stuff for each test).

tests/functional/conftest.py
```diff
@@ -17,15 +17,6 @@ def auto_use_dn(db_engine):
     return db_engine
 
 
-def _clean_database(engine):
-    tables = reversed(db.BASE.metadata.sorted_tables)
-    with contextlib.closing(engine.connect()) as conn:
-        tx = conn.begin()
-        tnames = ", ".join('"' + t.name + '"' for t in tables)
-        conn.execute("TRUNCATE {};".format(tnames))
-        tx.commit()
```
jon-betts: :wave:

seanh:
Yep, it's easy enough to verify, this fails, good spot:

```diff
diff --git a/tests/functional/test_sanity.py b/tests/functional/test_sanity.py
index bf499be1..cea665e5 100644
--- a/tests/functional/test_sanity.py
+++ b/tests/functional/test_sanity.py
@@ -1,4 +1,5 @@
 from tests.functional.base_class import TestBaseClass
+from lms.models import ApplicationInstance
 
 
 class TestSanity(TestBaseClass):
@@ -7,7 +8,7 @@ class TestSanity(TestBaseClass):
     def test_sanity(self, app):
         app.get("/not_a_url_at_all", status=404)
 
-    def test_we_can_read_and_send_params(self, app):
+    def test_we_can_read_and_send_params(self, app, db_session):
         # We aren't really running this test because we care, but just
         # to prove the testing framework
         result = app.get(
@@ -16,3 +17,16 @@ class TestSanity(TestBaseClass):
 
         assert result.json_body["bar"] == "Bar feature flag is enabled"
         assert result.json_body["foo"] == "Foo feature flag is disabled"
+
+        db_session.add(
+            ApplicationInstance(
+                consumer_key="consumer_key",
+                shared_secret="shared_secret",
+                lms_url="lms_url",
+                requesters_email="requesters_email",
+            )
+        )
+        db_session.commit()
+
+    def test_foo(self, db_session):
+        assert db_session.query(ApplicationInstance).count() == 0
```

tests/functional/conftest.py
seanh: That would work but only for tests that actually use the `app` fixture, which admittedly is probably most or all functests. We could make the `app` fixture `autouse=True` or move the `db.init(db_engine, drop=True)` into a separate `db_init` fixture with `autouse=True`. But I don't think we want to do that because recreating the whole DB schema is slow. We do want to recreate it at the beginning of each test run: the DB schema defined by the models might have changed since the last test run, so we want to recreate the schema from scratch based on the current models, before each test run. But between each test we just want to undo any data changes that might have been made by the previous test, not make any schema changes, because that's much faster. That's why h uses a `clean_database()` function that uses `TRUNCATE`: it's the fastest way to clean the DB between tests

tests/functional/conftest.py
seanh: I've put this back but as an `autouse` fixture rather than a helper function that the `app` fixture calls because we want the DB to be cleaned before every test, not just before tests that use the `app` fixture. And because cleaning the DB doesn't have anything to do with creating the WSGI app and shouldn't be mixed in with it. Also moved the `clean_database` fixture to the top of `conftest.py` so that it executes before any other fixtures.

tests/functional/conftest.py
```diff
@@ -11,15 +11,11 @@
 )
 
 
-@pytest.fixture(scope="session", autouse=True)
-def auto_use_dn(db_engine):
-    # TODO - Is this right?
-    return db_engine
```
seanh: I'm not sure what this `auto_use_dn` (`db`?) fixture was for. I think it might have been trying to make the `db_engine` fixture, which doesn't itself have `autouse=True`, into an autoused fixture? But it isn't necessary because other fixtures, some of which are autouse, depend on `db_engine`, so `db_engine` is always used

PR #1477, opened 2020-02-18
Remove leading _ from lms.resources._* modules

seanh:
Remove the leading `_` from the names of the
`lms.resources._default` and `lms.resources._lti_launch` modules.

In the future modules that are *not* intended to be public (that are
just private helpers for the resources) will be added with leading
underscores.

The resource classes are still imported into the `__init__.py` so that
they can be used more conveniently like this:

instead of having to do:

    from lms.resources.lti_launch import LTILaunchResource


seanh:
The `__all__` is already gone, already merged a PR to remove it

PR #1478, opened 2020-02-18
Move debug settings into JSConfig

seanh:
Move the `"debug"` settings that some views add to the JS config into the `JSConfig` class. For simplicity these are now just added for every request, not just some, which does no harm.

Just moving the code, not redesigning it.

Details:
    
`BasicLTILaunchViews` inserts some debug info, used by the Gherkin tests, into the JavaScript config:

https://github.com/hypothesis/lms/blob/a59fc96fd2c80aa7e015bb668c78e5041e3c5a46/lms/views/basic_lti_launch.py#L44-L51
    
Move this JavaScript config mutation into the `JSConfig` class and make it unconditional: the debug info is now added to the `JSConfig` for all requests (that have a `JSConfig`). The less variation of the `JSConfig` between different requests the better.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -27,6 +27,7 @@ class JSConfig:
         """
         return {
             "authToken": self._auth_token,
+            "debug": self._debug,
             "hypothesisClient": self._hypothesis_config,
             "rpcServer": self._rpc_server_config,
             "urls": self._urls,
@@ -41,6 +42,24 @@ class JSConfig:
         return BearerTokenSchema(self._request).authorization_param(
             self._request.lti_user
         )
+
+    @property
+    def _debug(self):
+        """
+        Return some debug information.
+
+        Currently used in the Gherkin tests.
+        """
+        debug_info = {}
+
+        if self._request.lti_user:
+            debug_info["tags"] = [
+                "role:instructor"
+                if self._request.lti_user.is_instructor
+                else "role:learner"
+            ]
+
+        return debug_info
 
     @property
     def _hypothesis_config(self):
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -7,7 +7,7 @@ from pyramid.httpexceptions import HTTPBadRequest
 
 from lms.resources import LTILaunchResource
 from lms.resources._js_config import JSConfig
-from lms.values import HUser
+from lms.values import HUser, LTIUser
 
 
 class TestJSConfig:
@@ -47,14 +47,28 @@ class TestJSConfigAuthToken:
         )
         assert authToken == bearer_token_schema.authorization_param.return_value
 
-    def test_it_is_None_for_non_lti_users(self, context, pyramid_request):
-        pyramid_request.lti_user = None
-
-        assert JSConfig(context, pyramid_request).config["authToken"] is None
+    @pytest.mark.usefixtures("no_lti_user")
+    def test_it_is_None_for_non_lti_users(self, authToken):
+        assert authToken is None
 
     @pytest.fixture
     def authToken(self, config):
         return config["authToken"]
+
+
+class TestJSConfigDebug:
+    """Unit tests for the "debug" sub-dict of JSConfig."""
+
+    def test_it_contains_debugging_info_about_the_users_role(self, config):
+        assert "role:learner" in config["tags"]
+
+    @pytest.mark.usefixtures("no_lti_user")
+    def test_its_empty_if_theres_no_lti_user(self, config):
+        assert config == {}
+
+    @pytest.fixture
+    def config(self, config):
+        return config["debug"]
 
 
 class TestJSConfigHypothesisClient:
@@ -94,12 +108,9 @@ class TestJSConfigHypothesisClient:
 
         assert groups == ["example_groupid"]
 
-    def test_it_is_empty_if_provisioning_feature_is_disabled(
-        self, context, pyramid_request
-    ):
-        context.provisioning_enabled = False
-
-        assert JSConfig(context, pyramid_request).config["hypothesisClient"] == {}
+    @pytest.mark.usefixtures("provisioning_disabled")
+    def test_it_is_empty_if_provisioning_feature_is_disabled(self, config):
+        assert config == {}
 
     def test_it_is_mutable(self, config):
         config.update({"a_key": "a_value"})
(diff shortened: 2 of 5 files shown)


lms/resources/_js_config.py
```diff
@@ -42,6 +43,24 @@ def _auth_token(self):
...
+    @property
+    def _debug(self):
+        """
+        Return some debug information.
+
+        Currently used in the Gherkin tests.
+        """
+        debug_info = {}
+
+        if self._request.lti_user:
```
seanh: Apparently there are some LTI launch requests where `request.lti_user` is `None`. Checks for `request.lti_user` being `None` already exist elsewhere in the `JSConfig` code, and the tests cover this case and will crash if it's not handled.

I'm not sure that this is actually necessary / actually happens. But not 100% sure that it doesn't either. So sticking with it for now

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -47,18 +47,30 @@ def test_it(
...
 
-    def test_it_is_None_for_non_lti_users(
-        self, authToken, BearerTokenSchema, context, pyramid_request
-    ):
-        pyramid_request.lti_user = None
-
-        assert JSConfig(context, pyramid_request).config["authToken"] is None
+    @pytest.mark.usefixtures("no_lti_user")
+    def test_it_is_None_for_non_lti_users(self, authToken):
+        assert authToken is None
```
seanh: The addition of the `no_lti_user` fixture enables tests like this one, which require the `pyramid_request.lti_user` to be set to `None` _before_ the `JSConfig` object is initialized, to use fixtures like the `authToken` fixture which initialize the `JSConfig` for the test.

Move more of the test set up code out of the tests themselves and into the fixtures. Makes that setup code reusable as fixtures. Makes the test methods themselves simpler.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -159,3 +168,15 @@ def context():
...
+@pytest.fixture
+def no_lti_user(pyramid_request):
+    """Modify the pyramid_request fixture so that request.lti_user is None."""
+    pyramid_request.lti_user = None
+
+
+@pytest.fixture
+def provisioning_disabled(context):
+    """Modify context so that context.provisioning_enabled is False."""
+    context.provisioning_enabled = False
```
seanh: These fixtures deliberately don't return anything. They're not meant to be used as parameters of test methods, but rather in `@pytest.mark.usefixtures` decorators. They're not test objects as such, they're little bits of test set up wrapped in fixtures.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -91,17 +103,14 @@ def test_it_includes_grant_token(self, config):
         assert before <= claims["nbf"] <= after
         assert claims["exp"] > before
 
-    def test_it_includes_the_group(self, pyramid_request, config):
+    def test_it_includes_the_group(self, config):
```
seanh: Randomly deleted an unused fixture here

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -91,17 +103,14 @@ def test_it_includes_grant_token(self, config):
...
 
-    def test_it_is_empty_if_provisioning_feature_is_disabled(
-        self, context, pyramid_request
-    ):
-        context.provisioning_enabled = False
-
-        assert JSConfig(context, pyramid_request).config["hypothesisClient"] == {}
+    @pytest.mark.usefixtures("provisioning_disabled")
+    def test_it_is_empty_if_provisioning_feature_is_disabled(self, config):
+        assert config == {}
```
seanh: The `provisioning_disabled` fixture is only used once (so far) but it's still worth moving it out into a fixture for how much it simplifies this test and makes this test more consistent with the other tests, by letting it use the `config` fixture instead of initializing its own `JSConfig`

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -47,16 +47,30 @@ def test_it(
...
         return config["authToken"]
 
 
+class TestJSConfigDebug:
+    """Unit tests for the "debug" sub-dict of JSConfig."""
+
+    def test_it_contains_debugging_info_about_the_users_role(self, config):
+        assert "role:learner" in config["tags"]
+
+    @pytest.mark.usefixtures("no_lti_user")
```
jon-betts: :+1: Having just read about what this is, I think there are a ton of places where we could benefit from using `usefixtures` more. We quite often include fixtures in the args just to make something happen which can lead to very long argument lists.

tests/unit/lms/resources/_js_config_test.py
jon-betts: I like this, it's nice and declarative. 

You also know that the fixture is supposed to _do_ something or setup some condition, rather than _be_ something. This generally makes it easier to understand, because you can trust it does what it says on the tin and carry on. With a fixture, you usually have to go read it to work out what it is, and if it changes anything.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -386,7 +386,6 @@ def test_it_sets_the_right_javascript_config_settings(
 
         assert context.js_config == {
             "mode": "content-item-selection",
-            "debug": Any.dict(),
```
jon-betts: Am I being daft. Why has this gone? Because it's not this guys job anymore?

lms/resources/_js_config.py
```diff
@@ -42,6 +43,24 @@ def _auth_token(self):
...
+            debug_info.setdefault("tags", []).append(
+                "role:instructor"
+                if self._request.lti_user.is_instructor
+                else "role:learner"
```
jon-betts: Why does this use `setdefault, append`? The tags are never present before we start?

This is more of a head scratcher than `debug_info["tags"] = [...]`

tests/unit/lms/resources/_js_config_test.py
seanh: It _should_ theoretically trigger a pylint "unused argument" warning when we do that unless we're not using pylint (as is the case in h, for example) or we've disabled that warning for the tests

tests/unit/lms/views/basic_lti_launch_test.py
seanh: Yep, not this guy's job anymore. And `context.js_config` in these tests isn't the real `JSConfig` it's a fake that contains only the minimum stuff that `BasicLTILaunchViews` cares about:

https://github.com/hypothesis/lms/blob/b9d0c2712bfd7efa04ffde6e8b8d59aa66a08994/tests/unit/lms/views/basic_lti_launch_test.py#L455

This of course is a fake that's duplicating part of the real `JSConfig` and could get out of sync with it. `mock.create_autospec()` will take care of that for us soon, once `JSConfig` no longer exposes a dict directly

lms/resources/_js_config.py
seanh: Hmm, indeed. There seems to be no reason for it. Removed

PR #1481, opened 2020-02-19
A11y dialog fixes

LMS007:
This diff covers 3 main things

1. The first commit is just pulling in 2 files from the client that I feel are valuable for a11y changes related to Dialog.js. But additionally, I think these modules will have repeated use in LMS. I think we'll eventually splinter off modules like this into a npm package that we can import. But for now, I'm just dumping them all into a subfolder called /common. The intent is that everything in common/ will be replaced with a package down the road. I do believe there are a few other stray modules in LMS that have been copied over (including the rpc stuff) and the intent is that they will all move the common. The RPC stuff will be a bit more involved because they have diverged a bit from the client to LMS, so I'll need to get the common parts of those in sync again first before moving them over.

2. Fix the a11y lint issues with Dialog. I took a close look at the models from w3 against what we had and eventually went with this model.
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html

All my instincts told me we should not be placing focus on the wrapper and instead lean on aria- labels. I did test this with screen readers and all seems okay. We diverge a bit from aforementioned example in that we only allow the dialog to close if an `onCancel` callback is provided. I think that is totally reasonable because we don't really have a valid state to dismiss the dialog when things need authorization or the UI is in an error state and needs a full reload. I also decide that `alertdialog` was more appropriate here than just `dialog` based on the mozilla docs:

> The difference with regular dialogs is that the alertdialog role should only used when an alert, error, or warning occurs. In other words, when a dialog's information and controls require the user's immediate attention alertdialog should be used instead of dialog. It is up to the developer to make this distinction though.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role

3. Fix a minor lint issue in BasicLtiLaunchApp.js  to add a `title` (open to other text ideas )

TODO: change the docs at the top of Dialog.js to reflect the changes if these are approved.

### DIFF
--- a/lms/static/scripts/postmessage_json_rpc/server/index.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/index.js
@@ -1,5 +1,5 @@
 import Server from './server';
-import { requestConfig } from './methods';
+import { requestConfig, groupsAsync } from './methods';
 
 let server = {}; // Singleton RPC server reference
 
@@ -9,6 +9,7 @@ let server = {}; // Singleton RPC server reference
 function startRpcServer() {
   server = new Server();
   server.register('requestConfig', requestConfig);
+  server.register('groupsAsync', groupsAsync);
 }
 
 /**
--- a/lms/static/scripts/postmessage_json_rpc/server/index.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/index.js
@@ -1,5 +1,5 @@
 import Server from './server';
-import { requestConfig, groupsAsync } from './methods';
+import { groupsAsync, requestConfig, requestFrame } from './methods';
 
 let server = {}; // Singleton RPC server reference
 
@@ -10,6 +10,7 @@ function startRpcServer() {
   server = new Server();
   server.register('requestConfig', requestConfig);
   server.register('groupsAsync', groupsAsync);
+  server.register('requestFrame', requestFrame);
 }
 
 /**
--- a/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
@@ -1,4 +1,4 @@
-import { requestConfig } from './methods';
+import { requestConfig, requestFrame } from './methods';
 
 describe('postmessage_json_rpc/methods#requestConfig', () => {
   let configEl;
@@ -20,5 +20,10 @@ describe('postmessage_json_rpc/methods#requestConfig', () => {
   it('returns the config object', async () => {
     const result = await requestConfig();
     assert.deepEqual(result, { foo: 'bar' });
+  });
+
+  it('returns the parameters passed to requestFrame', async () => {
+    const result = await requestFrame({ bar: 1 });
+    assert.deepEqual(result, { bar: 1 });
   });
 });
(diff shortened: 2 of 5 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -135,7 +125,10 @@ Dialog.propTypes = {
    *
    * The "Cancel" button is added automatically if the `onCancel` prop is set.
    */
-  buttons: propTypes.arrayOf(propTypes.element),
+  buttons: propTypes.oneOfType([
```
LMS007: Note: This still causes an error until the node/element proeType polyfill" can be added

robertknight: That PR has landed, so this should now be resolved.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -189,13 +189,12 @@ export default function BasicLtiLaunchApp() {
...
   if (ltiLaunchState.state === 'fetched-url') {
     const iFrame = (
-      // FIXME-A11Y
-      // eslint-disable-next-line jsx-a11y/iframe-has-title
       <iframe
         width="100%"
         height="100%"
         className="js-via-iframe"
         src={ltiLaunchState.contentUrl}
+        title="Course assignment with annotation viewer"
```
robertknight: Can I suggest moving this change to a separate PR since I think it should be quicker to review and merge than the rest of the changes. I'm not certain about the best title here, maybe ask on Slack for suggestions.

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -42,48 +43,37 @@ export default function Dialog({
...
     //
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   return (
-    // FIXME-A11Y
-    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
     <div
-      role="dialog"
+      role="alertdialog"
```
robertknight: The `Dialog` control is used in multiple contexts, some of which `alertdialog` is appropriate for (eg. `ErrorDialog`) but many where I think `dialog` would make more sense (eg. the dialogs that appear in the file picker app when selecting a source to pick a file from). I'd suggest that we make this a configurable prop.

What are the effects of using `alertdialog` vs `dialog`?

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -42,48 +43,37 @@ export default function Dialog({
...
       aria-labelledby="Dialog__title"
-      onKeyDown={handleKey}
-      // FIXME-A11Y
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex="0"
-      ref={rootEl}
+      aria-modal="true"
+      aria-describedby="Dialog__content"
```
robertknight: Hmm, is it common to use `aria-describedby` for the entire content of the dialog? Also, I'd suggest that it probably would be a good idea to adopt a different syntax for IDs as opposed to class names to avoid mixing them up (eg. where someone adds an `aria-` attribute but then sets a class rather than an ID on the target element).

LMS007: I think since the content is fully customized, it may not be valuable to even add aria-describedby, and perhaps the aria-labelledby is enough. Thoughts?

I'll wait on the other PR to land which deals with the ids, then I'll update this one again.

robertknight:
I've had a quick look over the updated PR. The code looks good. I want to do some manual testing.

@LMS007 is this still WIP or is it ready for a "proper" review and OK to merge after that?

PR #1483, opened 2020-02-20
Various a11y lint fixes

LMS007:
This is by no means complete, but getting there. I still see lots of areas we could improve and converge on in terms of CSS, but I'd rather reserve that work in the pattern lib than duplicate whats been done on the client.

In this PR:
- Change ValidationMessage to be a button and give is a focus state
-Change SubmitGradeForm to listen to onInput instead of onKeydown to dismiss validation message
- Give SubmitGradeForm a unique id counter for axe validation
- Add onBlur to StudentSelector.js
- Remove erroneous `<a>` tag in LMSFilePicker
- Add informative comment about enableLmsFilePicker prop
- Fix missing alt attribute in FileList

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FileList.js
+++ b/lms/static/scripts/frontend_apps/components/FileList.js
@@ -39,14 +39,13 @@ export default function FileList({
         renderItem={(file, isSelected) => (
           <Fragment>
             <td aria-label={file.display_name} className="FileList__name-col">
-              {/* FIXME-A11Y */}
-              {/* eslint-disable-next-line jsx-a11y/alt-text */}
               <img
                 className={classnames(
                   'FileList__icon',
                   isSelected && 'is-selected'
                 )}
                 src="/static/images/file-pdf.svg"
+                alt="PDF icon"
               />
               <span className="FileList__name">{file.display_name}</span>
             </td>
--- a/lms/static/scripts/frontend_apps/components/StudentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/StudentSelector.js
@@ -59,8 +59,12 @@ export default function StudentSelector({
 
     return (
       <span className="StudentsSelector__students">
-        {/* FIXME-A11Y */}
-        {/* eslint-disable-next-line jsx-a11y/no-onchange */}
+        {/* 
+        This lint issue may have arisen from browser inconsistency issues with 
+        `onChange` which have since been fixed. See browser compatibility note here:
+        https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event#annotations:xeC6ClQsEequhUdih2lXzw
+        */}
+        {/* eslint-disable-next-line jsx-a11y/no-onchange*/}
         <select
           className="StudentsSelector__students-select"
           onChange={e => {
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -14,6 +14,7 @@ import ErrorDialog from './ErrorDialog';
 import Spinner from './Spinner';
 import SvgIcon from './SvgIcon';
 import { fetchGrade, submitGrade } from '../utils/grader-service';
+import { useUniqueId } from '../utils/hooks';
 import { trustMarkup } from '../utils/trusted';
 import { formatToNumber, scaleGrade, validateGrade } from '../utils/validation';
 import ValidationMessage from './ValidationMessage';
@@ -90,6 +91,8 @@ export default function SubmitGradeForm({ disabled = false, student }) {
   const [showValidationError, setValidationError] = useState(false);
   // The actual validation error message.
   const [validationMessage, setValidationMessageMessage] = useState('');
+  // Unique id attribute for <input>
+  const gradeId = useUniqueId('SubmitGradeForm__grade:');
 
   const { authToken } = useContext(Config);
 
@@ -150,7 +153,7 @@ export default function SubmitGradeForm({ disabled = false, student }) {
           setValidationError(false);
         }}
       />
-      <label className="SubmitGradeForm__label" htmlFor="lms-grade">
+      <label className="SubmitGradeForm__label" htmlFor={gradeId}>
         Grade (Out of 10)
       </label>
       <span className="SubmitGradeForm__grade-wrapper">
@@ -159,9 +162,9 @@ export default function SubmitGradeForm({ disabled = false, student }) {
             'is-saved': gradeSaved,
           })}
           disabled={disabled}
-          id="lms-grade"
+          id={gradeId}
           ref={inputRef}
-          onKeyDown={handleKeyDown}
+          onInput={handleKeyDown}
           type="input"
           defaultValue={grade}
           key={student.LISResultSourcedId}
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
@@ -152,9 +152,7 @@ export default function LMSFilePicker({
         <ErrorDisplay
           message={
             <Fragment>
-              {/* FIXME-A11Y */}
-              {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
-              <a>{`Failed to authorize with the ${lmsName} instance at `}</a>
+              {`Failed to authorize with the ${lmsName} instance at `}
               <a href={`${lmsUrl}`}>{`${lmsUrl}`}</a>
             </Fragment>
           }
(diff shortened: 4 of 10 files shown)


lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -31,6 +31,9 @@ export default function FilePickerApp({
     authToken,
     authUrl,
     courseId,
+    // It is assumed that enableLmsFilePicker is only used by LMS apps
```
LMS007: pulled this comment from the py side, I found it useful when trying to figure out how to put the client into local file picker mode.

lms/static/scripts/frontend_apps/components/StudentSelector.js
```diff
@@ -59,13 +59,15 @@ export default function StudentSelector({
...
       <span className="StudentsSelector__students">
-        {/* FIXME-A11Y */}
-        {/* eslint-disable-next-line jsx-a11y/no-onchange */}
         <select
           className="StudentsSelector__students-select"
           onChange={e => {
             onSelectStudent(parseInt(e.target.value));
           }}
+          // See https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md
+          onBlur={e => {
```
LMS007: I found this has no ill wanted side effect, but I still don't like it, I'm almost leaning towards adding the lint disable comment back. Thoughts?

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -23,6 +23,9 @@ import ValidationMessage from './ValidationMessage';
 // back to [0-1] when saving it to the api service.
 const GRADE_MULTIPLIER = 10;
 
+// Global counter used to create a unique ids
+let idCounter = 0;
```
LMS007: me thinks axe adds multiple instances of the component into each test, so I did this? This seems to be an emerging pattern so maybe there is a way to encapsulate this.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -159,9 +170,9 @@ export default function SubmitGradeForm({ disabled = false, student }) {
             'is-saved': gradeSaved,
           })}
           disabled={disabled}
-          id="lms-grade"
+          id={`lms-grade-${instanceId}`}
           ref={inputRef}
-          onKeyDown={handleKeyDown}
+          onInput={handleKeyDown}
```
LMS007: This allows the user to keyboard tab back to the validation message and have it in focus, otherwise as soon as you press tab, the message hides because onKeyDown was triggered. This more appropriately changes it to only an input change.

lms/static/scripts/frontend_apps/components/FileList.js
```diff
@@ -39,14 +39,13 @@ export default function FileList({
...
             <td aria-label={file.display_name} className="FileList__name-col">
-              {/* FIXME-A11Y */}
-              {/* eslint-disable-next-line jsx-a11y/alt-text */}
               <img
                 className={classnames(
                   'FileList__icon',
                   isSelected && 'is-selected'
                 )}
                 src="/static/images/file-pdf.svg"
+                alt="File icon"
```
robertknight: Perhaps this should be more specific and say "PDF icon". On the other hand the most likely future change that I can think of is that we add folders to this list in which case the most important differentiation will be between files and folders.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -31,6 +31,9 @@ export default function FilePickerApp({
...
+    // for which we do not have an integration with the file storage.
```
robertknight: This should say "for which we *do* have an integration" surely? I'd also clarify that "the file storage" means "the LMS's file storage API".

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -31,6 +31,9 @@ export default function FilePickerApp({
...
+    // (currently only Canvas supports this).
```
robertknight: This comment is true at the time of writing, but we definitely do intend to support file storage for other LMSes in future and Jon and Sean have been actively looking into it.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
robertknight: Extract a hook perhaps, eg. `const id = useUniqueId(prefix)`?

robertknight: > me thinks axe adds multiple instances of the component into each test, so I did this?

axe tests do run [one at a time](https://github.com/hypothesis/lms/blob/08e1e251fef66d92b72abbc405cc0b5eb183dd8d/lms/static/scripts/test-util/accessibility.js#L61) and elements [are removed](https://github.com/hypothesis/lms/blob/08e1e251fef66d92b72abbc405cc0b5eb183dd8d/lms/static/scripts/test-util/accessibility.js#L38) after each test. However it is possible that axe is remembering IDs it has already seen internally.

In any case, I think it is good practice to design almost all components so that you could have multiple instances in a document, even if the current UI only requires one.

lms/static/scripts/frontend_apps/components/StudentSelector.js
robertknight: According to a [comment in the MDN docs](https://hyp.is/XcWxEFPMEeq0tYNWatw4eA/developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) the "change" event is now triggered consistently across all modern browsers (including Firefox as of v63, from October 2018). On that basis, assuming `onChange` is triggered when we want it to be, when using either keyboard or mouse navigation, I'd be OK with putting the lint disable comment back. Do include a note about why we're ignoring it if so.

lms/static/styles/components/_ValidationMessage.scss
```diff
@@ -37,6 +40,7 @@
   &:hover {
     cursor: pointer;
   }
+  @include mixins.input-focus;
```
robertknight: I generally prefer to put mixins at the top of a CSS rule because that makes it easier to reason about the final values of properties, as it avoids a mixin overriding a property defined specifically for a CSS rule which is almost always going to be a mistake.

LMS007:
@robertknight made enough changes where I would like 1 more round, thanks!

lms/views/basic_lti_launch.py
```diff
@@ -220,8 +220,8 @@ def unconfigured_basic_lti_launch(self):
         self.context.js_config.update(
             {
                 "mode": "content-item-selection",
-                # It is assumed that this view is only used by LMSes for which
-                # we do not have an integration with the LMS's file storage.
+                # It is assumed that this view is only used by LMS apps
+                # for which we do have an integration with the file storage.
```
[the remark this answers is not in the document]
LMS007: Gotcha, sorry, I'll re-create this comment in the config.js file

lms/static/scripts/frontend_apps/components/ValidationMessage.js
```diff
@@ -33,11 +34,13 @@ export default function ValidationMessage({
   });
 
   return (
-    // FIXME-A11Y
-    // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
-    <div onClick={closeValidationError} className={errorClass}>
+    <button
+      aria-label="Validation error"
```
robertknight: What is the reason for the `aria-label` here? This will end up replacing the presumably more specific `{message}` content. My understanding is that it is generally preferable not to override the UI text with something else unless the meaning of the UI text would be unclear for some reason.

LMS007: I think you're right, can't remember why I added it

lms/static/scripts/frontend_apps/components/ValidationMessage.js
```diff
@@ -33,11 +34,18 @@ export default function ValidationMessage({
   });
 
   return (
-    // FIXME-A11Y
-    // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
-    <div onClick={closeValidationError} className={errorClass}>
+    <button
+      // Add the aria-hidden="true" to inform axe tests to ignore it when
```
LMS007: @robertknight Got to the button of this. I added the aria-label to make the axe happy w/o thinking too much about it because it told me it wanted a label, but it also said a few other things that spoke to the root of the issue.  When the validation message is closed, its not actually hidden because it needs to slide closed. So I needed to add 2 things to make this work correctly. 

Observing the docs here https://www.w3.org/TR/using-aria/#4thrule
Its seems reasonable to add `aria-hidden` directly yon the button. In this case, I don't think we need to cast to a string because I don't care about that attribute being set to "false" (that is an aside). But because its also a button, I needed to disable to the focus state when it was hidden so I added a dynamic tabindex too. axe tests now pass.

PR #1489, opened 2020-02-23
Move all JavaScript config settings into JSConfig (draft)

seanh:


seanh:
This is just a draft version of this for easy reference, not intended to be merged

PR #1494, opened 2020-02-24
Change RPC server methods to be async

LMS007:
- We need to add the ability for an RPC method to be async so it can fetch a result from the lms service. These changes allow the methods to return to the result asynchronously and delay the postMessage response according.

- Move the _resolveSidebarWindow() call until after the _jsonRPCResponse returns. This ensures we have the frame ready only after we respond to the client that send it.

### DIFF
--- a/lms/static/scripts/postmessage_json_rpc/server/server.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/server.js
@@ -61,7 +61,7 @@ export default class Server {
    * Receive a postMessage event and, if it's a JSON-RPC request from an
    * allowed origin, post back either a result or an error response.
    */
-  _receiveMessage(event) {
+  async _receiveMessage(event) {
     if (!this._allowedOrigins.includes(event.origin)) {
       return;
     }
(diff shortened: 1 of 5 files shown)


lms/static/scripts/postmessage_json_rpc/server/server.js
```diff
@@ -61,22 +61,24 @@ export default class Server {
...
+    event.source.postMessage(result, event.origin);
+
     // Resolve the promise we created in the constructor with the saved
     // sidebar frame and origin.
     this._resolveSidebarWindow({
       frame: event.source,
       origin: event.origin,
     });
-
-    event.source.postMessage(this._jsonRPCResponse(event.data), event.origin);
```
LMS007: This stumped me for a while! It was causing the changeFocusModeUser in LMSGrader.js to timeout and not resolve but only on the first request at load time. The effect was benign since the default user was not in focus, but bothersome. 

Turns out once I moved the rpc methods to async, it delayed that first request to fetch the config (initiated from the client)  too much and that meant that the changeFocusModeUser went out before the RPC response came back from that first request, and this caused the changeFocusModeUser to never make it to the client's rpc server. 

forcing the _resolveSidebarWindow to happen only after we await on the _jsonRPCResponse fixed this issue.

LMS007: Upon further review, this may not be directly addressing the issue. Turns out the client's rpc server's listener may not be started before the `changeFocusModeUser` request is sent. May need to refactor the hand shake parts some more. to be continued...

PR #1496, opened 2020-02-25
Fix some test fixtures

seanh:
Fix a bunch of `BasicLTILaunchViews` test fixtures that were either missing, were not autouse when they should be, or were not module-scope when they should be.

* This makes the tests more correct (all the tests are now unit tests for `basic_lti_launch_test.py`, which is what they're supposed to be according to our testing approach, whereas previously some of them were also testing code from other modules)

* It also brings the code in line with our general testing patterns: the unit tests for a module always patch imports of other modules _from our own codebase_, and import patch fixtures are always module-level and autouse

* It's going to make the test code easier to refactor. I'm going to need to move a bunch of tests around between different test classes. Module-level autouse fixtures will just work as I do this without any changes, whereas class-level and non-autouse fixtures would require much juggling as I refactor the tests.

### DIFF
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -375,7 +375,7 @@ class TestConfigureModuleItem(ConfiguredLaunch):
 
 class TestUnconfiguredBasicLTILaunch:
     def test_it_sets_the_right_javascript_config_settings(
-        self, context, pyramid_request
+        self, BearerTokenSchema, bearer_token_schema, context, pyramid_request
     ):
         pyramid_request.params.update({"some_random_rubbish": "SOME_RANDOM_RUBBISH"})
 
@@ -402,9 +402,13 @@ class TestUnconfiguredBasicLTILaunch:
 
         # Test that we pass through parameters from the request made to us
         # onto the configure module item call
+        BearerTokenSchema.assert_called_once_with(pyramid_request)
+        bearer_token_schema.authorization_param.assert_called_once_with(
+            pyramid_request.lti_user
+        )
         assert form_fields == Any.dict.containing(
             {
-                "authorization": Any.string(),
+                "authorization": bearer_token_schema.authorization_param.return_value,
                 "resource_link_id": "TEST_RESOURCE_LINK_ID",
                 "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
                 "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
@@ -412,22 +416,6 @@ class TestUnconfiguredBasicLTILaunch:
                 "context_id": "TEST_CONTEXT_ID",
                 "some_random_rubbish": "SOME_RANDOM_RUBBISH",
             }
-        )
-
-        self._assert_authorization_valid_jwt(
-            form_fields["authorization"],
-            {
-                "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
-                "roles": "TEST_ROLES",
-                "user_id": "TEST_USER_ID",
-            },
-        )
-
-    def _assert_authorization_valid_jwt(self, authorization, expected_values):
-        assert authorization.startswith("Bearer ")
-        assert (
-            decode_jwt(authorization[len("Bearer ") :], "test_secret")
-            == expected_values
         )
 
     @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -169,14 +169,6 @@ def test_it_does_not_call_grading_info_upsert_if_canvas(
 
         grading_info_service.upsert_from_request.assert_not_called()
 
-    @pytest.fixture
-    def frontend_app(self, patch):
-        return patch("lms.views.basic_lti_launch.frontend_app")
-
-    @pytest.fixture(autouse=True)
-    def via_url(self, patch):
-        return patch("lms.views.basic_lti_launch.via_url")
```
seanh: These have been moved to module-level autouse fixtures, below.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -191,14 +183,6 @@ def lti_h_service(self, pyramid_config):
         pyramid_config.register_service(lti_h_service, name="lti_h")
         return lti_h_service
 
-    @pytest.fixture
-    def LtiLaunches(self, patch):
-        return patch("lms.views.basic_lti_launch.LtiLaunches")
-
-    @pytest.fixture(autouse=True)
-    def ModuleItemConfiguration(self, patch):
-        return patch("lms.views.basic_lti_launch.ModuleItemConfiguration")
```
seanh: These have been moved to module-level autouse fixtures, below.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -422,20 +406,13 @@ def test_it_sets_the_right_javascript_config_settings(
...
+            pyramid_request.lti_user
         )
-
-    def _assert_authorization_valid_jwt(self, authorization, expected_values):
-        assert authorization.startswith("Bearer ")
         assert (
-            decode_jwt(authorization[len("Bearer ") :], "test_secret")
-            == expected_values
+            form_fields["authorization"]
+            == bearer_token_schema.authorization_param.return_value
```
seanh: We don't care _what_ the authorization string is (that's `BearerTokenSchema`'s job, and it has its own unit tests) we only care that it came from `BearerTokenSchema`

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -422,22 +410,6 @@ def test_it_sets_the_right_javascript_config_settings(
...
-                "user_id": "TEST_USER_ID",
-            },
-        )
-
-    def _assert_authorization_valid_jwt(self, authorization, expected_values):
-        assert authorization.startswith("Bearer ")
-        assert (
-            decode_jwt(authorization[len("Bearer ") :], "test_secret")
-            == expected_values
-        )
```
seanh: We don't care _what_ the authorization string is or whether it decodes successfully as a JWT with a given secret. That's `BearerTokenSchema`'s job, and it has its own unit tests. The fact that the authorization string is a JWT is an internal implementation detail of `BearerTokenSchema`, not part of its public API. In `BasicLTILaunchViews` we only care that it came from `BearerTokenSchema`

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -410,9 +394,13 @@ def test_it_sets_the_right_javascript_config_settings(
...
         # Test that we pass through parameters from the request made to us
         # onto the configure module item call
+        BearerTokenSchema.assert_called_once_with(pyramid_request)
+        bearer_token_schema.authorization_param.assert_called_once_with(
+            pyramid_request.lti_user
+        )
         assert form_fields == Any.dict.containing(
             {
-                "authorization": Any.string(),
+                "authorization": bearer_token_schema.authorization_param.return_value,
```
seanh: We simply test that `BasicLTILaunchViews` sets `"authorization_param"` to what `BearerTokenSchema` returned. Testing the  implementation details of what `BearerTokenSchema` returns (a JWT encoded with a secret) is done in `BearerTokenSchema`'s own unit tests. This is much simpler than these tests decoding the JWT to a dict, doesn't duplicate existing `BearerTokenSchema` unit tests, and doesn't couple `BasicLTILaunchViews` tests to `BearerTokenSchema`'s implementation details (only its API), doesn't allow `BasicLTILaunchViews` tests to fail if there's a bug in `BearerTokenSchema`.

PR #1498, opened 2020-02-26
WIP- Rpc improvements

LMS007:
Relates to https://github.com/hypothesis/client/pull/1843

This PR covers  3 changes.

1. Add ability to have async rpc methods

We need to add the ability for an RPC method to be async so it can fetch a result from the lms service. These changes allow the methods to return to the result asynchronously and delay the postMessage response according. This includes the changes from this PR https://github.com/hypothesis/lms/pull/1494

2. Add requestFrame method
`requestFrame` is used to discover which ancestor is the embedding frame as opposed to guessing at which frame is the right ancestor each call.

3. Allow RPC server methods to accept parameters

This matches what the client RPC server can already do. Parameters must be in array format otherwise they are ignored.

e.g.
```
{
      jsonrpc: '2.0',
      id: 'test_id',
      method: 'registeredMethod',
      params: ['one' {two: 'two’}]
      timeout: 2000
};
```

Note: I will need to squash these commits into 3, some of the messages are outdated.

### DIFF
--- a/lms/static/scripts/postmessage_json_rpc/server/index.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/index.js
@@ -1,5 +1,5 @@
 import Server from './server';
-import { requestConfig } from './methods';
+import { requestConfig, groupsAsync } from './methods';
 
 let server = {}; // Singleton RPC server reference
 
@@ -9,6 +9,7 @@ let server = {}; // Singleton RPC server reference
 function startRpcServer() {
   server = new Server();
   server.register('requestConfig', requestConfig);
+  server.register('groupsAsync', groupsAsync);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -9,6 +9,8 @@ let server = {}; // Singleton RPC server reference
 function startRpcServer() {
   server = new Server();
   server.register('requestConfig', requestConfig);
+  server.register('groupsAsync', groupsAsync);
```
LMS007: comment this out or remove it from this PR

lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -9,8 +9,37 @@
 /**
  * Return a Hypothesis client config object for the current LTI request.
  */
-export function requestConfig() {
+export async function requestConfig() {
   const configEl = document.querySelector('.js-config');
   const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
-  return clientConfigObj;
+  return Promise.resolve(clientConfigObj);
```
LMS007: don't need to return a resolved promise, async already does that. Just return the result.

PR #1502, opened 2020-02-26
Remove test class inheritance

seanh:
Factor out some test class inheritance in `basic_lti_launch_test.py`
that was being used to run one base class of tests against a few
different view methods.
Replace this with a pytest parametrized fixture instead.

Test class inheritance is un-idiomatic in pytest and I think it might be
incompatible with some pytest features. I also found it quite confusing
(I didn't spot the inheritance, and it sent me spinning for a while) and
my experience with test class inheritance (in past lives when I worked
on projects that used nosetests) is that it usually leads to test
maintenance horror by tightly coupling test classes to each other, and
by making code harder to understand as you have to refer to multiple
classes to follow it. The main problem though is that the inheritance
and coupling is making it hard for me to refactor these tests in the way
that I need to, because I need to split and join test classes, move
tests between classes, etc.

Replace the inheritance with a simpler and more explicit fixture-based
approach that makes the tests much easier to maintain and to refactor
as the test classes are now all separate from each other,
and is more idiomatic in pytest.

## How these test classes worked before

* Each of the test classes `TestCanvasFileBasicLTILaunch`,
  `TestDBConfiguredBasicLTILaunch`, `TestURLConfiguredBasicLTILaunch` and
  `TestConfigureModuleItem` had a `make_request(context, pyramid_request)`
  method that would call the appropriate view method.

* Each of the four test classes also overrode the `pyramid_request`
  fixture and added the request params necessary to call that view method.
  For example `canvas_file_basic_lti_launch()` only works if `"file_id"`
  is in `request.params` (and only ever gets called if `"file_id"` is in
  `request.params`), and so on.

* Tests common to all four view methods were in a base class named
  `ConfiguredLaunch`. `ConfiguredLaunch` would not itself be picked up
  by pytest as a test class (name doesn't start with "Test"). Rather,
  each of `TestCanvasFileBasicLTILaunch`,
  `TestDBConfiguredBasicLTILaunch`, `TestURLConfiguredBasicLTILaunch`
  and `TestConfigureModuleItem` inherited from `ConfiguredLaunch` and so
  inherited the test methods in `ConfiguredLaunch`.

* `ConfiguredLaunch`'s test methods called `self.make_request(context,
  request)` in order to call the view method under test, and by
  inheritance this would be the `make_request` of
  `TestCanvasFileBasicLTILaunch`, `TestDBConfiguredBasicLTILaunch`,
  `TestURLConfiguredBasicLTILaunch` or `TestConfigureModuleItem`
  depending on which test class was being run, so `ConfiguredLaunch`'s
  tests would all get run for all four view methods.

* `ConfiguredLaunch` wasn't a great name since these tests are also run
  against the `configure_module_item()` view which is an _un_-configured
  launch

## How these test classes work now

* For each view method there's a "caller" function at the top of
  `basic_lti_launch_test.py` that sets up the necessary request params
  and then calls the view:

  ```python
  def canvas_file_basic_lti_launch_caller(context, pyramid_request):
      pyramid_request.params["file_id"] = "TEST_FILE_ID"
      views = BasicLTILaunchViews(context, pyramid_request)
      return views.canvas_file_basic_lti_launch()

  def db_configured_basic_lti_launch_caller(context, pyramid_request):
      ...

  def url_configured_basic_lti_launch_caller(context, pyramid_request):
      ...

  def configure_module_item_caller(context, pyramid_request):
      ...
  ```

* The four test classes no longer have their `make_request()` methods
  nor their `pyramid_request` fixtures, as both things are now done by
  the caller function. Each test class calls the appropriate caller function
  when it wants to call the view.

* The tests common to all four view methods are in a `TestCommon` class
  that uses a `view_caller` parametrized fixture to run all of its tests
  four times:

  ```python
  class TestCommon:
      def test_something(self, context, pyramid_request, view_caller):
          ...

          view_caller(context, pyramid_request)

          ...

    @pytest.fixture(
        params=[
            canvas_file_basic_lti_launch_caller,
            db_configured_basic_lti_launch_caller,
            url_configured_basic_lti_launch_caller,
            configure_module_item_caller,
        ]
    )
    def view_caller(self, request):
        return request.param
  ```

* Module-level functions are the appropriate place for these caller functions, which are helper functions shared by multiple test classes in the module. Same as how shared pytest fixtures or any other shared helper functions would be module-level functions, if they were used by multiple test classes. The test classes don't use each other in any way, but they can all use the module-level helpers and fixtures.

* This is simple and explicit, there's no inheritance, and the test
  classes are decoupled from each other

### DIFF
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -6,8 +6,6 @@ from h_matchers import Any
 from lms.resources import LTILaunchResource
 from lms.resources._js_config import JSConfig
 from lms.services import HAPIError
-from lms.services.h_api import HAPI
-from lms.validation.authentication._helpers._jwt import decode_jwt
 from lms.values import HUser, LTIUser
 from lms.views.basic_lti_launch import BasicLTILaunchViews
 
@@ -305,7 +303,7 @@ class TestDBConfiguredBasicLTILaunch:
         )
 
     def test_it_configures_frontend_grading(
-        self, context, pyramid_request, frontend_app, via_url, ModuleItemConfiguration,
+        self, context, pyramid_request, frontend_app
     ):
         db_configured_basic_lti_launch_caller(context, pyramid_request)
         frontend_app.configure_grading.assert_called_once_with(
@@ -328,13 +326,7 @@ class TestURLConfiguredBasicLTILaunch:
         )
 
     def test_it_configures_frontend_grading(
-        self,
-        context,
-        pyramid_request,
-        frontend_app,
-        lti_outcome_params,
-        via_url,
-        ModuleItemConfiguration,
+        self, context, pyramid_request, frontend_app, lti_outcome_params,
     ):
         pyramid_request.params = lti_outcome_params
 
@@ -365,7 +357,7 @@ class TestConfigureModuleItem:
         assert context.js_config.config["urls"]["via_url"] == via_url.return_value
 
     def test_it_configures_frontend_grading(
-        self, context, pyramid_request, frontend_app, via_url, ModuleItemConfiguration,
+        self, context, pyramid_request, frontend_app
     ):
         configure_module_item_caller(context, pyramid_request)
 
@@ -420,7 +412,7 @@ class TestUnconfiguredBasicLTILaunch:
         )
 
     @pytest.fixture
-    def pyramid_request(self, context, pyramid_request):
+    def pyramid_request(self, pyramid_request):
         pyramid_request.params = {
             "user_id": "TEST_USER_ID",
             "resource_link_id": "TEST_RESOURCE_LINK_ID",
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -12,6 +12,75 @@ from lms.values import HUser, LTIUser
 from lms.views.basic_lti_launch import BasicLTILaunchViews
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -12,6 +12,75 @@
 from lms.views.basic_lti_launch import BasicLTILaunchViews
 
 
+def canvas_file_basic_lti_launch_caller(context, pyramid_request):
```
seanh: The "caller" naming style is from the stdlib's `methodcaller`: https://docs.python.org/3/library/operator.html#operator.methodcaller

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -128,36 +203,55 @@ def test_it_reports_lti_launches(self, context, pyramid_request, LtiLaunches):
...
+        """
+        Return a function that calls the view method to be tested.
 
-class TestCanvasFileBasicLTILaunch(ConfiguredLaunch):
+        This is a parametrized fixture. A test that uses this fixture will be
+        run multiple times, once for each parametrized version of this fixture.
+
+        See https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures
+        """
+        return request.param
```
seanh: `request` here is a pytest thing FYI, not a Pyramid request object. This is why our `pyramid_request` fixture has that awkward name instead of just `request`: pytest already has a builtin fixture named `request` which it uses to pass in fixture params and various other things.

PR #1504, opened 2020-02-26
Add resource_link_id param to pyramid_request fixture

seanh:
Rather than individual tests adding this standard LTI launch param to
pyramid_request.params just add it to the global pyramid_request fixture
as we already do for other standard params.

Also removed some unnecessary setting of the tool_consumer_instance_guid
param in tests: this param is _already_ present in the global
pyramid_request fixture.

### DIFF
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -223,9 +223,7 @@ class TestDBConfiguredBasicLTILaunch(ConfiguredLaunch):
         self.make_request(context, pyramid_request)
 
         ModuleItemConfiguration.get_document_url.assert_called_once_with(
-            pyramid_request.db,
-            "TEST_TOOL_CONSUMER_INSTANCE_GUID",
-            "TEST_RESOURCE_LINK_ID",
+            pyramid_request.db, "TEST_GUID", "TEST_RESOURCE_LINK_ID",
         )
         via_url.assert_called_once_with(pyramid_request, "TEST_DOCUMENT_URL")
         assert context.js_config.config["urls"]["via_url"] == via_url.return_value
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -279,11 +267,7 @@ def test_it_configures_frontend_grading(
         via_url,
         ModuleItemConfiguration,
     ):
-        pyramid_request.params = {
-            "resource_link_id": "TEST_RESOURCE_LINK_ID",
-            "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
-            **lti_outcome_params,
-        }
+        pyramid_request.params = lti_outcome_params
```
jon-betts: All other calls to this in the file do:

`pyramid_request.params.update(lti_outcome_params)`

Should that be the same here?

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -279,11 +267,7 @@ def test_it_configures_frontend_grading(
         via_url,
         ModuleItemConfiguration,
     ):
-        pyramid_request.params = {
-            "resource_link_id": "TEST_RESOURCE_LINK_ID",
-            "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
-            **lti_outcome_params,
```
seanh: In this case the view being tested doesn't even need `"resource_link_id"` or `"tool_consumer_instance_guid"`. The test was adding them for no apparent reason

seanh: Actually it doesn't need the `lti_outcome_params` either

PR #1505, opened 2020-02-26
Factor out lti_outcome_params fixture

seanh:
In preparation for replacing the lti_outcome_params fixture (in a future
PR) remove some unnecessary uses of it.

This reduced coverage so I increased it again by omitting some files
that really don't need to be tested.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,5 +10,5 @@ omit =
 [report]
 show_missing = True
 precision = 2
-fail_under = 98.56
+fail_under = 98.68
 skip_covered = True
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,8 +7,13 @@ source =
 omit =
     lms/scripts/*
 
+    # Don't bother covering these files as they just contain feature flags test
+    # views, not really user-facing code.
+    lms/extensions/feature_flags/views/test.py
+    lms/views/feature_flags_test.py
+
 [report]
 show_missing = True
 precision = 2
-fail_under = 98.68
+fail_under = 99.16
 skip_covered = True
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -10,15 +10,14 @@ from lms.models import ApplicationInstance
 )
 def create_application_instance(request):
     """Create application instance in the databse and respond with key and secret."""
-    # Check to make sure that both developer key and secret are present. If not set to None.
+
+    # Default developer_key and developer_secret to None rather than letting
+    # them be empty strings.
     developer_key = request.params["developer_key"].strip()
     developer_secret = request.params["developer_secret"].strip()
-    if developer_key == "":
-        developer_key = None
 
-    if developer_secret == "":
-        developer_secret = None
-
+    # If either one of developer_key or developer_secret is missing, then we
+    # don't save the other one either.
     if not developer_key or not developer_secret:
         developer_key = None
         developer_secret = None
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -87,12 +87,10 @@ class TestBasicLTILaunchViewsInit:
 
         assert context.js_config.config["mode"] == "basic-lti-launch"
 
-    def test_it_adds_report_submission_config_if_required_params_present(
-        self, context, pyramid_request, lti_outcome_params
+    def test_it_adds_report_submission_config_if_the_LMS_is_Canvas(
+        self, context, canvas_request
     ):
-        pyramid_request.params.update(lti_outcome_params)
-
-        BasicLTILaunchViews(context, pyramid_request)
+        BasicLTILaunchViews(context, canvas_request)
 
         assert context.js_config.config["submissionParams"] == {
             "h_username": context.h_user.username,
@@ -109,9 +107,8 @@ class TestBasicLTILaunchViewsInit:
         ],
     )
     def test_it_doesnt_add_report_submission_config_if_required_param_missing(
-        self, context, pyramid_request, lti_outcome_params, key
+        self, context, pyramid_request, key
     ):
-        pyramid_request.params.update(lti_outcome_params)
         del pyramid_request.params[key]
 
         BasicLTILaunchViews(context, pyramid_request)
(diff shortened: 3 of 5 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -52,9 +52,8 @@ def test_it_doesnt_add_report_submission_config_if_required_param_missing(
         assert "submissionParams" not in context.js_config.config
 
     def test_it_doesnt_add_report_submission_config_if_lms_not_canvas(
-        self, context, pyramid_request, lti_outcome_params
+        self, context, pyramid_request
     ):
-        pyramid_request.params.update(lti_outcome_params)
         pyramid_request.params.update(
             {"tool_consumer_info_product_family_code": "whiteboard"}
         )
```
jon-betts: This is a little OTT for setting a single value :woman_shrugging:

.coveragerc
```diff
@@ -6,9 +6,11 @@ source =
     tests/unit
 omit =
     lms/scripts/*
+    lms/extensions/feature_flags/views/test.py
+    lms/views/feature_flags_test.py
```
jon-betts: Why are these files not subject to coverage? A comment here would be nice, otherwise it just lives in heads and is hard to question.

.coveragerc
```diff
@@ -6,9 +6,11 @@ source =
...
 
 [report]
 show_missing = True
 precision = 2
-fail_under = 98.49
+fail_under = 98.98
```
jon-betts: If we are dropping covering in this test, is it worth just fessing up?

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -248,16 +247,8 @@ def test_it_configures_via_url(
...
 
     def test_it_configures_frontend_grading(
-        self,
-        context,
-        pyramid_request,
-        frontend_app,
-        lti_outcome_params,
-        via_url,
-        ModuleItemConfiguration,
+        self, context, pyramid_request, frontend_app, via_url, ModuleItemConfiguration,
```
jon-betts: I freaking love it when this happens!

.coveragerc
seanh: They're just weird little test views for the feature flags mechanism. They're not anything actually being used by users. They could even just be deleted but they can be handy now and then

.coveragerc
seanh: Naah, let us always move towards 100%. Once we hit 100% it will no longer be possible for a PR to reduce coverage by deleting more lines than it adds and stuff like that, so things will be a little more convenient

.coveragerc
seanh: Added a comment

tests/unit/lms/views/basic_lti_launch_test.py
seanh: Fixed

seanh:
I've added some new commits to this PR:

* Increase test coverage by adding tests for `views/application_instances.py`

* Remove the `lti_outcome_params` fixture. The `"lis_result_sourcedid"` and `"lis_outcome_service_url"` params that it added have been added to a general `pyramid_request` fixture in the module since all launch requests actually have these standard params.

  Certain specific requests -- e.g. a launch by a teacher, or a SpeedGrader launch -- are missing one or another of these params. Tests for those specific cases delete those params from the request.

  These weren't added to the `pyramid_request` in `conftest.py` because doing so breaks some unrelated tests for reasons I don't want to deal with now.

* The `"tool_consumer_info_product_family_code"` that `lti_outcome_params` added has been added to the universal `pyramid_request` in `conftest.py`. Every launch request has this. It's set to `"whiteboard"` (i.e. not Canvas) by default.

* Added a `canvas_request` fixture that sets `"tool_consumer_info_product_family_code"` to `"canvas"`, and changed tests for Canvas-specific behaviour to use it

lms/views/application_instances.py
```diff
@@ -10,18 +10,11 @@
...
-    developer_secret = request.params["developer_secret"].strip()
-    if developer_key == "":
-        developer_key = None
-
-    if developer_secret == "":
-        developer_secret = None
-
-    if not developer_key or not developer_secret:
-        developer_key = None
-        developer_secret = None
```
seanh: Amusing code here

lms/views/application_instances.py
```diff
@@ -10,15 +10,14 @@
...
+
+    # Default developer_key and developer_secret to None rather than letting
+    # them be empty strings.
     developer_key = request.params["developer_key"].strip()
     developer_secret = request.params["developer_secret"].strip()
-    if developer_key == "":
-        developer_key = None
-
-    if developer_secret == "":
-        developer_secret = None
```
seanh: If either of these is falsey (e.g. an empty string) then they both get set to `None` right below, so these four lines are unnecessary. (The fact that it silently deletes the given value for either one of these fields if the other field is empty doesn't seem great but I'm not fixing this page right now, just adding test coverage.)

tests/unit/lms/views/test_application_instance.py
```diff
@@ -1,25 +1,68 @@
...
+    def test_if_developer_secret_is_missing_it_doesnt_save_developer_key_either(
+        self, pyramid_request
+    ):
+        pyramid_request.params["developer_key"] = "example_key"
+
+        create_application_instance(pyramid_request)
+
+        ai = pyramid_request.db.query(ApplicationInstance).one()
+        assert ai.developer_key is None
+        assert ai.developer_secret is None
```
jon-betts: These two tests look like they could be a parameterised `test_if_developer_key_or_secret_is_missing_the_other_isnt_saved` or something by iterating over the key to change.

tests/unit/lms/views/test_application_instance.py
```diff
@@ -1,25 +1,68 @@
...
-            .filter(ApplicationInstance.lms_url == pyramid_request.params["lms_url"])
-            .count()
-        )
-        create_application_instance(pyramid_request)
-        new_count = (
-            pyramid_request.db.query(ApplicationInstance)
-            .filter(ApplicationInstance.lms_url == pyramid_request.params["lms_url"])
-            .count()
-        )
-        assert new_count == initial_count + 1
```
jon-betts: I'm assuming this isn't necessary because we truncate the DB now?

seanh: The unit tests always truncated the DB. Whoever wrote this just didn't know what they were doing

tests/unit/lms/views/test_application_instance.py
seanh: Done

PR #1507, opened 2020-02-26
Add ability to have async RPC methods handlers

LMS007:
Relates to https://github.com/hypothesis/client/pull/1843

**Add ability to have async RPC methods handlers** 

We need to add the ability for an RPC method to be async so it can fetch a result from the lms service and potentially block on the client (e.g groups). These changes allow the methods to return to the result asynchronously and delay the postMessage response according.

### DIFF
--- a/lms/static/scripts/postmessage_json_rpc/server/server.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/server.js
@@ -61,7 +61,7 @@ export default class Server {
    * Receive a postMessage event and, if it's a JSON-RPC request from an
    * allowed origin, post back either a result or an error response.
    */
-  _receiveMessage(event) {
+  async _receiveMessage(event) {
     if (!this._allowedOrigins.includes(event.origin)) {
       return;
     }
@@ -76,7 +76,8 @@ export default class Server {
       origin: event.origin,
     });
 
-    event.source.postMessage(this._jsonRPCResponse(event.data), event.origin);
+    const result = await this._jsonRPCResponse(event.data);
+    event.source.postMessage(result, event.origin);
   }
 
   /**
@@ -99,7 +100,7 @@ export default class Server {
   /**
    * Return a JSON-RPC response object for the given JSON-RPC request object.
    */
-  _jsonRPCResponse(request) {
+  async _jsonRPCResponse(request) {
     // Return an error response if the request id is invalid.
     // id must be a string, number or null.
     const id = request.id;
@@ -123,6 +124,15 @@ export default class Server {
     }
 
     // Call the method and return the result response.
-    return { jsonrpc: '2.0', result: method(), id: request.id };
+    try {
+      const result = await method();
+      return { jsonrpc: '2.0', result: result, id: request.id };
+    } catch (e) {
+      return {
+        jsonrpc: '2.0',
+        id: request.id,
+        error: { code: -32600, message: e.message },
+      };
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -9,8 +9,22 @@
 /**
  * Return a Hypothesis client config object for the current LTI request.
  */
-export function requestConfig() {
+export async function requestConfig() {
   const configEl = document.querySelector('.js-config');
   const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
-  return clientConfigObj;
+  return Promise.resolve(clientConfigObj);
```
robertknight: Functions declared `async` always return a Promise, so you don't need to wrap the result in `Promise.resolve(..)` explicitly.

lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -9,8 +9,22 @@
...
+}
+
+/**
+ * Temporary method that blocks for a little while to simulate
+ * waiting for canvas groups to be ready.
+ *
+ * TODO: replace this with a real request to lms/
+ */
+export async function groupsAsync() {
```
robertknight: If all of the methods in this file are going to be async, I'd suggest the `Async` suffix can be avoided. I would suggest only adding that if for some reason there are two methods that differ only in whether they are sync or async (like Node's "fs" module has sync and async versions of many functions and adds a `Sync` suffix to the synchronous ones).

lms/static/scripts/postmessage_json_rpc/server/server.js
```diff
@@ -61,22 +61,24 @@ export default class Server {
...
     if (!this._allowedOrigins.includes(event.origin)) {
       return;
     }
 
     if (!this._isJSONRPCRequest(event)) {
       return;
     }
+
+    const result = await this._jsonRPCResponse(event.data);
+    event.source.postMessage(result, event.origin);
```
robertknight: I don't think it will make a practical difference, but was there a reason for moving this up from before the `this._resolveSidebarWindow` call?

lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -1,5 +1,5 @@
 import Server from './server';
-import { requestConfig } from './methods';
+import { requestConfig, groupsAsync } from './methods';
```
robertknight: Are you using `groupsAsync` for anything locally or was it just code for you to test your changes locally, if so I'd omit it from this PR.

lms/static/scripts/postmessage_json_rpc/server/server.js
LMS007: Good q. Yes, and no!

In short, I will move this back to the previous line to avoid confusion.  And to be clear, its the next bit of code that moved down which was the important part:

```
this._resolveSidebarWindow
```

_Backstory.._
At first I thought it fixed a timeout issue I saw after making these calls `async`... and in fact, it did, but it only masked the problem. Which makes more sense when I more closely understood was what going on in the client. We need to build out the `readyToReceive` RPC call that the client sends to LMS to only after its own RPC server is spun up. Once I build that out, this issue will be resolved. PR to come.

LMS007:
Made this more of a minimal diff and removed the group method. 
Also, with `await`, we don't need to return a promise at all, so it can just be optional for each method to be `async` or not. This is more straight forward and each method that needs to be async, can be. Otherwise it does not have to.

PR #1511, opened 2020-02-28
Change get_by_assignment() to just return GradingInfos

seanh:
Change `GradingInfoService.get_by_assignment()` to just return
`GradingInfo` objects instead of `(GradingInfo, HUser)` 2-tuples.

The calling code (in `frontend_app.py`) is now responsible for creating
the `HUser` object corresponding to each `GradingInfo` which it needs to
do in order to create the "student" dicts that it puts into the JS
config.

Overall this ends up making the tests for both `GradingInfoService` and
`frontend_app.py` simpler, and is a better division of work between the
two.

### DIFF
--- a/lms/views/helpers/frontend_app.py
+++ b/lms/views/helpers/frontend_app.py
@@ -1,4 +1,5 @@
 """Helpers for configuring the front-end JavaScript application."""
+from lms.values import HUser
 
 __all__ = ("configure_grading",)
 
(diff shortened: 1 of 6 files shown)


lms/views/helpers/frontend_app.py
```diff
@@ -26,19 +27,32 @@ def configure_grading(request, js_config):
...
+                # The keys here end up in the JavaScript config object,
+                # hence the JavaScript casingStyle.
                 "userid": h_user.userid,
                 "displayName": h_user.display_name,
                 "LISResultSourcedId": grading_info.lis_result_sourcedid,
                 "LISOutcomeServiceUrl": grading_info.lis_outcome_service_url,
             }
-            for (grading_info, h_user) in grading_infos
-        ],
+        )
```
seanh: A plain old loop is a lot easier to read than a comprehension in this case

tests/unit/lms/app_test.py
```diff
@@ -1,6 +1,28 @@
+from unittest import mock
+
 import pytest
 
-from lms.app import create_app
+from lms.app import configure_jinja2_assets, create_app
+
+
+class TestConfigureJinja2Assets:
```
seanh: For coverage reasons I ended up adding these random tests

tests/unit/lms/services/grading_info_test.py
```diff
@@ -72,14 +98,7 @@ def test_it_creates_new_record_if_no_matching_exists(
...
     def test_it_updates_existing_record_if_matching_exists(
-        self,
-        svc,
-        pyramid_request,
-        h_user,
-        lti_user,
-        grading_info,
-        db_session,
-        lti_params,
+        self, svc, pyramid_request, h_user, lti_user, db_session,
```
seanh: Random unused fixtures being removed here

tests/unit/lms/views/helpers/frontend_app_test.py
```diff
@@ -35,89 +28,75 @@ def test_it_disables_grading_if_grading_is_disabled_for_assignment(
...
+    def grading_info_service(self, grading_info_service):
+        grading_info_service.get_by_assignment.return_value = [
+            mock.create_autospec(
+                GradingInfo,
+                instance=True,
+                spec_set=True,
+                lis_result_sourcedid=f"test_lis_result_sourcedid_{i}",
+                lis_outcome_service_url=f"test_lis_outcomes_service_url_{i}",
+                h_username=f"test_h_username_{i}",
+                h_display_name=f"test_h_display_name_{i}",
```
seanh: One nice thing about using a mock here is you only need to set the `GradingInfo` attributes that `frontend_app.py` actually uses and not the others

tests/unit/lms/services/grading_info_test.py
```diff
@@ -10,46 +10,72 @@
...
-            oauth_consumer_key=lti_user.oauth_consumer_key,
-            context_id=lti_params["context_id"],
-            resource_link_id=lti_params["resource_link_id"],
+            oauth_consumer_key, context_id, resource_link_id
         )
 
-        assert not grading_infos
-
-    @pytest.fixture
-    def another_grading_info(self, lti_params, db_session):
```
jon-betts: Was this just not used? How did I miss this?

PR #1512, opened 2020-02-28
Tidy up JSConfig

seanh:
A bunch of small JSConfig tidy-ups to pave the way for coming JSConfig changes:

* More docstrings
* More code comments
* Change things that don't need to be `@property`'s to just methods
* Change things that don't need to be either `@property`'s or methods to just attributes or inline them into `config()`
* `JSConfig._hypothesis_config` is renamed to `_hypothesis_client` to match its name in the config dict, and has `@functools.lru_cache()` added to it because future `JSConfig` code is going to need to mutate `self._hypothesis_client` as a dict

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -14,6 +14,9 @@ class JSConfig:
         self._context = context
         self._request = request
 
+        # A dict of URLs for the frontend to use.
+        self._urls = {}
+
     @property
     @functools.lru_cache()
     def config(self):
@@ -119,14 +122,3 @@ class JSConfig:
             self._request.registry.settings["h_jwt_client_secret"],
             algorithm="HS256",
         ).decode("utf-8")
-
-    @property
-    def _urls(self):
-        """
-        Return a dict of URLs for the frontend to use.
-
-        For example: API endpoints for the frontend to call would go in
-        here.
-
-        """
-        return {}
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -23,13 +23,18 @@ class JSConfig:
         """
         Return the configuration for the app's JavaScript code.
 
-        This is a mutable config dict. It can be accessed, for example by
-        views, and they can mutate it or add their own view-specific config
-        settings. The modified config object will then be passed to the
-        JavaScript code in the response page.
-
         :rtype: dict
         """
+        # This is a lazy-computed property so that if it's going to raise an
+        # exception that doesn't happen until someone actually reads it.
+        # If it instead crashed in JSConfig.__init__() that would happen
+        # earlier in the request processing pipeline and could change the error
+        # response.
+        #
+        # We cache this property (@functools.lru_cache()) so that it's
+        # mutable. You can do self.config["foo"] = "bar" and the mutation will
+        # be preserved.
+
         return {
             # The auth token that the JavaScript code will use to authenticate
             # itself to our own backend's APIs.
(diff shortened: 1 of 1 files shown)


lms/resources/_js_config.py
```diff
@@ -14,28 +14,52 @@ def __init__(self, context, request):
...
         Return the configuration for the app's JavaScript code.
 
-        This is a mutable config dict. It can be accessed, for example by
-        views, and they can mutate it or add their own view-specific config
-        settings. The modified config object will then be passed to the
-        JavaScript code in the response page.
+        This is a cached property (with @functools.lru_cache()) so that it's
+        mutable. Code outside of this method can do self.config["foo"] = "bar"
+        and, because self.config always returns the same cached dict and not a
+        new dict every time, the mutation will be preserved.
```
seanh: This mutability will be going away eventually

lms/resources/_js_config.py
```diff
@@ -64,19 +87,34 @@ def _debug(self):
...
+
+        This is implemented as a property rather than having the __init__()
+        method initialize the self._hypothesis_client dict as a simple
+        attribute, because generating the dict can raise. If it's going to
+        raise we don't want that to happen until as late as possible (not until
+        someone actually tries to access the dict). If JSConfig.__init__() were
+        to raise that would change the location of the crash in Pyramid's
+        request processing pipeline and could change the type of error response
+        as a result.
```
seanh: These docstrings are a bit verbose but the reason for why this needs to be a cached property and not just either a method or `self` attr initialized in `__init__()` is quite obscure so I thought it worth spelling out

lms/resources/_js_config.py
```diff
@@ -14,28 +14,52 @@ def __init__(self, context, request):
...
+        and, because self.config always returns the same cached dict and not a
+        new dict every time, the mutation will be preserved.
+
+        This is implemented as a property rather than having the __init__()
+        method initialize the self.config dict as a simple attribute, because
+        generating the dict can raise.  If it's going to raise we don't want
+        that to happen until as late as possible (not until someone actually
+        tries to access the dict). If JSConfig.__init__() were to raise that
+        would change the location of the crash in Pyramid's request processing
+        pipeline and could change the type of error response as a result.
```
jon-betts: This is very wordy and seems more like a comment than doc string. Would something shorter suffice like:

```
Lazily create this on demand so that any errors raised happen during access in
the views, not earlier in the Pyramid request pipeline.

We cache the return value so this property is externally mutable.
```

lms/resources/_js_config.py
```diff
@@ -14,28 +14,52 @@ def __init__(self, context, request):
...
+            # The config object for our JSON-RPC server.
+            "rpcServer": {
+                "allowedOrigins": self._request.registry.settings[
+                    "rpc_allowed_origins"
+                ],
+            },
+            # A dict of URLs for the frontend to use.
+            # For example: API endpoints for the frontend to call would go in
+            # here.
             "urls": self._urls,
```
jon-betts: If `rpcServer` above can be inlined I think the `_urls` dict can be do. Unless I'm missing something about why this is in the constructor?

lms/resources/_js_config.py
jon-betts: Yeah, it's kind of a copy of the comment above. Similar thoughts on this one. I think a little more thought could slim this down. Or maybe a note at the top about the cached properties, or a decorator to house the string. 

Doesn't really matter but I always groan when I see a wall of text like, because it's _probably_ not relevant to what I'm up to right now, but I'd have to read the thing every time to be sure.

lms/resources/_js_config.py
```diff
@@ -14,28 +14,52 @@ def __init__(self, context, request):
         self._context = context
         self._request = request
 
+        # A dict of URLs for the frontend to use.
+        self._urls = {}
```
jon-betts: I don't understand what this is up to.

seanh: This is initialized to an empty dict and then, in future, new code in `JSConfig` will mutate it before `config()` returns it

lms/resources/_js_config.py
seanh: It's because `urls` will get mutated in future whereas `rpcServer` won't

lms/resources/_js_config.py
seanh: Changed these to comments and reworded them to be shorter

PR #1515, opened 2020-02-28
Remove un-used test imports

jon-betts:
This is due to a gap in our linting at present


jon-betts:
Just wait for the tests to go through on this one to make sure.

Also Pycharm FTW. There's an auto clean up my imports which is handy for this sort of thing.

PR #1516, opened 2020-02-28
Centralize JSConfig 1/n: Move the "lmsName" config setting into JSConfig

seanh:
The `"lmsName"` setting is the `"Canvas"` string on the <kbd>Select PDF from Canvas</kbd> button when creating an assignment. It is always `"Canvas"`, never takes any other value.

This PR changes things so that this setting is now present in the JSConfig all the time whereas previously it was only present for certain views. But this doesn't do any harm.

An alternative approach to this one would be to remove the setting entirely, and just have the frontend code hardcode the single `"Canvas"` value instead of using a false setting that never changes.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -48,6 +48,9 @@ class JSConfig:
             # Our JSON-RPC server passes this to the Hypothesis client over
             # postMessage.
             "hypothesisClient": self._hypothesis_client,
+            # The LMS name to use in user-facing messages.
+            # Shown on the "Select PDF from Canvas" button label.
+            "lmsName": "Canvas",
             # The config object for our JSON-RPC server.
             "rpcServer": {
                 "allowedOrigins": self._request.registry.settings[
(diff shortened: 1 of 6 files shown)


lms/resources/_js_config.py
```diff
@@ -48,6 +48,9 @@ def config(self):
             # Our JSON-RPC server passes this to the Hypothesis client over
             # postMessage.
             "hypothesisClient": self._hypothesis_client,
+            # The LMS name to use in user-facing messages.
+            # Shown on the "Select PDF from Canvas" button label.
+            "lmsName": "Canvas",
```
jon-betts: This is super weird, because a lot of the time this is wrong. Previously presumably this wouldn't have been the case as it was only added in a Canvas files context?

We don't happen to have a string around in the LTI parameters that would actually tell is the LMS name do we?

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/1533

PR #1517, opened 2020-02-28
Centralize JSConfig 3/n: Move the "mode" JavaScript setting into JSConfig

seanh:
Depends on https://github.com/hypothesis/lms/pull/1522 for the tests to pass.

The "mode" config setting tells the JavaScript code whether it should do a normal assignment launch (`"basic-lti-launch"` mode) or whether it should show the UI for the teacher to configure the assignment (selecting the document for the assignment) (`"content-item-selection"` mode).

The mode is now changed to default to `"basic-lti-launch"` mode, and it gets explicitly changed to `"content-item-selection"` in two specific places.

Previously the default was `"content-item-selection"` (but this default was implemented in JavaScript not Python) and it had to be explicitly set to `"basic-lti-launch"` and then in one place it had to be explicitly set back to `"content-item-selection"` again _after_ changing it to `"basic-lti-launch"`.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -48,6 +48,11 @@ class JSConfig:
             # Our JSON-RPC server passes this to the Hypothesis client over
             # postMessage.
             "hypothesisClient": self._hypothesis_client,
+            # What "mode" to put the JavaScript code in.
+            # For example in "basic-lti-launch" mode the JavaScript code
+            # launches its BasicLtiLaunchApp, whereas in
+            # "content-item-selection" mode it launches its FilePickerApp.
+            "mode": "basic-lti-launch",
             # The config object for our JSON-RPC server.
             "rpcServer": {
                 "allowedOrigins": self._request.registry.settings[
@@ -59,6 +64,15 @@ class JSConfig:
             # here.
             "urls": self._urls,
         }
+
+    def enable_content_item_selection_mode(self):
+        """
+        Put the JavaScript code into "content item selection" mode.
+
+        This mode shows teachers an assignment configuration UI where they can
+        choose the document to be annotated for the assignment.
+        """
+        self.config["mode"] = "content-item-selection"
 
     def _auth_token(self):
         """Return the authToken setting."""
--- a/lms/static/scripts/frontend_apps/index.js
+++ b/lms/static/scripts/frontend_apps/index.js
@@ -15,12 +15,10 @@ startRpcServer();
 const rootEl = document.querySelector('#app');
 const config = JSON.parse(document.querySelector('.js-config').textContent);
 
-const mode = config.mode || 'content-item-selection';
-
 render(
   <Config.Provider value={config}>
-    {mode === 'basic-lti-launch' && <BasicLtiLaunchApp />}
-    {mode === 'content-item-selection' && <FilePickerApp />}
+    {config.mode === 'basic-lti-launch' && <BasicLtiLaunchApp />}
+    {config.mode === 'content-item-selection' && <FilePickerApp />}
   </Config.Provider>,
   rootEl
 );
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -37,13 +37,6 @@ class BasicLTILaunchViews:
         self.context = context
         self.request = request
 
-        self.context.js_config.config.update(
-            {
-                # Configure the front-end mini-app to run.
-                "mode": "basic-lti-launch",
-            }
-        )
-
         if self.is_launched_by_canvas():
             self.initialise_canvas_submission_params()
             self.set_canvas_focused_user()
@@ -200,6 +193,7 @@ class BasicLTILaunchViews:
         we'll save it in our DB. Subsequent launches of the same assignment
         will then be DB-configured launches rather than unconfigured.
         """
+        self.context.js_config.enable_content_item_selection_mode()
 
         params = self._extract_lti_params(self.request)
 
@@ -211,7 +205,6 @@ class BasicLTILaunchViews:
         # Add the config needed by the JavaScript document selection code.
         self.context.js_config.config.update(
             {
-                "mode": "content-item-selection",
                 # It is assumed that this view is only used by LMSes for which
                 # we do not have an integration with the LMS's file storage.
                 # (currently only Canvas supports this).
--- a/lms/views/content_item_selection.py
+++ b/lms/views/content_item_selection.py
@@ -51,6 +51,8 @@ def content_item_selection(context, request):
     lti_h_service.upsert_h_user()
     lti_h_service.upsert_course_group()
 
+    context.js_config.enable_content_item_selection_mode()
+
     context.js_config.config.update(
         {
             # The URL that the JavaScript code will open if it needs the user to
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -18,6 +18,11 @@ class TestJSConfig:
 
         assert config["a_key"] == "a_value"
 
+    def test_enable_content_item_selection_mode(self, js_config):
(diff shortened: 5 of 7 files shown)


lms/resources/_js_config.py
```diff
@@ -60,6 +65,15 @@ def config(self):
...
         }
 
+    def enable_content_item_selection_mode(self):
+        """
+        Put the JavaScript code into "content item selection" mode.
+
+        This mode shows teachers an assignment configuration UI where they can
+        choose the document to be annotated for the assignment.
+        """
+        self.config["mode"] = "content-item-selection"
```
seanh: There's no need for an `enable_basic_lti_launch_mode()` because that's the default.

lms/static/scripts/frontend_apps/index.js
```diff
@@ -15,12 +15,10 @@ startRpcServer();
 const rootEl = document.querySelector('#app');
 const config = JSON.parse(document.querySelector('.js-config').textContent);
 
-const mode = config.mode || 'content-item-selection';
```
seanh: There's no need for JavaScript to have a fallback default for the mode -- the Python always puts the mode in the JSConfig

lms/views/basic_lti_launch.py
```diff
@@ -37,13 +37,6 @@ def __init__(self, context, request):
         self.context = context
         self.request = request
 
-        self.context.js_config.config.update(
-            {
-                # Configure the front-end mini-app to run.
-                "mode": "basic-lti-launch",
-            }
-        )
```
seanh: "basic-lti-launch" is now the default, so no need to set this here

lms/views/basic_lti_launch.py
```diff
@@ -211,7 +205,6 @@ def unconfigured_basic_lti_launch(self):
         # Add the config needed by the JavaScript document selection code.
         self.context.js_config.config.update(
             {
-                "mode": "content-item-selection",
```
seanh: This is now done by the `self.context.js_config.enable_content_item_selection_mode()` call above. This is the one view in `BasicLTILaunchViews` that needs `"content-item-selection"` mode instead of `"basic-lti-launch"` mode

lms/views/content_item_selection.py
```diff
@@ -51,6 +51,8 @@ def content_item_selection(context, request):
     lti_h_service.upsert_h_user()
     lti_h_service.upsert_course_group()
 
+    context.js_config.enable_content_item_selection_mode()
```
seanh: This view previously relied JavaScript defaulting the mode to `"content-item-selection"`. It didn't set any mode in the JS config. Now JSConfig always contains a mode and it defaults to `"basic-lti-launch"`, so this view now has to change it to `"content-item-selection"`

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -155,8 +162,13 @@ def bearer_token_schema(BearerTokenSchema):
 
 
 @pytest.fixture
-def config(context, pyramid_request):
-    return JSConfig(context, pyramid_request).config
+def js_config(context, pyramid_request):
+    return JSConfig(context, pyramid_request)
```
seanh: Sometimes it's convenient for tests to receive the `JSConfig.config` dict as a fixture directly, sometimes they need the containing `JSConfig` object itself. So I've made two separate fixtures for them

lms/resources/_js_config.py
jon-betts: Would this be nicer as something which accepted a value (maybe an Enum)? We only have two values right now, but any additions would cause a proliferation of `enable_<mode>_mode()` functions. A `@property` setter would allow something along the lines of:

```
config.mode = ConfigMode.CONTENT_ITEM_SELECTION
```

lms/static/scripts/frontend_apps/index.js
jon-betts: :+1: This is the kind of stuff that makes it very hard to reason about when debugging. You send through `None` expecting things to break and it carries on but behaves differently.

lms/views/basic_lti_launch.py
jon-betts: It's technically unnecessary, but I wonder if it would be nice to have to set it here as well (not like this of course). This would mean:

 * Everything behaved the same regardless of mode
 * You'd know the mode existed and what value it would be set to just by looking at the view
    *  Rather than having to know it existed and that there was a default set in the JSConfig

This would be easily accommodated with a parameterised `mode` or `set_mode` etc.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -344,7 +339,6 @@ def test_it_sets_the_right_javascript_config_settings(
         BasicLTILaunchViews(context, pyramid_request).unconfigured_basic_lti_launch()
 
         assert context.js_config.config == {
-            "mode": "content-item-selection",
```
jon-betts: Just spotted the test below re-adds the coverage this line was providing before.

lms/resources/_js_config.py
seanh: In future PRs this method is actually gonna grow a large amount of stuff that needs to be set whenever we go into content item selection mode (whenever we call `js_config.enable_content_item_selection_mode()`), beyond just setting the `"mode"` string. And `enable_content_item_selection_mode()` is also gonna grow more params beyond the `mode` param that are specific to content item selection mode. So if we had an `enable_mode(mode)` method it'd have to be something like:

```python
def enable_mode(self, mode, *args, **kwargs):
    self.config["mode"] = mode

    if "mode" == "content-item-selection":
        self._enable_content_item_selection_mode(*args, **kwargs)

def _enable_content_item_selection_mode(self, ...):
    # Sets loads of things.
```

For now, I think I'll opt to leave it as it is and deal with it in the future if/when we have a proliferation of different modes, because that will be the time when we'll be able to know which code design will work best for whatever is needed to enable all those different modes.

lms/views/basic_lti_launch.py
seanh: It would be a little weird though because `BasicLTILaunchViews.__init__()` would be explicitly setting the mode to `"basic-lti-launch"` and then `BasicLTILaunchViews.unconfigured_basic_lti_launch()` would be changing it to `"content-item-selection"`.

I think it'd just be more code. `BasicLTILaunchViews.__init__()` would need to set the mode and need a test that it sets the mode. And `JSConfig` would need a way of setting the mode to  `"basic-lti-launch"`, whether a separate method or by complicating the `enable_content_item_selection_mode()` method (which in future is going to grow a bunch of content item selection-specific params), and tests for that.

seanh:
It might be useful to look at when `enable_content_item_selection()` is going to look like, roughly, once all the config has been moved into `JSConfig`:

```python
    def enable_content_item_selection_mode(
        self, form_action, form_fields, lti_launch_url=None
    ):
        """
        Put the JavaScript code into "content item selection" mode.

        This mode shows teachers an assignment configuration UI where they can
        choose the document to be annotated for the assignment.
        """
        self.config.update(
            {
                "enableLmsFilePicker": False,
                "formAction": form_action,
                "formFields": form_fields,
                "googleClientId": self._request.registry.settings["google_client_id"],
                "googleDeveloperKey": self._request.registry.settings[
                    "google_developer_key"
                ],
                # Pass the URL of the LMS that is launching us to our JavaScript code.
                # When we're being launched in an iframe within the LMS our JavaScript
                "customCanvasApiDomain": self._context.custom_canvas_api_domain,
                "lmsUrl": self._context.lms_url,
            }
        )

        if lti_launch_url:
            self.config["ltiLaunchUrl"] = lti_launch_url

        # Enable the "LMS file picker" (Canvas file picker) if it's available.
        if self._canvas_files_available():
            self.config["enableLmsFilePicker"] = True
            self.config["courseId"] = self._request.params["custom_canvas_course_id"]
```

PR #1518, opened 2020-02-29
Centralize JSConfig 4/n: Move the "authUrl" JavaScript setting into JSConfig

seanh:
The `"authUrl"` setting is the URL of the popup window that the
JavaScript code opens when it needs the user to get us a new Canvas API
access token by (re-)authorizing us with Canvas. It's actually a URL of
the LMS app itself, not a Canvas URL, but it redirects the popup window
to the appropriate Canvas URL.

This setting should probably be named something else given that it's
actually specific to the Canvas API.

It also belongs in the `"urls"` JavaScript setting (which is a dict of
all the URLs the JavaScript code needs to know about) not as its own
top-level setting.

But this PR limits itself to moving the `"authUrl"` setting into the
`JSConfig` class.

This change means that `"authUrl"` exists in the JS config all the time
not just some of the time, which does no harm.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -42,6 +42,9 @@ class JSConfig:
             # The auth token that the JavaScript code will use to authenticate
             # itself to our own backend's APIs.
             "authToken": self._auth_token(),
+            # The URL that the JavaScript code will open if it needs the user to
+            # authorize us to request a new Canvas access token.
+            "authUrl": self._request.route_url("canvas_api.authorize"),
             # Some debug information, currently used in the Gherkin tests.
             "debug": self._debug(),
             # The config object for the Hypothesis client.
--- a/lms/views/content_item_selection.py
+++ b/lms/views/content_item_selection.py
@@ -55,9 +55,6 @@ def content_item_selection(context, request):
 
     context.js_config.config.update(
         {
-            # The URL that the JavaScript code will open if it needs the user to
-            # authorize us to request a new access token.
-            "authUrl": request.route_url("canvas_api.authorize"),
             # The URL that we'll POST the ContentItemSelection form submission
             # (containing the user's selected document) to.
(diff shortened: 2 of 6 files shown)


lms/resources/_js_config.py
```diff
@@ -42,6 +42,9 @@ def config(self):
             # The auth token that the JavaScript code will use to authenticate
             # itself to our own backend's APIs.
             "authToken": self._auth_token(),
+            # The URL that the JavaScript code will open if it needs the user to
+            # authorize us to request a new access token.
+            "authUrl": self._request.route_url("canvas_api.authorize"),
```
jon-betts: Worth adding a note to mention this is Canvas specific?

lms/views/content_item_selection.py
```diff
@@ -53,9 +53,6 @@ def content_item_selection(context, request):
 
     context.js_config.config.update(
         {
-            # The URL that the JavaScript code will open if it needs the user to
-            # authorize us to request a new access token.
-            "authUrl": request.route_url("canvas_api.authorize"),
```
jon-betts: Nice to see this duplication go

tests/unit/lms/views/content_item_selection_test.py
```diff
@@ -8,16 +8,6 @@
...
-    def test_it_sets_the_authUrl_javascript_config_setting(
-        self, context, pyramid_request
-    ):
-        content_item_selection(context, pyramid_request)
-
-        assert (
-            context.js_config.config["authUrl"]
-            == "http://example.com/TEST_AUTHORIZE_URL"
-        )
-
```
jon-betts: :+1: Ditto here reducing duplication. Let's not do the same thing _very_ slightly different.

lms/resources/_js_config.py
seanh: Added a comment

PR #1520, opened 2020-02-29
Centralize JSConfig 5/n: Move is_canvas into LTILaunchResource

seanh:
Move `BasicLTILaunchViews.is_launched_by_canvas()` to
`LTILaunchResource.is_canvas` (a property).

This is necessary because future `JSConfig` code is going to need to
know whether the LMS is Canvas, and `JSConfig` can't access a
`BasicLTILaunchViews` method.

One thing that both `BasicLTILaunchViews` and `JSConfig` have access to
is the `LTILaunchResource` context, so they can both read
`context.is_canvas`.

Whether or not the LMS is Canvas also seems fitting as part of the
"context".

This is also in line with keeping views thin in general. One of the
reasons for keeping code out of views is that view code isn't reusable.
Context code is usable by the context itself, context helpers like
`JSConfig`, views, services that receive the context, etc.

### DIFF
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -82,10 +82,11 @@ def configure_module_item_caller(context, pyramid_request):
 class TestBasicLTILaunchViewsInit:
     """Unit tests for BasicLTILaunchViews.__init__()."""
 
+    @pytest.mark.usefixtures("is_canvas")
     def test_it_adds_report_submission_config_if_the_LMS_is_Canvas(
-        self, context, canvas_request
+        self, context, pyramid_request
     ):
-        BasicLTILaunchViews(context, canvas_request)
+        BasicLTILaunchViews(context, pyramid_request)
 
         assert context.js_config.config["submissionParams"] == {
             "h_username": context.h_user.username,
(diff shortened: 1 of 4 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -123,29 +125,31 @@ def test_it_doesnt_add_report_submission_config_if_lms_not_canvas(
         assert "submissionParams" not in context.js_config.config
 
     def test_it_configures_client_to_focus_on_user_if_in_canvas_and_param_set(
-        self, context, canvas_request, h_api
+        self, context, pyramid_request, h_api
     ):
+        context.is_canvas = True
```
jon-betts: This kind of feels like a step backwards because of the duplication, but on the other hand it's very clear about what makes this tick. So on balance I think I like the explicit nature of this in the class where some of these are canvas calls and some aren't.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -228,25 +234,31 @@ def view_caller(self, request):
 
 
 class TestCanvasFileBasicLTILaunch:
-    def test_it_configures_frontend(self, context, canvas_request):
-        canvas_file_basic_lti_launch_caller(context, canvas_request)
+    def test_it_configures_frontend(self, context, pyramid_request):
+        context.is_canvas = True
```
jon-betts: As this is the `TestCanvasFile...` I wonder if we should override the context fixture for this class to add this in.

tests/unit/lms/views/basic_lti_launch_test.py
seanh: I've replaced this with a fixture, so it how works like this:

```python
    @pytest.mark.usefixtures("is_canvas")
    def test_it_configures_client_to_focus_on_user_if_in_canvas_and_param_set(
        self, context, pyramid_request, h_api
    ):
        ...

...

@pytest.fixture
def is_canvas(context):
    """Set the LMS that launched us to Canvas."""
    context.is_canvas = True
```

Tests still have to explicitly declare that they want Canvas with `@pytest.mark.usefixtures("is_canvas")` but the definition of what constitutes "is canvas" is no longer duplicated, and more test set up is moved out of tests and into fixtures

tests/unit/lms/views/basic_lti_launch_test.py
seanh: Replaced this with a class-level `@pytest.mark.usefixtures("is_canvas")`

seanh:
> The changes here are solid, but I think we've missed a duplicated version of this in lms.views.helpers.frontend_app.py which has:

I'm gonna ignore this for now because an upcoming PR moves the entire `frontend_app.py` into `JSConfig` and removes this duplication, so I don't want to spend time refactoring `frontend_app.py` (to take a `context` param) now

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -222,22 +226,23 @@ def view_caller(self, request):
         return request.param
 
 
+@pytest.mark.usefixtures("is_canvas")
```
jon-betts: This is cool

jon-betts:
Oops merge boo boo. I'm rebasing and testing this again now.

PR #1523, opened 2020-02-29
Change from Prospector to PyLint

seanh:
Remove the Prospector meta-linter and change `make lint` to call pylint, pydocstyle and pycodestyle directly instead.

Prospector wasn't serving us well. It doesn't work with PyLint's multiprocessing support so it's a lot slower -- running PyLint on one CPU`*`. It also seems to mess with the linters it runs: for example it seems to have been disabling a lot of useful PyLint rules that we didn't tell it to disable, resulting in a lot of messages if you run PyLint directly.

What this PR does:

* Get rid of Prospector and just run the underlying tools directly. This is faster and gives us better control of the tools.

* Add PyLint, pycodestyle and pydocstyle configs

* I've also removed `make docstrings` and `make checkdocstrings`. These were being used to validate the docstrings, but PyLint's `docparams` extension can do at least some of that validation for us: it checks for missing or incorrect `:param`'s and `:raise`'s etc in docstrings. Just use that instead.

`*`: (Note that this PR doesn't actually make `make lint` any faster compared to master though. I think because it enables a lot more PyLint stuff so it's doing a lot more checking. And also because it has to run PyLint twice and then each of pycodestyle and pydocstyle separately as well.)

Usage
-----

* `make lint` will pylint the code and tests and also run pycodestyle and pydocstyle, and also create venvs for all these tools and install them, update the versions of the dependencies that pylint is checking against, etc. `make sure` also runs it.

* `tox -qe lint --run-command 'pylint lms'` will still allow tox to do everything, but just run pylint over the code only and not run everything else. Faster. Any of the other individual commands can be run this way as well. You can also lint just certain directories or files this way.

* `.tox/lint/bin/pylint lms` will run tox's copy of pylint but without any tox involvement at all. Even faster, but you lose all tox's help. E.g. you may be linting against old versions of the app's dependencies. Again, you can run any of the commands this way, and can use it to lint only certain files or dirs. Won't work unless you have run one of the two above commands first so that tox's copy of pylint exists. There's literally no simpler or faster way to run pylint than this, since you're literally just running it directly without tox or make.

* `/PATH/TO/YOUR/pylint ...` run your own installed copy of pylint (maybe just `pylint` if it's on your `$PATH`). No reason this won't work, and it will pick up our config.


seanh:
[List of tools that Prospector currently runs for us](https://prospector.readthedocs.io/en/master/supported_tools.html) (Prospector doesn't itself do any linting, it just runs other linters):

* PyLint
* [pycodestyle](https://pycodestyle.pycqa.org/en/latest/). I'm not sure how much we need this since we use Black. But I'd like to keep it because I don't know that _everything_ pycodestyle checks for is handled by Black
* [Pyflakes](https://launchpad.net/pyflakes). I don't we should run this -- PyLint is better. I don't know of anything Pyflakes would give us that we aren't getting from PyLint
* [McCabe](https://pypi.org/project/mccabe/) (cyclomatic complexity checker)
* [Dodgy](https://github.com/landscapeio/dodgy), tries to check for passwords and secrets in your code. Looks like it's in early stages of development. May not bother with this
* [pydocstyle](http://www.pydocstyle.org/en/5.0.1/), we do want this

We'll want to make `make lint` to run each of those, instead of running Prospector. Some of them may be runnable as PyLint plugins rather than as separate commands

lms/assets.py
```diff
@@ -133,20 +133,20 @@ def wrapper(context, request):
...
     """Read an asset bundle config from a file object."""
     parser = configparser.ConfigParser()
-    parser.read_file(fp)
+    parser.read_file(file_)
     return {k: aslist(v) for k, v in parser.items("bundles")}
 
 
 # Site assets
-ASSETS_VIEW = static_view("lms:../build", cache_max_age=None, use_subpath=True)
-ASSETS_VIEW = _add_cors_header(ASSETS_VIEW)
```
seanh: PyLint doesn't like it when you reassign a variable (`ASSETS_VIEW`) to a different type. So I just combined these two into one line

lms/assets.py
```diff
@@ -133,20 +133,20 @@ def wrapper(context, request):
...
+ASSETS_VIEW = _add_cors_header(
+    static_view("lms:../build", cache_max_age=None, use_subpath=True)
+)
 
 
 def includeme(config):
-    # TODO: figure out auto reload if we want it
```
seanh: TODO and FIXME comments trigger PyLint warnings. And I actually don't think it adds much value, to have these kinds of comments littering the code, so I just deleted them.

This means you can put TODOs and FIXMEs in your WIP code for things you intend to fix before merging it, and `make lint` will be failing until you remove them

lms/authentication/password_hash.py
```diff
@@ -34,7 +34,7 @@ def hash_password(password: str, salt: str = ""):
     if isinstance(password, str):
         password = password.encode("utf8")
 
-    if salt == "":
```
seanh: PyLint wants you to use the falsiness of False, None, empty string, empty list, etc, instead of comparing to them

lms/extensions/feature_flags/views/_predicates.py
```diff
@@ -20,8 +20,8 @@ def __init__(self, feature_flag, _config):
         :arg feature_flag: the feature flag that this predicate requires,
             for example ``"my_feature"``
         :type feature_flag: str
-        :arg config: the Pyramid config object
-        :type config: pyramid.config.Configurator
+        :arg _config: the Pyramid config object
+        :type _config: pyramid.config.Configurator
```
seanh: PyLint pointed out that the param name in the `:arg` was wrong

lms/services/_helpers/canvas_api.py
```diff
@@ -160,6 +160,11 @@ def validated_response(request, schema=None, access_token=None):
           and the response had a 2xx HTTP status.
         :type schema: a subclass of :cls:`lms.validation.RequestsResponseSchema`
 
+        :arg access_token: the access token to use in the
+            "Authorization: Bearer <ACCESS_TOKEN>" header to authorize the
+            request
+        :type access_token: str
```
seanh: PyLint pointed out that this argument wasn't documented

lms/services/lti_outcomes.py
```diff
@@ -61,6 +61,8 @@ def record_result(  # pylint:disable=no-self-use
             Defined as required by the LTI spec but is optional in Canvas if
             an `lti_launch_url` is set.
         :param pre_record_hook: Hook to allow modification of the request
+
+        :raise TypeError: if the given pre_record_hook returns a non-dict
```
seanh: PyLint points out when a method raises an exception and you haven't documented that with a `:raise`. I doubt that it spots all exceptions that you might raise (even ones you fail to catch in methods you call directly or indirectly) but it finds some and that's useful

lms/validation/authentication/_helpers/_jwt.py
```diff
@@ -18,11 +18,10 @@ def decode_jwt(jwt_str, secret):
...
 
-    :raise lms.validation.authentication.ExpiredJWTError: if decoding fails because
-        the JWT's timestamp has expired
-    :raise lms.validation.authentication.InvalidJWTError: if decoding fails for any
-        other reason (for example ``jwt_str`` is invalid, or ``secret`` is
-        wrong)
+    :raise ExpiredJWTError: if decoding fails because the JWT's timestamp has
+        expired
+    :raise InvalidJWTError: if decoding fails for any other reason (for example
+        ``jwt_str`` is invalid, or ``secret`` is wrong)
```
seanh: PyLint doesn't know what you're talking about if you do `:raise lms.validation.authentication.InvalidJWTError:`. I just changed it to the simpler version that PyLint does understand

pylintrc.tests
```diff
@@ -0,0 +1,95 @@
+[MASTER]
```
seanh: We need some different PyLint config for tests, that we wouldn't want to use for main code. I don't think a pylintrc file can inherit from another

tests/functional/api/basic_lti_launch.py
```diff
@@ -68,7 +68,7 @@ def oauth_sign_params(cls, url, params):
     # Fixtures ------------------------------------------------------------- #
 
     @pytest.fixture(autouse=True)
-    def application_instance(self, db_session, app):
```
seanh: PyLint is now finding unused fixtures in test :+1:

tests/unit/lms/authentication/_lti_policy_test.py
```diff
@@ -27,7 +27,7 @@ def test_it_returns_None_if_theres_no_lti_user(self, method_name, pyramid_reques
         policy = LTIAuthenticationPolicy()
         method = getattr(policy, method_name)
 
-        assert policy.authenticated_userid(pyramid_request) is None
+        assert method(pyramid_request) is None
```
seanh: PyLint pointed out that the `method` fixture was unused. It turned out the test was wrong: it was meant to be calling `method` but wasn't

tox.ini
```diff
@@ -10,6 +10,37 @@ requires =
...
+ignore =
+    # Disable pycodestyle errors and warnings that we don't care about because
+    # Black formats our code for us.
+    E203,  # Whitespace before ':',
+    E231,  # Missing whitespace after ',',
+    E501,  # Line too long,
+    W503,  # Line break before binary operator,
+
+    # Bare except. PyLint finds these for us so we don't need pycodestyle to.
+    E722,
```
seanh: [pycodestyle supports reading config from `setup.cfg` or `tox.ini`](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration) and we don't have a `setup.cfg`, so our pycodestyle is in `tox.ini`. This config still works if pycodestyle is run without tox: pycodestyle itself reads the `[pycodestyle]` section in the `tox.ini` file

tox.ini
```diff
@@ -10,6 +10,37 @@ requires =
...
+    D202,
+
+    # "1 blank line required before class docstring" conflicts with another
+    # pydocstyle rule D211 "No blank lines allowed before class docstring".
+    D203,
+
+    # Multi-line docstring summary should start at the first line, this
+    # conflicts with another pycodestyle rule "Multi-line docstring summary
+    # should start at the second line".
+    D212,
```
seanh: pydocstyle actually does [support its own separate config file](http://www.pydocstyle.org/en/5.0.1/usage.html#configuration-files) but I just put its config here in `tox.ini` so it's next to pycodestyle's. Again: pydocstyle reads this file directly, it doesn't have to be run in tox

tox.ini
```diff
@@ -78,8 +111,10 @@ commands =
     tests: coverage report
     functests: pytest {posargs:tests/functional/}
     bddtests: behave {posargs:tests/bdd/}
-    lint: prospector
-    lint: prospector tests
+    lint: pylint lms
+    lint: pylint --rcfile=pylintrc.tests tests/unit tests/functional
```
seanh: It's necessary to run two separate pylint commands if we want to apply different linting rules to the code and tests

Makefile
```diff
@@ -17,8 +17,6 @@ help:
 	@echo "make test              Run the unit tests"
 	@echo "make functests         Run the functional tests"
 	@echo "make bddtests          Run the gherkin tests"
-	@echo "make docstrings        View all the docstrings locally as HTML"
-	@echo "make checkdocstrings   Crash if building the docstrings fails"
```
seanh: PyLint's `docparams` extension finds missing or incorrect `:param`'s, `:raise`'s etc in docstrings. Just use that instead of using Sphinx to check the docstrings for us.

jon-betts:
Is there any chance the changes to pylint and the fixes could be separated into different commits? This is kind of hard to review as it stands.

Is there anything that is fixed in this PR that would break the previous linting regime? Could these be two separate PRs? Lint fixes first, lint changes second? I realise it would be difficult to develop in that order, but it can be committed that way I'd think.

Makefile
jon-betts: +100 to the above

Makefile
```diff
@@ -17,8 +17,6 @@ help:
 	@echo "make test              Run the unit tests"
 	@echo "make functests         Run the functional tests"
 	@echo "make bddtests          Run the gherkin tests"
-	@echo "make docstrings        View all the docstrings locally as HTML"
```
jon-betts: Why has this gone away? I haven't used this, but it doesn't seem directly related.

.pylintrc
```diff
@@ -0,0 +1,54 @@
...
+    missing-type-doc,
+
+    # We use isort to sort and group our imports, so we don't need PyLint to
+    # check them for us.
+    ungrouped-imports,
+
+    # We use Black to format our code automatically, so we don't need PyLint to
+    # check formatting for us.
+    bad-continuation,
+    line-too-long,
```
jon-betts: We have a few more exceptions here than we do in our cookie cutters. I wonder if we want to look at removing one or two in future to get them inline. Certainly better than trying to fix this all now.

It's already plenty better than before.

PR #1533, opened 2020-03-04
Centralize JSConfig 1/n: Remove the lmsName JavaScript setting

seanh:
This setting is only ever set to "Canvas". Simplify both the backend and
frontend code by just using "Canvas" as a hardcoded string and removing
the setting.


lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -137,7 +136,7 @@ describe('FilePickerApp', () => {
   it('shows LMS file dialog when "Select PDF from <LMS Name>" is clicked', () => {
```
jon-betts: The description here (`<LMS Name>`) looks like it should be just "Canvas".

seanh: Fixed

PR #1534, opened 2020-03-04
Add `make pylint`

seanh:
Add a new `make pylint` command that runs pylint, pydocstyle and
pycodestyle directly, instead of running prospector.

The old, prospector-based `make lint` command still exists and is still
used by `make sure` and by CI. `make pylint` isn't actually used by
anything yet.

`make pylint` is currently failing because of lots of unfixed pylint
warnings in the code and tests.

In the future, once the tests are fixed, `make lint` will be removed and
`make pylint` will become `make lint`.

The `pylintrc` file had to be named `pylintrc.code` to prevent
prospector from picking it up (which would cause prospector to hang
forever). It'll be renamed to just `pylintrc` once prospector is gone.

Usage
-----

* `make pylint` will pylint the code and tests and also run pycodestyle
  and pydocstyle, and also create venvs for all these tools and install
  them, update the versions of the dependencies that pylint is checking
  against, etc. This is what CI and `make sure` will run in future.

* `tox -qe pylint --run-command 'pylint lms'` will still allow tox to do
  everything, but just run pylint over the code only and not run
  everything else. Faster. Any of the other individual commands can be
  run this way as well. You can also lint just certain directories or
  files this way.

* `.tox/pylint/bin/pylint lms` will run tox's copy of pylint but without
  any tox involvement at all. Even faster, but you lose all tox's help.
  E.g. you may be linting against old versions of the app's
  dependencies.  Again, you can run any of the commands this way, and
  can use it to lint only certain files or dirs. Won't work unless you
  have run one of the two above commands first so that tox's copy of
  pylint exists. There's literally no simpler or faster way to run
  pylint than this, since you're literally just running it directly
  without tox or make.

* `/PATH/TO/YOUR/pylint --rcfile=pylintrc.code ...` run your own
  installed copy of pylint (maybe just `pylint` if it's on your
  `$PATH`). No reason this won't work, and it will pick up our config.

### DIFF
--- a/pylintrc.tests
+++ b/pylintrc.tests
@@ -1,0 +1,80 @@
+[MASTER]
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=0
+
+load-plugins=pylint.extensions.bad_builtin,
+             pylint.extensions.check_elif,
+             pylint.extensions.comparetozero,
+             pylint.extensions.docparams,
+             pylint.extensions.emptystring,
+             pylint.extensions.overlapping_exceptions,
+             pylint.extensions.redefined_variable_type,
+
+[MESSAGES CONTROL]
+disable=
+    # Docstrings are encouraged but we don't want to enforce that everything
+    # must have a docstring.
+    missing-docstring,
+
+    # We don't always want to have to put a `:return:` in a docstring.
+    missing-return-doc,
+
+    # We don't always want to have to put an `:rtype:` in a docstring.
+    missing-return-type-doc,
+
+    # We don't want to have to document the type of every parameter with a
+    # `:type:` in the docstring.
+    missing-type-doc,
+
+    # We use isort to sort and group our imports, so we don't need PyLint to
+    # check them for us.
+    ungrouped-imports,
+
+    # We use Black to format our code automatically, so we don't need PyLint to
+    # check formatting for us.
+    bad-continuation,
+    line-too-long,
+
+    # Because of how pytest fixtures work it's frequently necessary for
+    # parameters to redefine outer names.
+    redefined-outer-name,
+
+    # Lots of test methods don't use self, but we still want to group our tests
+    # into classes.
+    no-self-use,
+    too-few-public-methods,
+    too-many-public-methods,
+
+    too-many-arguments,
+
+# Just disable PyLint's name style checking for the tests, because we
+# frequently use lots of argument names that don't conform.
+# For example we frequently create pytest fixtures that aren't named in
+# snake_case, such as a fixture that returns a mock of the FooBar class would
+# be named FooBar in CamelCase.
+argument-naming-style=any
+class-naming-style=any
+function-naming-style=any
+method-naming-style=any
+
+good-names=
+    # PyLint's default good names.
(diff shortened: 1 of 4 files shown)


pylintrc.tests
```diff
@@ -0,0 +1,80 @@
+[MASTER]
```
seanh: We need a separate pylint config file for tests because we need different rules for tests. Unfortunately there's no way for a single pylintrc file to apply different rules to different files/directories. Nor can pylintrc files inherit from each other

tox.ini
```diff
@@ -10,6 +10,37 @@ requires =
...
+ignore =
+    # Disable pycodestyle errors and warnings that we don't care about because
+    # Black formats our code for us.
+    E203,  # Whitespace before ':',
+    E231,  # Missing whitespace after ',',
+    E501,  # Line too long,
+    W503,  # Line break before binary operator,
+
+    # Bare except. PyLint finds these for us so we don't need pycodestyle to.
+    E722,
```
seanh: pycodestyle's config can either go in `setup.cfg` or in `tox.ini`, it doesn't have its own separate config file, and we don't have a `setup.cfg`, so it's in here. This doesn't mean it has to be run in tox though -- pycodestyle run outside of tox style reads the `[pycodestyle]` in `tox.ini`

tox.ini
```diff
@@ -10,6 +10,37 @@ requires =
...
+    D202,
+
+    # "1 blank line required before class docstring" conflicts with another
+    # pydocstyle rule D211 "No blank lines allowed before class docstring".
+    D203,
+
+    # Multi-line docstring summary should start at the first line, this
+    # conflicts with another pycodestyle rule "Multi-line docstring summary
+    # should start at the second line".
+    D212,
```
seanh: The pydocstyle config is in here so that it's in the same place as the pycodestyle config

pylintrc.tests
jon-betts: These two aren't following the pattern we established in [h-cookiecutter-pypackage](https://github.com/hypothesis/h-cookiecutter-pypackage/) which is to have a `.pylintrc` and `tests/.pylintrc`. I think the idea was:

 * It's the default filename, and so is unsurprising / we didn't have much imagination
 * Running pylint manually from either directory has a descent chance of picking up the right config

I think it would be nicer to have everything going in the same direction if possible.

The comments say that `prospector` will pick this up and run `pylint` if it's about, but I think we could avoid that by adding `--without-tool pylint`, or positively listing the tools we do want to run (which might be nice anyway).

tox.ini
jon-betts: Is there any harm in having a `setup.cfg`? Having two files might make each easier to read.

seanh:
I don't think we're gonna need this PR...

pylintrc.tests
seanh: I've changed these to `.pylintrc` and `tests/.pylintrc` in [the other pr](https://github.com/hypothesis/lms/pull/1523)

tox.ini
seanh: I've moved these into `setup.cfg` in [the other pr](https://github.com/hypothesis/lms/pull/1523)

PR #1535, opened 2020-03-04
Fix all pylint warnings

seanh:
If we merge https://github.com/hypothesis/lms/pull/1534 then rebase this, then you can verify it and play with it by running `make pylint`.

After this is merged a follow-up PR will change `make lint` to do what `make pylint` does and run pylint on CI

### DIFF
--- a/lms/validation/authentication/_oauth.py
+++ b/lms/validation/authentication/_oauth.py
@@ -101,12 +101,12 @@ class CanvasOAuthCallbackSchema(PyramidRequestSchema):
         Return the :class:`lms.values.LTIUser` authenticated by the current
         request's ``state`` query parameter.
 
-        :raise lms.validation.MissingStateParamError: if the request has no
-            ``state`` query parameter
-        :raise lms.validation.ExpiredStateParamError: if the request's
-            ``state`` param has expired
-        :raise lms.validation.InvalidStateParamError: if the request's
-            ``state`` param is invalid
+        :raise MissingStateParamError: if the request has no ``state`` query
+            parameter
+        :raise ExpiredStateParamError: if the request's ``state`` param has
+            expired
+        :raise InvalidStateParamError: if the request's ``state`` param is
+            invalid
         :rtype: str
         """
         request = self.context["request"]
@@ -149,7 +149,7 @@ class CanvasAccessTokenResponseSchema(RequestsResponseSchema):
 
     @marshmallow.validates("expires_in")
     def validate_quantity(self, expires_in):  # pylint:disable=no-self-use
-        if not expires_in > 0:
+        if expires_in <= 0:
             raise marshmallow.ValidationError("expires_in must be greater than 0")
 
 
--- a/tests/bdd/steps/the_response.py
+++ b/tests/bdd/steps/the_response.py
@@ -2,7 +2,7 @@
 
 import re
 
-from behave import then
+from behave import then  # pylint:disable=no-name-in-module
 
 from tests.bdd.step_context import StepContext
 from tests.bdd.steps.the_url import TheURL
@@ -13,6 +13,7 @@ class WebTestResponse(StepContext):
     ephemeral = True
 
     def __init__(self, response, **kwargs):
+        super().__init__(**kwargs)
         self.response = response
 
     def get_body(self):
@@ -52,7 +53,7 @@ def the_response_body_matches(context, regex):
 
 
 @then("the response body does not match '{regex}'")
-def the_response_body_matches(context, regex):
+def the_response_body_does_not_match(context, regex):
     body = context.the_response.get_body()
 
     assert not re.compile(regex).search(body), f'The body does not match "{regex}"'
--- a/tests/unit/lms/authentication/_lti_policy_test.py
+++ b/tests/unit/lms/authentication/_lti_policy_test.py
@@ -14,7 +14,7 @@ class TestLTIAuthenticationPolicy:
         policy = LTIAuthenticationPolicy()
         method = getattr(policy, method_name)
 
-        userid = policy.authenticated_userid(pyramid_request)
+        userid = method(pyramid_request)
 
         _helpers.authenticated_userid.assert_called_once_with(pyramid_request.lti_user)
         assert userid == _helpers.authenticated_userid.return_value
@@ -27,7 +27,7 @@ class TestLTIAuthenticationPolicy:
         policy = LTIAuthenticationPolicy()
         method = getattr(policy, method_name)
 
-        assert policy.authenticated_userid(pyramid_request) is None
+        assert method(pyramid_request) is None
 
     def test_effective_principals_when_theres_an_lti_user(
         self, pyramid_request, _helpers
--- a/tests/unit/lms/resources/lti_launch_test.py
+++ b/tests/unit/lms/resources/lti_launch_test.py
@@ -156,6 +156,7 @@ class TestLTILaunchResource:
             HTTPBadRequest,
             match='Required parameter "tool_consumer_instance_guid" missing from LTI params',
         ):
+            # pylint:disable=expression-not-assigned
             LTILaunchResource(pyramid_request).h_authority_provided_id
 
     def test_h_authority_provided_id_raises_if_theres_no_context_id(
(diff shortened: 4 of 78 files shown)


lms/validation/authentication/_oauth.py
```diff
@@ -149,7 +149,7 @@ class CanvasAccessTokenResponseSchema(RequestsResponseSchema):
 
     @marshmallow.validates("expires_in")
     def validate_quantity(self, expires_in):  # pylint:disable=no-self-use
-        if not expires_in > 0:
```
jon-betts: Glorious

tests/bdd/steps/the_response.py
```diff
@@ -52,7 +53,7 @@ def the_response_body_matches(context, regex):
 
 
 @then("the response body does not match '{regex}'")
-def the_response_body_matches(context, regex):
+def the_response_body_does_not_match(context, regex):
```
jon-betts: How on earth did this come up in linting? Or did you just spot this yourself?

tests/unit/lms/authentication/_lti_policy_test.py
```diff
@@ -14,7 +14,7 @@ def test_it_returns_the_lti_userid(self, method_name, pyramid_request, _helpers)
         policy = LTIAuthenticationPolicy()
         method = getattr(policy, method_name)
 
-        userid = policy.authenticated_userid(pyramid_request)
+        userid = method(pyramid_request)
```
jon-betts: Oops. Linting is good I suppose

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -286,6 +291,7 @@ def test_h_user_raises_if_user_id_is_missing(self, pyramid_request):
         with pytest.raises(
             HTTPBadRequest, match='Required parameter "user_id" missing from LTI params'
         ):
+            # pylint:disable=expression-not-assigned
```
jon-betts: Is the approach here to always do it at the line level? In other tests I think we've put very frequently occurring ones at the top of the file.

tests/unit/lms/validation/_exceptions_test.py
```diff
@@ -34,10 +34,9 @@ def test_we_produce_a_string_message(self, redirect):
         assert redirect.detail == Any.string.matching(".*field_name.*error_name.*")
 
     @pytest.mark.parametrize(
-        "description,messages",
-        [("string", "a string"), ("list", []), ("dict of string", {"a": "b"})],
+        "messages", ["a string", [], {"a": "b"}],
     )
-    def test_is_requires_well_formatted_messages(self, description, messages):
+    def test_it_requires_well_formatted_messages(self, messages):
```
jon-betts: The above does stringify nicely when running the tests. Would this be worth changing to `_description`?

tests/unit/lms/views/test_reports.py
```diff
@@ -29,7 +29,11 @@ def build_ai_from_pair(pair):
                 pair[0], pair[1], None, None, None
             )
 
-        app_instances = list(map(build_ai_from_pair, zip(test_urls, test_emails)))
+        app_instances = list(
+            map(  # pylint:disable=bad-builtin
+                build_ai_from_pair, zip(test_urls, test_emails)
+            )
+        )
```
jon-betts: Is this different from?

```[build_ai_from_pair(pair) for pair in zip(test_urls, test_emails)]```

File: lms/validation/authentication/_oauth.py (hunk 1 above)
Comment (seanh) [reply lms/validation/authentication/_oauth.py]: Yeah, this was pretty good

File: tests/bdd/steps/the_response.py (hunk 2 above)
Comment (seanh) [reply tests/bdd/steps/the_response.py]: Linting noticed that two test methods has the exact same name. What that was doing to the tests, I don't know

File: tests/unit/lms/resources/lti_launch_test.py (hunk 4 above)
Comment (seanh) [reply tests/unit/lms/resources/lti_launch_test.py]: I tend to be as minimal as possible about the scope of the ignore, though sometimes it's sensible to make an exception to that.

I was too lazy to make the real change here which is that these properties that may raise should probably be methods, not properties. I think pylint may be indirectly pointing out to us that it's not good for properties to raise.

File: tests/unit/lms/validation/_exceptions_test.py (hunk 5 above)
Comment (seanh) [reply tests/unit/lms/validation/_exceptions_test.py]: pytest actually has a way of adding descriptions (which it calls "ids") to the tests, when you want these to actually be used as the test names in pytest so that they show up in pytest's output (as opposed to just putting code comments next to each parameter, which we often do, but obviously that doesn't do anything to pytest's output).

I'll change it to use pytest ids properly instead.

The way it works is that normally pytest auto-generates an id for each parametrized test case based on the param names and values, but it also provides a few different ways for you to set the ids yourself if you want to:

You can pass a list of id strings which must be in the same order as the test cases in `testdata`:

```python
@pytest.mark.parametrize("a,b,expected", testdata, ids=["forward", "backward"])
```

Alternatively you can use `pytest.param()` so that the ids are grouped together with the test data:

```python
@pytest.mark.parametrize(
    "a,b,expected",
    [
        pytest.param(
            datetime(2001, 12, 12), datetime(2001, 12, 11), timedelta(1), id="forward"
        ),
        pytest.param(
            datetime(2001, 12, 11), datetime(2001, 12, 12), timedelta(-1), id="backward"
        ),
    ],
)
```

Another alternative: you provide a function that generates the ID for each test case:

```python
def idfn(val):
    if isinstance(val, (datetime,)):
        # note this wouldn't show any hours/minutes/seconds
        return val.strftime("%Y%m%d")

@pytest.mark.parametrize("a,b,expected", testdata, ids=idfn)
def test_timedistance_v2(a, b, expected):
    ...
```

* * * 

Parametrize is actually a lot more powerful than I usually remember.
See <https://docs.pytest.org/en/latest/parametrize.html> and
<https://docs.pytest.org/en/latest/example/parametrize.html#deferring-the-setup-of-parametrized-resources>. Some examples:

Apply a mark to just one of the parametrized instances of a test. Here the
`(6*9, 42)` version of the test is marked as expected to fail:

```python
@pytest.mark.parametrize(
    "test_input,expected",
    [("3+5", 8), ("2+4", 6), pytest.param("6*9", 42, marks=pytest.mark.xfail)],
)
```

Similarly you can mark just one test case as expected to raise a specific
error:

```python
@pytest.mark.parametrize(
    "example_input,expectation",
    [
        (3, does_not_raise()),
        (2, does_not_raise()),
        (1, does_not_raise()),
        (0, pytest.raises(ZeroDivisionError)),
    ],
)
def test_division(example_input, expectation):
    ...
```

You can defer the parametrization of a test function, which normally happens at
test collection time, until test run time with `indirect=True` or
`indirect=["x"]` to defer the generation of the parametrized `x` argument only:
https://docs.pytest.org/en/latest/example/parametrize.html#deferring-the-setup-of-parametrized-resources

You can get pytest to generate the cross-product, e.g. this generates every
combination of the different values for x and y:

```python
@pytest.mark.parametrize("x", [0, 1])
@pytest.mark.parametrize("y", [2, 3])
def test_foo(x, y):
    pass
```

There's something called `pytest_generate_tests` that's for when you need to do
more powerful test generation: https://docs.pytest.org/en/latest/parametrize.html#basic-pytest-generate-tests-example

seanh: I've changed this to:

```python
    @pytest.mark.parametrize(
        "messages",
        [
            pytest.param("a string", id="string"),
            pytest.param([], id="list"),
            pytest.param({"a": "b"}, id="dict of string"),
        ],
    )
    def test_it_requires_well_formatted_messages(self, messages):
        with pytest.raises(ValueError):
            LTIToolRedirect("http://example.com", messages)
```

Test output now:

```terminal
> .tox/tests/bin/pytest tests/unit/lms/validation/_exceptions_test.py -vv
...
tests/unit/lms/validation/_exceptions_test.py::TestLTIToolRedirect::test_it_requires_well_formatted_messages[string] PASSED [ 71%]
tests/unit/lms/validation/_exceptions_test.py::TestLTIToolRedirect::test_it_requires_well_formatted_messages[list] PASSED [ 85%]
tests/unit/lms/validation/_exceptions_test.py::TestLTIToolRedirect::test_it_requires_well_formatted_messages[dict of string] PASSED [100%]
...
```

tests/unit/lms/views/test_reports.py
seanh: I don't think so, I've replaced it with that

PR #1539, opened 2020-03-07
Stub out readyToReceive RPC method

LMS007:
New method to allow delaying outgoing RPC requests to the client until its ready to receive them. Implementation details will need to happen after client changes, therefore this method is only a stub.

relates to https://github.com/hypothesis/client/pull/1885


LMS007:
see https://github.com/hypothesis/client/pull/1961

PR #1540, opened 2020-03-08
wip - Add readyToReceive method

LMS007:
readyToReceive calls setSidebarResolved which saves the current incoming frame event (from the client) and resolves a promise. This delays any outgoing RPC requests until the client sends the `readyToReceive` message first to prevent outgoing RCP requests from arriving too early and causing a timeout.


LMS007:
see https://github.com/hypothesis/client/pull/1961

PR #1545, opened 2020-03-09
Add a timeout to Canvas API requests

seanh:
Looking at New Relic, some Canvas API's seem to take up to a second to
respond sometimes, so I've made the timeout pretty generous at 3
seconds.

### DIFF
--- a/lms/services/_helpers/canvas_api.py
+++ b/lms/services/_helpers/canvas_api.py
@@ -179,7 +179,7 @@ class CanvasAPIHelper:
             request.headers["Authorization"] = f"Bearer {access_token}"
 
         try:
-            response = requests.Session().send(request)
+            response = requests.Session().send(request, timeout=3)
             response.raise_for_status()
         except RequestException as err:
             CanvasAPIError.raise_from(err)
--- a/lms/services/_helpers/canvas_api.py
+++ b/lms/services/_helpers/canvas_api.py
@@ -179,7 +179,7 @@ class CanvasAPIHelper:
             request.headers["Authorization"] = f"Bearer {access_token}"
 
         try:
-            response = requests.Session().send(request, timeout=3)
+            response = requests.Session().send(request, timeout=9)
             response.raise_for_status()
         except RequestException as err:
             CanvasAPIError.raise_from(err)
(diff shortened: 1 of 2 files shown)


jon-betts:
Do we know what our gunicorn timeout is and how many of these requests we make in a single call? If we only make one, and the gunicorn timeout is 30 seconds, we can safely set this to 15 seconds or so.

I'd very much err on the side of picking the longest timeout that gets the job done, as the closer we trim the time down by harder the long tail of probabilities will start to exclude people who were previously working.

Long story short, people who's requests are taking 30 seconds are already broken right now, but people who's requests take 3.1 seconds aren't. By changing this value we don't really help the former, but we do hurt the latter.

seanh:
I don't see us setting gunicorn's worker timeout anywhere, so I think we're getting the [default of 30 secs](https://docs.gunicorn.org/en/stable/settings.html#timeout).

Generally our proxy API for the Canvas API is a 1:1 mapping from a request to our proxy API to a request to a Canvas API endpoint _but_ with automatic refresh and retry, so it's up to three Canvas API requests per request (first failed Canvas API request, refresh token request, then re-try the Canvas API request again). The grading-related proxy APIs violate this 1:1 design, using different logic and sometimes calling multiple different LMS APIs in one request, but at a glance they don't look like they send more than two requests per request.

So I think we could get away with, say, a 9 second request timeout, putting us comfortably below the 30 second worker timeout at 27 seconds.

In New Relic I can't easily find any LMS API response times slower than about 1.1 seconds, so 9 seconds should give us plenty of headroom, although New Relic doesn't seem to be able to tell us "slowest ever response time" (in a given time period), only average response time: https://rpm.newrelic.com/accounts/1385283/applications/76030703/externals?tw%5Bend%5D=1583775535&tw%5Bstart%5D=1583170735#id=5b2245787465726e616c2f6d6f6f646c652e6175612e616d2f616c6c222c22225d&sort_by=average_call_time

It's worth keeping in mind that a request blocking a gunicorm worker for over 27 seconds is likely to cause queuing and timing out of _other_ requests that would have been fine, because it's blocking up finite resources. So by allowing really long timeouts because we don't want to cause those slow requests to fail, we may end up allowing the slow requests to cause other requests to randomly fail. We've had a bad case of that happening with h in the past.

I don't _know_ of any evidence that LMS has request queuing issues currently though.

I've changed the timeout from 3 to 9 seconds, let's try that.

seanh:
If a `requests` timeout happens I think this will catch it and send an error response back to our frontend, which will show an error dialog with a <kbd>Try again</kbd> button:

https://github.com/hypothesis/lms/blob/23ccd5427d9f52cea337ddfdfdba72599c4e4347/lms/services/_helpers/canvas_api.py#L181-L185

PR #1546, opened 2020-03-09
Move file picker settings into JSConfig

seanh:
This moves a bunch of file picker-related settings into `JSConfig`. Specifically, they all move into `JSConfig.enable_content_item_selection_mode()` ("content item selection mode" and "file picker" are kind of two different names for the same thing). These settings are all used by the UI that's shown to teachers when they're configuring an assignment, that lets them paste in a URL or select a file from Google Drive or Canvas:

* `enableLmsFilePicker` is whether or not to show the <kbd>Select PDF from Canvas</kbd> button

* `courseId` is needed when `enableLmsFilePicker` is `True`. The course ID is used to request the list of files for that course from the Canvas API

* `formAction` is the URL to submit the form to, once the document (whether a URL, Google Drive file or Canvas file) has been selected. In Canvas this is a Canvas URL, we submit the document back to Canvas for saving. In non-Canvas LMS's this is the URL of our own `configure_module_item()` view

* `formFields` is the fields that will be added to the form that we submit to either Canvas or ourselves. This is a completely different set of fields when we're submitting to Canvas, vs when we're submitting to ourselves.

* `ltiLaunchUrl` is needed when submitting the chosen document to Canvas rather than to ourselves. It's needed to form the value of one of the form fields we have to submit

* `googleClientId` and `googleDeveloperKey` are needed for the Google Drive picker

* `customCanvasApiDomain` and `lmsUrl` are also used by the Google Drive picker. We have to provide the domain of the top-level frame (the LMS's frame) to Google Picker otherwise Google Picker refuses to let itself be iframed. But we have no way of getting that top-level domain. So the backend needs to pass the top-level domain to the frontend. The `custom_canvas_api_domain` launch param is the most reliable way to get this top-level domain but, failing that (in non-Canvas LMS's), we use the `lms_url` from our `application_instances` table in the DB.

Previously all these fields were set by `BasicLTILaunchViews.unconfigured_basic_lti_launch()` (the view that's used when a teacher launches an unconfigured assignment in a non-Canvas LMS). And the same fields were also set differently by `content_item_selection()` (the view that's used when configuring an assignment in Canvas).

Now, both of those views call `enable_content_item_selection_mode()` and it does all the work. But the views do need to pass the `form_action` and `form_fields` to `enable_content_item_selection_mode()` as those differ between the two views.


lms/resources/_js_config.py
```diff
@@ -93,14 +94,42 @@ def add_document_url(self, document_url):
         self.config["urls"]["via_url"] = via_url(self._request, document_url)
         self._add_canvas_submission_params(document_url=document_url)
 
-    def enable_content_item_selection_mode(self):
+    def enable_content_item_selection_mode(self, form_action, form_fields):
```
jon-betts: Could do with documenting these parameters.

lms/resources/_js_config.py
```diff
@@ -93,14 +94,42 @@ def add_document_url(self, document_url):
...
+                # needs to pass this URL (which is the URL of the top-most page) to Google
+                # Picker, otherwise Picker refuses to launch inside an iframe.
+                "customCanvasApiDomain": self._context.custom_canvas_api_domain,
+                "lmsUrl": self._context.lms_url,
+                # The "content item selection" that we submit to Canvas's
+                # content_item_return_url is actually an LTI launch URL with
+                # the selected document URL or file_id as a query parameter. To
+                # construct these launch URLs our JavaScript code needs the
+                # base URL of our LTI launch endpoint.
+                "ltiLaunchUrl": self._request.route_url("lti_launches"),
```
jon-betts: It would be good to break up the items which are:

 * For canvas only
 * For google only
 * Shared

For example, we set `courseId` as it's only needed for Canvas below, but it would seem there's no reason to set `ltiLaunchUrl` if we aren't going to enable it.

I think a good end state would be to have `google.clientId` and `google.developerKey` as a nested dict, and the same for Canvas, but this would disrupt the tests and make this a very big diff. In the mean time, we could set the stage by building the dict in three parts. Intialise with the common bits, update with the Google bits and then optionally update with the Canvas bits.

lms/resources/_js_config.py
```diff
@@ -185,6 +214,24 @@ def _auth_token(self):
             self._request.lti_user
         )
 
+    def _canvas_files_available(self):
+        """Return True if the Canvas Files API is available to this request."""
+
+        if not self._is_canvas():
```
jon-betts:
```suggestion
        if not self.context.is_canvas():
```

lms/resources/_js_config.py
```diff
@@ -257,3 +304,10 @@ def _hypothesis_client(self):
                 }
             ]
         }
+
+    def _is_canvas(self):
+        """Return True if Canvas is the LMS that launched us."""
+        return (
+            self._request.params.get("tool_consumer_info_product_family_code")
+            == "canvas"
+        )
```
jon-betts: I don't think this is needed as it looks like a dupe of `context.is_canvas()`

lms/views/basic_lti_launch.py
```diff
@@ -159,49 +159,23 @@ def unconfigured_basic_lti_launch(self):
...
         will then be DB-configured launches rather than unconfigured.
         """
-        self.context.js_config.enable_content_item_selection_mode()
-
-        params = self._extract_lti_params(self.request)
+        form_fields = {
+            param: value
+            for param, value in self.request.params.items()
+            if param not in ["oauth_nonce", "oauth_timestamp", "oauth_signature"]
+        }
```
jon-betts: I think in-lining this here is really good. It stops you having to immediately jump to another function upon reading this one.

It's possible to do this without this becoming overwhelming because so much has been extracted below. At this point you could probably get away with just using this dict in the call to `enable_content_item_selection_mode()` as we do in the `content_item_selection` view.

The comment below about why we are removing these particular parameters was helpful though (because they aren't re-usable)

lms/resources/_js_config.py
jon-betts: BTW. Good opportunity here to make this easier to understand by renaming the vars later. A lot of these option names are really hostile to understanding what it is.

lms/views/content_item_selection.py
```diff
@@ -51,45 +49,12 @@ def content_item_selection(context, request):
...
-    # For Canvas Picker support our JavaScript needs the ID of the Canvas
-    # course, as this is a required param of the API it'll call to get the list
-    # of files in the course.
-    if helpers.canvas_files_available(request):
-        context.js_config.config["enableLmsFilePicker"] = True
-        context.js_config.config["courseId"] = request.params["custom_canvas_course_id"]
-    else:
-        context.js_config.config["enableLmsFilePicker"] = False
-
     return {}
```
jon-betts: Not for now, but this really seems like it has a lot in common with the LTI launches. It even states it is one in the comments above. It even has (most of) the h upsert stuff from the other launch views.

If that could be unified, I suspect we could remove the many separate calls to this stuff, which will help us later on with this course stuff.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -241,54 +240,27 @@ def test_it_adds_the_document_url(self, context, pyramid_request):
...
+                "user_id": "TEST_USER_ID",
                 "resource_link_id": "TEST_RESOURCE_LINK_ID",
-                "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
                 "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
-                "user_id": "TEST_USER_ID",
+                "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
                 "context_id": "TEST_CONTEXT_ID",
-                "some_random_rubbish": "SOME_RANDOM_RUBBISH",
-            }
+                "authorization": bearer_token_schema.authorization_param.return_value,
```
jon-betts: Not easy to see in the diff, but there's a lot of duplication here with the form fields from the fixture. I think this might benefit from a `form_fields` fixture used by both.

There's also a testing gap here. We don't check that we remove any of the `oauth_nonce` and friends.

tests/unit/lms/views/content_item_selection_test.py
```diff
@@ -108,10 +44,3 @@ def context(self):
```
jon-betts: Can't really comment the right line, but the `routes` fixture here doesn't appear to make any difference any more and can be removed.

The `custom_canvas_api_domain` and `custom_canvas_course_id` also don't appear to have a bearing on this test.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -22,10 +22,89 @@ def test_it_is_mutable(self, config):
...
+    def test_it(self, context, js_config):
+        js_config.config["mode"] = "foo"
+
+        js_config.enable_content_item_selection_mode(
+            "test_form_action", "test_form_fields"
+        )
 
         assert js_config.config["mode"] == "content-item-selection"
+        assert js_config.config["formAction"] == "test_form_action"
+        assert js_config.config["formFields"] == "test_form_fields"
```
jon-betts: Shouldn't test form fields be a dict? Arguably this shouldn't pass. I don't really believe in the test as documentation thing, but this is misleading too.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -22,10 +22,89 @@ def test_it_is_mutable(self, config):
...
+        self, js_config, pyramid_request
+    ):
+        del pyramid_request.params["custom_canvas_course_id"]
+
+        js_config.enable_content_item_selection_mode(
+            "test_form_action", "test_form_fields"
+        )
+
+        assert not js_config.config["enableLmsFilePicker"]
+        assert "courseId" not in js_config.config
```
jon-betts: Gah. I don't know how you'd do it exactly, but the duplication in the four above kind of kills me. I can't see how you'd parameterise it. Maybe an `_assert_lms_file_picker(js_config, enabled=False)` function instead? Then all of these tests would become some variation of:

```python
del pyramid_request.params["custom_canvas_course_id"]

self._assert_lms_file_picker(js_config, enabled=False)
```

lms/resources/_js_config.py
seanh: Oh yeah, I merged a PR to move this method (https://github.com/hypothesis/lms/pull/1520) and then forgot about it. Good catch

lms/resources/_js_config.py
seanh: Gonna get the whole lot in one place in `JSConfig` first, and then take a lot at reorganising it (if we have time).

With things like `ltiLaunchUrl` I'm thinking it's simpler to just set them all the time, in all LMS's and all types of assignment, even if they're not used all the time. It's just the base LTI launch URL, so it's at least potentially a setting of general interest. If we ever need to use it in a new circumstance, it'll be there. But mostly it's just less logic, less variation, and less tests in the `JSConfig` code.

seanh: Nested dicts within the `JSConfig` dict does seem like the direction to go in, yes. Once it's all in one place it'll hopefully be clear _which_ nested dicts would work best

lms/resources/_js_config.py
seanh: Yup

tests/unit/lms/views/basic_lti_launch_test.py
seanh: Ah yes, because the `pyramid_request` fixture doesn't have the `oauth` params in the first place. I'll add them.

I think the duplication can be removed by doing `"user_id": pyramid_request.params["user_id"]` etc instead of `"user_id": "TEST_USER_ID"`. I'm not sure that adds much value with "sentinel strings" like `"TEST_USER_ID"` but I'll do it anyway. It sort of makes it clearer what it's testing

tests/unit/lms/views/content_item_selection_test.py
seanh: Nope, well spotted

tests/unit/lms/resources/_js_config_test.py
seanh: I've replaced these with `mock.sentinel.form_fields`.

I don't really want to use a dict because, even though it would be a dict really, `JSConfig` just treats it as an opaque value and inserts it into the config.

But strings like `"test_form_fields"` are confusing because they make it look like the value is supposed to be a string.

`mock.sentinel` is a nice way to explicitly say "I'm not implying that this value would actually be a `mock.sentinel` object in production (obviously), I'm explicitly saying that the code treats this thing as an opaque value."

tests/unit/lms/resources/_js_config_test.py
seanh: I think one of those `method_caller()` fixtures would do it, I'll add that.

seanh: Oh, and for the two duplicated asserts an `asser_lms_file_picker_not_enabled()` fixture will work. If you make it a pytest fixture then you don't need to pass `js_config` to it 'cos it can itself depend on `js_config` as a fixture. It ends up meaning that the tests themselves don't need `js_config` as a parameter anymore

seanh: Did https://github.com/hypothesis/lms/pull/1546/commits/e87eb93e2841b0bb54069afe7fa45b0a92fb4875 and https://github.com/hypothesis/lms/pull/1546/commits/294852dada5039c3946100bd071cdd8850698c3f. End result:

```python
class TestEnableContentItemSelectionMode:
    def test_it(self, context, js_config, enable_content_item_selection_mode):
        js_config.config["mode"] = "foo"

        enable_content_item_selection_mode()

        ... (lots of asserts)

    def test_it_doesnt_enable_the_lms_file_picker_if_the_lms_isnt_Canvas(
        self,
        context,
        enable_content_item_selection_mode,
        assert_lms_file_picker_not_enabled,
    ):
        context.is_canvas = False

        enable_content_item_selection_mode()

        assert_lms_file_picker_not_enabled()

    def test_it_doesnt_enable_the_lms_file_picker_if_the_consumer_key_isnt_found_in_the_db(
        self,
        ai_getter,
        enable_content_item_selection_mode,
        assert_lms_file_picker_not_enabled,
    ):
        ai_getter.developer_key.side_effect = ConsumerKeyError()

        enable_content_item_selection_mode()

        assert_lms_file_picker_not_enabled()

    def test_it_doesnt_enable_the_lms_file_picker_if_we_dont_have_a_developer_key(
        self,
        ai_getter,
        enable_content_item_selection_mode,
        assert_lms_file_picker_not_enabled,
    ):
        ai_getter.developer_key.return_value = None

        enable_content_item_selection_mode()

        assert_lms_file_picker_not_enabled()

    def test_it_doesnt_enable_the_lms_file_picker_if_theres_no_custom_canvas_course_id(
        self,
        pyramid_request,
        enable_content_item_selection_mode,
        assert_lms_file_picker_not_enabled,
    ):
        del pyramid_request.params["custom_canvas_course_id"]

        enable_content_item_selection_mode()

        assert_lms_file_picker_not_enabled()

    @pytest.fixture
    def assert_lms_file_picker_not_enabled(self, js_config):
        def assert_lms_file_picker_not_enabled():
            assert not js_config.config["enableLmsFilePicker"]
            assert "courseId" not in js_config.config

        return assert_lms_file_picker_not_enabled

    @pytest.fixture
    def enable_content_item_selection_mode(self, js_config):
        """
        Return a wrapper method for enable_content_item_selection_mode.

        Passing standard test arguments to the method.
        """

        def enable_content_item_selection_mode():
            js_config.enable_content_item_selection_mode(
                mock.sentinel.form_action, mock.sentinel.form_fields
            )

        return enable_content_item_selection_mode
```

lms/resources/_js_config.py
seanh: Done

tests/unit/lms/views/basic_lti_launch_test.py
jon-betts: This is better, but it's not quite what I meant. Not worth going over again I'd say, but I was thinking something like:

```python
        ...
        context.js_config.enable_content_item_selection_mode.assert_called_once_with(
            form_action="http://example.com/module_item_configurations",
            form_fields=dict(form_params, authorization=bearer_token_schema.authorization_param.return_value)
        )

    @pytest.fixture
    def form_params(self):
        return {
            "user_id": "TEST_USER_ID",
            "resource_link_id": "TEST_RESOURCE_LINK_ID",
            "oauth_consumer_key": "TEST_OAUTH_CONSUMER_KEY",
            "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
            "context_id": "TEST_CONTEXT_ID",
        }

    @pytest.fixture
    def pyramid_request(self, pyramid_request, form_params):
        pyramid_request.params = dict(
            form_params,
            oauth_nonce="TEST_OAUTH_NONCE",
            oauth_timestamp="TEST_OAUTH_TIMESTAMP",
            oauth_signature="TEST_OAUTH_SIGNATURE",
        )
        
        return pyramid_request
```

lms/resources/_js_config.py
jon-betts: Yep, one for later.

tests/unit/lms/resources/_js_config_test.py
jon-betts: Hmm, I don't think this is better. We don't have to pass `js_config` into the tests, but seeing as these are tests about `js_config`, I'm not sure that's good. It's also longer than it was before.

Pretty much all of that additional complexity has been introduced to avoid passing a parameter. I think just because we can do this with `pytest` doesn't mean we should.

As an example here is the last test (with it's fixtures) in the current style:

```python
def test_it_doesnt_enable_the_lms_file_picker_if_theres_no_custom_canvas_course_id(
        self,
        pyramid_request,
        enable_content_item_selection_mode,
        assert_lms_file_picker_not_enabled,
    ):
        del pyramid_request.params["custom_canvas_course_id"]

        enable_content_item_selection_mode()

        assert_lms_file_picker_not_enabled()

    @pytest.fixture
    def assert_lms_file_picker_not_enabled(self, js_config):
        def assert_lms_file_picker_not_enabled():
            assert not js_config.config["enableLmsFilePicker"]
            assert "courseId" not in js_config.config

        return assert_lms_file_picker_not_enabled

    @pytest.fixture
    def enable_content_item_selection_mode(self, js_config):
        """
        Return a wrapper method for enable_content_item_selection_mode.

        Passing standard test arguments to the method.
        """

        def enable_content_item_selection_mode():
            js_config.enable_content_item_selection_mode(
                mock.sentinel.form_action, mock.sentinel.form_fields
            )

        return enable_content_item_selection_mode
```

Here is the same thing with the condition inline and with a plain function:

```python
    def test_it_doesnt_enable_the_lms_file_picker_if_theres_no_custom_canvas_course_id(
        self, pyramid_request, js_config,
    ):
        del pyramid_request.params["custom_canvas_course_id"]

        js_config.enable_content_item_selection_mode(
            mock.sentinel.form_action, mock.sentinel.form_fields
        )

        self.assert_lms_file_picker_not_enabled(js_config)

    def assert_lms_file_picker_not_enabled(self, js_config):
        assert not js_config.config["enableLmsFilePicker"]
        assert "courseId" not in js_config.config
```

There's just a lot less to read and understand there.

I was originally thinking to move the setup into the function too:

```python
    def test_it_doesnt_enable_the_lms_file_picker_if_theres_no_custom_canvas_course_id(
        self, pyramid_request, js_config,
    ):
        del pyramid_request.params["custom_canvas_course_id"]

        self.assert_lms_file_picker_not_enabled(js_config)

    def assert_lms_file_picker_not_enabled(self, js_config):
        js_config.enable_content_item_selection_mode(
            mock.sentinel.form_action, mock.sentinel.form_fields
        )

        assert not js_config.config["enableLmsFilePicker"]
        assert "courseId" not in js_config.config
```

This would lead to very short tests over all, but it looses the setup / do something / assert split, which is quite nice.

seanh:
Rebased

tests/unit/lms/resources/_js_config_test.py
seanh: Ah ok, I've gone ahead and done this: https://github.com/hypothesis/lms/pull/1546/commits/89b7fe47fef64db96a4829a05e9e1c7355df7fca

tests/unit/lms/views/basic_lti_launch_test.py
seanh: Gone ahead and done this: https://github.com/hypothesis/lms/pull/1546/commits/8162e84682102c01a6da3a742680eede8afb7cc1

It's cool that you can make a dict like this, btw: `dict(other_dict, **additional_items)`

PR #1547, opened 2020-03-09
Move frontend_app.py into JSConfig

seanh:
* CI will be failing because this needs the `_is_canvas()` method from https://github.com/hypothesis/lms/pull/1546/

This moves a bunch of grading-related settings that were in `frontend_app.py` into `JSConfig`. This removes `frontend_app.py` entirely: it only contained a few grading settings. These settings are used by our non-Canvas grading UI, not by our Canvas SpeedGrader support:

* `lmsGrader` is whether or not to show the non-Canvas grading UI when launching an assignment
* `["grading"]["courseName"]`, `["grading"]["assignmentName"]` and `["grading"]["students"]` are settings used by the grading UI. The course name and assignment name are used in the UI. The students are used to render the dropdown menu of students to grade (each student is an object with `"userid"`, `"displayName"`, `"LISResultSourcedId"`, `"LISOutcomeServiceUrl"`, which are the settings needed to display that student, tell the client to focus on that student, and call the APIs to retrieve and submit grades for that student)


lms/resources/_js_config.py
```diff
@@ -140,6 +143,31 @@ def _add_canvas_submission_params(self, **kwargs):
         # Add the given document_url or canvas_file_id.
         self.config["submissionParams"].update(kwargs)
 
+    def maybe_enable_grading(self):
```
jon-betts: This still kills me, but I can't think of anything better. `try_to_enable_grading()`? At least we commit to trying?

lms/resources/_js_config.py
```diff
@@ -202,6 +230,40 @@ def _debug(self):
...
+                display_name=grading_info.h_display_name,
+            )
+            students.append(
+                {
+                    "userid": h_user.userid,
+                    "displayName": h_user.display_name,
+                    "LISResultSourcedId": grading_info.lis_result_sourcedid,
+                    "LISOutcomeServiceUrl": grading_info.lis_outcome_service_url,
+                }
+            )
```
jon-betts: Same feedback as before here. If we yielded out the student and wrapped the call with `list` this would be a little snappier.

lms/resources/_js_config.py
```diff
@@ -140,6 +143,31 @@ def _add_canvas_submission_params(self, **kwargs):
         # Add the given document_url or canvas_file_id.
         self.config["submissionParams"].update(kwargs)
```
jon-betts: Hard to see in the diff here, but `_add_canvas_submission_params` is out and about with the public methods. This should probably be shoved to the bottom.

lms/resources/_js_config.py
seanh: It's not "try to enable grading if we can", though, it's "enable grading if appropriate".

A name like `enable_grading_if_not_canvas()` might work but the _if_ here is if the user is an instructor, if the assignment is gradeable, and if the LMS is not Canvas. So it would end up having to be `enable_grading_if_appropriate()` or something.

Or you split it into two functions:

```python
if context.js_config.should_enable_grading():
    context.js_config.enable_grading()
```

But I don't really want to do that.

I actually don't mind the "maybe" naming style, I think it may be the best approach.

lms/resources/_js_config.py
seanh: Would this really make it faster? I'm not sure I understand how yielding the values and then turning them into a list right after is faster than just returning the list in the first place?

lms/resources/_js_config.py
seanh: How did that happen? My precious alphabetical ordering. I'll fix it

lms/resources/_js_config.py
jon-betts: Snappier as in short and sweet, rather than as in code execution-wise

lms/resources/_js_config.py
jon-betts: To be clear, I mean kills me in that I think it's really funny. I don't object to it at all.

lms/resources/_js_config.py
seanh: https://github.com/hypothesis/lms/pull/1547/commits/eb705f6f3d1dca38aeb2966f121e4b04f9442797

lms/resources/_js_config.py
seanh: https://github.com/hypothesis/lms/pull/1547/commits/4c9d043f56a645d49ef4ed94da96c2726500a0a6

PR #1551, opened 2020-03-11
Fix a11y for Dialog

LMS007:
This is a continuation of this PR https://github.com/hypothesis/lms/pull/1481

In summary 
- Dialog may now have either a “dialog” or an “alertdialog” role depending what its used for.
- Dialog no longer has an `onKeyDown` handler. (a11y issue)
- `Escape` keyboard event is now handled with useElementShouldClose
- Add aria-modal=“true” prop to Dialog
- Dialog `id` prop is now generated with useUniqueId
- Expand Dialog `button` prop types for better accuracy

dom.js and use-elemenet-should-close.js are direct copies from the client added to a new common/ folder until we can get an imported package for such things.

### DIFF
--- a/lms/static/scripts/frontend_apps/common/dom.js
+++ b/lms/static/scripts/frontend_apps/common/dom.js
@@ -1,0 +1,23 @@
+/**
+ * Attach listeners for one or multiple events to an element and return a
+ * function that removes the listeners.
+ *
+ * @param {Element}
+ * @param {string[]} events
+ * @param {(event: Event) => any} listener
+ * @param {boolean} [options.useCapture]
+ * @return {function} Function which removes the event listeners.
+ */
+export function listen(element, events, listener, { useCapture = false } = {}) {
+  if (!Array.isArray(events)) {
+    events = [events];
+  }
+  events.forEach(event =>
+    element.addEventListener(event, listener, useCapture)
+  );
+  return () => {
+    events.forEach(event =>
+      element.removeEventListener(event, listener, useCapture)
+    );
+  };
+}
--- a/lms/static/scripts/frontend_apps/common/test/dom-test.js
+++ b/lms/static/scripts/frontend_apps/common/test/dom-test.js
@@ -1,0 +1,60 @@
+import { listen } from '../dom';
+
+describe('common/dom', () => {
+  const createElement = () => ({
+    addEventListener: sinon.stub(),
+    removeEventListener: sinon.stub(),
+  });
+
+  describe('listen', () => {
+    [true, false].forEach(useCapture => {
+      it('adds listeners for specified events', () => {
+        const element = createElement();
+        const handler = sinon.stub();
+
+        listen(element, ['click', 'mousedown'], handler, { useCapture });
+
+        assert.calledWith(
+          element.addEventListener,
+          'click',
+          handler,
+          useCapture
+        );
+        assert.calledWith(
+          element.addEventListener,
+          'mousedown',
+          handler,
+          useCapture
+        );
+      });
+    });
+
+    [true, false].forEach(useCapture => {
+      it('removes listeners when returned function is invoked', () => {
+        const element = createElement();
+        const handler = sinon.stub();
+
+        const removeListeners = listen(
+          element,
+          ['click', 'mousedown'],
+          handler,
+          { useCapture }
+        );
+        removeListeners();
+
+        assert.calledWith(
+          element.removeEventListener,
+          'click',
+          handler,
+          useCapture
+        );
+        assert.calledWith(
+          element.removeEventListener,
+          'mousedown',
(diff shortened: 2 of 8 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -60,23 +52,18 @@ export default function Dialog({
...
-      aria-labelledby="Dialog__title"
-      onKeyDown={handleKey}
-      // FIXME-A11Y
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex="0"
+      role={role}
+      aria-labelledby={dialogId}
+      aria-modal="true"
       ref={rootEl}
+      tabIndex="-1"
```
LMS007: Allows it to be programmatically focused, but not via keyboard navigation

LMS007: Decided to change this to -1 so it would not be something you can arrive at by keyboard. We really just use focus here when it opens as a fallback when a button or child element is not specified.

lms/static/scripts/frontend_apps/common/test/dom-test.js
```diff
@@ -0,0 +1,60 @@
+import { listen } from '../dom';
+
+describe('sidebar/util/dom', () => {
```
robertknight: Can you update this to match the local path of the module relative to `frontend_apps` (ie. `common/dom`)

lms/static/scripts/frontend_apps/common/test/use-element-should-close-test.js
```diff
@@ -0,0 +1,129 @@
+import { mount } from 'enzyme';
+import { createElement } from 'preact';
+import { useRef } from 'preact/hooks';
+import { act } from 'preact/test-utils';
+import propTypes from 'prop-types';
+
+import useElementShouldClose from '../use-element-should-close';
+
+describe('hooks.useElementShouldClose', () => {
```
robertknight:
```suggestion
describe('useElementShouldClose', () => {
```

There is no `hooks` directory here. Since this module only exports one thing, the name can just be the name of that thing.

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -4,53 +4,45 @@ import propTypes from 'prop-types';
 import classNames from 'classnames';
 
 import Button from './Button';
+import useElementShouldClose from '../common/use-element-should-close';
 import { zIndexScale } from '../utils/style';
+import { useUniqueId } from '../utils/hooks';
 
 /**
- * Accessibility notes:
+ * A model dialog wrapper with a title. The wrapper sets initial focus to itself
```
robertknight:
```suggestion
 * A modal dialog wrapper with a title. The wrapper sets initial focus to itself
```

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -4,53 +4,45 @@ import propTypes from 'prop-types';
...
- * Resources:
- *
- * 1. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role
- * 2. https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/
- * 3. https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/
- *
- * [3] is the most useful resource IMO as it highlights the essentials from
- * the perspective of a blind engineer.
+ * https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html
+ * https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html
```
robertknight: 👍 for references

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -4,53 +4,45 @@ import propTypes from 'prop-types';
...
   contentClass,
   initialFocus,
   onCancel,
+  role = 'dialog',
   title,
   buttons,
 }) {
-  const handleKey = event => {
-    event.stopPropagation();
+  const dialogId = useUniqueId('dialog');
```
robertknight: This is used to refer to the _title_ of the dialog rather than the dialog itself, so perhaps `dialogTitleId` and adjust the prefix accordingly.

lms/static/scripts/frontend_apps/common/dom.js
```diff
@@ -0,0 +1,23 @@
+/**
```
robertknight: I didn't review the additions to `frontend_apps/common` because the canonical location for that is the client repository.

PR #1552, opened 2020-03-11
Add via.config_frame_ancestor_level to via_url

LMS007:
This is needed to inform the client which ancestor frame to send RCP requests to. This PR needs to land before we can make any changes to via so it won't cause errors downstream.

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -37,6 +37,7 @@ def via_url(request, document_url):
     ]
     query_string_as_list.append(("via.open_sidebar", "1"))
     query_string_as_list.append(("via.request_config_from_frame", request.host_url))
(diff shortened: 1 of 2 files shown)


lms/views/helpers/_via.py
```diff
@@ -37,6 +37,7 @@ def via_url(request, document_url):
     ]
     query_string_as_list.append(("via.open_sidebar", "1"))
     query_string_as_list.append(("via.request_config_from_frame", request.host_url))
+    query_string_as_list.append(("via.config_frame_ancestor_level", 2))
```
robertknight: Ah, I didn't know that integer values would be automatically converted to strings, although that's not too surprising.

robertknight:
I'm going to go ahead and deploy this to speed up getting the Via change shipped later.

PR #1553, opened 2020-03-12
Make JSConfig.config private

seanh:
Change the mutable property JSConfig.config to a private, mutable
property JSConfig._config (it's still mutated by code within JSConfig).

Add a new JSConfig.asdict() method for the one place where code outside
of JSConfig needs to actually access the config dict (in the base
template, where it gets rendered to JSON).

Since the new asdict() just returns self._config anyway, this doesn't
make any real difference. But it might clarify the intent: code outside
of JSConfig isn't supposed to read or write the underlying config dict
anymore.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -24,57 +24,6 @@ class JSConfig:  # pylint:disable=too-few-public-methods
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
 
-    @property
-    @functools.lru_cache()
-    def config(self):
-        """
-        Return the configuration for the app's JavaScript code.
-
-        :raise HTTPBadRequest: if a request param needed to generate the config
-            is missing
-
-        :rtype: dict
-        """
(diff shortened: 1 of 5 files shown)


lms/resources/_js_config.py
```diff
@@ -94,9 +43,20 @@ def add_document_url(self, document_url):
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
-        self.config["urls"]["via_url"] = via_url(self._request, document_url)
+        self._config["urls"]["via_url"] = via_url(self._request, document_url)
         self._add_canvas_submission_params(document_url=document_url)
 
+    def asdict(self):
```
jon-betts: Why isn't this `as_dict()`?

seanh: `asdict()` rather than `as_dict` is used in the standard library so I figured I'd follow that, e.g. https://docs.python.org/3/library/dataclasses.html?highlight=asdict#dataclasses.asdict and https://docs.python.org/3/library/collections.html?highlight=asdict#collections.somenamedtuple._asdict

PR #1554, opened 2020-03-12
Make a small code simplification

seanh:


lms/resources/_js_config.py
```diff
@@ -231,16 +231,12 @@ def _add_canvas_submission_params(self, **kwargs):
...
-        )
-
-        # Add the given document_url or canvas_file_id.
-        self.config["submissionParams"].update(kwargs)
+        self.config["submissionParams"] = {
+            "h_username": self._context.h_user.username,
+            "lis_result_sourcedid": lis_result_sourcedid,
+            "lis_outcome_service_url": lis_outcome_service_url,
+            **kwargs,
+        }
```
seanh: This method is the only place (in all of the Python code) where `"submissionParams"` is mentioned. So `"submissionParams"` is never already set when this method runs. We can just set it, no need to do `setdefault(...).update(...)`

Issue #1555, opened 2020-03-13
Add stubbed version of new RPC server 'requestGroups' method


seanh:
## Current situation

The LMS app's frontend code contains a [JSON-RPC-over-postMessage server](https://github.com/hypothesis/lms/tree/master/lms/static/scripts/postmessage_json_rpc/server) that receives JSON-RPC requests from the Hypothesis client and responds to them.

Currently this JSON-RPC server only supports [one method](https://github.com/hypothesis/lms/blob/375bef26a184530042149aa2b3334476f426ef47/lms/static/scripts/postmessage_json_rpc/server/methods.js#L9-L16) (`'requestConfig'`) which allows the client to request and receive its config object.

<details>
<summary>Background on postMessage-JSON-RPC</summary>

#### postMessage

[postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) is the browser's API for allowing JavaScript code in different frames to intercommunicate. The way postMessage works is:

The sender has to have a reference to the receiver's `window` object (the Hypothesis client has a technique for getting a reference to the LMS app's `window`), and then can send a data object to the receiver like this:

```javascript
targetWindow.postMessage(data, "lms.hypothes.is");
```

(The second argument is the origin of the receiving window. If this does not match `targetWindow`'s origin then the browser will not send the message.)

Here's [the function in the client that posts a message to the LMS app and waits for the response](https://github.com/hypothesis/client/blob/ada96df5ff8d7782c31a6175e8c980a2e86a8ca9/src/sidebar/util/postmessage-json-rpc.js#L17-L98).

The receiver must have registered a "message" event listener (callback function) with the browser, and on receiving the `postMessage` the browser will call the receiver's callback function:

```javascript
window.addEventListener('message', receiveMessage, false);

function receiveMessage(event) {
  // Called by the browser when the sender calls targetWindow.postMessage(...).
}
```

Here's [where we register a postMessage event listener in the LMS app's server](https://github.com/hypothesis/lms/blob/375bef26a184530042149aa2b3334476f426ef47/lms/static/scripts/postmessage_json_rpc/server/server.js#L34), and here's [the LMS app's server's callback function for receiving messages](https://github.com/hypothesis/lms/blob/375bef26a184530042149aa2b3334476f426ef47/lms/static/scripts/postmessage_json_rpc/server/server.js#L58-L81).

[...]

PR #1557, opened 2020-03-13
Add RPC server requestGroups method with artificial delay in dev

seanh:
Nothing actually calls this yet. But I have hacked the client to call it locally, and verified that it returns the list of one group ID.

Fixes https://github.com/hypothesis/lms/issues/1555

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -1,6 +1,7 @@
 [app:main]
 use = call:lms.app:create_app
 debug = true
+dev = true
 
 pyramid.reload_templates = true
 pyramid.debug_authorization = false
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -255,6 +255,8 @@ class JSConfig:  # pylint:disable=too-few-public-methods
             "authUrl": self._request.route_url("canvas_api.authorize"),
             # Some debug information, currently used in the Gherkin tests.
             "debug": self._debug(),
+            # Tell the JavaScript code whether we're in "dev" mode.
+            "dev": self._request.registry.settings["dev"],
             # The config object for the Hypothesis client.
             # Our JSON-RPC server passes this to the Hypothesis client over
             # postMessage.
--- a/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
@@ -39,7 +39,7 @@ describe('postmessage_json_rpc/methods#requestGroups', () => {
     configEl.parentNode.removeChild(configEl);
   });
 
-  it('returns the list of groups', () => {
-    assert.deepEqual(requestGroups(), ['groupid1', 'groupid2']);
+  it('returns the list of groups', async () => {
+    assert.deepEqual(await requestGroups(), ['groupid1', 'groupid2']);
   });
 });
--- a/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods-test.js
@@ -1,4 +1,4 @@
-import { requestConfig } from './methods';
+import { requestConfig, requestGroups } from './methods';
 
 describe('postmessage_json_rpc/methods#requestConfig', () => {
   let configEl;
@@ -19,5 +19,27 @@ describe('postmessage_json_rpc/methods#requestConfig', () => {
 
   it('returns the config object', () => {
     assert.deepEqual(requestConfig(), { foo: 'bar' });
+  });
+});
+
+describe('postmessage_json_rpc/methods#requestGroups', () => {
+  let configEl;
+
+  beforeEach('inject the client config into the document', () => {
+    configEl = document.createElement('script');
+    configEl.setAttribute('type', 'application/json');
+    configEl.classList.add('js-config');
+    configEl.textContent = JSON.stringify({
+      hypothesisClient: { services: [{ groups: ['groupid1', 'groupid2'] }] },
+    });
+    document.body.appendChild(configEl);
+  });
+
+  afterEach('remove the client config from the document', () => {
+    configEl.parentNode.removeChild(configEl);
+  });
+
+  it('returns the list of groups', () => {
+    assert.deepEqual(requestGroups(), ['groupid1', 'groupid2']);
   });
 });
--- a/lms/static/scripts/postmessage_json_rpc/server/methods.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods.js
@@ -14,3 +14,12 @@ export function requestConfig() {
   const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
   return clientConfigObj;
 }
+
+/**
+ * Return the groups for the Hypothesis client to show.
+ */
+export function requestGroups() {
+  const configEl = document.querySelector('.js-config');
+  const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
+  return clientConfigObj.services[0].groups;
+}
--- a/lms/static/scripts/postmessage_json_rpc/server/methods.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods.js
@@ -30,7 +30,11 @@ export async function requestGroups() {
   if (configObj.dev === true) {
     // Artifically sleep to simulate how long this will take when it needs to
     // send real API requests to get the groups.
-    await sleep(1900);
+    // The artificial delay is only inserted 50% of the time (at random) so we
+    // don't somehow accidentally end up relying on the slowness.
+    if (Math.random() < 0.5) {
+      await sleep(1900);
+    }
   }
 
   return configObj.hypothesisClient.services[0].groups;
--- a/lms/static/scripts/postmessage_json_rpc/server/methods.js
+++ b/lms/static/scripts/postmessage_json_rpc/server/methods.js
@@ -18,8 +18,20 @@ export function requestConfig() {
 /**
  * Return the groups for the Hypothesis client to show.
  */
-export function requestGroups() {
-  const configEl = document.querySelector('.js-config');
-  const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
-  return clientConfigObj.services[0].groups;
+export async function requestGroups() {
+  const configObj = JSON.parse(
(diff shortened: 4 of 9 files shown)


lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -14,3 +14,12 @@ export function requestConfig() {
...
   return clientConfigObj;
 }
+
+/**
+ * Return the groups for the Hypothesis client to show.
+ */
+export function requestGroups() {
+  const configEl = document.querySelector('.js-config');
+  const clientConfigObj = JSON.parse(configEl.textContent).hypothesisClient;
+  return clientConfigObj.services[0].groups;
```
seanh: This is hacky of course (reaching into the client config object to extract the groups) but I think that's fine because it's temporary. We're working as fast as possible to replace this with something else entirely (getting the groups from the sync API response), so I don't think it's worth doing anything like redesigning the JavaScript config to better support this temporary hack in the meantime.

lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -14,3 +14,24 @@ export function requestConfig() {
...
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  if (configObj.dev === true) {
+    // Artifically sleep to simulate how long this will take when it needs to
+    // send real API requests to get the groups.
+    await sleep(1900);
+  }
+
+  return configObj.hypothesisClient.services[0].groups;
```
seanh: This is hacky of course (reaching into the client config object to extract the groups) but I think that's fine because it's temporary. We're working as fast as possible to replace this with something else entirely (getting the groups from the sync API response), so I don't think it's worth doing anything like redesigning the JavaScript config to better support this temporary hack in the meantime.

conf/development.ini
```diff
@@ -1,6 +1,7 @@
 [app:main]
 use = call:lms.app:create_app
 debug = true
+dev = true
```
seanh: Added a new "dev" config setting so that we know when we're in dev mode. It's true only in `make dev`, not on QA, in prod, or in tests,

lms/resources/_js_config.py
```diff
@@ -54,6 +54,8 @@ def config(self):
             "authUrl": self._request.route_url("canvas_api.authorize"),
             # Some debug information, currently used in the Gherkin tests.
             "debug": self._debug(),
+            # Tell the JavaScript code whether we're in "dev" mode.
+            "dev": self._request.registry.settings["dev"],
```
jon-betts: Should this be part of the `debug` dict?

lms/static/scripts/postmessage_json_rpc/server/methods.js
```diff
@@ -14,3 +14,24 @@ export function requestConfig() {
...
+  );
+
+  function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  if (configObj.dev === true) {
+    // Artifically sleep to simulate how long this will take when it needs to
+    // send real API requests to get the groups.
+    await sleep(1900);
```
jon-betts: Might be a better test if this was quick or slow randomly, in case we end up relying on the slow behavior.

lms/resources/_js_config.py
seanh: I don't think it has to do with debugging in particular, so maybe not? Maybe `debug` should be under a `dev` dict, if `debug` is only meant for use in dev?

seanh:
Rebased

lms/static/scripts/postmessage_json_rpc/server/methods.js
seanh: Done. Also increased the artificial delay so that it's longer than the client's current RPC response timeout of two seconds (a timeout that is going to have to be extended or removed, for sections)

lms/resources/_js_config.py
jon-betts: That works too, they seem related. The debug dict doesn't have to be anywhere in particular to work, the contents just need to appear in the page content.

lms/static/scripts/postmessage_json_rpc/server/methods-test.js
```diff
@@ -21,3 +21,25 @@ describe('postmessage_json_rpc/methods#requestConfig', () => {
     assert.deepEqual(requestConfig(), { foo: 'bar' });
   });
 });
+
+describe('postmessage_json_rpc/methods#requestGroups', () => {
```
LMS007: Jon, i don't want to put too much JS burden on you, so if you want to pass this off, I can refactor a bit.

We can restructure this test file like this and reuse a bit of the `beforeEach` stuff

```describe('postmessage_json_rpc/methods#requestConfig', () => {
  let configEl;

  beforeEach('inject the client config into the document', () => {
    configEl = document.createElement('script');
    configEl.setAttribute('type', 'application/json');
    configEl.classList.add('js-config');
    configEl.textContent = JSON.stringify({
      hypothesisClient: { foo: 'bar' },
    });
    document.body.appendChild(configEl);
  });

  afterEach('remove the client config from the document', () => {
    configEl.parentNode.removeChild(configEl);
  });

  describe('#requestConfig', () => {
    it(' the config returnsobject', () => {
      assert.deepEqual(requestConfig(), { foo: 'bar' });
    });
  });

  describe('#requestGroups', () => {
    it('returns the list of groups filter', async () => {
      configEl.textContent = JSON.stringify({
        hypothesisClient: { services: [{ groups: ['groupid1', 'groupid2'] }] },
      });
      assert.deepEqual(await requestGroups(), ['groupid1', 'groupid2']);
    });

    it.only('returns the list of groups filter in dev mode', async () => {
      const clock = sinon.useFakeTimers();

      configEl.textContent = JSON.stringify({
        dev: true,
        hypothesisClient: { services: [{ groups: ['groupid1', 'groupid2'] }] },
      });

      const promise = requestGroups();
      clock.tickAsync(4000);
      const groups = await promise;
      assert.deepEqual(groups, ['groupid1', 'groupid2']);
    });
  });
});```

I also added a test for the sleep() method you made. If we're gonna be REALLY picky, we would stub out the math.random too, but I think thats overkill for just a placeholder function anyway

File: lms/static/scripts/postmessage_json_rpc/server/methods.js [hunk 7]
```diff
@@ -14,3 +14,28 @@ export function requestConfig() {
...
+export async function requestGroups() {
+  const configObj = JSON.parse(
+    document.querySelector('.js-config').textContent
+  );
+
+  function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  if (configObj.dev === true) {
```
LMS007: super dumb question. do we really need "dev" flag? I'm just thinking ahead here, does it even make sense to get the client code calling `requestGroups` before its done? I'm sure its useful to have perhaps a client branch that makes requests to `requestGroups` but I'm just not sure its valuable to put that into master branch before its done. Was the intent here we get the client making the groups call, then backfill the requestGroups with a real method since the sync/ api (or similar is done) ?

seanh:
> I want to firm up my understanding of whats to come on the client repo. As of right now, the services: [{ groups: [...}] in the js-config object represents a filter, but is not in fact the list of groups the client will display. That list rather comes from the /groups service call to h.
> 
> Will that service call no longer work in the same capacity? As of right now, the only was I see requestGroups() working, is if the list of groups already exist from the /group api call (from h) and that this additional RCP call (requestGroups) filters them down to the actual group(s) that a canvas sidebar would need out of the larger set from h.
> 
> To put it another way: If in fact there are group ids from the requestGroups() call that are not present in the h groups/ call, then this will not work without some more significate refactoring on the client.

Deets are in the client issue: https://github.com/hypothesis/client/issues/1908. The `/groups` service call to h is going to continue to work in exactly the same way as it works currently, that isn't changing. The `services: [{ groups: [...}]` in the `js-config` object will continue to represent a filter and will continue to work in the same way as it does now. The list of groups will continue to already exist in the `/group` API call from h in the same way as it does now.

The _only_ thing that's changing in the client is that it needs to get the groups from the new `requestGroups` RPC call instead of getting them from the `requestConfig` RPC call.

lms/static/scripts/postmessage_json_rpc/server/methods.js
seanh: Yes, let's get the client work done before the LMS work is done. That will allow us to parallelize the client work and the LMS work, and so release the sections feature sooner. And let's actually merge that client PR into master and release it, so the whole work of reviewing/merging/releasing the client changes can be done in parallel with ongoing LMS changes.

The `requestGroups` method in this PR _is a real method_ that really does return the list of one course group. It will later be changed to return a list of more than one section groups.

lms/resources/_js_config.py
seanh: I'm gonna leave this where it is. The `dev` setting is for indicating that we're in dev, i.e. `make dev`, as opposed to QA, production, _or running the tests_. The `debug` setting is only used in the tests, not in dev. If anything `debug` should maybe be renamed to `tests`. Unless it one day gets extended to also contain stuff that we use for debugging things in the dev server

LMS007:
> > I want to firm up my understanding of whats to come on the client repo. As of right now, the services: [{ groups: [...}] in the js-config object represents a filter, but is not in fact the list of groups the client will display. That list rather comes from the /groups service call to h.
> > Will that service call no longer work in the same capacity? As of right now, the only was I see requestGroups() working, is if the list of groups already exist from the /group api call (from h) and that this additional RCP call (requestGroups) filters them down to the actual group(s) that a canvas sidebar would need out of the larger set from h.
> > To put it another way: If in fact there are group ids from the requestGroups() call that are not present in the h groups/ call, then this will not work without some more significate refactoring on the client.
> 
> Deets are in the client issue: [hypothesis/client#1908](https://github.com/hypothesis/client/issues/1908). The `/groups` service call to h is going to continue to work in exactly the same way as it works currently, that isn't changing. The `services: [{ groups: [...}]` in the `js-config` object will continue to represent a filter and will continue to work in the same way as it does now. The list of groups will continue to already exist in the `/group` API call from h in the same way as it does now.
> 
> The _only_ thing that's changing in the client is that it needs to get the groups from the new `requestGroups` RPC call instead of getting them from the `requestConfig` RPC call.

@seanh perfect, thank you for clarification

PR #1573, opened 2020-03-19
Swap Via 3 to be the default with legacy Via as a feature flag

jon-betts:
You can now enable legacy Via with:

```bash
export FEATURE_FLAG_USE_LEGACY_VIA=true
```
----

This requires rebasing after a few other things:

 * https://github.com/hypothesis/lms/pull/1572
 * https://github.com/hypothesis/lms/pull/1571

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -16,13 +16,14 @@ session_cookie_secret = "notasecret"
 
 # The secret string that's used to sign the feature flags cookie.
 feature_flags_cookie_secret = "notasecret"
-feature_flags_allowed_in_cookie = use_via3
+feature_flags_allowed_in_cookie = use_legacy_via
 
 # The secret string that's used to sign the OAuth2 state param.
 oauth2_state_secret = "notasecret"
 
-via_url = http://localhost:9080
-via3_url = http://localhost:9082
+# Set the default to Via 3 for dev (original via: 9080)
+via_url = http://localhost:9082
+legacy_via_url = http://localhost:9080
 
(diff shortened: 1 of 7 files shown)


jon-betts:
We want all hands on deck when deploying this in case it goes horribly wrong. It's a bit late in the day in the UK for that so we'll have another crack tomorrow.

conf/development.ini
```diff
@@ -16,13 +16,14 @@ session_cookie_secret = "notasecret"
...
 feature_flags_cookie_secret = "notasecret"
-feature_flags_allowed_in_cookie = use_via3
+feature_flags_allowed_in_cookie = use_legacy_via
 
 # The secret string that's used to sign the OAuth2 state param.
 oauth2_state_secret = "notasecret"
 
-via_url = http://localhost:9080
-via3_url = http://localhost:9082
+# Set the default to Via 3 for dev (original via: 9080)
```
seanh:
```suggestion
```

I'm not sure this comment's adding anything

tests/unit/lms/views/helpers/_via_test.py
```diff
@@ -14,6 +14,16 @@ class TestViaURL:
         "via.config_frame_ancestor_level": "2",
     }
 
+    def test_if_creates_the_correct_via_url(self, pyramid_request):
```
seanh:
```suggestion
    def test_it_creates_the_correct_via_url(self, pyramid_request):
```

Issue #1565, opened 2020-03-18
Reorganize the JavaScript config


seanh:
Now that the JavaScript config has been gathered into one place (https://github.com/hypothesis/lms/issues/1435), reorganise it better. See https://github.com/hypothesis/lms/issues/1435#issuecomment-598738736 for the specific reorganisation we're planning

Issue #1435, opened 2020-02-05
Refactor: Gather all JavaScript config code into a service


seanh:
Gather all the code that generates and mutates JavaScript config objects into one place, in a service. Then do any tidying or simplifying that can be done.

seanh:
Now that we've gathered everything together into one place, we can take a look at what the shape of our JavaScript config is:

```
authToken
mode
debug:
  tags
hypothesisClient:  # Is an empty dict when provisioning not enabled
  services[0]:
    apiUrl
    authority
    enableShareLinks
    grantToken groups
    # Only present if there was a focused_user param, i.e. only in SpeedGrader.
    focus:
        user:
            username
            displayName
rpcServer:
  allowedOrigins
urls:
  via_url  # Not present for Canvas Files assignments (via_url_callback must be used)
  via_url_callback

# Only present in content item selection mode.
formAction
formFields

# Only present in content item selection mode. Used for Google Picker.
googleClientId
googleDeveloperKey
customCanvasApiDomain  # Actually for Google Picker, when in Canvas.
lmsUrl  # For Google Picker, when not in Canvas.

# Only present in content item selection mode.
# True when Canvas Files are available for this application instance, otherwise
# False.
enableLmsFilePicker
# Only present when Canvas Files are available for this application instance.
ltiLaunchUrl
courseId

# Non-Canvas grading UI.
# This stuff is only present when not in Canvas, the user is an instructor,
# and there's a lis_outcome_service_url (i.e. the assignment is gradeable)
lmsGrader  # Always either True or just missing from the config.
grading:
  courseName
  assignmentName
  students [userid, displayName, LISResultSourcedId, LISOutcomeServiceUrl]

# Canvas only.
authUrl

# Canvas SpeedGrader.
# Actually used during normal student launches, but for SpeedGrader support.
# Only present in Canvas and when there's a lis_result_sourcedid and
# lis_outcome_service_url (i.e. only for learner launches of gradeable
# assignments)
submissionParams:
  h_username
  lis_result_sourcedid
  lis_outcome_service_url
  canvas_file_id  # Only for Canvas file assignments
  document_url    # For non-Canvas file assignments
```

seanh:
Based on the above, we might reorganize things under these top-level keys:

1. mode
2. authToken (what the frontend uses to authenticate to our backend APIs)
3. debug
3. rpcServer
4. urls
5. hypothesisClient
6. filePicker (with googlePicker and canvasPicker sub-objects)
[...]

PR #1577, opened 2020-03-23
JS config reorganization 1/n: viaUrl

seanh:

This PR moves `urls.via_url` to a top-level setting and renames it to `viaUrl`. The `urls` sub-object will eventually be going away.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -43,7 +43,7 @@ class JSConfig:  # pylint:disable=too-few-public-methods
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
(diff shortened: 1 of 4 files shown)


lms/resources/_js_config.py
```diff
@@ -43,7 +43,7 @@ def add_document_url(self, document_url):
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
-        self._config["urls"]["via_url"] = via_url(self._request, document_url)
+        self._config["viaUrl"] = via_url(self._request, document_url)
```
jon-betts: I'm kind of surprised we didn't break any tests here. Some missing coverage?

PR #1583, opened 2020-03-23
JS config reorganization 2/n: api 

seanh:

This PR introduces the new `"api"` sub-object and moves the `"via_url_callback"` setting into it, and renames it to `"viaCallbackUrl"`. The previously top-level `"authToken"` setting also moves under `"api"` (this is the token for authenticating API requests).

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -49,24 +49,24 @@ const INITIAL_LTI_LAUNCH_STATE = {
  */
 export default function BasicLtiLaunchApp() {
   const {
-    api: { authToken },
+    api: {
+      authToken,
+      // API callback to use to fetch the URL to show in the iframe. This is
+      // needed if resolving the content URL involves potentially slow calls
+      // to third party APIs (eg. the LMS's file storage).
+      viaCallbackUrl,
+    },
     authUrl,
     grading,
     lmsGrader,
     submissionParams,
-    urls: {
-      // API callback to use to fetch the URL to show in the iframe. This is
-      // needed if resolving the content URL involves potentially slow calls
-      // to third party APIs (eg. the LMS's file storage).
-      via_url_callback: viaUrlCallback,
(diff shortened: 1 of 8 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -49,24 +49,26 @@ const INITIAL_LTI_LAUNCH_STATE = {
...
     grading,
     lmsGrader,
     submissionParams,
     urls: {
       // Content URL to show in the iframe.
       via_url: viaUrl,
-      // API callback to use to fetch the URL to show in the iframe. This is
-      // needed if resolving the content URL involves potentially slow calls
-      // to third party APIs (eg. the LMS's file storage).
-      via_url_callback: viaUrlCallback,
```
jon-betts: If PR #1 is merged, this will leave an empty dict here. So this will likely need to be fixed when that is merged before this one can go in (as both remove different halves). This is true in a few places in the python too, but this is the only place I could put a comment about it.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -30,7 +30,9 @@ const GRADE_MULTIPLIER = 10;
  * and error status of that fetch request.
  */
 const useFetchGrade = student => {
-  const { authToken } = useContext(Config);
+  const {
+    api: { authToken },
+  } = useContext(Config);
```
jon-betts: I assume this is destructuring? Not a big Javascript head.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -4,4 +4,4 @@ import { createContext } from 'preact';
  * Configuration object for the file picker application, read from a JSON
  * script tag injected into the page by the backend.
  */
-export const Config = createContext({});
+export const Config = createContext({ api: {} });
```
jon-betts: What did this break when it wasn't included? Do we have to do this for a lot of other stuff?

PR #1585, opened 2020-03-23
JS config reorganization 4/n: grading

seanh:

This PR moves the `lmsGrader` setting into the already-existing `grading` sub-object of the JavaScript config object with the rest of the grading-related settings, and renames it to `enabled`. All grading settings are now gathered together in the `grading` sub-object.


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -197,7 +196,7 @@ export default function BasicLtiLaunchApp() {
       />
     );
 
-    if (lmsGrader) {
+    if (grading && grading.enabled) {
```
jon-betts: This is _sooo_ much easier to understand.

PR #1586, opened 2020-03-23
JS config reorganization 5/n: canvas

seanh:

This PR moves Canvas-specific settings into a `"canvas"` sub-object. `authUrl` is now `canvas.authUrl`, and `submissionParams` is now `canvas.speedGrader.submissionParams`.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -59,7 +59,7 @@ export default function BasicLtiLaunchApp() {
     grading,
     // Content URL to show in the iframe.
     viaUrl,
-    canvas: { authUrl, speedGrader },
+    canvas,
   } = useContext(Config);
 
   const [ltiLaunchState, setLtiLaunchState] = useState({
@@ -125,7 +125,7 @@ export default function BasicLtiLaunchApp() {
     // If a teacher launches an assignment or the LMS does not support reporting
     // outcomes or grading is not enabled for the assignment, then no submission
     // URL will be available.
-    if (!speedGrader.submissionParams) {
+    if (!canvas.speedGrader.submissionParams) {
       return;
     }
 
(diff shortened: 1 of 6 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -62,6 +60,7 @@ export default function BasicLtiLaunchApp() {
       // to third party APIs (eg. the LMS's file storage).
       via_url_callback: viaUrlCallback,
     },
+    canvas: { authUrl, speedGrader },
```
jon-betts: If we are going to pop these out into variables, they should probably be `canvasAuthUrl` and `canvasSpeedGrader` otherwise we are back to using generic terms for Canvas specific things.

Less so for Speed Grader, but it still requires some domain specific knowledge, so why not just always map the names directly from the path?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -39,6 +38,7 @@ export default function FilePickerApp({
     customCanvasApiDomain,
     lmsUrl,
     ltiLaunchUrl,
+    canvas: { authUrl },
```
jon-betts: Ditto here regarding param names: `canvasAuthUrl`

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -465,6 +457,11 @@ def config(js_config):
     return js_config.asdict()
 
 
+@pytest.fixture
+def get_submission_params(config):
```
jon-betts:
```suggestion
def submission_params(config):
```

seanh: Done

lms/static/scripts/frontend_apps/components/FilePickerApp.js
seanh: Done

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
seanh: Done. Note that generally, though, the frontend is absolutely rife with generic names for specific things and it's quite a pain to work with, refactoring it would take too long and the risk of breaking things would be too high, so I'm going to be keeping frontend changes to a minimum

PR #1588, opened 2020-03-24
Split up a test class into multiple test classes

seanh:
`LTILaunchResource` is too large and as a result its test class is too large. We have cards coming up that will involve necessarily splitting `LTILaunchResource` up into multiple classes. For now, and in preparation, split its tests up into one test class per method.

This makes the tests easier to add to which we'll be doing soon.

Also removes some duplication from the tests by moving code into class-level fixtures.

Also clarifies the tests by having each test class have its own `pyramid_request` fixture, so you can see which request params each method being tested actually depends on.

### DIFF
--- a/tests/unit/lms/resources/lti_launch_test.py
+++ b/tests/unit/lms/resources/lti_launch_test.py
@@ -219,11 +219,8 @@ class TestHGroupName:
         "context_title,expected_group_name",
         (
             ("Test Course", "Test Course"),
-            (" Test Course", "Test Course"),
-            ("Test Course ", "Test Course"),
             (" Test Course ", "Test Course"),
             ("Test   Course", "Test   Course"),
-            ("Object Oriented Programming 101", "Object Oriented Programm…"),
             ("Object Oriented Polymorphism 101", "Object Oriented Polymorp…"),
             ("  Object Oriented Polymorphism 101  ", "Object Oriented Polymorp…"),
         ),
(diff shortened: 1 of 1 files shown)


tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -240,17 +263,22 @@ def test_h_provider_raises_if_tool_consumer_instance_guid_is_missing(
...
+            "tool_consumer_instance_guid": "test_tool_consumer_instance_guid",
+        }
+        return pyramid_request
 
+
+class TestHProviderUniqueID:
+    def test_it_just_returns_the_user_id(self, pyramid_request):
         provider_unique_id = LTILaunchResource(pyramid_request).h_provider_unique_id
 
-        assert provider_unique_id == "4533***70d9"
```
jon-betts: wat

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -174,50 +179,65 @@ def test_h_authority_provided_id_raises_if_theres_no_context_id(
...
         "context_title,expected_group_name",
         (
             ("Test Course", "Test Course"),
-            (" Test Course", "Test Course"),
-            ("Test Course ", "Test Course"),
             (" Test Course ", "Test Course"),
             ("Test   Course", "Test   Course"),
-            ("Object Oriented Programming 101", "Object Oriented Programm…"),
             ("Object Oriented Polymorphism 101", "Object Oriented Polymorp…"),
             ("  Object Oriented Polymorphism 101  ", "Object Oriented Polymorp…"),
```
jon-betts: I couldn't put a comment because these weren't introduced here, but these extra cases aren't really adding anything

Issue #1410, opened 2020-01-30
Add a function for generating an h group ID from a Canvas section


seanh:
## How the LMS app currently generates the h groupid for a course

* It's generated based entirely on standard LTI launch params

* It's deterministic. Given the same LTI launch params again it will generate the same `groupid`

* It's globally unique: no two courses (either within the same LMS or across different LMS's) will get the same `groupid`

* [`LTILaunchResource.h_groupid()`](https://github.com/hypothesis/lms/blob/6eaab7dbd3128a955ed1f4b61937b2de695a64e7/lms/resources/_lti_launch.py#L101-L116) is the method that generates the `groupid`

* `h_groupid()` just returns `"group:<AUTHORITY_PROVIDED_ID>@<AUTHORITY>"` where:

* `<AUTHORITY>` is the `H_AUTHORITY` deployment setting (e.g. `"lms.hypothes.is"` on production)

* `<AUTHORITY_PROVIDED_ID>` comes from [`LTILaunchResource.h_authority_provided_id()`](https://github.com/hypothesis/lms/blob/6eaab7dbd3128a955ed1f4b61937b2de695a64e7/lms/resources/_lti_launch.py#L75-L99). It's a SHA 1 hash of the standard LTI launch params `"tool_consumer_instance_guid"` (a globally unique ID for the LMS instance, e.g. the particular Canvas site) and `"context_id"` (an ID for the course, unique within the LMS instance).

## Generating the h groupid for a section

The `groupid` for a group corresponding to a Canvas course section must:

* Be deterministic--the same Canvas section must always generate the same `groupid`

* Be in the format `group:<AUTHORITY_PROVIDED_ID>@<AUTHORITY>`

* Be globally unique across all h groups. The `@<AUTHORITY>` part means they will never clash with non-LMS groups. The `<AUTHORITY_PROVIDED_ID>` part has to take care of the rest

* The  `@<AUTHORITY>` part must be the `H_AUTHORITY` deployment setting (e.g. `"lms.hypothes.is"` on production)

* The `<AUTHORITY_PROVIDED_ID>` part must:

  * Contain only the characters `[a-zA-Z0-9._\-+!~*()']`

  * Be between 1 and 1024 chars in total

  * Be unique within the LMS authority. That means unique across all sections, courses and LMS's. No two LMS groups can have the same `<AUTHORITY_PROVIDED_ID>` even if they're in different courses or different LMS's

[...]

PR #1589, opened 2020-03-24
Add unique section groupid function

seanh:
Add a method, `h_section_groupid()`, for deterministically generating a unique h groupid for a Canvas course section.

Fixes https://github.com/hypothesis/lms/issues/1410

See https://github.com/hypothesis/lms/issues/1410#issuecomment-603248954 for the design of the IDs

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -106,6 +106,34 @@ class LTILaunchResource:
         """
         return f"group:{self.h_authority_provided_id}@{self._authority}"
 
+    def h_section_groupid(self, tool_consumer_instance_guid, context_id, section):
+        """
+        Return a unique h groupid for a given Canvas course section.
+
+        The groupid is deterministic and is unique to the course section.
+        Calling this function again with params representing the same course
+        section will always return the same groupid. Calling this function with
+        different params will always return a different groupid.
+
+        :param tool_consumer_instance_guid: the tool_consumer_instance_guid LTI
+            launch param from the Canvas instance that the section belongs to.
+            This is a unique identifier for the Canvas instance
+        :type tool_consumer_instance_guid: str
+
+        :param context_id: the context_id LTI launch param from the Canvas
+            course that the section belongs to. This is a unique identifier for
+            the course within the Canvas instance
+        :type context_id: str
+
+        :param section: a section dict as received from the Canvas API
+        :type section: dict
+        """
+        hash_object = hashlib.sha1()
+        hash_object.update(tool_consumer_instance_guid.encode())
+        hash_object.update(context_id.encode())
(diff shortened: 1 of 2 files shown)


lms/resources/lti_launch.py
```diff
@@ -106,6 +106,36 @@ def h_groupid(self):
         """
         return f"group:{self.h_authority_provided_id}@{self._authority}"
 
+    def h_section_groupid(self, tool_consumer_instance_guid, context_id, section_id):
```
seanh: `tool_consumer_instance_guid` and `context_id` can of course be retrieved from `self.request.params` but they won't be able to be when this function is actually called--it's going to be called by an API view not an LTI launch request and `self.request.params` won't contain all the launch params. Plus it's kind of better to pass them in as arguments anyway.

Many of the other methods in this class are soon going to need to be refactored similarly

seanh:
Note that this method is being added to `LTILaunchResource` for now because that's where all these types of methods go currently, but we have future cards in later sections milestones for factoring many of these methods out into a separate class. So this is just a temporary home for the new method for now

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -186,6 +186,16 @@ def test_h_groupid(self, lti_launch):
...
 
+    def test_h_section_groupid(self, lti_launch):
+        groupid = lti_launch.h_section_groupid(
+            "test_tool_consumer_instance_guid", "test_context_id", "test_section_id"
+        )
+
+        assert (
+            groupid
+            == "group:section-720d7708a7960df6bbfcaae1f3d4ade736544dbf@TEST_AUTHORITY"
+        )
```
jon-betts: I think it might be worth having a parameterised test here to prove we are sensitive to each of the inputs.

```
@pytest.mark.paramterised('args,groupid', (
    (
        ['tool_consumer_instance_guid', 'context_id', 'section_id'],
        ""group:section-..."
    ),
    (
        ['DIFFERENT', 'context_id', 'section_id'],
        ""group:section-..."
    ),
   (
        ['tool_consumer_instance_guid', 'DIFFERENT', 'section_id'],
        ""group:section-..."
    ),
    (
        ['tool_consumer_instance_guid', 'context_id', 'DIFFERENT'],
        ""group:section-..."
    ),
))
```

seanh: I've changed the method to accept a section dict rather than a section ID: I think it makes sense for this method to be responsible for extracting the section ID (or any other parts of the raw section dict that it needs from Canvas) itself, rather than requiring the calling code to.

Also changed this test to be a parametrized one that fully tests every line of the method: you can't break a single line of the method under test without a test failing now

Issue #1567, opened 2020-03-18
Collect all calls to sync data to H into a single interface


jon-betts:
Currently our code in the views calls separate methods to upsert groups, users and user membership. This should all be collected together in a single interface to prepare for bulk calling.

At the end of this ticket:

 * There will be a fascade for bulk calling H
 * This will accept all of the details to update in one go
 * The code will have been altered to use it
 * Internally it will use existing mechanisms to update H as we do now

----

This might be a good time to look at the group upsert weirdness (we seem to create then upsert or something?)

PR #1594, opened 2020-03-24
Change LTIHService to look like it will with the bulk API

jon-betts:
_From linked ticket https://github.com/hypothesis/lms/issues/1567:_

> Currently our code in the views calls separate methods to upsert groups, users and user membership. This should all be collected together in a single interface to prepare for bulk calling.
> 
> At the end of this ticket:
> 
> There will be a fascade for bulk calling H
> This will accept all of the details to update in one go
> The code will have been altered to use it
> Internally it will use existing mechanisms to update H as we do now
> 

The word 'facade' is used here to indicate that although we are accepting data _as if_ we are the bulk API, we are not actually implementing any bulk processing at this point. The existing `LITHService` was chosen as the target.

This is to allow work to progress in tandem with the Bulk API work.

A lot of the tests and work here will get swept away when the guts of this change, so minimal efforts have been made to keep them working and preserve coverage.

Refactoring LTIHService so that:

1. It's interface now looks more like how it will look once it's using the bulk API library
2. The LMS app's actual end-to-end behaviour hasn't changed at all

### DIFF
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -94,9 +94,9 @@ class HAPI:
         except HAPINotFoundError:
             self.create_user(h_user, provider, provider_unique_id)
 
-    def create_group(self, group_id, group_name, creator):
+    def upsert_group(self, group_id, group_name, creator):
         """
-        Create a group in H.
+        Update or create a group in H.
 
         :param group_id: The id of the group
         :param group_name: The display name for the group
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -1,26 +1,10 @@
-from functools import wraps
+from collections import namedtuple
 
 from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
 
 from lms.services import HAPIError, HAPINotFoundError
 
-
-def lti_h_action(function):
-    """Disable an action if provisioning is not enabled and catch HAPIError."""
-
-    @wraps(function)
-    def wrapper(self, *args, **kwargs):
-        # Pylint doesn't understand that self is "us" in this context
-        if not self._context.provisioning_enabled:  # pylint: disable=protected-access
-            return None
-
-        try:
-            return function(self, *args, **kwargs)
-
-        except HAPIError as err:
-            raise HTTPInternalServerError(explanation=err.explanation) from err
-
-    return wrapper
+Group = namedtuple("Group", "name groupid")
 
 
 class LTIHService:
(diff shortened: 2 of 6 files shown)


lms/services/h_api.py
```diff
@@ -94,9 +94,9 @@ def upsert_user(self, h_user, provider, provider_unique_id):
         except HAPINotFoundError:
             self.create_user(h_user, provider, provider_unique_id)
 
-    def create_group(self, group_id, group_name, creator):
+    def upsert_group(self, group_id, group_name, creator):
```
jon-betts: This just isn't the create call, that is a `POST`.

lms/services/lti_h.py
```diff
@@ -1,26 +1,10 @@
...
-        if not self._context.provisioning_enabled:  # pylint: disable=protected-access
-            return None
-
-        try:
-            return function(self, *args, **kwargs)
-
-        except HAPIError as err:
-            raise HTTPInternalServerError(explanation=err.explanation) from err
-
-    return wrapper
```
jon-betts: Now we only use this decorator in one place, it wasn't required anymore.

tests/unit/lms/services/lti_h_test.py
```diff
@@ -1,42 +1,59 @@
...
+        h_api.upsert_user.side_effect = HAPIError
+
+        with pytest.raises(HTTPInternalServerError):
+            lti_h_svc.sync(groups=[group])
+
+    @pytest.fixture
+    def group(self):
+        return Group(name="test_group_name", groupid="test_groupid")
 
-        h_api.update_group.assert_not_called()
```
jon-betts: Below here all the tests are as before. I've just changed the previous individual method tests to be called from the top level. These will all be gone soon.

lms/services/lti_h.py
```diff
@@ -44,71 +28,64 @@ def __init__(self, _context, request):
...
+        for group in groups:
+            self.h_api.add_user_to_group(h_user=h_user, group_id=group.groupid)
 
-    def _upsert_h_group(self, group_id, group_name, creator):
+    def _upsert_h_group(self, group, creator):
         """Update the group and create it if the user is an instructor."""
 
         try:
-            self.h_api.update_group(group_id=group_id, group_name=group_name)
+            self.h_api.update_group(group_id=group.groupid, group_name=group.name)
```
jon-betts: We could push this value object down into h_api, but soon enough all these methods will be gone, so why bother?

seanh:
From the description I'm unclear on what this PR does. Can you clarify? This doesn't just add a new bulk API facade. Rather, it refactors the existing `LTIHService`. Does this change the existing behaviour in any way? Or is it just changing `LTIHService`'s interface in a way that doesn't change its behaviour at all.

I'm curious what might ultimately happen to the `single_group_sync()` method? In the future something is going to have to either create a list of one course group _or_ call the Canvas API and create a list of one-or-more section groups, and pass the list of groups to `LTIHService.sync()`. But who will be responsible for making that course-groups-or-section-groups decision and generating the list of course or section groups (calling the Canvas API if necessary)? Will `LTIHService` do that itself? Or something else?

seanh:
Testing locally everything still seems to work as normal

lms/services/lti_h.py
```diff
@@ -1,26 +1,10 @@
...
-            return None
-
-        try:
-            return function(self, *args, **kwargs)
-
-        except HAPIError as err:
-            raise HTTPInternalServerError(explanation=err.explanation) from err
-
-    return wrapper
+Group = namedtuple("Group", "name groupid")
```
jon-betts: This fella will probably die in time when the library exists, but who knows?

lms/services/lti_h.py
```diff
@@ -118,6 +96,6 @@ def _upsert_h_group(self, group_id, group_name, creator):
                 raise HTTPBadRequest("Instructor must launch assignment first.")
 
         # Try to create the group with the current instructor as its creator.
-        self.h_api.create_group(
-            group_id=group_id, group_name=group_name, creator=creator
+        self.h_api.upsert_group(
+            group_id=group.groupid, group_name=group.name, creator=creator
```
jon-betts: There's something super fishy going on with the logic in this function. It's more obvious with the rename to `upsert_group`. I think we are quietly always an instructor by this point, which means we probably try a pointless "create" for instructors above.

If we flipped this around to test whether you were an instructor first, we could probably skip that. This doesn't matter if we end up deleting the lot though.

jon-betts:
I don't really know. I suppose we'll cross that bridge when we get to it. @seanh What state is this in? Is there anything more you need to know?

Issue #1382, opened 2020-01-29
Refactor: merge h_groupid and h_group_name


seanh:
`LTILaunchResource` contains two properties [`h_groupid` and `h_group_name`](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/resources/_lti_launch.py#L101-L139). They're both [used in the same place](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/lti_h.py#L95-L99) by `LTIHService`.

Since we're now going to be dealing with multiple groups at a time not just one, these two properties need to merge into a single `h_group` property that returns a dict with `"id"` and `"name"` keys. So that in future `h_group` can easily become `h_groups` and return a list of dicts and we'll know which group id's go with which keys.

Note that `h_groupid` is also [used on its own](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/lti_h.py#L60) in one other place, but that will be easy to handle (`h_group["id"]`).

See also:

* Making the new `h_groupid` property plural, which follows on from this card: https://github.com/hypothesis/lms/issues/1407

Issue #1407, opened 2020-01-30
Refactor: Make h_group plural


seanh:
Depends on:

* Merging `h_groupid` and `h_group_name` into a single new property named `h_group`

We're now going to be dealing with multiple Hypothesis groups in a single LTI launch, not just one group all the time. So `h_group` needs to become `h_groups` and return a list. And everything that uses `h_group(s)` needs to deal with it being a list (e.g. iterate over the list and process each group in it).

For example `LTIHService.upsert_course_group()` will have to become `upsert_course_groups()` and accept a list. `HAPIClient`s `create_group()` will have to become `create_groups()` and accept a list.

Note that there won't actually be more than one group in the list until we have support for section groups. It'll just be a list of length 1.

PR #1596, opened 2020-03-25
Combine the groupid and group_name methods

seanh:
Combine the `h_groupid` and `h_group_name` properties into a single
`h_course_group` property, and `h_section_groupid()` and
`h_section_group_name()` into a single `h_section_group()` method.

Fixes https://github.com/hypothesis/lms/issues/1382

In the future these might become plural: `h_course_groups` and
`h_section_groups()` returning lists of one or more groups (see
https://github.com/hypothesis/lms/issues/1407).

We may ultimately want an `h_groups()` method that decides whether a
course group or section groups is needed and returns one or the other.
Or we may want to leave that up to code elsewhere, to decide whether to
call `h_course_groups` or `h_section_groups()`. I'm not sure yet, so I'm
leaving them as separate methods for now.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -376,7 +376,7 @@ class JSConfig:  # pylint:disable=too-few-public-methods
                     "authority": self._request.registry.settings["h_authority"],
                     "enableShareLinks": False,
                     "grantToken": self._grant_token(api_url),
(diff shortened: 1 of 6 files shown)


seanh:
Closing this as I don't think I'm going to use it

PR #1601, opened 2020-03-27
Fix some pylint warnings

seanh:
I'm not sure why `make sure` isn't picking this up but running pylint directly on this file discovered these warnings

### DIFF
--- a/tests/unit/lms/services/_helpers/canvas_api_test.py
+++ b/tests/unit/lms/services/_helpers/canvas_api_test.py
@@ -1,7 +1,8 @@
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/_helpers/canvas_api_test.py
```diff
@@ -1,7 +1,8 @@
 from unittest import mock
 
 import pytest
-from requests import ConnectionError, HTTPError, ReadTimeout, Request, TooManyRedirects
+from requests import ConnectionError as requests_ConnectionError
```
jon-betts: Maybe use `requests.ConnectionError`?

Issue #1405, opened 2020-01-30
Canvas API client: Add support for getting an arbitrary user's sections for a course


seanh:
See also:

* Canvas API client pagination support: https://github.com/hypothesis/lms/issues/1406
* [How can I get a user's sections for a course, from the Canvas API?](https://community.canvaslms.com/thread/43825-how-can-i-get-a-users-sections-for-a-course-from-the-canvas-api) on Canvas's forums

When a teacher launches an assignment in SpeedGrader we want the client's groups menu to contain the Hypothesis groups for all of the course sections that **the student being graded** belongs to (not that the teacher belongs to) (for the current course).

Add support to our Canvas API client for a request that returns a list of all an arbitrary user's sections, including both the section IDs and names, for a given course.

This is different than getting a list of the authenticated user's sections (https://github.com/hypothesis/lms/issues/1403) because in SpeedGrader the authenticated user is the teacher but the user whose sections we want is a different user: the student.

As far as I know it's not possible to do this in a single Canvas API request. You have to send one request to get the user's sections:

    http "https://FOO.instructure.com/api/v1/courses/<COURSE_ID>/users/<USER_ID>?include[]=enrollments" Authorization:"Bearer 9382~1sN***oaZ"

but it'll only return the section IDs not the names. You then need to send a second request to get _all_ of the course's sections (which will return the section names along with the IDs) and you can find the section IDs from your first request in the second request and discover the corresponding names. Both requests need pagination support as well.

Issue #1404, opened 2020-01-30
Canvas API client: Add support for getting all a course's sections


seanh:
See also:

* Canvas API client pagination support: https://github.com/hypothesis/lms/issues/1406

When a teacher launches an assignment we want the client's groups menu to contain the Hypothesis groups for _all_ of the course sections for the current course, regardless of which sections the teacher belongs to.

Add support to our Canvas API client for a request that returns a list of all the course's sections, including both the section IDs and names, for a given course.

Here's an example of the Canvas API request it needs to make:

```
http GET 'https://hypothesis.instructure.com/api/v1/courses/<COURSE_ID>/sections' Authorization:'Bearer 9382~p5X***L9y'
```

Issue #1403, opened 2020-01-30
Canvas API client: Add support for getting the authenticated user's sections for a course


seanh:
See also:

* Canvas API client pagination support: https://github.com/hypothesis/lms/issues/1406

When a student launches an assignment we want the client's groups menu to contain the Hypothesis groups for all of the course sections that the student belongs to (for the current course).

Add support to our Canvas API client for a request that returns a list of all the authenticated user's sections, including both the section IDs and names, for a given course.

Here's an example of the Canvas API request it needs to make:

```
$ http GET 'https://hypothesis.instructure.com/api/v1/courses/<COURSE_ID>?include[]=sections' Authorization:'Bearer 9382~c7I***A8v'
{
  "id": 100,
  "name": "Sections 101",
  ...
  "sections": [
    {
      "id": 107,
      "name": "Section 1",
      "start_at": null,
      "end_at": null,
      "enrollment_role": "StudentEnrollment"
    },
    {
      "id": 108,
      "name": "Section 2",
      "start_at": null,
      "end_at": null,
      "enrollment_role": "StudentEnrollment"
    }
  ],
  ...
}
```

PR #1608, opened 2020-03-30
Add support for getting sections from the Canvas API

seanh:
Add support for getting sections from the Canvas API in various ways.

Fixes https://github.com/hypothesis/lms/issues/1403

Fixes https://github.com/hypothesis/lms/issues/1404

Fixes https://github.com/hypothesis/lms/issues/1405

Note that there are no proxy API views/endpoints for these new Canvas API calls. There are only service methods for making the calls. These APIs aren't meant to be called by our frontend code directly. They'll be called by our backend code as appropriate (in response to sync API calls from the frontend)

There are three different circumstances under which we're going to need to retrieve sections from the Canvas API, in three different ways:

1. When a teacher (or any non-student) launches an assignment we want them to see the section groups for _all_ of the course sections in the Hypothesis client.

   We need to retrieve the full list of all of the course's sections including both the section IDs and names, so that we can both create those section groups in h and configure the client to show them. 

   The list of sections might change at any time (sections can be added, deleted or renamed in Canvas at any time) so we need to retrieve the list of sections and upsert the section groups into h on every launch.

   For this case we use **Canvas's sections API**, which has an endpoint for listing sections by course: https://canvas.instructure.com/doc/api/sections.html#method.sections.index

2. When a student launches an assignment we only want the client to show the section groups for the Canvas course sections that the student currently belongs to.

   We need to retrieve the student's sections including both the section IDs and names, so that we can both create those section groups in h and configure the client to show them.

   The list of sections might change at any time (sections can be added, deleted or renamed in Canvas at any time, and students can also be added to and removed from sections at any time) so we need to retrieve the list of sections and upsert the section groups into h on every launch.

   For this case we have to use **Canvas' _courses_ API** with `?include[]=sections` param.
This allow us to retrieve **the authenticated user's** sections only and not other sections. And since the student is the authenticated user, this works: https://canvas.instructure.com/doc/api/courses.html#method.courses.show

3. When a teacher launches an assignment in SpeedGrader then the teacher is the authenticated user but we want the client to show the section groups for the sections of the student who is currently being graded.

   In this case we can probably get away with not upserting the section groups into h. We can assume that all of the section groups will already exist in h or at least that if the student now belongs to any new sections that don't exist in h yet, the student can't have created any annotations in those groups yet so it would be okay not to show those groups in SpeedGrader.

   If we don't need to upsert the groups into h then we don't need the section names.

   We do need the section IDs, in order to configure the client to show those groups.

   If a student got removed from a section (or a section got deleted) after the student had already made some annotations for the assignment in that section group, those annotations would no longer show up in SpeedGrader. This would probably be acceptable, it's pretty edge casey.

   For this case we're using the **courses API endpoint** for getting information about a given user in the context of a given course: https://canvas.instructure.com/doc/api/courses.html#method.courses.user And we're using the `?include[]=enrollments` param to ask Canvas to include the student's enrollments in the course in that info. Students seem to have one course enrollment per course section they belong to. And the enrollment dicts contain `"course_section_id"` fields which are the section IDs. They do not contain the section names.

   I don't know of any way to get both the section IDs and names for all of an arbitrary user (not the authenticated user)'s sections for a course, without stitching together data from multiple different Canvas API endpoints.

### DIFF
--- a/lms/services/_helpers/canvas_api.py
+++ b/lms/services/_helpers/canvas_api.py
@@ -91,6 +91,72 @@ class CanvasAPIHelper:
             },
         ).prepare()
 
+    def authenticated_users_sections_request(self, access_token, course_id):
+        """
+        Return a prepared "authenticated users sections" request.
+
+        Return a server-to-server request to the Canvas API that gets a list of
+        the authenticated user's sections for the given course (course_id).
+
+        For documentation of this request see:
+
+        https://canvas.instructure.com/doc/api/courses.html#method.courses.index
+
+        :arg access_token: the access token to authenticate the request with
+        :type access_token: str
+
+        :arg course_id: the Canvas course_id of the course to look in
+        :type course_id: str
+
+        :rtype: requests.PreparedRequest
+        """
+        # Canvas's sections API
+        # (https://canvas.instructure.com/doc/api/sections.html) only allows
+        # you to get _all_ of a course's sections, it doesn't provide a way to
+        # get only the sections that the authenticated user belongs to. So we
+        # have to get the authenticated user's sections from part of the
+        # response from a courses API endpoint instead.
+        #
+        # Canvas's "Get a single course" API is capable of doing this if the
+        # ?include[]=sections query param is given:
+        #
+        #    https://canvas.instructure.com/doc/api/courses.html#method.courses.show
+        #
+        # The ?include[]=sections query param is documented elsewhere (in the
+        # "List your courses" API:
+        # https://canvas.instructure.com/doc/api/courses.html#method.courses.index)
+        # as:
+        #
+        #    "Section enrollment information to include with each Course.
+        #    Returns an array of hashes containing the section ID (id), section
+        #    name (name), start and end dates (start_at, end_at), as well as the
+        #    enrollment type (enrollment_role, e.g. 'StudentEnrollment')."
+        #
+        # In practice ?include[]=sections seems to add a "sections" key to the
+        # API response that is a list of section dicts, one for each section
+        # the authenticated user is currently enrolled in, each with the
+        # section's "id" and "name" among other fields.
+        #
+        # **We don't know what happens if the user belongs to a really large
+        # number of sections**. Does the list of sections embedded within the
+        # get course API response just get really long? Does it get truncated?
+        # Can you paginate through it somehow? This seems edge-casey enough
+        # that we're ignoring it for now.
+        url = urlunparse(
+            (
+                "https",
+                self._canvas_url,
+                f"/api/v1/courses/{course_id}" "",
+                "",
+                "include[]=sections",
+                "",
+            )
+        )
+
+        return requests.Request(
+            "GET", url, headers={"Authorization": f"Bearer {access_token}"},
+        ).prepare()
+
     def list_files_request(self, access_token, course_id):
         """
         Return a prepared list files request.
--- a/lms/services/_helpers/canvas_api.py
+++ b/lms/services/_helpers/canvas_api.py
@@ -146,7 +146,7 @@ class CanvasAPIHelper:
             (
                 "https",
                 self._canvas_url,
-                f"/api/v1/courses/{course_id}" "",
+                f"/api/v1/courses/{course_id}",
                 "",
                 "include[]=sections",
                 "",
@@ -180,7 +180,7 @@ class CanvasAPIHelper:
             (
                 "https",
                 self._canvas_url,
-                f"/api/v1/courses/{course_id}/sections" "",
+                f"/api/v1/courses/{course_id}/sections",
                 "",
                 "",
                 "",
--- a/lms/services/_helpers/canvas_api.py
+++ b/lms/services/_helpers/canvas_api.py
@@ -191,6 +191,44 @@ class CanvasAPIHelper:
             "GET", url, headers={"Authorization": f"Bearer {access_token}"},
         ).prepare()
 
+    def users_sections_request(self, access_token, user_id, course_id):
+        """
+        Return a prepared "user's course sections" request.
+
+        Return a server-to-server request to the Canvas API that gets a list of
+        all the given user's (user_id) sections for the given course
+        (course_id).
+
+        For documentation of this request see:
+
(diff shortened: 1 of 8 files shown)


lms/services/_helpers/canvas_api.py
```diff
@@ -91,6 +91,72 @@ def refresh_token_request(self, refresh_token):
...
+        url = urlunparse(
+            (
+                "https",
+                self._canvas_url,
+                f"/api/v1/courses/{course_id}" "",
+                "",
+                "include[]=sections",
+                "",
+            )
+        )
```
seanh: I'm just gonna stick with `urlunparse` here because that's what the rest of the methods in this module do

lms/services/canvas_api.py
```diff
@@ -65,6 +69,35 @@ def get_refreshed_token(self, refresh_token):
...
+
+        :raise lms.services.CanvasAPIAccessTokenError: if we can't get the list
+            of sections because we don't have a working Canvas API access token
+            for the user
+        :raise lms.services.CanvasAPIServerError: if we do have an access token
+            but the Canvas API request fails for any other reason
+
+        :return: a list of raw section dicts as received from the Canvas API
+        :rtype: list(dict)
+        """
```
seanh: Adding support for a new Canvas API endpoint involves adding code in three separate places:

1. In the Canvas API helper (above)
2. In the Canvas API service (here)
3. In the Canvas API validation schemas (below)

It'd be nice if these three separate locations could be brought together so that adding a new endpoint only involved adding code in one place, and didn't involve having to explain the same thing three times in three separate docstrings.

The code design is perfectly workable though and I'm more focused on making forward progress with the sections feature right now, so I'm not going to refactor the code now.

tests/unit/lms/validation/_canvas_test.py
```diff
@@ -131,12 +145,90 @@ def test_it_raises_ValidationError_if_the_response_body_isnt_valid_json(
...
+    )
+    def test_invalid(self, schema, invalid_body, error_messages):
+        schema.context["response"].json.return_value = invalid_body
+
+        with pytest.raises(ValidationError) as exc_info:
+            schema.parse()
+
+        assert exc_info.value.messages == {
+            "sections": error_messages,
+        }
```
seanh: Only about six lines of test code are needed to test `CanvasAuthenticatedUsersSectionsResponseSchema` (counting the lines in this test method plus the one other test method above), and then there are about 50 lines of declarative pytest fixtures covering all the possible failure cases. I think this is fine: there's not much actual test _code_ here, and I'm fine with throwing a large number of cases at the schema. Each of the test cases tests something about the `CanvasAuthenticatedUsersSectionsResponseSchema` class code itself, each test case could be broken by a change to `CanvasAuthenticatedUsersSectionsResponseSchema`

tests/unit/lms/validation/_canvas_test.py
```diff
@@ -131,12 +145,90 @@ def test_it_raises_ValidationError_if_the_response_body_isnt_valid_json(
...
+        ]
+
+    @pytest.mark.parametrize(
+        "invalid_body,error_messages",
+        [
+            pytest.param(
+                {},
+                ["Missing data for required field."],
+                id="There must be a sections key",
+            ),
```
seanh: It's important that we test the actual error messages from Marshmallow. Otherwise a test for one failure case could be triggering a different failure case instead and the test could be passing, when in fact the failure case that was _supposed_ to be tested might _not_ actually be failing. This kind of test mistake happens quite often if you don't use specific-enough asserts.

It so happens that the way marshmallow distinguishes one test failure from another is by its dicts-of-lists-of-strings error message format and not by, for example, raising a different exception class for each possible failure. So that's what we have to work with. (And marshmallow's design here makes good sense for a validation lib.)

lms/services/_helpers/canvas_api.py
```diff
@@ -91,6 +91,72 @@ def refresh_token_request(self, refresh_token):
...
+        # In practice ?include[]=sections seems to add a "sections" key to the
+        # API response that is a list of section dicts, one for each section
+        # the authenticated user is currently enrolled in, each with the
+        # section's "id" and "name" among other fields.
+        #
+        # **We don't know what happens if the user belongs to a really large
+        # number of sections**. Does the list of sections embedded within the
+        # get course API response just get really long? Does it get truncated?
+        # Can you paginate through it somehow? This seems edge-casey enough
+        # that we're ignoring it for now.
```
seanh: We're having to do somewhat obscure stuff with the Canvas API to get the sections info that we need, unfortunately. I don't think there's any other way. So I think the extensive code comment is warranted

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -198,6 +198,74 @@ def old_oauth2_token(self, db_session, pyramid_request):
...
+
+        # It sends the prepared request.
+        send_with_refresh_and_retry.assert_called_once_with(
+            canvas_api_helper.authenticated_users_sections_request.return_value,
+            CanvasAuthenticatedUsersSectionsResponseSchema,
+            "test_refresh_token",
+        )
+
+        # It returns the sections.
+        assert sections == send_with_refresh_and_retry.return_value
```
seanh: Great big integration test here because the method under test is an integration method that does nothing other than integrate a few other things together. It'd be nicer if the code design was such that an integration method wasn't needed, or at least was much smaller, then this integration test might not be needed or could be much simpler. Again, though, resisting the urge to redesign at this point

seanh:
This _could_ be reviewed now, but I'm also going right to work on adding support for two more of these endpoints so it might make sense to group them all into one PR when done

seanh:
Added a second commit that adds the "all of a course's sections" API (https://github.com/hypothesis/lms/issues/1404)

lms/services/_helpers/canvas_api.py
```diff
@@ -91,6 +91,144 @@ def refresh_token_request(self, refresh_token):
...
+        url = urlunparse(
+            (
+                "https",
+                self._canvas_url,
+                f"/api/v1/courses/{course_id}" "",
```
jon-betts:
```suggestion
                f"/api/v1/courses/{course_id}",
```

PR #1612, opened 2020-04-01
Refactor h_api_test.py

seanh:
Tidy up/simplify/make-consistent this test file, to make it easier to edit when we edit `h_api.py`

### DIFF
--- a/tests/unit/lms/services/h_api_test.py
+++ b/tests/unit/lms/services/h_api_test.py
@@ -3,7 +3,6 @@ from unittest.mock import call, patch, sentinel
 import pytest
 import requests as requests_
 from h_matchers import Any
-from pyramid.testing import DummyRequest
 from requests import (
     HTTPError,
     ReadTimeout,
@@ -15,9 +14,6 @@ from requests import (
 from lms.services import HAPIError, HAPINotFoundError
 from lms.services.h_api import HAPI
 from lms.values import HUser
-
-BASE_URL = "http://example.com/base/"
-AUTHORITY = "test-authority"
 
 
 class TestHAPI:
@@ -34,11 +30,11 @@ class TestHAPI:
         user = h_api.get_user("username")
 
         h_api._api_request.assert_called_once_with(
-            "GET", path="users/acct:username@test-authority"
+            "GET", path="users/acct:username@TEST_AUTHORITY"
         )
 
         assert user == Any.instance_of(HUser)
-        assert user.authority == AUTHORITY
+        assert user.authority == "TEST_AUTHORITY"
         assert user.username == "username"
         assert user.display_name == sentinel.display_name
 
@@ -51,7 +47,7 @@ class TestHAPI:
             data={
                 "username": h_user.username,
                 "display_name": h_user.display_name,
-                "authority": AUTHORITY,
+                "authority": "TEST_AUTHORITY",
                 "identities": [
                     {
                         "provider": sentinel.provider,
@@ -105,7 +101,7 @@ class TestHAPI:
 
         h_api._api_request.assert_called_once_with(
             "POST",
-            "groups/sentinel.group_id/members/acct:sentinel.username@test-authority",
+            "groups/sentinel.group_id/members/acct:sentinel.username@TEST_AUTHORITY",
         )
 
     @pytest.fixture
@@ -113,12 +109,12 @@ class TestHAPI:
         return HUser(
             username=sentinel.username,
             display_name=sentinel.display_name,
-            authority=AUTHORITY,
+            authority="TEST_AUTHORITY",
         )
 
     @pytest.yield_fixture
-    def h_api(self, p_request):
-        h_api = HAPI(sentinel.context, p_request)
+    def h_api(self, pyramid_request):
+        h_api = HAPI(sentinel.context, pyramid_request)
 
         with patch.object(HAPI, "_api_request"):
             yield h_api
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/h_api_test.py
```diff
@@ -16,42 +15,39 @@
 from lms.services.h_api import HAPI
 from lms.values import HUser
 
-BASE_URL = "http://example.com/base/"
-AUTHORITY = "test-authority"
```
seanh, line 20: The tests now just use the global `pyramid_request` fixture, instead of creating their own `p_request` fixture and adding these constants to it, so these constants are no longer needed. Also good to remove constants floating around at the top of test files.

tests/unit/lms/services/h_api_test.py
```diff
@@ -16,42 +15,39 @@
...
 
 class TestHAPI:
 
     # We're accessing h_api._api_request a lot in this test class, so disable
     # protected-access messages.
     # pylint: disable=protected-access
 
-    def test_get_user_works(self, h_api):
-        h_api._api_request.return_value.json.return_value = {
+    def test_get_user_works(self, h_api, _api_request):
```
seanh, line 25: The new `_api_request` fixture is a mocked/patched version of `HAPI._api_request()`, and is used by all the tests that want a mocked version of that method to assert against. Tests that _don't_ want `_api_request()` to be mocked just don't use this fixture

tests/unit/lms/services/h_api_test.py
```diff
@@ -16,42 +15,39 @@
...
+        _api_request.return_value.json.return_value = {
             "display_name": sentinel.display_name
         }
 
         user = h_api.get_user("username")
 
-        h_api._api_request.assert_called_once_with(
-            "GET", path="users/acct:username@test-authority"
+        _api_request.assert_called_once_with(
+            "GET", path="users/acct:username@TEST_AUTHORITY"
```
seanh, line 33: The value of authority has changed because we're now using the global `pyramid_request` fixture instead of the test module creating its own similar-but-not-quite-the-same `p_request` fixture

tests/unit/lms/services/h_api_test.py
```diff
@@ -61,17 +57,15 @@ def test_create_user_works(self, h_api, h_user):
...
 
-        h_api._api_request.assert_called_once_with(
+        _api_request.assert_called_once_with(
             "PATCH",
             "users/sentinel.username",
             data={"display_name": h_user.display_name},
         )
 
-    @patch.object(HAPI, "create_user")
-    @patch.object(HAPI, "update_user")
```
seanh, line 74: Replaced these with pytest fixtures to be more consistent with how we patch things throughout our tests (including other patch fixtures in this same test module) and because pytest fixtures are more powerful as they can compose with other fixtures

tests/unit/lms/services/h_api_test.py
```diff
@@ -83,83 +77,64 @@ def test_upsert_calls_calls_create_if_update_fails(
...
+                # It requests the given URL.
+                url="https://example.com/private/api/dummy-path",
+                # It adds the authentication headers.
+                auth=("TEST_CLIENT_ID", "TEST_CLIENT_SECRET"),
+                # Requests time out after 10 seconds.
+                timeout=10,
+                # It adds the Hypothesis-Application header.
+                headers={"Hypothesis-Application": "lms"},
+            )
+        ]
```
seanh, line 121: This was previously a few separate tests plus a helper method (with matcher usage!) to help them make assertions. I just combined them into one simple, standard mock assert

tests/unit/lms/services/h_api_test.py
```diff
@@ -83,83 +77,64 @@ def test_upsert_calls_calls_create_if_update_fails(
...
+
+    def test_if_given_a_data_dict__api_request_dumps_it_to_a_json_string(
+        self, h_api, requests
+    ):
         h_api._api_request(sentinel.method, "dummy-path", {"a": 1, "b": [2]})
 
-        self._assert_called_requests_with(requests, data='{"a":1,"b":[2]}')
+        assert requests.request.call_args[1]["data"] == '{"a":1,"b":[2]}'
```
seanh, line 128: Replacing the `self._assert_called_requests_with()` helper with normal mock library usage

tests/unit/lms/services/h_api_test.py
```diff
@@ -83,83 +77,64 @@ def test_upsert_calls_calls_create_if_update_fails(
...
+    def test_if_given_custom_headers__api_request_adds_them(self, h_api, requests):
         h_api._api_request(
             sentinel.method, "dummy-path", headers={"X-Header": sentinel.header}
         )
 
-        self._assert_called_requests_with(
-            requests,
-            headers={"X-Header": sentinel.header, "Hypothesis-Application": "lms"},
-        )
+        assert requests.request.call_args[1]["headers"]["X-Header"] == sentinel.header
```
seanh, line 135: Replacing the `self._assert_called_requests_with()` helper with normal mock library usage

tests/unit/lms/services/h_api_test.py
```diff
@@ -187,44 +162,40 @@ def test_it_raises_HAPIError_for_other_http_errors(
...
+            display_name=sentinel.display_name,
+            authority="TEST_AUTHORITY",
+        )
 
-        values.update(extras)
+    @pytest.fixture
+    def h_api(self, pyramid_request):
+        return HAPI(sentinel.context, pyramid_request)
 
-        requests.request.assert_called_once_with(**values)
```
seanh, line 207: I'm not a fan of test helpers like `_assert_called_requests_with()` except when really needed. Certainly not when they can easily be replaced with standard mock library asserts, as in this case. They make tests harder to understand (you now need to scroll down to this undocumented test helper method and understand what it does, before you can understand the test that uses). They also complicate test code, and clever test code (usually added in the name of avoiding test code duplication) often results in buggy tests. If not careful they can easily end up coupling tests together in undesirable ways too.

tests/unit/lms/services/h_api_test.py
```diff
@@ -83,83 +77,64 @@ def test_upsert_calls_calls_create_if_update_fails(
...
-    @pytest.fixture
-    def h_user(self):
-        return HUser(
-            username=sentinel.username,
-            display_name=sentinel.display_name,
-            authority=AUTHORITY,
+            "groups/sentinel.group_id/members/acct:sentinel.username@TEST_AUTHORITY",
         )
 
-    @pytest.yield_fixture
```
seanh, line 119: FYI `yield_fixture` is deprecated

Issue #1400, opened 2020-01-30
Automatically create a "machine user" in h and use it to create groups


seanh:
See also:

* Migrating existing LMS group creators in h to be the new machine user: https://github.com/hypothesis/lms/issues/1400

When creating groups in h the LMS app [uses an `X-Forwarded-User` header](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/h_api.py#L109) to tell h which user it is creating the group "on behalf of" (see the [h API docs about `X-Forwarded-User`](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/h_api.py#L109)). The forwarded user ends up as the group's creator in h.

The user who we actually pass as the `X-Forwarded-User` is just [whoever is the authenticated user launching the LMS app](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/lti_h.py#L98). The sequence of events ensures that this is always a teacher on the course, not a student: since LMS's ensure that only teachers can create assignments, and you have to launch our LMS app in order to create or configure an assignment, it's always a teacher who is first to launch our app. If a student somehow launches our app without the course's group having been created yet we see that as an "unconfigured assignment" and show an error page explaining that a teacher must launch the assignment first.

There are some problems with this first-teacher-creates-group approach:

* Using the first teacher as the group's creator is a little odd, since an LMS course can have multiple teachers and this first teacher can also leave the course and no longer be a teacher

* The sequence of events that it relies on no longer works when we're creating one h group for every course section rather than just one h group for the course.

  New sections can be added to a course at any time, including mid-course when Hypothesis assignments have already been created and launched by teachers and students. Whenever anyone--teacher or student--launches our app at any time and we retrieve the list of course sections from the Canvas API that list may contain new sections that we haven't seen before and that we need to create new groups in h for.

  We can't start showing all students the "Teacher must launch assignment first" error page (breaking all assignments in the course until a teacher launches one of them) every time a new section was added mid-course.

To get around this problem the LMS app needs to create a ["Machine User"](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) and use that user to create groups rather than using a real teacher. This will mean that we can create new section groups in response to any LTI launch, whether from a student or a teacher.

The LMS app should:

[...]

PR #1614, opened 2020-04-01
Use a machine user to create groups

seanh:
Automatically create a machine user in h and use it as the group's creator when creating groups.

Fixes https://github.com/hypothesis/lms/issues/1400

### DIFF
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -94,19 +94,18 @@ class HAPI:
         except HAPINotFoundError:
             self.create_user(h_user, provider, provider_unique_id)
 
-    def upsert_group(self, group_id, group_name, creator):
+    def upsert_group(self, group_id, group_name):
         """
         Update or create a group in H.
 
         :param group_id: The id of the group
         :param group_name: The display name for the group
-        :param creator: The user creating the group
         """
         self._api_request(
             "PUT",
             f"groups/{group_id}",
             data={"groupid": group_id, "name": group_name},
-            headers={"X-Forwarded-User": creator.userid},
+            headers={"X-Forwarded-User": f"acct:lms@{self._authority}"},
         )
 
     def update_group(self, group_id, group_name):
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -101,12 +101,25 @@ class HAPI:
         :param group_id: The id of the group
         :param group_name: The display name for the group
         """
-        self._api_request(
-            "PUT",
-            f"groups/{group_id}",
-            data={"groupid": group_id, "name": group_name},
-            headers={"X-Forwarded-User": f"acct:lms@{self._authority}"},
-        )
+
+        def do_upsert_group():
+            """Send an upsert group request to h."""
+            self._api_request(
+                "PUT",
+                f"groups/{group_id}",
+                data={"groupid": group_id, "name": group_name},
+                headers={"X-Forwarded-User": f"acct:lms@{self._authority}"},
+            )
+
+        try:
+            do_upsert_group()
+        except HAPINotFoundError:
+            # If we get a 404 when trying to upsert a group that must mean
+            # that the lms user doesn't exist in h yet.
+            self.create_user(
+                HUser(self._authority, "lms"), provider="lms", provider_unique_id="lms",
+            )
+            do_upsert_group()
 
     def update_group(self, group_id, group_name):
         """
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -121,19 +121,6 @@ class HAPI:
             )
             do_upsert_group()
 
-    def update_group(self, group_id, group_name):
-        """
-        Update a group in H.
-
-        Currently this only updates the name of the group.
-
-        :param group_id: The id of the group
-        :param group_name: The display name for the group
-        """
-        self._api_request(
-            "PATCH", f"groups/{group_id}", data={"name": group_name},
-        )
-
     def add_user_to_group(self, h_user, group_id):
         """
         Add the user as a member of the group.
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -1,6 +1,6 @@
 from collections import namedtuple
 
-from pyramid.httpexceptions import HTTPBadRequest, HTTPInternalServerError
+from pyramid.httpexceptions import HTTPInternalServerError
 
 from lms.services import HAPIError, HAPINotFoundError
 
@@ -75,14 +75,8 @@ class LTIHService:
                 self.h_api.update_group(group_id=group.groupid, group_name=group.name)
             except HAPINotFoundError:
                 # The group hasn't been created in h yet.
-                if not self._request.lti_user.is_instructor:
-                    raise HTTPBadRequest("Instructor must launch assignment first.")
-
-                # Try to create the group with the current instructor as its creator.
                 self.h_api.upsert_group(
-                    group_id=group.groupid,
-                    group_name=group.name,
-                    creator=self._context.h_user,
+                    group_id=group.groupid, group_name=group.name,
                 )
 
             self.group_info_service.upsert(
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -2,7 +2,7 @@ from collections import namedtuple
 
 from pyramid.httpexceptions import HTTPInternalServerError
 
-from lms.services import HAPIError, HAPINotFoundError
+from lms.services import HAPIError
 
 Group = namedtuple("Group", "name groupid")
 
@@ -71,13 +71,7 @@ class LTIHService:
         )
 
         for group in groups:
-            try:
-                self.h_api.update_group(group_id=group.groupid, group_name=group.name)
-            except HAPINotFoundError:
-                # The group hasn't been created in h yet.
-                self.h_api.upsert_group(
-                    group_id=group.groupid, group_name=group.name,
-                )
+            self.h_api.upsert_group(group_id=group.groupid, group_name=group.name)
 
             self.group_info_service.upsert(
                 authority_provided_id=self._context.h_authority_provided_id,
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -71,7 +71,19 @@ class LTIHService:
         )
 
         for group in groups:
-            self._upsert_h_group(group, creator=self._context.h_user)
+            try:
+                self.h_api.update_group(group_id=group.groupid, group_name=group.name)
+            except HAPINotFoundError:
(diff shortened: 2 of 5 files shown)


lms/services/h_api.py
```diff
@@ -94,20 +94,32 @@ def upsert_user(self, h_user, provider, provider_unique_id):
...
+
+        try:
+            do_create_group()
+        except HAPINotFoundError:
+            # If we get a 404 when trying to create a new group that must mean
+            # that the lms user doesn't exist in h yet.
+            self.create_user(
+                HUser(self._authority, "lms"), provider="lms", provider_unique_id="lms",
+            )
+            do_create_group()
```
seanh, line 122: The create-machine-user logic is specific to h's create-group endpoint. That endpoint 404s if the X-Forwarded-User doesn't exist, so we can detect that situation and create the user. I don't know whether the same is true for other API endpoints. In practice we only use X-Forwarded-User when creating groups. So this logic is implemented here in `create_group()`

lms/services/lti_h.py
```diff
@@ -91,33 +91,17 @@ def upsert_course_group(self):
...
-                raise HTTPBadRequest("Instructor must launch assignment first.")
-
-        # Try to create the group with the current instructor as its creator.
-        self.h_api.create_group(
-            group_id=group_id, group_name=group_name, creator=creator
+        self.group_info_service.upsert(
+            authority_provided_id=self._context.h_authority_provided_id,
+            consumer_key=self._request.lti_user.oauth_consumer_key,
+            params=self._request.params,
         )
```
seanh, line 107: This method got refactored a bit, simplifying things by moving a separate private helper method into the main method. The only behaviour change is that it no longer passes `creator` to `create_group()`

tests/unit/lms/services/lti_h_test.py
```diff
@@ -68,23 +68,6 @@ def test_it_raises_if_creating_the_group_fails(
...
-            roles="learner",
-        )
-        h_api.update_group.side_effect = HAPINotFoundError()
-
-        with pytest.raises(
-            HTTPBadRequest, match="Instructor must launch assignment first"
-        ):
-            lti_h_svc.upsert_course_group()
-
-        h_api.create_group.assert_not_called()
```
seanh, line 86: We used to show an error page if the group didn't exist in h yet and the user wasn't a teacher. I'm not sure how often that actually happened in practice, but the code is simpler now as the error case has been removed.

lms/services/lti_h.py
```diff
@@ -91,33 +91,17 @@ def upsert_course_group(self):
...
+        group_id = self._context.h_groupid
+        group_name = self._context.h_group_name
 
         try:
             self.h_api.update_group(group_id=group_id, group_name=group_name)
-            return
-
         except HAPINotFoundError:
             # The group hasn't been created in h yet.
+            self.h_api.create_group(group_id=group_id, group_name=group_name)
```
seanh, line 101: Since we no longer have different logic for students vs teachers (both could update the group but we would only _create_ the group if it was a teacher) this could now use the [existing group upsert API](https://h.readthedocs.io/en/latest/api-reference/v2/#tag/groups/paths/~1groups~1{id}/put) which would simplify this code. But there seems little point as the whole thing is going to move to the bulk API soon anyway. And I don't think that existing group upsert API has had much exercise in production

lms/services/h_api.py
```diff
@@ -94,20 +94,32 @@ def upsert_user(self, h_user, provider, provider_unique_id):
         except HAPINotFoundError:
             self.create_user(h_user, provider, provider_unique_id)
 
-    def create_group(self, group_id, group_name, creator):
+    def create_group(self, group_id, group_name):
```
jon-betts, line 97:
```suggestion
    def upsert_group(self, group_id, group_name):
```

I think this change has finally brought to the surface some weirdness that's been going on here for a while. The `PUT ` method for groups is actually upsert, not create:

 * https://h.readthedocs.io/en/latest/api-reference/v1/#tag/groups/paths/~1groups~1{id}/put

lms/services/h_api.py
jon-betts: Unless I'm missing something, doesn't this happen once ever, globally? If that's the case I don't think we should be handling this in every call.

Would this be better as a database migration step to add the user?

lms/services/lti_h.py
jon-betts, line 101:
```suggestion
            self.h_api.upsert_group(group_id=group_id, group_name=group_name)
```

This is the part that the upsert makes very weird now. We first try to update the group, and if that fails we then upsert it. I don't think we need the first call at all anymore.

seanh:
I'd rather stick to the approach here than change this to a DB migration script:

* It's not doing anything on every call. Only when the call fails because the user doesn't exist, does it handle that by creating the user
* This will just work in dev environments both when we first merge this, and in the future when people wipe their dev DBs. We don't have to send a PR to the devdata repo, people won't have to run `make devdata` again
* This will just work in QA and prod when we deploy it, we won't have to write a DB migration script, deploy that, run it, and then deploy this separately
* If anyone ever deployed a new instance of the the LMS app to a new production environment it would just work, without them having to manually create the lms user in h

The cost is a _little_ more complexity in the code and tests. But it's more robust, and easier to handle in terms of deployments and managing dev envs

lms/services/lti_h.py
seanh: Hmm. Actually I think this might have made sense before. The logic was:

1. Whether the user is a teacher or a student, try to update the group
2. If the update fails:
   1. If the user is a student show an error page, since students couldn't create (or upsert) groups
   2. If the user is a teacher upsert the group

The user of upsert rather than update in the final step looks odd but upsert does not have the concurrency issues that create has so in fact it made sense to use upsert rather than create here.

And it _did_ have to be update and then upsert, because of the student vs teacher issue. 

It could have been done differently, and this would have been less confusing: for a student it could have been update then show error page if it fails, whereas for a teacher it could have just gone straight to update. I think the reason it ended up like it is, is probably because it originally used _create_ for the teacher because there was no upsert, and then that was changed to upsert later :man_shrugging: 

The naming of the `HAPI` method as `create_group()` when it actually does an upsert is wrong. But that name was introduced in a recent refactoring that I think probably got confused by the above and got the name wrong: https://github.com/hypothesis/lms/commit/d07495be2d1ab12f9281a74e5dcfad5eb0355ec4

seanh: I want to minimise unnecessary changes to `LTIHService` and `HAPI` since they're scheduled to be completely changed to use the new bulk API ASAP. So any changes in the meantime just seem like unnecessary risk. Nonetheless, now that we're using the machine user it does seem to make sense just to call the upsert method right away so I'll do that

jon-betts: I'm pretty sure I introduced the wrong naming here, as I had to reverse engineer what all these actions were doing when they were un-named. Win some lose some.

I'm not 100% sure if we're on the same page here, but I understood why it used to make sense, but I'm saying it no longer does.

jon-betts:
If the authority changes from dev, QA to live is this going to break the bulk stuff which expects your authority to be exactly "lms.hypothes.is"?

`make devdata` is a really good tool, and I think we need to get ourselves in a place where we feel free to use it. 

If it's a case of keeping people up to date with these changes, I'm happy to do that. It should be a case of letting people know and then expecting them to do it. The PRs required are not big. I think we can / should expect people to do this sort of thing regardless. It's a good opportunity to push `superdev` as well, as it does this all for you.

I suspect we're coding to avoid talking here, and I'm happy to do the chatting.

seanh:
I'm gonna stick to my guns here and say that the solution implemented in this PR (which is the same solution as outlined on the issue back in January: https://github.com/hypothesis/lms/issues/1400) is the best solution here, for a few reasons:

1. It's a good solution:
   1. It does add the user once only and never re-adds or checks for it again. It's not checking for the user on every call
   1. It's really simple. It involved adding a total of 4 lines of code. And some tests.
   1. It'll just work in people's dev envs when we merge this, without devs having to do anything (such as re-run `make devdata`)
   1. It'll just work in people's dev envs after a DB reset, without devs having to do anything.
   1. It'll just work on QA and prod after a deploy, without us having to do anything (such as deploy and run a DB migration)
1. `make devdata` isn't quite as good a solution as this one, for people's dev envs. It would require us to go away and add this to `make devdata`; require devs to re-run `make devdata` (and us to communicate that to them); would only work for our team not external developers. We should limit usage of `make devdata` to when it's really needed
1. A DB migration isn't as good a solution as this one, for QA and prod. It would require us to go away and write a migration script, deploy it and run it. It's also an abuse of DB migration scripts: those are for _migrating_ the DB schema and existing data. Migration scripts aren't meant for creating required initial data. If someone ever deployed a new instance of the LMS app to a new production environment (which is something we may very well find ourselves doing one day for various reasons) the app would be broken by the missing lms user (migration scripts don't get run on new deploys, they're not used to initialize the DB)
1. It's a separate conversation but based on my experience of using it I actually don't think superdev is quite ready for prime-time yet and I wouldn't push it to the team as a whole. I had to stop using it myself. I think a more incremental approach to bringing superdev in is needed. Separate discussion. But I don't think we should bring superdev into the scope of this issue
1. The solution in this PR has already been implemented, so there's definitely only time to be spent, and no time to be saved, by now implementing a different solution instead
1. I'm not doing this to try to avoid talking to people for the sake of avoiding it but I do think introducing the need to coordinate with other devs is always a downside to a solution, if it can be avoided. "Just works" is better

Basically: when upserting a group if the LMS app receives an error message from h that says "the lms user doesn't exist" then it executes two lines of code to create the lms user and then re-try upserting the group. I think that solves the problem at hand for all environments and cases.

seanh:
> If the authority changes from dev, QA to live is this going to break the bulk stuff which expects your authority to be exactly "lms.hypothes.is"?

Actually no. The LMS app's authority string comes from its `H_AUTHORITY` envvar so it can in theory be different in prod, QA and dev. But in fact `H_AUTHORITY` is set to `"lms.hypothes.is"` in all three.

I don't think it theoretically _has_ to be set the same in QA and prod and can't remember why it is, but there seems no reason to change it no

lms/services/lti_h.py
seanh: Ok:

* Renamed create_group() to upsert_group(): https://github.com/hypothesis/lms/pull/1614/commits/c589abb911a4c6e4143fec29eb6d854060b2738f
* Simplified the now-unnecessary update-then-upsert logic to just upsert: https://github.com/hypothesis/lms/pull/1614/commits/8ab3a4389940920f8f19610f129f4615e94cd215
* Re-tested and it still seems to work fine

seanh:
Rebased

PR #1618, opened 2020-04-03
Canvas client merge

jon-betts:
This is a fundamental rethink of the approach of the canvas API to make it smaller, simpler to understand and more localised (it's all in one file now). The total result is to make the code around 2-3 times smaller.

Major changes:

 * The separate 'helper' which previously prepared and executed requests has been absorbed into the client
 * The schemas (which are private to this object) have all been inlined
 * We now treat the public functions of the client as the unit of test (most of the tests don't mock or know about schemas inside etc.)

More minor changes:

 * Integration allowed the removal of a lot of duplication and false generality in methods
 * Simpler and re-used URL generation internally
 * Simpler calling logic and flow for HTTP requests
 * Made functions private which are not called outside of the API class
 * Cleaned up and merged doc strings where possible
 
Some style changes to follow new practice:

 * No `__all__`
 * Collaborators in-file
 * Remove Sphinx style docstrings

----

### What was it like before

Previously the logic for this client was spread out over:

 * `lms/services/canvas_api.py` - Public interface and 1/2 the code to send requests
 * `lms/services/_helpers/canvas_api.py` - Prepared requests and the other 1/2 of code to send them
 * `lms/validation/_canvas.py` - All private schema to the client API (used no-where else)
 * `lms/validation/authentication/_oauth.py` - A single private schema in here

The documentation was largely duplicated between the client and the client helper. Each function had a corresponding function in the other.

The tests mocked each layer heavily, making the flow hard to see.

To understand a single method:

 * Start in the main client file to read the args for the given method
 * Switch to the helper to see how these are made into a request
 * Switch back the the main client file for a call to "send_with_refresh_and_retry"
 * Switch back to the helper which implements "validated_response" which executes the query
 * Switch to the correct schema file to read what the expected return value looks like
 * Switch back to the client API to see how these return values are manipulated before returning to caller

### What it's like now

* `lms/services/canvas_api.py` - Contains all code

To understand a single method:

 * Read the method to see the arguments and how they are turned into a request
 * Read the schema immediately below the function to see what the expected return value looks like
 * Read the function to see if these values are modified
 * Read the relevant function to see how the request is made

----

# Reviewing notes

 * Most of the changes are just files or content being deleted
 * If the code went somewhere, it went into the client file
 * Each of the original refactors to get to the end-state are in separate commits to see each one
 * This is useful to follow to see this does what it did before, but the actual review of the code is likely better viewing the file as a new file
 * The tests were not changed until the refactoring was done - This results in a totally different set of tests which were basically rewritten from scratch. I would recommend reading the tests as a new file, instead of a diff
 * The tests contain a prototype h-matcher for requests, which will be shipped out in a later PR (artificially increasing their size by around 50 lines)


jon-betts:
Looks like permissions issues in Jenkins killing the build here. It can't delete the workspace

lms/services/canvas_api.py
```diff
@@ -1,30 +1,45 @@
...
     def __init__(self, _context, request):
-        self._helper = CanvasAPIHelper(
-            request.find_service(name="ai_getter"), request.route_url,
-        )
+        ai_getter = request.find_service(name="ai_getter")
+
+        self._client_id = ai_getter.developer_key()
+        self._client_secret = ai_getter.developer_secret()
+        self._canvas_url = urlparse(ai_getter.lms_url()).netloc
+        self._redirect_uri = request.route_url("canvas_oauth_callback")
```
jon-betts, line 41: This was previously the "helpers" responsibility, but we need to do it now

lms/services/canvas_api.py
```diff
@@ -33,202 +48,351 @@ def get_token(self, authorization_code):
         """
         Get an access token for the current LTI user.
 
-        Posts to the Canvas API to get the access token and returns it.
```
jon-betts, line 36: There were lots of comments saying we posted to the Canvas API, or made HTTP requests or soemthing similar. This seems obvious from the context

lms/services/canvas_api.py
```diff
@@ -33,202 +48,351 @@ def get_token(self, authorization_code):
...
-
-        :arg authorization_code: The Canvas API OAuth 2.0 authorization code to
+        :param authorization_code: The Canvas API OAuth 2.0 authorization code to
             exchange for an access token
-        :type authorization_code: str
```
jon-betts, line 40: We aren't mentioning parameter data types, because it doesn't do anything, and also this is Python

lms/services/canvas_api.py
```diff
@@ -33,202 +48,351 @@ def get_token(self, authorization_code):
...
+                params={
+                    "grant_type": "authorization_code",
+                    "client_id": self._client_id,
+                    "client_secret": self._client_secret,
+                    "redirect_uri": self._redirect_uri,
+                    "code": authorization_code,
+                    "replace_tokens": True,
+                },
+            ).prepare(),
+            self._CanvasTokenResponseSchema,
```
jon-betts, line 72: You can now see the request we are going to make. The token calls are a little weird as they can't use the nicer "_make_authenticated_request" interface which creates a request for you.

The schema `_CanvasTokenResponseSchema` mentioned here is directly below these functions so it can be read on the same page as this function.

lms/services/canvas_api.py
```diff
@@ -1,30 +1,45 @@
 import datetime
```
jon-betts, line 1: I very much recommend:

 * Reading through the commits if you are interested how we got here
 * Then reading the file (not in diff format) to see what it's like now

lms/services/canvas_api.py
```diff
@@ -33,202 +48,351 @@ def get_token(self, authorization_code):
...
+        # In practice ?include[]=sections seems to add a "sections" key to the
+        # API response that is a list of section dicts, one for each section
+        # the authenticated user is currently enrolled in, each with the
+        # section's "id" and "name" among other fields.
+        #
+        # **We don't know what happens if the user belongs to a really large
+        # number of sections**. Does the list of sections embedded within the
+        # get course API response just get really long? Does it get truncated?
+        # Can you paginate through it somehow? This seems edge-casey enough
+        # that we're ignoring it for now.
```
jon-betts, line 164: This is a belting great long comment that was previously in the helper. I wonder if a README or something would be better for some of thi.

lms/services/canvas_api.py
```diff
@@ -33,202 +48,351 @@ def get_token(self, authorization_code):
...
+            schema=self._AuthenticatedUsersSectionsSchema,
+        )
+
+    class _AuthenticatedUsersSectionsSchema(RequestsResponseSchema):
+        """Schema for the "authenticated user's sections" responses."""
+
+        sections = fields.List(
+            fields.Nested(_SectionSchema),
+            validate=validate.Length(min=1),
+            required=True,
```
jon-betts, line 179: A good demonstration here of the:

 * Here's how we make the call
 * Here's what we expect back

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -1,643 +1,482 @@
+import json
```
jon-betts, line 1: I'd really recommend not reading this as a diff:

 * The file has been rewritten again from scratch 
 * It covers all of the code where as previously it was only 1/4
 * It's still somehow 200 lines shorter

jon-betts:
I was a bit torn about the testing of the underlying bits. I was tempted to test it all from the outside, but I thought it might duplicate a lot of tests unnecessarily. There's a definite split between the auth token methods and everything else. If these were two sub-classes it would be a lot easier.

Hopefully anything like this is now easier to push through with it being smaller.

We've had an aspiration to match requests in general for a while. This would probably in the end need to match `requests.Request`, `requests.PreparedRequest` and `pyramid.Request` at least. I'll add some extra comments to it. When we've done this before we've added a "just enough" matcher where it's used, then extracted and really gone to down documenting and testing it in `h-matchers`.

PR #1634, opened 2020-04-08
Add some convenience variables

seanh:
Just make some code more concise by adding a few convenience variables
to `self`.

One test had to be fixed up because it was replacing `request.lti_user`
with a new object _after_ the request had already been passed to
`JSConfig`, but `JSConfig` now reads `request.lti_user` in its
`__init__()` so it still had the old `lti_user`. The fix makes the test
more correct anyway.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -16,6 +16,10 @@ class JSConfig:  # pylint:disable=too-few-public-methods
     def __init__(self, context, request):
         self._context = context
         self._request = request
+        self._lti_user = request.lti_user
+        self._consumer_key = (
+            request.lti_user.oauth_consumer_key if request.lti_user else None
+        )
 
         self._grading_info_service = request.find_service(name="grading_info")
         self._ai_getter = request.find_service(name="ai_getter")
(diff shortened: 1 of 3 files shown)


tests/unit/lms/resources/_js_config_test.py
```diff
@@ -222,11 +222,8 @@ def test_it_adds_the_grading_settings(self, js_config, grading_info_service):
             ],
         }
 
-    def test_it_does_nothing_if_the_user_isnt_an_instructor(
-        self, js_config, pyramid_request
-    ):
-        pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Learner")
-
+    @pytest.mark.usefixtures("learner_request")
```
jon-betts, line 225: This is better even if when we aren't forced to do it

lms/resources/_js_config.py
```diff
@@ -212,9 +214,7 @@ def _canvas_files_available(self):
             return False
 
         try:
-            developer_key = self._ai_getter.developer_key(
-                self._request.params.get("oauth_consumer_key")
```
jon-betts, line 216: This seems like the win here, not getting the key in two ways

PR #1636, opened 2020-04-08
Separate ai_getter service from service factory

seanh:
Separate `application_instance_getter_service_factory()` from the `ApplicationInstanceGetter` class itself.

This is because I'm expecting to have to register two different `ApplicationInstanceGetter` factories in two different contexts, in a future PR.

We have two different ways of doing pyramid_services in h and lms: in h the service class itself is "pure" and the factory is a separate function. In lms the service class's `__init__()` is used as the factory, so the service class itself has a dependency on `context` and `request`. I think h's way is better here and we should start migrating lms's services to it: it allows service classes to be decoupled from Pyramid, and it allows different service factories for the same service class to be registered in different contexts. Or you can have different classes that provide the same service, and a single service factory can contain if/else logic that instantiates different classes under different circumstances. Finally, it make constructing a service class directly easier, if you ever have reason to do that.

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -31,7 +31,7 @@ __all__ = (
 
 def includeme(config):
     config.register_service_factory(
-        "lms.services.application_instance_getter.ApplicationInstanceGetter",
+        "lms.services.application_instance_getter.application_instance_getter_service_factory",
         name="ai_getter",
     )
     config.register_service_factory(
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -10,9 +10,9 @@ __all__ = ["ApplicationInstanceGetter"]
 class ApplicationInstanceGetter:
     """Methods for getting properties from application instances."""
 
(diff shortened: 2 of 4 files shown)


lms/services/__init__.py
```diff
@@ -31,7 +31,7 @@
 
 def includeme(config):
     config.register_service_factory(
-        "lms.services.application_instance_getter.ApplicationInstanceGetter",
+        "lms.services.application_instance_getter.application_instance_getter_service_factory",
```
jon-betts, line 34: I know it's not current practice, but can you just put the actual function / a lambda here? Or does that break something to do with the design where you might not even import the code etc?

Imports vs. strings are nice, as the code won't even compile if you change the name on one side etc.

seanh: I'm not sure whether pyramid_services accepts callables directly, or whether it only works with strings.

I know that the authors prefer passing strings to this and other Pyramid registries because it's easier because the registering code doesn't even have to import the thing it's registering

PR #1639, opened 2020-04-10
Move values into models

seanh:
Instead of having a weird top-level `values.py` file hanging around,
let's just put these domain models in `lms/models/` with the other
models.

This means that `lms/models/` now contains both SQLAlchemy ORM classes
and non-DB model classes.


lms/authentication/_helpers.py
```diff
@@ -26,15 +26,15 @@ def authenticated_userid(lti_user):
 
 def get_lti_user(request):
     """
-    Return an :class:`~lms.values.LTIUser` for the authenticated LTI user.
+    Return an models.LTIUser for the authenticated LTI user.
```
seanh, line 29: I'm removing these Sphinx `:class:` references and other similar things. They aren't checked or used by anything (e.g. we don't use any tooling that parses them) and it's hard to get people to write them consistently. The syntax is hard to remember.

jon-betts: Yeah, it's a pain and we don't make any use of it.

seanh:
One thing about the two non-ORM models that we have so far is that they're immutable because they use [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple). That might be a good idea as it sort of enforces simplicity.

I think both ORM and non-ORM models tend to float around the whole codebase and also tend to not get mocked in tests, which means we should probably tend to keep them quite simple and move logic into services instead. Most of what was said in these [thin models slides](https://docs.google.com/presentation/d/1O-rl1qr0diNNAwACLAe7StLeZEi4kuIvhfwrX83AMKM/edit#slide=id.g4c957d27dd_0_566) (and the services section that follows it) back in Jan 2019 is probably still true. The later [boundaries section](https://docs.google.com/presentation/d/1O-rl1qr0diNNAwACLAe7StLeZEi4kuIvhfwrX83AMKM/edit#slide=id.g4b03785266_0_72) of the same presentation (and the _value objects_ section that follows it) is also relevant.

For specific guidelines I think maybe simple, single-model SQLAlchemy queries (`get_by_*()`), simple computed properties (like the ones `LTIUser` and `HUser` already have), simple factory methods, are examples of the kinds of things that are probably okay in models. [<q>A method that finds in a model all the data that it needs... really wants to belong to that model.</q>](http://dev.nando.audio/2014/04/01/large_apps_with_sqlalchemy__architecture.html) But we should avoid any model class or method getting very big or complicated at all.

PR #1640, opened 2020-04-10
Get LMS unit tests to 100% percent code and branch coverage

seanh:
I keep accidentally reducing coverage by deleting lines, causing the coverage test to fail, which wastes developer time. Also if I really have added lines with no coverage, I have to look for my lines in amongst the not-covered-on-master lines in the coverage report.

Fix both issues by getting coverage to 100%.

In some cases to increase coverage I fixed broken tests or added missing tests, or fixed bad code. See individual commit messages for details.

In the final commit I covered a handful of things that aren't worth testing or would be a big pain to test with `# pragma: no cover` comments. The 100% coverage is a practical benefit in itself, for the reasons above.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,5 +1,6 @@
 [![Build Status](https://travis-ci.org/hypothesis/lms.svg?branch=master)](https://travis-ci.org/hypothesis/lms)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
+[![Coverage: 100%](https://img.shields.io/badge/Coverage-100%25-brightgreen)](https://github.com/hypothesis/lms/blob/master/.coveragerc#L18)
 
 # Hypothesis LMS App
 
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -103,7 +103,7 @@ def make_engine(settings):
     return sqlalchemy.create_engine(settings["sqlalchemy.url"])
 
 
-def _session(request):
+def _session(request):  # pragma: no cover
     engine = request.registry["sqlalchemy.engine"]
     session = SESSION(bind=engine)
 
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -46,14 +46,13 @@ class BaseClass:
         if skip_keys is None:
             skip_keys = {"id"}
 
-        columns = {col.key for col in self.iter_columns()}
-        if skip_keys:
-            if not isinstance(skip_keys, set):
-                raise TypeError(
-                    f"Expected a set of keys to skip but found '{type(skip_keys)}'"
-                )
+        if not isinstance(skip_keys, set):
+            raise TypeError(
+                f"Expected a set of keys to skip but found '{type(skip_keys)}'"
+            )
 
-            columns -= skip_keys
+        columns = {col.key for col in self.iter_columns()}
+        columns -= skip_keys
 
         for key in columns:
             if key in data:
--- a/lms/views/api/lti.py
+++ b/lms/views/api/lti.py
@@ -114,8 +114,8 @@ class CanvasPreRecordHook:
 
         if parsed_params.get("document_url"):
             params["url"] = parsed_params.get("document_url")
-
-        elif parsed_params.get("canvas_file_id"):
+        else:
+            assert parsed_params.get("canvas_file_id")
             params["canvas_file"] = "true"
             params["file_id"] = parsed_params["canvas_file_id"]
 
--- a/lms/views/api/lti.py
+++ b/lms/views/api/lti.py
@@ -115,7 +115,10 @@ class CanvasPreRecordHook:
         if parsed_params.get("document_url"):
             params["url"] = parsed_params.get("document_url")
         else:
-            assert parsed_params.get("canvas_file_id")
+            assert parsed_params.get("canvas_file_id"), (
+                "All Canvas launches should have either a 'document_url' or "
+                "a 'canvas_file_id' parameter."
+            )
             params["canvas_file"] = "true"
             params["file_id"] = parsed_params["canvas_file_id"]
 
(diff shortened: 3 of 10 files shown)


lms/views/api/lti.py
```diff
@@ -114,8 +114,8 @@ def get_speedgrader_launch_url(self):
 
         if parsed_params.get("document_url"):
             params["url"] = parsed_params.get("document_url")
-
-        elif parsed_params.get("canvas_file_id"):
+        else:
+            assert parsed_params.get("canvas_file_id")
```
jon-betts, line 118: I seem to remember having a conversation about using asserts in production code, and how it's not a thing we do.

I'm totally fine with it for sanity style this code should never get here style things. It seems to me that this could quite easily get in this state though if we get a `document_url` and no `canvas_file_id`.

What's the error handling for `AssertionError`? Will it result in a 400? Previously this wouldn't have failed.

tests/unit/lms/services/grading_info_test.py
```diff
@@ -100,17 +100,26 @@ def test_it_creates_new_record_if_no_matching_exists(
...
+            lis_result_sourcedid="lis_result_sourcedid",
+            lis_outcome_service_url="lis_outcomes_service_url",
+            oauth_consumer_key=lti_user.oauth_consumer_key,
+            user_id=lti_user.user_id,
+            context_id=pyramid_request.params["context_id"],
+            resource_link_id=pyramid_request.params["resource_link_id"],
+            tool_consumer_info_product_family_code="tool_consumer_info_product_family_code",
+            h_username="h_username",
+            h_display_name="h_display_name",
+        )
```
jon-betts, line 113: This test is hard to grok as it's not clear which arguments are relevant. Using a fixture or method to create this would give an opportunity to give it an helpful name. If you used a method, it's also an opportunity for that method to only accept the parts which are relevant for this test.

My guess is it's only the user name and display name?

lms/db/__init__.py
```diff
@@ -46,14 +46,13 @@ def update_from_dict(self, data, skip_keys=None):
...
-        columns = {col.key for col in self.iter_columns()}
-        if skip_keys:
-            if not isinstance(skip_keys, set):
-                raise TypeError(
-                    f"Expected a set of keys to skip but found '{type(skip_keys)}'"
-                )
+        if not isinstance(skip_keys, set):
+            raise TypeError(
+                f"Expected a set of keys to skip but found '{type(skip_keys)}'"
+            )
```
jon-betts, line 52: It was possible before to squeeze through this condition with `skip_keys` being an empty dict. It seems better to me to just fail.

So this eliminates a code path, and probably a bug too.

README.markdown
```diff
@@ -1,5 +1,6 @@
 [![Build Status](https://travis-ci.org/hypothesis/lms.svg?branch=master)](https://travis-ci.org/hypothesis/lms)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
+[![Coverage: 100%](https://img.shields.io/badge/Coverage-100%25-brightgreen)](https://github.com/hypothesis/lms/blob/master/.coveragerc#L18)
```
jon-betts, line 3: I like that this doesn't depend on this actually being true.

lms/views/api/lti.py
seanh: No, I think `assert`s are fine in production code as internal consistency checks and documentation, as long as they're used for unrecoverable errors / internal self-checks about conditions that should be impossible (and not e.g. for validation). Python's `-O` and `-OO` command line arguments disable asserts, but we don't use those. We didn't have a conversation in which I said we don't do them. This is a good post on it: https://dbader.org/blog/python-assert-tutorial

An uncaught `AssertionError` will be caught by the `Exception` exception view, so it'll be a 500 Internal Server Error, the user will get the HTML or JSON error response, and it'll be both logged to Papertrail and reported to Sentry.

Previously this code wouldn't have crashed but it would have generated an invalid SpeedGrader launch URL (one with neither a `url` or a `file_id` query param) and sent it to SpeedGrader, and SpeedGrader would then launch us with that invalid URL, the SpeedGrader launch-handling code isn't expecting such invalid launches so it would either misbehave or crash. Crashing early seems a lot less bad.

I'll definitely test SpeedGrader before deploying this though

tests/unit/lms/services/grading_info_test.py
seanh: I think a `GradingInfo` test factory would simplify and clean up these tests quite a bit, but we'd need the [test factories PR](https://github.com/hypothesis/lms/pull/1642) merged first

seanh: Done (adding a `GradingInfo` factory)

seanh: Ending up moving the `GradingInfo` factory into a separate PR, because it turns out to be a slightly larger thing: https://github.com/hypothesis/lms/pull/1650

To get this PR merged I just added a quickie helper method to remove noise from this one test: https://github.com/hypothesis/lms/pull/1640/commits/11409dc7335c6e2edb72d59cf02fda8cd369b426

Note there are already other `GradingInfo` factory methods with different features elsewhere in this same test module! Trying to merge them into a single helper method with all the features would be too much work and comes too close to reimplementing the Factory Boy-based `GradingInfo` factory in #1650 with my own code. I'll just get #1650 finished up and merged soon, instead

seanh:
I think this might need a rebase because https://github.com/hypothesis/lms/pull/1637 was merged

tests/unit/lms/services/grading_info_test.py
jon-betts: Yeah no worries. Looks like I guessed the wrong fields incidentally, so there we go.

lms/views/api/lti.py
jon-betts: This is coming from Canvas anyway, so it's not like a user is typing this stuff in. If the previous state was a failure state, that doesn't worry me so much. I just wanted to check we weren't introducing one, as this code didn't previously make it obvious this wasn't an acceptable path.

The message is good. We can look for it in sentry, and fix it if it ever happens.

PR #1642, opened 2020-04-12
Add LTIUser test factory

seanh:
In a future PR I need to add a new required argument to `LTIUser()`, which will mean updating everywhere that `LTIUser()` is called. `LTIUser()` is only called in a few places in the code, but it's called in many places in the tests, and those would all need to be updated.

Instead, add an `LTIUser` test factory so that `LTIUser()`'s arguments can be changed in future without having to change a lot of test code. Instead of calling `values.LTIUser(...)` and having to pass every required param every time, tests now call `factories.LTIUser()` with no arguments and the factory generates values for all of `values.LTIUser()`'s parameters. Or, if the test needs to pass a specific value for one of `LTIUser`'s params, it does `factories.LTIUser(some_param="some_value")`, passing only the arguments it cares about, and the factory generates the rest.

This means that required arguments can be added to `LTIUser()` and only the test factory needs to be updated. Individual tests won't need to change.

The factory also brings some other improvements to the test code:

* Fixes "too much information" problems in tests. A lot of tests are doing `LTIUser("TEST_USER_ID", "TEST_OAUTH_CONSUMER_KEY", "TEST_ROLES")` when the test doesn't actually care about any of those arguments. This is noise, and makes the tests less clear. When the tests actually do care about the value of one of the arguments, it obscures it.

  Tests can now do `factories.LTIUser()` or, if one of the arguments is actually relevant to the test, then include just that argument: `factories.LTIUser(oauth_consumer_key="...")`.

* Fixes "too little information" problems in tests. Often tests are using a distant `LTIUser` fixture (such as `pyramid_request.lti_user` in the global `pyramid_request` fixture) and are either explicitly reliant on that `LTIUser`s fields (e.g. they'll contain things like `some_service.some_method.assert_called_once_with("TEST_OAUTH_CONSUMER_KEY")` or, worse, the tests is implicitly reliant on the `LTIUser` having a particular attribute value like `"TEST_OAUTH_CONSUMER_KEY"` but that value isn't mentioned anywhere in the test at all. In both cases, the tests are depending on an `LTIUser` having been created with certain arguments, but creating the `LTIUser` with those arguments doesn't happen in the test. This again makes the tests less clear.

  Because `factories.LTIUser()` randomly generates new arguments to pass to `LTIUser` every time it's called, it's no longer possible to write tests that invisibly rely on fixed arguments having been passed to `LTIUser`. Tests either have to call `factories.LTIUser(...)` and pass specific values for arguments that the test depends on, or else the test needs to access things like `pyramid_request.lti_user.oauth_consumer_key` instead of just hard-coding `"TEST_OAUTH_CONSUMER_KEY"`.

* Establishes a pattern. Test factories are a best practice for the reasons above and others (e.g. they tend to result in more realistic values being used in tests, which can find bugs that wouldn't be found otherwise). Adding a `tests.factories` package establishes this best practice as a pattern that more tests can follow in future. Since the test factories are always used as `factories.Foo()` you can always see when a test factory is used, and since test factories all behave in broadly the same way you already know what's going on and probably don't need to look up the factory code.

* The test factory is implemented correctly once and reusable throughout all of the tests, which saves test fixture duplication and avoids buggy or less-than-ideal test fixture behaviour when different tests all implement their own ways of creating test objects.

  Correctly means things like generating realistic-looking values, generating randomised values so that tests can't invisibly depend on hard-coded values, using predictable randomness so flakey tests aren't produced, generating unique values when uniqueness is required (doesn't apply to `LTIUser`), providing a consistent way for individual tests to override just the values they care about, etc.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@ import functools
 import os
 from unittest import mock
 
+import factory.random
 import pytest
 import sqlalchemy
 from sqlalchemy.orm import sessionmaker
@@ -71,3 +72,11 @@ def autopatcher(request, target, **kwargs):
 @pytest.fixture
 def patch(request):
     return functools.partial(autopatcher, request)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def factory_boy_random_seed():
+    # Set factory_boy's random seed so that it produces the same random values
+    # in each run of the tests.
+    # See: https://factoryboy.readthedocs.io/en/latest/index.html#reproducible-random-values
+    factory.random.reseed_random("hypothesis/lms")
--- a/tests/factories/__init__.py
+++ b/tests/factories/__init__.py
@@ -1,0 +1,1 @@
+from tests.factories.lti_user import LTIUser
--- a/tests/factories/lti_user.py
+++ b/tests/factories/lti_user.py
@@ -1,0 +1,10 @@
+from factory import Faker, make_factory
+
+from lms import models
+
+LTIUser = make_factory(  # pylint:disable=invalid-name
+    models.LTIUser,
+    user_id=Faker("hexify", text="^" * 40),
+    oauth_consumer_key=Faker("hexify", text="Hypothesis" + "^" * 32),
+    roles=Faker("random_element", elements=["Learner", "Instructor"]),
+)
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -7,7 +7,7 @@ import sqlalchemy
 from pyramid import testing
 from pyramid.request import apply_request_extensions
 
-from lms.models import HUser, LTIUser
+from lms.models import HUser
 from lms.services.application_instance_getter import ApplicationInstanceGetter
 from lms.services.canvas_api import CanvasAPIClient
 from lms.services.grading_info import GradingInfoService
@@ -17,6 +17,7 @@ from lms.services.launch_verifier import LaunchVerifier
 from lms.services.lti_h import LTIHService
 from lms.services.lti_outcomes import LTIOutcomesClient
 from lms.services.oauth1 import OAuth1Service
+from tests import factories
 from tests.conftest import SESSION, TEST_SETTINGS, get_test_database_url
 
 TEST_SETTINGS["sqlalchemy.url"] = get_test_database_url(
@@ -54,15 +55,18 @@ def pyramid_request(db_session):
     pyramid_request.feature = mock.create_autospec(
         lambda feature: False, return_value=False  # pragma: no cover
     )
-    pyramid_request.lti_user = LTIUser(
-        "TEST_USER_ID", "TEST_OAUTH_CONSUMER_KEY", "TEST_ROLES"
-    )
+    pyramid_request.lti_user = factories.LTIUser()
 
     # The DummyRequest request lacks a content_type property which the real
     # request has
     pyramid_request.content_type = None
 
     return pyramid_request
+
+
+@pytest.fixture
+def user_is_learner(pyramid_request):
+    pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Learner")
 
 
 def configure_jinja2_assets(config):
--- a/tests/unit/lms/authentication/_helpers_test.py
+++ b/tests/unit/lms/authentication/_helpers_test.py
@@ -1,8 +1,8 @@
 import pytest
 
 from lms.authentication._helpers import authenticated_userid, get_lti_user, groupfinder
-from lms.models import LTIUser
 from lms.validation import ValidationError
+from tests import factories
 
 
 class TestAuthenticatedUserID:
@@ -10,16 +10,16 @@ class TestAuthenticatedUserID:
         "lti_user,expected_userid",
         [
             (
-                LTIUser(
-                    "sam", "Hypothesisf301584250a2dece14f021ab8424018a", "TEST_ROLES"
+                factories.LTIUser(
+                    user_id="sam",
+                    oauth_consumer_key="Hypothesisf301584250a2dece14f021ab8424018a",
                 ),
                 "c2Ft:Hypothesisf301584250a2dece14f021ab8424018a",
             ),
             (
-                LTIUser(
-                    "Sam:Smith",
-                    "Hypothesisf301584250a2dece14f021ab8424018a",
-                    "TEST_ROLES",
+                factories.LTIUser(
+                    user_id="Sam:Smith",
+                    oauth_consumer_key="Hypothesisf301584250a2dece14f021ab8424018a",
                 ),
                 "U2FtOlNtaXRo:Hypothesisf301584250a2dece14f021ab8424018a",
             ),
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -198,12 +198,14 @@ class TestAddCanvasFileIDAddDocumentURLCommon:
 
 
 class TestMaybeEnableGrading:
-    def test_it_adds_the_grading_settings(self, js_config, grading_info_service):
+    def test_it_adds_the_grading_settings(
+        self, js_config, grading_info_service, pyramid_request
+    ):
         js_config.maybe_enable_grading()
 
         grading_info_service.get_by_assignment.assert_called_once_with(
             context_id="test_course_id",
-            oauth_consumer_key="TEST_OAUTH_CONSUMER_KEY",
+            oauth_consumer_key=pyramid_request.lti_user.oauth_consumer_key,
             resource_link_id="TEST_RESOURCE_LINK_ID",
         )
         assert js_config.asdict()["grading"] == {
@@ -221,7 +223,7 @@ class TestMaybeEnableGrading:
             ],
         }
 
-    @pytest.mark.usefixtures("learner_request")
+    @pytest.mark.usefixtures("user_is_learner")
     def test_it_does_nothing_if_the_user_isnt_an_instructor(self, js_config):
         js_config.maybe_enable_grading()
 
@@ -481,12 +483,6 @@ def pyramid_request(pyramid_request):
     pyramid_request.params[
         "lis_outcome_service_url"
     ] = "example_lis_outcome_service_url"
-    return pyramid_request
-
-
-@pytest.fixture
-def learner_request(pyramid_request):
-    pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Learner")
     return pyramid_request
 
 
--- a/tests/unit/lms/services/grading_info_test.py
+++ b/tests/unit/lms/services/grading_info_test.py
@@ -3,9 +3,10 @@ from unittest import mock
 import pytest
 from h_matchers import Any
 
-from lms.models import GradingInfo, HUser, LTIUser
+from lms.models import GradingInfo, HUser
 from lms.resources import LTILaunchResource
 from lms.services.grading_info import GradingInfoService
+from tests import factories
 
 
 class TestGetByAssignment:
@@ -158,7 +159,7 @@ class TestUpsertFromRequest:
 
 @pytest.fixture
 def lti_user():
-    return LTIUser("test_user_id", "matching_oauth_consumer_key", "test_roles")
+    return factories.LTIUser(oauth_consumer_key="matching_oauth_consumer_key")
 
 
 @pytest.fixture
--- a/tests/unit/lms/models/lti_user_test.py
+++ b/tests/unit/lms/models/lti_user_test.py
@@ -1,8 +1,6 @@
-from unittest import mock
-
 import pytest
 
-from lms.models import LTIUser
+from tests import factories
 
 
 class TestLTIUser:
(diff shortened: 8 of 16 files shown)


tests/conftest.py
```diff
@@ -71,3 +72,11 @@ def autopatcher(request, target, **kwargs):
...
 def patch(request):
     return functools.partial(autopatcher, request)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def factory_boy_random_seed():
+    # Set factory_boy's random seed so that it produces the same random values
+    # in each run of the tests.
+    # See: https://factoryboy.readthedocs.io/en/latest/index.html#reproducible-random-values
+    factory.random.reseed_random("hypothesis/lms")
```
seanh, line 82: This is to make sure that if the tests pass when I run them, the tests will pass again if I run them again (without changing the code). Or if they pass for me, they'll pass for you and on CI too. We don't want flakey tests, so we don't want actual randomness in tests. The "predictable randomness" is enough to discourage tests from invisibly depending on fixed values having been passed by factories to models, while keeping the tests deterministic.

tests/factories/__init__.py
```diff
@@ -0,0 +1 @@
+from tests.factories.lti_user import LTIUser
```
seanh, line 1: It turns out that LMS already had `factory-boy` as a dependency but wasn't using it. I've added a `tests.factories` package to provide a place to put shared test factories and establish a pattern.

There's no `factories` pytest fixture like h has. LMS's one factory so far doesn't use the DB so it isn't necessary, and just importing `tests.factories` directly allows them to be used in parametrize.

We might add an h-like `factories` fixture when LMS needs factories for DB models, or figure out a way to do that without requiring the factories to be accessed via a `factories` fixture.

tests/factories/lti_user.py
```diff
@@ -0,0 +1,10 @@
+import factory
+
+from lms import values
+
+LTIUser = factory.make_factory(  # pylint:disable=invalid-name
+    values.LTIUser,
+    user_id=factory.Faker("hexify", text="^" * 40),
+    oauth_consumer_key=factory.Faker("hexify", text="Hypothesis" + "^" * 32),
+    roles=factory.Faker("random_element", elements=["Learner", "Instructor"]),
+)
```
seanh, line 10: For this simple factory I'm using Factory Boy's [lightweight factory declaration](https://factoryboy.readthedocs.io/en/latest/reference.html#lightweight-factory-declaration), which unfortunately requires a `# pylint:disable=invalid-name`.

The longer-form way, which you need to use if your factory requires advanced Factory Boy features, would be:

```python
class LTIUser(factory.Factory):
    class Meta:
        model = values.LTIUser

    user_id = factory.Faker("hexify", text="^" * 40)
    oauth_consumer_key = factory.Faker("hexify", text="Hypothesis" + "^" * 32)
    roles = factory.Faker("random_element", elements=["Learner", "Instructor"])
```

This also has an unfortunate PyLint shortcoming: PyLint understandably thinks that `factories.LTIUser()` returns an object of class `factories.LTIUser`. When you then try to access `values.LTIUser` fields on that object (e.g. `lti_user.is_learner`) PyLint complains that the field doesn't exist.

seanh: The Factory Boy docs tend to name the factory classes with a `...Factory` suffix like `UserFactory` etc. You'd then do:

```python
from tests.factories import UserFactory

test_user = UserFactory(...)
```

I've followed h's approach of just calling the factory `User`:

```python
from tests import factories

test_user = factories.User(...)
```

This works well when you **have** to access factories via a `factories` fixture as you do for DB factories in h and may in future need to for future DB factories in LMS. You can just write `factories.User()` not `factories.UserFactory()`

seanh: The `Faker` stuff deserves some explanation:

* Factory Boy integrates a library called [Faker](https://faker.readthedocs.io/) for randomly generating realistic looking fake values.

* Faker comes with a set of "providers": functions for returning different types of fake values. See [Faker's built-in providers](https://faker.readthedocs.io/en/latest/providers.html) and [Faker's community providers](https://faker.readthedocs.io/en/latest/communityproviders.html)

* The way you call Faker via Factory Boy is `factory.Faker("provider", **kwargs)` where `"provider"` is the name of the Faker provider method as a string and `**kwargs` are the kwargs for Factory Boy to pass to the provider method (see the docs for each individual provider method for what kwargs it supports).

  So `factory.Faker("hexify", text="Hypothesis" + "^" * 32)` calls Faker's [`hexify()`](https://faker.readthedocs.io/en/latest/providers/baseprovider.html#faker.providers.BaseProvider.hexify) with `text="Hypothesis" + "^" * 32`.

* `hexify()` is one of a few flexible Faker methods for generating random strings. It replaces every `"^"` in the given string with a random hex character. So passing it `text="Hypothesis" + "^" * 32` causes it go generate strings like `"Hypothesis<32_RANDOM_HEX_CHARS>"`, which look like the real consumer keys that the LMS app generates.

  There's also `lexify()` which replaces `"?"`'s with random letters, `numerify()` which replaces `"#"`'s, `"%"`'s, `"!"`'s and `"@"`'s with different kinds of digits, and `bothify()` which combines `lexify()` and `numerify()`.

* Factory Boy passes its fixed random seed to Faker and everything else that Factory Boy uses that requires a randomness source, so they all use the same predictable randomness.

* You can also add your own providers to Faker. See Faker's [docs on writing your own provider](https://faker.readthedocs.io/en/latest/index.html#how-to-create-a-provider) and Factory Boy's [docs on adding a provider to Factory Boy's Faker instance](https://factoryboy.readthedocs.io/en/latest/reference.html#factory.Faker.add_provider).

  It seems unlikely you'll need to add a provider though. Faker comes with a pretty good set. And if there's an existing provider that's close to what you want but not quite right you can use various Factory Boy tricks to adjust what Faker returns, the easiest of which is probably Factory Boy's [`_adjust_kwargs()` hook](https://factoryboy.readthedocs.io/en/latest/reference.html#factory.Factory._adjust_kwargs)

See Factory Boy's [docs about its Faker integration](https://factoryboy.readthedocs.io/en/latest/reference.html#faker)

tests/unit/conftest.py
```diff
@@ -54,9 +55,7 @@ def pyramid_request(db_session):
     pyramid_request.feature = mock.create_autospec(
         lambda feature: False, return_value=False  # pragma: no cover
     )
-    pyramid_request.lti_user = LTIUser(
-        "TEST_USER_ID", "TEST_OAUTH_CONSUMER_KEY", "TEST_ROLES"
-    )
+    pyramid_request.lti_user = factories.LTIUser()
```
seanh, line 58: `pyramid_request.lti_user` is now a different randomly-generated `LTIUser` each time, so tests can no longer invisibly depend on `"TEST_USER_ID"`, "TEST_OAUTH_CONSUMER_KEY"` and `"TEST_ROLES"` having been passed to it

tests/factories/lti_user.py
seanh: The factory of course generates realistic-looking random `LTIUser`'s. They look more realistic than what all the tests are using currently: they have a realistic, randomly generated `user_id` and `oauth_consumer_key` etc:

```terminal
$ .tox/tests/bin/python
Python 3.6.9 (default, Jan  7 2020, 12:11:53) 
[GCC 9.2.1 20191008] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from tests import factories
>>> factories.LTIUser()
LTIUser(user_id='52d26561232aa962634d245a1fec2804f00a3804', oauth_consumer_key='Hypothesis1371da4380dc333f2075fcbcfc7680bb', roles='Instructor')
>>> factories.LTIUser()
LTIUser(user_id='2b402c900d752a4867223f0e8651e5d3a7784e98', oauth_consumer_key='Hypothesis5321d75bd9f8ee356b890140262b6ea5', roles='Learner')
>>> factories.LTIUser()
LTIUser(user_id='8fad39e652ded224dcf5d70765ca71988f4a98b5', oauth_consumer_key='Hypothesis53dc1cb8beb0c86a1ea05682e377e26f', roles='Instructor')
>>> 
```

tests/unit/lms/authentication/_helpers_test.py
```diff
@@ -2,24 +2,24 @@
...
 class TestAuthenticatedUserID:
     @pytest.mark.parametrize(
         "lti_user,expected_userid",
         [
             (
-                LTIUser(
-                    "sam", "Hypothesisf301584250a2dece14f021ab8424018a", "TEST_ROLES"
+                factories.LTIUser(
+                    user_id="sam",
+                    oauth_consumer_key="Hypothesisf301584250a2dece14f021ab8424018a",
```
seanh, line 15: This is a good example of how the factory forces tests to include relevant information but not include noise. These tests do depend on the particular `user_id` and `oauth_consumer_key` values but don't care about `roles`

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -199,12 +199,14 @@ def method(self, js_config, request):
...
-    def test_it_adds_the_grading_settings(self, js_config, grading_info_service):
+    def test_it_adds_the_grading_settings(
+        self, js_config, grading_info_service, pyramid_request
+    ):
         js_config.maybe_enable_grading()
 
         grading_info_service.get_by_assignment.assert_called_once_with(
             context_id="test_course_id",
-            oauth_consumer_key="TEST_OAUTH_CONSUMER_KEY",
+            oauth_consumer_key=pyramid_request.lti_user.oauth_consumer_key,
```
seanh, line 209: Tests can no longer rely on `"TEST_OAUTH_CONSUMER_KEY"` having been hard-coded when `pyramid_request.lti_user` was created. In most cases accessing `pyramid_request.lti_user.oauth_consumer_key` instead of the literal value is the best solution.

I am starting to want to rename `pyramid_request` to `req` though

tests/unit/lms/services/grading_info_test.py
```diff
@@ -159,7 +160,7 @@ def pyramid_request(self, pyramid_request, lti_params):
 
 @pytest.fixture
 def lti_user():
-    return LTIUser("test_user_id", "matching_oauth_consumer_key", "test_roles")
+    return factories.LTIUser(oauth_consumer_key="matching_oauth_consumer_key")
```
seanh, line 163: These tests don't care what the user_id and roles are, noise removed

tests/factories/lti_user.py
```diff
@@ -0,0 +1,10 @@
+import factory
```
jon-betts, line 1: I think it's ok to import the constituent parts of factory boy that we use. It's all we're going to do in this file, so it's Ok to be _all about that_.  Should be a little easier on the eye.

tests/factories/lti_user.py
jon-betts: `bothify()` is a spectacular name.

tests/unit/conftest.py
```diff
@@ -65,6 +64,12 @@ def pyramid_request(db_session):
     return pyramid_request
 
 
+@pytest.fixture
+def learner_request(pyramid_request):
+    pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Learner")
+    return pyramid_request
```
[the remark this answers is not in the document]
seanh: Changed to `user_is_learner`.

Returning nothing is a fairly strong hint.

I'm not sure about a consistent naming pattern for this kind of fixture, can't think of anything

tests/unit/lms/services/application_instance_getter_test.py
```diff
@@ -11,15 +11,20 @@
 
 
 class TestApplicationInstanceGetter:
-    def test_developer_key_returns_the_developer_key(self, ai_getter):
-        assert ai_getter.developer_key("TEST_CONSUMER_KEY") == "TEST_DEVELOPER_KEY"
+    def test_developer_key_returns_the_developer_key(self, ai_getter, pyramid_request):
+        assert (
+            ai_getter.developer_key(pyramid_request.lti_user.oauth_consumer_key)
+            == "TEST_DEVELOPER_KEY"
+        )
```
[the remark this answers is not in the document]
seanh: I think this is because there's [another PR awaiting re-review](https://github.com/hypothesis/lms/pull/1637) that changes `ApplicationInstanceGetter` so that it _is_ specific to the Pyramid request LTI user consumer key (at which point the consumer key arg to this method goes away). So if I changed this test now, I'd immediately have to change it back again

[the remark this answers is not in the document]
seanh: This really is as easy as just reverting the entire file though, so I've gone ahead and done it anyway

tests/factories/lti_user.py
seanh: Done

seanh:
Rebased

tests/unit/lms/services/application_instance_getter_test.py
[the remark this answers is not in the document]
jon-betts: Aw right, gotcha. The PR train will destroy you!

tests/factories/__init__.py
jon-betts: Unless we need to I think it's better without, it's kind unnecessarily roundabout if you want to find where it came from.

PR #1650, opened 2020-04-15
Add GradingInfo factory

seanh:
Add a test factory for `models.GradingInfo`. Since this is the first test factory for a DB model to be added to LMS, it requires integrating Factory Boy with the DB session.

When you do `factories.GradingInfo()` (or `factories.GradingInfo.create_batch(n)`) it automatically adds them to the DB session for you.

I haven't tested but I don't think this would work in an `@parametrize` because, even though the factory classes are just imported directly not accessed via a `factories` fixture like in h, the DB session won't have been set up at the time the `@parametrize` gets evaluated so it'll crash.

_But_ you can do `factories.GradingInfo.build()` or `build_batch(n)` which just creates `GradingInfo`'s without adding them to the DB. This means that your test can use the factory without using the DB if it doesn't actually need the DB (which would be a speed advantage if we weren't always initializing the DB for every test anyway) and it will also probably work in an `@parametrize` (though I haven't tried it).

Also, you could use `factories.GradingInfo.build()` in an `@parametrize` and then, if you also needed the object to be in the DB, add it manually with `db_session.add()` in your test.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -4,6 +4,7 @@ from unittest import mock
 import httpretty
 import pytest
 import sqlalchemy
+from factory.alchemy import SQLAlchemyModelFactory
 from pyramid import testing
 from pyramid.request import apply_request_extensions
 
@@ -135,12 +136,53 @@ def db_session(db_engine):
         ):  # pylint:disable=protected-access
             session.begin_nested()
 
+    set_factory_boy_sqlalchemy_session(session)
+
     try:
         yield session
     finally:
+        clear_factory_boy_sqlalchemy_session()
         session.close()
         trans.rollback()
         conn.close()
+
+
+def sqlalchemy_factory_classes():
+    # Return all the SQLAlchemy factory classes from tests.factories.
+    for factory_class in factories.__dict__.values():
+        try:
+            is_sqla_factory = issubclass(factory_class, SQLAlchemyModelFactory)
+        except TypeError:
+            is_sqla_factory = False
+
+        if is_sqla_factory:
+            yield factory_class
+
+
+def set_factory_boy_sqlalchemy_session(session):
+    # Set the Meta.sqlalchemy_session option on all our SQLAlchemy test factory
+    # classes. We can't do it in the normal Factory Boy way:
+    #
+    #     class MyFactory:
+    #         class Meta:
+    #             sqlalchemy_session = session
+    #
+    # Because we don't have `session` available to us at import time.
+    # So we have to do it this way instead.
+    #
+    # See:
+    # https://factoryboy.readthedocs.io/en/latest/orms.html#sqlalchemy
+    # https://factoryboy.readthedocs.io/en/latest/reference.html#factory.Factory._meta
+    for factory_class in sqlalchemy_factory_classes():
+        # pylint:disable=protected-access
+        factory_class._meta.sqlalchemy_session = session
+
+
+def clear_factory_boy_sqlalchemy_session():
+    # Delete the sqlalchemy session from all our test factories.
+    # Just in case, so we don't have references to the session hanging about.
+    for factory_class in sqlalchemy_factory_classes():
+        factory_class._meta.sqlalchemy_session = None  # pylint:disable=protected-access
 
 
 @pytest.fixture
--- a/tests/unit/lms/services/grading_info_test.py
+++ b/tests/unit/lms/services/grading_info_test.py
@@ -1,3 +1,4 @@
+from operator import attrgetter
 from unittest import mock
 
 import pytest
@@ -17,7 +18,9 @@ class TestGetByAssignment:
             "matching_resource_link_id",
         )
 
-        assert list(grading_infos) == matching_grading_infos
+        assert sorted(grading_infos, key=attrgetter("id")) == sorted(
+            matching_grading_infos, key=attrgetter("id")
+        )
 
     @pytest.mark.parametrize(
         "oauth_consumer_key,context_id,resource_link_id",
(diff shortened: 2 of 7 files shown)


tests/unit/conftest.py
```diff
@@ -135,14 +136,40 @@ def restart_savepoint(session, transaction):  # pylint:disable=unused-variable
         ):  # pylint:disable=protected-access
             session.begin_nested()
 
+    set_factory_boy_sqlalchemy_session(session)
```
seanh, line 139: So if you're gonna use `factories.GradingInfo()` your test needs to use the `db_session` fixture or it'll crash. Which can be in an `@pytest.mark.usefixtures("db_session")` or in a module-level `pytestmark = pytest.mark.usefixtures("db_session")`.

Or you can use `factories.GradingInfo.build()` or `build_batch()` if you just need `GradingInfo`'s but don't actually need them in the DB.

In fact LMS always uses the DB session fixture for every test so you don't need to worry about it.

But if we ever fixed it so that not every test uses the DB session, then tests that want to use DB factories would need to depend on it.

tests/unit/lms/services/grading_info_test.py
```diff
@@ -50,31 +54,29 @@ def test_it_returns_an_empty_list_if_there_are_no_matching_GradingInfos(
...
+            oauth_consumer_key="matching_oauth_consumer_key",
+            context_id="matching_context_id",
+            resource_link_id="matching_resource_link_id",
+        )
+
+        return [
+            matching_grading_info(),
+            matching_grading_info(),
+            matching_grading_info(),
+        ]
```
seanh, line 70: I've fixed this locally to but GitHub won't let me push it. Should be:

```python
        return factories.GradingInfo.create_batch(
            size=3,
            oauth_consumer_key="matching_oauth_consumer_key",
            context_id="matching_context_id",
            resource_link_id="matching_resource_link_id",
        )
```

tests/unit/lms/services/grading_info_test.py
```diff
@@ -50,31 +54,29 @@ def test_it_returns_an_empty_list_if_there_are_no_matching_GradingInfos(
...
-            resource_link_id=f"{prefix}_resource_link_id",
-            tool_consumer_info_product_family_code=f"{prefix}_tool_consumer_info_product_family_code_{index}",
-            h_username=f"{prefix}_h_username_{index}",
-            h_display_name=f"{prefix}_h_display_name_{index}",
-        )
+        return [
+            factories.GradingInfo(),
+            factories.GradingInfo(),
+            factories.GradingInfo(),
+        ]
```
seanh, line 79: I've fixed this locally to but GitHub won't let me push it. Should be just:

```python
    @pytest.fixture(autouse=True)
    def noise_grading_infos(self):
        """Add some GradingInfo's that should *not* match the test query."""
        return factories.GradingInfo.create_batch(3)
```

tests/unit/conftest.py
```diff
@@ -135,14 +136,40 @@ def restart_savepoint(session, transaction):  # pylint:disable=unused-variable
...
+def sqlalchemy_factory_classes():
+    # Return all the SQLAlchemy factory classes from tests.factories.
+    for factory_class in factories.__dict__.values():
+        try:
+            is_sqla_factory = issubclass(factory_class, SQLAlchemyModelFactory)
+        except TypeError:
+            is_sqla_factory = False
+
+        if is_sqla_factory:
+            yield factory_class
```
jon-betts, line 159: I _think_ this can be replaced with:

```suggestion
    return SQLAlchemyModelFactory.__subclasses__()
```

If so, it might be worth in-lining below?

tests/unit/conftest.py
```diff
@@ -135,14 +136,40 @@ def restart_savepoint(session, transaction):  # pylint:disable=unused-variable
...
+
+def set_factory_boy_sqlalchemy_session(session):
+    for factory_class in sqlalchemy_factory_classes():
+        # pylint:disable=protected-access
+        factory_class._meta.sqlalchemy_session = session
+
+
+def clear_factory_boy_sqlalchemy_session():
+    for factory_class in sqlalchemy_factory_classes():
+        factory_class._meta.sqlalchemy_session = None  # pylint:disable=protected-access
```
jon-betts, line 170: Might be nice to have a short comment for these guys saying why we're doing this.

tests/unit/conftest.py
jon-betts: Would be very nice to stop this from being automatic. We could turn it on for whole directories at first and get more fine grained later.

tests/unit/lms/services/grading_info_test.py
jon-betts: Yeah that's a lot nicer

tests/unit/lms/services/grading_info_test.py
jon-betts: The previous stuff was so gross.

tests/unit/conftest.py
seanh: We don't want _all_ subclasses of `SQLAlchemyModelFactory` though, we just want our own subclasses. If for example Factory Boy subclasses that itself internally for some reason we wouldn't want to mess with that.

This would work if we had our own `SQLAlchemyModelFactory` subclass as a base class for all our SQLAlchemy models but then we'd need to implement that base class and remember to always use it instead of the default

tests/unit/conftest.py
seanh: Added comments

tests/unit/conftest.py
seanh: Yep

tests/unit/conftest.py
jon-betts: I'm pretty sure that it doesn't at the moment, but fair enough. Likely a list comprehension that would filter those out too if you're worried.

tests/unit/conftest.py
jon-betts: How would you normally be expected to have a session about like that?

seanh: They seem to expect you to have a module-level session. See their example here: https://factoryboy.readthedocs.io/en/latest/orms.html#sqlalchemy

PR #1655, opened 2020-04-15
Use request.lti user not request.parsed_params

seanh:
Means this method can now be used during any request that has a request.lti_user, not just during LTI launch requests.


seanh:
This got merged as part of another PR

PR #1656, opened 2020-04-15
Add factories.HUser

seanh:
Similar to `factories.LTIUser`, this is going to reduce the amount of test refactoring needed when we add new required parameters to `models.HUser` in the future, and it also forces us to improve our tests:

Add `factories.HUser`, a test factory for `models.HUser`, and change all the tests to use it instead of creating `HUser` objects directly.

This forces us to improve the tests by not specifying specific field values for test `HUser` objects when the test doesn't care about the specific values, but still specifying specific values for certain fields when the test _does_ care about them.

### DIFF
--- a/tests/factories/h_user.py
+++ b/tests/factories/h_user.py
@@ -1,0 +1,10 @@
+import factory
+
+from lms import models
+
+HUser = factory.make_factory(  # pylint:disable=invalid-name
+    models.HUser,
+    authority="lms.hypothes.is",
+    username=factory.Faker("hexify", text="^" * 30),
+    display_name=factory.Faker("name"),
+)
--- a/tests/unit/lms/models/h_user_test.py
+++ b/tests/unit/lms/models/h_user_test.py
@@ -1,8 +1,8 @@
-from lms.models import HUser
+from tests import factories
 
 
 class TestHUser:
     def test_userid(self):
-        h_user = HUser("test_authority", "test_username")
+        h_user = factories.HUser(username="test_username", authority="test_authority")
 
         assert h_user.userid == "acct:test_username@test_authority"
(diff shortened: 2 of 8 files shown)


tests/factories/h_user.py
```diff
@@ -0,0 +1,10 @@
+import factory
+
+from lms import models
+
+HUser = factory.make_factory(  # pylint:disable=invalid-name
+    models.HUser,
+    authority="lms.hypothes.is",
+    username=factory.Faker("hexify", text="^" * 30),
+    display_name=factory.Faker("name"),
+)
```
seanh, line 10: The `username` field looks like one of the usernames that LMS generates for its h users: 30 random hex chars.

tests/unit/lms/models/h_user_test.py
```diff
@@ -1,8 +1,8 @@
-from lms.models import HUser
+from tests import factories
 
 
 class TestHUser:
     def test_userid(self):
-        h_user = HUser("test_authority", "test_username")
+        h_user = factories.HUser(username="test_username", authority="test_authority")
```
jon-betts, line 6: I think using the `kwargs` style is nicer for the reader.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -300,7 +303,7 @@ def test_it_sets_the_focused_user_if_theres_a_focused_user_param(
         assert js_config.asdict()["hypothesisClient"]["focus"] == {
             "user": {
                 "username": "example_h_username",
-                "displayName": "example_h_display_name",
+                "displayName": h_api.get_user.return_value.display_name,
```
jon-betts, line 306: This is nice as it actually tells you why this value is this, rather than just what it is

tests/unit/lms/models/h_user_test.py
seanh: By the way, there is a way to get Factory Boy to pass arguments to the underlying model class as pargs (while still allowing the test code to pass arguments to the test factory as kwargs). You might have to use this if the underlying thing that your test factory calls has positional-only args: https://factoryboy.readthedocs.io/en/latest/reference.html#factory.FactoryOptions.inline_args

PR #1657, opened 2020-04-15
Add new fields to LTIUser

seanh:
Add a few more fields to `models.LTIUser` (the `request.lti_user` object). Specifically, this adds all the fields necessary to generate a `models.HUser` object from _just_ a `models.LTIUser` object, and future PRs will refactor the `HUser`-generating code to do just that.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -40,13 +40,13 @@ class LTILaunchResource:
 
         def display_name():
             """Return the h display name for the current request."""
-            params = self._request.parsed_params
+            lti_user = self._request.lti_user
 
-            display_name = params.get("lis_person_name_full", "").strip()
+            display_name = lti_user.full_name.strip()
 
             if not display_name:
-                given_name = params.get("lis_person_name_given", "").strip()
-                family_name = params.get("lis_person_name_family", "").strip()
+                given_name = lti_user.given_name.strip()
+                family_name = lti_user.family_name.strip()
 
                 display_name = " ".join((given_name, family_name)).strip()
 
@@ -169,12 +169,12 @@ class LTILaunchResource:
     @property
     def h_provider(self):
         """Return the h "provider" string for the current request."""
-        return self._request.parsed_params["tool_consumer_instance_guid"]
+        return self._request.lti_user.tool_consumer_instance_guid
 
     @property
     def h_provider_unique_id(self):
         """Return the h provider_unique_id for the current request."""
-        return self._request.parsed_params["user_id"]
+        return self._request.lti_user.user_id
 
     @property
     def is_canvas(self):
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -38,31 +38,10 @@ class LTILaunchResource:
             username_hash_object.update(self.h_provider_unique_id.encode())
             return username_hash_object.hexdigest()[:30]
 
-        def display_name():
-            """Return the h display name for the current request."""
-            lti_user = self._request.lti_user
-
-            display_name = lti_user.full_name.strip()
-
-            if not display_name:
-                given_name = lti_user.given_name.strip()
-                family_name = lti_user.family_name.strip()
-
-                display_name = " ".join((given_name, family_name)).strip()
-
-            if not display_name:
-                return "Anonymous"
-
-            # The maximum length of an h display name.
-            display_name_max_length = 30
-
-            if len(display_name) <= display_name_max_length:
-                return display_name
-
-            return display_name[: display_name_max_length - 1].rstrip() + "…"
-
         return HUser(
-            authority=self._authority, username=username(), display_name=display_name()
+            authority=self._authority,
+            username=username(),
+            display_name=self._request.lti_user.display_name,
         )
 
     @property
--- a/lms/validation/authentication/_bearer_token.py
+++ b/lms/validation/authentication/_bearer_token.py
@@ -60,6 +60,7 @@ class BearerTokenSchema(PyramidRequestSchema):
     user_id = marshmallow.fields.Str(required=True)
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
+    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
 
     def __init__(self, request):
         super().__init__(request)
--- a/lms/validation/authentication/_bearer_token.py
+++ b/lms/validation/authentication/_bearer_token.py
@@ -61,9 +61,7 @@ class BearerTokenSchema(PyramidRequestSchema):
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
     tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
-    given_name = marshmallow.fields.Str(required=True)
-    family_name = marshmallow.fields.Str(required=True)
-    full_name = marshmallow.fields.Str(required=True)
+    display_name = marshmallow.fields.Str(required=True)
 
     def __init__(self, request):
         super().__init__(request)
--- a/lms/validation/authentication/_bearer_token.py
+++ b/lms/validation/authentication/_bearer_token.py
@@ -61,6 +61,9 @@ class BearerTokenSchema(PyramidRequestSchema):
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
     tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    given_name = marshmallow.fields.Str(required=True)
+    family_name = marshmallow.fields.Str(required=True)
+    full_name = marshmallow.fields.Str(required=True)
 
     def __init__(self, request):
         super().__init__(request)
--- a/lms/validation/authentication/_launch_params.py
+++ b/lms/validation/authentication/_launch_params.py
@@ -31,6 +31,9 @@ class LaunchParamsAuthSchema(PyramidRequestSchema):
     user_id = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
     tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    lis_person_name_given = marshmallow.fields.Str(missing="")
+    lis_person_name_family = marshmallow.fields.Str(missing="")
+    lis_person_name_full = marshmallow.fields.Str(missing="")
 
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     oauth_nonce = marshmallow.fields.Str(required=True)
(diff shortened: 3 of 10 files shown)


lms/validation/authentication/_bearer_token.py
```diff
@@ -60,6 +60,10 @@ class BearerTokenSchema(PyramidRequestSchema):
     user_id = marshmallow.fields.Str(required=True)
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
+    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    given_name = marshmallow.fields.Str(required=True)
+    family_name = marshmallow.fields.Str(required=True)
+    full_name = marshmallow.fields.Str(required=True)
```
seanh, line 66: Adding these to `BearerTokenSchema` means that:

1. When the backend serializes `request.lti_user` into a bearer token to go in the JS config, that seralized bearer token now contains these additional fields

2. When the frontend uses that bearer token to authenticate an API request and the backend deserializes the bearer token into a `request.lti_user` object, that `LTIUser` object will now have the new fields

lms/validation/authentication/_launch_params.py
```diff
@@ -30,6 +30,10 @@ class LaunchParamsAuthSchema(PyramidRequestSchema):
 
     user_id = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
+    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    lis_person_name_given = marshmallow.fields.Str(missing="")
+    lis_person_name_family = marshmallow.fields.Str(missing="")
+    lis_person_name_full = marshmallow.fields.Str(missing="")
```
seanh, line 36: Adding these to `LaunchParamsAuthSchema` means that when we receive an LTI launch request and deserialize a `request.lti_user` from the verified launch params, that `LTIUser` object will now have the new fields

jon-betts:
As far as I can tell the only thing we use `given_name`, `family_name`, and `full_name` for is to generate the display name. I don't think we ever use them individually outside of `lti_launch.LTILaunchResource.h_user()`'s inner function `display_name()`. So it would seem we're doing a lot of effort to preserve the strings just to splodge them all together later.

This would suggest that we have a `display_name` proxy on the `LTIUser`, but this doesn't seem quite right as `display_name` is a property of the `HUser`.

I think the solution might be to introduce a `name` model to the LTIUser like this:

```
class DisplayName(NamedTuple):
    given: str
    """The user's given name from the lis_person_name_given LTI param."""

    family: str
    """The user's family name from the lis_person_name_family LTI param."""

    full: str
    """The user's full name from the lis_person_name_full LTI param."""

    def __str__(self):
        # ... display_name() logic here

   @classmethod
   def from_string(...):
        # etc.
   ...
```

This object could contain a stringify method which effectively contains the `display_name()` logic. If we can also create it from a string then the creation of the `LTIUser` object could either be:

```
LTIUser(
    user_id=...,
    oauth_consumer_key=...,
    ...,
    # Could put the `DisplayName` object here, or let `LTIUser` deal with that for us
    name={
        "given": ...,
        "family": ...,
        "full": ...,
    }
)
```

... when we are coming from LTI params and:

```
LTIUser(
    user_id=...,
    oauth_consumer_key=...,
    ...,
    name="already concatenated`
)
```
... the rest of the time (for bearer token etc).

Both the `HUser` and `LTIUser` could directly share this object as their `name` and `display_name` respectively.

This is Ok to do for the `LTIUser` because it's a domain model, rather than a schema. It represents our logical understanding of the object rather than the exact fields. As a nice bonus also gives us the flexibility for example to have a different name object should LTI 1.3 be different etc.

This would kind of fundamentally change the PR, so it's probably worth talking about before we go any further.

seanh:
@jon-betts In order to introduce the `display_name` field I had to do add a bunch of commits:

1. This test needed to be refactored: https://github.com/hypothesis/lms/pull/1657/commits/fb6a16c988fcb42f4ccda7302b6a75f0bf3a536c

2. Changed the `display_name`-generating code (still in its previous location in `LTILaunchResource`, where it generates the `display_name` for an `HUser`) to read params from `request.lti_user` instead of from `request.parsed_params`: https://github.com/hypothesis/lms/pull/1657/commits/431e20d6dd3c5025f268b988af869b1fe4b2fb08

3. Put all the `display_name` tests in one place: https://github.com/hypothesis/lms/pull/1657/commits/40f57ce7b6d2b91766b610842494e731904b64ec

4. Finally, simplify `LTIUser` to have just `display_name`

lms/models/_display_name.py
```diff
@@ -0,0 +1,27 @@
...
+    if not name:
+        return "Anonymous"
+
+    # The maximum length of an h display name.
+    display_name_max_length = 30
+
+    if len(name) <= display_name_max_length:
+        return name
+
+    return name[: display_name_max_length - 1].rstrip() + "…"
```
seanh, line 27: This is `_display_name.py` because the module only contains collaborators used internally within `lms.models`, it's not meant to be used outside of `lms.models`. The function is named `display_name()` (no underscore) because it's part of the `lms.models._display_name` module's public interface: it's meant to be called by code outside of the module.

lms/resources/lti_launch.py
```diff
@@ -38,31 +38,10 @@ def username():
...
-            if len(display_name) <= display_name_max_length:
-                return display_name
-
-            return display_name[: display_name_max_length - 1].rstrip() + "…"
-
         return HUser(
-            authority=self._authority, username=username(), display_name=display_name()
+            authority=self._authority,
+            username=username(),
+            display_name=self._request.lti_user.display_name,
```
seanh, line 44: We no longer need to generate a `display_name` when we create an `HUser`. It's now generated earlier, when we create the `LTIUser` (in `LaunchParamsAuthSchema`)

lms/validation/authentication/_bearer_token.py
```diff
@@ -60,6 +60,8 @@ class BearerTokenSchema(PyramidRequestSchema):
     user_id = marshmallow.fields.Str(required=True)
     oauth_consumer_key = marshmallow.fields.Str(required=True)
     roles = marshmallow.fields.Str(required=True)
+    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    display_name = marshmallow.fields.Str(required=True)
```
seanh, line 64: `BearerTokenSchema` can just continue doing what it always did to create an `LTIUser`: just `LTIUser(**data)`. But `data` will now contain the `tool_consumer_instance_guid` and `display_name` fields.

lms/validation/authentication/_launch_params.py
```diff
@@ -52,7 +56,15 @@ def lti_user(self):
...
-        return LTIUser(kwargs["user_id"], kwargs["oauth_consumer_key"], kwargs["roles"])
+        return LTIUser.from_names(
+            kwargs["lis_person_name_given"],
+            kwargs["lis_person_name_family"],
+            kwargs["lis_person_name_full"],
+            user_id=kwargs["user_id"],
+            oauth_consumer_key=kwargs["oauth_consumer_key"],
+            roles=kwargs["roles"],
+            tool_consumer_instance_guid=kwargs["tool_consumer_instance_guid"],
+        )
```
seanh, line 67: So this is where we now generate the display name (`LaunchParamsAuthSchema` calls `from_names()`)

tests/unit/lms/models/lti_user_test.py
```diff
@@ -31,3 +33,22 @@ def test_is_learner(self, roles, is_learner):
...
+            given_name="Gareth", family_name="Smith", full_name="", **kwargs,
+        )
+
+        display_name.assert_called_once_with("Gareth", "Smith", "")
+        assert user.display_name == display_name.return_value
+
+
+@pytest.fixture(autouse=True)
+def display_name(patch):
+    return patch("lms.models.lti_user.display_name")
```
seanh, line 54: `LTIUser` and `display_name()` are a pretty small-scale case of separate units, but I still decided to patch `display_name()` in `LTIUser`'s tests because:

* `display_name()` has its own separate unit tests and those should be responsible for `display_name()`'s test coverate. `LTIUser`'s unit tests aren't responsible for testing `display_name()`. We don't want `LTIUser`'s unit tests contributing to `display_name()`'s test coverage, and potentially hiding coverage holes in `display_name()`'s actual unit tests

* `display_name()` has its own separate unit tests and those are responsible for failing or crashing if there's a bug in `display_name()`. We don't want `LTIUser`'s unit tests failing or crashing because of bugs in `display_name()`. They should only fail or crash because of bugs in `LTIUser`

* It makes the tests read more clearly about what `LTIUser`'s responsibilities actually are. `assert user.display_name == display_name.return_value` tells you that `LTIUser`'s job is just to blindly use whatever `display_name()` returns as the display name. Something like `assert user.display_name == "Gareth Smith"`, which depends on the particular value returned by 'display_name()`'s actual implementation, confuses what the `LTIUser` tests are about

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -250,23 +201,19 @@ def test_it(self, pyramid_request, parsed_params, is_canvas):
...
-    def test_userid(self, pyramid_request):
-        userid = LTILaunchResource(pyramid_request).h_user.userid
-
-        assert userid == "acct:16aa3b3e92cdfa53e5996d138a7013@TEST_AUTHORITY"
+    def test_it(self, pyramid_request):
+        assert LTILaunchResource(pyramid_request).h_user == HUser(
+            authority=pyramid_request.registry.settings["h_authority"],
+            username="16aa3b3e92cdfa53e5996d138a7013",
+            display_name=pyramid_request.lti_user.display_name,
+        )
```
seanh, line 209: This is what remains of the tests for `LTILaunchResource.h_user()`: just one tiny test is enough

tests/unit/lms/validation/authentication/_launch_params_test.py
```diff
@@ -1,18 +1,38 @@
...
+        LTIUser.from_names.assert_called_once_with(
+            "TEST_GIVEN_NAME",
+            "TEST_FAMILY_NAME",
+            "TEST_FULL_NAME",
+            user_id="TEST_USER_ID",
+            oauth_consumer_key="TEST_OAUTH_CONSUMER_KEY",
+            roles="TEST_ROLES",
+            tool_consumer_instance_guid="TEST_TOOL_CONSUMER_INSTANCE_GUID",
+        )
+        assert lti_user == LTIUser.from_names.return_value
```
seanh, line 22: Previously `LTIUser` was just a simple value class, but now its grown a non-trivial factory method (which calls a non-trivial private collaborator) so I decided to patch `LTIUser` in `LaunchParamsAuthSchema`'s unit tests. Some reasoning as for patching `display_name()` in `LTIUser`s unit tests

tests/unit/lms/models/lti_user_test.py
jon-betts: I think this is the downside of having the factory function. The length cost isn't the main code, but the tests. By creating this interface between `LTIUser` and `display_name()` and coupling them, we now have to test and police it.

If it was simply the users choice to use `display_name()` and provide the string to `LTIUser()` init method, we wouldn't have to.

I believe (correct me if wrong), this code will only ever be called once, at which point we will presumably have to test _that_ boundary too, mocking `from_names`. If that caller did both we could kill two birds with one stone there.

tests/unit/lms/validation/authentication/_launch_params_test.py
jon-betts: This is another point where adding the factory method has imposed a cost. With the other approach the LTIUser object would remain a plain value object.

PR #1661, opened 2020-04-17
Add provider and provider_unique_id fields to HUser

seanh:
`LTILaunchResource` currently has three separate properties:

1. `LTILaunchResource.h_user`. Returns a `models.HUser` object.

2. `LTILaunchResource.h_provider`

3. `LTILaunchResource.h_provider_unique_id`

`h_provider` and `h_provider_unique_id` are really part of the `HUser`:
they're only really used by `LTILaunchResource.h_user`.

Add two new attributes `provider` and `provider_unique_id` to `models.HUser`,
change `LTILaunchResource.h_user` to set them, and delete
`LTILaunchResource.h_provider` and `LTILaunchResource.h_provider_unique_id`.

Code that was previously reading `context.h_provider` and
`context.h_provider_unique_id` can now read `context.h_user.provider`
and `context.h_user.provider_unique_id` instead.

This also simplifies the `HAPI` service as `HAPI.create_user()` and
`HAPI.upsert_user()` no longer need to take a separate `provider` and
`provider_unique_id` params in addition to the `h_user` param.

`provider` and `provider_unique_id` are added to as optional fields to `HUser` because there are a couple of places other than `LTILaunchResource` that create `HUser` objects and setting a `provider` and `provider_unique_id` doesn't make sense in those places.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -31,17 +31,25 @@ class LTILaunchResource:
     def h_user(self):
         """Return the h user for the current request."""
 
+        # The h "provider" string for the current request.
+        provider = self._request.lti_user.tool_consumer_instance_guid
+
+        # The h "provider_unique_id" string for the current request.
+        provider_unique_id = self._request.lti_user.user_id
+
         def username():
             """Return the h username for the current request."""
             username_hash_object = hashlib.sha1()
-            username_hash_object.update(self.h_provider.encode())
-            username_hash_object.update(self.h_provider_unique_id.encode())
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
             return username_hash_object.hexdigest()[:30]
 
         return HUser(
             authority=self._authority,
             username=username(),
             display_name=self._request.lti_user.display_name,
+            provider=provider,
+            provider_unique_id=provider_unique_id,
         )
 
     @property
@@ -144,16 +152,6 @@ class LTILaunchResource:
             name = name[: group_name_max_length - 1].rstrip() + "…"
 
         return name
-
-    @property
-    def h_provider(self):
-        """Return the h "provider" string for the current request."""
-        return self._request.lti_user.tool_consumer_instance_guid
-
-    @property
-    def h_provider_unique_id(self):
-        """Return the h provider_unique_id for the current request."""
-        return self._request.lti_user.user_id
 
     @property
     def is_canvas(self):
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -42,21 +42,22 @@ class HAPI:
             display_name=user_info["display_name"],
         )
 
-    def create_user(self, h_user, provider, provider_unique_id):
+    def create_user(self, h_user):
         """
         Create a user in H.
 
         :arg h_user: the user to be created in h
         :type h_user: HUser
-        :param provider: the "provider" string to send to h
-        :param provider_unique_id:  the "provider_unique_id" string to send to h
         """
         user_data = {
             "username": h_user.username,
             "display_name": h_user.display_name,
             "authority": self._authority,
             "identities": [
-                {"provider": provider, "provider_unique_id": provider_unique_id,}
+                {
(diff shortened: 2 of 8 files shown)


lms/resources/lti_launch.py
```diff
@@ -31,17 +31,25 @@ def __init__(self, request):
...
+        provider_unique_id = self._request.lti_user.user_id
+
         def username():
             """Return the h username for the current request."""
             username_hash_object = hashlib.sha1()
-            username_hash_object.update(self.h_provider.encode())
-            username_hash_object.update(self.h_provider_unique_id.encode())
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
             return username_hash_object.hexdigest()[:30]
```
jon-betts, line 45: Is this moving into the user object in a different PR?

lms/services/h_api.py
```diff
@@ -42,21 +42,22 @@ def get_user(self, username):
...
 
-    def create_user(self, h_user, provider, provider_unique_id):
+    def create_user(self, h_user):
         """
         Create a user in H.
 
         :arg h_user: the user to be created in h
         :type h_user: HUser
-        :param provider: the "provider" string to send to h
-        :param provider_unique_id:  the "provider_unique_id" string to send to h
```
jon-betts, line 52: Nice!

lms/services/h_api.py
```diff
@@ -42,21 +42,22 @@ def get_user(self, username):
...
         """
         user_data = {
             "username": h_user.username,
             "display_name": h_user.display_name,
             "authority": self._authority,
```
jon-betts, line 55:
```suggestion
            "authority": h_user.authority,
```

I think this works, and it would make it clearer that this is all about the user. Might pave the way for removing `_authority` from this object in future if it's not used.

lms/services/h_api.py
```diff
@@ -77,7 +78,7 @@ def update_user(self, h_user):
             data={"display_name": h_user.display_name},
         )
```
jon-betts, line 79: ~I think in future it would be totally Ok for the HUser to know how to turn itself into an H compatible dict (`as_dict()`?) as all the parts come from it.~

Yeah ignore that. This is the H-API. While this is in charge of making API calls, we don't want to have to go hunting around for what the payload is in other files.

This might be more appropriate when the Bulk API is in place.

lms/services/h_api.py
```diff
@@ -77,7 +78,7 @@ def update_user(self, h_user):
...
 
-    def upsert_user(self, h_user, provider, provider_unique_id):
+    def upsert_user(self, h_user):
```
jon-betts, line 81: A much nicer interface again

lms/services/h_api.py
```diff
@@ -86,13 +87,11 @@ def upsert_user(self, h_user, provider, provider_unique_id):
...
-        :param provider: the "provider" string to send to h
-        :param provider_unique_id:  the "provider_unique_id" string to send to h
         """
         try:
             self.update_user(h_user)
         except HAPINotFoundError:
-            self.create_user(h_user, provider, provider_unique_id)
+            self.create_user(h_user)
 
     def upsert_group(self, group_id, group_name):
```
jon-betts, line 96: A group object like the user would work really well here too.

seanh:
Rebased

lms/services/h_api.py
seanh: Done. We're unlikely to be able to remove `authority` from `HAPIService` though, it's used for a bunch of things

lms/services/h_api.py
seanh: namedtuples do come with an `_asdict()` method for free and the resulting dict happens to be close to (but not exactly) what's needed for the h create user API. But not what the update user API needs. It's probably fine for `HAPIService` to handle serializing things into the necessary formats for the different API calls though, as you say

lms/resources/lti_launch.py
seanh: That's coming, yes

PR #1662, opened 2020-04-17
Extract HUser creation from LTILaunchResource

seanh:
Extract `HUser` creation out of `LTILaunchResource`. This is so that
requests outside of an `LTILaunchResource` context (e.g. API requests) can
create `HUsers`.

First, add a new classmethod `models.HUser.from_lti_user(lti_user,
authority)` that returns a new `HUser` based on a given `LTIUser` and the
`"h_authority"` setting. Move most of the code that was previously
`LTILaunchResource.h_user()` into this new classmethod.

Next, add a new property `LTIHService.h_user` which returns the `HUser` for
the current request, just: `models.HUser.from_lti_user(request.lti_user,
request.registry.settings["h_authority")`. That replaces the remaining
code in `LTILaunchResource.h_user`. Code can now access
`request.find_service(name="lti_h").h_user` instead of `context.h_user`.
`LTILaunchResource.h_user` can be deleted.

Finally, change all the places that were reading `context.h_user` to read
`request.find_service(name="lti_h").h_user` instead: `LTIHService` itself (now reads `self.h_user`)
`JSConfig`, and `BasicLTILaunchViews`.

### DIFF
--- a/lms/models/h_user.py
+++ b/lms/models/h_user.py
@@ -1,3 +1,4 @@
+import hashlib
 from typing import NamedTuple
 
 
@@ -32,6 +33,26 @@ class HUser(NamedTuple):
 
     provider_unique_id: str = ""
     """The "provider_unique_id" string to pass to the h API for this user."""
+
+    @classmethod
+    def from_lti_user(cls, lti_user, authority):
+        provider = lti_user.tool_consumer_instance_guid
+        provider_unique_id = lti_user.user_id
+
+        def username():
+            """Return the h username for the current request."""
+            username_hash_object = hashlib.sha1()
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
+            return username_hash_object.hexdigest()[:30]
+
+        return cls(
+            authority=authority,
+            username=username(),
+            display_name=lti_user.display_name,
+            provider=provider,
+            provider_unique_id=provider_unique_id,
+        )
 
     @property
     def userid(self):
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -10,7 +10,7 @@ from lms.validation.authentication import BearerTokenSchema
 from lms.views.helpers import via_url
 
 
-class JSConfig:  # pylint:disable=too-few-public-methods
+class JSConfig:  # pylint:disable=too-few-public-methods,too-many-instance-attributes
     """The config for the app's JavaScript code."""
 
     def __init__(self, context, request):
@@ -24,6 +24,7 @@ class JSConfig:  # pylint:disable=too-few-public-methods
         self._grading_info_service = request.find_service(name="grading_info")
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
+        self._lti_h = request.find_service(name="lti_h")
 
     def add_canvas_file_id(self, canvas_file_id):
         """
@@ -193,7 +194,7 @@ class JSConfig:  # pylint:disable=too-few-public-methods
 
         self._config["canvas"]["speedGrader"] = {
             "submissionParams": {
-                "h_username": self._context.h_user.username,
+                "h_username": self._lti_h.h_user.username,
                 "lis_result_sourcedid": lis_result_sourcedid,
                 "lis_outcome_service_url": lis_outcome_service_url,
                 **kwargs,
@@ -330,7 +331,7 @@ class JSConfig:  # pylint:disable=too-few-public-methods
         claims = {
             "aud": urlparse(api_url).hostname,
             "iss": self._request.registry.settings["h_jwt_client_id"],
-            "sub": self._context.h_user.userid,
+            "sub": self._lti_h.h_user.userid,
             "nbf": now,
             "exp": now + datetime.timedelta(minutes=5),
         }
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -2,6 +2,7 @@ from collections import namedtuple
 
 from pyramid.httpexceptions import HTTPInternalServerError
 
+from lms.models import HUser
 from lms.services import HAPIError
 
 Group = namedtuple("Group", "name groupid")
@@ -24,11 +25,17 @@ class LTIHService:
     def __init__(self, _context, request):
         self._context = request.context
         self._request = request
+        self._authority = self._request.registry.settings["h_authority"]
         self._lti_user = request.lti_user
 
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
         self._group_info_service = request.find_service(name="group_info")
+
+    @property
+    def h_user(self):
+        """Return the h user for the current request's LTI user."""
+        return HUser.from_lti_user(self._lti_user, self._authority)
 
     def single_group_sync(self):
         """
@@ -64,9 +71,7 @@ class LTIHService:
             raise HTTPInternalServerError(explanation=err.explanation) from err
 
     def _sync_to_h(self, groups):
-        h_user = self._context.h_user
-
-        self._h_api.upsert_user(h_user=h_user)
+        self._h_api.upsert_user(h_user=self.h_user)
 
         for group in groups:
             self._h_api.upsert_group(group_id=group.groupid, group_name=group.name)
@@ -78,4 +83,4 @@ class LTIHService:
             )
 
         for group in groups:
-            self._h_api.add_user_to_group(h_user=h_user, group_id=group.groupid)
+            self._h_api.add_user_to_group(h_user=self.h_user, group_id=group.groupid)
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -4,7 +4,6 @@ import hashlib
 
 from pyramid.security import Allow
 
-from lms.models import HUser
 from lms.resources._js_config import JSConfig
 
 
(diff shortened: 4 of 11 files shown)


lms/models/h_user.py
```diff
@@ -33,6 +34,26 @@ class HUser(NamedTuple):
...
+    def from_lti_user(cls, lti_user, authority):
+        provider = lti_user.tool_consumer_instance_guid
+        provider_unique_id = lti_user.user_id
+
+        def username():
+            """Return the h username for the current request."""
+            username_hash_object = hashlib.sha1()
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
+            return username_hash_object.hexdigest()[:30]
```
jon-betts, line 47: I think this would be better as a cached property like `userid`. Then this function would be:
```python
    @classmethod
    def from_lti_user(cls, lti_user, authority):
        return cls(
            authority=authority,
            display_name=lti_user.display_name,
            provider=lti_user.tool_consumer_instance_guid,
            provider_unique_id=lti_user.user_id,
        )
```

lms/resources/_js_config.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, context, request):
         self._grading_info_service = request.find_service(name="grading_info")
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
+        self._lti_h = request.find_service(name="lti_h")
```
jon-betts, line 27: This seems a bit squiffy. This class doesn't want the LTI-H service, just an `HUser`. I think we're using the `LTI-H` service here because it's gained a `h_user` property, which just calls `from_lti_user`,but we only need _that_ because the `LTIUser` doesn't know the authority.

I think we can simplify this as per our discussion:

 * Let the `LTIUser` conceptually be all user data in an LTI context (rather than a user explictly created from or representing LTI params). We started this by adding the `display_name`
 * Add authority to the `LTIUser`
 * Move `h_user` to be a property of `LTIUser`

The result in this class would be `self._lti_user.h_user`

lms/services/lti_h.py
```diff
@@ -24,12 +25,18 @@ class LTIHService:
     def __init__(self, _context, request):
         self._context = request.context
         self._request = request
+        self._authority = self._request.registry.settings["h_authority"]
         self._lti_user = request.lti_user
```
jon-betts, line 29: If the `LTIUser` already contained the authority we would not need to add this here.

lms/services/lti_h.py
```diff
@@ -24,12 +25,18 @@ class LTIHService:
...
 
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
         self._group_info_service = request.find_service(name="group_info")
 
+    @property
+    def h_user(self):
+        """Return the h user for the current request's LTI user."""
+        return HUser.from_lti_user(self._lti_user, self._authority)
+
```
jon-betts, line 39: We also wouldn't need this

lms/models/h_user.py
jon-betts: As per later in this PR, if the `lti_user` had `authority` this could just be an `h_user` property on the `lti_user` as well

seanh: Unfortunately there are five places in the code where `HUser`'s are created, and the `username` only comes from `provider` and `provider_unique_id` in one of those places, so it can't be a computed property:

1. `LTIHService` is the one place where `from_lti_user()` is called and `username` is computed from `provider` and `provider_unique_id`:

   https://github.com/hypothesis/lms/blob/a9cd1fd77aa8f046bcd8e6989a06a3f7876bf76e/lms/services/lti_h.py#L35-L38

2. `JSConfig` creates `HUser`'s whose `username` comes from the `lis_result_sourcedid` (`GradingInfo`) table in the DB, and who do not have a `provider` or `provider_unique_id`:

   https://github.com/hypothesis/lms/blob/a9cd1fd77aa8f046bcd8e6989a06a3f7876bf76e/lms/resources/_js_config.py#L315-L319

3. `HAPIService` creates the `"lms"` machine user with a hardcoded `username`:

   https://github.com/hypothesis/lms/blob/a9cd1fd77aa8f046bcd8e6989a06a3f7876bf76e/lms/services/h_api.py#L118-L120

4. `HAPIService` creates an `HUser` with a given `username` and no `provider` or `provider_unique_id`, in order to generate a `userid` from it. This is related to the non-Canvas grading feature again:

   https://github.com/hypothesis/lms/blob/a9cd1fd77aa8f046bcd8e6989a06a3f7876bf76e/lms/services/h_api.py#L36

5. `HAPIService` creates an `HUser` with a `username` and `display_name` received from the h API and no `provider` or `provider_unique_id`. Non-Canvas grading, again:

   https://github.com/hypothesis/lms/blob/a9cd1fd77aa8f046bcd8e6989a06a3f7876bf76e/lms/services/h_api.py#L39-L43

I think the factory method approach is more flexible and, given how `HUser` is used, is appropriate:

1. If you have an `LTIUser` and want an `HUser` from it you use `HUser.from_lti_user()` and `username` is computed for you from `provider` and `provider_unique_id`

2. If you want to create an `HUser` with a given `username` (and potentially without a `provider` or `provider_unique_id`) you do `HUser(username="...", ...)`

lms/resources/_js_config.py
seanh: Yes, I've expanded the responsibility of `LTIHService` to include generating `HUser`'s from `LTIUser`'s, which seems an appropriate responsibility for the LTI -> H translation service. It's just `HUser.from_lti_user(request.lti_user, request.registry.settings["h_authority"])` though, so this property could be put anywhere that has access to the `request`. For example `request.h_user` should be doable.

I think adding `authority` to `LTIUser` is controversial in a couple of ways:

1. `authority` is an h concept, not an LTI concept, and `LTIUser` is `LTIUser` not `HUser`. It's specifically about the LTI user in what it represents and how it's used, currently.

   When an `HUser` is generated from an `LTIUser` that seems like the appropriate time for `authority` to come in to the picture, because LTI users don't have authorities but h users do. So `HUser.from_lti_user(lti_user, authority)` seems appropriate to me. Those are the correct two ingredients: everything needed for an LMS app h user comes from two ingredients, an LTI user and the LMS app's authority setting. Conceptually it's actually _not_ the case that you can generate an h user from just an LTI user.

   Turning `LTIUser` into just `User` seems like a much bigger move to me, and I'm not sure about it. `LTIUser` is specifically about LTI users in what it represents and what it's used for. I think it would need more thinking through. It certainly goes beyond what I need to do right now (which is to decouple `LTIHService` from `LTILaunchResource` so that the sync API can call it).

2. Even though it has crept into `models.HUser` already, I think `authority` should likely be treated as a deployment setting, not as part of the domain model. There's only ever one authority in play, which is `request.registry.settings["h_authority"]` (and is always `"lms.hypothes.is"`). There aren't different users and groups with different authorities. It's a setting.

   This might be the correct way to enable an `LTIUser.h_user` property to exist: remove the `authority` attribute from `HUser` so it's gone from the domain model entirely. Just have services that actually need to pass the authority to h (`HAPIService`) read the authority setting. But that's a bigger refactoring that I don't want to bite off yet

lms/models/h_user.py
jon-betts: Fair enough, it seems it's more often _not_ created this way than created this way. This gives me a few thoughts:

I can't see any reason not to do what we did on the previous PR here. So remove the factory method, have a free floating `username_from_provider`. This would also mean we don't have the "chase the code" in LTI-H. If LTI-H is going to be responsible for this it should probably do the work. It would end up pretty short, and eliminate another interface causing testing woes. Maybe something like:

```
@property
def h_user(self):
    """Return the h user for the current request's LTI user."""
    provider = self._lti_user.tool_consumer_instance_guid,
    provider_unique_id = self._lti_user.user_id,

    return HUser(
        authority=self._authority,
        username=username_from_provider(provider, provider_unique_id),
        display_name=self._lti_user.display_name,
        provider=provider,
        provider_unique_id=provider_unique_id,
    )
```

I think if we hadn't started with a `NamedTuple` we wouldn't be having the trouble we are. A normal class could accomodate all of the access patterns above, but creating username when it's null. Maybe something like:

```
class HUser:
    def __init__(self, authority, display_name, provider, provider_unique_id, username=None):
        self.authority = authority
        self.display_name = display_name
        self.provider = provider
        self.provider_unique_id = provider_unique_id

        self.username = self._make_username(provider, provider_unique_id) \
            if username is None else username

    @staticmethod
    def _make_username(provider, provider_unique_id):
        digest = hashlib.sha1()
        digest.update(provider.encode())
        digest.update(provider_unique_id.encode())

        return digest.hexdigest()[:30]
```

lms/resources/_js_config.py
jon-betts: If we can get rid of authority from both, then I think that's the best outcome. I recommended renaming "LTIUser" to get around this issue. It would become what we need it to be to get the job done, rather than being specific to LTI. But as you say, it's probably more akin to a global variable, rather than copying it into every group and user.

I'm less happy about expanding `LTIH`. It was originally called "LTIHSyncService" or something more long winded, and is/was specifically about ferrying data from LMS to H for it to store. A lot of the time that's not what the `HUser` is used for, so I think it sticks out like a sore thumb. `LTIH `currently has a pretty tightly defined role, and I don't think it's made better here. It feels like this needed a home, and this was the closest.

This is probably a small, but distinct service. As per our discussion in H, as this only returns one thing, we would follow the service factory pattern where you just get an `HUser` right away:

```self._h_user = request.find_service(name="h_user")```

If we eliminate the factory method on the user object it won't seem quite as pathetic.

seanh:
Closing this in favor of https://github.com/hypothesis/lms/pull/1667

PR #1664, opened 2020-04-20
Remove authority from models.HUser

seanh:
Remove the `authority` attribute from `models.HUser`, and change the
`userid()` property to a `userid(authority)` instance method.

This makes creating `HUser`'s easier because you don't need
`request.registry.settings["h_authority"]` to create one.

This comes at the cost of making getting an `HUser`'s `userid` more
difficult: you now need to pass in
`request.registry.settings["h_authority"]` to the `userid` method.
Previously you could have accessed `h_user.userid` without passing in
the authority, if you had been given an already-created `HUser` from
somewhere else. But if you had to create the `HUser` yourself then you
would have needed the `"h_authority"` setting to create it anyway.

In practice I think the code is overall simpler and more flexible if we
delay needing the `"h_authority"` setting until if and when it's needed:
if and when something needs the `userid`. It turns out only two places
at the "edges" of the code actually need the `userid`:

1. `JSConfig` needs it when generating the JavaScript config to be
   passed to the client

2. `HAPIService` needs it to pass to the h API

Rather than forcing `"h_authority"` to be needed whenever an `HUser` is
created. This way, code that doesn't need the `userid`, can create
`HUser`'s without ever needing the `"h_authority"` setting.

I think it also makes sense to remove the concept of authority from our
domain model. There is only ever one authority: the value of the
`request.registry.settings["h_authority"]` setting. All domain model
objects would always have the same authority. It's a setting, not part
of the domain model.

### DIFF
--- a/lms/models/h_user.py
+++ b/lms/models/h_user.py
@@ -10,13 +10,6 @@ class HUser(NamedTuple):
     usernames.
     """
 
-    authority: str
-    """
-    The authority which the user belongs to.
-
-    This should always match the h_authority setting.
-    """
-
     username: str
     """
     Generated username for h user.
@@ -33,6 +26,5 @@ class HUser(NamedTuple):
     provider_unique_id: str = ""
     """The "provider_unique_id" string to pass to the h API for this user."""
 
-    @property
-    def userid(self):
-        return f"acct:{self.username}@{self.authority}"
+    def userid(self, authority):
+        return f"acct:{self.username}@{authority}"
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -10,7 +10,7 @@ from lms.validation.authentication import BearerTokenSchema
 from lms.views.helpers import via_url
 
 
-class JSConfig:  # pylint:disable=too-few-public-methods
+class JSConfig:  # pylint:disable=too-many-instance-attributes
     """The config for the app's JavaScript code."""
 
     def __init__(self, context, request):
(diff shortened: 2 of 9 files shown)


lms/models/h_user.py
```diff
@@ -33,6 +26,5 @@ class HUser(NamedTuple):
     provider_unique_id: str = ""
     """The "provider_unique_id" string to pass to the h API for this user."""
 
-    @property
-    def userid(self):
-        return f"acct:{self.username}@{self.authority}"
+    def userid(self, authority):
+        return f"acct:{self.username}@{authority}"
```
seanh, line 30: We'll follow exactly the same technique with the new `models.HGroup` in a future PR: it'll have a `groupid(authority)` method

lms/resources/_js_config.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, context, request):
         self._consumer_key = (
             request.lti_user.oauth_consumer_key if request.lti_user else None
         )
+        self._authority = request.registry.settings["h_authority"]
```
seanh, line 23: `JSConfig` was already using the authority setting (see below), I just added a convenience variable for it. `JSConfig` is one of the two places where it makes sense for the authority setting to be used: it's one of the "edges" of the app where we render userid's and groupid's for passing to the Hypothesis client, and those IDs must contain the authority

lms/services/h_api.py
```diff
@@ -33,14 +33,10 @@ def get_user(self, username):
 
         :rtype: HUser
         """
-        userid = HUser(self._authority, username).userid
+        userid = HUser(username).userid(self._authority)
```
seanh, line 36: `HAPIService` was also already using the authority. This is the other "edge" of the app, where we have to render userid's and groupid's containing the authority for sending to the h API

jon-betts: I'm guessing this is out of sequence with the other PRs, as this has all shipped out!

lms/models/h_user.py
jon-betts: Makes sense to me.

PR #1667, opened 2020-04-20
Move LTILaunchResource.h_user to request.lti_user.h_user

seanh:
Move the `LTILaunchResource.h_user` (`context.h_user`) property, which
returns the `HUser` for the current request, to
`request.lti_user.h_user`:

1. Add a new factory classmethod `models.HUser.from_lti_user(lti_user)`
   that constructs an `HUser` from an `LTIUser`

2. As a convenience add a new instance property `models.LTIUser.h_user`
   which is the `HUser` for the given `LTIUser`

3. This means that instead of `context.lti_user` which only works when
   the context is an `LTILaunchResource`, code can now do
   `request.lti_user.h_user` instead, which works whenever
   `request.lti_user` isn't `None`. So code that uses the `h_user` is
   now decoupled from `LTILaunchResource` and can be used in other
   contexts.

4. Delete `LTILaunchResource.h_user` now that it's no longer used

### DIFF
--- a/lms/models/h_user.py
+++ b/lms/models/h_user.py
@@ -1,3 +1,4 @@
+import hashlib
 from typing import NamedTuple
 
 
@@ -28,3 +29,22 @@ class HUser(NamedTuple):
 
     def userid(self, authority):
         return f"acct:{self.username}@{authority}"
+
+    @classmethod
+    def from_lti_user(cls, lti_user):
+        provider = lti_user.tool_consumer_instance_guid
+        provider_unique_id = lti_user.user_id
+
+        def username():
+            """Return the h username for the current request."""
+            username_hash_object = hashlib.sha1()
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
+            return username_hash_object.hexdigest()[:30]
+
+        return cls(
+            username=username(),
+            display_name=lti_user.display_name,
+            provider=provider,
+            provider_unique_id=provider_unique_id,
+        )
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,5 +1,7 @@
 from typing import NamedTuple
 
+from lms.models import HUser
+
 
 class LTIUser(NamedTuple):
     """An LTI user."""
@@ -18,6 +20,11 @@ class LTIUser(NamedTuple):
 
     display_name: str
     """The user's display name."""
+
+    @property
+    def h_user(self):
+        """Return a models.HUser generated from this LTIUser."""
+        return HUser.from_lti_user(self)
 
     @property
     def is_instructor(self):
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -17,6 +17,7 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
         self._context = context
         self._request = request
         self._lti_user = request.lti_user
+        self._h_user = request.lti_user.h_user
         self._consumer_key = (
             request.lti_user.oauth_consumer_key if request.lti_user else None
         )
@@ -194,7 +195,7 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
 
         self._config["canvas"]["speedGrader"] = {
             "submissionParams": {
-                "h_username": self._context.h_user.username,
+                "h_username": self._h_user.username,
                 "lis_result_sourcedid": lis_result_sourcedid,
                 "lis_outcome_service_url": lis_outcome_service_url,
                 **kwargs,
@@ -203,9 +204,6 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
 
     def _auth_token(self):
         """Return the authToken setting."""
-        if not self._lti_user:
-            return None
-
         return BearerTokenSchema(self._request).authorization_param(self._lti_user)
 
     def _canvas_files_available(self):
@@ -262,7 +260,13 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                 "authUrl": self._request.route_url("canvas_api.authorize"),
             },
             # Some debug information, currently used in the Gherkin tests.
-            "debug": self._debug(),
+            "debug": {
+                "tags": [
+                    "role:instructor"
+                    if self._lti_user.is_instructor
+                    else "role:learner"
+                ],
+            },
             # Tell the JavaScript code whether we're in "dev" mode.
             "dev": self._request.registry.settings["dev"],
             # The config object for the Hypothesis client.
@@ -281,21 +285,6 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                 ],
             },
         }
-
-    def _debug(self):
-        """
-        Return some debug information.
-
-        Currently used in the Gherkin tests.
-        """
-        debug_info = {}
-
-        if self._lti_user:
-            debug_info["tags"] = [
-                "role:instructor" if self._lti_user.is_instructor else "role:learner"
-            ]
-
-        return debug_info
 
     def _get_students(self):
         """
@@ -330,7 +319,7 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
         claims = {
             "aud": urlparse(api_url).hostname,
             "iss": self._request.registry.settings["h_jwt_client_id"],
-            "sub": self._context.h_user.userid(self._authority),
+            "sub": self._h_user.userid(self._authority),
             "nbf": now,
             "exp": now + datetime.timedelta(minutes=5),
         }
--- a/tests/unit/lms/models/lti_user_test.py
+++ b/tests/unit/lms/models/lti_user_test.py
@@ -5,6 +5,14 @@ from tests import factories
 
 
 class TestLTIUser:
+    def test_h_user(self, HUser):
+        lti_user = factories.LTIUser()
+
+        h_user = lti_user.h_user
+
+        HUser.from_lti_user.assert_called_once_with(lti_user)
+        assert h_user == HUser.from_lti_user.return_value
+
     @pytest.mark.parametrize(
         "roles,is_instructor",
         [
@@ -68,3 +76,8 @@ def test_display_name(
     display_name_ = display_name(given_name, family_name, full_name)
 
     assert display_name_ == expected_display_name
+
+
+@pytest.fixture(autouse=True)
+def HUser(patch):
+    return patch("lms.models.lti_user.HUser")
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -4,7 +4,6 @@ import hashlib
 
 from pyramid.security import Allow
 
-from lms.models import HUser
 from lms.resources._js_config import JSConfig
 
 
@@ -26,30 +25,6 @@ class LTILaunchResource:
         """Return the context resource for an LTI launch request."""
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
-
-    @property
-    def h_user(self):
-        """Return the h user for the current request."""
-
-        # The h "provider" string for the current request.
-        provider = self._request.lti_user.tool_consumer_instance_guid
-
-        # The h "provider_unique_id" string for the current request.
-        provider_unique_id = self._request.lti_user.user_id
-
-        def username():
-            """Return the h username for the current request."""
-            username_hash_object = hashlib.sha1()
-            username_hash_object.update(provider.encode())
-            username_hash_object.update(provider_unique_id.encode())
-            return username_hash_object.hexdigest()[:30]
-
-        return HUser(
-            username=username(),
-            display_name=self._request.lti_user.display_name,
-            provider=provider,
-            provider_unique_id=provider_unique_id,
-        )
 
     @property
     def h_authority_provided_id(self):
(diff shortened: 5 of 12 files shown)


lms/models/h_user.py
```diff
@@ -28,3 +29,22 @@ class HUser(NamedTuple):
...
+            username_hash_object.update(provider.encode())
+            username_hash_object.update(provider_unique_id.encode())
+            return username_hash_object.hexdigest()[:30]
+
+        return cls(
+            username=username(),
+            display_name=lti_user.display_name,
+            provider=provider,
+            provider_unique_id=provider_unique_id,
+        )
```
seanh, line 50: The method for constructing `HUser`'s belongs on the `HUser` class, but I've also added `LTIUser.h_user` (below) as a convenience

lms/resources/_js_config.py
```diff
@@ -203,9 +204,6 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
 
     def _auth_token(self):
         """Return the authToken setting."""
-        if not self._lti_user:
-            return None
-
```
seanh, line 208: I don't think `LTILaunchResource` is ever used when `request.lti_user` is `None` so I've just simplified this

lms/resources/_js_config.py
```diff
@@ -282,21 +286,6 @@ def _config(self):
...
-        Currently used in the Gherkin tests.
-        """
-        debug_info = {}
-
-        if self._lti_user:
-            debug_info["tags"] = [
-                "role:instructor" if self._lti_user.is_instructor else "role:learner"
-            ]
-
-        return debug_info
```
seanh, line 298: I don't think `LTILaunchResource` is ever used when `request.lti_user` is `None` so I've just simplified this

lms/models/h_user.py
jon-betts: Getting some deja-vu here, but I think the inner method should just be a plain method like `lti_user.py/display_name()`.

Also, once this is true, this factory method (and the attendant boundary and tests) doesn't need to exist. The `display_name()` part probably belongs here, but the rest can be in the `h_user()` property on the `lti_user`, as it would just be a call to the initialiser, with a call to this function first.

lms/resources/_js_config.py
jon-betts: :+1:

lms/models/lti_user.py
```diff
@@ -19,6 +21,11 @@ class LTIUser(NamedTuple):
     display_name: str
     """The user's display name."""
 
+    @property
+    def h_user(self):
+        """Return a models.HUser generated from this LTIUser."""
+        return HUser.from_lti_user(self)
```
jon-betts, line 27: If the above `username()` function were public this would be something like:
```
return HUser(
    username=username(
        self.tool_consumer_instance_guid, self.lti_user.user_id
    ),
    display_name=self.display_name,
    provider=self.tool_consumer_instance_guid,
    provider_unique_id=self.lti_user.user_id,
)
```

tests/unit/lms/models/lti_user_test.py
```diff
@@ -68,3 +76,8 @@ def test_display_name(
     display_name_ = display_name(given_name, family_name, full_name)
 
     assert display_name_ == expected_display_name
+
+
+@pytest.fixture(autouse=True)
+def HUser(patch):
+    return patch("lms.models.lti_user.HUser")
```
jon-betts, line 83: This test would change fairly radically if the above suggestions were in place, but I think this might be a good time to allow a deeper kind of testing. We can probably just check that we get the HUser object we expect, without mocking here.

lms/models/lti_user.py
seanh: Hmm. I'm not sure I prefer this. I don't like it because:

1. It's dividing up the knowledge of how to construct an `HUser` from an `LTIUser` across two different places. There's a `models.h_user.h_username()` function that knows how to create the username from parts of an `LTIUser`, then `models.LTIUser.h_user`, in a different files, has the _rest_ of the knowledge about what parts of an `LTIUser` line up with what parts of the `HUser`.

2. It's encoding a bunch of h-specific knowledge (specifically: knowledge about how we create h users with valid `provider`'s and `provider_unique_id`'s etc) in the `LTIUser` class, which seems like the wrong place for it. Specifically, it means the `LTIUser` class knows that we use an `LTIUser`'s `tool_consumer_instance_guid` for the `HUser`'s `provider` and knows that we use an `LTIUser`'s `user_id` for the `HUser`s `provider_unique_id`. But `provider` and `provider_unique_id` are h-specific concepts that have nothing to do with LTI users. Shouldn't the `HUser` class be the class that knows about how to create a valid `HUser`?

   When all `LTIUser` has to do is `HUser.from_lti_user(self)`, then `LTIUser` is freed of any specific knowledge about what parts an `HUser` has or how they're constructed

```python
diff --git a/lms/models/h_user.py b/lms/models/h_user.py
index 5975aade..957316d4 100644
--- a/lms/models/h_user.py
+++ b/lms/models/h_user.py
@@ -30,21 +30,9 @@ class HUser(NamedTuple):
     def userid(self, authority):
         return f"acct:{self.username}@{authority}"
 
-    @classmethod
-    def from_lti_user(cls, lti_user):
-        provider = lti_user.tool_consumer_instance_guid
-        provider_unique_id = lti_user.user_id
-
-        def username():
-            """Return the h username for the current request."""
-            username_hash_object = hashlib.sha1()
-            username_hash_object.update(provider.encode())
-            username_hash_object.update(provider_unique_id.encode())
-            return username_hash_object.hexdigest()[:30]
-
-        return cls(
-            username=username(),
-            display_name=lti_user.display_name,
-            provider=provider,
-            provider_unique_id=provider_unique_id,
-        )
+
+def h_username(lti_user):
+    username_hash_object = hashlib.sha1()
+    username_hash_object.update(lti_user.tool_consumer_instance_guid.encode())
+    username_hash_object.update(lti_user.user_id.encode())
+    return username_hash_object.hexdigest()[:30]
diff --git a/lms/models/lti_user.py b/lms/models/lti_user.py
index 24ae9091..cbf0978c 100644
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,6 +1,6 @@
 from typing import NamedTuple
 
-from lms.models import HUser
+from lms.models.h_user import HUser, h_username
 
 
 class LTIUser(NamedTuple):
@@ -24,7 +24,12 @@ class LTIUser(NamedTuple):
     @property
     def h_user(self):
         """Return a models.HUser generated from this LTIUser."""
-        return HUser.from_lti_user(self)
+        return HUser(
+            username=h_username(self),
+            display_name=self.display_name,
+            provider=self.tool_consumer_instance_guid,
+            provider_unique_id=self.user_id,
+        )
 
     @property
     def is_instructor(self):
```

tests/unit/lms/models/lti_user_test.py
seanh: We actually didn't remove the mocks when we made the `display_name()` extraction with `LTIUser`, I just mocked `display_name()` instead:

https://github.com/hypothesis/lms/blob/b51a54d545e0dc1c7ef38a755e2778f9bf48c2a4/tests/unit/lms/validation/authentication/_launch_params_test.py#L11-L31

And I think for good reason. Take a look at how the `LTIUser.h_user` tests would look if they didn't mock `HUser` (if `username()` were extracted out of `HUser` and not mocked in `LTIUser.h_user`'s tests the results would be equivalent):

```diff
diff --git a/tests/unit/lms/models/lti_user_test.py b/tests/unit/lms/models/lti_user_test.py
index 2a60e4ce..f39a8054 100644
--- a/tests/unit/lms/models/lti_user_test.py
+++ b/tests/unit/lms/models/lti_user_test.py
-        HUser.from_lti_user.assert_called_once_with(lti_user)
-        assert h_user == HUser.from_lti_user.return_value
+        assert h_user == HUser(
+            username="2f42fed00de3039b0c720859d03b47",
+            display_name=lti_user.display_name,
+            provider=lti_user.tool_consumer_instance_guid,
+            provider_unique_id=lti_user.user_id,
+        )
```

The mock here isn't a problem in and of itself. For example, I don't think the tests without the mock are any deeper. Specifically, I don't think there's any way to break the code such that the with-mock tests would still be passing, whereas the without-mock tests would catch it. The with-mock tests provide complete coverage.

And the mock actually provides a bunch of advantages:

* The test "speaks" a lot more clearly about what `LTIUser.h_user` is actually responsible for doing, and what we're actually trying to test about it:

  ```python
   assert h_user == HUser.from_lti_user.return_value
   ```

   Compared to:

   ```python
   assert h_user == HUser(username="2f42fed00de3039b0c720859d03b47", ...)
   ```

   What the heck is this actually testing? What is `"2f42fed00de3039b0c720859d03b47"`? It's impossible to see from the test, but that opaque string is  SHA-1 hash of `tool_consumer_instance_guid` and `user_id`, which is an implementation detail of `HUser` that has nothing to do with `LTIUser` or to do with what this test is actually for.

* It also removes `display_name=lti_user.display_name`, `provider=lti_user.tool_consumer_instance_guid`, and `provider_unique_id=lti_user.user_id` from the `LTIUser` tests, all of which are `HUser` implementation details that have nothing to do with `LTIUser` or with what this test is supposed to be for

* It avoids duplicating tests, makes the tests less brittle, requires fewer test changes when you change the code. The without-mock version of the test above duplicates stuff that's already tested by `HUser.from_lti_user()`'s own unit tests. If you changed `HUser`'s internal implementation details, without the mock you'd have to update both `HUser` and `LTIUser`'s tests

* It means that `LTIUser`'s tests don't contribute to `HUser()`'s code coverage. `HUser` has its own unit tests and we want those to be responsible for its code coverage. We don't want misleadingly inflated coverage.

* It means that if there's a bug in `HUser.from_lti_user()` then `HUser.from_lti_user()`'s tests will be failing, but `LTIUser.h_user`'s tests will not be failing unless there's a bug in `LTIUser.h_user`. This is a good thing: a test should have only one reason to fail. This is easier to see if you imagine the pattern repeated again and again for the tests across an entire codebase. It's easier to pinpoint and debug test failures if only a few unit tests for the specific unit that has the bug are failing. If you frequently run the tests and get a lot of failing tests across a lot of different units, this wastes time trying to pinpoint where the bug is. Multiply the time lost again and again over the maintenance life of the codebase. Isolated tests scale better, they're easier to maintain.

seanh:
@jon-betts Can you take another look?

I'm not sure that I want to extract the classmethod into a standalone function, as this seems to put h-specific knowledge into the LTI user class. See comment.

I'm also not sure that I want to remove the mock. This is a pretty small-scale instance of isolated units, but as they are `models/h_user.py` and `models/lti_user.py` are two separate units as far as our isolated unit testing goes. In the unit tests for a given module we always mock other public modules that've been imported from elsewhere in our won code, and neither of `h_user.py` or `lti_user.py` is an internal close-collaborator for the other. And I think the tests with mocks are actually better than without. See comment.

seanh:
One nuance about the isolated unit testing is that model objects are usually an exception from our "you should mock any public names imported from elsewhere in our own code" rule. We don't usually mock SQLAlchemy models in the unit tests for the services that use them, for example. And this is also true of the non-DB models like `LTIUser` and `HUser`, dating back to when they were in `values.py` and called "value objects": part of the point of these objects is that they're so simple you don't need to bother mocking them.

Once a non-trivial method like `from_lti_user()` gets added to `HUser`, though, then I do find it beneficial to mock `HUser` in the `LTIUser` tests that're calling `HUser.from_lti_user()`. But I still wouldn't bother to mock `HUser` in other tests for code that is just using `HUser` as a value object and not calling `from_lti_user()`.

If the `from_lti_user()` method were not part of the `HUser` class it might make the to-mock-or-not-to-mock decision less nuanced, as you could simply say "don't mock models". Maybe it'd be a standalone `models.h_user.from_lti_user()` function.

On the other hand I do feel that factory functions are just screaming to be `@classmethod`'s on their respective classes

seanh:
Another approach here would be to simply say that `LTIUser` and `HUser` are close collaborators of each other and move them both into a single `models/user.py` file, and then apply your suggestion of extracting the `classmethod` into a separate `_h_username()` function that's called by `LTIUser.h_user`.

That would enable complete code coverage without either mocks or test duplication: everything about creating `HUser`'s from `LTIUser`'s would be covered by `LTIUser.h_user`'s tests, `HUser` would just be a dumb value object. And it eases the objection about `LTIUser` containing h-specific knowledge, because we've kind of said that it's all one unit by running `LTIUser` and `HUser` together into a single module anyway.

Something like:

```python
class HUser(NamedTuple):
    username: str
    display_name: str = ""
    provider: str = ""
    provider_unique_id: str = ""

    def userid(self, authority):
        return f"acct:{self.username}@{authority}"


class LTIUser(NamedTuple):
    user_id: str
    oauth_consumer_key: str
    roles: str
    display_name: str

    @property
    def h_user(self):
        return HUser(
            username=_h_username(self),
            display_name=self.display_name,
            provider=self.tool_consumer_instance_guid,
            provider_unique_id=self.user_id,
        )

    @property
    def is_instructor(self):
        ...

    @property
    def is_learner(self):
        ...


def display_name(given_name, family_name, full_name):
    ...


def _h_username(lti_user):
    ...
```

On the other hand, this `user.py` module would now contain `LTIUser`, `HUser`, `display_name()` and `_h_username()`, and would be getting pretty long, over 100 lines with all the docstrings and details filled out. I think I might prefer to smaller isolated modules, than one big one.

seanh:
I think I'd vote for leaving this PR as it currently is. The alternatives that I can think of have their pros and cons, but none of them seems a slam dunk

PR #1669, opened 2020-04-21
Use the BulkAPI in LTI-H and H-API

jon-betts:
This isn't ready to rock yet. I still need to:

 * Test it practically
 * Chop it up a bit more sensibly probably


jon-betts:
This is dead. The code here has been chopped into smaller PRs and superseded.

PR #1671, opened 2020-04-21
Add GradingInfo.__repr__()

seanh:
Add a `__repr__()` to `GradingInfo`. I found this useful when debugging some test stuff.

Might be a nicer way of implementing it using [`reprlib`](https://github.com/hypothesis/lms/pull/new/add-GradingInfo-repr)?

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -58,6 +58,15 @@ class BaseClass:
             if key in data:
                 setattr(self, key, data[key])
 
+    def __repr__(self):
+        return "{class_}({kwargs})".format(
+            class_=self.__class__.__name__,
+            kwargs=", ".join(
+                f"{kwarg}={repr(getattr(self, kwarg))}"
+                for kwarg in self.__table__.columns.keys()  # pylint:disable=no-member
+            ),
+        )
+
 
 BASE = declarative_base(
     # Create a default metadata object with naming conventions for indexes and
--- a/tests/unit/lms/db/__init___test.py
+++ b/tests/unit/lms/db/__init___test.py
@@ -52,6 +52,22 @@ class TestBase:
         with pytest.raises(TypeError):
             ModelClass().update_from_dict({}, skip_keys=["a"])
 
+    def test_repr(self):
+        model = ModelClass(id=23, column=46)
+
+        assert repr(model) == ("ModelClass(" "id=23, " "column=46)")
+
+    def test_repr_is_valid_python(self):
+        model = ModelClass(id=23, column=46)
+
+        deserialized_model = eval(repr(model))  # pylint:disable=eval-used
+
+        for attr in (
+            "id",
+            "column",
+        ):
+            assert getattr(deserialized_model, attr) == getattr(model, attr)
+
     @pytest.fixture
     def model(self):
         return ModelClass(id=1234, column="original_value")


lms/models/grading_info.py
```diff
@@ -69,3 +69,11 @@ class GradingInfo(BASE):  # pylint:disable=too-few-public-methods
...
     h_username = sa.Column(sa.UnicodeText(), nullable=False)
     h_display_name = sa.Column(sa.UnicodeText(), nullable=False)
+
+    def __repr__(self):
+        return "GradingInfo({kwargs})".format(
+            kwargs=", ".join(
+                f"{kwarg}={repr(getattr(self, kwarg))}"
+                for kwarg in self.__table__.columns.keys()
+            )
+        )
```
jon-betts, line 79: Could this go in `lms.db.BaseClass` and benefit all of our models? It could easily not be grading info specific by using `self.__class__.__name__` in place of `GradingInfo`.

There's also `iter_columns` in there which looks like it might be helpful for this? Maybe not. If we had an `as_dict()` which returned the `kwargs`, and then `__repr__` used that it would be pretty neat.

tests/unit/lms/models/grading_info_test.py
```diff
@@ -57,6 +57,44 @@ def test_it_enforces_uniqueness_constraint(
...
+            "user_id",
+            "context_id",
+            "resource_link_id",
+            "tool_consumer_info_product_family_code",
+            "h_username",
+            "h_display_name",
+        ):
+            assert getattr(deserialized_grading_info, attr) == getattr(
+                grading_info, attr
+            )
```
jon-betts, line 96: Did you find this property handy during debugging? I feel a bit strange about writing a test to maintain a property we don't make use of. I think this property is generally broken enough in the wild that it's not really a thing.

These tests would probably disappear to be replaced with something shorter in `BaseClass` anyway if you liked that idea, so maybe it doesn't really matter either way.

seanh: The fact that it's valid Python? No that's not really useful but I just think it's the right style

seanh: (See the Python docs: [<q>If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value</q>](https://docs.python.org/3/reference/datamodel.html?highlight=__str__#object.__repr__). It's supposed to be <q>information-rich and unambiguous</q> for debugging purposes and a Python expression is just a good style for achieving that. The fact that you could literally copy-paste-execute it in the debugger to create another instance, might come in handy sometimes but isn't really the point.)

lms/models/grading_info.py
seanh: Yep, moved this onto the base class

tests/unit/lms/models/grading_info_test.py
jon-betts: I know, I was talking about the tests not the style. In general we shouldn't write tests to preserve properties and behavior we don't use. It's pain if you make a change which you know is fine but the tests fail anyway because they fix some prorperty the product doesn't actually rely on.

But as I said, it's so much shorter when done generally that I don't think it matters here, and if you have to have coverage of `__repr__` in your tests, unless it only states it returns a string (might be a good idea actually), it's going to be sensitive to the contents anyway.

seanh:
Rebased

PR #1674, opened 2020-04-21
Merge h_authority_provided_id, h_groupid and h_group_name into h_group

seanh:
Add a new `models.HGroup` domain model class with
`authority_provided_id`, `groupid` and `group_name` attributes and
combine the three separate `LTILaunchResource` properties
`h_authority_provided_id`, `h_groupid` and `h_group_name` into a single
`h_group` property that returns an `HGroup`.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -363,7 +363,7 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                     "authority": self._authority,
                     "enableShareLinks": False,
                     "grantToken": self._grant_token(api_url),
-                    "groups": [self._context.h_groupid],
+                    "groups": [self._context.h_group.groupid(self._authority)],
                 }
             ]
         }
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -4,6 +4,7 @@ import hashlib
 
 from pyramid.security import Allow
 
+from lms.models import HGroup
 from lms.resources._js_config import JSConfig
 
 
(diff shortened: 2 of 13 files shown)


lms/services/h_api.py
```diff
@@ -114,16 +111,18 @@ def do_upsert_group():
...
         :param h_user: the user to add to the group
         :type h_user: HUser
-        :param group_id: The id of the group
+        :param h_group: the group to add the user to
+        :type h_group: HGroup
         """
         self._api_request(
-            "POST", f"groups/{group_id}/members/{h_user.userid(self._authority)}"
+            "POST",
+            f"groups/{h_group.groupid(self._authority)}/members/{h_user.userid(self._authority)}",
```
jon-betts, line 125: This looks outwardly daft as we just pluck out things we could be given, but the interface is really clean as a result. Shame we're about to delete it :).

lms/services/lti_h.py
```diff
@@ -51,7 +43,7 @@ def sync(self, groups):
         This will upsert the provided list of groups, the current user and
         make that user a member of each group.
 
-        :param groups: A list of Group objects.
+        :param groups: A list of models.Group objects.
```
jon-betts, line 46:
```suggestion
        :param groups: A list of models.HGroup objects.
```

lms/services/lti_h.py
```diff
@@ -68,13 +60,13 @@ def _sync_to_h(self, groups):
...
             self._group_info_service.upsert(
-                authority_provided_id=self._context.h_authority_provided_id,
+                authority_provided_id=self._context.h_group.authority_provided_id,
                 consumer_key=self._lti_user.oauth_consumer_key,
                 params=self._request.params,
             )
 
         for group in groups:
-            self._h_api.add_user_to_group(h_user=self._h_user, group_id=group.groupid)
+            self._h_api.add_user_to_group(self._h_user, group)
```
jon-betts, line 72: This is going to clash like a doozy in a mo.

lms/resources/_js_config.py
```diff
@@ -363,7 +363,7 @@ def _hypothesis_client(self):
                     "authority": self._authority,
                     "enableShareLinks": False,
                     "grantToken": self._grant_token(api_url),
-                    "groups": [self._context.h_groupid],
+                    "groups": [self._context.h_group.groupid(self._authority)],
```
jon-betts, line 366: :+1:

lms/resources/lti_launch.py
```diff
@@ -49,21 +59,12 @@ def h_authority_provided_id(self):
...
-        The groupid is deterministic and is unique to the LTI course. Calling this
-        function again with params representing the same LTI course will always
-        return the same groupid. Calling this function with different params will
-        always return a different groupid.
-        """
-        return f"group:{self.h_authority_provided_id}@{self._authority}"
+        return HGroup(
+            name=self._group_name(self._request.parsed_params["context_title"].strip()),
+            authority_provided_id=authority_provided_id,
+        )
```
jon-betts, line 67: I can't really see how this is all going to come together with the various methods below. This also seems like a different pattern for creating `HGroups`, from both `LTIUsers` and `HUsers`.

I'm going to assume this is half way through a chain of stuff and doesn't make much sense now.

jon-betts:
Weird. I seem to have lost a message. I think github is playing up... which is a shame. But basically if the `HGroup` looked like this:

```python
class HGroup:
    # The maximum length of an h group name.
    group_name_max_length = 25

    def __init__(self, name, authority_provided_id):
        self.name = self._clean_name(name)
        self.authority_provided_id

    @classmethod
    def _clean_name(cls, name):
        """Return an h group name from the given course or section name."""

        name = name.strip()

        if len(name) > cls.group_name_max_length:
            name = name[: cls.group_name_max_length - 1].rstrip() + "…"

        return name
```        

Then it would tuck away the logic into the object. It also means it's _not possible_ to create a group with an invalid name. This is the difference between a POJO which doesn't form a border around it's state and an active model.

This would also simplify the callers by having them just pass in the string they want (we'd trim and fix it)

lms/resources/lti_launch.py
seanh: Haven't really considered it yet at all

PR #1676, opened 2020-04-21
Remove single_group_sync() and decouple LTIHService from LTILaunchResource (context)

seanh:
This finishes decoupling `LTIHService` from `LTILaunchResource` (`self._context`) and means that `LTIHService` should now be callable by the sync API:

Remove `LTIHService.single_group_sync()` which was just a convenience method for doing `self.sync([self._context.h_group])`. Just have the views call `lti_h_svc.sync([context.h_group])` instead. The sync API view will call `lti_h_svc.sync([group_created_in_some_other_way])`

### DIFF
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -3,7 +3,7 @@ from pyramid.httpexceptions import HTTPInternalServerError
 from lms.services import HAPIError
 
 
-class LTIHService:
+class LTIHService:  # pylint:disable=too-few-public-methods
     """
     Copy LTI users and courses to h users and groups.
 
@@ -18,7 +18,6 @@ class LTIHService:
     """
 
     def __init__(self, _context, request):
-        self._context = request.context
         self._request = request
         self._lti_user = request.lti_user
         self._h_user = request.lti_user.h_user
@@ -26,15 +25,6 @@ class LTIHService:
         self._ai_getter = request.find_service(name="ai_getter")
         self._h_api = request.find_service(name="h_api")
         self._group_info_service = request.find_service(name="group_info")
-
-    def single_group_sync(self):
-        """
-        Sync standard data to H for an LTI launch.
-
-        This will read a single group from the context object, upsert it, the
-        current user and make that user a member of the group.
-        """
-        self.sync(groups=[self._context.h_group])
 
     def sync(self, groups):
         """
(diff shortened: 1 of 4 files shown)


lms/services/lti_h.py
```diff
@@ -3,7 +3,7 @@
 from lms.services import HAPIError
 
 
-class LTIHService:
+class LTIHService:  # pylint:disable=too-few-public-methods
```
seanh, line 6: I think we may want to globally disable this pylint warning

lms/services/lti_h.py
```diff
@@ -18,7 +18,6 @@ class LTIHService:
     """
 
     def __init__(self, _context, request):
-        self._context = request.context
```
seanh, line 21: Yay!

lms/services/lti_h.py
jon-betts: Yeah I agree. It's clearly part of our style

lms/services/lti_h.py
```diff
@@ -63,7 +53,7 @@ def _sync_to_h(self, groups):
             self._h_api.upsert_group(group)
 
             self._group_info_service.upsert(
-                authority_provided_id=self._context.h_group.authority_provided_id,
+                authority_provided_id=group.authority_provided_id,
```
jon-betts, line 56: Oops. I spotted that one too in the bulk one.

seanh, line 56: Yikes. This looks like it was wrong before: sending the same `authority_provided_id` for every group! But it worked because of course there is only one group and it's always `self._context.h_group`. It wouldn't work with section groups now. Anyway, this fixes the latent bug

lms/views/basic_lti_launch.py
```diff
@@ -45,7 +45,7 @@ def sync_lti_data_to_h(self):
         and group corresponding to the LTI user and course.
         """
 
-        self.request.find_service(name="lti_h").single_group_sync()
+        self.request.find_service(name="lti_h").sync([self.context.h_group])
```
seanh, line 48: This makes sense. `context` is a web-layer thing. It's right for views to access `context`. `context` is not a services-layer thing, services shouldn't be accessing it themselves.

jon-betts: Apart from anything else, it kills off a fairly content free method.

Issue #1566, opened 2020-03-18
Add sync API settings to the JavaScript


seanh:
Add a `config.api.sync` object to the JavaScript config containing the URL, method and body for the sync API request that the JS code is going to have to make in future. See https://github.com/hypothesis/lms/issues/1435#issuecomment-598738736 for deets

Issue #1389, opened 2020-01-30
Add a Canvas->h sync API


seanh:
Depends on:

- [ ] Making various LTILaunchResource properties available to API requests: https://github.com/hypothesis/lms/issues/1430

See also:

* Separate card for changing the frontend to actually call this API: https://github.com/hypothesis/lms/issues/1390
* Separate card for _removing_ the synchronous Canvas->h syncing, once it's no longer needed: https://github.com/hypothesis/lms/issues/1391

Add a Canvas->sync API (name tbd) to the backend, for the frontend to call.

Currently we have an [`LTIHService`](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/services/lti_h.py#L26-L123) with methods for upserting users into h, upserting course groups, and adding users to groups (all by calling the h API). This service is called synchronously during LTI launches. With Canvas sections it's no longer going to be possible to sync data to h synchronously during launches because part of the data that we need to sync--the IDs and names of the h groups that we need to sync--has to be generated from Canvas sections that we have to retrieve asynchronously from the Canvas API.

So we need to add a backend API, for our frontend code to call asynchronously, that does the LTI->h sync (calls all three of `LTIHService`'s methods).

This card is just about adding the sync API. It's not about changing any code to actually call the sync API yet. Code will still be calling `LTIHService` synchronously for now.

## Params

The [params that the proof of concept sent to the sync API](https://github.com/hypothesis/lms/blob/76cc3a103fa8082e2803fa82931cd1fd795cb5d5/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js#L105-L115) were:

* tool_consumer_instance_guid
* lis_person_name_full
* lis_person_name_full
* lis_person_name_given
* lis_person_name_family
* context_id
* context_title
* custom_canvas_course_id
* student_custom_canvas_user_id
* student_lti_user_id

This is in addition to the bearer token in the `Authorization` header, which deserializes into an [`LTIUser`](https://github.com/hypothesis/lms/blob/7e1b85ff0c15690673e3dd471f5cf890b465a5da/lms/values.py#L7-L25) object: `user_id` (the LTI user ID), `oauth_consumer_key` and `roles`.

The long flat list of params should probably be grouped into compound objects, maybe something like:

```
{
  "lms": {
    "tool_consumer_instance_guid": "..."
  },
  "user": {
    "lis_person_name_full": "...",
[...]

PR #1680, opened 2020-04-23
Add feature-flagged sync API

seanh:
Add the new sync API and, if the `section_groups` feature flag is on, add the JavaScript config necessary to tell the frontend to call the sync API.

Fixes https://github.com/hypothesis/lms/issues/1566
Fixes https://github.com/hypothesis/lms/issues/1389

This PR does not include the frontend changes necessary to make it respond to the new config and actually call the sync API.

## What this PR does

1. Adds a `context.should_use_section_groups` boolean. For now, sections groups are enabled if you're in Canvas and you have the `section_groups` feature flag on.

1. Adds the new sync API: route, ACL and view.

   When called, the sync API fetches the sections from Canvas, generates groups from them, upserts the groups into h, and returns the list of groupids.

1. Adds new sync API settings to the JavaScript config.

   This tells the JS what path to call the sync API on, and what params to pass to it. The JS is dumb about calling the sync API: it just passes it the params the backend told it to pass. (This JS code that dumbly calls the sync API is _not in this PR_.)

   This new sync API config is only added to the JS config when sections are in use. Otherwise it's omitted (and the JS will just use the course group from the JS config and will not call the sync API).

1. When section groups are in use the `"groups"` in the JavaScript config object that we send **to the Hypothesis client** will be the string `"$rpc:requestGroups"` instead of the list of group IDs. This is what we have to do to tell the client to fetch the list of group IDs over postMessage-JSON-RPC.

   Note: this is the config for the Hypothesis client, that will be sent over postMessage-JSON-RPC to the Hypothesis client, it's _not_ the config for the LMS app's own frontend code.

   When section groups are not in use the `"groups"` config is what it was before: a list of one course group ID.

## How to test this PR

You can't really test this PR without the frontend changes that make it actually call the sync API, and pass the groups from the sync API to the client. And that frontend code isn't ready yet. But there is a hacky version of that frontend PR: https://github.com/hypothesis/lms/pull/1681

So the easiest way to test this PR is to cherry-pick #1681 onto this branch locally (`git cherry-pick make-the-frontend-call-the-sync-api`). That PR will both enable the `section_groups` feature flag by default in dev and change the frontend code to use the section groups. You can then launch assignments as teachers and students and see section groups in the client.

There are https://hypothesis.instructure.com/ test users on 1Password named **Sections 101 Teacher**, **Sections 101 Student 1**, **Sections 101 Student 2** and **Sections 101 Student 3** (the different students are members of different sections of the course). If you log in to Canvas as one of these users you'll see a test course named **Sections 101** that contains pre-created test sections.

- [ ] When launching an assignment as a teacher you should see all of the course's sections in the client's groups menu

- [ ] When launching an assignment as a student you should only see that student's sections, which may only be one section (or it might be more than one if the student belongs to multiple sections)

- [ ] If you delete all of the Canvas API access tokens from your LMS DB (`make sql` and then `delete from oauth2_token;`) and then launch any assignment, you should see that it asks you to authorize with Canvas and _then_ launches and shows the section groups.

  Previously this Canvas authorization would only have happened when launching a Canvas files assignment because only Canvas files assignments had to call the Canvas API. Now it'll happen with any type of assignment, since they all need to call the Canvas API to get the sections.

- [ ] In dev tools you should be able to see the sync API request and response containing the section group IDs

- [ ] You could also look at the groups being created in h's DB with `make sql` in h

- [ ] If you set `feature_flags.section_groups` to `false` in `development.ini` you should then see course groups again not section groups, and no sync API request in dev tools

## Known issues

1. When more than one section group is shown in the client, the groups menu includes buttons for leaving the groups: https://hypothes-is.slack.com/archives/C1M8NH76X/p1587752985223300 These buttons shouldn't be shown in the LMS context. We'll fix this in a separate PR

2. This will be totally broken in SpeedGrader. We'll fix that in later PRs

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -250,6 +250,7 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                 # The auth token that the JavaScript code will use to
                 # authenticate itself to the API.
                 "authToken": self._auth_token(),
+                "sync": self._sync_api(),
             },
             # The URL that the JavaScript code will open if it needs the user to
             # authorize us to request a new Canvas access token.
@@ -363,7 +364,33 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                     "authority": self._authority,
                     "enableShareLinks": False,
                     "grantToken": self._grant_token(api_url),
-                    "groups": [self._context.h_group.groupid(self._authority)],
+                    "groups": self._groups(),
                 }
             ]
+        }
+
+    def _groups(self):
+        if self._context.should_use_section_groups:
+            return "$rpc:requestGroups"
+        return [self._context.h_group.groupid(self._authority)]
+
+    def _sync_api(self):
+        if not self._context.should_use_section_groups:
+            return None
+
+        req = self._request
+
+        return {
+            "path": req.route_path("canvas_api.sync"),
+            "data": {
+                "lms": {
+                    "tool_consumer_instance_guid": req.params[
+                        "tool_consumer_instance_guid"
+                    ],
+                },
+                "course": {
+                    "context_id": req.params["context_id"],
+                    "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+                },
+            },
         }
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -66,43 +66,6 @@ class LTILaunchResource:
             authority_provided_id=authority_provided_id,
         )
 
-    def h_section_groupid(self, tool_consumer_instance_guid, context_id, section):
-        """
-        Return a unique h groupid for a given Canvas course section.
-
-        The groupid is deterministic and is unique to the course section.
-        Calling this function again with params representing the same course
-        section will always return the same groupid. Calling this function with
-        different params will always return a different groupid.
-
-        :param tool_consumer_instance_guid: the tool_consumer_instance_guid LTI
-            launch param from the Canvas instance that the section belongs to.
-            This is a unique identifier for the Canvas instance
-        :type tool_consumer_instance_guid: str
-
-        :param context_id: the context_id LTI launch param from the Canvas
-            course that the section belongs to. This is a unique identifier for
-            the course within the Canvas instance
-        :type context_id: str
-
-        :param section: a section dict as received from the Canvas API
-        :type section: dict
-        """
-        hash_object = hashlib.sha1()
-        hash_object.update(tool_consumer_instance_guid.encode())
-        hash_object.update(context_id.encode())
-        hash_object.update(section["id"].encode())
-        return f"group:section-{hash_object.hexdigest()}@{self._authority}"
-
-    @staticmethod
-    def h_section_group_name(section):
-        """
-        Return the h group name for the given Canvas course section.
-
-        :param section: a section dict as received from the Canvas API
-        """
-        return h_group_name(section["name"])
-
     @property
     def is_canvas(self):
         """Return True if Canvas is the LMS that launched us."""
@@ -134,3 +97,11 @@ class LTILaunchResource:
         Canvas.
         """
         return self._request.parsed_params.get("custom_canvas_api_domain")
+
+    @property
+    def should_use_section_groups(self):
+        """Return True if section groups rather than a course group should be used."""
+        if not self.is_canvas:
+            return False
+
+        return self._request.feature("section_groups")
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -30,7 +30,8 @@ def includeme(config):
         "canvas_api.courses.files.list", "/api/canvas/courses/{course_id}/files"
     )
     config.add_route("canvas_api.files.via_url", "/api/canvas/files/{file_id}/via_url")
-    config.add_route("lti_api.submissions.record", "/api/lti/submissions")
+    config.add_route("canvas_api.sync", "/api/canvas/sync", request_method="POST")
 
+    config.add_route("lti_api.submissions.record", "/api/lti/submissions")
     config.add_route("lti_api.result.read", "/api/lti/result", request_method="GET")
     config.add_route("lti_api.result.record", "/api/lti/result", request_method="POST")
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -102,11 +102,6 @@ class TestEnableContentItemSelectionMode:
         assert not js_config.asdict()["filePicker"]["canvas"]["enabled"]
         assert "courseId" not in js_config.asdict()
 
-    @pytest.fixture
-    def pyramid_request(self, pyramid_request):
-        pyramid_request.params["custom_canvas_course_id"] = "test_course_id"
-        return pyramid_request
-
 
 class TestAddCanvasFileID:
     """Unit tests for JSConfig.add_canvas_file_id()."""
@@ -266,7 +261,6 @@ class TestMaybeEnableGrading:
     @pytest.fixture
     def pyramid_request(self, pyramid_request):
         pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Instructor")
-        pyramid_request.params["context_id"] = "test_course_id"
         pyramid_request.params["context_title"] = "test_course_name"
         pyramid_request.params["resource_link_title"] = "test_assignment_name"
         return pyramid_request
@@ -330,6 +324,44 @@ class TestJSConfigAuthToken:
         return config["api"]["authToken"]
 
 
+class TestJSConfigAPISync:
+    """Unit tests for the api.sync sub-dict of JSConfig."""
+
+    @pytest.mark.usefixtures("section_groups_on")
+    def test_it(self, sync):
+        assert sync == {
+            "path": "/api/canvas/sync",
+            "data": {
+                "course": {
+                    "context_id": "test_context_id",
+                    "custom_canvas_course_id": "test_custom_canvas_course_id",
+                },
+                "lms": {
+                    "tool_consumer_instance_guid": "test_tool_consumer_instance_guid"
+                },
+            },
+        }
+
+    def test_its_None_if_section_groups_isnt_enabled(self, sync):
+        assert sync is None
+
+    @pytest.fixture
+    def sync(self, config):
+        return config["api"]["sync"]
+
+    @pytest.fixture
+    def pyramid_request(self, pyramid_request):
+        pyramid_request.params.clear()
+        pyramid_request.params.update(
+            {
+                "context_id": "test_context_id",
+                "custom_canvas_course_id": "test_custom_canvas_course_id",
+                "tool_consumer_instance_guid": "test_tool_consumer_instance_guid",
+            }
+        )
+        return pyramid_request
+
+
 class TestJSConfigDebug:
     """Unit tests for the "debug" sub-dict of JSConfig."""
 
@@ -381,6 +413,10 @@ class TestJSConfigHypothesisClient:
         groups = config["services"][0]["groups"]
 
         assert groups == [context.h_group.groupid.return_value]
+
+    @pytest.mark.usefixtures("section_groups_on")
+    def test_it_configures_the_client_to_fetch_the_groups_over_RPC(self, config):
+        assert config["services"][0]["groups"] == "$rpc:requestGroups"
 
     @pytest.mark.usefixtures("provisioning_disabled")
     def test_it_is_empty_if_provisioning_feature_is_disabled(self, config):
@@ -443,15 +479,25 @@ def context():
         instance=True,
         h_group=mock.create_autospec(HGroup, instance=True, spec_set=True),
         is_canvas=True,
+        should_use_section_groups=False,
     )
 
 
 @pytest.fixture
+def section_groups_on(context):
+    context.should_use_section_groups = True
+
+
+@pytest.fixture
 def pyramid_request(pyramid_request):
-    pyramid_request.params["lis_result_sourcedid"] = "example_lis_result_sourcedid"
-    pyramid_request.params[
-        "lis_outcome_service_url"
-    ] = "example_lis_outcome_service_url"
+    pyramid_request.params.update(
+        {
+            "lis_result_sourcedid": "example_lis_result_sourcedid",
+            "lis_outcome_service_url": "example_lis_outcome_service_url",
+            "context_id": "test_course_id",
+            "custom_canvas_course_id": "test_course_id",
+        }
+    )
     return pyramid_request
 
 
--- a/tests/unit/lms/resources/lti_launch_test.py
+++ b/tests/unit/lms/resources/lti_launch_test.py
@@ -41,63 +41,6 @@ class TestHGroup:
         )
 
 
-class TestHSectionGroupID:
-    @pytest.mark.parametrize(
-        "settings,args,expected_groupid",
-        (
-            (
-                {},
-                {},
-                "group:section-298e68dd4befff42f9a9ff7edffc542b2ba1f782@TEST_AUTHORITY",
-            ),
-            (
-                {},
-                {
-                    "tool_consumer_instance_guid": "DIFFERENT_tool_consumer_instance_guid"
-                },
-                "group:section-e690bef466ef1d42587ae88b6b76f553d5e718e7@TEST_AUTHORITY",
-            ),
-            (
-                {},
-                {"context_id": "DIFFERENT_context_id"},
-                "group:section-1ca5d17b0eab0e02784964c21aca33d99167402f@TEST_AUTHORITY",
-            ),
-            (
-                {},
-                {"section": {"id": "DIFFERENT_section_id"}},
-                "group:section-15348ff2dbeb6e250d029c363007b8357bde7eea@TEST_AUTHORITY",
-            ),
-            (
-                {"h_authority": "DIFFERENT_authority"},
-                {},
-                "group:section-298e68dd4befff42f9a9ff7edffc542b2ba1f782@DIFFERENT_authority",
-            ),
-        ),
-    )
-    def test_it(
-        self, settings, args, expected_groupid, pyramid_request,
-    ):
-        pyramid_request.registry.settings.update(**settings)
-        lti_launch_resource = LTILaunchResource(pyramid_request)
-        args.setdefault("tool_consumer_instance_guid", "tool_consumer_instance_guid")
-        args.setdefault("context_id", "context_id")
-        args.setdefault("section", {"id": "section_id"})
-
-        groupid = lti_launch_resource.h_section_groupid(**args)
-
-        assert groupid == expected_groupid
-
-
-class TestHSectionGroupName:
-    def test_it(self, lti_launch, h_group_name):
-        # A section dict as received from the Canvas API (except this one only
-        # has the keys that we actually use).
-        section = {"name": "test_section_name"}
-
-        assert lti_launch.h_section_group_name(section) == h_group_name.return_value
-        h_group_name.assert_called_once_with("test_section_name")
-
-
 class TestIsCanvas:
     @pytest.mark.parametrize(
         "parsed_params,is_canvas",
@@ -156,6 +99,30 @@ class TestJSConfig:
 
         JSConfig.assert_called_once_with(lti_launch, pyramid_request)
         assert js_config == JSConfig.return_value
+
+
+class TestShouldUseSectionGroups:
+    @pytest.mark.parametrize(
+        "is_canvas,is_feature_flag_enabled,should_use_section_groups",
+        [
+            pytest.param(True, True, True, id="Canvas with feature flag on"),
+            pytest.param(True, False, False, id="Canvas with feature flag off"),
+            pytest.param(False, True, False, id="Non-Canvas with feature flag on",),
+            pytest.param(False, True, False, id="Non-Canvas with feature flag off"),
+        ],
+    )
+    def test_it(
+        self,
+        lti_launch,
+        pyramid_request,
+        is_canvas,
+        is_feature_flag_enabled,
+        should_use_section_groups,
+    ):
+        pyramid_request.feature.return_value = is_feature_flag_enabled
+
+        with mock.patch.object(LTILaunchResource, "is_canvas", is_canvas):
+            assert lti_launch.should_use_section_groups == should_use_section_groups
 
 
 @pytest.fixture
--- a/lms/resources/default.py
+++ b/lms/resources/default.py
@@ -9,6 +9,7 @@ class DefaultResource:  # pylint:disable=too-few-public-methods
         (Allow, "report_viewers", "view"),
         (Allow, "lti_user", "canvas_api"),
         (Allow, "lti_user", "lti_outcomes"),
+        (Allow, "lti_user", "sync_api"),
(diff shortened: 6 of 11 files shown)


lms/resources/_js_config.py
```diff
@@ -367,3 +368,24 @@ def _hypothesis_client(self):
                 }
             ]
         }
+
+    def _sync_api(self):
+        if not self._context.should_use_section_groups:
+            return None
```
seanh, line 374: It's gonna be up to the backend to decide when the sync API should be called and when not. The backend will either put the sync API settings in the JS config or not, and the frontend will just call the sync API if the settings are present

lms/resources/lti_launch.py
```diff
@@ -62,57 +62,10 @@ def h_group(self):
...
-        hash_object.update(section["id"].encode())
-        return f"group:section-{hash_object.hexdigest()}@{self._authority}"
-
-    def h_section_group_name(self, section):
-        """
-        Return the h group name for the given Canvas course section.
-
-        :param section: a section dict as received from the Canvas API
-        """
-        return self._group_name(section["name"].strip())
```
seanh, line 103: These methods for generating an `HGroup` from a Canvas section just got inlined directly into the sync API view function itself, which is the only place they're needed

lms/resources/lti_launch.py
```diff
@@ -62,57 +62,10 @@ def h_group(self):
...
-    @staticmethod
-    def _group_name(name):
-        """Return an h group name from the given course or section name."""
-        # The maximum length of an h group name.
-        group_name_max_length = 25
-
-        if len(name) > group_name_max_length:
-            name = name[: group_name_max_length - 1].rstrip() + "…"
-
-        return name
```
seanh, line 114: This got moved into `models.h_group`

lms/resources/lti_launch.py
```diff
@@ -144,3 +97,11 @@ def custom_canvas_api_domain(self):
...
         """
         return self._request.parsed_params.get("custom_canvas_api_domain")
+
+    @property
+    def should_use_section_groups(self):
+        """Return True if section groups rather than a course group should be used."""
+        if not self.is_canvas:
+            return False
+
+        return self._request.feature("section_groups")
```
seanh, line 107: The question of whether or not section groups should be used is going to have to become a lot more complicated before we can release this. For now you just have to be in Canvas and have the feature flag on, which will allow us to deploy and test intermediate versions of the sync API

lms/views/api/sync.py
```diff
@@ -0,0 +1,35 @@
...
+        hash_object.update(tool_consumer_instance_guid.encode())
+        hash_object.update(context_id.encode())
+        hash_object.update(str(section["id"]).encode())
+        authority_provided_id = hash_object.hexdigest()
+
+        groups.append(HGroup(HGroup.group_name(section["name"]), authority_provided_id))
+
+    lti_h_svc.sync(groups)
+
+    return groups
```
seanh, line 35: :man_shrugging: Aside from the 5 lines for generating `authority_provided_id` this view is just plugging together a bunch of already existing stuff:

* The sync API params that the frontend posted, as placed in the JavaScript config by `JSConfig`

* The Canvas API client's methods for getting sections from Canvas. There are actually three different methods for getting the sections in different ways under different circumstances, and this view is going to have to grow some logic for choosing between them. For now it's just using the all-of-the-courses-sections method all the time

* The `LTIHService`'s `sync()` method

* It then returns the groups to the frontend, which will need to pass them to the client

So far no logic in the view, just one loop. It will grow a little logic for selecting the right Canvas API client method to call, but I'm expecting it to remain simple enough for the code to get away with just all living in the view directly

seanh: We should probably add a validation schema to the view for good measure, though it's not like the frontend is ever going to be calling it invalidly anyway

lms/views/api/sync.py
```diff
@@ -0,0 +1,35 @@
...
+@view_config(
+    route_name="api.sync", request_method="POST", renderer="json", permission="sync_api"
+)
+def sync(request):
+    canvas_api_svc = request.find_service(name="canvas_api_client")
+    lti_h_svc = request.find_service(name="lti_h")
+
+    context_id = request.json["course"]["context_id"]
+    custom_canvas_course_id = request.json["course"]["custom_canvas_course_id"]
+    tool_consumer_instance_guid = request.json["lms"]["tool_consumer_instance_guid"]
```
seanh, line 17: We should probably add a validation schema to the view for good measure, rather than just assuming all the params are present and correct, but it's not like the frontend is ever going to be calling it invalidly anyway

tests/unit/lms/views/api/sync_test.py
```diff
@@ -0,0 +1,47 @@
...
+            authority_provided_id="d0f36006728f2277f228b74c8a7f620305bfb3e7",
+        ),
+        HGroup(
+            name="Section Name 2",
+            authority_provided_id="d99674b9700f4a40a2b301d2949b61339c58236c",
+        ),
+    ]
+
+    assert sync(pyramid_request) == expected_groups
+    lti_h_service.sync.assert_called_once_with(expected_groups)
```
seanh, line 24: Since the sync API view doesn't contain any logic, there's only one test needed

lms/views/api/sync.py
seanh: There are arguably missing domain model objects here: `LMS` and `LTICourse`. A validation schema could even create them for the view so that `request.parsed_params["course"]` would be a `models.LTICourse` object and so on. But I'm not sure it's really worthwhile introducing them at this point.

lms/resources/_js_config.py
```diff
@@ -363,7 +364,33 @@ def _hypothesis_client(self):
...
-                    "groups": [self._context.h_group.groupid(self._authority)],
+                    "groups": self._groups(),
                 }
             ]
         }
+
+    def _groups(self):
+        if self._context.should_use_section_groups:
+            return "$rpc:requestGroups"
+        return [self._context.h_group.groupid(self._authority)]
```
seanh, line 375: This goes into **the Hypothesis client's config object**, not the config for the LMS app's own frontend code. We either send the client the list of one course group's `groupid` as before or, if section groups are in use, we send the client the special string `"$rpc:requestGroups"` which tells it to request the groups over postMessage-JSON-RPC.

lms/resources/_js_config.py
```diff
@@ -363,7 +364,33 @@ def _hypothesis_client(self):
...
+                "lms": {
+                    "tool_consumer_instance_guid": req.params[
+                        "tool_consumer_instance_guid"
+                    ],
+                },
+                "course": {
+                    "context_id": req.params["context_id"],
+                    "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+                },
+            },
```
seanh, line 395: These are the params that the sync API needs, in its request body

lms/resources/lti_launch.py
```diff
@@ -66,43 +66,6 @@ def h_group(self):
...
-        return f"group:section-{hash_object.hexdigest()}@{self._authority}"
-
-    @staticmethod
-    def h_section_group_name(section):
-        """
-        Return the h group name for the given Canvas course section.
-
-        :param section: a section dict as received from the Canvas API
-        """
-        return h_group_name(section["name"])
```
seanh, line 104: These two methods just ended up getting inlined into the sync API view itself, below

lms/views/api/sync.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if request.lti_user.is_learner:
+        # For learners we only want the client to show the sections that the
+        # student belongs to, so fetch only the user's sections.
+        sections = canvas_api_svc.authenticated_users_sections(custom_canvas_course_id)
+    else:
+        # For non-learners (e.g. instructors, teaching assistants) we want the
+        # client to show all of the course's sections.
+        sections = canvas_api_svc.course_sections(custom_canvas_course_id)
```
seanh, line 27: SpeedGrader will be a third case here, but we're not supporting that yet

jon-betts:
I've just gone through this from a manual testing point of view first:

* When there's only one section no drop down - Tick
* Auth popup when not authed - Tick
* No auth pop when authed - Tick
* Students only see their own sections - Tick
* Teachers see all sections - Tick
* Student annotations are stored - Tick
* Students see their own annotations - Tick
* Teachers see student annotations - Per section yes. You have to swap between, but that's probably a good thing. This is more of a grading concern than review one?
* Check sync request made - Tick
* Check feature flag - Tick (sections annotations appear missing)

The only oddity I could see is a violation of my expectations around the "Post to Section x" button for submitting annotations. I expected the drop down to include other sections, but it only has a private annotation option. This is just the way it works though.

seanh:
> The only oddity I could see is a violation of my expectations around the "Post to Section x" button for submitting annotations. I expected the drop down to include other sections, but it only has a private annotation option. This is just the way it works though.

Yeah that's by design. It's the cause of quite a few UX problems in the client, and showing all available groups on the <kbd>Post</kbd> button's dropdown menu has long been a suggested solution. There are even designs. See:

* https://trello.com/c/B7e15ukT/40-simplified-scope-and-post-controls
* https://docs.google.com/document/d/1jJrMa5rLCZZX3FEzn1b6VTBjPHi83MDtO3oskUBvMAI/edit#heading=h.gvse2blu4nei

tests/unit/lms/views/api/sync_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+    canvas_api_client_method = getattr(canvas_api_client, canvas_api_client_method_name)
+    canvas_api_client_method.assert_called_once_with("test_custom_canvas_course_id")
+
+    lti_h_service.sync.assert_called_once_with(expected_groups)
+
+    authority = pyramid_request.registry.settings["h_authority"]
+    assert returned_groups == [
+        f"group:{group.authority_provided_id}@{authority}" for group in expected_groups
+    ]
```
jon-betts, line 68: I think this would be better as two tests instead of a parameterised one.

This would remove:

 * The `is_learner` path which is specific to one case on either side
 * The `canvas_api_client_method getattr` stuff for a direct assertion
 * The parameterisation and param id

Is this also an opportunity for factory use to collapse the large `HGroup` declarations?

lms/views/api/sync.py
jon-betts: Very clear code here. The combination of properties and well encapsulated and named functions makes this very easy to understand.

lms/views/api/sync.py
```diff
@@ -0,0 +1,43 @@
...
+
+    groups = []
+
+    for section in sections:
+        # Generate a unique authority_provided_id for the section group.
+        hash_object = hashlib.sha1()
+        hash_object.update(tool_consumer_instance_guid.encode())
+        hash_object.update(context_id.encode())
+        hash_object.update(str(section["id"]).encode())
+        authority_provided_id = hash_object.hexdigest()
```
jon-betts, line 37: This sticks out as kind of obscure compared to the simplicity of the rest. I think readability would be improved by popping this out to a separate function.

On a related note, we've had a rash of functions very like this. I wonder if we are missing something in models like:

```python
def hashed_id(*parts):
    hash_object = hashlib.sha1()
    
    for part in parts:
        hash_object.update(str(part).encode())

    return hash_object.hexdigest()
    

# ... meanwhile

from models.? import hashed_id


def _section_id(instance_guid, context_id, section):
    # At this point should we even bother?
    return hashed_id(instance_guid, context_id, section['id'])
```

lms/resources/_js_config.py
seanh: Because of the way [the frontend's API client](https://github.com/hypothesis/lms/blob/master/lms/static/scripts/frontend_apps/utils/api.js) works, it has to be in this format: it doesn't accept a `method` argument but instead [sets the method depending on whether there's a body or not](https://github.com/hypothesis/lms/blob/809b24db432e8bd61ebd294b06e7c5372b5e0456/lms/static/scripts/frontend_apps/utils/api.js#L59) and it [assumes the body is an object not a string](https://github.com/hypothesis/lms/blob/809b24db432e8bd61ebd294b06e7c5372b5e0456/lms/static/scripts/frontend_apps/utils/api.js#L54)

seanh: I think there's some work to be done on the frontend here:

* Not all API requests work by the backend passing all the request details in the JS config like this. In most cases the frontend constructs the URL and params itself. I think they should all be changed to work this way: the backend passing the request deets and the frontend just being dumb

* The backend could also stringify the body and set the method and the headers. The frontend could be dumb about all that too

Should result in overall simpler frontend code.

For another time maybe

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -330,6 +324,44 @@ def authToken(self, config):
...
+    @pytest.fixture
+    def pyramid_request(self, pyramid_request):
+        pyramid_request.params.clear()
+        pyramid_request.params["context_id"] = "test_context_id"
+        pyramid_request.params[
+            "custom_canvas_course_id"
+        ] = "test_custom_canvas_course_id"
+        pyramid_request.params[
+            "tool_consumer_instance_guid"
+        ] = "test_tool_consumer_instance_guid"
```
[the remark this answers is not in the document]
seanh: You can use `update`, done

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -158,6 +101,60 @@ def test_it_returns_the_js_config(self, pyramid_request, JSConfig):
...
+            ),
+            pytest.param(
+                False,
+                True,
+                False,
+                id="If we're not in Canvas and the section_groups feature "
+                "flag is off then should_use_section_groups is False",
+            ),
+        ],
+    )
```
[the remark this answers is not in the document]
seanh: Unfortunately you can't use keyword arguments with `pytest.param`, which would help a lot. That seems like the kind of thing that might be quite easy to do in pytest, might be successful as a feature request to pytest. For now I've rewritten it so that they each format onto one line as you suggested

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -158,6 +101,60 @@ def test_it_returns_the_js_config(self, pyramid_request, JSConfig):
...
+        should_use_section_groups,
+    ):
+        if feature_flag:
+            pyramid_request.feature.side_effect = (
+                lambda feature: feature == "section_groups"
+            )
+        else:
+            pyramid_request.feature.side_effect = (
+                lambda feature: feature != "section_groups"
+            )
```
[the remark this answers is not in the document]
seanh: That doesn't work!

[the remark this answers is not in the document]
seanh: Went for this instead:

```python
        def feature(feature_flag):
            if feature_flag == "section_groups":
                return is_feature_flag_enabled
            return mock.DEFAULT

        pyramid_request.feature.side_effect = feature
```

tests/unit/lms/views/api/sync_test.py
seanh: Split this into two separate tests with an assertion helper.

I wasn't using factories here because the tests actually do pass all two kwargs to `HGroup` and actually do depend on the values of both kwargs. But a factory is still potentially beneficial because, if a new required kwarg is added to `HGroup` in future that these tests don't care about, if they're using a factory the tests won't need to be updated. So I've changed it to a factory as well

lms/resources/_js_config.py
jon-betts: Yeah, I didn't know if this was using a new pathway written for this, but if it's existing stuff, so be it. I do think it would be nice in future, but it's not exactly burning the house down.

tests/unit/lms/resources/lti_launch_test.py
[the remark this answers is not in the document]
jon-betts: What's the boo boo?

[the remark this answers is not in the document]
seanh: PyTest ended up pointing out that the `if` above is always taken, so I ended up simplifying this further:

```python
pyramid_request.feature.return_value = is_feature_flag_enabled
```

I think if we want a test mock that _only_ enables the `section_groups` feature flag rather than all feature flags, we'll end up with what I originally had. It's really not necessary though: just setting the boolean return value is fine

lms/views/api/sync.py
```diff
@@ -0,0 +1,43 @@
...
+@view_config(
+    route_name="api.sync", request_method="POST", renderer="json", permission="sync_api"
+)
+def sync(request):
+    authority = request.registry.settings["h_authority"]
+    canvas_api_svc = request.find_service(name="canvas_api_client")
+    lti_h_svc = request.find_service(name="lti_h")
+
+    context_id = request.json["course"]["context_id"]
+    custom_canvas_course_id = request.json["course"]["custom_canvas_course_id"]
```
jon-betts, line 17: With this guy, was "custom_" added to distinguish it from standard LTI parameters, or does Canvas actually send us a request with "custom_canvas_course_id"?

I was just asking, because I wonder:

* If it's for transport only whether "canvas_course_id" would be the better name in the code
* ... or it's better to have the two match, even if it doesn't read as nice?

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -443,15 +479,23 @@ def context():
...
+
+
 @pytest.fixture
 def pyramid_request(pyramid_request):
     pyramid_request.params["lis_result_sourcedid"] = "example_lis_result_sourcedid"
     pyramid_request.params[
         "lis_outcome_service_url"
     ] = "example_lis_outcome_service_url"
+    pyramid_request.params["context_id"] = "test_course_id"
+    pyramid_request.params["custom_canvas_course_id"] = "test_course_id"
```
jon-betts, line 498: Maybe another opportunity to update here.

jon-betts, line 498:
```suggestion
    pyramid_request.params.update(
        {
            "lis_result_sourcedid": "example_lis_result_sourcedid",
            "lis_outcome_service_url": "example_lis_outcome_service_url",
            "context_id": "test_course_id",
            "custom_canvas_course_id": "test_course_id",
        }
    )
```

Also is the one above duplicating "context_id" and "custom_canvas_course_id" or does it not inherit from this?

lms/views/api/sync.py
seanh: I think generating the `HGroup` for a section is quite specific to this view (it depends on the view's `tool_consumer_instance_guid` and `context_id` params, as well as on the section dicts retrieved from `canvas_api_svc`) so I don't think it bears extracting out into a service or anything. So I think we're just talking about a collaborator function for the view, as in your example.

I hate to say it but I think an inner function is nicer here due to the ability to use closures so you don't have to pass long arguments. Here it is with an outer helper function:

```python
    groups = [
        _group(tool_consumer_instance_guid, context_id, section) for section in sections
    ]

    lti_h_svc.sync(groups)

    return [group.groupid(authority) for group in groups]


def _group(tool_consumer_instance_guid, context_id, section):
    hash_object = hashlib.sha1()
    hash_object.update(tool_consumer_instance_guid.encode())
    hash_object.update(context_id.encode())
    hash_object.update(str(section["id"]).encode())
    authority_provided_id = hash_object.hexdigest()

    return HGroup(h_group_name(section["name"]), authority_provided_id,)

```

And here's the inner-function version, which I think reads nicer:

```python
    def authority_provided_id(section):
        hash_object = hashlib.sha1()
        hash_object.update(tool_consumer_instance_guid.encode())
        hash_object.update(context_id.encode())
        hash_object.update(str(section["id"]).encode())
        return hash_object.hexdigest()

    groups = [
        HGroup(h_group_name(section["name"]), authority_provided_id(section),)
        for section in sections
    ]

    lti_h_svc.sync(groups)

    return [group.groupid(authority) for group in groups]
```

seanh: I think a `models.hashed_id()` function might be a good idea but I'd like to leave that for a separate PR

tests/unit/lms/resources/lti_launch_test.py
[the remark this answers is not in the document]
jon-betts: That's the Gordian knot solution right there. Honestly it's undefined whether we should be returning true for false for the rest anyway.

tests/unit/lms/resources/_js_config_test.py
seanh: :+1: The one above clears and then re-sets the params for clarity/independence

jon-betts:
Just went through this properly again. Can't really see anything more except possibly a tiny bit of duplication in `pyramid_request` fixture in `_js_config_test.py`, but this could be for clarity.

As far as I can tell a fixture and the fixture it inherits from both set `context_id` and `custom_canvas_course_id`.

tests/unit/lms/resources/_js_config_test.py
jon-betts: Righto

lms/views/api/sync.py
seanh: https://github.com/hypothesis/lms/pull/1680/commits/0529e0a550d086b45b9a57a270235ed640bfc68a

tests/unit/lms/views/api/sync_test.py
seanh: https://github.com/hypothesis/lms/pull/1680/commits/35dd24f0de7e482f9d613a7f240cf09871d6bca3

tests/unit/lms/resources/lti_launch_test.py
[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/1680/commits/42a52bd474a2079a2593e5ced6195c67874ebaa5

tests/unit/lms/resources/lti_launch_test.py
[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/1680/commits/ff5647662013f052a5ea7a3fad269ae9f726929c

lms/views/api/sync.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pyramid.view import view_config
+
+from lms.models import HGroup, h_group_name
+
+
+@view_config(
+    route_name="api.sync", request_method="POST", renderer="json", permission="sync_api"
+)
+def sync(request):
```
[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/1680/commits/00dd58287a341e5bf4e980d6822a515e61443c22

tests/unit/lms/resources/_js_config_test.py
seanh: https://github.com/hypothesis/lms/pull/1680/commits/2cef2fcab4fe143404c2936265bc7579864f4a64

seanh:
Reminder to self: change the path to /api/canvas/sync

lms/routes.py
```diff
@@ -34,3 +34,5 @@ def includeme(config):
 
     config.add_route("lti_api.result.read", "/api/lti/result", request_method="GET")
     config.add_route("lti_api.result.record", "/api/lti/result", request_method="POST")
+
+    config.add_route("api.sync", "/api/sync", request_method="POST")
```
jon-betts, line 38: Maybe "/api/canvas/sync" and "api.canvas.sync" here?

PR #1683, opened 2020-04-24
Extract group_name() into model

seanh:
Extract the `_group_name()` helper method out of `LTILaunchResource` into the domain model. This reduces some test duplication, and also makes it so that future code outside of `LTILaunchResource` can call `models.h_group_name()` without duplicating any code.

Same approach as with `display_name()` for `models.LTIUser`

### DIFF
--- a/tests/unit/lms/resources/lti_launch_test.py
+++ b/tests/unit/lms/resources/lti_launch_test.py
@@ -3,7 +3,6 @@ from unittest import mock
 import pytest
 from pyramid.authorization import ACLAuthorizationPolicy
 
-from lms.models import HGroup
 from lms.resources import LTILaunchResource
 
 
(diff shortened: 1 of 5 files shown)


tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -27,25 +26,18 @@ def test_it_allows_the_correct_users_to_launch_LTI_assignments(
...
         }
 
-        assert lti_launch.h_group == HGroup(
-            expected_group_name, "d55a3c86dd79d390ec8dc6a8096d0943044ea268",
+        assert lti_launch.h_group == HGroup.return_value
+        h_group_name.assert_called_once_with("test_context_title")
+        HGroup.assert_called_once_with(
+            name=h_group_name.return_value,
+            authority_provided_id="d55a3c86dd79d390ec8dc6a8096d0943044ea268",
         )
```
jon-betts, line 41: Caused some good separation of concerns here. Just FYI, I think for small pure pythons stuff like this I'd be equally happy to black box test it too.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -97,22 +89,13 @@ def test_it(
...
-    @pytest.mark.parametrize(
-        "section_name,group_name",
-        [
-            ("Test Section", "Test Section"),
-            (" Test Section ", "Test Section"),
-            ("Test   Section", "Test   Section"),
-            ("Object Oriented Polymorphism 101", "Object Oriented Polymorp…"),
-            ("  Object Oriented Polymorphism 101  ", "Object Oriented Polymorp…"),
-        ],
-    )
```
jon-betts, line 109: Removal of duplication too!

PR #1685, opened 2020-04-24
Skip Via 3 routing for Google Drive documents

jon-betts:
Got a little fancy here and introduced a simple model class here to help re-use some of the logic. The idea is to make the central bit nice and declarative (if is_pdf do this).

The document object also paves the way for us to specify that something is a PDF for Canvas docs (in the next PR)

### What's happened here?

The logic has been split into some model object so represent:

 * The document 
 * Via itself, making URLs based the document

The document contains a "content_type" which you can set, but if you don't it will use the same regex we have in Via 3 to guess that the doc is a PDF when we see a Google Drive doc.

The objects are tested as blackbox private components of `via_url`.

### How to test

This is mostly about observing what URLs are constructed when we look at various docs by LMS. So here are some URLs to follow on the local Canvas instance, along with with the expected result. In each case filter by "via.client" in the dev tools:

Run "make dev" in superdev, then visit:

* https://hypothesis.instructure.com/courses/83/assignments/462 - Google drive localhost assignment
   - You should see a single call to "pdf?url=" showing Via 3 handling the PDF viewing
   - You should _not_ see a call to "route" showing we have skipped the routing
* https://hypothesis.instructure.com/courses/83/assignments/460 - HTML assignment
   - You should see "route?url=" showing Via 3 routing is in place
   - You should see "example.com?..." showing original flavour Via is handling the HTML
* https://hypothesis.instructure.com/courses/83/assignments/461 - PDF assignment
   - You should see a call to "route?url=" and "pdf?url=" showing Via 3 routing and viewing is in place

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -1,7 +1,81 @@
 """Via-related view helpers."""
+import re
 from urllib.parse import parse_qsl, urlencode, urlparse
 
 __all__ = ["via_url"]
+
+# pylint: disable=too-few-public-methods
+
+
+class _ViaDoc:
+    """A doc we want to proxy with content type."""
+
+    GOOGLE_DRIVE_REGEX = re.compile(
+        r"^https://drive.google.com/uc\?id=(.*)&export=download$", re.IGNORECASE
+    )
+
+    def __init__(self, url, content_type=None):
+        self.url = url
+
+        if content_type is None and self.GOOGLE_DRIVE_REGEX.match(url):
+            content_type = "pdf"
+
+        self._content_type = content_type
+
+    @property
+    def is_pdf(self):
+        return self._content_type == "pdf"
+
+    @property
+    def is_html(self):
+        return self._content_type == "html"
+
+
+class _ViaClient:
+    """A small wrapper to make calling Via easier."""
+
+    def __init__(self, service_url, legacy_service_url, host_url, legacy_mode=False):
+        self.service_url = urlparse(service_url)
+        self.legacy_service_url = legacy_service_url
+
+        # Default via parameters
+        self.options = {
+            "via.client.openSidebar": "1",
+            "via.client.requestConfigFromFrame.origin": host_url,
+            "via.client.requestConfigFromFrame.ancestorLevel": "2",
+        }
+        self.legacy_mode = legacy_mode
+
+    def url_for(self, doc):
+        if not isinstance(doc, _ViaDoc):
+            raise TypeError(f"Expected doc to be a ViaDoc, not '{type(doc)}'")
+
+        if self.legacy_mode or doc.is_html:
+            return self._legacy_via_url(doc.url)
+
+        if doc.is_pdf:
+            return self._url_for("/pdf", doc.url)
+
+        return self._url_for("/route", doc.url)
+
+    def _url_for(self, path, doc_url):
+        options = {"url": doc_url}
+        options.update(self.options)
+
+        return self.service_url._replace(path=path, query=urlencode(options)).geturl()
+
+    def _legacy_via_url(self, doc_url):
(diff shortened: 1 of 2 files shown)


lms/views/helpers/_via.py
```diff
@@ -1,8 +1,82 @@
...
+
+        self._content_type = content_type
+
+    @property
+    def is_pdf(self):
+        return self._content_type == "pdf"
+
+    @property
+    def is_html(self):
+        return self._content_type == "html"
```
jon-betts, line 31: I know it's a bit _extra_, but go with me. I wanted the core function below to be very declarative.

The idea is we can also specify the content type if we know it when we create this object and the right thing (tm) should happen as a result.

lms/views/helpers/_via.py
```diff
@@ -1,8 +1,82 @@
...
+        if not isinstance(doc, _ViaDoc):
+            raise TypeError(f"Expected doc to be a ViaDoc, not '{type(doc)}'")
+
+        if self.legacy_mode or doc.is_html:
+            return self._legacy_via_url(doc.url)
+
+        if doc.is_pdf:
+            return self._url_for("/pdf", doc.url)
+
+        return self._url_for("/route", doc.url)
```
jon-betts, line 59: This is the core logic that was getting a bit convoluted, so I wanted properties to make it readable. We're not testing this directly (so we might get some coverage issues I'll have to fix).

This is all (so far) a private collaborator for `via_url()` below.

lms/views/helpers/_via.py
```diff
@@ -1,8 +1,82 @@
...
+
+    def _url_for(self, path, doc_url):
+        options = {"url": doc_url}
+        options.update(self.options)
+
+        return self.service_url._replace(path=path, query=urlencode(options)).geturl()
```
jon-betts, line 65: This is actually the second change here (that isn't stylistic) as we needed to be able to construct both `/pdf` and `/route` URLs here.

lms/views/helpers/_via.py
```diff
@@ -18,28 +92,9 @@ def via_url(request, document_url):
...
-    params = [kv for kv in parse_qsl(parsed_url.query) if not kv[0].startswith("via.")]
-    params.extend(options.items())
-
-    return via_service_url + parsed_url._replace(query=urlencode(params)).geturl()
+    return _ViaClient(
+        service_url=request.registry.settings["via_url"],
+        legacy_service_url=request.registry.settings["legacy_via_url"],
+        host_url=request.host_url,
+        legacy_mode=request.feature("use_legacy_via"),
+    ).url_for(_ViaDoc(document_url))
```
jon-betts, line 100: This is the opportunity to construct the doc with `content_type` and have that be passed in to `via_url()`. Hopefully we can then set that when using a Canvas doc or the like.

lms/views/helpers/_via.py
jon-betts: Note also, that `is_pdf() != not is_html()`. If we have a `content_type` of `None` we don't know what it is, and will have to route to find out.

lms/views/helpers/_via.py
```diff
@@ -1,8 +1,81 @@
...
+        self.options = {
+            "via.client.openSidebar": "1",
+            "via.client.requestConfigFromFrame.origin": host_url,
+            "via.client.requestConfigFromFrame.ancestorLevel": "2",
+        }
+        self.legacy_mode = legacy_mode
+
+    def url_for(self, doc):
+        assert isinstance(doc, _ViaDoc), "Provided doc is a ViaDoc"
+
```
seanh, line 51:
```suggestion
```

I don't think we should bother with these kinds of checks in Python, just rely on Duck Typing (i.e. don't bother)

Issue #1701, opened 2020-04-29
Surface teacher names for LMS groups


klemay:
Previously our Education team was was using the creator of LMS groups to associate teacher names with LMS groups. In the course of recent Sections work, we've changed our group creator from "the first teacher to launch Hypothesis in an LMS course site" to a generic LMS user. 

We've discussed adding the names of group members with instructor roles to a table in the LMS database.

judell:
From Slack:

> seanh
> @judell I'm wondering whether the group_info table is really being used to track courses, and just relying on the fact that it's one h group per course?
> With sections we'll be creating one h group for each section of the course, so if we keep the group_info table as it is you'll get multiple groups per course in there, for Canvas courses that have multiple sections. Is that going to work? 

> judell
> Everything that depends on https://jonudell.info/h/analytics/doc/views.html#lms_course_groups should be fine, but naturally we'll want to be able to provide a higher-level whole-course view.
> So, perhaps an identifier for a course, and course_info table keyed on that id that bundles sets of groups? (edited)

> seanh
> Re the LMS group teacher names: will adding these to the group_info table in the LMS DB (or more likely a related table in the LMS DB) work?

> judell
> Although I had proposed group_info in #1684, a course_info table in the LMS DB seems better in light of the above.

PR #1703, opened 2020-04-30
Add instructor info to group_info table

seanh:
Fixes https://github.com/hypothesis/lms/issues/1701

Depends on https://github.com/hypothesis/lms/pull/1704

Record info about instructors who launch our app in the `group_info`
table:

* Add a new JSON column `group_info.info`.

  So far this is a JSON object with a single top-level key "instructors"
  that contains a list of instructor dicts.

  `group_info.info` can be extended with more info (besides instructors)
  in future without a DB migration.

  The instructor dicts can also be extended to hold more field in future
  without a DB migration.

* Add a `GroupInfo.instructors` read-write property for convenience, and
  a `GroupInfo.upsert_instructor(dict)` method the creates-or-updates an
  instructor dict in a `GroupInfo`'s instructors list.

* When an instructor launches an assignment, have `GroupInfoService`
  upsert their user details into `GroupInfo.info["instructors"]` for
  each of the assignment's course or section groups

### DIFF
--- a/lms/models/group_info.py
+++ b/lms/models/group_info.py
@@ -1,4 +1,6 @@
 import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
 
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
 
     #: The value of the custom_canvas_course_id param this group was last launched with.
     custom_canvas_course_id = sa.Column(sa.UnicodeText())
+
+    #: A dict of info about this group.
+    info = sa.Column(MutableDict.as_mutable(JSONB))
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("info", {})
+        kwargs["info"].setdefault("instructors", [])
+        super().__init__(*args, **kwargs)
+
+    @property
+    def instructors(self):
+        return self.info["instructors"]
+
+    @instructors.setter
+    def instructors(self, new_instructors):
+        self.info["instructors"] = new_instructors
+
+    def upsert_instructor(self, new_instructor):
+        updated_instructors = []
+        found = False
+
+        for existing_instructor in self.instructors:
+            if existing_instructor["username"] == new_instructor["username"]:
+                updated_instructors.append(new_instructor)
+                found = True
+            else:
+                updated_instructors.append(existing_instructor)
+
+        if not found:
+            updated_instructors.append(new_instructor)
+
+        if updated_instructors != self.instructors:
+            self.instructors = updated_instructors
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -17,6 +17,7 @@ class GroupInfoService:  # pylint:disable=too-few-public-methods
 
     def __init__(self, _context, request):
         self._db = request.db
+        self._lti_user = request.lti_user
 
     def upsert(self, authority_provided_id, consumer_key, params):
         """
@@ -52,4 +53,9 @@ class GroupInfoService:  # pylint:disable=too-few-public-methods
             self._db.add(group_info)
 
         group_info.consumer_key = consumer_key
-        group_info.update_from_dict(params, skip_keys={"authority_provided_id", "id"})
+        group_info.update_from_dict(
+            params, skip_keys={"authority_provided_id", "id", "info"}
+        )
+
+        if self._lti_user.is_instructor:
+            group_info.upsert_instructor(self._lti_user.h_user._asdict())
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -58,4 +58,6 @@ class GroupInfoService:  # pylint:disable=too-few-public-methods
         )
 
         if self._lti_user.is_instructor:
-            group_info.upsert_instructor(self._lti_user.h_user._asdict())
+            group_info.upsert_instructor(
+                dict(email=self._lti_user.email, **self._lti_user.h_user._asdict())
+            )
--- a/tests/unit/lms/models/group_info_test.py
+++ b/tests/unit/lms/models/group_info_test.py
@@ -1,3 +1,4 @@
+import factory
 import pytest
 from sqlalchemy.exc import IntegrityError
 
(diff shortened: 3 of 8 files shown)


lms/models/group_info.py
```diff
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
...
+        kwargs["info"].setdefault("instructors", [])
+        super().__init__(*args, **kwargs)
+
+    @property
+    def instructors(self):
+        return self.info["instructors"]
+
+    @instructors.setter
+    def instructors(self, new_instructors):
+        self.info["instructors"] = new_instructors
```
seanh, line 98: This is just a convenience property, makes the method below read nicer

lms/models/group_info.py
```diff
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
...
+    def upsert_instructor(self, new_instructor):
+        updated_instructors = []
+        found = False
+
+        for existing_instructor in self.instructors:
+            if existing_instructor["username"] == new_instructor["username"]:
+                updated_instructors.append(new_instructor)
+                found = True
+            else:
+                updated_instructors.append(existing_instructor)
```
seanh, line 109: The loop is done this way so that `self.instructors` remains in order of first seen

lms/models/group_info.py
```diff
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
...
+
+        if not found:
+            updated_instructors.append(new_instructor)
+
+        if updated_instructors != self.instructors:
+            self.instructors = updated_instructors
```
seanh, line 115: The `if` is to avoid making SQLAlchemy think that `self.instructors` has changed when it actually hasn't. I'm not sure whether it's really needed or not

lms/models/group_info.py
```diff
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
 
     #: The value of the custom_canvas_course_id param this group was last launched with.
     custom_canvas_course_id = sa.Column(sa.UnicodeText())
+
+    #: A dict of info about this group.
+    info = sa.Column(MutableDict.as_mutable(JSONB))
```
seanh, line 85: I tried but wasn't able to write a unit test that fails without the `MutableDict` thing here. But without it (with just `sa.Column(MutableDict.as_mutable(JSONB))`) I found that, if `GroupInfo.info` already contained one instructor, then if a second instructor launched an assignment they would _not_ be added in to `GroupInfo.info`. The Python would _would_ add them to the instructors list, but it wouldn't get saved to the DB.

This is a hazard with using mutable types like JSON with SQLAlchemy. See **Detecting Changes in JSON columns when using the ORM** on this page: <https://docs.sqlalchemy.org/en/13/core/type_basics.html>. We've had to do the same fix in a couple of places in h, see: https://github.com/hypothesis/h/pull/3417

seanh:
I've just pushed a second commit to this that adds recording of email addresses plus some missing tests

tests/unit/lms/models/group_info_test.py
```diff
@@ -43,6 +44,59 @@ def test_application_instance_cant_be_None(self, db_session, group_info):
...
+            existing_instructors[2],
+        ]
+
+    def test_upsert_instructor_when_existing_equal_instructor(self):
+        group_info = GroupInfo()
+        existing_instructors = group_info.instructors = [
+            factories.HUser()._asdict(),
+            factories.HUser()._asdict(),
+            factories.HUser()._asdict(),
+        ]
```
jon-betts, line 94: I can't remember what it is, but there's a way of doing 3 in a batch isn't there?

lms/models/group_info.py
```diff
@@ -78,3 +80,36 @@ class GroupInfo(BASE):  # pylint:disable=too-few-public-methods
...
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("info", {})
+        kwargs["info"].setdefault("instructors", [])
+        super().__init__(*args, **kwargs)
+
+    @property
+    def instructors(self):
+        return self.info["instructors"]
```
jon-betts, line 94: So by the looks of it, to interact with a JSON field you just use it like data structure. That's pretty neat.

lms/services/group_info.py
```diff
@@ -52,4 +53,9 @@ def upsert(self, authority_provided_id, consumer_key, params):
             self._db.add(group_info)
 
         group_info.consumer_key = consumer_key
-        group_info.update_from_dict(params, skip_keys={"authority_provided_id", "id"})
+        group_info.update_from_dict(
+            params, skip_keys={"authority_provided_id", "id", "info"}
+        )
+
+        if self._lti_user.is_instructor:
+            group_info.upsert_instructor(self._lti_user.h_user._asdict())
```
jon-betts: Obvious spot to add students too if they turn out to be required.

tests/unit/lms/models/group_info_test.py
```diff
@@ -43,6 +44,59 @@ def test_application_instance_cant_be_None(self, db_session, group_info):
...
+        assert group_info.instructors == existing_instructors + [new_instructor]
+
+    def test_upsert_instructor_when_existing_matching_instructor(self):
+        group_info = GroupInfo()
+        updated_instructor = factories.HUser()._asdict()
+        existing_instructors = group_info.instructors = [
+            factories.HUser()._asdict(),
+            factories.HUser(username=updated_instructor["username"])._asdict(),
+            factories.HUser()._asdict(),
+        ]
```
jon-betts, line 78: I wonder if we could flip this one around and have all three of these test functions have accept "group_with_instructors" which has three instructors set, then make our `updated_instructor` read it's username from the instructor in position 1?

tests/unit/lms/services/group_info_test.py
```diff
@@ -83,7 +83,9 @@ def group_info_svc(self, pyramid_request):
     @pytest.fixture
     def params(self):
         return {
-            field.key: f"TEST_{field.key.upper()}" for field in GroupInfo.iter_columns()
+            field.key: f"TEST_{field.key.upper()}"
+            for field in GroupInfo.iter_columns()
+            if field.key not in ["info"]
```
jon-betts, line 121: Fun facts, last time I checked, using a list is more efficient than a set when the number of items is very low, as the hashing costs exceed the benefits when searching.

This would probably be fine as a "!=" though, which would be faster than both.

lms/models/group_info.py
seanh: Yeah, you just get a Python dict or list (I didn't check exactly what type of object it really is)

tests/unit/lms/models/group_info_test.py
seanh: It's complicated by the fact that I need dicts (I was calling `_asdict()` but:

```diff
diff --git a/tests/unit/lms/models/group_info_test.py b/tests/unit/lms/models/group_info_test.py
index d062cc42..149d49db 100644
--- a/tests/unit/lms/models/group_info_test.py
+++ b/tests/unit/lms/models/group_info_test.py
@@ -1,3 +1,4 @@
+import factory
 import pytest
 from sqlalchemy.exc import IntegrityError
 
@@ -87,11 +88,9 @@ class TestGroupInfo:
 
     def test_upsert_instructor_when_existing_equal_instructor(self):
         group_info = GroupInfo()
-        existing_instructors = group_info.instructors = [
-            factories.HUser()._asdict(),
-            factories.HUser()._asdict(),
-            factories.HUser()._asdict(),
-        ]
+        existing_instructors = group_info.instructors = factory.build_batch(
+            dict, 3, FACTORY_CLASS=factories.HUser
+        )
 
         group_info.upsert_instructor(existing_instructors[1])
```

Or of course you could do something like `[u._asdict() for u in factories.HUser.build_batch(3)]`

tests/unit/lms/models/group_info_test.py
seanh: Done

PR #1704, opened 2020-04-30
Add DB migration for GroupInfo.info column

seanh:
This is the DB migration for https://github.com/hypothesis/lms/pull/1703

### DIFF
--- a/lms/migrations/versions/37710e6bcb66_add_group_info_info_column.py
+++ b/lms/migrations/versions/37710e6bcb66_add_group_info_info_column.py
@@ -1,0 +1,30 @@
+"""
+Add group_info.info column.
+
+Revision ID: 37710e6bcb66
+Revises: a930adadac74
+Create Date: 2020-04-30 12:21:20.645402
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
+
+# revision identifiers, used by Alembic.
+revision = "37710e6bcb66"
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/37710e6bcb66_add_group_info_info_column.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = "a930adadac74"
+
+
+def upgrade():
+    op.add_column(
+        "group_info", sa.Column("info", MutableDict.as_mutable(JSONB)),
+    )
+    group_info_table = sa.table(
+        "group_info", sa.Column("info", MutableDict.as_mutable(JSONB))
+    )
```
seanh, line 25: The migration only needs to know about the `"info"` column, doesn't care about the rest of `group_info`'s columns.

We define our own `group_info_table` in the migration rather than importing `lms.models.GroupInfo` because the migration needs the table's schema as it was at this particular point in the migration sequence, not as it's defined in the production code at whatever time when the script is run (which will be missing the `info` column at the time when this migration is run)

lms/migrations/versions/37710e6bcb66_add_group_info_info_column.py
```diff
@@ -0,0 +1,30 @@
...
+    op.execute(group_info_table.update().values(info={"instructors": []}))
```
seanh, line 26: It's normally a good idea to do schema and data migrations separately but I'm being lazy and combining these. There are only 2600 rows in the production `group_info` table so I think we're okay to update them all at once (it won't lock the DB for too long)

Issue #1409, opened 2020-01-30
Deal with GroupInfoService


seanh:
This is a stub issue for now but we're gonna have to address how GroupInfoService needs to change now that we have multiple section groups per course.

seanh:
There might not be much we need to change here. Still investigating. One thing we might want to do is tag a group as either a course group or a section group

PR #1705, opened 2020-04-30
Make GroupInfoService work with section groups

seanh:
Fixes https://github.com/hypothesis/lms/issues/1409, although I still want to add a "type" field to distinguish course from section groups

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -17,18 +17,13 @@ class BaseClass:
     """Functions common to all SQLAlchemy models."""
 
     @classmethod
-    def iter_properties(cls):
-        """Iterate over all declared SQLAlchemy properties."""
-        return inspect(cls).iterate_properties
-
-    @classmethod
-    def iter_columns(cls):
-        """Iterate over all declared SQLAlchemy columns."""
-        return (
-            property
-            for property in cls.iter_properties()
+    def columns(cls):
+        """Return a list of all declared SQLAlchemy column names."""
+        return [
+            property.key
+            for property in inspect(cls).iterate_properties
             if isinstance(property, ColumnProperty)
-        )
+        ]
 
     def update_from_dict(self, data, skip_keys=None):
         """
@@ -51,7 +46,7 @@ class BaseClass:
                 f"Expected a set of keys to skip but found '{type(skip_keys)}'"
             )
 
-        columns = {col.key for col in self.iter_columns()}
+        columns = set(self.columns())
         columns -= skip_keys
 
         for key in columns:
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -4,7 +4,7 @@ from urllib.parse import urlparse
 
 import jwt
 
-from lms.models import HUser
+from lms.models import GroupInfo, HUser
 from lms.services import ConsumerKeyError, HAPIError
 from lms.validation.authentication import BearerTokenSchema
 from lms.views.helpers import via_url
@@ -391,6 +391,11 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
                 "course": {
                     "context_id": req.params["context_id"],
                     "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+                },
+                "group_info": {
+                    key: value
+                    for key, value in req.params.items()
+                    if key in GroupInfo.columns()
                 },
             },
         }
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -48,7 +48,7 @@ class LTIHService:  # pylint:disable=too-few-public-methods
         except HAPIError as err:
             raise HTTPInternalServerError(explanation=err.explanation) from err
 
-        # Keep a note of the groups locally for reporting purposes
+        # Keep a note of the groups locally for reporting purposes.
         for h_group in h_groups:
             self._group_info_service.upsert(
                 authority_provided_id=h_group.authority_provided_id,
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -28,15 +28,18 @@ class LTIHService:  # pylint:disable=too-few-public-methods
         self._h_api = request.find_service(name="h_api")
         self._group_info_service = request.find_service(name="group_info")
 
-    def sync(self, h_groups):
+    def sync(self, h_groups, group_info_params):
         """
-        Sync standard data to H for an LTI launch with the provided groups.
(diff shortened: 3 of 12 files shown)


lms/resources/_js_config.py
```diff
@@ -392,5 +392,6 @@ def _sync_api(self):
                     "context_id": req.params["context_id"],
                     "custom_canvas_course_id": req.params["custom_canvas_course_id"],
                 },
+                "group_info": dict(req.params),
```
seanh, line 395: Should probably reduce this to just the necessary subset of params

jon-betts: Are these "group_info" or are they "analytics_info" or something? I feel like group info in general is not the right name.

lms/db/__init__.py
```diff
@@ -17,18 +17,13 @@ class BaseClass:
...
-    def iter_columns(cls):
-        """Iterate over all declared SQLAlchemy columns."""
-        return (
-            property
-            for property in cls.iter_properties()
+    def columns(cls):
+        """Return a list of all declared SQLAlchemy column names."""
+        return [
+            property.key
+            for property in inspect(cls).iterate_properties
```
seanh, line 24: I changed this to return a list so that I don't have to worry about situations where a generator might not work. But the main change is to return the property keys (column name strings) rather than returning SQLAlchemy column objects--everywhere this method is used it's the keys that're actually wanted.

See https://github.com/hypothesis/lms/pull/1705/commits/7a2511b0f7d052d1cf910f59dab4b4c6fcce6443

lms/resources/_js_config.py
```diff
@@ -392,5 +392,10 @@ def _sync_api(self):
                     "context_id": req.params["context_id"],
                     "custom_canvas_course_id": req.params["custom_canvas_course_id"],
                 },
+                "group_info": {
+                    key: value
+                    for key, value in req.params.items()
+                    if key in GroupInfo.columns()
+                },
```
seanh, line 399: I changed this to only add into `group_info` that request params that're relevant to the `group_info` table by filtering based on `GroupInfo.columns()`. Previously it was adding _all_ the request params here, including things like the OAuth 1 signature

lms/resources/_js_config.py
seanh: I think `group_info` is fine: the table is one row per group (whether a course group or section group) and it contains the group's `authority_provided_id` (intended to be used to look up the group in the h DB) plus a bunch of "contextual" information about the group: the LTI params the group was last launched with, and a list of instructors who've launched the groups.

seanh:
I fixed this to filter the params that're being added to the `JSConfig` down to only those params that're actually needed for the `GroupInfo` table rather than _all_ the request's params, so that it's not putting things like the OAuth 1 signature into the JS config.

lms/db/__init__.py
jon-betts: Returning the column names is a reasonable change, but should the name be `column_names` or something similar, as it doesn't actually return columns as you might expect from an SQLAlchemy models?

The previous method had the advantage of being useful for more than we need, at the cost of being annoying every time we actually used it, so I suppose we can go the other way one day if we ever need it, and benefit from being simpler in the mean time.

The generator should go back in. We don't need a list in any of the locations this function is used. This results in us creating a list from an iterator just to iterate over it and then discard it which loses a (very small) amount of time and memory. In general Python 3 has made a move away from using lists in favour of generators (`iter_items -> items`, `xrange -> range` etc). The underlying `SQLAlchemy` methods also return generators, presumably for the same reasons and I don't see anything here that suggests we have a special case that would warrant going against the grain.

Given the general move to make iterators the default, I like removing the `iter_` from the front. I don't think this is something we have to warn people about.

lms/db/__init__.py
```diff
@@ -51,7 +46,7 @@ def update_from_dict(self, data, skip_keys=None):
                 f"Expected a set of keys to skip but found '{type(skip_keys)}'"
             )
 
-        columns = {col.key for col in self.iter_columns()}
+        columns = set(self.columns())
```
jon-betts, line 49: This would have been a good change before as well:

` set(self.iter_colums())`

lms/db/__init__.py
seanh: I'm going to take the liberty of leaving this as it is (still changing the generator to a list) and merging it, because there's quite a lot I want to get delivered this sprint and don't want to get stuck on this.

I think as a general rule we should avoid returning iterators where lists would do (e.g. where the efficiency gain is insignificant), because:

* Iterators lack various list methods that people are likely to expect (subscripting etc)

* It's very likely that, when mocking a method that returns an iterator, developers will set the mock's `return_value` to a list not an iterator. It's just much easier to do `some_mock.return_value = [1, 2, 3]` and I think developers are unlikely to remember in every case to make it an iterator instead. If the code calls list methods like subscripting that will then work with the mock but not work in production, so the tests could be passing when the code is wrong.

* Even if you do remember to, the return values of generator functions are difficult to mock correctly, very easy to get it wrong in a way that can make the tests fail when they shouldn't be failing. Or the tests might be passing now but then some correct code change is made in future and then a poorly mocked generator causes tests to start failing when they shouldn't.

seanh: I think this might be what you'd have to do to mock it:

1. Use `side_effect` rather than `return_value` when creating the mock
2. Set `side_effect` to a function that returns an iterator
3. Use `side_effect()` (you have to call it every time) rather than `return_value` when querying the mock in tests

Seems to work. Many of the more obvious possibilities don't work for various reasons.

```diff
diff --git a/tests/unit/lms/resources/_js_config_test.py b/tests/unit/lms/resources/_js_config_test.py
index c0b4d00a..5a3cb002 100644
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -342,7 +342,7 @@ class TestJSConfigAPISync:
                 "group_info": {
                     key: value
                     for key, value in pyramid_request.params.items()
-                    if key in GroupInfo.columns.return_value
+                    if key in GroupInfo.columns.side_effect()
                 },
             },
         }
@@ -520,5 +520,7 @@ def via_url(patch):
 @pytest.fixture(autouse=True)
 def GroupInfo(patch):
     group_info_class = patch("lms.resources._js_config.GroupInfo")
-    group_info_class.columns.return_value = ["context_id", "custom_canvas_course_id"]
+    group_info_class.columns.side_effect = lambda: iter(
+        ["context_id", "custom_canvas_course_id"]
+    )
     return group_info_class
```

PR #1709, opened 2020-05-04
Add requestGroups RPC call

LMS007:
`requestGroups` is a non-standard RPC handler method in that its value is pre-loaded before the client makes the request. This value is resolved through an api endpoint call (sync) on app load. The awaited result of that call is returned to the client via  RPC. e.g. `requestGroups`.

The pre-loaded result is done in BasicLtiLauchApp.js and has retry support if for any reason the auth token expires or is invalided.

BasicLtiLauchApp will also wait to render its own iframe until the sync api endpoint (which returns the groups) is resolved.

This feature is primary used to support canvas sections.

----------

Mostly the difficulty with this change involved keeping BasicLtiLaunch as close to it originally was but allowing either the callback request or the groups request to independently fetch, succeed or fail. 

Part of the reasons for the refactor came from needed independent control over the fetch and error handling from not 1 but 2 requests on load. The useCallback and hooks were causing some issues with looping updates when setting state as I was building this out so I moved the api calls out of useCallback and they are now just called either onload or on auth retry, but not for any other reason. This helped me get this working, but there may yet be an opportunity to pull things back into hooks -- I have not explored that.

In an attempt to simplify the flow. I got rid of `failedAction` and just use errorState as the logic to control which Dialog needs to be displayed. The Spinner will show if either of the 2 request are still pending, but like before, it does not operate on the submission request for students.

Currently, this will avoid the situation where if one request succeeds after the first fails, then it wont override the UI state error state. If its a retry-able error, then both the groups and callback requests re-try.

**Update:** Groups no longer blocks `<iframe>`.

This now will allow the `<iframe>` to be built when the contentUrl is resolved (either passed in at the outset via prop) or resolved with the viaCallbackUrl api request. I confirmed this works by artificially. The spinner will still remain until 1. viaCallbackUrl and the 2. /sync (groups) resolve or error.

1. Adding a test for this: `renders the iframe hidden after contentUrl succeeds and groups remains pending`
2. Manually preventing groups from resolving and then inspecting the network requests to see the assignment through via request be made along with the sidebar's usual requests.


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -177,20 +217,20 @@ export default function BasicLtiLaunchApp() {
...
       await authWindow.current.authorize();
-      await fetchContentUrl();
+      await Promise.all([fetchContentUrl(), fetchGroups()]);
     } finally {
-      // eslint-disable-next-line require-atomic-updates
       authWindow.current = null;
     }
-  }, [authToken, canvas, fetchContentUrl]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [authToken, canvas.authUrl]);
```
LMS007, line 225: I have not flushed out a comment for this, but its worth discussing here in the PR. To fix this, it needs both fetchContentUrl and fetchGroups, and then it needs those to be wrapped in useCallback along with handleError with their respective dependencies. Doing all this causes an infinite update loop. 

I do need to take a closer look to see if there is a way to simplify the logic or we just need to force the linter to ignore `react-hooks/exhaustive-deps` and control the flow in a more manual way like this.

robertknight:
I checked out this branch and was able to get login working. I applied the following diff locally, which looks identical to https://github.com/hypothesis/lms/pull/1720/ as far as I can see, and the app still worked as expected with and without existing authorization:

```diff
diff --git a/conf/development.ini b/conf/development.ini
index 10887810..977894b9 100644
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -18,6 +18,9 @@ session_cookie_secret = "notasecret"
 feature_flags_cookie_secret = "notasecret"
 feature_flags_allowed_in_cookie = use_legacy_via
 
+# TESTING
+feature_flags.section_groups = true
+
 # The secret string that's used to sign the OAuth2 state param.
 oauth2_state_secret = "notasecret"
 
diff --git a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
index 06d0d79f..c9754177 100644
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -104,7 +104,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
    *  "error-authorizing", or "error-report-submission"
    * @param {boolean} [retry=true] - Can the request be retried?
    */
-  const handleError = (e, errorState, retry = true) => {
+  const handleError = useCallback((e, errorState, retry = true) => {
     setFetched();
     if (e instanceof ApiError && !e.errorMessage && retry) {
       setErrorState('error-authorizing');
@@ -112,12 +112,12 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
       setError(e);
       setErrorState(errorState);
     }
-  };
+  }, []);
 
   /**
    * Fetch the groups from the sync endpoint if `sync` object exists.
    */
-  const fetchGroups = async () => {
+  const fetchGroups = useCallback(async () => {
     if (apiSync) {
       try {
         setFetching();
@@ -132,7 +132,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         handleError(e, 'error-fetch');
       }
     }
-  };
+  }, [apiSync, authToken, handleError, rpcServer]);
 
   /**
    * Fetch the URL of the content to display in the iframe if `viaCallbackUrl`
@@ -140,7 +140,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
    *
    * This will typically be a PDF URL proxied through Via.
    */
-  const fetchContentUrl = async () => {
+  const fetchContentUrl = useCallback(async () => {
     if (!viaCallbackUrl) {
       // If no "callback" URL was supplied for the frontend to use to fetch
       // the URL, then the backend must have provided the Via URL in the
@@ -158,7 +158,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     } catch (e) {
       handleError(e, 'error-fetch');
     }
-  };
+  }, [authToken, handleError, viaCallbackUrl]);
 
   /**
    * Fetch the assignment content URL and groups when the app is initially displayed.
@@ -166,8 +166,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   useEffect(() => {
     fetchContentUrl();
     fetchGroups();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [fetchContentUrl, fetchGroups]);
 
   /**
    * Report a submission to the LMS, with the LMS-provided metadata needed for
@@ -221,8 +220,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     } finally {
       authWindow.current = null;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [authToken, canvas.authUrl]);
+  }, [authToken, canvas.authUrl, fetchContentUrl, fetchGroups]);
 
   if (showIframe()) {
     const iFrame = (
```

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,105 +25,155 @@ import Spinner from './Spinner';
...
+
+  // When the app is initially displayed, it will use the Via URL if given
+  // or invoke the API callback to fetch the URL otherwise.
+  const [contentUrl, setContentUrl] = useState(viaUrl ? viaUrl : null);
+
+  // How many current blocking request are pending
+  const [fetchCount, setFetchCount] = useState(0);
+
+  // Var to keep track of the fetchCount
+  const count = useRef(0);
```
robertknight, line 67: You can replace this with a functional state updater:

```js
const incFetchCount = () => {
  setFetchCount(count => count + 1);
  setErrorState(null);
};
...
const decFetchCount = () => setFetchCount(count => count - 1)
```

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,105 +25,155 @@ import Spinner from './Spinner';
...
+
+  // `AuthWindow` instance, set only when waiting for the user to approve
+  // the app's access to the user's files in the LMS.
+  const authWindow = useRef(null);
+
+  // Show the assignment when there are no pending requests
+  // and errorState is falsely
+  const showIframe = () => {
```
robertknight, line 75: Since this function and the one below take no arguments, they can just be plain values (ie. `const showIframe = fetchCount === 0 && !errorState`)

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,105 +25,155 @@ import Spinner from './Spinner';
...
 
   /**
-   * Fetch the assignment content URL when the app is initially displayed.
+   * Fetch the assignment content URL and groups when the app is initially displayed.
    */
   useEffect(() => {
     fetchContentUrl();
-  }, [fetchContentUrl]);
+    fetchGroups();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
robertknight, line 169: In my brief manual testing today, I wasn't able to reproduce any issues when removing the two lint suppressions in this file and adding the dependencies and `useCallback(...)` hooks as directed by the lint errors.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
robertknight: I couldn't reproduce this in testing today. I did run into an issue with authorization was failing because Sean was testing some stuff (see https://hypothes-is.slack.com/archives/CBN3DGW02/p1588787487144200) but that resolved itself once he fixed the devdata keys.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,105 +25,155 @@ import Spinner from './Spinner';
...
+        setFetching();
+        const groups = await apiCall({
+          authToken,
+          path: apiSync.path,
+          data: apiSync.data,
+        });
+        rpcServer.resolveGroupFetch(groups);
+        setFetched();
+      } catch (e) {
+        handleError(e, 'error-fetch');
```
robertknight, line 132: If an exception is thrown, this will leave the `fetchCount` state with a value > 0. One way to resolve that would be decrementing the counter in a `finally` block.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,105 +25,155 @@ import Spinner from './Spinner';
...
-        });
-      } else {
-        setLtiLaunchState({
-          ...INITIAL_LTI_LAUNCH_STATE,
-          state: 'error',
-          error: e,
-          failedAction: 'fetch-url',
-        });
-      }
+      handleError(e, 'error-fetch');
```
robertknight, line 159: This `try ... catch` block has the same issue as above in that the `fetchCount` value will not be decremented if an exception occurs.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -209,31 +249,30 @@ export default function BasicLtiLaunchApp() {
...
-      )}
-      {ltiLaunchState.state === 'error' &&
-        ltiLaunchState.failedAction === 'fetch-url' && (
+  } else {
+    // Render either a Spinner or Dialog.
+    return (
+      <Fragment>
+        {showSpinner() && <Spinner className="BasicLtiLaunchApp__spinner" />}
+        {errorState === 'error-authorizing' && (
+          <Dialog
```
robertknight, line 258: It might be useful to extract a helper component out of these dialogs to ensure consistency for the parts that are supposed to be the same across all of them.

lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -1,15 +1,32 @@
 import Server from './server';
-import { requestConfig, requestGroups } from './methods';
+import { requestConfig } from './methods';
```
robertknight, line 2: I would recommend to just move the remaining contents of `methods` into this file, since it is 4 lines of trivial code and I don't see any value in having a separation between `index.js` and `methods.js`.

As an aside, Jon B had to do a similar refactoring on the backend recently.

lms/static/scripts/postmessage_json_rpc/server/index.js
```diff
@@ -1,15 +1,32 @@
...
+   * object. In order to speed this up even more, the groups are pre-loaded
+   * before this RPC endpoint (`requestGroups`) is queried -- This method
+   * simply returns an awaited promise to the client.
+   */
+  const groupsPromise = new Promise(resolve => {
+    server.resolveGroupFetch = resolve;
+  });
+  server.register('requestGroups', async () => {
+    return await groupsPromise;
+  });
```
robertknight, line 27: I can't see a test for this. Can you add one?

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: That may be the case, I didn't circle back around that after I got everything working, but I'll test all the error paths too. Thanks Rob.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: Awe indeed, it will skip the setFetched(); I will use finally.

lms/static/scripts/postmessage_json_rpc/server/index.js
LMS007: makes sense

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: Looked closely, but I'm gonna leave that for another pass, each dialog is slightly different than the other and I have another ticket to refactor the Dialog's css so it may be a good time to revisit this for readability when doing that ticket.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
+import { startRpcServer, getSidebarWindow, $imports } from '../index';
```
LMS007, line 1: @robertknight 
Was not sure what our standard was for testing index.js files, but I added these tests in any case.

lms/static/scripts/postmessage_json_rpc/server/index.js
LMS007: done

LMS007:
Still not currently testable in canvas do the auth issues, but I hope that will be resolved tomorrow.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
robertknight: OK, that's fine by me.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,6 +305,70 @@ describe('BasicLtiLaunchApp', () => {
...
+        path: '/api/sync',
+        data: {
+          course: {
+            context_id: '12345',
+            custom_canvas_course_id: '101',
+          },
+        },
+      });
+    });
+    it('renders the spinner until both groups and contentUrl requests finish', async () => {
```
robertknight, line 343: Nit. Can you add a blank line between tests for readability please? (Just the tests in this block are affected)

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,6 +305,70 @@ describe('BasicLtiLaunchApp', () => {
     });
   });
 
+  describe('multiple fetches', () => {
```
robertknight, line 308:
```suggestion
  describe('concurrent fetching of groups and content URL', () => {
```

I think a slightly more descriptive name here will make these tests easier to follow.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,6 +305,70 @@ describe('BasicLtiLaunchApp', () => {
...
+      await contentUrl;
+      // Spinner should not go away after first request
+      assert.isTrue(wrapper.exists('Spinner'));
+
+      await groups;
+      wrapper.update();
+      // Spinner should hide after second request finishes
+      assert.isFalse(wrapper.exists('Spinner'));
+    });
+    it('shows an error dialog if the first request fails and second succeeds', async () => {
```
robertknight, line 358: What about the case where the first request succeeds and the second one fails?

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+
+describe('postmessage_json_rpc/server', () => {
```
robertknight, line 3:
```suggestion
describe('postmessage_json_rpc/server/index', () => {
```

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
robertknight: I suggest that in cases where we do write tests for a file named `index.js`, we treat it just like any other module.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+      register: FakeRegister,
+      sidebarWindow: 'FakeSidebarWindow',
+    });
+
+    $imports.$mock({
+      './server': FakeServer,
+    });
+  });
+
+  afterEach('remove the server config from the document', () => {
```
robertknight, line 20:
```suggestion
  afterEach(() => {
```

The first argument was obsolete here, as that is now handled in a separate `afterEach` below.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+    $imports.$restore();
+  });
+
+  describe('#startRpcServer', () => {
```
robertknight, line 24:
```suggestion
  describe('startRpcServer', () => {
```

There are many existing tests that do this wrongly but `#` should be reserved for instance properties and methods (following JSDoc conventions). If you're just testing a function, just use the function name without a `#` prefix.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+    it('returns a server object', async () => {
```
robertknight, line 25: Does this test need to be async?

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+      const server = startRpcServer();
+      assert.isTrue(FakeServer.called);
+      assert.isOk(server);
+    });
+  });
+
+  describe('#getSidebarWindow', () => {
+    it('returns the value of server.sidebarWindow', async () => {
```
robertknight, line 33: It doesn't look like this test needs to be async?

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
...
+        configEl.setAttribute('type', 'application/json');
+        configEl.classList.add('js-config');
+        configEl.textContent = JSON.stringify({
+          hypothesisClient: { foo: 'bar' },
+        });
+        document.body.appendChild(configEl);
+      });
+
+      afterEach('remove the client config from the document', () => {
+        configEl.parentNode.removeChild(configEl);
```
robertknight, line 68: You can simplify this to `configEl.remove()`. Using `removeChild` used to be necessary when the tests ran in an older version of PhantomJS, but now they run in a modern version of Chrome.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: I will, I missed those.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: I didn't bother because its the same result as a single request just failing, but I'll add that test in any case.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
LMS007: This is my new "RCP", lol. I'm gonna make a ticket to clean this up because this has confused me a few times..

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
LMS007: I will do that for ./server-test.js too.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,118 @@ describe('BasicLtiLaunchApp', () => {
...
+        data: {
+          course: {
+            context_id: '12345',
+            custom_canvas_course_id: '101',
+          },
+        },
+      });
+    });
+
+    it.skip('renders the spinner until both groups and contentUrl requests finish', async () => {
```
LMS007, line 349: This test was a false negative and was in fact not correct before.

I changed it to use `waitForElementToRemove` and that needs to fail. I would in fact use `assert.rejects` but that has not been ported to LMS yet. But in any case, this test fails.

It seems that the groups and contentUrl just resolve right away without needing to call `await`. That means I can't pause on an in-between state between `groups` and `contentUrl` to check that the spinner is still spinning.

`waitForElementToRemove` is a bit more accurate than just calling update() once, and in this case it does prove to be different. What I really want to show is: if `groups` does not reject or resolve, then still show a spinner.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,118 @@ describe('BasicLtiLaunchApp', () => {
...
+      const wrapper = renderLtiLaunchApp();
+      await contentUrl;
+      // Should show an error after the first request fails
+      await waitForElement(wrapper, 'FakeDialog[title="Authorize Hypothesis"]');
+      await groups;
+      // Should still show an error even if the second request does not fail
+      await waitForElement(wrapper, 'FakeDialog[title="Authorize Hypothesis"]');
+    });
+
+    it.skip('shows an error dialog if the first request succeeds and second fails', async () => {
```
LMS007, line 392: same problem as the other skipped test

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,53 +25,109 @@ import Spinner from './Spinner';
...
-  const [ltiLaunchState, setLtiLaunchState] = useState({
-    ...INITIAL_LTI_LAUNCH_STATE,
-    state: viaCallbackUrl ? 'fetching-url' : 'fetched-url',
-    contentUrl: viaUrl ? viaUrl : null,
-  });
+  // The current state of the error.
+  // One "error-fetch", "error-authorizing", or "error-report-submission"
+  const [errorState, setErrorState] = useState(null);
+
+  // Any current error event object thrown.
```
robertknight, line 56: I don't think its obvious what "error event object" refers to here. Do you just mean the `Error` object associated with `errorState` if it is non-null?

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,53 +25,109 @@ import Spinner from './Spinner';
...
+  // the app's access to the user's files in the LMS.
+  const authWindow = useRef(null);
+
+  // Show the assignment when the contentUrl has resolved and errorState
+  // is falsely
+  const showIframe = contentUrl && !errorState;
+
+  // Show the loader is there are any pending requests and
+  // and errorState is falsely
+  const showSpinner = fetchCount > 0;
```
robertknight, line 76: This comment is out of date (it mentions `errorState` which is handled by the logic for `showIframe` above). Can I suggest just removing it since it doesn't really tell me anything not clear from the code.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,53 +25,109 @@ import Spinner from './Spinner';
...
+
+  // Increment the fetch counter by 1 and clear any
+  // previous error state.
```
robertknight, line 79: I think the purpose and effect `incFetchCount` and `decFetchCount` is clear from the name and the content, so I would just remove the comments above them.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
robertknight: Update: We discussed this on Slack. The intent here was to only ever execute this on first mount. Having said that, it looks to me like you'd get that even if you did specify dependencies, since `fetchContentUrl` and `fetchGroups` never change. That might avoid confusion if for some reason in future we make a change where these callbacks' dependencies do change and we do want them to re-run.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,109 @@ describe('BasicLtiLaunchApp', () => {
...
+      const contentUrl = contentUrlCall.resolves({
+        via_url: 'https://via.hypothes.is/123',
+      });
+      const groups = groupsCall.resolves(['group1', 'group2']);
+      const wrapper = renderLtiLaunchApp();
+      await contentUrl;
+      // Spinner should not go away after first request
+      wrapper.update();
+      assert.isTrue(wrapper.find('Spinner').exists());
+      await groups;
```
robertknight, line 359: Is this `await groups` doing anything any more given the `await` below?

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,109 @@ describe('BasicLtiLaunchApp', () => {
...
+      const contentUrl = contentUrlCall.resolves({
+        via_url: 'https://via.hypothes.is/123',
+      });
+      const groups = groupsCall.rejects(new ApiError(400, {}));
+      const wrapper = renderLtiLaunchApp();
+      await contentUrl;
+      // Should not show an error yet
+      assert.isFalse(
+        wrapper.find('FakeDialog[title="Authorize Hypothesis"]').exists()
+      );
```
robertknight, line 398: Asserting that something is not rendered "yet" is a bit tricky because it is possible that some batched state updates haven't yet been flushed (Preact currently does this internally using something equivalent to `queueMicrotask`). The pattern I would probably use is something like:

1. Put the `groups` Promise into a pending state where it won't resolve or reject
2. Wait one macro-task (`setTimeout` with a delay of 0). This will let any pending renders flush.
3. Check that the auth dialog hasn't rendered
4. Reject the `groups` promise
5. Now wait for the "Authorize Hypothesis" dialog to appear

robertknight: Also, be aware that you need to call `wrapper.update(...)` again before querying it here.

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
```diff
@@ -0,0 +1,80 @@
+import { startRpcServer, getSidebarWindow, $imports } from '../index';
+
+describe('postmessage_json_rpc/server/index', () => {
+  let FakeServer;
+  let FakeRegister;
```
robertknight, line 5:
```suggestion
  let fakeRegister;
```

The convention in our code is to use `FakeXYZ` if the thing you are faking has a capitalized name or `fakeXYZ` otherwise, and in this case the faked method is just called `register`.

lms/static/scripts/test-util/wait.js
```diff
@@ -61,3 +61,26 @@ export function waitForElement(wrapper, selector, timeout = 10) {
...
+
+/**
+ * Wait up to `timeout` ms for an element to be removed.
+ *
+ * @param {CommonWrapper} wrapper - Root Enzyme wrapper
+ * @param {string|Function} selector - Selector string or function to pass to `wrapper.find`
+ * @param {number} timeout
+ * @return {Promise<CommonWrapper>}
+ */
+export function waitForElementToRemove(wrapper, selector, timeout = 10) {
```
robertknight, line 73:
```suggestion
export function waitForElementToBeRemoved(wrapper, selector, timeout = 10) {
```

The meaning of the previous name was obvious, this is just me being a grammar pedant.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,109 @@ describe('BasicLtiLaunchApp', () => {
...
+          course: {
+            context_id: '12345',
+            custom_canvas_course_id: '101',
+          },
+        },
+      });
+    });
+
+    it('renders the spinner until both groups and contentUrl requests finish', async () => {
+      const contentUrl = contentUrlCall.resolves({
```
robertknight, line 350: I don't think `contentUrl` is what you are expecting it to be. `stub.resolves(...)` returns _the stub_ for chaining, not the Promise. In other words:

```js
const stub = sinon.stub()
stub.resolves('foobar') === stub // true. `resolves(...)` returns `this`.
```

robertknight: The same applies to other statements of the form:

```js
const someVar = someStub.resolves(...)
```

Below.

robertknight: To show what I mean, the tests still pass if I replace all uses of `await contentUrl` or `await groups` with `await true`:

```diff
diff --git a/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js b/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
index be6044e6..948a64d4 100644
--- a/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
@@ -347,16 +347,16 @@ describe('BasicLtiLaunchApp', () => {
     });
 
     it('renders the spinner until both groups and contentUrl requests finish', async () => {
-      const contentUrl = contentUrlCall.resolves({
+      contentUrlCall.resolves({
         via_url: 'https://via.hypothes.is/123',
       });
-      const groups = groupsCall.resolves(['group1', 'group2']);
+      groupsCall.resolves(['group1', 'group2']);
       const wrapper = renderLtiLaunchApp();
-      await contentUrl;
+      await true;
       // Spinner should not go away after first request
       wrapper.update();
       assert.isTrue(wrapper.find('Spinner').exists());
-      await groups;
+      await true;
       // Spinner should go away after the second request
       await waitForElementToRemove(wrapper, 'Spinner');
       // iframe should be visible
@@ -364,39 +364,39 @@ describe('BasicLtiLaunchApp', () => {
     });
 
     it('renders the iframe hidden after contentUrl succeeds and groups remains pending', async () => {
-      const contentUrl = contentUrlCall.resolves({
+      contentUrlCall.resolves({
         via_url: 'https://via.hypothes.is/123',
       });
       const wrapper = renderLtiLaunchApp();
-      await contentUrl;
+      await true;
       wrapper.update();
       assert.equal(wrapper.find('iframe').prop('style').visibility, 'hidden');
     });
 
     it('shows an error dialog if the first request fails and second succeeds', async () => {
-      const contentUrl = contentUrlCall.rejects(new ApiError(400, {}));
-      const groups = groupsCall.resolves(['group1', 'group2']);
+      contentUrlCall.rejects(new ApiError(400, {}));
+      groupsCall.resolves(['group1', 'group2']);
       const wrapper = renderLtiLaunchApp();
-      await contentUrl;
+      await true;
       // Should show an error after the first request fails
       await waitForElement(wrapper, 'FakeDialog[title="Authorize Hypothesis"]');
-      await groups;
+      await true;
       // Should still show an error even if the second request does not fail
       await waitForElement(wrapper, 'FakeDialog[title="Authorize Hypothesis"]');
     });
 
     it('shows an error dialog if the first request succeeds and second fails', async () => {
-      const contentUrl = contentUrlCall.resolves({
+      contentUrlCall.resolves({
         via_url: 'https://via.hypothes.is/123',
       });
-      const groups = groupsCall.rejects(new ApiError(400, {}));
+      groupsCall.rejects(new ApiError(400, {}));
       const wrapper = renderLtiLaunchApp();
-      await contentUrl;
+      await true;
       // Should not show an error yet
       assert.isFalse(
         wrapper.find('FakeDialog[title="Authorize Hypothesis"]').exists()
       );
-      await groups;
+      await true;
       // Should show an error after failure
       await waitForElement(wrapper, 'FakeDialog[title="Authorize Hypothesis"]');
     });
```

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: correct, how about "error object"

lms/static/scripts/postmessage_json_rpc/server/test/index-test.js
LMS007: gotcha!

LMS007:
@robertknight, I hear you points, and I think this caused a lot of confusion on my part. There may be other tests sprinkled in our source (on the client) that could use a second look. In any case, I think my `fixup` here may be a more appropriate way to test this, but if there simpler approach (or this is still invalid) I'm all ears.

Basically instead of using sinon to reject or resolve. I instead return a custom `new Promise`, then save the resolver and rejector as local vars.

Then I call those in the middle of the test exactly when I need the promise to reject or resolve but not before. I think the bulk of the struggles I had with these tests is that there is no way to set up sinon to to resolve/reject but only after we give it some que to do so. E.g. we need to testing the pending state first, then proceed with its resolve/reject business. 

Thanks!

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,53 +25,104 @@ import Spinner from './Spinner';
...
+   * Helper to handle thrown errors from from API requests.
+   *
+   * @param {Error} e - Error object from request.
+   * @param {string} errorState - One of "error-fetch",
+   *  "error-authorizing", or "error-report-submission"
+   * @param {boolean} [retry=true] - Can the request be retried?
+   */
+  const handleError = (e, errorState, retry = true) => {
+    if (e instanceof ApiError && !e.errorMessage && retry) {
+      setErrorState('error-authorizing');
```
robertknight, line 95: I found an issue in testing that if the server returns 500 when the sync API is called, the frontend just prompts for authorization instead of indicating that an unexpected error occurred. The logic for checking the error type is the same as before, so I suggest we deal with that separately.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -23,53 +25,104 @@ import Spinner from './Spinner';
...
     canvas,
   } = useContext(Config);
 
-  const [ltiLaunchState, setLtiLaunchState] = useState({
-    ...INITIAL_LTI_LAUNCH_STATE,
-    state: viaCallbackUrl ? 'fetching-url' : 'fetched-url',
-    contentUrl: viaUrl ? viaUrl : null,
-  });
+  // The current state of the error.
+  // One "error-fetch", "error-authorizing", or "error-report-submission"
```
robertknight, line 53:
```suggestion
  // One of "error-fetch", "error-authorizing", or "error-report-submission"
```

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -254,11 +310,124 @@ describe('BasicLtiLaunchApp', () => {
...
+    });
+
+    it('renders the spinner until both groups and contentUrl requests finish', async () => {
+      const wrapper = renderLtiLaunchApp();
+
+      // Spinner should not go away after first request
+      contentUrlResolve({
+        via_url: 'https://via.hypothes.is/123',
+      });
+      assert.isTrue(wrapper.find('Spinner').exists());
```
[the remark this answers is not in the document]
LMS007: I wonder if would have a similar effect?

`assert.rejects(await waitForSpinnerToBeRemoved(...))`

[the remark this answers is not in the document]
robertknight: As long as you set a very short timeout (say 1ms) on the `waitForXXX` call, and the `waitForXXX` call is calling `wrapper.update()` before each. check, then yes this could work.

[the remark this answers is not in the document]
robertknight: (If you don't set a short timeout, `waitForXXX` will of course just hang for several seconds waiting for something that doesn't happen).

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: will do

PR #1710, opened 2020-05-04
Remove the un-used functions which are just for the serial API

jon-betts:
This strips out the calls to individual methods in the H-API now the bulk API has been in place for a while.


lms/services/h_api.py
```diff
@@ -61,96 +61,9 @@ def get_user(self, username):
...
-
-        This is equivalent to calling `update_user()` then `create_user()` if
-        that fails.
-
-        :param h_user: the updated user value
-        :type h_user: HUser
-        """
-        try:
-            self.update_user(h_user)
-        except HAPINotFoundError:
```
jon-betts, line 112: The removed code to do with upserting was the only place we caught HAPINotFoundError and it serves no purpose with these gone.

lms/services/h_api.py
```diff
@@ -159,43 +72,33 @@ def _api_request(self, method, path, data=None, headers=None, body=None):
...
         :raise HAPIError: if the request fails for any other reason
         :return: the response from the h API
         :rtype: requests.Response
         """
         headers = headers or {}
         headers["Hypothesis-Application"] = "lms"
 
-        request_args = {"headers": headers}
-
+        request_args = {}
```
jon-betts, line 85: It was a bit weird to set headers like this, when we set everything else directly below. The reason to do this is when we don't know if the arg will be set at all (like "data" below). We always set headers so I've used a named parameter to the call to request below.

tests/unit/lms/services/h_api_test.py
```diff
@@ -221,26 +86,8 @@ def test_if_given_custom_headers__api_request_adds_them(self, h_api, requests):
...
-        with pytest.raises(HAPINotFoundError) as exc_info:
-            h_api._api_request(sentinel.method, "dummy-path")
-
-        # It records the requests exception that caused the HAPIError.
-        assert exc_info.value.__cause__ == exception
-
-    @pytest.mark.parametrize(
-        "exception_class",
-        [requests_.ConnectionError, HTTPError, ReadTimeout, TooManyRedirects],
-    )
```
jon-betts, line 239: No point in enumerating different sub-classes of RequestException. We catch them all

lms/services/h_api.py
```diff
@@ -61,18 +59,13 @@ def get_user(self, username):
...
         Send any kind of HTTP request to the h API and return the response.
 
-        :param method: the HTTP request method to use, for example "GET",
-                       "POST", "PUT", "PATCH" or "DELETE"
+        :param method: the HTTP request method to use, (e.g. "GET")
         :param path: the h API path to post to, relative to
-                     ``settings["h_api_url_private"]``, for example:
-                     ``"users"`` or ``"groups/<GROUPID>/members/<USERID>"``
-        :param data: the data to post as JSON in the request body
+                     `settings["h_api_url_private"]` (e.g. "users")
```
jon-betts, line 68: Slimmed down the examples a bit here. It's odd to add examples we don't actually have in the code anymore

PR #1717, opened 2020-05-05
Add "type" to group_info

seanh:
Add a new `type` field to `group_info` that records whether a group is a course group or a section group.

### DIFF
--- a/lms/models/h_group.py
+++ b/lms/models/h_group.py
@@ -4,6 +4,7 @@ from typing import NamedTuple
 class HGroup(NamedTuple):
     name: str
     authority_provided_id: str
+    type: str = "course_group"
 
     def groupid(self, authority):
         return f"group:{self.authority_provided_id}@{authority}"
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -68,4 +68,4 @@ class GroupInfoService:  # pylint:disable=too-few-public-methods
                 dict(email=self._lti_user.email, **self._lti_user.h_user._asdict())
             )
 
-        group_info.info["type"] = type_
+        group_info.type = type_
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -19,7 +19,7 @@ class GroupInfoService:  # pylint:disable=too-few-public-methods
         self._db = request.db
         self._lti_user = request.lti_user
 
-    def upsert(self, authority_provided_id, consumer_key, params, type_):
+    def upsert(self, h_group, consumer_key, params):
         """
         Upsert a row into the ``group_info`` DB table.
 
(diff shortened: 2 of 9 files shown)


tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -15,6 +15,7 @@ def test_sync_when_the_user_is_a_learner(
         factories.HGroup(
             name="Section 2",
             authority_provided_id="d99674b9700f4a40a2b301d2949b61339c58236c",
+            type="section_group",
```
jon-betts, line 18: There's no "type" field in the corresponding factory class so I'm guessing factoryboy just lets you add any extra params you send?

For compatibility with the underlying object, it might be nice if it had the same default of `course_group`.

lms/services/group_info.py
```diff
@@ -61,3 +67,5 @@ def upsert(self, authority_provided_id, consumer_key, params):
             group_info.upsert_instructor(
                 dict(email=self._lti_user.email, **self._lti_user.h_user._asdict())
             )
+
+        group_info.info["type"] = type_
```
jon-betts, line 71: This little guy is a bit mysterious on his own. As we interact with instructors above using a method (which internally uses a property over `info`) should this be a property too? Otherwise you have to jump between the two files to see how `info` is built up.

lms/services/group_info.py
```diff
@@ -36,9 +36,15 @@ def upsert(self, authority_provided_id, consumer_key, params):
...
+
         :param consumer_key: the ``GroupInfo.consumer_key`` value to set
+
         :param params: the other ``GroupInfo`` columns to set
         :type params: dict
+
+        :param type_: the type of the group, e.g. "course_group" or "section_group"
+        :type type_: str
+
         """
```
jon-betts, line 48: Maybe go the other way and remove the types?

```suggestion
        :param authority_provided_id: the ``authority_provided_id`` of the
            ``GroupInfo`` to create or update
        :param consumer_key: the ``GroupInfo.consumer_key`` value to set
        :param params: the other ``GroupInfo`` columns to set
        :param type_: the type of the group, e.g. "course_group" or "section_group"
        """
```

lms/models/h_group.py
```diff
@@ -4,6 +4,7 @@
 class HGroup(NamedTuple):
     name: str
     authority_provided_id: str
+    type: str = "course_group"
```
jon-betts, line 7: How do we feel about `enum`s for this sort of thing?

lms/services/group_info.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, _context, request):
         self._db = request.db
         self._lti_user = request.lti_user
 
-    def upsert(self, authority_provided_id, consumer_key, params):
+    def upsert(self, authority_provided_id, consumer_key, params, type_):
```
jon-betts, line 22:
```suggestion
    def upsert(self, h_group, consumer_key, params):
```

Now we have the value object, we could use it to simplify this call

tests/unit/lms/views/api/canvas/sync_test.py
seanh: I actually hadn't considered the factory but yes it does seem to "just work" and the default from the model applies:

```python
>>> from tests import factories
>>> factories.HGroup()
HGroup(name='Test Group 0', authority_provided_id='166be8f2ded4070d27f993c52ad833ffe58d31b5', type='course_group')
>>> factories.HGroup(type="foo")
HGroup(name='Test Group 1', authority_provided_id='de2de2d96421ebde0da4a466c05045fff210f481', type='foo')
```

lms/services/group_info.py
seanh: Done c24dc94e6ab1768ee52ca34e11ed91eb42713f37

lms/models/h_group.py
seanh: Not sure really. In this case it's probably easier not to bother

lms/services/group_info.py
seanh: Done https://github.com/hypothesis/lms/pull/1717/commits/d6146b2fbe1f39bd89eafd29c1c1972f80c8d42b

lms/services/group_info.py
seanh: I've simplified this docstring: https://github.com/hypothesis/lms/pull/1717/commits/856ec13e8112002fcbfd86b90cd50f66af5eb1a8

I do think the `:type`'s on the `h_group` and `params` arguments are adding something so I left those in, but the one on `consumer_key` was just duplicating info from the `GroupInfo` model so I removed that one

Issue #1718, opened 2020-05-05
Don't crash when Canvas API developer key lacks a requested scope


seanh:
1. Log in to <https://hypothesis.instructure.com/> as an admin and edit the "localhost (make devdata)" developer key (<kbd>Admin</kbd> &rarr; <kbd>University of Hypothesis</kbd> &rarr; <kbd>Developer keys</kbd> &rarr; <kbd>Show all keys</kbd>, then find the "localhost (make devdata)" key and click its <kbd>Edit</kbd> icon):

   1. Toggle <kbd>Enforce Scopes</kbd> on
   2. Un-check one of the scopes that our app needs. For example <samp>Courses > url:GET|/api/v1/files/:id/public_url</samp>
   3. Click <kbd>Save</kbd>

   **Note:** this will edit the shared localhost dev key for everyone, it'll break localhost Canvas testing for everyone.

2. Delete any existing dev tokens from your DB:

   ```terminal
   tox -qe docker-compose -- exec postgres psql -U postgres -c 'delete from oauth2_token;'
   ```

2. Go to Test Course 1 and launch a Canvas Files test assignment

3. It should open the Canvas API authorization popup window, which will crash:

   ![Screenshot from 2020-05-05 13-00-20](https://user-images.githubusercontent.com/22498/81063793-65c56b00-8ed0-11ea-8ccd-b6c679353a13.png)

What's happened is that Canvas has redirected the popup window back to our Canvas OAuth callback endpoint (`http://localhost:8001/canvas_oauth_callback?error=invalid_scope&error_description=A+requested+scope+is+invalid...`) with an error message in the URL query params. And our callback endpoint has crashed:

```
Traceback (most recent call last):
  File "/home/seanh/Projects/lms/.tox/dev/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 134, in handle
    self.handle_request(listener, req, client, addr)
  File "/home/seanh/Projects/lms/.tox/dev/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 175, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/home/seanh/Projects/lms/.tox/dev/lib/python3.6/site-packages/sentry_sdk/integrations/pyramid.py", line 123, in sentry_patched_wsgi_call
    environ, start_response
  File "/home/seanh/Projects/lms/.tox/dev/lib/python3.6/site-packages/sentry_sdk/integrations/wsgi.py", line 129, in __call__
[...]

PR #1721, opened 2020-05-05
Fix the error page when our Canvas developer key lacks required scopes

seanh:
Fixes https://github.com/hypothesis/lms/issues/1718

**Note**: You'll want to delete all your Canvas API access tokens from your dev DB when testing this, to force it to open the authorization popup window to try to get a new token. Run: `tox -qe docker-compose -- exec postgres psql -U postgres -c 'delete from oauth2_token;'`

**Note**: We currently only use the Canvas API for Canvas Files assignments, so you'll want to launch a Canvas Files test assignment to test this.

When creating or launching a Canvas Files assignment, if we don't already have a working Canvas API access token for the user, we open a popup window to get the user to authorize us with Canvas and save a new access token to the DB.

There are two ways that this authorization popup window dance can fail:

1. Canvas might redirect the popup window back to us successfully with an authorization code in a query param but, when we try to exchange the authorization code for an access token this server-to-server request to the Canvas API fails.

   The easiest way to simulate this is to hack our `access_token_request()` function to send invalid requests to Canvas, which will force an error response from the Canvas API. For example:

   ```diff
   diff --git a/lms/services/_helpers/canvas_api.py b/lms/services/_helpers/canvas_api.py
   index 66fc16bb..1b54206c 100644
   --- a/lms/services/_helpers/canvas_api.py
   +++ b/lms/services/_helpers/canvas_api.py
   @@ -68,7 +68,7 @@ class CanvasAPIHelper:
                params={
                    "grant_type": "authorization_code",
                    "client_id": self._client_id,
   -                "client_secret": self._client_secret,
   +                "client_secret": "wrong",
                    "redirect_uri": self._redirect_uri,
                    "code": authorization_code,
                    "replace_tokens": True,
   ```

   When this happens the popup window will show an error page with a generic error message and <kbd>Cancel</kbd> (closes window) and <kbd>Try again</kbd> (redirects the popup window back to Canvas again) buttons:

   Screenshot from 2020-05-05 19-50-41

   **This error page is pre-existing behavior**, it's preserved by this PR but wasn't created by this PR.

2. If our Canvas developer key is missing one or more of the Canvas API scopes that we request, then Canvas redirects the popup window back to us with an error message instead of the authorization code in the query params. Previously we would crash when this happened (https://github.com/hypothesis/lms/issues/1718), now we show an error page.

   To simulate this you need to remove a required scope from our test dev key. Log in to <https://hypothesis.instructure.com/> as an admin and edit the "localhost (make devdata)" developer key (<kbd>Admin</kbd> &rarr; <kbd>University of Hypothesis</kbd> &rarr; <kbd>Developer keys</kbd> &rarr; <kbd>Show all keys</kbd>, then find the "localhost (make devdata)" key and click its <kbd>Edit</kbd> icon):

   1. Toggle <kbd>Enforce Scopes</kbd> on
   2. Un-check one of the scopes that our app needs. For example <samp>Courses > url:GET|/api/v1/files/:id/public_url</samp>
   3. Click <kbd>Save</kbd>

   **Note:** this will edit the shared localhost dev key for everyone, it'll break localhost Canvas testing for everyone.

   Then try to launch a Hypothesis assignment and you should see an error page with a specific error message listing the scopes we need. This error page just has a <kbd>Close</kbd> button, no <kbd>Try again</kbd>, because we've unavoidably lost the login session within the popup window. (The app in the underlying Canvas tab from whence the popup window came will still be working fine though.)

   Screenshot from 2020-05-05 19-56-09

## Known Issue

There's a CSS issue when you expand the **Details**. I'll try to get a frontend dev to fix this in a follow-up PR:

Screenshot from 2020-05-05 19-56-40

### DIFF
--- a/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
+++ b/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
@@ -8,7 +8,7 @@
 
       <p>Something went wrong when authorizing Hypothesis.</p>
 
-      <p>If the problem persists
+      <p>To get help from Hypothesis
       <a href="mailto:support@hypothes.is?subject=Hypothesis%20LMS%20Support" target="_blank" rel="noopener noreferrer">send us an email</a>
       or <a href="https://web.hypothes.is/get-help/" target="_blank" rel="noopener noreferrer">open a support ticket</a>.
       You can also visit our <a href="https://web.hypothes.is/help/" target="_blank" rel="noopener noreferrer"> help documents</a>.</p>
--- a/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
+++ b/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
@@ -15,16 +15,23 @@
 
       <div class="u-stretch"></div>
       <div class="Dialog__actions">
-        <button class="Button Button--cancel"
-                type="button"
-                onclick="window.close()">
-          Cancel
-        </button>
-        <button class="Button"
-                type="button"
-                onclick="window.location.href = '{{ authorize_url }}'">
-          Try again
-        </button>
+        {% if authorize_url %}
+          <button class="Button Button--cancel" type="button"
+                  onclick="window.close()">
+            Cancel
+          </button>
+          <button class="Button"
+                  type="button"
+                  onclick="window.location.href = '{{ authorize_url }}'">
+            Try again
+          </button>
+        {% else %}
+          <button class="Button"
+                  type="button"
+                  onclick="window.close()">
+            Close
+          </button>
+        {% endif %}
       </div>
     </div>
   </div>
--- a/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
+++ b/lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
@@ -4,14 +4,41 @@
   <div class="Dialog__background" style="z-index: 1;"></div>
   <div class="Dialog__container" style="z-index: 2;">
     <div class="Dialog__content LMSFilePicker__dialog">
-      <h1 class="Dialog__title">Authorization failed</h1>
+      {% if invalid_scope %}
+        <h1 class="Dialog__title">Developer Key Scopes Missing</h1>
 
-      <p>Something went wrong when authorizing Hypothesis.</p>
+        <p>
+          A Canvas admin needs to edit Hypothesis's developer key and add these
+          scopes:
+        </p>
+
+        <ol>
+          {% for scope in scopes %}
+            <li><code>{{ scope }}</code></li>
+          {% endfor %}
+        </ol>
+
+        <p>For more information see:
+        <a target="_blank" rel="noopener noreferrer" href="https://github.com/hypothesis/lms/wiki/Canvas-API-Endpoints-Used-by-the-Hypothesis-LMS-App">Canvas API Endpoints Used 
+y the Hypothesis LMS App</a>.
+        </p>
+      {% else %}
+        <h1 class="Dialog__title">Authorization Failed</h1>
+
+        <p>Something went wrong when authorizing Hypothesis.</p>
+      {% endif %}
 
       <p>To get help from Hypothesis
       <a href="mailto:support@hypothes.is?subject=Hypothesis%20LMS%20Support" target="_blank" rel="noopener noreferrer">send us an email</a>
       or <a href="https://web.hypothes.is/get-help/" target="_blank" rel="noopener noreferrer">open a support ticket</a>.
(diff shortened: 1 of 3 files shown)


lms/views/api/canvas/authorize.py
```diff
@@ -20,25 +20,33 @@
...
+
+    #: The Canvas API scopes that we request.
+    scopes = (
+        "url:GET|/api/v1/courses/:course_id/files",
+        "url:GET|/api/v1/files/:id/public_url",
+    )
+
     def __init__(self, request):
         self.request = request
-        self.ai_getter = request.find_service(name="ai_getter")
```
seanh, line 25: `request.find_service(name="ai_getter")` here in the `__init__()` was crashing because, when the view is `CanvasAPIAuthorizeViews.oauth2_redirect_error()`, `request.lti_user` is `None` and this causes `ai_getter`'s service factory to crash. Moved this line into the one specific view method that actually uses `ai_getter` instead. That fixes the crash

lms/views/api/canvas/authorize.py
```diff
@@ -73,13 +81,18 @@ def oauth2_redirect(self):
...
-            "authorize_url": self.request.route_url(
-                "canvas_api.authorize", _query=[("authorization", authorization_param)]
+        if self.request.lti_user:
+            authorization_param = BearerTokenSchema(self.request).authorization_param(
+                self.request.lti_user
             )
-        }
+            template_variables["authorize_url"] = self.request.route_url(
+                "canvas_api.authorize", _query=[("authorization", authorization_param)],
+            )
```
seanh, line 96: If we're still logged in (`request.lti_user` is still set) then sent the `authorize_url` so that we remain logged in. This makes the <kbd>Try again</kbd> button keep working. But sometimes we've already lost `request.lti_user` by this point and can't do this

tests/unit/lms/views/api/canvas/authorize_test.py
```diff
@@ -8,45 +9,41 @@
 
 
 class TestAuthorize:
-    def test_it_redirects_to_the_right_Canvas_endpoint(
-        self, ai_getter, pyramid_request
-    ):
-        response = CanvasAPIAuthorizeViews(pyramid_request).authorize()
+    def test_it_redirects_to_the_right_Canvas_endpoint(self, ai_getter, views):
+        response = views.authorize()
```
seanh, line 13: I added this `views` fixture, which simplifies a lot of these tests

lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
```diff
@@ -4,27 +4,61 @@
...
-      <p>Something went wrong when authorizing Hypothesis.</p>
+        <p>
+          A Canvas admin needs to edit Hypothesis's developer key and add these
+          scopes:
+        </p>
 
-      <p>If the problem persists
+        <ol>
+          {% for scope in scopes %}
+            <li><code>{{ scope }}</code></li>
```
LMS007, line 17: If this is not code, I would recommend `<pre>{{ scope }}</pre>`

jon-betts:
Just finished a manual test. All the behavior mentioned above is working for me locally.

tests/unit/lms/views/api/canvas/authorize_test.py
```diff
@@ -110,6 +102,66 @@ def test_it_passes_authorize_url_to_the_template(
...
+
+        assert template_variables["details"] == expected_details
+
+    def test_it_passes_our_required_API_scopes_to_the_template(self, views):
+        template_variables = views.oauth2_redirect_error()
+
+        assert template_variables["scopes"] == (
+            "url:GET|/api/v1/courses/:course_id/files",
+            "url:GET|/api/v1/files/:id/public_url",
+        )
```
jon-betts, line 158:
```suggestion
        assert template_variables["scopes"] == views.scopes
```

Do we want to say the exact scopes, or just that we pass through "the" scopes? It means not changing this in future if we do, but it's less specific. I lean towards just asserting they are the same, otherwise our tests tend towards an MD5 of our code.

tests/unit/lms/views/api/canvas/authorize_test.py
```diff
@@ -110,6 +102,66 @@ def test_it_passes_authorize_url_to_the_template(
...
+        template_variables = views.oauth2_redirect_error()
+
+        assert template_variables["invalid_scope"] == invalid_scope
+
+    @pytest.mark.parametrize(
+        "params,expected_details",
+        [
+            (
+                {"error_description": mock.sentinel.error_description},
+                mock.sentinel.error_description,
```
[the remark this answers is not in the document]
seanh: Yeah I'm fine with that

PR #1725, opened 2020-05-06
Split the Canvas API into a data part and authentication part

jon-betts:
This is to prepare for splitting these into separate files which work together.

This PR:

 * Keeps everything in the same files
 * Moves all the basic calling machinery and authentication to a base class
 * Has all the data calls implemented on top of that
 * Removes some duplicate mentions (every function can raise the same errors)

### DIFF
--- a/lms/services/canvas_api.py
+++ b/lms/services/canvas_api.py
@@ -31,7 +31,18 @@ class _SectionSchema(Schema):
     name = fields.String(required=True)
 
 
-class CanvasAPIClient:
+class _CanvasAPIAuthenticatedClient:
+    """
+    A client for making authenticated calls to the Canvas API.
+
+    All methods in the authenticated client may raise:
+
+    :raise CanvasAPIAccessTokenError: if the request fails because our
+         Canvas API access token for the user is missing, expired, or has
+         been deleted
+    :raise CanvasAPIServerError: if the request fails for any other reason
+    """
+
     def __init__(self, _context, request):
         ai_getter = request.find_service(name="ai_getter")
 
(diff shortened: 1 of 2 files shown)


lms/services/canvas_api.py
```diff
@@ -115,17 +124,130 @@ def validate_quantity(self, expires_in):  # pylint:disable=no-self-use
             if expires_in <= 0:
                 raise marshmallow.ValidationError("expires_in must be greater than 0")
 
-    def authenticated_users_sections(self, course_id):
+    def make_authenticated_request(self, method, path, schema, params=None):
```
jon-betts, line 127: These functions have been lifted up verbatim from the bottom of the file

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -244,6 +244,10 @@ def data_method(self, request, api_client):
 
         return lambda: getattr(api_client, method)(*args)
 
+    @pytest.fixture
+    def api_client(self, pyramid_request):
+        return CanvasAPIClient(sentinel.context, pyramid_request)
```
jon-betts, line 249: Made a distinction between the public `api_client` and the `base_client`. This should help us separate the tests later on if we split the files up.

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -465,18 +469,18 @@ def access_token_no_refresh(self, db_session, access_token_fields):
         return access_token
 
     @pytest.fixture
-    def api_client(self, api_client):
-        with mock.patch.object(api_client, "_get_refreshed_token"):
-            with mock.patch.object(api_client, "_validated_response"):
-                yield api_client
+    def base_client(self, base_client):
+        with mock.patch.object(base_client, "_get_refreshed_token"):
+            with mock.patch.object(base_client, "_validated_response"):
```
seanh, line 474: I think if some of the tests for `FooClass` are having to patch some of `FooClass`'s own methods, that suggests that there's more than one separate class here. I wonder if `_get_refreshed_token` and `_validated_response` belong in separate collaborator classes, or just as separate standalone functions.

Maybe for a later PR

jon-betts:
Ok, from a chat with @seanh we're going ahead with this for now. Any fixes can come later. It works, it's just not ideal.

I want to split it out into files anyway, which will make any future work more sane.

PR #1729, opened 2020-05-06
Add pagination support to the Canvas API

jon-betts:
Add support for following "rel=next" links provided by Canvas for pagination.

This will automatically apply to all end-points which use a `many=True` schema:

 * This will add the `per_page=` query param to a large value
 * Follow next links up to a pre-defined limit
 * If a query fails for authentication reasons, it will be started from scratch

This only currently applies to `list_files` and `course_sections` as the other sections calls do not support pagination (I think).

This also enables re-use of the HTTP session which should speed up repeated calls. Based on local testing I had rough performance of:

 * 600ms per call for calls without the session
 * 400ms up front, then 200ms per call with a session

This will effect paginated calls, but also any situations where we attempt to retry and make multiple calls.

----

This should be merged after: https://github.com/hypothesis/lms/pull/1725

### DIFF
--- a/tests/unit/lms/services/canvas_api_test.py
+++ b/tests/unit/lms/services/canvas_api_test.py
@@ -1,19 +1,19 @@
 import json
 from io import BytesIO
 from unittest import mock
-from unittest.mock import call, sentinel
+from unittest.mock import call, create_autospec, sentinel
 
 import pytest
 from _pytest.mark import param
 from h_matchers import Any
 from h_matchers.decorator import fluent_entrypoint
 from h_matchers.matcher.core import Matcher
-from requests import PreparedRequest, RequestException, Response
+from requests import PreparedRequest, Request, RequestException, Response
 
 from lms.models import ApplicationInstance, OAuth2Token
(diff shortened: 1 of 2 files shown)


jon-betts:
Getting some coverage weirdness. It's saying we aren't covering fixtures we clearly need...

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -519,3 +600,23 @@ def access_token(db_session, access_token_fields):
...
+    return Schema
+
+
+@pytest.fixture
+def PaginatedSchema(Schema):
+    Schema.many = True
+
+    Schema.return_value.parse.side_effect = (
+        [f"item_{i}"] for i in range(100)
+    )  # pragma: no cover
```
seanh, line 620: Is there something weird about setting `side_effect` to a generator?

jon-betts: No it's intended behavior. It causes the side effect to return a new item each time you call it.

PR #1730, opened 2020-05-06
DB migration to add ApplicationInstance.settings column

seanh:
This is the DB migration for https://github.com/hypothesis/lms/pull/1731

### DIFF
--- a/lms/migrations/versions/f0859cd029fe_add_application_instances_canvas_.py
+++ b/lms/migrations/versions/f0859cd029fe_add_application_instances_canvas_.py
@@ -1,0 +1,29 @@
+"""
+Add application_instances.canvas_sections_enabled column.
+
+Revision ID: f0859cd029fe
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/f0859cd029fe_add_application_instances_canvas_.py
```diff
@@ -0,0 +1,29 @@
...
+down_revision = "37710e6bcb66"
+
+
+def upgrade():
+    op.add_column(
+        "application_instances",
+        sa.Column(
+            "canvas_sections_enabled",
+            sa.Boolean(),
+            server_default=sa.sql.expression.false(),
```
seanh, line 22: The `server_default` will set the column to `false` for all existing rows when the migration is run, which is what we want: we want sections disabled for all pre-existing application instances

Issue #1724, opened 2020-05-06
Enable sections by default for new application instances (feature-flagged)


seanh:
New instances should either get sections or not depending on the feature flag.

Issue #1697, opened 2020-04-27
Don't use sections for pre-existing application instances


seanh:
For pre-existing application instances that were created before the sections feature was released, fall back to using course groups instead of sections.

Issue #1696, opened 2020-04-27
Show error page if the dev key lacks scopes for sections


seanh:
This is going to be handled by showing an error page explaining that we do need the scopes, rather than by falling back to course groups. If the application instance has a developer key, but the developer key lacks required scopes, the app won't work

seanh:
Note to self: also need to deal with the _Allow Include Parameters_ checkbox

Issue #1695, opened 2020-04-27
Don't use sections if the application instance doesn't have a developer key


seanh:
Fall back to course groups

PR #1731, opened 2020-05-06
Enable and disable sections in various circumstances

seanh:
Depends on this DB migration: https://github.com/hypothesis/lms/pull/1730/

## Intended Behaviour

1. Pre-existing application instances (either created before the DB migration is run and before this branch is deployed; _or_ created after the DB migration and deploying this branch but before the `"section_groups"` feature flag is enabled) should **not** get the sections feature.

2. Pre-existing application instances that have a developer key should continue to get the Canvas Files feature but **not** get the sections feature.

   These pre-existing instances shouldn't request the sections API scopes, so they should keep working with a scoped dev key that only has the Files scopes.

   If the dev key lacks even the Files scopes then these instances will show an error page when you try to create or launch a Canvas Files assignment. The error page will tell the user that we need both the Files **and** Sections scopes.

3. **New** application instances (created after the DB migration has been run, this branch has been deployed, and the feature flag has been turned on) should get both the Files and Sections features.

   These new instances will always request both the Files and Sections scopes, and will show an error page if any of the scopes are missing. The error page will show whenever anyone tries to create or launch any assignment, and will ask for both the Files and Sections scopes.

4. New app instances that don't have a developer key won't get either the files or sections features.

## Testing

* To test this you need to cherry-pick frontend sections support:

  ```terminal
  git cherry-pick make-the-frontend-call-the-sync-api
  ```

* You'll want to repeatedly delete all the OAuth 2 tokens from your DB when testing this, to force it to re-authorize with Canvas (changing the developer key's scopes in Canvas also often forces a re-authorize, but not always):

   ```terminal
  tox -qe docker-compose -- exec postgres psql -U postgres -c 'delete from oauth2_token;'
  ```

### Test Cases

* App instances created before the DB migration:

  The `"section_groups"` feature flag should be **off** when creating the app instances for this test, and then using the app instances should be tested with the flag both **off** and **on**.

  Create the app instances on the master branch without the DB migration having been run so there's no `ApplicationInstance.canvas_sections_enabled` column. Then run the DB migration, then switch to this branch, and then test the app instance's behaviour.

  - App instance created without a developer key:
    - [x] Should work with no Canvas files or sections feature
  - App instance created with a developer key:
    - [x] Should have the Canvas files feature but not sections
    - [x] Should work with a scoped dev key even if the key lacks the sections scopes
    - [x] With a scoped dev key that lacks the _files_ scopes, should show an error page asking for both the sections and files scopes
  - [x] All of the above should work with the feature flag both off and on

* App instances created after the DB migration, but before the feature flag has been turned on:

  The `"section_groups"` feature flag should be **off** when creating the app instances for this test, and then using the app instances should be tested with the flag both **off** and **on**.

  Use this branch for the entire test (before creating the app instances and testing them).

  - App instance created without a developer key:
    - [x] Should work with no Canvas files or sections feature
  - App instance created with a developer key:
    - [x] Should have the Canvas files feature but not sections
    - [x] Should work with a scoped dev key even if the key lacks the sections scopes
    - [x] With a scoped dev key that lacks the _files_ scopes, should show an error page asking for both the sections and files scopes
  - [x] All of the above should work with the feature flag both off and on

* App instances created after the DB migration and after the feature flag has been turned on:

    The `"section_groups"` feature flag should be **on** for this test.

  Use this branch for the entire test (before creating the app instances and testing them).

  - App instance created without a developer key:
    - [x] Should work with no Canvas files or sections feature
  - App instance created with a developer key:
    - [x] Should have both the Canvas files and sections features
    - [x] Should work with a scoped dev key as long as the key has both the files and sections scopes
    - [x] With a scoped dev key that lacks the sections scopes, should show an error page asking for both the sections and files scopes
    - [x] With a scoped dev key that lacks the files scopes, should show an error page asking for both the sections and files scopes

## Implementation

1. Adds a new boolean DB column `ApplicationInstance.canvas_sections_enabled`. This is exactly the same as the existing `ApplicationInstance.provisioning`, which also toggles a feature

2. _If the `"section_groups"` feature flag is on_ enables sections by default for new application instances that have a developer key.

   Fixes https://github.com/hypothesis/lms/issues/1724.

   This is done when the `/welcome` form is submitted. Logic is added to the `create_application_instance()` view (which is the view that receives the form submission) to set `canvas_sections_enabled` to `True` if the feature flag is on and a dev key was submitted.

3. Disables sections if the application instance doesn't have a developer key. Fixes https://github.com/hypothesis/lms/issues/1695

4. Disables sections for pre-existing application instances.

   Fixes https://github.com/hypothesis/lms/issues/1697.

   This is done in two ways:

   1. The DB migration in https://github.com/hypothesis/lms/pull/1730/ will set `canvas_sections_enabled` to `False` for all preexisting rows
   2. The logic in the `create_application_instance()` view mentioned above will set `canvas_sections_enabled` to `False` for all new application instances created after the DB migration, as long as the `"section_groups"` feature flag is off. (Once the flag is on `create_application_instance()` starts setting `canvas_sections_enabled` to `True` for new instances.)

5. Adds the new Canvas API scopes that we need for sections, to the list of scopes that we request from the Canvas API.

   This will trigger the recently implemented scopes error page, if the dev key lacks any of the new scopes.

   Fixes https://github.com/hypothesis/lms/issues/1696

   The new scopes have already been documented here https://github.com/hypothesis/lms/wiki/Canvas-API-Endpoints-Used-by-the-Hypothesis-LMS-App

6. Does not request the sections scopes from the Canvas API if the application instance doesn't have the `canvas_sections_enabled` flag on. Even if the app instance does have a developer key. Instead requests the Files scopes only.

   This is so that preexisting app instances that have scoped dev keys with only the Files scopes, can continue working with only the Files feature (not Sections), and won't start showing an error page because we're now requesting the Sections scopes and the dev key doesn't have them.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -4,6 +4,8 @@ from datetime import datetime
 import sqlalchemy as sa
 from Crypto import Random
 from Crypto.Cipher import AES
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
 
@@ -30,12 +32,23 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
-    canvas_sections_enabled = sa.Column(
-        sa.Boolean(),
-        default=False,
-        server_default=sa.sql.expression.false(),
-        nullable=False,
-    )
+    _settings = sa.Column("settings", MutableDict.as_mutable(JSONB))
+
+    @property
+    def settings(self):
+        """
+        Get a settings object which can get and set grouped settings.
+
+        The object can be queried with `settings.get(group, key)` or
+        `settings.set(group, key, value)`.
+
+        :return: _ApplicationSettings object.
+        """
+
+        if self._settings is None:
+            self._settings = {}
+
+        return _ApplicationSettings(self._settings)
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
@@ -56,13 +69,7 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
 
     @classmethod
     def build_from_lms_url(  # pylint:disable=too-many-arguments
-        cls,
-        lms_url,
-        email,
-        developer_key,
-        developer_secret,
-        encryption_key,
-        canvas_sections_enabled,
+        cls, lms_url, email, developer_key, developer_secret, encryption_key, settings,
     ):
         """Instantiate ApplicationInstance with lms_url."""
         encrypted_secret = developer_secret
@@ -82,7 +89,7 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
             developer_secret=encrypted_secret,
             aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
-            canvas_sections_enabled=canvas_sections_enabled,
+            _settings=settings,
         )
 
 
@@ -105,3 +112,30 @@ def _encrypt_oauth_secret(oauth_secret, key, init_v):
     """Encrypt an oauth secrety via AES encryption."""
     cipher = AES.new(key, AES.MODE_CFB, init_v)
     return cipher.encrypt(oauth_secret)
+
+
+class _ApplicationSettings:
+    """Model for accessing and updating application settings."""
+
+    def __init__(self, data):
+        self.data = data
+
+    def get(self, group, key):
+        """
+        Get a specific settings or None if it doesn't exist.
+
+        :param group: The name of the group of settings
+        :param key: The key in that group
+        :return: The value or None
+        """
+        return self.data.get(group, {}).get(key)
+
+    def set(self, group, key, value):
+        """
+        Set a specific setting in a group.
+
+        :param group: The name of the group of settings
+        :param key: The key in that group
+        :param value: The value to set
+        """
+        self.data.setdefault(group, {})[key] = value
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -56,12 +56,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
 
     @classmethod
     def build_from_lms_url(  # pylint:disable=too-many-arguments
-        cls, lms_url, email, developer_key, developer_secret, encryption_key=None
+        cls, lms_url, email, developer_key, developer_secret, encryption_key
     ):
         """Instantiate ApplicationInstance with lms_url."""
         encrypted_secret = developer_secret
         aes_iv = None
-        if encryption_key is not None and developer_secret and developer_key:
+        if developer_secret and developer_key:
             aes_iv = _build_aes_iv()
             encrypted_secret = _encrypt_oauth_secret(
                 developer_secret, encryption_key, aes_iv
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -56,7 +56,13 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
 
     @classmethod
     def build_from_lms_url(  # pylint:disable=too-many-arguments
-        cls, lms_url, email, developer_key, developer_secret, encryption_key
+        cls,
+        lms_url,
+        email,
+        developer_key,
+        developer_secret,
+        encryption_key,
+        canvas_sections_enabled,
     ):
         """Instantiate ApplicationInstance with lms_url."""
         encrypted_secret = developer_secret
@@ -76,6 +82,7 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
             developer_secret=encrypted_secret,
             aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
+            canvas_sections_enabled=canvas_sections_enabled,
         )
 
 
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -30,6 +30,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
+    canvas_sections_enabled = sa.Column(
+        sa.Boolean(),
+        default=False,
+        server_default=sa.sql.expression.false(),
+        nullable=False,
+    )
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -106,7 +106,7 @@ class LTILaunchResource:
 
         ai_getter = self._request.find_service(name="ai_getter")
 
-        if not ai_getter.canvas_sections_enabled():
+        if not ai_getter.developer_key() or not ai_getter.canvas_sections_enabled():
             return False
 
         return self._request.feature("section_groups")
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -104,4 +104,9 @@ class LTILaunchResource:
         if not self.is_canvas:
             return False
 
+        ai_getter = self._request.find_service(name="ai_getter")
+
+        if not ai_getter.canvas_sections_enabled():
+            return False
+
         return self._request.feature("section_groups")
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -99,14 +99,8 @@ class LTILaunchResource:
         return self._request.parsed_params.get("custom_canvas_api_domain")
 
     @property
-    def should_use_section_groups(self):
-        """Return True if section groups rather than a course group should be used."""
-        if not self.is_canvas:
-            return False
-
+    def canvas_sections_enabled(self):
+        """Return True if Canvas sections is enabled for this request."""
         ai_getter = self._request.find_service(name="ai_getter")
 
-        if not ai_getter.developer_key() or not ai_getter.canvas_sections_enabled():
-            return False
-
-        return self._request.feature("section_groups")
+        return self.is_canvas and ai_getter.canvas_sections_enabled()
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -70,6 +70,12 @@ class ApplicationInstanceGetter:
             provisioning = False
         return provisioning
 
+    def canvas_sections_enabled(self):
+        try:
+            return self._get_by_consumer_key().canvas_sections_enabled
+        except ConsumerKeyError:
+            return False
+
     def shared_secret(self):
         """
         Return the LTI/OAuth 1 shared secret for the current request.
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -81,7 +81,10 @@ class ApplicationInstanceGetter:
         except ConsumerKeyError:
             return False
 
-        return bool(app_instance.developer_key and app_instance.canvas_sections_enabled)
+        return bool(
+            app_instance.developer_key
+            and app_instance.settings.get("canvas", "sections_enabled")
+        )
 
     def shared_secret(self):
         """
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -11,10 +11,11 @@ __all__ = ["ApplicationInstanceGetter"]
 class ApplicationInstanceGetter:
     """Methods for getting properties from application instances."""
 
-    def __init__(self, db, aes_secret, consumer_key):
+    def __init__(self, db, aes_secret, consumer_key, section_groups_feature_flag):
         self._db = db
         self._aes_secret = aes_secret
         self._consumer_key = consumer_key
+        self._section_groups_feature_flag = section_groups_feature_flag
 
     def developer_key(self):
         """
@@ -71,10 +72,16 @@ class ApplicationInstanceGetter:
         return provisioning
 
     def canvas_sections_enabled(self):
+        """Return True if the application instance has Canvas sections is enabled."""
+        if not self._section_groups_feature_flag:
+            return False
+
         try:
-            return self._get_by_consumer_key().canvas_sections_enabled
+            app_instance = self._get_by_consumer_key()
         except ConsumerKeyError:
             return False
+
+        return bool(app_instance.developer_key and app_instance.canvas_sections_enabled)
 
     def shared_secret(self):
         """
@@ -114,4 +121,5 @@ def application_instance_getter_service_factory(_context, request):
         request.db,
         request.registry.settings["aes_secret"],
         request.lti_user.oauth_consumer_key,
+        section_groups_feature_flag=request.feature("section_groups"),
     )
--- a/lms/views/api/canvas/authorize.py
+++ b/lms/views/api/canvas/authorize.py
@@ -20,25 +20,31 @@ from lms.validation.authentication import BearerTokenSchema, CanvasOAuthCallback
 
 @view_defaults(request_method="GET", route_name="canvas_oauth_callback")
 class CanvasAPIAuthorizeViews:
+
+    # The Canvas API scopes that we need for our Canvas Files feature.
+    files_scopes = (
+        "url:GET|/api/v1/courses/:course_id/files",
+        "url:GET|/api/v1/files/:id/public_url",
+    )
+
+    # The Canvas API scopes that we need for our Sections feature.
+    sections_scopes = (
+        "url:GET|/api/v1/courses/:id",
+        "url:GET|/api/v1/courses/:course_id/sections",
+        "url:GET|/api/v1/courses/:course_id/users/:id",
+    )
+
     def __init__(self, request):
         self.request = request
-
-        #: The Canvas API scopes that we request.
-        self.scopes = (
-            "url:GET|/api/v1/courses/:course_id/files",
-            "url:GET|/api/v1/files/:id/public_url",
-        )
-
-        if request.feature("section_groups"):
-            self.scopes = self.scopes + (
-                "url:GET|/api/v1/courses/:id",
-                "url:GET|/api/v1/courses/:course_id/sections",
-                "url:GET|/api/v1/courses/:course_id/users/:id",
-            )
 
     @view_config(permission="canvas_api", route_name="canvas_api.authorize")
     def authorize(self):
         ai_getter = self.request.find_service(name="ai_getter")
+
+        if ai_getter.canvas_sections_enabled():
+            scopes = self.files_scopes + self.sections_scopes
+        else:
+            scopes = self.files_scopes
 
         authorize_url = urlunparse(
             (
@@ -52,7 +58,7 @@ class CanvasAPIAuthorizeViews:
                         "response_type": "code",
                         "redirect_uri": self.request.route_url("canvas_oauth_callback"),
                         "state": CanvasOAuthCallbackSchema(self.request).state_param(),
-                        "scope": " ".join(self.scopes),
+                        "scope": " ".join(scopes),
                     }
                 ),
                 "",
@@ -90,7 +96,7 @@ class CanvasAPIAuthorizeViews:
         template_variables = {
             "invalid_scope": self.request.params.get("error") == "invalid_scope",
             "details": self.request.params.get("error_description"),
-            "scopes": self.scopes,
+            "scopes": self.files_scopes + self.sections_scopes,
         }
 
         if self.request.lti_user:
--- a/lms/views/api/canvas/authorize.py
+++ b/lms/views/api/canvas/authorize.py
@@ -20,15 +20,21 @@ from lms.validation.authentication import BearerTokenSchema, CanvasOAuthCallback
 
 @view_defaults(request_method="GET", route_name="canvas_oauth_callback")
 class CanvasAPIAuthorizeViews:
-
-    #: The Canvas API scopes that we request.
-    scopes = (
-        "url:GET|/api/v1/courses/:course_id/files",
-        "url:GET|/api/v1/files/:id/public_url",
-    )
-
     def __init__(self, request):
         self.request = request
+
+        #: The Canvas API scopes that we request.
+        self.scopes = (
+            "url:GET|/api/v1/courses/:course_id/files",
+            "url:GET|/api/v1/files/:id/public_url",
+        )
+
+        if request.feature("section_groups"):
+            self.scopes = self.scopes + (
+                "url:GET|/api/v1/courses/:id",
+                "url:GET|/api/v1/courses/:course_id/sections",
+                "url:GET|/api/v1/courses/:course_id/users/:id",
+            )
 
     @view_config(permission="canvas_api", route_name="canvas_api.authorize")
     def authorize(self):
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -28,6 +28,7 @@ def create_application_instance(request):
         developer_key,
         developer_secret,
         request.registry.settings["aes_secret"],
+        canvas_sections_enabled=developer_key and request.feature("section_groups"),
     )
     request.db.add(instance)
 
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -28,7 +28,13 @@ def create_application_instance(request):
         developer_key,
         developer_secret,
         request.registry.settings["aes_secret"],
-        canvas_sections_enabled=developer_key and request.feature("section_groups"),
+        settings={
+            "canvas": {
+                "sections_enabled": bool(
+                    developer_key and request.feature("section_groups")
+                )
+            }
+        },
     )
     request.db.add(instance)
 
--- a/tests/unit/lms/models/test_application_instance.py
+++ b/tests/unit/lms/models/test_application_instance.py
@@ -33,6 +33,32 @@ class TestApplicationInstance:
 
         assert application_instance.provisioning is not None
 
+    def test_canvas_sections_enabled_defaults_to_False(
+        self, db_session, application_instance
+    ):
+        db_session.add(application_instance)
+
+        db_session.flush()
+
+        assert not application_instance.canvas_sections_enabled
+
+    def test_canvas_sections_can_be_enabled(self, db_session, application_instance):
+        application_instance.canvas_sections_enabled = True
+        db_session.add(application_instance)
+        db_session.flush()
+
+        assert application_instance.canvas_sections_enabled is True
+
+    def test_canvas_sections_enabled_is_not_nullable(
+        self, db_session, application_instance
+    ):
+        application_instance.canvas_sections_enabled = None
+        db_session.add(application_instance)
+
+        db_session.flush()
+
+        assert application_instance.canvas_sections_enabled is not None
+
     def test_consumer_key_cant_be_null(self, db_session, application_instance):
         application_instance.consumer_key = None
         db_session.add(application_instance)
--- a/tests/unit/lms/views/api/canvas/authorize_test.py
+++ b/tests/unit/lms/views/api/canvas/authorize_test.py
@@ -42,28 +42,26 @@ class TestAuthorize:
             "http://example.com/canvas_oauth_callback"
         ]
 
-    def test_it_includes_the_scope_in_a_query_param(self, views):
+    def test_it_includes_the_scopes_in_a_query_param(self, views):
         response = views.authorize()
 
         query_params = parse_qs(urlparse(response.location).query)
 
-        assert query_params["scope"] == [
-            "url:GET|/api/v1/courses/:course_id/files url:GET|/api/v1/files/:id/public_url"
-        ]
-
-    @pytest.mark.usefixtures("sections_enabled")
-    def test_it_requests_sections_scopes_if_the_sections_feature_flag_is_on(
-        self, views
-    ):
-        response = views.authorize()
-
-        query_params = parse_qs(urlparse(response.location).query)
         assert query_params["scope"] == [
             "url:GET|/api/v1/courses/:course_id/files "
             "url:GET|/api/v1/files/:id/public_url "
             "url:GET|/api/v1/courses/:id "
             "url:GET|/api/v1/courses/:course_id/sections "
             "url:GET|/api/v1/courses/:course_id/users/:id"
+        ]
+
+    @pytest.mark.usefixtures("sections_disabled")
+    def test_it_doesnt_request_sections_scopes_if_sections_is_disabled(self, views):
+        response = views.authorize()
+
+        query_params = parse_qs(urlparse(response.location).query)
+        assert query_params["scope"] == [
+            "url:GET|/api/v1/courses/:course_id/files url:GET|/api/v1/files/:id/public_url"
         ]
 
     def test_it_includes_the_state_in_a_query_param(
@@ -170,15 +168,6 @@ class TestOAuth2RedirectError:
         assert template_variables["scopes"] == (
             "url:GET|/api/v1/courses/:course_id/files",
             "url:GET|/api/v1/files/:id/public_url",
-        )
-
-    @pytest.mark.usefixtures("sections_enabled")
-    def test_it_passes_sections_scopes_if_the_feature_flag_is_on(self, views):
-        template_variables = views.oauth2_redirect_error()
-
-        assert template_variables["scopes"] == (
-            "url:GET|/api/v1/courses/:course_id/files",
-            "url:GET|/api/v1/files/:id/public_url",
             "url:GET|/api/v1/courses/:id",
             "url:GET|/api/v1/courses/:course_id/sections",
             "url:GET|/api/v1/courses/:course_id/users/:id",
@@ -215,11 +204,8 @@ def canvas_oauth_callback_schema(CanvasOAuthCallbackSchema):
 
 
 @pytest.fixture
-def sections_enabled(pyramid_request):
-    def feature(flag):
-        return flag == "section_groups"
-
-    pyramid_request.feature.side_effect = feature
+def sections_disabled(ai_getter):
+    ai_getter.canvas_sections_enabled.return_value = False
 
 
 pytestmark = pytest.mark.usefixtures("ai_getter", "canvas_api_client")
--- a/tests/unit/lms/views/api/canvas/authorize_test.py
+++ b/tests/unit/lms/views/api/canvas/authorize_test.py
@@ -51,6 +51,21 @@ class TestAuthorize:
             "url:GET|/api/v1/courses/:course_id/files url:GET|/api/v1/files/:id/public_url"
         ]
 
+    @pytest.mark.usefixtures("sections_enabled")
+    def test_it_requests_sections_scopes_if_the_sections_feature_flag_is_on(
+        self, views
+    ):
+        response = views.authorize()
+
+        query_params = parse_qs(urlparse(response.location).query)
+        assert query_params["scope"] == [
+            "url:GET|/api/v1/courses/:course_id/files "
+            "url:GET|/api/v1/files/:id/public_url "
+            "url:GET|/api/v1/courses/:id "
+            "url:GET|/api/v1/courses/:course_id/sections "
+            "url:GET|/api/v1/courses/:course_id/users/:id"
+        ]
+
     def test_it_includes_the_state_in_a_query_param(
         self,
         pyramid_request,
@@ -157,6 +172,18 @@ class TestOAuth2RedirectError:
             "url:GET|/api/v1/files/:id/public_url",
         )
 
+    @pytest.mark.usefixtures("sections_enabled")
+    def test_it_passes_sections_scopes_if_the_feature_flag_is_on(self, views):
+        template_variables = views.oauth2_redirect_error()
+
+        assert template_variables["scopes"] == (
+            "url:GET|/api/v1/courses/:course_id/files",
+            "url:GET|/api/v1/files/:id/public_url",
+            "url:GET|/api/v1/courses/:id",
+            "url:GET|/api/v1/courses/:course_id/sections",
+            "url:GET|/api/v1/courses/:course_id/users/:id",
+        )
+
 
 @pytest.fixture
 def views(pyramid_request):
@@ -185,6 +212,14 @@ def canvas_oauth_callback_schema(CanvasOAuthCallbackSchema):
     schema = CanvasOAuthCallbackSchema.return_value
     schema.state_param.return_value = "test_state"
     return schema
+
+
+@pytest.fixture
+def sections_enabled(pyramid_request):
+    def feature(flag):
+        return flag == "section_groups"
+
+    pyramid_request.feature.side_effect = feature
 
 
 pytestmark = pytest.mark.usefixtures("ai_getter", "canvas_api_client")
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -370,12 +370,12 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
         }
 
     def _groups(self):
-        if self._context.should_use_section_groups:
+        if self._context.canvas_sections_enabled:
             return "$rpc:requestGroups"
         return [self._context.h_group.groupid(self._authority)]
 
     def _sync_api(self):
-        if not self._context.should_use_section_groups:
+        if not self._context.canvas_sections_enabled:
             return None
 
         req = self._request
--- a/tests/unit/lms/models/application_instance_test.py
+++ b/tests/unit/lms/models/application_instance_test.py
@@ -2,6 +2,9 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 from lms.models import ApplicationInstance
+from lms.models.application_instance import _ApplicationSettings
+
+# pylint: disable=protected-access
 
 
 class TestApplicationInstance:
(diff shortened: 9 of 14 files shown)


lms/models/application_instance.py
```diff
@@ -30,6 +30,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
+    canvas_sections_enabled = sa.Column(
+        sa.Boolean(),
+        default=False,
+        server_default=sa.sql.expression.false(),
+        nullable=False,
+    )
```
seanh, line 38: Exactly the same as the `provisioning` column above

lms/models/application_instance.py
```diff
@@ -50,12 +56,18 @@ def get_by_consumer_key(cls, db, consumer_key):
...
+        developer_key,
+        developer_secret,
+        encryption_key,
+        canvas_sections_enabled,
     ):
         """Instantiate ApplicationInstance with lms_url."""
         encrypted_secret = developer_secret
         aes_iv = None
-        if encryption_key is not None and developer_secret and developer_key:
+        if developer_secret and developer_key:
```
seanh, line 70: I made `encryption_key` non-optional, it's never not passed

lms/views/api/canvas/authorize.py
```diff
@@ -20,16 +20,22 @@
...
+            "url:GET|/api/v1/courses/:course_id/files",
+            "url:GET|/api/v1/files/:id/public_url",
+        )
+
+        if request.feature("section_groups"):
+            self.scopes = self.scopes + (
+                "url:GET|/api/v1/courses/:id",
+                "url:GET|/api/v1/courses/:course_id/sections",
+                "url:GET|/api/v1/courses/:course_id/users/:id",
+            )
```
seanh, line 37: Once the feature flag is removed it'll just be one hard-coded list of scopes that we always request. If you have a developer key / if we're talking to the Canvas API for you at all, then we require both the Files and Sections scopes.

For now, though, we don't want to request the sections scopes when the feature flag is off.

lms/views/application_instances.py
```diff
@@ -28,6 +28,7 @@ def create_application_instance(request):
         developer_key,
         developer_secret,
         request.registry.settings["aes_secret"],
+        canvas_sections_enabled=developer_key and request.feature("section_groups"),
```
seanh, line 31: This logic can easily be moved into the model once it no longer depends on the feature flag

lms/models/application_instance.py
seanh: I don't want to get in to refactoring `ApplicationInstance` now, but some notes:

There's no such thing as a "Canvas application instance" or a "Moodle application instance" etc. There are just application instances, and one of the things that an application instance can have is a `canvas_sections_enabled` flag. They also have a `canvas_developer_key` and `canvas_developer_secret` but the `canvas_` prefix is missing from those Canvas-specific fields.

An application instance is just a thing that gets created when you fill in the https://lms.hypothes.is/welcome form. At the time when you submit the form we don't know whether you're going to use it in Canvas, Moodle, or something else. You don't even have to use it in just one: you could use it first in Canvas and then install it in Moodle.

What an application instance really is, is an LTI 1 consumer key and shared secret. You enter some info into `/welcome`--a domain name (that isn't really used for anything), email address (just for our records), and optionally a Canvas developer ID and key--and we generate a consumer key and shared secret for you and associate the bits of into you submitted with that consumer key and shared secret.

A correct domain model for this might be something like:

1. A `consumer_key` table (with `key` and `secret` columns)
2. Some registration info (domain and email) possibly in a separate table with a foreign key to the `consumer_key` table or possibly just in extra columns on the `consumer_key` table
3. A Canvas developer key (with `id` and `key` columns), possibly in a separate `canvas_developer_key` table with a foreign key to the `consumer_key` table or possibly just in extra columns on the `consumer_key` table

It might be worth a look through what the LTI specs and guides call this though, for naming and modeling ideas.

But the consumer keys (a.k.a. "application instances") themselves aren't LMS-specific. There aren't different types of application instance, like Canvas app instance and Moodle app instance etc.

lms/models/application_instance.py
seanh: This classmethod is a bit of a horror but I opted not to get in to refactoring it any more than that

lms/services/application_instance_getter.py
```diff
@@ -70,6 +70,12 @@ def provisioning_enabled(self):
             provisioning = False
         return provisioning
 
+    def canvas_sections_enabled(self):
+        try:
+            return self._get_by_consumer_key().canvas_sections_enabled
+        except ConsumerKeyError:
+            return False
```
seanh, line 77: I think the `ApplicationInstanceGetter` service is a bit silly and should just return the `ApplicationInstance` model itself to the calling code. But I don't want to get into refactoring it now

tests/unit/lms/views/api/canvas/authorize_test.py
```diff
@@ -187,4 +214,12 @@ def canvas_oauth_callback_schema(CanvasOAuthCallbackSchema):
     return schema
 
 
+@pytest.fixture
+def sections_enabled(pyramid_request):
+    def feature(flag):
+        return flag == "section_groups"
+
+    pyramid_request.feature.side_effect = feature
```
seanh, line 222: This fixture will be deleted when the feature flag is removed

lms/views/api/canvas/authorize.py
seanh: Urgh, just realised that if there's a preexisting application instance that does have a developer key, then it needs to request the Canvas files scopes but not the sections ones (even if the feature flag is on). So there's a little more work to do here

lms/resources/lti_launch.py
```diff
@@ -99,9 +99,8 @@ def custom_canvas_api_domain(self):
...
-    def should_use_section_groups(self):
-        """Return True if section groups rather than a course group should be used."""
-        if not self.is_canvas:
-            return False
+    def canvas_sections_enabled(self):
+        """Return True if Canvas sections is enabled for this request."""
+        ai_getter = self._request.find_service(name="ai_getter")
 
-        return self._request.feature("section_groups")
+        return self.is_canvas and ai_getter.canvas_sections_enabled()
```
seanh, line 106: During an LTI launch request we use `context.canvas_sections_enabled`. This ensures that even if the `"section_groups"` feature flag is on and the application instance has a Canvas developer key and has `canvas_sections_enabled` on, we _still_ won't try to use sections if we're being launched by a non-Canvas LMS.

This is an edge-case: a user would have to give us a Canvas developer key and secret on the `/welcome` form and then use the LTI consumer key and secret that we generate for them in a _non-Canvas_ LMS. But it could happen, and if we tried to enable sections the results would be really confusing for both the user and for us/support.

lms/services/application_instance_getter.py
```diff
@@ -70,6 +71,18 @@ def provisioning_enabled(self):
...
+        """Return True if the application instance has Canvas sections is enabled."""
+        if not self._section_groups_feature_flag:
+            return False
+
+        try:
+            app_instance = self._get_by_consumer_key()
+        except ConsumerKeyError:
+            return False
+
+        return bool(app_instance.developer_key and app_instance.canvas_sections_enabled)
```
seanh, line 84: This is the "main" logic for "Is Canvas sections enabled?" The application instance has to both have a developer key and have `canvas_sections_enabled`.

It _is_ possible to have a developer key but _not_ have `canvas_sections_enabled`: this will happen for preexisting app instances that have a developer key.

Also, sections is off if the feature flag is off. The fact that `ApplicationInstanceGetter` is a service turned out to be quite handy 'cos the service factory can inject the feature flag.

lms/views/api/canvas/authorize.py
```diff
@@ -21,19 +21,31 @@
...
         self.request = request
 
     @view_config(permission="canvas_api", route_name="canvas_api.authorize")
     def authorize(self):
         ai_getter = self.request.find_service(name="ai_getter")
 
+        if ai_getter.canvas_sections_enabled():
+            scopes = self.files_scopes + self.sections_scopes
+        else:
+            scopes = self.files_scopes
```
seanh, line 47: We're calling `ai_getter.canvas_sections_enabled()` here because we _can't_ call `LTILaunchResource.canvas_sections_enabled()`: `LTILaunchResource` is not the context resource for this view. We can't extract `LTILaunchResource.canvas_sections_enabled()` out of that class either: it depends in `is_canvas`, which depends on checking certain LTI launch params to decide whether we're in Canvas or not, and those launch params are only present during the original launch request.

Here, in the authorization popup window, we're not "in Canvas" (where "in Canvas" is defined as the request has certain launch params that identify it as from Canvas). We're in a standalone popup window.

Fortunately this popup window would never be opened in a non-Canvas LMS anyway (it's the Canvas API authorization popup) so it's not necessary to check whether we're in Canvas here.

seanh: If we had tried to prematurely generalize this we'd be completely fucked right now

lms/views/api/canvas/authorize.py
```diff
@@ -84,7 +96,7 @@ def oauth2_redirect_error(self):
         template_variables = {
             "invalid_scope": self.request.params.get("error") == "invalid_scope",
             "details": self.request.params.get("error_description"),
-            "scopes": self.scopes,
+            "scopes": self.files_scopes + self.sections_scopes,
```
seanh, line 99: This is the list of scopes that we pass to the "You have missing scopes" error page's template. We always list both the files and sections scopes on this error page, even if sections is disabled, because:

1. In this exception view you can't always tell whether sections is enabled or not. Depending on exactly what happened `request.lti_user` is sometimes `None` when this view is called and you don't have the consumer key and can't look up the application instance.

2. We want people to give us all the scopes anyway. Even if they aren't using the sections feature right now, it'll be easier for them if/when they decide to ask us to enable sections for them if they already have the sections scopes. There's no advantage in _not_ asking people to set the sections scopes, this will only create more work for us and them.

lms/models/_application_settings.py
```diff
@@ -0,0 +1,51 @@
+"""Model classes representing application settings."""
+
+
+class ApplicationSettings:
```
jon-betts, line 4: This is very boring right now and doesn't have any business logic. But if we were going to implement any, this would be where we would.

For example, if we wanted different defaults, or reading from a "defaults" group or something.

We could also easily extend this with properties for the expected groups so settings could be accessed like:

```settings.canvas.sections_enabled```

File: tests/unit/lms/models/test_application_instance.py [hunk 12]
```diff
@@ -33,6 +34,45 @@ def test_provisioning_is_not_nullable(self, db_session, application_instance):
 
         assert application_instance.provisioning is not None
 
+    def test_settings_defaults_to_an_empty_dict(self, db_session, application_instance):
+        settings = application_instance.settings
+
+        assert isinstance(settings, ApplicationSettings)
+        assert settings.data == {}
```
seanh, line 41: I think `ApplicationSettings` should either have its own direct unit tests in an `_application_settings_test.py` file and the `ApplicationInstance` tests should patch `ApplicationSettings`, or `ApplicationSettings` should become an internal `_ApplicationSettings` helper within `application_instance.py` and then it can continue to be tested indirectly via `ApplicationInstance`.

Otherwise it's a bit strange, and breaks the usual layout pattern of our tests:

It's unclear where to expect to find the `ApplicationSettings` tests. It looks like it doesn't have any at first because there's no `_application_settings_test.py`, but in fact its tests can be found in another module's test file.

If `ApplicationSettings` might be re-used by another model (not just `ApplicationInstance`) in future, will the `ApplicationSettings` tests need to be repeated in the tests for the new model? If not, then it becomes even weirder that `ApplicationSettings` has two users and its tests can be found in the tests for one of its users but not the other.

If `ApplicationSettings` _isn't_ going to be reused then it should be an internal helper inside `application_instance.py`, then there's no problem with testing it indirectly via `ApplicationInstance`.

seanh:
Looks like the **Make the frontend call the sync API** commit got pushed to this branch. We need to remove that before merging this

tests/unit/lms/models/test_application_instance.py
jon-betts: I wanted to move this into a direction with application instance to make it explicit it's a private helper, or make it public but a few things stopped me in either direction:

 * I thought the PR would be rejected for being "too big" if I did move the application instance model
 * I think it actually could be used outside, for example a Canvas tool consumer object which is just passed the canvas portion for example...
 * ...but on the other hand I want to have deep tests that prove that manipulating the real settings object modifies the underlying SQL Alchemy. At this point if it was public we'd probably get stuck on enforced mocking, as I don't think we've resolved the "don't bother mocking PP model objects idea" yet. I thought the easiest way to avoid the conversation for longer would be to make it private for now.

So I'm kind of stuck with the current legal frame work.

My preferred solution would be:

 * Make the model public, test it in depth publicly
 * Define the unit of work to include the model: allow non mocking to prove changing the settings object modifies the underlying SQLAlchemy object

seanh:
Here's one alternative to the `ApplicationSettings` class. This maintains the JSON column, so you can add new settings without having to do a DB migration. But it also maintains the simple public interface of the `ApplicationInstance` model:

* `canvas_sections_enabled` is just a read/write boolean attr. Really simple to use and to mock: `application_instance.canvas_sections_enabled = True`

* `canvas_sections_enabled` also works as an init kwarg like other model attrs: `ApplicationInstance(canvas_sections_enabled=True)`

* This is about 70 lines of code shorter

* It does mean adding a new Python `@property` for every new setting, but I suspect a helper function might be able to remove the boilerplate duplication there somehow

```diff
diff --git a/lms/models/_application_settings.py b/lms/models/_application_settings.py
deleted file mode 100644
index e17fc0de..00000000
--- a/lms/models/_application_settings.py
+++ /dev/null
@@ -1,51 +0,0 @@
-"""Model classes representing application settings."""
-
-
-class ApplicationSettings:
-    """Model for accessing and updating application settings."""
-
-    def __init__(self, data):
-        self.data = data
-
-    @classmethod
-    def from_dict(cls, data):
-        """
-        Create a new settings object from the provided data.
-
-        The data should have dicts in dicts. With the first level defining
-        groups, and the second defining keys and values in that group.
-
-        :param data: Dicts in dicts with settings
-        :return: ApplicationSettings object
-        """
-        settings = cls({})
-        settings.update(data)
-
-        return settings
-
-    def update(self, data):
-        """Update all data from the provided settings.
-
-        :param data: Dicts in dicts with settings
-        """
-        for group, sub_settings in data.items():
-            for key, value in sub_settings.items():
-                self.set(group, key, value)
-
-    def get(self, group, key):
-        """Get a specific settings or None if it doesn't exist.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :return: The value or None
-        """
-        return self.data.get(group, {}).get(key)
-
-    def set(self, group, key, value):
-        """Set a specific setting in a group.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :param value: The value to set
-        """
-        self.data.setdefault(group, {})[key] = value
diff --git a/lms/models/application_instance.py b/lms/models/application_instance.py
index 61b2c9ec..94786a77 100644
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -8,7 +8,6 @@ from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
-from lms.models._application_settings import ApplicationSettings
 
 __all__ = ["ApplicationInstance"]
 
@@ -33,14 +32,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
-    _settings = sa.Column("settings", MutableDict.as_mutable(JSONB))
-
-    @property
-    def settings(self):
-        if self._settings is None:
-            self._settings = {}
-
-        return ApplicationSettings(self._settings)
+    settings = sa.Column(
+        "settings",
+        MutableDict.as_mutable(JSONB),
+        server_default=sa.func.jsonb("{}"),
+        nullable=False,
+    )
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
@@ -54,6 +51,24 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         "GroupInfo", back_populates="application_instance"
     )
 
+    def __init__(self, **kwargs):
+        canvas_sections_enabled = kwargs.pop("canvas_sections_enabled", False)
+
+        super().__init__(**kwargs)
+
+        self.settings = self.settings or {}
+        self.canvas_sections_enabled = canvas_sections_enabled
+
+    @property
+    def canvas_sections_enabled(self):
+        return self.settings.get("canvas", {}).get("sections_enabled", False)
+
+    @canvas_sections_enabled.setter
+    def canvas_sections_enabled(self, canvas_sections_enabled):
+        self.settings.setdefault("canvas", {})[
+            "sections_enabled"
+        ] = canvas_sections_enabled
+
     @classmethod
     def get_by_consumer_key(cls, db, consumer_key):
         """Return the ApplicationInstance with the given consumer_key or None."""
@@ -61,7 +76,13 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
 
     @classmethod
     def build_from_lms_url(  # pylint:disable=too-many-arguments
-        cls, lms_url, email, developer_key, developer_secret, encryption_key, settings,
+        cls,
+        lms_url,
+        email,
+        developer_key,
+        developer_secret,
+        encryption_key,
+        canvas_sections_enabled,
     ):
         """Instantiate ApplicationInstance with lms_url."""
         encrypted_secret = developer_secret
@@ -72,9 +93,6 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
                 developer_secret, encryption_key, aes_iv
             )
 
-        # Run the settings through the object to enable it validate it
-        settings = ApplicationSettings.from_dict(settings or {}).data
-
         return cls(
             consumer_key=_build_unique_key(),
             shared_secret=_build_shared_secret(),
@@ -84,7 +102,7 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
             developer_secret=encrypted_secret,
             aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
-            _settings=settings,
+            canvas_sections_enabled=canvas_sections_enabled,
         )
 
 
diff --git a/lms/services/application_instance_getter.py b/lms/services/application_instance_getter.py
index 5385317a..7e82ae54 100644
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -81,10 +81,7 @@ class ApplicationInstanceGetter:
         except ConsumerKeyError:
             return False
 
-        return bool(
-            app_instance.developer_key
-            and app_instance.settings.get("canvas", "sections_enabled")
-        )
+        return bool(app_instance.developer_key and app_instance.canvas_sections_enabled)
 
     def shared_secret(self):
         """
diff --git a/lms/views/application_instances.py b/lms/views/application_instances.py
index 4064dba6..e0ca1fad 100644
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -28,11 +28,9 @@ def create_application_instance(request):
         developer_key,
         developer_secret,
         request.registry.settings["aes_secret"],
-        settings={
-            "canvas": {
-                "sections_enabled": developer_key and request.feature("section_groups")
-            }
-        },
+        canvas_sections_enabled=bool(
+            developer_key and request.feature("section_groups")
+        ),
     )
     request.db.add(instance)
 
diff --git a/tests/unit/lms/models/test_application_instance.py b/tests/unit/lms/models/test_application_instance.py
index 33672ec7..798454fa 100644
--- a/tests/unit/lms/models/test_application_instance.py
+++ b/tests/unit/lms/models/test_application_instance.py
@@ -2,7 +2,6 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 
 
 class TestApplicationInstance:
@@ -34,44 +33,18 @@ class TestApplicationInstance:
 
         assert application_instance.provisioning is not None
 
-    def test_settings_defaults_to_an_empty_dict(self, db_session, application_instance):
-        settings = application_instance.settings
-
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == {}
-
-    @pytest.mark.parametrize(
-        "group,key,expected_value",
-        (
-            ("group", "key", "value"),
-            ("NEW", "key", None),
-            ("group", "NEW", None),
-            ("NEW", "NEW", None),
-        ),
-    )
-    def test_settings_can_be_retrieved(
-        self, db_session, application_instance, group, key, expected_value
-    ):
-        data = {"group": {"key": "value"}}
-        application_instance._settings = data
-
-        settings = application_instance.settings
+    def test_canvas_sections_enabled_defaults_to_False(self, application_instance):
+        assert not application_instance.canvas_sections_enabled
 
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == data
-        assert settings.get(group, key) == expected_value
+    def test_canvas_sections_enabled_can_be_initialized_to_True(self):
+        application_instance = ApplicationInstance(canvas_sections_enabled=True)
 
-    @pytest.mark.parametrize(
-        "group,key",
-        (("group", "key"), ("NEW", "key"), ("group", "NEW"), ("NEW", "NEW"),),
-    )
-    def test_can_update_settings(self, db_session, application_instance, group, key):
-        application_instance._settings = {"group": {"key": "value"}}
+        assert application_instance.canvas_sections_enabled is True
 
-        application_instance.settings.set(group, key, "new_value")
+    def test_canvas_sections_enabled_can_be_changed(self, application_instance):
+        application_instance.canvas_sections_enabled = True
 
-        assert application_instance.settings.get(group, key) == "new_value"
-        assert application_instance._settings[group][key] == "new_value"
+        assert application_instance.canvas_sections_enabled is True
 
     def test_consumer_key_cant_be_null(self, db_session, application_instance):
         application_instance.consumer_key = None
diff --git a/tests/unit/lms/services/application_instance_getter_test.py b/tests/unit/lms/services/application_instance_getter_test.py
index b174938f..a3af5a1b 100644
--- a/tests/unit/lms/services/application_instance_getter_test.py
+++ b/tests/unit/lms/services/application_instance_getter_test.py
@@ -3,7 +3,6 @@ from unittest import mock
 import pytest
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 from lms.models.application_instance import _build_aes_iv, _encrypt_oauth_secret
 from lms.services import ConsumerKeyError
 from lms.services.application_instance_getter import (
@@ -98,9 +97,7 @@ class TestApplicationInstanceGetter:
             mock.sentinel.context, pyramid_request
         )
         test_application_instance.developer_key = developer_key
-        test_application_instance._settings = {
-            "canvas": {"sections_enabled": canvas_sections_enabled}
-        }
+        test_application_instance.canvas_sections_enabled = canvas_sections_enabled
 
         assert ai_getter.canvas_sections_enabled() == expected_result
 
diff --git a/tests/unit/lms/views/test_application_instance.py b/tests/unit/lms/views/test_application_instance.py
index e4857d90..97e25d3c 100644
--- a/tests/unit/lms/views/test_application_instance.py
+++ b/tests/unit/lms/views/test_application_instance.py
@@ -69,10 +69,7 @@ class TestCreateApplicationInstance:
         create_application_instance(pyramid_request)
 
         application_instance = pyramid_request.db.query(ApplicationInstance).one()
-        assert (
-            bool(application_instance.settings.get("canvas", "sections_enabled"))
-            == canvas_sections_enabled
-        )
+        assert application_instance.canvas_sections_enabled == canvas_sections_enabled
 
     @pytest.fixture
     def pyramid_request(self, pyramid_request):
```

seanh:
Another possible alternative. This one:

* Is over 70 lines shorter

* Keeps the JSON column so you don't need to do a DB migration to add new settings

* You also don't need to add new Python code to the model to add new settings. Just start using a setting via the `get_setting()` and `set_setting()` methods

* This does complicate the interface of the `ApplicationInstance` class somewhat. But it might be simple enough:

  * `canvas_sections_enabled` is no longer a simple boolean property. Instead of `application_instance.canvas_sections_enabled` and `application_instance.canvas_sections_enabled = True` it's now:

        application_instance.get_setting("canvas", "sections_enabled")
        application_instance.set_setting("canvas", "sections_enabled", True)

    You also can't just pass `ApplicationInstance(canvas_sections_enabled=True)` as an init kwarg. You would have to do `ApplicationInstance(settings={"canvas": {"sections_enabled": True }})`. But you could alternatively do:

        application_instance = ApplicationInstance()
        application_instance.set_setting("canvas", "sections_enabled", True)

    That's still more complicated and different from the other "normal" settings but maybe less bad as an alternative to passing a nested dict into the init.

* The dict is exposed directly as `application_instance.settings`, as well as having the `get_setting()` and `set_setting()` convenience methods

```diff
diff --git a/lms/models/_application_settings.py b/lms/models/_application_settings.py
deleted file mode 100644
index e17fc0de..00000000
--- a/lms/models/_application_settings.py
+++ /dev/null
@@ -1,51 +0,0 @@
-"""Model classes representing application settings."""
-
-
-class ApplicationSettings:
-    """Model for accessing and updating application settings."""
-
-    def __init__(self, data):
-        self.data = data
-
-    @classmethod
-    def from_dict(cls, data):
-        """
-        Create a new settings object from the provided data.
-
-        The data should have dicts in dicts. With the first level defining
-        groups, and the second defining keys and values in that group.
-
-        :param data: Dicts in dicts with settings
-        :return: ApplicationSettings object
-        """
-        settings = cls({})
-        settings.update(data)
-
-        return settings
-
-    def update(self, data):
-        """Update all data from the provided settings.
-
-        :param data: Dicts in dicts with settings
-        """
-        for group, sub_settings in data.items():
-            for key, value in sub_settings.items():
-                self.set(group, key, value)
-
-    def get(self, group, key):
-        """Get a specific settings or None if it doesn't exist.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :return: The value or None
-        """
-        return self.data.get(group, {}).get(key)
-
-    def set(self, group, key, value):
-        """Set a specific setting in a group.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :param value: The value to set
-        """
-        self.data.setdefault(group, {})[key] = value
diff --git a/lms/models/application_instance.py b/lms/models/application_instance.py
index 61b2c9ec..4ccdc480 100644
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -8,7 +8,6 @@ from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
-from lms.models._application_settings import ApplicationSettings
 
 __all__ = ["ApplicationInstance"]
 
@@ -33,14 +32,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
-    _settings = sa.Column("settings", MutableDict.as_mutable(JSONB))
-
-    @property
-    def settings(self):
-        if self._settings is None:
-            self._settings = {}
-
-        return ApplicationSettings(self._settings)
+    settings = sa.Column(
+        "settings",
+        MutableDict.as_mutable(JSONB),
+        server_default=sa.func.jsonb("{}"),
+        nullable=False,
+    )
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
@@ -54,6 +51,16 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         "GroupInfo", back_populates="application_instance"
     )
 
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.settings = self.settings or {}
+
+    def get_setting(self, group, key):
+        return self.settings.get(group, {}).get(key)
+
+    def set_setting(self, group, key, value):
+        self.settings.setdefault(group, {})[key] = value
+
     @classmethod
     def get_by_consumer_key(cls, db, consumer_key):
         """Return the ApplicationInstance with the given consumer_key or None."""
@@ -72,9 +79,6 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
                 developer_secret, encryption_key, aes_iv
             )
 
-        # Run the settings through the object to enable it validate it
-        settings = ApplicationSettings.from_dict(settings or {}).data
-
         return cls(
             consumer_key=_build_unique_key(),
             shared_secret=_build_shared_secret(),
@@ -84,7 +88,7 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
             developer_secret=encrypted_secret,
             aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
-            _settings=settings,
+            settings=settings,
         )
 
 
diff --git a/lms/services/application_instance_getter.py b/lms/services/application_instance_getter.py
index 5385317a..da18d536 100644
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -83,7 +83,7 @@ class ApplicationInstanceGetter:
 
         return bool(
             app_instance.developer_key
-            and app_instance.settings.get("canvas", "sections_enabled")
+            and app_instance.get_setting("canvas", "sections_enabled")
         )
 
     def shared_secret(self):
diff --git a/lms/views/application_instances.py b/lms/views/application_instances.py
index 4064dba6..32a451af 100644
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -30,8 +30,10 @@ def create_application_instance(request):
         request.registry.settings["aes_secret"],
         settings={
             "canvas": {
-                "sections_enabled": developer_key and request.feature("section_groups")
-            }
+                "sections_enabled": bool(
+                    developer_key and request.feature("section_groups")
+                ),
+            },
         },
     )
     request.db.add(instance)
diff --git a/tests/unit/lms/models/test_application_instance.py b/tests/unit/lms/models/test_application_instance.py
index 33672ec7..7f1f6c6e 100644
--- a/tests/unit/lms/models/test_application_instance.py
+++ b/tests/unit/lms/models/test_application_instance.py
@@ -2,7 +2,6 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 
 
 class TestApplicationInstance:
@@ -34,44 +33,18 @@ class TestApplicationInstance:
 
         assert application_instance.provisioning is not None
 
-    def test_settings_defaults_to_an_empty_dict(self, db_session, application_instance):
-        settings = application_instance.settings
-
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == {}
-
-    @pytest.mark.parametrize(
-        "group,key,expected_value",
-        (
-            ("group", "key", "value"),
-            ("NEW", "key", None),
-            ("group", "NEW", None),
-            ("NEW", "NEW", None),
-        ),
-    )
-    def test_settings_can_be_retrieved(
-        self, db_session, application_instance, group, key, expected_value
-    ):
-        data = {"group": {"key": "value"}}
-        application_instance._settings = data
-
-        settings = application_instance.settings
+    def test_settings_default_to_None(self, application_instance):
+        assert application_instance.get_setting("group", "key") is None
 
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == data
-        assert settings.get(group, key) == expected_value
+    def test_settings_can_be_initialized(self):
+        application_instance = ApplicationInstance(settings={"group": {"key": "value"}})
 
-    @pytest.mark.parametrize(
-        "group,key",
-        (("group", "key"), ("NEW", "key"), ("group", "NEW"), ("NEW", "NEW"),),
-    )
-    def test_can_update_settings(self, db_session, application_instance, group, key):
-        application_instance._settings = {"group": {"key": "value"}}
+        assert application_instance.get_setting("group", "key") == "value"
 
-        application_instance.settings.set(group, key, "new_value")
+    def test_settings_can_be_changed(self, application_instance):
+        application_instance.set_setting("group", "key", "value")
 
-        assert application_instance.settings.get(group, key) == "new_value"
-        assert application_instance._settings[group][key] == "new_value"
+        assert application_instance.get_setting("group", "key") == "value"
 
     def test_consumer_key_cant_be_null(self, db_session, application_instance):
         application_instance.consumer_key = None
diff --git a/tests/unit/lms/services/application_instance_getter_test.py b/tests/unit/lms/services/application_instance_getter_test.py
index b174938f..d632e55a 100644
--- a/tests/unit/lms/services/application_instance_getter_test.py
+++ b/tests/unit/lms/services/application_instance_getter_test.py
@@ -3,7 +3,6 @@ from unittest import mock
 import pytest
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 from lms.models.application_instance import _build_aes_iv, _encrypt_oauth_secret
 from lms.services import ConsumerKeyError
 from lms.services.application_instance_getter import (
@@ -98,9 +97,9 @@ class TestApplicationInstanceGetter:
             mock.sentinel.context, pyramid_request
         )
         test_application_instance.developer_key = developer_key
-        test_application_instance._settings = {
-            "canvas": {"sections_enabled": canvas_sections_enabled}
-        }
+        test_application_instance.set_setting(
+            "canvas", "sections_enabled", canvas_sections_enabled
+        )
 
         assert ai_getter.canvas_sections_enabled() == expected_result
 
diff --git a/tests/unit/lms/views/test_application_instance.py b/tests/unit/lms/views/test_application_instance.py
index e4857d90..1ff46fa5 100644
--- a/tests/unit/lms/views/test_application_instance.py
+++ b/tests/unit/lms/views/test_application_instance.py
@@ -70,7 +70,7 @@ class TestCreateApplicationInstance:
 
         application_instance = pyramid_request.db.query(ApplicationInstance).one()
         assert (
-            bool(application_instance.settings.get("canvas", "sections_enabled"))
+            application_instance.get_setting("canvas", "sections_enabled")
             == canvas_sections_enabled
         )
 ```

**Comment by seanh** [comment]
Here's a third variant. This makes `ApplicationInstance`'s API more similar to with your `ApplicationSettings` commit (`.settings.set()`), but still retains some of the advantages I'm looking for.

```python
application_instance.settings.set("canvas", "sections_enabled", True)
application_instance.settings.get("canvas", "sections_enabled")
application_instance._settings  # The raw dict is there if you need it, albeit marked internal.

# You can't pass settings to __init__ in a nice way but you can do it in this horrible way:
ApplicationInstance(_settings={"canvas": {"sections_enabled": True}})

# Or just use two lines:
application_instance = ApplicationInstance()
application_instance.settings.set("canvas", "sections_enabled", True)
```

```diff
diff --git a/lms/models/_application_settings.py b/lms/models/_application_settings.py
deleted file mode 100644
index e17fc0de..00000000
--- a/lms/models/_application_settings.py
+++ /dev/null
@@ -1,51 +0,0 @@
-"""Model classes representing application settings."""
-
-
-class ApplicationSettings:
-    """Model for accessing and updating application settings."""
-
-    def __init__(self, data):
-        self.data = data
-
-    @classmethod
-    def from_dict(cls, data):
-        """
-        Create a new settings object from the provided data.
-
-        The data should have dicts in dicts. With the first level defining
-        groups, and the second defining keys and values in that group.
-
-        :param data: Dicts in dicts with settings
-        :return: ApplicationSettings object
-        """
-        settings = cls({})
-        settings.update(data)
-
-        return settings
-
-    def update(self, data):
-        """Update all data from the provided settings.
-
-        :param data: Dicts in dicts with settings
-        """
-        for group, sub_settings in data.items():
-            for key, value in sub_settings.items():
-                self.set(group, key, value)
-
-    def get(self, group, key):
-        """Get a specific settings or None if it doesn't exist.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :return: The value or None
-        """
-        return self.data.get(group, {}).get(key)
-
-    def set(self, group, key, value):
-        """Set a specific setting in a group.
-
-        :param group: The name of the group of settings
-        :param key: The key in that group
-        :param value: The value to set
-        """
-        self.data.setdefault(group, {})[key] = value
diff --git a/lms/models/application_instance.py b/lms/models/application_instance.py
index 61b2c9ec..000e3a0c 100644
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -8,7 +8,6 @@ from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
-from lms.models._application_settings import ApplicationSettings
 
 __all__ = ["ApplicationInstance"]
 
@@ -33,14 +32,12 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
-    _settings = sa.Column("settings", MutableDict.as_mutable(JSONB))
-
-    @property
-    def settings(self):
-        if self._settings is None:
-            self._settings = {}
-
-        return ApplicationSettings(self._settings)
+    _settings = sa.Column(
+        "settings",
+        MutableDict.as_mutable(JSONB),
+        server_default=sa.func.jsonb("{}"),
+        nullable=False,
+    )
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
@@ -54,6 +51,11 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
         "GroupInfo", back_populates="application_instance"
     )
 
+    @property
+    def settings(self):
+        self._settings = self._settings or {}
+        return _Settings(self._settings)
+
     @classmethod
     def get_by_consumer_key(cls, db, consumer_key):
         """Return the ApplicationInstance with the given consumer_key or None."""
@@ -72,9 +74,6 @@ class ApplicationInstance(BASE):  # pylint:disable=too-few-public-methods
                 developer_secret, encryption_key, aes_iv
             )
 
-        # Run the settings through the object to enable it validate it
-        settings = ApplicationSettings.from_dict(settings or {}).data
-
         return cls(
             consumer_key=_build_unique_key(),
             shared_secret=_build_shared_secret(),
@@ -107,3 +106,14 @@ def _encrypt_oauth_secret(oauth_secret, key, init_v):
     """Encrypt an oauth secrety via AES encryption."""
     cipher = AES.new(key, AES.MODE_CFB, init_v)
     return cipher.encrypt(oauth_secret)
+
+
+class _Settings:
+    def __init__(self, settings):
+        self.settings = settings
+
+    def get(self, group, key):
+        return self.settings.get(group, {}).get(key)
+
+    def set(self, group, key, value):
+        self.settings.setdefault(group, {})[key] = value
diff --git a/lms/views/application_instances.py b/lms/views/application_instances.py
index 4064dba6..32a451af 100644
--- a/lms/views/application_instances.py
+++ b/lms/views/application_instances.py
@@ -30,8 +30,10 @@ def create_application_instance(request):
         request.registry.settings["aes_secret"],
         settings={
             "canvas": {
-                "sections_enabled": developer_key and request.feature("section_groups")
-            }
+                "sections_enabled": bool(
+                    developer_key and request.feature("section_groups")
+                ),
+            },
         },
     )
     request.db.add(instance)
diff --git a/tests/unit/lms/models/test_application_instance.py b/tests/unit/lms/models/test_application_instance.py
index 33672ec7..88af0c07 100644
--- a/tests/unit/lms/models/test_application_instance.py
+++ b/tests/unit/lms/models/test_application_instance.py
@@ -2,7 +2,6 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 
 
 class TestApplicationInstance:
@@ -34,44 +33,20 @@ class TestApplicationInstance:
 
         assert application_instance.provisioning is not None
 
-    def test_settings_defaults_to_an_empty_dict(self, db_session, application_instance):
-        settings = application_instance.settings
-
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == {}
-
-    @pytest.mark.parametrize(
-        "group,key,expected_value",
-        (
-            ("group", "key", "value"),
-            ("NEW", "key", None),
-            ("group", "NEW", None),
-            ("NEW", "NEW", None),
-        ),
-    )
-    def test_settings_can_be_retrieved(
-        self, db_session, application_instance, group, key, expected_value
-    ):
-        data = {"group": {"key": "value"}}
-        application_instance._settings = data
-
-        settings = application_instance.settings
+    def test_settings_default_to_None(self, application_instance):
+        assert application_instance.settings.get("group", "key") is None
 
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == data
-        assert settings.get(group, key) == expected_value
+    def test_settings_can_be_initialized(self):
+        application_instance = ApplicationInstance(
+            _settings={"group": {"key": "value"}}
+        )
 
-    @pytest.mark.parametrize(
-        "group,key",
-        (("group", "key"), ("NEW", "key"), ("group", "NEW"), ("NEW", "NEW"),),
-    )
-    def test_can_update_settings(self, db_session, application_instance, group, key):
-        application_instance._settings = {"group": {"key": "value"}}
+        assert application_instance.settings.get("group", "key") == "value"
 
-        application_instance.settings.set(group, key, "new_value")
+    def test_settings_can_be_changed(self, application_instance):
+        application_instance.settings.set("group", "key", "value")
 
-        assert application_instance.settings.get(group, key) == "new_value"
-        assert application_instance._settings[group][key] == "new_value"
+        assert application_instance.settings.get("group", "key") == "value"
 
     def test_consumer_key_cant_be_null(self, db_session, application_instance):
         application_instance.consumer_key = None
diff --git a/tests/unit/lms/services/application_instance_getter_test.py b/tests/unit/lms/services/application_instance_getter_test.py
index b174938f..16364c4d 100644
--- a/tests/unit/lms/services/application_instance_getter_test.py
+++ b/tests/unit/lms/services/application_instance_getter_test.py
@@ -3,7 +3,6 @@ from unittest import mock
 import pytest
 
 from lms.models import ApplicationInstance
-from lms.models._application_settings import ApplicationSettings
 from lms.models.application_instance import _build_aes_iv, _encrypt_oauth_secret
 from lms.services import ConsumerKeyError
 from lms.services.application_instance_getter import (
@@ -98,9 +97,9 @@ class TestApplicationInstanceGetter:
             mock.sentinel.context, pyramid_request
         )
         test_application_instance.developer_key = developer_key
-        test_application_instance._settings = {
-            "canvas": {"sections_enabled": canvas_sections_enabled}
-        }
+        test_application_instance.settings.set(
+            "canvas", "sections_enabled", canvas_sections_enabled
+        )
 
         assert ai_getter.canvas_sections_enabled() == expected_result
 
diff --git a/tests/unit/lms/views/test_application_instance.py b/tests/unit/lms/views/test_application_instance.py
index e4857d90..2b2293e0 100644
--- a/tests/unit/lms/views/test_application_instance.py
+++ b/tests/unit/lms/views/test_application_instance.py
@@ -70,7 +70,7 @@ class TestCreateApplicationInstance:
 
         application_instance = pyramid_request.db.query(ApplicationInstance).one()
         assert (
-            bool(application_instance.settings.get("canvas", "sections_enabled"))
+            application_instance.settings.get("canvas", "sections_enabled")
             == canvas_sections_enabled
         )
 
```

Issue #1748, opened 2020-05-13
Improve UI when /sync is loading groups


LMS007:
When using canvas with section groups enabled, 2 issues can happen.

**Issue 1:** It's not important to show the spinner when the groups are loading via /sync call. It would be better to just load the PDF and sidebar w/o the groups and the lazily populate when the endpoint resolved.

![spinner](https://user-images.githubusercontent.com/3939074/81874588-02bc7f80-9533-11ea-899f-172b12558131.gif)

**Issue 2.** When the /sync API returns an auth error , the spinner shows while the dialog is also rendered requesting authorization for the user. It would be better to just hide the spinner in this case.

![spinner-and-dialog](https://user-images.githubusercontent.com/3939074/81874594-06500680-9533-11ea-8fd4-5dc6ad877855.gif)

robertknight:
Original context from @seanh in Slack:

First issue: https://hypothes-is.slack.com/archives/C1M8NH76X/p1589384535367400
Second issue: https://hypothes-is.slack.com/archives/C1M8NH76X/p1589384651368200

PR #1749, opened 2020-05-14
BasicLtiLaunchApp UI improvments

LMS007:
- Don’t show spinner if there is an error
- Don’t show spinner when /sync is pending
- Render the <iframe> no matter what, but hidden if content URL is null or there is an error
- Remove waitForElementToBeRemoved as its no longer needed
- Various test adjustments for the above
- Fix several false negative tests
- Add `hide` prop to Spinner for easier negative testing
----

One shortcoming is it will override an error if another comes in. In order to prevent that, it will need some rethinking about `errorState`, but I think this is a minor and in most cases, the same error will be produced (auth error).  

fixes https://github.com/hypothesis/lms/issues/1748

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -303,7 +303,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
 
   return (
     <span className="BasicLtiLaunchApp">
-      <Spinner hide={!showSpinner} className="BasicLtiLaunchApp__spinner" />
+      {showSpinner && <Spinner className="BasicLtiLaunchApp__spinner" />}
       {errorDialog}
       {content}
     </span>
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -117,8 +117,6 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         handleError(e, 'error-fetch');
       }
     }
-    // ignore errorState changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [apiSync, authToken, rpcServer]);
 
   /**
@@ -146,8 +144,6 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     } finally {
       decFetchCount();
     }
-    // ignore errorState changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [authToken, viaCallbackUrl]);
 
   /**
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import { Fragment, createElement } from 'preact';
 import propTypes from 'prop-types';
 
@@ -50,7 +51,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   } = useContext(Config);
 
   // The current state of the error.
-  // One of "error-fetch", "error-authorizing", or "error-report-submission"
+  // One of "error-fetch", "error-authorizing", or "error-report-submission" or null
   const [errorState, setErrorState] = useState(null);
 
   // Any current error thrown.
@@ -71,7 +72,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   // is falsely
   const showIframe = contentUrl && !errorState;
 
-  const showSpinner = fetchCount > 0;
+  const showSpinner = fetchCount > 0 && !errorState;
 
   const incFetchCount = () => {
     setFetchCount(count => count + 1);
@@ -86,16 +87,16 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
    * Helper to handle thrown errors from from API requests.
    *
    * @param {Error} e - Error object from request.
-   * @param {string} errorState - One of "error-fetch",
+   * @param {string} newErrorState - One of "error-fetch",
    *  "error-authorizing", or "error-report-submission"
    * @param {boolean} [retry=true] - Can the request be retried?
    */
-  const handleError = (e, errorState, retry = true) => {
+  const handleError = (e, newErrorState, retry = true) => {
     if (e instanceof ApiError && !e.errorMessage && retry) {
       setErrorState('error-authorizing');
     } else {
       setError(e);
-      setErrorState(errorState);
+      setErrorState(newErrorState);
     }
   };
 
@@ -105,7 +106,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   const fetchGroups = useCallback(async () => {
     if (apiSync) {
       try {
-        incFetchCount();
+        setErrorState(null);
         const groups = await apiCall({
           authToken,
           path: apiSync.path,
@@ -114,10 +115,10 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         rpcServer.resolveGroupFetch(groups);
       } catch (e) {
         handleError(e, 'error-fetch');
-      } finally {
-        decFetchCount();
       }
     }
+    // ignore errorState changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [apiSync, authToken, rpcServer]);
 
   /**
@@ -145,6 +146,8 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     } finally {
       decFetchCount();
     }
+    // ignore errorState changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [authToken, viaCallbackUrl]);
 
   /**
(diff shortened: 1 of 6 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -209,43 +202,11 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
...
-        >
-          {iFrame}
-        </LMSGrader>
-      );
-    } else {
-      return iFrame;
-    }
-  } else {
-    // Render either a Spinner or Dialog.
+  // Render
```
LMS007, line 205: I simplified the render() logic to the following

1. If there is an error , render the error
2. If there is a spinner, render the spinner
3. Else Render the iFrame

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -105,7 +101,6 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   const fetchGroups = useCallback(async () => {
```
LMS007, line 101: fetchGroups no longer is part of fetchCount. I'm hesitate to throw incFetchCount and decFetchCount away even through its only used for the fetchContentUrl.  There is still value in those functions for cleaning up error so I'd like to keep it.

robertknight:
I tested this locally with the following diff against the LMS app. This is not the correct method of enabling the feature flag but it was the easiest thing for me to do locally before I looked into the correct way to do it. Note that in addition to enabling section groups I have also made the `/sync` API always have a 10 second delay:

```diff
diff --git a/conf/development.ini b/conf/development.ini
index 10887810..27d1bb7e 100644
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -18,6 +18,8 @@ session_cookie_secret = "notasecret"
 feature_flags_cookie_secret = "notasecret"
 feature_flags_allowed_in_cookie = use_legacy_via
 
+feature_flags.section_groups = true
+
 # The secret string that's used to sign the OAuth2 state param.
 oauth2_state_secret = "notasecret"
 
diff --git a/lms/services/application_instance_getter.py b/lms/services/application_instance_getter.py
index 5385317a..007de31a 100644
--- a/lms/services/application_instance_getter.py
+++ b/lms/services/application_instance_getter.py
@@ -72,6 +72,9 @@ class ApplicationInstanceGetter:
         return provisioning
 
     def canvas_sections_enabled(self):
+        # TESTING
+        return True
+
         """Return True if the application instance has Canvas sections is enabled."""
         if not self._section_groups_feature_flag:
             return False
diff --git a/lms/views/api/canvas/sync.py b/lms/views/api/canvas/sync.py
index d646a9d3..157a804b 100644
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -1,4 +1,5 @@
 import hashlib
+import time
 
 from pyramid.view import view_config
 
@@ -12,6 +13,8 @@ from lms.models import HGroup, h_group_name
     permission="sync_api",
 )
 def sync(request):
+    time.sleep(10)
+
     authority = request.registry.settings["h_authority"]
     canvas_api_svc = request.find_service(name="canvas_api_client")
     lti_h_svc = request.find_service(name="lti_h")
```

(nb. to apply this locally on macOS, copy the above to the clipboard and then use `pbpaste | git apply` or save the diff to a file and use `git apply <filename>`).

The first test I did was to start a session with an existing Canvas API authorization. That resulted in the spinner no-longer showing while the sync API call was being made. In most cases the sync API is likely to complete before the client has fully loaded. However if the sync API is still in-flight at that time, the client UI shows a "blank state" which doesn't indicate that anything is currently loading:

Screenshot 2020-05-14 12 34 01

This is an existing issue with the client. I filed this as https://github.com/hypothesis/client/issues/2161.

The second test I did was to revoke my OAuth token approval for the "localhost (make devdata)" app under https://hypothesis.instructure.com/profile/settings and then try to reload the assignment.

This time the iframe appeared and then once the sync completed, it disappeared and an auth dialog was shown. After approving the auth dialog, nothing appeared to happen. A sync request was triggered however and once that eventually completed, an error was displayed in the console:

Screenshot 2020-05-14 12 40 20

After reloading the assignment, it worked.

The final test I did was to remove my OAuth tokens from the LMS DB using:

```
make sql
...
postgres=# delete from oauth2_token;
```

And then reloading the assignment. This produced a similar / the same result as above.

robertknight:
It looks to me like the problem above where the app gets stuck at the Authorize screen was happening because the `errorState` state is not reset after authorization completes. Adding the following diff resolved the issue:

```diff
diff --git a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
index f7faf473..ff5dee05 100644
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -196,6 +196,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
 
     try {
       await authWindow.current.authorize();
+      setErrorState(null);
       await Promise.all([fetchContentUrl(), fetchGroups()]);
     } finally {
       authWindow.current = null;
```

Something that is not clear to me yet - why did this work before? Since the error state was not being reset after `authorize()` returned previously.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
robertknight: Since it is possible to transition from state (3) to state (2) or (1), it is possible that the iframe gets created, the client loads and then it gets destroyed. There are some places in the LMS app that can end up with references to the iframe which would need to handle it being destroyed:

1. If the iframe goes away while the JSON-RPC server is handling a `requestGroups` request from it, then the attempt to `postMessage` a response back to the iframe (`event.source.postMessage(...)`) will fail in https://github.com/hypothesis/lms/blob/86923468995891023c16f869e35d50256e5f8a92/lms/static/scripts/postmessage_json_rpc/server/server.js#L79.
2. The `Server` object exposes a Promise for the sidebar window via `sidebarWindow`. Promises can only be resolved once, so if the promise gets resolved with an initial iframe and that iframe is later destroyed and a new one is created, the `sidebarWindow` Promise will continue to try to return the original iframe (which will become `null` once it is destroyed).

So to resolve this, we either need to make the code in various places in the LMS frontend cope with the iframe being destroyed and re-created, or re-work the render code here to keep the iframe around once it is rendered but just hide it.

Keeping the iframe rendered-but-hidden is going to be a better user experience because it will show up instantly when the error/authorization step is completed, rather than needing to be reloaded from scratch.

LMS007: > Something that is not clear to me yet - why did this work before? Since the error state was not being reset after `authorize()` returned previously.

Likely because `showIframe` (now removed) was true and it didnt matter that the errorState have a value in it. And in spite of that working, it seems confusing, best to clean it up

LMS007:
- Tests broken atm, but this will render a hidden iframe from the outside.
- iframe will remain hidden as long as there is either an error or the contentUrl is null
- the /sync call will not affect the spinner or the iframe's rendered state, nor will it invoke the spinner 
- like before, any errors will take precedent, and if there are 2 successive errors, they will stack like before unless they are the same (e.g. same errorState value)
-

lms/static/scripts/frontend_apps/components/Spinner.js
```diff
@@ -1,21 +1,26 @@
...
  */
-export default function Spinner({ className }) {
+export default function Spinner({ className, hide }) {
   return (
-    <SvgIcon
-      className={className}
-      src={trustMarkup(require('../../../images/spinner.svg'))}
-      inline={true}
-    />
+    <Fragment>
```
LMS007, line 11: The spinner has an option `hide` prop that makes testing a bit easier since the component does not remove itself from the DOM.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -36,6 +31,17 @@ describe('BasicLtiLaunchApp', () => {
     );
   };
 
+  function contentHidden(wrapper) {
+    return waitForElement(wrapper, '.BasicLtiLaunchApp__content.is-hidden');
```
LMS007, line 35: two new helper methods to test when the iframe is visible.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -114,10 +115,10 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         rpcServer.resolveGroupFetch(groups);
       } catch (e) {
         handleError(e, 'error-fetch');
-      } finally {
-        decFetchCount();
       }
     }
+    // ignore errorState changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
robertknight: This lint suppression and the one below isn't needed since the effect doesn't read `errorState`. It does update `errorState`, but state setters don't need to be listed as dependencies since the setter function never changes.

lms/static/scripts/frontend_apps/components/Spinner.js
robertknight: I'm not super keen on this approach since it means adding a prop like this to each component where you need to wait for it to be removed.

In this branch I've pushed a change which reverts to adding/removing the `Spinner`, but simplifies waiting for the spinner to be shown or hidden by creating a couple of utilities:

```js
function spinnerVisible(wrapper) {
  return waitForElement(wrapper, 'Spinner');
}

function spinnerHidden(wrapper) {
  return waitFor(() => {
    wrapper.update();
    return !wrapper.exists('Spinner');
  });
}
```

The advantage is that this is easier to adapt for any new use cases we have that need to do the same thing.

I think the `waitForElementToBeRemoved` utility that you had implemented previously was also a fine approach here.

PR #1762, opened 2020-05-18
POC: Automate make services more bash-like example

jon-betts:
This is far more bash oriented and relies on the fact that all bash commands have a return status.

This allows us to do things like `if /bin/services/are-up; then...` Which makes the scripts very short and concise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ help:
 .PHONY: services
 services: args?=up -d
 services: python
-	@tox -qe docker-compose -- $(args)
+	@./bin/services/are-up || (tox -qe docker-compose -- $(args) && ./bin/services/wait-until-up)
 
 .PHONY: dev
 dev: build/manifest.json python
--- a/bin/services/are-up
+++ b/bin/services/are-up
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# Check the status
+docker exec -it lms_postgres_1 /opt/hypothesis/is-ready 1>/dev/null 2>/dev/null;
--- a/bin/services/wait-until-up
+++ b/bin/services/wait-until-up
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+while ! ./bin/services/are-up
+do
+  echo "Waiting for services..."
+  sleep 0.01
+done
+
+echo "Services up"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,3 +4,5 @@ services:
     image: postgres:11.5-alpine
     ports:
       - '127.0.0.1:5433:5432'
+    volumes:
+      - ./bin/services/postgres:/opt/hypothesis
--- a/tox.ini
+++ b/tox.ini
@@ -68,11 +68,14 @@ setenv =
     OBJC_DISABLE_INITIALIZE_FORK_SAFETY = YES
 whitelist_externals =
     tests,functests,bddtests: sh
-commands =
-    dev: {posargs:pserve conf/development.ini --reload}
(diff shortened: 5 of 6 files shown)


jon-betts:
An example of what I meant about running bits separately. Currently this doesn't take Jenkins into account, as the `dev` tox command now runs make services.

The simplest way to fix this would be to have the `dev` make target have `services-up` as a dependency, but this would lose the property that Sean would like where running tox commands directly starts the services. I'm sure we could work something out.

Anyway, _don't_ merge this as I'm sure it will create merry havok, but it demonstrates the kind of thing I mean.

bin/services/are-up
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Check the status
+docker exec -it lms_postgres_1 /opt/hypothesis/is-ready 1>/dev/null 2>/dev/null;
```
jon-betts, line 4: This could be a loop over docker image names

bin/services/wait-until-up
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+while ! ./bin/services/are-up
+do
+  echo "Waiting for services..."
+  sleep 0.01
```
jon-betts, line 6: The time in the sleep is actually basically irrelevant, it could probably go. Connecting via docker limits this to about 2 per second anyway.

docker-compose.yml
```diff
@@ -4,3 +4,5 @@ services:
     image: postgres:11.5-alpine
     ports:
       - '127.0.0.1:5433:5432'
+    volumes:
+      - ./bin/services/postgres:/opt/hypothesis
```
jon-betts, line 8: I tried putting in the file, but it behaved weirdly. I think it might interpret things as a directory if they aren't already there. Not sure, this is actually kind of nicer anyway.

tox.ini
```diff
@@ -68,11 +68,14 @@ setenv =
     OBJC_DISABLE_INITIALIZE_FORK_SAFETY = YES
 whitelist_externals =
     tests,functests,bddtests: sh
-commands =
-    dev: {posargs:pserve conf/development.ini --reload}
+    dev,tests,functests,bddtests: make
+commands_pre =
+    {dev,tests,functests,bddtests}: make -C {toxinidir} services
```
jon-betts, line 73: This is the part we'd need to skip for jenkins. Could just be a case of sh `$JENKINS_VAR || make...` or something?

Makefile
```diff
@@ -29,7 +29,7 @@ help:
 .PHONY: services
 services: args?=up -d
 services: python
-	@tox -qe docker-compose -- $(args)
+	@./bin/services/are-up || (tox -qe docker-compose -- $(args) && ./bin/services/wait-until-up)
```
jon-betts, line 32: This could easily be split into `services` and `services-up`. If you do them together you can skip a second check for whether they are up which would happen if they were.

bin/services/are-up
jon-betts, line 4:
```suggestion
docker exec -it lms_postgres_1 /opt/hypothesis/is-ready 2>&1 >/dev/null;
```
?

seanh:
Closing this in favour of https://github.com/hypothesis/lms/pull/1834. I think https://github.com/hypothesis/lms/pull/1834 meets the requirements of having lots of separate scripts instead of one big one, and being easily extendable to adding more services.

It's also a bit nicer because it avoids any service-specific scripts (e.g. Postgres-specific), instead keeping all the per-container specific commands in `docker-compose.yml`

PR #1779, opened 2020-05-26
Dialog css fixes

LMS007:
- Increase max-width of dialogs to 700px
- Allow dialogs to expand to 90% of the height to take advantage of extra space
- Allow extremely long content in dialogs to scroll, but not hide any bottom buttons out of the view
- Coverage & delete some CSS settings among various dialogs used
- Fix bug where non-json error details text was not formatted pretty in <ErrorDisplay>
- Fix bug that set “undefined” class in <Dialog> when not passing a `contentClass` prop.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -57,6 +57,13 @@ export default function Dialog({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  const contentClasses = {
+    Dialog__content: true,
+  };
+  if (contentClasses) {
+    contentClasses[contentClass] = true;
+  }
+
   return (
     <div
       role={role}
@@ -70,12 +77,7 @@ export default function Dialog({
         style={{ zIndex: zIndexScale.dialogBackground }}
       />
       <div className="Dialog__container" style={{ zIndex: zIndexScale.dialog }}>
-        <div
-          className={classNames({
-            Dialog__content: true,
-            [contentClass]: true,
-          })}
-        >
+        <div className={classNames(contentClasses)}>
           <h1 className="Dialog__title" id={dialogTitleId}>
             {title}
             <span className="u-stretch" />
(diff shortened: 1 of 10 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -57,6 +57,13 @@ export default function Dialog({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  const contentClasses = {
+    Dialog__content: true,
+  };
+  if (contentClasses) {
+    contentClasses[contentClass] = true;
```
LMS007, line 64: fixes the `undefined` class error

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -13,12 +13,15 @@ function emailLink({ address, subject = '', body = '' }) {
...
+  if (typeof error.details === 'object') {
+    try {
       details = JSON.stringify(error.details, null, 2 /* indent */);
+    } catch (e) {
+      // ignore
     }
-  } catch (e) {
-    // Ignored
+  } else {
+    details = error.details;
```
LMS007, line 24: this renders with `white-space: pre-wrap` and looks much better for plain text. 

TODO: Do we need escaping?

lms/static/styles/components/_Dialog.scss
```diff
@@ -26,10 +20,13 @@
 }
 
 .Dialog__content {
+  display: flex;
```
LMS007, line 23: moving the display flex here and setting the max height to 90vh allows for the dialogs to fill up the screen if they need to, but not beyond.

lms/static/styles/components/_Dialog.scss
```diff
@@ -26,10 +20,13 @@
...
+  flex-direction: column;
+  max-height: 90vh;
+  max-width: 700px;
   background-color: white;
   border-radius: 3px;
-  margin-left: auto;
-  margin-right: auto;
+  margin: auto;
```
LMS007, line 29: need auto margins on the top and bottom now too

lms/static/styles/components/_FileList.scss
```diff
@@ -14,6 +14,7 @@ $file-list-icon-size: 24px;
 
 .FileList__spinner {
   position: absolute;
+  margin: 15px;
```
LMS007, line 17: centers the spinner better

PR #1788, opened 2020-06-01
Record course settings in db

seanh:
Depends on https://github.com/hypothesis/lms/pull/1791

### DIFF
--- a/lms/models/course.py
+++ b/lms/models/course.py
@@ -1,0 +1,35 @@
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
+
+from lms.db import BASE
+from lms.models.application_settings import ApplicationSettings
+
+
+class Course(BASE):
+    """An LTI course."""
+
+    __tablename__ = "course"
+
+    #: The LTI consumer_key (oauth_consumer_key) of the application instance
+    #: that these course settings belong to.
+    consumer_key = sa.Column(
+        sa.String(),
+        sa.ForeignKey("application_instances.consumer_key", ondelete="cascade"),
+        primary_key=True,
+    )
+
+    #: The authority_provided_id that uniquely identifies the course that these
+    #: settings belong to.
+    authority_provided_id = sa.Column(sa.UnicodeText(), primary_key=True)
+
+    _settings = sa.Column("settings", MutableDict.as_mutable(JSONB), nullable=False)
+
+    @property
+    def settings(self):
+        """
+        Return this course's settings.
+
+        :rtype: models.ApplicationSettings
+        """
+        return ApplicationSettings(self._settings)
--- a/lms/models/course_groups_exported_from_h.py
+++ b/lms/models/course_groups_exported_from_h.py
@@ -1,0 +1,17 @@
+import sqlalchemy as sa
+
+from lms.db import BASE
+
+
+class CourseGroupsExportedFromH(BASE):
+    """
+    A read-only table holding a one-time export of groups from h.
+
+    See: https://github.com/hypothesis/lms/issues/1787
+
+    """
+
+    __tablename__ = "course_groups_exported_from_h"
+
+    authority_provided_id = sa.Column(sa.UnicodeText(), primary_key=True)
+    created = sa.Column(sa.DateTime, nullable=False)
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -34,6 +34,7 @@ class BasicLTILaunchViews:
     def __init__(self, context, request):
         self.context = context
         self.request = request
+        self.course_service = request.find_service(name="course")
 
         self.context.js_config.maybe_set_focused_user()
 
@@ -83,6 +84,7 @@ class BasicLTILaunchViews:
         """
         self.sync_lti_data_to_h()
         self.store_lti_data()
+        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
         self.context.js_config.add_canvas_file_id(self.request.params["file_id"])
         return {}
 
@@ -100,6 +102,7 @@ class BasicLTILaunchViews:
         """
         self.sync_lti_data_to_h()
         self.store_lti_data()
+        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
 
         self.context.js_config.maybe_enable_grading()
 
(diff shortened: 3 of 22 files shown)


lms/models/course.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        return ApplicationSettings(self._settings)
+
+    @classmethod
+    def get(cls, db, authority_provided_id, consumer_key):
+        """Return the course matching `authority_provided_id` and `consumer_key` or None."""
+        try:
+            return cls._get(db, authority_provided_id, consumer_key)
+        except NoResultFound:
+            return None
```
jon-betts, line 44: We only use the public version of `get` in the tests. I think we should collapse the two into one. I think we could do without the private one, and just check for return `None` instead of catching `NoResultFound` in `insert_if_not_exists` below.

lms/models/course.py
```diff
@@ -0,0 +1,77 @@
...
+        course = cls(
+            authority_provided_id=authority_provided_id,
+            consumer_key=consumer_key,
+            _settings=settings,
+        )
+
+        try:
+            cls._get(db, authority_provided_id, consumer_key)
+        except NoResultFound:
+            db.add(course)
```
jon-betts, line 67: The above feels like a service method to me. If we had a service for the getting and setting of courses, you could put the  `record_course` call in there too and manage the transfer of state (from application instance to course) in one place instead of the split here and with the LTI views.

lms/views/basic_lti_launch.py
```diff
@@ -227,3 +232,13 @@ def configure_module_item(self):
...
+
+
+def record_course(context, request):
+    """Add the current course to the course table in the DB."""
+    Course.insert_if_not_exists(
+        request.db,
+        context.h_group.authority_provided_id,
+        request.lti_user.oauth_consumer_key,
+        {"canvas": {"sections_enabled": context.canvas_sections_enabled}},
+    )
```
jon-betts, line 244: I think this might belong with the `insert_if_not_exists` in a course service. It would then be accessible from `content_item_selection` more naturally without being imported from here.

It would also keep the logic in one place. Currently it's split between here and the DB model.

I was surprised to not see the settings not being wholesale copied from the application instance here.

lms/models/course.py
```diff
@@ -0,0 +1,77 @@
...
+    #: settings belong to.
+    authority_provided_id = sa.Column(sa.UnicodeText(), primary_key=True)
+
+    #: The LTI consumer_key (oauth_consumer_key) of the application instance
+    #: that these course settings belong to.
+    consumer_key = sa.Column(
+        sa.String(),
+        sa.ForeignKey("application_instances.consumer_key", ondelete="cascade"),
+        primary_key=True,
+    )
```
jon-betts, line 25: Why don't we use the primary key ("id") of the application instance here? / Why does an application instance have one if the consumer key is the de-facto primary key?

seanh: I copied what we'd done elsewhere (in `models.oauth2_token`), but I think we want to be able to easily look up a `models.Course` by `request.lti_user.consumer_key` and `context.h_group.authority_provided_it`, and that's more convenient if `consumer_key` is directly on the `course` table rather than in a related table.

I dunno why `application_instance` has an `id` and doesn't just use the `consumer_key` as its primary key

lms/models/course.py
seanh: `record_course` is specific to a particular context (`LTILaunchResource`) though. I think you could register a service that's only available in that context. I haven't tried it before but I think something like this might work:

```python
    record_course_service_factory,
    context=LTILaunchResource,
    name="record_course",
)
```

Then I think the views would do something like:

```python
request.find_service(name="record_course").record_course()
```

The point is that any `request.find_service(name="record_course")` outside of an `LTILaunchResource` request would fail. (And in future other `"record_course"` service implementations could potentially be registered for other contexts.)

lms/views/basic_lti_launch.py
seanh: Would we want to assume that _all_ application instance settings (now and going forward) are overridable on a per-course basis?

lms/models/course.py
seanh: Yeah that'd work

seanh:
@jon-betts I found one issue with this PR, added a new commit to fix it. See the commit message: https://github.com/hypothesis/lms/pull/1788/commits/5c95e09689f37ecb7090db2c326149302b9c47cc

lms/views/basic_lti_launch.py
jon-betts: Which ever way you go will be wrong in a different circumstance, so we can't know which is the right path. Both could change but it makes the rule simple for now? It's easier to think about.

lms/models/course.py
jon-betts: Ah well. I wonder if we could just drop the column if it isn't actually used? (Not for now obvs)

lms/models/course.py
jon-betts: I doubt we'd ever need to register it twice, but I think a "course" service in general would be good.

There a comment on one of the commits:

> The way this is going to work in future is that
> ApplicationInstanceGetter.canvas_sections_enabled() is going to have two
> modes:
> 
> 1. ApplicationInstanceGetter.canvas_sections_enabled(course_settings)
>    will tell you whether sections should be enabled in the context of a
>    given course. This is going to be used to decide whether or not to
>    use sections when launching our app in a course.
> 
> 2. ApplicationInstanceGetter.canvas_sections_enabled() (with no
>    course_settings) argument will just tell you whether sections is
>    enabled for this application instance, without reference to a
>    particular course. This is what we should use to set the course's
>    canvas.sections_enabled setting the first time we see a new course.

I think this would be the natural home for the sections enabled question. Instead of asking the AIGetter to ask the course, ask the course directly (which may internally need to ask the AIGetter... or not)

lms/models/course_groups_exported_from_h.py
```diff
@@ -0,0 +1,17 @@
...
+
+class CourseGroupsExportedFromH(BASE):
+    """
+    A read-only table holding a one-time export of groups from h.
+
+    See: https://github.com/hypothesis/lms/issues/1787
+
+    """
+
+    __tablename__ = "course_groups_exported_from_h"
```
jon-betts, line 14: Love this ugly name! Should remind us to tidy up one day. Or will it just become part of the furniture. Who knows!

PR #1802, opened 2020-06-02
Refactor a parametrize and a feature flag fixture

seanh:
1. Make an `@parametrize` easier to read by having it use a params dict instead of long tuples, and by having it have default values for the items in the params dict so that each parametrized case only needs to contain the values the differ from the defaults. This is going to help with future changes to this test which are going to need to add more parameters and more cases: it'll now be possible to add these without making readability even worse.

2. Add a module level `enable_section_groups_feature_flag()` helper function and `section_groups_feature_flag_enabled` fixture. These just make it easier for tests that need to turn the flag on and off. In the future we're going to have to add more such tests to this module.

### DIFF
--- a/tests/unit/lms/services/application_instance_getter_test.py
+++ b/tests/unit/lms/services/application_instance_getter_test.py
@@ -69,26 +69,45 @@ class TestApplicationInstanceGetter:
         assert not ai_getter.provisioning_enabled()
 
     @pytest.mark.parametrize(
-        "section_groups_feature_flag,developer_key,canvas_sections_enabled,expected_result",
+        "params",
         [
-            (True, "test_developer_key", True, True),
-            (False, "test_developer_key", True, False),
-            (True, None, True, False),
-            (True, "test_developer_key", False, False),
-            (False, None, False, False),
+            dict(
+                feature_flag=True,
+                developer_key="test_developer_key",
+                canvas_sections_enabled=True,
+                expected_result=True,
+            ),
+            dict(
+                feature_flag=False,
+                developer_key="test_developer_key",
+                canvas_sections_enabled=True,
+                expected_result=False,
+            ),
+            dict(
+                feature_flag=True,
+                developer_key=None,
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/application_instance_getter_test.py
```diff
@@ -69,40 +69,38 @@ def test_provisioning_returns_False_if_consumer_key_unknown(self, ai_getter):
...
         [
-            (True, "test_developer_key", True, True),
-            (False, "test_developer_key", True, False),
-            (True, None, True, False),
-            (True, "test_developer_key", False, False),
-            (False, None, False, False),
+            dict(),
+            dict(feature_flag=False, expected_result=False),
+            dict(developer_key=None, expected_result=False),
+            dict(canvas_sections_enabled=False, expected_result=False),
```
jon-betts, line 77: What's the difference between this and a dict literal? 

Weird aside: I was looking it up and apparently literals are 2x the speed. Not relevant here, but interesting anyway.

jon-betts: As `expected_result` is mandatory it might be nice to keep that outside of the dict so we have an input output divide:

```python
({}, True),
({"feature_flag": False}, False),
({"developer_key": None}, False),
...
```

tests/unit/lms/services/application_instance_getter_test.py
```diff
@@ -69,40 +69,38 @@ def test_provisioning_returns_False_if_consumer_key_unknown(self, ai_getter):
...
-        def feature(flag):
-            if section_groups_feature_flag:
-                return flag == "section_groups"
-            return False
+        default_params = dict(
+            feature_flag=True,
+            developer_key="test_developer_key",
+            canvas_sections_enabled=True,
+            expected_result=True,
+        )
```
jon-betts, line 88: I was going to say you could do `get(..., default)` where it's needed, but it easier to compare with the parameterization above this way.

PR #1804, opened 2020-06-02
Minor styling and bug fixes for grader

LMS007:
- Remove custom focus styling on grader buttons.  Use `:focus-visible` instead.

- Fix bug where the success background color (green) only changed on the first submit, but not on successive submits.

- Fix  bug where submitting grade via keyboard `Enter` didn’t work after an error. This was due to the fact that 2 <buttons> were inside the same <form> (the second was the error validation). The error needs to stay in the DOM for animation purposes. The fix is to use <input> with type=“submit” vs type=“button” so it know which one to default to.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ValidationMessage.js
+++ b/lms/static/scripts/frontend_apps/components/ValidationMessage.js
@@ -34,9 +34,12 @@ export default function ValidationMessage({
   });
 
   return (
-    <button onClick={closeValidationError} className={errorClass}>
-      {message}
-    </button>
+    <input
+      type="button"
+      onClick={closeValidationError}
+      className={errorClass}
+      value={message}
+    />
   );
 }
 
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -145,6 +145,7 @@ export default function SubmitGradeForm({ disabled = false, student }) {
    */
   const handleKeyDown = () => {
     setValidationError(false);
+    setGradeSaved(false);
   };
 
   return (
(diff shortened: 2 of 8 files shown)


robertknight:
> dude to the fact we dynamically add this class, it causes a re-render and really has some less than ideal behavior with the focus state. It momentarily removes focus (because the input is re-built) and won't add it back again until keyboard input is detected.

I'm not clear what you are referring to here. Can you clarify what exactly the undesirable behavior is from a user's point of view? A screenshot might help.

In general, a property change for an element that was present in the previous render will just update the existing DOM element rather than re-create it. In other words, if a component rendered `<input class="a"/>` and then a subsequent state change causes it to render `<input class="a b"/>` that would just result in the existing DOM element being modified using `domElement.classList = "a b"`. Where this doesn't happen is if the element moves around in the DOM tree (say a `<li>` list item gets moved from one `<ul>` list to a different `<ul>` list). In that case it will get re-created.

lms/static/scripts/frontend_apps/components/ValidationMessage.js
```diff
@@ -34,9 +34,12 @@ export default function ValidationMessage({
...
   return (
-    <button onClick={closeValidationError} className={errorClass}>
-      {message}
-    </button>
+    <input
+      type="button"
+      onClick={closeValidationError}
+      className={errorClass}
+      value={message}
+    />
```
robertknight, line 42: This works, although you can set the `type` attribute on the `button` as well rather than change to `<input type="button">`. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/button) `<input type="button">` still works, although using a `<button>` element is preferred.

LMS007: Initially i tried that and I don't think it worked, but I will double check. Its quite possible I did something wrong and then just moved on to <input type=""

PR #1808, opened 2020-06-03
POC for decoupling enabled from supported for sections

jon-betts:
* The AI getter now reports if sections can be enabled
 * Separately you can get the AI and ask it if it is
 * The LTI launch resource now has a similar split and determines if it can be enabled
 * It then asks the course service for the course to see if it is


seanh:
Closing this because I've integrated it into https://github.com/hypothesis/lms/pull/1788

The only part that hasn't been integrated into https://github.com/hypothesis/lms/pull/1788 is actually checking the new per-course setting to decide whether or not to enable sections for a launch. This needs to be done in a separate PR 'cos it needs to be deployed separately:

```python
        course_service = self._request.find_service(name="course")
        course = course_service.get_or_create(self.h_group.authority_provided_id)
        return course.settings.get("canvas", "sections_enabled")
```

PR #1813, opened 2020-06-04
VitalSource integration prototype

robertknight:
This PR contains the current prototype of an integration with VitalSource. It adds a new type of assignment content (a VitalSource book) alongside the three existing types (URL, Google Drive, Canvas/LMS file storage):

Screenshot 2020-06-04 19 03 47

When the user selects "_Select a book from VitalSource_" we need to display some UI to select the book and content within it. This currently uses an existing course browser UI that VS has:

Screenshot 2020-06-04 19 04 16

It is currently TBD whether we continue to use this UI or use a data API and build a custom UI for browsing the content (if that's possible). Depending on the LMS, we may be very space constrained in this view. In the current prototype there is obviously a lot of space taken up by the dialog border which can be eliminated, but still, we won't have a lot of space to work with.

We need to somehow capture the book that was selected and the location within it. Currently the _Select_ button just saves a hardcoded book ID.

When an assignment using VitalSource is launched, instead of rendering an iframe that loads a web page or PDF proxied through Via,  we instead render an iframe that displays the VitalSource book viewer:

Screenshot 2020-06-04 19 02 40

The Hypothesis viewer is currently embedded in the viewer with the default configuration, hence it isn't using account integration.

Authentication for both VitalSource's book viewer and content selection web apps are handled via an LTI launch:

1. Our LTI tool frontend renders an iframe with a local URL (`/vitalsource/book_launch` or `/vitalsource/book_selection_launch`)
2. This URL renders a tiny web page with an HTML form which has hidden parameters for an LTI launch, which is automatically submitted as soon as the page loads. This process is similar to how Canvas launches us
3. When the form is submitted, VS's web app verifies the LTI launch parameters and then shows their viewer with Hypothesis embedded.

----

**TODO items**

1. Update the Hypothesis configuration in the viewer so that it fetches configuration from the parent iframe.
2. Decide on an approach for the book + chapter selection (API vs. existing VS UI for browsing books)
3. The VS viewer currently always loads the _production_ Hypothesis client, whereas for other assignment types it will load the local development, QA or prod clients as appropriate. We need to figure out how we'll handle this.
4. We need to make sure the client captures appropriate document location metadata so that we can re-anchor annotations reliably in future
5. We need to figure out how to enable the client to navigate the viewer to annotations that are on different pages than the one that is currently visible.
6. The prototype currently uses a single pair of OAuth credentials to launch the VitalSource viewer. We'll need to instead figure out how to use credentials which enable the user to access appropriate materials depending on the user/institution etc.

### DIFF
--- a/lms/config/__init__.py
+++ b/lms/config/__init__.py
@@ -62,6 +62,10 @@ def configure(settings):
         # like this:
         #     python3 -c 'import secrets; print(secrets.token_hex())'
         "oauth2_state_secret": sg.get("OAUTH2_STATE_SECRET"),
+        # Default OAuth 1.0 key and secret for VitalSource LTI launches
+        "vitalsource_launch_key": sg.get("VITALSOURCE_LAUNCH_KEY"),
+        "vitalsource_launch_secret": sg.get("VITALSOURCE_LAUNCH_SECRET"),
+        "vitalsource_api_key": sg.get("VITALSOURCE_API_KEY"),
     }
 
     env_settings["dev"] = asbool(env_settings["dev"])
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -75,6 +75,17 @@ class JSConfig:
             self._config["viaUrl"] = via_url(self._request, document_url)
             self._add_canvas_speedgrader_settings(document_url=document_url)
 
+    def add_vitalsource_launch_url(self, book_id, cfi=None):
+        vitalsource_svc = self._request.find_service(name="vitalsource")
+        launch_url, launch_params = vitalsource_svc.get_launch_params(
+            book_id, cfi, self._request.lti_user
+        )
+        self._config["vitalSource"] = {
+            "launchUrl": launch_url,
+            "launchParams": launch_params,
+        }
+        # TODO - Add SpeedGrader settings for Canvas.
+
     def asdict(self):
         """
         Return the configuration for the app's JavaScript code.
--- a/lms/resources/default.py
+++ b/lms/resources/default.py
@@ -8,6 +8,7 @@ class DefaultResource:
     __acl__ = [
         (Allow, "report_viewers", "view"),
         (Allow, "lti_user", "api"),
+        (Allow, "lti_user", "vitalsource_api"),
     ]
 
     def __init__(self, request):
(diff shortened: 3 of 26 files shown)


judell:
Nice start!

robertknight:
When the instructor creates an assignment with Hypothesis, there is a new VitalSource option, as in the earlier iteration:

vs-content-type-selection

When selected, that now shows a new UI that presents the list of available books, which we fetch from VitalSource's "products" API:

vs-book-selection

After selecting a book, the table of contents for that book is fetched and the user is prompted to select a chapter to use for the assignment:

vs-chapter-selection

After the selection is confirmed the assignment is created. When launched, our LTI tool will launch the VitalSource viewer with the selected book, opened at the chosen chapter:

vs-viewer-selecting-text

When the user clicks "Annotate", the Hypothesis client opens and the user can create an annotation:

vs-viewer-creating-annotation

In the screenshot above the default "Public" group is selected, but what should happen is that the Hypothesis client gets logged into an account that corresponds to the user's LMS identity. That requires a small piece of configuration in the VitalSource book viewer.

In terms of the reading/annotating experience, there are a number of issues we need to resolve:

1. VitalSource's own text selection popup gets in the way of Hypothesis's popup
2. We need to decide what identifiers the Hypothesis client should extract from the page (eg. ISBN or book ID, chapter etc.) and store those as metadata with the annotation
3. We need to decide what should happen if the user navigates away from the original chapter
4. ... plus we need to look at any other polish to the overall experience

robertknight:
So the whole thing is now working end-to-end on this branch, except for the missing client configuration in the VitalSource viewer - hopefully that's only a small change. The next step I think is to polish the design a little, tidy up the code and get it landed behind a feature flag so that various stakeholders can try it for themselves and we can have a conversation about what to do next.

robertknight:
Collection of design feedback from Lyza so I don't forget: https://hypothes-is.slack.com/archives/C07NXBDNW/p1591974968069600?thread_ts=1591973079.064500&cid=C07NXBDNW

robertknight:
Status update 2020-07-31: Pieces of this PR are currently being extracted and integrated separately. I'm going to keep this proof-of-concept PR open while that is done.

robertknight:
There is an ongoing Slack discussion about extracting a smaller PR from this one which just adds the minimum functionality to the LMS app to enable creating a VitalSource assignment which launches a specific, hard-coded book at a specific chapter. This will enable the VS team to see / test the integration for themselves.

robertknight:
The work in this prototype is going to be broken out into smaller chunks and shipped incrementally. The first PR in the series is https://github.com/hypothesis/lms/pull/2374. I will leave the branch around for a little while so that I can cherry-pick pieces of it into subsequent PRs.

PR #1820, opened 2020-06-08
GroupInfo info defaults should be applied on read, not create

jon-betts:
Not sure how we got here, but groups infos in the wild don't have the defaults set. In general I think we must always assume this value is blank.

The result of this is reading a GroupInfo with "info = None" will cause a crash. ~This isn't an issue with using JSON fields, it's an issue of not using them correctly. We must never assume a key is present, or have a much better strategy for trying to setup the defaults.~ Seems we should have set this, although I still feel this is a risky way to go.

~I'm not sure if the DB migration was supposed to catch this. Working out how we got here can probably wait.~ It was, so how this has happened is a mystery. A real possibility is something we aren't thinking of is resulting in this getting blanked.

### DIFF
--- a/lms/models/group_info.py
+++ b/lms/models/group_info.py
@@ -84,6 +84,12 @@ class GroupInfo(BASE):
     #: A dict of info about this group.
     info = sa.Column(MutableDict.as_mutable(JSONB))
 
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("info", {})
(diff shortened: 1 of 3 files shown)


jon-betts:
It looks like the PR that added the column should have set the instructors key in the dict as well.

jon-betts:
Yep, not sure what's up. I'll get it out. My worry is that without understanding we're probably fixing the wrong bug.

PR #1822, opened 2020-06-09
Don't crash when dev key lacks scopes

seanh:
Fixes https://github.com/hypothesis/lms/issues/1821

`CanvasAPIAuthorizeViews` is a little confusing and error-prone because:

* Some of the class's view methods (`authorize()` and
  `oauth2_redirect()`) are authenticated: `request.lti_user` is set for
  these views

* But the class's third view method (`oauth2_redirect_error()`) is not
  authenticated: `request.lti_user` is `None` for requests to this view.

The lack of authentication for the error view is unavoidable: the
request comes from Canvas (which redirects the popup window to us) and
is out of our control, and the request doesn't contain any kind of
parameter that we could use to authenticate the user.

The crash in https://github.com/hypothesis/lms/issues/1821 was caused
when a seemingly harmless refactor
(https://github.com/hypothesis/lms/pull/1812/) moved these two lines:

    self.request.find_service(name="ai_getter")
    self.request.find_service(name="course")

from the body of a view method into
`CanvasAPIAuthorizeViews.__init__()`. The `__init__()` is executed
whenever any of the class's views is requested, so this change meant
that all three views now depended on the `ai_getter` and `course`
services.

Both the `ai_getter` and `course` service factories depend on
`request.lti_user` being set. Since `request.lti_user` is `None` for the
`oauth2_redirect_error()` view, this meant that the error view now
crashes.

The `CanvasAPIAuthorizeViews` class has caused other confusions in the
past as well. For example many of its `@view_defaults` only apply to
some of its views and are overridden by others. And the inherting of
view defaults by exception views, though perfectly logical, has
surprised us in the past as well.

So I've decided to fix this by turning the three views into unrelated
view functions, and getting rid of the containing class.

### DIFF
--- a/lms/views/api/canvas/authorize.py
+++ b/lms/views/api/canvas/authorize.py
@@ -12,107 +12,106 @@ authenticated server-to-server requests to Canvas.
 from urllib.parse import urlencode, urlparse, urlunparse
 
 from pyramid.httpexceptions import HTTPFound, HTTPInternalServerError
-from pyramid.view import exception_view_config, view_config, view_defaults
+from pyramid.view import exception_view_config, view_config
 
 from lms.services import CanvasAPIServerError
 from lms.validation.authentication import BearerTokenSchema, CanvasOAuthCallbackSchema
 
+#: The Canvas API scopes that we need for our Canvas Files feature.
+FILES_SCOPES = (
+    "url:GET|/api/v1/courses/:course_id/files",
+    "url:GET|/api/v1/files/:id/public_url",
+)
 
-@view_defaults(request_method="GET", route_name="canvas_oauth_callback")
-class CanvasAPIAuthorizeViews:
+#: The Canvas API scopes that we need for our Sections feature.
+SECTIONS_SCOPES = (
+    "url:GET|/api/v1/courses/:id",
+    "url:GET|/api/v1/courses/:course_id/sections",
+    "url:GET|/api/v1/courses/:course_id/users/:id",
+)
 
-    # The Canvas API scopes that we need for our Canvas Files feature.
-    files_scopes = (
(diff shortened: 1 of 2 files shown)


lms/views/api/canvas/authorize.py
```diff
@@ -12,107 +12,106 @@
...
+
+    scopes = FILES_SCOPES
+
+    if ai_getter.canvas_sections_supported() and (
+        # If the instance could add a new course with sections...
+        ai_getter.settings().get("canvas", "sections_enabled")
+        # ... or any of it's existing courses have sections
+        or course_service.any_with_setting("canvas", "sections_enabled", True)
+    ):
+        scopes += SECTIONS_SCOPES
```
seanh, line 49: This block was previously `self._required_scopes()`. I just inlined it back into here, which is the one place that uses it. It doesn't seem worth having a separate `_required_scopes()` helper function, especially since you'd have to pass `ai_getter` and `course_service` into it

lms/views/api/canvas/authorize.py
```diff
@@ -12,107 +12,106 @@
...
+#: The Canvas API scopes that we need for our Sections feature.
+SECTIONS_SCOPES = (
+    "url:GET|/api/v1/courses/:id",
+    "url:GET|/api/v1/courses/:course_id/sections",
+    "url:GET|/api/v1/courses/:course_id/users/:id",
+)
+
+
+@view_config(
+    request_method="GET", route_name="canvas_api.authorize", permission="canvas_api"
```
seanh, line 35: Some of these view config params are duplicated on the different functions now that we can no longer use `@view_defaults`. Things are less confusing as a result, though

Issue #1826, opened 2020-06-09
Screen "flicker" when closing authorization popup window


seanh:
I don't think this used to happen: when I close the authorization popup window you can see that the entire iframe flashes (would be the entire browser tab if the assignment were configured to open in a new tab):

![Peek 2020-06-09 14-51](https://user-images.githubusercontent.com/22498/84155844-b9dbe600-aa60-11ea-83fe-afc050c67ef5.gif)

robertknight:
There is an open PR in https://github.com/hypothesis/lms/pull/1844 but it is awaiting re-review from me.

robertknight:
https://github.com/hypothesis/lms/pull/1844 resolved the flicker, however there is an issue that after the authorization popup is completed, the "Authorize Hypothesis" dialog briefly remains visible until both the groups and content are fetched, but there is no clear indication that anything is happening:

![Transition after authorizing](https://user-images.githubusercontent.com/2458/90742127-46a7e900-e2c7-11ea-8b25-918950d789be.gif)

The "Authorize" button in the dialog gets disabled while the iframe content URL is fetched after OAuth authorization completes, but not while the groups are being fetched. This means that if the content URL is already available or is fetched more quickly than the groups then the "Authorize" button remains enabled.

The issue in the code is to do with a difference between the `fetchGroups` and `fetchContentUrl` functions in `BasicLtiLaunchApp`. The `fetchContentUrl` unconditionally updates a counter of "API requests that block display of the content" (the `fetchContent` state) whereas `fetchGroups` does not, because whether fetching groups blocks display of the content depends on whether it happens on the initial launch (where it does not) or afterwards (where it does).

PR #1833, opened 2020-06-12
UI improvements for dialog, spinner and error state of LMS app

LMS007:
The motivation for the extent of these changes comes from fixing a flickering issue when the error state jumps no error to an error between waiting for authorization of the popup window and receiving a second error from the api.

The BasicLtiLaunchApp is reworked to not reset error state on outgoing requests and instead resets error state value only after subsequent requests come back and we know there are no new errors.

The groups and contentUrl fetchers were previously fully atomic and took care of resetting the error state. They now return an error code which we can use to prevent flicker because the caller (authorizeAndFetchUrl) can now clear the error state rather than these two de-coupled requests.

Additionally some improvements were made to the spinner. These are mostly superficial but improve the UI. There is now an additional <Spinner> in the error dialog in addition to the full page spinner in BasicLtiLaunchApp. This has the advantage of showing the pending
requests while a previous error (if triggered) is still being displayed without rendering a huge spinner over the dialog.

One minor blind spot is if the initial requests fail and invoke an error. We don't keep track of which requests the loading states were triggered by. So for example, if the groups
errors first, then later the contentUrl resolves or errors, there will be a very short time the small spinner is rendered on the error dialog. This is unintended and cant not be solved unless we have a way of structuring the loading state with the errors that triggered them. This could be done either by a transaction or some queue or list. This would only add complexity and could be fixed later.

Additional fixes:

  - Spinner no longer operates by being removed from DOM. Now has `visible` prop. This gives it the advantage of having css animations. In some cases where we render a full background the spinner, it still gets removed from the DOM --that can be improved later to facilitate animations.
  - Allow styling to be set on the spinner's wrapper to adjust position from the caller component's css.
  - Refactor all instances of the spinner's css to use mixins.spinner()
  - Minor refactors on several tests that needed changes

----------

Test with these two

   https://hypothesis.instructure.com/courses/121/assignments/851

fixes https://github.com/hypothesis/lms/issues/1826

You can see when slowing down the network request, we don't see the double loader when I refresh the whole page. But when it's faster (no throttling), it can be seen. Thats just that "blind spot" I was talking about. 

canvas_851

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -68,15 +68,18 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   // the app's access to the user's files in the LMS.
   const authWindow = useRef(null);
 
+  // Helpers
+  const hasError = !!errorState;
+
   // Show the assignment when the contentUrl has resolved and errorState
   // is falsely
-  const showIframe = contentUrl && !errorState;
+  const showIframe = contentUrl && !hasError;
 
-  const showSpinner = fetchCount > 0 && !errorState;
+  const showPageSpinner = fetchCount > 0 && !hasError;
+  const showDialogSpinner = fetchCount > 0 && hasError;
 
   const incFetchCount = () => {
     setFetchCount(count => count + 1);
-    setErrorState(null);
   };
 
   const decFetchCount = () => {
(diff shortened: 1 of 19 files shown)


seanh:
Testing using the Canvas File assignment in the Missing Developer Key Scopes test course: https://hypothesis.instructure.com/courses/121/assignments/851


Peek 2020-06-15 09-06


I see two visual issues when loading an assignment if the access token is missing:

1. It's too fast to see in the GIF but the "full screen" spinner actually shows very very briefly when the iframe first loads, and causes a flicker. (Doesn't happen on master.)

2. This is visible in the GIF: the new in-button spinner appears briefly, but is already fading out when it first appears.

In the case shown in the GIF the API responds very quickly (the LMS app doesn't have a token in the DB so it responds with an error immediately, without sending any requests to Canvas). The response is too quick for any spinner to make sense.

The correct behaviour in this case should be that neither spinner is shown at all, and that seems to be what happens on master

seanh:
Testing using the Canvas Files Assignment in the Sections Disabled test course: https://hypothesis.instructure.com/courses/124/assignments/868

Peek 2020-06-15 09-15

1. The full-screen loading spinner has changed compared to master. It's much bigger and a different colour.

2. The full-screen spinner remains visible, fading out, after PDF.js, the PDF and the client have already begun appearing. I don't think this is necessary

seanh:
Testing using the Canvas Files Assignment in the Sections Enabled test course: https://hypothesis.instructure.com/courses/125/assignments/875

Also, my refresh token has expired which causes an authorization error on assignment load.

Peek 2020-06-15 09-19

1. The full-screen spinner and in-button spinner are shown at the same time. There should never be two spinners at once

2. The in-button spinner shouldn't be shown at all in this case. When the "something went wrong" dialog appears it has already received an error response from the API. There isn't an API request in-flight at this time. So I think the in-button spinner is being shown incorrectly

seanh:
When loading the list of Canvas Files when creating an assignment, the spinner remains visible after the files have loaded because it's fading out:

Peek 2020-06-15 09-26

seanh:
The spinner has changed and has gotten really big when creating a Google Drive assignment:

Peek 2020-06-15 09-28

seanh:
One thing that seems to contribute to a number of the issues above is that it looks like both of the loading spinners now fade _out_ over a few hundred milliseconds.

I think there _may_ be an argument for having the spinners fade _in_ when they first appear. There's an issue with loading spinners that they can flicker if you show the spinner and then the thing you're waiting for (network response or whatever) comes immediately after you just showed the spinner so you immediately hide the spinner again. Having the spinner fade in may reduce the flickering effect in these instances. (I'm not sure, I think I'd have to see it.)

But I'm not sure that there is ever a case for having a spinner fade _out_? This seems to just introduce cases where the thing that was loading (the assignment, the list of files, the "something went wrong" dialog, etc) is already shown and the spinner is still visible over top of it because it's taking time to fade out. It seems that you just want the spinner to disappear immediately in these cases.

I don't think fading out is the only issue here but I think it may be a contributing factor in multiple places.

robertknight:
Hi @LMS007 - My understanding from the PR description is that this PR combines a fix for #1826 with cosmetic changes to the spinner which are related to the issues @seanh mentioned above. Can you split out these changes into separate PRs so that it is easier for me to review them individually?

LMS007:
> e correct behaviour in this case should be that neither spinner is shown at all, and that seems to be what happens on master

The drive assignment was missed for sure, I don't think I have that set up locally, need to do that.

I'm not sure how to solve the flicker. The spinner does not cause the flicker it is the white background changing to grey on the dialogs. We could make the default background grey. I'm also not sure how to guess when an error will happen. We can delay the loader 100sm or so, but in theory, the error could come at 101ms and we're still seeing the issue, but perhaps less?

In any case, I'm closing this PR as it does not fix issues.

PR #1740, opened 2020-05-12
Automate and remove "make services"

seanh:
Whenever you run a command that needs the DB (dev server, tests, functests, bddtests, ...) have it start the Postgres container automatically if it isn't already running. Works whether you're running `make` commands or running `tox` commands directly.

This will check off _you shouldn't have to run make services manually_ on the dev env roadmap: https://github.com/hypothesis/product-backlog/issues/1044

The same solution will also work for h, but we'll need to add support for Elasticsearch and RabbitMQ.

When the containers are already running the script adds very little overhead:

```terminal
$ time bin/start-services

________________________________________________________
Executed in  258.44 millis    fish           external 
   usr time   58.97 millis    0.00 micros   58.97 millis 
   sys time   34.13 millis  1451.00 micros   32.68 millis 
```

Some handy commands for testing this:

Stop the docker container:

    tox -qqe docker-compose --run-command 'docker-compose stop'

Stop the container and also delete your DB:

    tox -qqe docker-compose --run-command 'docker-compose down'

Run just a couple of tests:

    tox -qq --run-command 'pytest tests/unit/lms/app_test.py'

PR #1834, opened 2020-06-15
Automate `make services`

seanh:
Make `make` and `tox` commands that require the services (Postgres) to be running automatically start the services if they aren't already running, wait for the services, and then continue.

This eliminates the need to manually type `make services`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,6 @@
 .PHONY: help
 help:
 	@echo "make help              Show this help message"
-	@echo 'make services          Run the services that `make dev` requires'
-	@echo "                       (Postgres) in Docker"
 	@echo "make dev               Run the entire app (web server and other processes)"
 	@echo "make web               Run the web server on its own (useful for debugging the "
 	@echo "                       Python code with pdb)"
--- a/README.markdown
+++ b/README.markdown
@@ -56,12 +56,6 @@ installation process:
 
     cd lms
 
-### Run the services with Docker Compose
-
-Start the services that the LMS app requires using Docker Compose:
-
-    make services
-
 ### Create the development data and settings
 
 Create the database contents and environment variable settings needed to get lms
--- a/bin/check-services
+++ b/bin/check-services
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+"""
+Exit with 0 if all the given services are healthy, 1 otherwise.
+
+Usage:
+
+    bin/check-services lms_postgres_1 ...
+
+"""
+from sys import argv, exit
+from subprocess import check_output, DEVNULL
+
+
+def check_service(service):
+    try:
+        status = (
+            check_output(
+                [
+                    "docker",
+                    "inspect",
+                    "-f",
+                    "{{.State.Health.Status}}",
+                    service,
+                ],
+                stderr=DEVNULL,
+            )
+            .decode()
+            .strip()
+        )
+    except:
+        pass
+    else:
+        if status == "healthy":
+            return True
+
+    return False
+
+
+if __name__ == "__main__":
+    for service in argv[1:]:
+        if not check_service(service):
+            exit(1)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,6 @@ services:
       - '127.0.0.1:5433:5432'
     healthcheck:
         test: ["CMD", "pg_isready", "-U", "postgres"]
-        interval: 0.1s
+        interval: 1s
         timeout: 1s
-        retries: 10
+        retries: 1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,5 +7,3 @@ services:
     healthcheck:
         test: ["CMD", "pg_isready", "-U", "postgres"]
         interval: 1s
-        timeout: 1s
-        retries: 1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,3 +4,8 @@ services:
     image: postgres:11.5-alpine
     ports:
       - '127.0.0.1:5433:5432'
+    healthcheck:
+        test: ["CMD", "pg_isready", "-U", "postgres"]
+        interval: 0.1s
+        timeout: 1s
+        retries: 10
--- a/bin/start-and-wait-for-services
+++ b/bin/start-and-wait-for-services
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+"""
+Start the given services and wait for them to become healthy.
+
+Usage:
+
+    bin/start-and-wait-for-services lms_postgres_1 ...
+
+"""
+from os import environ
+from sys import argv, exit
+from subprocess import run
+
+if "JENKINS_URL" in environ:
+    exit()
+
+for command in ("start-services", "wait-for-services"):
+    run([f"bin/{command}"] + argv[1:])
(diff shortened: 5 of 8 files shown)


Makefile
```diff
@@ -1,8 +1,6 @@
 .PHONY: help
 help:
 	@echo "make help              Show this help message"
-	@echo 'make services          Run the services that `make dev` requires'
-	@echo "                       (Postgres) in Docker"
```
seanh, line 5: `make services` still exists but is now undocumented because you never need to run it manually

README.markdown
```diff
@@ -56,12 +56,6 @@ installation process:
 
     cd lms
 
-### Run the services with Docker Compose
-
-Start the services that the LMS app requires using Docker Compose:
-
-    make services
-
```
seanh, line 64: You don't need to run this manually as part of the installation process anymore

bin/check-services
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
seanh, line 1: I wrote all the scripts in Python because it's just so much easier to write. Maybe they'd be a bit faster in bash? Not sure. They'd certainly be much harder to write

bin/check-services
```diff
@@ -0,0 +1,42 @@
...
+            check_output(
+                [
+                    "docker",
+                    "inspect",
+                    "-f",
+                    "{{.State.Health.Status}}",
+                    service,
+                ],
+                stderr=DEVNULL,
+            )
```
seanh, line 26: This `docker inspect` command can be used to check the `healthcheck` of a container:

```terminal
$ docker inspect -f {{.State.Health.Status}} lms_postgres_1
health
```

docker-compose.yml
```diff
@@ -4,3 +4,8 @@ services:
     image: postgres:11.5-alpine
     ports:
       - '127.0.0.1:5433:5432'
+    healthcheck:
+        test: ["CMD", "pg_isready", "-U", "postgres"]
+        interval: 0.1s
+        timeout: 1s
+        retries: 10
```
seanh, line 11: Using Docker / Docker Compose's healthcheck thing allows each container's healthcheck command to be specified with the container in `docker-compose.yml`.

Also, it turns out Postgres has a builtin `pg_isready` for checking it.

seanh:
> The built in mechanism for docker health looks good, but 10x per second seems aggressive. Presumably this happens all the time when running in dev

Tweaked these: https://github.com/hypothesis/lms/pull/1834/commits/daeb55aaacff56e799fc4a6840ecb5334cbb11e4 and https://github.com/hypothesis/lms/pull/1834/commits/8e6a03f3bbe671ea65aeca9e21459c7200be89ba

* Interval (how long after the container starts until the first healthcheck is run, also how long until each successive healthcheck): 1s. I think there's a tradeoff here: a longer interval period would mean checks are happening less frequently all the time, but it also means that waiting for services to become healthy when the container first starts up takes longer. A longer interval only introduces a delay when containers first start up though. Subsequent `make <something>` or `tox <something>` commands that run after the container already started, will be fast regardless of the interval

* Timeout (how long it will wait for a healthcheck command to return): 30s. Our healthcheck command is always very fast, but there seems little point in making the timeout tight, if it one time runs slow for some reason it seems better to wait for it and then work than to timeout and fail. I actually removed `timeout` from the file because 30s is the default

* Retries (number of failed healthchecks until it's considered unhealthy): 3. Same reasoning as for timeout. Also removed retries from the file since 3 is the default

Docs for healthcheck options: https://docs.docker.com/engine/reference/builder/#healthcheck

Timings:

`bin/start-and-wait-for-services` seems to take about 5s to start and then wait for Postgres if it's not already running. When Postgres is already running (which is what will happen most times when you run `make <something>` or `tox <something>`) it seems to take 250-500ms and vary a lot. That's what this will add to every `make <something>` or `tox <something>` command

seanh:
> `bin/start_services` basically is just make services and can go

This isn't quite true. `bin/start_services`:

1. Accepts a list of one or more services as command line arguments
2. Checks whether the services are already running (by calling `bin/check-services` forwarding the command line arguments to it)
3. If any of the services aren't already running then runs `make services` (which is just `docker-compose up -d`)

Whereas `make services`:

1. Accepts the args to be passed to docker-compose as command line args. Defaults to `up -d` if no args are given
2. Runs `tox -qe docker-compose -- <ARGS>`

So for example I often do `make services args=down` to take down the containers (instead of running `tox -qe docker-compose -- down`), although I could easily get out of that habit. `bin/create-db` also uses this, it runs `make services args="exec postgres psql ..."`. I think those are the only places where passing args to `make services` is currently used.

seanh:
> As per https://github.com/hypothesis/lms/pull/1762/files#diff-b67911656ef5d18c4ae36cb6741b7965R32 - make services can also easily be adapted to remove the need for bin/start-and-wait-for-services (it's a one liner does that all)

I think you may want to make it a multi-line script within `Makefile` if you did that, rather than a one-liner. The overall logic of `bin/start-and-wait-for-services` (including the `bin/start-services` and `bin/wait-for-services` scripts that it calls) is something like:

1. Exit successfully if `JENKINS_URL` envvar is present
2. Exit successfully if `bin/check-services <services>` exits successfully (the services are already running)
3. Run `tox -qe docker-compose -- up -d`
4. Repeatedly run `bin/check-services <services>` until it exits successfully and then exit successfully. Or exit unsuccessfully after a set number of repeats

But this is the same single-script logic as in my [`bin/start-services`](https://github.com/hypothesis/lms/blob/c2fca7a3a9a49e87e9d3819ac30d091e38cff0d5/bin/start-services) script in my previous PR for this, but I thought you [wanted it to be split up into separate scripts for separate concerns?](https://github.com/hypothesis/lms/pull/1740#pullrequestreview-413597077)

We could edit `make services` to replace `bin/start-and-wait-for-services` _only_ and not the other scripts. So an inline script within the `Makefile` like:

1. Exit successfully if `JENKINS_URL` envvar is present
2. Run `bin/start-services <services>`
3. Run `bin/wait-for-services <services>`

But then `tox.ini` would have to run `make services`, which would introduce a dependency of tox on make. Currently make depends on tox, not the other way round. I think in some cases that might matter. For example IIRC h's `Jenkinsfile` calls `tox` directly because `make` isn't installed in the environment that's in use on Jenkins. We might be able to fix that but I think a tox -> make dependency might be generally undesirable?

seanh:
I find bash scripts very difficult to write as soon as they become even remotely non-trivial. I just don't know bash, and also bash is pretty horrible to be fair. So things like how to strip a string and then check whether it's `"healthy"`, loops, reading command line args and passing them through to other scripts, running another script and then branching depending on that script's exit status, etc, are all a lot easier for me to write in Python.

At the least I'd like to request that we leave these in Python for now and consider a future PR to rewrite them in bash. But to be honest I think we'd have easier lives if we wrote shell scripts in Python generally

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/1846.

We've decided _not_ to automate `make services` in the individual repos because doing so would add a 250-500ms delay to every `make` and `tox` command to check whether or not Postgres is running. The delay is due to Docker and Postgres themselves, not due to our own scripting (tox or Python or anything). In h where there are three services not one (Postgres, Elasticsearch and RabbitMQ) the delay would be even worse.

However, we did decide to add `healthcheck` configs to the services in `docker-compose.yml` so that https://github.com/hypothesis/superdev can use these. https://github.com/hypothesis/lms/pull/1846 does that

Issue #1824, opened 2020-06-09
Error details aren't shown when "Error Details" button is clicked


seanh:
Go to the [Missing Developer Key Scopes test course](https://hypothesis.instructure.com/courses/121) and launch a Canvas Files assignment. You'll be asked to authorize, and in the popup window you'll be taken to an error dialog.

Clicking on the _Error Details_ button doesn't reveal the error details. It does expand the _Error Details_ section but, since it expands off the bottom of the window, the error details aren't actually in view until the user notices that they have to scroll down.

![Peek 2020-06-09 14-39](https://user-images.githubusercontent.com/22498/84154434-04f4f980-aa5f-11ea-82aa-439ba8b75900.gif)

You can also see in this GIF that the scrollbar is shown when it isn't needed. I made a separate issue for that: https://github.com/hypothesis/lms/issues/1823

**You can also see that the width and height of the white dialog both change a little bit when the error details are expanded**, which is weird.

robertknight:
I've marked this issue as closed. However there is a caveat.

The issue is fixed in most instances of the error dialog but I realised it does not apply for the specific case of the error shown in the screenshot above because that dialog is server-rendered HTML rather than the Preact component we use it most places. As such it doesn't inherit any of the logic built into the `ErrorDialog` component.

robertknight:
I'm reopening this while I look into the above.

robertknight:
This was resolved by https://github.com/hypothesis/lms/pull/1835, which switches the dialog shown in the issue to the client-rendered one. The client rendered dialog has logic to scroll the "Error details" box into view when expanded.

PR #1835, opened 2020-06-15
Convert Canvas auth error dialog to a frontend mini-app

robertknight:
The Canvas API authorization error dialog re-used the visual style of
error dialogs elsewhere in our frontend applications but was server
rendered, re-using some of the same CSS and markup structure from client
components.

As the styling and markup of our base `Dialog` and `ErrorDisplay`
components has become more complex, we've encountered several scenarios
where the server and client rendered versions got out of sync or
functionality added to the client-rendered version was not present in
the server-rendered one.

This commit resolves the issue by converting the Canvas auth error
dialog into a mini-SPA which re-uses the existing frontend UI
components.

Original context: https://github.com/hypothesis/lms/issues/1824#issuecomment-644152263

----

**User-visible changes:**

Error seen when auth fails due to missing scopes:

canvas-auth-error-missing-scopes

New feature due to re-using the existing frontend components, clicking "Error details" scrolls the content into view:

canvas-auth-error-details

Example of a non-scope related error:

canvas-auth-error-wrong-dev-secret

One change compared from before is that the previous version of the dialog had code to show a "Try again" button in certain circumstances. I couldn't reproduce this with the various test courses and assignments that we have, and I also don't think it is that important as the user can simply close the dialog and click the "Authorize" button again in the window behind it. Therefore I have removed that button.

**Developer-facing changes**

1. The `JSConfig` resource helper no longer defaults to the `basic-lti-launch` mode. Instead it will only populate generic properties which are applicable to all frontend apps when initialized. The properties specific to the basic LTI launch are only set when `js_config.enable_lti_launch_mode()` is called, which is now done in the `BasicLTILaunchViews` constructor
2. The `JSConfig` resource now has an additional mode for the Canvas auth error dialog, and there is an associated `FrontendAppResource` context type
3. The Canvas auth error views now configure the frontend app rather than rendering the dialog as HTML in the templates
4. A new mode/route has been added to the frontend app in `index.js` which renders the `CanvasAuthErrorApp` root component

I think the `JSConfig` class could further refactoring in future as putting the logic for configuring all the different frontend apps into one route is going to get increasingly messy as additional frontend mini-apps are added. I didn't try to do that here to limit the size of the PR.


lms/views/api/canvas/authorize.py
```diff
@@ -89,6 +89,16 @@ def oauth2_redirect(request):
...
 
 
+# Stripped down version of `lms.resources._js_config.JSConfig` implementing
+# the interface used by the `base.html.jinja2` template.
+class JSConfig:
+    def __init__(self, config):
+        self.config = config
+
+    def asdict(self):
+        return self.config
```
robertknight, line 99: The `lms.resources._js_config.JSConfig` class is heavily geared towards the LTI launch scenarios and sets a lot of properties that would not be relevant if we were re-using `context.js_config` as a way for the Python backend app to pass configuration to the frontend in non-LTI launch contexts. If we are going to do this, I could use input on how best to make `context.js_config` more general.

lms/static/scripts/frontend_apps/index.js
```diff
@@ -6,21 +6,33 @@ import { createElement, render } from 'preact';
...
+    {config.mode === 'canvas-auth-error-dialog' && (
+      <CanvasAuthErrorDialog
+        authorizeUrl={config.authorizeUrl}
+        details={config.details}
+        invalidScope={config.invalidScope}
+        scopes={config.scopes}
+      />
+    )}
   </Config.Provider>,
   rootEl
```
robertknight, line 37: We currently bundle all of the frontend apps into one bundle. This currently works OK as the bundle isn't that large. Also, when the user sees the OAuth error dialog, their browser will have already recently fetched and cached the bundle as part of the LTI launch. If however we added a bunch more mini-apps, especially ones with a substantial amount of code, we would want to split the bundle up by route.

robertknight:
@seanh - This can wait until after the off-site if you're busy, but I could use your input on [the js_config related parts of this](https://github.com/hypothesis/lms/pull/1835#discussion_r440252873). Also if you have any comments or questions on the general direction, feel free to add them.

seanh:
Our approach has been to centralize `JSConfig` and related logic (around how the config changes; when different settings appear and disappear, etc) in the `resources._js_config.JSConfig` class.

I think as a general approach:

1. Config settings and logic should live in `resources._js_config.JSConfig` not elsewhere

2. Just because a given config setting isn't needed or isn't relevant in a given context, doesn't mean it has to be missing from the config in that context. If the setting can be generated, and does no harm, then just let it be in the config. This keeps the code simpler as you don't need logic about toggling the setting into and out of the config

3. If a setting _can't_ be in the config in a given context (for example maybe the info needed to generate the setting's value isn't available in that context) then you need to add logic to `resources._js_config.JSConfig` to remove (or not insert) that setting (or block of settings) in that context

When a setting or block of settings is only relevant in a given context, the approach has been to add a high-level method to `JSConfig` called something like `add_foobar_settings()` and have the view(s) for the appropriate context(s) call that method.

The "content item selection" settings (for creating and editing assignments in Canvas) are a good example of settings that don't always appear in the config. `JSConfig` has a [`enable_content_item_selection_mode()`](https://github.com/hypothesis/lms/blob/9f3216664301659f87b3d201f444313e71086080/lms/resources/_js_config.py#L63-L106) method that contains code for generating the content item selection settings. Views for contexts where content item selection is relevant call this method, for example: https://github.com/hypothesis/lms/blob/9f3216664301659f87b3d201f444313e71086080/lms/views/content_item_selection.py#L57-L63

We did just do a several weeks technical debt project to collect config-generation logic that was scattered throughout the app, particularly in the views, into one place in the `JSConfig` class. So I think we'd appreciate _not_ adding config into the views again.

One additional difficulty that you're going to run into with what you're trying to do: the `JSConfig` instance is usually accessed as `context.JSConfig` but that property is [provided by `LTILaunchResource`](https://github.com/hypothesis/lms/blob/9f3216664301659f87b3d201f444313e71086080/lms/resources/lti_launch.py#L83-L86). In OAuth error dialog requests the `context` object will not be an `LTILaunchResource`. I think it'll be `None`. You'll have to add a context with a `js_config` property to [the relevant route](https://github.com/hypothesis/lms/blob/9f3216664301659f87b3d201f444313e71086080/lms/routes.py#L27). See how that's done for the routes that do use `LTILaunchResource`: https://github.com/hypothesis/lms/blob/9f3216664301659f87b3d201f444313e71086080/lms/routes.py#L14-L26. At first glance I think you probably want a new context class that uses `JSConfig` in the same way that `LTILaunchResource` does

jon-betts:
It might not be optimal, I think it's worth over-correcting in the direction we are going for a while (lump it all together). Then if we get to a point where it's too big, we can split it into some logical lumps if needs be.

The scattered stuff was a total nightmare when answering questions like "what gets sent to the front-end" or "what's our config structure". It was really difficult to make a sensible stab.

lms/templates/api/canvas/oauth2_redirect_error.html.jinja2
```diff
@@ -1,76 +1,19 @@
...
 {% endblock %}
 
-
 {% block styles %}
   {% for url in  asset_urls("frontend_apps_css") %}
     <link rel="stylesheet" href="{{ url }}">
   {% endfor %}
 {% endblock %}
+
+{% block scripts %}
```
LMS007, line 14: tiny nit here, since this now the third times this block is used, it might be best to make it a separate file and import it in the file_picker, basic_lti_launch and here.

lms/static/scripts/frontend_apps/index.js
```diff
@@ -6,21 +6,33 @@ import { createElement, render } from 'preact';
...
+}
+
 render(
   <Config.Provider value={config}>
     {config.mode === 'basic-lti-launch' && (
       <BasicLtiLaunchApp rpcServer={rpcServer} />
     )}
     {config.mode === 'content-item-selection' && <FilePickerApp />}
+    {config.mode === 'canvas-auth-error-dialog' && (
+      <CanvasAuthErrorDialog
```
LMS007, line 29: Question here: I'm a bit stumped when to rely on the provider or directly pass the props in via config. Won't these all be available via the provider already?

lms/static/scripts/frontend_apps/components/CanvasAuthErrorDialog.js
```diff
@@ -0,0 +1,76 @@
...
+
+import Button from './Button';
+import ErrorDisplay from './ErrorDisplay';
+import Dialog from './Dialog';
+
+/**
+ * Error dialog displayed when authorization of Canvas API access via OAuth
+ * fails.
+ */
+export default function CanvasAuthErrorDialog({
```
LMS007, line 12: This looks solid

lms/views/api/canvas/authorize.py
```diff
@@ -110,8 +124,10 @@ def oauth2_redirect_error(request):
         authorization_param = BearerTokenSchema(request).authorization_param(
             request.lti_user
         )
-        template_variables["authorize_url"] = request.route_url(
+        js_config["authorizeUrl"] = request.route_url(
```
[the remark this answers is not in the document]
robertknight: I just copied the existing logic over from the previous template here. I haven't actually checked which circumstances `request.lti_user` will be set.

robertknight:
Thanks for the feedback Sean and Jon. I'm going to attempt to rework the JS config along these lines.

lms/static/scripts/frontend_apps/index.js
robertknight: You're right - I hadn't thought too deeply about this when I put this draft together. I think it would be better to get it from the config where possible for consistency.

lms/resources/frontend_app.py
```diff
@@ -0,0 +1,20 @@
+import functools
+
+from lms.resources._js_config import JSConfig
+
+
+class FrontendAppResource:
```
robertknight, line 6: Perhaps this should be a base class for the `LTILaunchResource` resource?

lms/views/api/canvas/authorize.py
```diff
@@ -100,18 +101,13 @@ def oauth2_redirect(request):
...
+    # In an exception view the `context` is usually the exception that occurred.
+    # We're going to cheat here and create a different object instead to serve
+    # as the `context` in templates.
+    context = FrontendAppResource(request)
+    context.js_config.enable_oauth_error_mode(
+        error_details=request.params.get("error_description"),
+        is_scope_invalid=request.params.get("error") == "invalid_scope",
+        requested_scopes=FILES_SCOPES + SECTIONS_SCOPES,
+    )
+    return {"context": context}
```
robertknight, line 113: Re. the comment above, feel free to suggest alternative approaches to creating the "context" like this.

robertknight:
I have revised this per @seanh's suggestions above. Please see the updated description for an overview of the user-facing and internal changes. There were a couple of things that I wasn't sure about that I've called out in comments on the diff.

lms/resources/_js_config.py
```diff
@@ -16,17 +16,20 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
...
 
+        self._lti_user = request.lti_user
+        if self._lti_user:
+            self._h_user = self._lti_user.h_user
+            self._consumer_key = self._lti_user.oauth_consumer_key
+            self._ai_getter = request.find_service(name="ai_getter")
+        else:
+            self._h_user = None
+            self._consumer_key = None
+            self._ai_getter = None
```
seanh, line 31: I'd like to avoid if/else login in `__init__` methods.

What's the reason for this change?

I'm guessing that, for authorization error dialog requests, `request.find_service(name="ai_getter")` crashes? And `request.lti_user` would be `None` and `request.lti_user.h_user` would crash.

Could we instead:

* Leave `self._lti_user = request.lti_user` alone. I think this one is fine? `self._lti_user` will just be `None` for auth error dialog requests

* Move `request.find_service(name="ai_getter")` and `request.lti_user.h_user` and `self._lti_user.oauth_consumer_key` into properties:

  ```python
  @property
  def _ai_getter(self):
      return self._request.find_service(name="ai_getter")

  @property
  def _h_user(self):
      return self._lti_user.h_user

  ...
  ```

  You could add docstrings explaining that these properties should only be accessed during authenticated requests.

  Accessing the properties during an auth error dialog request would crash. But the auth error dialog view presumably does not access these properties.

  It's not necessary to implement caching for these properties. The efficiency saving would be negligible. pyramid_services already caches services.

lms/resources/_js_config.py
```diff
@@ -60,6 +63,46 @@ def asdict(self):
...
+        :type is_scope_invalid: bool
+        :param requested_scopes: List of Canvas API scopes that were requested
+        :type requested_scopes: list(str)
+        """
+        self._config.update(
+            {
+                "mode": "canvas-auth-error-dialog",
+                "invalidScope": is_scope_invalid,
+                "errorDetails": error_details,
+                "scopes": requested_scopes or [],
```
seanh, line 88:
```suggestion
                "oauth2_redirect_error": {
                    "invalidScope": is_scope_invalid,
                    "errorDetails": error_details,
                    "scopes": requested_scopes or [],
                },
```

Can we move these three settings into an `"oauth2_redirect_error"` sub-object? We've been trying to tidy up the config by grouping things into sub-objects

lms/resources/_js_config.py
```diff
@@ -60,6 +63,46 @@ def asdict(self):
         """
         return self._config
 
+    def enable_oauth_error_mode(
```
seanh, line 66:
```suggestion
    def enable_oauth2_redirect_error_mode(
```

Just naming this after the `oauth2_redirect_error()` view for consistency

lms/resources/_js_config.py
```diff
@@ -60,6 +63,46 @@ def asdict(self):
...
+        self, error_details, is_scope_invalid=False, requested_scopes=None
+    ):
+        """
+        Put the JavaScript code into "OAuth authorization failed" mode.
```
seanh, line 70:
```suggestion
        Put the JavaScript code into "OAuth 2 redirect error" mode.
```

lms/resources/_js_config.py
```diff
@@ -60,6 +63,46 @@ def asdict(self):
...
+        self._config["api"]["sync"] = self._sync_api()
+
+        # The config object for the Hypothesis client.
+        # Our JSON-RPC server passes this to the Hypothesis client over
+        # postMessage.
+        self._config["hypothesisClient"] = self._hypothesis_client
+
+        self._config["rpcServer"] = {
+            "allowedOrigins": self._request.registry.settings["rpc_allowed_origins"]
+        }
```
seanh, line 104: I'm scratching my head about this one. Are these lines of code that would crash for oauth redirect error requests? Or could these settings actually remain the defaults without harm?

lms/resources/_js_config.py
```diff
@@ -200,7 +243,7 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
                 "lis_outcome_service_url": lis_outcome_service_url,
                 "learner_canvas_user_id": self._request.params["custom_canvas_user_id"],
                 **kwargs,
-            },
+            }
```
seanh, line 246:
```suggestion
            },
```

lms/resources/_js_config.py
```diff
@@ -244,50 +291,40 @@ def _config(self):
...
+                "authToken": self._auth_token()
             },
             # The URL that the JavaScript code will open if it needs the user to
             # authorize us to request a new Canvas access token.
             "authUrl": self._request.route_url("canvas_api.authorize"),
             "canvas": {
                 # The URL that the JavaScript code will open if it needs the user to
                 # authorize us to request a new Canvas access token.
-                "authUrl": self._request.route_url("canvas_api.authorize"),
+                "authUrl": self._request.route_url("canvas_api.authorize")
```
seanh, line 308:
```suggestion
                "authUrl": self._request.route_url("canvas_api.authorize"),
```

lms/resources/_js_config.py
```diff
@@ -388,7 +425,7 @@ def _sync_api(self):
                 "lms": {
                     "tool_consumer_instance_guid": req.params[
                         "tool_consumer_instance_guid"
-                    ],
+                    ]
```
seanh, line 428:
```suggestion
                    ],
```

lms/resources/_js_config.py
```diff
@@ -404,7 +441,7 @@ def _sync_api(self):
 
         if "learner_canvas_user_id" in req.params:
             sync_api_config["data"]["learner"] = {
-                "canvas_user_id": req.params["learner_canvas_user_id"],
+                "canvas_user_id": req.params["learner_canvas_user_id"]
```
seanh, line 444:
```suggestion
                "canvas_user_id": req.params["learner_canvas_user_id"],
```

lms/resources/frontend_app.py
seanh: I don't think we should introduce a `FrontendAppResource`.

It's not actually true that this is the context resource for views that serve frontend apps (as the docstring below states). In fact, in most cases where a frontend app is served `LTILaunchResource` is the context resource. This `FrontendAppResource` is only used by the `oauth2_redirect_error` view.

I think we should just rename this to `OAuth2RedirectErrorResource`

lms/resources/frontend_app.py
```diff
@@ -0,0 +1,20 @@
...
+    resources.
+    """
+
+    def __init__(self, request):
+        self._request = request
+
+    @property
+    @functools.lru_cache()
+    def js_config(self):
+        return JSConfig(self, self._request)
```
seanh, line 20: What happens if you do this instead? _If_ this works it's simpler

```suggestion
    def __init__(self, request):
        self.js_config = JSConfig(self, request)
```

lms/views/api/canvas/authorize.py
seanh: You first need to configure `FrontendAppResource` to be used as the context resource for non-error views for this view's route. Something like:

```diff
diff --git a/lms/routes.py b/lms/routes.py
index 3ac63406..24014b3f 100644
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -25,7 +25,11 @@ def includeme(config):
         factory="lms.resources.LTILaunchResource",
     )
     config.add_route("canvas_oauth_callback", "/canvas_oauth_callback")
-    config.add_route("canvas_api.authorize", "/api/canvas/authorize")
+    config.add_route(
+        "canvas_api.authorize",
+        "/api/canvas/authorize",
+        factory="lms.resources.FrontendAppResource",
+    )
     config.add_route(
         "canvas_api.courses.files.list", "/api/canvas/courses/{course_id}/files"
     )
```

Now, even though the `context` argument that gets passed to exception views is the exception not the original context, the original context is still accessible as `request.context`:

```
index 131403c9..30310425 100644
@@ -101,13 +101,9 @@ def oauth2_redirect(request):
     renderer="lms:templates/api/canvas/oauth2_redirect_error.html.jinja2",
 )
 def oauth2_redirect_error(request):
-    # In an exception view the `context` is usually the exception that occurred.
-    # We're going to cheat here and create a different object instead to serve
-    # as the `context` in templates.
-    context = FrontendAppResource(request)
-    context.js_config.enable_oauth_error_mode(
+    request.context.js_config.enable_oauth_error_mode(
         error_details=request.params.get("error_description"),
         is_scope_invalid=request.params.get("error") == "invalid_scope",
         requested_scopes=FILES_SCOPES + SECTIONS_SCOPES,
     )
-    return {"context": context}
+    return {}
```

seanh, line 113: Since you're returning `return {"context": context}` to the template, I'm guessing some template somewhere does `context.js_config` but, if you don't `return {"context": context}`, then `context` is the exception not the `FrontendAppResource`?

If so, can the template just do `request.context.js_config` instead?

lms/resources/_js_config.py
robertknight: I formatted these files with `make format`. Does that not deal with this as expected?

robertknight: I think I may have had an older version of black installed globally and used by my editor, although that doesn't explain why `make format` didn't correct it if so.

robertknight:
I have updated this PR based on @seanh's feedback:

- The context resource is now named `CanvasOAuthRedirectResource` instead of `FrontendAppResource` and is used via the normal way of specifying a route context rather than the hack I had previously.  Two points in particular to check:
 - For the non-exception views we don't want/need the frontend app. Therefore `context.js_config` returns `None` if no exception has occurred
 - I copied the `CanvasOAuthRedirectResource.__acl__` attribute from part of `DefaultResource.__acl__`. Is this correct?
- The branches in `JSConfig` that would set various properties to `None` if `request.lti_user` was `None` has been replaced with `@property` methods
- The JSConfig values that are specific to the new route are grouped under a `canvasOAuth2RedirectError` key (camel-cased, since that is the convention for `js-config` keys)
-

seanh:
Seeing this:

Screenshot from 2020-07-01 13-59-07

seanh:
Launching assignments seems to be broken on this branch. @robertknight I think this needs to be tested in Canvas

robertknight:
>  @robertknight I think this needs to be tested in Canvas

With all due respect, how do you think I've been testing this?

My guess is that you are encountering an issue associated with the recent changes in Via 3. I've cherry-picked a recent commit from master from Jon which fixes an issue when PDF.js fetches PDFs through the latest version of Via 3.

robertknight:
> Seeing this:

Which course/assignment were you testing with? If you were looking at the "Wrong Developer Secret" course then that is a known issue explained in [this comment](https://github.com/hypothesis/lms/pull/1835#issuecomment-652279934).

seanh:
> > Seeing this:
> > 
> > Screenshot from 2020-07-01 13-59-07
>
> Which course/assignment were you testing with? If you were looking at the "Wrong Developer Secret" course then that is a known issue explained in [this comment](https://github.com/hypothesis/lms/pull/1835#issuecomment-652279934).

First delete all your auth tokens (to force the authorization dialog to appear):

```terminal
$ tox -qqe docker-compose -- exec postgres psql -U postgres -c "delete from oauth2_token;"
```

Then launch any assignment that requires Canvas API access (for example any assignment in the _Sections Enabled_ course, or any Canvas Files assignment in any course). In the authorization popup window click <kbd>Cancel</kbd> instead of <kbd>Authorise</kbd> and you'll get the **Error Details** with `null`

robertknight:
> Then launch any assignment that requires Canvas API access (for example any assignment in the Sections Enabled course, or any Canvas Files assignment in any course). In the authorization popup window click Cancel instead of Authorise and you'll get the Error Details with null

Thanks for the clarification. I did originally say that I wanted to address that separately. I've pushed [a fix and tests](https://github.com/hypothesis/lms/pull/1835/commits/3a5f9425648161586a12d1de28b7ebbdf77b6133) to this branch. This issue may have affected other dialogs, although I don't know of a way to reproduce.

robertknight:
@seanh - That was an intentional change that I called out in the PR description. For the moment the user is going to need to close the dialog and click “Authorize” in the window behind.

seanh:
Can we put the <kbd>Try again</kbd> button back? That seems like a UX regression to me.

I think the <kbd>Try again</kbd> button was just always shown in the popup window error dialog. You can reproduce it on master by clicking <kbd>Cancel</kbd> on Canvas's auth page for example, or you can hack the server-side code to send an invalid access token request to Canvas so the request will fail.

seanh:
If you click the _send us an email_ link on this branch it opens an email with _Error message: undefined_. On master the email body does contain an actual error message string that it gets from somewhere (not sure where).

robertknight:
> Can we put the Try again button back? That seems like a UX regression to me.

OK. I was trying to reduce the size of the initial PR but I'll look at adding it back in.

seanh:
I think it's worth doing the work to add it back in. Consider this workflow:

1. User opens authorization popup window and tries to authorize, gets "Authorization Failed" dialog without try again button.
2. User goes back to original tab (but does not close the popup window, maybe they minimized it, or it just went behind the main browser window, etc)
3. In the main tab the user clicks <kbd>Authorize</kbd> again which just re-raises the popup window without re-trying and with no button to retry.

It's true that the user can force it to retry by first clicking either <kbd>X</kbd> or <kbd>Cancel</kbd> in the popup window and then clicking <kbd>Authorize</kbd> again back in the main window but I'd say that's pretty non-obvious. I think it would be pretty easy for the user to not know how to re-try here.

Peek 2020-07-08 10-11

robertknight:
> I think it's worth doing the work to add it back in. Consider this workflow:

OK. I have re-implemented the "Try again" button in the client-rendered dialog. See https://github.com/hypothesis/lms/pull/1835/commits/8c1abfaa4bb5d70699027e67d0da3b8bbd25330b.

> If you click the send us an email link on this branch it opens an email with:

This was actually an existing issue with the support email link in all of the client-rendered dialogs. I have made a change to omit sections from the email body where there is no content. See https://github.com/hypothesis/lms/pull/1835/commits/3747a66d6691e20ef93302424285ccb38d8957ef.

With this change the email body is empty in most cases on this branch as it is on master, except that when a scope error occurs the "Technical details" section is populated, whereas on master it is not.

robertknight:
On both master and this branch the error dialog doesn't have any details if the developer secret is wrong. I'm not sure how much information we can get from Canvas in this case, but I filed https://github.com/hypothesis/lms/issues/1887 in case we can do something better.

seanh:
The dialog shouldn't have an <kbd>X</kbd> button in its top-right when in the popup window. It's not there on master, and it makes no sense (the OS window's own <kbd>X</kbd> button is right next to it and we also show a <kbd>Cancel</kbd> button as well):

close

closeagain

lms/resources/_js_config.py
```diff
@@ -16,16 +16,23 @@ class JSConfig:  # pylint:disable=too-many-instance-attributes
...
+    def _h_user(self):
+        return self._lti_user.h_user
+
+    @property
+    def _consumer_key(self):
+        return self._lti_user.oauth_consumer_key
+
+    @property
+    def _ai_getter(self):
+        # nb. `ai_getter` service is only available in LTI launches.
```
seanh, line 34:
```suggestion
```

This comment belongs in a docstring: it's information that the person calling the method wants to know. "Restrictions on when it can be called" are one of the things that belong in a multiline method docstring according to PEP 257.

Also, the property is available during all authenticated requests not just LTI launches.

The same restriction applies to the `_h_user()` and `_consumer_key()` properties above.

I think simply deleting the comment would also be acceptable since they're private methods (but any public methods that call these should note the restriction in their docstrings).

lms/resources/canvas_oauth2_redirect.py
```diff
@@ -0,0 +1,24 @@
...
+    @property
+    def js_config(self):
+        if not self._request.exception:
+            # The normal views do not need to configure the frontend, only
+            # the exception views.
+            return None
+
+        if not self._js_config:
+            self._js_config = JSConfig(self, self._request)
+        return self._js_config
```
seanh, line 24:
```suggestion
    def __init__(self, request):
        self.js_config = JSConfig(self, request)
```

I think this works although some tests will need to be updated. Let's keep the code simple if we can

robertknight: The problem with the suggested approach is that it will cause the `base.html` template to load the client JS during the non-exception view, which we don't want. See https://github.com/hypothesis/lms/blob/3747a66d6691e20ef93302424285ccb38d8957ef/lms/templates/base.html.jinja2#L39

lms/resources/_js_config.py
robertknight: > Also, the property is available during all authenticated requests not just LTI launches.

I'm wary of writing that because there may be requests in future where there the caller is "authenticated" in some way but `request.lti_user` is not set. For now I've just removed this comment since, as you noted, it is not entirely correct.

robertknight:
I have removed the "X" button from the dialog and replaced "Cancel" with "Close", since that seems more appropriate to me in this context:

OAuth error dialog close only

lms/resources/canvas_oauth2_redirect.py
robertknight: In the context of the `oauth2_redirect_error` view this happens to work from the user's point of view because the `oauth2_redirect.html` template calls `window.close()` as soon as the page loads.

The client frontend scripts load synchronously though so the `window.close()` block will not currently run until after the frontend JS has been fetched and run. Again, these are small so the user is unlikely to notice, but it still adds a very small unnecessary delay.

robertknight: Actually, nevermind. I see that `oauth2_redirect.html` overrides the default `scripts` block so the above comment is wrong.

robertknight:
I realized I forgot to fix up a test after I changed `js_config` from an `@property` which is present on auto-mocked objects to an attribute assigned in `__init__` which is not. I'll fix that up and get this merged and deployed.

PR #1844, opened 2020-06-19
Fix flicker issue when successive errors occur

LMS007:
- Error state is not reset until after all requests have succeeded.
- The full page spinner no longer is rendered while an error is present. As a follow up, a smaller in-element spinner could be added to the dialog component
- Add missing test for auth window

----------

This is a follow up from this PR
https://github.com/hypothesis/lms/pull/1833

In short, I simply removed the spinner changes and just addressed the flickering.

Follow ups to this could be:
- how to show loading while the error is present?
- do we need to delay or throttle the onset of the spinner for very fast server errors?

fixes https://github.com/hypothesis/lms/issues/1826

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -84,7 +84,6 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
 
   const incFetchCount = () => {
     setFetchCount(count => count + 1);
-    setErrorState(null);
   };
 
   const decFetchCount = () => {
@@ -111,18 +110,20 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
    * Fetch the groups from the sync endpoint if `sync` object exists.
    */
   const fetchGroups = useCallback(async () => {
-    if (apiSync) {
-      try {
-        setErrorState(null);
-        const groups = await apiCall({
-          authToken,
-          path: apiSync.path,
-          data: apiSync.data,
-        });
-        clientRpc.setGroups(groups);
-      } catch (e) {
-        handleError(e, 'error-fetch');
-      }
+    if (!apiSync) {
+      return true;
+    }
+    try {
+      const groups = await apiCall({
+        authToken,
+        path: apiSync.path,
+        data: apiSync.data,
+      });
+      clientRpc.setGroups(groups);
+      return true;
+    } catch (e) {
+      handleError(e, 'error-fetch');
+      return false;
     }
   }, [apiSync, authToken, clientRpc]);
 
@@ -137,20 +138,23 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
       // If no "callback" URL was supplied for the frontend to use to fetch
       // the URL, then the backend must have provided the Via URL in the
       // initial request, which we'll just use directly.
-      return;
+      return true;
     }
+    let result;
+    incFetchCount();
     try {
-      incFetchCount();
       const { via_url: contentUrl } = await apiCall({
-        authToken,
+        authToken: authToken,
         path: viaCallbackUrl,
       });
       setContentUrl(contentUrl);
+      result = true;
     } catch (e) {
       handleError(e, 'error-fetch');
-    } finally {
-      decFetchCount();
+      result = false;
     }
+    decFetchCount();
+    return result;
   }, [authToken, viaCallbackUrl]);
 
   /**
@@ -211,7 +215,13 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
 
     try {
       await authWindow.current.authorize();
-      await Promise.all([fetchContentUrl(), fetchGroups()]);
+      const [fetchedContent, fetchedGroups] = await Promise.all([
+        fetchContentUrl(),
+        fetchGroups(),
+      ]);
+      if (fetchedContent && fetchedGroups) {
+        setErrorState(null);
+      }
     } finally {
       // @ts-ignore - The `current` field is incorrectly marked as not-nullable.
       authWindow.current = null;
@@ -271,6 +281,7 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
           role="alertdialog"
           buttons={[
             <Button
+              disabled={fetchCount > 0}
               buttonRef={focusedDialogButton}
               onClick={authorizeAndFetchUrl}
               className="BasicLtiLaunchApp__button"
@@ -290,6 +301,7 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
           role="alertdialog"
           buttons={[
             <Button
+              disabled={fetchCount > 0}
               buttonRef={focusedDialogButton}
               onClick={authorizeAndFetchUrl}
               className="BasicLtiLaunchApp__button"
--- a/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
@@ -187,7 +187,6 @@ describe('BasicLtiLaunchApp', () => {
         'Button[label="Authorize"]'
       );
       assert.isTrue(authButton.exists());
-      await contentHidden(wrapper);
 
       // Click the "Authorize" button and verify that authorization is attempted.
       fakeApiCall.reset();
@@ -202,6 +201,27 @@ describe('BasicLtiLaunchApp', () => {
         wrapper.find('iframe').prop('src'),
         'https://via.hypothes.is/123'
       );
+    });
+
+    it('does not create a second auth window when Authorize button is clicked twice', async () => {
+      // Make the initial URL fetch request reject with an unspecified `ApiError`.
+      fakeApiCall.rejects(new ApiError(400, {}));
+
+      const wrapper = renderLtiLaunchApp();
+      const authButton = await waitForElement(
+        wrapper,
+        'Button[label="Authorize"]'
+      );
+
+      // Click the "Authorize" button
+      fakeApiCall.reset();
+      fakeApiCall.resolves({ via_url: 'https://via.hypothes.is/123' });
+      authButton.prop('onClick')();
+      assert.calledOnce(FakeAuthWindow);
+
+      // Click the "Authorize" button again
+      authButton.prop('onClick')();
+      assert.calledOnce(FakeAuthWindow);
     });
 
     [
@@ -340,21 +360,7 @@ describe('BasicLtiLaunchApp', () => {
     let groupsCallResolve;
     let groupsCallReject;
 
-    beforeEach(() => {
-      // Will attempt to fetch the 1. content url and 2. groups.
-      fakeConfig.api = {
-        authToken: 'dummyAuthToken',
-        viaCallbackUrl: 'https://lms.hypothes.is/api/files/1234',
-        sync: {
-          data: {
-            course: {
-              context_id: '12345',
-              custom_canvas_course_id: '101',
-            },
-          },
-          path: '/api/sync',
-        },
-      };
+    const resetApiCalls = () => {
       fakeApiCall
         .withArgs({
           authToken: 'dummyAuthToken',
@@ -384,6 +390,26 @@ describe('BasicLtiLaunchApp', () => {
             groupsCallReject = reject;
           })
         );
+    };
+
+    beforeEach(() => {
+      // When BasicLtiLaunchApp is rendered, it will attempt to fetch:
+      //  1. content url
+      //  2. groups
+      fakeConfig.api = {
+        authToken: 'dummyAuthToken',
+        viaCallbackUrl: 'https://lms.hypothes.is/api/files/1234',
+        sync: {
+          data: {
+            course: {
+              context_id: '12345',
+              custom_canvas_course_id: '101',
+            },
+          },
+          path: '/api/sync',
+        },
+      };
+      resetApiCalls();
     });
 
     it('renders the spinner until contentUrl requests finish', async () => {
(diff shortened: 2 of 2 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -91,60 +90,68 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
...
       setError(e);
       setErrorState(newErrorState);
     }
-  };
+  }, []);
 
   /**
    * Fetch the groups from the sync endpoint if `sync` object exists.
+   *
+   * @return {boolean} - true if an error occurred, false otherwise
```
robertknight, line 105: Logically it seems confusing to me to have a function called `fetchGroups` which returns `true` if it _failed_ rather than succeeded. The same applies to `fetchContentUrl` below.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -202,7 +209,33 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
 
     try {
       await authWindow.current.authorize();
-      await Promise.all([fetchContentUrl(), fetchGroups()]);
+      // In the re-try case, block only the contentUrl request. At
```
robertknight, line 212: I think this block could be implemented in a simpler way. Supposing the `fetchContentUrl` and `fetchGroups` functions were changed to return `true` on success as suggested above:

```js
const [fetchedContent, fetchedGroups] = Promise.all([fetchContentUrl(), fetchGroups()]);
if (fetchedContent && fetchedGroups) {
  setErrorState(null);
}
```

An alternative approach would be for `fetchContentUrl` and `fetchGroups` to re-throw any error that occurs. Then in this block the code would just be:

```js
try {
  await Promise.all([fetchContentUrl(), fetchGroups()]);
  setErrorState(null);
} catch (e) {
 // Ignore error here.
} finally {
  authWindow.current = null;
}
```

If you went for this second route, you'd also have to ignore any exceptions when `fetchContentUrl` and `fetchGroups` are called after the component is initially rendered.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,57 @@ describe('BasicLtiLaunchApp', () => {
...
+          'Button[label="Authorize"]'
+        );
+
+        // Click the "Authorize" button.
+        resetApiCalls();
+        // contentUrlReject rejects again
+        contentUrlReject(new ApiError(400, {}));
+        groupsCallReject(new ApiError(400, {}));
+        // Error prompt should remain.
+        authButton.prop('onClick')();
```
robertknight, line 477: Is this line the "Click" that the "Click the "Authorize" button" comment above refers to?

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,57 @@ describe('BasicLtiLaunchApp', () => {
...
+        await waitForElement(
+          wrapper,
+          'FakeDialog[title="Authorize Hypothesis"]'
```
robertknight, line 480: This test passes if run against the version of `BasicLtiLaunchApp.js` from the `master` branch. I guess that is happening because the "Authorize" dialog is still visible immediately after the "Authorize" button is clicked.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,57 @@ describe('BasicLtiLaunchApp', () => {
...
+        // groups still fails, but contentUrl does not.
+        groupsCallReject(new ApiError(400, {}));
+        contentUrlResolve({
+          via_url: 'https://via.hypothes.is/123',
+        });
+        // Error prompt should remain.
+        authButton.prop('onClick')();
+        await waitForElement(
+          wrapper,
+          'FakeDialog[title="Authorize Hypothesis"]'
```
robertknight, line 505: This test passes if run against the version of `BasicLtiLaunchApp.js` from the `master` branch. I guess that is happening because the "Authorize" dialog is still visible immediately after the "Authorize" button is clicked.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -91,60 +90,68 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
    *  "error-authorizing", or "error-report-submission"
    * @param {boolean} [retry=true] - Can the request be retried?
    */
-  const handleError = (e, newErrorState, retry = true) => {
+  const handleError = useCallback((e, newErrorState, retry = true) => {
```
robertknight, line 93: I don't think the change to `handleError` is strictly necessary here, although it doesn't do any harm to use `useCallback`. When I tested on the `master` branch, the linting seemed smart enough to realize that it doesn't matter if the `handleError` callback was not a dependency because it doesn't read any state.

LMS007:
> Hi Kyle,
> 
> Apologies for taking so long to get through this. I think the basic approach of not resetting the error state until after both the content URL and groups are fetched seems reasonable, but I had a few thoughts:
> 
> 1. The tests on this branch pass if run against the version of `BasicLtiLaunchApp.js` on the master branch, so they don't actually verify the behavior change here.
> 2. I found it confusing that `fetchContentUrl` and `fetchGroups` return `true` if they _fail_.
> 3. From a UX perspective, there is a hazard that the user can click buttons in the error dialog a second time while requests are being made. I think you had a plan to address this separately, but a simple intermediate option might be to disable the buttons with `<Button disabled={fetchCount > 0} ...>`.

I'll take another look first thing tomorrow, I need to refresh a bit on this branch. 
With respect to #2. I originally used a numeric value as the return value, then changed to a boolean. The idea was the pull from the old "returns a non-zero result when it fails" standard. e.g. a 0 would be passing, 1 is failing. That's where true/false came from but I now see that appears odd without the context.

LMS007:
For now, I just fixed the new 2 new tests to ensure they fail on master and pass on this branch. I still don't have a solid understanding how to simplify the promises so I'm going to leave this for next week and give it some more thought or may be just punt the issue. But at least there is a set of tests now that will detect the transient disappearance of the dialog and fail.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: yes, fixed

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
LMS007: @robertknight , just wanted to get some clarification on this block once more.

Just glancing at your suggestions here.

**Suggestion A.**
This solution appears to not block on either request. Is that correct? Was there supposed to be an `await` before Promise.all ? If they don't block, then both will return right away so I'm a bit confused by that.

**Suggestion B.**
This solution appears to block on both requests, but this is different than what I have done.  In my solution, I specifically did not block on the groups request --only on the contentUrl. This means if the groups  request take a long time, and the contentUrl does not, then we arrive at the promise `then` block for the contentUrl's promise before the groups are done. Presumable, the return code for groups (`groupError`) would still be false and we would happily set error to null.

After that point, two outcomes may occur a few moments later on. 1. The groups finishes w/o an error, no UI changes occur. 2. The groups call errors,  sets an error state, and the popup jumps right back on screen. 

The other path this code handles is if the groups call actaul errors before the contentUrl. In that case, we use the variable `groupError` and check that value in the contentUrl's `then` block. In that case we know not to clear the error state.

I'm a bit confused how the examples you proposed handle that. But maybe I'm way overthinking this code block or just not understanding something that could be a whole lot more simple and I apologize for that.

robertknight: > Was there supposed to be an await before Promise.all ?

Yes, indeed there was. It should have been `await Promise.all([...])`.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -384,6 +389,23 @@ describe('BasicLtiLaunchApp', () => {
             groupsCallReject = reject;
           })
         );
+    };
+    beforeEach(() => {
```
robertknight, line 393: Nit: Can you add a new line before the `beforeEach` block here.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,71 @@ describe('BasicLtiLaunchApp', () => {
...
     });
+
+    context('when auth fails multiple times in a row', () => {
+      // Helper to ensure the dialog is initially present but also remains so
+      // for a short duration longer without disappearing.
+      async function dialogShouldRemain(wrapper) {
+        // Add an artificial delay to ensure that all the api calls resolve or
+        // reject before proceeding.
+        await true;
+        await true;
```
robertknight, line 466: This is not really safe as it assumes that API calls will take a certain number of microtasks before being resolved. In my experience this can make tests brittle and is liable to cause unexpected failures with apparently innocent refactoring.

Having a think about the original issue, the problem was that the auth dialog would be hidden and the iframe shown briefly when the LMS app was in the following state:

- User has clicked the "Authorize" button
- API calls to the backend to fetch content URL or fetch groups have not yet completed

In this branch, the dialog remains visible until both those calls complete. Therefore what I would suggest to do in the tests is instead of making the content URL / groups API calls reject, instead make them not resolve immediately and then check whether the dialog is visible while the API calls are "in flight". Then resolve the API calls after this assert passes.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,71 @@ describe('BasicLtiLaunchApp', () => {
...
+        // Error prompt should remain.
+        assert.isTrue(wrapper.find('FakeDialog').exists());
+        // Error prompt should also not go away after a short while (e.g.flicker)
+        try {
+          await waitFor(() => {
+            wrapper.update();
+            const el = wrapper.find('FakeDialog');
```
robertknight, line 473: If you want to test whether a selector has any matches `wrapper.exists('Selector')` is an easier shorthand.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +455,71 @@ describe('BasicLtiLaunchApp', () => {
...
+            if (el.length === 0) {
+              throw new Error();
```
robertknight, line 475: I would recommend to always set the `message` parameter when creating an `Error` to make debugging easier.

Edit: Oh I see, you're using this to test what error was reported in the "catch" clause.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: sounds reasonable, I'll give it a go!

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: I'll keep that in mind from now on :)

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -36,10 +36,7 @@ describe('BasicLtiLaunchApp', () => {
   }
 
   function spinnerHidden(wrapper) {
-    return waitFor(() => {
-      wrapper.update();
-      return !wrapper.exists('Spinner');
-    });
+    return !wrapper.exists('Spinner');
```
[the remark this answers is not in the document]
LMS007: I'm sure what happened here, I'm going to revert because clearly this looks wrong.

robertknight:
Rebased

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -131,26 +133,31 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
    * exists.
    *
    * This will typically be a PDF URL proxied through Via.
+   *
+   * @return {boolean} - false if an error occurred, true otherwise
```
robertknight, line 137: This `@return` has no effect because it is not attached to a named function or variable that is directly assigned to a closure. I would just remove this and let TS infer it, or use `@type {() => Promise<boolean>}` if you wanted to provide information for the reader.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -111,17 +110,20 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
    * Fetch the groups from the sync endpoint if `sync` object exists.
    */
   const fetchGroups = useCallback(async () => {
-    if (apiSync) {
+    if (!apiSync) {
+      return true;
+    } else {
```
robertknight, line 115: You can avoid the `else` here since the `if` branch returns. This avoids an extra level of indentation. ie.

```js
if (!apiSync) {
  return true;
}

try {
  ...
}
```

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -204,6 +204,28 @@ describe('BasicLtiLaunchApp', () => {
...
+    it('does not create a second auth window when clicked twice', async () => {
+      // Make the initial URL fetch request reject with an unspecified `ApiError`.
+      fakeApiCall.rejects(new ApiError(400, {}));
+
+      const wrapper = renderLtiLaunchApp();
+      const authButton = await waitForElement(
+        wrapper,
+        'Button[label="Authorize"]'
+      );
+      await contentHidden(wrapper);
```
robertknight, line 216: Is this needed? It seems that the rest of the test only needs to verify that the `authButton` element is shown.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -384,6 +392,24 @@ describe('BasicLtiLaunchApp', () => {
             groupsCallReject = reject;
           })
         );
+    };
+
+    beforeEach(() => {
+      // Will attempt to fetch the 1. content url and 2. groups.
```
robertknight, line 398: It's not immediately clear what "Will attempt to fetch" refers to here. I presume it should say something like "When `BasicLtiLaunchApp` is rendered, it will attempt..."?

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +459,67 @@ describe('BasicLtiLaunchApp', () => {
...
+            }
+            return null;
+          }, 10);
+        } catch (e) {
+          if (!e.message) {
+            throw new Error('The dialog disappeared');
+          }
+        }
+      }
+      it('shows an error dialog if the first requests rejects and second also rejects', async () => {
```
robertknight, line 484: Can you add a new line above this `it` block.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +459,67 @@ describe('BasicLtiLaunchApp', () => {
...
       await contentHidden(wrapper);
     });
+
+    context('when auth fails multiple times in a row', () => {
+      // Helper to ensure the dialog is initially present but also remains so
+      // for a short duration longer without disappearing.
+      async function dialogShouldRemain(wrapper) {
+        // Error prompt should be present.
+        assert.isTrue(wrapper.find('FakeDialog').exists());
+        // Error prompt should also not go away after a short while (e.g.flicker)
```
robertknight, line 469: I suggest to replace `(e.g. flicker)` with `(see issue #1826)` which describes the issue that this test intends to guard against in more detail.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -433,6 +459,67 @@ describe('BasicLtiLaunchApp', () => {
...
+        await dialogShouldRemain(wrapper);
+        // contentUrlReject rejects again
+        contentUrlReject(new ApiError(400, {}));
+        groupsCallReject(new ApiError(400, {}));
+        // Click the "Authorize" button.
+        authButton.prop('onClick')();
+        // Dialog prompt should remain.
+        await dialogShouldRemain(wrapper);
+      });
+      it('shows an error dialog if contentUrl succeeds but groups rejects first', async () => {
```
robertknight, line 504: Can you add a new line above this `it` block

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -204,6 +204,28 @@ describe('BasicLtiLaunchApp', () => {
       );
     });
 
+    it('does not create a second auth window when clicked twice', async () => {
```
robertknight, line 207: It isn't immediately obvious what "when clicked twice" refers to. Perhaps `when "Authorize" button is clicked twice`?

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
robertknight, line 484:
```suggestion
      it('shows an error dialog if the initial content/groups requests reject and second attempt also rejects', async () => {
```

I suggested to reword this slightly to clarify what "requests reject" refers to.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: Its redundant along with the the previous test. Can be removed in both places.

Issue #1751, opened 2020-05-14
Make sure that a newly initialized DB is stamped with current schema version


robertknight:
When a developer sets up a new lms install locally, the current schema version is not stamped into the database. This means that when the next schema migration happens, attempting to run `alembic upgrade head` will try to run migrations that have already been applied. This was resolved in h in https://github.com/hypothesis/h/pull/5388. Perhaps we can do something similar in the lms app?

PR #1853, opened 2020-06-23
Automate `alembic stamp`

seanh:
Automatically stamp the DB with the latest version on startup if it
isn't stamped already.

Fixes https://github.com/hypothesis/lms/issues/1751

Similar to what we already do in h:
https://github.com/hypothesis/h/blob/928fd87a4b2132716850733c99d8c016fda5356e/h/cli/commands/init.py#L33-L46

To test this:

1. Delete your DB:

       $ make services args='down'
       $ make services

2. Run `make dev` to recreate your DB:

       $ make dev

3. Query your current DB version:

       $ tox -qqe dev --run-command 'alembic -c conf/alembic.ini current'

4. Try to upgrade your DB to the latest version:

       $ tox -qqe dev --run-command 'alembic -c conf/alembic.ini upgrade head'

On master (3) would print out nothing because the DB isn't stamped and
(4) would try to run all the DB migrations from the first one and crash.

On this branch (3) will print out the latest version and (4) will run no
migrations since it knows you're already on the latest version.

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -1,5 +1,7 @@
 import logging
 
+import alembic.command
+import alembic.config
 import sqlalchemy
 import zope.sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
@@ -102,6 +104,20 @@ def init(engine, drop=False):
     BASE.metadata.create_all(engine)
 
 
+def _stamp_db(engine):  # pragma: nocover
+    """
+    Stamp the database with the latest revision if it isn't already stamped.
+
+    This is convenient in development environments to automatically stamp a new
+    database after initializing it.
+
+    """
+    try:
+        engine.execute("select 1 from alembic_version")
+    except sqlalchemy.exc.ProgrammingError:
+        alembic.command.stamp(alembic.config.Config("conf/alembic.ini"), "head")
+
+
 def make_engine(settings):
     """Construct a sqlalchemy engine from the passed ``settings``."""
     return sqlalchemy.create_engine(settings["sqlalchemy.url"])
@@ -145,6 +161,9 @@ def includeme(config):
     engine = make_engine(config.registry.settings)
     config.registry["sqlalchemy.engine"] = engine
     init(engine)
+
+    if config.registry.settings["dev"]:  # pragma: nocover
+        _stamp_db(engine)
 
     # Add a property to all requests for easy access to the session. This means
     # that view functions need only refer to ``request.db`` in order to


.coveragerc
```diff
@@ -6,6 +6,7 @@ source =
     tests/unit
 omit =
     lms/scripts/*
+    lms/migrations/*
```
seanh, line 9: Something about the code below causes coverage to notice `lms/migrations/`, which it didn't notice before, and fail because the migration scripts and co don't have test coverate

lms/db/__init__.py
```diff
@@ -1,5 +1,7 @@
 import logging
 
+import alembic.command
+import alembic.config
```
seanh, line 4: `import alembic` doesn't work, you have to do it like this

lms/db/__init__.py
```diff
@@ -101,6 +103,11 @@ def init(engine, drop=False):
         BASE.metadata.drop_all(engine)
     BASE.metadata.create_all(engine)
 
+    try:
+        engine.execute("select 1 from alembic_version")
+    except sqlalchemy.exc.ProgrammingError:
+        alembic.command.stamp(alembic.config.Config("conf/alembic.ini"), "head")
```
seanh, line 109: As far as I can tell the `alembic` Python module doesn't provide any easy way to find out whether the DB is already stamped. `engine.execute("select 1 from alembic_version")` is much simpler and has been working for us in h

lms/db/__init__.py
```diff
@@ -146,6 +155,9 @@ def includeme(config):
     config.registry["sqlalchemy.engine"] = engine
     init(engine)
 
+    if config.registry.settings["dev"]:  # pragma: nocover
+        stamp_db(engine)
```
seanh, line 159: The `alembic` command added above caused problems with database connections on Jenkins. I couldn't figure out why, and debugging it / trying things out was very slow. So I just put the stamping in a separate function and only called it in dev. We don't want to do this in tests, CI or QA and prod anyway, so this is probably a good thing. I don't think it's worth adding unit tests for something that's only used in dev / I think the unit test you'd likely write for this would probably be a bit silly

lms/db/__init__.py
```diff
@@ -102,6 +104,13 @@ def init(engine, drop=False):
     BASE.metadata.create_all(engine)
 
 
+def stamp_db(engine):  # pragma: nocover
```
seanh, line 107: I don't think it's worth adding unit tests for something that's only used in dev / I think the unit test you'd likely write for this would probably be a bit silly

jon-betts:
I can confirm this breaks as described in master and works as described in this branch, now for a code review...

lms/db/__init__.py
jon-betts: Is this because we're trying to reach one extra down the hierarchy (alembic > command > stamp) or does this just not work full stop?

lms/db/__init__.py
jon-betts: I'm pretty sure we'll test it practically enough by running it a million times a day.

Add a docstring saying what this does and that it's only to be used in dev? Also is this private? Should it be '_stamp_db'?

lms/db/__init__.py
jon-betts: The relative path is a bit annoying here. You can fix it with `pkg_resources.resource_filename` but it doesn't really fix the issue as the `script_location` in the config is relative to root as well.

To get this to work correctly, I think you'd need a layout like:

 * `lms/lms/alembic/`
   * `conf.ini`
   * `migrations/`

Then the relative path in the conf would work ok. In this case I'd say it's more effort than it's worth, as we don't do this all the time. But it will explode with different working directories.

But in other places this can make running tests in particular very annoying and should probably be a thing we try to avoid.

lms/db/__init__.py
seanh: I dunno:

```python
>>> import alembic
>>> alembic.command
AttributeError
```

It's always done as `import alembic.command`

:man_shrugging: Not worth worrying about

lms/db/__init__.py
seanh: You can't easily run `make dev` (which is what will run this code) from any other directory anyway and if you do go out of your way to do so, lots of existing stuff in `make dev` already fails :man_shrugging: I don't think it's a concern, no one ever does it

lms/db/__init__.py
seanh: Added the `_` and a docstring

PR #1854, opened 2020-06-23
Add a convenient `make db` command

### LINKED ISSUE #1708: Add `make db` command
Often enough we end to edit rows in SQL tables that otherwise cause breaking changes with local dbs for developers. Adding a `make db` command that can run a series of updates would be very useful.

We could also have a local file with a number that is generated only by `make db` and a versioned file that is the same thing. If they ever get out of sync, then `make dev` could give a warning or auto run `make db` which may provide more useful output than whatever broken thing may happened as a result of db out of sync.

From Sean
-----------

 People need to upgrade their LMS dev DBs after we merged a DB migration yesterday. The command to do this without resetting your DB is tox -qe dev --run-command 'alembic -c conf/alembic.ini upgrade head' <- @developers
I'd like to make this easier:
1. I think we should add a make db so you don't have to remember that long tox command. This will also paper over that the DB upgrade command is different in h and lms
2. I'd like to merge make devdata into make db. So make db will both upgrade your DB schema (if it's out of date) and (re-)load devdata.
This'll mean that if anything seems to be wrong with your DB, whether crashes like the above or the apps not working correctly due to missing dev data, just running make db should fix it, and will be the same command across apps.
It's also a very simple incremental improvement on the current set up, that should be quick and easy to add to all our apps, and a very small change to developers existing practices: you just need to start using the new make db command when you have a problem.

seanh:
Depends on #1853 in order to work properly.

Fixes #1708

# 1. Add a script for initializing the DB

To test it:

1. Delete your DB:

       $ make services args='down'
       $ make services

2. Verify that your DB hasn't been initialized:

       $ docker exec -it lms_postgres_1 psql -U postgres -c "\d"
       Did not find any relations.

3. Run the script:

       $ tox -qqe dev --run-command 'initdb conf/development.ini'

4. Verify that now your DB has been initialized:

       $ docker exec -it lms_postgres_1 psql -U postgres -c "\d"
                               List of relations
        Schema |               Name                |   Type   |  Owner
       --------+-----------------------------------+----------+----------
        public | application_instances             | table    | postgres
        public | application_instances_id_seq      | sequence | postgres
        public | course                            | table    | postgres
        ...

# 2. Add a convenient `make db` command

This is just so that:

1. You don't have to remember
   `tox -qe dev --run-command 'alembic -c conf/alembic.ini upgrade head'`,
   it's now just `make db`

2. You also don't have to remember that the long tox command for
   upgrading your DB is different in h than in lms. I'll send a similar
   PR to h that also adds `make db` to h

To test this:

1. Delete your DB:

       $ make services args='down'
       $ make services

2. Run `make db`, this will both initialize your DB and upgrade it to the latest version:

       $ make db

3. Run `make db` again. Since your DB it already initialized and at the latest version this won't do anything, but it's safe to run:

       $ make db

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,15 +36,15 @@ services: python
 db: args?=upgrade head
 db: python
 	@tox -qqe dev --run-command 'initdb conf/development.ini'
-	@tox -qe dev --run-command 'alembic -c conf/alembic.ini $(args)'
+	@tox -qe dev  --run-command 'alembic -c conf/alembic.ini $(args)'
 
 .PHONY: dev
 dev: build/manifest.json python
-	@tox -qe dev -- honcho start ${processes}
+	@tox -qe dev --run-command 'honcho start ${processes}'
 
 .PHONY: devdata
 devdata: python
-	@tox -qe dev -- devdata conf/development.ini
+	@tox -qe dev --run-command 'devdata conf/development.ini'
 
 .PHONY: web
 web: python
@@ -58,7 +58,7 @@ assets:
 
 .PHONY: shell
 shell: python
-	@tox -qe dev -- pshell conf/development.ini
+	@tox -qe dev --run-command 'pshell conf/development.ini'
 
 .PHONY: sql
 sql: python
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ services: python
 .PHONY: db
 db: args?=upgrade head
 db: python
-	@tox -qqe dev --run-command 'initdb conf/development.ini'
+	@tox -qqe dev --run-command 'initdb conf/development.ini'  # See setup.py for what initdb is.
 	@tox -qe dev  --run-command 'alembic -c conf/alembic.ini $(args)'
 
 .PHONY: dev
@@ -44,7 +44,7 @@ dev: build/manifest.json python
 
 .PHONY: devdata
 devdata: python
-	@tox -qe dev --run-command 'devdata conf/development.ini'
+	@tox -qe dev --run-command 'devdata conf/development.ini'  # See setup.py for what devdata is.
 
 .PHONY: web
 web: python
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ help:
 	@echo "make help              Show this help message"
 	@echo 'make services          Run the services that `make dev` requires'
 	@echo "                       (Postgres) in Docker"
+	@echo 'make db                Upgrade the DB schema to the latest version'
 	@echo "make dev               Run the entire app (web server and other processes)"
 	@echo "make web               Run the web server on its own (useful for debugging the "
 	@echo "                       Python code with pdb)"
@@ -30,6 +31,12 @@ help:
 services: args?=up -d
 services: python
 	@tox -qe docker-compose -- $(args)
+
+.PHONY: db
+db: args?=upgrade head
+db: python
+	@tox -qqe dev --run-command 'initdb conf/development.ini'
+	@tox -qe dev --run-command 'alembic -c conf/alembic.ini $(args)'
 
 .PHONY: dev
 dev: build/manifest.json python
--- a/lms/scripts/__init__.py
+++ b/lms/scripts/__init__.py
@@ -1,3 +1,2 @@
 from lms.scripts.devdata import devdata
-
-__all__ = ["devdata"]
+from lms.scripts.initdb import initdb
--- a/lms/scripts/initdb.py
+++ b/lms/scripts/initdb.py
@@ -1,0 +1,16 @@
+"""
+A CLI command to initialize the database if it isn't already initialized.
+
+Usage:
+
+    initdb conf/development.ini
+"""
+import sys
+
+from pyramid.paster import bootstrap
+
+
+def initdb():
+    # Initialise the pyramid environment, which is enough to trigger the
+    # initialisation code in `lms.db.__init__.py` to setup the DB for us.
+    bootstrap(sys.argv[1])
--- a/setup.py
+++ b/setup.py
@@ -24,5 +24,6 @@ setup(
     entry_points="""\
     [console_scripts]
     devdata = lms.scripts:devdata
+    initdb = lms.scripts:initdb
     """,
 )


Makefile
```diff
@@ -31,6 +32,12 @@ services: args?=up -d
 services: python
 	@tox -qe docker-compose -- $(args)
 
+.PHONY: db
+db: args?=upgrade head
+db: python
+	@tox -qqe dev --run-command 'initdb conf/development.ini'
+	@tox -qe dev --run-command 'alembic -c conf/alembic.ini $(args)'
```
seanh, line 39: The first command has `-qq` so that tox doesn't print out "commands succeeded congratulations :)" after the first command, because then it would print it out twice. The second command has just one `-q`, the same as every other `tox` command in our `Makefile`'s, so tox _does_ print out its summary of the second command. If the first command fails tox will fail without running the second command and will print out the error from the first command

lms/scripts/initdb.py
```diff
@@ -0,0 +1,14 @@
...
+
+    initdb conf/development.ini
+"""
+import sys
+
+from pyramid.paster import bootstrap
+
+
+def initdb():
+    bootstrap(sys.argv[1])
```
seanh, line 14: Initializing the DB (if it isn't already) is already done as part of booting up the app. For example `make dev` would do it. So all we need for a standalone script is to call `bootstrap` and then exit

seanh: P.S. The way we've added a command line script to the app here is the same as how lms's `devdata` script is already done, and is just the standard way of doing command line scripts in Pyramid apps

Makefile
```diff
@@ -31,6 +32,12 @@ services: args?=up -d
 services: python
 	@tox -qe docker-compose -- $(args)
 
+.PHONY: db
+db: args?=upgrade head
```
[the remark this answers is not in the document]
jon-betts: I really wish we could pass args normally...

Makefile
jon-betts: The mechanism we are using to run the script is the same as `devdata`, but it's a bit obscure. I think we could make it clearer they are the same by having the lines look the same. So either change the `devdata` line to use `--run command` or to remove it here.

```
tox -qqe dev --run-command 'initdb conf/development.ini'

# vs.

tox -qe dev -- devdata conf/development.ini
tox -qe dev -- pshell conf/development.ini
```

lms/scripts/initdb.py
jon-betts, line 14:
```suggestion
    # Initialise the pyramid environment, which is enough to trigger the initialisation
    # code in `lms.db.__init__.py` to setup the DB for us.
    bootstrap(sys.argv[1])
```

lms/scripts/__init__.py
```diff
@@ -1,3 +1,2 @@
 from lms.scripts.devdata import devdata
-
-__all__ = ["devdata"]
+from lms.scripts.initdb import initdb
```
jon-betts, line 2: The other approach I've seen to this is to have all the scripts have `main` so the target is `lms.scripts.initdb:main`

Makefile
```diff
@@ -31,6 +32,12 @@ services: args?=up -d
...
+db: python
+	@tox -qqe dev --run-command 'initdb conf/development.ini'
```
jon-betts, line 38:
```suggestion 
	# Run lms/scripts/initdb.py
	@tox -qqe dev --run-command 'initdb conf/development.ini'
```

Makefile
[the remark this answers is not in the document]
seanh: I think what we'd need is something equivalent to Make, that we could easily translate our existing Makefiles to, and that provides a better CLI. Maybe http://www.pyinvoke.org/ or something. Realistically it's probably not worth bothering, make is both terrible and just fine

Makefile
seanh: I think `--run-command` is the correct usage.

The `posargs` form (`tox -e <ENV> -- <POSARGS>`) is for when you want to provide an easy way to just replace part of one command that the env would run, and the env has to have explicit support for it in `tox.ini`. For example the tests command in `tox.ini` is `tests: coverage run -m pytest -v -Werror {posargs:tests/unit/}`, which means that to run just one test you can do `tox tests/unit/foo/bar.py` (no `-e tests` is needed because it's the default). Any _other_ commands in the tests env (aside from the command containing the `{posargs}` will still be run unmodified as well.

`--run-command`, on the other hand, completely replaces _all) of the env's commands from `tox.ini`, and is more suitble for cases like devdata, alembic and pshell, where we're reusing the `dev` env but to run a different command. Any `pre_commands` or `post_commands` for the env in `tox.ini` will still be run though, only `commands` are replaced.

The reason why various things in `Makefile`'s still use posargs when they should probably be using `--run-command` is that they were written before we had added the run-command plugin

seanh: Added a commit to tidy this up

seanh:
>  * We call `make` to call `tox` which executes a command 
> * That command is provided by us and links to a script defined in `setup.py`
> * That script calls `pyramid.paster.bootstrap`
> * Which initialises the Pyramid environment
> * Which results in calling the code in `lms.db.__init__`
>
> I don't think there's anything particularly wrong with that, but we could do ourselves a solid by having a few bread crumbs in the comments telling you the next place to look if you are trying to debug this. It's very hard to guess.

Yeah I think it's all probably justified / for the best:

1. `make` exists in or projects to do little more than provide a convenient UI on top of tox and gulp and other underlying tools that we use. Apart from _occassional_ actual usage of make's powers it's basically just a bunch of aliases. People _do not_ seem to like it at all when I try to remove make commands and tell people to run tox or docker-compose or whatever directly.

2. `tox` is definitely providing value. If we removed it, you'd have to create virtualenvs, install and update deps, manage envvars, use pyenv directly, etc etc. Things would not get any simpler or more direct without tox, they'd be equally indirect (you'd still need  pyenv plus a venv) and much more complicated.

3. Our `setup.py` script and it's calling of bootstrap to initialise the Pyramid environment, is just the standard way of adding commands to Pyramid. Straight from the docs: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html Having seen what happens when we ignore that and invent our own ways of adding CLI's to our apps (the `bin/hypothesis` command in h), I think we should just do this in the normal Pyramid way

I added a couple of comments to `Makefile` pointing you to `setup.py`, since this is failry non-obvious. Though to be honest, if you just grep `initdb` or `devdata` you will find all very easily:

```terminal
$ rg initdb
Makefile
38:	@tox -qqe dev --run-command 'initdb conf/development.ini'  # See setup.py for what initdb is.

lms/scripts/__init__.py
2:from lms.scripts.initdb import initdb

lms/scripts/initdb.py
6:    initdb conf/development.ini
13:def initdb():

setup.py
27:    initdb = lms.scripts:initdb
```

lms/scripts/initdb.py
seanh: Done

Issue #1741, opened 2020-05-12
LMS error payload sends either `message` and `error_message`


LMS007:
~~If for any reason apiCall returns a 500 error, it will put the UI in a state to re-request authorization. This is somewhat meaningless and any re-request would likely result in the same error over and over. In this case we should instead put the UI into the `error-report-submission` state where it informs the user to reload the page.~~

**Update**

Updating this ticket to focus only on fixing error payload. Errors returned from LMS should have a variable `message` but not `error_message` 

Currently the client does not handle both cases correctly due to a logical bug in `utils/api.js` and moreso, we don't need to handle both cases but rather should just have 1 consistent way of packaging the error.

jon-betts:
Quick note on HTTP statuses, I think practically you're right. If it's dead it's probably going to be dead the next time.

Semantically though 5xx errors suggest that retrying _might_ actually work. 4xx errors are to say it _definitely_ won't. 

It might be good to see what the error is in case it really should be something else. What's the error condition?

seanh:
> If for any reason apiCall returns a 500 error, it will put the UI in a state to re-request authorization. This is somewhat meaningless and any re-request would likely result in the same error over and over. In this case we should instead put the UI into the error-report-submission state where it informs the user to reload the page.

I don't think this is correct. Showing an error dialog with a **Try again** button was deliberate and by design. What the **Try again** button does is try the whole process again from the start in order to give things the best chance of succeeding on re-try: beginning with getting a brand new access token (in case there was anything wrong with the access or refresh token, there are many things wrong with these) and then re-doing the action (launching the assignment or whatever) with the new access token.

In many cases when there is a 500 error this re-try will in fact succeed.

In some cases it may not succeed--for example if our code simply has a bug that means it's going to fail every time. But we can't detect those cases. Putting the user in an infinite **Try again** loop but with an error dialog that says **If the problem persists...** (followed by support links) was a deliberate design decision.

It looks to me like there isn't an issue to fix here? I think the suggested fix would take a deliberate behaviour and make it worse: removing the **Try again** button deprives users of the ability to kick off a re-try that would sometimes succeed.

> Semantically though 5xx errors suggest that retrying might actually work. 4xx errors are to say it definitely won't.

I actually don't think it's true that _4xx errors are to say it definitely won't_ (succeed) on re-try either. Given that re-trying gets a whole new access token from scratch and then re-tries, it could in fact succeed.

seanh:
[...]

PR #1857, opened 2020-06-25
Add support for 500 errors

LMS007:
- Add new category of error dialog for any 500 error from the server from any request. This provides a dialog informing the user to reload the page.

- Fix test that incorrectly did not contain an error_message but assumed a non-auth error would render. A submission error must contain an error_message otherwise it is assumed that it is a regular authorization error.

--------
fixes https://github.com/hypothesis/lms/issues/1741

I didn't try mundging the error dialogs together and instead whet for readability and just added a 4th dialog. I realize its very redundant but I think the consolidation step can happen later. I see an opportunity to create a component wrapper around the 4 error dialogs inside BasicLtiLaunchApp.js so that we can call just the 1 wrapper with different params, but I would want to keep that wrapper inside BasicLtiLaunchApp.js as its very specific to that file.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -51,7 +51,8 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   } = useContext(Config);
 
   // The current state of the error.
-  // One of "error-fetch", "error-authorizing", or "error-report-submission" or null
+  // One of "error-fetch", "error-authorizing", "error-report-submission",
+  // "error-unrecoverable" or null
   const [errorState, setErrorState] = useState(null);
 
   // Any current error thrown.
@@ -87,12 +88,13 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
    * Helper to handle thrown errors from from API requests.
    *
    * @param {Error} e - Error object from request.
-   * @param {string} newErrorState - One of "error-fetch",
-   *  "error-authorizing", or "error-report-submission"
-   * @param {boolean} [retry=true] - Can the request be retried?
+   * @param {string} newErrorState - One of "error-fetch" or "error-report-submission"
    */
-  const handleError = (e, newErrorState, retry = true) => {
-    if (e instanceof ApiError && !e.errorMessage && retry) {
+  const handleError = (e, newErrorState) => {
+    if (e.status === 500) {
+      setError(e);
+      setErrorState('error-unrecoverable');
+    } else if (e instanceof ApiError && !e.errorMessage) {
       setErrorState('error-authorizing');
     } else {
       setError(e);
@@ -107,6 +109,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     if (apiSync) {
       try {
         setErrorState(null);
+
         const groups = await apiCall({
           authToken,
           path: apiSync.path,
@@ -138,6 +141,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         authToken,
         path: viaCallbackUrl,
       });
+
       setContentUrl(contentUrl);
     } catch (e) {
       handleError(e, 'error-fetch');
@@ -180,8 +184,7 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
       // If reporting the submission failed, replace the content with an error.
       // This avoids the student trying to complete the assignment without
       // knowing that there was a problem, and the teacher then not seeing a
-      // submission.
-      handleError(e, 'error-report-submission', false);
+      handleError(e, 'error-report-submission');
     }
   }, [authToken, canvas.speedGrader, contentUrl]);
 
@@ -293,6 +296,19 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
         >
           <ErrorDisplay
             message="There was a problem submitting this Hypothesis assignment"
+            error={error}
+          />
+          <b>To fix this problem, try reloading the page.</b>
+        </Dialog>
+      )}
+      {errorState === 'error-unrecoverable' && (
+        <Dialog
+          title="Something went wrong"
+          contentClass="BasicLtiLaunchApp__dialog"
+          role="alertdialog"
+        >
+          <ErrorDisplay
+            message="There was an unrecoverable problem with the server"
             error={error}
           />
           <b>To fix this problem, try reloading the page.</b>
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -270,7 +280,7 @@ describe('BasicLtiLaunchApp', () => {
     });
 
     it('displays an error if reporting the submission fails', async () => {
-      const error = new ApiError(400, {});
+      const error = new ApiError(400, { error_message: 'Server error' });
```
LMS007, line 283: This test failed after these changes, but I think the test was incorrect. It failed because there is no longer a `retry=false` param that forces a particular dialog that has no retry button. By the looks of it, we don't need that param, so its just simpler to remove it and just assume any error without a message body is an auth error -- this is the assumption I am going on, my my memory is a bit hazy on this.

I did some quick tests and threw faulty params (as a student) and both the 500 and 422 submission errors I saw did  have `error_message` in their response which supports my raional.

robertknight: Looking at https://github.com/hypothesis/lms/blob/e8ac7fc990716185514827a8d0a1d4b476759d94/lms/views/api/error.py#L13, some 4xx errors have an `error_message` property and some only have a `message` property. What happens here if we get an error with a `message` property.

The fact that different JSON 4xx errors use different property names for the error message makes it harder to handle consistently on the frontend. Is that an oversight @seanh?

robertknight:
Screenshot of what the user will see if a 5xx error occurs during one of the API calls while launching an assignment:

Screenshot 2020-06-26 12 24 00

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -51,7 +51,8 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
   } = useContext(Config);
 
   // The current state of the error.
-  // One of "error-fetch", "error-authorizing", or "error-report-submission" or null
+  // One of "error-fetch", "error-authorizing", "error-report-submission",
+  // "error-unrecoverable" or null
   const [errorState, setErrorState] = useState(null);
```
robertknight, line 56: The `errorState` property has become overloaded to have two different meanings here:

- What activity failed? (`error-fetch`, `error-authorizing` and `error-report-submission` are all about this)
- What _kind_ of error occurred? (`error-uncoverable` is about this)

In the case of `error-unrecoverable` this is also duplicates information that is available in the `error` state.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -298,6 +301,19 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
...
+        <Dialog
+          title="Something went wrong"
+          contentClass="BasicLtiLaunchApp__dialog"
+          role="alertdialog"
+        >
+          <ErrorDisplay
+            message="There was an unrecoverable problem with the server"
+            error={error}
+          />
+          <b>To fix this problem, try reloading the page.</b>
```
robertknight, line 314: An unfortunate problem here is that we've lost information about the context of what the frontend was trying to do that failed.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
LMS007: My preference would be to make assumptions from the status code itself if possible.

LMS007:
@robertknight 

100% agree about the semantic problem with the errorState var. I'm going to make an effort here to restructure some of that so it's more clear,  readable and meaning is not overloaded.

I see your screenshot of the 500 error but I suspect the details are hidden for another reason. The error (from python) returns "message" rather than "error_message". Looking at api.js and then at <DisplayError> and in error.py, I can see that "message" and "error_message" are both used so I think it's worth taking a higher level look to see if we have the correct structure for these errors. This is really where the jsdocs are going to pay off.  However, I added question in slack and I'm hoping we can just use "message" and do away with "error_message" entirely.  I don't know how to simulate that 500 error so I'm making these assumptions just looking at code. By my estimation,  if a 500 error is thrown, then api.js converts that to an ApiError, but it only looks for `error_message`   and not `message`. 

e.g.

`this.errorMessage = data.error_message || null;`

Although in its constructor, it does pass `message` to its base class. Which is a bit confusing to me. 

In any case, <ErrorDisplay> looks for an errorMessage member, and sees none and _does not render the details_. I think that can be fixed by simply tightening up / fixing the error payload in error.py and api.js and so on to allow <ErrorDisplay> to work.

Next, my proposal is to break down the error handler in BasicLtiLaunch by first dividing up the category of errors into 2 types

**1. fetch 
2. submit**

Then individually looking at the error code and rendering slightly different dialogs depending on what we want to do. I think we can also use the same code (with variables) to render all dialog permutations. 

- 1. fetch 
  - 401
    - hide details & ticket urls
    - show auth button
  - 4xx / 5xx
    - show "fetch" type error in title
    - show details & ticket urls
    - show retry button
 

- 2. submit
  - 401
    - this is a bit odd because the fetches would likely fail first so I can't imagine this sort of case. I'm tempted to treat this error just like a fetch 401 and hide the details and show a retry button, but I could use some input.
  - 4xx / 5xx
    - show "submit" type error in title
    - show details & ticket urls
    - hide retry button & prompt user to reload


Current LMS throws a 400 for the auth error but I made a suggestion to turnt hat into a 401. If we can't do that, then we can proceed with the existing way of detecting an auth error and checking for an empty "error_message" (or just "message" if we can get that consolidated too)

LMS007:
I'm also currently blocked on the previous PR because this refactor will affect the same code and vars.
https://github.com/hypothesis/lms/pull/1844/files

seanh:
As [commented on the issue](https://github.com/hypothesis/lms/issues/1741#issuecomment-650564074) I don't think we should make this change. The behaviour for 500 errors that's on production currently was deliberate and I think it's correct. I think we should close this

seanh:
> I'm also not sure about the assumption that it is not worth enabling a retry of 5xx errors whereas it is worth enabling a retry of 4xx errors. A 5xx could be caused by a temporary networking issue on the server end for example. See also @jon-betts comment on the original issue.

Right: the current error dialog with a **Try again** button makes sense for 5xx's and this was done deliberately. We don't want a different dialog for 5xx's

seanh:
I've got a feeling this might not work on production anyway because Cloudflare would intercept and replace the API's 500 response.

PR #1868, opened 2020-06-30
Fix error payload to use `message` rather than `error_message`

LMS007:
Some of the error responses use `error_message` rather than just `message`. This inconsistency created an issue where in some cases, error details were passed but ignored on the client because ApiError was only setting its `errorMessage` param to `error_message` rather than `message`.

fixes https://github.com/hypothesis/lms/issues/1741

This PR is related to closing https://github.com/hypothesis/lms/pull/1857

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/api.js
+++ b/lms/static/scripts/frontend_apps/utils/api.js
@@ -8,7 +8,8 @@
  */
 export class ApiError extends Error {
   constructor(status, data) {
-    const message = data.error_message || data.message || 'API call failed';
+    // If message is omitted, pass a default error message.
+    const message = data.message || 'API call failed';
     super(message);
 
     /**
@@ -26,7 +27,7 @@ export class ApiError extends Error {
      *
      * @type {string|null}
      */
-    this.errorMessage = data.error_message || null;
+    this.errorMessage = data.message || null;
 
     /**
      * Server-provided details of the error.
--- a/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
@@ -207,7 +207,7 @@ describe('BasicLtiLaunchApp', () => {
     [
       {
         description: 'a specific server error',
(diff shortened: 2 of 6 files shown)


lms/static/scripts/frontend_apps/utils/api.js
```diff
@@ -4,7 +4,8 @@
  */
 export class ApiError extends Error {
   constructor(status, data) {
-    const message = data.error_message || data.message || 'API call failed';
```
LMS007, line 7: This was causing the actual bug, it was passing either error_message or message to its base class here, but down on line 25 it was only setting `error_message` to `errorMessage` which is the actual value checked in BasicLtiLaunchApp which will determine if it should render the basic Auth dialog or the more verbose dialog.

lms/static/scripts/frontend_apps/utils/api.js
```diff
@@ -4,7 +4,8 @@
...
+    // If message is omitted, pass undefined to base class, not null.
+    const message = data.message || undefined;
     super(message);
```
LMS007, line 9: passing null here causes `this.message` to be the string "null". undefined seems more semantic which produces a result of ""

lms/static/scripts/frontend_apps/utils/api.js
robertknight, line 7: The default value of "API call failed" was provided so that if code which doesn't know about API errors specifically ends up receiving this error it will have a useful `message` property. That is distinct from `errorMessage` which exists to allow callers who are specifically aware of `ApiError` to extract just the message that came from the server.

lms/static/scripts/frontend_apps/utils/api.js
```diff
@@ -26,7 +27,8 @@ export class ApiError extends Error {
      *
      * @type {string|null}
      */
-    this.errorMessage = data.error_message || null;
+
```
robertknight, line 30: Can you remove the blank line that was added here between the JSDoc comment and the `this.errorMessage` assignment.

robertknight:
I have made the trivial change mentioned in my review and I'm going to go ahead and merge this once the tests pass to avoid conflicts with future work.

PR #1872, opened 2020-07-02
Add test factories to `make shell`

seanh:
Make the test factories available in `make shell`'s environment so you do things like this:

```python
$ make shell
Python 3.6.9 (default, May 26 2020, 13:51:02)

...

In [1]: f.LTIUser()                                                                                                                                                                                     
Out[1]: LTIUser(user_id='8955e3587d3831542127e503dbdcafac8125e98d', oauth_consumer_key='Hypothesise8ef1f48811f42990fadfa0147696bbe', roles='Learner', tool_consumer_instance_guid='f34f2010df2144cc496bf8ff87ecdd5d6a8f5e25', display_name='Karen Paul', email='')
```

### DIFF
--- a/lms/pshell.py
+++ b/lms/pshell.py
@@ -1,3 +1,4 @@
+import sys
 from contextlib import suppress
 
 from transaction.interfaces import NoTransaction
@@ -6,6 +7,12 @@ from lms import models
 
 
 def setup(env):
+    sys.path = ["."] + sys.path
+
+    from tests import factories  # pylint:disable=import-outside-toplevel
+
+    sys.path = sys.path[1:]
+
     request = env["request"]
 
     request.tm.begin()
@@ -18,6 +25,10 @@ def setup(env):
 
     env["m"] = env["models"] = models
     env["m"].__doc__ = "The lms.models package."
+
+    env["f"] = env["factories"] = factories
+    env["f"].__doc__ = "The test factories for quickly creating objects."
+    factories.set_sqlalchemy_session(request.db)
 
     try:
         yield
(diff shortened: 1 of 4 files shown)


lms/pshell.py
```diff
@@ -6,6 +7,12 @@
 
 
 def setup(env):
+    sys.path = ["."] + sys.path
+
+    from tests import factories  # pylint:disable=import-outside-toplevel
+
+    sys.path = sys.path[1:]
```
seanh, line 14: The library `pylti` [contains a `tests` package inside the `pylti` package](https://github.com/mitodl/pylti/tree/master/pylti/tests) and this gets installed at `tox/dev/lib/python3.6/site-packages/tests/`. This means that `import tests` or `from tests import factories` tries to import from pylti's tests dir. This makes it impossible to import our own test factories. So I had to hack the import path

tests/factories/__init__.py
```diff
@@ -1,5 +1,47 @@
...
+def _sqlalchemy_factory_classes():
+    # Return all the SQLAlchemy factory classes from tests.factories.
+    for factory_class in sys.modules[__name__].__dict__.values():
+        try:
+            is_sqla_factory = issubclass(factory_class, SQLAlchemyModelFactory)
+        except TypeError:
+            is_sqla_factory = False
+
+        if is_sqla_factory:
+            yield factory_class
```
seanh, line 47: Moved these methods from `conftest.py` so that `make shell` doesn't have to import `conftest.py` (`conftest.py` imports a bunch of tests-only dependencies that aren't installed in the environment that `make shell` runs in)

PR #1886, opened 2020-07-07
Try and prevent error details from escaping from the box

jon-betts:
Very long single words / tokens can break out of the details area and cause
horizontal scrolling and hidden content.

## To test

This is kind of tricky. You need one of the exceptions that has a massive stack trace in it to see the difference. You could also get any old error and bodge the text to see it.

If I find an easy way I'll add it.

## Results

Before:
image

After:
image

### DIFF
--- a/lms/static/styles/components/_ErrorDisplay.scss
+++ b/lms/static/styles/components/_ErrorDisplay.scss
@@ -10,8 +10,8 @@
   padding-right: 5px;
   margin-bottom: 10px;
   max-width: 100%;
-  
-  &__details {  
+
+  &__details {
     padding: 4px; // prevents focus ring from getting cut off
     &-summary {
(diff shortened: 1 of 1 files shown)


LMS007:
I'm not sure this helps with readability. In some ways it's nice having everything on a single line even if you can't see the end without scrolling. On the other hand, it may not be worth the trouble and what we really want is just the whole thing at once. If this is helpful to you, I say merge it!

But I also want to follow up about the "error" json. It looks like its got extra escaping and I just wonder why? It makes it harder to read? I wonder if we're incorrectly parsing it in the client.

jon-betts:
I think horizontal scrolling is a no, no. It also just looks messy. This allows you to see all the content on the page in one go too.

The error JSON double escaping is a result of the "body" being a string... which also happens to be JSON.

PR #1893, opened 2020-07-09
Remove prototype matcher for h-matchers version

jon-betts:
This PR takes a matcher which was written here and sub-sequently moved out to `h-matchers` and replaces it with the new version now in `h-matchers`. 

## Review notes

* The tests haven't been changed at all, the request matcher is just gone
* There was some tomfoolery with the imports I fixed as well

### DIFF
--- a/tests/unit/lms/services/canvas_api_test.py
+++ b/tests/unit/lms/services/canvas_api_test.py
@@ -4,11 +4,9 @@ from unittest import mock
 from unittest.mock import call, create_autospec, sentinel
 
 import pytest
-from _pytest.mark import param
 from h_matchers import Any
-from h_matchers.decorator import fluent_entrypoint
-from h_matchers.matcher.core import Matcher
-from requests import PreparedRequest, Request, RequestException, Response
+from pytest import param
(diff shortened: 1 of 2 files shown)


tests/unit/lms/services/canvas_api_test.py
```diff
@@ -251,9 +196,9 @@ def test_list_files(self, api_client, http_session):
         response = api_client.list_files("COURSE_ID")
 
         assert response == files
-        AnyRequest.assert_on_comparison = True
```
jon-betts, line 254: This was some debugging stuff left in. This is still present in the new matcher, but currently private as `_assert_on_comparison`.

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -4,11 +4,9 @@
 from unittest.mock import call, create_autospec, sentinel
 
 import pytest
-from _pytest.mark import param
```
jon-betts, line 7: Eugh. I need to watch out for this. Auto import in Pycharm appears to detect this private package, which is probably where this is actually defined, above the main exported version.

Issue #1753, opened 2020-05-14
SpeedGrader Canvas API sync call finishing touches


jon-betts:
We _think_ the code can be safely simplified by:

 * [x] Making the call to get the sections as we do
 * [x] Using the users sections to filter them
 * [x] Unifying the path from there
 * [x] Separating out the authority provided id function (This part has happened in #1894)

The tests could also be simplified by:

 * [x] Mocking the authority function above
 * ~Separating id generation from other assertions by using `Any.object()` calls~ Doesn't matter now
 * [x] Try to remove the parts with duplication or where hex strings are the only links
 * [x] Using more descriptive names for the sections to describe what they are for in the tests

jon-betts:
Finished off the rest of the ideas in: https://github.com/hypothesis/lms/pull/1895

jon-betts:
This has been merged out now, although I might have spotted an existing bug: https://hypothes-is.slack.com/archives/CBN3DGW02/p1594666915274300

PR #1894, opened 2020-07-10
Centralise model object creation and refactor Canvas sync tests

jon-betts:
Partially achieves: https://github.com/hypothesis/lms/issues/1753

 * We now have a re-usable "hashed_id" function which the models use to create ids
 * This replaces all the places which performed hashed based ids before
 * All construction logic has been centralised into the models
 * Services are responsible for getting the right values, and the model is responsible for combining them
 * This simplifies test and removes lots of "magic" hex values which opaquely line up with input strings
 * Went to town as a result of this on the Canvas sync tests to try and make them easier to understand

## Review notes

 * The changes have been broken up into four commits
 * The commits do not work individually, but hopefully help when reading it
 * The biggest changes are to the sync tests (which was the motivator for this)

### DIFF
--- a/lms/models/_hashed_id.py
+++ b/lms/models/_hashed_id.py
@@ -1,0 +1,18 @@
+import hashlib
+
+
+def hashed_id(*parts):
+    """
+    Create a hashed id from multiple stringifying parts.
+
+    It is crucial that if any one of the parts changes meaningfully that it's
+    stringification changes too.
+
+    :param parts: An objects which can be converted to strings
+    :return: A string which can be used as an id.
+    """
+    hash_object = hashlib.sha1()
+    for part in parts:
+        hash_object.update(str(part).encode())
+
+    return hash_object.hexdigest()
--- a/lms/models/_hashed_id.py
+++ b/lms/models/_hashed_id.py
@@ -8,8 +8,8 @@ def hashed_id(*parts):
     It is crucial that if any one of the parts changes meaningfully that it's
     stringification changes too.
 
-    :param parts: An objects which can be converted to strings
-    :return: A string which can be used as an id.
+    :param parts: An iterable of objects which can be converted to strings
+    :return: A string which can be used as an id
     """
     hash_object = hashlib.sha1()
     for part in parts:
--- a/lms/models/h_group.py
+++ b/lms/models/h_group.py
@@ -2,63 +2,60 @@ from typing import NamedTuple
 
 from lms.models._hashed_id import hashed_id
 
+MAX_GROUP_NAME_LENGTH = 25
+
 
 class HGroup(NamedTuple):
     name: str
     authority_provided_id: str
-    type: str = "course_group"
+    type: str
 
     def groupid(self, authority):
         return f"group:{self.authority_provided_id}@{authority}"
 
     @classmethod
-    def from_lti_parts(
-        # pylint: disable=too-many-arguments
-        cls,
-        name,
-        tool_consumer_instance_guid,
-        context_id,
-        section_id=None,
-        type_=type,
-    ):
+    def course_group(cls, course_name, tool_consumer_instance_guid, context_id):
         """
-        Create an HGroup from LMS specific parts.
+        Create an HGroup for a course.
 
-        :param name: The name of the course
+        :param course_name: The name of the course
         :param tool_consumer_instance_guid: Tool consumer GUID
         :param context_id: Course id
-        :param section_id: A section id for a section group
-        :param type_: Group type (defaults to "course_group")
         """
         return HGroup(
-            name=cls._name(name) if name is not None else None,
-            authority_provided_id=cls._authority_provided_id(
-                tool_consumer_instance_guid, context_id, section_id
-            ),
-            type=type_,
+            cls._name(course_name),
+            hashed_id(tool_consumer_instance_guid, context_id),
+            type="course_group",
+        )
+
+    @classmethod
+    def section_group(
+        cls, section_name, tool_consumer_instance_guid, context_id, section_id
+    ):
+        """
+        Create an HGroup for a course section.
+
+        :param section_name: The name of the section
+        :param tool_consumer_instance_guid: Tool consumer GUID
+        :param context_id: Course id the section is a part of
+        :param section_id: A section id for a section group
+        """
+        return HGroup(
+            cls._name(section_name),
+            hashed_id(tool_consumer_instance_guid, context_id, section_id),
+            type="section_group",
         )
 
     @classmethod
     def _name(cls, name):
         """Return an h-compatible group name from the given string."""
 
+        if name is None:
+            return None
+
         name = name.strip()
 
-        # The maximum length of an h group name.
-        group_name_max_length = 25
-
-        if len(name) > group_name_max_length:
-            name = name[: group_name_max_length - 1].rstrip() + "…"
+        if len(name) > MAX_GROUP_NAME_LENGTH:
+            return name[: MAX_GROUP_NAME_LENGTH - 1].rstrip() + "…"
 
         return name
-
-    @classmethod
-    def _authority_provided_id(
-        cls, tool_consumer_instance_guid, context_id, section_id=None
-    ):
-        """Return an h-compatible authority_provided_id from the LTI parts."""
-
-        if section_id is None:
-            return hashed_id(tool_consumer_instance_guid, context_id)
-
-        return hashed_id(tool_consumer_instance_guid, context_id, section_id)
--- a/lms/models/h_group.py
+++ b/lms/models/h_group.py
@@ -1,5 +1,7 @@
 from typing import NamedTuple
 
+from lms.models._hashed_id import hashed_id
+
 
 class HGroup(NamedTuple):
     name: str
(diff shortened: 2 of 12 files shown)


tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,140 +1,148 @@
...
-    assert returned_groupids == [
-        f"group:{group.authority_provided_id}@TEST_AUTHORITY"
-        for group in expected_groups
-    ]
+pytestmark = pytest.mark.usefixtures("canvas_api_client", "lti_h_service")
 
 
-pytestmark = pytest.mark.usefixtures("canvas_api_client", "lti_h_service")
+@pytest.fixture
+def assert_sync_and_return(lti_h_service, request_json):
```
jon-betts, line 62: This is the main guy that pulls all of this together. It roughly maps to the previous `assert_that_it_called_sync_and_returned_the_groupids` but with the following changes:

 * It accepts sections and converts them into the right groups itself
 * It uses groupid function instead of replicating the logic
 * It accepts two lists of sections, instead of one, allowing it to be used in all three tests (the speedgrader one has different values)

tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,140 +1,148 @@
...
+    def assert_return_values(groupids, synced_sections, returned_sections):
+        lti_h_service.sync.assert_called_once_with(
+            _groups_for(synced_sections), request_json["group_info"]
+        )
+
+        assert groupids == [
+            group.groupid(TEST_SETTINGS["h_authority"])
+            for group in _groups_for(returned_sections)
+        ]
+
```
jon-betts, line 87: A caller pattern fixture which has a closure around a closure has to be some kind of `pytest` bingo win.

tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,140 +1,148 @@
...
+
+@pytest.fixture
+def is_speedgrader(request_json):
+    request_json["learner"] = {"canvas_user_id": "user_id"}
+
+
+@pytest.fixture
+def sections():
+    class Sections:
```
jon-betts, line 108: Put all the section data into a fixture that we can directly refer to in the tests. This is another part of making this easier to understand the flow of.

Previously we had these values in the `canvas_api_client` alone, and then we constructed separate groups that happened to correspond to them with hash values. 

Now we pass these around directly so instead of creating a set of groups we say `_groups_for(sections.user)` etc.

tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,140 +1,148 @@
...
 @pytest.fixture
-def pyramid_request(pyramid_request):
-    pyramid_request.json = {
+def pyramid_request(pyramid_request, request_json):
+    pyramid_request.json = request_json
+    return pyramid_request
+
+
+@pytest.fixture
+def request_json():
```
jon-betts, line 140: Similar to splitting the sections data out of `canvas_api_client` the request JSON body has been separated out to make it a bit easier to access.

lms/models/_hashed_id.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def hashed_id(*parts):
+    """
+    Create a hashed id from multiple stringifying parts.
+
+    It is crucial that if any one of the parts changes meaningfully that it's
+    stringification changes too.
+
+    :param parts: An objects which can be converted to strings
```
seanh, line 11:
```suggestion
    :param parts: An iterable of objects which can be converted to strings
```

lms/models/_hashed_id.py
```diff
@@ -0,0 +1,18 @@
...
+    :return: A string which can be used as an id.
```
seanh, line 12:
```suggestion
    :return: A string which can be used as an id
```

lms/models/_hashed_id.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    hash_object = hashlib.sha1()
+    for part in parts:
+        hash_object.update(str(part).encode())
+
+    return hash_object.hexdigest()
```
seanh, line 18: :+1:

lms/models/h_group.py
```diff
@@ -9,15 +11,54 @@ class HGroup(NamedTuple):
...
+        :param section_id: A section id for a section group
+        :param type_: Group type (defaults to "course_group")
+        """
+        return HGroup(
+            name=cls._name(name) if name is not None else None,
+            authority_provided_id=cls._authority_provided_id(
+                tool_consumer_instance_guid, context_id, section_id
+            ),
+            type=type_,
+        )
```
seanh, line 39: I don't think "lti parts" is quite the right term:

* Name isn't always an LTI thing (it does come from an LTI param for course groups, for section groups it comes from a non-LTI Canvas API param)
* `section_id` is a Canvas API thing not an LTI thing
* `type` is unrelated to LTI

But I'm not sure what a good name would be. Something like "group params" isn't very good.

I think PyLint's too-many-arguments warning may suggest a missing class:

```python
class HGroupParams(NamedTuple):
    """Params from which an `HGroup` can be created."""
    name: str
    context_id: str
    section_id: str=None
    type: str = "course_group"  # `type` default can now be removed from HGroup
```

But adding a class just to hold the params of a method, seems wrong.

So to be honest, at the cost of just a little duplication (almost none really), I think you can neatly avoid both issues and have nice meaningful `course_group()` and `section_group()` method names:

```python
    @classmethod
    def course_group(cls, name, tool_consumer_instance_guid, context_id):
        return HGroup(
            cls._name(name),
            hashed_id(tool_consumer_instance_guid, context_id),
        )

    @classmethod
    def section_group(cls, name, tool_consumer_instance_guid, context_id, section_id):
        return HGroup(
            cls._name(name),
            hashed_id(tool_consumer_instance_guid, context_id, section_id),
            type="section_group",
        )

    @classmethod
    def _name(cls, name):
        """Return an h-compatible group name from the given string."""
        if name is None:
            return None

        ...
```

(And note that `HGroup._authority_provided_id()` is now no longer needed.)


This would look pretty nice in the code that calls it:

```python
return HGroup.course_group(
    name=params["context_title"],
    tool_consumer_instance_guid=params["tool_consumer_instance_guid"],
    context_id=params["context_id"],
)

return HGroup.section_group(
    name=section.get("name"),
    tool_consumer_instance_guid=tool_consumer_instance_guid,
    context_id=context_id,
    section_id=section["id"],
)
```

lms/models/_hashed_id.py
```diff
@@ -0,0 +1,18 @@
...
+
+    It is crucial that if any one of the parts changes meaningfully that it's
+    stringification changes too.
+
+    :param parts: An objects which can be converted to strings
+    :return: A string which can be used as an id.
+    """
+    hash_object = hashlib.sha1()
+    for part in parts:
+        hash_object.update(str(part).encode())
```
seanh, line 16: The centralisation of both `str()` and `encode()` is pretty nice

lms/models/h_user.py
```diff
@@ -35,15 +36,8 @@ def from_lti_user(cls, lti_user):
...
 
-        def username():
-            """Return the h username for the current request."""
-            username_hash_object = hashlib.sha1()
-            username_hash_object.update(provider.encode())
-            username_hash_object.update(provider_unique_id.encode())
-            return username_hash_object.hexdigest()[:30]
-
         return cls(
-            username=username(),
```
seanh, line 46: :+1:

lms/resources/lti_launch.py
```diff
@@ -54,16 +53,13 @@ def h_group(self):
...
-        hash_object = hashlib.sha1()
-        hash_object.update(
-            self._request.parsed_params["tool_consumer_instance_guid"].encode()
-        )
-        hash_object.update(self._request.parsed_params["context_id"].encode())
-        authority_provided_id = hash_object.hexdigest()
 
-        return HGroup(
-            name=h_group_name(self._request.parsed_params["context_title"]),
-            authority_provided_id=authority_provided_id,
```
seanh, line 66: :+1:

lms/views/api/canvas/sync.py
```diff
@@ -30,20 +26,19 @@ def sync(request):
...
-        hash_object.update(context_id.encode())
-        hash_object.update(str(section["id"]).encode())
-        authority_provided_id = hash_object.hexdigest()
-
-        if "name" in section:
-            name = h_group_name(section["name"])
-        else:
-            name = None
-
-        return HGroup(name, authority_provided_id, type="section_group")
```
seanh, line 46: :+1:

lms/models/_hashed_id.py
jon-betts: "an objects"...

lms/models/h_group.py
jon-betts: Yeah, that's better

jon-betts: This is split up now and has a few benefits:

 * Type management is now an internal affair. You can't create a course group with a section type, or a random string
 * We can now name "name" as "course_name" / "section_name" which is more self documenting
 * It's also more accurate, as it's _not_ the eventual group name that is derived from it in all cases
 * Makes reading the places that create it much more obvious in intent as well

PR #1895, opened 2020-07-10
Simpler clearer Canvas Sync API

jon-betts:
By unifying the flow of speedgrader with the other ways of syncing we can now:

 * Split each section up into discrete parts
 * Remove an inner function (as it's only used once)
 * Remove the path which allowed `HGroups` to be created without a name

As a result the main sync function is now:

```python
def sync(self):
    groups = self._to_groups(self._get_sections())

    self._sync_to_h(groups)

    authority = self._request.registry.settings["h_authority"]
    return [group.groupid(authority) for group in groups]
```

This is for: https://github.com/hypothesis/lms/issues/1753

### DIFF
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -3,49 +3,69 @@ from pyramid.view import view_config
 from lms.models import HGroup
 
 
-@view_config(
-    route_name="canvas_api.sync",
-    request_method="POST",
-    renderer="json",
-    permission="sync_api",
-)
-def sync(request):
-    authority = request.registry.settings["h_authority"]
-    canvas_api_svc = request.find_service(name="canvas_api_client")
-    lti_h_svc = request.find_service(name="lti_h")
+class Sync:
+    def __init__(self, request):
+        self._request = request
 
-    course_id = request.json["course"]["custom_canvas_course_id"]
-    group_info = request.json["group_info"]
+    @view_config(
+        route_name="canvas_api.sync",
+        request_method="POST",
+        renderer="json",
+        permission="sync_api",
+    )
+    def sync(self):
+        groups = self._to_groups(self._get_sections())
 
-    if request.lti_user.is_learner:
-        # For learners we only want the client to show the sections that the
-        # student belongs to, so fetch only the user's sections.
-        sections = canvas_api_svc.authenticated_users_sections(course_id)
-    else:
+        self._sync_to_h(groups)
+
+        authority = self._request.registry.settings["h_authority"]
+        return [group.groupid(authority) for group in groups]
+
+    @property
+    def _is_speedgrader(self):
+        return "learner" in self._request.json
+
+    def _get_sections(self):
+        canvas_api = self._request.find_service(name="canvas_api_client")
+        course_id = self._request.json["course"]["custom_canvas_course_id"]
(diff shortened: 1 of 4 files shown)


tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,55 +1,47 @@
...
-    groupids = sync(pyramid_request)
+    groupids = Sync(pyramid_request).sync()
 
     course_id = request_json["course"]["custom_canvas_course_id"]
     canvas_api_client.authenticated_users_sections.assert_called_once_with(course_id)
 
-    assert_sync_and_return(
-        groupids,
-        synced_sections=sections.authenticated_user,
-        returned_sections=sections.authenticated_user,
```
jon-betts, line 20: As all methods now return what they sync, these two things can be collapsed

jon-betts:
Looks like this is failing on formatting. Some divergence between local and remote here. Looks all good on my side.

jon-betts:
Eugh, just "ran it again" without any changes and it was fine.

lms/views/api/canvas/sync.py
```diff
@@ -3,50 +3,70 @@
...
-
-    tool_consumer_instance_guid = request.json["lms"]["tool_consumer_instance_guid"]
-    context_id = request.json["course"]["context_id"]
+        sections = canvas_api.course_sections(course_id)
+        if not self._is_speedgrader:
+            return sections
 
-    def group(section):
-        """Return an HGroup from the given Canvas section dict."""
+        # Speedgrader requests are made by the teacher, but we want the
```
seanh, line 44:
```suggestion
        # SpeedGrader requests are made by the teacher, but we want the
```

tests/unit/lms/models/h_group_test.py
```diff
@@ -45,32 +44,32 @@ def test_sections_group(self, hashed_id):
...
-                lambda name: HGroup.course_group(name, "blah", "blah"),
-                id="course_group",
-            ),
-            param(
-                lambda name: HGroup.section_group(name, "blah", "blah", "blah"),
-                id="section_group",
-            ),
-        ),
-    )
-    def test_contructors_truncate_the_name(self, constructor, name, expected_result):
```
jon-betts, line 69: Contructors untie!

jon-betts:
I've changed it a bit based on modification of the underlying PR, but it hasn't changed code wise. Unlike the previous PR, which was straight refactoring, this is a change in behavior, so I need to give this a test before it goes any further.

jon-betts:
Testing this locally has highlighted what I think is an esoteric and unrelated bug. On my system trying to access SpeedGrader shows all user names as "Couldn't fetch student name". I think this is because the users haven't launched it ever locally (I may have blown away my DB at some point). 

When the teacher does a SpeedGrader launch, their name is synced, but the users name is not. If the students name was synced instead / in addition it would probably fix this. I can't think of a situation where this matters though. If a user has annotated with us then their name will appear. I suppose this could close a gap if the student never did anything to do with us at all? The teacher would then see their name currently (with no annotations).

I can confirm doing anything as the student relating to us fixes it (just look at _any_ assignment).

I'm assuming this is fine and I'm doing to deploy this out.

PR #1904, opened 2020-07-14
Split up the Canvas API into multiple parts with composition (original)

jon-betts:
Previously we had one very big file with two major components which inherited from each other. This caused some issues:

 * The tests were horribly long
 * It's hard to find anything as the code is very long
 * The code is sufficiently big that we need to mock inner parts out
 * Inheritance made this a bit more tricky using our usual patching style

This instead:

 * Splits into multiple discrete components which are composed
 * The components are more fine grained than before
 * All Pyramid stuff is kept out of them up-front with a factory method that puts them all together

## Review notes

This is currently hilariously not finished. This is in this state just to give it a sense check before I plough a lot of effort into testing.


jon-betts:
Eugh, this is kind of doing my head in. I have a persistent lack of coverage over lines I'm pretty much 100% sure are covered. I'm not sure if this is a coverage bug or if I'm doing something wrong.

Anyway, _aside_ from that, there's 100% coverage again. This could do with a documentation pass, but then it's done. Sadly at that point I really need to try and carve this up into sensible commits, because it's absolutely massive. There's no getting around that. Some very big files have been split into 5 or more each, and in some cases heavily rewritten.

jon-betts:
During going over this again for documentation I spotted some weird loopy calling we had, and I've streamlined this a bit now so:

 * The `BasicClient` has no `make_request` which you call to pass to `send_and_validate`
 * There is just `send()` now
 * This means from the top:
    * `CanvasAPIClient` calls `AuthenticatedClient.send()`
   * `AuthenticatedClient` calls `BasicClient.send()`
   * `BasicClient` calls `request.Session.send()`
   * All very orderly!
 * Required some more test rewriting
 * I also moved a big chunk of text out of the classes and into a separate README.

PR #1906, opened 2020-07-14
Fix Google API loader typecheck errors

robertknight:
Add the `@types/gapi` package recommended in the Google API loader's
README [1]. This fixes the errors reported in this file when running `yarn typecheck`.

[1] https://github.com/google/google-api-javascript-client#features

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/google-api-client.js
+++ b/lms/static/scripts/frontend_apps/utils/google-api-client.js
@@ -12,8 +12,8 @@ async function loadGAPI() {
     gapiScript.onload = () => {
       resolve(window.gapi);
     };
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/utils/google-api-client.js
```diff
@@ -12,8 +12,8 @@ async function loadGAPI() {
     gapiScript.onload = () => {
       resolve(window.gapi);
     };
-    gapiScript.onerror = ({ error }) => {
```
robertknight, line 15: This is a mistake caught by TS. The `onerror` handler gets a [generic `Event`](https://w3c.github.io/uievents/#event-type-error) with no `error` property.

PR #1907, opened 2020-07-14
Add ApplicationInstance.settings and Course.settings server_defaults

seanh:
...and also make `ApplicationSettings.settings` not-nullable like `Course.settings` already is.

Make `ApplicationInstance.settings` default to an empty dict rather than `NULL` at the Postgres level, and make it not nullable at the Postgres level.

In practice there are no NULLs in the QA or prod DBs. Let's enforce that at the DB level so that we don't need to worry about enforcing it (or potentially failing to enforce it) at the Python level (including both Python app code and future migration scripts).


seanh:
This is gonna need a DB migration to make these schema changes in QA and prod...

PR #1909, opened 2020-07-14
Make ApplicationSettings a SQLAlchemy type

seanh:
**Depends on** https://github.com/hypothesis/lms/pull/1907 for the tests to pass.

This means that instead of doing this:

https://github.com/hypothesis/lms/blob/2e7133a6714a06347d1fbc3d080d602e755ec44e/lms/models/application_instance.py#L34-L50

which is repeated in both [`ApplicationInstance`](https://github.com/hypothesis/lms/blob/2e7133a6714a06347d1fbc3d080d602e755ec44e/lms/models/application_instance.py#L34-L50) and [`Course`](https://github.com/hypothesis/lms/blob/2e7133a6714a06347d1fbc3d080d602e755ec44e/lms/models/course.py#L26-L35), both places (and any future ones) can now just do this one-liner:

https://github.com/hypothesis/lms/blob/90d8df42328c7a594a981964dd55f34832684154/lms/models/application_instance.py#L34

# Benefits

* The boilerplate code that's required in each model class to use `ApplicationSettings` is much shorter
* `ApplicationSettings` attributes on models behave like you would expect them to: the same as any other SQLAlchemy attribute. This differs from before in a bunch of ways
* Models no longer have separate `_settings`, `settings` and `settings.data` attributes. There's now just one `settings` attr which is an `ApplicationSettings` object

# Demo

## New "transient" models get `settings=None` by default

When you create a new `ApplicationInstance` (for example) that isn't in the DB yet, `settings` is initially `None` and so trying to use it can crash:

```python
>>> ai = models.ApplicationInstance()                                                                                                                                                               
>>> ai.settings is None                                                                                                                                                                             
True

>>> ai.settings.get("canvas", "sections_enabled")                                                                                                                                                   
AttributeError: 'NoneType' object has no attribute 'get'
```

This could be fixed by adding an `__init__()` that sets a default value. Something like:

```python
class ApplicationInstance(Base):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.settings = self.settings or ApplicationSettings()
```

But I don't think we should do this. Defaulting to `None` is okay because this is how SQLAlchemy model attributes normally work. For example the `consumer_key` attribute (a string) is the same:

```python
>>> ai = models.ApplicationInstance()                                                                                                                                                               
>>> ai.consumer_key is None                                                                                                                                                                         
True

>>> ai.consumer_key.upper()                                                                                                                                                                         
AttributeError: 'NoneType' object has no attribute 'upper'
```

## Passing a dict to `settings` turns it into an `ApplicationSettings`

I'm not actually sure how / why this works :) But if you initialize an `ApplicationInstance` passing a plain dict as `settings` it gets turned into an `ApplicationSettings`:

```python
>>> ai = models.ApplicationInstance(settings={"canvas": {"sections_enabled": True}})                                                                                                                
>>> ai.settings                                                                                                                                                                                    
ApplicationSettings({'canvas': {'sections_enabled': True}})

>>> ai.settings.get("foo", "bar") is None                                                                                                                                                          
True
```

Assigning to `settings` also does the same:

```python
>>> ai = models.ApplicationInstance()                                                                                                                                                              
>>> ai.settings = {"canvas": {"sections_enabled": True}}                                                                                                                                           
>>> ai.settings                                                                                                                                                                                    
ApplicationSettings({'canvas': {'sections_enabled': True}})
```

## Persistent objects with `NULL` in the `settings` column get `settings=None`

The `application_instances.settings` column is currently nullable in the DB. If you retrieve a row that has `NULL` in the column it gets `settings=None`:

```python
>>> ai = db.query(models.ApplicationInstance).filter_by(consumer_key='Hypothesis84a316e31206259c3c58cbf7e8b0cbc8').one()                                                                           
>>> ai.settings is None                                                                                                                                                                            
True
```

This is annoying because it means that any code that wants to use `settings` has to account for the fact that it might be `None`. This is fixable though: we can make `settings` not nullable and default to an empty dict at the Postgres level: https://github.com/hypothesis/lms/pull/1907. This seems like the right way to handle this, as this is how you would normally make any other type of column required and have a default. It's also quite nice and explicit: it gives the default value and `nullable=False` right on the model attribute in the class (which is where you'd normally expect to find defaults and `nullable=False` for any other type of SQLAlchemy attribute. The `ApplicationSettings` and `ApplicationInstance` classes (and the Python code generally) don't need to deal with null values or defaults except for specifying them in the attribute declaration:

```python
class ApplicationInstance:
    settings = Column(ApplicationSettings.as_mutable(JSONB), server_default=text("'{}'::jsonb"), nullable=False)
```

## Persistent objects with non-NULL in the `settings` column get `ApplicationSettings` objects

No surprise here:

```python
>>> ai = db.query(models.ApplicationInstance).filter_by(consumer_key='Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b').one()                                                                           
>>> ai.settings                                                                                                                                                                                    
ApplicationSettings({'canvas': {'sections_enabled': True}})
```

## You can mutate the dict directly and then call `changed()`

If you use `ApplicationSettings`'s `set()` method then changes will automatically be persisted.

`ApplicationSettings` is also a `dict` subclass so you can mutate it with normal `dict` methods. Normally any top-level mutations will be persisted (since `ApplicationSettings` subclasses SQLAlchemy's `MutableDict` which supports top-level mutations) but any nested mutations will not be persisted unless you call `changed()`:

```python
>>> ai = db.query(models.ApplicationInstance).filter_by(consumer_key='Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b').one()                                                                           
>>> ai.settings["canvas"]["sections_enabled"] = False                                                                                                                                              
>>> ai.settings.changed()
>>> tm.commit()
```

Direct mutation also enables you to make the dict invalid, but I think that's fixable. Covered further down below.

## All of the normal `dict` methods are accessible

`ApplicationSettings` no longer composes a dict (it was previously accessible as `ai.data`), it now subclasses `MutableDict` (which subclasses `dict`). This means that all of the normal methods for reading and writing dicts, comparisons, etc are available.

We _could_ override all of the standard `dict` methods in `ApplicationSettings` and disable them, forcing use of the safe `get()` and `set()` methods. But I don't think we should bother. I think the more Pythonic thing to do is just document "good behaviour" in the `ApplicationSettings` docstring and leave it up to the user.

It's a little weird that `ApplicationSettings` behaves in every way like a normal `dict` _except_ that its `get()` method has a different signature and behaviour from `dict.get()`.

## Python's normal copy protocol works

Since `ApplicationSettings` is a dict subclass you can copy (clone) it in the same way you'd clone a normal dict or other standard Python object: by using the `copy` module. The custom `clone()` method that `ApplicationSettings` used to have is no longer needed. There's probably all sorts of other standard Python stuff that now works too:

```python
>>> import copy                                                                                                                                                                                    
>>> ai.settings                                                                                                                                                                                    
ApplicationSettings({'foo': 'bar', 'canvas': {'sections_enabled': False}})

>>> settings_copy = copy.deepcopy(ai.settings)                                                                                                                                                     
>>> settings_copy                                                                                                                                                                                  
ApplicationSettings({'foo': 'bar', 'canvas': {'sections_enabled': False}})

>>> settings_copy.set("canvas", "sections_enabled", True)                                                                                                                                          
>>> settings_copy                                                                                                                                                                                  
ApplicationSettings({'foo': 'bar', 'canvas': {'sections_enabled': True}})

>>> ai.settings                                                                                                                                                                                    
ApplicationSettings({'foo': 'bar', 'canvas': {'sections_enabled': False}})
```

## The `Course` test factory

The `Course` test factory creates an empty `ApplicationSettings` by default but lets you specify it using a dict:

```python
>>> course = factories.Course()                                                                                                                                                                     
>>> course.settings                                                                                                                                                                                 
ApplicationSettings({})

>>> course = factories.Course(settings={"foo": "bar"})                                                                                                                                              
>>> course.settings                                                                                                                                                                                 
ApplicationSettings({'foo': 'bar'})
```
## You can set `settings` to a malformed dict

Because you can directly assign `settings` using a dict and can also mutate it using `dict` methods directly, it's possible to break things. I think this is probably another case of "you should go through the `get()` and `set()` methods and if you don't it's your responsibility":

```python
>>> ai = models.ApplicationInstance(settings={"foo": 23})                                                                                                                                           
>>> ai.settings.get("foo", "bar")                                                                                                                                                                   
AttributeError: 'int' object has no attribute 'get'
```

We _could_ prevent this by adding validation to the `ApplicationSettings` class. `ApplicationSettings` would have to override `__init__()` and all `dict` mutator methods and validate the value (e.g. raise if a top-level key is being set to a non-dict value). I don't think there's a SQLAlchemy way of doing this--I think we'd have to do it in a Python way, by overriding all the dict mutator methods in `ApplicationSettings`.

I'm not sure whether this is worthwhile?

I'd like to do it in a separate PR or not at all.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -5,7 +5,6 @@ import sqlalchemy as sa
 from Crypto import Random
 from Crypto.Cipher import AES
 from sqlalchemy.dialects.postgresql import JSONB
-from sqlalchemy.ext.mutable import MutableDict
 
 from lms.db import BASE
 from lms.models.application_settings import ApplicationSettings
@@ -31,28 +30,13 @@ class ApplicationInstance(BASE):
         server_default=sa.sql.expression.true(),
         nullable=False,
     )
-    _settings = sa.Column(
+
+    settings = sa.Column(
         "settings",
-        MutableDict.as_mutable(JSONB),
+        ApplicationSettings.as_mutable(JSONB),
         server_default=sa.text("'{}'::jsonb"),
         nullable=False,
     )
-
-    @property
-    def settings(self):
-        """
-        Get a settings object which can get and set grouped settings.
-
-        The object can be queried with `settings.get(group, key)` or
-        `settings.set(group, key, value)`.
-
-        :rtype: models.ApplicationSettings
-        """
-
-        if self._settings is None:
-            self._settings = {}
-
-        return ApplicationSettings(self._settings)
 
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
@@ -93,7 +77,7 @@ class ApplicationInstance(BASE):
             developer_secret=encrypted_secret,
             aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
-            _settings=settings,
+            settings=settings,
         )
 
 
--- a/lms/models/application_settings.py
+++ b/lms/models/application_settings.py
@@ -2,11 +2,41 @@ from sqlalchemy.ext.mutable import MutableDict
 
 
 class ApplicationSettings(MutableDict):
-    """Model for accessing and updating application settings."""
+    """
+    Model for accessing and updating application settings.
+
+    This is a dict subclass, but you should use the custom get() and set()
+    methods below to get and set values because they're safer and more
+    convenient.
+
+    If you mutate a nested dict or list in-place this change won't be detected
+    by SQLAlchemy and your change **will not be saved**.
+
+    For example in the code below the change to "sections_enabled" will not be
+    saved when the transaction is committed:
+
+    >>> ai = db.query(models.ApplicationInstance).filter_by(...).one()
+    >>> ai.settings["canvas"]["sections_enabled"] = False
+
+    The safe thing to do is to use the custom set() method because it makes
+    sure that your changes are saved.
+
+    But alternatively you can call changed() after making your changes, then
+    they will be saved. In this example the changes to both "sections_enabled"
+    and "bar" will be saved:
+
+    >>> ai = db.query(models.ApplicationInstance).filter_by(...).one()
+    >>> ai.settings["canvas"]["sections_enabled"] = False
+    >>> ai.settings["foo"]["bar"] = "gar"
+    >>> # Notify SQLAlchemy that ai.settings has changed so it knows to save it.
+    >>> ai.settings.changed()
+    """
 
     def get(self, group, key):
         """
-        Get a specific setting or None if it doesn't exist.
+        Return the requested setting or None.
+
+        Will return None if *either* "group" or "key" is missing from the dict.
 
         :param group: The name of the group of settings
         :param key: The key in that group
@@ -17,6 +47,10 @@ class ApplicationSettings(MutableDict):
     def set(self, group, key, value):
         """
         Set a specific setting in a group.
+
+        Will create the sub-dict "group" for you if it's missing from the dict,
+        and will notify SQLAlchemy of your change so that it gets saved to the
+        DB.
 
         :param group: The name of the group of settings
         :param key: The key in that group
--- a/lms/models/application_settings.py
+++ b/lms/models/application_settings.py
@@ -28,3 +28,9 @@ class ApplicationSettings(MutableDict):
 
     def clone(self):
         return ApplicationSettings(deepcopy(self))
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({super().__repr__()})"
+
+    def __str__(self):
+        return repr(self)
--- a/lms/models/application_settings.py
+++ b/lms/models/application_settings.py
@@ -1,11 +1,10 @@
 from copy import deepcopy
 
+from sqlalchemy.ext.mutable import MutableDict
 
-class ApplicationSettings:
+
+class ApplicationSettings(MutableDict):
     """Model for accessing and updating application settings."""
-
-    def __init__(self, data):
-        self.data = data
 
     def get(self, group, key):
         """
@@ -15,7 +14,7 @@ class ApplicationSettings:
         :param key: The key in that group
         :return: The value or None
         """
-        return self.data.get(group, {}).get(key)
+        return super().get(group, {}).get(key)
 
     def set(self, group, key, value):
         """
@@ -25,7 +24,7 @@ class ApplicationSettings:
         :param key: The key in that group
         :param value: The value to set
         """
-        self.data.setdefault(group, {})[key] = value
+        super().setdefault(group, {})[key] = value
 
     def clone(self):
-        return ApplicationSettings(deepcopy(self.data))
+        return ApplicationSettings(deepcopy(self))
--- a/lms/models/application_settings.py
+++ b/lms/models/application_settings.py
@@ -1,5 +1,3 @@
-from copy import deepcopy
-
 from sqlalchemy.ext.mutable import MutableDict
 
 
@@ -25,9 +23,6 @@ class ApplicationSettings(MutableDict):
         :param value: The value to set
         """
         super().setdefault(group, {})[key] = value
-
-    def clone(self):
-        return ApplicationSettings(deepcopy(self))
 
     def __repr__(self):
         return f"{self.__class__.__name__}({super().__repr__()})"
--- a/lms/services/course.py
+++ b/lms/services/course.py
@@ -28,8 +28,7 @@ class CourseService:
         return bool(
             self._db.query(Course)
             .filter(Course.consumer_key == self._consumer_key)
-            # pylint: disable=protected-access
-            .filter(Course._settings[group][key] == json.dumps(value))
+            .filter(Course.settings[group][key] == json.dumps(value))
             .limit(1)
             .count()
         )
@@ -52,7 +51,7 @@ class CourseService:
         course = Course(
             consumer_key=self._consumer_key,
             authority_provided_id=authority_provided_id,
-            _settings=course_settings.data,
+            settings=course_settings,
         )
 
         self._db.add(course)
--- a/lms/services/course.py
+++ b/lms/services/course.py
@@ -1,4 +1,5 @@
 import json
+from copy import deepcopy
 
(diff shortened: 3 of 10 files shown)


lms/models/application_settings.py
```diff
@@ -25,7 +24,13 @@ def set(self, group, key, value):
...
 
     def clone(self):
-        return ApplicationSettings(deepcopy(self.data))
+        return ApplicationSettings(deepcopy(self))
+
+    def __repr__(self):
+        return f"ApplicationSettings({super().__repr__()})"
+
+    def __str__(self):
+        return repr(self)
```
seanh, line 36: Added these to make `ApplicationSettings` objects easily distinguishable from normal dicts

lms/models/application_settings.py
```diff
@@ -25,7 +22,10 @@ def set(self, group, key, value):
...
         :param value: The value to set
         """
-        self.data.setdefault(group, {})[key] = value
+        super().setdefault(group, {})[key] = value
+
+    def __repr__(self):
+        return f"ApplicationSettings({super().__repr__()})"
 
-    def clone(self):
-        return ApplicationSettings(deepcopy(self.data))
```
seanh, line 31: The clone method is no longer needed, you can just use `copy.deepcopy` directly:

```python
settings_copy = copy.deepcopy(ai.settings)
```

`settings` will be a new `ApplicationSettings` object

lms/models/application_settings.py
```diff
@@ -25,7 +22,10 @@ def set(self, group, key, value):
...
+    def __str__(self):
+        return repr(self)
```
seanh, line 31: Added these to make `ApplicationSettings` objects easily distinguishable from normal dicts

lms/services/course.py
```diff
@@ -29,7 +30,7 @@ def any_with_setting(self, group, key, value=True):
             self._db.query(Course)
             .filter(Course.consumer_key == self._consumer_key)
             # pylint: disable=protected-access
-            .filter(Course._settings[group][key] == json.dumps(value))
+            .filter(Course.settings[group][key] == json.dumps(value))
```
seanh, line 33: This `Course.settings[group][key]` syntax is really weird and I wonder what happens if `group` is missing? But :man_shrugging:

tests/unit/lms/models/application_instance_test.py
```diff
@@ -35,23 +35,22 @@ def test_provisioning_is_not_nullable(self, db_session, application_instance):
...
+    def test_settings_defaults_to_None(self, application_instance):
         settings = application_instance.settings
 
-        assert isinstance(settings, ApplicationSettings)
-        assert settings.data == {}
+        assert settings is None
 
     def test_settings_can_be_retrieved(self, application_instance):
-        application_instance._settings = {"group": {"key": "value"}}
+        application_instance.settings = {"group": {"key": "value"}}
```
seanh, line 44: This actually sets `settings` to an `ApplicationSettings` object, not a dict, which is pretty handy

lms/models/application_instance.py
```diff
@@ -31,29 +30,14 @@ class ApplicationInstance(BASE):
...
-        `settings.set(group, key, value)`.
-
-        :rtype: models.ApplicationSettings
-        """
-
-        if self._settings is None:
-            self._settings = {}
-
-        return ApplicationSettings(self._settings)
-
```
jon-betts, line 56: :+1: Yep, this is much nicer.

lms/models/application_settings.py
```diff
@@ -1,21 +1,54 @@
...
+
+    application_settings.get("group", "key")
+    : will return None if *either* "group" or "key" is missing from the dict
+
+    application_settings.set("group", "key", "value")
+    : will create the sub-dict "group" for you if it's missing from the dict,
+      and will notify SQLAlchemy of your change so that it gets saved to the DB
+
+    If you mutate a nested dict or list in-place this change won't be detected
+    by SQLAlchemy and your change **will not be saved**.
```
jon-betts, line 20: You have been told

lms/models/application_settings.py
```diff
@@ -1,21 +1,54 @@
...
+    But alternatively you can call changed() after making your changes, then
+    they will be saved. In this example the changes to both "sections_enabled"
+    and "bar" will be saved:
+
+    >>> ai = db.query(models.ApplicationInstance).filter_by(...).one()
+    >>> ai.settings["canvas"]["sections_enabled"] = False
+    >>> ai.settings["foo"]["bar"] = "gar"
+    >>> # Notify SQLAlchemy that ai.settings has changed so it knows to save it.
+    >>> ai.settings.changed()
+
```
jon-betts, line 40:
```suggestion
```

lms/models/application_settings.py
```diff
@@ -25,7 +58,10 @@ def set(self, group, key, value):
         :param key: The key in that group
         :param value: The value to set
         """
-        self.data.setdefault(group, {})[key] = value
+        super().setdefault(group, {})[key] = value
+
+    def __repr__(self):
+        return f"ApplicationSettings({super().__repr__()})"
```
jon-betts, line 64: I like doing this:

```suggestion
        return f"{self.__class__.__name__}({super().__repr__()})"
```

As it behaves nicely when sub-classed.

lms/models/application_settings.py
```diff
@@ -1,21 +1,54 @@
...
+    This is a dict subclass, but you should use the custom get() and set()
+    methods below to get and set values because they're safer and more
+    convenient:
+
+    application_settings.get("group", "key")
+    : will return None if *either* "group" or "key" is missing from the dict
+
+    application_settings.set("group", "key", "value")
+    : will create the sub-dict "group" for you if it's missing from the dict,
+      and will notify SQLAlchemy of your change so that it gets saved to the DB
```
jon-betts, line 17: Some of this seems like it might be better as documentation for the specific methods below. Also cut down on the content here. Keep this to more of "to be safe use set() and get() detailed below".

seanh: Moved these into the methods

PR #1911, opened 2020-07-15
Convert `Dialog` props from prop-types to JSDoc and fix ref usage in `URLPicker`

robertknight:
This PR converts the prop documentation for `Dialog` from prop-types to JSDoc and makes some improvements to it in the process. In doing this I uncovered a mistake in `URLPicker` where the wrong function was used to create refs (`createRef` instead of `useRef`). This mistake went unnoticed because of the way the refs were used, but in general is not safe because `createRef` creates a new object every time that it is called, whereas `useRef` returns the same object on each render.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -13,7 +13,8 @@ import { useUniqueId } from '../utils/hooks';
  *
  * @typedef DialogProps
  * @prop {Object} children - The content of the dialog.
- * @prop {Object} [initialFocus] - A ref associated with the child element to focus when the dialog is rendered.
+ * @prop {import("preact/hooks").Ref<HTMLElement>} [initialFocus] -
+ *   Child element to focus when the dialog is rendered.
  * @prop {JSXElement[]} [buttons] -
  *   Additional `Button` elements to display at the bottom of the dialog.
  *   A "Cancel" button is added automatically if the `onCancel` prop is set.
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -58,7 +58,7 @@ export default function Dialog({
   const rootEl = useRef(/** @type {HTMLDivElement|null} */ (null));
 
   useElementShouldClose(rootEl, true, () => {
-    if (typeof onCancel === 'function') {
+    if (onCancel) {
       onCancel();
     }
   });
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -35,7 +55,7 @@ export default function Dialog({
   buttons,
 }) {
   const dialogTitleId = useUniqueId('dialog');
-  const rootEl = useRef();
+  const rootEl = useRef(/** @type {HTMLDivElement|null} */ (null));
 
   useElementShouldClose(rootEl, true, () => {
     if (typeof onCancel === 'function') {
```
LMS007, line 61: not. could be be simplified to `if(onCancel)` now?

seems like it would either be the right thing or undefined.

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -8,6 +8,24 @@ import useElementShouldClose from '../common/use-element-should-close';
...
 import { useUniqueId } from '../utils/hooks';
 
+/**
+ * @typedef {import("preact").JSX.Element} JSXElement
+ *
+ * @typedef DialogProps
+ * @prop {Object} children - The content of the dialog.
+ * @prop {import("preact/hooks").Ref<HTMLElement>} [initialFocus] -
+ *   Child element to focus when the dialog is rendered.
+ * @prop {JSXElement[]} [buttons] -
```
LMS007, line 18: Since we just generalize the `children` prop  as Object, are we going to concern ourselves with getting this exactly right? I'm all for it if we are, but I wonder if we should also do that for the `children`prop in components that use it?

also, does  this need to be JSX.Element|Element or does  JSX.Element capture both?

lms/static/scripts/frontend_apps/components/Dialog.js
robertknight: Yes, I think so. I can't see a reason to use the more verbose check.

lms/static/scripts/frontend_apps/components/Dialog.js
robertknight: For components that just insert `children` somewhere in their output and don't do anything else with it, I think `Object` is fine. Preact will report an error at runtime if `children` or some entry in `children` is something that can't be rendered.

Although I don't think there is a whole lot of value in being more specific here, a look through the React and Preact type definitions for mentions of `children` shows that they have some included types for this. React has [`ReactNode`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/650cddcab0d9b50cc30bfe526bbe5bf7f8e4c861/types/react/index.d.ts#L239) and Preact has [`ComponentChildren`](https://github.com/preactjs/preact/blob/5a9c453d78fd885fa3db40b21a5ec4b9a6f7da48/src/index.d.ts#L54). Note that the definition of these types includes null, undefined, primitives and arrays of any of those things.

PR #1912, opened 2020-07-15
Fix typecheck errors in `BasicLtiLaunchApp`

robertknight:
- Ignore return value of `reportSubmission` rather than returning it
   from `useEffect`. The return value of `useEffect` should be an effect
   cleanup function if provided.
 - Add missing JSDoc comments, in particular clarify what `errorState`
   is for
 - Suppress a couple of TS errors where we know things TS doesn't

Some of the changes here conflict with https://github.com/hypothesis/lms/pull/1844. I will take care of updating that PR as it is on my plate to review.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -50,12 +50,16 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     canvas,
   } = useContext(Config);
 
-  // The current state of the error.
-  // One of "error-fetch", "error-authorizing", or "error-report-submission" or null
-  const [errorState, setErrorState] = useState(null);
+  /**  @typedef {'error-fetch'|'error-authorizing'|'error-report-submission'} ErrorState */
 
-  // Any current error thrown.
-  const [error, setError] = useState(null);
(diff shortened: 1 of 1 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -185,7 +188,9 @@ export default function BasicLtiLaunchApp({ rpcServer }) {
     }
   }, [authToken, canvas.speedGrader, contentUrl]);
 
-  useEffect(reportSubmission, [reportSubmission]);
+  useEffect(() => {
+    reportSubmission();
+  }, [reportSubmission]);
```
robertknight, line 193: This change might look like a no-op, but it ensures that the return value of `reportSubmission` is not passed back `useEffect`, which was not intended.

LMS007: 👍

robertknight:
> Rob, this looks reasonable, but I'm still seeing an error on line 259 & 275 with <Dialog>... due to require props (or at least it thinks some props are required). I'm assuming that will be fixed up with Dialog changes coming later?

Yes, indeed. I should have clarified that I didn't fix all the outstanding errors in the files touched by this PR.

Issue #1874, opened 2020-07-02
Copy LMS backend checks from Jenkins to GitHub Actions


seanh:
The idea is that -- when this card is finished -- all of LMS's backend checks will be running on both GitHub Actions and Jenkins on every PR. This will let us test out GitHub Actions: how hard is it to get working? How fast is it? How reliable is it?

If GitHub Actions turns out to be unreliable and builds are failing when they shouldn't we can just ignore it and merge the PRs as long as the Jenkins builds passed.

If things go well then we can write another issue in the future to remove the checks from Jenkins.

Done when:

- [x] `make checkformatting` runs on GitHub Actions for every PR (I think this might actually check both backend and frontend formatting)
- [x] `make backend-lint` runs on GitHub Actions for every PR
- [x] `make backend-tests` runs on GitHub Actions for every PR
- [x] `make functests-only` runs on GitHub Actions for every PR
- [x] `make bddtests` runs on GitHub Actions for every PR

PR #1914, opened 2020-07-16
Copy all CI checks to GitHub Actions

seanh:
Fixes https://github.com/hypothesis/lms/issues/1874

This gets all the CI checks that we run on Jenkins (or that `make sure` runs) working on GitHub Actions as well: backend and frontend formatting, linting and unit, functional and BDD tests.

See the checks running on GitHub Actions here: https://github.com/hypothesis/lms/actions?query=branch%3Agha

This runs two GitHub Actions jobs named backend and frontend in parallel. Within the backend job, after installing everything, the backend formatting, linting, and unit, functional and BDD tests all run in parallel:

* In parallel:
  * backend:
    * In serial:
      - Start Postgres Docker container
      - Install Python 3.6.9
      - Update pip
      - Install tox
      - Checkout the lms git repo
      - Create the test databases
      - Install the frontend dependencies
      - Build the frontend assets
      - In parallel:
        - Run the backend code formatting
        - Run the backend linting
        - Run the backend unit tests
        - Run the backend functional tests
        - Run the backend BDD tests
  * frontend:
    * In serial:
      - Checkout the lms git repo
      - Install the frontend dependencies
      - Run the frontend code formatting
      - Run the frontend linting
      - Run the frontend tests

Here's a list of everything that's preinstalled in the Ubuntu VM that GHA runs our jobs in: https://github.com/actions/virtual-environments/blob/ubuntu18/20200709.0/images/linux/Ubuntu1804-README.md

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,66 @@
+name: Continuous integration
+on: [push]
+jobs:
+  backend:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11.5-alpine
+        ports:
+        - 5433:5432
+    env:
+      TOX_PARALLEL_NO_SPINNER: 1
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6.9
+    - name: Update pip
+      run: python -m pip install --upgrade pip
+    - name: Install tox
+      run: python -m pip install tox
+    - name: Checkout git repo
+      uses: actions/checkout@v2
+    - name: Create test databases
+      run: |
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_test'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_functests'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_bddtests'
+    - name: Cache the .tox dir
+      uses: actions/cache@v2
+      with:
+        path: .tox
+        key: ${{ runner.os }}-tox-${{ hashFiles('tox.ini') }}-${{ hashFiles('requirements*') }}-${{ hashFiles('setup.py') }}-${{ hashFiles('setup.cfg') }}
+        restore-keys: |
+          .tox-
+    - name: Cache the node_modules dir
+      uses: actions/cache@v2
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock') }}
+    - name: yarn install
+      run: yarn install --frozen-lockfile
+    - name: yarn build
+      run: yarn build
+    - name: Run tox
+      run: tox --parallel auto -e checkformatting,lint,tests,functests,bddtests
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Cache the node_modules dir
+      uses: actions/cache@v2
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-node_modules-${{ hashFiles('yarn.lock') }}
+    - name: Install
+      run: yarn install --frozen-lockfile
+    - name: Format
+      run: yarn checkformatting
+    - name: Lint
+      run: yarn lint
+    - name: Typecheck
+      run: yarn typecheck
+    - name: Test
+      run: gulp test
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@ tox_pip_extensions_ext_venv_update = true
 tox_pyenv_fallback = false
 
 [testenv]
+parallel_show_output = true
 skip_install = true
 passenv =
     HOME


.github/workflows/checks.yml
```diff
@@ -0,0 +1,46 @@
...
+        - 5433:5432
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6.9
+    - name: Update pip
+      run: python -m pip install --upgrade pip
+    - name: Install tox
+      run: python -m pip install tox
```
seanh, line 19: See: https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#running-tests-with-tox

Normally GitHub actions runs **jobs** (like the "backend" and "frontend" jobs in this workflow) in parallel but it cannot run steps within a job in parallel. And each job runs in its own isolated GHA "environment" (VM?) so if we were to make separate GHA jobs for formatting, linting, tests, functests and bddtests, then we'd have to install Postgres, Python, checkout the git repo, update pip, build the frontend assets, etc five times over (albeit in parallel but they might throttle us). Instead, the tox command below will use [tox's parallel mode](https://tox.readthedocs.io/en/latest/example/basic.html?highlight=parallel#parallel-mode) to run a bunch of stuff in parallel within a single GHA job.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Checkout git repo
+      uses: actions/checkout@v2
+    - name: Create test databases
+      run: |
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_test'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_functests'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_bddtests'
```
seanh, line 26: psql seems to be already installed

.github/workflows/checks.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Build frontend assets
+      run: |
+        yarn install
+        yarn build
```
seanh, line 30: yarn seems to be already installed

.github/workflows/checks.yml
```diff
@@ -0,0 +1,46 @@
...
+      uses: actions/checkout@v2
+    - name: Install
+      run: |
+        yarn install
+    - name: Format
+      run: yarn checkformatting
+    - name: Lint
+      run: yarn lint
+    - name: Test
+      run: gulp test
```
seanh, line 46: These three run in serial because I don't know of a frontend command to run them parallel like we can do with tox for the backend checks. It doesn't really matter: these three are all really quick, and the backend job always takes much longer than the frontend job.

@robertknight Could use your feedback on whether `yarn install` is really necessary here or can something smaller be done?

.github/workflows/checks.yml
seanh: @robertknight Could use your feedback on this: here I'm running `yarn install` and `yarn build` for the sake of the _backend_ functional tests, not the frontend stuff (these commands are being run in the backend VM not the frontend one). The backend functests crash with `assets.py` complaining about missing files if I don't do this. Is it really necessary to run a full `yarn` `install` and `build` here?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+        - 5433:5432
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6.9
+    - name: Update pip
+      run: python -m pip install --upgrade pip
+    - name: Install tox
+      run: python -m pip install tox
```
seanh, line 19: See: https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#running-tests-with-tox

Normally GitHub actions runs **jobs** (like the "backend" and "frontend" jobs in this workflow) in parallel but it cannot run steps within a job in parallel. And each job runs in its own isolated GHA "environment" (VM?) so if we were to make separate GHA jobs for formatting, linting, tests, functests and bddtests, then we'd have to install Postgres, Python, checkout the git repo, update pip, build the frontend assets, etc five times over (albeit in parallel but they might throttle us). Instead, the tox command below will use [tox's parallel mode](https://tox.readthedocs.io/en/latest/example/basic.html?highlight=parallel#parallel-mode) to run a bunch of stuff in parallel within a single GHA job. So you can install everything just once and then run the checks in parallel.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Checkout git repo
+      uses: actions/checkout@v2
+    - name: Create test databases
+      run: |
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_test'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_functests'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_bddtests'
```
seanh, line 26: psql seems to be already installed

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Build frontend assets
+      run: |
+        yarn install
+        yarn build
```
seanh, line 30: yarn seems to be already installed

seanh: @robertknight Could use your feedback on this: here I'm running `yarn install` and `yarn build` for the sake of the _backend_ functional tests, not the frontend stuff (these commands are being run in the backend VM not the frontend one). The backend functests crash with `assets.py` complaining about missing files if I don't do this. Is it really necessary to run a full `yarn` `install` and `build` here?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+      uses: actions/checkout@v2
+    - name: Install
+      run: |
+        yarn install
+    - name: Format
+      run: yarn checkformatting
+    - name: Lint
+      run: yarn lint
+    - name: Test
+      run: gulp test
```
seanh, line 46: These three run in serial because I don't know of a frontend command to run them parallel like we can do with tox for the backend checks. It doesn't really matter: these three are all really quick, and the backend job always takes much longer than the frontend job.

@robertknight Could use your feedback on whether `yarn install` is really necessary here or can something smaller be done?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_test'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_functests'
+        psql -U postgres -h localhost -p 5433 -c 'CREATE DATABASE lms_bddtests'
+    - name: Build frontend assets
+      run: |
+        yarn install
+        yarn build
+    - name: Run tox
+      run: tox --parallel all --parallel-live -e checkformatting,lint,tests,functests,bddtests
```
seanh, line 32: This is tox running five things in five different venvs in parallel

.github/workflows/continuous-integration.yml
robertknight: You could run a shell script that creates dummy files within the backend tests for all the files listed in `lms/assets.ini`. In other words, the same files will exist in the same locations as they would in the real app, but I'm guessing the `assets.py` tests don't care about the actual content.

.github/workflows/continuous-integration.yml
robertknight: `yarn install` is correct, although in CI you should use the `--frozen-lockfile` flag per https://stackoverflow.com/a/58525708. In frontend projects using npm, you should use `npm ci` rather than `npm install`.

In Travis builds we used to configure caching of the `node_modules/` folder. It may be possible to do the same for GHA.

seanh:
Note to self: I think there may be a way to improve the dependency caching.

See GHA's caching docs: https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

I think there's a problem with the current setup, which is that every PR shares the same cached `.tox` dir. So this can happen:

1. A PR is sent that changes the requirements (e.g. a Dependabot PR)
2. The PR updates the cached tox dir with the new requirements
3. The next time a commit is built on master or on any other PR branch it will get the cached tox dir from PR 1, which contains PR 1's requirements. `venv_update` will have to re-sync the tox dir to match the current commit's requirements

Basically what's happening is that whenever you have a PR or PRs open that change the requirements, they "fight" over the cached tox dir, each PR changing the tox dir to contain its own requirements each time it builds (and then writing that to the cache).

The thrashing will be minimal because `venv_update` computes the minimal necessary sequence of package installs, upgrades, downgrades and removes to mutate the venv in-place to match the current requirements. So nothing is reinstalling all of the requirements repeatedly. Just the changes.

But still.

I think you can fix this by using a hash of the requirements files (`tox.ini`, `requirements.txt`, etc) as the cache key. Something like this:

```yaml
key: .tox-${{ hashFiles("tox.ini") }}-${{ hashFiles("requirements*.txt") }}
restore-keys: |
  .tox-
```

Each build will use `key` as the cache key when writing to the cache, and `key` contains hashes of all the files that contain requirements.

When _reading_ from the cache a build first looks for a cached dir matching its `key` exactly. So if the `.tox` dir has ever been built from these exact requirements files before (whether by an earlier commit on this branch, or by a commit on master or any other branch) it will be found and restored, and `venv_update` won't have any work to do.

If there's no exact match for `key` then the build looks for cached dirs that match any of the `restore-keys`, using **pattern matching** not exact matches. The restore key `.tox-` will find all cached tox dirs made from any version of the requirements files. It then picks the most recent matching one to restore. This most recently cached tox dir may have been built from any version of the requirements, but most likely it's a version pretty similar to the current commit's requirements. `venv_update` will then compute the minimal mutations to make this tox dir match the commit's requirements, and the resulting tox dir will then be cached using `key` so that the same mutations won't need to be repeated in future.

seanh:
@robertknight Are there any files or directories we should cache that might speed up the `yarn install`, `yarn build`, or `gulp test` commands?

For example for Python we're going to cache the `.tox` dir that contains the venvs with the dependencies installed in them, and the `.pip/cache` dir that contains pip's downloaded and compiled Python packages.

Keep in mind that we can use hashes of files to trigger cache hits and misses. So for example we could use hashes of `package.json` and/or `yarn.lock` to decide whether cached directories should be re-used for the current commit. (See my comment above for how this might work for Python.)

robertknight:
> @robertknight Are there any files or directories we should cache that might speed up the yarn install, yarn build, or gulp test commands?

Caching `node_modules/` and using `yarn.lock` as a key will work I think. The build + test steps don't currently have any files that you can easily cache to speed things up, we would need to look into alternative JS build tooling to enable that.

robertknight:
I pushed a small change to the frontend steps to enforce that the Yarn lockfile is not changed during `yarn install` (see notes at https://classic.yarnpkg.com/en/docs/cli/install/) and also run the `yarn typecheck` step that we now run in Jenkins.

In the process it was necessary to rebase this PR on master.

Issue #1752, opened 2020-05-14
Finish refactoring of Canvas API client


jon-betts:
The client could do with some additional love to get it finished:

 * Split the parts of the client into separate files
 * Look for other splits in the parts (composition rather than inheritance etc)
 * Any additional madness reductions

jon-betts:
I've started tinkering with this, but nothing much to show yet.

jon-betts:
This is approved now, but we're going to wait for Monday in case it causes trouble.

PR #1923, opened 2020-07-17
Split up the Canvas API into multiple parts with composition

jon-betts:
For: https://github.com/hypothesis/lms/issues/1752

This is a PR MkII (from: https://github.com/hypothesis/lms/pull/1904) to try and break it up into reviewable chunks. It's big, but I'd like to try and review it in it's current state, as I don't think there's any sensible way to "slide" from the old code to this in a series of PRs without rewriting the same tests a lot.

## Feeback summary

A quick note of things to go over for me:

-  [x] Structural changes - file layout
- [x] Remove testing comment
- [x] Readme
- ~[ ]  ApplicationInstanceGetter - lms_host()~ I think this is overkill. It makes the factory pure plumbing... but why is that important?
- [x] TokenResponseSchema to bottom of file
- [x] New line indentation doc strings
- ~[ ] requests.Session()~ This makes patching a pain, Added a comment instead
- [x] _link_headers -> link_headers
- ~[ ]  TokenResponseSchema -> validate_quantity static?~ - This gets rid of the lint warning, but the validation doesn't work
- [x]  TokenResponseSchema - AuthenticatedClient testing is nonsense
- [x]  _api -> _client

## Review notes

I've tried to split this up into reviewable bits, but the whole thing doesn't work until they are all present, so you shouldn't expect to be able to stop at any particular part and have all the test pass etc. The tests in question should work though.

The commits are done in a green fields kind of way (it just adds new files) with the old files removed at the end. The original code is kept until the last commit so it can be compared with it's replacement. I don't think diffs were going to be readable given the changes. 

The commits include discrete parts and the tests for them (along with any test factories created):

1. Add TokenStore
    - OAuth2Token factory
    - ApplicationInstance factory
2. Basic client
    - requests.Response factory
3. Authenticated client
4. Public client
    - Added readme from long in file comments
5. Add service factory method
6. Removed original code

The tests for public client are more integration style than the rest. They only mock out the `requests.Session`. I started the other way, but the tests were just asserting the code was the shape it was, and not really that it all held together. I think this is a decent balance between separate underlying parts having decoupled tests, and the external part having good coverage. It also has the benefit of listing the HTTP calls we are trying to make (go to this URL) rather than the code calls (call this method with).

This means that the original tests look a bit weird as they export lots of their fixtures out to a `conftest.py`. Why this is happening doesn't make sense until the final batch of tests.

## About the major change of splitting up

 * Pulled all Pyramid specific stuff out of all of the classes and into a separate factory
     * I think we might want to think about doing this in general
     * It forces you to think about what you are passing around (e.g. 8 params! vs. `request` :shrug: )
     * This prompts thoughts about fault lines and good seams to split objects on
     * It makes for easier to follow code (you can see what needs what data, rather than just passing the god object around)
     * It makes for easier tests on the whole as initialisation is simpler
 * Removed "Canvas..." from the front of everything bar the public interface (this is obvious by namespace)
 * Make previously private methods public on objects (that themselves are private now)

## Other changes

Apart from the obvious changes which split the files up, some other opportunities were taken (once the code could be more easily considered):

 * Renamed some methods to have shorter snapper names
 * The old `make_request` pass that back to `send_and_validate` is now just `send`.
 * This means we have a direct chain of calls:
    * Public API method
    * `AuthenticatedClient.send()`
    * `BasicClient.send()`
    * `request.Session.send()`
  * The result of this is we are no longer passing prepared request objects all over the place
 * Some methods which were previously public are now private, as the objects have clearer boundaries


lms/services/canvas_api/token_store.py
```diff
@@ -0,0 +1,52 @@
...
+        oauth2_token.refresh_token = refresh_token
+        oauth2_token.expires_in = expires_in
+        oauth2_token.received_at = datetime.datetime.utcnow()
+
+    def get(self):
+        """
+        Return the user's saved access and refresh tokens from the DB.
+
+        :raise CanvasAPIAccessTokenError: if we don't have an access token for the user
+        """
```
jon-betts, line 41: This is as before pretty much, but I initially thought we could cache the token we save above as an optimisation. I'm pretty sure we can't however, as any other process could cause the token to be refreshed. So we are forced to get it every time.

tests/unit/lms/services/canvas_api/conftest.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from lms.services.canvas_api import TokenStore
+from tests import factories
+
```
jon-betts, line 7: Putting all of this out side of the test looks very strange, but they are re-used in later tests for the public client.

lms/services/canvas_api/basic.py
```diff
@@ -0,0 +1,118 @@
...
+        Make a request to the Canvas API and apply a schema to the response.
+
+        :param method: HTTP method to use
+        :param path: Path fragment to add to the end of the URL
+        :param schema: A schema object which this request must
+        :param params: URL parameters to include
+        :param headers: Headers to include
+        :param url_stub: Path prefix
+        :return: The result of applying the schema to the response
+        :raise CanvasAPIError: For any validation or request errors
```
jon-betts, line 61: Grew another param here `url_stub`, but making this public avoided a lot of weird pain in the `AuthenticatedClient`.

Previously `path` and `params` were passed to `self.get_url()` which automatically included a `url_stub` for the API. The issue is the token paths don't use this so:

 * `AuthenticatedClient` had to call `get_url` itself
 * Then construct a new Request object itself
 * Then pass that to `send_and_validate`

If we just add this one parameter none of that is necessary, and the whole thing is simpler.

tests/factories/requests_.py
```diff
@@ -0,0 +1,50 @@
...
+
+        kwargs["raw"] = BytesIO(raw.encode(encoding)) if isinstance(raw, str) else raw
+
+        # Requests seems to store these lower case and expects them that way
+        kwargs["headers"] = {key.lower(): value for key, value in headers.items()}
+
+        return kwargs
+
+    @classmethod
+    def _create(cls, model_class, *_args, **kwargs):
```
jon-betts, line 41: Factoryboy kind of sucks for objects which aren't initialised using their constructor.

```python
Object(value=blah)   # All good!

# vs.

obj = Object()    # Do it yourself
obj.value = blah 
```

I tried a whole bunch of ways around this, but couldn't find any. There are `params` which are not directly mapped to constructor `kwargs` but I can't see any way of accessing them other than to define attributes, which doesn't help. So here we are.

tests/factories/requests_.py
```diff
@@ -0,0 +1,50 @@
+import json
+from io import BytesIO
+
+import requests
+from factory import Factory, Faker
+
+
+class Response(Factory):
```
jon-betts, line 8: We create `request.Response` objects all over the place, and it's quite hard to do, as you aren't supposed to (only `requests`). So this is helpful here, but I think there's probably similar code elsewhere.

Unfortunately for us `factoryboy` doesn't make this easy, so the code is _way_ more long winded than an equivalent custom class would be.

tests/factories/requests_.py
```diff
@@ -0,0 +1,50 @@
...
+        "random_element", elements=[200, 201, 301, 304, 401, 404, 500, 501],
+    )
+    headers = None
+
+    @classmethod
+    def _adjust_kwargs(cls, **kwargs):
+        headers = kwargs["headers"] or {}
+        encoding = kwargs["encoding"]
+        raw = kwargs["raw"]
+        json_data = kwargs["json_data"]
```
jon-betts, line 25: Pick your poison here. If we push these into the function signature, then we change the function signature, which prompts a warning. If we do this, it makes it hideously long... not sure which is worse.

tests/unit/lms/services/canvas_api/basic_test.py
```diff
@@ -0,0 +1,177 @@
...
+from h_matchers import Any
+
+from lms.services import CanvasAPIError
+from lms.services.canvas_api import BasicClient
+from lms.validation import RequestsResponseSchema, ValidationError
+from tests import factories
+
+
+@pytest.mark.usefixtures("http_session")
+class TestBasicClient:
```
jon-betts, line 13: There's only one public method nowadays, so the tests here all try to corner one behavior of it

lms/services/canvas_api/authenticated.py
```diff
@@ -0,0 +1,140 @@
...
+        if refresh_token:
+            params["refresh_token"] = refresh_token
+
+        parsed_params = self._api.send(
+            "POST",
+            "login/oauth2/token",
+            url_stub="",
+            params=params,
+            schema=CanvasTokenResponseSchema,
+        )
```
jon-betts, line 132: This part is massively simplified by adding the `url_stub` as an argument. This is where all the nonsense was before.

tests/unit/lms/services/canvas_api/authenticated_test.py
```diff
@@ -0,0 +1,159 @@
...
+    @pytest.fixture
+    def token_response(self):
+        return {
+            "access_token": "new_access_token",
+            "refresh_token": "new_refresh_token",
+            "expires_in": "new_expires_in",
+        }
+
+    @pytest.fixture
+    def token_store(self, oauth_token):
```
jon-betts, line 150: Override the `token_store` provided in `conftest.py` because we want to fake it here.

tests/unit/lms/services/canvas_api/authenticated_test.py
```diff
@@ -0,0 +1,159 @@
...
+    ):
+        basic_client.send.side_effect = (
+            CanvasAPIAccessTokenError,  # The first call should fail
+            token_response,  # Then we expect a refresh call
+            "success",  # Then finally our successful content request
+        )
+
+        call_args = ("METHOD", "/path", sentinel.schema, sentinel.params)
+        result = authenticated_client.send(*call_args)
+
```
jon-betts, line 51: This isn't very nice because we trigger ourselves to make a call in the middle, but I tried to use the main params to tell the story of: old param, refresh, new param

lms/services/canvas_api/README.md
```diff
@@ -0,0 +1,49 @@
...
+different levels. They all sit inside the `public_client.CanvasAPIClient` in
+a nested fashion:
+
+ * [`public.CanvasAPIClient`](public.py) - High level API access
+ * [`authenticated.AuthenticatedClient`](authenticated.py) - OAuth2 token handling
+ * [`basic.BasicClient`](basic.py) - HTTP marshalling, requests and applying schema
+
+To see how these are all bolted together see: [`__init__.py`](__init__.py)
+
+## Getting authenticated users sections
```
jon-betts, line 15: This section was snipped out of the code

lms/services/canvas_api/public.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+
+    class Meta:
+        unknown = EXCLUDE
+
+    id = fields.Int(required=True)
+    name = fields.String(required=True)
+
+
+class CanvasAPIClient:
```
jon-betts, line 26: This is the most copy paste of the files, with the difference that we are composed with the underlying client, rather than being one. Little here changed.

tests/unit/lms/services/canvas_api/public_test.py
```diff
@@ -0,0 +1,238 @@
...
+        http_session.send.assert_called_once_with(
+            Any.request(
+                "GET",
+                url=Any.url.with_path("api/v1/courses/COURSE_ID").with_query(
+                    {"include[]": "sections"}
+                ),
+            ),
+            timeout=Any(),
+        )
+
```
jon-betts, line 56: The following two tests were previously in big all in one tests with parameterisation. I think this might be a case of "actually do repeat yourself".

The parameterised tests weren't much shorter in lines (in code they were), but they were quite hard to read. This is dirt simple and keeps most of the tests for this function in one place.

This same thing has been done for all the other ones. So there are some duplicated-ish tests. This might not look like a genius move if we have to alter them all one day.

lms/services/canvas_api/public.py
```diff
@@ -0,0 +1,236 @@
...
+    def authenticated_users_sections(self, course_id):
+        """
+        Return the authenticated user's sections for the given course_id.
+
+        :param course_id: the Canvas course_id of the course to look in
+        :return: a list of raw section dicts as received from the Canvas API
+        :rtype: list(dict)
+        """
+
+        return self._ensure_sections_unique(
```
jon-betts, line 65: I think this looks like it should be a decorator (`@unique_sections` or something) we can drop on this function. This however is very much a job for another day.

tests/unit/lms/services/canvas_api/public_test.py
```diff
@@ -0,0 +1,238 @@
...
+        assert response == "public_url_value"
+        http_session.send.assert_called_once_with(
+            Any.request(
+                "GET", url=Any.url.with_path("api/v1/files/FILE_ID/public_url")
+            ),
+            timeout=Any(),
+        )
+
+
+class TestMetaBehavior:
```
jon-betts, line 198: As a result of pulling out the big parameterised tests from before there's a lot less in here. These functions test behavior common to many functions.

tests/unit/lms/services/canvas_api/public_test.py
```diff
@@ -0,0 +1,238 @@
+"""
+Tests for the Canvas public API client.
+
+These tests take an integration style approach, rather than the mocked approach
+for the sub-components. The only thing mocked here is the requests session, the
+rest is the real deal.
+
+This is to prevent these tests from just becoming an MD5 of the code, and to
+actually show what calls we are supposed to be making in the end, rather than
+just what we feed into the system.
```
jon-betts, line 10: ^ this

tests/unit/lms/services/canvas_api/public_test.py
```diff
@@ -0,0 +1,238 @@
...
+import pytest
+from h_matchers import Any
+
+from lms.services import CanvasAPIAccessTokenError, CanvasAPIError, CanvasAPIServerError
+from lms.services.canvas_api import CanvasAPIClient
+
+
+class TestCanvasAPIClientGetToken:
+    # This is the only test where we fake out the underlying class, because
+    # this _one_ call is just a pass through.
```
jon-betts, line 23: Although hilariously the very first test, actually _does_ need to mock out the `AuthenticatedClient`...

lms/services/canvas_api/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def canvas_api_client_factory(_context, request):
+    """
+    Get a CanvasAPIClient from a pyramid request.
+
+    :param _context: Pyramid context object
+    :param request: Pyramid request object
+    :return: An instance of CanvasAPIClient
+    """
```
jon-betts, line 16: Not sure how I feel about this completely. It 100% decouples Pyramid from the classes underneath, which is good. But it's super ugly stuff.

I think it's probably better to keep the ugly in one place on balance. It's also good documentation for how the objects all fit together.

tests/unit/lms/services/canvas_api/__init___test.py
```diff
@@ -0,0 +1,61 @@
+from unittest.mock import sentinel
+
+import pytest
+
+from lms.services.canvas_api import canvas_api_client_factory
+
+
+@pytest.mark.usefixtures("ai_getter")
+class TestCanvasAPIClientFactory:
```
jon-betts, line 9: These tests are dumb. These really are the MD5 your code kind of tests. I don't know how to do this better in our style.

This would be easier if the attributes were public on the objects, then we could just instantiate one and interrogate that all the attributes. We could technically do that now with something like:

```python
assert result == Any.object(CanvasAPIClient).with_attrs({
    '_api': Any.object(AuthenticatedClient).with_attrs({
        ... 
    })

But it feels a bit dirty, and we don't need these attributes to be public for any other reason.
})
```

tests/factories/_attributes.py
```diff
@@ -3,9 +3,8 @@
 from factory import Faker
 
 OAUTH_CONSUMER_KEY = Faker("hexify", text="Hypothesis" + "^" * 32)
+SHARED_SECRET = Faker("hexify", text="shared-secret-" + "^" * 16)
 
 USER_ID = Faker("hexify", text="^" * 40)
-
 H_USERNAME = Faker("hexify", text="^" * 30)
-
 H_DISPLAY_NAME = Faker("name")
```
jon-betts, line 10: Grouping like things here (auth-y things, user-y things)

seanh:
I'm about half way through reviewing this but I think I'm going to have to leave it for a few days and come back to it.

For a start I've split out the `ApplicationInstance` and `OAuth2Token` test factories into separate PRs so they can be reviewed in smaller chunks:

https://github.com/hypothesis/lms/pull/1941
https://github.com/hypothesis/lms/pull/1942

@jon-betts If you can review those two then we can merge them and then this PR can be rebased and become (slightly) smaller

seanh:
I've sent another PR to tidy up the `Response` factory on this branch: https://github.com/hypothesis/lms/pull/1946

tests/factories/requests_.py
seanh: I've sent a PR to simplify this factory: https://github.com/hypothesis/lms/pull/1946

tests/factories/requests_.py
seanh: Yeah, I'm not 100% but AFAIK the `_create()` method you have here is the right thing to do / is what you have to do.

Factory Boy's default `_create()` method passes all the generated values to the model's `__init__()` as kwargs. You can use `inline_args` to get it to pass some of them as pargs instead of kwargs if necessary. But you can't tell it to set some of them using `setattr` post-creation.

I don't think Factory Boy sucks here though. It provides a default `_create()` method that works for most objects, and if you need to create your object in a different way you override `_create()` with your own :man_shrugging: If you weren't using Factory Boy you would still need to create a `Response` and then set attrs on it so you would need to write the equivalent of this `_create()` code anyway.

tests/factories/requests_.py
seanh: https://github.com/hypothesis/lms/pull/1946/ gets rid of this

tests/unit/lms/services/canvas_api/__init___test.py
seanh: I don't think this is a problem with our test style. I think the problem is that `__init__.py` doesn't actually contain anything very interesting to unit test. It's integration code, and that kind of code just isn't very interesting to test. You want your tests to run the code to make sure it isn't completely broken. But beyond that there's not much of value to test. All the `assert_called_once_with()`'s are indeed kind of pointless.

I don't think there's much difference between with-mocks vs without-mocks here. A without-mocks version of this test doesn't look any more meaningful than the with-mocks version to me:

```python
assert isinstance(result, CanvasAPIClient)
assert isinstance(result._api, AuthenticatedClient)
```

As you say you'd have to either test the private `_api` attr or make `_api` public just for the test's sake, both of which feel icky. And even then, is this really any more meaningful than:

```python
CanvasAPIClient.assert_called_once_with(AuthenticatedClient.return_value)
assert result == CanvasAPIClient.return_value
```

Both seem equally meaningless to me. Either way you're just testing that it passed the `AuthenticatedClient` object to `CanvasAPIClient()` and returned the result. Both feel like MD5ing your code.

Here's a couple of options:

### With mocks

This is what I would do. I would spend too much time writing or thinking about it, because there's not much to test here. The test isn't actually testing much, it's just exercising the code so that we know it's not completely broken. I would keep the mocks for reasons explained in **Without mocks** below. The `BasicClient` assertion seems worth keeping since that's testing one thing the method actually does do (with `netloc`):

```python
from unittest.mock import sentinel

import pytest

from lms.services.canvas_api import canvas_api_client_factory


class TestCanvasAPIClientFactory:
    def test_it(self, ai_getter, pyramid_request, BasicClient, CanvasAPIClient):
        ai_getter.lms_url.return_value = "https://example.com/path"

        canvas_api = canvas_api_client_factory(sentinel.context, pyramid_request)

        BasicClient.assert_called_once_with("example.com")
        assert canvas_api == CanvasAPIClient.return_value


pytestmark = pytest.mark.usefixtures("ai_getter")


@pytest.fixture(autouse=True)
def BasicClient(patch):
    return patch("lms.services.canvas_api.BasicClient")


@pytest.fixture(autouse=True)
def TokenStore(patch):
    return patch("lms.services.canvas_api.TokenStore")


@pytest.fixture(autouse=True)
def AuthenticatedClient(patch):
    return patch("lms.services.canvas_api.AuthenticatedClient")


@pytest.fixture(autouse=True)
def CanvasAPIClient(patch):
    return patch("lms.services.canvas_api.CanvasAPIClient")
```

This test:

* Will fail if `canvas_api_client_factory` looks up the `ai_getter` service incorrectly
* Will fail if it doesn't do the `netloc` stuff correctly and pass it to `BasicClient`
* Will fail if it passes the wrong number of arguments, or an unknown keyword argument, to `BasicClient`, `TokenStore`, `AuthenticatedClient` or `CanvasAPIClient`
* Will fail if it doesn't return the `CanvasAPIClient`

It _won't_ fail if all the arguments it passes to everything are all completely wrong but still happen to have the right number of arguments and keyword-argument names for each method. But as you said, testing that really just feels like MD5ing your code and doesn't seem to add much value.

You _could_ add those asserts back in. I don't think they're doing much harm. I would just add them all into the one test, rather than a separate test for each assert. But as you say they're not adding much value

```python
    def test_it(
        self,
        ai_getter,
        pyramid_request,
        BasicClient,
        TokenStore,
        AuthenticatedClient,
        CanvasAPIClient,
    ):
        ai_getter.lms_url.return_value = "https://example.com/path"

        canvas_api = canvas_api_client_factory(sentinel.context, pyramid_request)

        BasicClient.assert_called_once_with("example.com")
        TokenStore.assert_called_once_with(
            user_id=pyramid_request.lti_user.user_id,
            db=pyramid_request.db,
        )
        AuthenticatedClient.assert_called_once_with(
            basic_client=BasicClient.return_value,
            token_store=TokenStore.return_value,
            client_id=ai_getter.developer_key(),
            client_secret=ai_getter.developer_secret(),
            redirect_uri=pyramid_request.route_url("canvas_oauth_callback"),
        )
        CanvasAPIClient.assert_called_once_with(AuthenticatedClient.return_value)
        assert canvas_api == CanvasAPIClient.return_value
```

### Without mocks

If I were writing this test and I disliked mocks I would probably do this:

```python
from unittest.mock import sentinel

import pytest

from lms.services.canvas_api import canvas_api_client_factory, CanvasAPIClient
from lms.services.canvas_api.public import CanvasAPIClient


class TestCanvasAPIClientFactory:
    def test_it(self, pyramid_request):
        canvas_api = canvas_api_client_factory(sentinel.context, pyramid_request)

        assert isinstance(canvas_api, CanvasAPIClient)


pytestmark = pytest.mark.usefixtures("ai_getter")
```

It's quite hard to compare with with-mocks and without-mocks versions given such a simple test, but:

* I don't think the without-mocks version has a good way to test that the factory does the right thing with `netloc` and passes it to `BasicClient`
* They both seem to run in the same amount of time
* Without-mocks is easier to write: you don't need the patch fixtures. (With-mocks is often easier to write because constructing real objects and getting them to behave as you want is often hard, whereas with mocks that's easy. But in this test, that doesn't apply.)
* The without-mocks version is less isolated. The test has more reasons to fail. If there are bugs in the `__init__()` methods of `BasicClient`, `TokenStore`, `AuthenticatedClient` or `CanvasAPIClient` or anything they call, they could cause this test to fail, which would just add noise and distraction to the test output, make debugging harder. We only want this test to fail if there's a bug in `canvas_api_client_factory()`. Bugs in the other classes should only cause the other classes' unit tests to fail.
* The without-mocks version messes up coverage reporting, making it look like those `__init__()` methods and anything they call have unit test coverage when they don't really. This harms coverage's ability to reveal bugs or test mistakes for us

### A more meaningful integration test?

You could try to write a test constructs a `CanvasAPIClient` without mocking its dependencies and then actually make one API call with that client and test (using a mocked out `requests` lib or whatever) that it actually made the expected HTTP call and returned the response.

I think you'd want to have all the details already covered with unit tests for the different components such that the integration test really is just finally testing that it all fits together just so we can feel safer about that, and no more than one or two integration tests are needed. And I think we'd want to exclude the integration test from counting towards test coverage.

lms/services/canvas_api/__init__.py
seanh: Yeah I think it's probably good to do it this way

tests/unit/lms/services/canvas_api/authenticated_test.py
```diff
@@ -0,0 +1,159 @@
...
+from h_matchers import Any
+
+from lms.services import CanvasAPIAccessTokenError
+from lms.services.canvas_api import BasicClient, TokenStore
+from lms.services.canvas_api.authenticated import TokenResponseSchema
+from lms.validation import ValidationError
+from tests import factories
+
+
+class TestCanvasTokenResponseSchema:
```
seanh, line 13:
```suggestion
class TestTokenResponseSchema:
```

tests/unit/lms/services/canvas_api/authenticated_test.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+        basic_client.send.assert_called_once_with(
+            "METHOD",
+            "/path",
+            sentinel.schema,
+            sentinel.params,
+            headers={"Authorization": f"Bearer {oauth_token.access_token}"},
+        )
+
```
seanh, line 37:
```suggestion
```

tests/unit/lms/services/canvas_api/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+    session = session()
+
+    def set_response(json_data=None, raw=None, status_code=200):
+        session.send.return_value = factories.requests.Response(
+            json_data=json_data, raw=raw, status_code=status_code
+        )
+
+    session.set_response = set_response
+
+    return session
```
seanh, line 35: You decided to mock requests instead of using httpretty?

lms/services/canvas_api/basic.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    PAGINATION_PER_PAGE = 1000
+    """The number of items to request at one time.
+
+     This only applies for calls which return more than one, and is subject
+     to a secret internal limit applied by Canvas (might be 100?)."""
+
+    PAGINATION_MAXIMUM_REQUESTS = 25
+    """The maximum number of calls to make before giving up."""
```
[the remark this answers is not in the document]
jon-betts: This is more about calling out configuration to the top of the class and making the code self document. By doing this changes to these values become very obviously intentional and it's easier to see what a particular piece of code is doing. Consider the meaning of the following two lines:

```python
if request_depth < 25:
# vs.
request_depth < self.PAGINATION_MAXIMUM_REQUESTS:
```

The intent of the second is much clearer.

tests/unit/lms/services/canvas_api/authenticated_test.py
jon-betts: The scars of refactoring... good catch.

tests/unit/lms/services/canvas_api/conftest.py
jon-betts: Not in this PR, but yep:

 * Mostly because it's what was already there
 * Lets you follow our standard stuff with mock objects and assertions: e.g.
    * Raising errors
    * Making assertions about arguments passed to send, but don't necessarily make it to a response (like timeout, stream etc.)

lms/services/canvas_api/authenticated.py
```diff
@@ -0,0 +1,140 @@
+"""Access to the authenticated parts of the Canvas API."""
+
+import marshmallow
+from marshmallow import fields
+
+from lms.services import CanvasAPIAccessTokenError
+from lms.validation import RequestsResponseSchema
+
+
+class TokenResponseSchema(RequestsResponseSchema):
```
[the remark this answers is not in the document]
jon-betts: I think moving the file to the bottom is good.

The new S/O question about close collaborators and your suggestions for marking the files private help here too: 

* Everything in this directory is a close collaborator (the mutuality definition), so we don't need to get into the weeds about _how_ close they all are inside the bounded context, or worry about who is private to who within.
* By following your suggestion and marking the file as private, everything in it is private too. The boundary that matters (inside or out of the context) has been nicely established by that, and more clarification isn't required.
* Part of the definition is that all items within the boundary could happily live in one file if they were small enough, or many files as convenience dictates. So we are freed from having to read much into the location of the classes within the context. Wherever makes sense and fits is fine. In this case it makes sense to keep it close.
* Choosing to "zoom" in temporarily on one item makes it the unit under test, but does not change the relationship between the code. So we are freed from having to read into this zooming behavior.

In this case the extra tests weren't because I thought it was too big to test, just that the tests would be easier and shorter if I did it this way (I could have gone from the outside, but it's annoying). In this case I was just trying to catch a little coverage shortfall.

----
### An interesting distinction between our "close collaborators" and a domain model

I think this does bring up a reasonable point though, which helps clarify the collaborator rules.

It's pretty clear that the `AuthenticatedClient` concept creates it's own sub-domain which triggers some of these conclusions. You can, and I did choose to "zoom" into the `AuthenticatedClient` as a unit of test, and then into the `TokenResponseSchema` after that. In a pure domain driven model, each of these could be considered nested sub-domains, and our code layout reflects that. So:

 * This is perfectly reasonable definition of this domain `(*OtherStuff, (AuthenticatedClient, (TokenResponseSchema)))`
 * But we are defining our collaborators as `(*OtherStuff, AuthenticatedClient, TokenResponseSchema)`

So how do we get from the first to the second using our rules?

Part of our definition of a close collaborators is that aren't useful independently. The `AuthenticatedClient` doesn't work without the `TokenResponseSchema` and in the reverse it's useless. This gets us to:

 `(*OtherStuff, (AuthenticatedClient, TokenResponseSchema))`

The second step is a bit more tricky. There are plenty of other indicators that work, but I'd like to talk about:

> `Style`: Being a close collaborator is a mutual arrangement between a set of code. 

The crucial part here is that this is a `Style` rather than a `Guide`. We are to some extent stating this to be true when we use these words, rather than detecting this situation (don't be silly etc.).

So to go whole hog here we could have defined our domain model like this:

 * `(factory, (CanvasAPIClient, (AuthenticatedClient, BasicClient, TokenStore, (TokenResponseSchema))))`

But that isn't our definition of close collaborators, that's just our class layout / call stack. So it's an accurate view of the code, but it's not the close collaborator concept we're discussing.

* `(factory, CanvasAPIClient, AuthenticatedClient, BasicClient, TokenStore, TokenResponseSchema)`

I think you're file layout is on the money too, which is to say that both the factory and the client form the interface together.

----

### What a lot of waffle: what's your point?

I think drawing an intentional circle round the code and saying "inside, outside, doorway" and not getting into any more detail frees us from some of the discussions and distinctions here. This is perhaps the feature of doing this.

If you have a bounded context containing (by definition) mutually close collaborators you are always free to treat the code exactly as if it wasn't in a bounded context:

 * Mock in all the ways you would
 * Layout your code in all the ways you would
 * Separate and unit test as you would

We draw this boundary to give us additional freedoms:

 * To move / compose and decompose items within at will
 * To put things in convenient places
 * To not mock if we don't want to

And to do all of the above without being "bad" or creating a mess

lms/services/canvas_api/authenticated.py
```diff
@@ -0,0 +1,140 @@
...
+
+    access_token = fields.Str(required=True)
+    refresh_token = fields.Str()
+    expires_in = fields.Integer()
+
+    @marshmallow.validates("expires_in")
+    def validate_quantity(self, expires_in):  # pylint:disable=no-self-use
+        if expires_in <= 0:  # pragma: no cover
+            # I have no idea what's going on with coverage. This absolutely is
+            # covered at time of writing. But it doesn't show up in coverage
```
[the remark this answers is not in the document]
jon-betts: This makes sense. The tests added for this specifically are intended to cover the unhappy path, but my assumption was that the happy path was covered by everything else being not dead.

Somewhat unfortunately, this is _literally the only end-point_ which isn't tested top to bottom like all the end-points in the main client. The upshot here, is a mocking gap. We test the schema, we test the schema is passed to basic client, and we mock return values from basic client, but at no point are all of these hooked together to see if they work.

The upshot is the `token_response` fixture is both never run through the schema, and total nonsense (has a string for the `expires_in`).

I think what I might do is:

 * Stop mocking the basic client and just test like we do in the main client
 * Have a look at testing the `TestCanvasTokenResponseSchema` as a part of this which also resolves some other concerns about it

lms/services/canvas_api/authenticated.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Create an AuthenticatedClient object for making authenticated calls.
+
+        :param basic_client: An instance of BasicClient
+        :param token_store: An instance of TokenStore
+        :param client_id: The OAuth2 client id
+        :param client_secret: The OAuth2 client secret
+        :param redirect_uri: The OAuth 2 redirect URI
+        """
+        self._api = basic_client
```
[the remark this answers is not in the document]
jon-betts: This is to make the layers more similar to each other. The `CanvasAPIClient` refers interally to the `AuthenticatedClient` as `_api`, as the `AuthenticatedClient` refers to the `BasicClient`.

`_authenticated_client` is super long winded, so I wonder if we can split the difference and go for '_client' which is both suitable vague and more accurate.

jon-betts:
A quick note of things to go over for me:

- Structural changes - file layout
- Remove testing comment
- Readme
- ApplicationInstanceGetter - lms_host()
- TokenResponseSchema
- New line indentation doc strings
- requests.Session()
- _link_headers -> link_headers
- TokenResponseSchema -> validate_quantity static?
- TokenResponseSchema - AuthenticatedClient testing is nonsense
- _api -> _client
- request.Session()

lms/services/canvas_api/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+There are a number of different API clients which deal with the API at varying
+different levels. They all sit inside the `public_client.CanvasAPIClient` in
+a nested fashion:
+
+ * [`public.CanvasAPIClient`](public.py) - High level API access
+ * [`authenticated.AuthenticatedClient`](authenticated.py) - OAuth2 token handling
+ * [`basic.BasicClient`](basic.py) - HTTP marshalling, requests and applying schema
+
+To see how these are all bolted together see: [`__init__.py`](__init__.py)
```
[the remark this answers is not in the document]
jon-betts: I think this is a good layout and both works with how we've described this stuff before, and also goes to shore up a few fuzzy ideas I was having.

The first is that in one sense `canvas_api_client_factory()` is the interface to this code, except... not at all. When this is working correctly, nobody at all interacts with that function. Everybody who is talking to this code is interacting with `CanvasAPIClient`.

Another mistake is "technically" the interface is defined in `__init__.py`. So it's true to say that we could not have underscores and you "are informed" of the public and private interface. But we don't have to make it hard. What's exptected to be exported in the `__init__.py` should probably reflect what's not underscored in the directoy.

By having the factory in a separate method, we also get around having to use `__all__` to make the distinction between things we imported into `__init__.py` because they are needed locally, and things which are intended for export. If it's in `__init__.py` it's because it's part of the public interface (import loops permitting).

This also allows us to work out what we should do in a few other cases. Lets say we had something like:

```python
__init__.py
   from factory import factory_method()
   from client import Client Object
factory.py
    factory_method()
client.py
    ClientObject()
    ClientObjectInFileCollaborator()
_client_sub_object.py
    ClientSubObject()
    _ClientSubObjectInFileCollaborator()
_client_return_model.py
   ReturnModel()
```

There a number of mistakes here:

 * `ClientObjectInFileCollaborator` should be private, as it's not part of the interface, and it is in a file marked as part of the interface
 * `_ClientSubObjectInFileCollaborator` should not be private, as it's already marked private by being in a private marked file. 
* Assuming `ReturnModel` is something sent back when interacting with `ClientObject` then it forms part of the surface area you interact with when using the code. This makes it part of the public interface too, even if it's at the bottom of the call stack. So `_client_return_model.py` should be public too. 
 * This also means `__init__.py` should import `ReturnModel` to further advertise it's public-ness and cut of outsiders from the exact location of things within the context (allowing refactoring)

jon-betts:
A summary of the changes are at the top. I've tried to address most of the suggestions. The changes are quite disruptive, as it involved a big reshuffle of the file layout, but the only areas that actually had significant changes are:

 * The README was inlined as a comment. I didn't like the docstring before, as it was so big you couldn't see the signature of the method and the content at the same time. It's above now
 * The AuthenticatedClient tests have been rewritten a bit to test the schema in context. We were missing this part before and it was total mince as a result

There were a few things I could get to work / didn't think were a great idea:

 * Making the schema method a static / class method doesn't seem to work. It suppresses the lint warning, but removes the validation
 * Using `requests.Session` in place of `Session` made mocking a pain. I added a comment to ward off any confusion instead
 * I didn't add `ApplicationInstance.lms_host()` - This would make our factory a "pure plumbing" factory... but why? Aside from scratching an itch, I'm not sure how that helps. It's a bunch of code and tests I think we can live without.

I _think_ I got everything else?

PR #1929, opened 2020-07-20
Convert remaining UI component docs to JSDoc

robertknight:
This surfaced an issue where several properties were missing from the
`StudentInfo` type listed in `frontend_apps/config.js`. These properties are
needed when calling the API to submit a grade update.

It also surfaced an issue where the `SubmitGradeForm` component redundantly took
a nullable `student` prop as well as a `disabled` prop.  Whether to disable the
form can be inferred from whether `student` is null or not (and _should_ be to
avoid a possible inconsistency).

### DIFF
--- a/lms/static/scripts/frontend_apps/components/LMSGrader.js
+++ b/lms/static/scripts/frontend_apps/components/LMSGrader.js
@@ -6,10 +6,7 @@ import StudentSelector from './StudentSelector';
 import SubmitGradeForm from './SubmitGradeForm';
 
 /**
- * @typedef User
- * @prop {string} displayName
- * @prop {string} userid
- *
+ * @typedef {import('../config').StudentInfo} StudentInfo
  * @typedef {import('../services/client-rpc').ClientRpc} ClientRpc
  */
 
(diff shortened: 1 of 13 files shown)


lms/static/scripts/frontend_apps/components/LMSGrader.js
```diff
@@ -125,10 +122,7 @@ export default function LMSGrader({
             />
           </li>
           <li className="LMSGrader__student-grade">
-            <SubmitGradeForm
-              student={getCurrentStudent()}
-              disabled={currentStudentIndex < 0}
-            />
+            <SubmitGradeForm student={getCurrentStudent()} />
```
LMS007, line 125: Awe, okay

we may need to change the state variable (line 36) from 

`const [currentStudentIndex, setCurrentStudentIndex] = useState(-1);`

to 

`const [currentStudentIndex, setCurrentStudentIndex] = useState(null);`

The negative index may be a bit misleading now since it has no purpose.

robertknight: I tried making this change but it turns out that several other places in this component and also `StudentSelector` make use of the fact that a negative `currentStudentIndex` is used to represent "no student selected". For the moment I have simply documented this fact in the `StudentSelector` props.

Using `null` would have the advantage that TS would be able to offer more help in terms of ensuring that you checked for it before indexing into the students list.

PR #1937, opened 2020-07-21
Clean up the functional tests

seanh:
Clean out some junk from our functional tests:

* Get rid of JSON fixtures.

  Let's just use pytest's fixtures system and keep our fixtures in
  Python code, like we do everywhere else. This is for consistency, but
  it's also much simpler, and it's going to allow pytest fixture
  composition (where one fixture uses values created by another fixture)
  to be used as normal in a future PR.

  The base class for all tests classes to inherit from with the
  `json_fixture()` method that streams JSON fixtures from disk has been
  deleted. The two JSON fixtures that were in use have each been inlined
  into the Python code in the one place that each was used.

  It's worth remembering that the JSON format was (indirectly) based on
  Python's list and dict literals format. The two formats are almost
  exactly the same. You can copy some JSON data and paste it into Python
  code with almost no modification. So there's really little point in
  putting data in separate JSON files and then writing code to read them
  from disk and parse them. You can just put the data in Python files.

  If you don't want test data cluttering up the test code, put the
  fixtures in a separate Python file. In this (and most) cases though,
  maintenance is easier if the fixtures live in the same file with the
  tests that use them.

  Putting complex fixtures in separate JSON files that're shared by all test modules also encourages coupling tests together. If lots of test modules all use the same big chunk of fixture data it can become impossible to modify that fixture without breaking tests all over the place.

* Delete a bunch of unused files.

  There was a `tests/functional/lti_certification/` dir with nothing in
  it.


tests/functional/test_sanity.py
```diff
@@ -10,9 +7,7 @@ def test_sanity(self, app):
     def test_we_can_read_and_send_params(self, app):
         # We aren't really running this test because we care, but just
         # to prove the testing framework
-        result = app.get(
-            "/flags/test", params=self.json_fixture("sanity/bar_on.json"), status=200
-        )
+        result = app.get("/flags/test", params={"feature_flags.bar": True}, status=200)
```
seanh, line 10: `params={"feature_flags.bar": True}` is a lot clearer than `params=self.json_fixture("sanity/bar_on.json")`, which required you to understand what `self.json_fixture()` was (you wouldn't find it in the file: it was defined in a base class in another file that the test class inherited from), find the `tests/functional/fixtures/sanity/bar_on.json` file, and open it to find `{"feature_flags.bar": true}`.

tests/functional/api/basic_lti_launch.py
```diff
@@ -99,7 +98,41 @@ def module_item_configuration(self, request, db_session):
...
+            "resource_link_title": "Link 1234",
+            "resourcelinkid": "rli-1234",
+            "roles": "Instructor",
+            "tool_consumer_info_product_family_code": "imsglc",
+            "tool_consumer_info_version": "1.1",
+            "tool_consumer_instance_description": "IMS Testing Description",
+            "tool_consumer_instance_guid": "IMS Testing",
+            "tool_consumer_instance_name": "IMS Testing Instance",
+            "user_id": "123456",
+        }
```
seanh, line 135: Having the params directly in the Python file with the code that uses them is a lot better. For example on another branch I was debugging a test that was crashing because it was still trying to use an old value for one of these params. I had changed the value of the param, and updated it everywhere I could find it mentioned in the source code. Eventually I had to track down this unrelated-looking `lti_params()` fixture, the `json_fixture()` method (defined in a base class in another file) and eventually the `good_params.json` file, to find the outdated param value. I still didn't find the value that I wanted because the params from the JSON file also get later post-processed by some Python code later on

PR #1941, opened 2020-07-21
Add ApplicationInstance test factory

seanh:
Extracted from https://github.com/hypothesis/lms/pull/1923/ but with some changes:

* Added `FACTORY_CLASS=SQLAlchemyModelFactory` to the factory so that Factory Boy adds instances to the DB session automatically.

* I've updated all the existing tests to use the new factory

* Also used the `ApplicationInstance` as a sub-factory in the `OAuth2Token` factory, separate PR: https://github.com/hypothesis/lms/pull/1942

* Also used the `ApplicationInstance` as a sub-factory in the `Course` factory, separate PR: https://github.com/hypothesis/lms/pull/1944

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -45,6 +45,9 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    #: A list of all the courses for this application instance.
+    courses = sa.orm.relationship("Course", back_populates="application_instance")
+
     #: A list of all the GroupInfo's for this application instance.
     group_infos = sa.orm.relationship(
         "GroupInfo", back_populates="application_instance"
--- a/lms/models/course.py
+++ b/lms/models/course.py
@@ -18,6 +18,11 @@ class Course(BASE):
         primary_key=True,
     )
 
+    #: The ApplicationInstance that this access token belongs to.
+    application_instance = sa.orm.relationship(
+        "ApplicationInstance", back_populates="courses"
+    )
+
     #: The authority_provided_id that uniquely identifies the course that these
     #: settings belong to.
     authority_provided_id = sa.Column(sa.UnicodeText(), primary_key=True)
--- a/tests/factories/__init__.py
+++ b/tests/factories/__init__.py
@@ -2,6 +2,7 @@ import sys
 
 from factory.alchemy import SQLAlchemyModelFactory
 
+from tests.factories.application_instance import ApplicationInstance
 from tests.factories.course import Course
 from tests.factories.grading_info import GradingInfo
 from tests.factories.h_group import HGroup
@@ -9,7 +10,7 @@ from tests.factories.h_user import HUser
 from tests.factories.lti_user import LTIUser
 
 
-def set_sqlalchemy_session(session):
+def set_sqlalchemy_session(session, persistence=None):
     # Set the Meta.sqlalchemy_session option on all our SQLAlchemy test factory
     # classes. We can't do it in the normal Factory Boy way:
     #
@@ -26,6 +27,9 @@ def set_sqlalchemy_session(session):
     for factory_class in _sqlalchemy_factory_classes():
         # pylint:disable=protected-access
         factory_class._meta.sqlalchemy_session = session
+
+        if persistence:
+            factory_class._meta.sqlalchemy_session_persistence = persistence
 
 
 def clear_sqlalchemy_session():
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -5,6 +5,7 @@ from webtest import TestApp
 
 from lms import db
 from lms.app import create_app
+from tests import factories
 from tests.conftest import SESSION, TEST_SETTINGS, get_test_database_url
 
 TEST_SETTINGS["sqlalchemy.url"] = get_test_database_url(
@@ -35,13 +36,17 @@ def app(pyramid_app, db_engine):
     return TestApp(pyramid_app)
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 def db_session(db_engine):
     """Get a standalone database session for preparing database state."""
 
     conn = db_engine.connect()
     session = SESSION(bind=conn)
 
-    yield session
+    factories.set_sqlalchemy_session(session, persistence="commit")
 
-    session.close()
+    try:
+        yield session
+    finally:
+        factories.clear_sqlalchemy_session()
+        session.close()
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -36,7 +36,7 @@ def app(pyramid_app, db_engine):
     return TestApp(pyramid_app)
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture
 def db_session(db_engine):
     """Get a standalone database session for preparing database state."""
 
--- a/tests/unit/lms/models/application_instance_test.py
+++ b/tests/unit/lms/models/application_instance_test.py
@@ -2,34 +2,32 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 from lms.models import ApplicationInstance, ApplicationSettings
+from tests import factories
 
 # pylint: disable=protected-access
 
 
 class TestApplicationInstance:
-    def test_it_persists_application_instance(self, db_session, application_instance):
+    def test_it_persists_application_instance(self, db_session):
         initial_count = db_session.query(ApplicationInstance).count()
-        db_session.add(application_instance)
+
+        factories.ApplicationInstance()
+
         new_count = db_session.query(ApplicationInstance).count()
         assert new_count == initial_count + 1
 
-    def test_provisioning_defaults_to_True(self, db_session, application_instance):
-        db_session.add(application_instance)
-
+    def test_provisioning_defaults_to_True(self, application_instance, db_session):
         db_session.flush()
 
         assert application_instance.provisioning is True
 
-    def test_provisioning_can_be_disabled(self, db_session, application_instance):
+    def test_provisioning_can_be_disabled(self, application_instance):
         application_instance.provisioning = False
-        db_session.add(application_instance)
-        db_session.flush()
 
         assert not application_instance.provisioning
 
     def test_provisioning_is_not_nullable(self, db_session, application_instance):
         application_instance.provisioning = None
-        db_session.add(application_instance)
 
         db_session.flush()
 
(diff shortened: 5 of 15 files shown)


lms/models/application_instance.py
```diff
@@ -45,6 +45,9 @@ class ApplicationInstance(BASE):
         "OAuth2Token", back_populates="application_instance"
     )
 
+    #: A list of all the courses for this application instance.
+    courses = sa.orm.relationship("Course", back_populates="application_instance")
```
seanh, line 49: This lets you do `application_instance.courses` to get a list of all the related courses. It also informs SQLAlchemy how to handle the related instances: if you add both a new `ApplicationInstance` and a new `Course` for that instance to the DB session at the same time, SQLAlchemy now knows to flush the `ApplicationInstance` to the DB before the `Course` to avoid an Integrity Error from Postgres

lms/models/course.py
```diff
@@ -18,6 +18,11 @@ class Course(BASE):
         primary_key=True,
     )
 
+    #: The ApplicationInstance that this access token belongs to.
+    application_instance = sa.orm.relationship(
+        "ApplicationInstance", back_populates="courses"
+    )
```
seanh, line 24: Lets you do `course.application_instance` and again informs SQLAlchemy of how to properly generate SQL for related objects

tests/factories/__init__.py
```diff
@@ -27,6 +28,9 @@ def set_sqlalchemy_session(session):
         # pylint:disable=protected-access
         factory_class._meta.sqlalchemy_session = session
 
+        if persistence:
+            factory_class._meta.sqlalchemy_session_persistence = persistence
```
seanh, line 32: The functional tests need Factory Boy to commit the DB session after creating each model instance, so they do `set_sqlalchemy_session(session, persistence="commit")`

tests/functional/conftest.py
```diff
@@ -35,13 +36,17 @@ def app(pyramid_app, db_engine):
...
 
-    yield session
+    factories.set_sqlalchemy_session(session, persistence="commit")
 
-    session.close()
+    try:
+        yield session
+    finally:
+        factories.clear_sqlalchemy_session()
+        session.close()
```
seanh, line 52: Had to add Factory Boy support to the functional tests, the same as we've done for the unit tests

tests/unit/lms/models/group_info_test.py
```diff
@@ -16,7 +16,7 @@ def test_persist_and_retrieve(self, db_session, group_info):
         assert (
             retrieved_group_info.authority_provided_id == "test_authority_provided_id"
         )
-        assert retrieved_group_info.consumer_key == "test_consumer_key"
+        assert retrieved_group_info.consumer_key == application_instance.consumer_key
```
jon-betts, line 19: I really like this style, as long as this is _why_ this value is this value, rather than just correlated. I made that mistake in a few places as the consumer key gets copied about quite a bit.

So not only does the consumer key _happen_ to equal the application instance consumer key, that's _why_ it's set like that.

tests/unit/lms/models/oauth2_token_test.py
```diff
@@ -23,7 +24,7 @@ def test_persist_and_retrieve_all_attrs(self, application_instance, db_session):
 
         token = db_session.query(OAuth2Token).one()
         assert token.user_id == "test_user_id"
-        assert token.consumer_key == "test_consumer_key"
+        assert token.consumer_key == application_instance.consumer_key
```
jon-betts, line 27:
```suggestion
```

Duplicate line of the one below

tests/unit/lms/services/application_instance_getter_test.py
```diff
@@ -121,46 +125,16 @@ def ai_getter(self, pyramid_request):
...
-                consumer_key="NOISE_CONSUMER_KEY_3",
-                developer_key="NOISE_DEVELOPER_KEY_3",
-                lms_url="NOISE_LMS_URL_3",
-                shared_secret="NOISE_SHARED_SECRET_3",
-                requesters_email="NOISE_EMAIL_3",
-            ),
-        ]
-        pyramid_request.db.add_all(application_instances)
-        return application_instances
+        return factories.ApplicationInstance.create_batch(size=3)
```
jon-betts, line 137: :partying_face: :balloon:

tests/unit/lms/services/course_test.py
```diff
@@ -109,24 +106,13 @@ def svc(self, pyramid_request):
...
-            consumer_key=consumer_key,
-            shared_secret="test_shared_secret",
-            lms_url="test_lms_url",
-            requesters_email="test_requesters_email",
-        )
-        pyramid_request.db.add(application_instance)
-        pyramid_request.db.flush()
-        return application_instance
+    def other_application_instance(self):
+        return factories.ApplicationInstance()
```
jon-betts, line 115:
```suggestion
        return factories.ApplicationInstance(
            consumer_key="something_different"
        )
```

Is this why these aren't the same? This isn't technically required, as the factoryboy default is a randomized hex, but this makes it explicit why these aren't the same.

tests/unit/lms/services/group_info_test.py
```diff
@@ -39,23 +43,25 @@ def test_it_updates_an_existing_GroupInfo_if_one_already_exists(
             factories.HGroup(
                 authority_provided_id=self.AUTHORITY, type="course_group",
             ),
-            consumer_key=self.CONSUMER_KEY,
+            consumer_key=application_instance.consumer_key,
```
jon-betts, line 46: This is doing the same work (removing a magical string), but better because it also tells you where it came from.

tests/unit/lms/services/course_test.py
```diff
@@ -91,15 +90,13 @@ def test_any_with_setting(
...
+    def add_courses_with_settings(self, pyramid_request):
         def add_courses_with_settings(settings_set, consumer_key=None):
             if consumer_key is None:
                 consumer_key = pyramid_request.lti_user.oauth_consumer_key
 
             for settings in settings_set:
-                db_session.add(
-                    factories.Course(consumer_key=consumer_key, settings=settings)
-                )
+                factories.Course(consumer_key=consumer_key, settings=settings)
```
jon-betts, line 99: Is this where we are making use of the new relationship? It's not obvious to me

lms/models/application_instance.py
seanh: It's needed for the factory to work. Otherwise SQLAlchemy will crash on saving a factory-created application instance and courses to the DB because it doesn't understand that it has to serialize the application instance first because the courses depend on it:

```python
>>> course = factories.Course(consumer_key=factories.ApplicationInstance().consumer_key)
>>> request.tm.commit()
IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "course" violates foreign key constraint "fk__course__consumer_key__application_instances"
DETAIL:  Key (consumer_key)=(Hypothesis69240600f272c337b13aa2e257bc5bbd) is not present in table "application_instances".
```

tests/functional/conftest.py
```diff
@@ -35,13 +36,17 @@ def app(pyramid_app, db_engine):
     return TestApp(pyramid_app)
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
```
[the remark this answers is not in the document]
seanh: Are there any functional tests that don't use the DB though? I suppose it might be technically possible. It seems pretty unlikely though

tests/unit/lms/models/application_instance_test.py
```diff
@@ -2,34 +2,32 @@
...
+        factories.ApplicationInstance()
+
         new_count = db_session.query(ApplicationInstance).count()
         assert new_count == initial_count + 1
 
-    def test_provisioning_defaults_to_True(self, db_session, application_instance):
-        db_session.add(application_instance)
-
+    def test_provisioning_defaults_to_True(self, application_instance, db_session):
         db_session.flush()
```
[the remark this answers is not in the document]
seanh: Yeah, this is testing a SQLAlchemy `default` / `server_default`, those don't get applied until flush. It's the same with all SQLAlchemy models

tests/unit/lms/models/application_instance_test.py
```diff
@@ -2,34 +2,32 @@
...
 
         assert application_instance.provisioning is True
 
-    def test_provisioning_can_be_disabled(self, db_session, application_instance):
+    def test_provisioning_can_be_disabled(self, application_instance):
         application_instance.provisioning = False
-        db_session.add(application_instance)
-        db_session.flush()
```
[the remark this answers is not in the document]
seanh: Good spot, I've put this one back

tests/unit/lms/models/oauth2_token_test.py
seanh: :+1:

tests/unit/lms/services/course_test.py
seanh: :+1:

tests/unit/lms/services/course_test.py
seanh: This would crash without the new relationship, yeah. Because SQLAlchemy spits out the SQL statements in the wrong order:

```
E           sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
E           (psycopg2.errors.ForeignKeyViolation) insert or update on table "course" violates foreign key constraint "fk__course__consumer_key__application_instances"
E           DETAIL:  Key (consumer_key)=(Hypothesis53fcdbb7055fb63c370fe829181b2e92) is not present in table "application_instances".
E           
E           [SQL: INSERT INTO course (consumer_key, authority_provided_id, settings) VALUES (%(consumer_key)s, %(authority_provided_id)s, %(settings)s)]
E           [parameters: ({'consumer_key': 'Hypothesis53fcdbb7055fb63c370fe829181b2e92', 'authority_provided_id': 'c146fc19badf85cb3a3a9a628b80e623eff9012b', 'settings': '{"group": {"key": "match"}}'}, {'consumer_key': 'other_consumer_key', 'authority_provided_id': '0a8decc90049cc68efc3310e8b10127dcabfd79d', 'settings': '{"group": {"key": "match"}}'})]
E           (Background on this error at: http://sqlalche.me/e/13/gkpj)
```

tests/functional/conftest.py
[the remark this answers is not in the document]
seanh: Moved this to a couple of tests directly instead

PR #1942, opened 2020-07-21
Add OAuth2Token test factory

seanh:
Extracted from https://github.com/hypothesis/lms/pull/1923/ but with some changes:

* Added `FACTORY_CLASS=SQLAlchemyModelFactory` to the factory so that Factory Boy adds instances to the DB session automatically.

* Added `application_instance=SubFactory(ApplicationInstance)` to the factory so that, when you do `factories.OAuth2Token()`, Factory Boy will also generate an `ApplicationInstance` object (using `factories.ApplicationInstance`) and set it as the new `OAuth2Token`'s application instance.

  `OAuth2Token` has a non-nullable foreign key to `ApplicationInstance` so you can't save an `OAuth2Token` to the DB without first creating a matching `ApplicationInstance`, so it's convenient for the factory to do this for you. This simplifies a lot of tests that were having to create `ApplicationInstance`'s for this reason, now they don't.

  If you _don't_ want an `ApplicationInstance` to be generated for you, you can override it:

  ```python
  factories.OAuth2Token(application_instance=None)
  ```

  If you already have an `ApplicationInstance` and want that to be used, you can pass it in:

  ```python
  factories.OAuth2Token(application_instance=my_application_instance)
  ```

  If you do want an `ApplicationInstance` to be generated but you want to specify some of its fields, you can do that:

  ```python
  factories.OAuth2Token(application_instance__consumer_key=my_consumer_key)
  ```

  For more see the [`SubFactory` docs](https://factoryboy.readthedocs.io/en/latest/reference.html?highlight=exclude#subfactory)

* Removed `consumer_key` from the factory. Use `application_instance` instead

* I've updated all the existing tests to use the new factory

### DIFF
--- a/tests/unit/lms/services/canvas_api_test.py
+++ b/tests/unit/lms/services/canvas_api_test.py
@@ -34,8 +34,7 @@ def _make_response(json_data=None, raw=None, status_code=200, headers=None):
 
 
 class TestDataCalls:
-    @pytest.mark.usefixtures("access_token")
-    def test_authenticated_users_sections(self, api_client, http_session):
+    def test_authenticated_users_sections(self, access_token, api_client, http_session):
         sections = [{"id": 1, "name": "name_1"}, {"id": 2, "name": "name_2"}]
         http_session.send.return_value = _make_response({"sections": sections})
 
@@ -48,13 +47,12 @@ class TestDataCalls:
                 url=Any.url.with_path("api/v1/courses/COURSE_ID").with_query(
                     {"include[]": "sections"}
                 ),
-                headers={"Authorization": "Bearer existing_access_token"},
+                headers={"Authorization": f"Bearer {access_token.access_token}"},
             ),
             timeout=Any(),
         )
 
-    @pytest.mark.usefixtures("access_token")
-    def test_course_sections(self, api_client, http_session):
+    def test_course_sections(self, access_token, api_client, http_session):
         sections = [
             {"id": 101, "name": "name_1"},
             {"id": 102, "name": "name_2"},
@@ -71,7 +69,7 @@ class TestDataCalls:
             Any.request(
                 "GET",
                 url=Any.url.with_path("api/v1/courses/COURSE_ID/sections"),
-                headers={"Authorization": "Bearer existing_access_token"},
+                headers={"Authorization": f"Bearer {access_token.access_token}"},
             ),
             timeout=Any(),
         )
@@ -85,8 +83,7 @@ class TestDataCalls:
         with pytest.raises(CanvasAPIError):
             api_client.course_sections("dummy")
 
-    @pytest.mark.usefixtures("access_token")
-    def test_users_sections(self, api_client, http_session):
+    def test_users_sections(self, access_token, api_client, http_session):
         http_session.send.return_value = _make_response(
             {
                 "enrollments": [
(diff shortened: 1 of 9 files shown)


tests/unit/lms/models/oauth2_token_test.py
```diff
@@ -8,118 +8,77 @@
...
 
-    def test_expires_in_defaults_to_None(self, db_session, init_kwargs):
-        token = OAuth2Token(**init_kwargs)
-        db_session.add(token)
+    def test_expires_in_defaults_to_None(self, db_session):
+        token = factories.OAuth2Token()
 
         db_session.flush()
 
         assert token.expires_in is None
```
jon-betts, line 77: I think the way this particular set of tests work is a no-no. We should not be testing the factory created version of an object in the tests of that object. I think this particular test shows the clearest example of why.

Currently the factory does not set a value for expires in, but it absolutely could / should. It would be totally reasonable to add a randomizer to it so that it was `None` 50% of the time. This would intermittently (based on the repeatable randomness) break this test. This shouldn't be possible.

This is because this test is testing the compound of the model class _and_ the factory object. The factory does not have to behave like the object, _particularly_ in the way the defaults are setup. So is _absolutely mandatory_ it not be used for testing initialisation behavior, but I'd say in general it's better not to use it to test _any_ behavior of the object in case the way it works is different in a more subtle way.

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -48,13 +47,12 @@ def test_authenticated_users_sections(self, api_client, http_session):
                 url=Any.url.with_path("api/v1/courses/COURSE_ID").with_query(
                     {"include[]": "sections"}
                 ),
-                headers={"Authorization": "Bearer existing_access_token"},
+                headers={"Authorization": f"Bearer {access_token.access_token}"},
```
jon-betts, line 50: :+1: More magical string value removal

tests/unit/lms/services/canvas_api_test.py
```diff
@@ -565,10 +563,12 @@ def test_it_raises_CanvasAPIAccessTokenError_if_refresh_token_is_None(
...
 
     @pytest.fixture
-    def access_token_no_refresh(self, db_session, access_token_fields):
-        access_token = OAuth2Token(**dict(access_token_fields, refresh_token=None))
-        db_session.add(access_token)
-        return access_token
+    def access_token_no_refresh(self, pyramid_request):
+        return factories.OAuth2Token(
+            user_id=pyramid_request.lti_user.user_id,
+            application_instance__consumer_key=pyramid_request.lti_user.oauth_consumer_key,
```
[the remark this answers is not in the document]
seanh: Sure. Generally I don't think we can comment stuff when just using standard features of libraries we use. For example we wouldn't want to put a comment every time we use this particular Factory Boy feature. But I'll add one this time. A better solution might be to have our own Factory Boy crash course or cheat sheet, focused on how we use Factory Boy / features of Factory Boy that we use

tests/unit/lms/models/oauth2_token_test.py
seanh: I've changed this back to using an `init_kwargs` fixture, but the fixture now uses a factory to generate a random dict instead of just returning a hardcoded dict

jon-betts: I think that gives us a best of both situation.

jon-betts:
Has this branch got messed up? I can't see the tests for OAuth2Token at all, and there's a bunch of attribute file rename changes?

seanh:
I decided to make `tests.factories._attributes` public, since they can be imported by test modules to use when creating their own within-module factories. No tests actually import them yet though.

I just reset the `models/oauth2_token.py` tests back to how they are on master--they just have an `init_kwargs` fixture with hard-coded attribute values. No factory

PR #1944, opened 2020-07-22
Add Course.application_instance sub-factory

seanh:
* Add `application_instance=SubFactory(ApplicationInstance)` to the `Course` factory so that, when you do `factories.Course()`, Factory Boy will also generate an `ApplicationInstance` object (using `factories.ApplicationInstance`) and set it as the new `Course`'s application instance.

  `Course` has a non-nullable foreign key to `ApplicationInstance` so you can't save a `Course` to the DB without first creating a matching `ApplicationInstance`, so it's convenient for the factory to do this for you.
  
  If you _don't_ want an `ApplicationInstance` to be generated for you, you can override it:

  ```python
  factories.Course(application_instance=None)
  ```

  If you already have an `ApplicationInstance` and want that to be used, you can pass it in:

  ```python
  factories.Course(application_instance=my_application_instance)
  ```

  If you do want an `ApplicationInstance` to be generated but you want to specify some of its fields, you can do that:

  ```python
  factories.Course(application_instance__consumer_key=my_consumer_key)
  ```

  For more see the [`SubFactory` docs](https://factoryboy.readthedocs.io/en/latest/reference.html?highlight=exclude#subfactory)

* Removed `consumer_key` from the factory. Use `application_instance` instead

* Updated the existing tests

### DIFF
--- a/tests/unit/lms/services/course_test.py
+++ b/tests/unit/lms/services/course_test.py
@@ -25,10 +25,10 @@ class TestCourseService:
         )
 
     def test_it_does_nothing_if_theres_already_a_matching_row(
-        self, ai_getter, pyramid_request, svc
+        self, ai_getter, application_instance, pyramid_request, svc
     ):
         existing_course = factories.Course(
-            consumer_key=pyramid_request.lti_user.oauth_consumer_key,
+            application_instance=application_instance,
             authority_provided_id="test_authority_provided_id",
             settings={},
         )
@@ -73,31 +73,26 @@ class TestCourseService:
         ),
     )
     def test_any_with_setting(
-        self,
-        svc,
-        settings_set,
-        value,
-        expected,
-        add_courses_with_settings,
-        other_application_instance,
+        self, svc, settings_set, value, expected, add_courses_with_settings,
     ):
         add_courses_with_settings(settings_set)
         # Add a matching course with a different consumer key
         add_courses_with_settings(
             [{"group": {"key": value}}],
-            consumer_key=other_application_instance.consumer_key,
+            application_instance=factories.ApplicationInstance(),
         )
 
         assert svc.any_with_setting("group", "key", value) is expected
 
     @pytest.fixture
-    def add_courses_with_settings(self, pyramid_request):
-        def add_courses_with_settings(settings_set, consumer_key=None):
-            if consumer_key is None:
-                consumer_key = pyramid_request.lti_user.oauth_consumer_key
-
+    def add_courses_with_settings(self, application_instance):
+        def add_courses_with_settings(
+            settings_set, application_instance=application_instance
+        ):
             for settings in settings_set:
-                factories.Course(consumer_key=consumer_key, settings=settings)
+                factories.Course(
+                    application_instance=application_instance, settings=settings
+                )
 
         return add_courses_with_settings
 
@@ -110,10 +105,6 @@ class TestCourseService:
         return factories.ApplicationInstance(
             consumer_key=pyramid_request.lti_user.oauth_consumer_key
         )
-
-    @pytest.fixture
-    def other_application_instance(self):
-        return factories.ApplicationInstance(consumer_key="other_consumer_key")
 
 
 pytestmark = pytest.mark.usefixtures("ai_getter")
--- a/tests/factories/course.py
+++ b/tests/factories/course.py
@@ -2,11 +2,11 @@ import factory
 from factory.alchemy import SQLAlchemyModelFactory
 
 from lms import models
-from tests.factories.attributes import OAUTH_CONSUMER_KEY
+from tests.factories import ApplicationInstance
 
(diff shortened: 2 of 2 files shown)


tests/unit/lms/services/course_test.py
```diff
@@ -28,7 +29,7 @@ def test_it_does_nothing_if_theres_already_a_matching_row(
         self, ai_getter, pyramid_request, svc
     ):
         existing_course = factories.Course(
-            consumer_key=pyramid_request.lti_user.oauth_consumer_key,
```
seanh, line 31: Previously this just created a `Course` with randomly-generated `consumer_key` string but did not create a corresponding `ApplicationInstance` that has that consumer key. `existing_course.application_instance` would be `None`. If you tried to save the course to the DB it would crash. _Except_ that there was a separate `application_instance` fixture (at the bottom of the file, with `autouse=True`) with the same `consumer_key` so it would happen to work.

I think the new version is better: create the `Course` and the `ApplicationInstance` with the matching `consumer_key` in one place.

tests/unit/lms/services/course_test.py
```diff
@@ -9,6 +9,7 @@
 
 
 class TestCourseService:
+    @pytest.mark.usefixtures("application_instance")
```
seanh, line 12: This `application_instance` fixture was previously `autouse`. Only a few tests need it now so I removed the `autouse` and added explicitly to the tests that actually need it

tests/unit/lms/services/course_test.py
```diff
@@ -28,7 +29,7 @@ def test_it_does_nothing_if_theres_already_a_matching_row(
         self, ai_getter, pyramid_request, svc
     ):
         existing_course = factories.Course(
-            consumer_key=pyramid_request.lti_user.oauth_consumer_key,
+            application_instance__consumer_key=pyramid_request.lti_user.oauth_consumer_key,
             authority_provided_id="test_authority_provided_id",
         )
```
jon-betts, line 34: I don't think we need to use the magic here, we can just instantiate this course like we do the others in the fixture using the `application_instance` fixture which contains the logic presented here. This makes it a simpler.

```suggestion
            application_instance=application_instance,
            authority_provided_id="test_authority_provided_id",
        )
```

I can't put the suggestion in full as github doesn't let you put suggestions over removed lines, but you'd need to pull in the fixture in the args for this to work.

tests/unit/lms/services/course_test.py
```diff
@@ -39,6 +40,7 @@ def test_it_does_nothing_if_theres_already_a_matching_row(
         existing_course = pyramid_request.db.query(Course).one()
         assert not existing_course.settings.get("canvas", "sections_enabled")
 
+    @pytest.mark.usefixtures("application_instance")
```
jon-betts, line 43: As a result of the above, we actually _do_ use the `application_instance` directly or indirectly in every test. It looks like `test_any_with_setting()` below doesn't, but it actually does by virtue of the `add_courses_with_settings` fixture.

So we could put back the `autouse=True` and remove these lines.

tests/unit/lms/services/course_test.py
```diff
@@ -72,47 +75,39 @@ def test_it_inserts_False_if_theres_a_matching_row_in_course_groups_exported_fro
...
     )
     def test_any_with_setting(
-        self,
-        svc,
-        settings_set,
-        value,
-        expected,
-        add_courses_with_settings,
-        other_application_instance,
+        self, svc, settings_set, value, expected, add_courses_with_settings,
```
jon-betts, line 78: :+1: I love it when this happens

tests/unit/lms/services/course_test.py
```diff
@@ -72,47 +75,39 @@ def test_it_inserts_False_if_theres_a_matching_row_in_course_groups_exported_fro
...
     ):
+        other_application_instance = factories.ApplicationInstance()
         add_courses_with_settings(settings_set)
         # Add a matching course with a different consumer key
         add_courses_with_settings(
             [{"group": {"key": value}}],
-            consumer_key=other_application_instance.consumer_key,
+            application_instance=other_application_instance,
         )
```
jon-betts, line 86: I can't give a suggestion 'cos a line has gone again... but the order here reads weirdly to me here. The first create actually uses the regular `application_instance`, but a quick reading looks like it might involve the `other_application_instance`. But that's just defined ahead of time. 

I think a better ordering might be:

```python
        add_courses_with_settings(settings_set)
        
        # Add a matching course with a different consumer key
        other_application_instance = factories.ApplicationInstance()
        add_courses_with_settings(
            [{"group": {"key": value}}],
            application_instance=other_application_instance,
        )
```

I think it's good to assign this to a variable in this case, as it gives us an opportunity to name it, and reveal intent.

jon-betts:
Just fixed up a conflict with master.

PR #1945, opened 2020-07-22
Remove an unnecessary `settings` attribute

seanh:
This does mean that if a test creates a new `Course` and then it wants to call `course.settings.set()` or `get()` before having added it to the DB, the test has to first set `settings={}` itself because the `server_default` only takes effect on saving to the DB. This is the same as how `models.Course.settings` works for production code, though, so I think it might be best for the factory _not_ to hide that fact.

I think maybe a good guide for how clever a factory should be / how many of the model's attributes a factory should autogenerate might be: it should generate the minimum set of attrs necessary to save the object to the DB. And `settings` isn't necessary for that. It's a balance between making factories convenient for tests to use (reducing how much work a test must do itself) versus factories potentially hiding bugs because they're generating a lot of properties that wouldn't exist on a new object in production code.

At least I think that guide should apply for default factories like `factories.Course()`. I think it would be fine to add specialized factories that do generate more fields. For example a `factories.CourseWithCanvasSectionsEnabled()` would be fine.

### DIFF
--- a/tests/factories/course.py
+++ b/tests/factories/course.py
@@ -9,5 +9,4 @@ Course = factory.make_factory(  # pylint:disable=invalid-name
     FACTORY_CLASS=SQLAlchemyModelFactory,
     consumer_key=OAUTH_CONSUMER_KEY,
     authority_provided_id=factory.Faker("hexify", text="^" * 40),
-    settings=factory.lazy_attribute(lambda o: {}),
 )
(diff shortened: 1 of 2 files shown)


jon-betts:
You also factoryboy has a way to do "bundles" of settings, so you could do something like "Course(canvas_sections=True)" and have that apply the correct thing for each attribute.

robertknight:
I'm currently trying to reduce the queue of open PRs. This one is trivial so I'm going to merge it.

Issue #1782, opened 2020-05-28
Improve accessibility for Canvas Files Authorize pop-up


klemay:
When a teacher or student launches an assignment created with Canvas files, they must authorize Hypothesis to launch the assignment. When using VoiceOver for Mac, that looks like:

<img width="1491" alt="Screen Shot 2020-05-28 at 2 40 36 PM" src="https://user-images.githubusercontent.com/12087705/83180383-5c0ade00-a0f1-11ea-8662-504441719f43.png">

The text at the bottom of this screenshot, starting with "Authorize Hypothesis and 2 other items..." is what is read aloud to the user. 

Ideally we would read the text of the Authorize Hypothesis pop-up and make it clear to the user what the Authorize button does.

# Steps to reproduce
1. Log into Canvas as a teacher. Create a Canvas assignment using the Canvas File Picker (follow [these instructions](https://web.hypothes.is/help/using-the-hypothesis-app-with-assignments-in-canvas/) and choose Option 3C for Step 3)
2. Log in to Canvas as a student. Turn on VoiceOver. 
3. Launch the assignment as a student

PR #1951, opened 2020-07-23
Improve screen readers on dialog errors

LMS007:
- Move the `aria-*` attributes down from the dialog root element to its content wrapper. This seems to prevent Voice Over from reading out “x items in group” after reading the header.

- Add `aria-describedby`. This facilitates reading of the actual content by the screen readers

- Pass a button ref as `initialFocus` for dialogs that have an authorize button in BasicLtiLaunchApp. This allows the screen reader to state the call to action after reading the content.

- These concepts are from https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/alertdialog.html

fixes https://github.com/hypothesis/lms/issues/1782

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -258,14 +258,20 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
     </span>
   );
 
+  const focusedDialogButton = useRef(
+    /** @type {HTMLButtonElement | null} */ (null)
+  );
+
   const errorDialog = (
     <Fragment>
       {errorState === 'error-authorizing' && (
         <Dialog
+          initialFocus={focusedDialogButton}
           title="Authorize Hypothesis"
           role="alertdialog"
           buttons={[
             <Button
+              buttonRef={focusedDialogButton}
               onClick={authorizeAndFetchUrl}
               className="BasicLtiLaunchApp__button"
               label="Authorize"
@@ -278,11 +284,13 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
       )}
       {errorState === 'error-fetch' && (
         <Dialog
+          initialFocus={focusedDialogButton}
           title="Something went wrong"
           contentClass="BasicLtiLaunchApp__dialog"
           role="alertdialog"
           buttons={[
             <Button
+              buttonRef={focusedDialogButton}
               onClick={authorizeAndFetchUrl}
               className="BasicLtiLaunchApp__button"
               label="Try again"
--- a/lms/static/scripts/frontend_apps/components/Button.js
+++ b/lms/static/scripts/frontend_apps/components/Button.js
@@ -3,6 +3,7 @@ import { createElement } from 'preact';
 
 /**
  * @typedef ButtonProps
+ * @prop {Object} [buttonRef]
  * @prop {string} [className]
  * @prop {boolean} [disabled]
  * @prop {string} label
@@ -18,6 +19,7 @@ export default function Button({
   disabled = false,
   label,
   onClick,
+  buttonRef,
   type = 'button',
 }) {
   return (
@@ -26,6 +28,7 @@ export default function Button({
       disabled={disabled}
       onClick={onClick}
       type={type}
+      ref={buttonRef}
     >
       {label}
     </button>
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -53,8 +53,9 @@ export default function Dialog({
   title,
   buttons,
 }) {
-  const dialogTitleId = useUniqueId('dialog');
-  const rootEl = useRef(/** @type {HTMLDivElement|null} */ (null));
+  const dialogTitleId = useUniqueId('dialog-title');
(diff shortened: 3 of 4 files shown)


lms/static/scripts/frontend_apps/components/Button.js
```diff
@@ -26,5 +28,6 @@ Button.propTypes = {
   disabled: propTypes.bool,
   label: propTypes.string,
   onClick: propTypes.func,
+  ref: propTypes.object,
```
LMS007, line 31: Effectively this is a pass through ref down to the button, but I could not find another example of doing this in the code which made me question it entirely. Otherwise, I think we would have to account for the whole `ref.current.base` vs `ref.current` between an JSX Element and Element and I'd rather not expose that to `<Dialog>`

lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -55,6 +55,7 @@ export default function Dialog({
   buttons,
 }) {
   const dialogTitleId = useUniqueId('dialog');
+  const dialogDescriptionId = useUniqueId('description');
```
robertknight, line 58:
```suggestion
  const dialogTitleId = useUniqueId('dialog-title');
  const dialogDescriptionId = useUniqueId('dialog-description');
```

Perhaps use a slightly more specific ID prefix here.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -20,6 +20,8 @@ import ErrorDisplay from './ErrorDisplay';
 import LMSGrader from './LMSGrader';
 import Spinner from './Spinner';
 
+/** @typedef {import("preact").JSX.Element} JSXElement */
```
robertknight, line 23: This import is unused.

robertknight:
I rebased this PR just now to see if would resolve a failure in the backend tests on this branch.

robertknight:
I've rebased again on top of seanh's fixes in master for the pylint issue.

robertknight:
I assume this is ready for review @LMS007?

lms/static/scripts/frontend_apps/components/Button.js
```diff
@@ -3,10 +3,11 @@ import { createElement } from 'preact';
 
 /**
  * @typedef ButtonProps
+ * @prop {Object} [buttonRef]
  * @prop {string} [className]
  * @prop {boolean} [disabled]
  * @prop {string} label
- * @prop {() => any} onClick
+ * @prop {(e: Event) => any} onClick
```
robertknight, line 10: What was the `onClick` change required for here?

robertknight:
On the dev sync call just now @LMS007 mentioned that he had some concerns about problems with the changes in this PR that are not mentioned in the description. Can you explain what those were @LMS007?

LMS007:
I went ahead and did a bit more testing. I looked back at the w3 spec for both dialog and alertdialog and all signs seem to indicate we want to set a focus on _something_ in initially. If this does not happen then VoiceOver will just scan all the elements in the dialog and barf everything out -- this tends to be confusing and does not teach the user about a call-to-action. More-so, both examples in w3 relating to modal dialogs set an initial focus. 

Having said that, VoiceOver is still fickle in is behavoir. It reads the dialog in canvas a bit differently inline vs a separate tab. 
Screen Shot 2020-08-03 at 4 15 03 PM

It also seems to be inconstant. At times it scans everything in the dialog as if nothing is in focus, but other times, it will just read things in a predictable way based on what is in focus. I think the tool itself is very inconsistent and I suspect it may have to do with what it was in the middle of previously doing when you triggered a refresh. 

In any case it also makes sense to assign `initialFocus` on the `LMSFilePicker` so I went ahead and did that too.

lms/static/scripts/frontend_apps/components/Button.js
LMS007: I think it's not needed. Going to remove it.

robertknight:
For future reference, the specific advice regarding keyboard interaction for dialogs is here: https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7:

> When a dialog opens, focus placement depends on the nature and size of the content.
In all circumstances, focus moves to an element contained in the dialog.
Unless a condition where doing otherwise is advisable, focus is initially set on the first focusable element.
If content is large enough that focusing the first interactive element could cause the beginning of content to scroll out of view, it is advisable to add tabindex=-1 to a static element at the top of the dialog, such as the dialog title or first paragraph, and initially focus that element.
If a dialog contains the final step in a process that is not easily reversible, such as deleting data or completing a financial transaction, it may be advisable to set focus on the least destructive action, especially if undoing the action is difficult or impossible. The Alert Dialog Pattern is often employed in such circumstances.
If a dialog is limited to interactions that either provide additional information or continue processing, it may be advisable to set focus to the element that is likely to be most frequently used, such as an OK or Continue button.

On the basis of the above, focusing the "Try again" or "Authorize" button in authorization dialogs seems reasonable to me. In the LMS file picker however I think we'd want to focus the file list rather than the "Select" button.

robertknight:
Hi @LMS007 - Did you leave this PR open because you wanted me to look at the changes in your last push? I did some more testing this morning and it looks fine to me. I'm going to go ahead and merge this.

PR #1952, opened 2020-07-24
Remove prop-types props documentation

robertknight:
Prop-types documentation + runtime prop checking has been superseded by JSDoc checked by TypeScript. This PR therefore removes prop-types usage from the lms frontend so we don't have to specify Preact component props in two different ways:

- Disable the ESLint prop-types checks
- Add missing JSDoc-based prop docs to the `Button` component
- Remove the prop-types based documentation from each UI component
- Remove our custom prop-types extensions
- Modify `mockImportedComponents` to work with function components that don't have a `propTypes` static property. The new approach involves a bit of a hack, but it is simple and easy to change in future if we need to.

A downside to removing prop-types is that there won't be such an obvious runtime error any more if code passes in a value of the wrong type. That can happen if the type of a value at runtime doesn't match the static type that TS checked (eg. due to an incorrect `/** @type {...} */` cast). We could potentially generate prop-types automatically using a Babel plugin as an extra layer of safety, but for now I'd like to see how we get on without it.


robertknight:
Thanks @lyzadanger. Let's see how we get on with this. If it turns out that we do have a reason for wanting runtime-checked props for any reason, I'll figure out a way to generate it automatically from the JSDoc.

PR #1984, opened 2020-08-06
Recompile all requirements/*.txt files

seanh:
Re-run pip-compile on every requirements/*.in files to re-compile all
the requirements/*.txt files. This just updates all the annotation comments etc on the requirements/*.txt files since they got moved into the requirements/ directory. This shouldn't actually change any of the requirements.


seanh:
For the record the command for doing this is `tox -e pipcompile --run-command 'pip-compile requirements/<SOMETHING>.in'`

PR #1994, opened 2020-08-11
Add a feature flag for enabling Via internal `pywb` rewriter

jon-betts:
This also includes the shortcut of going directly to the pywb end-point.

The flag is `use_via_rewriter` which can be enabled using cookies at: http://0.0.0.0:8001/flags

This should enable:

 * A short cut directly to the `pywb` end-point for known HTML documents
 * Adding the `via.rewrite` param for unknown types

This doesn't work right now, as the rewrite isn't implemented, but the routing should work

## Testing notes

Of course... you can't use the cookie flag due to the secure / not-secure cookie issue. So I think the only way to test is to set the environment variable: FEATURE_FLAG_USE_VIA_REWRITER

Then start everything and visit: https://hypothesis.instructure.com/courses/125/assignments/873

This should be broken, but when you inspect the location it should have `/html/v` in it.

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -32,7 +32,15 @@ class _ViaDoc:
 class _ViaClient:
     """A small wrapper to make calling Via easier."""
 
-    def __init__(self, service_url, legacy_service_url, host_url, legacy_mode=False):
+    def __init__(
+        # pylint: disable=too-many-arguments
+        self,
+        service_url,
+        legacy_service_url,
+        host_url,
+        legacy_mode=False,
+        via_rewriter_mode=False,
+    ):
         self.service_url = urlparse(service_url)
         self.legacy_service_url = legacy_service_url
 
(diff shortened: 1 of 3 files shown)


lms/views/helpers/_via.py
```diff
@@ -32,7 +32,15 @@ def is_html(self):
 class _ViaClient:
     """A small wrapper to make calling Via easier."""
 
-    def __init__(self, service_url, legacy_service_url, host_url, legacy_mode=False):
+    def __init__(
```
jon-betts, line 35: This is pretty much turning into the core of Via 3 routing logic.

tests/unit/lms/views/helpers/_via_test.py
```diff
@@ -44,16 +43,55 @@ def test_if_creates_the_correct_via_url(self, pyramid_request):
...
+    @pywb_test_params
+    def test_it_merges_params_correctly_for_legacy_via(
+        self, pyramid_request, params, expected_extras
+    ):
         final_url = via_url(
             pyramid_request, f"http://doc.example.com/?{urlencode(params, doseq=True)}"
         )
 
+        assert final_url == Any.url.with_host("test_legacy_via_server.is")
+        assert final_url == Any.url.with_path("/http://doc.example.com/")
```
jon-betts, line 57: I expanded these tests to check we were going to the right host and path. It was a bit wishy washy before

jon-betts:
Hmm, something blown up in CI. Might be a merge issue? I think this was working ok locally relatively recently

PR #1996, opened 2020-08-12
Remove an obsolete ESLint config setting

robertknight:
Remove the `ecmaVersion` setting which overrode settings from the
`eslint-config-hypothesis` package and caused an error if trying to use
ES language features from ES 2019+


robertknight:
I'm going to go ahead and merge this trivial PR.

PR #2023, opened 2020-08-19
Re-introduce accessible descriptions for dialogs

robertknight:
This is a follow-up to https://github.com/hypothesis/lms/pull/2018 which
re-introduces accessible descriptions for dialogs using a different
approach. Instead of wrapping all of the dialog contents in an element
which is marked as the description, a layout effect instead picks out
the first `<p>` element, if any, and marks it as the description.

Compared to the previous approach this has some benefits:

 - It only identifies static text as a description, not non-descriptive
   elements like tables or images

 - It doesn't wrap the children, which avoids the issue that was fixed
   in #2018

There is also a downside in that it won't update if the dialog's content
changes after it is initially displayed. I believe this is an acceptible
limitation.

There are other approaches we could consider such as requiring `Dialog`
consumers to explicitly specify a separate `description` prop. That
would be more React-y, although it may duplicate the content. The
approach taken here has the benefit of just-working with existing
dialogs.

----

**Testing:**

The easiest way to see the effect of this is to trigger a few dialogs in the LMS app, select the `Dialog__content` element in the dev tools "Elements" view and look at the element's computed accessibility properties in the "Accessibility" tab. The "Description" value should match the first `<p>` element in the dialog, if there is one. For example:

Dialog description

### DIFF
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -1,5 +1,5 @@
 import { createElement, Fragment } from 'preact';
-import { useEffect, useRef } from 'preact/hooks';
+import { useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 import classNames from 'classnames';
 
 import Button from './Button';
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -76,6 +77,19 @@ export default function Dialog({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  // If the content of the dialog contains a paragraph of text, mark it as the
+  // dialog's accessible description.
+  //
+  // A limitation of this approach is that it doesn't update if the dialog's
+  // content changes after the initial render.
+  useLayoutEffect(() => {
+    const description = rootEl.current.querySelector('p');
```
LMS007, line 86: hacky indeed, but works 👍

PR #2034, opened 2020-08-19
Remove duplicate authUrl setting

seanh:
authUrl was in the settings twice (as authUrl and canvas.authUrl).

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -311,9 +311,6 @@ class JSConfig:
                 # authenticate itself to the API.
                 "authToken": self._auth_token()
             },
-            # The URL that the JavaScript code will open if it needs the user to
-            # authorize us to request a new Canvas access token.
-            "authUrl": self._request.route_url("canvas_api.authorize"),
             "canvas": {
                 # The URL that the JavaScript code will open if it needs the user to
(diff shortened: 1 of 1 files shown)


lms/resources/_js_config.py
```diff
@@ -311,9 +311,6 @@ def _config(self):
                 # authenticate itself to the API.
                 "authToken": self._auth_token()
             },
-            # The URL that the JavaScript code will open if it needs the user to
-            # authorize us to request a new Canvas access token.
-            "authUrl": self._request.route_url("canvas_api.authorize"),
```
robertknight, line 316: Can you remove this from the client's type definitions here as well: https://github.com/hypothesis/lms/blob/a5debb5758dc61e6ed21d0f836f626bb0dbd71af/lms/static/scripts/frontend_apps/config.js#L82

Running `yarn typecheck` after that produces no errors, so it looks like nothing was using it on the client side.

robertknight (approved):
Looks fine, but the `authUrl` property also needs to be removed from `frontend_apps/config.js` which is the frontend's view of what configuration looks like.

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2048

PR #2037, opened 2020-08-20
Disable "Authorize" button when re-fetching groups

robertknight:
After completing OAuth authorization when launching an assignment, the
"Authorize" button should be disabled while the LMS frontend waits for
both content URL and group fetching to complete. Previously the button
was disabled while the API request to fetch the content URL was pending,
but not if the content URL had been fetched and group fetching was still
in progress.

This commit resolves that issue by adding conditional logic to the
`fetchGroups` helper to control whether it updates the `fetchCount`
state that tracks the number of API requests in flight that block
display of the content. This logic is conditional because fetching
groups does not block content display on the initial launch but it does
after authorization, to avoid a flicker (as explained in #1826) if
authorization fails a second time.

Part of #1826

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -69,7 +69,8 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
   // or invoke the API callback to fetch the URL otherwise.
   const [contentUrl, setContentUrl] = useState(viaUrl ? viaUrl : null);
 
-  // How many current blocking request are pending
+  // Count of pending API requests which must complete before the assignment
+  // content can be shown.
   const [fetchCount, setFetchCount] = useState(0);
 
   // `AuthWindow` instance, set only when waiting for the user to approve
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -217,7 +238,7 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
       await authWindow.current.authorize();
       const [fetchedContent, fetchedGroups] = await Promise.all([
         fetchContentUrl(),
-        fetchGroups(),
+        fetchGroups(true /* updateFetchCount */),
```
LMS007, line 241: okay, makes sense

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -521,6 +521,45 @@ describe('BasicLtiLaunchApp', () => {
         // Dialog prompt should remain.
         await dialogShouldRemain(wrapper);
       });
+
+      it('disables "Authorize" button while re-fetching content and groups', async () => {
```
LMS007, line 525: the tests for this file are getting crazy!

robertknight: > the tests for this file are getting crazy!

Indeed. I think it may be worth looking at splitting some functionality out of this component soon or otherwise refactoring how it works.

PR #2045, opened 2020-08-24
Only run backend functional tests on Jenkins

seanh:
Since they're running on GitHub Actions anyway it's not necessary to run the frontend, unit or BDD tests on Jenkins. Run only the functional tests as a quick sanity check.


robertknight:
> Since they're running on GitHub Actions anyway it's not necessary to run the frontend, unit or BDD tests on Jenkins.

What will happen if an integration issue occurs, such as a PR which passes on the branch but fails after a merge into a master branch which has since been updated with conflicting changes?

seanh:
This should fix that issue: https://github.com/hypothesis/lms/pull/2051

PR #2046, opened 2020-08-24
Don't show error dialogs when refresh tokens expire

seanh:
When a Canvas API refresh token expires (or the access token gets deleted from Canvas) Canvas sends us a "refresh_token not found" error response. When this happens just ask the user to re-authorize, don't show them an error dialog that makes it look like something went wrong.


robertknight (approved):
LGTM. This worked as expected with a quick manual test of deleting the refresh token from Canvas in the account settings page.

judell:
Thanks, @seanh.

PR #2047, opened 2020-08-24
Fail test if an async error happens in-between tests

robertknight:
If an uncaught exception is thrown or promise rejected in-between tests,
ensure the test suite fails. This applies a workaround for an issue in
karma-mocha which was earlier applied for the client in
https://github.com/hypothesis/client/pull/2421.

This caught several problems in the existing tests which are also fixed
in this commit:

 - `LSMGrader` did not mock the complex `SubmitGradeForm` component as
   we would normally do, and also did not do all the setup required for
   testing `SubmitGradeForm`. As a result, the tests ended up attempting
   real API calls which triggered warnings and produced exceptions which
   went uncaught. Resolve the issue by using `mockImportedComponents` to
   mock all imported components

 - One of the tests for `postmessage_json_rpc/client` did not mock a response
   from the Hypothesis client, resulting in a timeout exception

 - A mock in the `BasicLtiLaunchApp` tests was missing a required
   `focus` method. Also fix an issue where an invocation of a callback
   prop was not wrapped in `act`, as is best practice.

### DIFF
--- a/lms/static/scripts/bootstrap.js
+++ b/lms/static/scripts/bootstrap.js
@@ -12,3 +12,25 @@ configure({ adapter: new Adapter() });
 import { registerIcons } from './frontend_apps/components/SvgIcon';
 import iconSet from './frontend_apps/icons';
 registerIcons(iconSet);
+
+// Ensure that uncaught exceptions between tests result in the tests failing.
+// This works around an issue with mocha / karma-mocha, see
+// https://github.com/hypothesis/client/issues/2249.
+let pendingError = null;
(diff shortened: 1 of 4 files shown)


robertknight:
> This applies a workaround for an issue inkarma-mocha which was earlier applied for the client in hypothesis/client#2421.

It turns out the issue is not only karma-mocha. Mocha doesn't currently intercept the `unhandledrejection` event at all by default, so uncaught promise rejections don't result in any kind of failure from karma-mocha's point of view.

PR #2049, opened 2020-08-24
Move exception views into a view class

seanh:
No functional change, just refactoring.

The diff is bad due to indentation changing but these are the same exact views, just moved into a class.

This removes duplication of the `renderer="json"` in the view configs, and also makes the tests a little nicer.

It's going to make future changes to these exception views nicer too.

### DIFF
--- a/lms/views/api/error.py
+++ b/lms/views/api/error.py
@@ -4,6 +4,7 @@ from pyramid.view import (
     exception_view_config,
     forbidden_view_config,
     notfound_view_config,
+    view_defaults,
 )
 
 from lms.services import CanvasAPIAccessTokenError, CanvasAPIError, LTIOutcomesAPIError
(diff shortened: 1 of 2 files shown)


lms/views/api/error.py
```diff
@@ -12,52 +13,125 @@
...
+      we can't use 5xx statuses in production.
+
+    * 422 is used for validation errors (if the frontend sent invalid params to
+      a backend API) (in practice this should never happen)
+
+    * 404 Not Found is used for API endpoints that don't exist
+      (in practice this should never happen)
+
+    * 500 Server Error is used for unexpected exceptions (our code crashed)
+      (in practice this should never happen)
```
seanh, line 79: Took the opportunity to add a long docstring explaining how the API error responses work. This has caused confusion before so it seemed worth documenting.

lms/views/api/error.py
```diff
@@ -12,52 +13,125 @@
...
-@exception_view_config(context=CanvasAPIAccessTokenError, renderer="json")
-def canvas_api_access_token_error(request):
-    request.response.status_int = 400
-    # For a CanvasAPIAccessTokenError we don't send any error message or
-    # details to the frontend because we don't want the frontend to show any
-    # error message to the user in this case. Just the 400 status so the
-    # frontend knows that the request failed, and that it should show the user
-    # an [Authorize] button so they can get a (new) access token and try again.
-    return {"message": None, "details": None}
+    1. "auth_url": A URL to (re-)authenticate with the third-party API that's
```
robertknight, line 24: This key is not present in any of the responses currently, but I understand that you're going to add it shortly.

robertknight (approved):
LGTM

lms/views/api/error.py
seanh: I'll edit this out for now so that we can merge this. We don't know for sure yet whether we'll end up putting `"auth_url"` in the responses

PR #2050, opened 2020-08-24
Rename error.py files to exceptions.py

seanh:
We've discussed this in the past: Pyramid exception views are called exceptions views not just error views, and can catch any kind of exception and return any kind of response, not just errors.

Just implementing the name change now.


robertknight (commented):
The "h" repository still uses the name `errors.py` for its error views. Are you planning to make a similar change there? What about Via 3?

seanh:
When we get round to it we'll rename those as well yes. See the previously-written coding guideline about this: https://stackoverflow.com/c/hypothesis/questions/295

Issue #2061, opened 2020-08-27
Move all OAuth 2 API calls to config.api sub-objects with authUrls


seanh:
## Problem

The LMS app's JavaScript config object currently contains a single `authUrl` setting, which is the URL that the frontend should open in a popup window to direct the user to Canvas's authorization page to get a new access token for the user.

The frontend opens this URL when a Canvas proxy API request fails, before retrying the request.

Now that we're adding Blackboard API support there's no longer a single `authUrl` that the frontend might need to open in a popup window, there are two:

1. The Canvas API auth URL should be opened if a Canvas proxy API request fails
2. The Blackboard API auth URL should be opened if a Blackboard proxy API request fails

Technically the user can't be in Canvas and Blackboard at the same time so only one of these auth URLs is ever at play at a time, so we could still get away with the single `authUrl` setting.

This doesn't seem right though. You can imagine a future where we add support for a new cross-LMS file store that use OAuth 2. That file store could be in play at the same time as Canvas or Blackboard so we'd need two different auth URLs in the JS config.

We should fix the design to be able to handle different auth URLs for different APIs at once.

## Solution: per-endpoint `authUrl`'s in the JavaScript config

Proposed new JS config design for the sync, listFiles and viaUrl APIs:

```json
{
  "api": {
    "authToken": "Bearer ...",
    "sync": {
      "data": { ... },
      "path": "/api/canvas/sync",
      "authUrl": "https://lms.hypothes.is/api/canvas/authorize"
    },
    "viaUrl": {
      "path": "https://lms.hypothes.is/api/canvas/files/476/via_url",
      "authUrl": "https://lms.hypothes.is/api/canvas/authorize"
    }
  },
  "filePicker": {
    "canvas": {
      "listFiles": {
        "path": "/api/canvas/courses/23/files",
        "authUrl": "https://lms.hypothes.is/api/canvas/authorize"
      }
    },
    "blackboard": {
      "listFiles": {
        "path": "/api/blackboard/courses/23/files",
        "authUrl": "https://lms.hypothes.is/api/blackboard/authorize"
      }
    }
  }
}
```

Notes:

* Each API endpoint has its own `authUrl` setting. When an API request fails the frontend must open that API endpoint's `authUrl` in a popup window and then re-try.

  Of course most API endpoints have the same `authUrl` (all Canvas ones are the same; and all Blackboard ones are the same) so there will be duplicate strings in the JS config, but this is opaque to the frontend.

[...]

PR #2062, opened 2020-08-27
Make the authUrl setting per-endpoint

seanh:
Fixes https://github.com/hypothesis/lms/issues/2061

Instead of a single `authUrl` setting in the JavaScript config, change it to three `authUrl` settings: one for each OAuth 2.0 API endpoint used. Currently this is three endpoints:

1. The sync API
2. The Canvas files list API
3. The Canvas files Via URL API

This config tidy up is intended to make way for adding Blackboard API support in future.

This will break the frontend: it needs to be updated to work with these config changes.

Specific changes:

* Remove the `config.canvas.authUrl` setting from the JavaScript config.

  This has now been replaced by multiple per-endpoint `authUrl` settings, in the different API endpoint configs distributed throughout the config object.

* Also remove the `config.authUrl` setting.

  This was a duplicate of `config.canvas.authUrl` and was unused.

* Replace the `config.api.viaUrlCallback` string with a `config.api.viaUrl` object with `path` and `authUrl` strings.

* Replace the `config.filePicker.canvas.courseId` string with a `config.filePicker.canvas.listFiles` object with `path` and `authUrl` parts. The `courseId` string is now embedded in the `path` string.

* Add an `authUrl` string to the `config.api.sync` object.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -40,9 +40,12 @@ class JSConfig:
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
-        self._config["api"]["viaCallbackUrl"] = self._request.route_url(
-            "canvas_api.files.via_url", file_id=canvas_file_id
-        )
+        self._config["api"]["viaUrl"] = {
+            "authUrl": self._request.route_url("canvas_api.authorize"),
+            "path": self._request.route_path(
+                "canvas_api.files.via_url", file_id=canvas_file_id
+            ),
+        }
         self._add_canvas_speedgrader_settings(canvas_file_id=canvas_file_id)
 
     def add_document_url(self, document_url):
@@ -151,7 +154,13 @@ class JSConfig:
                 # construct these launch URLs our JavaScript code needs the
                 # base URL of our LTI launch endpoint.
                 "ltiLaunchUrl": self._request.route_url("lti_launches"),
-                "courseId": self._request.params.get("custom_canvas_course_id"),
+                "listFiles": {
+                    "authUrl": self._request.route_url("canvas_api.authorize"),
+                    "path": self._request.route_path(
+                        "canvas_api.courses.files.list",
+                        course_id=self._request.params.get("custom_canvas_course_id"),
+                    ),
+                },
             },
             "google": {
                 "clientId": self._request.registry.settings["google_client_id"],
@@ -311,14 +320,7 @@ class JSConfig:
                 # authenticate itself to the API.
                 "authToken": self._auth_token()
             },
-            # The URL that the JavaScript code will open if it needs the user to
-            # authorize us to request a new Canvas access token.
-            "authUrl": self._request.route_url("canvas_api.authorize"),
-            "canvas": {
-                # The URL that the JavaScript code will open if it needs the user to
-                # authorize us to request a new Canvas access token.
-                "authUrl": self._request.route_url("canvas_api.authorize"),
-            },
+            "canvas": {},
             # Some debug information, currently used in the Gherkin tests.
             "debug": {"tags": []},
             # Tell the JavaScript code whether we're in "dev" mode.
@@ -432,6 +434,7 @@ class JSConfig:
         req = self._request
 
         sync_api_config = {
+            "authUrl": req.route_url("canvas_api.authorize"),
             "path": req.route_path("canvas_api.sync"),
             "data": {
                 "lms": {
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -37,6 +37,9 @@ class JSConfig:
         """
         Set the document to the Canvas file with the given canvas_file_id.
 
+        This configures the frontend to make an API call to get the URL to use
+        as the src for the Via iframe.
+
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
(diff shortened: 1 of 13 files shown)


lms/resources/_js_config.py
```diff
@@ -40,9 +40,12 @@ def add_canvas_file_id(self, canvas_file_id):
...
             is missing
         """
-        self._config["api"]["viaCallbackUrl"] = self._request.route_url(
-            "canvas_api.files.via_url", file_id=canvas_file_id
-        )
+        self._config["api"]["viaUrl"] = {
+            "authUrl": self._request.route_url("canvas_api.authorize"),
+            "path": self._request.route_path(
+                "canvas_api.files.via_url", file_id=canvas_file_id
+            ),
```
jon-betts, line 47: All this stuff looks Canvas specific. Should it name-spaced accordingly? `api.viaUrl` seems a bit general.

It also seems like it might be a bit redunant. The paths we have here are:

 * `api.viaUrl.path`
 * `api.viaUrl.authUrl`

Might something like this be better?

* `api.canvas.via.url`
* `api.canvas.via.authUrl`

jon-betts (commented):
In a wonderfully charmed moment, I hit an authorise dialog on all methods of getting things from Canvas, so I can confirm this is working.

Before I get mad into the details, there are a couple of things that jump out:

 * `add_canvas_file_id()` adds settings to `api.viaURL`, but this is really `api.canvas.viaURL` or something presumably? They all look Canvas specific, or are there plans a foot here?
 * `enable_canvas_oauth2_redirect_error_mode()` also adds an auth URL, but the name is `authorizeUrl` which isn't consistent with the three new `authUrl` values

lms/resources/_js_config.py
seanh: I think the redundancy might be confusion:

* The `viaUrl` in `api.viaUrl` refers to the ["Via URL" API](https://github.com/hypothesis/lms/blob/626fecfe5942df22cb5291de1ec22472cedd2a53/lms/views/api/canvas/files.py#L23-L39), so named because it's an API that returns a `{ via_url: <VIA_URL> }` response where `<VIA_URL>` is a URL suitable for use as the `src` of the assignment's Via iframe. Currently the Via URL endpoint handles only Canvas files. In future there'll either be a second Via URL endpoint for Blackboard files or the single endpoint will be capable of handling either
* The `path` in `api.viaUrl.path` is the path _to the Via URL API_, for example `/api/canvas/files/23/via_url`. In future this will sometimes be the path to a Blackboard Via URL API instead. All API endpoints in the JS config have a `path` like this

So there's no redundancy -- the `viaUrl` and `path` refer to different things. `viaUrl` is the name of the API endpoint, `path` is the path to that endpoint. `path` is also just a path not a URL.

`viaUrl` actually _isn't_ Canvas specific. The way it works is:

* If the Via URL can be known up front then the backend puts a `config.viaUrl` setting in the JS config. Currently happens for public URL and Google Drive assignments in any LMS
* If the Via URL can't be known up front then the backend doesn't put a `config.viaUrl` setting in the JS config, but does put a `config.api.viaUrl` object instead. The frontend has to call this API to get the Via URL. Happens for Canvas files assignments and soon for Blackboard files assignments. This is because we need to call the Canvas or Blackboard API to get a temporary public URL for the file on every launch

So `config.api.viaUrl` will either contain a Canvas `authUrl` and `path` or a Blackboard `authUrl` and `path` depending. But this is opaque to the frontend. The frontend knows only `config.api.viaUrl` and needn't know whether it's Canvas or Blackboard. The backend just handles setting it appropriately and the frontend responds.

There can never be more than one `viaUrl` in play at once, so there's no need to support having two of them in different locations in the config (and doing so would complicate both the backend and frontend code)

jon-betts: Ok, so what's in there in a Canvas context is Canvas specific, but in different context you get different content in the same place? Kind of make sense.

I notice that the contents are totally different depending on whether this is set by `add_document_url()` or `add_canvas_file_id()`. I presume this is because the former is actually the URL to go to, and the later is a callback saying "go here to find out"?

If that's the case it makes sense, but some short comments to that effect would probably be good. It's not obvious. e.g.:

```
# Tell the frontend to call us back for a URL
# Give the frontend an exact URL to use for the document
```

Also `canvasOAuth2RedirectError.authorizeUrl?`

jon-betts: Also AFAIK, a path is just a non-fully qualified URL, so if we wanted to use the name it's accurate.

seanh: Added a couple of comments to those docstrings.

It's important to distinguish between "paths" and "URLs" here because various things in the JS config are just paths and the JS prepends the host to them, whereas other things are full URLs. We should probably fix this so that the backend always sends full URLs and the frontend doesn't prepend anything, at some point.

seanh: It's kind of an unrelated change so to reduce risk I've sent a separate PR to rename `authorizeUrl`, so that can be reviewed and tested separately: https://github.com/hypothesis/lms/pull/2092

jon-betts: Yep that all seems very sensible. The the front-end mucks about with it we need to mark it as one or the other. I'll take a look at the other PR.

jon-betts (approved):
The python side of this change is pretty small and sensible, so the real review here is probably the client stuff

PR #2063, opened 2020-08-27
Merge the different API permissions into one

seanh:
Merge the separate `"canvas_api"`, `"lti_outcomes"`, and `"sync_api"` permissions into a single `"api"` permission.

These permissions are all given to LTI users anyway so it doesn't make any difference, and there's really no need to give certain users access to only certain APIs: just giving LTI users access to the whole API is fine.

We're soon going to be adding Blackboard API endpoints that will need to be protected by permissions and we don't want to have to add another separare `"blackboard_api"` permission. Instead it'll just reuse this new `"api"` permission.

### DIFF
--- a/lms/resources/default.py
+++ b/lms/resources/default.py
@@ -7,9 +7,7 @@ class DefaultResource:
 
     __acl__ = [
         (Allow, "report_viewers", "view"),
-        (Allow, "lti_user", "canvas_api"),
-        (Allow, "lti_user", "lti_outcomes"),
-        (Allow, "lti_user", "sync_api"),
+        (Allow, "lti_user", "api"),
     ]
 
     def __init__(self, request):
--- a/tests/unit/lms/resources/default_test.py
+++ b/tests/unit/lms/resources/default_test.py
@@ -1,4 +1,3 @@
-import pytest
 from pyramid.authorization import ACLAuthorizationPolicy
 
 from lms.resources import DefaultResource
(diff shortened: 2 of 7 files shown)


lms/resources/default.py
```diff
@@ -7,9 +7,7 @@ class DefaultResource:
 
     __acl__ = [
         (Allow, "report_viewers", "view"),
-        (Allow, "lti_user", "canvas_api"),
-        (Allow, "lti_user", "lti_outcomes"),
-        (Allow, "lti_user", "sync_api"),
+        (Allow, "lti_user", "api"),
```
jon-betts, line 10: Yep. Seems better. If you get them all why distinguish? We can always go back and add distinctions if we ever get different user classes.

tests/unit/lms/resources/default_test.py
```diff
@@ -25,35 +24,20 @@ def test_it_doesnt_allow_others_to_view_the_reports_page(
...
         pyramid_config.testing_securitypolicy("someone_else", groupids=["others"])
         pyramid_config.set_authorization_policy(ACLAuthorizationPolicy())
 
         resource = DefaultResource(pyramid_request)
 
-        assert not pyramid_request.has_permission("canvas_api", resource)
-
-    @pytest.mark.parametrize(
-        "groupid,permission", [("lti_user", True), ("others", False)]
-    )
```
jon-betts, line 50: :+1:

jon-betts (approved):
This looks good and much simpler

PR #2064, opened 2020-08-27
Rename CanvasOAuth2RedirectResource to OAuth2RedirectResource

seanh:
There's nothing Canvas-specific about this. It only has the Canvas-specific name because it's only used in Canvas currently, but I'm pretty sure we're gonna end up using this for Blackboard OAuth 2 redirects as well.

The only reason this view exists is because the `"canvas_oauth_callback"` and `"canvas_api.authorize"` routes that use it (the two routes for the authorization popup window) need `context.js_config` to be set (because they use the Preact app for the error dialog that's shown if we receive a bad redirect from Canvas). They can't use the default resource because it doesn't have `context.js_config`. And `context.js_config` can't be added to the default resource because that resource is used by a bunch of views that don't and can't have `context.js_config`.

### DIFF
--- a/lms/resources/__init__.py
+++ b/lms/resources/__init__.py
@@ -8,6 +8,6 @@ See the traversal-related sections in the Pyramid docs:
 * https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/traversal.html
 * https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hybrid.html
 """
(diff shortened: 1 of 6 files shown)


jon-betts (approved):
This is a straight name change, looks fine to me. 

Even if it was Canvas specific, this would still be fine, as it's "Canvas-y" by context of being in a file called canvas. I.e. we don't need the full path in every class name

seanh:
Rebased

PR #2065, opened 2020-08-27
Rename the OAuth 2 templates

seanh:
There's nothing Canvas-specific about these. They just have Canvas-specific names because they're only currently used in Canvas. But they're soon going to be reused in Blackboard.


jon-betts (approved):
Seems very sensible. 

I checked the templates to see if they had anything canvas specific (naming etc), but there's nothing. Like really nothing, which leaves me a bit confused, but it's not due to this.

PR #2066, opened 2020-08-28
Remove unnecessary UI component mocks

robertknight:
**nb. This may conflict with https://github.com/hypothesis/lms/pull/2062, so I suggest that PR is merged first.**

Remove the `FakeFileList` and `FakeButton` component mocks in the `LMSFilePicker` tests. Mocking
these imports is already handled by `mockImportedComponents`.


LMS007 (approved):
👍  gtg Rob

PR #2074, opened 2020-08-31
Remove unneeded unicode_literals imports

seanh:
These are from Python 2.


jon-betts (approved):
Not much to say. All tests passing and a little less cruft :+1:

PR #2075, opened 2020-08-31
Remove an unnecessary test

seanh:
This test isn't adding anything -- the code in `routes.py` is entirely declarative so there's not much point in testing it. You could test that it at least doesn't crash given a real Pyramid config object, but this test uses a non-autospecced mock so it doesn't even do that. And in fact `routes.py` is already being executed as part of the `pyramid_config` fixture, so that
is already tests that it runs without crashing given a real config object:

https://github.com/hypothesis/lms/blob/d4ddd02113ae4e4e544eb41897106bb4be0293f6/tests/unit/conftest.py#L105

Hence why we still have 100% coverage after removing this test.

That also indirectly tests that at least some of the routes are set up correctly, since that fixture is what makes `request.route_url()` and `request.route_path()` calls work correctly in the tests.


jon-betts (commented):
I'm really not sure on this one. I think there a number of things we'd like to know are true:

 1. The route exists and relates to a view
 2. The route has the right URL
 3. The route and view have all the expected settings and parameters etc.
 4. The code and route work together to get the right thing done

Without this test, I think the unit-tests will have line coverage, but won't make any assertions about what's in the lines to give us any of the above (apart from in some cases where views call other views by name).

For example you can comment out most of the routes and the unit tests still work fine:

```
    # config.add_route("index", "/")
    # config.add_route("feature_flags_test", "/flags/test")
    # config.add_route("welcome", "/welcome")
    # config.add_route("assets", "/assets/*subpath") 
    # config.add_route("status", "/_status")
    # config.add_route("favicon", "/favicon.ico")
   ...
```

The functests catch some of this, as we can tell the app doesn't start without all the routes matching the code: "No route named welcome found for view registration". This covers (1), but I don't think that cares what URL the route resolves to (2), just that the decorators and names line up. 

The functests and BDD tests will go a good way to covering (2) - (4) where they apply, but I think they are patchy at best. If we had full functional testing coverage this would definitely cover it, but I don't think we do.

The current test gives us (1), (2) and the half of (3) that appears in the routing side, rather than the view side, which is better than nothing.

seanh:
As discussed gonna merge this for now but we'll have a go at writing functional tests for the new Blackboard API routes and views that we're going to be adding. If we do the work of creating the necessary fixtures etc for the new Blackboard API functests, it'll then be much less work to add functests retrospectively for the Canvas API and other views as well

PR #2088, opened 2020-09-02
Isolate TokenStore in unit tests

seanh:
Make it so that the only unit tests that instantiate a `TokenStore` object are the ones in `_token_store_test.py`. All others use a an autospecced mock `TokenStore`.

This is so that `TokenStore` can be moved from `lms.services.canvas_api._token_store.TokenStore` to a top-level service (`lms.services.token_store.TokenStore`) so that it can be shared by `CanvasAPIClient`, the upcoming `BlackboardAPIClient`, and possibly others in future.

When `TokenStore` is a top-level service it no longer makes sense for unit tests to consider it a close collaborator of `CanvasAPIClient` and not mock it.

`TokenStore` has a very simple API: it has only `save(access_token, refresh_token, expires_in)` and `get()` (no arguments) methods so it's very simple to mock and to assert that users like `CanvasAPIClient` and `BlackboardAPIClient` use it correctly.

The test isolation will also make the actual change to move `TokenStore` to a top-level service easier to make: the resulting test failures will be easier to understand and the tests will be easier to update.

### DIFF
--- a/tests/unit/lms/services/canvas_api/_authenticated_test.py
+++ b/tests/unit/lms/services/canvas_api/_authenticated_test.py
@@ -6,8 +6,6 @@ from h_matchers import Any
 from lms.services import CanvasAPIAccessTokenError, CanvasAPIServerError
 from lms.services.canvas_api._authenticated import TokenResponseSchema
 from lms.services.canvas_api._basic import BasicClient
-from lms.services.canvas_api._token_store import TokenStore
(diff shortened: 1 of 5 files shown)


jon-betts (approved):
This all looks pretty sensible to me. It looks like we weren't particularly relying on the object being the object in the tests anyway. :+1:

PR #2089, opened 2020-09-02
Make canvas_api._token_store a top-level service

seanh:
Move `services.canvas_api._token_store.TokenStore` to a top-level service (`services.token_store.TokenStore`) and register it with pyramid_services as the `"token_store"` service.

This is because `TokenStore` isn't actually Canvas-specific: it's just a get-and-set service for `models.OAuth2Token`, which is a generic model for storing OAuth 2 tokens. `TokenStore` is used by `CanvasAPIClient` and will soon be used by the upcoming `BlackboardAPIClient` and possibly others in future.

### DIFF
--- a/lms/services/canvas_api/factory.py
+++ b/lms/services/canvas_api/factory.py
@@ -2,7 +2,6 @@ from urllib.parse import urlparse
 
 from lms.services.canvas_api._authenticated import AuthenticatedClient
 from lms.services.canvas_api._basic import BasicClient
-from lms.services.canvas_api._token_store import TokenStore
 from lms.services.canvas_api.client import CanvasAPIClient
 
 
@@ -19,15 +18,9 @@ def canvas_api_client_factory(_context, request):
     canvas_host = urlparse(ai_getter.lms_url()).netloc
     basic_client = BasicClient(canvas_host)
 
-    token_store = TokenStore(
-        consumer_key=request.lti_user.oauth_consumer_key,
-        user_id=request.lti_user.user_id,
-        db=request.db,
-    )
-
     authenticated_api = AuthenticatedClient(
         basic_client=basic_client,
-        token_store=token_store,
+        token_store=request.find_service(name="token_store"),
         client_id=ai_getter.developer_key(),
         client_secret=ai_getter.developer_secret(),
         redirect_uri=request.route_url("canvas_oauth_callback"),
--- a/lms/services/token_store.py
+++ b/lms/services/token_store.py
@@ -1,5 +1,3 @@
-"""Access to OAuth2Tokens in the DB."""
-
 import datetime
 
 from sqlalchemy.orm.exc import NoResultFound
@@ -13,11 +11,11 @@ class TokenStore:
 
     def __init__(self, db, consumer_key, user_id):
         """
-        Construct a new TokenStore.
+        Return a new TokenStore.
 
-        :param db: An SQLAlchemy session
-        :param consumer_key: Consumer key to retrieve and save against
-        :param user_id: User id to retrieve and save against
+        :param db: the SQLAlchemy session
+        :param consumer_key: the LTI consumer key to use for tokens
+        :param user_id: the LTI user ID to user for tokens
         """
         self._db = db
         self._consumer_key = consumer_key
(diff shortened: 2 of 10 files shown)


lms/services/canvas_api/factory.py
```diff
@@ -19,15 +18,9 @@ def canvas_api_client_factory(_context, request):
     canvas_host = urlparse(ai_getter.lms_url()).netloc
     basic_client = BasicClient(canvas_host)
 
-    token_store = TokenStore(
-        consumer_key=request.lti_user.oauth_consumer_key,
-        user_id=request.lti_user.user_id,
-        db=request.db,
-    )
```
seanh, line 26: The factories for services like `CanvasAPIClient` and in future `BlackboardAPIClient` don't need to know how to create a `TokenStore` or what `TokenStore`'s dependencies are. They just find it with `request.find_service(name="token_store")`. This is nice because we don't end up with multiple places in the code where a `TokenStore` is created and its dependencies injected. `token_store_service_factory()` is the only place in the code where a `TokenStore` is created

lms/services/token_store.py
```diff
@@ -61,3 +59,9 @@ def get(self):
                 explanation="We don't have a Canvas API access token for this user",
                 response=None,
             ) from err
+
+
+def token_store_service_factory(_context, request):
+    return TokenStore(
+        request.db, request.lti_user.oauth_consumer_key, request.lti_user.user_id
+    )
```
seanh, line 67: It's nice that the code for creating a `TokenStore` and injecting its dependencies lives alongside `TokenStore` itself in `token_store.py`

tests/unit/conftest.py
```diff
@@ -245,6 +246,14 @@ def oauth1_service(pyramid_config):
     return oauth1_service
 
 
+@pytest.fixture
+def token_store_service(oauth_token, pyramid_config):
+    token_store_service = mock.create_autospec(TokenStore, instance=True, spec_set=True)
+    token_store_service.get.return_value = oauth_token
+    pyramid_config.register_service(token_store_service, name="token_store")
+    return token_store_service
```
seanh, line 254: Centralised service fixture, as we have for all the services. This means we only need one of these fixtures and tests for all code that uses `TokenStore` can use this fixture

tests/unit/conftest.py
```diff
@@ -291,3 +300,22 @@ def httpretty_():
...
+@pytest.fixture
+def lti_user(pyramid_request):
+    return pyramid_request.lti_user
+
+
+@pytest.fixture
+def oauth_token(lti_user, application_instance):
+    return factories.OAuth2Token(
+        user_id=lti_user.user_id, application_instance=application_instance
+    )
```
seanh, line 321: These three got moved up from `canvas_api/conftest.py` because the `token_store_service()` fixture needs them

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -72,7 +72,7 @@ def test_send_raises_CanvasAPIAccessTokenError_if_it_cannot_refresh(
             )
 
     def test_get_token(
-        self, authenticated_client, basic_client, token_store, token_response
+        self, authenticated_client, basic_client, token_store_service, token_response
```
seanh, line 75: Lots of renaming the `token_store` fixture to `token_store_service` in tests. But `token_store_service` was consistent with how our other service fixtures are named

tests/unit/lms/services/canvas_api/factory_test.py
```diff
@@ -5,7 +5,6 @@
 from lms.services.canvas_api.factory import canvas_api_client_factory
 
 
-@pytest.mark.usefixtures("ai_getter")
```
seanh, line 8: This got moved down below

tests/unit/lms/services/canvas_api/factory_test.py
```diff
@@ -48,14 +43,13 @@ def test_building_the_AuthenticatedClient(
...
     @pytest.fixture(autouse=True)
     def AuthenticatedClient(self, patch):
         return patch("lms.services.canvas_api.factory.AuthenticatedClient")
 
     @pytest.fixture(autouse=True)
     def CanvasAPIClient(self, patch):
         return patch("lms.services.canvas_api.factory.CanvasAPIClient")
+
+
+pytestmark = pytest.mark.usefixtures("ai_getter", "token_store_service")
```
seanh, line 55: Let's use these module level `pytestmark`'s for activating service fixtures in a test module. It's what all our other test modules use to activate other service fixtures, and it activates it for the entire test module not just one class or test like the `@pytest.mark.usefixtures` decorator. It may seem more efficient to try to activate the fixture only for those tests that need it, but in practice this saves little, adds noise  and easily leads to mistakes where a test that should use the fixture doesn't. It's just easier to slap one of these at the bottom of every test file that needs services

jon-betts (approved):
Yep this looks fine. I'm glad the refactoring of all the Canvas stuff has some side effects other than the immediate break up too. It was quite a lot of work, but I think this shows why it's a good idea to do that sort of thing.

Previously all of this was just mixed in with the rest. There was no separate object to re-use. I suspect we might find we can do similar things with more of it. I think in one of your POC's you re-used one of the lower client layers too.

PR #2090, opened 2020-09-02
Rename TokenStore to OAuth2TokenService

seanh:
Just to make it clearer what it actually is: it's a service wrapper for `models.OAuth2Token`. Getting "OAuth 2" in there is particularly important

### DIFF
--- a/lms/services/oauth2_token.py
+++ b/lms/services/oauth2_token.py
@@ -6,7 +6,7 @@ from lms.models import OAuth2Token
 from lms.services import CanvasAPIAccessTokenError
 
 
-class TokenStore:
+class OAuth2TokenService:
     """Save and retrieve OAuth2Tokens from the DB."""
 
     def __init__(self, db, consumer_key, user_id):
@@ -61,7 +61,7 @@ class TokenStore:
             ) from err
 
 
-def token_store_service_factory(_context, request):
-    return TokenStore(
+def oauth2_token_service_factory(_context, request):
+    return OAuth2TokenService(
         request.db, request.lti_user.oauth_consumer_key, request.lti_user.user_id
     )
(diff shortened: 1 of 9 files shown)


lms/services/oauth2_token.py
```diff
@@ -61,7 +61,7 @@ def get(self):
             ) from err
 
 
-def token_store_service_factory(_context, request):
-    return TokenStore(
+def oauth2_token_service_factory(_context, request):
+    return OAuth2TokenService(
```
jon-betts, line 65: This is totally fine, but I wonder if it's a bit titularly redundant. Presumably most of our services are in separate files from each other so what service it's a factory is for is clear, the fact it's a service factory rather than another kind is also pretty clear. When we refer to this in the config above it's:

 * `lms.services.oauth2_token.oauth2_token_service_factory`

I feel like we are repeating ourselves a lot, where this might do:

 * `lms.services.oauth2_token.factory`

Read backwards it's a factory, for the oauth2_token, service in LMS.

tests/unit/lms/services/oauth2_token_test.py
```diff
@@ -31,16 +31,16 @@ def test_save(self, token_store, db_session, application_instance, lti_user):
             }
         )
 
-    def test_get_returns_token_when_present(self, token_store, oauth_token):
-        result = token_store.get()
+    def test_get_returns_token_when_present(self, svc, oauth_token):
+        result = svc.get()
```
jon-betts, line 35: Our tests are pretty random about whether svc or the full name is prefered. I think either is fine honestly, but `svc` feels a bit pointlessly short.

jon-betts (approved):
I think this is good, but it did make me think about the factory names we have. I think they repeat information that's available in context and we could probably just get away with calling them `factory()` or something.

lms/services/oauth2_token.py
seanh: Yeah I think they could all just be called `factory()`

PR #2091, opened 2020-09-02
Rename AuthenticatedClient._token_store to oauth2_token_service

seanh:
I think it's a little weird for the name of the `"oauth2_token"` service to change to something completely different (`token_store`) when it gets passed to `AuthenticatedClient`. Makes the code hard to follow / makes it hard to know what `token_store` actually is. So rename `token_store` to `oauth2_token_service`. This is consistent with how references to other services are named, elsewhere in the code (`*_svc` is sometimes used but `*_service` is more common).


jon-betts (approved):
:+1: Very to the point

PR #2096, opened 2020-09-03
Don't raise CanvasAPIAccessTokenError from OAuth2TokenService

seanh:
It doesn't make sense to raise a Canvas-specific error from this generic OAuth 2 service. Add a new `NoOAuth2Token` exception and raise that instead.

Catch `NoOAuth2Token` in the (Canvas-specific) `AuthenticatedClient` that calls `OAuth2TokenService` and wrap it in `CanvasAPIAccessTokenError`, so the rest of the Canvas API services and views are unaffected

### DIFF
--- a/lms/services/canvas_api/_authenticated.py
+++ b/lms/services/canvas_api/_authenticated.py
@@ -3,7 +3,7 @@
 import marshmallow
 from marshmallow import fields
 
-from lms.services import CanvasAPIAccessTokenError
+from lms.services import CanvasAPIAccessTokenError, NoOAuth2Token
 from lms.validation import RequestsResponseSchema
 
 
@@ -57,16 +57,31 @@ class AuthenticatedClient:
         call_args = (method, path, schema, params)
 
         try:
-            auth_header = f"Bearer {self._oauth2_token_service.get().access_token}"
-            return self._client.send(*call_args, headers={"Authorization": auth_header})
+            oauth2_token = self._oauth2_token_service.get()
+        except NoOAuth2Token as err:
+            raise CanvasAPIAccessTokenError(
+                explanation="We don't have a Canvas API access token for this user",
+                response=None,
+            ) from err
 
+        access_token = oauth2_token.access_token
+        refresh_token = oauth2_token.refresh_token
+
+        try:
+            return self._client.send(
+                *call_args,
+                headers={"Authorization": f"Bearer {access_token}"},
+            )
         except CanvasAPIAccessTokenError:
-            refresh_token = self._oauth2_token_service.get().refresh_token
             if not refresh_token:
                 raise
 
-            auth_header = f"Bearer {self.get_refreshed_token(refresh_token)}"
-            return self._client.send(*call_args, headers={"Authorization": auth_header})
+        new_access_token = self.get_refreshed_token(refresh_token)
+
+        return self._client.send(
+            *call_args,
+            headers={"Authorization": f"Bearer {new_access_token}"},
+        )
 
     def get_token(self, authorization_code):
         """
--- a/lms/services/oauth2_token.py
+++ b/lms/services/oauth2_token.py
@@ -3,7 +3,7 @@ import datetime
 from sqlalchemy.orm.exc import NoResultFound
 
 from lms.models import OAuth2Token
-from lms.services import CanvasAPIAccessTokenError
+from lms.services import NoOAuth2Token
 
 
 class OAuth2TokenService:
@@ -31,7 +31,7 @@ class OAuth2TokenService:
         """
         try:
             oauth2_token = self.get()
-        except CanvasAPIAccessTokenError:
+        except NoOAuth2Token:
             oauth2_token = OAuth2Token(
                 consumer_key=self._consumer_key, user_id=self._user_id
             )
@@ -46,7 +46,7 @@ class OAuth2TokenService:
         """
         Return the user's saved OAuth 2 token from the DB.
 
-        :raise CanvasAPIAccessTokenError: if we don't have an access token for the user
+        :raise NoOAuth2Token: if we don't have an OAuth 2 token for the user
         """
         try:
             return (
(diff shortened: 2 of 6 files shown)


lms/services/canvas_api/_authenticated.py
```diff
@@ -57,16 +57,31 @@ def send(self, method, path, schema, params=None):
...
                 raise
 
-            auth_header = f"Bearer {self.get_refreshed_token(refresh_token)}"
-            return self._client.send(*call_args, headers={"Authorization": auth_header})
+            new_access_token = self.get_refreshed_token(refresh_token)
+
+            return self._client.send(
+                *call_args,
+                headers={"Authorization": f"Bearer {new_access_token}"},
+            )
```
seanh, line 84: I rewrote this method a little to make it clearer what it does:

1. It gets the `OAuth2Token` from the DB
2. If there's no `OAuth2Token` in the DB (`NoOAuth2Token`) it raises `CanvasAPIAccessTokenError`
3. It sends the request
4. If the request raises `CanvasAPIAccessTokenError` (e.g. our access token was deleted from Canvas, or is expired) it:
   1. Just raises the `CanvasAPIAccessTokenError` if we don't have a refresh token
   2. Gets a new access token if we do have a refresh token
   3. Re-sends the request with the new access token

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -24,6 +24,21 @@ def test_send(self, authenticated_client, basic_client, oauth_token):
...
+
+        with pytest.raises(
+            CanvasAPIAccessTokenError,
+            match="^We don't have a Canvas API access token for this user$",
+        ) as exc_info:
+            authenticated_client.send(
+                "METHOD", "/path", sentinel.schema, sentinel.params
+            )
+
+        assert exc_info.value.response is None
```
seanh, line 40: This was actually a missing test previously: there were two different ways that `CanvasAPIAccessTokenError` could be raised, both grouped under the same `try`, and only one of them had a test

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -24,6 +24,21 @@ def test_send(self, authenticated_client, basic_client, oauth_token):
...
         assert result == basic_client.send.return_value
 
+    def test_send_raises_CanvasAPIAccessTokenError_if_we_dont_have_an_access_token_for_the_user(
+        self, authenticated_client, oauth2_token_service
+    ):
+        oauth2_token_service.get.side_effect = NoOAuth2Token()
+
+        with pytest.raises(
+            CanvasAPIAccessTokenError,
+            match="^We don't have a Canvas API access token for this user$",
```
seanh, line 34: There are multiple ways that the method under test can raise `CanvasAPIAccessTokenError`, so I'm making sure that it raises the one that the test expects it to raise. Otherwise the code or test could be wrong and it could still pass

lms/services/canvas_api/_authenticated.py
```diff
@@ -57,16 +57,31 @@ def send(self, method, path, schema, params=None):
...
-            refresh_token = self._oauth2_token_service.get().refresh_token
+        access_token = oauth2_token.access_token
+        refresh_token = oauth2_token.refresh_token
+
+        try:
+            return self._client.send(
+                *call_args,
+                headers={"Authorization": f"Bearer {access_token}"},
+            )
+        except CanvasAPIAccessTokenError as err:
```
jon-betts, line 75:
```suggestion
        except CanvasAPIAccessTokenError:
```

lms/services/canvas_api/_authenticated.py
jon-betts: Not sure it makes much difference but we could out-dent the stuff after the raise to take advantage of the fast return for less nested code.

lms/services/oauth2_token.py
```diff
@@ -55,10 +55,7 @@ def get(self):
                 .one()
             )
         except NoResultFound as err:
-            raise CanvasAPIAccessTokenError(
-                explanation="We don't have a Canvas API access token for this user",
-                response=None,
-            ) from err
+            raise NoOAuth2Token() from err
```
jon-betts, line 58: It would be good if this had something here, otherwise it'll stringify to an empty string, which can be confusing.

jon-betts (commented):
This looks good, but the more I look at it, the less all of the `Canvas...Error` classes look like they have anything to do with Canvas at all.

If we transmuted somethings:

 * `CanvasAPIError` -> `ExternalAPIError`
 * `CanvasAPIAccessTokenError` -> `ExternalAPIAccessTokenError`
 * ...

This all looks quite generic (which solves some of the problems we were having in the chat). I don't think there's much that's canvas specific in `raise_from()` apart from the class names.

So, I think this PR is fine, but I also wonder if making these generic would free us up to do this differently. Honestly I think it might be relatively similar, as the `NoOAuthToken` exception is because we don't have one in the DB, this only becomes an `ExternalAPIAccessTokenError` in the context of being called from Canvas etc. So it would probably end-up similar.

lms/services/canvas_api/_authenticated.py
jon-betts: The more I look at `raise_from` the less I like it. This is accepting a soup of different types and duck typing the lot. I think it would be easier to pick apart if we could specify what types we are expecting, so a few more `if isinstance(...` and then pull bits off.

jon-betts: Yeah, this is dumb. It's only called in three places all inside _send_prepared. I think we're taking disparate situations, crushing them together, and then pulling them back apart again needlessly.

I think this might be worth a few refactoring PRs to try and knock back into a more sensible shape. I suspect a lot of this code would just dissappear. Some ideas:

 * Stop the duck typing stuff:
    * We are actually interested in `messages` if it's a `ValidationError`
    * `response` if it's a `pyramid.HTTPException` (which `ValidationError` is too)
 * Move the decision of what exception to raise into the basic client
 * Once the Canvas specific stuff is out move the logic for to `ExternalRequestError`

PR #2097, opened 2020-09-03
Tidy up some routes

seanh:
This is to make way for the Blackboard API routes to be added:

* Consistenly use `"canvas_api.*"` for Canvas API route names
* Consistently use `"canvas_api.oauth.*"` for Canvas API authorization route names
* Put the `"authorize"` route before the `"callback"` route because that's the order they get called in

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -24,14 +24,15 @@ def includeme(config):
         "/content_item_selection",
         factory="lms.resources.LTILaunchResource",
     )
-    config.add_route(
-        "canvas_api.oauth.callback",
-        "/canvas_oauth_callback",
-        factory="lms.resources.OAuth2RedirectResource",
-    )
+
     config.add_route(
         "canvas_api.oauth.authorize",
         "/api/canvas/oauth/authorize",
+        factory="lms.resources.OAuth2RedirectResource",
+    )
+    config.add_route(
+        "canvas_api.oauth.callback",
+        "/canvas_oauth_callback",
         factory="lms.resources.OAuth2RedirectResource",
     )
     config.add_route(
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -31,7 +31,7 @@ def includeme(config):
         factory="lms.resources.OAuth2RedirectResource",
     )
     config.add_route(
-        # Unfortunately we can't easily fix this URL to match the others as its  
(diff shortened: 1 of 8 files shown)


lms/routes.py
```diff
@@ -24,14 +24,15 @@ def includeme(config):
         "/content_item_selection",
         factory="lms.resources.LTILaunchResource",
     )
+
     config.add_route(
-        "canvas_oauth_callback",
-        "/canvas_oauth_callback",
+        "canvas_api.oauth.authorize",
+        "/api/canvas/oauth/authorize",
```
seanh, line 30: I fixed the URL on this one, which we can do since this is a URL that our own app opens itself

lms/routes.py
```diff
@@ -24,14 +24,15 @@ def includeme(config):
...
         factory="lms.resources.OAuth2RedirectResource",
     )
     config.add_route(
-        "canvas_api.authorize",
-        "/api/canvas/authorize",
+        "canvas_api.oauth.callback",
+        "/canvas_oauth_callback",
```
seanh, line 35: Unfortunately we can't easily fix this URL 'cos its been given out to Canvas instances

lms/routes.py
```diff
@@ -24,14 +24,15 @@ def includeme(config):
...
-        "canvas_oauth_callback",
-        "/canvas_oauth_callback",
+        "canvas_api.oauth.authorize",
+        "/api/canvas/oauth/authorize",
         factory="lms.resources.OAuth2RedirectResource",
     )
     config.add_route(
-        "canvas_api.authorize",
-        "/api/canvas/authorize",
+        "canvas_api.oauth.callback",
```
jon-betts, line 34:
```suggestion
        # Unfortunately we can't easily fix this URL to match the others as its been given 
        # out to Canvas instances
        "canvas_api.oauth.callback",
```

lms/routes.py
jon-betts: I was immediately thinking about a migration path, but presumably this would involve adding two paths or a redirect and asking customers to update settings.

At which point a customer could quite legitimately ask: "What's in it for me?" "Nothing"

jon-betts (approved):
Good to tidy up the naming. Especially if we are adding a bunch more in a similar place

PR #2098, opened 2020-09-03
Disable invalid-name warnings in tests

seanh:
As we've already done for argument, function, method and class names. This PyLint rule doesn't fit with our usual naming style in tests


jon-betts (approved):
I think this is safe. We're not going to start camel casing everything

PR #2106, opened 2020-09-07
Install pip-tools in each venv

seanh:
Remove the separate `pipcompile` virtualenv and install pip-tools in
each individual venv separately instead. The pip-tools docs say:

> `pip-compile` should be run from the same virtual environment as your project so conditional
> dependencies that require a specific Python version, or other environment markers, resolve
> relative to your project's environment.

So get tox to add `pip-tools` to every env after installing the env's
requirements file.

This means that the commands to manage Python dependencies are now:

To add a new dependency or to remove a dependency add the package name
to or remove the package name from the appropriate
`requirements/<ENV>.in` then run:

    tox -e <ENV> --run-command 'pip-compile requirements/<ENV>.in'

This will modify the corresponding `requirements/<ENV>.txt` file. Commit
these changes to git and send a pull request.

To upgrade a single package to the latest version run:

    tox -e <ENV> --run-command 'pip-compile --upgrade-package <PACKAGE> requirements/<ENV>.in'

To upgrade or downgrade a single package to a specific version run:

    tox -e <ENV> --run-command 'pip-compile --upgrade-package "<PACKAGE>==<VERSION>" requirements/<ENV>.in'


jon-betts (approved):
Not much to say, pip-compile is gone

PR #2111, opened 2020-09-08
Fix absolute path in requirements/dev.txt

seanh:
If you compile the dev requirements:

    tox -e dev --run-command 'pip-compile requirements/dev.in'

you'll see that it actually modifies `dev.txt` and adds an absolute path:

```diff
diff --git a/requirements/dev.txt b/requirements/dev.txt
index 772739b0..2732dc83 100644
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile requirements/dev.in
 #
--e .                      # via -r requirements/dev.in
+-e file:///home/seanh/Projects/lms  # via -r requirements/dev.in
 alembic==1.4.2            # via -r requirements/requirements.txt
 attrs==20.2.0             # via -r requirements/requirements.txt, jsonschema
 backcall==0.2.0           # via ipython
```

This seems to be an issue with pip-compile, fix it using this workaround: https://github.com/jazzband/pip-tools/issues/204

PR #2112, opened 2020-09-08
Remove `-e .` from requirements

seanh:
This seems to be unnecessary and causes problems.

See:

https://github.com/hypothesis/lms/pull/2111
https://github.com/hypothesis/via3/pull/262

### DIFF
--- a/requirements/dev.in
+++ b/requirements/dev.in
@@ -1,5 +1,4 @@
 factory-boy
 pyramid_ipython
 supervisor
--e .
 -r requirements.txt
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
Yah this looks good. I rebased it as penance for not looking at it earlier and letting it go stale.

I think the -e may be cargo cult from the packages where it's neccessary because we purposefully break the path by nesting the code in 'src'. This means it does serve a purpose there, and may therefore potentially cause trubs if we try and replicate this there.

PR #2146, opened 2020-09-24
Show explanation if no files are found

### LINKED ISSUE #765: Show an explanation when a Canvas course has no files
If you haven't uploaded any files to your Canvas course yet then trying to create a Canvas file assignment gives an empty table:

Screenshot from 2019-07-05 15-16-52

When this happens we should show a helpful error message in place of the table, telling the user that they need to add files to Canvas first, and giving them a <kbd>Try again</kbd> button.

esanzgar:
Resolves #765

image

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FileList.js
+++ b/lms/static/scripts/frontend_apps/components/FileList.js
@@ -6,6 +6,7 @@ import Table from './Table';
 
 /**
  * @typedef {import('../api-types').File} File
+ * @typedef {import("preact").VNode} VNode
  */
 
 /**
@@ -17,6 +18,7 @@ import Table from './Table';
  *   Callback invoked when the user clicks on a file
  * @prop {(f: File) => any} onUseFile -
  *   Callback invoked when the user double-clicks a file to indicate that they want to use it
+ * @prop {VNode} [noFiles] - component displayed when no files are available
  */
 
 /**
@@ -30,6 +32,7 @@ export default function FileList({
   selectedFile,
   onSelectFile,
   onUseFile,
+  noFiles,
 }) {
   const formatDate = isoString => new Date(isoString).toLocaleDateString();
   const columns = [
@@ -71,6 +74,7 @@ export default function FileList({
           </Fragment>
         )}
       />
+      {!isLoading && files.length === 0 && noFiles}
       {isLoading && <Spinner className="FileList__spinner" />}
     </div>
   );
--- a/lms/static/scripts/frontend_apps/components/FileList.js
+++ b/lms/static/scripts/frontend_apps/components/FileList.js
@@ -6,7 +6,7 @@ import Table from './Table';
 
 /**
  * @typedef {import('../api-types').File} File
- * @typedef {import("preact").VNode} VNode
+ * @typedef {import("preact").ComponentChildren} Children
  */
 
 /**
@@ -18,7 +18,7 @@ import Table from './Table';
  *   Callback invoked when the user clicks on a file
  * @prop {(f: File) => any} onUseFile -
  *   Callback invoked when the user double-clicks a file to indicate that they want to use it
- * @prop {VNode} [noFiles] - component displayed when no files are available
+ * @prop {Children} [noFilesMessage] - component displayed when no files are available
  */
 
 /**
@@ -32,7 +32,7 @@ export default function FileList({
   selectedFile,
   onSelectFile,
   onUseFile,
-  noFiles,
+  noFilesMessage,
 }) {
   const formatDate = isoString => new Date(isoString).toLocaleDateString();
   const columns = [
@@ -74,7 +74,7 @@ export default function FileList({
           </Fragment>
         )}
       />
-      {!isLoading && files.length === 0 && noFiles}
+      {!isLoading && files.length === 0 && noFilesMessage}
       {isLoading && <Spinner className="FileList__spinner" />}
     </div>
   );
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
@@ -27,6 +27,8 @@ import FileList from './FileList';
 /**
  * @typedef DialogState
  * @prop {'fetching'|'fetched'|'authorizing'|'error'} state
+ * @prop {string} title - Dialog title
+ * @prop {'select'|'authorize'|'authorize_retry'|'retry'|'none'} continueAction - Action for the continue button
  * @prop {File[]|null} files - List of fetched files
  * @prop {Error|null} error - Details of current error, if `state` is 'error'
  */
@@ -34,6 +36,8 @@ import FileList from './FileList';
 /** @type {DialogState} */
 const INITIAL_DIALOG_STATE = {
   state: 'fetching',
+  title: 'Select a file',
+  continueAction: 'select',
   files: null,
   error: null,
 };
@@ -41,7 +45,7 @@ const INITIAL_DIALOG_STATE = {
 const CanvasNoFiles = (
   <Fragment>
     <div className="FileList__no-files">
-      There are no supported files in the current Canvas course.
+      There are no available PDF files in the current course.
     </div>
     <p>
       Please see the Canvas help article for{' '}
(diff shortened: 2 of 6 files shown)


esanzgar:
@klemay when you know how the message should look like let me know.

klemay:
@esanzgar suggested copy: 

There are no supported files in the current Canvas course. Please see this Canvas help article for [uploading files to a course](https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-upload-a-file-to-a-course/ta-p/618) and try again.

esanzgar:
@klemay 

Is this acceptable?

image

esanzgar:
Latest look and feel:
image

lms/static/styles/components/_FileList.scss
```diff
@@ -58,3 +60,11 @@ $file-list-icon-size: 24px;
 .FileList__date-col {
   padding-right: 10px;
 }
+
+.FileList__no-files {
+  border: 1px solid var.$grey-6;
+  height: 40px;
+  vertical-align: middle;
+  padding: 12px;
+  background-color: #eee;
```
robertknight, line 69: Perhaps use `var.$grey-2` here (#ececec) for consistency with the disabled buttons in the form. Using color variables for foreground and background colors also makes it easier to check that contrast is high enough everywhere.

lms/static/scripts/frontend_apps/components/FileList.js
```diff
@@ -71,6 +71,24 @@ export default function FileList({
...
+            Please see the Canvas help article for{' '}
+            <a
+              href="https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-upload-a-file-to-a-course/ta-p/618"
+              target="_blank"
+              rel="noreferrer"
+            >
+              uploading files to a course
+            </a>{' '}
+            and try again.
+          </p>
```
robertknight, line 89: The text looks good. This text is specific to a particular LMS whereas the `FileList` component itself is currently not. One change that we plan to make soon is to re-use this component with other LMSes such as Blackboard.

What I would suggest is that we keep the basic structure here, but pass in the content of the `<div>` and the `<p>` as props from the parent `LMSFilePicker` component. The parent can then pass in the appropriate message text depending on the LMS.

robertknight (changes requested):
Hi Eduardo,

Visually it looks good. One change I would like to make to prepare for a feature that we plan to work on soon is to keep the `FileList` component agnostic of the current LMS and pass in the LMS-specific messages as props from the parent `LMSFilePicker` component.

lms/static/scripts/frontend_apps/components/test/FileList-test.js
```diff
@@ -60,6 +63,11 @@ describe('FileList', () => {
     assert.isFalse(wrapper.exists('.FileList__spinner'));
   });
 
+  it('renders a explanatory message when ', () => {
```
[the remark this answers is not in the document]
esanzgar: Done

lms/static/styles/components/_FileList.scss
```diff
@@ -38,7 +40,7 @@ $file-list-icon-size: 24px;
   &.is-selected {
     // nb. Argument must be passed to `invert` even though "1" is the default
     // to work around https://bugs.chromium.org/p/chromium/issues/detail?id=964696.
-    filter: #{"invert(1)"}; // Quoted to avoid conflict with SASS mixin.
+    filter: #{'invert(1)'}; // Quoted to avoid conflict with SASS mixin.
```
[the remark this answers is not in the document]
esanzgar: Unintentional.

Because .prettierrc has `"singleQuote": true` it changed the quote style when saving the file.

Is this something I should avoid in the future?

lms/static/styles/components/_FileList.scss
esanzgar: I was making it consistent with the look and feel of the zebra-stripe rows in the normal table (https://github.com/hypothesis/lms/blob/c0fa85c9a79186b3efda91e82a15ce5fb763350b/lms/static/styles/lms.css#L193).

Sure, I will make it `var.$grey-2`

lms/static/styles/components/_FileList.scss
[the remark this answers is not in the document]
robertknight: > Because .prettierrc has "singleQuote": true it changed the quote style when saving the file.

Ah. In the lms repo prettier currently only applies to JavaScript. See the `format` command in `package.json`. We already use Prettier to format .scss files in the client repo so it would make sense to use it here as well. See `package.json` in the client repo. If we're going to make that change it should be in a new PR which modifies the "format" and "checkformatting" scripts and also reformats the existing `.scss` files.

lms/static/scripts/frontend_apps/components/FileList.js
esanzgar: I implemented your suggestion.

Your approach is different from [Sean's proposal](https://github.com/hypothesis/lms/pull/2100) which make the FileList less agnostic.

lms/static/scripts/frontend_apps/components/FileList.js
```diff
@@ -6,6 +6,7 @@ import Table from './Table';
 
 /**
  * @typedef {import('../api-types').File} File
+ * @typedef {import("preact").VNode} VNode
```
esanzgar, line 9: I use VNode because it allows string, JSX components, and arrays of components.

esanzgar:
In the latest guidelines from @klemay, she requested to remove the 'Select' button when no files were available.

 Like this:
image

In order to remove the buttons in a more cleaner way, I introduced a second commit  (from [this draft PR](https://github.com/hypothesis/lms/pull/2147)) which added more flexibility to the `LMSFilePicker` component (to add or remove titles and buttons in a simpler way). 

A regression comment about the second commit. There is a small deviation from the previous behaviour: when the fetching of files fails generically (not because of authorization issues), it retries the fetching without bugging the user to re-authorise the access to files.

robertknight (commented):
Hi Eduardo,

Everything worked as expected in my manual testing. The code looked good. I requested some minor changes to a few details.

The main query I have is about the change to the behaviour when an error occurs with an associated error message. Can you remind me what prompted that change, as it isn't mentioned in the PR description?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -136,62 +175,82 @@ export default function LMSFilePicker({
...
+    case 'authorize':
+      buttons = [
+        <Button
+          key="showAuthWindow"
+          onClick={authorizeAndFetchFiles}
+          label="Authorize"
+        />,
+      ];
+      warningError = (
+        <p label="authorization warning">
```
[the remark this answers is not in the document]
esanzgar: You are correct. I will change them.

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -136,62 +175,82 @@ export default function LMSFilePicker({
   const useSelectedFile = () =>
     onSelectFile(/** @type {File} */ (selectedFile));
 
-  const title =
-    dialogState.state === 'authorizing' ? 'Allow file access' : 'Select a file';
+  let buttons;
+  let warningError;
```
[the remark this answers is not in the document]
esanzgar: I renamed to `warningOrError` to make it more generic, in case it needed to be extended to other type of warnings or errors. I hope it is OK.

esanzgar:
In response to @robertknight:

In order to remove the ‘Select’ file button I refactored the code to make it more flexible. During this refactor I realised that if the API returned an error not related to authorisation, the dialog warning and the action was authorisation related. That did not seem right to me. For example, in case of a network error during the API request, I think it is not suitable to display a warning about authorisation. Instead, I changed to display a generic warning and to retry the fetching of files. It only prompted the authorisation dialogue when the error is specific about authorisation.

However, as @robertknight mentioned, re-authorizing _maybe_ the right thing to do, therefore, I changed a line in the code so the behaviour is equivalent to the previous version. In the test there is a hint that may indicate the re-authorizing could be the right behaviour.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -114,8 +114,12 @@ describe('LMSFilePicker', () => {
     assert.equal(fileList.prop('files'), expectedFiles);
   });
 
-  it('shows the "Try again" prompt after a failed authorization attempt', async () => {
-    fakeApiCall.rejects(new ApiError('Not authorized', {}));
+  it('shows the "Authorize" and "Authorize again" buttons after 2 failed authorization requests', async () => {
+    fakeApiCall.rejects(
+      new ApiError('Not authorized', {
+        /** withouth message */
```
robertknight, line 120:
```suggestion
        /** withouth message */
```

Should this comment say something like "without `errorMessage` property"?

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -201,6 +236,38 @@ describe('LMSFilePicker', () => {
     });
   });
 
+  it('shows no buttons when the request returns an empty list of files', async () => {
```
robertknight, line 239: The dialog's built-in "Cancel" button will still be shown in this case. Perhaps "shows no continue button ..." or "shows no select button ...".

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -74,24 +97,45 @@ export default function LMSFilePicker({
...
+        });
       }
     }
-  }, [authToken, listFilesApi]);
+  }, [authToken, listFilesApi, authorizationAttempted]);
 
   // Execute the authorization flow in a popup window and then attempt to
   // fetch files.
   const authorizeAndFetchFiles = useCallback(async () => {
-    setDialogState({ ...INITIAL_DIALOG_STATE, state: 'authorizing' });
```
robertknight, line 93: There are still changes to the `authorizeAndFetchFiles` logic here. Are these still applicable or do they need to be reverted?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -118,11 +161,7 @@ export default function LMSFilePicker({
   // On the initial load, fetch files or prompt to authorize if we know that
   // authorization will be required.
   useEffect(() => {
-    if (dialogState.state === 'authorizing') {
-      authorizeAndFetchFiles();
-    } else {
-      fetchFiles();
-    }
+    fetchFiles();
```
robertknight, line 164: Is this change still applicable given that we're reverting the authorization error-handling changes in this PR?

robertknight (commented):
I had a couple of queries about the changes to `authorizeAndFetchFiles`.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
esanzgar: Yes, it should be.

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
esanzgar: This useEffect hook is only run when the component is mounted. At that time, the `dialogState.state` is never 'authorizing'.

This is my understanding on these lines. That's why I remove the lines from 121 to 125.

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
esanzgar: No, they don't need to be reverted.

The 'authorizing' state is set within the `fetchFiles` function, when an error is raised from the API.

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
robertknight: This makes sense. Yesterday I missed the fact that `dialogState.state` is always `fetching` here.

robertknight (approved):
LGTM 👍

PR #2158, opened 2020-09-30
Prettify scss files

esanzgar:


robertknight:
To provide context, we started using Prettier for SASS files in the client project last year in https://github.com/hypothesis/client/pull/1194. This makes the same change for the lms project.

robertknight (approved):

PR #2180, opened 2020-10-13
Move pytestmarks to tops of files

seanh:


jon-betts (approved):
I think this is a small but nice addition. 

`pytest` is good, but leads to complicated tests where it isn't clear where things come from. This gives the reader a heads up right away that certain things are in play, and that they are _probably_ not defined in this file. We can make that more certain by using `autouse=true` for local things, or the class decorator where it's a pain.

PR #2184, opened 2020-10-19
Don’t hardcode karma config port number

LMS007:
Don’t assume the port is 9876. If the client tests are running multiple instances, then the port number may not be 9876. This can happen if client tests are running at the same time lms tests are running. If lms starts second, then the port number may be 9877.

---------

I noticed these tests were failing for me and it was a bit hard to track down. I finally realized the port number was off by one, then it occured to me that I also had the client tests running in another instance of MS code. I did a grep in the client repo and it does not look like we are making this assumption there. I think this only was an issue with LMS.


robertknight (commented):
LGTM 👍

PR #2204, opened 2020-10-28
Synchronize Babel configuration with the client

robertknight:
We recently established a new minimum set of supported browsers for the
Hypothesis client [1] and updated the JS compilation settings accordingly.
This commit brings the LMS frontend's minimum supported browsers in line
with the client.

The related PR https://github.com/hypothesis/lms/pull/2203 adds a warning banner which will be displayed in browsers that don't support the modern JS syntax to which the LMS frontend is now compiled.

[1] https://github.com/hypothesis/client/pull/2659

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,12 +4,23 @@
       "pragma": "createElement"
     }],
     ["@babel/preset-env", {
+       "bugfixes": true,
        "targets": {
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
LGTM

PR #2209, opened 2020-10-29
Make browser support test work better in Edge Legacy

robertknight:
Add a browser feature test for [`Element.prepend()` method](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend). This fixes a
failure of the "unsupported browser" banner to show in older,
unsupported versions of Edge (<= 16). This test is also a good match for
our minimum Chrome version (available in v54, our minimum is Chrome
v55).

Part of https://github.com/hypothesis/product-backlog/issues/1141.


lyzadanger (approved):
LGTM

PR #2214, opened 2020-11-02
Remove the writer flag

jon-betts:
For: https://github.com/hypothesis/via3/issues/245

This removes the ability to switch on the "experimental rewriter" in Via 3, which is now gone.

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -32,15 +32,7 @@ class _ViaDoc:
 class _ViaClient:
     """A small wrapper to make calling Via easier."""
 
-    def __init__(
-        # pylint: disable=too-many-arguments
-        self,
-        service_url,
-        legacy_service_url,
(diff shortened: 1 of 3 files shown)


lms/views/helpers/_via.py
```diff
@@ -53,26 +45,17 @@ def __init__(
...
         if self.legacy_mode:
             return self._legacy_via_url(doc.url)
 
         if doc.is_html:
-            if self.via_rewriter_mode:
-                return self._pywb_style_url(
-                    self.service_url._replace(path="/html/v/").geturl(), doc.url
-                )
-
             return self._legacy_via_url(doc.url)
```
jon-betts, line 54: This means we don't direct straight to ViaHTML at any time, but this is going to change pretty soon, when we strip out all the legacy Via stuff.

seanh (approved):

PR #2215, opened 2020-11-03
Move browser support list to `browserslist` key in package.json

robertknight:
Move the browser support list from `.babelrc` to package.json, where it
will be seen by both CSS and JS processing tools. This results in
slightly smaller CSS bundles due to avoiding unnecessary vendor
prefixing.

See https://github.com/hypothesis/client/pull/2698


lyzadanger (approved):
👍

PR #2216, opened 2020-11-03
Enable `strict` option in tsconfig.json

robertknight:
This did not find any additional errors but could catch some in future.
`noImplicitAny` is left off because we still have un-annotated
parameters in the code that would need addressing first.


esanzgar (approved):
👍🏼

PR #2219, opened 2020-11-04
Polish the warning message if no files are available in the course

esanzgar:
Second iteration based on additional feedback from a Slack conversation:
https://app.slack.com/client/T03QZM0HN/C07NXBDNW/thread/C2BLQDKHA-1604440053.200400

In summary:

* Text has been simplified and condensed
* Text is centred on the table
* Spinner is shown on the table for at least 2s.
* A `Reload` button has been added to allow retry.

image

Spinner:

image

### DIFF
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
@@ -1,4 +1,4 @@
-import { createElement, Fragment } from 'preact';
+import { createElement } from 'preact';
 import { useCallback, useEffect, useRef, useState } from 'preact/hooks';
 
 import { ApiError, apiCall } from '../utils/api';
@@ -28,7 +28,7 @@ import FileList from './FileList';
  * @typedef DialogState
  * @prop {'fetching'|'fetched'|'authorizing'|'error'} state
  * @prop {string} title - Dialog title
- * @prop {'select'|'authorize'|'authorize_retry'|'retry'|'none'} continueAction - Action for the continue button
+ * @prop {'select'|'authorize'|'authorize_retry'|'retry'|'reload'} continueAction - Action for the continue button
  * @prop {File[]|null} files - List of fetched files
  * @prop {Error|null} error - Details of current error, if `state` is 'error'
  */
@@ -43,22 +43,19 @@ const INITIAL_DIALOG_STATE = {
 };
 
 const CanvasNoFiles = (
-  <Fragment>
-    <div className="FileList__no-files">
-      There are no available PDF files in the current course.
-    </div>
+  <div className="FileList__no-files-message">
     <p>
-      Please see the Canvas help article for{' '}
+      There are no PDFs in this course.{' '}
       <a
         href="https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-upload-a-file-to-a-course/ta-p/618"
         target="_blank"
         rel="noreferrer"
       >
-        uploading files to a course
+        Upload some files to the course
       </a>{' '}
       and try again.
     </p>
-  </Fragment>
+  </div>
 );
 
 /**
@@ -92,13 +89,20 @@ export default function LMSFilePicker({
   // Fetches files or shows a prompt to authorize access.
   const fetchFiles = useCallback(async () => {
     try {
-      setDialogState({ ...INITIAL_DIALOG_STATE, state: 'fetching' });
+      // Show the fetching state, but preserve the existing continueAction to
+      // prevent the button label changing. See:
+      // https://github.com/hypothesis/lms/pull/2219#issuecomment-721833947
+      setDialogState(({ continueAction }) => ({
+        ...INITIAL_DIALOG_STATE,
+        state: 'fetching',
+        continueAction,
+      }));
       const files = /** @type {File[]} */ (await apiCall({
         authToken,
         path: listFilesApi.path,
       }));
       const continueAction =
-        files.length === 0 ? 'none' : INITIAL_DIALOG_STATE.continueAction;
+        files.length === 0 ? 'reload' : INITIAL_DIALOG_STATE.continueAction;
       setDialogState({
         ...INITIAL_DIALOG_STATE,
         state: 'fetched',
@@ -181,7 +185,6 @@ export default function LMSFilePicker({
     case 'select':
       continueButton = (
         <Button
-          key="select"
           disabled={selectedFile === null}
           onClick={useSelectedFile}
           label="Select"
@@ -191,11 +194,7 @@ export default function LMSFilePicker({
       break;
     case 'authorize':
       continueButton = (
-        <Button
-          key="showAuthWindow"
-          onClick={authorizeAndFetchFiles}
-          label="Authorize"
-        />
+        <Button onClick={authorizeAndFetchFiles} label="Authorize" />
       );
       warningOrError = (
         <p data-testid="authorization warning">
@@ -205,11 +204,7 @@ export default function LMSFilePicker({
       break;
     case 'authorize_retry':
       continueButton = (
-        <Button
-          key="showAuthWindow"
-          onClick={authorizeAndFetchFiles}
-          label="Authorize again"
-        />
+        <Button onClick={authorizeAndFetchFiles} label="Authorize again" />
       );
       warningOrError = (
         <ErrorDisplay
@@ -221,7 +216,6 @@ export default function LMSFilePicker({
     case 'retry':
       continueButton = (
         <Button
-          key="retry"
           onClick={authorizeAndFetchFiles} // maybe it should use fetchFile function instead
           label="Try again"
         />
@@ -233,8 +227,16 @@ export default function LMSFilePicker({
         />
       );
       break;
+    case 'reload':
+      continueButton = (
+        <Button
+          disabled={dialogState.state === 'fetching'}
+          onClick={fetchFiles}
+          label="Reload"
+        />
+      );
+      break;
     default:
-      // 'none' case
       continueButton = null;
       warningOrError = null;
       break;
--- a/lms/static/scripts/frontend_apps/components/Table.js
+++ b/lms/static/scripts/frontend_apps/components/Table.js
@@ -100,7 +100,12 @@ export default function Table({
   };
 
   return (
-    <div className="Table__wrapper">
+    <div
+      className={classnames({
+        Table__wrapper: true,
+        'has-files': items.length > 0,
+      })}
+    >
       <table
         aria-label={accessibleLabel}
         className="Table__table"
--- a/lms/static/styles/components/_FileList.scss
+++ b/lms/static/styles/components/_FileList.scss
@@ -10,14 +10,7 @@ $file-list-icon-size: 24px;
 
   // Override the default minimum height of the table so that the file list
   // is fully visible even if available height is very small.
-  min-height: 100px;
-}
-
-.FileList__spinner {
-  position: absolute;
-  margin: 15px;
-  left: calc(50% - #{$file-list-icon-size / 2});
-  top: calc(50% - #{$file-list-icon-size / 2});
+  min-height: 130px;
 }
 
 .FileList__name-header {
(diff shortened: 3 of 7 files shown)


lyzadanger:
Hi @esanzgar ! To aid the PR review process, would you be so kind as to add screenshots and explanation of the interface changes to the PR description?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -43,22 +43,19 @@ const INITIAL_DIALOG_STATE = {
...
 
 const CanvasNoFiles = (
-  <Fragment>
-    <div className="FileList__no-files">
-      There are no available PDF files in the current course.
-    </div>
+  <div className="FileList__no-files">
     <p>
-      Please see the Canvas help article for{' '}
+      There are no PDFs in the course.{' '}
```
lyzadanger, line 48: Suggest "this" instead of "the" here, i.e. "There are no PDFs in this course."

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -43,22 +43,19 @@ const INITIAL_DIALOG_STATE = {
...
       <a
         href="https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-upload-a-file-to-a-course/ta-p/618"
         target="_blank"
         rel="noreferrer"
       >
-        uploading files to a course
+        Updload some files to the course
```
lyzadanger, line 54: Typo: "Upload"
Suggest also changing "the" to "this" here:

"Upload some files to this course"

robertknight, line 54:
```suggestion
        Upload some files to the course
```

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -43,22 +43,19 @@ const INITIAL_DIALOG_STATE = {
 };
 
 const CanvasNoFiles = (
-  <Fragment>
-    <div className="FileList__no-files">
-      There are no available PDF files in the current course.
-    </div>
+  <div className="FileList__no-files">
```
lyzadanger, line 46: Technically, using BEM syntax, I _believe_ this classname may be more correct as `FileList--no-files`, if I'm understanding its meaning correctly: this is a `FileList` in the context of `no-files` (i.e. would use "Modifier" syntax).

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
robertknight, line 54:
```suggestion
        Upload some files to the course
```

lyzadanger (commented):
Just leaving a few comments, not a specific approval (leaving that to @robertknight ).

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -191,11 +187,7 @@ export default function LMSFilePicker({
       break;
     case 'authorize':
       continueButton = (
-        <Button
-          key="showAuthWindow"
```
lyzadanger, line 195: Can you help me understand the removal of the `key` prop here?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -233,8 +220,21 @@ export default function LMSFilePicker({
         />
       );
       break;
+    case 'reload':
+      continueButton = (
+        <Button
+          onClick={() => {
```
lyzadanger, line 226: This anonymous function here has enough logic that it seems like extracting it to a named function in the component may be ideal...

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -233,8 +220,21 @@ export default function LMSFilePicker({
...
+            setDialogState({
+              ...INITIAL_DIALOG_STATE,
+              continueAction: 'reload',
+            });
+            setTimeout(fetchFiles, 2000);
```
robertknight, line 231: Why is there a delay here? If there was a need for one, we should take care that it is cleared if any change (eg. component being unmounted) means it should no longer be triggered.

lms/static/scripts/frontend_apps/components/Table.js
```diff
@@ -100,7 +100,12 @@ export default function Table({
   };
 
   return (
-    <div className="Table__wrapper">
+    <div
+      className={classnames({
+        Table__wrapper: true,
+        'has-files': items.length,
```
robertknight, line 106: I'm guessing that any truthy value produces the same result here, though I personally prefer using a comparison `items.length > 0` to make this more explicit. We don't have a firm convention within the team.

robertknight (changes requested):
Hi Eduardo,

I noticed a typo and had a query about the timeout.

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
robertknight: Using a component `__` is correct as this is a sub-element of the `FilesList`. Maybe `FileList__no-files-message` might be better as it is a more noun-y name.

seanh:
Can we make the empty table be the same height as the table when it has files in it? It's almost there but still a few pixels too short. You can see the height change when the files appear:

Peek 2020-11-04 15-54

It should also remain the same height when it has more than 3 files and has a vertical scrollbar, as well:

Peek 2020-11-04 15-59

By the way, here's some code that you can apply to the LMS app to make it return no files but then, when <kbd>Reload</kbd> is clicked, return some files. For local testing:

```diff
diff --git a/lms/views/api/canvas/files.py b/lms/views/api/canvas/files.py
index d978e1cc..922a65c5 100644
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -3,6 +3,8 @@ from pyramid.view import view_config, view_defaults
 
 from lms.views import helpers
 
+foo = True
+
 
 @view_defaults(permission="api", renderer="json")
 class FilesAPIViews:
@@ -18,6 +20,12 @@ class FilesAPIViews:
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
+        global foo
+        if foo:
+            foo = False
+            return []
+
+        foo = True
         return self.canvas_api_client.list_files(self.request.matchdict["course_id"])
 
     @view_config(request_method="GET", route_name="canvas_api.files.via_url")
```

seanh:
Could we make the <kbd>Reload</kbd> button change to <kbd>Select</kbd> at the same time as the files appear? Currently <kbd>Reload</kbd> changes to <kbd>Select</kbd> and then there's a brief pause and _then_ the files appear. It'd be good if the two visual changes could happen at the same time

Peek 2020-11-04 15-54

seanh:
Wording suggestion: saying "this" twice is a little odd IMO:

> There are no PDFs in this course. Upload some files to this course and try again.

Change the second "this" to "the":

> There are no PDFs in this course. Upload some files to the course and try again.

seanh:
When you click <kbd>Reload</kbd> and there are still no files found, the <kbd>Reload</kbd> button very briefly changes to <kbd>Select</kbd> and then changes back to <kbd>Reload</kbd>, causing a visible flicker. It shouldn't change from <kbd>Reload</kbd> at all if the file list comes back empty again.

Peek 2020-11-04 16-24

esanzgar:
Regarding the height of the table, we want the table to expand if the iframe is resized, in this way:

Edit-assignment (1)

If we set a maximum size won't be using the available space efficiently.

Regarding the buttons, I will try to see what I can do.

seanh:
Peek 2020-11-04 17-37

You need a `min-height`, not a `max-height`. The empty files list is too short, not too tall.

esanzgar:
I fixed the issue of the re-rendering of the button.

esanzgar:
Based on @seanh suggestion, I adjusted the `min-height`.

lms/static/styles/components/_Table.scss
```diff
@@ -10,6 +10,10 @@ $table-item-height: 40px;
   border: 1px solid var.$grey-6;
 
   margin-top: 10px;
+
+  &.has-files {
+    min-height: 176px;
```
robertknight, line 15: Where does this very specific value of 176px come from?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -92,13 +89,20 @@ export default function LMSFilePicker({
   // Fetches files or shows a prompt to authorize access.
   const fetchFiles = useCallback(async () => {
     try {
-      setDialogState({ ...INITIAL_DIALOG_STATE, state: 'fetching' });
+      // Set `continueAction` at the end of this function, depending of
+      // the outcome of the fetch. This prevents short re-renderings of the button,
+      // as seen mentioned here: https://github.com/hypothesis/lms/pull/2219#issuecomment-721833947
```
robertknight, line 94: I found it difficult to relate the comment to the change here. My understanding is that what this code is actually doing is preserving the `continueAction` from the previous state instead of resetting it as we did before. Is that correct?

If so I would suggest to change the comment to something like "Show the fetching state, but preserve the existing `continueAction` to prevent the button label changing. See https://github.com/hypothesis/lms/pull/2219#issuecomment-721833947".

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
robertknight: My understanding is that it is no longer required because the `buttons` prop of dialog now accepts a single button as an argument, which is set to `continueButton` here, rather than requiring the buttons to be wrapped in an array. When the `buttons` prop is a single JSX element rather than an array, a `key` is not required. This is just a standard Preact/React behavior.

Have I got this right @esanzgar?

robertknight (commented):
Hi @esanzgar - I had a couple of small queries, otherwise the code looks good. Can you give me a shout when you've had a change to look at them and then we should be able to get this merged soon.

esanzgar (commented):
I addressed previous suggestions

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -233,8 +227,16 @@ export default function LMSFilePicker({
...
+        <Button
+          disabled={dialogState.state === 'fetching'}
+          onClick={fetchFiles}
+          label="Reload"
+        />
+      );
+      break;
     default:
-      // 'none' case
       continueButton = null;
```
robertknight, line 240: Is this case no longer used?

lms/static/styles/components/_Table.scss
robertknight: OK, I think we should at least have a short comment here to explain what this rule is intended to do. Can you see if there a way of calculating this value from other variables or note how you determined it. Otherwise this could presumably get out of sync if other variables or properties are changed in future.

lms/static/styles/components/_FileList.scss
```diff
@@ -61,10 +54,14 @@ $file-list-icon-size: 24px;
...
-  height: 40px;
+.FileList__no-files-message {
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
   vertical-align: middle;
-  padding: 12px;
+  text-align: center;
+  width: 100%;
+  min-height: 149px;
```
robertknight, line 64: Where does the `min-height` come from here? Is there a way of calculating it from other data? Could it vary depending on the browser or system-specific settings like the font used?

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
robertknight: It doesn't look to me like the comment has been updated. Do you have local changes that haven't been pushed?

robertknight (commented):
Hi @esanzgar - I had a query about the comment that you had intended to update and also how the "magic numbers" used for `min-height` values were calculated, so we know what changes might require updating them in future (or better yet, create them with an expression which updates automatically).

robertknight (approved):

Issue #2217, opened 2020-11-03
Show error message for course-copied Canvas Files assignments


seanh:
See also:

* [EPIC: Course copy](https://github.com/hypothesis/lms/issues/2241)
* [Make it easier for instructors to manually fix course-copied Canvas File assignments](https://github.com/hypothesis/lms/issues/2259)
* [Automatically fix course-copied Canvas Files assignments](https://github.com/hypothesis/lms/issues/2258)
* Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1604408267009300
* Another Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1604420478029000
* And another: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1604420265043800

# Problem

If you use Canvas's course copy feature to copy a course that contains a Canvas Files assignment, the new copy of the assignment won't work for some users in the new course. Specifically:

* Users who are members of the original course will still be able to launch the assignment (so most of the time teachers will still be able to launch the assignment)
* Users who are _not_ members of the original course _won't_ be able to launch the assignment (so many students won't be able to launch the assignment)

When a user isn't a member of the old course and they try to launch the new Canvas Files assignment in the new course, they get into an authorization loop:

![Peek 2020-11-03 16-47](https://user-images.githubusercontent.com/22498/98015141-5d54db00-1df4-11eb-9bc1-e1d53b5108fe.gif)

# Diagnosis

Canvas file ID's are global, not per course. For example the ["public inline preview url" API](https://canvas.instructure.com/doc/api/files.html#method.files.public_url) that we use to get a Via-compatible URL for a Canvas file is:

```
GET /api/v1/files/:id/public_url
```

`:id` is an integer, e.g. `476`. Canvas uses incrementing integers for file IDs. The first file uploaded to a given Canvas instance (any course) gets ID `1`, the second file (whether uploaded to the same course or to a different course) gets ID `2`, etc.

When you copy a Canvas course the new course will contain _copies_ of all the uploaded files from the old courses and the copies _will have different ID's_.

[...]

PR #2223, opened 2020-11-05
Add custom error response for Canvas API permissions errors

seanh:
Part of https://github.com/hypothesis/lms/issues/2217

Depends on https://github.com/hypothesis/lms/pull/2222

Add a custom error response from our Canvas proxy API for Canvas API permissions errors. This error response is used:

* When a learner launches a Canvas Files assignment and the assignment's file is marked as _Unpublished_ in Canvas
* When a user launches a Canvas Files assignment in a course that was created from another course using Canvas's course copy feature, and the user isn't a member of the original course (so they don't have permission to read the original course's copy of the file)
* There might also be other ways to get a permissions error from the Canvas API?

Of course, adding a custom error response doesn't help any users until the frontend code has been updated to handle the new type of response by showing a new error dialog. This backend-only PR can still be merged now though:

1. It doesn't break the current UX for any other types of error

2. It doesn't change the current UX for Canvas Files permissions errors. The current behaviour (on master) when trying to launch a Canvas Files assignment whose file the user doesn't have permission to read is that we keep on asking the user to re-authorize us with Canvas forever and give no explanation as to why. This is obviously broken behaviour but it's no more broken on this branch than on master. This branch adds the new proxy API response that a future PR could use to fix this behaviour

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -81,8 +81,8 @@ class CanvasAPIError(ExternalRequestError):
     invalid or unexpected response is received from the Canvas API.
     """
 
-    @staticmethod
-    def raise_from(cause):
+    @classmethod
+    def raise_from(cls, cause):
         """
         Raise a :exc:`lms.services.CanvasAPIError` from the given ``cause``.
 
@@ -108,20 +108,8 @@ class CanvasAPIError(ExternalRequestError):
         ``<raised_exception>.response`` will be ``None``.
         """
         response = getattr(cause, "response", None)
-        status_code = getattr(response, "status_code", None)
 
-        exception_class = CanvasAPIServerError
-
-        if status_code == 401:
-            exception_class = CanvasAPIAccessTokenError
-        elif response is not None:
-            try:
-                json = response.json()
-            except ValueError:
-                pass
-            else:
-                if json.get("error_description") == "refresh_token not found":
-                    exception_class = CanvasAPIAccessTokenError
+        exception_class = cls._exception_class(response)
 
         details = {
             "validation_errors": getattr(cause, "messages", None),
@@ -143,6 +131,33 @@ class CanvasAPIError(ExternalRequestError):
             details=details,
         ) from cause
 
+    @staticmethod
+    def _exception_class(response):
+        """Return the exception class to raise for the given response."""
+        if response is None:
+            return CanvasAPIServerError
+
+        status_code = getattr(response, "status_code", None)
+
+        try:
+            response_json = response.json()
+        except ValueError:
+            response_json = {}
+
+        errors = response_json.get("errors", [])
+        error_description = response_json.get("error_description", "")
+
+        if {"message": "Invalid access token."} in errors:
+            return CanvasAPIAccessTokenError
+
+        if error_description == "refresh_token not found":
+            return CanvasAPIAccessTokenError
+
+        if status_code == 401:
+            return CanvasAPIPermissionError
+
+        return CanvasAPIServerError
+
 
 class CanvasAPIAccessTokenError(CanvasAPIError):
     """
@@ -155,6 +170,19 @@ class CanvasAPIAccessTokenError(CanvasAPIError):
     If we can put the user through the OAuth grant flow to get a new access
     token and then re-try the request, it might succeed.
     """
+
+
+class CanvasAPIPermissionError(CanvasAPIError):
+    """
+    A Canvas API permissions error.
+
+    This happens when the user's access token is fine but they don't have
+    permission to carry out the requested action. For example a user might not
+    have permission to get a public URL for a file if they don't have
+    permission to read that file.
+    """
+
+    error_code = "canvas_api_permission_error"
 
 
 class CanvasAPIServerError(CanvasAPIError):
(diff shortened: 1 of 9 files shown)


lms/services/exceptions.py
```diff
@@ -113,7 +113,10 @@ def raise_from(cause):
         exception_class = CanvasAPIServerError
 
         if status_code == 401:
-            exception_class = CanvasAPIAccessTokenError
+            if "status" in response.json():
```
robertknight, line 116: Can you clarify how the presence of a `status` field in the response allows us to distinguish between the two categories of error? I can make an educated guess, but I'd like to be sure. Are there any scenarios you can think of where this might produce an incorrect result?

lms/views/api/exceptions.py
```diff
@@ -26,18 +27,28 @@ class ExceptionViews:
...
     optional):
 
-    1. "message": An error message for the frontend to show to the user.
+    1. "error_code": A unique string that identifies the error to the frontend.
+
+       If "error_code" is present the frontend should show an error dialog
+       relevant to the particular error code. For example if "error_code" is
+       "canvas_api_permission_error" the frontend should show a
+       "You don't have permission" error dialog. The error dialog should
+       include a [Try again] button that opens the "canvas_api.oauth.authorize"
```
robertknight, line 36: Hmm, this comment implies different handling of the error than I have implemented in my PR. In my PR I assumed that a permission error indicates that the access token is fine and so there is no point in repeating the authorization part of the process. Can you think of a situation where `canvas_api_permission_error` will be returned but re-authorization will help?

robertknight (commented):
The code looks fine to me, although I had a couple of queries.

lms/services/exceptions.py
seanh: Just based on testing the Canvas API by calling it various ways the absence of a `"status"` field seems to be the only thing we can use to distinguish access token errors from other 401 errors. See this Slack thread for notes: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1604492209058900

I'm not aware of any non-access token 401 responses from Canvas that lack a `status` field, but there's no way to know for sure.

There's also a risk that Canvas could add a `status` to access token errors in the future.

But I don't think we have any choice -- this is the only that I could fine to way to distinguish them

lms/views/api/exceptions.py
seanh: :man_shrugging: Who knows? The general approach in the LMS app so far has been that if anything goes wrong with a Canvas API request, the <kbd>Try again</kbd> button re-starts the _whole_ process, beginning with getting a new access token, regardless of whether we think a new access token is necessary or not. This re-try-from-the-very-start approach gives <kbd>Try again</kbd> the best possible chance of succeeding, keeps the code simpler (no need to distinguish re-tries that do and don't require an access token), and avoids potential issues where our code thinks a new access token isn't required but in fact one is (and so the user is stuck in a loop due to our faulty logic)

lms/services/exceptions.py
robertknight: OK, I think it would be good to note this uncertainty in a comment. Let's roll with it for now.

tests/unit/lms/services/exceptions_test.py
```diff
@@ -151,9 +180,7 @@ def test_it_raises_CanvasAPIServerError_for_a_successful_but_invalid_response(
             "validation_errors": "The response was invalid.",
         }
 
-    def test_it_raises_truncates_the_body_if_it_is_very_long(
-        self, canvas_api_long_response
-    ):
+    def test_it_truncates_the_body_if_it_is_very_long(self, canvas_api_long_response):
```
seanh, line 183: Fixed the name of this test

tests/unit/lms/services/exceptions_test.py
```diff
@@ -75,42 +76,70 @@ def test_when_theres_a_response_it_uses_it_in_str(
...
-    def test_it_raises_CanvasAccessTokenError_for_refresh_token_not_found_responses(
-        self,
-    ):
-        cause = self._requests_exception(
-            status=400,
-            body=json.dumps({"error_description": "refresh_token not found"}),
-        )
-
-        self.assert_raises(cause, CanvasAPIAccessTokenError)
-
```
seanh, line 113: I moved this test up into the parametrized one above

lms/services/exceptions.py
```diff
@@ -173,3 +174,30 @@ class LTIOutcomesAPIError(ExternalRequestError):
     Raised whenever an LTI outcomes API request times out or when an
     unsuccessful, invalid or unexpected response is received from the API.
     """
+
+
+def _exception_class(response):
```
robertknight, line 179: This function looks to be entirely Canvas-specific to me, but the containing module is not. Perhaps rename this accordingly (eg. `_canvas_exception_class`)?

robertknight (approved):
LGTM, although I suggested giving the `_exception_class` helper a more specific name.

lms/services/exceptions.py
seanh: I've made it a method of the `CanvasAPIError` class that calls it

PR #2224, opened 2020-11-06
Show custom error dialog for Canvas file permission errors

robertknight:
This PR adds a more specific and helpful error message in LTI launches that fail due to the Canvas file used in the assignment not being accessible, following the proposal in https://github.com/hypothesis/lms/issues/2217#issuecomment-722281453.

When the backend reports that fetching the content URL for an assignment
launch fails due to a Canvas file permission error, indicated by
`"error_code": "canvas_api_permission_error"` in the API response, show
a specific dialog for that error. This dialog explains common causes of
the problem and how to rectify them. It also differs from the generic
error dialog in that clicking "Try again" does not retry authorization,
since we know that is not necessary.

Summary of changes in this PR:

- Add an `errorCode` property to the `ApiError` error class and populate it from the `error_code` field in the API response
- Add a new `error-fetching-canvas-file` error state to the `BasicLtiLaunchApp` component, set this state if the `error_code` is `canvas_api_permission_error` and then render a specific error message in this error state. Unlike other error states, the "Try again" button does not attempt re-authorization in this state
- Slightly modify the name of existing error states for consistency (`error-fetch` => `error-fetching` etc.)

Something I didn't address in this PR is that there is some repetition of markup across the dialogs shown for the various error states. I left that refactoring as a follow-up task.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -24,6 +24,15 @@ import Spinner from './Spinner';
  */
 
 /**
+ * The categories of error that can happen when launching an assignment.
+ *
+ * These affect the message that is shown to users and the actions
+ * offered to remedy the problem.
+ *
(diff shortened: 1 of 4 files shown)


seanh (approved):
I haven't reviewed the code but I've tested this locally and it seems to work :+1:

PR #2235, opened 2020-11-10
Course copy poc

seanh:


seanh:
One issue that's not dealt with in this POC: if we're now going to save Canvas assignment configurations in the DB and let those override the `file_id`'s in the launch URLs saved in Canvas then, if the teacher Canvas's the <kbd>Edit</kbd> and edits the assignment in the usual way, we need to make sure that that saves the new assignment configuration to the DB not to Canvas. Otherwise it'll be overridden

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2237

PR #2236, opened 2020-11-10
Speed up 'make dev'

esanzgar:
Avoid doing an additional build when starting the development server.

[Based on this PR on the client](https://github.com/hypothesis/client/pull/2701/)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ help:
 	@echo "                       (Postgres) in Docker"
 	@echo 'make db                Upgrade the DB schema to the latest version'
 	@echo "make dev               Run the entire app (web server and other processes)"
+	@echo "make build             Create a production build of the static assests"
 	@echo "make supervisor        Launch a supervisorctl shell for managing the processes "
 	@echo '                       that `make dev` starts, type `help` for docs'
 	@echo "make shell             Launch a Python shell in the dev environment"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -148,7 +149,8 @@ frontend-tests: node_modules/.uptodate
 
 DOCKER_TAG = dev
 
-build/manifest.json: node_modules/.uptodate
+.PHONY: build
+build: node_modules/.uptodate
```
robertknight, line 153: There are other references to this target in this file which will need to be updated. I also suggest naming this `frontend-build` for consistency with the other `frontend-*` targets and also to avoid confusion in this project that has both a frontend and backend component and where the main output is a Docker image.

Makefile
```diff
@@ -33,7 +34,7 @@ db: python
 	@tox -qe dev  --run-command 'alembic -c conf/alembic.ini $(args)'
 
 .PHONY: dev
-dev: build/manifest.json python
+dev: node_modules/.uptodate python
```
robertknight, line 37: Something I'm not sure about here is whether this could introduce a race condition if the Python app starts up and is ready to go before the first frontend build completes. I would suggest to check what happens when `build/manifest.json` does not exist at the point when `make dev` is run and then `build/manifest.json` is not built for the first time until after the Python app has started.

robertknight (changes requested):
There are other references to `build/manifest.json` in this file which need updating. More importantly, I think this change may introduce a race condition depending on whether the initial asset build completes before the Python app has started.

esanzgar:
It introduces race conditions with Python.

PR #2237, opened 2020-11-10
Show instructors an error when Canvas file ID not found

seanh:
Show instructors an error message when they're launching a Canvas files assignment and the assignment's file ID isn't found in the course:

Screenshot from 2020-11-11 15-11-30

This happens when:

* The assignment was copied from another course using Canvas's course-copy feature: the new copies of the assignment's still have the file ID from the original course's copy of the file
* The assignment's file has been deleted from Canvas

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -33,7 +33,7 @@ class JSConfig:
     def _ai_getter(self):
         return self._request.find_service(name="ai_getter")
 
-    def add_canvas_file_id(self, canvas_file_id):
+    def add_canvas_file_id(self, course_id, canvas_file_id):
         """
         Set the document to the Canvas file with the given canvas_file_id.
 
@@ -46,7 +46,7 @@ class JSConfig:
         self._config["api"]["viaUrl"] = {
             "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
             "path": self._request.route_path(
-                "canvas_api.files.via_url", file_id=canvas_file_id
+                "canvas_api.files.via_url", course_id=course_id, file_id=canvas_file_id
             ),
         }
         self._add_canvas_speedgrader_settings(canvas_file_id=canvas_file_id)
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -58,7 +58,10 @@ def includeme(config):
     config.add_route(
         "canvas_api.courses.files.list", "/api/canvas/courses/{course_id}/files"
     )
-    config.add_route("canvas_api.files.via_url", "/api/canvas/files/{file_id}/via_url")
+    config.add_route(
+        "canvas_api.files.via_url",
+        "/api/canvas/courses/{course_id}/files/{file_id}/via_url",
+    )
     config.add_route("canvas_api.sync", "/api/canvas/sync", request_method="POST")
 
     config.add_route("lti_api.submissions.record", "/api/lti/submissions")
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -3,7 +3,7 @@
 import marshmallow
 from marshmallow import EXCLUDE, Schema, fields, post_load, validate, validates_schema
 
-from lms.services import CanvasAPIError
+from lms.services import CanvasAPIError, CanvasFileNotFoundInCourse
 from lms.validation import RequestsResponseSchema
 
 
@@ -39,7 +39,7 @@ class CanvasAPIClient:
         """
         Create a new CanvasAPIClient.
 
-        :param authenticated_client: An instance of AuthenticatedClient.
+        :param authenticated_client: An instance of AuthenticatedClient
         """
         self._client = authenticated_client
 
@@ -226,12 +226,22 @@ class CanvasAPIClient:
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
 
+    def check_file_in_course(self, file_id, course_id):
+        """
+        Raise if the file with ID file_id isn't in the course with ID course_id.
+
+        :raise CanvasFileNotFoundInCourse: If the file isn't in the course
+        """
+        if not any(
+            str(file_["id"]) == str(file_id) for file_ in self.list_files(course_id)
+        ):
+            raise CanvasFileNotFoundInCourse(file_id)
+
     def public_url(self, file_id):
         """
         Get a new temporary public download URL for the file with the given ID.
 
         :param file_id: the ID of the Canvas file
-        :rtype: str
         """
         # For documentation of this request see:
         # https://canvas.instructure.com/doc/api/files.html#method.files.public_url
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -201,3 +201,13 @@ class LTIOutcomesAPIError(ExternalRequestError):
     Raised whenever an LTI outcomes API request times out or when an
     unsuccessful, invalid or unexpected response is received from the API.
     """
+
+
+class CanvasFileNotFoundInCourse(ServiceError):
+    """A Canvas file ID wasn't found in the current course."""
+
+    error_code = "canvas_file_not_found_in_course"
+
+    def __init__(self, file_id):
+        self.details = {"file_id": file_id}
+        super().__init__(self.details)
--- a/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
@@ -29,7 +29,7 @@ import Spinner from './Spinner';
  * These affect the message that is shown to users and the actions
  * offered to remedy the problem.
  *
- * @typedef {'error-fetching'|'error-authorizing'|'error-reporting-submission'|'error-fetching-canvas-file'} ErrorState
+ * @typedef {'error-fetching'|'error-authorizing'|'error-reporting-submission'|'error-fetching-canvas-file'|'canvas-file-not-found-in-course'} ErrorState
  */
 
 /**
@@ -122,6 +122,12 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
     ) {
       setError(e);
       setErrorState('error-fetching-canvas-file');
+    } else if (
+      e instanceof ApiError &&
+      e.errorCode === 'canvas_file_not_found_in_course'
+    ) {
+      setError(e);
+      setErrorState('canvas-file-not-found-in-course');
     } else if (e instanceof ApiError && !e.errorMessage && retry) {
       setErrorState('error-authorizing');
     } else {
@@ -378,6 +384,44 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
               edit this assignment and re-select the file.
             </li>
           </ul>
+          <ErrorDisplay error={/** @type {Error} */ (error)} />
+        </Dialog>
+      )}
+      {errorState === 'canvas-file-not-found-in-course' && (
+        <Dialog
+          initialFocus={focusedDialogButton}
+          title="Hypothesis couldn't find the file in the course"
+          role="alertdialog"
+          buttons={[
+            <Button
+              buttonRef={focusedDialogButton}
+              className="BasicLtiLaunchApp__button"
+              disabled={fetchCount > 0}
+              key="retry"
+              label="Try again"
+              onClick={refetchContentUrl}
+            />,
+          ]}
+        >
+          <p>This might have happened because:</p>
+
+          <ul>
+            <li>The file has been deleted from Canvas</li>
+            <li>The course was copied from another course</li>
+          </ul>
+
+          <p>
+            To fix the issue,{' '}
+            <a
+              target="_blank"
+              rel="noreferrer"
+              href="https://web.hypothes.is/help/fixing-a-broken-canvas-file-link/"
+            >
+              edit the assignment and re-select the file
+            </a>
+            .
+          </p>
+
           <ErrorDisplay error={/** @type {Error} */ (error)} />
         </Dialog>
       )}
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -28,6 +28,12 @@ class FilesAPIViews:
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
+        file_id = self.request.matchdict["file_id"]
+        course_id = self.request.matchdict["course_id"]
+
+        if self.request.lti_user.is_instructor:
+            self.canvas_api_client.check_file_in_course(file_id, course_id)
+
         public_url = self.canvas_api_client.public_url(
             self.request.matchdict["file_id"]
         )
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -126,15 +126,19 @@ class TestAddCanvasFileID:
     """Unit tests for JSConfig.add_canvas_file_id()."""
 
     def test_it_adds_the_viaUrl_api_config(self, js_config):
-        js_config.add_canvas_file_id("example_canvas_file_id")
+        js_config.add_canvas_file_id(
+            "example_canvas_course_id", "example_canvas_file_id"
+        )
 
         assert js_config.asdict()["api"]["viaUrl"] == {
             "authUrl": "http://example.com/api/canvas/oauth/authorize",
-            "path": "/api/canvas/files/example_canvas_file_id/via_url",
+            "path": "/api/canvas/courses/example_canvas_course_id/files/example_canvas_file_id/via_url",
         }
 
     def test_it_sets_the_canvas_file_id(self, js_config, submission_params):
-        js_config.add_canvas_file_id("example_canvas_file_id")
+        js_config.add_canvas_file_id(
+            "example_canvas_course_id", "example_canvas_file_id"
+        )
 
         assert submission_params()["canvas_file_id"] == "example_canvas_file_id"
 
@@ -213,7 +217,10 @@ class TestAddCanvasFileIDAddDocumentURLCommon:
 
     @pytest.fixture(
         params=[
-            {"method": "add_canvas_file_id", "args": ["example_canvas_file_id"]},
+            {
+                "method": "add_canvas_file_id",
+                "args": ["example_canvas_course_id", "example_canvas_file_id"],
+            },
             {"method": "add_document_url", "args": ["example_document_url"]},
         ]
     )
@@ -221,9 +228,7 @@ class TestAddCanvasFileIDAddDocumentURLCommon:
         """Return a function that calls the method-under-test with default args."""
 
         def method_caller():
-            method_name = request.param["method"]
-            args = request.param["args"]
-            return getattr(js_config, method_name)(*args)
+            return getattr(js_config, request.param["method"])(*request.param["args"])
 
         return method_caller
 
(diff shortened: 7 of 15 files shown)


lms/resources/_js_config.py
```diff
@@ -33,7 +33,7 @@ def _consumer_key(self):
     def _ai_getter(self):
         return self._request.find_service(name="ai_getter")
 
-    def add_canvas_file_id(self, canvas_file_id):
+    def add_canvas_file_id(self, course_id, canvas_file_id):
```
seanh, line 36: Added the `course_id` here because it's now needed to generate the URL for the `via_url` API call below

lms/resources/_js_config.py
```diff
@@ -46,7 +46,7 @@ def add_canvas_file_id(self, canvas_file_id):
         self._config["api"]["viaUrl"] = {
             "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
             "path": self._request.route_path(
-                "canvas_api.files.via_url", file_id=canvas_file_id
+                "canvas_api.files.via_url", course_id=course_id, file_id=canvas_file_id
```
seanh, line 49: The `via_url` API now includes the course ID in the path. This changes thepath that we pass to the frontend, so the frontend will not use the new path when calling the API

lms/routes.py
```diff
@@ -58,7 +58,10 @@ def includeme(config):
     config.add_route(
         "canvas_api.courses.files.list", "/api/canvas/courses/{course_id}/files"
     )
-    config.add_route("canvas_api.files.via_url", "/api/canvas/files/{file_id}/via_url")
+    config.add_route(
+        "canvas_api.files.via_url",
+        "/api/canvas/courses/{course_id}/files/{file_id}/via_url",
+    )
```
seanh, line 64: Changed the `via_url` API's path to include the course ID

lms/services/canvas_api/client.py
```diff
@@ -226,16 +227,29 @@ class _ListFilesSchema(RequestsResponseSchema):
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
 
-    def public_url(self, file_id):
+    def public_url(self, course_id, file_id, is_instructor):
```
seanh, line 230: I added `course_id` and `is_instructor` as arguments to this method.

`is_instructor` could have been added as an argument to `Client.__init__()` and then the service factory would pass in `request.lti_user.is_instructor`. But that would make `is_instructor` potentially a dependency of all of `Client`'s methods. Putting it here clarifies that only this one method uses it

lms/services/canvas_api/client.py
```diff
@@ -226,16 +227,30 @@ class _ListFilesSchema(RequestsResponseSchema):
...
+        :raise CanvasFileNotFoundInCourse: If there's no file with ID file_id
+            in the course with ID course_id
         """
         # For documentation of this request see:
         # https://canvas.instructure.com/doc/api/files.html#method.files.public_url
 
+        if is_instructor:
+            file_ids = [str(file_["id"]) for file_ in self.list_files(course_id)]
+            if file_id not in file_ids:
+                raise CanvasFileNotFoundInCourse(file_id)
```
seanh, line 252: We call the Canvas API to get a list of all the files in the course, with possible pagination and multiple requests, in order to determine whether the file is in the course.

We only want to do this for instructors because:

* We don't want to add these additional Canvas API calls to every Canvas Files assignment launch. Only to instructor launches
* The exception that we raise triggers and error page that tells the user to edit the assignment and re-select the file. Only instructors can do that
* We can't actually tell whether the file is in the course when a student launches an assignment anyway. The list files API will only list the files that the student has access to, which doesn't necessarily include all of the files in the course (instructors can control whether files are visible to students or not)

When a student launches an assignment if they don't have access to the file (which could be for a variety of reasons) we get a permissions error from the Canvas API and we show an error message to the student. That was done in a previous PR: https://github.com/hypothesis/lms/pull/2223

seanh: `self.list_files(course_id)` is _not_ a generator

lms/services/canvas_api/client.py
```diff
@@ -226,16 +227,31 @@ class _ListFilesSchema(RequestsResponseSchema):
...
+        :raise CanvasFileNotFoundInCourse: If there's no file with ID file_id
+            in the course with ID course_id
         """
         # For documentation of this request see:
         # https://canvas.instructure.com/doc/api/files.html#method.files.public_url
 
+        if is_instructor:
+            if not any(
+                str(file_["id"]) == file_id for file_ in self.list_files(course_id)
+            ):
```
seanh, line 252: The use of `any()` like this avoids iterating over all of the files returned by `self.list_files(course_id)` unless necessary. If `list_files()` were a generator this could avoid unnecessary Canvas API requests (but it's not).

We do `str(file_["id"])` on each file because anything can be turned into a string. If we instead did `int(file_id)` on the one `file_id` variable, we'd have to deal with `int()` potentially crashing if a non-int `file_id` were passed to the API

seanh: We call the Canvas API to get a list of all the files in the course, with possible pagination and multiple requests, in order to determine whether the file is in the course.

We only want to do this for instructors because:

* We don't want to add these additional Canvas API calls to every Canvas Files assignment launch. Only to instructor launches
* The exception that we raise triggers and error page that tells the user to edit the assignment and re-select the file. Only instructors can do that
* We can't actually tell whether the file is in the course when a student launches an assignment anyway. The list files API will only list the files that the student has access to, which doesn't necessarily include all of the files in the course (instructors can control whether files are visible to students or not)

When a student launches an assignment if they don't have access to the file (which could be for a variety of reasons) we get a permissions error from the Canvas API and we show an error message to the student. That was done in a previous PR: https://github.com/hypothesis/lms/pull/2223

lms/services/exceptions.py
```diff
@@ -201,3 +201,13 @@ class LTIOutcomesAPIError(ExternalRequestError):
...
+
+
+class CanvasFileNotFoundInCourse(ServiceError):
+    """A Canvas file ID wasn't found in the current course."""
+
+    error_code = "canvas_file_not_found_in_course"
+
+    def __init__(self, file_id):
+        self.details = {"file_id": file_id}
+        super().__init__(self.details)
```
seanh, line 213: Note that this isn't a `CanvasAPIError`: it's not a problem with a Canvas API request. It doesn't have an associated response, message and details. The response to the Canvas API request came back fine. It just didn't contain the file ID we were looking for

lms/services/canvas_api/client.py
seanh: We could do this check in the LTI launch instead, e.g. in [`canvas_file_basic_lti_launch()`](https://github.com/hypothesis/lms/blob/067943f29fd8b880566bd6be51fbee0e485e86db/lms/views/basic_lti_launch.py#L74-L90). Then we wouldn't need to add these params to `public_url`. That would break a pattern in the LMS app's code architecture though: all third-party API requests are made during proxy API requests, not during LTI launch requests. The main reason, though, is that by adding this to the proxy API call we can trigger the proxy API to return an error response with an error code, and the frontend already has a way to trigger an error dialog in response to an error code from the proxy API. If we did the check during the LTI launch instead, it'd be more work to make the frontend display an error message

lms/resources/_js_config.py
robertknight: I presume you meant "_will_ use the new path when calling the API"?

lms/services/canvas_api/client.py
robertknight: Do you have an idea about how much overhead this will add to instructor launches? I guess we could cache the file ID => course ID association, although that doesn't account for a file being deleted. If the file _is_ deleted however, then the subsequent call to get a URL for the file itself will presumably fail. Alternatively we could remove this check from this API and instead make the frontend make a separate request which can happen in parallel with fetching the file URL. If that second call eventually fails the frontend can switch to the error state.

lms/static/scripts/frontend_apps/components/BasicLtiLaunchApp.js
```diff
@@ -381,6 +387,44 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
...
+              href="https://web.hypothes.is/help/fixing-a-broken-canvas-file-link/"
+            >
+              edit the assignment and re-select the file
+            </a>
+            .
+          </p>
+
+          <ErrorDisplay error={/** @type {Error} */ (error)} />
+        </Dialog>
+      )}
```
robertknight, line 427: The frontend changes here look fine. There is some duplication across the various error dialogs but either myself or one of the other front-end devs can address that in a follow-up.

lms/views/api/canvas/files.py
```diff
@@ -29,7 +29,9 @@ def via_url(self):
             This exception is caught and handled by an exception view.
         """
         public_url = self.canvas_api_client.public_url(
-            self.request.matchdict["file_id"]
+            self.request.matchdict["course_id"],
+            self.request.matchdict["file_id"],
+            self.request.lti_user.is_instructor,
```
robertknight, line 34: Possibly worth using kwargs here to avoid any risk of a mismatch between the call and the definition.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -20,6 +20,10 @@ def canvas_file_basic_lti_launch_caller(context, pyramid_request):
     BasicLTILaunchViews.canvas_file_basic_lti_launch(), and return whatever
     BasicLTILaunchViews.canvas_file_basic_lti_launch() returns.
     """
+    # The custom_canvas_course_id param is always present when
+    # canvas_file_basic_lti_launch() is called: Canvas always includes this
+    # param because we request it in our config.xml.
```
robertknight, line 25: This assumes the integrator hasn't modified the custom fields in the LTI configuration. Very unlikely, but technically possible to do.

robertknight (approved):
This worked as expected in my testing. The frontend changes look good to me. The backend changes also looked fine to me - Jon may wish to have a look over as well. The main issue of course is that this adds some overhead to each assignment launch by an instructor, which can vary depending on the number of files in the course. I had some suggestions about how we might improve that in future. We should definitely land this iteration first though.

lms/resources/_js_config.py
seanh: Yep

lms/services/canvas_api/client.py
seanh: I think we can cache them in the DB, but that needs to be verified. As far as I know the course that a given file ID is associated with cannot change in Canvas. I don't see any way to move files between courses, for example. And course copy creates new copies of the files with new IDs. But I think we'd need to verify with greater certainty that a file ID always belongs to the same course. If so we could cache this result in the DB

seanh: Listing a site's files seems to take anything from 0.5 sec to 3 secs. It looks like the ceiling is more like about 1.5 secs though, with the slower ones being outliers: https://one.nr/0OwvO83R

seanh: We certainly could move this check to a separate API endpoint and that might speed things up as the list files request to Canvas could be happening in parallel with the public URL request. I'm not sure whether I think it's worthwhile or not. It might actually make the code a bit nicer. Something to try in a follow-up PR maybe.

Since the frontend already has the `file_id`, this could be a frontend-only change: it could call the existing list files API and check for the ID itself. The frontend would need to know whether the user was an instructor, which I'm not sure it does currently. But I think that would close off the possibility of caching the results in the DB in future.

jon-betts (changes requested):
A couple of things up front: I've not looked at the front-end stuff, as I'm assuming Rob has covered that off better than I could. I've also not gone through the tests with a super fine comb as I have a suggestion which would change them a bit.

Having talked through the design with Sean the large scale structure seems sensible to me. My intuition is to blow up much earlier instead of threading through this variable so we can blow up later. However Sean pointed out if we explode to early, there's no front-end to catch the error and display it.

My main suggestion is to have a separate `assert_file_in_course()` and call it from `lms.views.api.canvas.files.FileAPIViews.via_url()`. This keeps the knowledge of `is_instructor` in the view, and keeps the `CanvasAPI` calls single purpose and much more straight forward. The conditionality in the current `public_url()` feels like view work to me.

I think this would be nice from a few points of view:

 * Simpler single function API object methods
 * Less disruption to existing code as `public_url` wouldn't change
 * In general separate functions are nice, because you can use them separately in new circumstances

I'd also like to suggest we adopt a **Style** of not putting type annotations on parameters. There's significant swithering about this, but most of our code doesn't do it and I think it's clearer for it.

lms/services/canvas_api/client.py
```diff
@@ -226,16 +227,31 @@ class _ListFilesSchema(RequestsResponseSchema):
...
+                raise CanvasFileNotFoundInCourse(file_id)
+
```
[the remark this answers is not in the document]
seanh: I've moved this to a separate `Client.check_file_in_course()` method, which simplifies the `Client` tests nicely and keeps `is_instructor` out of the `Client` service.

This is also inline with Rob's suggestion of moving it to an entirely separate API / view. Once we do that the `course_id` will no longer need to be included in the `via_url` API request at all, so it'll be even cleaner. And it'll be faster because we'll have more parallelism, and the chances of request timeouts will be lower as we'll be making fewer external requests per request to us. But that's for a follow-up PR.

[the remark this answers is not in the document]
seanh: One observation: this meant that the test that if the user is an instructor it checks that the file is in the course got moved into the view's tests. Since the view uses a mock `canvas_api_client`, this test was also much easier to write:

```python
@pytest.mark.usefixtures("user_is_instructor")
def test_it_checks_whether_the_file_is_in_the_course(self, canvas_api_client, pyramid_request):
    FilesAPIViews(pyramid_request).via_url()

    canvas_api_client.check_file_in_course.assert_called_once_with("test_file_id", "test_course_id")
```

Checking what it does when the file _isn't_ in the course was easy too, as to create that situation I only need to do `     canvas_api_client.check_file_in_course.side_effect = CanvasFileNotFoundInCourse()`.

Because the Canvas API client's tests don't use mocks, the test was much harder to write when the logic was in the Canvas API client. The test has to know what Canvas API responses look like even though that doesn't have anything to do with `check_file_in_course()`'s code (this knowledge about what Canvas API responses looks like ends up duplicated in many places in the tests). The assertion is particularly egregious as we have to make an assertion about a network request that's made in `_basic.py` (which is called by `_authenticated.py`, which is called by `client.py`). All the method under test actually does is call `client.send()` but we can't assert about that because it's a real `AuthenticatedClient` not a mock:

```python
    json_data=[
        {
            "display_name": "display_name_1",
            "id": 1,
            "updated_at": "updated_at_1",
        },
        {
            "display_name": "display_name_2",
            "id": 2,
            "updated_at": "updated_at_2",
        },
    ],
    status_code=200,
)

response = canvas_api_client.public_url(course_id="test_course_id", file_id="1", is_instructor=True)

http_session.send.assert_called_once_with(
    Any.request(
        url=Any.url.with_path("api/v1/courses/test_course_id/files"),
    ),
    timeout=Any()
)
```

 It's much harder to tell what's wrong when this test fails as well, since the range of code that could contain the bug or that the exception could arise from is much wider. For the same reason it's much harder to tell whether the test is correct and is in fact testing what it should be, not giving false positive test passes. Much harder to tell whether everything has been tested, or whether there are cases missing. The chances of the tests passing even though there is a bug in the code, are much higher. And the tests are much harder to understand and maintain.

tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -170,6 +176,61 @@ def test_list_files(self, canvas_api_client, http_session):
...
+                    "updated_at": "updated_at_1",
+                },
+                {
+                    "display_name": "display_name_2",
+                    "id": 2,
+                    "updated_at": "updated_at_2",
+                },
+            ],
+            status_code=200,
+        )
```
jon-betts: Feels like an obvious fixture here duplicated above.

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -185,36 +189,50 @@ def test_it_sets_the_canvas_submission_params(
...
+    @pytest.fixture(
+        params=[
+            {
+                "method": "add_canvas_file_id",
+                "args": ["example_canvas_course_id", "example_canvas_file_id"],
+            },
+            {"method": "add_document_url", "args": ["example_document_url"]},
+        ]
+    )
+    def method_caller(self, js_config, request):
```
jon-betts, line 227: This is better than `method` but I feel like we've not zeroed on in it yet on something specific enough. `add_file_method_caller` pretty lanky... what do these two have in common? What makes them good to test together? I can't think of something good...

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -185,36 +189,50 @@ def test_it_sets_the_canvas_submission_params(
...
+        """Return a function that calls the method-under-test with default args."""
+
+        def method_caller():
+            method_name = request.param["method"]
+            args = request.param["args"]
+            return getattr(js_config, method_name)(*args)
```
jon-betts, line 233:
```suggestion
            return getattr(js_config, request.param["method"])(*request.param["args"])
```

Unless this is for debug purposes?

lms/services/canvas_api/client.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, authenticated_client):
         """
         Create a new CanvasAPIClient.
 
-        :param authenticated_client: An instance of AuthenticatedClient.
+        :param authenticated_client: An instance of AuthenticatedClient
+
         """
```
jon-betts, line 44:
```suggestion
        :param authenticated_client: An instance of AuthenticatedClient
        """
```

tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -170,6 +176,31 @@ def test_list_files(self, canvas_api_client, http_session):
             timeout=Any(),
         )
 
+    @pytest.mark.usefixtures("list_files_response")
+    @pytest.mark.parametrize("file_id", [1, "1"])
```
jon-betts, line 180: The link between the number here and the fixture is a little subtle. But given:

 * It has to be an int
 * You want to test variations

I'm not really sure what you could / should do about it. The only thing I can think of is have `FILE_ID` as a constant and then do `[FILE_ID, str(FILE_ID)]` but ain't nobody got time for that.

jon-betts (approved):
This all seems pretty legit.

There are two things I think could be better, but I don't actually have any concrete idea as to how, but just in case you do:

 * `method_caller` is better than `method` but still very vague. They surely have more in common than just being methods.
 * The link between the `1` file id and the fixture is a bit subtle. I can't think of much better, but maybe just a longer number so it stands out more like `12345` or something? `12345` looks like an intentional choice, `1` and `2` might be required to be that.

tests/unit/lms/services/canvas_api/client_test.py
seanh: Yeah I thought (briefly) about but it wasn't easy enough so I get sad and gave up. The constant was also my only idea

tests/unit/lms/resources/_js_config_test.py
seanh: I think `method_caller()` is ok because it's in the `TestAddCanvasFileIDAddDocumentURLCommon` class. If a test class contains a fixture named `method_caller()` it must be a caller for the method(s) that the class tests.

When we move the check to a separate API I think this might go away anyway

PR #2242, opened 2020-11-16
Enable `noImplicitAny` option in TypeScript config

robertknight:
Fix the remaining issues that prevented the lms frontend from being
compiled with `noImplicitAny` enabled and remove the override in
`tsconfig.json`. Note that this setting is implicitly enabled by default when
`strict` is `true`.

The result is that the types of all identifiers must be either specified
or inferred, improving documentation for the reader and helping to catch
some minor bugs.


esanzgar (approved):
👍🏼

PR #2246, opened 2020-11-16
Extract `LaunchErrorDialog` component

robertknight:
Extract the various error dialogs that can be shown by the `BasicLtiLaunchApp` component into a separate `LaunchErrorDialog` component. This will help keep `BasicLtiLaunchApp` readable as new error cases are added. It also makes testing easier as the `BasicLtiLaunchApp` tests can focus on checking _which_ error dialog is shown and the `LaunchErrorDialog` component tests check _what_ is shown in a particular error dialog.

There should be no visual changes to the dialogs, except for a minor change in the layout of the text for the case when creating the speedgrader submission fails in Canvas due to fixing an inconsistency wrt. the other dialogs.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
+++ b/lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
@@ -1,0 +1,176 @@
+import { createElement } from 'preact';
+import { useRef } from 'preact/hooks';
+
+import Button from './Button';
+import Dialog from './Dialog';
+import ErrorDisplay from './ErrorDisplay';
+
+/**
+ * @typedef {import("preact").ComponentChildren} Children
+ * @typedef {import('./BasicLtiLaunchApp').ErrorState} ErrorState
+ */
+
+/**
+ * Common structure for LTI launch error dialogs.
+ *
+ * This is a "full-screen", non-cancelable dialog consisting of:
+ *
+ *  - An explanation of what went wrong and how to resolve it
+ *  - An optional "Try again" button
+ *  - An optional detailed error message for use when contacting support
+ *
+ * @param {Object} props
+ * @param {boolean} props.busy
+ * @param {Children} props.children
+ * @param {Error|null} [props.error]
+ * @param {() => void} [props.onRetry]
+ * @param {string} [props.retryLabel]
+ * @param {string} [props.title]
+ */
+function BaseDialog({
+  busy,
+  children,
+  error = null,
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -0,0 +1,176 @@
...
+ *
+ *  - An explanation of what went wrong and how to resolve it
+ *  - An optional "Try again" button
+ *  - An optional detailed error message for use when contacting support
+ *
+ * @param {Object} props
+ * @param {boolean} props.busy
+ * @param {Children} props.children
+ * @param {Error|null} [props.error]
+ * @param {() => void} [props.onRetry]
```
esanzgar, line 26: You can use `VoidFunction` if you wish.

lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -0,0 +1,176 @@
...
+    case 'error-fetching-canvas-file':
+      return (
+        <BaseDialog
+          busy={busy}
+          error={error}
+          onRetry={onRetry}
+          title="Couldn't get the file from Canvas"
+        >
+          <p>
+            Hypothesis couldn&apos;t get the assignment&apos;s file from Canvas.
```
esanzgar, line 112: I believe it should be the same as the more readable  `Hypothesis couldn't get the assignment file from Canvas.`

I think the second apostrophe is incorrect.

esanzgar (approved):
👍🏼

lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
robertknight: The text here is the same as on master. It isn't grammatically incorrect, although I might have phrased it more verbosely ("Hypothesis couldn't get the PDF for this assignment from Canvas"). I'd like to leave revising it out of scope for this PR though.

lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
robertknight: The `VoidFunction` interface appears to be an interface that is defined in web API specs and is used by things like `queueMicrotask`. I don't think it is intended for general use outside of that, but please let me know if you've  seen/heard otherwise.

Issue #2261, opened 2020-11-24
Fix course copy in Blackboard


seanh:
Course-copied assignments don't work in Blackboard: launching the new assignment makes the instructor re-select the assignment's document (and if students try to launch the assignment before the teacher has, they get an error message). Make it so that course-copied assignments remember the selected document.

See also:

* [EPIC: Course copy](https://github.com/hypothesis/lms/issues/2241)
* Investigation of why course copy doesn't work in Blackboard, and how to fix it: https://github.com/hypothesis/lms/issues/2218#issuecomment-732332176

PR #2262, opened 2020-11-24
Fix course copy in Blackboard

seanh:
Fixes [course copy in Blackboard](https://github.com/hypothesis/lms/issues/2261).

## Problem

When you copy a course in Blackboard, Blackboard copies all of the course's Hypothesis assignments but the new copies of the assignments all get new `resource_link_id`'s. Since the `resource_link_id` launch param is what we use to look up the assignment's settings (`ModuleItemConfiguration`) we fail to find the settings for the new copy of the assignment: to us it just looks like a brand new assignment, a new `resource_link_id` that we've never seen before. So we show the dialog asking to choose the document for the assignment.

## Solution

It turns out that Blackboard also includes an undocumented and non-standard launch param `resource_link_id_history`, which is the `resource_link_id` of the original assignment that this assignment was copied from. Look up the `resource_link_id_history`'s `ModuleItemConfiguration` in the DB and make a new copy of it with the new `resource_link_id`.

Note that the presence of a `resource_link_id_history` launch param doesn't guarantee that we will have a `ModuleItemConfiguration` for that `resource_link_id_history` in the DB. For example if you copy a course, and then don't launch the copied course's assignments, and then make a copy of the copy of the course, and launch that second copy's assignments, then we'll receive a launch request with both a `resource_link_id` and a `resource_link_id_history` that are both missing from our DB.

I didn't make any of the code in this PR Blackboard-specific: if it happens that any other LMS also sends the `resource_link_id_history` param then we may as well use it. (I don't think any other LMS's do support it though.)

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -40,6 +40,20 @@ class BasicLTILaunchViews:
         self.context.js_config.enable_lti_launch_mode()
         self.context.js_config.maybe_set_focused_user()
 
+    def basic_lti_launch(self, document_url=None, grading_supported=True):
+        """Do a basic LTI launch with the given document_url."""
+        self.sync_lti_data_to_h()
+        self.store_lti_data()
+        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
+
+        if grading_supported:
+            self.context.js_config.maybe_enable_grading()
+
+        if document_url is not None:
+            self.context.js_config.add_document_url(document_url)
+
+        return {}
+
     def sync_lti_data_to_h(self):
         """
         Sync LTI data to H.
@@ -84,14 +98,11 @@ class BasicLTILaunchViews:
         Via. We have to re-do this file-ID-for-download-URL exchange on every
         single launch because Canvas's download URLs are temporary.
         """
-        self.sync_lti_data_to_h()
-        self.store_lti_data()
-        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
         self.context.js_config.add_canvas_file_id(
             self.request.params["custom_canvas_course_id"],
             self.request.params["file_id"],
         )
-        return {}
+        return self.basic_lti_launch(grading_supported=False)
 
     @view_config(db_configured=True)
     def db_configured_basic_lti_launch(self):
@@ -105,23 +116,43 @@ class BasicLTILaunchViews:
         in the LMS and passing it back to us in each launch request. Instead we
         retrieve the document URL from the DB and pass it to Via.
         """
-        self.sync_lti_data_to_h()
-        self.store_lti_data()
-        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
-
-        self.context.js_config.maybe_enable_grading()
-
         # The ``db_configured=True`` view predicate ensures that this view
         # won't be called if there isn't a matching document_url in the DB. So
         # here we can safely assume that the document_url exists.
-        self.context.js_config.add_document_url(
-            self.assignment_service.get_document_url(
-                self.request.params["tool_consumer_instance_guid"],
-                self.request.params["resource_link_id"],
-            )
+        tool_consumer_instance_guid = self.request.params["tool_consumer_instance_guid"]
+        resource_link_id = self.request.params["resource_link_id"]
+        document_url = self.assignment_service.get_document_url(
+            tool_consumer_instance_guid, resource_link_id
+        )
+        return self.basic_lti_launch(document_url)
+
+    @view_config(blackboard_copied=True)
+    def blackboard_copied_basic_lti_launch(self):
+        """
+        Respond to a launch of a newly-copied Blackboard assignment.
+
+        Find the document_url for resource_link_id_history and make a copy of
+        it with the new resource_link_id, then lanuch the assignment as normal.
+
+        For more about Blackboard course copy and the resource_link_id_history
+        param see the BlackboardCopied predicate's docstring.
+        """
+        # The ``blackboard_copied=True`` view predicate ensures that this view
+        # won't be called if there isn't a resource_link_id_history param and a
+        # matching document_url in the DB.
+        tool_consumer_instance_guid = self.request.params["tool_consumer_instance_guid"]
+        resource_link_id = self.request.params["resource_link_id"]
+        resource_link_id_history = self.request.params["resource_link_id_history"]
+
+        document_url = self.assignment_service.get_document_url(
+            tool_consumer_instance_guid, resource_link_id_history
         )
 
-        return {}
+        self.assignment_service.set_document_url(
+            tool_consumer_instance_guid, resource_link_id, document_url
+        )
+
+        return self.basic_lti_launch(document_url)
 
     @view_config(url_configured=True, schema=URLConfiguredBasicLTILaunchSchema)
     def url_configured_basic_lti_launch(self):
@@ -135,12 +166,7 @@ class BasicLTILaunchViews:
         LMS, which passes it back to us in each launch request. All we have to
         do is pass the URL to Via.
         """
-        self.sync_lti_data_to_h()
-        self.store_lti_data()
-        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
-        self.context.js_config.maybe_enable_grading()
-        self.context.js_config.add_document_url(self.request.parsed_params["url"])
-        return {}
+        return self.basic_lti_launch(self.request.parsed_params["url"])
 
     @view_config(
         authorized_to_configure_assignments=True,
--- a/lms/views/predicates/_lti_launch.py
+++ b/lms/views/predicates/_lti_launch.py
@@ -1,14 +1,6 @@
 """Predicates for use with LTI launch views."""
 from lms.views.predicates._helpers import Base
 
-__all__ = [
-    "DBConfigured",
-    "CanvasFile",
-    "URLConfigured",
-    "Configured",
-    "AuthorizedToConfigureAssignments",
-]
-
 
 class DBConfigured(Base):
     """
@@ -48,6 +40,65 @@ class DBConfigured(Base):
         )
 
         return has_document_url == self.value
+
+
+class BlackboardCopied(Base):
+    """
+    Allow invoking an LTI launch view for newly course-copied Blackboard assignments.
+
+    When a user uses Blackboard's Course Copy feature to copy a course that
+    contains Hypothesis assignments, Blackboard copies the Hypothesis
+    assignments into the new course. Blackboard gives new
+    resource_link_id's to the new copies of the assignments and includes
+    the original resource_link_id in a resource_link_id_history launch
+    param.
+
+    Pass blackboard_copied=True to a view's configuration to allow invoking the
+    view only for newly course-copied Blackboard assignments where we don't yet
+    have a document_url for the new resource_link_id but we do have a
+    document_url for the resource_link_id_history.
+
+    For example:
+
+        @view_config(..., blackboard_copied=True)
+        def blackboard_course_copied_assignment_launch_view(context, request):
+            ...
+
+    resource_link_id_history is a non-standard and undocumented param
+    that's only used by Blackboard as far as we know, but nothing in this
+    code actually prevents us from using resource_link_id_history if
+    another LMS sends it to us.
+    """
+
+    name = "blackboard_copied"
+
+    def __init__(self, value, config):
+        super().__init__(value, config)
+        self.db_configured = DBConfigured(True, config)
+
+    def __call__(self, context, request):
+        if self.db_configured(context, request):
+            # We already have a document URL in the DB for this resource_link_id,
+            # so it's not a newly copied assignment.
+            is_newly_copied = False
+        else:
+            resource_link_id_history = request.params.get("resource_link_id_history")
+            tool_consumer_instance_guid = request.params.get(
+                "tool_consumer_instance_guid"
+            )
+
+            if not resource_link_id_history or not tool_consumer_instance_guid:
+                is_newly_copied = False
+            else:
+                # Look for the document URL of the previous assignment that
+                # this one was copied from.
+                assignment_service = request.find_service(name="assignment")
+                previous_document_url = assignment_service.get_document_url(
+                    tool_consumer_instance_guid, resource_link_id_history
+                )
+                is_newly_copied = bool(previous_document_url)
+
+        return is_newly_copied == self.value
 
 
 class CanvasFile(Base):
@@ -122,6 +173,7 @@ class Configured(Base):
         self.canvas_file = CanvasFile(True, config)
         self.url_configured = URLConfigured(True, config)
         self.db_configured = DBConfigured(True, config)
+        self.blackboard_copied = BlackboardCopied(True, config)
 
     def __call__(self, context, request):
         configured = any(
@@ -129,6 +181,7 @@ class Configured(Base):
                 self.canvas_file(context, request),
                 self.url_configured(context, request),
                 self.db_configured(context, request),
+                self.blackboard_copied(context, request),
             ]
         )
         return configured == self.value
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -50,6 +50,18 @@ def db_configured_basic_lti_launch_caller(context, pyramid_request):
     return views.db_configured_basic_lti_launch()
 
 
+def blackboard_copied_basic_lti_launch_caller(context, pyramid_request):
+    """
+    Call BasicLTILaunchViews.blackboard_copied_basic_lti_launch().
+
+    Set up the appropriate conditions and then call
+    BasicLTILaunchViews.blackboard_copied_basic_lti_launch(), and return
+    whatever BasicLTILaunchViews.blackboard_copied_basic_lti_launch() returns.
+    """
+    views = BasicLTILaunchViews(context, pyramid_request)
+    return views.blackboard_copied_basic_lti_launch()
+
+
 def url_configured_basic_lti_launch_caller(context, pyramid_request):
     """
     Call BasicLTILaunchViews.url_configured_basic_lti_launch().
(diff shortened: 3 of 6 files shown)


lms/views/basic_lti_launch.py
```diff
@@ -113,14 +113,26 @@ def db_configured_basic_lti_launch(self):
...
+            document_url = ModuleItemConfiguration.get_document_url(
+                self.request.db, tool_consumer_instance_guid, previous_resource_link_id
+            )
+
+            ModuleItemConfiguration.set_document_url(
+                self.request.db,
+                tool_consumer_instance_guid,
+                resource_link_id,
+                document_url,
+            )
```
seanh, line 134: Basically what this does is, if we don't have an assignment configuration for this assignment's ID, then it looks for an assignment configuration for the `resource_link_id_history` (the ID of the original assignment that this assignment was copied from) instead and copies that assignment configuration to the new assignment's ID.

If we don't find an assignment configuration (`ModuleItemConfiguration`) for the `resource_link_id` launch param in our DB then we look for a `ModuleItemConfiguration` for the `resource_link_id_history` launch param instead. If we find one, then we copy its `document_url` into a new `ModuleItemConfiguration` for the `resource_link_id`.

It needs to be a bit more careful about the cases where there's no `resource_link_id_history` param, or where there is one but we don't find it in our DB.

lms/views/predicates/_lti_launch.py
```diff
@@ -41,13 +41,21 @@ def __call__(self, context, request):
...
+        if document_url:
+            has_module_item_configuration = True
+        else:
+            previous_resource_link_id = request.params.get("resource_link_id_history")
+
+            previous_document_url = ModuleItemConfiguration.get_document_url(
+                request.db, tool_consumer_instance_guid, previous_resource_link_id
+            )
+
+            has_module_item_configuration = bool(previous_document_url)
```
seanh, line 57: `get_document_url()` gets called four times in total during a DB-configured launch (in the worst case). I think the solution is to add caching to `get_document_url()` though, rather than complicate the code by trying to pass its return value around. I'd have to think about caching SQLAlchemy query results and how that might interact with DB transactions. Also it might be that SQLAlchemy is already caching this for us. An optimisation to look into for in a follow-up PR.

seanh: So the end result of this code is that a view with `@view_config(db_configured=True)` will only be called if:

* The request has a `resource_link_id` param and we do have a `ModuleItemConfiguration` for it in the DB
* Or, failing that, the request has a `resource_link_id_history` param and we do have a `ModuleItemConfiguration` for that in the DB

The view will not be called if there isn't a `ModuleItemConfiguration` matching either the `resource_link_id` or the `resource_link_id_history` in the DB, even if both params are in the request.

lms/views/predicates/_lti_launch.py
```diff
@@ -41,13 +41,21 @@ def __call__(self, context, request):
...
-                request.db, tool_consumer_instance_guid, resource_link_id
-            )
-            is not None
+        document_url = ModuleItemConfiguration.get_document_url(
+            request.db, tool_consumer_instance_guid, resource_link_id
         )
 
+        if document_url:
+            has_module_item_configuration = True
+        else:
```
jon-betts, line 50: We really need to document the hell out of the fact this is Blackboard specific. Especially as this is undocumented, there's just no handle to grab onto in future to work out what this does and why.

jon-betts (commented):
I'm going to stop this one a little early so I can learn more. There are some things here which are leaping out at me right away which I don't understand:

The whole `get_document_url` / `set_document_url` thing is weird and horrible:

* Internally they have a `_get()` method that looks up a `ModuleItemConfiguration` 
* 9 times out of 10 this looks like what we actually want and we use `get_document_url` as a proxy to get to it to see if a `ModuleItemConfiguration` exists
* It looks like these methods would be better removed and replaced with `find()` and `find_or_create()` 
* The finding/making of a `ModuleItemConfiguration` could then be separated from the getting/setting of document urls

Split of responsibilities between the predicate and the view:

* The set of assumptions and situational awareness the view needs has increased a lot because what "DB configured" means has become two things instead of one
    * Either it's actually configured and ready to go, or it's in this new state
     * This is confusing and complicated (leading to all the comments and code)
* I'd prefer a solution that keeps the knowledge of this all in one place. For example:
    * We could let the `unconfigured` view catch this case. Configure the module and then call the `configured` view. Only one view needs to change here
   * We could let the predicate handle it all and ensure the module is conventionally configured. Only the predicate has to change
   * Something else? A new predicate? A new view?

We should indicate where this came from:

 * The code is written in such a way that any LMS could provide this parameter and it would work
 * This is kind of _ok_ I think, but not actually great. Given the general spaghetti and lack of any formal system for dealing with different LMS's in different ways it seems pragmatic
 * I don't think this LMS agnostic solution has to extend to the _comments_ though. We should absolutely document the hell out of the fact this is Blackboard specific.

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -352,6 +403,7 @@ def is_canvas(context):
 def pyramid_request(pyramid_request):
     pyramid_request.params.update(
         {
+            "resource_link_id_history": "test_resource_link_id_history",
```
jon-betts, line 406: Can this be a private override for the relevant tests rather than coupling?

tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -51,6 +52,84 @@ def test_when_request_params_are_missing(self, pyramid_request, value, expected)
...
+            # If there's a ModuleItemConfiguration for resource_link_id_history but
+            # not yet one for resource_link_id then blackboard_copied is True.
+            dict(
+                resource_link_id_exists=False,
+                resource_link_id_history_exists=True,
+                expected_result=True,
+            ),
+        ],
+    )
+    def test_it(self, params, pyramid_request):
```
jon-betts, line 87: For table like content, I think lists can be easier to grok here:

```python
    @pytest.mark.parametrize(
        "link_exists,history_exists,expected_result",
        [
            param(True, True, False, id="already copied")",
            param(True, False, False, id="normal db configured"),
            param(False, False, False, id="no history"),
            param(False, True, True, id="history copy required")
        ]
    )
    def test_it(self, link_exists, history_exists, expected_result, pyramid_request):
```

Another (possibly better) alternative is to directly encode the logic into the test as a declarative rule:

```python
    @pytest.mark.parametrize("link_exists", [True, False])
    @pytest.mark.parametrize("history_exists", [True, False])
    def test_it(self, link_exists, history_exists, pyramid_request):
        # We only need to copy if there is a history and no link has been created
         expected_result == history_exists and not link_exists
```

This has the benefit of being very compact, and actually describing the rule in play, rather than leaving it to the reader to work backwards from the examples.

tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -51,6 +52,84 @@ def test_when_request_params_are_missing(self, pyramid_request, value, expected)
...
+        )
+
+    def test_with_request_params_missing(self, pyramid_request):
+        pyramid_request.params = {}
+
+        predicate = BlackboardCopied(True, mock.sentinel.config)
+        assert not predicate(mock.sentinel.context, pyramid_request)
+
+        predicate = BlackboardCopied(False, mock.sentinel.config)
+        assert predicate(mock.sentinel.context, pyramid_request)
```
jon-betts, line 123: What's the first truthy argument here?

jon-betts (changes requested):
## Aboot the main code

The inclusion of the new function makes this _much_ better. I think the split is better too.

There's one spot with a `settings.get()` in the new view that I can't immediately see why it shouldn't be a direct subscript, which makes me think I'm missing something?

The way the predicates work is really unfortunate, as we end up executing the DB configured one four times by my count, and the new one twice due to the way the composition works. I don't know enough to know if there's a way around this. But adding the new one is going to add 4 new queries from what I can see.

If there was a way to make these atomic of keep the result that would be good, as as far as I can tell this will be adding lots of queries to every call.

## Aboot the tests

A small suggestion about splitting a query and assertion to make each clearer.

A suggestion for how to express some parameterisation in a much more compact way. I think beyond a certain length it becomes very hard to work out what rule they are trying to encode.

tests/unit/lms/views/predicates/_lti_launch_test.py
seanh: It's the value that was passed to the predicate: `@view_config(blackboard_copied=True)`. You can also do `@view_config(blackboard_copied=False)`

lms/views/predicates/_lti_launch.py
```diff
@@ -123,13 +173,15 @@ def __init__(self, value, config):
...
         self.db_configured = DBConfigured(True, config)
+        self.blackboard_copied = BlackboardCopied(True, config)
 
     def __call__(self, context, request):
         configured = any(
             [
                 self.canvas_file(context, request),
                 self.url_configured(context, request),
                 self.db_configured(context, request),
+                self.blackboard_copied(context, request),
```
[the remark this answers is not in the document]
seanh: You can turn on debug logging of SQL queries like this:

```diff
diff --git a/lms/app.py b/lms/app.py
index 613eb4f4..7df719a4 100644
--- a/lms/app.py
+++ b/lms/app.py
@@ -98,4 +98,7 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
 
     config.scan()
 
+    import logging
+    logging.getLogger("sqlalchemy.engine").setLevel(logging.DEBUG)
+
     return config.make_wsgi_app()
```

In the case where the `resource_link_id_history` does exist in the DB but the `resource_link_id` doesn't yet (which I think is the worst case) this is what I see (excerpt of the log output):

* The same application instance is `SELECT`'ed twice. (That's a separate issue, but we could cache it in the service layer.)
* The non-existent `ModuleItemConfiguration` for the `resource_link_id` (`_155_1`) is `SELECT`'ed 6 times
* The `ModuleItemConfiguration` for the `resource_link_id_history` (`_144_1`) is `SELECT`'ed 4 times
* A new `ModuleItemConfiguration` for `resource_link_id` is `INSERT`'ed
* There's also a `GroupInfo` `SELECT`, an `LtiLaunches` `INSERT` and a `Course` `SELECT`
* 16 DB queries in total (for the entire LTI launch, not just these views and predicates). This could be reduced to 8 if the `ModuleItemConfiguration` lookups were cached, 7 if the `ApplicationInstance` lookups were also cached

```
BEGIN (implicit)

SELECT application_instances.id ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_155_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_155_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_144_1', ...

SELECT module_item_configurations.id ...
{'resource_link_id_1': '_155_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_155_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_144_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_155_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_144_1', ...

SELECT application_instances.id ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_144_1', ...

SELECT module_item_configurations ...
{'resource_link_id_1': '_155_1', ...

INSERT INTO module_item_configurations ...
{'resource_link_id': '_155_1', ...

SELECT group_info.info ...

INSERT INTO lti_launches ...

SELECT course.consumer_key ...

COMMIT
```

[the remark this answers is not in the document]
seanh: Did a quick test of moving `set_document_url()` and `get_document_url()` into a service with `@lru_cache()` and it does indeed reduce the `SELECT module_item_configurations`'s to just two and caching works per request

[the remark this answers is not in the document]
seanh: The `ApplicationInstance` `SELECT` is actually [already cached in the services](https://github.com/hypothesis/lms/blob/3dbffc74550e7ba63847c8abb2b5e8152447bc19/lms/services/application_instance_getter.py#L99-L116) but the `LaunchVerifier` service [calls the model directly](https://github.com/hypothesis/lms/blob/3dbffc74550e7ba63847c8abb2b5e8152447bc19/lms/services/launch_verifier.py#L58-L60) instead. That's the reason for the two `ApplicationInstance` lookups. IIRC that was because `LaunchVerifier` couldn't depend on `ApplicationInstanceGetter` because it would create a circular dependency between the services. So if we wanted to optimise that duplicate `ApplicationInstance` `SELECT` away we'd have to resolve that circular dependency in the service layer (I think by splitting out a third service)

lms/views/predicates/_lti_launch.py
```diff
@@ -51,6 +43,64 @@ def __call__(self, context, request):
...
+    that's only used by Blackboard as far as we know, but nothing in this
+    code actually prevents us from using resource_link_id_history if
+    another LMS sends it to us.
+    """
+
+    name = "blackboard_copied"
+
+    def __init__(self, value, config):
+        super().__init__(value, config)
+        self.db_configured = DBConfigured(True, config)
```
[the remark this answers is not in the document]
seanh: Yeah. I think this is always the case with using composition instead of inheritance: composition creates more boilerplate code as you have to create the composed object in our `__init__()` (or have it passed in to you) and save it on `self`. But I think composition is nonetheless a better pattern: it keeps classes more decoupled, and code easier to understand

lms/views/basic_lti_launch.py
```diff
@@ -104,26 +115,47 @@ def db_configured_basic_lti_launch(self):
...
 
-        return {}
+        Find the ModuleItemConfiguration for resource_link_id_history and make
+        a copy of it with the new resource_link_id, then lanuch the assignment
+        as normal.
+
+        For more about Blackboard course copy and the resource_link_id_history
+        param see the BlackboardCopied predicate's docstring.
+        """
+        resource_link_id = self.request.params.get("resource_link_id")
```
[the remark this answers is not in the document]
seanh: Yep, there's a validation schema attached to this view class that ensures that it can't be blank

lms/views/predicates/_lti_launch.py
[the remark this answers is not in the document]
seanh: This has been rebased on top of https://github.com/hypothesis/lms/pull/2269 which adds the caching

tests/unit/lms/views/predicates/_lti_launch_test.py
seanh: :+1: The doubled `parametrize` decorators work pretty nicely I think

seanh:
This is ready for re-review. I've rebased it on top of https://github.com/hypothesis/lms/pull/2269 which adds caching to remove the duplicate DB queries

lms/views/basic_lti_launch.py
```diff
@@ -40,6 +40,20 @@ def __init__(self, context, request):
...
+        self.store_lti_data()
+        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
+
+        if grading_supported:
+            self.context.js_config.maybe_enable_grading()
+
+        if document_url is not None:
+            self.context.js_config.add_document_url(document_url)
+
+        return {}
```
seanh, line 55: Helper method added to deduplicate the various different LTI launch methods below

lms/views/basic_lti_launch.py
```diff
@@ -84,14 +98,11 @@ def canvas_file_basic_lti_launch(self):
...
         """
-        self.sync_lti_data_to_h()
-        self.store_lti_data()
-        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
         self.context.js_config.add_canvas_file_id(
             self.request.params["custom_canvas_course_id"],
             self.request.params["file_id"],
         )
-        return {}
+        return self.basic_lti_launch(grading_supported=False)
```
seanh, line 105: Just refactored to use the new helper method

lms/views/basic_lti_launch.py
```diff
@@ -105,23 +116,43 @@ def db_configured_basic_lti_launch(self):
...
-            self.assignment_service.get_document_url(
-                self.request.params["tool_consumer_instance_guid"],
-                self.request.params["resource_link_id"],
-            )
+        tool_consumer_instance_guid = self.request.params["tool_consumer_instance_guid"]
+        resource_link_id = self.request.params["resource_link_id"]
+        document_url = self.assignment_service.get_document_url(
+            tool_consumer_instance_guid, resource_link_id
         )
+        return self.basic_lti_launch(document_url)
```
seanh, line 127: Just refactored to use the new helper method

lms/views/basic_lti_launch.py
```diff
@@ -105,23 +116,43 @@ def db_configured_basic_lti_launch(self):
...
+
+        document_url = self.assignment_service.get_document_url(
+            tool_consumer_instance_guid, resource_link_id_history
+        )
+
+        self.assignment_service.set_document_url(
+            tool_consumer_instance_guid, resource_link_id, document_url
+        )
+
+        return self.basic_lti_launch(document_url)
```
seanh, line 155: This is the new view method for a new type of assignment launch: a "Blackboard copied" assignment launch. The predicates (below) will direct launch requests to this view as appropriate

lms/views/basic_lti_launch.py
```diff
@@ -135,12 +166,7 @@ def url_configured_basic_lti_launch(self):
         LMS, which passes it back to us in each launch request. All we have to
         do is pass the URL to Via.
         """
-        self.sync_lti_data_to_h()
-        self.store_lti_data()
-        self.course_service.get_or_create(self.context.h_group.authority_provided_id)
-        self.context.js_config.maybe_enable_grading()
-        self.context.js_config.add_document_url(self.request.parsed_params["url"])
-        return {}
+        return self.basic_lti_launch(self.request.parsed_params["url"])
```
seanh, line 169: Just refactored to use the new helper method

robertknight (approved):
LGTM. This makes sense to me. I think the most important next steps are to look at course copy in other LMSes and figure out whether they can re-use anything from the code path here or if we need to do something completely different.

PR #2265, opened 2020-11-25
Add ModuleItemConfiguration test factory

seanh:
Add a `ModuleItemConfiguration` test factory. Tests in a [follow-up PR](https://github.com/hypothesis/lms/pull/2266) will use it.

PR #2266, opened 2020-11-25
Remove ModuleItemConfiguration test mocks

seanh:
Depends on [Add ModuleItemConfiguration test factory](https://github.com/hypothesis/lms/pull/2265).

Mocking model classes is always optional, and replacing these mocks with the real model / DB and a test factory makes the tests better.

It also makes some test changes that I need to make in a [following pull request](https://github.com/hypothesis/lms/pull/2262/commits) much easier.

Note that I didn't remove _all_ uses of mocked `ModuleItemConfiguration`'s: in some tests mocking it was actually better.

### DIFF
--- a/tests/unit/lms/views/predicates/_lti_launch_test.py
+++ b/tests/unit/lms/views/predicates/_lti_launch_test.py
@@ -16,29 +16,23 @@ from tests import factories
 class TestDBConfigured:
     @pytest.mark.parametrize("value,expected", [(True, True), (False, False)])
     def test_when_theres_a_matching_assignment_config_in_the_db(
-        self, pyramid_request, value, expected, ModuleItemConfiguration
+        self, pyramid_request, value, expected
     ):
-        ModuleItemConfiguration.get_document_url.return_value = "test_document_url"
+        mic = factories.ModuleItemConfiguration()
         pyramid_request.params = {
-            "resource_link_id": "test_resource_link_id",
-            "tool_consumer_instance_guid": "test_tool_consumer_instance_guid",
+            "resource_link_id": mic.resource_link_id,
+            "tool_consumer_instance_guid": mic.tool_consumer_instance_guid,
         }
         predicate = DBConfigured(value, mock.sentinel.config)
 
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -16,55 +16,38 @@
...
         }
         predicate = DBConfigured(value, mock.sentinel.config)
 
         result = predicate(mock.sentinel.context, pyramid_request)
 
-        ModuleItemConfiguration.get_document_url.assert_called_once_with(
-            pyramid_request.db,
-            "test_tool_consumer_instance_guid",
-            "test_resource_link_id",
-        )
```
jon-betts, line 34: I _massively_ prefer this way of going about things. We aren't specifying _how_ it get this done any more, just that it does. If this class is refactored to do this in a different way this test won't fail.

tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -16,55 +16,38 @@
...
     @pytest.mark.parametrize("value,expected", [(True, False), (False, True)])
     def test_when_theres_no_matching_assignment_config_in_the_db(
-        self, pyramid_request, value, expected, ModuleItemConfiguration
+        self, pyramid_request, value, expected
     ):
-        ModuleItemConfiguration.get_document_url.return_value = None
         pyramid_request.params = {
             "resource_link_id": "test_resource_link_id",
             "tool_consumer_instance_guid": "test_tool_consumer_instance_guid",
         }
```
jon-betts, line 39: These strings look like magic values, but I presume are actually irrelevant. Could we change them if so to say so. Something like:

```suggestion
        pyramid_request.params = {
            "resource_link_id": "missing_id",
            "tool_consumer_instance_guid": "missing_guid",
        }
```

jon-betts (approved):
These look a lot better. We're definitely over mocking here. I think partly mocking is popular because it requires so little thought. You can just mock the code line for line without understanding it and call it a day.

I had one suggestion to never let a good string go to waste in one of the tests.

PR #2269, opened 2020-11-26
Add AssignmentService

seanh:
Add an `AssignmentService` for getting and setting assignment document URLs, instead of using classmethods on the model.

The main reason for doing this is that it makes the code more efficient because the service can cache SQL query results on a per-request basis (due to the design of pyramid_services: it lazily instantiates one singleton instance of each service _per request_) and avoid repeated DB queries for the same `ModuleItemConfiguration` within a single request.

For more on the motivation for this caching, see this discussion thread on another PR: https://github.com/hypothesis/lms/pull/2262#discussion_r530555013

Different parts of the code can blindly call the service's methods without regard to whether another part of the code might have called the same method in the same request, and the service-layer caching will ensure that the DB queries aren't repeated.

This also simplifies the interface because `get_document_url()` and `set_document_url()` don't need to have the `db` param passed in to them. It can be injected by the service factory.

## Design

I've limited this PR to fixing the duplicate SQL queries by moving `get_document_url()` and `set_document_url()` from the model to the services layer and adding caching. I've resisted the urge to make other design changes that've been discussed because I think those are going to raise more questions and issues. So it's still `get_document_url()` and `set_document_url()` methods, rather than methods that return and accept whole `ModuleItemConfiguration` objects. And the `get_document_url()` and `set_document_url()` methods still require `tool_consumer_instance_guid` and `resource_link_id` arguments, instead of having the current request params injected.

### Returning whole `ModuleItemConfiguration` objects instead of just `document_url` strings

I think this _might_ be a better design but I'm not sure, and it raises issues. Currently the code can just call `set_document_url(tool_consumer_instance_guid, resource_link_id, document_url)` and this will create a new `ModuleItemConfiguration` with `tool_consumer_instance_guid` and `resource_link_id` if none exists, or overwrite the `document_url` of an existing one. If the service just returned `ModuleItemConfiguration` objects instead, you'd need something like a `get_or_create(tool_consumer_instance_guid, resource_link_id)` method, used like so:

```python
mic = assignment_svc.get_or_create(tool_consumer_instance_guid, resource_link_id)
mic.document_url = document_url
```

There are some issues with this:

* Is it actually simpler than a single `set_document_url(tool_consumer_instance_guid, resource_link_id, document_url)` call? Or is the resulting view code and view tests more complex?
* You'd still need a plain `get()` method as sometimes view code wants to get the document URL or `None` without potentially creating a new record in the DB
* When `get_or_create()` creates a new `ModuleItemConfiguration` it has created a `ModuleItemConfiguration` with no `document_url` which will cause the DB session to crash if something doesn't then do `mic.document_url = document_url` (because `document_url` is a non-nullable column in the DB). But if `get_or_create()` returns an existing `ModuleItemConfiguration` this isn't necessary. (To tell the difference you'd have to test whether `mic.document_url` is `None`). Seems a bit odd maybe?

I think it's controversial enough that it's worth addressing in a separate PR, if at all.

### Injecting the `tool_consumer_instance_guid` and `resource_link_id` params into the service

The `AssignmentService` factory can inject `request.params["tool_consumer_instance_guid"]` and `request.params["resource_link_id"]` into `AssignmentService` which enables a simpler interface: `assignment_svc.get_document_url()` (no arguments) and `assignment_svc.set_document_url(document_url)` (just the one argument) instead of the views having to pass `tool_consumer_instance_guid` and `resource_link_id` params into these methods. But this raises some issues:

* `AssignmentService` does get called by view predicates for requests that don't have `tool_consumer_instance_guid` and `resource_link_id` params. It can't just crash when these params are missing, otherwise we have BDD tests failing. It probably needs to return a version of the service with `None` injected for these params, or something. Which is weird.

* Although `get_document_url()` and `set_document_url()` can _default_ to using the `tool_consumer_instance_guid` and `resource_link_id` from the request, they still need to allow different ones to be passed in on a per-call basis, because sometimes the code wants to use the `resource_link_id_history` param instead of `resource_link_id`. This complicates things the service code and tests.

* I'm not sure that service factories injecting request params is a good way to design services. Services and their factories shouldn't be coupled to the current request in this way. I think when services need request params, having the views pass them in to the service method calls might be better.

  For example views don't get called until _after_ their attached validation schemas have run. So a view can read `request.parsed_params["tool_consumer_instance_guid"]` and `request.parsed_params["resource_link_id"]` (in order to pass them into a service method as arguments) safe in the knowledge that those params must be present and valid, otherwise the validation layer would have prevented the view from being called.

  Service factories, on the other hand, can be executed by the resources layer, view predicates, and other parts that execute _before_ a validation schema has been selected and executed. So they can't rely on the validation schema to have validated the params, and end up having to deal with the possibility of missing or invalid params. The point of our validation layer is to simplify our code by _not_ having to account for  potentially invalid params and to avoid bugs when code forgets to account for them. By using service factories to inject request params, we're unable to lean on our validation layer properly.

  The right place for the "current `ModuleItemConfiguration`" is probably in the `context` rather than in a service. Remember that what `models.ModuleItemConfiguration` should really be called `models.Assignment`. A `models.Assignment` object seems like the appropriate "context" for an LTI assignment launch request. Or rather a `resources.AssignmentResource` object on which `context.assignment` is the `models.Assignment` object. This seems to make much more sense than the current `LTILaunchResource` class (what the hell is an "LTI launch resource"?). But this would be a larger refactoring.

* We could probably fairly easily add a read-write `assignment_svc.current_document_url` property that works with injected `request.params["tool_consumer_instance_guid"]` and `request.params["resource_link_id"]` on top of the `get_document_url(tool_consumer_instance_guid, resource_link_id)` and `set_document_url(tool_consumer_instance_guid, resource_link_id, document_url)` methods, so you can have it both ways: views can call the methods and pass in  `tool_consumer_instance_guid` and `resource_link_id` or they can use the property. I think there'd still be some oddness around what that properties does when the request params don't exist.

Again, I think it's controversial enough that it's worth addressing in a separate PR, if at all.


lms/services/assignment.py
```diff
@@ -0,0 +1,95 @@
...
+
+class AssignmentService:
+    """A service for getting and setting assignments (ModuleItemConfiguration's)."""
+
+    def __init__(self, db, tool_consumer_instance_guid, resource_link_id):
+        self._db = db
+        self._tool_consumer_instance_guid = tool_consumer_instance_guid
+        self._resource_link_id = resource_link_id
+
+    def get_document_url(self, tool_consumer_instance_guid=None, resource_link_id=None):
```
seanh, line 14: I'm leaving these as `get_document_url()` and `set_document_url()` for now as changing that is going to raise questions that can be dealt with in a separate PR

lms/services/assignment.py
```diff
@@ -0,0 +1,95 @@
...
+            )
+            .one_or_none()
+        )
+
+
+def factory(_context, request):
+    return AssignmentService(
+        db=request.db,
+        tool_consumer_instance_guid=request.params.get("tool_consumer_instance_guid"),
+        resource_link_id=request.params.get("resource_link_id"),
```
seanh, line 94: The `.get()`'s are necessary here because this service does get used by view predicates when serving requests that don't have these params, and if the factory crashes that causes BDD tests to fail. If the params are missing from the request then these will be `None` and `get_document_url()` will always return `None`

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -234,17 +235,14 @@ def test_it_enables_frontend_grading(self, context, pyramid_request):
...
-
+    def test_it_adds_the_document_url(
+        self, assignment_service, context, pyramid_request
+    ):
         db_configured_basic_lti_launch_caller(context, pyramid_request)
 
-        context.js_config.add_document_url.assert_called_once_with(mic.document_url)
+        context.js_config.add_document_url.assert_called_once_with(
+            assignment_service.get_document_url.return_value
+        )
```
seanh, line 245: It was nice when we improved these tests by replacing the previous mock with a factory. Now we've made them even nicer by replacing the factory with a new mock. The difference is that the code that we're mocking and the code that we're testing is now simpler because the service offers a simpler interface than the model classmethods did: it's just `get_document_url()` with no args. That's really easy to write isolated mock-based tests for

lms/services/assignment.py
```diff
@@ -0,0 +1,67 @@
+from functools import lru_cache
+
+from lms.models import ModuleItemConfiguration
```
robertknight, line 3: Do you have a plan to resolve the terminology conflict of `Assignment`s vs `ModuleItemConfiguration`s?

lms/services/assignment.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        Return the matching document URL or None.
+
+        Return the document URL for the assignment with the given
+        tool_consumer_instance_guid and resource_link_id, or None.
+        """
+        mic = self._get(tool_consumer_instance_guid, resource_link_id)
+
+        if mic:
+            return mic.document_url
```
robertknight, line 22: I don't know what the current convention is on the backend but my preference is always to have early-exits in a function be failure/no match cases (of which there may be many) and the success/match found path be the last return in the function

lms/services/assignment.py
```diff
@@ -0,0 +1,67 @@
...
+                    tool_consumer_instance_guid=tool_consumer_instance_guid,
+                )
+            )
+
+        # Clear the cache (@lru_cache) on self._get because we've changed the
+        # contents of the DB. (Python's @lru_cache doesn't have a way to remove
+        # just one key from the cache, you have to clear the entire cache.)
+        self._get.cache_clear()
+
+    @lru_cache(maxsize=None)
```
robertknight, line 54: According to https://docs.python.org/3/library/functools.html, `@cache` is a "smaller, faster" alternative than `@lru_cache` when there is no `maxsize`.

robertknight (approved):
LGTM. The main question I have is whether there is a plan to make the naming of the service (`AssignmentService`) and the model objects it works with (`ModuleItemConfiguration`) more consistent in future.

Regarding the design questions mentioned in this PR, I agree it makes sense to leave them out of scope here. I'll leave you to discuss that with @jon-betts when he is back next week.

lms/services/assignment.py
seanh: Maybe rename the model class at some point. I think renaming the actual DB table might be difficult

lms/services/assignment.py
seanh: Changed this to `return mic.document_url if mic else None`

lms/services/assignment.py
seanh: That was only added in Python 3.9 unfortunately

PR #2271, opened 2020-11-27
Add types for query-string

esanzgar:


robertknight (approved):
The npm page for the `@types/query-string` package has a big notice at the top saying that the package is deprecated and that `query-string` provides its own types: https://www.npmjs.com/package/@types/query-string. However that is only true of `query-string` v6 and we're currently using query-string v5.x.

The latest version of query-string (v6) will probably work now that we've dropped support for IE. Unfortunately the [README](https://github.com/sindresorhus/query-string) is a bit vague about browser support ("This module targets Node.js 6 or later and the latest version of Chrome, Firefox, and Safari."). We could switch to using the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) API in future which definitely supports all of our target browsers.

I think we can merge this PR now and revisit the above later though.

esanzgar:
Totally agree!

PR #2284, opened 2020-12-08
Improve the behavior or grade fetch errors

LMS007:
- Add a dialog that the alerts the user that an error has occurred when fetching a students grade.
- Kill off the grade loading spinner in the grade input field after an error.
- Add cancelLabel to ErrorDialog  / Dialog to override the button's label for more appropriate action naming.

-----

This is part of an effort to make the erros a bit more understandable from the customer's point of view to help our own triage efforts. We are seeing that GET requests to `api/lti/result?lis_result_sourcedid` failing in some customer environments. It's unclear why, but it was initially unclear what the problem was because failed GET requests were not handled in a reasonable way on the UI. It was previously assuming that such things were unlikely to fail. Since that is not the case, this simply adds the familiar error dialog when a GET request fails and kills off the loading spinner so the UI does not look like its still stuck.

To fake an error from a GET request, you can add

`if(data === undefined) {
     throw new ApiError(result.status, resultJson);
  }`

inside of `apiCall` in api.js

error

https://hypothesisuniversity.moodlecloud.com/mod/lti/view.php?id=110

relates to https://github.com/hypothesis/support/issues/124

### DIFF
--- a/lms/static/scripts/frontend_apps/components/Dialog.js
+++ b/lms/static/scripts/frontend_apps/components/Dialog.js
@@ -23,6 +23,7 @@ import { useUniqueId } from '../utils/hooks';
  * @prop {() => any} [onCancel] -
  *   A callback to invoke when the user cancels the dialog. If provided, a
  *   "Cancel" button will be displayed.
+ * @prop {string} [cancelLabel] - Label for the cancel button
  */
 
 /**
@@ -54,6 +55,7 @@ export default function Dialog({
   contentClass,
   initialFocus,
   onCancel,
+  cancelLabel = 'Cancel',
   role = 'dialog',
   title,
   buttons,
@@ -132,7 +134,7 @@ export default function Dialog({
               <Button
                 className="Button--cancel"
                 onClick={onCancel}
-                label="Cancel"
+                label={cancelLabel}
               />
             )}
             {buttons}
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -32,8 +32,9 @@ const GRADE_MULTIPLIER = 10;
  * and error status of that fetch request.
  *
  * @param {StudentInfo|null} student
+ * @param {(value: string) => void} setFetchGradeError
  */
-const useFetchGrade = student => {
+const useFetchGrade = (student, setFetchGradeError) => {
   const {
     api: { authToken },
   } = useContext(Config);
@@ -42,20 +43,22 @@ const useFetchGrade = student => {
 
   useEffect(() => {
     /** @type {boolean} */
-    let didCancel;
+    let ignoreResults;
     if (student) {
       // Fetch the grade from the service api
       // See https://www.robinwieruch.de/react-hooks-fetch-data for async in useEffect
       const fetchData = async () => {
         setGradeLoading(true);
         setGrade(''); // Clear previous grade so we don't show the wrong grade with the new student
-
-        const response = await fetchGrade({ student, authToken });
-        if (!didCancel) {
-          // Only set these values if we didn't cancel this request
-          if (response.currentScore) {
-            setGrade(scaleGrade(response.currentScore, GRADE_MULTIPLIER));
+        try {
+          const { currentScore } = await fetchGrade({ student, authToken });
+          if (!ignoreResults && currentScore) {
+            // Only set these values if we didn't cancel this request
+            setGrade(scaleGrade(currentScore, GRADE_MULTIPLIER));
           }
+        } catch (e) {
+          setFetchGradeError(e.errorMessage ? e.errorMessage : 'Unknown error');
+        } finally {
           setGradeLoading(false);
         }
       };
@@ -66,10 +69,10 @@ const useFetchGrade = student => {
     }
     // Called when unmounting the component
     return () => {
-      // Set a flag to cancel the the fetchGrade response from saving to state
-      didCancel = true;
+      // Set a flag to ignore the the fetchGrade response from saving to state
+      ignoreResults = true;
     };
-  }, [student, authToken]);
+  }, [student, authToken, setFetchGradeError]);
   return { grade, gradeLoading };
 };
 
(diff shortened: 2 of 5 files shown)


lms/static/scripts/frontend_apps/components/Dialog.js
```diff
@@ -132,7 +134,7 @@ export default function Dialog({
               <Button
                 className="Button--cancel"
                 onClick={onCancel}
-                label="Cancel"
+                label={cancelLabel}
```
LMS007, line 137: Adding a way to override this label because  "Cancel" does not make sense when presenting an error when the only action is to close the dialog.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -86,7 +91,8 @@ const useFetchGrade = student => {
  */
 export default function SubmitGradeForm({ student }) {
   // State for loading the grade
-  const { grade, gradeLoading } = useFetchGrade(student);
+  const [fetchGradeError, setFetchGradeError] = useState('');
+  const { grade, gradeLoading } = useFetchGrade(student, setFetchGradeError);
```
LMS007, line 95: The source of truth needs to be SubmitGradeForm, not useFetchGrade. So to make this work, the setFetchGradeError setter gets passed to `useFetchGrade`.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -215,11 +221,22 @@ export default function SubmitGradeForm({ student }) {
       </button>
       {!!submitGradeError && (
         <ErrorDialog
-          title="Error"
+          title="Submit Grade Error"
```
LMS007, line 224: This text helps differentiate between fetch and submit errors from customers.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -49,13 +50,17 @@ const useFetchGrade = student => {
...
+        try {
+          const response = await fetchGrade({ student, authToken });
+          if (!didCancel) {
+            // Only set these values if we didn't cancel this request
+            if (response.currentScore) {
+              setGrade(scaleGrade(response.currentScore, GRADE_MULTIPLIER));
+            }
           }
+        } catch (e) {
+          setFetchGradeError(e.errorMessage ? e.errorMessage : 'Unknown error');
```
esanzgar, line 62:
```suggestion
          setFetchGradeError(e.errorMessage ?? 'Unknown error');
```

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -49,13 +50,17 @@ const useFetchGrade = student => {
...
-          if (response.currentScore) {
-            setGrade(scaleGrade(response.currentScore, GRADE_MULTIPLIER));
+        try {
+          const response = await fetchGrade({ student, authToken });
+          if (!didCancel) {
+            // Only set these values if we didn't cancel this request
+            if (response.currentScore) {
+              setGrade(scaleGrade(response.currentScore, GRADE_MULTIPLIER));
+            }
           }
```
esanzgar, line 60: I haven't tested this, but I wonder if this is simpler:

```suggestion

       if (didCancel) { return; }
       
        try {
          const response = await fetchGrade({ student, authToken });
          if (response.currentScore) {
              setGrade(scaleGrade(response.currentScore, GRADE_MULTIPLIER));
          }
```

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -86,7 +91,8 @@ const useFetchGrade = student => {
  */
 export default function SubmitGradeForm({ student }) {
   // State for loading the grade
-  const { grade, gradeLoading } = useFetchGrade(student);
+  const [fetchGradeError, setFetchGradeError] = useState('');
```
esanzgar, line 94: To be consistent with the rest of the previous code this line should be inside the `useFetchGrade` and the `fetchGradeError` should be returned from that function.

esanzgar (approved):
I find a little bit unusual to have `useState` wrapped on the `useFetchGrade` function

esanzgar:
On a follow up PR, I suggest to:

* combine the three different states (grade, error and loading) into a single one `const fetchState = useState(initialState)`  and `const initialState = {loading: false, grade: null, error: null}`

* debounce the fetch requests so that quick changes in the student prop result in a fetch request.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -69,7 +74,7 @@ const useFetchGrade = student => {
       // Set a flag to cancel the the fetchGrade response from saving to state
       didCancel = true;
```
esanzgar, line 75: I would rename `didCancel` to `ignoreResults`, because the fetch requests is not actually cancelled.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
LMS007: I think thats reasonable!

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
LMS007: I think that won't work because errorMessage could be just an empty string IIRC.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
LMS007: Thats was the main issue I had in the first place. I'm not convinced encapsulating useFetchGrade is working for us anymore because I need to call setFetchGradeError when the cancel button is pressed in order to close out the dialog. But I also need to call it inside of useFetchGrade when an error is first detected. I supposed the right thing to do is to get rid of the custom effect unless there is a simpler solution I'm not seeing?

Issue #2280, opened 2020-12-07
Make the dialog buttons for URL and FilePicker the same


LMS007:
Currently the Submit button is disabled for the file picker, but not for the URL picker. Instead, the URL picker has a client-side popup message explaining that it won't work until you pick a URL.

We should adopt this behavior for both dialogs. The file picker is a bit confusing because the disabled styling looks like the enabled cancel button. So its unclear it cannot be clicked and then if you click it it, it's also unclear why it does not work. Additionally, making these 2 dialogs behave the same improves consistency.

![Dec-07-2020 14-40-04](https://user-images.githubusercontent.com/3939074/101414486-0337b180-389b-11eb-9514-cfa508d3cee5.gif)

esanzgar:
This [PR](https://github.com/hypothesis/lms/pull/2288/files) solves one of the aspects of this issue: the style of the disabled 'Select' button being identical to the 'Cancel' button.

It leaves the other aspect to be discussed: implementation of a 'popup' message warning the instructor to select a file from the list.

PR #2288, opened 2020-12-09
Improve UX of dialog buttons

LMS007:
- Improve UX of dialog buttons and make them more consistent with both the sidebar and the grader
- Solve issue where it was impossible to know if the submit button was disabled in the file picker

-------

In solving the original ticket. I thought initially it would be best to add the `required` attribute to the file picker, but that was not really the right solution as the UI was fundamentally different. The URL picker is just a field missing and the other was a selection of items from a list and the native `required` attribute would not work the same (or as plug-and-play'ish). Furthermore it does make sense to disable the button while things are loading and I think we should keep that behavoir.  _I think a follow up could be to use the client validator and replace the native required UX for the URLPicker as it does not really match much of anything currently._ 

The issues that became more apparent were how we were 1. dealing with the styling of a disabled button and 2. these dialogs are not at all consistent with the any other of our buttons in the sidebar or grader and I can't recall any reasoning for this? 

In any case, This changes the behavior to make buttons darker on hover (like our sidebar and grader) and removes the hover effect for disabled buttons (like our sidebar and grader). Note: WCAG compatibility does not have special requirement either since its decorative in case anyone was wondering.

https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html#:~:text=Incidental%3A%20Text%20or%20images%20of,content%2C%20have%20no%20contrast%20requirement.

Furthermore, the brand color (~red) seems to only be used for activated menu buttons in our sidebar e.g. (which I do like)

Screen Shot 2020-12-09 at 1 25 20 PM

But Since we really didn't have those cases in LMS, I just remove that styling for now.

dialog_buttons

Fixes https://github.com/hypothesis/lms/issues/2280

### DIFF
--- a/lms/static/styles/_variables.scss
+++ b/lms/static/styles/_variables.scss
@@ -6,6 +6,11 @@ $grey-2: #ececec;
 $grey-3: #dbdbdb;
 $grey-4: #a6a6a6;
 $grey-5: #7a7a7a;
+// Interim color variable for migration purposes, as the step between `$grey-5`
+// and `$grey-6` is large. Represents `base-mid` in proposed future palette,
+// minus blue tint.
+// This is the lightest grey admissible on $grey-2, $grey-3
+$grey-mid: #595959;
 $grey-6: #3f3f3f;
 $grey-7: #202020;
 
--- a/lms/static/styles/components/_Button.scss
+++ b/lms/static/styles/components/_Button.scss
@@ -6,7 +6,7 @@
 
   min-height: 30px;
 
-  background-color: var.$grey-6;
+  background-color: var.$grey-mid;
   border: none;
   border-radius: 2px;
   color: var.$white;
@@ -16,16 +16,9 @@
   padding-right: 15px;
   white-space: nowrap;
 
-  &:hover {
-    background-color: var.$brand;
-  }
-
-  &[disabled] {
-    color: var.$grey-6;
-    background-color: var.$grey-2;
-    &:hover {
-      background-color: var.$grey-2;
-    }
+  &:hover:not([disabled]),
+  &:focus:not([disabled]) {
+    background-color: var.$grey-6;
   }
 }
 
@@ -37,14 +30,10 @@
   color: var.$grey-6;
   background-color: var.$grey-2;
 
-  &:hover {
+  &:hover:not([disabled]),
+  &:focus:not([disabled]) {
     background-color: var.$grey-3;
-    color: var.$brand;
   }
-}
-
-.Button--danger {
-  background-color: var.$brand;
 }
 
 @include mixins.touch-input {
--- a/lms/static/styles/components/_Dialog.scss
+++ b/lms/static/styles/components/_Dialog.scss
@@ -42,19 +42,20 @@
   @include focus.outline-on-keyboard-focus;
   border: 0;
   background: none;
-  color: var.$grey-6;
+  color: var.$grey-mid;
 
   // Given the button a large hit target and ensure the 'X' label is large
   // enough to see easily and aligned with the right edge of the dialog.
   // Add negative margins so that the button does not force the dialog to
   // grow in height.
   font-size: 20px;
+  font-weight: 600;
   padding: 5px;
   margin: -10px 0px;
 
   &:hover {
     cursor: pointer;
-    color: var.$brand;
+    color: var.$grey-7;
   }
 }
 


lms/static/styles/_variables.scss
```diff
@@ -6,6 +6,11 @@ $grey-2: #ececec;
 $grey-3: #dbdbdb;
 $grey-4: #a6a6a6;
 $grey-5: #7a7a7a;
+// Interim color variable for migration purposes, as the step between `$grey-5`
+// and `$grey-6` is large. Represents `base-mid` in proposed future palette,
+// minus blue tint.
+// This is the lightest grey admissible on $grey-2, $grey-3
+$grey-mid: #595959;
```
LMS007, line 13: Pulled this terrible color in form the sidebar, It really need it  unfortunately  since it is the button color over there and its just too light if we use $grey-5.

lms/static/styles/components/_Button.scss
```diff
@@ -16,16 +16,9 @@
...
-  }
-
-  &[disabled] {
-    color: var.$grey-6;
-    background-color: var.$grey-2;
-    &:hover {
-      background-color: var.$grey-2;
-    }
+  &:hover:not([disabled]),
+  &:focus:not([disabled]) {
```
LMS007, line 20: This is how we do disabled styling in the sidebar

lms/static/styles/components/_Dialog.scss
```diff
@@ -42,19 +42,20 @@
...
   background: none;
-  color: var.$grey-6;
+  color: var.$grey-mid;
 
   // Given the button a large hit target and ensure the 'X' label is large
   // enough to see easily and aligned with the right edge of the dialog.
   // Add negative margins so that the button does not force the dialog to
   // grow in height.
   font-size: 20px;
+  font-weight: 600;
```
LMS007, line 52: beefed up the "X" button, It just looked too light to me and since the hover-color is no longer red, it needs more weight to be able to see that effect.

lms/static/styles/components/_Button.scss
```diff
@@ -37,16 +30,12 @@
...
 
-  &:hover {
+  &:hover:not([disabled]),
+  &:focus:not([disabled]) {
     background-color: var.$grey-3;
-    color: var.$brand;
   }
 }
 
-.Button--danger {
```
LMS007, line 46: This was unused css.

lms/static/styles/_variables.scss
esanzgar: Some alternative:

```suggestion
$grey-5\.5: #595959;
```

```suggestion
$grey-5-5: #595959;
```

```suggestion
$grey-5-tmp: #595959;
```

esanzgar: Why is this an interim color? Why don't we fully adopt it?

If we would adopt this new grey value, it may be makes sense to convert all:
*  `grey-7` to `grey-8`
* `grey-6` to `grey-7`

esanzgar (approved):
It is challenging to work with a monochromatic grey palette!

I was taking a look at [google's material design](https://material.io/components/buttons#hierarchy-and-placement) and they have a few good ideas that we could adopt. It maybe worth discussing these.

seanh:
This makes all the previously black buttons throughout the app grey instead (and they only turn fully black on hover), was that intentional?

Currently on master:

Screenshot from 2020-12-11 14-35-24

On this branch:

Screenshot from 2020-12-11 14-34-54
Screenshot from 2020-12-11 14-36-59
Screenshot from 2020-12-11 14-36-19
Screenshot from 2020-12-11 14-35-57

robertknight:
Hello - Just a heads up. To ensure consistency across the client and LMS frontend I think we should review this collectively. In particular we should get input from @lyzadanger who has done much of the related work on the client side. We also have a number of other button-related issues we need to discuss.

robertknight (changes requested):
I haven't had a chance to look at this in detail yet, but can we hold off merging this for the moment per my previous comment. There was also a related Slack thread here: https://hypothes-is.slack.com/archives/C1M8NH76X/p1607695129241700.

LMS007:
> This makes all the previously black buttons throughout the app grey instead (and they only turn fully black on hover), was that intentional?


Yes, you're exactly correct and this also matches the style in the sidebar and grader. It has been changed for consistency and I did not question validity. Perhaps we should at this juncture? We have a lot of gray and it might serve us well to have some colors here and there.

lms/static/styles/_variables.scss
LMS007: I think we do revisit this rule next week. It was dropped in to the client as  a shim many moons ago, but we did not expect it to be long lived. We were at that time in the process of developing a new color palette but that was put on hold so its long outlived it expected life span.

We should keep the name the same as the client (if we do end up using it) OR move to a new pallet entirely and then abandon it.

PR #2289, opened 2020-12-10
Add custom validation message to URLPicker

LMS007:
This matches our validation UI for the grader and brings constancy to the app.

-------

This is mostly a cosmetic change that came about from this work:
https://github.com/hypothesis/lms/pull/2288

This just removes the use of the native form popup error and instead replaces it with our own branded validation popup that matches the grader. (somewhat still experimental) 

Current:
old

Changed: (also shows new CSS of buttons in prev. PR)
new

### DIFF
--- a/lms/static/scripts/frontend_apps/components/URLPicker.js
+++ b/lms/static/scripts/frontend_apps/components/URLPicker.js
@@ -1,8 +1,9 @@
 import { createElement } from 'preact';
-import { useRef } from 'preact/hooks';
+import { useRef, useState } from 'preact/hooks';
 
 import Button from './Button';
 import Dialog from './Dialog';
+import ValidationMessage from './ValidationMessage';
 
 /**
  * @typedef URLPickerProps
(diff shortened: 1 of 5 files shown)


lms/static/scripts/frontend_apps/components/URLPicker.js
```diff
@@ -20,23 +21,52 @@ import Dialog from './Dialog';
...
+  /**
+   * If any input is detected, close the ValidationMessage.
+   */
+  const handleKeyDown = () => {
+    setValidationError(false);
+  };
+
+  const submitButton = (
+    <span className="URLPicker__buttons">
+      {validationMessage && (
```
LMS007, line 50: This code is mostly just borrowed from SubmitGradeForm, we have to do a tiny bit of trickery here and check if the message exists the very first time things are rendered to prevent the animation from quickly showing, and then collapsing the red valitation banner. For this reason it helps to keep tract of both the open/closed state and a message state which is initially a blank string.

lms/static/styles/components/_Dialog.scss
```diff
@@ -66,8 +66,6 @@
 
   .Button {
     @include focus.outline-on-keyboard-focus;
-    &:not(:first-child) {
-      margin-left: 5px;
-    }
+    margin-left: 5px;
```
LMS007, line 69: It's not perfect, but it also It does not hurt to render the margin on the first button. This rule was causing a problem with the submit button since it was nested in a new parent wrapper. The 5px margin was only being applied after the validation message opened. The super simple fix is to just apply the rule on all action buttons in the dialog. I think the most we ever see is 3 buttons IIRC.

lms/static/styles/components/_URLPicker.scss
```diff
@@ -0,0 +1,39 @@
+@keyframes validationErrorOpen--url {
```
LMS007, line 1: I had to override some of the css to get the message to fit better on the URLPicker.
300px vs 200px on the width.

Things are still a bit hard coded due to absolute pos, perhaps there is opportunity to improve, and generalize, but not at this juncture.

lms/static/styles/components/_URLPicker.scss
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+.URLPicker {
+  &__buttons {
+    position: relative;
+  }
+  .ValidationMessage {
+    line-height: inherit;
+    height: 100%;
+    right: 73px;
```
LMS007, line 29: Get it to render just 1px inside the submit buttons left edge.

seanh:
Should the error message point to or be attached to the URL `<input>` (which is where the mistake is) rather than to the <kbd>Submit</kbd> button?

LMS007:
In thinking more about this, we likely need to attach it to the field as Sean suggested, but it becomes a bit wide and awkward. I think an arrow might be the best approach in combination with putting the popup above the field (or below) and that would require some design which is way beyond the scope of this little fix. Nothing worth chasing down atm.

Issue #2282, opened 2020-12-08
Fix course copy in Brightspace


seanh:

seanh:
It looks like Brightspace includes the `resource_link_id` of the original assignment in a non-standard and undocumented `ext_d2l_resource_link_id_history` LTI 1 launch param. We can use that to make course copy work in the same way as we did for Blackboard with its `resource_link_id_history` param.

https://hypothes-is.slack.com/archives/CBN3DGW02/p1607605568143400

PR #2293, opened 2020-12-10
Add Brightspace course copy support

seanh:
Add Brightspace course copy support. Fixes https://github.com/hypothesis/lms/issues/2282

Course copy works exactly the same in Brightspace as it does in Blackboard: the copied assignments get a new `resource_link_id` so we can't find their settings in our DB, but the original `resource_link_id` is supplied in an additional launch param. In Blackboard this additional launch param is called `resource_link_id_history`, in Brightspace it's `ext_d2l_resource_link_id_history`.

There's some work in this PR to make previously Blackboard-specific code generic. The actual amount of new code we'd have to add if we came across a third LMS that worked in a new way (but with a third param name) is pretty minimal / the amount of Blackboard or Brightspace-specific code is pretty minimal. It's roughly this:

```python
# views/basic_lti_launch.py:

@view_config(blackboard_copied=True)
def blackboard_copied_basic_lti_launch(self):
    return self.course_copied_basic_lti_launch(BlackboardCopied.get_original_resource_link_id(self.request))

@view_config(brightspace_copied=True)
def brightspace_copied_basic_lti_launch(self):
    return self.course_copied_basic_lti_launch(BrightspaceCopied.get_original_resource_link_id(self.request))

# predicates/_lti_launch.py:

class BlackboardCopied(_CourseCopied):
    name = "blackboard_copied"
    param_name = "resource_link_id_history"

class BrightspaceCopied(_CourseCopied):
    name = "brightspace_copied"
    param_name = "ext_d2l_resource_link_id_history"
```

So far the original `resource_link_id` can always be found by just reading another launch param, and only the name of the param changes. But the above code design should extend to other LMS's that need to look up the original launch param in another way. For example:

```python
# views/basic_lti_launch.py:

@view_config(moodle_copied=True)
def moodle_copied_basic_lti_launch(self):
    # Since we have a separate *_copied_basic_lti_launch() view for each LMS we
    # _could_ do other Moodle-specific stuff here if it were necessary.

    return self.course_copied_basic_lti_launch(MoodleCopied.get_original_resource_link_id(self.request))

# predicates/_lti_launch.py:

class MoodleCopied(_CourseCopied):
    name = "moodle_copied"

    @staticmethod
    def get_original_resource_link_id(request):
        moodle_api_svc = request.find_service(name="moodle_api")
        return moodle_api_svc.get_original_resource_link_id(request)
```

### Longer-term code design

Each view and corresponding view predicate are closely related. I think a code reorganization that gets them into the same files might be an improvement. This'd be a lot more work and raise all sorts of questions so I haven't tackled it yet.

```
lms/
  views/
    basic_lti_launch/
      _base.py:
        class BasicLTILaunchViewBaseClass:
            ...

      blackboard.py:
        class BlackboardCopiedViewPredicate:
            ...

        class BlackboardBasicLTILaunchView(BasicLTILaunchViewBaseClass):
            @view_config(blackboard_copied=True)
            def blackboard_copied_basic_lti_launch(self):
                ...

      brightspace.py:
        class BrightspaceCopiedViewPredicate:
            ...

        class BrightspaceBasicLTILaunchView(BasicLTILaunchViewBaseClass):
            @view_config(brightspace_copied=True)
            def brightspace_copied_basic_lti_launch(self):
                ...
```

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -21,6 +21,7 @@ from lms.validation import (
     URLConfiguredBasicLTILaunchSchema,
 )
 from lms.validation.authentication import BearerTokenSchema
+from lms.views.predicates import BlackboardCopied, BrightspaceCopied
 
 
 @view_defaults(
@@ -131,21 +132,42 @@ class BasicLTILaunchViews:
         """
         Respond to a launch of a newly-copied Blackboard assignment.
 
-        Find the document_url for resource_link_id_history and make a copy of
-        it with the new resource_link_id, then lanuch the assignment as normal.
-
-        For more about Blackboard course copy and the resource_link_id_history
-        param see the BlackboardCopied predicate's docstring.
+        For more about Blackboard course copy see the BlackboardCopied
+        predicate's docstring.
         """
-        # The ``blackboard_copied=True`` view predicate ensures that this view
-        # won't be called if there isn't a resource_link_id_history param and a
-        # matching document_url in the DB.
+        return self.course_copied_basic_lti_launch(
+            BlackboardCopied.get_original_resource_link_id(self.request)
+        )
+
+    @view_config(brightspace_copied=True)
+    def brightspace_copied_basic_lti_launch(self):
+        """
+        Respond to a launch of a newly-copied Brightspace assignment.
+
+        For more about Brightspace course copy see the BrightspaceCopied
+        predicate's docstring.
+        """
+        return self.course_copied_basic_lti_launch(
+            BrightspaceCopied.get_original_resource_link_id(self.request)
+        )
+
+    def course_copied_basic_lti_launch(self, original_resource_link_id):
+        """
+        Respond to a launch of a newly-copied assignment.
+
+        Find the document_url for the original assignment and make a copy of it
+        with the new resource_link_id, then launch the assignment as normal.
+
+        Helper method for the *_copied_basic_lti_launch() methods above.
+
+        :param original_resource_link_id: the resource_link_id of the original
+            assignment that this assignment was copied from
+        """
         tool_consumer_instance_guid = self.request.params["tool_consumer_instance_guid"]
         resource_link_id = self.request.params["resource_link_id"]
-        resource_link_id_history = self.request.params["resource_link_id_history"]
 
         document_url = self.assignment_service.get_document_url(
-            tool_consumer_instance_guid, resource_link_id_history
+            tool_consumer_instance_guid, original_resource_link_id
         )
 
         self.assignment_service.set_document_url(
--- a/lms/views/predicates/_lti_launch.py
+++ b/lms/views/predicates/_lti_launch.py
@@ -1,4 +1,6 @@
 """Predicates for use with LTI launch views."""
+from abc import ABC
+
 from lms.views.predicates._helpers import Base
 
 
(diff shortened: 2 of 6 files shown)


lms/views/predicates/_lti_launch.py
```diff
@@ -42,7 +44,7 @@ def __call__(self, context, request):
         return has_document_url == self.value
 
 
-class BlackboardCopied(Base):
+class _CourseCopied(Base, ABC):  # pragma: nocover
```
robertknight, line 47: Why is the `# pragma: nocover` needed here? Is that because coverage reports will otherwise complain about the abstract methods not being executed?

lms/views/basic_lti_launch.py
```diff
@@ -137,12 +137,36 @@ def blackboard_copied_basic_lti_launch(self):
...
+    def course_copied_basic_lti_launch(self, param_name):
+        """
+        Respond to a launch of a newly-copied assignment.
+
+        Helper method for the *_copied_basic_lti_launch() methods above.
+
+        :param param_name: the name of the launch param that contains the
+        resource_link_id of the original assignment that this assignment was
+        copied from
+        """
```
seanh, line 166: Added a helper method for `blackboard_copied_basic_lti_launch()` and `brightspace_copied_basic_lti_launch()`, since the two methods are exactly the same except for the `param_name`. I prefer having two separate views that call a helper method, rather than a single view that looks for both param names, the code works out simpler.

lms/views/predicates/_lti_launch.py
```diff
@@ -42,7 +44,40 @@ def __call__(self, context, request):
         return has_document_url == self.value
 
 
-class BlackboardCopied(Base):
+class _CourseCopied(Base, ABC):
```
seanh, line 47: Added a base class for `BlackboardCopied` and `BrightspaceCopied`, since they're the same except for the param name.

I'm not sure that I think we should actually be using `ABC`'s, we're just writing an app not a framework, and I think we can just do "informal" protocols and abstract base classes. Perhaps with `raise NotImplementedError()`. But the pre-existing `Base` class is already an `ABC` and that forces us to make this new abstract class an `ABC` too otherwise PyLint complains. For the sake of keeping this PR small I just went along with it. Refactoring `Base` to not be an `ABC` could be done separately if we want to

lms/views/basic_lti_launch.py
```diff
@@ -131,21 +132,42 @@ def blackboard_copied_basic_lti_launch(self):
...
+
+        Find the document_url for the original assignment and make a copy of it
+        with the new resource_link_id, then launch the assignment as normal.
+
+        Helper method for the *_copied_basic_lti_launch() methods above.
 
-        For more about Blackboard course copy and the resource_link_id_history
-        param see the BlackboardCopied predicate's docstring.
+        :param original_resource_link_id: the resource_link_id of the original
+            assignment that this assignment was copied from
```
seanh, line 164: Added a helper method for `blackboard_copied_basic_lti_launch()` and `brightspace_copied_basic_lti_launch()`, since the two methods are exactly the same except for the `param_name`. I prefer having two separate views that call a helper method, rather than a single view that looks for both param names, the code works out simpler.

lms/views/predicates/_lti_launch.py
```diff
@@ -71,34 +130,38 @@ def blackboard_course_copied_assignment_launch_view(context, request):
...
+    For example:
+
+        @view_config(..., brightspace_copied=True)
+        def brightspace_course_copied_assignment_launch_view(context, request):
+            ...
+
+    ext_d2l_resource_link_id_history is a non-standard and undocumented param
+    that's only used by Brightspace, but nothing in this code actually prevents
+    us from using ext_d2l_resource_link_id_history if another LMS sends it to
+    us.
```
jon-betts, line 160: I think these comments are a kind of code repetition / maintainance problem. The base class could probably do with a description of what it is as well.

We can probably have our cake and eat it by genericising this a bit and adding it to the base class, and then replacing this and the Blackboard ones with something like:

```
Brightspace custom parameter based course copy detection (see base class for details)

Usage:

        @view_config(..., brightspace_copied=True)
        def brightspace_course_copied_assignment_launch_view(context, request):
            ...

jon-betts (changes requested):
The actual code and testing here all looks good to me. This seems a pretty straight generalisation of what we already had, which is good!

My feedback is that there are two pretty much copy and paste comments in the classes and none in the base class. I think we'd be better off putting a generalised version of the very big docstring in the base class, and short descriptions in the sub-classes. Basically to follow the style of the code with the documentation too.

jon-betts (approved):
Much better!

Issue #2258, opened 2020-11-23
Automatically fix course-copied Canvas Files assignments


seanh:
See also:

* [EPIC: Course copy](https://github.com/hypothesis/lms/issues/2241)
* [Show error message for course-copied Canvas Files assignments](https://github.com/hypothesis/lms/issues/2217)
* [Make it easier for instructors to manually fix course-copied Canvas File assignments](https://github.com/hypothesis/lms/issues/2259)

In https://github.com/hypothesis/lms/issues/2217 we implemented a basic course copy solution for Canvas Files assignments: when a student or teacher launches a Canvas Files assignment that has been broken by course copy they're shown an error message (for teachers the error message explains how they can manually fix the assignment).

This solution has some limitations:

1. Teachers have to actually launch every Hypothesis assignment in the course to see the error message, if they don't test all the assignments teachers might not realise the assignment is broken until students start running into an error message
2. Even when teachers do see the error message, it requires them to manually fix each assignment (following instructions that we provide). We could be more helpful than this

This card is to make course-copied Canvas Files assignments just work without teachers having to do anything to fix them:

### Problem

When a Canvas course is copied any Canvas Files assignment are broken. Students run into this error message:

![Screenshot from 2020-11-23 13-41-55](https://user-images.githubusercontent.com/22498/99968923-b24f9580-2d91-11eb-901d-2fd124b37a5f.png)

Teachers run into this alternate error message:

![Screenshot from 2020-11-23 13-44-44](https://user-images.githubusercontent.com/22498/99969208-0fe3e200-2d92-11eb-8258-cfe77286d8f0.png)

### Solution

Course-copied Canvas Files assignments should just work.

So far our only idea for how to implement this is:

* When a teacher launches a Canvas Files assignment we can detect whether it might be a broken course-copied assignment by looking for the assignment's file's ID in the list of all files in the course. If the file ID isn't in the course, the assignment is broken. (This is how we currently decide to show teachers the error message shown above.)

* When we detect that a teacher is launching what might be a broken course-copied Canvas Files assignment we can get the list of all the files on the course and look for a file whose path and filename exactly matches that of the old file from the previous course. If we find a match we can silently and automatically fix the assignment by updating it to use the matching file's ID instead

  * This would require us to have recorded the previous file's path and name somewhere (e.g. in our DB), which we don't do currently
  * Technically false-positives are possible (two _different_ files with the same exact name) but this is probably unlikely enough that we can ignore it
[...]

PR #2311, opened 2020-12-17
Fix course copied Canvas files assignments (proof of concept)

seanh:
Fixes https://github.com/hypothesis/lms/issues/2258 (proof of concept only)

## Testing

- Launching a normal (non-copied) Canvas file assignment:
  - As an instructor:
    - [x] Without existing file records in the DB (launches correctly and adds the file records to the DB)
    - [x] With existing file records in the DB (launches correctly)
  - As a student:
    - [x] Without existing file records in the DB (launches correctly and adds the file records to the DB)
    - [x] With existing file records in the DB (launches correctly)
- Launching a copied Canvas file assignment
  - When the original assignment has not been launched on this branch (so we don't have a record of the original file in the DB):
    - [x] As a teacher (shows appropriate error dialog) 
    - [x] As a student (shows appropriate error dialog) 
  - When the original assignment has been launched on this branch (so we do have a record of the original file in the DB):
    - As an instructor:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
    - As a student:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
- Launching a Canvas file assignment that was _copied from a copy_ of an original assignment:
  - When the original assignment has not been launched on this branch (so we don't have a record of the original file in the DB):
    - [x] As a teacher (shows appropriate error dialog) 
    - [x] As a student (shows appropriate error dialog) 
  - When the original assignment has been launched on this branch (so we do have a record of the original file in the DB):
    - As an instructor:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
    - As a student:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
  - Now trying launching the intermediate assignment (the first copy, rather than the copy-of-the-copy):
    - As an instructor:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
    - As a student:
      - [x] Without existing file records for the _new_ course in the DB (launches correctly and adds the file records to the DB)
      - [x] With existing file records for the _new_ course in the DB (launches correctly)
- Launching an assignment whose file is unpublished
  - [x] As a teacher (it should work fine because the teacher can access the file)
  - [x] As a student (they should see an error page because they can't access the file)
  - [x] As a student, when the course also contains a _published_ file with the same filename and size (it does not change the assignment to use the other file)
- Assignment whose file has been deleted
- [x] When there are two files with the same name and size in the same course

### DIFF
--- a/lms/models/__init__.py
+++ b/lms/models/__init__.py
@@ -10,6 +10,7 @@ from lms.models.lti_launches import LtiLaunches
 from lms.models.lti_user import LTIUser, display_name
 from lms.models.module_item_configuration import ModuleItemConfiguration
 from lms.models.oauth2_token import OAuth2Token
+from lms.models.canvas_file import CanvasFile
 
 
 def includeme(_config):
--- a/lms/models/canvas_file.py
+++ b/lms/models/canvas_file.py
@@ -1,0 +1,21 @@
+from sqlalchemy import Column, UnicodeText, String, Integer
+
+from lms.db import BASE
+
+
+class CanvasFile(BASE):
+    __tablename__ = "canvas_file"
+
+    id = Column(Integer, autoincrement=True, primary_key=True)
+
+    consumer_key = Column(String)
+    tool_consumer_instance_guid = Column(UnicodeText)
+
+    # The course ID (the context_id launch param).
+    course_id = Column(UnicodeText)
+
+    # Metadata about the file, from the Canvas API.
+    filename = Column(UnicodeText)
+    size = Column(UnicodeText)
+    file_id = Column(Integer)
+    file_id_override = Column(UnicodeText)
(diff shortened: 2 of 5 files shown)


seanh:
Closing this proof-of-concept, will break it up into a series of production PRs

PR #2313, opened 2020-12-18
WIP: Refactor submit grade form

esanzgar:


robertknight:
Hi @esanzgar - When you have time for a context switch could you add some notes here about what the PR aims to achieve and how it works? Can you also decide whether you want to finish this up or close it for the time being?

I can see that there is a switch from using several independent state values to a `useReducer` and in general I'm supportive of the use of this Redux-lite technique in places where it helps to prevent bugs or otherwise simplifies the code.

esanzgar:
This PR was a proposal to simplify the handling of the component state. It was based on these two articles:

- https://www.robinwieruch.de/react-hooks-fetch-data
- https://dev.to/housinganywhere/matching-your-way-to-consistent-states-1oag

PR #2314, opened 2020-12-18
Add temporary logging to help debug XML error

LMS007:
Add logging to help see the XML response from customer's LMS that may be potentially causing the issue related to "Malformed LTI outcome response"

-------

The goal here is to see the XML response (as a dict) of the customer's LMS when requesting a grade. 
relates to https://github.com/hypothesis/support/issues/124


seanh:
@LMS007 I've fixed this up to log the actual XML body as a string, as part of logging the exception. Should work

seanh (approved):

PR #2315, opened 2020-12-18
Prevent error support text from having scroll bars

LMS007:
Allow long text that details how to open a support ticket to wrap in small dialogs rather than displaying a horizontal scrollbar

Screen Shot 2020-12-18 at 1 03 36 PM


seanh:
Looks good. I noticed that there's a bunch of differences between this particular dialog and the rest of the dialogs in the app. Opened a separate issue for those: https://github.com/hypothesis/lms/issues/2320

seanh (approved):

PR #2325, opened 2020-12-22
Record Canvas files in the DB

seanh:
Depends on: https://github.com/hypothesis/lms/pull/2323, https://github.com/hypothesis/lms/pull/2327

Save records of the Canvas files that we've seen in our DB, whenever we receive data about files from the Canvas API

### DIFF
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -3,6 +3,7 @@
 import marshmallow
 from marshmallow import EXCLUDE, Schema, fields, post_load, validate, validates_schema
 
+from lms.models import CanvasFile
 from lms.services import CanvasAPIError, CanvasFileNotFoundInCourse
 from lms.validation import RequestsResponseSchema
 
@@ -35,13 +36,17 @@ class CanvasAPIClient:
         Canvas API request fails for any other reason
     """
 
-    def __init__(self, authenticated_client):
+    def __init__(self, authenticated_client, canvas_files_svc, lti_user):
         """
         Create a new CanvasAPIClient.
 
-        :param authenticated_client: An instance of AuthenticatedClient
+        :param authenticated_client: an instance of AuthenticatedClient
+        :param canvas_files_svc: the "canvas_files" service
+        :param lti_user: the current models.LTIUser
         """
         self._client = authenticated_client
+        self._canvas_files_svc = canvas_files_svc
+        self._lti_user = lti_user
 
     def get_token(self, authorization_code):
         """
@@ -210,12 +215,26 @@ class CanvasAPIClient:
         # For documentation of this request see:
         # https://canvas.instructure.com/doc/api/files.html#method.files.api_index
 
-        return self._client.send(
+        files = self._client.send(
             "GET",
             f"courses/{course_id}/files",
             params={"content_types[]": "application/pdf"},
             schema=self._ListFilesSchema,
         )
+
+        for file_dict in files:
+            self._canvas_files_svc.upsert(
+                CanvasFile(
+                    consumer_key=self._lti_user.oauth_consumer_key,
+                    tool_consumer_instance_guid=self._lti_user.tool_consumer_instance_guid,
+                    course_id=course_id,
+                    file_id=file_dict["id"],
+                    filename=file_dict["filename"],
+                    size=file_dict["size"],
+                )
+            )
+
+        return files
 
     class _ListFilesSchema(RequestsResponseSchema):
         """Schema for the list_files response."""
@@ -223,8 +242,10 @@ class CanvasAPIClient:
         many = True
 
         display_name = fields.Str(required=True)
+        filename = fields.Str(required=True)
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
+        size = fields.Int(required=True)
 
     def check_file_in_course(self, file_id, course_id):
         """
--- a/lms/services/canvas_files.py
+++ b/lms/services/canvas_files.py
@@ -1,0 +1,30 @@
+from lms.models import CanvasFile
+
+
+class CanvasFilesService:
+    """Methods for working with the models.CanvasFile DB model."""
+
+    def __init__(self, db):
+        self._db = db
+
+    def upsert(self, canvas_file):
+        existing_file = (
+            self._db.query(CanvasFile)
+            .filter_by(
+                consumer_key=canvas_file.consumer_key,
+                tool_consumer_instance_guid=canvas_file.tool_consumer_instance_guid,
+                course_id=canvas_file.course_id,
+                file_id=canvas_file.file_id,
+            )
+            .one_or_none()
+        )
+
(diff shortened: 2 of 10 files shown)


lms/services/canvas_files.py
```diff
@@ -0,0 +1,30 @@
...
+                file_id=canvas_file.file_id,
+            )
+            .one_or_none()
+        )
+
+        if existing_file:
+            existing_file.filename = canvas_file.filename
+            existing_file.size = canvas_file.size
+        else:
+            self._db.add(canvas_file)
```
seanh, line 26: This is a single-method service for now but in https://github.com/hypothesis/lms/pull/2328 it'll grow additional methods for searching for and getting files from the DB

tests/factories/canvas_file.py
```diff
@@ -0,0 +1,17 @@
...
+
+CanvasFile = make_factory(
+    models.CanvasFile,
+    FACTORY_CLASS=SQLAlchemyModelFactory,
+    application_instance=SubFactory(ApplicationInstance),
+    tool_consumer_instance_guid=TOOL_CONSUMER_INSTANCE_GUID,
+    file_id=Faker("random_int"),
+    course_id=Faker("random_int"),
+    filename=Faker("file_name", extension="pdf"),
+    size=Faker("random_int"),
```
seanh, line 16: TODO: Add `file_id_history` here

tests/unit/lms/services/canvas_files_test.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+
+from lms.models import CanvasFile
+from lms.services.canvas_files import factory
+from tests import factories
+
+
+class TestUpsert:
+    def test_if_theres_no_record_in_the_db_it_adds_one(self, db_session, svc):
+        canvas_file = factories.CanvasFile.build()
```
seanh, line 12: This creates a `CanvasFile` object _without_ adding it to the DB session (factory boy's `build()` method)

tests/unit/lms/services/canvas_files_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_if_theres_a_record_in_the_db_it_updates_it(self, db_session, svc):
+        existing_file = factories.CanvasFile()
+        new_file = factories.CanvasFile.build(
+            consumer_key=existing_file.application_instance.consumer_key,
+            tool_consumer_instance_guid=existing_file.tool_consumer_instance_guid,
+            course_id=existing_file.course_id,
+            file_id=existing_file.file_id,
+            filename="NEW" + existing_file.filename,
+        )
```
seanh, line 37: `existing_file` _is_ saved to the DB, `new_file` isn't

lms/services/canvas_files.py
seanh: Note on the design of this `upsert()` method:

## Fields as separate params :negative_squared_cross_mark: 

In the past we've often written service methods that encapsulate the model classes and just accept the fields as separate arguments, like this:

```python
def upsert(self, consumer_key, tool_consumer_instance_guid, course_id, file_id, filename, size):
    ...
```

This doesn't work well because it creates methods with long argument lists. PyLint rightly complains. Adding a new field to the model class necessitates adding another argument to the method and then updating every place that calls the method (including every test).

## Combine params into value object :negative_squared_cross_mark: 

You can beat the too-many-arguments problem by providing a value class to hold them all in a single object. You could put the value class in `canvas_files.py` or even inside the `CanvasFilesService` class as an inner class:

```python
class CanvasFilesService:

    class CanvasFile(NamedTuple):
        consumer_key: str
        course_id: int
        file_id: int
        filename: str
        size: int

    def upsert(self, canvas_file):
        ...
```

The view code would then do something like:

```python
canvas_files_svc.upsert(CanvasFilesService.CanvasFile(...))
```

The problem with this is that the `CanvasFilesService.CanvasFile` is pretty much a duplicate of `models.CanvasFile` and we'll end up needing these value classes for many of our model classes

## Combine params into dict :negative_squared_cross_mark: 

Defeat the too-many-arguments problem by putting them all in a data dict:

```python
def upsert(self, data):
    existing_file = (
        self._db.query(CanvasFile)
        .filter_by(
            consumer_key=data["consumer_key"],
            tool_consumer_instance_guid=data["tool_consumer_instance_guid"],
            course_id=data["course_id"],
            file_id=data["file_id"],
        )
        .one_or_none()
    )

    ...
```

But data dicts suck because they just obfuscate the method's actual params. The method effectively does still have a long argument list but you now have to read the entire method body to find out what all the arguments are. This approach also tends to lead to methods that pass their data dicts into other methods wholesale, which is even worse.

## Just use the model! :heavy_check_mark: 

Services don't have to encapsulate models. It's okay for views to instantiate model objects and pass them into services. This is the solution I went for:

```python
def upsert(self, canvas_file):
    ...
```

The view then does:

```python
canvas_files_svc.upsert(models.CanvasFile(consumer_key="foo", ...))
```

This works particularly well with our thin models approach because:

* All model objects have the same interface when it comes to constructing them. You can always pass all the fields as kwargs:

  ```python
  canvas_file = CanvasFile(consumer_key="foo", ...)
  ```

  You can also always assign all the fields:

  ```python
  canvas_file = CanvasFile()
  canvas_file.consumer_key = "foo"
  ...
  ```

  This is the normal behaviour of SQLAlchemy ORM model classes and of Factory Boy factories, and our non-SQLAlchemy models also behave the same.

* You can create a model object without adding it to the DB. You can create a model object that will _never_ get added to the DB and just pass it around as a value. `canvas_file = CanvasFile()` doesn't add anything to the DB session.

* You can create an invalid model object. For example you can create a `CanvasFile` whose `consumer_key` doesn't match any `ApplicationInstance` in the DB, or one that has no `tool_consumer_instance_guid` even though that column isn't nullable, etc. This is totally fine as long as you're just using it as a value and not adding it to the DB (or as long as someone fills out the missing fields before it does get added to the DB).

* Because of our thin models approach (no `__init__()` methods, and few or no methods in general on model classes) our models are close to being value objects (like `NamedTuple`'s) and we can get away with treating them as such

lms/services/canvas_api/client.py
```diff
@@ -210,21 +215,37 @@ def list_files(self, course_id):
...
+                    consumer_key=self._lti_user.oauth_consumer_key,
+                    tool_consumer_instance_guid=self._lti_user.tool_consumer_instance_guid,
+                    course_id=course_id,
+                    file_id=file_dict["id"],
+                    filename=file_dict["filename"],
+                    size=file_dict["size"],
+                )
+            )
+
+        return files
```
seanh, line 237: I think we might want to combine this into a single call: `self._canvas_files_svc.upsert([CanvasFile(...) for file_dict in files])`. That way the service method (which now accepts multiple `CanvasFile`'s for upserting) can be more efficient and do just a single `SELECT` and a single `UPSERT`, instead of one per file.

PR #2327, opened 2020-12-22
Split up a large test class

seanh:
This doesn't actually change any of the tests, it just splits the massive `TestCanvasAPIClient` class into separate classes, one for each `CanvasAPIClient` method that's tested.

This is going to make future commits that need to add more tests to this file much nicer.

It also has a few advantages even just for the existing tests:

* Smaller classes are easier to understand than big ones

* The scope of fixtures is reduced. Most of the tests don't actually use any of the fixtures defined in this file, and now that they're in separate, small test classes that clearly don't have any fixtures in the test class, that's easier to see. It makes it clear that the tests are as simple as they actually are

* Reducing the scope of fixtures also means that for test methods that _do_ use fixtures, those fixtures are defined much closer to where they're used (inside the one small test class that uses them)

* We can use class-level autouse fixtures when all tests for a given method, but not the tests for any other method, should use certain fixtures. Previously this required an `@pytest.mark.usefixtures(...)` decorator on each individual test method for the given source method.

* It makes test names shorter and less repetitive. Instead of:

      class TestCanvasAPIClient:
          def test_authenticated_users_sections_does_x(self):
              ...

          def test_authenticated_users_sections_does_y(self):
              ...

  It's now:

      class TestAuthenticatedUsersSections:
          def test_it_does_x(self):
              ...

          def test_it_does_y(self):
              ...

  Leaves a lot more space for the `does_x` and `does_y` part of the test name (the important part) to be longer and more expressive.

### DIFF
--- a/tests/unit/lms/services/canvas_api/client_test.py
+++ b/tests/unit/lms/services/canvas_api/client_test.py
@@ -12,12 +12,11 @@ from lms.services import (
 from lms.services.canvas_api.client import CanvasAPIClient
 from tests import factories
 
+pytestmark = pytest.mark.usefixtures("http_session", "oauth_token")
 
-class TestCanvasAPIClientGetToken:
-    # This is the only test where we fake out the underlying class, because
-    # this _one_ call is just a pass through.
 
-    def test_get_token(self, canvas_api_client, authenticated_client):
+class TestGetToken:
+    def test_it(self, canvas_api_client, authenticated_client):
         token = canvas_api_client.get_token(sentinel.authorization_code)
 
         authenticated_client.get_token.assert_called_once_with(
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -12,12 +12,11 @@
...
 from tests import factories
 
+pytestmark = pytest.mark.usefixtures("http_session", "oauth_token")
 
-class TestCanvasAPIClientGetToken:
-    # This is the only test where we fake out the underlying class, because
-    # this _one_ call is just a pass through.
 
-    def test_get_token(self, canvas_api_client, authenticated_client):
+class TestGetToken:
```
seanh, line 18: This is our naming style for test classes:

* If the class-under-test is small enough to cover with a single test class, it's just `TestFooClass`
* If the class-under-test is large, then it's one `TestBarMethod` class for each `FooClass.bar_method()`

tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -12,12 +12,11 @@
...
+    def test_it(self, canvas_api_client, authenticated_client):
```
seanh, line 19: There's no point in having a test called `TestGetToken.test_get_token()`, our style is to just call it `TestGetToken.test_it()` in these cases

jon-betts (approved):
Looks good and still passes. I think the two nicest features of this are:

* Pulling relevant fixtures nearer to the methods that use them
* Making names a little more concise

This code has been on a fabulous voyage over the last year. If only I could say the same.

Issue #2320, opened 2020-12-21
Error dialog when fetching a grade fails is different from other error dialogs


seanh:
## Problem

The error dialog that's shown when fetching a grade fails is different from other dialogs in the app:

![Screenshot from 2020-12-21 11-15-07](https://user-images.githubusercontent.com/22498/102771458-d6e16200-437d-11eb-9e9c-4eb9e2280646.png)

![Screenshot from 2020-12-21 11-14-51](https://user-images.githubusercontent.com/22498/102771459-d779f880-437d-11eb-9368-8fd17e1765e7.png)

## Reproducing

Apply the diff below to cause fetching grades to fail. Then log in to hypothesisuniversity.moodlecloud.com as a teacher, launch an assignment, and go to a student.

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 5bf7cd77..bccd62d6 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -134,6 +134,7 @@ class LTIOutcomesClient:
             status = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
                 "imsx_codeMajor"
             ]
+            foo = data["foo"]
 
         except KeyError as err:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
```

## Solution

This dialog should re-use the HTML and CSS used for other dialogs elsewhere in the app, including:

- [x] It shouldn't have the **: (**
- [x] The font should be smaller
- [x] The font color should be darker
- [x] The dialog should be less wide
- [x] The dialog should be less tall

There are probably other differences too. If it can re-use the HTML and CSS that should catch them all.

LMS007:
Repurposing this ticket to change the fetch/submit error dialogs to behave more like LaunchErrorDialogs with retry buttons that re-send the appropriate requests and can not be dismissed on click (and possible don't have a "X") -- TBD.

Also, the CSS/Styling of all dialogs should be consistent across the app. Some of those changes were addressed here that still need to be merged https://github.com/hypothesis/lms/pull/2329/files

Issue #2295, opened 2020-12-11
Keyboard focus ring is activated when not using the keyboard


seanh:
The keyboard focus ring (black border) is sometimes shown around buttons even though I haven't used the keyboard:

![Screenshot from 2020-12-11 13-38-33](https://user-images.githubusercontent.com/22498/101910035-528d2300-3bb6-11eb-90dd-11ca56b74ca7.png)

1. Delete all the access tokens from your DB, to force re-authorization:

   ```terminal
   tox -qe dockercompose -- exec postgres psql -U postgres -c 'DELETE FROM oauth2_token;'
   ```

2. Launch the [localhost (make devdata) Canvas Files Assignment](https://hypothesis.instructure.com/courses/125/assignments/875)

3. You should see the **Authorize Hypothesis** dialog and, even though you haven't used the keyboard, the focus ring black border will be shown around the <kbd>Authorize</kbd> button. Note that the focus ring is difficult to notice because it's black and the button is also black (that's a separate issue), but if you hover the mouse pointer over the <kbd>Authorize</kbd> button the button will turn red and the black focus ring will be revealed.

I _think_ there are also other situations where the focus ring is being shown even though the keyboard isn't being used. The one particular situation described above is just the one way I've found to reproduce it.

robertknight:
It looks like the `Button` component's styles don't use the `outline-on-keyboard-focus` mixin as they do in the client. What we probably need to do here is sync up the LMS frontend to use the current version of the button styles from the client.

seanh:
Another example of this: https://github.com/hypothesis/lms/pull/2329#issuecomment-770751830

I think I've seen other places in the UI where the focus ring is around a dialog as soon as the dialog appears, as well

LMS007:
Closing this out as all dialog buttons have been converted to the new shared button component.

Focus will be visible if the browser determines if the keyboard was used to trigger the action that opened the dialog, otherwise no focus will be present on buttons.

PR #2329, opened 2020-12-22
Make error dialogs more consistent

LMS007:
Set font size for Dialog content
Do not override font color in LMSGrader
Add default font color to body (grey-7)
Use standard font sizes for LMSGrader
Fixup Dialog title text

-----------

fixes https://github.com/hypothesis/lms/issues/2320

dialogs now look a lot more similar. The main issue here was that the LMSGrader was overriding some of the default styling that dialogs were taking on when spawned from the app root (BasicLtiLauchApp)

A few things are done here. I moved the default color to the body, so now everything is dark grey. Previously, all the error dialogs seen from BasicLtiLauchApp were colored black which is not ideal. Now everything is off-black (better). Then I removed the color override from LMSGrader which was causing the major differences we were seeing. Additionally I forced the Dialog to set its own font size and then finally I removed some of the hard-coded px font sizes in the Grader. In some cases fonts few by 1px and in other they shrank, but this is  pretty insignificant. 

Screen Shot 2020-12-22 at 12 16 01 PM

Screen Shot 2020-12-22 at 12 19 12 PM

### DIFF
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -219,7 +219,9 @@ export default function SubmitGradeForm({ student }) {
       </button>
       {!!submitGradeError && (
         <ErrorDialog
-          title="Submit Grade Error"
+          title={`There was a problem submitting ${
+            student ? student.displayName : ' the student'
+          }'s grade.`}
           error={{ message: submitGradeError }}
           onCancel={() => {
             setSubmitGradeError('');
@@ -229,7 +231,9 @@ export default function SubmitGradeForm({ student }) {
       )}
       {!!fetchGradeError && (
         <ErrorDialog
-          title="Fetch Grade Error"
+          title={`There was a problem fetching ${
+            student ? student.displayName : ' the student'
+          }'s grade.`}
           error={{ message: submitGradeError }}
           onCancel={() => {
             setFetchGradeError('');
--- a/lms/static/styles/components/_ErrorDisplay.scss
+++ b/lms/static/styles/components/_ErrorDisplay.scss
@@ -11,7 +11,6 @@
 
   &__links {
     white-space: normal;
-    line-height: 1.25em;
   }
 
   &__details {
--- a/lms/static/styles/components/_LMSGrader.scss
+++ b/lms/static/styles/components/_LMSGrader.scss
@@ -3,11 +3,10 @@
 .LMSGrader {
   display: flex;
   flex-direction: column;
-  font-size: 16px;
+  font-size: var.$input-font-size;
   height: 100%;
   width: 100%;
   position: fixed;
-  color: var.$grey-6;
 }
 
 .LMSGrader__grading-components {
@@ -39,12 +38,12 @@
 }
 
 .LMSGrader__assignment {
-  font-size: 18px;
+  font-size: var.$title-font-size;
   margin: 0 0 10px 0;
 }
 
 .LMSGrader__name {
-  font-size: 16px;
+  font-size: var.$subtitle-font-size;
   margin: 0;
 }
 
--- a/lms/static/scripts/frontend_apps/components/ErrorDialog.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDialog.js
@@ -21,7 +21,7 @@ export default function ErrorDialog({ onCancel, title, error, cancelLabel }) {
   return (
     <Dialog
       role="alertdialog"
-      title="Something went wrong :("
+      title="Something went wrong"
       onCancel={onCancel}
       cancelLabel={cancelLabel}
     >
--- a/lms/static/styles/components/_Dialog.scss
+++ b/lms/static/styles/components/_Dialog.scss
@@ -1,3 +1,4 @@
+@use "../mixins";
 @use "../mixins/focus";
 @use "../variables" as var;
 
@@ -21,6 +22,7 @@
 }
 
 .Dialog__content {
+  @include mixins.font-normal;
   display: flex;
   flex-direction: column;
   max-height: 90vh;
@@ -48,7 +50,7 @@
   // enough to see easily and aligned with the right edge of the dialog.
   // Add negative margins so that the button does not force the dialog to
   // grow in height.
-  font-size: 20px;
+  font-size: var.$title-font-size;
   padding: 5px;
   margin: -10px 0px;
 
--- a/lms/static/styles/frontend_apps.scss
+++ b/lms/static/styles/frontend_apps.scss
@@ -32,4 +32,5 @@ body {
   font-family: var.$sans-font-family;
   font-size: var.$normal-font-size;
   line-height: var.$normal-line-height;
+  color: var.$grey-7;
 }


lms/static/styles/components/_ErrorDisplay.scss
```diff
@@ -11,7 +11,6 @@
 
   &__links {
     white-space: normal;
-    line-height: 1.25em;
```
LMS007, line 14: Since the font is smaller, this no longer word wraps, but even if it did, now the default font-size / line height is better matched so we don't need to override this.

lms/static/styles/components/_LMSGrader.scss
```diff
@@ -3,11 +3,10 @@
 .LMSGrader {
   display: flex;
   flex-direction: column;
-  font-size: 16px;
+  font-size: var.$input-font-size;
   height: 100%;
   width: 100%;
   position: fixed;
-  color: var.$grey-6;
```
LMS007, line 10: let the color come from body

seanh (changes requested):
Can we replace <q>Fetch Grade Error</q> with something better? At least an actual sentence?

For example the other error dialog shown in the screenshot says <q>There was a problem fetching this Hypothesis assignment.</q> Similarly, maybe this dialog could say <q>There was a problem fetching Paul Smith's grade from Blackboard.</q> Replacing _Paul Smith_ with the student's name and _Blackboard_ with the name of the LMS (Blackboard, Moodle, Brightspace, ...).

Getting the correct LMS name in there might be more work so maybe <q>There was a problem fetching Paul Smith's grade.</q> would be good enough for now.

LMS007:
That's reasonable Sean. I think some of my confusion came from this prop being called "title" where I just assumed it was intended to be a short header on the error dialog and that the message itself would be returned from the server. But in this case there is no message.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -219,7 +219,7 @@ export default function SubmitGradeForm({ student }) {
       </button>
       {!!submitGradeError && (
         <ErrorDialog
-          title="Submit Grade Error"
+          title={`There was a problem submitting ${student ? student.displayName : ' the student'}'s grade.`}
```
LMS007, line 222: I could add the `LISOutcomeServiceUrl` but I'm not sure there is a lot of value as I would assume a customer would know what LMS they were using. A screen could show that too.

lms/static/styles/components/_ErrorDisplay.scss
lyzadanger: BTW, auditing for `line-height` and `font-size/line-height` combinations is on my longish mental TODO list for (S)CSS tech debt and pattern-clarity cleanup.

lms/static/styles/components/_LMSGrader.scss
```diff
@@ -3,11 +3,10 @@
 .LMSGrader {
   display: flex;
   flex-direction: column;
-  font-size: 16px;
+  font-size: var.$input-font-size;
```
lyzadanger, line 6: Good. Every time we eliminate a magic number in our SCSS source, a fairy gets their wings.

lyzadanger (commented):
From a strictly code-side perspective, I don't see anything alarming with these changes; in the SCSS, it's good to see the removal of an unneeded override and the use of variables for font sizing.

_I have not looked at these changes in context_ as there are other team members who have quicker access to do so right now.

I'm also happy to work with the team to look at LMS design-application consistency overall as we move toward having shareable patterns and components. It may be a little premature for me to wade in too much. Happy to field any specific questions or concerns.

LMS007:
If there are no objections, I'd like to merge this

seanh:
@LMS007 Can you post instructions for how to test this locally?

LMS007:
> @LMS007 Can you post instructions for how to test this locally?

No prob Sean. Force an error, then you can click next student in this assignment to see the grade fetch error.

error_test

patch


```
From 2ae97bcebe5782d3f580d35ccf177b31ebafeb97 Mon Sep 17 00:00:00 2001
From: Kyle Keating <kkeating@hypothes.is>
Date: Fri, 29 Jan 2021 11:57:48 -0800
Subject: [PATCH] fake error

---
 lms/services/lti_outcomes.py                                    | 1 +
 .../scripts/frontend_apps/components/BasicLtiLaunchApp.js       | 2 +-
 2 files changed, 2 insertions(+), 1 deletion(-)

index 2e73bd67..9a06f779 100644
@@ -133,6 +133,7 @@ class LTIOutcomesClient:
         """Return the POX body element, checking for errors."""
 
         try:
             body = data["imsx_POXEnvelopeResponse"]["imsx_POXBody"]
index 5fd95d39..0ec2583d 100644
@@ -64,7 +64,7 @@ export default function BasicLtiLaunchApp({ clientRpc }) {
   // Indicates what the application was doing when the error indicated by
   // `error` occurred.
   const [errorState, setErrorState] = useState(
-    /** @type {ErrorState|null} */ (null)
+    /** @type {ErrorState|null} */ ('error-fetching')
   );
 
   // The most recent error that occurred when launching the assignment.
-- 
2.24.3 (Apple Git-128)

```

seanh:
@LMS007 Do you know why the keyboard focus ring is showing on this dialog as soon as it appears, even though I'm not using the keyboard?

Screenshot from 2021-02-01 10-26-13

It looks similar to a problem with some other LMS controls, and Rob suggested a solution here: https://github.com/hypothesis/lms/issues/2295

seanh:
When I try to submit a grade and get an error there's a `.` and `:` together: `There was a problem submitting Sean Student's grade.: Malformed LTI outcome response.` Should be the `:` only

Screenshot from 2021-02-01 10-31-00

seanh:
When I submit a grade and get an error, the submitted grade is still present in the UI even though it hasn't actually been saved. It should probably clear the grade from the input field if it hasn't been saved. I think this could be moved to a separate issue

Peek 2021-02-01 10-33

seanh:
I noticed that clicking anywhere outside of this dialog causes the dialog to disappear. Other LMS dialogs don't do that and I think this one probably shouldn't either. It could lead to the user accidentally closing the dialog. They can just click the <kbd>Close</kbd> or <kbd>x</kbd> button if they want to close the dialog

seanh:
I think the dialog should have a <kbd>Try again</kbd> button that re-tries fetching the grade and re-shows the dialog if it fails again

seanh (changes requested):
I gave this a quick test and found a few issues, but I think most of them can be moved into separate issues. I've noted each issue as a separate comment above so they can easily be converted into issues

Should be fixed in this PR:

- [ ] The `.:` in the text when getting an error when submitting a grade
- [ ] The dialog should not disappear when you click outside of the dialog

Could be moved to separate issues:

- [ ] The keyboard focus ring appearing as soon as the dialog appears (I think this affects multiple buttons and dialogs throughout the LMS app)
- [ ] It should clear the grade from the input after getting an error when submitting a grade
- [ ] The dialog should have a try again button

LMS007:
> When I submit a grade and get an error, the submitted grade is still present in the UI even though it hasn't actually been saved. It should probably clear the grade from the input field if it hasn't been saved. I think this could be moved to a separate issue
> 
> Peek 2021-02-01 10-33

We can address that separately, i'd like to just land some specific issues related to the ticket. In terms of clearing a field on client error.. this i something I generally don't like. If there is an error, it's presented and its assumed the user fixes the error. As a real-world example, Gmail does not clear the email when there is an invalid email address, it just informs you its wrong and does not send anything. There may be validity and convince to reverting the grade since there could be a previous one, but we do give affordance when its saved by making the background green for a second.  It should be obvious when that action succeeds.  Lets move this to a separate ticket/PR.

LMS007:
> I gave this a quick test and found a few issues, but I think most of them can be moved into separate issues. I've noted each issue as a separate comment above so they can easily be converted into issues
> 
> Should be fixed in this PR:
> 
> * [ ]  The `.:` in the text when getting an error when submitting a grade
> * [ ]  The dialog should not disappear when you click outside of the dialog
> 
> Could be moved to separate issues:
> 
> * [ ]  The keyboard focus ring appearing as soon as the dialog appears (I think this affects multiple buttons and dialogs throughout the LMS app)
> * [ ]  It should clear the grade from the input after getting an error when submitting a grade
> * [ ]  The dialog should have a try again button

_"The dialog should not disappear when you click outside of the dialog"_ 

That behavior is baked into the Dialog components that have an `onCancel` callback, or rather the ability to dismiss the dialog. A change here, would be carried over to other dialogs that can be dismissed and so we would need to probably just remove`useElementShouldClose` from LMS entirely. Its a bigger UX change then I would like to make here, which I hoped would just be some clean-up of CSS. I can file a ticket to change the Dialog behavior and remove useElementShouldClose from LMS.

LMS007:
@seanh 

I think I understand more what you're asking with the Dialog.

1. Don't use an `onCancel` callback for `ErrorDialog`, (_that allows it to close on click-outside_) but possibly still allow it to have an "X" to dismiss it. We can make that change, but that will be a new adjustment to the component. 

2. Use a Try again button (similar to `LaunchErrorDialog`) and make that retry the fetch (or post) request to get/set grades.



Going to be more major changes for that, but makes sense.  I'll close out this PR for now.

PR #2332, opened 2020-12-24
Implement CanvasFile polymorphic model

seanh:
TODO:

* Either make the `type` column an enum or get rid of it and use the
  `file_id_override` column (whether it's NULL or not) as the
  polymorphic_on switch

* Make file_id_override a foreign key to file_id (or to id, since
  file_id isn't a primary key or unique) with ondelete=cascade
  (this is known as an "adjacency list" pattern

Known issues:

* There's nothing at the model level to stop file_id_override from
  pointing to another CanvasFileOverride rather than to a CanvasFile or
  to prevent loops (circular references)

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -55,15 +55,6 @@ class BaseClass:
             if key in data:
                 setattr(self, key, data[key])
 
-    def __repr__(self):
-        return "{class_}({kwargs})".format(
-            class_=self.__class__.__name__,
-            kwargs=", ".join(
-                f"{kwarg}={repr(getattr(self, kwarg))}"
-                for kwarg in self.__table__.columns.keys()  # pylint:disable=no-member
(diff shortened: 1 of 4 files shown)


lms/db/__init__.py
```diff
@@ -55,15 +55,6 @@ def update_from_dict(self, data, skip_keys=None):
...
                 setattr(self, key, data[key])
 
-    def __repr__(self):
-        return "{class_}({kwargs})".format(
-            class_=self.__class__.__name__,
-            kwargs=", ".join(
-                f"{kwarg}={repr(getattr(self, kwarg))}"
-                for kwarg in self.__table__.columns.keys()  # pylint:disable=no-member
-            ),
-        )
```
seanh, line 65: Had to delete this because it breaks SQLAlchemy polymorphism

PR #2374, opened 2021-01-26
Add MVP for launching VitalSource book viewer from LMS app

robertknight:
This PR adds a "VitalSource Book" option to the file picker app which creates an assignment/activity using a hard-coded VitalSource book and chapter. When that assignment is launched, the VitalSource book viewer is loaded inside an iframe showing the specified book and chapter.

The purpose of this PR is to enable the VS team to see Hypothesis running inside VS for themselves so they can resolve various integration issues.

In future our LMS app will query the VitalSource API for a list of available books and chapters within a selected book, for use as the text for an assignment.

The initial VS integration adds two new configuration variables: VITALSOURCE_LAUNCH_KEY and VITALSOURCE_LAUNCH_SECRET. These are credentials used to sign the OAuth 1.0-authenticated LTI launch request used to load the VitalSource book viewer. In future there will also be an API key for authenticating access to the VS API.

Changes in detail:

 - Add a backend service in `lms/services/vitalsource.py`  to handle VitalSource LTI launches
 - Add a "Select book from VitalSource" button to the file picker app which submits the file picker form with an LTI launch URL for a VS book.
 - Add logic to the LTI launch views to handle launching an assignment configured to show a VS book
 - Add logic to the frontend in `BasicLtiLaunchApp` and `VitalSourceBookViewer` to load the VS book viewer inside an iframe.
 - Whitelist the `VITALSOURCE_LAUNCH_KEY` and `VITALSOURCE_LAUNCH_SECRET` keys in tox.ini. These keys are needed by the VitalSource launch request
 
For Hypothesis devs, the `VITALSOURCE_LAUNCH_KEY` and `VITALSOURCE_LAUNCH_SECRET` values are in our usual password store.

Note that in development, the Hypothesis client will only partially load inside the book viewer. You'll see the Annotate/Highlight popup appear when selecting text in the book, but it is partially obscured by the viewer's native annotation tools. The sidebar is also not visible - that's because the sidebar is currently hardcoded by VS to always talk to `https://lms.hypothes.is`, instead of the QA or dev LMS app.

---

**TODO:**

- [x] Discuss the changes here with a backend dev
- [ ] Write tests for the frontend and backend
- [ ] Hide the VitalSource book viewer button in the file picker app behind a feature flag


robertknight:
For ease of review, I'm going to split this into two separate pull requests:

- One that adds the "Select book from VitalSource" option in the file picker
- One that handles LTI launches configured to load a VitalSource book

PR #2377, opened 2021-01-29
Add @hypothesis/frontend-shared

LMS007:
Replace mixins/focus.scss with FES version


robertknight (approved):
LGTM. I checked that the `focus.scss` file in the package is exactly the same as the one that was removed from this repository.

PR #2384, opened 2021-02-02
Import `useElementShouldClose` hook from frontend-shared package

robertknight:
Replace the version of the `useElementShouldClose` hook in this repository with one from the @hypothesis/frontend-shared package.

The version in the package is the same except that it only supports being used with element refs and not Preact component refs.

This hook is currently intended for non-modal popups but in the LMS we also use it for dialogs that should probably be either modal or conditionally modal if a form within the dialog has unsaved changes. This is something we can revisit later.

This depends on https://github.com/hypothesis/client/pull/2951 and https://github.com/hypothesis/client/pull/2955.


lyzadanger (approved):
LGTM

PR #2387, opened 2021-02-04
Remove unused `trustMarkup` function

robertknight:
This was used at one point for SVG icons, but that was replaced with an
alternate method of ensuring that only trusted SVG markup is used.


lyzadanger (approved):
LGTM

PR #2390, opened 2021-02-04
Support launching assignments using VitalSource books

robertknight:
This commit adds support for loading the VitalSource book viewer in an LTI launch if the assignment URL contains `vitalsource_enabled=true` in the query params, similar to how Canvas files assignments are configured.

The VitalSource book viewer is loaded via a secondary LTI 1.1 launch. When an assignment is configured to use VitalSource, the backend generates form parameters for an LTI launch and signs them using a key and secret specified by two new env vars: `VITALSOURCE_LAUNCH_KEY` and `VITALSOURCE_LAUNCH_SECRET`. These form parameters are rendered into the configuration for the LMS frontend.

When this configuration exists, the LMS frontend creates an empty iframe, renders a form with these parameters into the iframe and auto-submits it. As a result of the submission, the VitalSource book viewer app will load in the iframe at the specified book and chapter with Hypothesis embedded.

There is currently an issue that the VS app always tries to load the production Hypothesis client, which won't work properly with the development LMS app.

Changes in this PR in detail:

 - Whitelist the `VITALSOURCE_LAUNCH_KEY` and `VITALSOURCE_LAUNCH_SECRET` env vars in tox.ini.
 - Add a `VitalSourceService` service in `lms/services/vitalsource.py` to generate the form parameters for the VS LTI launch
 - Add a new LTI launch type, guarded by a `vitalsource_book` predicate, which responds to launches where the URL contains `vitalsource_book=true`. It uses the `VitalSourceService` service to generate the necessary configuration for the LMS frontend to load the VS viewer.
 - Add logic to the frontend in `BasicLtiLaunchApp` to check for this config and respond by rendering a `VitalSourceBookViewer` component instead of the normal Via iframe. The `VitalSourceBookViewer` component in turn renders an iframe containing a form with the LTI launch parameters and auto-submits the form. This results in the iframe then displaying the VitalSource book viewer
   
----

**Testing:**

1. Configure the `VITALSOURCE_LAUNCH_KEY` and `VITALSOURCE_LAUNCH_SECRET` env vars in your local lms dev environment. H staff can find the credentials in the "VitalSource LTI and API keys" entry in 1Password.
2. Create an assignment using the LMS app in Canvas and manually set the assignment's URL to http://localhost:8001/lti_launches?vitalsource_book=true&book_id=BOOKSHELF-TUTORIAL&cfi=%2F6%2F8%5B%3Bvnd.vst.idref%3Dvst-70a6f9d3-0932-45ba-a583-6060eab3e536%5D. Once https://github.com/hypothesis/lms/pull/2388 is merged, it will be possible to use the file picker UI to configure a VitalSource assignment
3. Launch the assignment. It should load the VitalSource book viewer app in an iframe instead of Via as for other assignments. I suggest clicking the option to open VS in a new window, since the book viewer is difficult to use in the iframe Canvas provides.
4. You may be assigned to enter VitalSource account details when the viewer first appears. Either create an account for yourself or ping me and I can give you credentials to use. You may see a "No, skip this step" option at the login screen, in which case just click that.
5. You may see a bunch of popups and other cruft in the viewer that need to be dismissed before you can get to the text. We are going to work with the VS team to remove this

After this, you should see the the viewer load a tutorial book at the following page:

vitalsource-book

You won't see the Hypothesis sidebar yet, but if you select some text you will see the Annotate/Highlight toolbar appear, although depending on where and in which direction you create the selection, it may be obscured by VitalSource's own annotation tools.


lms/services/vitalsource.py
```diff
@@ -0,0 +1,60 @@
...
+
+        launch_url = f"https://bc.vitalsource.com/books/{book_id}"
+        book_location = "/cfi" + cfi
+
+        launch_params = _launch_params(
+            user_id=lti_user.user_id,
+            roles=lti_user.roles,
+            # Set a dummy `context_id`. This needs to be replaced with the real
+            # course in future.
+            context_id="testcourse",
```
robertknight, line 25: According to the linked documentation above, the course information is optional but useful for analytics purposes:

> Context information will also allow VitalSource to report accurately and be able to distinguish between courses. Without the context information, VitalSource will see all launches as the same and without context, and therefore cannot provide any analytics about course activity.

lms/services/vitalsource.py
```diff
@@ -0,0 +1,60 @@
...
+        self._oauth_secret = request.registry.settings["vitalsource_launch_secret"]
+
+    def get_launch_params(self, book_id, cfi, lti_user):
+        """
+        Launch the VitalSource book viewer using an LTI launch.
+
+        See https://developer.vitalsource.com/hc/en-us/articles/215612237-POST-LTI-Create-a-Bookshelf-Launch
+        """
+
+        launch_url = f"https://bc.vitalsource.com/books/{book_id}"
```
robertknight, line 17: The book ID can be included either in the launch URL or in launch params via the `custom_vbid` field. I went with using the launch URL because that seemed the most obvious thing to do. Putting in the params in future could have the potential advantage of allowing the launch URL to be removed from the Python code and just hardcoded in the LMS frontend.

lms/services/vitalsource.py
```diff
@@ -0,0 +1,60 @@
...
+        "lti_version": "LTI-1p0",
+        "lti_message_type": "basic-lti-launch-request",
+        # User data. These should be proxied from the LTI launch.
+        "user_id": user_id,
+        "roles": roles,
+        "context_id": context_id,
+        # Book presentation and location options
+        "launch_presentation_document_target": "window",
+        "custom_book_location": location,
+    }
```
robertknight, line 58: There are also some other fields which are not required but the linked article above recommends setting, such as `lis_person_name_full`. I plan to add this in future.

lms/static/scripts/frontend_apps/components/VitalSourceBookViewer.js
```diff
@@ -0,0 +1,61 @@
...
+    launchForm.action = launchUrl;
+
+    Object.entries(launchParams).forEach(([key, value]) => {
+      const field = iframeDoc.createElement('input');
+      field.type = 'hidden';
+      field.name = key;
+      field.value = value;
+      launchForm.appendChild(field);
+    });
+    iframeDoc.body.appendChild(launchForm);
```
robertknight, line 45: I opted to build the form elements manually here because it is so simple, although I could have used Preact to render it.

lms/views/predicates/_lti_launch.py
```diff
@@ -178,6 +178,13 @@ def __call__(self, context, request):
         return ("canvas_file" in request.params) == self.value
 
 
+class VitalSourceBook(Base):
+    name = "vitalsource_book"
+
+    def __call__(self, context, request):
+        return ("vitalsource_book" in request.params) == self.value
```
robertknight, line 185: This check is consistent with how the `canvas_file` predicate above works, but note that I'm not checking for the existence of the required `book_id` and `cfi` parameters here. Similarly the `canvas_file` predicate above doesn't check for the required `file_id` parameter. Where is the best place to put such validation?

tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -130,6 +130,25 @@ def configure_module_item_caller(context, pyramid_request):
...
 
 
+def vitalsource_lti_launch_caller(context, pyramid_request):
+    """
+    Call BasicLTILaunchViews.vitalsource_lti_launch().
+
+    Set up the appropriate conditions and then call
+    BasicLTILaunchViews.vitalsource_lti_launch(), and return whatever
+    BasicLTILaunchViews.vitalsource_lti_launch() returns.
+    """
```
robertknight, line 140: This comment follows the pattern of the similar functions above, although it feels highly redundant as it just restates the code.

lms/resources/_js_config.py
```diff
@@ -75,6 +75,19 @@ def add_document_url(self, document_url):
...
+        launch_url, launch_params = vitalsource_svc.get_launch_params(
+            book_id, cfi, self._request.lti_user
+        )
+        self._config["vitalSource"] = {
+            "launchUrl": launch_url,
+            "launchParams": launch_params,
+        }
+        self._add_canvas_speedgrader_settings(
+            vitalsource_book_id=book_id, vitalsource_cfi=cfi
+        )
```
robertknight, line 89: I'm filling out these SpeedGrader params here in a way that is consistent with how Canvas files launches are handled. However I haven't actually tested the VS viewer with SpeedGrader yet. We could just leave that out of scope for this PR.

lms/static/scripts/frontend_apps/components/VitalSourceBookViewer.js
```diff
@@ -0,0 +1,61 @@
...
+
+    willSubmitLaunchForm?.(launchForm);
```
LMS007, line 47: Maybe a comment that this is used to stub out a "submit" function that can be used to verify that it was called. If I have that correct?

lms/static/scripts/frontend_apps/config.js
```diff
@@ -41,6 +41,14 @@ import { createContext } from 'preact';
  * @prop {Object} submissionParams
  */
 
+/**
+ * Data needed to display the VitalSource book viewer.
+ *
+ * @typedef VitalsourceConfig
+ * @prop {string} launchUrl
+ * @prop {Object.<string,string>} launchParams
```
LMS007, line 49: Or link to docs here? Now wondering if we should have a separate typedef for this...

See https://developer.vitalsource.com/hc/en-us/articles/206156238-General-LTI-Usage

LMS007 (approved):
👍 
JS parts of the PR look solid to me, just had some minor suggestions for comments.

Also, do we intend to use this with non-canvas LMS's with our own grader? I did not test that out yet.

lms/static/scripts/frontend_apps/components/VitalSourceBookViewer.js
```diff
@@ -0,0 +1,61 @@
...
+
+    launchForm.submit();
```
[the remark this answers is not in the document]
robertknight: That's correct. This is just a regular HTML form post which will load the HTML response returned by the server. In this case that means a redirect to the VitalSource app viewer. I'll add a comment to clarify.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -358,6 +358,29 @@ describe('BasicLtiLaunchApp', () => {
...
+          book_id: 'somebook',
+          location: 'chapter-2',
+        },
+      };
+
+      const wrapper = renderLtiLaunchApp();
+
+      const vsViewer = wrapper.find('VitalSourceBookViewer');
+      assert.isTrue(vsViewer.exists());
+      assert.isFalse(wrapper.exists('iframe'));
```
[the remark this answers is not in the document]
robertknight: That's correct - I'm just checking that we rendered only the VS viewer and not the iframe.

lms/static/scripts/frontend_apps/components/VitalSourceBookViewer.js
robertknight: Kind-of. This is a "hook" that is used to observe the form just before it is submitted.

lms/static/scripts/frontend_apps/config.js
robertknight: The intent here is that the frontend doesn't need to know about what the form parameters mean, it just renders a form given whatever parameters the backend supplies and submits it. The supported parameters are documented in the backend code that generates the them.

robertknight:
I have separated the frontend and backend changes out into separate PRs so that I can land the frontend changes. The frontend changes have been merged and the backend changes are in https://github.com/hypothesis/lms/pull/2416.

PR #2391, opened 2021-02-04
Log LTIOutcomesAPIError exceptions from _get_body

LMS007:
Log LTIOutcomesAPIError exceptions so they are picked up by sentry for easier debugging.

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -1,3 +1,4 @@
+import logging
 from xml.parsers.expat import ExpatError
 
 import requests
@@ -5,6 +6,8 @@ import xmltodict
 from requests import RequestException
 
 from lms.services.exceptions import LTIOutcomesAPIError
+
+log = logging.getLogger(__name__)
 
 __all__ = ["LTIOutcomesClient"]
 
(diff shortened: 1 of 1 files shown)


lms/services/lti_outcomes.py
```diff
@@ -126,6 +129,7 @@ def _send_request(self, request_body):
             return self._get_body(data)
         except LTIOutcomesAPIError as err:
             err.response = response
+            log.exception(err)
```
robertknight, line 132: Does passing an `LTIOutcomesAPIError` option as the `msg` argument to `log.exception` work? The [Python docs](https://docs.python.org/3/library/logging.html#logging.Logger.exception) and [Sentry docs](https://docs.sentry.io/platforms/python/guides/logging/) to me suggests that the argument should be a string and then exception details will be captured automatically.

robertknight (changes requested):
Are you sure about the usage of `log.exception` here @LMS007? It doesn't match the examples in the Python or Sentry docs or the one other example of `log.exception` that I found in the code.

lms/services/lti_outcomes.py
LMS007: Awe so if I understand that correctly, just log a message with .exception and the error will follow too.

robertknight: Yes, exactly.

robertknight (approved):
LGTM, except that you're going to need to run `make format`. Strings are double-quoted in our Python code.

PR #2393, opened 2021-02-06
Dialog visual fixes

### LINKED ISSUE #2320: Error dialog when fetching a grade fails is different from other error dialogs

LMS007:
### Commits

Fix incorrect fetch grade message …

The fetch grade error was incorrectly showing the wrong message

------------

- Remove ":(" from ErrorDialog's title text for submit and fetch errors
- Set font size for Dialog content
- Remove focus outline when Dialog's content pane is in focus
- Do not override font color in LMSGrader
- ~Add default font color to body ($grey-7)~
- Use standard font sizes for LMSGrader

-----------

_Updates_

**Refactored this PR to only include CSS/text issues to dialogs. Pulled out the retry feature and will include that in another PR.
** removed body color, (will fix in follow up PR)

This PR builds off the orignal here
https://github.com/hypothesis/lms/pull/2329

fixes #2320

**Before**
Screen Shot 2021-02-10 at 2 39 24 PM

**After**
Screen Shot 2021-02-10 at 2 39 59 PM

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ErrorDialog.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDialog.js
@@ -21,7 +21,7 @@ export default function ErrorDialog({ onCancel, title, error, cancelLabel }) {
   return (
     <Dialog
       role="alertdialog"
-      title="Something went wrong :("
+      title="Something went wrong"
       onCancel={onCancel}
       cancelLabel={cancelLabel}
     >
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -230,7 +230,7 @@ export default function SubmitGradeForm({ student }) {
       {!!fetchGradeError && (
         <ErrorDialog
           title="Fetch Grade Error"
-          error={{ message: submitGradeError }}
+          error={{ message: fetchGradeError }}
           onCancel={() => {
             setFetchGradeError('');
           }}
--- a/lms/static/styles/components/_Dialog.scss
+++ b/lms/static/styles/components/_Dialog.scss
@@ -1,5 +1,6 @@
 @use "@hypothesis/frontend-shared/styles/mixins/focus";
 
+@use "../mixins";
 @use "../variables" as var;
 
 .Dialog__container {
@@ -22,6 +23,8 @@
 }
 
 .Dialog__content {
+  @include mixins.font-normal;
+  @include focus.outline-on-keyboard-focus;
   display: flex;
   flex-direction: column;
   max-height: 90vh;
@@ -49,7 +52,7 @@
   // enough to see easily and aligned with the right edge of the dialog.
   // Add negative margins so that the button does not force the dialog to
   // grow in height.
-  font-size: 20px;
+  font-size: var.$title-font-size;
   padding: 5px;
   margin: -10px 0px;
 
--- a/lms/static/styles/components/_ErrorDisplay.scss
+++ b/lms/static/styles/components/_ErrorDisplay.scss
@@ -12,7 +12,6 @@
 
   &__links {
     white-space: normal;
-    line-height: 1.25em;
   }
 
   &__details {
--- a/lms/static/styles/components/_LMSGrader.scss
+++ b/lms/static/styles/components/_LMSGrader.scss
@@ -3,7 +3,7 @@
 .LMSGrader {
   display: flex;
   flex-direction: column;
-  font-size: 16px;
+  font-size: var.$input-font-size;
   height: 100%;
   width: 100%;
   position: fixed;
@@ -39,12 +39,12 @@
 }
 
 .LMSGrader__assignment {
-  font-size: 18px;
+  font-size: var.$title-font-size;
   margin: 0 0 10px 0;
 }
 
 .LMSGrader__name {
-  font-size: 16px;
+  font-size: var.$subtitle-font-size;
   margin: 0;
 }
 


lms/static/scripts/frontend_apps/components/ErrorDialog.js
```diff
@@ -1,14 +1,21 @@
 import { createElement } from 'preact';
+import { useRef } from 'preact/hooks';
 
+import Button from './Button';
 import ErrorDisplay from './ErrorDisplay';
 import Dialog from './Dialog';
 
 /**
  * @typedef ErrorDialogProps
+ * @prop {boolean} [busy] -
```
LMS007, line 10: This is actually not used yet, but could be if we replace BaseDialog.. I sorta brought it in with the other retry onRetry prop

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -225,16 +226,21 @@ export default function SubmitGradeForm({ student }) {
...
+            setSubmitGradeError('');
+            onSubmitGrade(e);
+          }}
         />
       )}
       {!!fetchGradeError && (
         <ErrorDialog
           title="Fetch Grade Error"
-          error={{ message: submitGradeError }}
+          error={{ message: fetchGradeError }}
```
LMS007, line 238: This was actually a bug that I caused! It's why "Malformed LTI..." message was not reporting on fetch errors and we were not able to distinguish between any type of fetch error.

lms/static/styles/components/_Dialog.scss
```diff
@@ -22,6 +23,8 @@
 }
 
 .Dialog__content {
+  @include mixins.font-normal;
+  @include focus.outline-on-keyboard-focus;
```
LMS007, line 27: This prevents the Dialog box from having that big blue (or grey) outline ring.

lms/static/styles/frontend_apps.scss
```diff
@@ -32,4 +32,5 @@ body {
   font-family: var.$sans-font-family;
   font-size: var.$normal-font-size;
   line-height: var.$normal-line-height;
+  color: var.$grey-7;
```
LMS007, line 35: There was no default color...now there is!

LMS007:
Quick note about focus ring and the "Try again" or "Authorize" button on the dialog. It won't have focus if the last action you performed was with a mouse. e.g. If you click to change students to fetch a new grade, and there is an error, then the button won't show focus (it will indeed be focused, but visibly hidden). However if you did the same thing with the keyboard, then it will show focus. 

The one exception is when you first load up the app. and you get a launch error such "Authorize". I believe in this case it will default to showing the focus ring. e.g. keyboard focus. I think that may be different from browser to browser, but more testing needs to be done.

robertknight:
I've just started looking at this PR and it was initially confusing because there is quite a gap in scope between the original issue in https://github.com/hypothesis/lms/issues/2320 and what is included in this PR.

Reading the PR description it looks to me like there are three orthogonal sets of changes here:

1. Regression fixes
   - Fixed a bug where the fetch grade error was incorrectly showing the wrong message
2. Styling / text updates
   - Remove ":(" from ErrorDialog's title text for submit and fetch errors
   - Set font size for Dialog content
   - Remove focus outline when Dialog's content pane is in focus
   - Do not override font color in LMSGrader
   - Add default font color to body ($grey-7)
   - Use standard font sizes for LMSGrader
3. Adding retry functionality to this dialog
   - Extend the ErrorDialog component to have the ability to render a retry button and busy override similar to LaunchErrorDialog. nb. This could eventually replace BaseDialog inside LaunchErrorDialog.
   - Give grading fetch and submit errors a retry button.
   - Change LMSGrader so it can rebuild SubmitGradeForm component on onRetry callback to refetch student grades.

It looks to me like earlier PR https://github.com/hypothesis/lms/pull/2329 tried to address (1) and (2). I'm sorry to be a pain but I would prefer to review a PR that addresses (1) and (2) before (3). I know that Sean added a bunch of feedback in https://github.com/hypothesis/lms/pull/2329 which wasn't strictly on the original issue, but I still think it makes sense to try and land the fixes for the problems that were on the original issue, plus anything very closely connected, before adding new functionality.

Regarding this comment in the PR description:

>  did not remove the click-to-close feature were dialogs close when clicking outside of the content div because that has broad changing effects to every dialog in the app and we probably need to be consistent about how that works with cancelable dialogs.

I agree. This is a legitimate concern but let's not address that here as it affects other dialogs in the app.

robertknight (changes requested):
I really would prefer to review and land the fixes for the style inconsistencies and incorrect error message separately from the change to add retry functionality to the dialog. Would you mind splitting these two sets of changes into different PRs.

LMS007:
> I really would prefer to review and land the fixes for the style inconsistencies and incorrect error message separately from the change to add retry functionality to the dialog. Would you mind splitting these two sets of changes into different PRs.

Not a prob, and I agree.

robertknight:
On master:

fetch-grade-error-master

This branch:

fetch-grade-error

Example other error dialog:

auth-failed-error

This is obviously much better, though I notice the dialog title has a different shade. The "Fetch Grade Error" text is something that needs improving. That can be done separately.

robertknight (changes requested):
I suggest to remove the `body` color change from this PR and I had a query about the `line-height` change. Otherwise the changes here look good.

lms/static/styles/components/_ErrorDisplay.scss
```diff
@@ -12,7 +12,6 @@
 
   &__links {
     white-space: normal;
-    line-height: 1.25em;
```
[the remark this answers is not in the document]
LMS007: It was a change done originally to make the word wrapping of the links look better because the default line-height is fixed at `px` rather than `em`, but now since the font is back to default size, things look okay when wrapped. I just decided to remove the override.

LMS007:
Follow up PR
https://github.com/hypothesis/lms/issues/2421

robertknight (approved):
LGTM

PR #2395, opened 2021-02-08
Sign via3 urls

marcospri:
Need to update h-vialib version to https://github.com/hypothesis/h-vialib/pull/9/files

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -1,5 +1,5 @@
 """Via-related view helpers."""
-from h_vialib import ViaClient, ViaDoc
+from h_vialib import ViaClient
 
 __all__ = ["via_url"]
 
@@ -18,9 +18,8 @@ def via_url(request, document_url, content_type=None):
     :param content_type: Either "pdf" or "html" if known, None if not
     :return: A URL string
     """
-
-    doc = ViaDoc(document_url, content_type)
-
     return ViaClient(
-        service_url=request.registry.settings["via_url"], host_url=request.host_url
-    ).url_for(doc)
+        request.registry.settings["via_url"],
+        request.host_url,
+        request.registry.settings["via_secret"],
+    ).url_for(document_url, content_type)
--- a/tests/unit/lms/views/helpers/_via_test.py
+++ b/tests/unit/lms/views/helpers/_via_test.py
@@ -18,9 +18,12 @@ class TestViaURL:
 
         final_url = via_url(pyramid_request, url)
 
+        url_params = dict(self.DEFAULT_OPTIONS)
+        url_params["url"] = url
+        url_params["via.sec"] = Any.string()
         assert final_url == Any.url.matching(
             "http://test_via3_server.is/route"
-        ).with_query(dict(self.DEFAULT_OPTIONS, url=url))
+        ).with_query(url_params)
 
     pywb_test_params = pytest.mark.parametrize(
         "params,expected_extras",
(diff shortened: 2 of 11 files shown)


lms/views/helpers/_via.py
```diff
@@ -18,9 +19,10 @@ def via_url(request, document_url, content_type=None):
...
     :return: A URL string
     """
-
     doc = ViaDoc(document_url, content_type)
 
     return ViaClient(
-        service_url=request.registry.settings["via_url"], host_url=request.host_url
+        request.registry.settings["via_url"],
+        request.host_url,
+        ViaSecureURLToken(request.registry.settings["via_secret"]),
```
jon-betts, line 27: My suggestions in https://github.com/hypothesis/h-vialib/pull/9/files would mean you don't have to import the `ViaSecureURLToken`, that would be an internal detail of the client. We'd just pass the secret here.

tests/unit/lms/views/helpers/_via_test.py
```diff
@@ -20,7 +20,13 @@ def test_if_creates_the_correct_via_url(self, pyramid_request):
 
         assert final_url == Any.url.matching(
             "http://test_via3_server.is/route"
-        ).with_query(dict(self.DEFAULT_OPTIONS, url=url))
+        ).with_query(
+            dict(
+                self.DEFAULT_OPTIONS,
+                url=url,
+                sec="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1cmwiOiJodHRwOi8vVEVTVF9WSUEzX1NFUlZFUi5pcy9yb3V0ZT91cmw9aHR0cCUzQSUyRiUyRmV4YW1wbGUuY29tIiwiZXhwIjoxNjEyNzc4NDAwfQ.X1ydoVIX59VZivCdswfe65t8by1k75RJYUI-hE_EC9k",
```
jon-betts, line 27: I think we might want to vague this up a bit, unless we care about the specific value or implementation for some reason.

`Any.string()`?

lms/views/helpers/_via.py
```diff
@@ -18,9 +18,8 @@ def via_url(request, document_url, content_type=None):
...
     """
-
-    doc = ViaDoc(document_url, content_type)
-
     return ViaClient(
-        service_url=request.registry.settings["via_url"], host_url=request.host_url
-    ).url_for(doc)
+        request.registry.settings["via_url"],
+        request.host_url,
+        request.registry.settings["via_secret"],
```
jon-betts, line 24: I think as we add more options here it's good to keep the kwargs to say what they all are. Easy to swap two and not be able to spot it.

I'll be adding another one for ViaHTML pretty soon.

This probably should be a service... but who's got time for that?

tox.ini
```diff
@@ -61,6 +61,7 @@ setenv =
     functests: TEST_DATABASE_URL = {env:TEST_DATABASE_URL:postgresql://postgres@localhost:5433/lms_functests}
     bddtests: TEST_DATABASE_URL = {env:TEST_DATABASE_URL:postgresql://postgres@localhost:5433/lms_bddtests}
     OBJC_DISABLE_INITIALIZE_FORK_SAFETY = YES
+    VIA_SECRET = not_secret
```
jon-betts, line 64: I don't know why this bugs me, but we use "notasecret" in `tests/conftest.py` and in most products it's "not_a_secret"

jon-betts (approved):
Holy moley that requirements update is a doozy. I think the removal of ViaDoc as part of the interface is the right play here. It's much nicer now.

My only comment here is we use "not_a_secret" in other products and places, and we need these to line up between products for our dev environment to work, so it might be a good idea to be consistent.

jon-betts:
I should say, my approval is conditional on getting the CI passing etc!

PR #2415, opened 2021-02-10
Support VitalSource bookshelf launches in the LMS frontend

robertknight:
This PR contains the frontend changes only from https://github.com/hypothesis/lms/pull/2390 which @LMS007 has already reviewed. I'm going to land those and change the scope of https://github.com/hypothesis/lms/pull/2390 to only the backend changes which still need review.

----

This commit adds support in the frontend for loading the VitalSource book viewer
in an LTI launch, when the backend provides the necessary configuration.
The backend logic needed to generate the form parameters is not yet
present and will be added separately.

Changes in this commit:

 - Check for VitalSource configuration from the backend in
   `BasicLtiLaunchApp` and respond by rendering a
   `VitalSourceBookViewer` instead of the normal Via iframe
 - Add a `VitalSourceBookViewer` component which creates an empty
   iframe, renders form parameters from the backend into it and
   auto-submits the form.

See https://github.com/hypothesis/lms/pull/2390 for an overview of the VitalSource launch process.


robertknight:
The frontend changes were reviewed by Kyle here: https://github.com/hypothesis/lms/pull/2390#pullrequestreview-584608290.

PR #2438, opened 2021-02-18
Separate VitalSource launch factory

seanh:
This tweaks the Vital Source factory and tests to use our latest approach to services:

Add a separate `factory()` function for `VitalSourceService` so that the `VitalSourceService` class itself is more "pure" and doesn't depend on the Pyramid context and request.


robertknight (approved):
LGTM. I'll get this merged into https://github.com/hypothesis/lms/pull/2416 and then address the remaining feedback.

PR #2441, opened 2021-02-19
Reformat requirements files on make requirements

marcospri:
To test it just change any dependency, for example change:

`alembic`
to 
`alembic==1.5.2`

on `requirements/requirements.in`

and `make requirements`

the resulting diff should just be individual lines changing alembic's version (instead of that + reformatting)

### DIFF
--- a/requirements/format.txt
+++ b/requirements/format.txt
@@ -4,14 +4,25 @@
 #
 #    pip-compile requirements/format.in
 #
-appdirs==1.4.4            # via black
-black==20.8b1             # via -r requirements/format.in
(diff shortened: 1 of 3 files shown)


requirements/format.txt
```diff
@@ -4,14 +4,25 @@
 #
 #    pip-compile requirements/format.in
 #
-appdirs==1.4.4            # via black
```
marcospri, line 7: This file hasn't seen an update since the new formatting landed on pip-tools

seanh (approved):

Issue #2451, opened 2021-02-24
Replace pycrypto


robertknight:
The LMS app currently depends on pycrypto which is unmaintained and does not work with the newest Python versions (3.9 at the time of writing).

We currently use this package for:

- A secure random number generator. This could be replaced with the builtin `secrets` package in Python
- Encrypting and decrypting the Canvas developer secret that is stored in the database with a server-instance specific secret - the `LMS_SECRET` var. I presume the logic was to limit the risk to third parties if an unauthorized copy of a database dump was obtained.

Possible replacements include [pycryptodome](https://github.com/Legrandin/pycryptodome), intended as a drop-in replacement for pycrypto, and [cryptography](https://cryptography.io/en/latest/).

seanh:
I think this was done

PR #2452, opened 2021-02-24
Fix resource leaks that cause test failures in newer Python versions

robertknight:
This PR fixes a couple of unclosed files that caused tests to fail with `ResourceWarning` errors in newer Python versions. It also accommodates a trivial change in the text of an error message between Python 3.6.x and Python 3.9.x.

Following this change all of the unit tests which do not depend on PyCrypto pass under Python 3.6.x and Python 3.9.x. The remaining cause of test failures with newer Python versions in the lms app is the unmaintained PyCrypto package which we'll need to replace. See https://github.com/hypothesis/lms/issues/2451.

See also https://github.com/hypothesis/h/pull/6508

### DIFF
--- a/lms/assets.py
+++ b/lms/assets.py
@@ -54,8 +54,9 @@ class _CachedFile:
 
         current_mtime = os.path.getmtime(self.path)
         if not self._mtime or self._mtime < current_mtime:  # pragma: no cover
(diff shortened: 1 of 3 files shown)


marcospri (approved):
Nice one, will fix the warning but also the actual possible leak :+1:

PR #2458, opened 2021-02-26
Fix `configured` predicate test not checking whether assignment is URL-configured

robertknight:
Fix a regression introduced in https://github.com/hypothesis/lms/pull/2416 where the check for an assignment being URL-configured in the `configured` predicate was accidentally removed. Also fix a flaw in the tests that caused this issue to go unnoticed. See commit message for details.

I also added a test for the assignment being configured via a VitalSource book. The functionality was implemented but not tested in https://github.com/hypothesis/lms/pull/2416.

### DIFF
--- a/tests/unit/lms/views/predicates/_lti_launch_test.py
+++ b/tests/unit/lms/views/predicates/_lti_launch_test.py
@@ -163,7 +163,11 @@ class TestConfigured:
         assert predicate(mock.sentinel.context, pyramid_request) is expected
 
     @pytest.mark.parametrize("value,expected", [(True, True), (False, False)])
-    def test_when_assignment_is_db_configured(self, pyramid_request, value, expected):
+    def test_when_assignment_is_db_configured(
+        self, pyramid_request, assignment_service, value, expected
+    ):
+        assignment_service.get_document_url.return_value = "https://example.com"
+
         predicate = Configured(value, mock.sentinel.config)
 
         assert predicate(mock.sentinel.context, pyramid_request) is expected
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -163,7 +163,20 @@ def test_when_assignment_is_canvas_file(self, pyramid_request, value, expected):
         assert predicate(mock.sentinel.context, pyramid_request) is expected
 
     @pytest.mark.parametrize("value,expected", [(True, True), (False, False)])
-    def test_when_assignment_is_db_configured(self, pyramid_request, value, expected):
+    def test_when_assignment_is_db_configured(
+        self, pyramid_request, assignment_service, value, expected
+    ):
+        assignment_service.get_document_url.return_value = "https://example.com"
```
seanh, line 169:
```suggestion
        assignment_service.get_document_url.return_value = mock.sentinel.document_url
```

tests/unit/lms/views/predicates/_lti_launch_test.py
```diff
@@ -186,6 +199,13 @@ def pyramid_request(self, pyramid_request):
         }
         return pyramid_request
 
+    @pytest.fixture
```
seanh, line 202: Make sure this applies to all the tests in this class (not just the ones that explicitly use the `assignment_service` fixture):

```suggestion
    @pytest.fixture(autouse=True)
```

You could also make this fixture a module-level function instead of a class method, then it'd apply to all tests in this module. (Would require some small tweaks to `TestDBConfigured`).

seanh (changes requested):

seanh (approved):

PR #2501, opened 2021-03-17
Refresh grant token if necessary when Hypothesis client fetches configuration

robertknight:
~~**Depends on https://github.com/hypothesis/lms/pull/2500**~~

During an LTI launch the LMS backend generates a grant token for use by the Hypothesis client that is valid for 5 minutes.

If the Hypothesis client attempts to fetch its configuration after this time has elapsed then the client will fail to log in. With the sidebar application this rare because the sidebar launches as soon as Via loads. It could happen however if the LMS frontend did not load the assignment content immediately but instead paused for Canvas OAuth and the user took more than 5 minutes to respond. With the client's "Notebook" application this is expected to be much more common however, because the notebook isn't loaded until the user explicitly chooses the "Open notebook" menu option.

Fix the issue by modifying the LMS frontend to request a new grant token if the current one has expired when responding
to a `requestConfig` RPC request from the Hypothesis client.

 - Add a new `/api/grant_token` view in the backend (see `lms.views.api.grant_token`) that generates a fresh login token on demand. This is authenticated using the same JWT token as other LMS frontend API calls
 - Add a `JWT` JS class (`frontend_apps/utils/jwt`) that wraps a server-generated JWT and provides methods to check whether it is still valid.
 - Modify `ClientRpc` JS class to use the `JWT` helper to check if the JWT is still valid before returning config to the client, and fetch a new grant token otherwise.

Part of https://github.com/hypothesis/client/issues/2801

----

**Testing:**

1. Make a couple of local changes to shorten grant token lifetimes:

```diff
diff --git a/lms/services/grant_token.py b/lms/services/grant_token.py
index cbd054a9..45cd8dc0 100644
--- a/lms/services/grant_token.py
+++ b/lms/services/grant_token.py
@@ -38,7 +38,7 @@ class GrantTokenService:
             "iss": self._h_jwt_client_id,
             "sub": h_user.userid(self._h_authority),
             "nbf": now,
-            "exp": now + datetime.timedelta(minutes=5),
+            "exp": now + datetime.timedelta(seconds=10),
         }
 
         return jwt.encode(
diff --git a/lms/static/scripts/frontend_apps/services/client-rpc.js b/lms/static/scripts/frontend_apps/services/client-rpc.js
index 619f168e..a8133fb2 100644
--- a/lms/static/scripts/frontend_apps/services/client-rpc.js
+++ b/lms/static/scripts/frontend_apps/services/client-rpc.js
@@ -58,7 +58,7 @@ export class ClientRpc {
     // A convervative estimate of when the grant token was issued.
     // When this is older than the true value, the frontend will just consider it
     // to "expire" earlier than it really does.
-    const issuedAt = Date.now() - 30 * 1000;
+    const issuedAt = Date.now() - 1000;
     let grantToken = new JWT(clientConfig.services[0].grantToken, issuedAt);
 
     // Handle the requests for configuration from the Hypothesis client.
```

2. Launch an assignment with the browser devtools open at the Network tab and the "XHR" filter selected.
3. During the initial assignment launch there should be no `/api/grant_token` API call as the token rendered by the backend during the initial launch is still valid (nb. If you _do_ see this call unexpectedly, try restarting the lms dev server just in case it didn't rebuild when the branch was checked out)
4. Wait at least 10 seconds, then click on the user account menu icon in the client's sidebar and press the "n" key to launch the notebook. There should be an `/api/grant_token` API call followed by the Notebook app appearing.

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -25,6 +25,8 @@ def includeme(config):
         factory="lms.resources.LTILaunchResource",
     )
 
+    config.add_route("api.grant_token", "/api/grant_token", request_method="GET")
+
     config.add_route(
         "blackboard_api.oauth.authorize",
         "/api/blackboard/oauth/authorize",
--- a/lms/static/scripts/frontend_apps/services/client-rpc.js
+++ b/lms/static/scripts/frontend_apps/services/client-rpc.js
@@ -14,15 +14,17 @@ import { JWT } from '../utils/jwt';
  */
 
 /**
- * The subset of the Hypothesis client configuration that `ClientRpc` references.
+ * The subset of the Hypothesis client configuration that `ClientRpc` directly references.
  *
- * The backend will set other configuration which is just forwarded to the client
- * and not touched by `ClientRpc`.
+ * The backend will add other properties which are intentionally not included here,
+ * even as an index signature. This other configuration is simply forwarded to
+ * the client but not touched by `ClientRpc`.
  *
  * See https://h.readthedocs.io/projects/client/en/latest/publishers/config/.
  *
  * @typedef ClientConfig
- * @prop {[ServiceConfig]} services
+ * @prop {[ServiceConfig]} services - Annotation configuration for the client.
+ *   In the LMS context this will always consist of exactly one entry.
  */
 
 /**
@@ -55,7 +57,7 @@ export class ClientRpc {
   constructor({ allowedOrigins, authToken, clientConfig }) {
     this._server = new Server(allowedOrigins);
 
-    // A convervative estimate of when the grant token was issued.
+    // A conservative estimate of when the grant token was issued.
     // When this is older than the true value, the frontend will just consider it
     // to "expire" earlier than it really does.
     const issuedAt = Date.now() - 30 * 1000;
--- a/lms/static/scripts/frontend_apps/services/client-rpc.js
+++ b/lms/static/scripts/frontend_apps/services/client-rpc.js
@@ -65,11 +65,17 @@ export class ClientRpc {
     this._server.register('requestConfig', async () => {
       if (grantToken.hasExpired()) {
         const issuedAt = Date.now();
-        const response = await apiCall({
-          authToken,
-          path: '/api/grant_token',
-        });
-        grantToken = new JWT(response.grant_token, issuedAt);
+        try {
+          const response = await apiCall({
+            authToken,
+            path: '/api/grant_token',
+          });
+          grantToken = new JWT(response.grant_token, issuedAt);
+        } catch (err) {
+          throw new Error(
+            'Unable to fetch Hypothesis login. Please reload the assignment.'
+          );
+        }
       }
 
       clientConfig.services[0].grantToken = grantToken.value();
--- a/lms/static/scripts/frontend_apps/services/client-rpc.js
+++ b/lms/static/scripts/frontend_apps/services/client-rpc.js
@@ -1,9 +1,28 @@
 import { Server, call as rpcCall } from '../../postmessage_json_rpc';
+import { apiCall } from '../utils/api';
+import { JWT } from '../utils/jwt';
 
 /**
  * @typedef User
  * @prop {string} displayName
  * @prop {string} userid
+ */
+
+/**
+ * @typedef ServiceConfig
+ * @prop {string} grantToken
+ */
+
+/**
+ * The subset of the Hypothesis client configuration that `ClientRpc` references.
+ *
+ * The backend will set other configuration which is just forwarded to the client
+ * and not touched by `ClientRpc`.
+ *
+ * See https://h.readthedocs.io/projects/client/en/latest/publishers/config/.
+ *
+ * @typedef ClientConfig
+ * @prop {[ServiceConfig]} services
  */
 
 /**
@@ -24,17 +43,38 @@ export class ClientRpc {
    * @param {Object} options
    *   @param {string[]} options.allowedOrigins -
    *     Origins that are allowed to request client configuration
-   *   @param {Object} options.clientConfig -
+   *   @param {string} options.authToken -
+   *     Auth token used in LMS backend API calls to refresh the grant token
+   *     in the `clientConfig` if it has expired
+   *   @param {ClientConfig} options.clientConfig -
    *     Configuration for the Hypothesis client. Whatever is provided here is
    *     passed directly to the client via `window.postMessage` when it requests
    *     configuration. It should be a subset of the config options specified at
    *     https://h.readthedocs.io/projects/client/en/latest/publishers/config/.
    */
-  constructor({ allowedOrigins, clientConfig }) {
+  constructor({ allowedOrigins, authToken, clientConfig }) {
     this._server = new Server(allowedOrigins);
 
-    // Handle the initial request for configuration from the Hypothesis client.
-    this._server.register('requestConfig', () => clientConfig);
+    // A convervative estimate of when the grant token was issued.
+    // When this is older than the true value, the frontend will just consider it
+    // to "expire" earlier than it really does.
+    const issuedAt = Date.now() - 30 * 1000;
+    let grantToken = new JWT(clientConfig.services[0].grantToken, issuedAt);
+
+    // Handle the requests for configuration from the Hypothesis client.
+    this._server.register('requestConfig', async () => {
+      if (grantToken.hasExpired()) {
+        const issuedAt = Date.now();
+        const response = await apiCall({
+          authToken,
+          path: '/api/grant_token',
+        });
+        grantToken = new JWT(response.grant_token, issuedAt);
+      }
+
+      clientConfig.services[0].grantToken = grantToken.value();
+      return clientConfig;
+    });
 
     const groups = new Promise(resolve => {
       this._resolveGroups = resolve;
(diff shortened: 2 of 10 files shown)


robertknight:
Noting an issue that needs to be resolved: If the JWT token used by the LMS frontend to authenticate to the LMS backend has expired by the time that the notebook requests its grant token, it currently displays blank output:

Load failed notice

From a user-facing point of view, I think the best we can do at this point is tell the user to re-launch the assignment.

Within the client we ought to have better handling of configuration fetch errors anyway. Currently we log something to the console but there is no visual output.

On the basis of the above, I think the fix could involve:

 1. Catching the auth error on the backend and returning a more meaningful error to the client
 2. Adding logic in the client to show to the user any errors that were received when fetching configuration

robertknight:
The error handling has been improved so that, together with https://github.com/hypothesis/client/pull/3167, a helpful error message will be shown to the user if the grant token fetch fails for any reason.

The most likely reason for it to fail currently is that the JWT expired. This is addressed in https://github.com/hypothesis/lms/pull/2502.

lms/static/scripts/frontend_apps/utils/jwt.js
```diff
@@ -0,0 +1,54 @@
...
+    this._token = token;
+
+    const [, payloadBase64] = token.split('.');
+    this._payload = /** @type {JWTPayload} */ (JSON.parse(atob(payloadBase64)));
+
+    // Estimated offset of server's clock relative to client.
+    //
+    // +ve if the server's clock is ahead of the client or -ve if the server's
+    // clock is behind.
+    const skew = this._payload.nbf * 1000 - issuedAt;
```
esanzgar, line 30: Maybe useful to add a comment that `nbf` stands for `not before`?
https://tools.ietf.org/html/rfc7519#page-10

lms/static/scripts/frontend_apps/services/client-rpc.js
```diff
@@ -24,17 +43,44 @@ export class ClientRpc {
...
+  constructor({ allowedOrigins, authToken, clientConfig }) {
     this._server = new Server(allowedOrigins);
 
-    // Handle the initial request for configuration from the Hypothesis client.
-    this._server.register('requestConfig', () => clientConfig);
+    // A convervative estimate of when the grant token was issued.
+    // When this is older than the true value, the frontend will just consider it
+    // to "expire" earlier than it really does.
+    const issuedAt = Date.now() - 30 * 1000;
+    let grantToken = new JWT(clientConfig.services[0].grantToken, issuedAt);
```
esanzgar, line 62: This could raise an exception if the JWT is malformed. Maybe it could be wrap into a try/catch?

lms/static/scripts/frontend_apps/utils/jwt.js
```diff
@@ -0,0 +1,54 @@
...
+   * @param {number} [now] - Current timestamp in milliseconds
+   */
+  hasExpired(now = Date.now()) {
+    return now > this._validUntil;
+  }
+
+  value() {
+    if (this.hasExpired()) {
+      throw new Error('Tried to use an expired JWT token');
+    }
```
esanzgar, line 47: I think the responsibility of the backend server to validate the token, including the expire date.

esanzgar (commented):
I just leave a few comments for you to consider.

lms/static/scripts/frontend_apps/utils/jwt.js
esanzgar: According to the standard, it maybe better if the backend generates the JWT with a `nbf` with a little margin: 

```
Implementers MAY provide for some small leeway, usually no more than a few minutes,
```

lms/routes.py
```diff
@@ -25,6 +25,8 @@ def includeme(config):
         factory="lms.resources.LTILaunchResource",
     )
 
+    config.add_route("api.grant_token", "/api/grant_token", request_method="GET")
```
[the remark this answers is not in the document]
robertknight: In my mind one reason to use `POST` is to signal that backend state is being modified in some way. For an OAuth refresh token that is true (new tokens are generated and stored in the DB, the previous one should be invalidated). In this case however fetching a new grant token doesn't invalidate the previous one or modify any backend state. I see this API as more like reading a clock.

lms/views/api/grant_token.py
```diff
@@ -0,0 +1,15 @@
...
+@view_config(permission="api", renderer="json", route_name="api.grant_token")
+def grant_token(request):
+    """
+    Return a grant token that the Hypothesis client can use to log in to H.
+
+    See https://h.readthedocs.io/en/latest/publishers/authorization-grant-tokens/.
+    """
+
+    grant_token_svc = request.find_service(name="grant_token")
+    h_user = request.lti_user.h_user
```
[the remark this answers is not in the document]
robertknight: Are you referring to verification of the login token that this API call generates, or the JWT token used to authenticate this call?

- The h server verifies the grant token generated by the LMS backend
- The lms backend verifies the JWT token used to authenticate the `/api/grant_token` call

lms/static/scripts/frontend_apps/services/client-rpc.js
```diff
@@ -1,11 +1,30 @@
...
+/**
+ * The subset of the Hypothesis client configuration that `ClientRpc` references.
+ *
+ * The backend will set other configuration which is just forwarded to the client
+ * and not touched by `ClientRpc`.
+ *
+ * See https://h.readthedocs.io/projects/client/en/latest/publishers/config/.
+ *
+ * @typedef ClientConfig
+ * @prop {[ServiceConfig]} services
```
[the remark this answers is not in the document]
robertknight: The fact that `services` is an array is part of the client's configuration: https://h.readthedocs.io/projects/client/en/latest/publishers/config/#cmdoption-arg-services

The reason this is typed as an array of one item is because in the LMS context we know that this array will always be generated on the backend with one entry. I'll add a comment to clarify.

lms/static/scripts/frontend_apps/services/client-rpc.js
```diff
@@ -24,17 +43,44 @@ export class ClientRpc {
...
    *     configuration. It should be a subset of the config options specified at
    *     https://h.readthedocs.io/projects/client/en/latest/publishers/config/.
    */
-  constructor({ allowedOrigins, clientConfig }) {
+  constructor({ allowedOrigins, authToken, clientConfig }) {
     this._server = new Server(allowedOrigins);
 
-    // Handle the initial request for configuration from the Hypothesis client.
-    this._server.register('requestConfig', () => clientConfig);
+    // A convervative estimate of when the grant token was issued.
```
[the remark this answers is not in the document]
robertknight: Whoops!

lms/static/scripts/frontend_apps/services/client-rpc.js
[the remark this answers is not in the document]
robertknight: In this context it is intentional that there is no index signature. The `ClientConfig` type _only_ documents the configuration that `ClientRpc` directly uses:

```
 * The subset of the Hypothesis client configuration that `ClientRpc` references.
 *
 * The backend will set other configuration which is just forwarded to the client
 * and not touched by `ClientRpc`.
```

This ensures that a type checking error is used if `ClientRpc` tries to reference any other properties. I'll reword the comment to clarify.

lms/static/scripts/frontend_apps/utils/jwt.js
robertknight: Hmm - thinking about this it would be better to set an `iat` (issued at) timestamp on the backend rather than rely on `nbf` being the issued-at timestamp.

lms/static/scripts/frontend_apps/utils/jwt.js
robertknight: The h server does validate the token, but only once the client has received it and submitted it to the backend. The purpose of this code is to quickly check whether a refresh is needed before sending the token on to the client or for checking if an authorization failure may have been specifically caused by token expiry.

I'll update the `hasExpired` docs to clarify this.

robertknight:
Thanks for the feedback @esanzgar. I have made a few changes in response:

- The client side expiry checking is now based on `iat` (Issued At) rather than `nbf` (which is currently equal to `iat`, but is not intended for the purpose of determining JWT age because of leeway). The `iat` field has been added to the token on the server side.
- I've updated the `JWT` class docs to add links to relevant resources
- I've added a note on why `ClientConfig` only includes one field, why `services` is typed as it is and why there is no index signature
- I've changed the `JWT` class on the client side to throw more helpful errors if the token is missing needed fields

lms/static/scripts/frontend_apps/services/client-rpc.js
robertknight: If the JWT is malformed here then it means there is a programming error on the backend, in which case we just want a helpful exception so we can locate the problem. I've modified the `JWT` class so that it reports a more helpful error (eg. "Failed to parse JWT" rather than a generic "JSON parse error" or similar).

lms/static/scripts/frontend_apps/services/client-rpc.js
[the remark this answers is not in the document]
esanzgar: I see... it is clear.

esanzgar (approved):

PR #2512, opened 2021-03-25
Canvas Groups POC

marcospri:
Head to 

https://hypothesis.instructure.com/courses/125/assignments/1069

as Hypothesis 101 Student you should see one annotation and the group name at the top should be "others"

as an admin/teacher you should see both groups in the selector ("Seans" & "others") and toggling between them should hide/show one annotation.

I haven't figured out how to the authentication nicely so the process is not great, my code on the view doesn't seem to trigger the oauth flow, but triggering it without will request a token with the right scopes so that works.

- Deleting all the tokens in postgres with

```
truncate oauth2_token;
````

- setting `False` on basic_lti_launch.py:201

- open the assignment and complete the authorization workflow

https://hypothesis.instructure.com/courses/125/assignments/1069

- set `True` again on basic_lti_launch.py:201

The assignment should now be a "group assignment", unfortunately you need to do this process again if testing with another user.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -465,6 +465,9 @@ class JSConfig:
             g.groupid(self._authority) for g in groups
         ]
 
+    def yolo_group_categories(self, group_categories):
+        self._config["filePicker"]["canvas"]["groupCategories"] = group_categories
+
     def _groups(self):
         if self._context.canvas_sections_enabled:
             return "$rpc:requestGroups"
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -4,7 +4,9 @@ from urllib.parse import urlparse
 
 import jwt
 
-from lms.models import GroupInfo, HUser
+from typing import List
+
+from lms.models import GroupInfo, HUser, HGroup
 from lms.services import ConsumerKeyError, HAPIError
 from lms.validation.authentication import BearerTokenSchema
 from lms.views.helpers import via_url
@@ -457,6 +459,11 @@ class JSConfig:
                 }
             ]
         }
+
+    def set_groups(self, groups: List[HGroup]):
+        self._hypothesis_client["services"][0]["groups"] = [
+            g.groupid(self._authority) for g in groups
+        ]
 
     def _groups(self):
         if self._context.canvas_sections_enabled:
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -6,6 +6,13 @@ from marshmallow import EXCLUDE, Schema, fields, post_load, validate, validates_
 from lms.services import CanvasAPIError, CanvasFileNotFoundInCourse
 from lms.validation import RequestsResponseSchema
 
+from lms.services.canvas_api.schemas import (
+    ListGroups,
+    ListGroupCategories,
+    ListGroupMemberships,
+    Profile,
+)
+
 
 class _SectionSchema(Schema):
     """
@@ -215,6 +222,35 @@ class CanvasAPIClient:
             f"courses/{course_id}/files",
             params={"content_types[]": "application/pdf"},
             schema=self._ListFilesSchema,
+        )
+
+    def get_course_groups(self, course_id, only_own_groups=True):
+        return self._client.send(
+            "GET",
+            f"courses/{course_id}/groups",
+            params={"only_own_group": only_own_groups},
+            schema=ListGroups,
+        )
+
+    def get_course_group_categories(self, course_id):
+        return self._client.send(
+            "GET", f"courses/{course_id}/group_categories", schema=ListGroupCategories
+        )
+
+    def get_groups_in_group_category(self, group_category_id):
+        return self._client.send(
+            "GET", f"group_categories/{group_category_id}/groups", schema=ListGroups
+        )
+
+    def get_group_members(self, group_id):
+        return self._client.send(
+            "GET", f"groups/{group_id}/memberships", schema=ListGroupMemberships
+        )
+
+    def get_user_profile(self, user_id):
+        # https://community.canvaslms.com/t5/Question-Forum/lti-user-id-attribute-is-missed-in-Profile-API-response/td-p/145093/page/2
+        return self._client.send(
+            "GET", f"users/{user_id}/profile?as_user_id={user_id}", schema=Profile
         )
 
     class _ListFilesSchema(RequestsResponseSchema):
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -228,7 +228,7 @@ class CanvasAPIClient:
         return self._client.send(
             "GET",
             f"courses/{course_id}/groups",
-            params={"only_own_group": only_own_groups},
+            params={"only_own_groups": only_own_groups},
             schema=ListGroups,
         )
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: lms/resources/_js_config.py [hunk 1]
```diff
@@ -458,6 +460,11 @@ def _hypothesis_client(self):
             ]
         }
 
+    def set_groups(self, groups: List[HGroup]):
+        self._hypothesis_client["services"][0]["groups"] = [
```
marcospri, line 464: Not sure about the `[0]` part here, seems unreliable

lms/services/canvas_api/client.py
```diff
@@ -217,6 +224,35 @@ def list_files(self, course_id):
             schema=self._ListFilesSchema,
         )
 
+    def get_course_groups(self, course_id, only_own_groups=True):
```
marcospri, line 227: There's more method here than are actually needed but will trim them when everything it's more clear

lms/services/canvas_api/schemas.py
```diff
@@ -0,0 +1,53 @@
+from enum import Enum
```
marcospri, line 1: Added this file here as I initially thought this were the types being returning but they are only validated against them.

Well clean the ones not used and move them back to `canvas/client`

lms/services/canvas_api/schemas.py
```diff
@@ -0,0 +1,53 @@
...
+    id = fields.Integer(required=True)
+
+    name = fields.Str(required=True)
+    description = fields.String(default=None, allow_none=True)
+
+    role = fields.Str(required=False, allow_none=True)
+    content_type = fields.Str(required=False, allow_none=True)
+
+
+class MembershipState(Enum):
```
marcospri, line 30: I wanted to filter by these but it's not actually needed for now

lms/views/api/canvas/authorize.py
```diff
@@ -31,6 +31,16 @@
 )
 
 
+#: The Canvas API scopes that we need for our groups feature.
+GROUPS_SCOPES = (
+    "url:GET|/api/v1/courses/:course_id/groups",
```
marcospri, line 36: Again, more that we need here, will clean as I clean the methods on canvas/client

lms/views/basic_lti_launch.py
```diff
@@ -188,7 +194,64 @@ def url_configured_basic_lti_launch(self):
         LMS, which passes it back to us in each launch request. All we have to
         do is pass the URL to Via.
         """
-        return self.basic_lti_launch(self.request.parsed_params["url"])
+        groups = None
+
+        lti_user = self.request.lti_user
+        canvas = self.request.find_service(name="canvas_api_client")
+        if True:  # "is_group_assigment":
```
marcospri, line 201: Not very nice code ahead but it just gets the courses a student belongs to and finds the one in the group set. Syncs to H and sets `groups` to make it available in the client.

lms/views/basic_lti_launch.py
```diff
@@ -188,7 +194,64 @@ def url_configured_basic_lti_launch(self):
...
+                        self.request.params["tool_consumer_instance_guid"],
+                        course_id,
+                    )
+                ]
+
+                self.request.find_service(name="lti_h").sync(
+                    groups,
+                    self.request.params,
+                )
+            elif lti_user.is_instructor:
```
marcospri, line 234: Very similar to the student part but here it gets all the groups in the group set.

lms/views/basic_lti_launch.py
```diff
@@ -188,7 +194,64 @@ def url_configured_basic_lti_launch(self):
         LMS, which passes it back to us in each launch request. All we have to
         do is pass the URL to Via.
         """
-        return self.basic_lti_launch(self.request.parsed_params["url"])
+        groups = None
```
marcospri, line 197: This doesn't really belong to `url_configured_basic_lti_launch` but I found I would trigger the authorization flow if it was here (to get the Canvas API credentials)

seanh:
I'm not sure why this isn't triggering the frontend's authorization flow. In case it helps here's how it's supposed to work:

1. The API views (or services called by the views, like the Canvas API client service) raise exceptions like [`NoOAuth2Token`](https://github.com/hypothesis/lms/blob/f8274bbd288aa350437b070f1ebb81b7b59d66eb/lms/services/exceptions.py#L27-L28), [`CanvasAPIAccessTokenError`](https://github.com/hypothesis/lms/blob/f8274bbd288aa350437b070f1ebb81b7b59d66eb/lms/services/exceptions.py#L168-L178), and others in that file. (Often the views just call service methods and do not catch the exceptions, just let them be raised through.)

   For example if we have no access token for this user in the DB then `NoOAuth2Token` will be raised. If we do have an access token but it doesn't work (maybe it's expired, or lacks the right scopes) then `CanvasAPIAccessTokenError` will be raised.

2. When a view raises an exception Pyramid catches it and passes it to our [exception views](https://github.com/hypothesis/lms/blob/master/lms/views/api/exceptions.py). These format JSON error responses for our frontend in particular ways depending on the exception. See [this docstring](https://github.com/hypothesis/lms/blob/f8274bbd288aa350437b070f1ebb81b7b59d66eb/lms/views/api/exceptions.py#L26-L89). 

3. Any `400 Bad Request` response should cause the frontend to show an <kbd>Authorize</kbd> button for the user to get us a new access token. The frontend will either present this as an error ("Something went wrong, press this <kbd>Try again</kbd> button) or not as an error ("You need to authorize us, press this <kbd>Authorize</kbd> button") depending on whether the JSON body of the `400 Bad Request` contains an error message or not

seanh:
This [`CanvasAPIError.raise_from()` method](https://github.com/hypothesis/lms/blob/f8274bbd288aa350437b070f1ebb81b7b59d66eb/lms/services/exceptions.py#L76-L165) (and the `_exception_class()` helper below it) are important in deciding which `lms.services` exception to raised based on an originating exception that happened when trying to call the API

lms/resources/_js_config.py
```diff
@@ -458,6 +460,14 @@ def _hypothesis_client(self):
...
+            g.groupid(self._authority) for g in groups
+        ]
+
+    def yolo_group_categories(self, group_categories):
```
marcospri, line 468: I endup sending the actual list of group categories instead of the URL for an API call is it was easier for me to setup.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -1,10 +1,11 @@
-import { createElement } from 'preact';
+import { Fragment, createElement } from 'preact';
```
marcospri, line 1: Everything in this file is just me trying to make it work with as less changes as possible so it's wrong in many ways.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -124,9 +127,14 @@ export default function FilePickerApp({
   const selectURL = url => {
     cancelDialog();
     setUrl(url);
-    submit(true);
+    //submit(true);
```
marcospri, line 130: No logic here, just not send it on manual URL assignments and do send it on group set selection

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -247,6 +257,19 @@ export default function FilePickerApp({
               onClick={selectVitalSourceBook}
             />
           )}
+
+          <input name="canvas_group_set_id" type="hidden" value={groupSet || ''} />
+          { true && (
+            <Table 
+                 columns={[{label: "Group set", "className": "FileList__name-header"}]}
+                 items={[...[{"name": "Not a group assignment", "id": null}],...groupCategories]}
```
marcospri, line 265: Stackoverflow "how to add two lists in JS "

PR #2534, opened 2021-03-30
Fix two flakey tests

seanh:
These two tests are failing for me locally on master:

    ________________________________________________ TestEncodeJWT.test_it_uses_one_hour_lifetime_by_default ________________________________________________

    self = <tests.unit.lms.validation.authentication._helpers._jwt_test.TestEncodeJWT object at 0x7f4deea63c50>

        def test_it_uses_one_hour_lifetime_by_default(self):
            jwt_str = _jwt.encode_jwt({}, "test_secret")
            decoded_payload = jwt.decode(jwt_str, "test_secret", algorithms=["HS256"])
            lifetime_minutes = self._jwt_lifetime(decoded_payload) / 60

    >       assert round(lifetime_minutes) == 60
    E       assert 120 == 60
    E        +  where 120 = round(119.98486773173015)

    tests/unit/lms/validation/authentication/_helpers/_jwt_test.py:95: AssertionError
    ______________________________________________________ TestEncodeJWT.test_it_uses_custom_lifetime _______________________________________________________

    self = <tests.unit.lms.validation.authentication._helpers._jwt_test.TestEncodeJWT object at 0x7f4deec06eb8>

        def test_it_uses_custom_lifetime(self):
            jwt_str = _jwt.encode_jwt(
                {}, "test_secret", lifetime=datetime.timedelta(hours=24)
            )
            decoded_payload = jwt.decode(jwt_str, "test_secret", algorithms=["HS256"])
            lifetime_hours = self._jwt_lifetime(decoded_payload) / (60 * 60)

    >       assert round(lifetime_hours) == 24
    E       assert 25 == 24
    E        +  where 25 = round(24.99973412030273)

    tests/unit/lms/validation/authentication/_helpers/_jwt_test.py:104: AssertionError

I think it's probably best to mock `datetime` for these time-based tests. I've also mocked `jwt` because if you don't it looks like `jwt` makes its own calls to `datetime` so you still get time-related failures. The two tests in question don't really need to be tested against the real `jwt` anyway: they're just tests for what expiry time our code passes in the payload to `jwt.encode()`. Mocking `jwt` is easier than decoding the payload and then trying to calculate the
expected expiry timestamp.

### DIFF
--- a/tests/unit/lms/validation/authentication/_helpers/_jwt_test.py
+++ b/tests/unit/lms/validation/authentication/_helpers/_jwt_test.py
@@ -1,5 +1,5 @@
 import copy
-import datetime
+import datetime as datetime_
 
 import jwt
 import pytest
@@ -32,7 +32,7 @@ class TestDecodeJWT:
         jwt_str = self.encode_jwt(
             {
                 "TEST_KEY": "TEST_VALUE",
-                "exp": datetime.datetime.utcnow() - datetime.timedelta(hours=1),
+                "exp": datetime_.datetime.utcnow() - datetime_.timedelta(hours=1),
             }
         )
 
@@ -61,7 +61,7 @@ class TestDecodeJWT:
             # Insert an unexpired expiry time into the given payload dict, if
             # it doesn't already contain an expiry time.
             payload.setdefault(
-                "exp", datetime.datetime.utcnow() + datetime.timedelta(hours=1)
+                "exp", datetime_.datetime.utcnow() + datetime_.timedelta(hours=1)
             )
 
         jwt_str = jwt.encode(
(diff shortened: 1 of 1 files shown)


robertknight:
I can reproduce the failures. Its not flakiness though: Look at the difference between the actual and expected values. They are one hour out. I would guess this has to do with the recent change to daylight savings in the UK (where you and I are running the tests) which puts us one hour ahead of UTC. The production server's clock is always UTC so it probably isn't affected, but I'd like to make sure.

robertknight:
This was indeed a timezone handling issue, due to a lack of depth of knowledge on my part about Python datetime handling. The fix is a one-liner: https://github.com/hypothesis/lms/pull/2541.

Given that we are passing a datetime instance to PyJWT and then relying on it to correctly encode this internally, I do think it is better not to mock here, on the basis that we want to find out if that behavior ever changes in future.

I'm going to close this PR for now in favor of https://github.com/hypothesis/lms/pull/2541. Please shout if you object.

PR #2536, opened 2021-03-31
Improve lti outcome error detail

LMS007:
In trying to track down the grading issue we have seen in moodle with several customers, I had a difficult time finding details about the request errors beecause paper trail cuts off after so many days -- and I can't see logs in newrelic. I think It probably would be useful to capture to the details with the error so they can be preserved in a screen shot with a hubspot ticket. I'd like to hand this off as a start point to gaining more error information so we can track down moodle issues with grading. 

This will still not relay the school or LMS service and it may be nice to also do that, but that will probably be easy to figure out from the ticket in any case.  If there is anything else that may be useful to relay to the error, let's add that too.

Relates to https://github.com/hypothesis/support/issues/169

TODO:
- [ ] - fix client tests
- [ ] - should we relay additional details to the client?

Screen Shot 2021-03-31 at 4 37 17 PM

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -146,7 +146,7 @@ class LTIOutcomesClient:
         except KeyError as err:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
-        if status == "success":
+        if status != "success":
             raise LTIOutcomesAPIError(explanation="LTI outcome request failed", response=data, details=data)
 
         return body
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -146,8 +146,8 @@ class LTIOutcomesClient:
         except KeyError as err:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
-        if status != "success":
-            raise LTIOutcomesAPIError("LTI outcome request failed")
+        if status == "success":
+            raise LTIOutcomesAPIError(explanation="LTI outcome request failed", response=data, details=data)
 
         return body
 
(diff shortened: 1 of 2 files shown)


lms/services/lti_outcomes.py
```diff
@@ -147,7 +147,7 @@ def _get_body(cls, data):
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
         if status != "success":
-            raise LTIOutcomesAPIError("LTI outcome request failed")
+            raise LTIOutcomesAPIError(explanation="LTI outcome request failed", response=data, details=data)
```
LMS007, line 150: Do we need to also pass `response=` here? I think that will change the logs a bit, but not exactly clear how.

robertknight: The `response` is already set by the caller of `_get_body` above when catching a `LTIOutcomesAPIError` exception, so it doesn't need to be set here.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -220,19 +221,19 @@ export default function SubmitGradeForm({ student }) {
       {!!submitGradeError && (
         <ErrorDialog
           title="Submit Grade Error"
-          error={{ message: submitGradeError }}
+          error={ submitGradeError }
```
robertknight, line 224: While this code is being updated, can you update the `title` prop of this dialog and the one below so that it better fits how error descriptions appear elsewhere.

I would suggest something like "Unable to submit grade" and "Unable to fetch grade". Note the use of "Sentence case", rather than "Title Case" and the wording follows the form "Unable to <operation>" or "Failed to <operation>" rather than "<operation> error".

robertknight (commented):
I think that including the entire response from the server in the "Error details" panel shown to the user is actually a useful thing to do in this context since it may have information that will be of use to IT admins at the institution. Including the request URL and data, but not the authentication token, might also be useful.

Since it is useful to do this in more contexts rather than just the specific scenario where `status != "success"` perhaps populating the `details` field could be done by the `LTIOutcomesAPIError` class if `details` is not set.

The frontend part of this change looks separate to me from the backend part and there is less discussion that may happen around it, so it might help to split that out into a separate PR which can be merged first.

PR #2537, opened 2021-04-02
Render more details with LTI outcome errors

LMS007:
In order to help pinpoint moodle grading errors, it will help to relay more information to the client so that it may be captured in a screen shot at the time the error occurs. This helps prevent losing metadata when the logs expire in paper trail after too much time.

--------

This is just the client change in this PR
https://github.com/hypothesis/lms/pull/2536

Screen Shot 2021-03-31 at 4 37 17 PM

### DIFF
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -24,6 +24,7 @@ const GRADE_MULTIPLIER = 10;
 
 /**
  * @typedef {import('../config').StudentInfo} StudentInfo
+ * @typedef {import('./ErrorDisplay').ErrorLike} ErrorLike
  */
 
 /**
@@ -32,7 +33,7 @@ const GRADE_MULTIPLIER = 10;
  * and error status of that fetch request.
  *
  * @param {StudentInfo|null} student
- * @param {(value: string) => void} setFetchGradeError
+ * @param {(value: ErrorLike) => void} setFetchGradeError
  */
 const useFetchGrade = (student, setFetchGradeError) => {
   const {
@@ -57,7 +58,7 @@ const useFetchGrade = (student, setFetchGradeError) => {
             setGrade(scaleGrade(currentScore, GRADE_MULTIPLIER));
           }
         } catch (e) {
-          setFetchGradeError(e.errorMessage ? e.errorMessage : 'Unknown error');
+          setFetchGradeError(e);
         } finally {
           setGradeLoading(false);
         }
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -220,19 +225,19 @@ export default function SubmitGradeForm({ student }) {
       {!!submitGradeError && (
         <ErrorDialog
           title="Submit Grade Error"
```
robertknight, line 227: Can you update this title to "Unable to submit grade" while you're improving the error info.

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -220,19 +225,19 @@ export default function SubmitGradeForm({ student }) {
...
           onCancel={() => {
-            setSubmitGradeError('');
+            setSubmitGradeError(null);
           }}
           cancelLabel="Close"
         />
       )}
       {!!fetchGradeError && (
         <ErrorDialog
           title="Fetch Grade Error"
```
robertknight, line 237: Can you update this title to "Unable to fetch grade" while you're improving the error info.

robertknight (changes requested):
Hi Kyle,

Since there are no backend changes in this PR to actually set the `details` property of the error returned to the client, I assume the screenshot is not actually from this branch and a separate PR will be needed to set that property?

The changes to `SubmitGradeForm` look fine, but the fact that the tests didn't require any changes shows that they aren't checking what gets passed to the `ErrorDialog` at all. Can you update them to check that the error is passed through appropriately.

I also suggested that it would be a good idea to update the title of the error dialogs while we're improving the content.

LMS007:
> t it would be a good idea to update the title of the error dialogs while we're improving the content.

Yep, that makes sense, I think you mentioned that in the previous PR, and I admit I forgot! will do.

robertknight (approved):
The revised tests work, but I think it would be better to check that the error object is forwarded directly, rather than just testing that something with the same `errorMessage` and `details` properties is passed to `ErrorDialog`. Assuming you're on board with that change, I'm good with merging this.

I tested this locally with the following diff:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 2377b87c..a6fee0c6 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -130,6 +130,7 @@ class LTIOutcomesClient:
         except LTIOutcomesAPIError as err:
             err.response = response
             log.exception("An error occurred in the LTI Outcomes response")
+            err.details = f"Response data:\n\n{response.text}"
             raise
 
     @classmethod
@@ -146,6 +147,9 @@ class LTIOutcomesClient:
         except KeyError as err:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
+        # TESTING
+        status = "failure"
+
         if status != "success":
             raise LTIOutcomesAPIError("LTI outcome request failed")
```

The error details were then correctly displayed in the frontend when trying to view a user's grade in Moodle.

PR #2539, opened 2021-04-02
Improve test task

LMS007:
- Use new commander options for test task exactly how the client is configured.
  - Add --grep, --browser, --no-browser, and --watch options to `test` task
  - Remove `test-watch` task (replaced with `test --watch`)

-----

more or less matches the task from the client and frontend-shared


robertknight (approved):
👍. It might be helpful to extract some of the common Karma-related code into a shared package in future. This is fine in the interim.

PR #2565, opened 2021-04-09
Pass the requestConfigFromFrame options to h-vialib

seanh:
Depends on https://github.com/hypothesis/h-vialib/pull/23/

### DIFF
--- a/lms/views/helpers/_via.py
+++ b/lms/views/helpers/_via.py
@@ -20,6 +20,13 @@ def via_url(request, document_url, content_type=None):
     """
     return ViaClient(
         service_url=request.registry.settings["via_url"],
-        host_url=request.host_url,
         secret=request.registry.settings["via_secret"],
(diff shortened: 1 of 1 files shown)


lms/views/helpers/_via.py
```diff
@@ -20,6 +20,13 @@ def via_url(request, document_url, content_type=None):
...
         service_url=request.registry.settings["via_url"],
-        host_url=request.host_url,
         secret=request.registry.settings["via_secret"],
-    ).url_for(document_url, content_type, blocked_for="lms")
+    ).url_for(
+        document_url,
+        content_type,
+        options={
+            "via.client.requestConfigFromFrame.origin": request.host_url,
+            "via.client.requestConfigFromFrame.ancestorLevel": 2,
```
seanh, line 29: These two are totally LMS-specific options: created for LMS, only used by LMS. Non-LMS uses specifically need these options _not_ to be present. So having them live in LMS seems right

PR #2574, opened 2021-04-13
Update to pyramid 2.0

marcospri:
Followed the same conventions as the same topic PR on checkmate.

Keep in mind that there will another issue about integrating google oauth here as well so it will be even closer to checkmate's implementation after that.

Remove all ACLs and replaced them by permission assigned directly on the policies.


lms/config/__init__.py
```diff
@@ -95,13 +95,7 @@ def configure(settings):
...
     settings.update(env_settings)
 
-    config = Configurator(
-        settings=settings, root_factory="lms.resources.DefaultResource"
-    )
-
-    # Security policies
-    authz_policy = ACLAuthorizationPolicy()
-    config.set_authorization_policy(authz_policy)
+    config = Configurator(settings=settings)
```
marcospri, line 98: I don't think the default resource had any effect, but please consider it.

lms/security.py
```diff
@@ -0,0 +1,176 @@
+import base64
```
marcospri, line 1: This is a relativly large file but I think still make sense to keep all here unless it keeps growing, we could later split int lms/security/policies etc

tests/unit/lms/security_test.py
```diff
@@ -0,0 +1,486 @@
...
+        AuthTktCookieHelper.return_value.remember.assert_called_once()
+
+    def test_forget(self, pyramid_request, AuthTktCookieHelper):
+        TktSecurityPolicy("TEST_LMS_SECRET").forget(pyramid_request)
+
+        AuthTktCookieHelper.return_value.forget.assert_called_once()
+
+    @pytest.fixture
+    def AuthTktCookieHelper(self, patch):
+        return patch("lms.security.AuthTktCookieHelper")
```
seanh, line 117: We should probably try to write the tests without mocking this, though I know the original tests already mocked it

tests/unit/lms/security_test.py
```diff
@@ -0,0 +1,486 @@
...
+                factories.LTIUser(
+                    user_id="Sam:Smith",
+                    oauth_consumer_key="Hypothesisf301584250a2dece14f021ab8424018a",
+                ),
+                "U2FtOlNtaXRo:Hypothesisf301584250a2dece14f021ab8424018a",
+            ),
+        ],
+    )
+    def test_it(self, lti_user, expected_userid):
+        assert authenticated_userid(lti_user) == expected_userid
```
seanh, line 335: I wonder if this belongs in `TestLTISecurityPolicy` (as a test for `LTISecurityPolicy.authenticated_userid()`) since I think that's the only thing that calls the `authenticated_userid()` function, and I think the function is really a private helper

seanh (changes requested):

lms/security.py
```diff
@@ -0,0 +1,176 @@
...
+    permissions: List[str]
+
+
+class Permissions(Enum):
+    LTI_LAUNCH_ASSIGNMENT = "lti_launch_assignment"
+    REPORTS_VIEW = "report_viewers"
+    API = "api"
+
+
+def authenticated_userid(lti_user):
```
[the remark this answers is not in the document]
marcospri: I left it as it own `_` function on the module

tests/functional/views/reports_test.py
```diff
@@ -0,0 +1,31 @@
...
+    def test_it_doesnt_allow_any_user(self, app):
+        cookie_token = AuthTicket(
+            "TEST_LMS_SECRET", "some-user", "0.0.0.0", hashalg="sha512"
+        ).cookie_value()
+
+        app.set_cookie("auth_tkt", cookie_token)
+
+        response = app.post(
+            "/reports",
+            status=200,
```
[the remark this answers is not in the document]
marcospri: It's return a 200 currently, moved this bits into its own PR

https://github.com/hypothesis/lms/pull/2579

tests/functional/views/reports_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+class TestReports:
+    def test_it_allows_whitelisted_user(self, app):
+        cookie_token = AuthTicket(
+            "TEST_LMS_SECRET", "report_viewer", "0.0.0.0", hashalg="sha512"
+        ).cookie_value()
+
+        app.set_cookie("auth_tkt", cookie_token)
+
+        response = app.post(
```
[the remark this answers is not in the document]
marcospri: It should, move these into its own PR https://github.com/hypothesis/lms/pull/2579

tests/unit/lms/security_test.py
marcospri: You mean setting a cookie on the DummyRequest to get the actual value from the helper?

seanh (approved):

PR #2575, opened 2021-04-13
Add imsx_description to LTIOutcomesAPIError details

LMS007:
When users see a grading error, its helpful to capture to error details at the same time so we can more quickly discover causes for these errors. Passing the response data to the LTIOutcomesAPIError will allow it to propagate to the client and render in the error dialog.

### DIFF
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -147,7 +147,19 @@ class LTIOutcomesClient:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
         if status != "success":
-            raise LTIOutcomesAPIError("LTI outcome request failed")
+            description = None
+            try:
+                # Look for an imsx_description to pass along to the client, but it is possible this field
+                # may not exist.
+                description = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
+                    "imsx_description"
+                ]
+            except KeyError as err:
+                pass
+
+            raise LTIOutcomesAPIError(
+                explanation="LTI outcome request failed", details=description
+            )
 
         return body
 
--- a/tests/unit/lms/services/lti_outcomes_test.py
+++ b/tests/unit/lms/services/lti_outcomes_test.py
@@ -127,24 +127,46 @@ class TestLTIOutcomesClient:
             content_type="application/json",
             priority=1,
         )
-        with pytest.raises(LTIOutcomesAPIError):
+        with pytest.raises(
+            LTIOutcomesAPIError,
+            match="Unable to parse XML response from LTI Outcomes service",
+        ):
             svc.read_result(self.GRADING_ID)
 
     def test_requests_fail_if_no_status(self, svc, respond_with):
         respond_with(include_status=False)
-        with pytest.raises(LTIOutcomesAPIError):
+        with pytest.raises(LTIOutcomesAPIError, match="Malformed LTI outcome response"):
+            svc.read_result(self.GRADING_ID)
+
+    def test_requests_fail_if_response_is_malformed(self, svc, respond_with):
+        respond_with(malformed=True)
+
+        with pytest.raises(LTIOutcomesAPIError, match="Malformed LTI outcome response"):
             svc.read_result(self.GRADING_ID)
 
     def test_requests_fail_if_status_is_not_success(self, svc, respond_with):
         respond_with(status_code="failure")
 
-        with pytest.raises(LTIOutcomesAPIError):
+        # LTI outcome request failed
+        with pytest.raises(
+            LTIOutcomesAPIError,
+            match="<imsx_description>An error occurred.</imsx_description>",
+        ):
+            svc.read_result(self.GRADING_ID)
+
+    def test_requests_fail_and_no_description_returned(self, svc, respond_with):
+        respond_with(status_code="failure", include_description=False)
+
+        # imsx_description is missing
+        with pytest.raises(LTIOutcomesAPIError, match="LTI outcome request failed"):
             svc.read_result(self.GRADING_ID)
 
     def test_it_gracefully_handles_RequestException(self, requests, svc):
         requests.post.side_effect = RequestException
 
-        with pytest.raises(LTIOutcomesAPIError):
+        with pytest.raises(
+            LTIOutcomesAPIError, match="Error calling LTI Outcomes service"
+        ):
             svc.read_result(self.GRADING_ID)
 
     @classmethod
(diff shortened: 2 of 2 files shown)


robertknight (changes requested):
Can you revise this to make sure that we gracefully handle the `imsx_description` field not being present (eg. by defaulting to an empty string), since it is not critical for the rest of the code.

Can you please also update the tests to check the `explanation` and `details` fields are set as expected.

lms/services/lti_outcomes.py
```diff
@@ -139,6 +139,9 @@ def _get_body(cls, data):
         try:
             body = data["imsx_POXEnvelopeResponse"]["imsx_POXBody"]
             header = data["imsx_POXEnvelopeResponse"]["imsx_POXHeader"]
+            description = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
+                "imsx_description"
+            ]
```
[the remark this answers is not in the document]
LMS007: i agree

lms/services/lti_outcomes.py
```diff
@@ -147,7 +150,9 @@ def _get_body(cls, data):
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
```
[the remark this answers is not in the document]
LMS007: roger

lms/services/lti_outcomes.py
```diff
@@ -146,8 +146,20 @@ def _get_body(cls, data):
...
+                # may not exist.
+                description = data["imsx_POXEnvelopeResponse"]["imsx_POXHeader"][
+                    "imsx_POXResponseHeaderInfo"
+                ]["imsx_statusInfo"]["imsx_description"]
+            except KeyError as err:
+                pass
+            finally:
+                raise LTIOutcomesAPIError(
+                    explanation="LTI outcome request failed", details=description
+                )
```
robertknight, line 162: Can you please make sure the new behavior is covered by a test.

robertknight (changes requested):
Hi Kyle - The tests are currently failing due to the change in the `if status` branch. We also need a test for the new behavior here to make sure that works and for the case where the description field is missing.

lms/services/lti_outcomes.py
```diff
@@ -146,8 +146,20 @@ def _get_body(cls, data):
         except KeyError as err:
             raise LTIOutcomesAPIError("Malformed LTI outcome response") from err
 
-        if status != "success":
-            raise LTIOutcomesAPIError("LTI outcome request failed")
+        if status == "success":
```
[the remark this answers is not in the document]
LMS007: That was me testing and forgetting to replace it! thats why all these tests failed, doh!

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -195,6 +210,19 @@ def make_response(cls, score, include_score, include_status, status_code):
             }
         )
 
+    @classmethod
+    def make_malformed_response(cls):
+        return xmltodict.unparse(
+            {
+                # "imsx_POXEnvelopeResponse" is the expected key
+                "imsx_POXEnvelopeRequest": {
```
LMS007, line 218: Does not much matter what this value is as long as its not imsx_POXEnvelopeResponse, but this is actually what blackbaud returns (incorrectly), so I figured I would just match its brokenness

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -135,12 +135,24 @@ def test_requests_fail_if_no_status(self, svc, respond_with):
         with pytest.raises(LTIOutcomesAPIError):
             svc.read_result(self.GRADING_ID)
 
+    def test_requests_fail_if_response_is_malformed(self, svc, respond_with):
```
LMS007, line 138: Added this missing test too

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -135,12 +135,24 @@ def test_requests_fail_if_no_status(self, svc, respond_with):
...
+            svc.read_result(self.GRADING_ID)
+
     def test_requests_fail_if_status_is_not_success(self, svc, respond_with):
         respond_with(status_code="failure")
 
         with pytest.raises(LTIOutcomesAPIError):
             svc.read_result(self.GRADING_ID)
 
+    def test_requests_fail_and_no_description_returned(self, svc, respond_with):
+        respond_with(status_code="failure", include_description=False)
```
LMS007, line 151: This test ensures we consume (hide) the KeyError from missing description

tests/unit/lms/services/lti_outcomes_test.py
robertknight: Do you have an example of a complete response from Blackbaud? Does it always return an unexpected response structure or only in certain scenarios?

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -135,12 +135,24 @@ def test_requests_fail_if_no_status(self, svc, respond_with):
...
+        with pytest.raises(LTIOutcomesAPIError):
+            svc.read_result(self.GRADING_ID)
+
     def test_requests_fail_if_status_is_not_success(self, svc, respond_with):
         respond_with(status_code="failure")
 
         with pytest.raises(LTIOutcomesAPIError):
             svc.read_result(self.GRADING_ID)
 
+    def test_requests_fail_and_no_description_returned(self, svc, respond_with):
```
robertknight, line 150: Can you add a check to the above test to verify that the expected description is set on the response when a description is present.

robertknight (commented):
The test changes generally look good. I requested adding a check to one of the existing tests to verify that the description is set on the `LTIOutcomesError` when expected. I also had a query about the malformed response from Blackbaud that you mentioned. Is there an issue that describes when it happens and do we know what the full response looks like in that case?

LMS007:
> The test changes generally look good. I requested adding a check to one of the existing tests to verify that the description is set on the `LTIOutcomesError` when expected. I also had a query about the malformed response from Blackbaud that you mentioned. Is there an issue that d

i have not combed through all the papertrail logs, but a quick search check suggests that this is consistently broken.  I wrote up some notes on the original ticket. 

https://github.com/hypothesis/support/issues/124

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -167,17 +189,20 @@ def assert_sent_header_ok(cls):
...
         if include_status:
             header_info["imsx_statusInfo"] = {
                 "imsx_codeMajor": status_code,
                 "imsx_severity": "status",
-                "imsx_description": "Result read",
                 "imsx_messageRefIdentifier": "999999123",
                 "imsx_operationRefIdentifier": "readResult",
             }
+            if include_description:
+                header_info["imsx_statusInfo"]["imsx_description"] = "Result read"
```
robertknight, line 205: The "Result read" description implies success, although we're actually interested in this for error cases. It doesn't matter for correctness, but it might help for documentation to use a different description when the status is an error.

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -195,6 +220,19 @@ def make_response(cls, score, include_score, include_status, status_code):
             }
         )
 
+    @classmethod
+    def make_malformed_response(cls):
+        return xmltodict.unparse(
+            {
+                # "imsx_POXEnvelopeResponse" is the expected key
```
robertknight, line 227: Can you add a note here that this malformed response is "inspired" by Blackbaud. On that note, has anyone gotten in touch with Blackbaud to alert them to the issue? Have we looked at working around it yet?

robertknight (approved):
Changes LGTM. I've verified that the error details are presented as expected to the user:

Moodle grade error details

tests/unit/lms/services/lti_outcomes_test.py
LMS007: yep!

PR #2595, opened 2021-04-23
Add course group set endpoint to canvas API

marcospri:
# Testing

(there's probably a more straightforward method to test this)

### Enable groups on the application_instances

```make sql```

An run something simlar to this, use a `where id = X` or similar if you rather not enable it in all instances locally.

```update application_instances set settings = jsonb_set(settings, '{canvas,groups_enabled}', 'true');```

### Remove existing oauth tokens

```make sql``` (or same shell as before)

Delete all oauth tokens locally, again, use `delete where ...` if you rather target an specific one.
```truncate  oauth2_token;``` 

### Re-authorize you user

head to `https://hypothesis.instructure.com/courses/125/assignments/1575` open any assigment and click "authorize"

### Copy the authorization key

I found the easiest to copy it from any other API call on the browerer, for example after opening the assigment, open the network tab on the brower and copy the Authorization header from the call to `http://localhost:8001/api/canvas/sync`

### Call the new endpoint

Using the auth key, use a curl command or similar to call the new endpoint

```
curl 'http://localhost:8001/api/canvas/courses/125/group_categories' \
  -H 'Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJlbWFpbCI6ImVuZytjYW52YXN0ZWFjaGVyQGh5cG90aGVzLmlzIiwib2F1dGhfY29uc3VtZXJfa2V5IjoiSHlwb3RoZXNpc2IzYmUwYjMzZDBiNWU0YjFjZjNhYWYwNGUwZTE4MTlhIiwidXNlcl9pZCI6Ijk2OWZjMjk0ZWUyNmUzZjFkMGM5NzE0YWYzMzUxZGM0ZWFjZmQ2ZGYiLCJ0b29sX2NvbnN1bWVyX2luc3RhbmNlX2d1aWQiOiJWQ1N5OERBS0NGSVFZYUl0cFNHeVhoT1hJbGs1QTZOV2R1VlZHMXUzOmNhbnZhcy1sbXMiLCJkaXNwbGF5X25hbWUiOiJIeXBvdGhlc2lzIDEwMSBUZWFjaGVyIiwicm9sZXMiOiJJbnN0cnVjdG9yIiwiZXhwIjoxNjE5MjU4MjU0fQ.JJ2Q8xQT70MDCjDTRPQOIOZQgl-yl4NYJQ97f-5K3Y0' \
  -H 'Content-Type: application/json; charset=UTF-8' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  --compressed
```

results in:

```[{"id": 121, "name": "Test group set"}, {"id": 122, "name": "Empty group set"}, {"id": 125, "name": "PoC"}]```

### DIFF
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -255,6 +255,19 @@ class CanvasAPIClient:
 
         public_url = fields.Str(required=True)
 
+    def course_group_categories(self, course_id):
+        return self._client.send(
+            "GET",
+            f"courses/{course_id}/group_categories",
+            schema=self._ListGroupCategories,
+        )
+
+    class _ListGroupCategories(RequestsResponseSchema):
+        many = True
+
+        id = fields.Integer(required=True)
+        name = fields.Str(required=True)
+
     @classmethod
     def _ensure_sections_unique(cls, sections):
         """
--- a/lms/views/api/canvas/groups.py
+++ b/lms/views/api/canvas/groups.py
@@ -1,0 +1,25 @@
+"""Proxy API views for group-related Canvas API endpoints."""
+from pyramid.view import view_config, view_defaults
+
+from lms.security import Permissions
+
+
+@view_defaults(permission=Permissions.API, renderer="json")
+class GroupsAPIViews:
+    def __init__(self, request):
+        self.request = request
+        self.canvas_api_client = request.find_service(name="canvas_api_client")
+
+    @view_config(
+        request_method="GET", route_name="canvas_api.courses.group_categories.list"
+    )
+    def course_group_categories(self):
+        """
+        Return the list of group categories in the given course.
+
+        :raise lms.services.CanvasAPIError: if the Canvas API request fails.
+            This exception is caught and handled by an exception view.
+        """
+        return self.canvas_api_client.course_group_categories(
+            self.request.matchdict["course_id"]
+        )
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -64,6 +64,11 @@ def includeme(config):
         "canvas_api.files.via_url",
         "/api/canvas/courses/{course_id}/files/{file_id}/via_url",
     )
+    config.add_route(
+        "canvas_api.courses.group_categories.list",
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: lms/services/canvas_api/client.py [hunk 1]
```diff
@@ -255,6 +255,24 @@ class _PublicURLSchema(RequestsResponseSchema):
...
+    class _ListGroupCategories(RequestsResponseSchema):
+        many = True
+
+        id = fields.Integer(required=True)
+
+        name = fields.Str(required=True)
+        description = fields.String(default=None, allow_none=True)
+
+        role = fields.Str(required=False, allow_none=True)
+        content_type = fields.Str(required=False, allow_none=True)
```
seanh, line 274: Do we need `description`, `role` and `content_type` here? I don't think we're using them so I think we can just omit them from the schema and they'll be ignored. This works because we configure Marshmallow to ignore unknown fields: https://github.com/hypothesis/lms/blob/d220299bd95b8f4d0262a45c169e35ab97948a1e/lms/validation/_base.py#L45-L46

lms/views/api/canvas/groups.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, request):
+        self.request = request
+        self.canvas_api_client = request.find_service(name="canvas_api_client")
+
+    @view_config(
+        request_method="GET", route_name="canvas_api.courses.group_categories.list"
+    )
+    def course_group_categories(self):
+        """
+        Return the list of files in the given course.
```
seanh, line 18:
```suggestion
        Return the list of group categories in the given course.
```

seanh (approved):
Looks good.

One thing I'm wondering is whether it's possible for a course to have so many group sets that they don't fit in a single API response? What will happen in that case? Will our Canvas API client's pagination automatically kick in?

lms/services/canvas_api/client.py
marcospri: Yup, those were useless at the moment. Removed.

PR #2596, opened 2021-04-26
HGroup now a DB model

marcospri:
Includes https://github.com/hypothesis/lms/pull/2607 (will rebase after merging that )

Build fails on CI (I think will be fixed on https://github.com/hypothesis/lms/pull/2616)

Just open some assignments on canvas, it will create a group per section and one per course on the DB

```
postgres=# select * from h_group;
-[ RECORD 1 ]---------+--------------------------------------------
created_at            | 2021-04-29 08:12:51.003277
updated_at            | 
id                    | 6
_name                 | Developer Test Course with Sections Enabled
authority_provided_id | 653e9620a656a954c684942f6443fa3e3410c03a
type                  | course_group
-[ RECORD 2 ]---------+--------------------------------------------
created_at            | 2021-04-29 08:12:51.164646
updated_at            | 
id                    | 7
_name                 | Section 1
authority_provided_id | d2871e5e1c36b58f207d9ef5b22dcb7829f52e61
type                  | section_group
-[ RECORD 3 ]---------+--------------------------------------------
created_at            | 2021-04-29 08:12:51.164646
updated_at            | 
id                    | 8
_name                 | Section 2
authority_provided_id | f453c6124e4ead4471896f26e896c3e5bbf0bb36
type                  | section_group
-[ RECORD 4 ]---------+--------------------------------------------
created_at            | 2021-04-29 08:12:51.164646
updated_at            | 
id                    | 9
_name                 | Section 3
authority_provided_id | 83e64baad215f63057387910a1b41debc716a150
type                  | section_group
```


lms/models/h_group.py
```diff
@@ -1,61 +1,30 @@
...
+    UNIQUE_CONSTRAINT_NAME = "unique_h_group"
 
-class HGroup(NamedTuple):
-    name: str
-    authority_provided_id: str
-    type: str
+    __tablename__ = "h_group"
+    __table_args__ = (
+        sa.UniqueConstraint(
+            "name", "authority_provided_id", name=UNIQUE_CONSTRAINT_NAME
```
seanh, line 12: I think you just omit the `name` here. There's a naming convention in `db/__init__.py` that gets applied automatically: https://github.com/hypothesis/lms/blob/d220299bd95b8f4d0262a45c169e35ab97948a1e/lms/db/__init__.py#L79

lms/models/h_group.py
```diff
@@ -1,61 +1,30 @@
...
+    name = sa.Column(
+        sa.String(),
+        nullable=False,
+    )
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
+
+    type = sa.Column(
+        sa.String(),
+        nullable=False,
+    )
```
seanh, line 27: I think we should add `created` and `updated` columns

lms/services/h_group.py
```diff
@@ -0,0 +1,79 @@
...
+            .on_conflict_do_nothing(
+                constraint=HGroup.UNIQUE_CONSTRAINT_NAME,
+            )
+        )
+        self._db.execute(upsert_statement)
+        return (
+            self._db.query(HGroup)
+            .filter_by(name=name, authority_provided_id=authority_provided_id)
+            .one()
+        )
```
seanh, line 28: I wonder if we can use the ORM here? It'd be something like:

```python
h_group = self._db.query(HGroup).filter_by(authority_provided_id=...)

if not h_group:
    h_group = HGroup(authority_provided_id=..., type=...)
    self._db.add(h_group)

# Update any fields that might have changed.
h_group.name = name

return h_group
```

Note the distinction between fields that're used to identify the group for looking it up in the DB (`authority_provided_id`) versus fields that we just create/update (`name`)

seanh (changes requested):

lms/services/h_group.py
seanh: FYI I'm not sure this method implements an upset as written. I think this is a create-if-not-exists. For an upsert you still might need to update certain fields on an pre-existing row in the DB. `name` is the only field that can change I think.

marcospri: Switched to ORM only :+1:

lms/models/h_group.py
```diff
@@ -1,61 +1,24 @@
...
+import sqlalchemy as sa
 
-from lms.models._hashed_id import hashed_id
+from lms.db import BASE, TimestampedModelMixin
 
-MAX_GROUP_NAME_LENGTH = 25
 
+class HGroup(TimestampedModelMixin, BASE):
+    __tablename__ = "h_group"
+    __table_args__ = (sa.UniqueConstraint("name", "authority_provided_id"),)
```
marcospri, line 8: Is this right @seanh ? You mention something in slack identifying columns vs "updateable" ones. I'm using `authority_provided_id` as the only unique one and allowing name and type to change

lms/services/h_group.py
```diff
@@ -0,0 +1,79 @@
...
+        :param section_id: A section id for a section group
+        """
+        return self.upsert(
+            self._name(section_name),
+            hashed_id(tool_consumer_instance_guid, context_id, section_id),
+            "section_group",
+        )
+
+    @staticmethod
+    def _name(name):
```
marcospri, line 64: This currently stores the truncated names in the DB, an alternative would be to have _name as the DB backed attribute, storing there the full name and having `.name` as a property doing the truncation on self._name

lms/db/__init__.py
```diff
@@ -88,6 +89,13 @@ def __repr__(self):
 SESSION = sessionmaker()
 
 
+class TimestampedModelMixin:
+    created_at = sqlalchemy.Column(sqlalchemy.DateTime, default=func.now())
+    updated_at = sqlalchemy.Column(
+        sqlalchemy.DateTime(timezone=True), onupdate=func.now()
+    )
```
seanh, line 96: A mixin class seems like a good idea for this :+1: We have a similar thing in h: https://github.com/hypothesis/h/blob/5101ff6045f76b64f9358d366e68e4cbf6c64a3b/h/db/mixins.py#L8-L21 And in Checkmate (though not for timestamps): https://github.com/hypothesis/checkmate/blob/main/checkmate/models/db/mixins.py

Might be nicer in `models/_mixins.py` rather than in `db`.

Are there any existing models for LMS that could be changed to use this mixin?

I'm not sure what the correct / best way to do timestamps in the DB is:

* Should they have `timezone=True`?

* Why does `updated_at` have `timezone=True` but not `created_at`?

* I think `created_at` should probably have `server_default=sa.func.now()` rather than `default`, and `updated_at` should also have `server_default=sa.func.now()`, that way the user of a model class doesn't have to do anything about `created_at` and `updated_at` they just get set to the right values automatically.

  (Generally I think you're supposed to use `server_default` rather than `default` in SQLAlchemy unless you have a reason not to. `server_default` actually sets the default in the Postgres schema so direct SQL that doesn't go through the models will also use the default.)

* Should it be `sa.func.now()` or `sa.func.utcnow()`?

* They should probably have `nullable=False`, no reason these should ever be null

lms/migrations/versions/c349a812ca82_add_h_group_table.py
```diff
@@ -0,0 +1,36 @@
...
+        sa.PrimaryKeyConstraint("id", name=op.f("pk__h_group")),
+        sa.UniqueConstraint("authority_provided_id", name=op.f("uq__h_group__name")),
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table("h_group")
+    # ### end Alembic commands ###
```
seanh, line 36: It's super nice to have migration auto-generation working.

We should probably delete the generated comments.

This migration will need to be split out into a separate PR because it needs to be deployed and run before we can deploy the Python code that uses the new model

lms/models/h_group.py
```diff
@@ -1,59 +1,36 @@
...
-    type: str
+class HGroup(TimestampedModelMixin, BASE):
+    __tablename__ = "h_group"
+    __table_args__ = (sa.UniqueConstraint("authority_provided_id"),)
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    # Full name of the group in the LMS
+    _name = sa.Column(
+        sa.String(),
```
seanh, line 16: We should probably use `UnicodeText` everywhere rather than `String`?

lms/models/h_group.py
```diff
@@ -1,59 +1,36 @@
...
 
-class HGroup(NamedTuple):
-    name: str
-    authority_provided_id: str
-    type: str
+class HGroup(TimestampedModelMixin, BASE):
+    __tablename__ = "h_group"
+    __table_args__ = (sa.UniqueConstraint("authority_provided_id"),)
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
```
seanh, line 12: :+1: I think having a surrogate unique key is a good idea

lms/models/h_group.py
```diff
@@ -1,59 +1,36 @@
...
+
+    # Full name of the group in the LMS
+    _name = sa.Column(
+        sa.String(),
+        nullable=False,
+    )
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
```
seanh, line 19:
```suggestion
    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False, unique=True)
```

lms/models/h_group.py
```diff
@@ -1,59 +1,36 @@
...
 MAX_GROUP_NAME_LENGTH = 25
 
 
-class HGroup(NamedTuple):
-    name: str
-    authority_provided_id: str
-    type: str
+class HGroup(TimestampedModelMixin, BASE):
+    __tablename__ = "h_group"
+    __table_args__ = (sa.UniqueConstraint("authority_provided_id"),)
```
seanh, line 10:
```suggestion
```

I think you can just add `unique=True` to the `authority_provided_id = sa.Column(...)` below

lms/models/h_group.py
```diff
@@ -1,59 +1,36 @@
...
-        return HGroup(
-            cls._name(section_name),
-            hashed_id(tool_consumer_instance_guid, context_id, section_id),
-            type="section_group",
-        )
-
-    @classmethod
-    def _name(cls, name):
+    @property
+    def name(self):
```
seanh, line 30: I think there might be something not quite right about the `HGroup` model.

We're storing the name of the LMS course, section or group that the `HGroup` represents in an `h_group._name` column on the DB, then we have this `name` property that generates the `HGroup`'s actual name from the course, section or group name.

I think we may need to start recording LMS courses, Canvas sections and Canvas groups in the DB first. Courses are actually already done (https://github.com/hypothesis/lms/blob/74bdda7ae901cdd75139574ca5d035c1a7912474/lms/models/course.py) but Canvas sections and Canvas groups aren't.

Each course, Canvas section and Canvas group would be related to exactly one `HGroup` by a  non-nullable foreign key.

Then instead of `_name`, the `name()` property would generate the h group name by looking up the name of this `HGroup`'s related LTI course, Canvas section or Canvas group.

I'm not sure how exactly to implement this foreign key relation though. We want each group to be related to _exactly one_ LTI course, Canvas section or Canvas group in a `1:1` relation. But courses, sections and groups will probably be in three separate DB tables. I'm sure it can be done somehow

seanh (changes requested):
I think we might need to put our heads together and think about the modeling here. I'm not sure what the right solution is, but I've got a feeling that LTI courses, Canvas sections and Canvas groups should all be recorded in the DB, and then the `h_group` table should be related to these tables. See comment

lms/db/__init__.py
marcospri: I'll do the Mixin in a different PR :+1: 
https://github.com/hypothesis/lms/pull/2622

lms/models/h_group.py
marcospri: Yep initially needed to name the index to use the `on_conflict` but after updating to use the ORM that's not needed

PR #2604, opened 2021-04-26
Replace Button in Dialog.js

LMS007:
Depends on https://github.com/hypothesis/lms/pull/2594

This is the first go at replacing Button with LabeledButton. This is more or less a straight conversion as simple as it gets.

### Commit

Replace Button in Dialog.js …

- Include @hypothesis/frontend-shared/styles in css main file
- Replace Button with LabeledButton

-----------

Only the cancel button has changed.

**Before**
Screen Shot 2021-04-26 at 3 17 51 PM

**After**
Screen Shot 2021-04-26 at 3 14 39 PM

### DIFF
--- a/lms/static/styles/components/_Dialog.scss
+++ b/lms/static/styles/components/_Dialog.scss
@@ -1,3 +1,4 @@
+@use '@hypothesis/frontend-shared/styles/components/buttons';
 @use "@hypothesis/frontend-shared/styles/mixins/focus";
 
 @use "../mixins";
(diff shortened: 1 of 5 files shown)


lms/static/styles/components/_Dialog.scss
```diff
@@ -68,8 +69,7 @@
   justify-content: flex-end;
   margin-top: 10px;
 
-  .Button {
-    @include focus.outline-on-keyboard-focus;
+  * {
```
lyzadanger, line 72: Can you help me understand this universal selector here? It is making me nervous :)

LMS007:
Updated scss to make this a direct generic selector... much less dangerous!

Screen Shot 2021-04-29 at 8 50 46 AM

lyzadanger (approved):
GTG

PR #2607, opened 2021-04-27
Alembic config changes

marcospri:
- Change any DB backed model on `models/*.py`
- `hev alembic revision --autogenerate -m "change"`
- Review autogenerated migration

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -145,7 +145,7 @@ def _session(request):  # pragma: no cover
     # So as a precaution add our own callback here to make sure db sessions are
     # always closed.
     @request.add_finished_callback
-    def close_the_sqlalchemy_session(_request):  # pylint: disable=unused-variable
+    def close_the_sqlalchemy_session(_request):
         connections = (
             session.transaction._connections  # pylint:disable=protected-access
         )
--- a/lms/migrations/env.py
+++ b/lms/migrations/env.py
@@ -4,6 +4,11 @@ from logging.config import fileConfig
 from alembic import context
 from sqlalchemy import MetaData, engine_from_config, pool
 
+# Import all model modules here in order to populate the metadata
+# for 'autogenerate' support
+from lms import models  # noqa
+from lms.db import BASE
+
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
 config = context.config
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -30,7 +30,6 @@ services: python
 .PHONY: db
 db: args?=upgrade head
 db: python
-	@tox -qqe dev --run-command 'initdb conf/development.ini'  # See setup.py for what initdb is.
```
marcospri, line 33: I don't know if this will break any existing workflows but at the moment if trying to auto generate a migration for a new model 

- initdb will silently create the table
- alembic will see the new table already in the DB and create an empty migration.

It might be that make db it's trying to do many things? both "init the db" and "any alembic command"?

seanh: I think `make db` needs to continue running `initdb` because the purpose of `make db` is to give team members a foolproof way to fix their DB when it might not be initialized or might not be on the latest migration version. Without it some members of the frontend team become helplessly confused when they haven't initialized their DB for some reason or when we've merged a DB migration. See this thread for example: <https://hypothes-is.slack.com/archives/CBN3DGW02/p1588329090111400?thread_ts=1588283479.110200&cid=CBN3DGW02>. Similar things had happened again and again.

But I wonder if `make db` should just hardcode the `alembic upgrade head` command and there should be a separate `make alembic` or `make migrate` for passing custom commands into alembic?

seanh (changes requested):

Makefile
```diff
@@ -4,6 +4,7 @@ help:
 	@echo 'make services          Run the services that `make dev` requires'
 	@echo "                       (Postgres) in Docker"
 	@echo 'make db                Upgrade the DB schema to the latest version'
+	@echo 'make alembic           Run alembic (DB migrations) commands. Default to `history`
```
marcospri, line 7: Added the new alembic target, left the other one unchanged in case someone has muscle memory to use that for alembic stuff.

Makefile
marcospri: Kept make db unchanged and added `make alembic`

Makefile
```diff
@@ -4,6 +4,7 @@ help:
 	@echo 'make services          Run the services that `make dev` requires'
 	@echo "                       (Postgres) in Docker"
 	@echo 'make db                Upgrade the DB schema to the latest version'
+	@echo 'make alembic           Run alembic (DB migrations) commands.
```
seanh, line 7:
```suggestion
	@echo 'make alembic           Run alembic (DB migrations) commands'
```

These docstrings don't have periods on the end, but mostly you need to add the `'` to fix a syntax error

seanh (approved):
Can you copy `make alembic` to h and Checkmate as well? It's something we'll want to move into `hdev` of course (along with `make db`) but since we only have three projects that have a DB (LMS, h and Checkmate) it seems quick enough to copy `make alembic` into all three for now. Or alternatively feel free to move `make alembic` into `hdev` now instead.

Once that's done can you also update these SO4T posts?

[How do I create and test a database migration for LMS or Checkmate?](https://stackoverflow.com/c/hypothesis/questions/317/)

[How do I create and test a database migration for `h`?](https://stackoverflow.com/c/hypothesis/questions/28/29#29)

Note that you can edit other people's posts on SO4T and when a question has an answer it's usually best to treat it like a wiki and update the existing answer rather than add a new answer

seanh (approved):

Issue #2590, opened 2021-04-21
Upgrade LMS to Python 3.8


jon-betts:
For: https://github.com/hypothesis/product-backlog/issues/1199

Our deps:

 * [x] `h-api` - https://github.com/hypothesis/h-api/issues/6
 
Some LMS specific deps:

 * [x] `pycrypto` - https://github.com/hypothesis/lms/pull/2616
 * [x] `PyLTI` - https://github.com/hypothesis/lms/pull/2610
 * [x] `xmltodict` - 3.7 - Locally tests fine out to 3.9
 * [x] `requests_oauthlib` - 3.7 - Locally tests fine out to 3.9 (worth a PR for this one?)
 * [x] `behave` - 3.6 - This one is kind of dicey, but our tests still pass... 

![app_lms](https://user-images.githubusercontent.com/7131143/115562154-2002dd80-a2ae-11eb-8f4c-b008b84f4ec3.png)

jon-betts:
I've run the `behave` tests locally, and they don't pass for basically anything... but they don't pass the same for 3.6 and 3.7. I think what we might be able to do here is move the behave tests alone to 3.7? Ultimately in our case, if our tests still pass under 3.8 it's all gravy.

There is support for 3.9 in master last updated 4 months ago, but there hasn't been a release since 2018.

jon-betts:
Done!

PR #2608, opened 2021-04-27
Upgrade python to Python 3.8

jon-betts:
For: https://github.com/hypothesis/lms/issues/2590


.github/workflows/continuous-integration.yml
```diff
@@ -1,66 +1,108 @@
...
+      python_headline_version: ${{ steps.list-python-versions.outputs.python_headline_version }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
LMS007, line 17: I take it 3.8.9 was not an option in github actions?

LMS007 (commented):
I did run `brew upgrade`  and I installed python 3.8.9 in pyenv on my mac, but I was not able to start the app with 3.8.9. I'm not really sure what went wrong because the errors are numerous.  I have saved dev-2.log locally in case its helpful to look at? but I can't make much sense of it.

.github/workflows/continuous-integration.yml
jon-betts: It's an option, but we don't need to run on an older version of Python for this stage. So it's one less thing to try and fix up.

jon-betts:
Hi @LMS007, yeah it might be a good idea to wait until we've worked through this before diving in, but we do need to get this working for you. I'm going to be honest, I'm not really sure what `brew` is...

This is currently failing on some coverage by the looks of it, but this is very much enough for today.

LMS007:
> Hi @LMS007, yeah it might be a good idea to wait until we've worked through this before diving in, but we do need to get this working for you. I'm going to be honest, I'm not really sure what `brew` is...
> 
> This is currently failing on some coverage by the looks of it, but this is very much enough for today.

Fair enough, brew is more or less a a way to install these packages on mac and make them somewhat consistent for everyone. I think of it a a wrapper around chaos, but it does mean that things do work a bit differently on mac. it may just mean I need upgrade or change some installed packages as we hit errors.

marcospri (approved):

PR #2610, opened 2021-04-27
Replace `PyLTI` with `oauthlib`

jon-betts:
For: https://github.com/hypothesis/lms/issues/2590

PyLTI is kind of unmaintained and the tests are very broken. This replaces it with `oauthlib` which we already use.

## Review notes

* The way `oauthlib` works is you have to sub-class objects and fill in loads of methods
* The major difficulty in this was previously our code and tests were pretty confused about whether this was a POST or a GET. It's always a POST. This mostly just confused me.
* There's also some weirdness where the URL we get isn't necessarily the one that was signed against. We have to strip any GET params prior to checking the signature
* Due to the important thing being the request body, and the tests previously written around the `params` exclusively the tests were quite heavily rewritten

## Testing notes

Nothing really to do but try it against a number of LTI tool consumers. I've tried Canvas and Moodle.

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -10,7 +10,6 @@ from lms.services.exceptions import (
     LTILaunchVerificationError,
     LTIOAuthError,
     LTIOutcomesAPIError,
-    NoConsumerKey,
     NoOAuth2Token,
     ServiceError,
 )
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -12,10 +12,6 @@ class LTILaunchVerificationError(ServiceError):
     """
 
 
-class NoConsumerKey(LTILaunchVerificationError):
-    """Raised when a launch request has no ``oauth_consumer_key`` parameter."""
-
-
 class ConsumerKeyError(LTILaunchVerificationError):
     """Raised when a given ``consumer_key`` doesn't exist in the DB."""
 
--- a/lms/services/launch_verifier.py
+++ b/lms/services/launch_verifier.py
@@ -1,13 +1,8 @@
 """LTI launch request verifier service."""
-import pylti.common
+from oauthlib.oauth1 import RequestValidator, SignatureOnlyEndpoint
 
 from lms import models
-from lms.services import (
-    ConsumerKeyError,
-    LTILaunchVerificationError,
-    LTIOAuthError,
-    NoConsumerKey,
-)
+from lms.services import ConsumerKeyError, LTILaunchVerificationError, LTIOAuthError
 
 __all__ = ["LaunchVerifier"]
 
@@ -17,6 +12,10 @@ class LaunchVerifier:
 
     def __init__(self, _context, request):
         self._request = request
+        self._oauth1_endpoint = SignatureOnlyEndpoint(
+            OAuthRequestValidator(db_session=self._request.db)
+        )
+
         self._request_verified = False
         self._exception = None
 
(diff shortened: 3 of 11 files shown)


lms/services/__init__.py
```diff
@@ -10,7 +10,6 @@
     LTILaunchVerificationError,
     LTIOAuthError,
     LTIOutcomesAPIError,
-    NoConsumerKey,
```
jon-betts, line 13: This isn't something we retrieve any more, `oauthlib` does it for us. This means we never have an opportunity to raise this.

lms/services/exceptions.py
```diff
@@ -12,10 +12,6 @@ class LTILaunchVerificationError(ServiceError):
     """
 
 
-class NoConsumerKey(LTILaunchVerificationError):
-    """Raised when a launch request has no ``oauth_consumer_key`` parameter."""
```
jon-betts, line 16: This isn't something we retrieve any more, `oauthlib` does it for us. This means we never have an opportunity to raise this.

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,78 @@ def verify(self):
...
     def _verify(self):
-        try:
-            consumer_key = self._request.params["oauth_consumer_key"]
-        except KeyError as err:
-            raise NoConsumerKey() from err
+        is_valid, _request = self._oauth1_endpoint.validate_request(
+            uri=self._request.url,
+            http_method=self._request.method,
+            body=self._request.body,
+            headers=dict(self._request.headers),
```
jon-betts, line 56: The behavior of this method is quite different from the library we used before and is the source of all the change in the tests. Previously this accepted `uri`, `params` and `body`, now it only accepts `uri` and `body` and the params are expected to be in the `uri`.

This causes quite a lot of annoyance, as our tests relied on being able to modify the params separately quite a bit. More on this later...

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,78 @@ def verify(self):
...
+    def __init__(self, db_session):
+        super().__init__()
+        self.db_session = db_session
+
+    def check_client_key(self, client_key):
+        """Check that the client key only contains safe characters."""
+
+        # Just a style check for chars and length etc... we don't care about
+        # the format, as long as our lookup later works
+        return True
```
jon-betts, line 80: It's mandatory to implement lots of methods on this, or the process crashes, but I don't actually think we want to enforce much about the style of the key.

`oauthlib`s design is very weird to me as well, as we implement all these different cases, but ultimately it just crushes them all down to `False` if anything fails. I'm not sure what the advantage in distinguishing is.

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,78 @@ def verify(self):
...
+        timestamp,
+        nonce,
+        request,
+        request_token=None,
+        access_token=None,
+    ):
+        """Validate that the nonce has not been used before."""
+
+        # We don't appear to care about the nonce value
+
```
jon-betts, line 103:
```suggestion
        # We don't appear to care about the nonce value
```

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,78 @@ def verify(self):
...
+        return True
```
jon-betts, line 104:
```suggestion
        # We don't appear to care about the nonce value
        return True
```

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,77 @@ def verify(self):
...
+
+    def validate_client_key(self, client_key, request):
+        """Validate that supplied client key is registered and valid."""
+
+        # This is slightly incorrect, but we are just going to say the client
+        # key is fine! Even though this might not be true. This is because
+        # we would have to look up the DB to find out, and we _have_ to do
+        # that in the next step `get_client_secret()`
+
+        return True
```
jon-betts, line 113: We are being slightly naughty here and checking this in the wrong place. This function is intended to see if a `client_key` is real after the previous checks that it's the right format.

The way we would do that is to look it up in the DB. The next step is to get a client secret for the key, which we do by... looking it up in the DB. I wasn't sure if SQLAlchemy would cache the request or not, so I thought we'd just do both in the next step.

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,77 @@ def verify(self):
...
+    def get_client_secret(self, client_key, request):
+        """Retrieve the client secret associated with the client key."""
 
         application_instance = models.ApplicationInstance.get_by_consumer_key(
-            self._request.db, consumer_key
+            self.db_session, client_key
         )
 
         if not application_instance:
             raise ConsumerKeyError()
```
jon-betts, line 123: If we want any feedback about what's going wrong in this process we have to raise exceptions or `oauthlib` will eat all the feedback.

requirements/bddtests.txt
```diff
@@ -42,11 +42,6 @@ h-pyramid-sentry==1.2.3
     # via -r requirements/requirements.txt
 h-vialib==1.0.15
     # via -r requirements/requirements.txt
-httplib2==0.19.0
-    # via
-    #   -r requirements/requirements.txt
-    #   oauth2
-    #   pylti
```
jon-betts, line 49: Some nice eliminations of dependencies of dependencies. Our whole set of deps has gotten quite a bit smaller throughout the Python 3.6 upgrade process.

tests/unit/lms/services/launch_verifier_test.py
```diff
@@ -139,28 +152,19 @@ def test_it_caches_a_failed_verification_result(self, launch_verifier, pylti):
...
-        pyramid_request.url += "?" + urlencode(params)
-
-        launch_verifier = LaunchVerifier(mock.sentinel.context, pyramid_request)
+        # This test is probably not relevant any more as we don't use PyLTI
+        pyramid_request.params[
+            "url"
+        ] = "https://en.wikipedia.org/wiki/G%C3%B6reme_National_Park"
+
+        oauth1.sign()
+
```
jon-betts, line 167: I don't know if this test makes any sense anymore, as it looks to be covering a specific `PyLTI` bug.

tests/unit/lms/services/launch_verifier_test.py
```diff
@@ -169,48 +173,71 @@ def launch_verifier(self, pyramid_request):
...
-        # Add the OAuth 1 params (version, nonce, timestamp, ...)
-        oauthlib_client = oauthlib.oauth1.Client(
-            pyramid_request.params["oauth_consumer_key"]
-        )
-        oauthlib_request = oauthlib.common.Request(
-            pyramid_request.url, pyramid_request.method
-        )
-        pyramid_request.params = dict(
-            oauthlib_client.get_oauth_params(oauthlib_request)
-        )
```
jon-betts, line 188: We were doing very similar things twice in this and the `sign` method, and they both need to be different now, so I've separated all the OAuth1 stuff into a class we use as a fixture to get this stuff done.

tests/unit/lms/services/launch_verifier_test.py
```diff
@@ -169,48 +173,71 @@ def launch_verifier(self, pyramid_request):
...
+        launch_verifier._oauth1_endpoint = oauth1_endpoint
 
-def sign(pyramid_request):
-    """Add or replace pyramid_request's OAuth 1 signature param."""
-    oauthlib_client = oauthlib.oauth1.Client(
-        pyramid_request.params["oauth_consumer_key"], "TEST_SECRET"
-    )
-    oauthlib_request = oauthlib.common.Request(
-        pyramid_request.url, pyramid_request.method, body=pyramid_request.params
-    )
```
jon-betts, line 208: :point_up: This right here is some odd behavior. The method in this case is `GET` we then take the params and pass them as a body, which `oauthlib` interprets as form fields and signs them.

It just so happens the way you sign the form fields and get params is identical, so it all sort of works out...

However, our new method requires the params to actually be on the URL we received _or_ for the request to be a proper POST with form fields.

tests/unit/lms/services/launch_verifier_test.py
```diff
@@ -169,48 +173,71 @@ def launch_verifier(self, pyramid_request):
...
+        parsed_url = parsed_url._replace(query=None)
+
+        return oauthlib.common.Request(
+            # This all seems very wrong, for actual requests we expect the URL to
+            # contain the query string, here we remove it
+            parsed_url.geturl(),
+            self.pyramid_request.method,
+            # And set the query string as the body... in any case it works
+            body=self.pyramid_request.params,
+        )
```
jon-betts, line 225: We're keeping the old weird way of doing the signing but we will need to put the results back on the URL later...

tests/unit/lms/services/launch_verifier_test.py
```diff
@@ -169,48 +173,71 @@ def launch_verifier(self, pyramid_request):
...
+        self.pyramid_request.params = dict(
+            self.client.get_oauth_params(self._get_request())
+        )
+
+    def sign(self):
+        self.pyramid_request.params[
+            "oauth_signature"
+        ] = self.client.get_oauth_signature(self._get_request())
+
+        self.pyramid_request.set_url_from_params()
```
jon-betts, line 237: Here we add the params to the URL.

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,91 @@ def verify(self):
...
+        """Check that the client key only contains safe characters."""
+
+        # Just a style check for chars and length etc... we don't care about
+        # the format, as long as our lookup later works
+        return True
+
+    def check_nonce(self, nonce):
+        """Check that the nonce only contains only safe characters."""
+
+        # Just a style check for chars and length etc... we don't care about
```
marcospri, line 99: Is this comment right? Maybe the docstring one is enought.

lms/services/launch_verifier.py
```diff
@@ -50,44 +49,91 @@ def verify(self):
...
+
+    def check_client_key(self, client_key):
+        """Check that the client key only contains safe characters."""
+
+        # Just a style check for chars and length etc... we don't care about
+        # the format, as long as our lookup later works
+        return True
+
+    def check_nonce(self, nonce):
+        """Check that the nonce only contains only safe characters."""
```
marcospri, line 97: Only x 2

marcospri (approved):
You can tell when I only have comments about the code comments that my level of understanding of LTI stuff it's not there yet but I can see it works on Canvas and understanding what `OAuthRequestValidator` does it's clear :+1:

lms/services/launch_verifier.py
jon-betts: Yep. I'll remove these

PR #2615, opened 2021-04-28
Application Instance admin pages

marcospri:
Test with

make devdata
make dev

Navigate to http://localhost:8001/admin and complete the oauth flow

Find/Edit ApplicationIinstances


lms/services/application_instance.py
```diff
@@ -0,0 +1,54 @@
+from typing import Optional
+
+from sqlalchemy import or_
+
+from lms.models import ApplicationInstance
+
+
+class ApplicationInstanceService:
```
marcospri, line 8: There's `ApplicationInstanceGetter` service but that seems to be very focused on the application instance currently in use so I don't think these fit there

lms/services/application_instance.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, db):
+        self._db = db
+
+    def find(self, query: str) -> Optional[ApplicationInstance]:
```
marcospri, line 12: I not sure if this belong here, should this type of query go into the model itself? Then will need boilerplate code here on the service to call them 1 service method per 1 model one.

lms/services/application_instance.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+    # pylint: disable=invalid-name
+    def get(self, pk: int) -> Optional[ApplicationInstance]:
+        """Query an ApplicationInstance by primary key."""
+
+        return self._db.query(ApplicationInstance).get(pk)
+
+    @staticmethod
+    def update_settings(
```
marcospri, line 32: I did this because I didn't wan the caller needing to know where to store the stuff "canvas" -> "group_enabled" vs just passing a boolean here

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{% block title %}LMS Admin{% endblock %}</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma@0.9.2/css/bulma.min.css">
```
marcospri, line 7: I googled a bit and this one didn't seem to assume you were going to use some java script and just makes moving things around easier.

lms/templates/admin/installation.html.jinja2
```diff
@@ -0,0 +1,93 @@
+{% extends "lms:templates/admin/base.html.jinja2" %}
+{% block header %}
+Installation {{ installation.id }} / {{installation.consumer_key}}
```
marcospri, line 3: I'm calling this installation here, Application Instance it's not a great name but just "Consumer key" might be a bit reductive.

Happy to rename back to something else thought.

lms/templates/admin/index.html.jinja2
```diff
@@ -0,0 +1,16 @@
+{% extends "lms:templates/admin/base.html.jinja2" %}
+
+
+{% block content %}
```
marcospri, line 4: There's a loner input field here on the index page :shrug:

lms/templates/admin/base.html.jinja2
seanh: Can we host this file ourselves (in `lms/static/styles/`) instead of linking to `jsdelivr.net`?

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
+from pyramid.httpexceptions import HTTPFound
+from pyramid.view import forbidden_view_config, view_config, view_defaults
+
+from lms.security import Permissions
+
+
+@view_defaults(request_method="GET")
```
seanh, line 7:
```suggestion
@view_defaults(request_method="GET", permission=Permissions.ADMIN)
```

Requires moving the forbidden view (see other comment)

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @view_config(
+        route_name="admin.installations",
+        permission=Permissions.ADMIN,
+        request_method="POST",
+    )
+    def find_installation(self):
+        if "query" not in self.request.params:
+            self.request.session.flash("Missing mandatory 'query'", "errors")
+            return HTTPFound(location=self.request.route_url("admin.installations"))
```
seanh, line 37: We're probably okay to just `raise httpexceptions.HTTPBadRequest()` here (without the flash message), in practice only our own form ever `POST`'s to this view and it does always include `query`

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+        installation = self.application_instance.find(self.request.params["query"])
+        if installation:
+            return HTTPFound(
+                location=self.request.route_url(
+                    "admin.installation", id=installation.id
+                ),
+            )
+        self.request.session.flash(
+            f'No installation found for {self.request.params["query"]}', "errors"
+        )
```
seanh, line 48: :+1:

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+        return HTTPFound(location=self.request.route_url("admin.installations"))
+
+    @view_config(
+        route_name="admin.installation",
+        renderer="lms:templates/admin/installation.html.jinja2",
+        permission=Permissions.ADMIN,
+    )
+    def show_installation(self):
+
```
seanh, line 57:
```suggestion
```

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+        installation = self.application_instance.get(self.request.matchdict["id"])
```
seanh, line 58: This crashes if the `id` in the URL isn't an int or if it's an int that doesn't exist in the DB.

Also, I think this should look up installations by `consumer_key` rather than by `id`, as `id` is supposed to be an internal database ID

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
...
+          href="{{ request.route_url("admin.installations") }}" >
+        Installations
+      </a>
+
+    <div class="navbar-end">
+      <div class="navbar-item">
+        <div class="buttons">
+          <a class="button is-light" href="{{ request.route_url("pyramid_googleauth.logout") }}" >
+            Log out
+          </a>
```
seanh, line 31: The log out button seems to disappear if the window is too thin :)

lms/templates/admin/index.html.jinja2
seanh: I think maybe the `/admin/` page should just redirect to `/admin/installations/`, then it doesn't need a template. Even when we have more than one admin page in future, `/admin/` and still redirect to `/admin/installations/` and if people want a different page they can use the navbar at the top to navigate to where they want to go

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+        self.application_instance.update_settings(
+            installation,
+            canvas_sections_enabled=sections_enabled,
+            canvas_groups_enabled=groups_enabled,
+        )
+        self.request.session.flash(
+            f"Updated installation {installation.id}", "messages"
+        )
+
+        return HTTPFound(location=self.request.route_url("admin.installations"))
```
seanh, line 87: It'd be nice if this redirected to `route_url(admin.installation`, id=installation.id)` instead so I could see the checkboxes reloaded with their new values

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
...
+  </head>
+  <body>
+
+  {% block base_navbar %}
+  <nav class="navbar" role="navigation" aria-label="main navigation">
+  <div class="navbar-brand">
+    <a class="navbar-item" href="{{ request.route_url("admin.index") }}">
+      <p class="is-size-4">ADMIN</p>
+    </a>
+  </div>
```
seanh, line 17:
```suggestion
```

If the `/admin/` page just redirected to `/admin/installations/` (as suggested in another comment) then this link could just be removed

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
...
+
+  <div id="navbarBasicExample" class="navbar-menu">
```
seanh, line 19:
```suggestion
  <div class="navbar-menu">
```

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
...
+    <div class="navbar-start" style="flex-grow: 1; justify-content: center;">
```
seanh, line 20: It seems odd that we need our own CSS here?

lms/templates/admin/installation.html.jinja2
seanh: Hmm. Might be worth some wider discussion in Slack?

* "Application instance" is clearly terrible, although it is what's already used in the DB and source and known by support staff
* "Installation" is better but I think it's actually incorrect, since a single `ApplicationInstance` can be installed multiple times in multiple LMS's, or I think even multiple times in the _same_ LMS, or not installed anywhere at all. For example changing the settings of an `ApplicationInstance` will change the settings for _all_ installations that use that consumer key
* Personally I think "consumer key" is correct as that's what it actually is: a consumer key and related fields (shared secret and other stuff)
* Maybe there's another name that better captures that this isn't _just_ the consumer key but also other related fields and settings associated with this consumer key?

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @forbidden_view_config()
+    def logged_out(self):
+        return HTTPFound(location=self.request.route_url("pyramid_googleauth.login"))
```
seanh, line 91: I'm not sure this forbidden view is doing what you might think. Since no arguments are passed to `@forbidden_view_config()` (except the `request_method="GET"` from the `@view_defaults`) this forbidden view is being registered for _all_ `GET` requests that fail a permissions check, for any route. So this will cause all permissions failures to redirect to the admin login page.

The forbidden view is also _not_ being registered for the `update_installion()` form submission because it's a `POST`.

I think you need a forbidden view that's registered for the admin pages only. You need to find the right view config predicates: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html#predicate-arguments The easiest option is probably `@forbidden_view_config(path_info="/admin/*")`. I know people don't like to reference parts of the URL path in the code but I think it's the only way to get the correct behaviour.

I think the forbidden view should be moved out of the `AdminViews` class as well, which will allow the `permission=Permissions.ADMIN` to be moved into the `@view_defaults`

lms/views/admin.py
```diff
@@ -0,0 +1,91 @@
...
+from pyramid.view import forbidden_view_config, view_config, view_defaults
+
+from lms.security import Permissions
+
+
+@view_defaults(request_method="GET")
+class AdminViews:
+    def __init__(self, request):
+        self.request = request
+        self.application_instance = request.find_service(name="application_instance")
```
seanh, line 11:
```suggestion
        self.application_instance_service = request.find_service(name="application_instance")
```

or `_svc_

seanh (changes requested):
Looking great! A few requests (see comments for more details):

* I think we can simplify the admin page by having `/admin/` just redirect to `/admin/installations/`
* I think the `update_installation()` view should redirect back to `/admin/installation/<ID>` after updating
* I think everything should work based on the consumer key alone, and we should hide `ApplicationInstance.id` as an internal DB column (it's only really meant for use in foreign key relationships and the like)
* Found a couple of crashes (see comments)
* I don't think the forbidden view is right (see comments)

lms/templates/admin/index.html.jinja2
marcospri: Yeah, I'll do that, current index.html it's pointless

lms/views/admin.py
marcospri: Yep that was completely broken, `path_info` seems the most straightforward fix :+1:

lms/services/application_instance.py
marcospri: Now they share the code for the `.get`

lms/templates/admin/base.html.jinja2
marcospri: Responsive design :stuck_out_tongue:

lms/templates/admin/installation.html.jinja2
marcospri: Renamed to ApplicationInstance

lms/templates/admin/base.html.jinja2
```diff
@@ -0,0 +1,85 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{% block title %}LMS Admin{% endblock %}</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma@0.9.2/css/bulma.min.css">
+  </head>
+  <body>
+
+  {% block base_navbar %}
+  <nav class="navbar" role="navigation" aria-label="main navigation">
+  <div class="navbar-brand">
+    <a class="navbar-item" href="{{ request.route_url("admin.index") }}">
```
marcospri, line 14: I think this link and the "instances" one apparently taking you to the same place it's an effect of having only one option currently. 

I rather leave as is, will make adding a second option more obvious.

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -0,0 +1,76 @@
...
+        ai = db_session.query(ApplicationInstance).get(test_application_instance.id)
+        assert (
+            # pylint: disable=compare-to-zero
+            ai.settings.get(*setting_location.split("."))
+            is False
+        )
+
+    @pytest.fixture
+    def ai_service(self, pyramid_request):
+        return factory(mock.sentinel.context, pyramid_request)
```
seanh, line 63: We usually call this fixture just `svc` in service tests

seanh (changes requested):
This is looking great :+1:

I have one big question which is whether the form submission needs CSRF protection? See https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/security.html#csrf-protection

I'm seeing a bunch of lint warning on this branch (which I'm not seeing on master) but for some reason `make lint` isn't failing. They're actually all `useless-suppression` warnings, so maybe that doesn't trigger PyLint to fail. We should configure PyLint to fail on all warnings:

```
************* Module lms.security
lms/security.py:151:0: I0021: Useless suppression of 'unused-argument' (useless-suppression)
************* Module lms.db.__init__
lms/db/__init__.py:148:0: I0021: Useless suppression of 'unused-variable' (useless-suppression)
************* Module lms.views.admin
lms/views/admin.py:31:0: I0021: Useless suppression of 'no-self-use' (useless-suppression)
************* Module tests.unit.conftest
tests/unit/conftest.py:144:0: I0021: Useless suppression of 'unused-variable' (useless-suppression)
```

Other than that mostly a bunch of small suggestions about the tests

lms/views/admin.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, request):
+        self.request = request
+        self.application_instance_service = request.find_service(
+            name="application_instance"
+        )
+
+    @view_config(
+        route_name="admin.index",
+    )  # pylint: disable=no-self-use
```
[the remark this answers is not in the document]
marcospri: Yeah, that had a `permission=` arg before but never went back and change it to fit one line :+1:

tests/unit/lms/views/admin_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+        with pytest.raises(HTTPBadRequest):
+            AdminViews(pyramid_request).find_instance()
+
+    def test_find_instance_not_found(
+        self, pyramid_request, application_instance_service
+    ):
+        application_instance_service.get.return_value = None
+        pyramid_request.params["query"] = "some-value"
+        response = AdminViews(pyramid_request).find_instance()
```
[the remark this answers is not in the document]
marcospri: Added the space but if modifying `pyramid_request` seems clearer to construct AdminViews here

tests/unit/lms/views/admin_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+        assert pyramid_request.session.peek_flash("errors")
+        assert response == temporary_redirect_to(
+            pyramid_request.route_url("admin.instances")
+        )
+
+    def test_find_instance_found(self, pyramid_request, application_instance_service):
+        application_instance_service.get.return_value = Mock(consumer_key="XXX")
```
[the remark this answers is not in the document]
marcospri: Added on the fixture :+1:

marcospri:
> 
> I have one big question which is whether the form submission needs CSRF protection? See https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/security.html#csrf-protection
> 

Added this on the last commit, to test it just go into any of the templates and change the value or remove `<input type="hidden" name="csrf_token"`

tests/unit/conftest.py
```diff
@@ -171,6 +173,22 @@ def ai_getter(pyramid_config):
...
 
+@pytest.fixture
+def application_instance_service(pyramid_config):
+    application_instance_service = mock.create_autospec(
+        ApplicationInstanceService, instance=True, spec_set=True
+    )
+
+    application_instance_service.get.return_value = mock.create_autospec(
+        ApplicationInstance, consumer_key=mock.sentinel.consumer_key
+    )
```
seanh, line 184: You need to pass `instance=True` otherwise `create_autospec` creates a mock of the _class_ rather than a mock of an instance of the class, and `spec_set=True` so that tests can't _set_ attributes on the mock that don't exist on the real object

seanh (approved):
One small thing about the tests, otherwise looks good

PR #2616, opened 2021-04-28
Use `pycryptodomex` instead of `pycryptodome`

jon-betts:
Born of: https://github.com/hypothesis/lms/pull/2611 for: https://github.com/hypothesis/lms/issues/2590

The 'x' varient changes the name of the installed package so it's no longer a direct replacement for `pycrypto`. This means it can't conflict with it and both can be installed. It _also_ means that tools which inspect the virtual env can see it's not installed and don't get confused between the two.

I _think_ this fixes some bugs I've been having locally when switching branches where it says "AES" cannot be found. I think this is also affecting CI.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -2,8 +2,8 @@ import secrets
 from datetime import datetime
 
 import sqlalchemy as sa
-from Crypto import Random
-from Crypto.Cipher import AES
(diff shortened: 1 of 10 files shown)


marcospri (approved):
Interesting that they provide the 'x' version, :+1:

PR #2617, opened 2021-04-28
Update StudentSelector icons

LMS007:
This is continuation of https://github.com/hypothesis/lms/pull/2594 but broken up into mostly atomic commits with specific messages.

The final code result is almost identical to the previous PR with one notable difference: I did not include the `cancel`svg as that is not related to the immediate changes here. _That icon will need be be added later when the dialogs are converted to IconButtons._


lyzadanger (approved):
Thanks for working through all this and getting those commits tight. You have said there are no user-visible changes here, so this should be GTG. 👍🏻

PR #2622, opened 2021-04-29
Timestamped model mixin

marcospri:
Using the existing definitions of created and & updated

Running alembic revision --autogenerate results in an empty migration.

There's more models that have `created` but not updated, does it make sense to have CreatedMixin?

### DIFF
--- a/lms/models/_mixins.py
+++ b/lms/models/_mixins.py
@@ -1,0 +1,15 @@
+import sqlalchemy as sa
+
+
+class CreatedUpdatedMixin:
+    created = sa.Column(
+        sa.DateTime(),
+        server_default=sa.func.now(),
+        nullable=False,
+    )
+    updated = sa.Column(
+        sa.DateTime(),
+        server_default=sa.func.now(),
+        onupdate=sa.func.now(),
+        nullable=False,
+    )
--- a/lms/models/__init__.py
+++ b/lms/models/__init__.py
@@ -1,3 +1,4 @@
+from lms.models._mixins import CreatedUpdatedMixin
 from lms.models.application_instance import ApplicationInstance
 from lms.models.application_settings import ApplicationSettings
 from lms.models.course import Course
--- a/lms/models/grading_info.py
+++ b/lms/models/grading_info.py
@@ -1,13 +1,12 @@
-import datetime
-
 import sqlalchemy as sa
 
 from lms.db import BASE
+from lms.models import CreatedUpdatedMixin
 
 __all__ = ["GradingInfo"]
 
 
-class GradingInfo(BASE):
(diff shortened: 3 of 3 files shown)


lms/models/_mixins.py
```diff
@@ -0,0 +1,19 @@
...
+        sa.DateTime(),
+        server_default=sa.func.now(),
+        default=datetime.datetime.utcnow,
+        nullable=False,
+    )
+    updated = sa.Column(
+        sa.DateTime(),
+        default=datetime.datetime.utcnow,
+        server_default=sa.func.now(),
+        onupdate=sa.func.now(),
```
marcospri, line 17: Only change over the existing one is onupdate using func.now() vs datetime

seanh (approved):

lms/models/_mixins.py
jon-betts: There's some prior art for this in `h`, but I don't think we should take that too seriously: https://github.com/hypothesis/h/blob/master/h/db/mixins.py

One thing to note is the mixed use of Python / Postgres functions for specifying the date. From what I can see (https://docs.sqlalchemy.org/en/14/core/defaults.html#client-invoked-sql-expressions) you can use the server side version for the `default` and `onupdate`. Without knowing the exact outcome, it seems better to have an identical mechanism for setting them all if we can.

This insulates us from any drift between Postgres and pPthon

lms/models/_mixins.py
```diff
@@ -0,0 +1,15 @@
+import sqlalchemy as sa
+
+
+class TimestampedModelMixin:
```
jon-betts, line 4: Super nit pick, but this isn't necessarily a model / the fact it's a model comes from elsewhere. So I think this is `TimestampedMixin` or `WithTimestamps` or something.

marcospri: :+1:  for something like `CreatedUpdatedMixin`

marcospri: @jon-betts We could do `CreatedMixin` and `UpdatedMixin` but I don't we need those independently as we might for Accessed, what do you think?

jon-betts: I think we should follow the usage. If created and updated always go together, then the class should too. I can't think of a good reason why we wouldn't want these both if we ever have either. If we do find that happens, it's easy to split up.

The `accessed` does seem like a different kettle of fish, as we have to code specifically to support it wherever it's used, so it makes sense to me for that not to be a part of it.

jon-betts (approved):
Yep, I think this looks good. Although I'll fully admit to be an "expert" on this in the way anyone is after 15 minutes poking about the SQLAlchemy docs.

PR #2636, opened 2021-05-04
Frontend changes for Blackboard Files POC

seanh:
This is a proof-of-concept showing the frontend changes that're needed
for the Blackboard Files feature. This is an updated version of
https://github.com/hypothesis/lms/pull/2100

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -43,6 +43,7 @@ export default function ContentSelector({
   const {
     api: { authToken },
     filePicker: {
+      blackboard: { enabled: blackboardEnabled, listFiles: blackboardListFilesApi },
       canvas: { enabled: canvasEnabled, listFiles: listFilesApi },
       google: {
         clientId: googleClientId,
@@ -115,6 +116,16 @@ export default function ContentSelector({
         />
       );
       break;
+    case 'blackboardLmsFile':
+      dialog = (
+        <LMSFilePicker
+          authToken={authToken}
+          listFilesApi={blackboardListFilesApi}
+          onCancel={cancelDialog}
+          onSelectFile={selectLMSFile}
+        />
+      );
+      break;
     default:
       dialog = null;
   }
@@ -172,6 +183,15 @@ export default function ContentSelector({
               data-testid="lms-file-button"
             >
               Select PDF from Canvas
+            </LabeledButton>
+          )}
+          {blackboardEnabled && (
+            <LabeledButton
+              onClick={() => selectDialog('blackboardLmsFile')}
+              variant="primary"
+              data-testid="lms-file-button"
+            >
+              Select PDF from Blackboard
             </LabeledButton>
           )}
           {googlePicker && (
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -93,6 +93,13 @@ export default function FilePickerApp({ onSubmit }) {
             formFields={formFields}
           />
         )}
+        {content && (
+          <input
+            name="document_url"
+            type="hidden"
+            value={content.file.id}
+          />
+        )}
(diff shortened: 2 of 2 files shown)


lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -93,6 +93,13 @@ export default function FilePickerApp({ onSubmit }) {
             formFields={formFields}
           />
         )}
+        {content && (
+          <input
+            name="document_url"
+            type="hidden"
+            value={content.file.id}
+          />
```
robertknight, line 101: The `document_uri` field, and other hidden form fields are now handled by the `FilePickerFormFields` component.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -174,6 +185,15 @@ export default function ContentSelector({
               Select PDF from Canvas
             </LabeledButton>
           )}
+          {blackboardEnabled && (
+            <LabeledButton
+              onClick={() => selectDialog('blackboardLmsFile')}
+              variant="primary"
+              data-testid="lms-file-button"
+            >
+              Select PDF from Blackboard
```
robertknight, line 194: This works for a proof of concept. For a production implementation I think we'll want to use the same button / dialog type for Canvas and Blackboard files but change the `listFilesApi` prop and button label according to the LMS being used.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -115,6 +116,16 @@ export default function ContentSelector({
         />
       );
       break;
+    case 'blackboardLmsFile':
```
robertknight, line 119: Since it is not possible to be using Canvas files and Blackboard files at the same time, we might be able to combine the `lmsFile` and `blackboardLmsFile` cases here and just switch the `listFilesApi` prop and button caption depending on the LMS.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -43,6 +43,7 @@ export default function ContentSelector({
   const {
     api: { authToken },
     filePicker: {
+      blackboard: { enabled: blackboardEnabled, listFiles: blackboardListFilesApi },
```
robertknight, line 46: Are there any differences between the Blackboard files and Canvas files APIs that the frontend needs to be aware of? Going further, could we consider merging them at some point?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: Related question: How does the backend know that the `document_url` field contains a file ID as opposed to a URL?

robertknight (commented):
I added some notes for whoever works on making a production version of this. I also had a couple of questions about the `document_url` field value and the Blackboard vs Canvas files APIs exposed to the LMS frontend.

From a user-facing perspective, are there any required changes needed for a shippable feature that are not included in this PoC?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
seanh: I think the frontend may have to do this for us.

When the `document_url` is a URL it needs to look like `https://example.com/foo/`.

When it's a Blackboard file ID it needs to look like `blackboard://content-resource/xyz123`.

I think it's the frontend that nows which type of document the user picked so the frontend is gonna need to format this `blackboard://content-resource/<ID>` string.

See https://stackoverflow.com/c/hypothesis/questions/356

seanh: Actually, maybe the frontend doesn't need to actually format the `blackboard://` URL. That URL format can remain a backend-internal thing. But the frontend does need to indicate to the backend that this is a Blackboard file ID rather than a URL or something else. For example by submitting a `blackboard_file_id` field (rather than `document_url`)

robertknight:
@LMS007 - Can this issue be closed now? I think https://github.com/hypothesis/lms/pull/2771 includes everything from this PR that is required.

PR #2637, opened 2021-05-04
Set default color in body

LMS007:
A few basic sass changes across the app.
- Fix the default color of the app to match the client with grey7.  This means the dialog text (previously black) is very slightly lighter and the grading bar text (prevently grey6) is very slightly darker. 

- Bake in a default cursor pointer effect for all buttons and button-like elements. This also matches the client's behavoir. Note that some of these sass changes will be wiped out when we convert the buttons over. Currently, the new buttons don't have a pointer styling so there is a lack of affordance for users when hovering over buttons--this fixes that.

### DIFF
--- a/lms/static/styles/_variables.scss
+++ b/lms/static/styles/_variables.scss
@@ -9,6 +9,9 @@ $grey-5: #7a7a7a;
 $grey-6: #3f3f3f;
 $grey-7: #202020;
 
+// Text colors
+$color-text: $grey-7;
+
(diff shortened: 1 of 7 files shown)


lyzadanger (approved):
This is nicely-structured, thank you. It is nice to bubble up the cursor stuff instead of sprinkling it throughout various component styles.

I trust that you've reviewed the effect of changing the default text color...seems like it should be fine at first glance.

PR #2641, opened 2021-05-05
Remove NoOAuth2Token from exception views

seanh:
`NoOAuth2Token` is never raised by any view, so it never gets caught by
the exception view.

`NoOAuth2Token` is only ever raised by `OAuth2TokenService.get()`

https://github.com/hypothesis/lms/blob/486f56a6324cd100735737f90787a9c8b8641b70/lms/services/oauth2_token.py#L58

`OAuth2TokenService` is only ever used by `CanvasAPIClient`:

https://github.com/hypothesis/lms/blob/486f56a6324cd100735737f90787a9c8b8641b70/lms/services/canvas_api/factory.py#L22

`CanvasAPIClient` (or more accurately its private helper class
`AuthenticatedClient`) catches `NoOAuth2Token` and turns it into
`CanvasAPIAccessTokenError`:

https://github.com/hypothesis/lms/blob/486f56a6324cd100735737f90787a9c8b8641b70/lms/services/canvas_api/_authenticated.py#L61-L65

### DIFF
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -13,7 +13,6 @@ from lms.services import (
     CanvasAPIPermissionError,
     CanvasFileNotFoundInCourse,
     LTIOutcomesAPIError,
-    NoOAuth2Token,
 )
 from lms.validation import ValidationError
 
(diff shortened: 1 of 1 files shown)


jon-betts (approved):
Ummm... YUP!

PR #2643, opened 2021-05-05
Add ProxyAPIError and ProxyAPIAccessTokenError

seanh:
This is some ground work before adding in Blackboard API support.

A new `ProxyAPIError` base class is added which both `CanvasAPIError`
and `LTIOutcomesAPIError` inherit from. In future a new
`BlackboardAPIError` will also inherit from this. The
`proxy_api_error()` exception view that was previously associated with
`CanvasAPIError` and `LTIOutcomesAPIError` is now associated with
`ProxyAPIError` so if future Blackboard code uses `ProxyAPIError` it'll
get the exception view behavior for free.

`CanvasAPIAccessTokenError` is renamed to `ProxyAPIError` so that it can
be reused by the Blackboard API code. For example
`CanvasAPIAccessTokenError`'s are filtered out from Sentry. Since that's
now been changed to a `ProxyAPIAccessTokenError` that the Blackboard
code can also use, the Blackboard code will get the Sentry filtering for
free. There was nothing actually Canvas-specific about
`CanvasAPIAccessTokenError`.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -69,13 +69,31 @@ class HAPIError(ExternalRequestError):
     """
 
 
-class CanvasAPIError(ExternalRequestError):
+class ProxyAPIError(ExternalRequestError):
     """
-    A problem with a Canvas API request.
+    A problem with a third-party API request.
 
-    Raised whenever a Canvas API request times out or when an unsuccessful,
-    invalid or unexpected response is received from the Canvas API.
+    Raised whenever a third-party API request times out or when an
+    unsuccessful, invalid or unexpected response is received from a third-party
+    API.
     """
+
+
+class ProxyAPIAccessTokenError(ProxyAPIError):
+    """
+    A problem with a third-party API access token.
+
+    Raised when a third-party API request fails because we don't have a
+    third-party access token for the user, or our access token is expired and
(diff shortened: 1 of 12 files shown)


lms/services/exceptions.py
```diff
@@ -144,36 +162,23 @@ def _exception_class(response):
...
         if error_description == "refresh_token not found":
-            return CanvasAPIAccessTokenError
+            return ProxyAPIAccessTokenError
 
         if (
             status_code == 401
             and {"message": "Insufficient scopes on access token."} in errors
         ):
-            return CanvasAPIAccessTokenError
+            return ProxyAPIAccessTokenError
```
seanh, line 174: Might want to move the `CanvasAPIError.raise_from()` class method. Previously it always raised `CanvasAPIError` or a subclass. Now it might raise `ProxyAPIAccessTokenError` which isn't a subclass of `CanvasAPIError`

jon-betts (approved):
This looks good, and I'm all for generalising stuff which isn't actually Canvas specific. This looks like it's going in the direction of LMS being a little more tool consumer agnostic, which is good.

I think you're right about moving the canvas exception `raise_from` handler, but it looks like a separate PR to me.

PR #2644, opened 2021-05-05
Add ApplicationInstance.lms_host()

seanh:
The Canvas API client uses the netloc of `ApplicationInstance.lms_url` as the base host to construct Canvas API URLs. The Blackboard API client is going to have to do the same, so move the netloc parsing into the model where both services can use it.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -1,5 +1,6 @@
 import secrets
 from datetime import datetime
+from urllib.parse import urlparse
 
 import sqlalchemy as sa
 from Cryptodome import Random
@@ -52,6 +53,25 @@ class ApplicationInstance(BASE):
     group_infos = sa.orm.relationship(
         "GroupInfo", back_populates="application_instance"
     )
+
+    def lms_host(self):
+        """
+        Return the hostname part of this ApplicationInstance's lms_url.
+
+        For example if application_instance.lms_url is
+        "https://example.com/lms/" then application_instance.lms_host() will
+        return "example.com".
+
+        :raise ValueError: if the ApplicationInstance's lms_url can't be parsed
+        """
+        # urlparse() or .netloc will raise ValueError for some invalid URLs.
+        lms_host = urlparse(self.lms_url).netloc
+
+        # For some URLs urlparse(url).netloc returns an empty string.
+        if not lms_host:
+            raise ValueError()
+
+        return lms_host
 
     @classmethod
     def get_by_consumer_key(cls, db, consumer_key):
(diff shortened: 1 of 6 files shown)


lms/services/application_instance_getter.py
```diff
@@ -57,6 +57,18 @@ def lms_url(self):
...
+        """
+        Return the hostname part of the current request's LMS URL.
+
+        :raise ConsumerKeyError: if the request's consumer key isn't in the DB
+        :raise ValueError: if the request's LMS URL can't be parsed
+
+        :return: the matching LMS host
+        :rtype: str
+        """
+        return self._get_by_consumer_key().lms_host()
```
seanh, line 70: `ApplicationInstanceGetter` is a bit silly but I'm not fixing that in this PR

lms/models/application_instance.py
```diff
@@ -53,6 +54,25 @@ class ApplicationInstance(BASE):
...
+        :raise ValueError: if the ApplicationInstance's lms_url can't be parsed
+        """
+        # urlparse() or .netloc will raise ValueError for some invalid URLs.
+        lms_host = urlparse(self.lms_url).netloc
+
+        # For some URLs urlparse(url).netloc returns an empty string.
+        if not lms_host:
+            raise ValueError()
+
+        return lms_host
```
seanh, line 74: This is a method not a property because it raises an exception. It's considered bad design for properties to raise exceptions (because it's surprising / unexpected), PyLint complains about it.

lms/services/application_instance_getter.py
jon-betts: ~I feel like `ApplicationInstanceGetter` could do with giving you back an application instance for example rather than awkwardly wrapping every method of one, but yes, fodder for another day.~

I've noticed that we've also grown an "ApplicationInstanceService" which has a get method. A small tweak to that service would seem to make this makes this unnecessary. The `ApplicationInstanceService` requires a `consumer_key`, but if that was optional and a default of `request.lti_user.oauth_consumer_key` was present, then this could be:

`application_instance_service.get().lms_host()`

The same would be true for a lot of these straight pass through functions. I'm not sure what the plan is with the two services, so I'll assume there is one and it's in progress?

jon-betts (commented):
I'm going to cut the review a little short as I'm a bit confused on the parallel `ApplicationInstanceService` and `ApplicationInstanceGetter`.

It seems to me that if `ApplicationInstanceService` could default the `consumer_key` value to the global value then the `ApplicationInstanceGetter` set of repetitions of the `lms_host` and attendent tests could be avoided.

Is this a partially executed plan I'm seeing the middle of here?

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2690

PR #2646, opened 2021-05-06
Upgrade file picker app buttons

LMS007:
~depends on https://github.com/hypothesis/frontend-shared/pull/48~
~depends on https://github.com/hypothesis/frontend-shared/pull/67~

### Commits
-------

Remove Button.js and corresponding scss

-------

Update buttons in FilePickerApp 

- Switch Button over to LabeledButton in FilePickerApp
- Rename several classes in FilePickerApp
- Adjust css and html of button wrapper to work as a column lay out rather than use inline-flex
- TODO: Replace FilePickerApp__actions* with exported classes from frontend-shared when ready.

-------

Replace data-test with data-testid 

This matches our convention for the testing id attribute

--------

### Discussion 

**Before**
May-05-2021 17-07-13

**After**
Screen Shot 2021-05-18 at 9 18 27 AM

Biggest visual difference is the button is not as dark (which matches are primary color) and there is a transition effect on hover. The padding is also a bit different and there is no more red hover color--this matches other LabeledButtons.


lyzadanger (changes requested):
Hiya! I think this can be accomplished with less customization—after a bit of experimentation, I think the left-alignment of the button text may be the only thing we need to override on the button itself. See what you think.

* I don't think the hover should be red (unless there is a compelling reason other than "it was red before") as this doesn't match our established button patterns.
* Several of the rules currently applied to the customized button class can be tackled in the parent element (more on this shortly).
* As the `layout-space` variables and handling of same are currently being re-established and re-considered in `frontend-shared`, let's avoid copying them around at this time.

Regarding some of the layout styling we're trying to achieve with these buttons:

* You can use a `flex-column` on a containing element to lay out the buttons in the columnar presentation we're aiming for here. The `.u-layout-column` utility style (from the `client` and just about to land in `frontend-shared`) has rules that do what you'd need:

```css
.button-parent-container-whatever {
    display: flex;
    flex-direction: column;
    justify-content: flex-start;
    align-items: stretch;
}
```

The width of the flex column container will establish the width of the buttons (they will all stretch to fill).
* Establish the vertical spacing between the buttons on the flex container element (not on the buttons themselves. Our general convention is that elements/components should not assert their own margins—that's the job of the parent container. See the client's `vertical-rhythm` mixin/utility style for implementation. 
* I can see you're overriding padding but I'm unsure why. You can use the `variant="large"` prop value to get a little more padding by default, which will match the established patterns. 

This leaves the alignment of text in the buttons, which may be a legitimate need for customization (`justify-content: flex-start;`).

lyzadanger:
Also: it looks like the cited dependency is merged. Can you update the PR description if so?

robertknight:
FYI a new release of the @hypothesis/frontend-shared package has been published and `LabeledButton` now sets `type="button"` by default. An issue has also been filed related to supporting [control over label alignment](https://github.com/hypothesis/frontend-shared/issues/63).

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -32,8 +29,9 @@
 }
 
 .FilePickerApp__document-source-buttons {
```
LMS007, line 31: I could have also added the vertical rhythm mixin here, but since it was unchanged from the util class I opted to include the actual exported class in the component directly. "u-vertical-rhythm"

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -32,8 +29,9 @@
...
-  align-items: stretch;
+  @include layout.column($align: 'flex-start');
```
LMS007, line 32: This endined being one way to keep the buttons from going full width. I had trouble if the display was set to "flex", I do wonder if there is an even simpler or better approach?

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -32,8 +29,9 @@
...
+  // Override layout.column's display from "flex" to "inline-flex" so the buttons don't
+  // take the full width
   display: inline-flex;
```
LMS007, line 35: This ended up being one way to keep the buttons from going full width. I had trouble if the display was set to "flex", I do wonder if there is an even simpler or better approach?

lms/static/styles/frontend_apps.scss
```diff
@@ -30,6 +30,7 @@
 
 // Shared styles from frontend-shared package
 @use '@hypothesis/frontend-shared/styles';
+@use '@hypothesis/frontend-shared/styles/util';
```
LMS007, line 33: Part of my thinks we should get all the exported classes in one go, this feels a bit confusing, but I realize it's still a new package.

lms/static/styles/components/_FilePickerApp.scss
lyzadanger: Danger zone! Recall from this PR yesterday: https://github.com/hypothesis/frontend-shared/pull/66 that all utility classes are about to get a prefix. Also, there is documentation here: https://github.com/hypothesis/frontend-shared/tree/main/styles about what the current package SASS API is, and (in the FAQ) section, specifically advises against using any utility classes directly. In short, this class won't exist after the next package release.

lms/static/styles/frontend_apps.scss
lyzadanger: Danger zone! Recall from this PR yesterday: https://github.com/hypothesis/frontend-shared/pull/66 that all utility classes are about to get a prefix. Also, there is documentation here: https://github.com/hypothesis/frontend-shared/tree/main/styles about what the current package SASS API is, and (in the FAQ) section, specifically advises against using any utility classes directly. In short, this class won't exist after the next package release.

lms/static/styles/components/_FilePickerApp.scss
lyzadanger: My gut hunch is that there probably is a more straightforward way, but I am not certain: I'd have to get in and poke around.

One suggestion: perhaps don't try to use any mixins or utilities and just see what CSS is needed to get the job done. A columnar flex layout does seem plausible...But the `inline-flex` display gives me pause.

lms/static/styles/components/_FilePickerApp.scss
lyzadanger: See comment further down: this will break on the next `frontend-shared` package release.

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -1,3 +1,6 @@
+@use '@hypothesis/frontend-shared/styles/mixins/layout';
+@use '@hypothesis/frontend-shared/styles/components/buttons';
```
lyzadanger, line 2: Is this used?

lyzadanger (changes requested):
Hiya! Alas, using any of the shared-package utility CSS classes is not currently feasible (see my inline comments). Feel free to ping me for a voice-chat or Slack if you want to run other approaches by me as you move forward.

lms/static/styles/components/_FilePickerApp.scss
LMS007: left over vestige from previous effort, good catch.

lms/static/styles/components/_FilePickerApp.scss
LMS007: After some careful consideration, I think what is here currently seems reasonable. We have a flex container set to vertical, but we don't want to take the full width, so we use inline-flex. Button items then flow and fill up the space that the longest one takes.

robertknight:
As part of https://github.com/hypothesis/lms/pull/2687 I'm planning to extract the part of `FilePickerApp` that deals with selecting content into a separate component. That will end up obsoleting the changes here. The issue in https://github.com/hypothesis/frontend-shared/issues/63 remains valid, so I would suggest resolving that first.

LMS007:
Updated PR top depend on button align fix in FES. I also adjusted the vertical spacing to match the soon-to-land column rule where it uses 1/2 REM rather than 1 REM. 

I was not able to replicate `hyp-actions--column` 
e.g.

```
@include layout.column;    
@include layout.vertical-rhythm($-space * 0.5);
```

here because that allows the container to expand the full width. In this case it's the form with is 600px.

LMS007:
Converting to draft until hypothesis/frontend-shared#67 lands

lyzadanger:
> I also adjusted the vertical spacing to match the soon-to-land column rule where it uses 1/2 REM rather than 1 REM.

Please don't! It's dangerous to start suddenly using `rem` units in one place without looking at the bigger picture.

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -1,18 +1,16 @@
+@use '@hypothesis/frontend-shared/styles/mixins/layout';
```
lyzadanger, line 1: I think this needs to be removed.

robertknight:
I have [opened a PR](https://github.com/hypothesis/lms/pull/2707) that extracts the content selection UI/logic out of `FilePickerApp` into a separate component as part of https://github.com/hypothesis/lms/issues/2524.

That PR conflicts with this one, so if this lands first we'll need to copy across the changes into that PR.

lyzadanger:
I have pushed a fixup commit: https://github.com/hypothesis/lms/pull/2646/commits/58eb5f0b12312801b902c9098b6779954adeecec that addresses HTML (containers) and styling for the LMS File Picker action buttons.

As of this commit, the file-picker buttons look like this:

Screen Shot 2021-05-19 at 12 30 15 PM

The left alignment will come when the relevant change is released in `frontend-shared`.

I have not resolved the (pre-existing) conflict between this branch and `master`—please let me know if I'm needed for that task.

I'm going to un-assign myself as reviewer for this PR.

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -31,9 +27,26 @@
...
 
-.FilePickerApp__document-source-buttons {
-  align-items: stretch;
-  display: inline-flex;
-  flex-direction: column;
-  margin-top: 30px;
+.FilePickerApp__actions {
+  // A flex container (row) with two children: a column of buttons and a
+  // "spacer" to take up the rest of the row space not needed by the buttons.
+  // The spacer prevents the button container from stretching full width.
```
robertknight, line 33: I think "stretching full width" is meant to say "stretching full height"?

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -1,18 +1,14 @@
 @use "../variables" as var;
 
+$-space: 1rem; // TODO: replace this with a real variable
```
robertknight, line 3: This variable is unused.

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -31,9 +27,26 @@
...
+  // The spacer prevents the button container from stretching full width.
+  display: flex;
+  // Add some white space before we start rendering the buttons
+  margin-top: 3em;
+
+  &-buttons {
+    // Display buttons in a column with a vertical margin between each button
+    display: flex;
+    flex-direction: column;
+    & > * + * {
```
robertknight, line 42: A possible way to achieve the same result without `!important`:

```scss
& > :not(:first-child) {
  margin-top: 0.5em;
}
```

lms/static/styles/components/_FilePickerApp.scss
```diff
@@ -29,9 +27,24 @@ $-space: 1rem; // TODO: replace this with a real variable
...
+      // Margin reset on button makes `!important` necessary here
+      margin-top: 0.5em !important;
+    }
+  }
+
+  &-spacer {
```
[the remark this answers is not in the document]
robertknight: Since this utility isn't doing anything besides `u-stretch` using the utility class would save a little code here. If a bunch of other rules were added to the selector then I think that would better justify a new class.

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -100,35 +100,35 @@ describe('FilePickerApp', () => {
...
     const wrapper = renderFilePicker();
-    assert.equal(wrapper.find('Button').length, 2);
+    assert.equal(wrapper.find('LabeledButton').length, 2);
   });
 
   it('renders Canvas file picker button if Canvas file picker enabled', () => {
     fakeConfig.filePicker.canvas.enabled = true;
     const wrapper = renderFilePicker();
-    assert.isTrue(wrapper.exists('Button[label="Select PDF from Canvas"]'));
+    assert.isTrue(wrapper.exists('LabeledButton[data-testid="pdf-button"]'));
```
robertknight, line 109: 👍 - Changing label matching to `data-testid` seems good here. We'll need to do this more widely when we add support for other languages which might happen later this year.

robertknight (commented):
Visually this looks good. The change to matching based on `data-testid`s rather than labels in the tests is I think a good change. I had some minor comments on the code:

- I found one unused SASS variable
- I had a couple of suggestions regarding styling (the `margin-top` override and use of `u-stretch`)

lms/static/styles/components/_FilePickerApp.scss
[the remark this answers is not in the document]
lyzadanger: What I was trying to get at in my comment is that I acknowledge that the `.u-stretch` class would have accomplished the same thing, but felt that ergonomically for the time being it was easier to understand if all of the rules were in one place in this particular case. It's easier to look at as a whole and squint and think: "Is this a pattern that should/can be extracted?" I see this styling as temporary, in that I hope we'll have a reusable columnar-actions pattern out of `frontend-shared` in the near- to medium-term. (Such a pattern already exists in `frontend-shared`, but it's still in flux as to how we'd provide this as part of the public SASS/CSS API).

Anyway, if you feel strongly, we can use `.u-stretch`. I'm not that passionate!

lms/static/styles/components/_FilePickerApp.scss
lyzadanger: This is a very interesting comment, thank you. A simple way to bump up specificity. Might I suggest:

* We leave this as-is for the moment, as the `!important` is caged in pretty well and is probably _safe_, and
* I have an issue where I'm tracking some work related to the existing layout utility classes and mixins: https://github.com/hypothesis/frontend-shared/issues/74 . This selector/rule is based on the `vertical-rhythm` mixin/pattern. I could evaluate at that time using this higher-specificity strategy for those mixins and utility classes.

In summary, I expect these new rules (here, in this LMS component's SASS) to have a short lifespan, to be (aspirationally) replaced by shared patterns soon.

lyzadanger:
@LMS007 @robertknight 

I've responded to a couple of @robertknight 's comments pertaining to SASS rules and strategy. I'm not going to make any code changes at this time and will leave it to @LMS007 to remove the unused SASS variable and any other changes needed to land this!

lms/static/styles/components/_FilePickerApp.scss
LMS007: aw.. missed that!

lms/static/styles/components/_FilePickerApp.scss
LMS007: The spacer is for width

```
Container text which is longer than the longest button
[button 1            ] [spacer.............................................
[b u t t o n 2       ]  
[b  u  t  t  o  n  3 ]                                                      ]
```

lms/static/styles/components/_FilePickerApp.scss
LMS007: Because this is going to be replaced shortly, I'm in favor of just leaving this. We may want to include Rob's suggestion in the util class in the shared  lib.

lms/static/styles/components/_FilePickerApp.scss
robertknight: Oh, right. I misread the first line that said "flex container (row)".

robertknight (approved):

robertknight:
> Lyza fixed her issues herself

That was poorly worded. What I meant is that Lyza addressed the issues she raised in an earlier review by making changes directly to the code.

PR #2648, opened 2021-05-06
Change `data-test` to `data-testid`

esanzgar:
Testing libraries uses `data-testid`. We have a fixture of both
`data-test` and `data-testid`. This changes makes consitent use of
`data-testid`.


LMS007 (approved):
Thank you, I will change my other PR too to match standard.

PR #2649, opened 2021-05-06
Specify frontend and backend tests options

esanzgar:
It is sometimes useful to run only the frontend or the backend tests. I
made that options more obvious by adding them on the `make help` output.

In addition I removed the reference to gulp and substitute it by a more
general `yarn test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,9 @@ help:
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make format            Correctly format the code"
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
-	@echo "make test              Run the unit tests"
+	@echo "make test              Run all unit tests"
+	@echo "make backend-tests     Run the backend unit tests"
+	@echo "make frontend-tests    Run the frontend unit tests"
 	@echo "make functests         Run the functional tests"
 	@echo "make bddtests          Run the gherkin tests"
 	@echo "make sure              Make sure that the formatter, linter, tests, etc all pass"
@@ -41,8 +43,6 @@ supervisor: python
 .PHONY: devdata
 devdata: build/manifest.json  python
 	@tox -qe dev --run-command 'python bin/update_dev_data.py conf/development.ini'
-
-GULP := node_modules/.bin/gulp
 
 .PHONY: shell
 shell: build/manifest.json python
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,7 +12,9 @@ help:
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make format            Correctly format the code"
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
-	@echo "make test              Run the unit tests"
+	@echo "make test              Run the all tests"
```
robertknight, line 15:
```suggestion
	@echo "make test              Run the unit tests"
```

Saying "run all tests" is misleading because this command doesn't run _all_ of the Python tests. Only the Python _unit_ tests. There are some other integration/E2E tests as well.

Makefile
```diff
@@ -144,7 +144,11 @@ sure: checkformatting backend-lint frontend-lint backend-tests frontend-tests fu
 
 .PHONY: frontend-tests
 frontend-tests: node_modules/.uptodate
-	@$(GULP) test
+ifdef ARGS
+	yarn test $(ARGS)
+else
+	yarn test
+endif
```
robertknight, line 151: I'm ambivalent about this. On the one hand it ensures that `yarn test` and `make frontend-tests` are consistent. On the other hand its slightly slower than running `gulp test` as it introduces an extra process. There is also that issue with SIGINT handling and running `yarn test` vs `gulp test` that we ran into in the client recently.

robertknight (changes requested):
I requested one small change to the documentation, otherwise LGTM.

Makefile
esanzgar: Good observation. It was fixed.

Makefile
esanzgar: I am ambivalent too. The client uses `yarn test`.

Makefile
```diff
@@ -12,7 +12,9 @@ help:
 	@echo "make lint              Run the code linter(s) and print any warnings"
 	@echo "make format            Correctly format the code"
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
-	@echo "make test              Run the unit tests"
+	@echo "make test              Run the all unit tests"
```
robertknight, line 15:
```suggestion
	@echo "make test              Run all unit tests"
```

robertknight (approved):
LGTM with one small grammatical correction.

PR #2650, opened 2021-05-06
Replaced switch by map

esanzgar:
This refactor replaces a switch statement for a map. The code is more
concise.


LMS007 (approved):
GTG, this was a bit long before.

PR #2651, opened 2021-05-06
Merge application instance services

seanh:
This replaces the `ApplicationInstanceGetter` and `ApplicationInstanceService` services with a single `ApplicationInstanceService` that only has one method: `get()`.

## End result

We now have just a single `ApplicationInstanceService` service, and it has only a single `get()` method:

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/services/application_instance.py#L7-L32

## Future work

We could probably make this even more convenient by replacing calls to `request.find_service(name="application_instance").get()` (where the default consumer key is used, which is the majority case) with `request.application_instance` or `context.application_instance`. But that's for another PR

### `get()` now raises `ConsumerKeyError`

`ApplicationInstanceService.get()` now raises `ConsumerKeyError` if the `consumer_key` doesn't exist in the DB, instead of returning `None`. This is what `ApplicationInstanceGetter` used to do and it's expected by all the code that used `ApplicationInstanceGetter` that now uses `ApplicationInstanceService`.

The only place where the `None`-returning `ApplicationInstanceService.get()` was used was in the admin pages views, and those have now been fixed to handle `ConsumerKeyError`:

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/views/admin.py#L32-L54

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/views/admin.py#L94-L98

### `get()` now uses the request's consumer key by default

The `consumer_key` argument to `ApplicationInstanceService.get()` is now optional and defaults to the current request's consumer key (`request.lti_user.oauth_consumer_key`). This was a behavior that `ApplicationInstanceGetter` had and it's pretty handy, most of the time the current request's `ApplicationInstance` is the one you want. This default consumer key is passed in to `ApplicationIntanceService` by its factory:

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/services/application_instance.py#L35-L41

When the consumer key is `None` `get()` will raise `ConsumerKeyError`.

### `ApplicationInstanceService.get()` now has caching

Same as `ApplicationInstanceGetter` used to have.

### `ApplicationInstanceGetter` forwarding methods replaced by direct model attr access

`ApplicationInstanceGetter` had a lot of methods that just forwarded to `ApplicationInstance`. For example `ApplicationInstanceGetter.developer_key()`:

https://github.com/hypothesis/lms/blob/705b76a25536107bde8d22edbec890bf62d32595/lms/services/application_instance_getter.py#L19-L28

This would raise `ConsumerKeyError` if there was no `ApplicationInstance` matching the request's consumer key in the DB (the `_get_by_consumer_key()` helper method did that) but otherwise it just returns `ApplicationInstance.developer_key`.

This has now been replaced with calls to `application_instance_service.get().developer_key`, which does the same thing.

Several `ApplicationInstanceGetter` methods were replaced in the same way.

### `ApplicationInstanceGetter.provisioning_enabled()` was removed

`ApplicationInstanceGetter` had a non-trivial `provisioning_enabled()` method:

https://github.com/hypothesis/lms/blob/705b76a25536107bde8d22edbec890bf62d32595/lms/services/application_instance_getter.py#L60-L71

This was only used in a couple of places in `JSConfig` and `LTIHService`, and these have just been replaced with `application_instance_service.get().provisioning`.

### `ApplicationInstanceGetter.canvas_sections_supported()` was removed

`ApplicationInstanceGetter` had a non-trivial `canvas_sections_supported()` method:

https://github.com/hypothesis/lms/blob/705b76a25536107bde8d22edbec890bf62d32595/lms/services/application_instance_getter.py#L73-L81

This was only used in a couple of places, in `LTILaunchResource` and the Canvas API authorize view. These got replaced with `bool(application_instance_service.get().developer_key)`, plus catching the `ConsumerKeyError`.

### `ApplicationInstanceGetter.developer_secret()` was moved into the model

`ApplicationInstanceGetter` had a non-trivial `developer_secret()` method:

https://github.com/hypothesis/lms/blob/705b76a25536107bde8d22edbec890bf62d32595/lms/services/application_instance_getter.py#L30-L47

This got moved into the `ApplicationInstance` model. It needs to be a method because `request.registry.settings["aes_secret"]` needs to be passed into it (unlike the service which has a factory for injecting settings). And it needed to be called `decrypted_developer_secret()` to prevent clashing with the existing `developer_secret` column:

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/models/application_instance.py#L56-L62

### `ApplicationInstanceService.update_settings()` was removed

`ApplicationInstanceService` had an `update_settings()` method:

https://github.com/hypothesis/lms/blob/705b76a25536107bde8d22edbec890bf62d32595/lms/services/application_instance.py#L15-L28

This was only used by the admin pages views. I just removed it and had the views use `ApplicationIntance.settings` directly instead:

https://github.com/hypothesis/lms/blob/afbf9f34214e51624bc4c54bd990fd1297b2e5c6/lms/views/admin.py#L69-L92


lms/models/application_instance.py
```diff
@@ -53,6 +53,14 @@ class ApplicationInstance(BASE):
         "GroupInfo", back_populates="application_instance"
     )
 
+    def decrypted_developer_secret(self, aes_secret):
+        if self.developer_secret is None:
+            return None
+
+        cipher = AES.new(aes_secret, AES.MODE_CFB, self.aes_cipher_iv)
+
+        return cipher.decrypt(self.developer_secret)
```
seanh, line 62: Moved `ApplicationInstanceGetter.developer_secret()` into the model. It has to be a method rather than a property because it needs `request.registry.settings["aes_secret"]` as an argument and, unlike services, models don't have factories to inject settings.

lms/services/application_instance.py
```diff
@@ -1,34 +1,41 @@
...
-        Update ApplicationInstance.settings.
+        consumer_key = consumer_key or self._default_consumer_key
 
-        For not nullable values only changes the params passed with a not None value
-        """
-        if canvas_sections_enabled is not None:
-            ai.settings.set("canvas", "sections_enabled", canvas_sections_enabled)
+        application_instance = ApplicationInstance.get_by_consumer_key(
+            self._db, consumer_key
+        )
```
seanh, line 27: `get()` now raises `ConsumerKeyError` if the `consumer_key` isn't found. That's what `ApplicationInstanceGetter` did, and the user code expects it. Also `consumer_key` is optional and defaults to the one from the request (very useful)

lms/resources/_js_config.py
```diff
@@ -26,8 +26,8 @@ def _consumer_key(self):
         return self._lti_user.oauth_consumer_key
 
     @property
-    def _ai_getter(self):
-        return self._request.find_service(name="ai_getter")
+    def _application_instance_service(self):
+        return self._request.find_service(name="application_instance")
```
jon-betts, line 30: Might simplify things a little to make this return the application instance rather than the service, as we always call `get()` on it anyway.

lms/resources/_js_config.py
```diff
@@ -167,7 +167,10 @@ def google_picker_origin():
             # When we're being launched in an iframe within the LMS our JavaScript
             # needs to pass this URL (which is the URL of the top-most page) to Google
             # Picker, otherwise Picker refuses to launch inside an iframe.
-            return self._context.custom_canvas_api_domain or self._ai_getter.lms_url()
+            return (
+                self._context.custom_canvas_api_domain
+                or self._application_instance_service.get().lms_url
```
jon-betts, line 172: So this would become:

```suggestion
                or self._application_instance().lms_url
```

etc.

lms/services/application_instance.py
```diff
@@ -1,34 +1,41 @@
...
 
 class ApplicationInstanceService:
-    def __init__(self, db):
+    def __init__(self, db, default_consumer_key):
         self._db = db
+        self._default_consumer_key = default_consumer_key
 
-    def get(self, consumer_key: str) -> Optional[ApplicationInstance]:
-        """Query one ApplicationInstance by consumer_key."""
+    @lru_cache
```
jon-betts, line 12: What's the default size on `lru_cache`?

lms/services/canvas_api/factory.py
```diff
@@ -12,16 +12,21 @@ def canvas_api_client_factory(_context, request):
...
-    canvas_host = urlparse(ai_getter.lms_url()).netloc
     basic_client = BasicClient(canvas_host)
 
     authenticated_api = AuthenticatedClient(
         basic_client=basic_client,
         oauth2_token_service=request.find_service(name="oauth2_token"),
-        client_id=ai_getter.developer_key(),
-        client_secret=ai_getter.developer_secret(),
+        client_id=application_instance.developer_key,
+        client_secret=developer_secret,
```
jon-betts, line 29: :+1: I think that makes _so_ much more sense

lms/services/canvas_api/factory.py
```diff
@@ -12,16 +12,21 @@ def canvas_api_client_factory(_context, request):
     :param request: Pyramid request object
     :return: An instance of CanvasAPIClient
     """
-    ai_getter = request.find_service(name="ai_getter")
+    application_instance_service = request.find_service(name="application_instance")
+    application_instance = application_instance_service.get()
```
jon-betts, line 16:
```suggestion
    application_instance = request.find_service(name="application_instance").get()
```

lms/services/canvas_api/factory.py
```diff
@@ -12,16 +12,21 @@ def canvas_api_client_factory(_context, request):
...
+
+    canvas_host = urlparse(application_instance.lms_url).netloc
```
jon-betts, line 18: Presumably this would be replaced with your new function?

lms/views/admin.py
```diff
@@ -71,16 +73,14 @@ def update_instance(self):
...
             "groups_enabled" in self.request.params
             and self.request.params["groups_enabled"] == "on"
         )
+        ai.settings.set("canvas", "groups_enabled", groups_enabled)
 
-        self.application_instance_service.update_settings(
-            ai,
-            canvas_sections_enabled=sections_enabled,
-            canvas_groups_enabled=groups_enabled,
-        )
```
jon-betts, line 83: I assume any interesting gate keeping that was happening in `update_settings()` can equally well happen in `set()`?

lms/views/api/canvas/authorize.py
```diff
@@ -40,31 +40,32 @@
     permission=Permissions.API,
 )
 def authorize(request):
-    ai_getter = request.find_service(name="ai_getter")
+    application_instance_service = request.find_service(name="application_instance")
+    application_instance = application_instance_service.get()
```
jon-betts, line 44:
```suggestion
    application_instance = request.find_service(name="application_instance").get()
```

lms/resources/lti_launch.py
```diff
@@ -101,7 +102,13 @@ def canvas_sections_supported(self):
             # Canvas course sections feature was released.
             return False
 
-        return self._request.find_service(name="ai_getter").canvas_sections_supported()
+        application_instance_service = self._request.find_service(
+            name="application_instance"
+        )
+        try:
+            return bool(application_instance_service.get().developer_key)
```
jon-betts, line 109:
```suggestion
        try:
            return bool(self._request.find_service(name="application_instance").get().developer_key)
```

Not sure if this is better or worse? Depends on how `black` handles it?

lms/services/application_instance.py
```diff
@@ -1,34 +1,41 @@
...
+        """
+        Return the ApplicationInstance with the given consumer_key.
 
-        return ApplicationInstance.get_by_consumer_key(self._db, consumer_key)
+        If no consumer_key is given return the current request's
+        ApplicationInstance (the ApplicationInstance whose consumer_key matches
+        request.lti_user.oauth_consumer_key).
 
-    @staticmethod
-    def update_settings(ai, canvas_sections_enabled=None, canvas_groups_enabled=None):
```
jon-betts, line 16: This did some work of hiding the details of how to set particular things. For example you can't just set "canvas.bananas".

On the other hand:

 * Seems annoying to collect here
 * If we want any enforcement the `set()` method looks like the place to me
 * Means this doesn't become a horrible bottle neck through which all settings flow at once

lms/services/application_instance.py
```diff
@@ -1,34 +1,41 @@
...
 
 
 def factory(_context, request):
-    return ApplicationInstanceService(
-        request.db,
-    )
+    if request.lti_user:
+        consumer_key = request.lti_user.oauth_consumer_key
+    else:
+        consumer_key = None
```
jon-betts, line 39:
```suggestion
    consumer_key = request.lti_user.oauth_consumer_key if request.lti_user else None
```

jon-betts (commented):
I'd be lying if I said I've gone through this in detail, but I'd say it looks a lot better. Generally speaking things of the form:

* `application_instance.property`

Make so much more sense than:

* `ai_getter().property`

`application_instance_service` is a bit of a mouthful, but it's very explicit, and if we do move to `account_service` or something in future will be totally fine.

I've put some random suggestions in as I was reading through. Nothing much. The only bigger suggestion I have right now is that maybe having a `get(raise=False)` would simplify some of the calls where we're catching `ConsumerKeyError`. You'd need to try it both ways to see if that helps though.

jon-betts:
I had another stupid idea about `get()`. We could have a singleton `DefaultApplicationInstance` which returns the default value for every key which doesn't even have to be a real application instance, just implement fixed values like:

```python
class DefaultApplicationInstance:
    provisioning=False
```

We could add a flag to `get()` to return instead of raising, which would greatly simplify calls which don't always care if a real AI is available like this:

```python
if application_instance_service.get(defaults=True).provisioning:
   ...

File: lms/resources/_js_config.py (hunk 4 above)
Comment (seanh) [reply lms/resources/_js_config.py:172]: :+1:

File: lms/services/canvas_api/factory.py (hunk 8 above)
Comment (seanh) [reply lms/services/canvas_api/factory.py:18]: Yep, when a later PR adds the new function

File: lms/views/admin.py (hunk 9 above)
Comment (seanh) [reply lms/views/admin.py:83]: Yeah, but I don't think we need the gate keeping

File: lms/views/api/canvas/authorize.py (hunk 10 above)
Comment (seanh) [reply lms/views/api/canvas/authorize.py:44]: :+1:

File: lms/services/application_instance.py (hunk 13 above)
Comment (seanh) [reply lms/services/application_instance.py:39]: :+1:

File: lms/resources/_js_config.py [hunk 14]
```diff
@@ -25,9 +25,13 @@ def _h_user(self):
...
-    @property
-    def _ai_getter(self):
-        return self._request.find_service(name="ai_getter")
+    def _application_instance(self):
+        """
+        Return the current request's ApplicationInstance.
+
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
+        """
+        return self._request.find_service(name="application_instance").get()
```
seanh, line 34: Since this helper method now returns the actual `ApplicationInstance` (rather than the service) it can raise `ConsumerKeyError`, which it didn't before. Added the docstring for that reason

lms/resources/_js_config.py
```diff
@@ -159,15 +170,28 @@ def enable_content_item_selection_mode(self, form_action, form_fields):
         :param form_fields: the fields (keys and values) to include in the
             HTML form that we'll use to submit the user's chosen document
         :type form_fields: dict
+
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
```
seanh, line 174: This could raise `ConsumerKeyError` before and it still can now. Added it to the docstring

lms/resources/_js_config.py
```diff
@@ -159,15 +170,28 @@ def enable_content_item_selection_mode(self, form_action, form_fields):
...
+            """
             # Pass the URL of the LMS that is launching us to our JavaScript code.
             # When we're being launched in an iframe within the LMS our JavaScript
             # needs to pass this URL (which is the URL of the top-most page) to Google
             # Picker, otherwise Picker refuses to launch inside an iframe.
-            return self._context.custom_canvas_api_domain or self._ai_getter.lms_url()
+            return (
+                self._context.custom_canvas_api_domain
+                or self._application_instance().lms_url
+            )
```
seanh, line 194: This is simply equivalent to the previous code and, like the previous code, it can raise `ConsumerKeyError` (which I added to the docstring)

lms/resources/_js_config.py
```diff
@@ -253,6 +277,7 @@ def maybe_set_focused_user(self):
         In theory, though, the focused_user param could work outside of Canvas
         as well if we ever want it to.
 
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
```
seanh, line 280: This could raise `ConsumerKeyError` before and it still can now, just added it to the docstring

lms/resources/_js_config.py
```diff
@@ -317,13 +342,17 @@ def _blackboard_files_available(self):
         return self._request.feature("blackboard_files")
 
     def _canvas_files_available(self):
-        """Return True if the Canvas Files API is available to this request."""
+        """
+        Return True if the Canvas Files API is available to this request.
+
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
+        """
```
seanh, line 349: This could raise `ConsumerKeyError` before and it still can now, just added it to the docstring

lms/resources/_js_config.py
```diff
@@ -418,6 +447,8 @@ def _hypothesis_client(self):
 
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
+
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
```
seanh, line 451: Could raise `ConsumerKeyError` before and still can now

lms/services/canvas_api/factory.py
```diff
@@ -12,16 +12,20 @@ def canvas_api_client_factory(_context, request):
     :param request: Pyramid request object
     :return: An instance of CanvasAPIClient
     """
-    ai_getter = request.find_service(name="ai_getter")
+    application_instance = request.find_service(name="application_instance").get()
```
seanh, line 15: So just _constructing_ the Canvas API client can raise `ConsumerKeyError`. Which was already the case before. It's not actually a problem because nothing tries to use this service without a valid consumer key

lms/services/course.py
```diff
@@ -67,7 +67,7 @@ def _is_pre_sections(self, authority_provided_id):
 
 def course_service_factory(_context, request):
     return CourseService(
-        request.find_service(name="ai_getter"),
+        request.find_service(name="application_instance").get(),
```
seanh, line 70: This means that _constructing_ the `CourseService` can now raise `ConsumerKeyError`, whereas previously it wouldn't have been raised until `get_or_create()` was called. `CourseService` is used in a few places, I've had a look through them all and I don't think this will make any difference.

lms/services/lti_h.py
```diff
@@ -41,7 +43,7 @@ def sync(self, h_groups, group_info_params):
         :raise HTTPInternalServerError: if we can't sync to h for any reason
         """
 
-        if not self._ai_getter.provisioning_enabled():
+        if not self._application_instance_service.get().provisioning:
```
seanh, line 46: This could have raised `ConsumerKeyError` before and it still can now, no change

seanh:
> I had another stupid idea about `get()`. We could have a singleton `DefaultApplicationInstance` which returns the default value for every key which doesn't even have to be a real application instance, just implement fixed values like:
> 
> ```python
> class DefaultApplicationInstance:
>     provisioning=False
> ```
> 
> We could add a flag to `get()` to return instead of raising, which would greatly simplify calls which don't always care if a real AI is available like this:
> 
> ```python
> if application_instance_service.get(defaults=True).provisioning:
>    ...
> ```

I might pass on this one for now. There's actually only two places in the code that call `ApplicationInstanceService.get()` and catch the `ConsumerKeyError` and fall back to a default value, so the benefit right now would be minimal. An `ApplicationInstance` actually has [quite a lot of attributes](https://github.com/hypothesis/lms/blob/402a6822a03607102aea9846884ba50219e835f5/lms/models/application_instance.py#L18-L54) (14, plus two methods) so I don't know if creating a separate `DefaultApplicationInstance` class to mirror it all would be practical.

lms/services/application_instance.py
jon-betts: I wonder if this is a case where having the service actually be persistent and cache across requests would be good, or terrible and confusing?

We'd need to actually do the interface thing though

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -2,65 +2,43 @@
...
 
+    @pytest.fixture(autouse=True)
+    def default_application_instance(self, db_session, pyramid_request):
+        ai = factories.ApplicationInstance(
+            consumer_key=pyramid_request.lti_user.oauth_consumer_key
+        )
+        db_session.flush()
+        return ai
+
     @pytest.fixture(autouse=True)
```
jon-betts, line 42: Not added by you, but I think the naming here could be tidied up a bit. We have `test_application_instance`. I don't think `test` is doing any work other than to distinguish it from the `application_instances` below. We have a comment to explain that these are "noise". Using the rule of naming that says if you have to immediately explain yourself it's a bad name, I'd suggest:

* `test_application_instance` -> `application_instance` - I think this is our normal policy for things to just name it after the thing when possible
* `application_instances` -> `noise_application_instances`

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -73,3 +51,25 @@ def application_instances(self):
...
+        pyramid_request.lti_user = None
+
+        application_instance_service = factory(mock.sentinel.context, pyramid_request)
+
+        ApplicationInstanceService.assert_called_once_with(pyramid_request.db, None)
+        assert application_instance_service == ApplicationInstanceService.return_value
+
+    @pytest.fixture(autouse=True)
+    def ApplicationInstanceService(self, patch):
+        return patch("lms.services.application_instance.ApplicationInstanceService")
```
jon-betts, line 75: :+1: Cover dem gaps!

A part of me is screaming to parameterise this but... nah, it would end up about as long with a big if statement in the middle

lms/services/course.py
```diff
@@ -5,8 +5,8 @@
 
 
 class CourseService:
-    def __init__(self, ai_getter, consumer_key, db):
-        self._ai_getter = ai_getter
+    def __init__(self, application_instance, consumer_key, db):
+        self._application_instance = application_instance
```
jon-betts, line 9: Do we actually want an application instance if we only access it's settings? Do we just want settings instead?

This isn't a flexibility we'd ever use in this case, but I think it's a nice idea in general if we really don't need or want other details about the parent object, as we can use this class with the sub object from any origin (not just the parent context).

lms/services/lti_h.py
```diff
@@ -23,7 +23,9 @@ def __init__(self, _context, request):
         self._h_user = request.lti_user.h_user
 
         self._authority = request.registry.settings["h_authority"]
-        self._ai_getter = request.find_service(name="ai_getter")
+        self._application_instance_service = request.find_service(
+            name="application_instance"
+        )
```
jon-betts, line 28: Like the service above, it seems what we really want to know is if provisioning is on, rather than actually keeping the whole application instance around.

Also I kind of hate having the service be it's own factory now we've started doing it the other way...

lms/services/oauth1.py
```diff
@@ -6,7 +6,9 @@ class OAuth1Service:
 
     def __init__(self, _context, request):
         self.request = request
-        self.ai_getter_service = request.find_service(name="ai_getter")
+        self.application_instance = request.find_service(
+            name="application_instance"
+        ).get()
```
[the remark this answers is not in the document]
jon-betts: Heh... guess what ma feedback is! Yeah, I think we probably want `self.shared_secret =` rather than keeping the whole application instance about.

lms/views/admin.py
```diff
@@ -47,20 +48,21 @@ def find_instance(self):
...
-        if ai:
+        try:
+            ai = self.application_instance_service.get(self.request.params["query"])
+        except ConsumerKeyError:
+            self.request.session.flash(
+                f'No application instance found for {self.request.params["query"]}',
+                "errors",
+            )
+            return HTTPFound(location=self.request.route_url("admin.instances"))
+        else:
```
jon-betts, line 59: I can't make a suggestion in Github as it spans deleted lines, but this might be clearer without the look before you leap, especially as the param `query` is kind of vague.

```python
try:
    consumer_key = self.request.params["query"]
except KeyError:
    raise HTTPBadRequest()

try:
    ai = self.application_instance_service.get(consumer_key)
except ConsumerKeyError:
    self.request.session.flash(
        f'No application instance found for {consumer_key}',
        "errors",
    )
```

lms/views/admin.py
```diff
@@ -81,16 +83,14 @@ def update_instance(self):
             "sections_enabled" in self.request.params
             and self.request.params["sections_enabled"] == "on"
         )
+        ai.settings.set("canvas", "sections_enabled", sections_enabled)
+
         groups_enabled = (
             "groups_enabled" in self.request.params
             and self.request.params["groups_enabled"] == "on"
         )
+        ai.settings.set("canvas", "groups_enabled", groups_enabled)
```
jon-betts, line 92: I couldn't spot any asserts related to these lines in the tests, but it's late and I could just be going cross eyed.

jon-betts (approved):
I've gone over this in a more fine grained (read pernickety) way and made more suggestions around the tests etc.

All of the suggestions fall into one of two camps:

* Pretty trivial suggestions to do with things in this PR
* Pretty trivial suggestions to do with things highlighted in this PR that existed before

I'd understand if you want to ignore all of the latter for speed. Some of the former might be worth considering. It all depends how sick you are of it.

Some areas I think that are probably worth looking at are:

* One place where we have a lot of things happening in one try statement which is disguised by fluent calling
* A look before you leap thing with "query" param which can be made nicer
* Some possible missing coverage with `ai.settings.set()` where clearly we're hitting the lines, but I can't see any assertions being made about it

lms/resources/_js_config.py
```diff
@@ -131,6 +135,13 @@ def enable_canvas_oauth2_redirect_error_mode(
         )
 
     def enable_lti_launch_mode(self):
+        """
+        Put the JavaScript code into "LTI launch" mode.
+
+        This mode launches an assignment.
+
+        :raise ConsumerKeyError: if request.lti_user.consumer_key isn't in the DB
+        """
```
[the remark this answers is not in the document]
seanh: I agree, but with the way the code is I found it quite helpful / necessary to trace all the places that could raise `ConsumerKeyError` before and that can raise it now in order to verify that I hadn't made any unintentional behaviour changes.

I think we should consider lots of stacked `:raise:`'s a code smell. I think there's two mitigations. One is that the code shouldn't have too many layers. In this case `ApplicationInstanceService` raises `ConsumerKeyError` and `JSConfig` calls `ApplicationInstanceService` so `JSConfig` also raises `ConsumerKeyError`. The only other layer (that calls `JSConfig`) is the view, and the particular views that call it would never trigger the `ConsumerKeyError` since authentication means that the view can't be called without a valid `ApplicationInstance` (or in the case of the views that _can_ be called without an `ApplicationInstance`, they don't call any of the `JSConfig` methods that might raise). So we come to a stop quite quickly, in tracing all the things that might raise this `ConsumerKeyError`. Still we could have one layer here instead of two.

The other thing I think is maybe a design issue with `JSConfig` where a context resource is wrapping a service, and has lots of methods that might raise. There may be a better design there, such as making sure that a `JSConfig` can't even be constructed in the first place if there isn't a valid `ApplicationInstance`. Or just having views call a service directly rather than using a resource (so make a `JSConfigService`).

I think Pyramid's "context resource" pattern is sometimes useful, but only in quite specific ways. For example the idea of an annotation update API view, where the `models.Annotation` to be updated (or an `AnnotationResource` that wraps `models.Annotation`) gets passed in to the view as the `context` argument and the view can then pass it to services. That seems quite useful. But I think with `LTILaunchResource` and `JSConfig` the LMS app is doing things in resources that really belong in services. Really, a lot of code makes more sense in services than anywhere else, and other layers like resources shouldn't be over-used.

tests/unit/lms/services/course_test.py
```diff
@@ -44,15 +46,15 @@ def test_it_does_nothing_if_theres_already_a_matching_row(
...
         db_session.add(
             CourseGroupsExportedFromH(
                 authority_provided_id="test_authority_provided_id",
                 created=datetime.datetime.utcnow(),
             )
         )
-        ai_getter.settings.return_value.set(
+        application_instance_service.get.return_value.settings.set(
             "canvas", "sections_enabled", canvas_sections_enabled
         )
```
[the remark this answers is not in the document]
seanh: The `svc` fixture could be changed to instantiate the service directly rather than through a factory. But the service instance would still be a fixture. If each individual test instantiated the service itself, then every test would have to do this:

```python
def test_foo(self, pyramid_request)
    svc = ApplicationInstanceService(pyramid_request.db, "test_consumer_key")
    ...
```

For services that need more arguments (things from the request, mocks of other services, ...) this can get a lot worse, as each test could have to depend on several fixtures (mock services etc) and make a complex service class `__init__()` call.

And if the service's arguments changed every test would have to be updated.

lms/services/lti_h.py
seanh: I think it's probably best not to do this because `ApplicationInstanceService.get()` can raise `ConsumerKeyError` so if `LTIHService.__init__()` calls `ApplicationInstanceService.get().provisioning` then `LTIHService.__init__()` can now raise `ConsumerKeyError`, which means that `request.find_service(name="lti_h")` can raise `ConsumerKeyError`. I think it's probably best that `find_service()` doesn't raise, as I think that'll be quite unexpected, easily missed. Even if the `:raise:` were documented in `LTIHService.__init__()` I think it'd be likely missed as you wouldn't think to look for it. Service factories shouldn't raise, which means service `__init__()`'s shouldn't raise unless the factory is going to handle that. It's better for `LTIHService.sync()` to raise `ConsumerKeyError` and document it in its docstrings. _Normal_ methods raising exceptions is something that developers expect and know to look out for.

`LTIHService` having a `self._application_instance_service` might come in handy in future if it needs to use any future `ApplicationInstanceService` methods.

tests/unit/conftest.py
```diff
@@ -184,6 +169,14 @@ def application_instance_service(pyramid_config):
         instance=True,
         spec_set=True,
         consumer_key=mock.sentinel.consumer_key,
+        lms_url="https://example.com",
+        shared_secret="TEST_SECRET",
+        provisioning=True,
+        settings=ApplicationSettings({}),
+    )
```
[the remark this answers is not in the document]
seanh: Hmm. This is making a mock `ApplicationInstance` not a real one. IIRC it is possible to get the real factory to actually return a mock or even just a data dict (which you could then pass to `create_autospec(..., **data)` or whatever). But I'm not actually sure _why_ a mock is being used here, so I just replaced it with a real model from the factory, which required fixing a bunch of tests: https://github.com/hypothesis/lms/pull/2651/commits/7751f3d6d93e6686e8d20bb56689550cf39eaa6b

lms/views/admin.py
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/64d67dab9082c0d94c7aa8c9b8c958756b1ed1fc

lms/services/course.py
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/bc3c8f7a6eabb42dcfbd26f7efb38974c60b4fd6

tests/unit/lms/services/application_instance_service_test.py
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/63b16573ec4251e6626dbdb930252d79a591986c

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -2,65 +2,43 @@
...
-
-        # Re-query again
-        ai = db_session.query(ApplicationInstance).get(test_application_instance.id)
-        assert (
-            # pylint: disable=compare-to-zero
-            ai.settings.get(*setting_location.split("."))
-            is False
-        )
+    def test_get_with_default_consumer_key(self, svc, default_application_instance):
+        assert svc.get() == default_application_instance
```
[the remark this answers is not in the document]
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/b448fefb2f675246217addef4d73aedc18d591a3

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -117,20 +118,25 @@ def test_its_support_for_speedgrader(
```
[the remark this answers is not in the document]
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/aa1d88c0212078c3b3643204bd34444681379fe4

tests/unit/lms/models/application_instance_test.py
```diff
@@ -92,6 +93,27 @@ def test_get_returns_None_if_theres_no_matching_ApplicationInstance(
             is None
         )
 
+    def test_decrypted_developer_secret_returns_the_decrypted_developer_secret(
+        self, application_instance, pyramid_request
+    ):
+        aes_secret = pyramid_request.registry.settings["aes_secret"]
+        application_instance.aes_cipher_iv = _build_aes_iv()
```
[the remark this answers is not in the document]
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/5b77c03a7f4f8e0a2f4305920ea1a5a9fb1117d0. Had to add some test coverage too: https://github.com/hypothesis/lms/pull/2651/commits/886bf07951daddb4ad0a2cb29995e7c596dba9eb

lms/services/oauth1.py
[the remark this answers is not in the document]
seanh: I think this actually needs to not even call `ApplicationInstanceService.get()` until the `get_client()` method, otherwise `OAuth1Service.__init__()` could raise `ConsumerKeyError` which'd cause `request.find_service(name="oauth1")` to raise `ConsumerKeyError` which I think would be unexpected and a likely source of bugs. Fixed that: https://github.com/hypothesis/lms/pull/2651/commits/d18f7485d44918ac9c5e90669fb35ebc781ede0e

lms/resources/lti_launch.py
```diff
@@ -101,7 +102,14 @@ def canvas_sections_supported(self):
...
 
-        return self._request.find_service(name="ai_getter").canvas_sections_supported()
+        try:
+            return bool(
+                self._request.find_service(name="application_instance")
+                .get()
+                .developer_key
+            )
+        except ConsumerKeyError:
+            return False
```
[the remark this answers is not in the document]
seanh: Done: https://github.com/hypothesis/lms/pull/2651/commits/be7cdd790301edde413d60672f90934f80b5292e

lms/services/course.py
seanh: Changed this back. I don't think it's a good idea for `CourseService.__init__()` to raise as that would cause `request.find_service(name="course")` to raise which I think would be unexpected and a likely source of errors: https://github.com/hypothesis/lms/pull/2651/commits/6ad80f0e01c485252a3c2fc52d502bd7afb2156f

lms/services/course.py
seanh: I ended up changing this again so that `CourseService.__init__()` doesn't raise: https://github.com/hypothesis/lms/pull/2651/commits/6ad80f0e01c485252a3c2fc52d502bd7afb2156f

lms/resources/_js_config.py
jon-betts: This PR + review is getting mega huge, so probably take this suggestion as an idea for a follow up (it would be nice to get this one done) but this feels very extra to me to define a function which just returns a single value and call it later _and_ to document it too. Could we put this in a variable instead, or just put the value inline later on?

```suggestion
google_picker_origin = (
    self._context.custom_canvas_api_domain
    or self._application_instance().lms_url
)
```

lms/views/admin.py
jon-betts: Yep, you can delete these lines and the tests pass, which isn't a good sign.

tests/unit/conftest.py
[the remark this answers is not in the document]
jon-betts: Sorry!

tests/unit/lms/services/course_test.py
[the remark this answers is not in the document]
jon-betts: Yeah, I'm just suggesting making the `svc` fixture instantiate it directly (not removing the fixture), and have separate tests for the factory, precisely because this would make it less end-to-end and more unit-y. If the service arguments change, you just change the one fixture.

This prevents the integration test type situation where you make one mistake in factory and every test blows up. This way only the factory tests would blow up, which would give you the strong hint it's the factory, not the class that's at fault.

Checkmate is a pretty good example of this style: https://github.com/hypothesis/checkmate/blob/main/tests/unit/checkmate/services/secure_link_test.py

jon-betts (approved):
I think this is pretty good. There are some remaining areas I think could be better, but honestly this PR is so massive the cost to work out what's changed since last view is incredibly high. I'd say any further changes to this really have to come after merging this.

I think it would have been good to see if any of this could have been done in a few stand alone PRs. This might have required having parallel code for things for a little while so we have a "setup new thing", "move things to it", "remove old" type PRs, which might be a little less huge.

The actual amount of lines of code changed here isn't so much the problem but the fact it's distributed 3 lines a go absolutely everywhere.

In any case, some things I think we could do in follow ups:

* In `lms.views.admin.update_instance()` there are two lines with `ai.settings.set(...`. You can delete those lines and the tests still pass
* I think our service testing style in other products like Checkmate is better, where we test the factories separately from the service. This is more of a unit friendly approach and means a fault in the factory doesn't cause massive explosions in all the unit tests for the service. It also means they are usually a little more direct as the fixture wiring is simpler (you don't have to put things on and off the request so often for example).

seanh:
Pushed one more commit to add those missing test assertions: https://github.com/hypothesis/lms/pull/2651/commits/9a2dbefcf56b95b3a5a643abc4b722f86f61c846

PR #2655, opened 2021-05-07
Update the student selector buttons

LMS007:
- Replace the student selector buttons with customized IconButton
- Add buttons.scss module for global lms-only buttons' styles
- Copy over the layout-space vars from the client in preparation for use

-------------

This mostly leaves the styling unchanged and migrated over to the new buttons. I'm not advocating we actually keep the styling, I just don't have a proposed change at this point and the defaults seem a bit too light and get washed out. 

relates to the button audit
https://github.com/hypothesis/lms/issues/2587

**Before**
Screen Shot 2021-05-06 at 5 28 02 PM

**After**
Screen Shot 2021-05-06 at 5 27 58 PM

_*The padding is slightly reduced_

### DIFF
--- a/lms/static/styles/_variables.scss
+++ b/lms/static/styles/_variables.scss
@@ -34,6 +34,16 @@ $subtitle-font-size: 15px;
 
 $input-font-size: 15px;
 
+// Layout spacing
+// -------------------------
+$layout-space: 1em;
+$layout-space--xxsmall: $layout-space * (1/4);
+$layout-space--xsmall: $layout-space * (1/2);
+$layout-space--small: $layout-space * (3/4);
+$layout-space--medium: $layout-space;
+$layout-space--large: $layout-space * 2;
+$layout-space--xlarge: $layout-space * 4;
+
 // Minimum font size for <input> fields on iOS. If the font size is smaller than
 // this, iOS will zoom into the field when focused.
 $touch-input-font-size: 16px;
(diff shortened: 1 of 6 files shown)


lms/static/styles/_variables.scss
```diff
@@ -34,6 +34,16 @@ $subtitle-font-size: 15px;
 
 $input-font-size: 15px;
 
+// Layout spacing
+// -------------------------
+$layout-space: 1em;
```
LMS007, line 39: We don't really need these yet, but I thought it might be good to start copies over the common vars

lms/static/styles/buttons.scss
```diff
@@ -0,0 +1,40 @@
...
+// An icon-only button that sits next to a text input field.
+@mixin InputButtonBase {
+  $-options: (
+    'colormap': $input-button-colors,
+    'withVariants': false,
+  );
+  @include buttons.IconButton($-options) {
+    border: 1px solid var.$grey-3;
+    border-radius: 0; // Turn off border-radius to align with <input> edges
+    height: 100%;
```
LMS007, line 26: Because we don;'t have `.u-layout-row` in LMS yet, this is a bit of a hack to allow the button to fill up the parent container and match the height of the adject input. 

Better, would be to use `u-layout-row`, but I was holding off on that for now

PR #2656, opened 2021-05-07
Use fullpath for scripts instead of using setup.py targets

marcospri:
Without this both `make db` and `make devdata` fail to import pyramid (are not using tox dev env?)

Using the full path of the script with `python` seems to fix it.


jon-betts (changes requested):
Yeah, I don't see what going through `setup.py` buys us except:

 * An extra hoop to jump through
 * Having to unnecessarily install the package
 * Feeling like big grown ups because we used another feature of `setup.py`?
 
 I think we should strip out the entry-points from `setup.py` and do a quick check to make sure they don't show up in CI or anywhere else.

jon-betts (approved):
Yep looks good!

PR #2660, opened 2021-05-07
Fix devdata script to acutally execute the main function

marcospri:
https://github.com/hypothesis/lms/pull/2656 Solved the issue around imports but removed the entry point for devdata which is doing nothing on master


jon-betts (approved):
Hmm, some egg on face for all concerned. Approved!

PR #2661, opened 2021-05-07
Fix gulpfile

LMS007:
Copied from the client...

### Fix runKarma

Karam expects the config to be of type `Config`. This can be done by calling the karma.config.parseConfig method which creates a Config object that can be passed to karma.Server()

----------

### Kill gulp-karma process on Ctrl-C

The only way to kill `yarn test --watch` is sending a SIGINT via a
<kbd>Ctrl</kbd>+<kbd>C</kbd> keyboard shorcut. Killing karma process in
this way sometimes leaves the parent gulp process orphan. That's because
when karma is killed by SIGINT sometimes doesn't run the `done` in the
callback, hence leaving the gulp process waiting for the `done` signal.

On more rare occasions, I have seen orphan karma process too.

The solution presented here registers a listener for SIGINT and call the
`done` function. It doesn't unregister the event.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -181,10 +181,10 @@ gulp.task('watch', gulp.parallel(['watch-js', 'watch-css', 'watch-manifest']));
 function runKarma(done) {
   const karma = require('karma');
   new karma.Server(
-    {
-      configFile: path.resolve(__dirname, './lms/static/scripts/karma.config.js'),
-      ...karmaOptions,
-    },
+    karma.config.parseConfig(
+      path.resolve(__dirname, './lms/static/scripts/karma.config.js'),
+      karmaOptions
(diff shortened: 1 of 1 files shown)


robertknight (approved):
These changes are good. However it would help me as a reviewer to include a reference to the original PRs in the client from which these changes were copied as they contain context about what the changes are for. Here is how I would have written the PR description:

This PR applies the changes from https://github.com/hypothesis/client/pull/3129 and https://github.com/hypothesis/client/pull/3239 to this project, fixing two issues when running frontend tests:

1. Interrupting `yarn test --watch` using Ctrl+C could leave an orphaned gulp process
2. A warning from Karma about deprecated usage of the `karma.Server` constructor

PR #2663, opened 2021-05-10
[WIP] DB Modeling test

marcospri:
Describes a potential new modeling for some LMS entities, there's no renaming or removal of tables yet

# New tables

## LMS

One particular LMS install identified by it's `application_instance_id` (FK) and the `tool_consumer_instance_guid`

## HGroup

A relationship between `application_instance_id` and unique (together) `authority_provided_id` representing groups in H

`CanvasGroup` and `CanvasSection` are both `hgroups` with a different value in the type column. They both don't add new columns

## _Course (replaces Course)

Also an hgroup but adding new columns (and stored in a new different, joined, table) https://docs.sqlalchemy.org/en/14/orm/inheritance.html#joined-table-inheritance

Data could be migrated from existing "course" to the new "_course", probably easier in a a multi step process with the two tables created on the DB.

## _LTIUser 

Existing LTIUser namedtuple as a DB model,  relates to lms. Would replace `LTIUser`

# New relationships

## ModuleItemConfiguration relates to lms 

I don't think this could be backfilled with just the information on this table so it's nullable atm. Maybe as LMS fills with data tool_consumer_instance_guid might be enough but I'd say the application_instance_id would be a best effort guess?

## GradingInfo (`lis_result_sourcedid`) relates to lms

Again not sure if possible to back fill easily, here we have the consumer_key but not the `tool_consumer_instance_guid`

## GradingInfo (`lis_result_sourcedid`) relates to _LTIUser

Same problem as for the lms relationship, here we have h_user and user_id but I think they could tecnically be duplicated (belonging to different instances).

# Other

No changes to `group_info` as it seems to be used mostly for analytics atm

mymodel

### DIFF
--- a/lms/models/_lms.py
+++ b/lms/models/_lms.py
@@ -1,0 +1,21 @@
+import sqlalchemy as sa
+
+from lms.db import BASE
+
+
+class LMS(BASE):
+    __tablename__ = "lms"
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "tool_consumer_instance_guid"),
+    )
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+    )
+
+    tool_consumer_instance_guid = sa.Column(sa.String, nullable=False)
+
+    tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText(), nullable=True)
--- a/lms/models/course.py
+++ b/lms/models/course.py
@@ -3,6 +3,7 @@ from sqlalchemy.dialects.postgresql import JSONB
 
 from lms.db import BASE
 from lms.models.application_settings import ApplicationSettings
+from lms.models.grouping import Grouping
 
 
 class Course(BASE):
@@ -26,6 +27,20 @@ class Course(BASE):
     #: The authority_provided_id that uniquely identifies the course that these
     #: settings belong to.
     authority_provided_id = sa.Column(sa.UnicodeText(), primary_key=True)
+
+    settings = sa.Column(
+        "settings",
+        ApplicationSettings.as_mutable(JSONB),
+        server_default=sa.text("'{}'::jsonb"),
+        nullable=False,
+    )
+
+
+class _Course(Grouping):
+    __tablename__ = "course_grouping"
+    __mapper_args__ = {"polymorphic_identity": "course"}
+
+    id = sa.Column(sa.Integer, sa.ForeignKey("grouping.id"), primary_key=True)
 
     settings = sa.Column(
         "settings",
--- a/lms/models/grading_info.py
+++ b/lms/models/grading_info.py
@@ -69,3 +69,11 @@ class GradingInfo(BASE):
     tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText(), nullable=True)
     h_username = sa.Column(sa.UnicodeText(), nullable=False)
     h_display_name = sa.Column(sa.UnicodeText(), nullable=False)
+
+    lms_id = sa.Column(
+        sa.Integer, sa.ForeignKey("lms.id", ondelete="cascade"), nullable=True
+    )
+
+    lti_user_id = sa.Column(
+        sa.Integer, sa.ForeignKey("lti_user.id", ondelete="cascade"), nullable=True
+    )
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -1,0 +1,30 @@
+import sqlalchemy as sa
+
+from lms.db import BASE
+
+
+class Grouping(BASE):
+    __tablename__ = "grouping"
+    __mapper_args__ = {"polymorphic_identity": "grouping", "polymorphic_on": "type"}
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
+    )
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+    )
+    authority_provided_id = sa.Column(sa.UnicodeText())
+    type = sa.Column(sa.String())
+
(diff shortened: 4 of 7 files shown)


lms/models/course.py
```diff
@@ -33,3 +34,17 @@ class Course(BASE):
         server_default=sa.text("'{}'::jsonb"),
         nullable=False,
     )
+
+
+class _Course(Grouping):
```
marcospri, line 39: This, I think, could be filled with data from the existing `course`

lms/models/course.py
```diff
@@ -33,3 +34,17 @@ class Course(BASE):
...
+    __tablename__ = "course_grouping"
+    __mapper_args__ = {"polymorphic_identity": "course"}
```
marcospri, line 41: Extends grouping adding new columns in a new table that would be joined to `grouping` 

https://docs.sqlalchemy.org/en/14/orm/inheritance.html#joined-table-inheritance

lms/models/grading_info.py
```diff
@@ -69,3 +69,7 @@ class GradingInfo(BASE):
     tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText(), nullable=True)
     h_username = sa.Column(sa.UnicodeText(), nullable=False)
     h_display_name = sa.Column(sa.UnicodeText(), nullable=False)
+
+    lms_id = sa.Column(
```
marcospri, line 73: Nullable, until (if) the existing data can be backfilled

lms/models/lti_user.py
```diff
@@ -1,8 +1,26 @@
...
+class _LTIUser(BASE):
+    __tablename__ = "lti_user"
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+
+    lms_id = sa.Column(
+        sa.Integer, sa.ForeignKey("lms.id", ondelete="cascade"), nullable=True
+    )
+
+    # provider = sa.Column(sa.UnicodeText(), nullable=False)  this is the lms.tool_consumer_instance_guid
```
marcospri, line 18: On the lms

lms/models/lti_user.py
```diff
@@ -1,8 +1,26 @@
...
+from lms.db import BASE
 from lms.models import HUser
 
 
+class _LTIUser(BASE):
+    __tablename__ = "lti_user"
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+
+    lms_id = sa.Column(
```
marcospri, line 14: I added the FK to lm_is in some place but didn't add the `relationship`, no reason, just lazyness

lms/models/_lms.py
```diff
@@ -0,0 +1,17 @@
+import sqlalchemy as sa
+
+from lms.db import BASE
+
+
+class LMS(BASE):
+    __tablename__ = "lms"
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    application_instance = sa.Column(
```
marcospri, line 10: This  should be `application_instance_id`

lms/models/lti_user.py
```diff
@@ -1,8 +1,26 @@
...
+        sa.Integer, sa.ForeignKey("lms.id", ondelete="cascade"), nullable=True
+    )
+
+    # provider = sa.Column(sa.UnicodeText(), nullable=False)  this is the lms.tool_consumer_instance_guid
+    provider_unique_id = sa.Column(sa.UnicodeText(), nullable=False)
```
marcospri, line 19: How should the h user id should be called on LTIUser? `authority_provided_id`

lms/models/grading_info.py
seanh: If we're deciding to group everything by `ApplicationInstance` and not by `LMS` I think we probably don't want to add this column here. `GradingInfo` already has an `oauth_consumer_key` column that is part of a unique constraint which groups it by `ApplicationInstance`

lms/models/grouping.py
```diff
@@ -0,0 +1,30 @@
...
+
+from lms.db import BASE
+
+
+class Grouping(BASE):
+    __tablename__ = "grouping"
+    __mapper_args__ = {"polymorphic_identity": "grouping", "polymorphic_on": "type"}
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
+    )
```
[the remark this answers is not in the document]
seanh: Update: in the end I think we probably do want to group by `application_instance_id` only, not `lms_id`. See https://hypothes-is.slack.com/archives/C1MA4E9B9/p1620824569443600?thread_ts=1620817529.437300&cid=C1MA4E9B9

lms/models/_lms.py
```diff
@@ -0,0 +1,21 @@
...
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+    )
+
+    tool_consumer_instance_guid = sa.Column(sa.String, nullable=False)
+
+    tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText(), nullable=True)
```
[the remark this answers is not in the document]
seanh: **Update:** I think we want a nullable, one-to-one relationship between `ApplicationInstance` and `LMS`. Not a one-to-many as you have it here.

1. When an `ApplicationInstance` is first created (when the user submits the https://lms.hypothes.is/welcome form) the `ApplicationInstance` hasn't been installed in an LMS or launched by an LMS yet. At this time `ApplicationInstance.lms_id` is `NULL`
2. The first time we receive an LTI launch for an `ApplicationInstance` whose `lms_id` is `NULL` we create a new `LMS` based on the launch params and set `ApplicationInstance.lms_id` to that LMS's ID. That `ApplicationInstance` is now "locked" to that LMS. Note that the `LMS` table is grouped by `ApplicationInstance`: if multiple `ApplicationInstance`'s are installed in the same LMS that LMS will have multiple entries in the `LMS` table
3. In the future (separate PR) we'll make it so that if we receive an LTI launch for an `ApplicationInstance` and the `tool_consumer_instance_guid` in the launch params doesn't match the `ApplicationInstance`'s related `LMS.tool_consumer_instance_guid` we refuse to launch and show an error message. You cannot install the same `ApplicationInstance` in multiple LMS's or change your LMS's GUID. We're going to show an error message in that case

PR #2666, opened 2021-05-10
Groups enabled and group categories endpoint on JS Config

marcospri:
Adds 
``` 
"filePicker": {
  "groupsEnabled":false,       
  "listGroupCategories":{
    "authUrl":"http://localhost:8001/api/canvas/oauth/authorize",
    "path":"/api/canvas/courses/125/group_categories"
  },
}
```

to JS config when in `content-item-selection` mode

Full config opening the content selection modal on local:

```{
   "api":{
      "authToken":"Bearer ey"
   },
   "canvas":{
      
   },
   "debug":{
      "tags":[
         "role:instructor"
      ]
   },
   "dev":true,
   "filePicker":{
      "blackboard":{
         "enabled":false,
         "listFiles":{
            "authUrl":"http://localhost:8001/api/blackboard/oauth/authorize",
            "path":"/api/blackboard/courses/XXXX/files"
         }
      },
      "canvas":{
         "enabled":true,
         "groupsEnabled":false,
         "listFiles":{
            "authUrl":"http://localhost:8001/api/canvas/oauth/authorize",
            "path":"/api/canvas/courses/125/files"
         },
         "listGroupCategories":{
            "authUrl":"http://localhost:8001/api/canvas/oauth/authorize",
            "path":"/api/canvas/courses/125/group_categories"
         },
         "ltiLaunchUrl":"http://localhost:8001/lti_launches"
      },
      "formAction":"https://hypothesis.instructure.com/courses/125/external_content/success/external_tool_dialog",
      "formFields":{
         "lti_message_type":"ContentItemSelection",
         "lti_version":"LTI-1p0"
      },
      "google":{
         "clientId":"XXXX",
         "developerKey":"XXXX",
         "origin":"hypothesis.instructure.com"
      },
      "vitalSource":{
         "enabled":false
      }
   },
   "mode":"content-item-selection"
}
```

PR #2687, opened 2021-05-17
Add group set selection to assignment configuration form

robertknight:
This branch implements group set selection in the assignment configuration form. The UI should eventually look like [these mocks](https://github.com/hypothesis/lms/issues/2523#issuecomment-821160185). The current state is that the overall flow is functional but the layout is not fully implemented.

When assignment configuration is complete, the the LTI launch URL that is submitted to Canvas and displayed in the Canvas assignment configuration will be the same as before if groups are disabled. If groups are enabled however there will be an additional `group_set` query parameter in the URL whose value is the ID of the Group Set in Canvas.

----

**Assignment configuration flow:**

The initial content screen looks the same as before:

group-set-select-1

group-set-select-2

Once content has been selected, the form submits as before if groups are disabled for an installation of the LMS app. If groups are enabled however, a new screen is shown to allow the user to choose the group configuration:

group-set-select-3

When the user checks the "Use groups" option, the list of Group Sets is fetched from the backend. If this fails, the user is prompted to authorize:

group-set-select-4

Once authorization succeeds, the list of groups is shown in a `<select>` and the "Continue" button is disabled until the user makes a choice:

group-set-select-5

group-set-select-6

group-set-select-7

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -210,6 +210,10 @@ class JSConfig:
             },
             "canvas": {
                 "enabled": self._canvas_files_available(),
+                "groupsEnabled": self._ai_getter.settings().get(
+                    "canvas", "groups_enabled"
+                )
+                or False,
                 # The "content item selection" that we submit to Canvas's
                 # content_item_return_url is actually an LTI launch URL with
                 # the selected document URL or file_id as a query parameter. To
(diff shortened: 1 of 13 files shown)


marcospri:
@robertknight Just merged https://github.com/hypothesis/lms/pull/2666

robertknight:
Thanks for the notice @marcospri. The work in this PR is being broken into a number of others: https://github.com/hypothesis/lms/pull/2706, https://github.com/hypothesis/lms/pull/2707 and one or possibly two more after that. Once those land, the flow should be usable end-to-end.

robertknight:
This draft has been superseded by https://github.com/hypothesis/lms/pull/2716.

PR #2689, opened 2021-05-17
Fix an admin pages crash

seanh:
`/admin/instance/<consumer_key>/` is crashing with:

    jinja2.exceptions.UndefinedError: 'lms.models.application_settings.ApplicationSettings object' has no attribute 'canvas'

Not sure what happened here. How was this every working?


marcospri (approved):
I think the `.` would of worked on AI that had a `canvas` section but not on `{}`.

Well spotted, thanks!

PR #2690, opened 2021-05-17
Add ApplicationInstance.lms_host()

seanh:
The Canvas API client uses the netloc of `ApplicationInstance.lms_url`
as the base host to construct Canvas API URLs. The Blackboard API client
is going to have to do the same, so move the netloc parsing into the
model where both services can use it.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -1,5 +1,6 @@
 import secrets
 from datetime import datetime
+from urllib.parse import urlparse
 
 import sqlalchemy as sa
 from Cryptodome import Random
@@ -60,6 +61,27 @@ class ApplicationInstance(BASE):
         cipher = AES.new(aes_secret, AES.MODE_CFB, self.aes_cipher_iv)
 
         return cipher.decrypt(self.developer_secret)
+
+    def lms_host(self):
+        """
+        Return the hostname part of this ApplicationInstance's lms_url.
+
+        For example if application_instance.lms_url is
+        "https://example.com/lms/" then application_instance.lms_host() will
+        return "example.com".
+
+        :raise ValueError: if the ApplicationInstance's lms_url can't be parsed
+        """
+        # urlparse() or .netloc will raise ValueError for some invalid URLs.
+        lms_host = urlparse(self.lms_url).netloc
+
+        # For some URLs urlparse(url).netloc returns an empty string.
+        if not lms_host:
+            raise ValueError(
+                f"Couldn't parse self.lms_url ({self.lms_url}): urlparse() returned an empty netloc"
+            )
+
+        return lms_host
 
     @classmethod
     def get_by_consumer_key(cls, db, consumer_key):
--- a/tests/unit/lms/models/application_instance_test.py
+++ b/tests/unit/lms/models/application_instance_test.py
@@ -75,6 +75,18 @@ class TestApplicationInstance:
         with pytest.raises(IntegrityError, match="requesters_email"):
             db_session.flush()
 
+    def test_lms_host(self, application_instance):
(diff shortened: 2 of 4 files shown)


lms/models/application_instance.py
```diff
@@ -61,6 +62,25 @@ def decrypted_developer_secret(self, aes_secret):
...
+        return "example.com".
+
+        :raise ValueError: if the ApplicationInstance's lms_url can't be parsed
+        """
+        # urlparse() or .netloc will raise ValueError for some invalid URLs.
+        lms_host = urlparse(self.lms_url).netloc
+
+        # For some URLs urlparse(url).netloc returns an empty string.
+        if not lms_host:
+            raise ValueError()
```
jon-betts, line 80:
```suggestion
            raise ValueError("No net location extracted from url")
```
... or good words... you know.

tests/unit/lms/models/application_instance_test.py
```diff
@@ -75,6 +75,18 @@ def test_requesters_email_cant_be_null(self, db_session, application_instance):
         with pytest.raises(IntegrityError, match="requesters_email"):
             db_session.flush()
 
+    def test_lms_host(self, application_instance):
+        application_instance.lms_url = "https://example.com/lms/"
+
+        assert application_instance.lms_host() == "example.com"
```
jon-betts, line 81: I assume we can always rely on this being fully qualified when we get it? As far as I can tell we don't actually enforce that. `urlparse` behavior is very iffy if you don't have a schema specified or implied with '//':

```python
for url in ("https://google.com:443/path", "google.com:443/path", "google.com"):
    print(url, '>>', urlparse(url).netloc or None)
    
# https://google.com:443/path >> google.com:443
# google.com:443/path >> None
# google.com >> None
```

jon-betts (approved):
This looks good, and better than adding more the the service. 

I think it's nice to add a string to the exception raised to give us a clue if we come across it in future without having to code dive to see why this happens.

A slightly more abstract point is `urllib` isn't very reliable with degenerate urls without schemes. My feelings on that are:

* This PR changes where this happens, but doesn't change that it happens (so it's no worse)
* It's possible a fully qualified URL is purposefully enforced somewhere I didn't spot
* It's possible a fully qualified URL is implicitly enforced because things fail without it

tests/unit/lms/models/application_instance_test.py
seanh: I'm pretty sure users can enter invalid values into the `/welcome` form for this! Even if they do enter a valid domain name that still doesn't mean that it's actually the correct domain name of their Canvas or Blackboard instance and will work when we try to use it to talk to the Canvas or Blackboard API. So we could do with better validation on the way in for sure

seanh: I think we do prepend `https://` if the user doesn't enter a schema though

seanh: https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/static/scripts/new-application-instance.js#L31-L42

seanh: https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/static/scripts/new-application-instance.js#L1-L6

seanh: There's no backend validation:

https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/views/application_instances.py#L27-L35

lms/models/application_instance.py
seanh: Added a string

PR #2692, opened 2021-05-17
Add the blackboard.files_enabled setting

seanh:
Replace the `"blackboard_files"` feature flag with a `blackboard.files_enabled` application instance setting. This allows us to enable Blackboard files for our test assignments in our test Blackboard instance, without enabling them for the public. Also once Blackboard files has been released to the public I think it might be opt-in at least at first, in which case we'll just keep using this setting.

One day we might want to enable Blackboard files by default at which point we'd need to detect whether the LMS is Blackboard (so as not to show the Blackboard files button in non-Blackboard LMS's) and then, when talking to the Blackboard API totally fails, detect that and show an error message explaining how the user needs to get their admin to set up an OAuth client in Blackboard.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -1,7 +1,7 @@
 import functools
 
 from lms.models import GroupInfo, HUser
-from lms.services import ConsumerKeyError, HAPIError
+from lms.services import HAPIError
 from lms.validation.authentication import BearerTokenSchema
 from lms.views.helpers import via_url
 
@@ -197,7 +197,7 @@ class JSConfig:
             "formAction": form_action,
             "formFields": form_fields,
             "blackboard": {
-                "enabled": self._blackboard_files_available(),
+                "enabled": self._blackboard_files_enabled(),
                 "listFiles": {
                     "authUrl": self._request.route_url(
                         "blackboard_api.oauth.authorize"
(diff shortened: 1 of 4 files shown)


lms/resources/_js_config.py
```diff
@@ -337,9 +337,13 @@ def _auth_token(self):
...
-    def _blackboard_files_available(self):
-        """Return True if the Blackboard Files API is available to this request."""
-        return self._request.feature("blackboard_files")
+    def _blackboard_files_enabled(self):
+        """
+        Return True if the Blackboard Files API is enabled for this request.
+
+        :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
+        """
+        return self._application_instance().settings.get("blackboard", "files_enabled")
```
jon-betts, line 346: I think this might be better in the body of the method it's used in. I think it's simple enough and possibly clearer to see where the settings are coming from inline. Maybe something like:

```python
setting = self._application_instance().setting
...
"blackboard: {
    "enabled": settings.get("blackboard", "files_enabled"),
    ...
}
```

_Irrelevant:_ Also... I still want that google inner function dead...

lms/views/admin.py
```diff
@@ -93,6 +93,12 @@ def update_instance(self):
             self.request.params.get("groups_enabled") == "on",
         )
 
+        ai.settings.set(
+            "blackboard",
+            "files_enabled",
+            self.request.params.get("blackboard_files_enabled") == "on",
```
jon-betts, line 99: We could call this param "blackboard.files_enabled" or "settings.blackboard.files_enabled" which would pave the way to fairly easily automate adding these if we had more. We don't have to do that now, but it's not much different.

jon-betts (approved):
I've tried this in the admin page, and it toggle on and it toggle off. Good toggling :+1:. 

This all looks good, but I think the extra private function is largely obscuring what's going on. I also had a tiny suggestion to have a formal link between the form param name and the setting by using "dot" separation. Makes it easy to know what to call them, and leaves a possibility of reading them in automatically if we add loads.

PR #2693, opened 2021-05-17
Raise ProxyAPIAccessTokenError not NoOAuth2Token

seanh:
You would need the frontend changes from
https://github.com/hypothesis/lms/pull/2636 to actually test this, but:

`ProxyAPIAccessTokenError` is the one that has an exception view:

https://github.com/hypothesis/lms/blob/e174b161e90918c94402b8e662e150b063b67b42/lms/views/api/exceptions.py#L107-L109

The exception view returns a response that triggers the frontend to show
a "You need to authorize" dialog.

If the view raises `NoOAuth2Token` instead that gets caught by the
catch-all exception view that triggers a "Something went wrong" dialog.

### DIFF
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -2,6 +2,7 @@
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
+from lms.services import NoOAuth2Token, ProxyAPIAccessTokenError
 from lms.views import helpers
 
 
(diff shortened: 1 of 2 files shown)


lms/views/api/blackboard/files.py
```diff
@@ -37,7 +46,15 @@ def list_files(self):
...
-        self.request.find_service(name="oauth2_token").get()
+        # Get the user's access token from the DB.
+        # We're not actually going to *use* the access token, since this view
+        # just returns hard-coded URLs. But we want to *get* the access token
+        # so that we can raise an exception if it's missing and trigger the
+        # authentication flow.
+        try:
+            self.request.find_service(name="oauth2_token").get()
+        except NoOAuth2Token as err:
+            raise ProxyAPIAccessTokenError() from err
```
jon-betts, line 57: Looks like some duplication here. Would this class benefit from a `_oauth2_token()` method?

seanh: Maybe, but none of this code is going to be here in the end anyway. Looking up access tokens in the DB and communicating with the Blackboard API will move into a `BlackboardAPIService`. These are just fake placeholder views, and they just need to raise this exception to trigger the authorization flow.

jon-betts: Fair enough. If it's a stepping stone.

jon-betts (approved):
If this code is going away then I guess it's ok for it to be duplicated for now. By the same token, I think it could equally well have been fixed from the off which doesn't run the risk of codus interuptus leaving this in place.

PR #2695, opened 2021-05-18
Allow "/" char when checking Base64 string in a test

robertknight:
Add missing "/" character to allowed chars list in a test checking for a
Base64 string. This caused [occassional failures](https://github.com/hypothesis/lms/pull/2686/checks?check_run_id=2599723103) in this test.

[1] https://en.wikipedia.org/wiki/Base64#Base64_table

### DIFF
--- a/tests/unit/lms/services/vitalsource_test.py
+++ b/tests/unit/lms/services/vitalsource_test.py
@@ -58,10 +58,11 @@ class TestVitalSourceService:
         # Ignore non-OAuth signature params in this test.
         params = {k: v for (k, v) in params.items() if k.startswith("oauth_")}
 
+        base64_char = "[0-9a-zA-Z+=/]"
         assert params == {
             "oauth_consumer_key": "lti_launch_key",
             "oauth_nonce": Any.string.matching("[0-9]+"),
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/vitalsource_test.py
```diff
@@ -58,10 +58,11 @@ def test_it_signs_lti_launch_form_params(self, svc, lti_user):
         # Ignore non-OAuth signature params in this test.
         params = {k: v for (k, v) in params.items() if k.startswith("oauth_")}
 
+        base64_char = "[0-9a-zA-Z+=/]"
         assert params == {
             "oauth_consumer_key": "lti_launch_key",
             "oauth_nonce": Any.string.matching("[0-9]+"),
-            "oauth_signature": Any.string.matching("[0-9a-zA-Z+=]+"),
+            "oauth_signature": Any.string.matching(f"{base64_char}+"),
```
robertknight, line 65: This check doesn't fully validate that the string is valid base 64, but I deemed it good enough for the purposes of this test.

seanh (approved):

PR #2696, opened 2021-05-18
Rename CanvasOAuthCallbackSchema to OAuthCallbackSchema

seanh:
There's nothing Canvas-specific about this schema and it's actually now used by the blackboard authorize() view as well.

### DIFF
--- a/lms/validation/authentication/_oauth.py
+++ b/lms/validation/authentication/_oauth.py
@@ -1,4 +1,3 @@
-"""Validation for OAuth views."""
 import secrets
 
 import marshmallow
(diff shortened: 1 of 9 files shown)


lms/validation/authentication/_oauth.py
```diff
@@ -1,4 +1,3 @@
-"""Validation for OAuth views."""
```
seanh, line 1: This docstring wasn't accurate: the validation schemas for most OAuth-related views are elsewhere, this module only contains one particular schema. There's nothing useful for a docstring on this module to say.

marcospri (approved):

PR #2697, opened 2021-05-18
Refactor selected content representation in `FilePickerApp`

robertknight:
Combine the various state fields indicating the type of content selected
for an assignment into a tagged union. This is a precursor to extracting
the content selection UI into a separate component. It also better
represents the fact that only one type of content can be selected for an
assignment.

This would have required a change to the hidden `document_url` field.
This field does not appear to be used for anything and is not a
field that appears in the specs for the Content-Item form [1], so I have
removed it.

[1] https://www.imsglobal.org/specs/lticiv1p0/specification

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -29,6 +29,23 @@ import URLPicker from './URLPicker';
  */
 
 /**
+ * @typedef FileContent
+ * @prop {'file'} type
+ * @prop {File} file
+ *
+ * @typedef URLContent
+ * @prop {'url'} type
+ * @prop {string} url
+ *
+ * @typedef VitalSourceBookContent
+ * @prop {'vitalsource'} type
+ * @prop {string} bookID
+ * @prop {string} cfi
+ *
+ * @typedef {FileContent|URLContent|VitalSourceBookContent} Content
+ */
+
+/**
  * An application that allows the user to choose the web page or PDF for an
  * assignment.
  *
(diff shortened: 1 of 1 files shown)


esanzgar (changes requested):
One thing to consider: because `useState` is async is there a possibility that `contentItem` could not be set by the time the form is posted? It is dangerous to relay on the timing of changing states.

Have you considered removing the `shouldSubmit` and `submit` tandem and relay directly on changes in `content`? Or maybe every click handler (URL, File, VitalSource) to set up `contentItem` directly.

Otherwise, the changes are sensible.

robertknight:
> One thing to consider: because useState is async is there a possibility that contentItem could not be set by the time the form is posted? It is dangerous to relay on the timing of changing states.

That's a reasonable question to ask. If a callback in a component does the following in the same tick of the event loop:

```
setStateA(newValueA);
setStateB(newValueB);
```

Then there are two separate guarantees you might be interested in:

1) Are state updates applied in order? Is `newValueA` guaranteed to be "seen"  at the same time as or before `newValueB`?
2) Are all pending state updates applied at the same time? Is `newValueB` guaranteed to be "seen" in the same render as `newValueA`?

Currently both of these are true in Preact and I think that is very unlikely to change [1]. It would be very surprising if state updates were not applied in order and since state updates are async, triggering separate renders to apply `newValueA` and `newValueB` would be very inefficient.

In this context we only rely on (1) since we always call `setContent` before `submit(true)`, but that's not strictly necessary.

> Have you considered removing the shouldSubmit and submit tandem and relay directly on changes in content? 

It would be possible to do this. However I'm going to make a change soon that will add an extra step to the dialog to choose group configuration for the assignment. At that point we won't be able to submit the form once the content is set and will have to also wait for the group configuration to be set.

esanzgar (approved):
Thanks for the clarifications.

PR #2699, opened 2021-05-18
Add LMS fields to ApplicationInstance

marcospri:
The standalone migration needs to be deployed first: https://github.com/hypothesis/lms/pull/2705

To test it

- Apply DB migration `hdev alembic upgrade head`
- Launch any assignment on canvas
- Check values no present on the new fields 

`select * from application_instances where tool_consumer_instance_guid is not null;`

- Manually change the GUID value to another 

`update application_instances set tool_consumer_instance_guid = 'UNIQUE GUID' where tool_consumer_instance_guid is not null;`

- Launch the assignment again, you should see an error logged on LMS' ouput

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -39,6 +39,35 @@ class ApplicationInstance(BASE):
         nullable=False,
     )
 
+    #: A unique identifier for the LMS instance.
+    tool_consumer_instance_guid = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: The LMS product name, e.g. "canvas" or "moodle".
+    tool_consumer_info_product_family_code = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: A plain text description of the LMS instance, e.g. "University of Hypothesis"
+    tool_consumer_instance_description = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: The URL of the LMS instance, e.g. "https://hypothesis.instructure.com".
+    tool_consumer_instance_url = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: The name of the LMS instance, e.g. "HypothesisU".
+    tool_consumer_instance_name = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: An contact email, e.g. "System.Admin@school.edu"
+    tool_consumer_instance_contact_email = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: Version of the LMS, e.g. "9.1.7081"
+    tool_consumer_info_version = sa.Column(sa.UnicodeText, nullable=True)
+
+    #: This Canvas custom variable substitution $Canvas.api.domain.
+    #: We request this in our config.xml file and name it "custom_canvas_api_domain":
+    #:
+    #: https://github.com/hypothesis/lms/blob/5394cf2bfb92cb219e177f3c0a7991add024f242/lms/templates/config.xml.jinja2#L20
+    #:
+    #: See https://canvas.instructure.com/doc/api/file.tools_variable_substitutions.html
+    custom_canvas_api_domain = sa.Column(sa.UnicodeText, nullable=True)
+
     #: A list of all the OAuth2Tokens for this application instance
     #: (each token belongs to a different user of this application
     #: instance's LMS).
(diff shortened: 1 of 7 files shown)


lms/models/application_instance.py
```diff
@@ -38,6 +41,24 @@ class ApplicationInstance(BASE):
         nullable=False,
     )
 
+    # LMS info linked to this consumer_key, description of each one from the spec
+    # https://www.imsglobal.org/specs/ltiv1p1p1/implementation-guide
+
+    # This is a unique identifier for the TC.  A common practice is to use the DNS of the organization or the DNS of the TC instance.
+    tool_consumer_instance_guid = sa.Column(sa.String, nullable=True)
```
marcospri, line 48: I don't think this is unique, it's "immutable" but we enforce that in the code

lms/models/application_instance.py
```diff
@@ -38,6 +41,24 @@ class ApplicationInstance(BASE):
...
+
+    # In order to better assist tools in using extensions and also making their user interface fit into the TC's user interface that they are being called from
+    tool_consumer_info_product_family_code = sa.Column(sa.String, nullable=True)
+
+    # This is a plain text user visible field – it should be about the length of a line
+    tool_consumer_instance_description = sa.Column(sa.String, nullable=True)
+
+    # This is the URL of the consumer instance
```
marcospri, line 56: Copied the comments from the spec, this one it's not great but left it for consistency

lms/models/application_instance.py
```diff
@@ -66,6 +87,41 @@ def get_by_consumer_key(cls, db, consumer_key):
         """Return the ApplicationInstance with the given consumer_key or None."""
         return db.query(cls).filter_by(consumer_key=consumer_key).one_or_none()
 
+    def update_lms_data(self, params: dict):
```
marcospri, line 90: Not a great interface but it's meant to be called from a couple of places and this is better that doing request.params['...'] for each one in both.

Arguably taking `tool_consumer_instance_guid` explicitly and then the rest could be a better one.

lms/views/basic_lti_launch.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, context, request):
         self.context.js_config.enable_lti_launch_mode()
         self.context.js_config.maybe_set_focused_user()
 
+        request.find_service(name="application_instance").get().update_lms_data(
```
marcospri, line 45: I'm not a fan of having "side effects" on a constructor but here we do some similar things and this is not an object we are going to construct manually most of the time so I'd say it makes sense in this case.

lms/models/application_instance.py
```diff
@@ -38,6 +41,24 @@ class ApplicationInstance(BASE):
...
+
+    # This is a plain text user visible field – it should be about the length of a line
+    tool_consumer_instance_description = sa.Column(sa.String, nullable=True)
+
+    # This is the URL of the consumer instance
+    tool_consumer_instance_url = sa.Column(sa.String, nullable=True)
+
+    # This is a plain text user visible field – it should be about the length of a column
+    tool_consumer_instance_name = sa.Column(sa.String, nullable=True)
+
```
seanh, line 61: Looking at [Potemkin](https://github.com/hypothesis/potemkin/blob/10d6b92666c58239309aabc9ee865bd0bf57637e/potemkin/templates/launch/_includes/builder_form.html.jinja2#L194-L229) I think there might be a couple more LMS-related launch params that we want to record: `tool_consumer_instance_contact_email` and `tool_consumer_info_version`. Searching [the LTI 1.1.1 spec](https://www.imsglobal.org/specs/ltiv1p1p1/implementation-guide) for `tool_consumer_` gives the same results. [Edu App Center](https://www.eduappcenter.com/docs/basics/post_parameters) is another good resource.

There's also a custom Canvas parameter `custom_canvas_api_domain` that [we request](https://github.com/hypothesis/lms/blob/d952bf408960374d71f5d72c08cfa25ba10f2350/lms/templates/config.xml.jinja2#L20), might be worth adding that too.

seanh: There's already a `created` timestamp on `ApplicationInstance`'s. I think we might also want to add an `updated` timestamp and update it whenever we add or update the LMS info. So we know how new the info is.

I wonder if we also want to track the last time the `ApplicationInstance` was _used_, even if the LMS info didn't need to be updated. That would let us know which `ApplicationInstance`'s are actually in use. And it'd also let us know how recently this LMS data was last seen, even if it didn't need to be updated when last seen.

lms/views/content_item_selection.py
```diff
@@ -49,6 +49,10 @@
     schema=ContentItemSelectionLTILaunchSchema,
 )
 def content_item_selection(context, request):
+    request.find_service(name="application_instance").get().update_lms_data(
+        request.params
+    )
```
seanh, line 54: This does make me realise that it might be worth moving `update_lms_data()` into the service because then `request.params` can be injected by the service factory rather than each view having to pass it in. For example it could be:

```python
request.find_service(name="application_instance").update()
```

seanh (changes requested):

lms/views/content_item_selection.py
marcospri: `request.find_service(name="application_instance").update()` That would make me scratch my head, "update how?, there's no params!"

I find explicitly passing the info in easier to understand but if you think this more pyramid-y happy to change it

lms/models/application_instance.py
marcospri: Added the extra fields. I'll tackle the `updated` and possibly `accessed` in a separate PR

lms/models/application_instance.py
```diff
@@ -38,6 +41,33 @@ class ApplicationInstance(BASE):
         nullable=False,
     )
 
+    # LMS info linked to this consumer_key, description of each one from the spec
+    # https://www.imsglobal.org/specs/ltiv1p1p1/implementation-guide
```
seanh, line 45: I think the comment is out of date because we no longer use the descriptions from the spec. It's probably not needed anyway.

```suggestion
```

If an ORM class has lots of columns and we want to separate them with "heading" comments like this I think it might be better to move them into a separate class that the `ApplicationInstance` class inherits from. That way you get class names and docstrings for documentation and the separate classes naturally structure things. Heading comments tend to get out of sync with the code. You could have something like:

```python
class ApplicationInstance(BASE, _ConsumerKeyColumns, _LMSColumns):
    __tablename__ = "application_instances"
    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
    ...

class _ConsumerKeyColumns:
    """ApplicationInstance columns related to the consumer key."""
    consumer_key = sa.Column(sa.String, unique=True, nullable=False)
    ...

class _LMSColumns:
    """ApplicationInstance columns related to the LMS."""
    ...
```

In this case I think just all in one class is fine though

lms/models/application_instance.py
seanh: Yeah, not unique

seanh (changes requested):
One small request about removing a comment otherwise this looks good :+1: The DB migration will have to be deployed as a separate PR and run first

seanh (approved):

PR #2700, opened 2021-05-18
Extract logic to show auth popup into `AuthButton` component

robertknight:
Extract the logic for showing an authorization prompt, when fetching
files from the LMS fails, into a separate `AuthButton` component.

This simplifies the `LMSFilePicker` component and tests and will enable the
"Authorize" button logic to be reused in other places.

----

**Manual testing:**

1. Delete your existing access token for the "localhost (make devdata)" installation of Hypothesis at https://hypothesis.instructure.com/profile/settings
2. Create an assignment and select the "Select PDF from Canvas" option
3. Click "Authorize" and verify that the auth popup appears
4. Check that clicking "Authorize" while the auth popup is open brings focus to the already-open window
5. Check that files are shown after the authorization flow in the popup completes
6. Repeat step (1) and (2). This time check that dismissing the auth popup without authorizing results in a dialog with a "Try again" button, and that clicking "Try again" shows the auth prompt again. Authorizing here should then result in the file list being shown

### DIFF
--- a/lms/static/scripts/frontend_apps/components/AuthButton.js
+++ b/lms/static/scripts/frontend_apps/components/AuthButton.js
@@ -1,0 +1,66 @@
+import { LabeledButton } from '@hypothesis/frontend-shared';
+import { createElement } from 'preact';
+import { useCallback, useEffect, useRef } from 'preact/hooks';
+
+import AuthWindow from '../utils/AuthWindow';
+
+/**
+ * @typedef AuthButtonProps
+ * @prop {string} authURL - Initial URL for the authorization popup. See `AuthWindow`.
+ * @prop {string} authToken - Auth token between the LMS frontend and backend. See `AuthWindow`.
+ * @prop {string} [label] - Custom label for the "Authorize" button
+ * @prop {() => void} onAuthComplete - Callback invoked when the authorization flow completes.
+ *   This does not guarantee that authorization was successful. Instead
+ *   the caller should retry whatever operation triggered the authorization
+ *   prompt and check whether it succeeds.
+ */
+
+/**
+ * Button that prompts the user to authorize the Hypothesis LMS app to access
+ * their data in the LMS (or another service).
+ *
+ * This component is typically shown to the user when an attempt to fetch data
+ * from the LMS via an LMS-specific API fails, requiring the user to complete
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/components/AuthButton.js
```diff
@@ -0,0 +1,66 @@
...
+  /** @type {{ current: AuthWindow|null }} */
+  const authWindow = useRef(null);
+
+  const authorize = useCallback(async () => {
+    if (authWindow.current) {
+      authWindow.current.focus();
+      return;
+    }
+
+    authWindow.current = new AuthWindow({ authToken, authUrl: authURL });
```
robertknight, line 43: `authURL` is the correct casing per https://github.com/hypothesis/frontend-toolkit/pull/34. We are gradually updating non-compliant identifier names of existing code.

lms/static/scripts/frontend_apps/components/AuthButton.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {AuthButtonProps} props
+ */
+export default function AuthButton({
+  authURL,
+  authToken,
+  label = 'Authorize',
+  onAuthComplete,
+}) {
+  /** @type {{ current: AuthWindow|null }} */
```
esanzgar, line 34: That's more or less equivalent to this:
```suggestion
  /** @type {import('preact').Ref<AuthWindow>} */
```

esanzgar (approved):
All worked well for me.

lms/static/scripts/frontend_apps/components/AuthButton.js
robertknight: That's almost true. `import('preact').Ref<T>` is the type used by the `ref` prop of components. It can be either a function or a callback.

`useRef` meanwhile returns a `import('preact/hooks').Ref<T>`. Elsewhere we typically do `useRef(/** @type {...} */ (someValue))` which is shorter. Unfortunately that strips `null` from the type, which is strictly incorrect but useful for the typical use case where a ref starts out as `null` and is then set to a non-null value as soon as the first render happens. Here however we need to reset the `ref` back to null, hence I've written out the type manually.

PR #2703, opened 2021-05-18
Get real access tokens from Blackboard instead of fake ones

seanh:
You'll need https://github.com/hypothesis/lms/pull/2636 to actually test this.

### DIFF
--- a/lms/services/blackboard_api.py
+++ b/lms/services/blackboard_api.py
@@ -1,0 +1,94 @@
+import datetime
+
+import requests
+from sqlalchemy.orm.exc import NoResultFound
+
+from lms.models import OAuth2Token
+from lms.services.exceptions import ProxyAPIAccessTokenError
+
+
+class BlackboardAPIClient:  # pylint:disable=too-many-instance-attributes
+    def __init__(
+        self,
+        blackboard_host,
+        client_id,
+        client_secret,
+        redirect_uri,
+        consumer_key,
+        user_id,
+        db,
+    ):  # pylint:disable=too-many-arguments
+        self.blackboard_host = blackboard_host
+        self.client_id = client_id
+        self.client_secret = client_secret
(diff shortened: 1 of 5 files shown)


lms/services/blackboard_api.py
```diff
@@ -0,0 +1,94 @@
...
+            data={
+                "grant_type": "authorization_code",
+                "redirect_uri": self.redirect_uri,
+                "code": authorization_code,
+            },
+            auth=(self.client_id, self.client_secret),
+        ).prepare()
+
+        response = self._session.send(request, timeout=9)
+        response.raise_for_status()
```
seanh, line 44: This stuff about calling `requests` in the right way is duplicated a few times in the LMS app now. I think we probably want to separate out a `RequestsService` rather than merging this

seanh: Also there needs to be validation and error handling here

lms/views/api/blackboard/authorize.py
```diff
@@ -27,11 +48,9 @@ def authorize(request):
     renderer="lms:templates/api/oauth2/redirect.html.jinja2",
 )
 def oauth2_redirect(request):
-    oauth2_token_service = request.find_service(name="oauth2_token")
+    authorization_code = request.params["code"]
```
seanh, line 51: Validation and error handling needed

lms/services/blackboard_api.py
jon-betts: When I did something similar in a Moodle POC I had a stack of:

* HTTPClient - Making requests + making requests with JSON
* OAuthHTTPClient - As above but calls the parent inserting `Authorization` headers and catching errors etc.
* ConcreteToolClient on top, which adds all the specific stuff

Could be worth a whirl.

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2739

PR #2705, opened 2021-05-19
Add migration adding new fields to application_instance

marcospri:
Just the alembic migration for deployment.

Usage on https://github.com/hypothesis/lms/pull/2699


seanh (changes requested):
This looks good except that `String` should be `UnicodeText` (https://hypothes-is.slack.com/archives/C1MA4E9B9/p1621508513066100?thread_ts=1621496877.065500&cid=C1MA4E9B9)

seanh (approved):

PR #2706, opened 2021-05-19
Extract FilePickerApp hidden form field rendering into separate component

robertknight:
Extract the rendering of hidden form fields in `FilePickerApp` into a separate `FilePickerFormFields` component. This will keep `FilePickerApp` more manageable as additional steps are added to the assignment configuration flow.

There are no functional changes in this PR.

---

**Testing:**

- Try configuring LMS assignments in Canvas using the various available options (Enter URL, select file from Canvas, select file from Google). The "External tool URL" that appears in Canvas should be the same as before.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
@@ -1,0 +1,85 @@
+import { Fragment, createElement } from 'preact';
+
+import {
+  contentItemForUrl,
+  contentItemForLmsFile,
+  contentItemForVitalSourceBook,
+} from '../utils/content-item';
+
+/**
+ * @typedef {import('./FilePickerApp').Content} Content
+ */
+
+/**
+ * @typedef FilePickerFormFieldsProps
+ * @prop {string} ltiLaunchURL - URL that the LMS should use to launch the
+ *   assignment.
+ * @prop {Content} content - Content for the assignment
+ * @prop {Record<string,string>} formFields - Form fields provided by the backend
+ *   that should be included in the response without any changes
+ */
+
+/**
+ * Generate an LTI launch URL for a given piece of content.
+ *
+ * @param {string} ltiLaunchURL
+ * @param {Content} content
+ */
+function contentItemString(ltiLaunchURL, content) {
+  let contentItem = null;
+
+  switch (content.type) {
+    case 'url':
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ *  - A "Content Item Selection" form when configuring an assignment. In this
+ *    case the form will look like "Section 3.2, Example Response" in
+ *    https://www.imsglobal.org/specs/lticiv1p0/specification
+ *  - When an assignment without any content configuration is launched.
+ *    See the `configure_module_item` view.
+ *
+ * @param {FilePickerFormFields} props
+ */
+export default function FilePickerFormFields({
```
esanzgar, line 63: Alternative name for this component could be `FilePickerHiddenFields`.

esanzgar (approved):
One of the types is incorrect, I am sure you will correct.

I added a few suggestions.

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -0,0 +1,90 @@
...
+  content,
+  formFields,
+  ltiLaunchURL,
```
esanzgar, line 66: Maybe instead of passing this `ltiLaunchIURL` around you could injected it here:

```
  const {  {filePicker: canvas: { ltiLaunchUrl }, }, } = useContext(Config);
```

esanzgar: Same for formFields.

robertknight: Hmm ... I'm not sure about this. Currently all the configuration extraction related to this form is in one place, which is useful for comprehension.

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
robertknight: I think it would need to be `FilePickerHiddenFormFields`, which would make sense, although it is pretty verbose. Currently all the fields which actually matter are hidden. I think I'm going to leave this as-is for the moment.

robertknight:
Thanks for the review Eduardo. I have applied all the suggestions except for pushing `formFields` extraction from the `Config` context down and renaming the component. See my responses for the rationale.

PR #2708, opened 2021-05-19
Store course as a grouping

marcospri:
## Testing notes

- Create a course on your local DB using the old schema (from `master` or from this branch but without running the migration) opening an assignment https://hypothesis.instructure.com/courses/125/

- Check the course on the `course` table

```
postgres=# select * from course;
                consumer_key                |          authority_provided_id           |                settings                
--------------------------------------------+------------------------------------------+----------------------------------------
 Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a | 653e9620a656a954c684942f6443fa3e3410c03a | {"canvas": {"sections_enabled": true}}
(1 row)
```

- Apply migration to create the grouping table `hdev alembic upgrade head`
- Check the new table 

```
postgres=# select * from grouping;
 created | updated | id | application_instance_id | authority_provided_id | parent_id | lms_id | type | lms_name | settings | extra 
---------+---------+----+-------------------------+-----------------------+-----------+--------+------+----------+----------+-------
(0 rows)
```

- Lunch any assignment again on the same course  https://hypothesis.instructure.com/courses/125/
- Check DB rows, existing `course` should have been removed and "moved" to the new grouping table

```
postgres=# select * from course;
 consumer_key | authority_provided_id | settings 
--------------+-----------------------+----------
(0 rows)

postgres=# select * from grouping;
-[ RECORD 1 ]-----------+-----------------------------------------------
created                 | 2021-05-24 13:10:06.738377
updated                 | 2021-05-24 13:10:06.738377
id                      | 2
application_instance_id | 8
parent_id               | 
settings                | {"canvas": {"sections_enabled": true}}
```

- Lunch an assigment from a course that didn't exist on `course` beforehand, https://hypothesis.instructure.com/courses/124
- A new row should appear on `grouping`

```
created                 | 2021-05-24 13:12:49.796875
updated                 | 2021-05-24 13:12:49.796875
id                      | 3
application_instance_id | 7
authority_provided_id   | cc5b6de979f437ed8c62105efd4fafb3b635e893
parent_id               | 
lms_id                  | 31e8b270384ca8176925a8d35a48f63a4a30ca5f
lms_name                | Developer Test Course with Sections Disabled
settings                | {"canvas": {"sections_enabled": false}}
extra                   | {"canvas": {"custom_canvas_course_id": "124"}}
```

### DIFF
--- a/lms/models/course.py
+++ b/lms/models/course.py
@@ -5,7 +5,7 @@ from lms.db import BASE
 from lms.models.application_settings import ApplicationSettings
 
 
-class Course(BASE):
+class _Course(BASE):
     """An LTI course."""
 
     __tablename__ = "course"
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -1,0 +1,80 @@
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+
+from lms.db import BASE
+from lms.models import CreatedUpdatedMixin
+from lms.models.application_settings import ApplicationSettings
+
+MAX_GROUP_NAME_LENGTH = 25
+
+
+class Grouping(CreatedUpdatedMixin, BASE):
+    __tablename__ = "grouping"
+    __mapper_args__ = {"polymorphic_identity": "grouping", "polymorphic_on": "type"}
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
+    )
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+        nullable=False,
+    )
+    application_instance = sa.orm.relationship("ApplicationInstance")
+
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
+
+    parent_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("grouping.id", ondelete="cascade"),
+        nullable=True,
+    )
+
+    #: ID on the LMS, not unique across LMS or even in the same LMS instance
+    lms_id = sa.Column(sa.String(), nullable=False)
+
+    #: Full name given on the LMS (e.g. "A course name 101")
+    lms_name = sa.Column(sa.UnicodeText(), nullable=False)
+
+    type = sa.Column(sa.String(), nullable=False)
+
+    settings = sa.Column(
+        "settings",
+        ApplicationSettings.as_mutable(JSONB),
+        server_default=sa.text("'{}'::jsonb"),
+        nullable=False,
+    )
+
+    extra = sa.Column(
+        "extra",
+        ApplicationSettings.as_mutable(JSONB),
+        server_default=sa.text("'{}'::jsonb"),
+        nullable=False,
+    )
+
+    @property
+    def name(self):
+        """Return an h-compatible group name."""
+        name = self.lms_name.strip()
+
+        if len(name) > MAX_GROUP_NAME_LENGTH:
+            return name[: MAX_GROUP_NAME_LENGTH - 1].rstrip() + "…"
+
+        return name
+
+    def groupid(self, authority):
+        return f"group:{self.authority_provided_id}@{authority}"
+
+
+class CanvasSection(Grouping):
+    __mapper_args__ = {"polymorphic_identity": "canvas_section"}
+
+
+class CanvasGroup(Grouping):
+    __mapper_args__ = {"polymorphic_identity": "canvas_group"}
+
+
+class Course(Grouping):
+    __mapper_args__ = {"polymorphic_identity": "course"}
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -24,15 +24,29 @@ class Grouping(CreatedUpdatedMixin, BASE):
     )
     application_instance = sa.orm.relationship("ApplicationInstance")
 
+    #: The authority_provided_id of the Group that was created for this Grouping in h's DB.
     authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
 
-    parent_id = sa.Column(
+        #: The id of the parent grouping that this grouping belongs to.
+        #:
+        #: For example if the grouping represents a Canvas section or group then parent_id
+        #: will reference the grouping for the course that the section or group belongs to.
+        parent_id = sa.Column(
         sa.Integer(),
         sa.ForeignKey("grouping.id", ondelete="cascade"),
         nullable=True,
     )
 
-    #: ID on the LMS, not unique across LMS or even in the same LMS instance
+    #: The LMS's ID for the grouping.
+    #:
+    #: For example for a course this is the value of the context_id launch param.
+    #: For a Canvas section or group this is the value of the section or group's id
+    #: from the Canvas API.
+    #:
+    #: lms_id may not be unique without `parent_id`. For example a Canvas instance may
+    #: have multiple sections or groups with the same id in different courses. In this
+    #: case multiple Grouping's would have the same lms_id but they will have different
+    #: parent_id's.
     lms_id = sa.Column(sa.String(), nullable=False)
 
     #: Full name given on the LMS (e.g. "A course name 101")
--- a/lms/models/h_group.py
+++ b/lms/models/h_group.py
@@ -14,21 +14,6 @@ class HGroup(NamedTuple):
         return f"group:{self.authority_provided_id}@{authority}"
 
     @classmethod
-    def course_group(cls, course_name, tool_consumer_instance_guid, context_id):
-        """
-        Create an HGroup for a course.
-
-        :param course_name: The name of the course
-        :param tool_consumer_instance_guid: Tool consumer GUID
-        :param context_id: Course id
-        """
-        return HGroup(
-            cls._name(course_name),
-            hashed_id(tool_consumer_instance_guid, context_id),
-            type="course_group",
-        )
-
-    @classmethod
     def section_group(
         cls, section_name, tool_consumer_instance_guid, context_id, section_id
     ):
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -69,7 +69,24 @@ class LTILaunchResource:
         return grouping_service.course_grouping(
             tool_consumer_instance_guid=params["tool_consumer_instance_guid"],
             context_id=params["context_id"],
+            extra=self.course_extra,
         )
+
+    @property
+    def course_extra(self):
+        """Extra information to store for courses."""
+        extra = {}
+
+        if self.is_canvas:
+            extra = {
+                "canvas": {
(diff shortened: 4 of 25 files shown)


lms/models/grouping.py
```diff
@@ -0,0 +1,71 @@
...
+        sa.Integer(),
+        sa.ForeignKey("grouping.id", ondelete="cascade"),
+        nullable=True,
+    )
+
+    lms_id = sa.Column(sa.String(), nullable=True)
+
+    type = sa.Column(sa.String(), nullable=False)
+
+    full_name = sa.Column(sa.UnicodeText(), nullable=False)
```
marcospri, line 39: Rename to `lms_name`

lms/models/grouping.py
```diff
@@ -0,0 +1,71 @@
...
+
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
+
+    parent_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("grouping.id", ondelete="cascade"),
+        nullable=True,
+    )
+
+    lms_id = sa.Column(sa.String(), nullable=True)
```
marcospri, line 35: This can be nullable to allow migrating existing courses to the new table

lms/models/grouping.py
```diff
@@ -0,0 +1,71 @@
...
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+        nullable=False,
+    )
+    application_instance = sa.orm.relationship("ApplicationInstance")
+
+    authority_provided_id = sa.Column(sa.UnicodeText(), nullable=False)
+
+    parent_id = sa.Column(
```
marcospri, line 29: `null` for courses. Will be the corresponding course_grouping.id for sections and groups.

lms/services/course.py
```diff
@@ -53,7 +53,16 @@ def _create(self, authority_provided_id):
...
+        return course_settings
+
+    def _create(self, authority_provided_id, name):
+        # By default we'll make our course setting have the same settings
+        # as the application instance
+        course_settings = self.course_settings(
+            self._application_instance, authority_provided_id
+        )
+
+        course = _Course(
```
marcospri, line 65: We only create the existing couse and rely on the HGroup.course_group -> grouping_service.curse_grouping to create rows int he new table.

lms/services/course.py
```diff
@@ -31,20 +31,20 @@ def any_with_setting(self, group, key, value=True):
...
             .count()
         )
 
-    def _get(self, authority_provided_id):
-        return self._db.query(Course).get((self._consumer_key, authority_provided_id))
+    def get(self, authority_provided_id):
+        return self._db.query(_Course).get((self._consumer_key, authority_provided_id))
 
-    def _create(self, authority_provided_id):
+    def course_settings(self, application_instance, authority_provided_id):
```
marcospri, line 44: It's own method as it's used from grouping service.

lms/services/course.py
```diff
@@ -31,20 +31,20 @@ def any_with_setting(self, group, key, value=True):
...
+            self._db.query(_Course)
             .filter(Course.consumer_key == self._consumer_key)
             .filter(Course.settings[group][key] == json.dumps(value))
             .limit(1)
             .count()
         )
 
-    def _get(self, authority_provided_id):
-        return self._db.query(Course).get((self._consumer_key, authority_provided_id))
+    def get(self, authority_provided_id):
```
marcospri, line 41: Used from grouping service.

lms/services/grouping.py
```diff
@@ -0,0 +1,73 @@
...
+        course_name,
+        course_lms_id,
+    ):
+        """
+        Get / update / create a Grouping for a course.
+
+        :param application_instance_id The ID of the AI this course belongs to
+        :param tool_consumer_instance_guid: Tool consumer GUID
+        :param course_name: The name of the course
+        :param context_id: Course id on the LMS
```
marcospri, line 41: Renamed

lms/services/grouping.py
```diff
@@ -0,0 +1,73 @@
...
+        self,
+        application_instance,
+        tool_consumer_instance_guid,
+        course_name,
+        course_lms_id,
+    ):
+        """
+        Get / update / create a Grouping for a course.
+
+        :param application_instance_id The ID of the AI this course belongs to
```
marcospri, line 38: Not ID, the full object

lms/services/grouping.py
```diff
@@ -0,0 +1,73 @@
...
+        :param application_instance: The AI this course belongs to
+        :param tool_consumer_instance_guid: Tool consumer GUID
+        :param course_name: The name of the course
+        :param course_lms_id: Course id on the LMS
+        """
+
+        authority_provided_id = hashed_id(tool_consumer_instance_guid, course_lms_id)
+
+        # We are currently maintaining two tables
+        # to represent Courses `_Course:course` and `Course:gruping`
```
marcospri, line 47: and `Course:grouping`

lms/models/grouping.py
seanh: I'm not sure about this column. There are multiple IDs involved here:

* The `context_id` launch param is a unique identifier for the course. This is a standard LTI launch param, I _think_ it works the same across all LMS's (they all use it as a unique course ID)

* In the Canvas API courses also have an `id`. This is not the same as the `context_id`! It's a different ID. Our https://lms.hypothes.is/config_xml uses `$Canvas.course.id` so that we receive the Canvas API course ID as a `custom_canvas_course_id` launch parameter. We need this `custom_canvas_course_id` to make API requests about the course, you can't use the `context_id` in API requests

* Canvas sections have their own ID that's only unique within the course

* I'm guessing Canvas groups also have their own ID that's only unique within the course

I've got a feeling we might want to store all for IDs separately

seanh: There are some other course-related launch params: `context_type`, `context_label` and `context_title`. I wonder if we want to store those as well? Although apparently we decided not to when we created the original `course` table

lms/models/grouping.py
```diff
@@ -0,0 +1,73 @@
...
+        sa.Integer(),
+        sa.ForeignKey("grouping.id", ondelete="cascade"),
+        nullable=True,
+    )
+
+    #: ID on the LMS, not unique acrooss LMS or even in the same LMS instance
+    lms_id = sa.Column(sa.String(), nullable=True)
+
+    #: Full name given on the LMS (e.g. "A course name 101")
+    lms_name = sa.Column(sa.UnicodeText(), nullable=False)
```
seanh, line 39: Is this gonna come from the `context_title` launch param? If so I wonder if we should call it `context_title` or maybe just `title`?

marcospri: Note that this not a `Course` table, its the grouping one with common fields to all of them, so I think `lms_name` works better to store a course name, a section name and a group name

lms/models/course.py
```diff
@@ -5,7 +5,7 @@
 from lms.models.application_settings import ApplicationSettings
 
 
-class Course(BASE):
+class _Course(BASE):
```
marcospri, line 8: I renamed as "_Course" and then been importing it `as LegacyCourse`, maybe I should just rename it to that?

lms/models/h_group.py
```diff
@@ -13,21 +13,6 @@ class HGroup(NamedTuple):
...
-        Create an HGroup for a course.
-
-        :param course_name: The name of the course
-        :param tool_consumer_instance_guid: Tool consumer GUID
-        :param context_id: Course id
-        """
-        return HGroup(
-            cls._name(course_name),
-            hashed_id(tool_consumer_instance_guid, context_id),
-            type="course_group",
```
marcospri, line 28: Where was this used?

lms/models/grouping.py
```diff
@@ -0,0 +1,80 @@
...
+    extra = sa.Column(
+        "extra",
+        ApplicationSettings.as_mutable(JSONB),
+        server_default=sa.text("'{}'::jsonb"),
+        nullable=False,
+    )
+
+    @property
+    def name(self):
+        """Return an h-compatible group nam."""
```
marcospri, line 59: `name`

lms/services/course.py
```diff
@@ -2,23 +2,22 @@
...
 
-    def get_or_create(self, authority_provided_id):
-        """
-        Add the current course to the `course` table if it's not there already.
-
-        :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
-        """
-        return self._get(authority_provided_id) or self._create(authority_provided_id)
+    def get_or_create(self, authority_provided_id, context_id, name, extra=None):
+        """Add the current course to the `course` table if it's not there already."""
```
marcospri, line 15: `grouping` table

lms/services/course.py
```diff
@@ -54,9 +90,12 @@ def _create(self, authority_provided_id):
             course_settings.set("canvas", "sections_enabled", False)
 
         course = Course(
-            consumer_key=self._consumer_key,
+            application_instance_id=self._application_instance.id,
```
marcospri, line 93: Use the instance directly here, not the ID (to allow this to work for newly created AIs

lms/resources/lti_launch.py
```diff
@@ -23,6 +22,17 @@ def __init__(self, request):
...
+    def get_or_create_course(self):
+        params = self._request.parsed_params
+        course = self._request.find_service(name="course").get_or_create(
+            self.h_group.authority_provided_id,
+            params["context_id"],
+            params["context_title"],
+            extra=self.course_extra,
+        )
+
+        return course
```
seanh, line 34: I think we probably want this to be a service method on either `CourseService` or `GroupingService`, rather than an `LTILaunchResource` method.

The problem with `LTILaunchResource` is that it's only available during requests that use `LTILaunchResource` as the `context` (i.e. LTI launch requests). It's not available for example during API requests made by the backend to the frontend. Services are available to all requests though.

seanh (changes requested):
I've added some comments but I think we might want to break this up into multiple smaller PRs so that we don't make any mistakes. Maybe:

One PR to add the new `Grouping` table.

One to rename `Course` to `LegacyCourse`.

One to start writing courses to the new `Grouping` table (while still also writing to and reading from the legacy `Course` table).

One to stop writing to the legacy `Course` table.

One to start reading from the new `Grouping` table instead of `Course`, but with a fallback to reading from `Course`.

One to start copying rows from `Course` to `Grouping` and deleting them from `Course` (this is a risky change).

I'm not sure whether that exact breakdown will be the one that makes sense, but something like that.

lms/services/course.py
```diff
@@ -2,23 +2,22 @@
 from copy import deepcopy
 
 from lms.models import Course, CourseGroupsExportedFromH
+from lms.models import _Course as LegacyCourse
 
 
 class CourseService:
```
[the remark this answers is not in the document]
marcospri: I think it makes sense that stays as Course, It uses both (new)Course and LegacyCourse but there's nothing legacy about creating/getting courses

lms/resources/lti_launch.py
marcospri: This is get_or_create_course and just calls course_service.get_or_create with the right argument  (during a lti_launch) I think it could stay here but I'll add a docstring to see if it make sense.

PR #2711, opened 2021-05-19
Deduplicate `requests` usage with a new service

seanh:
Problem
-------

There's a bit of logic that's repeated in a few places about using
`requests` to send a request, receive the response, and validate the
response using a `marshmallow` schema, with associated error handling
and tests, re-using a TCP session, making sure appropriate timeouts are
always used, etc. This is a basic version of the logic that's repeated:

```python
session = requests.Session()

try:
    response = session.request(method, url, timeout=9, ...)
    response.raise_for_status()
except requests.RequestException as err:
    raise CustomHTTPErrorClass(response) from err

validated_data = None

if schema:
    try:
        validated_data = schema(response).parse()
    except lms.validation.ValidationError as err:
        raise CustomHTTPValidationErrorClass(response) from err

return (response, validated_data)
```

This logic (or very similar) is repeated in `HAPIClient`,
`LTIOutcomesService` and `CanvasAPIClient` (in the `BasicClient` helper
class):

* https://github.com/hypothesis/lms/blob/b995d470acf28f70c342b03d0d4acd45b4c508c1/lms/services/h_api.py#L83-L99
* https://github.com/hypothesis/lms/blob/b995d470acf28f70c342b03d0d4acd45b4c508c1/lms/services/lti_outcomes.py#L101-L119
* https://github.com/hypothesis/lms/blob/b995d470acf28f70c342b03d0d4acd45b4c508c1/lms/services/canvas_api/_basic.py#L85-L95

The upcoming `BlackboardAPIClient` also needs to send HTTP requests
using `requests` and validate the responses using `marshmallow` in
exactly the same way, but that would mean duplicating this similar code
and tests for a fourth time.

Solution
--------

Add a new `HTTPService` to do the send-request-and-validate-response
bit. This will be implemented once, correctly, in `HTTPService` and
thoroughly unit tested there. Other services will just need to use a
very simple interface that `HTTPService` will provide, and their unit
tests will just need to work with a mock of that very simple interface.
A large amount of code and tests will be simplified and duplication
removed. Our code's behaviour (e.g. robustness) will be improved because
send-request-and-validate-response will be implemented and tested once
carefully and correctly, instead of several times with varying degrees
of correctness (e.g. timeouts missing in some places etc).

This PR adds the `HTTPService` and refactors `LTIOutcomesClient` and
`HAPI` and their tests to use `HTTPService`.

The new `BlackboardAPIClient` that'll use `HTTPService` will come in a
later PR.

Refactoring `CanvasAPIClient` to use `HTTPService` will also come
separately: that ones a bit more work.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -218,3 +218,15 @@ class CanvasFileNotFoundInCourse(ServiceError):
     def __init__(self, file_id):
         self.details = {"file_id": file_id}
         super().__init__(self.details)
+
+
+class HTTPError(ServiceError):
+    """A problem with an HTTP request to an external service."""
+
+    def __init__(self, response=None):
+        super().__init__(response)
+        self.response = response
+
+
+class HTTPValidationError(HTTPError):
+    """An invalid HTTP response from an external service."""
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -1,11 +1,9 @@
 """The H API service."""
 
-import requests
 from h_api.bulk_api import BulkAPI, CommandBuilder
-from requests import RequestException
 
 from lms.models import HUser
-from lms.services import HAPIError
+from lms.services import HAPIError, HTTPError
 
 __all__ = ["HAPI"]
 
@@ -25,6 +23,7 @@ class HAPI:
         self._authority = settings["h_authority"]
         self._http_auth = (settings["h_client_id"], settings["h_client_secret"])
         self._base_url = settings["h_api_url_private"]
+        self._http_service = request.find_service(name="http")
 
     def execute_bulk(self, commands):
         """
@@ -81,7 +80,7 @@ class HAPI:
             request_args["data"] = body
 
         try:
-            response = requests.request(
+            response = self._http_service.request(
                 method=method,
                 url=self._base_url + path.lstrip("/"),
                 auth=self._http_auth,
@@ -89,9 +88,7 @@ class HAPI:
                 headers=headers,
                 **request_args,
             )
-            response.raise_for_status()
-
-        except RequestException as err:
+        except HTTPError as err:
             response = getattr(err, "response", None)
 
             raise HAPIError("Connecting to Hypothesis failed", response) from err
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -84,7 +84,6 @@ class HAPI:
                 method=method,
                 url=self._base_url + path.lstrip("/"),
                 auth=self._http_auth,
-                timeout=10,
                 headers=headers,
                 **request_args,
             )
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -17,6 +17,21 @@ class HTTPService:
         # See https://docs.python-requests.org/en/latest/user/advanced/#session-objects
         self._session = _session or requests.Session()
 
+    def get(self, *args, **kwargs):
+        return self.request("GET", *args, **kwargs)
+
+    def put(self, *args, **kwargs):
+        return self.request("PUT", *args, **kwargs)
+
+    def post(self, *args, **kwargs):
+        return self.request("POST", *args, **kwargs)
+
+    def patch(self, *args, **kwargs):
+        return self.request("PATCH", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        return self.request("DELETE", *args, **kwargs)
+
     def request(
         self,
         method,
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -1,0 +1,161 @@
+import requests
+from requests import RequestException
+
+from lms.services.exceptions import HTTPError, HTTPValidationError
+from lms.validation import ValidationError
+
+
+class HTTPService:
+    """Send HTTP requests with `requests` and receive the responses."""
+
+    def __init__(self, _session=None):
+        # A requests session is used so that cookies are persisted across
+        # requests and urllib3 connection pooling is used (which means that
+        # underlying TCP connections are re-used when making multiple requests
+        # to the same host, e.g. pagination).
+        #
+        # See https://docs.python-requests.org/en/latest/user/advanced/#session-objects
+        self._session = _session or requests.Session()
+
+    def request(
+        self,
+        method,
+        url,
+        params=None,
+        data=None,
+        json=None,
+        headers=None,
+        auth=None,
+        timeout=9,
+        schema=None,
+        **kwargs,
+    ):  # pylint:disable=too-many-arguments
+        """
+        Send a request with `requests` and return the requests.Response object.
+
+        Also supports validating the response with a `marshmallow` schema. If a
+        `schema` argument is given the response will be validated using the
+        schema and the schema output will be added to the returned response as
+        response.validated_data.
+
+        The schema must be a RequestsResponseSchema sub-class.
+
+        HTTPValidationError will be raised if schema validation fails.
+
+        :param method: The HTTP method to use, one of "GET", "PUT", "POST",
+            "PATCH", "DELETE", "OPTIONS" or "HEAD"
+
+        :param url: The URL to request
+
+        :param params: Query params to append to the URL.
+
+            This can be a dict or list of tuples and it'll be serialized into a
+            query string.
+
+            Or it can be a byte string and it'll be used directly as the query
+            string.
+
+        :param data: Form data to send in the body of the request.
+
+            This can be a dict or list of tuples and it'll be serialized into a
+            form body.
+
+            Or it can be bytes or a file-like object (e.g. a JSON byte string)
+            it'll be used directly as the request body.
+
+        :param json: Data to send in the body of the request as JSON.
+
+            Using this changes the request's content-type to application/json.
+
+            This can't be used at the same time as the `data` argument.
+
+        :type json: Any JSON-serializable object (e.g. a list or dict)
+
+        :param headers: Headers to send in the request.
+
+            Header values should be byte strings not unicode.
+
+        :type headers: dict
+
+        :param auth: Authorization to send in the request's Authorization header.
+
+            This can be a (user, pass) 2-tuple which will be serialized using
+            HTTP Basic Authentication.
+
+            Or it can be a callable in order to implement other authentication
+            schemes. `requests` itself and libraries like `requests_oauthlib`
+            provide auth callables that you can use or you can write your own.
+
+            See: https://docs.python-requests.org/en/master/user/authentication/
+
+        :param timeout: How long (in seconds) to wait before raising an error.
+
+            This can be a (connect_timeout, read_timeout) 2-tuple or it can be
+            a single float that will be used as both the connect and read
+            timeout.
+
+            Note that the read_timeout is *not* a time limit on the entire
+            response download. It's a time limit on how long to wait *between
+            bytes from the server*. The entire download can take much longer.
+
+        :param schema: A schema class to use to validate the response
+        :type schema: lms.validation.RequestsResponseSchema
+
+        :param kwargs: Any other keyword arguments will be passed directly to
+            requests.Session().request():
+            https://docs.python-requests.org/en/latest/api/#requests.Session.request
+
+        :raise HTTPError: If sending the request or receiving the response
+            fails (DNS failure, refused connection, timeout, too many
+            redirects, etc).
+
+            The original exception from requests will be available as
+            HTTPError.__cause__.
+
+            In this case HTTPError.response will be None.
+
+        :raise HTTPError: If an error response (4xx or 5xx) is received.
+
+            The original exception from requests will be available as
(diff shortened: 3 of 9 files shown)


lms/services/http.py
```diff
@@ -0,0 +1,161 @@
...
+            The error response will be available as HTTPError.response.
+
+        :raise HTTPValidationError: If a `schema` argument is given and the
+            response fails validation with the schema.
+
+            The original lms.validation.ValidationError will be available as
+            HTTPValidationError.__cause__.
+
+            The invalid response will be available as
+            HTTPValidationError.response.
```
seanh, line 131: Long docstring mostly because `requests.Session.request()` has a lot of arguments and I picked out the arguments that our code already uses or that I think we're likely to want to use and documented them here. I think this is useful because `requests`'s own documentation is kind of hard to navigate and you often have to look in multiple places in the `requests` docs to find all the pertinent information about one parameter. So this docstring acts as a kind of cheat sheet.

It also documents our own `schema` parameter and setting `validated_data`, and our custom exception-raising behaviour (the different types of exception raised and setting `__cause__` and `response`).

lms/services/exceptions.py
```diff
@@ -218,3 +218,15 @@ class CanvasFileNotFoundInCourse(ServiceError):
...
+class HTTPError(ServiceError):
+    """A problem with an HTTP request to an external service."""
+
+    def __init__(self, response=None):
+        super().__init__(response)
+        self.response = response
+
+
+class HTTPValidationError(HTTPError):
+    """An invalid HTTP response from an external service."""
```
seanh, line 232: We normally define our `<APP>/<FOO>/exceptions.py` files, rather than in the individual `services/http.py` file. This is an attempt to achieve a balance between exception specificity versus exception reusability.
See https://stackoverflow.com/c/hypothesis/questions/295. In this case these two exceptions are currently only raised by `HTTPService`, but I put them here anyway because 1. It's where put put all our services exceptions; 2. It encourages re-use, if we have another HTTP-related service in future.

lms/services/h_api.py
```diff
@@ -81,17 +80,15 @@ def _api_request(self, method, path, body=None, headers=None):
...
                 url=self._base_url + path.lstrip("/"),
                 auth=self._http_auth,
                 timeout=10,
                 headers=headers,
                 **request_args,
             )
-            response.raise_for_status()
-
-        except RequestException as err:
+        except HTTPError as err:
```
seanh, line 91: :man_shrugging: Not much to see here, the new `HTTPService`'s API is very similar to the `requests` API that `HAPI` was already using. This does mean that `HAPI` has picked up some `HTTPService` features that it was lacking before, e.g. using a requests session.

lms/services/http.py
```diff
@@ -0,0 +1,161 @@
...
+        data=None,
+        json=None,
+        headers=None,
+        auth=None,
+        timeout=9,
+        schema=None,
+        **kwargs,
+    ):  # pylint:disable=too-many-arguments
+        """
+        Send a request with `requests` and return the requests.Response object.
```
seanh, line 34: Note that no attempt is made to hide `requests` here: `HTTPService.request()` accepts arbitrary `kwargs` that are passed directly to `requests.Session.request()` and it also returns `requests.Response` objects directly. We _do_ wrap `requests`'s many different exception classes in our own two exception classes that have particular semantic meaning for us (`HTTPError` is for a problem with sending the request or receiving the successful response, `HTTPValidationError` is for a problem with validating the response).

lms/services/http.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+            response.raise_for_status()
+        except RequestException as err:
+            raise HTTPError(response) from err
+
+        if schema:
+            try:
+                response.validated_data = schema(response).parse()
+            except ValidationError as err:
+                raise HTTPValidationError(response) from err
```
seanh, line 155: The schema argument isn't actually used by either `HAPI` or `LTIOutcomesClient`.

`HAPI` is sending requests to our own service so maybe we can more easily get away with just trusting the responses. But also I don't think it actually parses the response bodies anyway.

`LTIOutcomesClient` has had loads of customer-facing problems dealing with invalid or unexpected responses and would clearly benefit from better, schema-based validation. But that's beyond the scope of this PR. And LTI outcomes responses are XML so kind of a pain.

tests/unit/lms/services/http_test.py
```diff
@@ -0,0 +1,147 @@
...
+            "GET", url, body='{"test_response_key": "TEST_RESPONSE_VALUE"}'
+        )
+
+    @pytest.fixture
+    def session(self):
+        return create_autospec(requests.Session, instance=True, spec_set=True)
+
+    @pytest.fixture
+    def svc(self):
+        return HTTPService()
```
seanh, line 142: This is now much more thoroughly tested than any of the now-removed `requests`-using code was

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -97,32 +104,37 @@ def test_record_result_requires_dict_result(self, svc, hook_result):
                 self.GRADING_ID, pre_record_hook=lambda *args, **kwargs: hook_result
             )
 
-    def test_it_signs_request_with_oauth1(self, svc, requests, oauth1_service):
-        requests.post.side_effect = OSError()
+    def test_it_signs_request_with_oauth1(self, svc, http_service, oauth1_service):
+        http_service.request.side_effect = OSError()
```
seanh, line 108: I'm not sure why the original test was configuring `requests.post()` to raise `OSError` here, I don't think `requests.post` ever raises that. It seems like the test just wanted the `requests.post()` call to crash because it doesn't need the code after that call to execute because all it wants to do is assert about what `requests.post()` call was made :man_shrugging: I decided not to dig into it, and to just leave it the same

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -97,32 +104,37 @@ def test_record_result_requires_dict_result(self, svc, hook_result):
...
-            timeout=Any(),
         )
 
-    def test_requests_fail_if_http_status_is_error(self, svc, respond_with):
-        respond_with(status=400)
+    def test_requests_fail_if_the_third_party_request_fails(self, svc, http_service):
+        http_service.request.side_effect = HTTPError
+
+        with pytest.raises(LTIOutcomesAPIError):
+            svc.read_result(self.GRADING_ID)
```
seanh, line 127: `test_requests_fail_if_http_status_is_error()` was removed because checking the HTTP status code and raising if it's an error (4xx or 5xx) is no longer `LTIOutcomesService`'s job. `HTTPService` handles that now. `LTIOutcomesService` just needs to deal with `HTTPService` raising `HTTPError`. I added this new test for that

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -160,27 +172,19 @@ def test_requests_fail_and_no_description_returned(self, svc, respond_with):
         with pytest.raises(LTIOutcomesAPIError, match="LTI outcome request failed"):
             svc.read_result(self.GRADING_ID)
 
-    def test_it_gracefully_handles_RequestException(self, requests, svc):
-        requests.post.side_effect = RequestException
-
-        with pytest.raises(
-            LTIOutcomesAPIError, match="Error calling LTI Outcomes service"
-        ):
-            svc.read_result(self.GRADING_ID)
```
seanh, line 169: This is now covered by `test_requests_fail_if_the_third_party_request_fails()` above. Previously `LTIOutcomesClient` had to worry about both error responses (4xx or 5xx) and `requests` exceptions like `ConnectionError` etc. `HTTPService` now deals with that and raises `HTTPError` in either case, so only one test is needed in `LTIOutcomesClient`.

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -97,32 +104,31 @@ def test_record_result_requires_dict_result(self, svc, hook_result):
...
             data=Any(),
             headers=Any(),
             auth=oauth1_service.get_client.return_value,
-            timeout=Any(),
         )
 
-    def test_requests_fail_if_http_status_is_error(self, svc, respond_with):
-        respond_with(status=400)
+    def test_requests_fail_if_the_third_party_request_fails(self, svc, http_service):
+        http_service.request.side_effect = HTTPError
```
seanh, line 124: `test_requests_fail_if_http_status_is_error()` was removed because checking the HTTP status code and raising if it's an error (4xx or 5xx) is no longer `LTIOutcomesService`'s job. `HTTPService` handles that now. `LTIOutcomesService` just needs to deal with `HTTPService` raising `HTTPError`. I added this new test for that

tests/unit/lms/services/lti_outcomes_test.py
```diff
@@ -160,27 +166,19 @@ def test_requests_fail_and_no_description_returned(self, svc, respond_with):
...
-
-        with pytest.raises(
-            LTIOutcomesAPIError, match="Error calling LTI Outcomes service"
-        ):
-            svc.read_result(self.GRADING_ID)
-
     @classmethod
-    def sent_body(cls):
-        return xmltodict.parse(httpretty.last_request().body)
+    def sent_body(cls, http_service):
```
seanh, line 170: I added `http_service` as an argument to this classmethod and the others below, which means the tests that call these classmethods now need to pass in `http_service`. Alternatively I could have turned the classmethods into fixtures that could get `http_service` themselves using fixture injectin

lms/services/http.py
```diff
@@ -0,0 +1,161 @@
...
+    def request(
+        self,
+        method,
+        url,
+        params=None,
+        data=None,
+        json=None,
+        headers=None,
+        auth=None,
+        timeout=9,
```
marcospri, line 29: `timeout=timeout`

marcospri, line 29: Maybe a tuple is a better default here (connect_timeout, read_timeout)

https://2.python-requests.org/en/master/user/advanced/#timeouts

lms/services/http.py
```diff
@@ -0,0 +1,161 @@
...
+            raise HTTPError(response) from err
+
+        if schema:
+            try:
+                response.validated_data = schema(response).parse()
+            except ValidationError as err:
+                raise HTTPValidationError(response) from err
+
+        return response
+
```
marcospri, line 158: Would adding `post` and `get` utility functions here make sense?

Doc string could point to the docs on the main request func and the actual function bodies be pretty minimal

https://github.com/psf/requests/blob/master/requests/api.py#L64

seanh: Yeah, I think it could be easily added

lms/services/h_api.py
```diff
@@ -81,17 +80,15 @@ def _api_request(self, method, path, body=None, headers=None):
             request_args["data"] = body
 
         try:
-            response = requests.request(
+            response = self._http_service.request(
                 method=method,
                 url=self._base_url + path.lstrip("/"),
                 auth=self._http_auth,
                 timeout=10,
```
marcospri, line 87: timeout=10 is the default, unless clients have a good reason to overwrite it I'd leave it out

lms/services/http.py
seanh: Yeah, it might be better for readability to do `timeout=(10,10)` as it just clarifies, from the default value, that there are actually two separate timeouts not one.

Would we want the two timeouts to have different values?

I also wonder whether we should change the default from `9` to `10` because the requests docs say:

>  It’s a good practice to set connect timeouts to **slightly larger than a multiple of 3**, which is the default TCP packet retransmission window.
> 
> --- https://docs.python-requests.org/en/master/user/advanced/#timeouts

lms/services/h_api.py
seanh: Yeah, I think maybe make `10` the default (it's currently `9`) then remove it here

lms/services/http.py
seanh: Added these: https://github.com/hypothesis/lms/pull/2711/commits/7ca7666d4e3e41cee9201b2bd0838b444a9c4610

It's actually possible to do this in a fancy way in `HTTPService.__init__()`:

```python
for http_verb in ("get", "put", "post", "patch", "delete"):
    setattr(self, http_verb, partial(self.request, http_verb))
```

But I don't think it's worth it: it just means that `mock` fails to detect the `get()`, `put()` etc methods and you then have to work around that.

lms/services/h_api.py
seanh: Done: https://github.com/hypothesis/lms/pull/2711/commits/d54c0796f3b5b15119b7b59ea00dadff0ce54f1e

lms/services/http.py
seanh: Done: https://github.com/hypothesis/lms/pull/2711/commits/d54c0796f3b5b15119b7b59ea00dadff0ce54f1e

marcospri (approved):
I don't know what the "right" value for the timeouts is but at least there's now a clear place to tune it

PR #2713, opened 2021-05-20
Remove flag icon from annotation cards for LMS users

esanzgar:
Turn off the `allowFlagging` option to remove the flag icon from
annotation cards in the LMS context.

Relates to https://github.com/hypothesis/client/pull/3424


robertknight (approved):
LGTM

PR #2715, opened 2021-05-20
Add UI playground

LMS007:
### Commits

Add ui-playground …

- Add /ui-playground route and view
- Add static routes for js and css bundle files that ui-playground needs from the /build folder
- Add basic ui-playground html template
- Add js and css bundles for ui-playground app from frontend-shared
- Add eslint file to ignore arrow functions --otherwise causing errors in IDE

----------

http://localhost:8001/ui-playground

This is a prototype for adding the basic ui-playground to LMS and leveraging pyramid instead of using express.  This adds a new route called /ui-playground and also allows several static resources to be served from the temporary build/ dir that our frontend code uses to drop bundled resources.

This only sets up the ui playground but does not add any custom LMS pages or components. I could imagine we may have dialogs or pages that are specific to LMS eventually. 

I'd like to seek some help from the backend folks to see if I have done this in a reasonable way.

related to https://github.com/hypothesis/lms/issues/2675

TODO:

- [x] Add py test coverage for new view

### DIFF
--- a/lms/assets.ini
+++ b/lms/assets.ini
@@ -15,3 +15,9 @@ lms_css =
 
 reports_css =
   styles/reports.css
+
+ui_playground_css =
+  styles/ui-playground.css
+
+ui_playground_js =
+  scripts/ui-playground.bundle.js
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -5,6 +5,7 @@ def includeme(config):
     config.add_route("assets", "/assets/*subpath")
     config.add_route("status", "/_status")
     config.add_route("favicon", "/favicon.ico")
+    config.add_route("ui-playground", "/ui-playground/*remainder")
 
     config.add_route("login", "/login")
     config.add_route("logout", "/logout")
--- a/lms/templates/ui-playground.html.jinja2
+++ b/lms/templates/ui-playground.html.jinja2
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>UI component pattern library</title>
+  {% for url in asset_urls("ui_playground_css") %}
+    <link rel="stylesheet" href="{{ url }}"/>
+  {% endfor %}
+</head>
+<body>
+  <div id="app"></div>
+  {% for url in asset_urls("ui_playground_js") %}
+    <script src="{{ url }}"></script>
+  {% endfor %}
+</body>
+</html>
--- a/lms/views/ui_playground.py
+++ b/lms/views/ui_playground.py
@@ -1,0 +1,14 @@
+from pyramid.httpexceptions import HTTPNotFound
+from pyramid.view import notfound_view_config, view_config
+
+
+@view_config(
+    route_name="ui-playground", renderer="lms:templates/ui-playground.html.jinja2"
+)
+def ui_playground(_request):
+    return {}
+
+
+@notfound_view_config(path_info="/ui-playground", append_slash=True)
+def notfound(_request):
+    return HTTPNotFound()
--- a/scripts/gulp/create-style-bundle.js
+++ b/scripts/gulp/create-style-bundle.js
@@ -8,9 +8,16 @@ const postcss = require('postcss');
 const sass = require('sass');
 
 /**
+ * @typedef Options
+ * @prop {string} input - Input file path
+ * @prop {boolean} minify - Whether to minify the generated bundle
+ * @prop {string} output - Output file path
+ */
+
+/**
  * Compile a SASS file and postprocess the result.
  *
- * @param {options} options - Object specifying the input and output paths and
+ * @param {Options} options - Object specifying the input and output paths and
  *                  whether to minify the result.
  * @return {Promise} Promise for completion of the build.
  */
--- a/tests/unit/lms/views/test_ui_playground.py
+++ b/tests/unit/lms/views/test_ui_playground.py
@@ -1,0 +1,11 @@
+from lms.views import ui_playground
+
+
+def test_ui_playground(pyramid_request):
+    response = ui_playground.ui_playground(pyramid_request)
+    assert response == {}
+
+
+def test_not_found_view(pyramid_request):
+    response = ui_playground.notfound(pyramid_request)
+    assert response.status_code == 404
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -94,6 +94,10 @@ var bundles = [
     name: 'browser_check',
     entry: './lms/static/scripts/browser_check/index.js',
   },
+  {
+    name: 'ui-playground',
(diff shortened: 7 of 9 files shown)


lms/app.py
```diff
@@ -100,6 +100,9 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.include("lms.tweens")
     config.add_static_view(name="export", path="lms:static/export")
     config.add_static_view(name="static", path="lms:static")
+    # For ui-playground only
```
LMS007, line 103: I needed to expose these two static files which get built to this folder, is this the appropriate way to do it?

lms/routes.py
```diff
@@ -5,6 +5,8 @@ def includeme(config):
     config.add_route("assets", "/assets/*subpath")
     config.add_route("status", "/_status")
     config.add_route("favicon", "/favicon.ico")
+    config.add_route("ui-playground-root", "/ui-playground")
+    config.add_route("ui-playground", "/ui-playground/*subpath")
```
LMS007, line 9: Our UI has client side routes so I wanted to capture them all, is this the right way?
e.g
http://localhost:8001/ui-playground/foo
http://localhost:8001/ui-playground/bar

lms/routes.py
```diff
@@ -5,6 +5,8 @@ def includeme(config):
     config.add_route("assets", "/assets/*subpath")
     config.add_route("status", "/_status")
     config.add_route("favicon", "/favicon.ico")
+    config.add_route("ui-playground-root", "/ui-playground")
```
LMS007, line 8: I did not want a trailing slash to be a requirement so I added a separate route w/o the slash, I was not sure if there was a simpler way to do this. This also requires a second view :\

lms/templates/ui-playground.html.jinja2
```diff
@@ -0,0 +1,13 @@
...
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>UI component pattern library</title>
+  <link rel="stylesheet" href="/styles/ui-playground.css">
+</head>
+<body>
+  <div id="app"></div>
+  <script src="/scripts/ui-playground.bundle.js"></script>
+</body>
+</html>
```
LMS007, line 13: todo: add newline

scripts/gulp/create-style-bundle.js
```diff
@@ -7,10 +7,17 @@ const autoprefixer = require('autoprefixer');
 const postcss = require('postcss');
 const sass = require('sass');
 
+/**
```
LMS007, line 10: These changes are copied from frontend-shared, but not required at this time. Its worth noting that this file is still not typechecked and the sibling file create-bundle.js is still outdated and needs more refactoring to sync it with frontend-shared. I would suggest this is a TODO for later.

tests/unit/lms/views/test_ui_playground.py
```diff
@@ -0,0 +1,6 @@
+from lms.views import ui_playground
+
+
+def test_ui_playground(pyramid_request):
```
LMS007, line 4: This test is pretty useless because the return value is not important, but it's here for coverage. I think a better test would assert the routh path itself returns 200 okay, but I'm not sure how to do that. If its not too important, we could just leave this as is?

lms/routes.py
seanh: I think just `/ui-playground/{subpath}` would work but only if `subpath` does not contain any `/`'s

lms/views/ui_playground.py
```diff
@@ -0,0 +1,11 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    route_name="ui-playground", renderer="lms:templates/ui-playground.html.jinja2"
+)
+@view_config(
+    route_name="ui-playground-root", renderer="lms:templates/ui-playground.html.jinja2"
+)
+def ui_playground(request):  # pylint: disable=unused-argument
```
seanh, line 10:
```suggestion
def ui_playground(_request):
```

lms/views/ui_playground.py
```diff
@@ -0,0 +1,11 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    route_name="ui-playground", renderer="lms:templates/ui-playground.html.jinja2"
+)
+@view_config(
+    route_name="ui-playground-root", renderer="lms:templates/ui-playground.html.jinja2"
+)
```
seanh, line 9: If you use the `append_slash` thing I think you'll only need one `@view_config`

lms/routes.py
robertknight: Reading https://stackoverflow.com/a/15708788/434243 it looks to me like the idiomatic thing here is to declare only the route with a trailing slash (the one with the `/ui-playground/*subpath` pattern) and then handle requests for `/ui-playground` by making it redirect to the slash-appended route. The `/admin/` route already does this. You can access it at `/admin` and `/admin/`. Have a look in `admin.py` to see how this works.

lms/routes.py
robertknight: Using a `*remainder` pattern is the right approach here.

scripts/gulp/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
+    "prefer-arrow-callback": "off"
```
robertknight, line 3: The lint checks were passing on master (in CI builds at least) in previous commits. Has something changed on this branch or do you see an issue on master as well?

tests/unit/lms/views/test_ui_playground.py
robertknight: This trivial test makes sure the route doesn't crash at least and serves as a placeholder if we want to add more substantial logic later. I think that's fine for now. Checking for the HTTP status requires a functional test that involves more of the stack. I don't think its hugely important though given that this is not something that end users will be accessing.

robertknight (changes requested):
Hi Kyle,

I left some feedback regarding generation of CSS and JS asset URLs and also handling the `/ui-playground` vs. `/ui-playground/` issue: you want the former to redirect to the latter and there is an example of how to do this in the form of the `/admin` route.

lms/app.py
LMS007: Will do

scripts/gulp/.eslintrc
LMS007: It was related to my IDE , I removed the rule.

lms/assets.ini
```diff
@@ -15,3 +15,9 @@ lms_css =
 
 reports_css =
   styles/reports.css
+
+ui_playground_css =
+  styles/ui-playground.css
+
+ui_playground_js =
+    scripts/ui-playground.bundle.js
```
robertknight, line 23: Can you correct the indentation here so that it matches the other entries.

robertknight (approved):
I requested a couple of very minor changes, but otherwise LGTM.

PR #2717, opened 2021-05-21
Merge `contentItemFor*` functions

robertknight:
Combine the `contentItemFor<ContentType>` functions into a single
`contentItemForContent` function which accepts a `Content` object as an
argument, as well as additional query params to add to the launch URL.

Combining these functions removes the need to switch on the content type in other
places in the code.

The support for adding additional query params will be used to add the
`group_set` query param in assignments that are configured to use
groups.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/test/FilePickerFormFields-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/FilePickerFormFields-test.js
@@ -1,11 +1,7 @@
 import { mount } from 'enzyme';
 import { createElement } from 'preact';
 
-import {
-  contentItemForLmsFile,
-  contentItemForUrl,
-  contentItemForVitalSourceBook,
-} from '../../utils/content-item';
+import { contentItemForContent } from '../../utils/content-item';
 
 import FilePickerFormFields from '../FilePickerFormFields';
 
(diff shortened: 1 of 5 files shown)


lms/static/scripts/frontend_apps/utils/content-item.js
```diff
@@ -1,65 +1,66 @@
+/** @typedef {import('../api-types').File} File */
```
esanzgar, line 1: You have added these type definitions before the first import. I have not seen this before in our code, even for types that are imported and used in other files.

lms/static/scripts/frontend_apps/utils/test/content-item-test.js
```diff
@@ -1,73 +1,47 @@
...
+  [
+    {
+      content: { type: 'url', url: 'https://example.com?param_a=1&param_b=2' },
+      expectedURL: `${ltiLaunchURL}?url=${encodeURIComponent(
+        'https://example.com?param_a=1&param_b=2'
+      )}`,
+    },
+    {
+      content: { type: 'file', file: { id: 'foobar' } },
+      expectedURL: `${ltiLaunchURL}?canvas_file=true&file_id=foobar`,
```
esanzgar, line 15: We are relying in the order that `stringify` uses to serialise the parameters.

lms/static/scripts/frontend_apps/components/test/FilePickerFormFields-test.js
```diff
@@ -44,35 +40,13 @@ describe('FilePickerFormFields', () => {
...
-    it('renders content_items field for VitalSource book content', () => {
-      const formFields = createComponent({
-        content: { type: 'vitalsource', bookID: 'test-book', cfi: 'test-cfi' },
-      });
-      const contentItems = getContentItem(formFields);
-      assert.deepEqual(
-        contentItems,
-        contentItemForVitalSourceBook(launchURL, 'test-book', 'test-cfi')
-      );
+      assert.deepEqual(contentItems, contentItemForContent(launchURL, content));
```
esanzgar, line 49: Instead of this assertion, I suggest to check that the hidden form has the expected value property.

esanzgar (approved):
I made a suggestion on one test to check in the hidden input the correct `value` property.

lms/static/scripts/frontend_apps/utils/content-item.js
robertknight: Ah, indeed. I'll move the typedefs below the imports for consistency.

lms/static/scripts/frontend_apps/components/test/FilePickerFormFields-test.js
robertknight: I'm not quite sure I've understood. Checking the `value` property of the `content_item` field is what this test is doing, albeit with some indirection. Maybe you were expecting something like:

```js
const value = wrapper.find('[name="content_item"]'.prop('value');
assert.equal(value, "some string");
```

I'm using `contentItemForContent(...)` to generate the expected value since the structure is complex enough to make it worthwhile. I've inlined the `getContentItem(...)` helper though to make the test a bit more direct.

esanzgar: Indeed.

PR #2729, opened 2021-05-24
Fix some broken CanvasAPIClient tests

seanh:
Problem
-------

A bunch of tests in `tests/unit/lms/services/canvas_api/` aren't
actually testing what they think they're testing. If you break the
`tests/unit/lms/services/canvas_api_client/conftest.py::http_session()`
fixture:

```diff
diff --git a/tests/unit/lms/services/canvas_api/conftest.py b/tests/unit/lms/services/canvas_api/conftest.py
index 984f8113..9cf46dc6 100644
--- a/tests/unit/lms/services/canvas_api/conftest.py
+++ b/tests/unit/lms/services/canvas_api/conftest.py
@@ -18,9 +18,7 @@ def http_session(patch):
     session = session()

     def set_response(json_data=None, raw=None, status_code=200):
-        session.send.return_value = factories.requests.Response(
-            json_data=json_data, raw=raw, status_code=status_code
-        )
+        session.send.side_effect = RuntimeError()

     session.set_response = set_response
```

Then these tests still pass even though they use the `http_session`
fixture which we've now hacked to cause all HTTP requests to crash:

    # This still passes.
    $ tox -q tests/unit/lms/services/canvas_api/client_test.py::TestMetaBehavior

What's going wrong, in detail
-----------------------------

The Canvas API client tests deviate from our usual test style in ways
that're creating problems:

1. The amount of code being executed under a single unit test is too
   large. Tests for `CanvasAPIClient` use the real `AuthenticatedClient`
   and `BasicClient`. This makes whats-actually-happening when a given
   unit test much more difficult to trace, and much more likely to
   contain subtle problems, because there's just so much more going on.

   Our usual approach is to unit test one module at a time. The unit
   tests for `foo.py` will mock out any other modules from our own code
   that `foo.py` imports. This means that you can open `foo.py` and
   `foo_test.py` side-by-side in an editor and you have all the code
   that those tests are executing in view (more or less: modulo shared
   fixtures and the like).

   The Canvas API client tests execute the entire `canvas_api_client/`
   directory at once.

2. As a sort of side-effect of (1) when they do need to mock something
   the Canvas API client tests take a different-from-usual approach to
   patching.

   Our normal approach is **patch names within the module being
   tested**. So `foo_test.py` will patch names within `foo.py`. For
   example if `foo.py` did `from lms import bar` then `foo_test.py`
   would patch `"lms.foo.bar"`. `foo_test.py` would _not_ patch
   `"lms.bar.gar"`. This is called "patching at source" and there's a
   very good reason why we don't do it. See this old GitHub comment for
   a detailed explanation:
   https://github.com/hypothesis/via/pull/222#discussion_r469865055
   The Canvas API client tests have essentially the same problem.

**We really want to avoid patching at source**. Patching at source very
easily leads to tests that aren't really testing what they think they're
testing. For example the test will try to set up some method to return a
broken response, but fail to actually set this up correctly. The test
will then assert that the method-under-test handles the broken response
without failing. The method-under-test will succeed, because the broken
response was never actually set up. The test will pass. In fact the
method-under-test could be broken and failing under the intended test
conditions, but the test could be passing anyway since the conditions
aren't actually set up. That is what was happening with the Canvas API
client tests (see below for detailed explanation) and with other
instances of source-patching that we've had.

**These issues are hard to detect and fix**. Source-patching makes it
easy for these kinds of test bugs to creep in, and once in they're
difficult to detect and to fix. Get a whole bunch of these in your test
suite and you have an infestation that can only be fixed by a large
investment of time by a developer who really understands pytest and
mock.

Let's look at one particular unit test in detail to explain what's
happening. We'll look at
`tests/unit/lms/services/canvas_api/client_test.py::TestMetaBehavior::test_methods_raise_CanvasAPIServerError_if_the_response_json_has_the_wrong_format`.
What happens:

1. The test depends on the `data_method` fixture, so pytest executes
   this fixture first:

https://github.com/hypothesis/lms/blob/136d816e74340008b0173a7f2d56663ac54c63e2/tests/unit/lms/services/canvas_api/client_test.py#L267-L270
   The `data_method` fixture is defined further down in the same test
   class:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/tests/unit/lms/services/canvas_api/client_test.py#L294-L298

2. The `data_method` fixture depends on the `canvas_api_client` fixture:

 
 
 https://github.com/hypothesis/lms/blob/136d816e74340008b0173a7f2d56663ac54c63e2/tests/unit/lms/services/canvas_api/client_test.py#L294-L295

   The `canvas_api_client` fixture is defined further down in the same
   test class:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/tests/unit/lms/services/canvas_api/client_test.py#L301-L303

3. The `canvas_api_client` fixture depends on the `authenticated_client` fixture:

   https://github.com/hypothesis/lms/blob/136d816e74340008b0173a7f2d56663ac54c63e2/tests/unit/lms/services/canvas_api/client_test.py#L301-L302

   **Unlike our usual test style** the Canvas API client tests have
   their own `conftest.py` file containing shared fixtures. The
   `authenticated_client` fixture is defined in
   `tests/unit/lms/services/canvas_api/conftest.py`. Also **unlike our
   usual test style** the `authenticated_client` fixture returns a real
   `AuthenticatedClient` object rather than a mock, even though this is
   a unit test for `CanvasAPIClient` not `AuthenticatedClient`:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/tests/unit/lms/services/canvas_api/conftest.py#L30-L38

4. The `authenticated_client` fixture depends on the `basic_client`
   fixture:

   https://github.com/hypothesis/lms/blob/136d816e74340008b0173a7f2d56663ac54c63e2/tests/unit/lms/services/canvas_api/conftest.py#L30-L31

   The `basic_client` fixture is also defined in `conftest.py`. Again,
   **unlike our usual test style** the `basic_client` fixture returns a
   real `BasicClient` object rather than a mock, even though this is a
   unit test for `CanvasAPIClient` not `BasicClient`:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/tests/unit/lms/services/canvas_api/conftest.py#L10-L12

5. So at this point, during pytest's execution of the `data_method`
   fixtures (and the other fixtures that `data_method` depends on
   directly or indirectly), **the `BasicClient` instance has been
   constructed**. `BasicClient.__init__()` constructs a
   `requests.Session` object:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/lms/services/canvas_api/_basic.py#L32-L41

   So at this point `lms.services.canvas_api._basic.Session` has already
   been called, and a real `requests.Session` object has already been
   constructed.

6. Going all the way back up to the actual unit test that's being
   executed (`test_methods_raise_CanvasAPIServerError_if_the_response_json_has_the_wrong_format`),
   the test also depends on the `http_session` fixture:

   https://github.com/hypothesis/lms/blob/136d816e74340008b0173a7f2d56663ac54c63e2/tests/unit/lms/services/canvas_api/client_test.py#L267-L270

   `http_session` is defined in `conftest.py`. **Unlike our usual test
   style** the unit tests for `CanvasAPIClient` now patch a name within
   a different module: the `http_session` fixture patches
   `"lms.services.canvas_api._basic.Session"` and sets up a mock HTTP
   response. This is the dreaded patch-at-source:

   https://github.com/hypothesis/lms/blob/c51e43d41f2ac0201ff2515de266f16da8d5d0f7/tests/unit/lms/services/canvas_api/conftest.py#L15-L27

7. And that's the problem: at this point in the execution of fixtures
   patching `"_basic.Session"` has no effect. As we've already seen in
   step 5 above, `BasicClient()` has already been instantiated and it
   has already called `Session()` and created a real `requests.Session`
   object.  Replacing `Session()` with a mock after `Session()` has
   already been called (and is not going to be called again) does
   nothing. The unit test (and several others like it) is effectively
   not using the `http_session` fixture. When these tests call
   `http_session.set_response({})` or
   `http_session.set_response(raw="[broken json")` these calls **have no
   effect**. The tests are not actually setting up the scenarios that
   they think they are setting up.

Solution
--------

Add the requests session as an optional argument to
`BasicClient.__init__()` and have the `basic_client` fixture depend on
the `http_session` fixture and pass the `http_session` fixture in to
`BasicClient`:

```python
class BasicClient:
    def __init__(self, canvas_host, session=None):
        self._session = session or Session()
        ...

    ...

...

@pytest.fixture
def basic_client(http_session):
    return BasicClient(..., session=http_session)
```

The call to `patch()` can now be removed from the `http_session`
fixture: it no longer needs to patch anything.

This means that whenever any test uses `basic_client` pytest will
execute `http_session` _first_ and the requests session that
`http_session` returns will be passed to `BasicClient`.

This has a couple of consequences:

1. It's no longer possible to get a `BasicClient` with a real requests
   session from the `basic_client` fixture. In practice all the tests
   still pass so it doesn't look like any tests need this.

2. The `http_session.set_response()` method is no longer possible.

   The `http_session` fixture used to add a `set_response()` method
   to the mock `requests.Session`:

   ```python
   @pytest.fixture
   def http_session():
       session = patch("lms.services.canvas_api._basic.Session")
       session = session()

       def set_response(json_data=None, raw=None, status_code=200):
           session.send.return_value = factories.requests.Response(
               json_data=json_data, raw=raw, status_code=status_code
           )

       session.set_response = set_response

       return session
    ```

    Using `create_autospec(..., spec_set=True)` means you can't _set_
    any attribute on the mock that doesn't exist on the real
    `requests.Session` class, so `session.set_response()` can no longer
    be added.

    We could just remove the `spec_set=True` but using it is a mock best
    practice for a reason: generally you want tests to fail if code
    either reads or writes an attribute that doesn't exist

    I decided to just get rid of `set_response()` and replace calls like
    `http_session.set_response(...)` with
    `http_session.send.return_value = factories.requests.Response(...)`.
    This is longer but it's also arguably more readable: I can see
    exactly what the test is doing, rather than having to look up a
    shared fixture in a `conftest.py` file to understand it. This is
    similar to other recent decisions we made about fixtures: in some
    cases it can actually be better for test readability to make the
    tests do more work.

    I decided not to but for completeness sake it would actually be
    possible to re-implement `set_response()` as a standalone fixture:

    ```python
    @pytest.fixture
    def set_response(http_session):

        def set_response(json_data=None, raw=None, status_code=200):
               json_data=json_data, raw=raw, status_code=status_code
           )

        return set_response
    ```

Longer-term: if the Canvas API client tests used our usual approach to
unit testing and patching they wouldn't have this problem. The amount of
code-under-test per test method would also be much smaller making the
tests easier to understand, the tests for different modules would be
more decoupled making refactoring easier, and coverage reporting would
be more useful.

### DIFF
--- a/lms/services/canvas_api/_basic.py
+++ b/lms/services/canvas_api/_basic.py
@@ -29,15 +29,17 @@ class BasicClient:
     PAGINATION_MAXIMUM_REQUESTS = 25
     """The maximum number of calls to make before giving up."""
 
-    def __init__(self, canvas_host):
+    def __init__(self, canvas_host, session=None):
         """
         Create a new BasicClient for making calls to the Canvas API.
 
         :param canvas_host: Hostname of the Canvas instance
+        :param session: The requests Session to use
+        :type session: requests.Session
         """
 
         # This is a requests Session object, not a DB session etc.
-        self._session = Session()
+        self._session = session or Session()
         self._canvas_host = canvas_host
 
     def send(
--- a/tests/unit/lms/services/canvas_api/_authenticated_test.py
+++ b/tests/unit/lms/services/canvas_api/_authenticated_test.py
@@ -6,6 +6,7 @@ from h_matchers import Any
 from lms.services import CanvasAPIServerError, NoOAuth2Token, ProxyAPIAccessTokenError
 from lms.services.canvas_api._authenticated import TokenResponseSchema
 from lms.services.canvas_api._basic import BasicClient
+from tests import factories
 
 
 class TestAuthenticatedClient:
@@ -154,7 +155,9 @@ class TestAuthenticatedClientIntegrated:
     """Tests which include the real basic client and Schema."""
 
     def test_ok(self, token_method, http_session, token_response):
-        http_session.set_response(token_response)
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data=token_response
+        )
         token = token_method("code")
         assert token == token_response["access_token"]
 
@@ -162,7 +165,9 @@ class TestAuthenticatedClientIntegrated:
     def test_bad_expires_in(
         self, token_method, http_session, token_response, bad_value
     ):
-        http_session.set_response(dict(token_response, expires_in=bad_value))
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data=dict(token_response, expires_in=bad_value)
+        )
 
         with pytest.raises(CanvasAPIServerError):
             token_method("code")
--- a/tests/unit/lms/services/canvas_api/_basic_test.py
+++ b/tests/unit/lms/services/canvas_api/_basic_test.py
@@ -76,7 +76,7 @@ class TestBasicClient:
     def test_send_raises_CanvasAPIError_for_request_errors(
         self, basic_client, http_session, Schema
     ):
-        http_session.set_response(status_code=501)
+        http_session.send.return_value = factories.requests.Response(status_code=501)
 
         with pytest.raises(CanvasAPIError):
             basic_client.send("METHOD", "path/", schema=Schema)
@@ -126,7 +126,7 @@ class TestBasicClient:
 
     @pytest.fixture(autouse=True)
     def has_ok_response(self, http_session):
-        http_session.set_response()
+        http_session.send.return_value = factories.requests.Response(status_code=200)
 
     @pytest.fixture
     def Schema(self):
--- a/tests/unit/lms/services/canvas_api/client_test.py
+++ b/tests/unit/lms/services/canvas_api/client_test.py
@@ -34,7 +34,9 @@ class TestCanvasAPIClientGetToken:
 class TestCanvasAPIClient:
     def test_authenticated_users_sections(self, canvas_api_client, http_session):
         sections = [{"id": 1, "name": "name_1"}, {"id": 2, "name": "name_2"}]
-        http_session.set_response({"sections": sections})
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data={"sections": sections}
+        )
 
         response = canvas_api_client.authenticated_users_sections("COURSE_ID")
 
@@ -52,10 +54,12 @@ class TestCanvasAPIClient:
     def test_authenticated_users_sections_deduplicates_sections(
         self, canvas_api_client, http_session
     ):
-        http_session.set_response(
-            {"sections": [{"id": 1, "name": "name"}, {"id": 1, "name": "name"}]}
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200,
+            json_data={
+                "sections": [{"id": 1, "name": "name"}, {"id": 1, "name": "name"}]
+            },
         )
-
         sections = canvas_api_client.authenticated_users_sections("course_id")
 
         assert sections == [{"id": 1, "name": "name"}]
@@ -63,8 +67,11 @@ class TestCanvasAPIClient:
     def test_authenticated_users_sections_raises_CanvasAPIError_with_conflicting_duplicates(
         self, canvas_api_client, http_session
     ):
-        http_session.set_response(
-            {"sections": [{"id": 1, "name": "name"}, {"id": 1, "name": "DIFFERENT"}]}
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200,
+            json_data={
+                "sections": [{"id": 1, "name": "name"}, {"id": 1, "name": "DIFFERENT"}]
+            },
         )
 
         with pytest.raises(CanvasAPIError):
@@ -79,7 +86,9 @@ class TestCanvasAPIClient:
             dict(section, unexpected="ignored") for section in sections
         ]
 
-        http_session.set_response(sections_with_noise)
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data=sections_with_noise
+        )
 
         response = canvas_api_client.course_sections("COURSE_ID")
 
@@ -94,8 +103,9 @@ class TestCanvasAPIClient:
     def test_course_sections_deduplicates_sections(
         self, canvas_api_client, http_session
     ):
-        http_session.set_response(
-            [{"id": 1, "name": "name"}, {"id": 1, "name": "name"}]
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200,
+            json_data=[{"id": 1, "name": "name"}, {"id": 1, "name": "name"}],
         )
 
         sections = canvas_api_client.course_sections("course_id")
@@ -105,8 +115,9 @@ class TestCanvasAPIClient:
     def test_course_sections_raises_CanvasAPIError_with_conflicting_duplicates(
         self, canvas_api_client, http_session
     ):
-        http_session.set_response(
-            [{"id": 1, "name": "name"}, {"id": 1, "name": "DIFFERENT"}]
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200,
+            json_data=[{"id": 1, "name": "name"}, {"id": 1, "name": "DIFFERENT"}],
         )
 
         with pytest.raises(CanvasAPIError):
@@ -115,7 +126,9 @@ class TestCanvasAPIClient:
     def test_course_sections_raises_CanvasAPIError_with_too_few_returned(
         self, canvas_api_client, http_session
     ):
-        http_session.set_response([])
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data=[]
+        )
 
         with pytest.raises(CanvasAPIError):
             canvas_api_client.course_sections("dummy")
@@ -125,7 +138,9 @@ class TestCanvasAPIClient:
             {"id": 1, "name": "Group category 1"},
             {"id": 2, "name": "Group category 2"},
         ]
-        http_session.set_response(group_categories)
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200, json_data=group_categories
+        )
 
         response = canvas_api_client.course_group_categories("COURSE_ID")
 
@@ -142,13 +157,14 @@ class TestCanvasAPIClient:
         )
 
     def test_users_sections(self, canvas_api_client, http_session):
-        http_session.set_response(
-            {
+        http_session.send.return_value = factories.requests.Response(
+            status_code=200,
+            json_data={
                 "enrollments": [
                     {"course_section_id": 101, "unexpected": "ignored"},
                     {"course_section_id": 102, "unexpected": "ignored"},
                 ]
-            }
+            },
         )
 
         response = canvas_api_client.users_sections("USER_ID", "COURSE_ID")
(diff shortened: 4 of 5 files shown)


jon-betts (commented):
Currently this PR passes with or without the fix. Could you include some testing notes on how to reproduce the issue?

There's also honestly too much commentary on this ticket. I think there's a chance that too much spent time has been spent on thinking about why this proves certain practice to be globally the right way to do things rather than thinking how this could be written to be more test friendly.

## The commentary lacks balance

There is nothing wrong with choosing in certain circumstances to use more or less of the real stack as makes sense. We do this elsewhere. 

There's also nothing wrong with scoping re-used fixtures close to the tests that use then in `conftest.py` we do that elsewhere too. In fact, there's a good argument to suggest we should probably do it more in cases where we have a large number of fixtures dumped in the root which are only relevant to a subset of tests, or where we have a lot of duplicated fixtures in a cluster of tests. This is a fundamental part of how the `pytest` setup works.

We've also discussed in Slack why the present testing methodology isn't an accident or result of a lack of care. This is a result of the present architecture where the unit of behavior under test happens both locally and at the deepest layer of the stack. I think this is a reality of the present architecture and hiding it with mocks might lead to lower cognitive load in one direction (this is all the same as I'm used to) but worse in another (what is this actually testing, how does this code work?). I've made suggestions to address that too by pulling up the schema evaluation to the top level. At that point the behavior under test would be in a neater layer and our usual mocking strategy would be more appropriate.

## My suggestion

In any case, I'd suggest the following fix to make the code more test friendly without work arounds or major reworking.

```python
class BasicClient:
    ...
    def __init__(self, canvas_host, session=None):
        self._session = session or Session()
        ...
```

In `conftest.py`

```python
@pytest.fixture
def http_session(patch):
    session = create_autospec(Session, instance=True)
    ...
    
@pytest.fixture
def basic_client(http_session):
    return BasicClient("canvas_host", session=http_session)
```

This avoids any questions of patch location and fixture ordering.

seanh:
> ## My suggestion
>
> In any case, I'd suggest the following fix to make the code more test friendly without work arounds or major reworking.
> 
> ```python
> class BasicClient:
>     ...
>     def __init__(self, canvas_host, session=None):
>         self._session = session or Session()
>         ...
> ```
> 
> In `conftest.py`
> 
> ```python
> @pytest.fixture
> def http_session(patch):
>     session = create_autospec(Session, instance=True)
>     ...
>     
> @pytest.fixture
> def basic_client(http_session):
>     return BasicClient("canvas_host", session=http_session)
> ```
> 
> This avoids any questions of patch location and fixture ordering.

I think this is a better solution. The three options that've come to mind so far are:

1. Make the `basic_client` fixture depend on the `http_response` fixture. With this solution it's unclear why `basic_client` has an unused dependency on `http_session`, necessitating an explanatory code comment. The patch-at-source is still happening with this solution, it's just that the fixture ordering has been forced so that the patch-at-source happens to work. It could easily break again

2. Make the `http_response` fixture depend on `basic_client`, create a mock session with `create_autospec` and then set it as `basic_client._session`. This means writing a private attribute which necessitates disabling a PyLint warning. While this actually gets rid of the patch-at-source, it introduces another problem: a refactoring could rename `BasicClient._session` to something else. The fixture would still execute without crashing and set `BasicClient._session` but this would no longer have any effect. Like the source-patch this could result in test set ups not actually setting up what they're supposed to and tests passing when they should be failing

3. Add a `session` argument to `BasicClient.__init__()` as you suggested. This both gets rid of the source patch and avoids writing a private attribute. Doesn't need either a code comment or a PyLint disable. Avoids both ways in which test set ups could silently fail. This seems like the best solution

This did have one consequence: the `http_session.set_response()` method is no longer possible.

The `http_session` fixture used to add a `set_response()` method to the mock `requests.Session`:

```python
@pytest.fixture
def http_session():
    session = patch("lms.services.canvas_api._basic.Session")
    session = session()

    def set_response(json_data=None, raw=None, status_code=200):
        session.send.return_value = factories.requests.Response(
            json_data=json_data, raw=raw, status_code=status_code
        )

    session.set_response = set_response
    return session
```

Using `create_autospec(..., spec_set=True)` means you can't _set_ any attribute on the mock that doesn't exist on the real `requests.Session` class, so `session.set_response()` can no longer be added.

We could just remove the `spec_set=True` and sometimes we do, but using `spec_set=True` is a mock best practice for a reason: generally you want tests to fail if code either reads or writes an attribute that doesn't exist

I decided to just get rid of `set_response()` and replace calls like `http_session.set_response(...)` with
direct calls to `http_session.send.return_value = factories.requests.Response(...)`. This is longer but it's also arguably more readable: I can see exactly what the test is doing, rather than having to look up a distant shared fixture in a `conftest.py` file to understand it. This is similar to other recent decisions we made about fixtures: in some cases it can actually be better for test readability to make the tests do more work. If you have to learn a bunch of fixtures first before you can understand the tests, this can make things harder. (Very common fixtures like `pyramid_request` or factories are exempted, since those are already familiar.)

I decided not to but for completeness sake it would actually be possible to keep the `spec_set=True` and re-implement `set_response()` as a standalone fixture. If we wanted to have a `set_response()` helper I think this would be a more pytest way of doing this:

```python
@pytest.fixture
def set_response(http_session):

    def set_response(json_data=None, raw=None, status_code=200):
           json_data=json_data, raw=raw, status_code=status_code
       )

    return set_response
```

seanh:
Regarding the wider code design discussion here:

Testing notes are at the top of the PR description: if you break the `http_session` fixture the tests still pass because the fixture doesn't actually do anything.

I think we're going to need to move on from this because I need to make progress on the Blackboard Files project and don't have too much time to get bogged down right now. But just to make sure that I've stated my point of view:

I think the amount of commentary is appropriate: I'm actually very concerned about patch-at-source and integrated tests finding their way further into our codebase because I think these would have a bunch of very unfortunate consequences, so I want to make sure that's discussed and the issues made clear, and we're all on the same page. I'm afraid I really **do** think there's something wrong with mixing integrated tests into our unit tests, and I do think that isolated unit testing is a much better approach for several reasons. I also think that having one service use a different unit testing approach than everything else is expensive in terms of maintenance time.

I definitely don't agree that our present architecture necessitates and integrated tests approach. The architecture is absolutely amenable to isolated testing. That's the test approach that we currently use 90% of the time throughout our codebase.

I also **do** think there's something wrong with shared fixtures in `conftest.py` files. Not to say that we should never use any `conftest.py` files (of course). But from long experience with pytest: shared fixtures are often a source of problems. They tightly couple tests together, often making test code maintenance very difficult and introducing subtle bugs. We want our shared fixtures to be few and very generic in design. The `pyramid_request` fixture and mock service fixtures are good examples. One thing you definitely don't want to do in a `conftest.py` is use `patch()` because you'll be patching at source and that will very likely mean test bugs, as happened in this case.

While it's the best short-term fix here I don't think the `session` argument to `BasicClient.__init__()` is a solution we want to generalise. I don't think we want to start adding dependency injection arguments to all our `__init__()`'s and other methods and functions. We want to be able to patch imports. So we do need to stick to a test-isolation method that avoids patch-at-source.

marcospri (approved):

jon-betts:
I think this description would have been much easier to digest and to the point if it had something along the lines of:

 * There's a bug created by fixtures getting imported in the wrong order
 * We can fix it by enforcing fixture order (the original fix)
 * The use of a `conftest.py` here is a little unusual and I've seen this cause problems before
 * The use of `patch` within that `conftest.py` is a no-no (and here are some ways around it)
 * I don't like the integrated style of the tests and would like to use a regular patch style
 
Less is more here. This was a negative read and does not look like an invitation to discussion but more of a list of crimes.

With regards to `specset` I had this just disabled, but one way to keep it is with something like this:

```
class MySession(Session):
    def set_response(...)
    
create_autospec(MySession, ...)    
```

The only trick there is to re-instate the `set_response` body with defining the signature twice. I get the feeling you could transplant the original back on somehow.

PR #2730, opened 2021-05-24
Improve the httpretty fixture

seanh:
`conftest.py` has an autouse fixture that configures `httpretty` to
respond to any request with a blank response:

https://github.com/hypothesis/lms/blob/5394cf2bfb92cb219e177f3c0a7991add024f242/tests/unit/conftest.py#L317-L348

This is so that the unit tests can never send requests to real external
services -- `httpretty` always catches them.

There are two problems with this:

1. It means that by default any network request sent by code under test
   will receive a successful, blank response. This can hide problems
   with tests. It's better for attempted network requests to crash
   unless the individual test itself has explicitly set up a response.

2. It means that if a particular test uses
   [`httpretty.register_uri()`](https://httpretty.readthedocs.io/en/latest/api.html#register-uri)
   to register a particular response it won't work: `httpretty` will
   send the response from `conftest.py` instead.

   To work around this tests in `exceptions_test.py` and
   `lti_outcomes_test.py` have used `httpretty.register_uri(...,
   priority=1)` to override the default response:

   https://github.com/hypothesis/lms/blob/9c7eba7294916ee3177dcbd09f449a30222ec707/tests/unit/lms/services/exceptions_test.py#L215-L221

   https://github.com/hypothesis/lms/blob/9c7eba7294916ee3177dcbd09f449a30222ec707/tests/unit/lms/services/exceptions_test.py#L227-L229

   https://github.com/hypothesis/lms/blob/9c7eba7294916ee3177dcbd09f449a30222ec707/tests/unit/lms/services/exceptions_test.py#L235-L241

   https://github.com/hypothesis/lms/blob/f74ccccb7e31549d1ccd7cc50ede948f6f4ef231/tests/unit/lms/services/lti_outcomes_test.py#L123-L129

   The `priority` argument allows a `register_uri()` call to override
   previous ones: higher priorities override lower ones. It defaults to
   `0`.

Fix both problems by:

1. Removing the `register_uri()` calls from the global `httpretty_`
   fixture. Now there are no global responses to override those
   registered by tests

2. Adding `allow_net_connect=False` to the global `httpretty_` fixture.
   This means that any network request for which the test hasn't
   registered an httpretty response will raise an exception, rather than
   being allowed through to the real network.

So tests that don't set up their own httpretty responses for any network
requests will now crash. And tests can register their own httpretty
responses without having to pass `priority=1`.


jon-betts (approved):
This looks like a much better solution.

I think it might also be acceptable to just not do this, and apply the fixture on in cases where we think there's a chance of network access.

Currently this fixture will presumably take effect before and after each test run whether it's helping us or not.

It might be interesting to disable the auto-use and see if there's much of a difference in speed?

seanh:
Without the `autouse` I think it's too easy for some test to forget to set up an httpretty response (or to try to set it up but do it wrong) and to actually send real network requests. This can results in tests that should fail (because the code is actually broken) succeeding instead and hiding bugs. For example if the test tries to set up an httpretty error response but the test set up is wrong and the code is actually getting a successful response from example.com.

Once we have a central `HTTPService` that everything always uses to send HTTP requests then httpretty could be limited to the `HTTPService`'s tests and everything else could use the mock `HTTPService`. (Of course that breaks down as soon as someone goes and adds some code that calls `requests` directly instead of going through `HTTPService`.)

marcospri (approved):

PR #2738, opened 2021-05-26
Redirect to the real Blackboard authorization page

seanh:
When the user clicks the <kbd>Choose PDF from Blackboard</kbd> button, if we don't already have a Blackboard API access token for that user, the frontend opens a popup window to Blackboard's OAuth authorization page in order to start the process of getting an access token and saving it in our DB.

The view that the popup window opens to is `views/api/blackboard/authorize.py::authorize()` and our current implementation of that view is a fake: it just redirects directly to our OAuth redirect view instead of to Blackboard:

https://github.com/hypothesis/lms/blob/c4b4614aa7544ce3840150ca48da61d6cb1d2781/lms/views/api/blackboard/authorize.py#L9-L20

What's supposed to happen is that we redirect to Blackboard's authorization page, the user clicks <kbd>Allow</kbd>, and then _Blackboard_ redirects to our OAuth redirect view.

This PR changes the `authorize()` view so that it actually redirects to the real Blackboard authorization page instead of skipping it.

Note that the OAuth redirect view that gets called when the user clicks <kbd>Allow</kbd> is still a fake, so we still don't actually talk to Blackboard to get a real access token, we still just store a fake token in the DB. That's coming in a later PR.

## Testing

1. To see the <kbd>Choose PDF from Blackboard</kbd> you need to cherry-pick https://github.com/hypothesis/lms/pull/2636 onto this branch: `git cherry-pick b8a99c878c50a517c8ebdb76e4e54127e88b0500`

2. You need the latest dev data: run `make devdata`

3. The authorization only triggers if you don't already have a Blackboard API access token (real or fake) for the user in your DB. You can make sure of this by deleting them: `tox -qe dockercompose -- exec postgres psql --pset expanded=auto -U postgres -c 'DELETE FROM oauth2_token;'`

4. Also, the assignment configuration UI only triggers if you don't already have a `ModuleItemConfiguration` for the test assignment in your DB. Make sure of that too: `tox -qe dockercompose -- exec postgres psql --pset expanded=auto -U postgres -c "DELETE FROM module_item_configurations WHERE tool_consumer_instance_guid = '041c05b262b54282a0d9b61c42413ec0' and resource_link_id = '_39_1';"`

5. Go to our test Blackboard site: https://aunltd-test.blackboard.com/

6. Log in using the `blackboardteacher` account in 1Password

5. Revoke our LMS app's access to the `blackboardteacher` account.

   When you allow our app access to the Blackboard API, Blackboard remembers this decision and doesn't ask you again. Blackboard's authorization URL immediately redirects to our OAuth URL without showing the user anything. This has the effect of the authorization popup window popping up and then immediately closing again (authorization having been successful). Which sort of prevents you from actually testing it. And since we use shared user accounts _someone else_ might have already allowed the app access to the account. So you need to revoke any previously given access:

   Go to https://aunltd-test.blackboard.com/ultra/tools, click on **Application Authorization**, and revoke Hypothesis Dev.

7. Go to **Developer Test Course with Original Course View**: https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline

8. Launch **localhost (make devdata) Blackboard Files Test Assignment**, it'll open in a new tab

9. You should see this:

   Screenshot from 2021-05-26 18-54-08

10. Click <kbd>Select PDF from Blackboard</kbd>

11. You should see this:

    Screenshot from 2021-05-26 18-55-00

12. Click <kbd>Authorize</kbd>

13. You should see this:

    Screenshot from 2021-05-26 19-00-59

    **TODO:** The popup window is too short. You have to scroll down to see the <kbd>Allow</kbd> button.

14. Click <kbd>Allow</kbd>

15. You should see this:

    Screenshot from 2021-05-26 19-02-03

That's it. The code doesn't actually do anything with the authorization that you allowed (it doesn't actually get a real access token from Blackboard and save it in the DB) and the list of files is fake. That's all coming in later PRs. For now there's no need to complete the assignment creation.

If you click <kbd>Deny</kbd> instead of <kbd>Allow</kbd> it'll still act as if you clicked <kbd>Allow</kbd>. That's because the view that deals with your allow or deny is still a fake view at this point.

If you want to test it again you'll need to revoke the authorization in Blackboard again and delete the fake access token from `oauth2_token` table again. If you selected a file and completed the assignment configuration you'll need to delete the `module_item_configurations` row again too.

### DIFF
--- a/lms/config/__init__.py
+++ b/lms/config/__init__.py
@@ -68,6 +68,8 @@ def configure(settings):
         "vitalsource_lti_launch_secret": sg.get("VITALSOURCE_LTI_LAUNCH_SECRET"),
         "admin_auth_google_client_id": sg.get("ADMIN_AUTH_GOOGLE_CLIENT_ID"),
         "admin_auth_google_client_secret": sg.get("ADMIN_AUTH_GOOGLE_CLIENT_SECRET"),
+        "blackboard_api_client_id": sg.get("BLACKBOARD_API_CLIENT_ID"),
+        "blackboard_api_client_secret": sg.get("BLACKBOARD_API_CLIENT_SECRET"),
     }
 
     env_settings["dev"] = asbool(env_settings["dev"])
--- a/tests/unit/lms/views/api/blackboard/authorize_test.py
+++ b/tests/unit/lms/views/api/blackboard/authorize_test.py
@@ -1,30 +1,46 @@
-from urllib.parse import parse_qs, urlparse
-
 import pytest
+from h_matchers import Any
 
 from lms.services import NoOAuth2Token
 from lms.views.api.blackboard.authorize import authorize, oauth2_redirect
+from tests.matchers import temporary_redirect_to
 
-pytestmark = pytest.mark.usefixtures("oauth2_token_service")
+pytestmark = pytest.mark.usefixtures(
+    "application_instance_service", "oauth2_token_service"
+)
 
 
 class TestAuthorize:
-    def test_it_just_redirects_to_the_oauth2_redirect_view(self, pyramid_request):
-        response = authorize(pyramid_request)
-
-        assert response.status_int == 302
-        assert response.location.startswith(
-            pyramid_request.route_url("blackboard_api.oauth.callback")
-        )
-
-    def test_it_authenticates_the_redirect_with_an_OAuth2_state_param(
-        self, pyramid_request, OAuthCallbackSchema
+    def test_it_redirects_to_the_Blackboard_authorization_page(
+        self,
+        application_instance_service,
+        OAuthCallbackSchema,
+        oauth_callback_schema,
+        pyramid_request,
     ):
+        application_instance = application_instance_service.get.return_value
+
         response = authorize(pyramid_request)
 
         OAuthCallbackSchema.assert_called_once_with(pyramid_request)
-        state = parse_qs(urlparse(response.location).query).get("state")
-        assert state == ["test_state"]
+        assert response == temporary_redirect_to(
+            Any.url(
+                scheme="https",
+                host=application_instance.lms_host(),
+                path="learn/api/public/v1/oauth2/authorizationcode",
+                query={
+                    "client_id": pyramid_request.registry.settings[
+                        "blackboard_api_client_id"
+                    ],
+                    "response_type": "code",
+                    "redirect_uri": pyramid_request.route_url(
+                        "blackboard_api.oauth.callback"
+                    ),
+                    "state": str(oauth_callback_schema.state_param.return_value),
+                    "scope": "read offline",
+                },
+            )
+        )
 
 
 class TestOAuth2Redirect:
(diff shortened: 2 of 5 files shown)


lms/config/__init__.py
```diff
@@ -68,6 +68,8 @@ def configure(settings):
         "vitalsource_lti_launch_secret": sg.get("VITALSOURCE_LTI_LAUNCH_SECRET"),
         "admin_auth_google_client_id": sg.get("ADMIN_AUTH_GOOGLE_CLIENT_ID"),
         "admin_auth_google_client_secret": sg.get("ADMIN_AUTH_GOOGLE_CLIENT_SECRET"),
+        "blackboard_api_client_id": sg.get("BLACKBOARD_API_CLIENT_ID"),
+        "blackboard_api_client_secret": sg.get("BLACKBOARD_API_CLIENT_SECRET"),
```
seanh, line 72: `make devdata` will set these to the necessary values for you

tests/unit/lms/views/api/blackboard/authorize_test.py
```diff
@@ -1,30 +1,46 @@
...
+                    ],
+                    "response_type": "code",
+                    "redirect_uri": pyramid_request.route_url(
+                        "blackboard_api.oauth.callback"
+                    ),
+                    "state": str(oauth_callback_schema.state_param.return_value),
+                    "scope": "read offline",
+                },
+            )
+        )
```
seanh, line 43: h-matchers coming in handy here.

Not much of a test, and the test looks very similar to the code itself. That's because there's not much to test here: the view basically just returns a redirect to a hard-coded URL, with a couple of parts of the URL coming dynamically from settings and from the `ApplicationInstance` in the DB.

jon-betts (approved):
Sorry that took so long, I was being a fool with Blackboard.

Yes this looks good. I assume that `urlunparse` isn't the intended final state of that function? Presumably this is reliant on the base clients etc getting fancier.

marcospri (approved):

seanh:
>  I assume that urlunparse isn't the intended final state of that function? Presumably this is reliant on the base clients etc getting fancier.

You mean the Blackboard API client (when it exists) could provide some sort of `authorization_url()` function for the view to call, rather than the view having to create the URL itself? So something like `return HTTPFound(location=blackboard_api_client.authorization_url())`?

jon-betts:
No I was thinking it's an extremely common pattern to have something like:

`host_prefix + url_prefix + path` 

For all paths in an API, so you almost always want a function like:

`get_url(path, params)`

Which bolts it together for you. I was assuming such a thing was going to be in the base client.

seanh:
What do you mean by the base client here?

Pyramid actually provides a `route_url()` method that sounds similar to what you're describing, it takes path segments and a query dict as args. That's what you'd normally use with a redirect: `return HTTPFound(location=request.route_url("my_route", path_parts, query_dict))`. That's for redirecting to things the app has a route for which normally means redirecting to other URLs of the same app but you can actually add [external routes](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#external-routes) for generating routes to other apps, that might work here.

PR #2739, opened 2021-05-26
Get real Blackboard access tokens

seanh:
Actually call the Blackboard API to get a real access token and save it to the DB, instead of saving a fake one.

## Testing

1. To see the <kbd>Choose PDF from Blackboard</kbd> you need to cherry-pick https://github.com/hypothesis/lms/pull/2636 onto this branch: `git cherry-pick frontend-changes-for-blackboard-files-poc`

2. You need the latest dev data: run `make devdata`

3. The authorization only triggers if you don't already have a `ModuleItemConfiguration` for the test assignment or Blackboard API access token (real or fake) for the user in your DB. You can make sure of this by deleting them: `tox -qe dockercompose -- exec postgres psql -U postgres -c "'DELETE FROM oauth2_token; DELETE FROM module_item_configurations WHERE tool_consumer_instance_guid = \'041c05b262b54282a0d9b61c42413ec0\' and resource_link_id = \'_39_1\';'"`

5. Go to our test Blackboard site: https://aunltd-test.blackboard.com/

6. Log in using the `blackboardteacher` account in 1Password

5. Revoke our LMS app's access to the `blackboardteacher` account.

   When you allow our app access to the Blackboard API, Blackboard remembers this decision and doesn't ask you again. Blackboard's authorization URL immediately redirects to our OAuth URL without showing the user anything. This has the effect of the authorization popup window popping up and then immediately closing again (authorization having been successful). Which sort of prevents you from actually testing it. And since we use shared user accounts _someone else_ might have already allowed the app access to the account. So you need to revoke any previously given access:

   Go to https://aunltd-test.blackboard.com/ultra/tools, click on **Application Authorization**, and revoke Hypothesis Dev.

7. Go to **Developer Test Course with Original Course View**: https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline

8. Launch **localhost (make devdata) Blackboard Files Test Assignment**, it'll open in a new tab

9. You should see this:

   Screenshot from 2021-05-26 18-54-08

10. Click <kbd>Select PDF from Blackboard</kbd>

11. You should see this:

    Screenshot from 2021-05-26 18-55-00

12. Click <kbd>Authorize</kbd>

13. You should see this:

    Screenshot from 2021-05-26 19-00-59

    **Known issue:** The popup window is too short. You have to scroll down to see the <kbd>Allow</kbd> button.

14. Click <kbd>Allow</kbd>

15. You should see this:

    Screenshot from 2021-05-26 19-02-03

16. If you now look in your DB you should find that a real Blackboard access token has been saved: `tox -qe dockercompose -- exec postgres psql -U postgres -c "SELECT * FROM oauth2_token;"`

17. Assuming you didn't actually select a file and complete the assignment configuration you should be able to launch the assignment again and this time you'll get to the files list without being asked to authorize again because you already have an access token in your DB.

### Known issues

Error handling isn't done. If something goes wrong with the request to Blackboard and `HTTPService` raises an `HTTPError` or `HTTPValidationError` you'll end up with a JSON error message from our proxy API rendered directly to the user in the popup window. The same thing will happen if you click <kbd>Deny</kbd> instead of <kbd>Allow</kbd> on Blackboard's authorization page.

The desired behavior is our usual nice error dialog with <kbd>Try again</kbd> button.

The Canvas API code handles this with an exception view:

https://github.com/hypothesis/lms/blob/be46e8c356c0da97d86c53cedab28d4b31570cc4/lms/views/api/canvas/authorize.py#L100-L128

But there's a bunch of Canvas-specific stuff in that: the `"canvas_api.oauth.authorize"` route name is passed to `route_url()`, `js_config.enable_canvas_oauth2_redirect_error_mode()` is called, Canvas API scopes are passed to it.

We either need to write something very similar looking (but not exactly the same) for Blackboard or we need to do some refactoring to be able to share code. I'll create a separate issue for this.

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -21,6 +21,9 @@ from lms.services.exceptions import (
 def includeme(config):
     config.register_service_factory("lms.services.http.factory", name="http")
     config.register_service_factory(
+        "lms.services.blackboard_api.factory", name="blackboard_api_client"
+    )
+    config.register_service_factory(
         "lms.services.canvas_api.canvas_api_client_factory", name="canvas_api_client"
     )
     config.register_service_factory("lms.services.h_api.HAPI", name="h_api")
--- a/lms/services/blackboard_api.py
+++ b/lms/services/blackboard_api.py
@@ -1,0 +1,60 @@
+from lms.validation.authentication import OAuthTokenResponseSchema
+
+
+class BlackboardAPIClient:
+    def __init__(
+        self,
+        blackboard_host,
+        client_id,
+        client_secret,
+        redirect_uri,
+        http_service,
+        oauth2_token_service,
+    ):  # pylint:disable=too-many-arguments
+        self.blackboard_host = blackboard_host
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.redirect_uri = redirect_uri
+
+        self._http_service = http_service
+        self._oauth2_token_service = oauth2_token_service
+
+    def get_token(self, authorization_code):
+        """
+        Get an access token from Blackboard and save it in the DB.
+
+        :raise services.HTTPError: if something goes wrong with the access
+            token request to Blackboard
+        """
+        # Send a request to Blackboard to get an access token.
+        response = self._http_service.post(
+            f"https://{self.blackboard_host}/learn/api/public/v1/oauth2/token",
+            data={
+                "grant_type": "authorization_code",
+                "redirect_uri": self.redirect_uri,
+                "code": authorization_code,
+            },
+            auth=(self.client_id, self.client_secret),
+            schema=OAuthTokenResponseSchema,
+        )
+
+        # Save the access token to the DB.
+        self._oauth2_token_service.save(
+            response.validated_data["access_token"],
+            response.validated_data.get("refresh_token"),
+            response.validated_data.get("expires_in"),
+        )
+
+
+def factory(_context, request):
+    application_instance = request.find_service(name="application_instance").get()
+    settings = request.registry.settings
+
+    return BlackboardAPIClient(
+        blackboard_host=application_instance.lms_host(),
+        client_id=settings["blackboard_api_client_id"],
+        client_secret=settings["blackboard_api_client_secret"],
+        redirect_uri=request.route_url("blackboard_api.oauth.callback"),
+        http_service=request.find_service(name="http"),
+        oauth2_token_service=request.find_service(name="oauth2_token"),
+    )
--- a/lms/views/api/blackboard/authorize.py
+++ b/lms/views/api/blackboard/authorize.py
@@ -4,7 +4,6 @@ from pyramid.httpexceptions import HTTPFound
 from pyramid.view import view_config
 
 from lms.security import Permissions
-from lms.services import NoOAuth2Token
 from lms.validation.authentication import OAuthCallbackSchema
 
 
@@ -47,13 +46,8 @@ def authorize(request):
     route_name="blackboard_api.oauth.callback",
     permission=Permissions.API,
     renderer="lms:templates/api/oauth2/redirect.html.jinja2",
+    schema=OAuthCallbackSchema,
 )
 def oauth2_redirect(request):
-    oauth2_token_service = request.find_service(name="oauth2_token")
-
-    try:
-        oauth2_token_service.get()
-    except NoOAuth2Token:
-        oauth2_token_service.save("fake_access_token", "fake_refresh_token", 9999)
-
+    request.find_service(name="blackboard_api_client").get_token(request.params["code"])
     return {}
--- a/tests/unit/lms/services/blackboard_api_test.py
+++ b/tests/unit/lms/services/blackboard_api_test.py
@@ -1,0 +1,101 @@
+from unittest.mock import sentinel
+
+import pytest
+
+from lms.services import HTTPError, HTTPValidationError
+from lms.services.blackboard_api import BlackboardAPIClient, factory
+from lms.validation.authentication import OAuthTokenResponseSchema
+
+
+class TestBlackboardAPIClient:
+    def test_it(self, svc, http_service, oauth2_token_service):
+        http_service.post.return_value.validated_data = {
+            "access_token": sentinel.access_token,
+            "refresh_token": sentinel.refresh_token,
+            "expires_in": sentinel.expires_in,
+        }
+
+        svc.get_token(sentinel.authorization_code)
+
+        # It calls the Blackboard API to get the access token.
+        http_service.post.assert_called_once_with(
+            "https://blackboard.example.com/learn/api/public/v1/oauth2/token",
+            data={
+                "grant_type": "authorization_code",
+                "redirect_uri": sentinel.redirect_uri,
+                "code": sentinel.authorization_code,
+            },
+            auth=(sentinel.client_id, sentinel.client_secret),
+            schema=OAuthTokenResponseSchema,
+        )
+
+        # It saves the access token in the DB.
+        oauth2_token_service.save.assert_called_once_with(
+            sentinel.access_token, sentinel.refresh_token, sentinel.expires_in
+        )
+
+    @pytest.mark.parametrize("exception_class", [HTTPError, HTTPValidationError])
+    def test_it_raises_if_http_service_raises(self, svc, http_service, exception_class):
+        http_service.post.side_effect = exception_class
+
+        with pytest.raises(exception_class):
+            svc.get_token(sentinel.authorization_code)
+
+    def test_if_theres_no_refresh_token_or_expires_in(
+        self, svc, http_service, oauth2_token_service
+    ):
+        # refresh_token and expires_in are optional fields in
+        # OAuthTokenResponseSchema so get_token() has to still work if they're
+        # missing from the validated data.
+        http_service.post.return_value.validated_data = {
+            "access_token": sentinel.access_token
+        }
+
+        svc.get_token(sentinel.authorization_code)
+
+        oauth2_token_service.save.assert_called_once_with(
+            sentinel.access_token, None, None
+        )
+
+    @pytest.fixture
+    def svc(self, http_service, oauth2_token_service):
+        return BlackboardAPIClient(
+            blackboard_host="blackboard.example.com",
+            client_id=sentinel.client_id,
+            client_secret=sentinel.client_secret,
+            redirect_uri=sentinel.redirect_uri,
+            http_service=http_service,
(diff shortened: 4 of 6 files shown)


lms/views/api/blackboard/authorize.py
```diff
@@ -47,13 +46,8 @@ def authorize(request):
...
 )
 def oauth2_redirect(request):
-    oauth2_token_service = request.find_service(name="oauth2_token")
-
-    try:
-        oauth2_token_service.get()
-    except NoOAuth2Token:
-        oauth2_token_service.save("fake_access_token", "fake_refresh_token", 9999)
-
+    request.find_service(name="blackboard_api_client").get_token(request.params["code"])
```
seanh, line 52: Always good when a view is just a one-liner

tests/unit/lms/services/blackboard_api_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+        BlackboardAPIClient.assert_called_once_with(
+            blackboard_host=application_instance.lms_host(),
+            client_id=settings["blackboard_api_client_id"],
+            client_secret=settings["blackboard_api_client_secret"],
+            redirect_uri=pyramid_request.route_url("blackboard_api.oauth.callback"),
+            http_service=http_service,
+            oauth2_token_service=oauth2_token_service,
+        )
+        assert service == BlackboardAPIClient.return_value
```
seanh, line 97: This test looks pretty silly, but that's because there's nothing really to test about the `factory()` method: no logic or side effects or different inputs etc. The `factory()` method is just a hard-coded call to `BlackboardAPIClient()` with some of the arguments coming from settings etc. This test is pretty much just here to make sure the code doesn't crash and to get us 100% coverage. This is often the case with the tests for `factory()` methods, as long as we want to unit test the `factory()` methods separately from the services themselves.

lms/services/__init__.py
```diff
@@ -20,6 +20,9 @@
 
 def includeme(config):
     config.register_service_factory("lms.services.http.factory", name="http")
+    config.register_service_factory(
+        "lms.services.blackboard_api.factory", name="blackboard_api_client"
+    )
```
jon-betts, line 25: If we are registering a new service, this seems like a good time to use the normal interface style.
```suggestion
    config.register_service_factory(
        "lms.services.blackboard_api.factory", iface=BlackboardAPIClient
    )
```

lms/services/blackboard_api.py
```diff
@@ -0,0 +1,60 @@
...
+def factory(_context, request):
+    application_instance = request.find_service(name="application_instance").get()
+    settings = request.registry.settings
+
+    return BlackboardAPIClient(
+        blackboard_host=application_instance.lms_host(),
+        client_id=settings["blackboard_api_client_id"],
+        client_secret=settings["blackboard_api_client_secret"],
+        redirect_uri=request.route_url("blackboard_api.oauth.callback"),
+        http_service=request.find_service(name="http"),
```
jon-betts, line 58: How does the HTTP service benefit from being a service rather than a plain class?

lms/services/blackboard_api.py
```diff
@@ -0,0 +1,60 @@
...
+        # Send a request to Blackboard to get an access token.
+        response = self._http_service.post(
+            f"https://{self.blackboard_host}/learn/api/public/v1/oauth2/token",
+            data={
+                "grant_type": "authorization_code",
+                "redirect_uri": self.redirect_uri,
+                "code": authorization_code,
+            },
+            auth=(self.client_id, self.client_secret),
+            schema=OAuthTokenResponseSchema,
```
jon-betts, line 38: We've mentioned this a few times, but I think putting the responsibility for validating the data (the format of which is a very high level concept) into the lowest HTTP layer is a bit strange and causes some problems leading to the clarity vs. isolation design problems in the unit tests.

lms/services/blackboard_api.py
```diff
@@ -0,0 +1,60 @@
...
+        self._http_service = http_service
+        self._oauth2_token_service = oauth2_token_service
+
+    def get_token(self, authorization_code):
+        """
+        Get an access token from Blackboard and save it in the DB.
+
+        :raise services.HTTPError: if something goes wrong with the access
+            token request to Blackboard
+        """
```
jon-betts, line 28: This function looks like it's covering a lot of the same ground as `services.canvas_api._authenticated.AuthenticatedClient.get_token()`. Are we re-inventing the wheel here?

I know that class isn't instantly ready for use, but it was intended to be adaptable. There isn't much specific to Canvas in there apart from one hard coded reference to `login/oauth2/token`.

tests/unit/lms/services/blackboard_api_test.py
jon-betts: Yeah, the factory tests are always pure plumbing

jon-betts (commented):
I've followed the instructions here and they work. Functionally the code looks fine to me.

I have some questions about the overall design though. I feel like this is the start of rediscovering and recreating the structure already present in the Canvas API without any re-use there. Green fields are always nice, but I think there's a risk of (painfully) rediscovering why things were done, and of a missed opportunity to unify the systems. I'm sure the previous one looks complicated and this is nice and simple, but I suspect as you go on you'll converge on something similar.

I think it might be a good idea to take a moment to discuss the design (verbally). Some thoughts on that:

 * The base HTTP layer isn't doing as much as it could to be helpful to make calls: forming URLs is a pretty basic thing to want help with in an HTTP API caller
 * Reading between the lines, I think this is because I think there was a view that it could only talk to one target system. I'm not sure, but I wonder if this is idea came from working backwards from the current design's singleton? Normally, you'd just instantiate a separate one for each service. There's not much benefit to sharing the HTTP session, as the SSL caching is per host anyway.
  * I also think it's doing too much in applying Schema. This predated the previous design, and I think it's a mistake as I think applying schema could be orthogonal to retrieving JSON data. This would completely decouple the two concepts and allow them to be tested independantly
 * Ideally you should not need a factory to create one of these APIs, they should have an interface which allows them to be created with __init__ alone, so they are independent of the Pyramid ecosystem for the most part (probably still raising Pyramid type errors).
  
## TLDR

 * I think some high level thoughts about components would be good
 * Some separate components to me seem to be:
    * HTTP transport
    * OAuth Layer
    * Schema application
    * Final API specific component

jon-betts:
I've had a quick look at the `HTTP` service again and I think I've worked out what's bugging me about it. It's pretty much a requests Session object in disguise, especially without the schema (as I suggested):

```python
class HTTPSession(Session):
    def request(self, *args, **kwargs):
        try:
            response = super().request(*args, **kwargs)
            response.raise_for_status()
        except RequestException as err:
            raise HTTPError(response) from err

        return response
```

lms/services/blackboard_api.py
seanh: I'm not sure I agree. I think "send this request and validate the response using this schema" is a pretty logical unit, and something that's frequently needed.

`HTTPService` isn't responsible for the format of the data, it's just responsible for applying the given schema.

`response = http_service.post(..., schema=FooSchema)` seems clear to me. It makes it clear and explicit that `FooSchema` will have been applied to `response` and that the method can expect `response.validated_data` to contain the expected fields as defined by `FooSchema`, or a correctly-formatted exception to have been raised.

It's pretty analogous to guard schemas for views:

```python
@view_config(..., schema=FooSchema)
def foo_view(request):
    ...
```

The view can assume that `request.parsed_params` will contain the valid params as defined by `FooSchema` or an exception will have been raised.

lms/services/blackboard_api.py
seanh: It makes it easy for the view to get the `BlackboardAPIClient` because it can just do `request.find_service(name="blackboard_api")`. The view doesn't need to know how to construct the service, which is fairly complicated (lots of arguments, including other services).

It's nice that each service comes with its own factory method that knows how to construct it. Especially when a service depends on multiple other services as this one does. `BlackboardAPIClient`'s factory doesn't need to know how to construct `ApplicationInstanceService`, `HTTPService` or `OAuth2Service`, it just requests them.

The tests for `BlackboardAPIClient` benefit from the nice way that we have of injecting centralized mock services.

`BlackboardAPIClient` (and the other services it depends on) are all per-request singletons. So if different parts of the code need the `BlackboardAPIClient` they can each request it, you don't need to pass it around. `HTTPService` can ensure that a TCP session is re-used. `ApplicationInstanceService` can ensure that a DB query for the `ApplicationInstance` is cached. Etc.

Even though I do think there are several actual benefits to using a service here, also to be honest ervices are just what we use for these, I'm not looking for reasons to deviate. Consistency is a benefit itself.

seanh:
@jon-betts I might vote for keeping the design as-is for now and allowing it to evolve as needed, as we add the rest of the functionality.

To be fair the current code design enables the Blackboard API client so far to be extremely simple. I think that's hard to argue with for now. If the design needs to evolve as we add more functionality we can do that. The `BlackboardAPIClient` class in this PR is just a class with one method that has only two lines of code, it's pretty much this:

```python
class BlackboardAPIClient:
    def __init__(self, ...):
        ...

    def get_token(self, authorization_code):
        response = self._http_service.post(...)
        self._oauth2_token_service.save(response.validated_data["access_token"], ...)
```

It is similar to the `get_token()` method in `services/canvas_api/_authenticated.py::AuthenticatedClient`. That could perhaps be pulled out into an `HTTPService.get_token()` method that sends an OAuth 2 token request and saves the result in the DB. Then the `BlackboardAPIClient` could just be:

```python
class BlackboardAPIClient:
    ...

    def get_token(self, authorization_code):
        self._http_service.get_token(...)
```

It's a little difficult to extract stuff out of the Canvas API code currently though. It uses its own base client instead of `HTTPService` for sending requests and because of the integrated tests it's not easy to change it to use `HTTPService`, the tests expect real `requests` calls to be made and real Marshmallow schemas to be applied, I'd have to rewrite it a bit in order to get the mock `HTTPService` in there. I think for the sake of making forward progress this is probably best left alone for now

lms/services/blackboard_api.py
```diff
@@ -0,0 +1,60 @@
...
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.redirect_uri = redirect_uri
+
+        self._http_service = http_service
+        self._oauth2_token_service = oauth2_token_service
+
+    def get_token(self, authorization_code):
+        """
+        Get an access token from Blackboard and save it in the DB.
```
marcospri, line 24: I'd have assumed `get_token`  returns the `OAuth2Token` but neither does `.save` so don't bother changing it.

marcospri (approved):

PR #2742, opened 2021-05-27
We rely on marshmallow directly, not just as a dependency of webargs

jon-betts:
Installing webargs gets us marshmallow, but we could remove webargs
one day and continue using marshmallow. This means we list it as an independant
requirement, and that it therefore shows up in our dependency analysis.


robertknight (approved):
LGTM

PR #2743, opened 2021-05-28
Course service refactor

marcospri:
A few renames and new, not yet used, methods on the service

### DIFF
--- a/lms/services/course.py
+++ b/lms/services/course.py
@@ -1,22 +1,40 @@
 import json
 from copy import deepcopy
 
-from lms.models import CourseGroupsExportedFromH, LegacyCourse
+from lms.models import Course, CourseGroupsExportedFromH, LegacyCourse
 
 
 class CourseService:
     def __init__(self, application_instance_service, consumer_key, db):
-        self._application_instance_service = application_instance_service
+        self._application_instance = application_instance_service.get()
         self._consumer_key = consumer_key
         self._db = db
 
     def get_or_create(self, authority_provided_id):
-        """
-        Add the current course to the `course` table if it's not there already.
+        """Add the current course to the `course` table if it's not there already."""
+        return self._get_legacy(authority_provided_id) or self._create_legacy(
+            authority_provided_id
+        )
 
-        :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
-        """
-        return self._get(authority_provided_id) or self._create(authority_provided_id)
+    def upsert(
+        self, authority_provided_id, context_id, name, extra, settings=None
+    ):  # pylint: disable=too-many-arguments
+        course = self._get(authority_provided_id)
+
+        if not course:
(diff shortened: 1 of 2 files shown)


lms/services/course.py
```diff
@@ -6,7 +6,7 @@
 
 class CourseService:
     def __init__(self, application_instance_service, consumer_key, db):
-        self._application_instance_service = application_instance_service
+        self._application_instance = application_instance_service.get()
```
marcospri, line 9: I think this is a better pattern, if there's no AI something is very wrong, better find out soon. We are not handling ConsumerKeyError (but neither were we before). We might need to handle that globally if it ever becomes a problem.

lms/services/course.py
```diff
@@ -1,22 +1,39 @@
...
         self._consumer_key = consumer_key
         self._db = db
 
     def get_or_create(self, authority_provided_id):
-        """
-        Add the current course to the `course` table if it's not there already.
+        """Add the current course to the `course` table if it's not there already."""
```
marcospri, line 14: This is the existing one, just with the renamed `_legacy`  methods

lms/services/course.py
```diff
@@ -1,22 +1,39 @@
...
+        return self._get_legacy(authority_provided_id) or self._create_legacy(
+            authority_provided_id
+        )
 
-        :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
-        """
-        return self._get(authority_provided_id) or self._create(authority_provided_id)
+    def upsert(
```
marcospri, line 19: upserts `Course`, doesn't handle `LegacyCourse`

lms/services/course.py
```diff
@@ -38,15 +55,52 @@ def any_with_setting(self, group, key, value=True):
...
+
+    def _create(
+        self, authority_provided_id, context_id, name, extra, settings=None
+    ):  # pylint: disable=too-many-arguments
+        course = Course(
+            application_instance_id=self._application_instance.id,
+            authority_provided_id=authority_provided_id,
+            lms_id=context_id,
+            lms_name=name,
+            settings=settings or self._new_course_settings(authority_provided_id),
```
marcospri, line 92: For creating new `Courses` we might be creating this from an existing one so it take optional settings but defaults to create new ones from the AI

lms/services/course.py
```diff
@@ -1,22 +1,39 @@
...
+            course = self._create(
+                authority_provided_id,
+                context_id,
+                name,
+                extra,
+                settings,
+            )
+
+        # Update any values that might have changed
+        course.lms_name = name
```
marcospri, line 34: I think this is the only thing we want to be part of the update (settings are not updated on the app as far as I can tell)

seanh: Maybe `extra` should also be updated?

seanh (approved):

PR #2746, opened 2021-05-28
Refine group set configuration UI

robertknight:
This PR is the next step towards aligning the group set configuration UI with the mocks in https://github.com/hypothesis/lms/issues/2523#issuecomment-821160185.

- The first commit updates the layout and text in the main `FilePickerApp` component
- The second commit updates a checkbox label in the group set configuration step of the form
- The last commit fixes a bug where an authorization prompt in the group configuration step remained visible after de-selecting the checkbox

After these changes land there is some further work to be done on the group configuration UI elements (the checkbox, select widget and associated labels).

----

Initial content selection screen:

(_nb. There is a VitalSource option here because I have a feature flag for that enabled locally. Users won't see it in production._)

assignment-details-1

The dialogs that appear during content selection are the same as before:

assignment-details-2

Group configuration screen:

assignment-details-3

This is the screen with UI elements that need additional work, to come in later PRs.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -29,9 +29,9 @@ import GroupConfigSelector from './GroupConfigSelector';
  */
 
 /**
- * Return a human-readable description of the selected content for an assignment.
+ * Return a human-readable description of assignment content.
  *
- * @param {Content} content
+ * @param {Content} content - Type and details of assignment content
  * @return {string}
  */
 function contentDescription(content) {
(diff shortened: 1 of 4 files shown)


lyzadanger (approved):
I want to help you keep this moving! There's no question this is moving towards better and so I want to approve it without too much sweating-the-small-stuff right now.

In terms of styling, I think there is some tuning to be done with spacing/padding proportions, but, again, this is so much better than before that I don't want to throw that speed hump in right now. Probably the thing I'd want to attack soonest is the mixing of relative and absolute units in the layout bits. Not. Urgent.

As I mentioned to you in person, I think this work is highlighting some underlying potential visual patterns, e.g. a consistent layout grid in thirds...we'll circle back on this when it makes sense.

I have not reproduced the original auth-button-hanging-around-inappropriately error. Would you like me to take time to do so?

PR #2748, opened 2021-05-28
Store sections on the Grouping table

marcospri:
# Testing notes

- Truncate your courses and grouping tables to test the insertiong of a new course & sections
```truncate grouping, course;```

- Open an assignment with sections https://hypothesis.instructure.com/courses/125/assignments/873

- Check the grouping table, it should contain one row for the course and for each section. Pay attention to the parent_id relationship, sections should point to the course.

```
created                 | 2021-06-01 09:55:54.832308
updated                 | 2021-06-01 09:55:54.832308
id                      | 12
application_instance_id | 8
parent_id               | 
settings                | {"canvas": {"groups_enabled": true, "sections_enabled": true}, "blackboard": {"files_enabled": true}}
-[ RECORD 2 ]-----------+------------------------------------------------------------------------------------------------------
created                 | 2021-06-01 09:55:55.001108
updated                 | 2021-06-01 09:56:15.790005
id                      | 13
application_instance_id | 8
authority_provided_id   | d2871e5e1c36b58f207d9ef5b22dcb7829f52e61
parent_id               | 12
lms_id                  | 144
lms_name                | Section 1
type                    | canvas_section
settings                | {}
extra                   | null
-[ RECORD 3 ]-----------+------------------------------------------------------------------------------------------------------
created                 | 2021-06-01 09:55:55.001108
updated                 | 2021-06-01 09:56:15.790005
id                      | 14
application_instance_id | 8
authority_provided_id   | f453c6124e4ead4471896f26e896c3e5bbf0bb36
parent_id               | 12
lms_id                  | 145
lms_name                | Section 2
type                    | canvas_section
settings                | {}
extra                   | null
-[ RECORD 4 ]-----------+------------------------------------------------------------------------------------------------------
created                 | 2021-06-01 09:55:55.001108
updated                 | 2021-06-01 09:56:15.790005
id                      | 15
application_instance_id | 8
authority_provided_id   | 83e64baad215f63057387910a1b41debc716a150
parent_id               | 12
lms_id                  | 146
lms_name                | Section 3
type                    | canvas_section
settings                | {}
extra                   | null
```

- Reopen the assignment, not new sections or courses rows added.
- Open a non-sections assignment https://hypothesis.instructure.com/courses/124/assignments/868 there's should be a new course row but not new sections.

### CODE REVIEW

File: lms/models/grouping.py [hunk 1]
```diff
@@ -37,7 +37,7 @@ class Grouping(CreatedUpdatedMixin, BASE):
         sa.ForeignKey("grouping.id", ondelete="cascade"),
         nullable=True,
     )
-    parent = sa.orm.relationship("Grouping")
+    parent = sa.orm.relationship("Grouping", uselist=False)
```
marcospri, line 40: grouping.parent will return an object instead of a list with just one object.

lms/services/grouping.py
```diff
@@ -0,0 +1,67 @@
...
+from lms.models._hashed_id import hashed_id
+
+
+class GroupingService:
+    def __init__(self, db, application_instance_service, course_service):
+        self._db = db
+        self._application_instance = application_instance_service.get()
+        self._course_service = course_service
+
+    def upsert(self, grouping):
```
marcospri, line 11: Courses don't use this upsert yet as they have the added complexity of having to tables.

lms/services/grouping.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Create an HGroup for a course section.
+
+        :param section_name: The name of the section
+        :param tool_consumer_instance_guid: Tool consumer GUID
+        :param context_id: Course id the section is a part of
+        :param section_id: A section id for a section group
+        """
+
+        section_authority_provided_id = hashed_id(
```
marcospri, line 42: This needs more information to be unique

will something like `hashed_id(tool_consumer_instance_guid, context_id, "canvas_section", section_id)` work?

tests/unit/lms/services/grouping_test.py
```diff
@@ -0,0 +1,69 @@
...
+        grouping = svc.upsert(grouping)
+
+        db_grouping = svc._db.query(Grouping).one()
+        assert db_grouping.lms_name == "new_name"
+
+    def test_section_grouping_finds_course(
+        self, svc, course_service, application_instance_service
+    ):
+        application_instance_service.get.return_value = factories.ApplicationInstance()
+        course_service.get.return_value = factories.Course()
```
seanh, line 46: These `return_value`'s could maybe go as defaults on the fixtures in `conftest.py`, they seem pretty universal

tests/unit/lms/services/grouping_test.py
```diff
@@ -0,0 +1,69 @@
...
+        svc._db.flush()
+
+        grouping.lms_name = "new_name"
+
+        grouping = svc.upsert(grouping)
+
+        db_grouping = svc._db.query(Grouping).one()
+        assert db_grouping.lms_name == "new_name"
+
+    def test_section_grouping_finds_course(
```
seanh, line 42: The test should be named after the method that it tests:

```suggestion
    def test_canvas_section_finds_course(
```

seanh (changes requested):
Looks good, some issues with the tests

tests/unit/lms/services/group_info_test.py
```diff
@@ -14,7 +14,7 @@ def test_it_adds_a_new_GroupInfo_if_none_exists(
         self, application_instance, db_session, group_info_svc, params
     ):
         group_info_svc.upsert(
-            factories.HGroup(authority_provided_id=self.AUTHORITY, type="course_group"),
+            factories.Grouping(authority_provided_id=self.AUTHORITY, type="course"),
```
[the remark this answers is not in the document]
marcospri: Oh factory factories.Course already existed and does the right thing :+1:

tests/unit/lms/services/grouping_test.py
marcospri: For course_service that would create some random Course on the DB that conflicts a bit with the rest of the tests

lms/models/grouping.py
```diff
@@ -37,7 +37,7 @@ class Grouping(CreatedUpdatedMixin, BASE):
         sa.ForeignKey("grouping.id", ondelete="cascade"),
         nullable=True,
     )
-    parent = sa.orm.relationship("Grouping")
+    parent = sa.orm.relationship("Grouping", remote_side=[id])
```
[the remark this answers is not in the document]
marcospri: I added a link to the docs somewhere, (commit message?) but should've been here

Basically, as this refers to himself, if you didn't  specify remote_side it wouldn't know what Grouping to return, itself, or the one with primary key `parent_id`.

https://docs.sqlalchemy.org/en/14/orm/self_referential.html#adjacency-list-relationships

lms/services/grouping.py
```diff
@@ -0,0 +1,68 @@
...
+                type=grouping.type,
+            )
+            .one_or_none()
+        )
+        if not db_grouping:
+            self._db.add(grouping)
+        else:
+            # Update any fields that might have changed
+            db_grouping.lms_name = grouping.name
+            db_grouping.extra = grouping.extra
```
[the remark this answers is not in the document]
marcospri: I'd need a separate db_grouping or similar to do the update part, but witth the try: block it might be unbounded so I think it will complicate things.

`.lms_name = .name` was wrong in my code, should be `lms_name` in both ends

tests/unit/lms/services/grouping_test.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_upsert_updates(self, svc):
+        grouping = factories.Grouping()
+        svc._db.flush()
+
+        grouping.lms_name = "new_name"
+
+        grouping = svc.upsert(grouping)
+
+        db_grouping = svc._db.query(Grouping).one()
+        assert db_grouping.lms_name == "new_name"
```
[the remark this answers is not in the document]
marcospri: Yeah, good catch, fixed with `make_transient`

lms/models/grouping.py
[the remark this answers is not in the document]
seanh: :+1:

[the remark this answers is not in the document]
seanh: From the docs it sounds like we might be able to have both `parent` and `children` at once, something like:

```python
children = relationship("Node", backref=backref("parent", remote_side=[id]))
```

Not strictly necessary for this PR, but having `children` as well might come in handy one day

lms/services/grouping.py
```diff
@@ -0,0 +1,68 @@
...
+            .one_or_none()
+        )
+        if not db_grouping:
+            self._db.add(grouping)
+        else:
+            # Update any fields that might have changed
+            db_grouping.lms_name = grouping.lms_name
+            db_grouping.extra = grouping.extra
+
+        return db_grouping or grouping
```
seanh, line 28: So `GroupingService.upsert(grouping)` will either find a matching `Grouping` in the DB and update its `lms_name` and `extra` or, if no matching `Grouping` exists it'll add the giving `grouping` to the DB table.

Seems sensible and I think I like the idea of using the `models.Grouping` class itself as the argument to `upsert()`. It avoids the method having too many arguments like `grouping_svc.upsert(authority_provided_id, type, lms_name, extra)` when those arguments would just be exactly the same as the attributes of `models.Gruoping`. And we can have `upsert_*()` methods for upserting each different type of grouping (Canvas section etc) that call this base `upsert()` method. Those methods don't need to worry about adding the grouping to the DB session or looking for an existing grouping already in the DB. Seems pretty nice :+1:

lms/services/grouping.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def upsert(self, grouping):
+        db_grouping = (
+            self._db.query(Grouping)
+            .filter_by(
+                application_instance=grouping.application_instance,
+                authority_provided_id=grouping.authority_provided_id,
+                type=grouping.type,
+            )
+            .one_or_none()
```
seanh, line 19: I don't think two grouping of different types can have the same `authority_provided_id`, otherwise their groups in the h DB would clash, so the `type` might be unnecessary in this filter.

lms/services/grouping.py
```diff
@@ -0,0 +1,68 @@
...
+        else:
+            # Update any fields that might have changed
+            db_grouping.lms_name = grouping.lms_name
+            db_grouping.extra = grouping.extra
+
+        return db_grouping or grouping
+
+    def canvas_section(
+        self, tool_consumer_instance_guid, context_id, section_id, section_name
+    ):
```
seanh, line 32: I think this should be called `upsert_canvas_section()` to clarify that it actually does an upsert.

Makes sense to take individual arguments here (rather than a single `models.CanvasSection` as the argument) since this is a convenience method that's doing the work of generating the `section_authority_provided_id` and `course_authority_provided_id` from other fields.

lms/views/api/canvas/sync.py
```diff
@@ -57,11 +57,11 @@ def _to_groups(self, sections):
         context_id = self._request.json["course"]["context_id"]
 
         return [
-            HGroup.section_group(
-                section_name=section["name"],
+            self._grouping_service.canvas_section(
```
seanh, line 60: So whenever the sync API is called we end up upserting all the sections into the DB. The sync API is always used when launching an assignment with sections enabled, so this will always upsert the users sections into the DB whenever launching a sections assignment.

Previously this method used to return a list of `models.HGroup` namedtuples but now it's going to return a list of `models.CanvasSection` groupings, and those will be passed to the sync API. But I guess it still works because `CanvasSection` has the same interface as `HGroup` did.

Some naming tweaks could make it a little clearer what's happening here:

* `self._grouping_service.upsert_canvas_section(...)` would make it clearer that a section is being _upserted_ into the DB
* The method that this line is within is called `_to_groups()` but I think it's actually `_to_groupings()` now?

seanh (changes requested):
I've suggested renaming a couple of methods. It might also be nice to have a `children` attribute on the model, but not actually needed

lms/services/grouping.py
seanh: Yeah there's a unique constraint: https://github.com/hypothesis/lms/blob/88698f5ce32e87a59f66d9c87054885bfc02d119/lms/models/grouping.py#L15

seanh (approved):

PR #2763, opened 2021-06-01
Don't init the DB on startup. `make db` uses only alembic migrations

marcospri:
Looks like we use a hybrid approach to the one defined here:

https://alembic.sqlalchemy.org/en/latest/cookbook.html#building-an-up-to-date-database-from-scratch

the suggestion there is to use `create_all` and then stamp the DB. We do that + try to execute the migrations in an attempt to make it work on empty and not empty DB which will fail when new objects are created in them.

An alternative to this PR would be to just don't run the migrations and follow the cookbook suggesting but I think only using the migrations might be useful avoiding writing broken ones. 

We also seem to be using `init` on `includeme`  which would run `create_all` in every environment basically bypassing the alembic workflow.

# Testing

- Starting with an empty DB, swhich to some commit before the last migration, eg: `a2fe3341` and `make db`, all migrations are executed in order

```INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Running upgrade  -> f013a6b67f91, Add the oauth2_unvalidated_credentials table.
INFO  [alembic.runtime.migration] Running upgrade f013a6b67f91 -> d9e4bc3797b3, Add application_instance model.
INFO  [alembic.runtime.migration] Running upgrade d9e4bc3797b3 -> a6a78f338d4a, Add module_item_configurations table.
INFO  [alembic.runtime.migration] Running upgrade a6a78f338d4a -> 58f2693de313, Add application url and datestamp.
INFO  [alembic.runtime.migration] Running upgrade 58f2693de313 -> 3802f4d2ec5c, Add lti_launches table.
INFO  [alembic.runtime.migration] Running upgrade 3802f4d2ec5c -> 51889ae54178, Add oauth credentials to ApplicationInstance.
INFO  [alembic.runtime.migration] Running upgrade 51889ae54178 -> 64d9eacab937, Add oauth support.
INFO  [alembic.runtime.migration] Running upgrade 64d9eacab937 -> efcf8671f4d3, Add the course_groups table.
INFO  [alembic.runtime.migration] Running upgrade efcf8671f4d3 -> 7e4124035651, Add the provisioning column.
INFO  [alembic.runtime.migration] Running upgrade 7e4124035651 -> b99c4d910801, Disabled provisioning for non-whitelisted consumer keys.
INFO  [b99c4d910801_disable_provisioning_for_non_whitelisted__py] Disabled provisioning for 0 non-whitelisted application instances
INFO  [b99c4d910801_disable_provisioning_for_non_whitelisted__py] Left provisioning enabled for 0 whitelisted application instances
INFO  [alembic.runtime.migration] Running upgrade b99c4d910801 -> 83f18f61c76a, Remove the course_groups table.
INFO  [alembic.runtime.migration] Running upgrade 83f18f61c76a -> cfd94b1300ce, Add ApplicationInstance.consumer_key unique constraint.
INFO  [alembic.runtime.migration] Running upgrade cfd94b1300ce -> 7000bb8673c5, Make application_instances columns not nullable.
INFO  [alembic.runtime.migration] Running upgrade 7000bb8673c5 -> f36f9a0aadf4, Add oauth2_token table.
INFO  [alembic.runtime.migration] Running upgrade f36f9a0aadf4 -> 4ab4d06f7e3f, Add unique constraint to module_item_configurations.
INFO  [alembic.runtime.migration] Running upgrade 4ab4d06f7e3f -> db0e29272784, Make module_item_configurations columns not nullable.
INFO  [alembic.runtime.migration] Running upgrade db0e29272784 -> 2fdc9b46320a, Delete no-longer-used DB tables.
INFO  [alembic.runtime.migration] Running upgrade 2fdc9b46320a -> edab0e4610e0, Add the lis_result_sourcedid table.
INFO  [alembic.runtime.migration] Running upgrade edab0e4610e0 -> a930adadac74, Add group_info table.
INFO  [alembic.runtime.migration] Running upgrade a930adadac74 -> 37710e6bcb66, Add group_info.info column.
INFO  [alembic.runtime.migration] Running upgrade 37710e6bcb66 -> f0859cd029fe, Add application_instances.settings column.
INFO  [alembic.runtime.migration] Running upgrade f0859cd029fe -> 3517eb6254e7, Add the course table and course_groups_exported_from_h tables.
INFO  [alembic.runtime.migration] Running upgrade 3517eb6254e7 -> 7f9824ded172, Back-fill the course and course_groups_exported_from_h tables.
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Commit 1000
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Commit 2000
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Commit 3000
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Commit 4000
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Inserted 0 rows into course table with sections enabled
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Inserted 4230 rows into course_groups_exported_from_h table
INFO  [db3779df3f44_back_fill_course_and_course_groups_exported_from_h_tables_py] Inserted 4230 rows in total
INFO  [alembic.runtime.migration] Running upgrade 7f9824ded172 -> 5086e8b137b9, Add ApplicationInstance.settings and Course.settings server defaults and make them not-nullable.
INFO  [alembic.runtime.migration] Running upgrade 5086e8b137b9 -> 5f80062abcf4, Add LMS columns to the appication_instances table.
```

- Switch to master's head and run make db again

```INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Running upgrade 5f80062abcf4 -> bb11b5b06274, Adds new grouping table.
````

- Try the same procedure on the master branch, make db will fail the second time

### CODE REVIEW

File: lms/db/__init__.py [hunk 1]
```diff
@@ -104,20 +104,6 @@ def init(engine, drop=False):
...
-def _stamp_db(engine):  # pragma: nocover
-    """
-    Stamp the database with the latest revision if it isn't already stamped.
-
-    This is convenient in development environments to automatically stamp a new
-    database after initializing it.
-
-    """
-    try:
-        engine.execute("select 1 from alembic_version")
```
marcospri, line 116: Not stamping if there's already an `alembic_version` table means that `create_all` would create all new tables but alembic, which we run on `make db` wouldn't know, (alembic_version didn't change) and would try to create itself.

lms/db/__init__.py
```diff
@@ -160,10 +146,6 @@ def includeme(config):
     # Create the SQLAlchemy engine and save a reference in the app registry.
     engine = make_engine(config.registry.settings)
     config.registry["sqlalchemy.engine"] = engine
-    init(engine)
```
marcospri, line 163: I think this should've been also inside the `if config.registry.settings["dev"]:`

jon-betts (commented):
This certainly makes sense to me? I'm not a big expert on what's going on here so there could be implications I'm not aware of.

The procedure works as described and fails as described in main. So looks good to me?

seanh:
Do we know what exactly is going wrong? I think it might have to do with model changes being merged into git before their DB migrations? So you checkout `a2fe3341` and run `make db` and it creates the DB schema from the models which already contain the `grouping` table. But `a2fe3341` does not have the migration that adds the `grouping` table. Then you checkout `master` and run `make db`, master _does_ have the migration that adds the `grouping` table so it tries to run it but your DB already contains the `grouping` table because it was created from the models.

`make db` is supposed to:

1. Initialize your DB from scratch with the latest schema if your DB isn't initialized
2. If your DB _is_ already initialized the `make db` should just run any migrations that you haven't already run, bringing your schema up to date. (This is why it's necessary to stamp the DB in step 1.)

When initializing a DB from scratch I think we probably do want to create the DB schema from the models, not by running all the migrations in order. For example I could be working on a feature branch where I've made changes to the models and haven't written a migration for those changes yet. If `make db` recreates the DB from the models it'll do the right thing. If it runs the DB migration it'll create a DB that's out of sync with my models.

marcospri:
`make db` it's broken for new migrations added to the repository, the case everybody outside the backed team will encounter more often.

PR #2767, opened 2021-06-02
Canvas groups backend

marcospri:
# Testing notes

- Truncate your courses and grouping tables to test the insertion of a new course & groups
```truncate grouping, course;```

- Enable groups on the application instance

http://localhost:8001/admin/instance/Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a/

- Open an assignment with groups https://hypothesis.instructure.com/courses/125/assignments/873

- Check the grouping table, it should contain one row for the course and for each group (if opened as a teacher). Pay attention to the parent_id relationship, sections should point to the course.

```
-[ RECORD 1 ]-----------+-------------------------------------------------------------------------------------------------------
created                 | 2021-06-07 12:47:11.497763
updated                 | 2021-06-07 12:47:11.497763
id                      | 130
application_instance_id | 8
parent_id               | 
settings                | {"canvas": {"groups_enabled": true, "sections_enabled": true}, "blackboard": {"files_enabled": false}}
-[ RECORD 2 ]-----------+-------------------------------------------------------------------------------------------------------
created                 | 2021-06-07 12:47:11.619672
updated                 | 2021-06-07 12:47:11.619672
id                      | 131
application_instance_id | 8
authority_provided_id   | 381c455de84ae2f6d2fc84142f4e898b29d748f3
parent_id               | 130
lms_id                  | 132
type                    | canvas_group
lms_name                | another group
settings                | {}
extra                   | {"group_set_id": 121}
-[ RECORD 3 ]-----------+-------------------------------------------------------------------------------------------------------
created                 | 2021-06-07 12:47:11.619672
updated                 | 2021-06-07 12:47:11.619672
id                      | 132
application_instance_id | 8
authority_provided_id   | 102a25279bc609179f58cb4bf5c50d7dfc6f0929
parent_id               | 130
lms_id                  | 130
type                    | canvas_group
lms_name                | others
settings                | {}
extra                   | {"group_set_id": 121}
-[ RECORD 4 ]-----------+-------------------------------------------------------------------------------------------------------
created                 | 2021-06-07 12:47:11.619672
updated                 | 2021-06-07 12:47:11.619672
id                      | 133
application_instance_id | 8
authority_provided_id   | 802107eead62d51dcdef0f317f0eb0b9868eaa7c
parent_id               | 130
lms_id                  | 129
type                    | canvas_group
lms_name                | Seans
settings                | {}
extra                   | {"group_set_id": 121}
-[ RECORD 5 ]-----------+-------------------------------------------------------------------------------------------------------
created                 | 2021-06-07 12:47:11.619672
updated                 | 2021-06-07 12:47:11.619672
id                      | 134
application_instance_id | 8
authority_provided_id   | e8c66a347a5f547ae87928d4e92e5045908f8d54
parent_id               | 130
lms_id                  | 131
type                    | canvas_group
lms_name                | test
settings                | {}
extra                   | {"group_set_id": 121}
```

- The client should show all the groups in the top left menu.

- Open speed grader,  it should open the client for students that launched the assignment.

- Open the same assignment as a student. The menu should show only one group (no selection possible).

### DIFF
--- a/lms/models/__init__.py
+++ b/lms/models/__init__.py
@@ -5,7 +5,7 @@ from lms.models.course import LegacyCourse
 from lms.models.course_groups_exported_from_h import CourseGroupsExportedFromH
 from lms.models.grading_info import GradingInfo
 from lms.models.group_info import GroupInfo
-from lms.models.grouping import CanvasSection, Course, Grouping
+from lms.models.grouping import CanvasGroup, CanvasSection, Course, Grouping
 from lms.models.h_user import HUser
 from lms.models.lti_launches import LtiLaunches
 from lms.models.lti_user import LTIUser, display_name
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -337,6 +337,7 @@ class JSConfig:
                 "lis_result_sourcedid": lis_result_sourcedid,
                 "lis_outcome_service_url": lis_outcome_service_url,
                 "learner_canvas_user_id": self._request.params["custom_canvas_user_id"],
+                "group_set": self._request.params.get("group_set"),
                 **kwargs,
             },
         }
@@ -531,6 +532,7 @@ class JSConfig:
         if "learner_canvas_user_id" in req.params:
             sync_api_config["data"]["learner"] = {
                 "canvas_user_id": req.params["learner_canvas_user_id"],
+                "group_set": req.params.get("group_set"),
             }
 
         return sync_api_config
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -210,10 +210,7 @@ class JSConfig:
             },
             "canvas": {
                 "enabled": self._canvas_files_available(),
-                "groupsEnabled": self._application_instance().settings.get(
-                    "canvas", "groups_enabled"
-                )
-                or False,
+                "groupsEnabled": self._context.canvas_groups_enabled,
                 # The "content item selection" that we submit to Canvas's
                 # content_item_return_url is actually an LTI launch URL with
                 # the selected document URL or file_id as a query parameter. To
@@ -518,6 +515,7 @@ class JSConfig:
                 "course": {
                     "context_id": req.params["context_id"],
                     "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+                    "group_set": req.params.get("group_set"),
                 },
                 "group_info": {
                     key: value
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -498,7 +498,10 @@ class JSConfig:
         return [self._context.h_group.groupid(self._authority)]
 
     def _sync_api(self):
-        if not self._context.canvas_sections_enabled:
+        if (
+            not self._context.canvas_sections_enabled
+            and not self._context.canvas_groups_enabled
+        ):
             return None
 
         req = self._request
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -22,6 +22,9 @@ class LTILaunchResource:
         """Return the context resource for an LTI launch request."""
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
+        self._application_instance_service = self._request.find_service(
+            name="application_instance"
+        )
 
     def get_or_create_course(self):
         """Get the course this LTI launch based on the request's params."""
@@ -118,12 +121,8 @@ class LTILaunchResource:
             # Canvas course sections feature was released.
             return False
 
-        application_instance_service = self._request.find_service(
-            name="application_instance"
-        )
-
         try:
-            application_instance = application_instance_service.get()
+            application_instance = self._application_instance_service.get()
         except ConsumerKeyError:
             return False
 
@@ -143,16 +142,12 @@ class LTILaunchResource:
     @property
     def canvas_groups_enabled(self):
         """Return True if Canvas groups is enabled for this request."""
-        application_instance_service = self._request.find_service(
-            name="application_instance"
-        )
-
         try:
-            application_instance = application_instance_service.get()
+            application_instance = self._application_instance_service.get()
         except ConsumerKeyError:
             return False
 
-        return application_instance.settings.get("canvas", "groups_enabled") or False
+        return bool(application_instance.settings.get("canvas", "groups_enabled"))
 
     def _course_extra(self):
         """Extra information to store for courses."""
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -140,6 +140,20 @@ class LTILaunchResource:
 
         return legacy_course.settings.get("canvas", "sections_enabled")
 
+    @property
+    def canvas_groups_enabled(self):
+        """Return True if Canvas groups is enabled for this request."""
+        application_instance_service = self._request.find_service(
+            name="application_instance"
+        )
+
+        try:
+            application_instance = application_instance_service.get()
+        except ConsumerKeyError:
+            return False
+
+        return application_instance.settings.get("canvas", "groups_enabled") or False
+
     def _course_extra(self):
         """Extra information to store for courses."""
         extra = {}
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: lms/resources/lti_launch.py [hunk 1]
```diff
@@ -140,6 +140,20 @@ def canvas_sections_enabled(self):
...
+        application_instance_service = self._request.find_service(
+            name="application_instance"
+        )
+
+        try:
+            application_instance = application_instance_service.get()
+        except ConsumerKeyError:
+            return False
+
+        return application_instance.settings.get("canvas", "groups_enabled") or False
```
seanh, line 155: Could maybe be:

```suggestion
        return bool(application_instance.settings.get("canvas", "groups_enabled"))
```

seanh (commented):
:+1: General approach here looks good

lms/resources/_js_config.py
```diff
@@ -501,7 +499,10 @@ def _groups(self):
         return [self._context.h_group.groupid(self._authority)]
 
     def _sync_api(self):
-        if not self._context.canvas_sections_enabled:
+        if (
```
marcospri, line 502: Sync api either sections/groups enabled

lms/resources/_js_config.py
```diff
@@ -518,6 +519,7 @@ def _sync_api(self):
                 "course": {
                     "context_id": req.params["context_id"],
                     "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+                    "group_set": req.params.get("group_set"),
```
marcospri, line 522: For sync api

lms/resources/_js_config.py
```diff
@@ -530,6 +532,7 @@ def _sync_api(self):
         if "learner_canvas_user_id" in req.params:
             sync_api_config["data"]["learner"] = {
                 "canvas_user_id": req.params["learner_canvas_user_id"],
+                "group_set": req.params.get("group_set"),
```
marcospri, line 535: For speedgrader

lms/services/group_info.py
```diff
@@ -18,6 +18,7 @@ class GroupInfoService:
     GROUPING_TYPES = {
         "course": "course_group",
         "canvas_section": "section_group",
+        "canvas_group": "canvas_group_group",
```
marcospri, line 21: This is not a great name but the other two already are on the DB and using _grouping on the last one might be more confusing.

lms/services/grouping.py
```diff
@@ -58,6 +58,45 @@ def upsert_canvas_section(
...
+
+        :param tool_consumer_instance_guid: Tool consumer GUID
+        :param context_id: Course id the section is a part of
+        :param group_id: A section id for a section group
+        :param group_name: The name of the section
+        :param group_set_id: Id of the canvas group set this group belongs to
+        """
+
+        group_authority_provided_id = hashed_id(
+            tool_consumer_instance_guid, context_id, "canvas_group", group_id
```
marcospri, line 80: Needed `canvas_group` to generate different ID from sections

lms/validation/_api.py
```diff
@@ -32,6 +32,8 @@ class APIRecordSpeedgraderSchema(JSONPyramidRequestSchema):
     typically encodes the assignment context and LMS user.
     """
 
+    group_set = fields.Int(required=False)
```
marcospri, line 35: Present on speedgrader for a group assigment.

Needs doc string.

lms/models/__init__.py
```diff
@@ -5,7 +5,7 @@
 from lms.models.course_groups_exported_from_h import CourseGroupsExportedFromH
 from lms.models.grading_info import GradingInfo
 from lms.models.group_info import GroupInfo
-from lms.models.grouping import CanvasSection, Course, Grouping
+from lms.models.grouping import CanvasGroup, CanvasSection, Course, Grouping
```
seanh, line 8: The `CanvasGroup` import was just forgotten on master?

marcospri: It was not used anywhere (Canvas groups not yet implemented) I think `CanvasGroups` was prematurely added to master (but it was there to demonstrate `Grouping` usage)

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -509,9 +510,10 @@ def test_it(self, sync, pyramid_request, GroupInfo):
     def test_it_adds_learner_canvas_user_id_for_SpeedGrader_launches(self, sync):
         assert sync["data"]["learner"] == {
             "canvas_user_id": "test_learner_canvas_user_id",
+            "group_set": None,
         }
 
-    def test_its_None_if_section_groups_isnt_enabled(self, sync):
+    def test_its_None_if_section_and_groups_arent_enabled(self, sync):
```
seanh, line 516: :+1: Good catch

tests/unit/lms/views/api/canvas/sync_test.py
```diff
@@ -1,41 +1,250 @@
...
+@pytest.mark.usefixtures("user_is_instructor")
+def test_get_canvas_groups_instructor(pyramid_request):
+    # pylint: disable=protected-access
+    with mock.patch.object(
+        Sync, "_canvas_course_groups", autospec=True
+    ) as _canvas_course_groups:
+        groups = Sync(pyramid_request)._get_canvas_groups()
+
+        _canvas_course_groups.assert_called_once()
+        assert groups == _canvas_course_groups.return_value
```
seanh, line 102: The necessity of the `mock.patch.object()`'s might support the suggestion that these `_canvas_()` methods belong in the service

seanh (changes requested):
Code review looks good, only real suggestion is some wrong docstrings and some view helper methods that look like they'd probably be better as `CanvasAPIClient` service methods.

In trying to test this locally: after creating a groups assignment, when launching it I'm seeing a spinner in the top-right of the client spinning forever and the groups never appear in the client's groups menu. Not sure what's going wrong

lms/views/api/canvas/sync.py
```diff
@@ -26,19 +30,52 @@ def sync(self):
...
+        ).get()
+        if not application_instance.settings.get("canvas", "groups_enabled"):
+            return False
+
+        try:
+            self.group_set
+        except (KeyError, ValueError, TypeError):
+            return False
+        else:
+            return True
```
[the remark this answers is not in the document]
marcospri: Changed to a function :+1:

marcospri (commented):
@seanh 

- I moved the methods from the sync API view to the canvas API service, they make more sense there. I generalized them a bit (changed the names and they optionally take some of the arguments) so they could potentially be used in another context.

- Those methods all return list of groups (and they return an empty list for no groups, instead of None) which I think makes a nicer API now.

- In the context they are used, eg, "groups of this particular user in this particular group set" I don't think it could return more than one group (at least it doesn't seem possible to set that up on Canvas's UI

- This PR is broken for empty group set and some other edge cases, I think some of those will need UI to report a human readable message. We could tackle all of those in this separate PR https://github.com/hypothesis/lms/pull/2791

lms/services/grouping.py
marcospri: Yep, good idea, added a test that adds section & group with the same id

lms/services/canvas_api/client.py
```diff
@@ -276,15 +276,80 @@ def group_category_groups(self, group_category_id):
...
+            if group_category_id and group["group_category_id"] != group_category_id:
+                continue
+
+            for user in group["users"]:
+                if user["id"] == user_id:
+                    groups.append(group)
+                    if group_category_id:
+                        # If we are filtering by `group_category_id` there
+                        # can be only one group the user belongs to in the same group set
+                        return groups
```
seanh, line 335: I think we should maybe remove this. The code will be simpler. Even if we think it's correct, do we need to encode this assumption that a user can only belong to one group?

If it turns out (now or in the future) that a user can somehow belong to more than one group then:

* If we just return the list of multiple groups that the user belongs to then the client will just show all of those groups in its menu, which is the same thing that happens when a user belongs to multiple sections
* On the other hand if we've coded in the assumption that a user can only belong to one group then we'll end up only showing the first group even if the user belongs to multiple, which is weird

So even though this probably doesn't make a difference (since users probably can't belong to multiple groups anyway) it's safer not to assume it and might actually make the code simpler.

What do you think?

seanh (approved):
:+1:

Issue #2029, opened 2020-08-19
Blackboard Files frontend changes


seanh:
## Test site

We have a Blackboard test site at https://aunltd-test.blackboard.com/. There are test teacher and student accounts in 1Password. There is a course named **Developer Test Course with Original Course View** and a test assignment named **localhost (make devdata) Blackboard Files Test Assignment**.

Use `blackboardteacher`, `blackboardstudent` accounts
The Dev course is here
https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline

You will need to run `make devdata` in the LMS app to get this stuff working with your local LMS app.

Once you have configured a Blackboard assignment (choosing the file for the assignment) this gets saved in the DB and if you launch the assignment again it'll just launch. You won't be able to select the file again. To select the file again you need to delete the assignment config from your DB: `tox -qe dockercompose -- exec postgres psql -U postgres -c "DELETE FROM module_item_configurations WHERE tool_consumer_instance_guid = '041c05b262b54282a0d9b61c42413ec0' and resource_link_id = '_39_1';"`

## What is already implemented for Blackboard files (backend)

1. Blackboard Files is only enabled if the `blackboard.files_enabled` setting is turned on for the application instance. It's currently turned on for the test application instances on our test Blackboard site but not for any others. So real users won't see the Blackboard Files feature currently.

2. When the feature is enabled and an instructor launches an unconfigured assignment in Blackboard, the backend adds Blackboard file picker config to the JS config:

   ```json
   "filePicker":{
     "blackboard":{
       "enabled":true,
       "listFiles":{
         "authUrl":"http://localhost:8001/api/blackboard/oauth/authorize",
         "path":"/api/blackboard/courses/7c292e005ad54189869cbfb3aa896310/files"
       }
   },
   ```

* The Blackboard `listFiles` API and `authUrl` have been implemented and work the same as the Canvas ones. The `authUrl` is actually real (or it will be when <https://github.com/hypothesis/lms/pull/2739> is merged). The `listFiles` API returns a fake list of files.

* When a _configured_ Blackboard Files assignment is launched the backend sets the `api.viaUrl` setting in the JS config. This is the same setting as used when launching Canvas files assignments but it gets Blackboard-specific values (which should be opaque to the frontend):

  ```json
  "api": {
     "viaUrl":{
        "authUrl":"http://localhost:8001/api/blackboard/oauth/authorize",
[...]

PR #2771, opened 2021-06-03
Add Blackboard file picker dialog and button

LMS007:
Add Blackboard file picker dialog and button 

Add a Blackboard file picker similar to the Canvas PDF file picker. The Blackboard file picker enabled flag is passed in from config.blackboard.enabled. The file list is passed from config.blackboard.listFile.

The selected file is sent by passing a hidden field called document_url in the POST request and is set by passing a URLContent object to FilePickerFormField's `content` prop. Note that this is slightly different than the Canvas picker which uses a FileContent object instead and gets passed via content_items hidden field.

--------------

_Based from https://github.com/hypothesis/lms/pull/2636_

Test with `blackboardteacher`

Test content link:

Screen Shot 2021-06-02 at 5 49 06 PM

Screen Shot 2021-06-02 at 5 49 10 PM 

POST body
```

{
  ...
  content_items: {"@context":"http://purl.imsglobal.org/ctx/lti/v1/ContentItem","@graph":[{"@type":"LtiLinkItem","mediaType":"application/vnd.ims.lti.v1.ltilink","url":"http://localhost:8001/lti_launches?url=blackboard%3A%2F%2Fcontent-resource%2F123"}]}
  document_url: blackboard://content-resource/123
}

```
I am still unsure about this line of code which adds the `content` prop as a param to the http://localhost:8001/lti_launches url for the `content_items` key passed in the POST request.
 https://github.com/hypothesis/lms/blob/6360fb6b6161e9c5c54c02e08e3eaecd61b28703/lms/static/scripts/frontend_apps/components/FilePickerFormFields.js#L43

I think that the server does not need `url=` param attached here and only cares about `document_url` field in the blackboard case. As it stands now, it appears to be begin to pass it along, and it might further complicate this bit of code to have yet another exception to ignore the `url` param for this case.  I'm inclined to just leave this as it until we finalize the mechanism between the server and client for blackboard. I could use some direction on this.

fixes https://github.com/hypothesis/lms/issues/2029

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -18,7 +18,7 @@ import URLPicker from './URLPicker';
  * @prop {string} title
  * @prop {Error} error
  *
- * @typedef {'lmsFile'|'url'|null} DialogType
+ * @typedef {'blackboardFile'|'canvasFile'|'url'|null} DialogType
  *
  * @typedef {import('../utils/content-item').Content} Content
  */
@@ -43,7 +43,11 @@ export default function ContentSelector({
   const {
     api: { authToken },
     filePicker: {
-      canvas: { enabled: canvasEnabled, listFiles: listFilesApi },
+      blackboard: {
+        enabled: blackboardFilesEnabled,
+        listFiles: blackboardListFilesApi,
+      },
+      canvas: { enabled: canvasFilesEnabled, listFiles: listFilesApi },
       google: {
         clientId: googleClientId,
         developerKey: googleDeveloperKey,
@@ -95,9 +99,16 @@ export default function ContentSelector({
   };
 
   /** @param {File} file */
-  const selectLMSFile = file => {
+  const selectCanvasFile = file => {
     cancelDialog();
     onSelectContent({ type: 'file', file });
+  };
+
+  /** @param {File} file */
+  const selectBlackboardFile = file => {
+    cancelDialog();
+    // file.id shall be a url of the form blackboard://content-resource/{file_id}
+    onSelectContent({ type: 'url', url: file.id });
   };
 
   let dialog;
@@ -105,13 +116,23 @@ export default function ContentSelector({
     case 'url':
       dialog = <URLPicker onCancel={cancelDialog} onSelectURL={selectURL} />;
       break;
-    case 'lmsFile':
+    case 'canvasFile':
       dialog = (
         <LMSFilePicker
           authToken={authToken}
           listFilesApi={listFilesApi}
           onCancel={cancelDialog}
-          onSelectFile={selectLMSFile}
+          onSelectFile={selectCanvasFile}
+        />
+      );
+      break;
+    case 'blackboardFile':
+      dialog = (
+        <LMSFilePicker
+          authToken={authToken}
+          listFilesApi={blackboardListFilesApi}
+          onCancel={cancelDialog}
+          onSelectFile={selectBlackboardFile}
         />
       );
       break;
@@ -161,13 +182,22 @@ export default function ContentSelector({
           >
             Enter URL of web page or PDF
           </LabeledButton>
-          {canvasEnabled && (
+          {canvasFilesEnabled && (
             <LabeledButton
-              onClick={() => selectDialog('lmsFile')}
+              onClick={() => selectDialog('canvasFile')}
               variant="primary"
-              data-testid="lms-file-button"
+              data-testid="canvas-file-button"
             >
               Select PDF from Canvas
+            </LabeledButton>
+          )}
+          {blackboardFilesEnabled && (
+            <LabeledButton
+              onClick={() => selectDialog('blackboardFile')}
+              variant="primary"
+              data-testid="blackboard-file-button"
+            >
+              Select PDF from Blackboard
             </LabeledButton>
           )}
           {googlePicker && (
--- a/lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
@@ -27,11 +27,18 @@ describe('ContentSelector', () => {
     fakeConfig = {
       api: { authToken: 'dummy-auth-token' },
       filePicker: {
+        blackboard: {
+          enabled: true,
+          listFiles: {
+            authUrl: 'https://lms.anno.co/blackboard/authorize',
+            path: 'https://lms.anno.co/api/blackboard/files',
+          },
+        },
         canvas: {
           enabled: true,
           listFiles: {
-            authUrl: 'https://lms.anno.co/authorize',
-            path: 'https://lms.anno.co/api/files',
+            authUrl: 'https://lms.anno.co/canvas/authorize',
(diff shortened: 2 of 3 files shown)


robertknight:
Is there anything left in https://github.com/hypothesis/lms/pull/2636 which is not handled in this PR, or can that proof-of-concept one by closed?

LMS007:
> Is there anything left in #2636 which is not handled in this PR, or can that proof-of-concept one by closed?


@robertknight 

Not yet! This is not fully working. The next bits are to work out how to get the `file.id` passed as a hidden field for the `document_url`. I think we can make some small changes to ContentSelector to pipe that through and get FilePickerFormFields to pick it up. I'm still sorting this out.

robertknight (commented):
The UI changes and tests look good. I added a few minor comments/suggestions on the code. Once the changes have been made to populate the hidden form fields appropriately when submitting the form, this is good to go. Looking at the backend code in lms/views/api/blackboard/files.py the `id` of Blackboard files is in fact a URL with a custom scheme (`blackboard://content-resource/{id}`). This is why Sean's prototype just submitted the value of `content.file.id` in the existing `document_url` field. One option would be to do the same thing in this PR, and then we can revise both the frontend and backend later once the final mechanism to submit the file ID has been decided upon.

robertknight:
Note for future reference: When testing this I initially ran into an "unknown client ID" error in the Blackboard authorization popup. To fix this it was necessary to update my local database and environment variables using `make devadata` and then to restart my LMS dev server.

LMS007:
> Note for future reference: When testing this I initially ran into an "unknown client ID" error in the Blackboard authorization popup. To fix this it was necessary to update my local database and environment variables using `make devadata` and then to restart my LMS dev server.

awe, that was covered in Seans first POC, I forgot to repost that. Thanks

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -127,48 +134,80 @@ describe('ContentSelector', () => {
...
+        name: 'Blackboard',
+        buttonTestId: 'blackboard-file-button',
+        files: () => fakeConfig.filePicker.blackboard.listFiles,
+      },
+    ].forEach(test => {
+      it(`shows LMS file dialog when "Select PDF from ${test.name}" is clicked`, () => {
+        const wrapper = renderContentSelector();
+
+        assert.isFalse(
+          wrapper.find('.ContentSelector__loading-backdrop').exists()
```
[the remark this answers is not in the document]
LMS007: roger that

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -95,23 +99,39 @@ export default function ContentSelector({
...
 
   /** @param {File} file */
-  const selectLMSFile = file => {
+  const selectCanvasSFile = file => {
     cancelDialog();
     onSelectContent({ type: 'file', file });
   };
 
+  /** @param {File} file */
+  const selectBlackBoardFile = file => {
```
LMS007, line 108: This is the newest piece that was added since the draft which adds the `document_url` param by passing up a `URLContent` object to FilePickerApp, and then that of course goes to FilePickerFormFields . The draft version never fully worked because it was not passing the hidden `document_url` parm.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -95,23 +99,39 @@ export default function ContentSelector({
   };
 
   /** @param {File} file */
-  const selectLMSFile = file => {
+  const selectCanvasSFile = file => {
```
robertknight, line 102:
```suggestion
  const selectCanvasFile = file => {
```

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -130,48 +140,80 @@ describe('ContentSelector', () => {
...
+        file: { id: 123 },
+        result: {
+          type: 'file',
+          file: { id: 123 },
+        },
+      },
+      {
+        name: 'blackboard',
+        dialogName: 'blackboardFile',
+        file: { id: 'blackboard://content-resource/123' },
```
robertknight, line 196: 👍 for using "realistic" looking file IDs here in the sense that they match the shape of values from the actual app.

lms/static/scripts/frontend_apps/components/ContentSelector.js
robertknight, line 108:
```suggestion
  const selectBlackboardFile = file => {
```

For consistency with the rest of the code and Blackboard's product documentation.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -95,23 +99,39 @@ export default function ContentSelector({
...
     cancelDialog();
     onSelectContent({ type: 'file', file });
   };
 
+  /** @param {File} file */
+  const selectBlackBoardFile = file => {
+    cancelDialog();
+    onSelectContent({ type: 'url', url: file.id });
```
robertknight, line 110: Can you add a comment to indicate the expected format of `file.id` values here.

lms/static/scripts/frontend_apps/components/ContentSelector.js
robertknight: To elaborate on this comment to check my understanding:

- When the `FilePickerApp` form is submitted, the frontend indicates to Blackboard the chosen URL in a hidden `document_url` form field rendered by `FilePickerFormFields`
- When the user selects a Blackboard file, this is currently indicated by using a `blackboard://content-resource/{file_id}` URL, rather than a separate form field
- Therefore in order to render the correct form fields we need to separate the code paths for Canvas and Blackboard files here. For Blackboard files we behave as if the user selected a URL, but instead set the URL to the `file.id` value provided by the backend, which is already a `blackboard://content-resource/{file_id}` URL

robertknight (changes requested):
I requested some minor fixes. Otherwise LGTM. Please give me a ping as soon as you've had a chance to look at those.

The form fields that the frontend submits here (`document_url` with the `blackboard://...` URL) match what the backend currently expects AFAIK, so I think we should land this. I had seen some discussion in Slack (with @seanh?) about potentially using different fields to indicate a Blackboard file. Do you know if any conclusion was reached about that?

lms/static/scripts/frontend_apps/components/ContentSelector.js
LMS007: good catch! thx

LMS007:
"Therefore in order to render the correct form fields we need to separate the code paths for Canvas and Blackboard files here. For Blackboard files we behave as if the user selected a URL, but instead set the URL to the file.id value provided by the backend, which is already a blackboard://content-resource/{file_id} URL"

Thats correct Rob, it's not perfect because the file.id is the url--which is a little awkward, but it works for now and we're going to stick with using `document_url` on the backend for the foreseeable future.

LMS007:
- Fixed variable typos
- Added a comment before `onSelectContent({ type: 'url', url: file.id });` to explain the file.id format.

I also confirmed with Sean, we're going to use `document_url` param with the value "blackboard://..." for the time being.

Issue #2768, opened 2021-06-02
Store records of files in a tool consumer agnostic way


jon-betts:
For: https://github.com/hypothesis/lms/issues/2764

Primarily for Canvas course copy, but potentially for other purposes too, we should keep a record of the files we serve. 

### What we should store

At a minimum we should store:

 * The course id (context_id)
 * File id
 * File name
 * File size (if available?)

Some other things which would be good if available:

 * File hash - this would be very nice
 * Anything else that could be useful for file matching
 * The assignment id (resource_link_id) <- ??? Does this make sense? Is this a last seen type value? Are files actually always in a specific course/assignment etc? Presumably not

### When we should store it

 * Whenever we access a file for Canvas to serve it
 * Whenever we pull a full list of files from Canvas (optional?)
 * We should update the table whenever we get new data (i.e. if the filename changed)

### Other features

 * The table shouldn't have anything particularly Canvas specific to it so we can use it with other tool consumers
 * In particular, although Canvas file ids are globally unique, we can't rely on that at a schema level

## Release notes

**This should be released as soon as is practical**, as this sets the time limit for which assignments we can fix after a course copy without any intervention. Any assignments which are accessed after this is released will have the file data stored, and although we won't use it yet, these will be able to be fixed without requiring the user to launch the old content.

For this reason we should see if it's better / easier / quicker to release part of this work first. For example if it's simpler to **release the single file access part first**, we should do that. This is also the most likely to be triggered in the regular course of activity. Listing the full contents of the course only happens in special circumstances.

PR #2777, opened 2021-06-03
Add a file table and update it when we retrieve a list of Canvas files

jon-betts:
For: https://github.com/hypothesis/lms/issues/2768

Use a custom event and subscriber to store the data to try and keep view-ish looking stuff out of the API class. Not sure how I feel about that yet.

This adds a `File` table which is keyed to the application instance. We could probably get away with making this a shared table, but I feel like that's going against the grain of what we're doing nowadays.

### Some things I learned along the way

* We can't easily store information when getting single files as we don't have that information (bummer)
* For single files we ask for a public URL but there are no other details
* We could make another call, but this might be a new end-point (and so new permission?) or trigger the full file pull down
* Recently in `h` we removed `db.init()` in favour of using alembic exclusively, that might not really be very good for development, as you can't change the DB code and have it be present in the DB without constantly updating a migration for it? That's not really how we tend to work (write the model class first, migration when it's done).

### The approach taken in this PR

 * I've intercepted the call to list_files so there are no ways of accessing this data that don't cause the data to be stored in the DB
 * This is a little icky, because the API could be doing double duty here (getting and persisting)
 * To get around this I've made the API register a custom `FilesDiscoveredEvent` instead which technically decouples it
 * It's not that clean, as we need the request to get our work done, so we have to pass that down
 * We could have the API call the DB code directly, or have it look up a service etc, but it all feels a bit icky and view-ish to be doing in the bowels of the API. We could make the call in the view itself, but some views call methods which result in a call to list_files but don't actually have access to the underlying data at any point.
 
The code could be much nicer if we could use `get_current_request()` inside the subscriber so the caller is really decoupled from the request object and can raise a more plain event, but:

 * You can't really do that as you have to call request.registry.notify to send the event anyway
 * The docs heavily imply you really shouldn't use that function

This feels like it's a nice-ish pattern, as we could potentially notify that we've found all sorts of things in our API's and code, and selectively chose which to persist etc. It's also relatively clear how you'd slot this in other places (not that anything else is particularly hard).

Even though it's not technically different from calling the code directly or a service (it's all synchronous) I think this has the right semantics. The API raises the event, but doesn't care if anyone is listening. We could conceivably have the event handled out of band (with celery), discard it, or ignore it's failure and the calling code should be none the wiser.

### Testing notes

 * To trigger this you need to be a teacher, and access an assignment with a Canvas file
 * Once done try: `make sql`
 * `select * from files;`
 * You should also be able to re-load the assigment (causing an update) without anything going wrong

### DIFF
--- a/lms/db/_session_extension.py
+++ b/lms/db/_session_extension.py
@@ -1,0 +1,29 @@
+class SessionExtension:
+    """
+    A self returning descriptor to adding functionality to a session.
+
+    To use this, sub-class it and then add your object to the session:
+
+    class CustomSession(Session):
+        my_extension = MyExtension()
+
+    The `name` and `session` attributes will be available to your instance.
+    """
+
+    session = None
+    name = None
+
+    def __set_name__(self, owner, name):
+        # Store the name for more informative errors
+        self.name = name
+
+    def __get__(self, instance, owner):
+        if not instance:
+            raise TypeError(
+                f"You can only call '{self.name}' on instances not the bare "
+                f"class: '{owner.__name__}'"
+            )
+
+        # Store the session and return ourselves
+        self.session = instance
+        return self
--- a/lms/events/subscriber.py
+++ b/lms/events/subscriber.py
@@ -1,0 +1,29 @@
+from pyramid.events import subscriber
+
+from lms.models.file import File
+
+
+class FilesDiscoveredEvent:
+    """Notify that a file has been found in an LMS."""
+
+    def __init__(self, request, values):
+        """
+        Initialize the event.
+
+        :param request: Pyramid request object
+        :param values: List of dicts of file information
+        """
+        self.request = request
+        self.values = values
+
+
+@subscriber(FilesDiscoveredEvent)
+def files_discovered(event):
+    """Record discovered files in the DB."""
+
+    application_instance = event.request.find_service(name="application_instance").get()
+
+    for value in event.values:
+        value["application_instance_id"] = application_instance.id
+
+    event.request.db.bulk.upsert(File, values=event.values)
--- a/lms/models/file.py
+++ b/lms/models/file.py
@@ -1,0 +1,49 @@
+import sqlalchemy as sa
+
+from lms.db import BASE, BulkAction
+from lms.models._mixins import CreatedUpdatedMixin
+
+
+class File(CreatedUpdatedMixin, BASE):
+    """A record of files we've seen in LMS's."""
+
+    __tablename__ = "file"
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "lms_id", "type", "course_id"),
+    )
+
+    # Enable bulk actions
+    BULK_CONFIG = BulkAction.Config(
+        upsert_index_elements=[
+            "application_instance_id",
+            "lms_id",
+            "type",
+            "course_id",
+        ],
+        upsert_update_elements=["name", "size"],
+    )
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+    """Primary key"""
+
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+        nullable=False,
+    )
+    """Link to the application instance."""
+
+    type = sa.Column(sa.String(), nullable=False)
+    """What type of file is this? e.g. 'canvas_file'."""
+
+    lms_id = sa.Column(sa.String(), nullable=False)
+    """The natural id of the file in the LMS."""
+
+    course_id = sa.Column(sa.String())
+    """If the file is associated with a specific course set it here."""
+
+    name = sa.Column(sa.String())
+    """The user facing file name."""
+
+    size = sa.Column(sa.Integer())
+    """The size of the file in bytes."""
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,0 +1,42 @@
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CanvasService:
+    """A high level Canvas service."""
+
+    api = None
+
+    def __init__(self, canvas_api):
+        self.api = canvas_api
+
+    def public_url_for_file(self, file_id, course_id, check_in_course=False):
+        """
+        Get a public URL for a Canvas file.
+
+        This will also attempt to check if the file is in the course to detect
+        course copy situations and fix it if we can.
+
+        :param file_id: The file to look up
+        :param course_id: The course the file should be in
+        :param check_in_course: Enable pre-emptive checking for the file in the
+            course regardless of whether the user has permission to see it
+        :return: A URL suitable for public presentation of the file
+        :raises  CanvasFileNotFoundInCourse: If we determine the file should
+            be in the course but isn't.
+        """
+
+        if check_in_course:
+            if not self._file_in_course(file_id, course_id):
+                # Looks like a course copy
+                raise CanvasFileNotFoundInCourse(file_id)
+
+        return self.api.public_url(file_id)
+
+    def _file_in_course(self, file_id, course_id):
+        return any(
+            str(file_["id"]) == str(file_id) for file_ in self.api.list_files(course_id)
+        )
+
+
+def factory(_context, request):
+    return CanvasService(canvas_api=request.find_service(name="canvas_api_client"))
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -3,6 +3,7 @@
 import marshmallow
 from marshmallow import EXCLUDE, Schema, fields, post_load, validate, validates_schema
 
+from lms.events import FilesDiscoveredEvent
 from lms.services import CanvasAPIError
 from lms.validation import RequestsResponseSchema
 
@@ -35,13 +36,15 @@ class CanvasAPIClient:
         Canvas API request fails for any other reason
     """
 
-    def __init__(self, authenticated_client):
+    def __init__(self, authenticated_client, request):
         """
         Create a new CanvasAPIClient.
 
         :param authenticated_client: An instance of AuthenticatedClient
+        :param request: For reporting events
         """
         self._client = authenticated_client
+        self._request = request
 
     def get_token(self, authorization_code):
         """
@@ -210,12 +213,31 @@ class CanvasAPIClient:
         # For documentation of this request see:
         # https://canvas.instructure.com/doc/api/files.html#method.files.api_index
 
-        return self._client.send(
+        files = self._client.send(
             "GET",
             f"courses/{course_id}/files",
             params={"content_types[]": "application/pdf"},
             schema=self._ListFilesSchema,
         )
+
+        # Notify that we've found some files
+        self._request.registry.notify(
+            FilesDiscoveredEvent(
+                request=self._request,
+                values=[
+                    {
+                        "type": "canvas_file",
+                        "course_id": course_id,
+                        "lms_id": file["id"],
+                        "name": file["display_name"],
+                        "size": file["size"],
+                    }
+                    for file in files
+                ],
+            )
+        )
+
+        return files
 
     class _ListFilesSchema(RequestsResponseSchema):
         """Schema for the list_files response."""
@@ -225,6 +247,7 @@ class CanvasAPIClient:
         display_name = fields.Str(required=True)
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
+        size = fields.Integer(required=True)
 
     def public_url(self, file_id):
         """
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -3,7 +3,7 @@
 import marshmallow
 from marshmallow import EXCLUDE, Schema, fields, post_load, validate, validates_schema
 
-from lms.services import CanvasAPIError, CanvasFileNotFoundInCourse
+from lms.services import CanvasAPIError
 from lms.validation import RequestsResponseSchema
 
 
@@ -225,17 +225,6 @@ class CanvasAPIClient:
         display_name = fields.Str(required=True)
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
-
-    def check_file_in_course(self, file_id, course_id):
-        """
-        Raise if the file with ID file_id isn't in the course with ID course_id.
-
-        :raise CanvasFileNotFoundInCourse: If the file isn't in the course
-        """
-        if not any(
-            str(file_["id"]) == str(file_id) for file_ in self.list_files(course_id)
-        ):
-            raise CanvasFileNotFoundInCourse(file_id)
 
     def public_url(self, file_id):
         """
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -2,6 +2,7 @@
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
+from lms.services.canvas import CanvasService
 from lms.views import helpers
 
 
@@ -9,7 +10,7 @@ from lms.views import helpers
 class FilesAPIViews:
     def __init__(self, request):
         self.request = request
-        self.canvas_api_client = request.find_service(name="canvas_api_client")
+        self.canvas = request.find_service(CanvasService)
 
     @view_config(request_method="GET", route_name="canvas_api.courses.files.list")
     def list_files(self):
@@ -19,7 +20,7 @@ class FilesAPIViews:
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
-        return self.canvas_api_client.list_files(self.request.matchdict["course_id"])
+        return self.canvas.api.list_files(self.request.matchdict["course_id"])
 
     @view_config(request_method="GET", route_name="canvas_api.files.via_url")
     def via_url(self):
@@ -29,14 +30,13 @@ class FilesAPIViews:
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
-        file_id = self.request.matchdict["file_id"]
-        course_id = self.request.matchdict["course_id"]
-
-        if self.request.lti_user.is_instructor:
-            self.canvas_api_client.check_file_in_course(file_id, course_id)
-
-        public_url = self.canvas_api_client.public_url(
-            self.request.matchdict["file_id"]
+        public_url = self.canvas.public_url_for_file(
+            file_id=self.request.matchdict["file_id"],
+            course_id=self.request.matchdict["course_id"],
+            # For teachers we can do a pre-emptive check to see if the file is
+            # in the course or not, to catch course copy scenarios even though
+            # the teacher has permission to see the file
+            check_in_course=self.request.lti_user.is_instructor,
         )
 
         # Currently we only let users pick PDF files, so we can save a little
--- a/lms/app.py
+++ b/lms/app.py
@@ -98,6 +98,7 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.include("lms.services")
     config.include("lms.validation")
     config.include("lms.tweens")
+    config.scan("lms.events")
     config.add_static_view(name="export", path="lms:static/export")
     config.add_static_view(name="static", path="lms:static")
 
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -6,10 +6,12 @@ import sqlalchemy
 import zope.sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session, sessionmaker
 from sqlalchemy.orm.properties import ColumnProperty
 
-__all__ = ("BASE", "init")
+from lms.db._bulk_action import BulkAction
+
+__all__ = ("BASE", "init", "BulkAction")
 
 
 LOG = logging.getLogger(__name__)
@@ -85,9 +87,6 @@ BASE = declarative_base(
 )
 
 
-SESSION = sessionmaker()
-
-
 def init(engine, drop=False):
     """
     Create all the database tables if they don't already exist.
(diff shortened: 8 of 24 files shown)


lms/models/file.py
```diff
@@ -0,0 +1,47 @@
...
+from lms.db import BASE
+from lms.models._mixins import BulkUpsertMixin, CreatedUpdatedMixin
+
+
+class File(CreatedUpdatedMixin, BulkUpsertMixin, BASE):
+    """A record of files we've seen in LMS's."""
+
+    __tablename__ = "file"
+    __table_args__ = (
+        sa.UniqueConstraint("application_instance_id", "lms_id", "type", "course_id"),
```
jon-betts, line 12: I think this index ordering will support the query patterns we are likely to have:

 * We'll always search by `application_instance_id`
 * We're mostly interested in single files, so the ID seems very likely
 * The type seems like the next most likely part to be in any query
 * Only the course part is probably likely to be dropped

As long as we use the columns from the start, we should be able to leverage this for our queries

lms/services/canvas_api/client.py
```diff
@@ -200,23 +201,45 @@ def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
...
 
-        return self._client.send(
+        files = self._client.send(
             "GET",
             f"courses/{course_id}/files",
             params={"content_types[]": "application/pdf"},
             schema=self._ListFilesSchema,
         )
 
+        if request:
```
jon-betts, line 223: Should this just be mandatory? Seems like this is a nice way to try and keep Pyramid from being a mandatory part of using this.

lms/views/api/canvas/files.py
```diff
@@ -19,7 +19,9 @@ def list_files(self):
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
-        return self.canvas_api_client.list_files(self.request.matchdict["course_id"])
+        return self.canvas_api_client.list_files(
+            self.request.matchdict["course_id"], self.request
+        )
```
jon-betts, line 24: The actual part where we trigger search and so storage

lms/services/canvas_api/client.py
marcospri: I don't have a better suggestion (a persist: bool parameter and the request seems worse) but seems odd that passing a `request` object implies persisting the object

lms/services/canvas_api/client.py
```diff
@@ -200,23 +201,45 @@ def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
...
+            # Notify that we've found some files
+            request.registry.notify(
+                FilesDiscoveredEvent(
+                    request=request,
+                    values=[
```
marcospri, line 228: Would it make sense that FilesDiscoveredEvent takes the full files objects instead of just some `values`? Then it's up to the event what to do with them

lms/services/canvas_api/client.py
seanh: Can't the service factory inject `request.registry.notify` into the service?

lms/services/canvas_api/client.py
jon-betts: It would ensure they have the right shape, but we'd need to marshall them back down to dicts again I think. So it's nice from the point of view of using the model code to enforce the schema, but it might add extra code.

I'll give it a shot and see if it works. Maybe SQLAlchemy will just accept it?

seanh (commented):
Looks good to me.

I think the mixin would be better as a service. If nothing else I think this is more harmonious with our usual approach. But I think it might also be a bit nicer as it keeps the models thinner and is more flexible and decoupled.

I think the event subscriber approach seems fine. To be honest I would probably have just had `CanvasAPIClient.list_files()` store the files in the DB directly. But :man_shrugging: 

@marcospri Can you handle proper testing and reviewing of this?

lms/events/subscriber.py
```diff
@@ -0,0 +1,29 @@
...
+def files_discovered(event):
+    """Record discovered files in the DB."""
+
+    application_instance_id = (
+        event.request.find_service(name="application_instance").get().id
+    )
+    for value in event.values:
+        value["application_instance_id"] = application_instance_id
+
+    File.bulk_upsert(session=event.request.db, values=event.values)
```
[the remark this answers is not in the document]
jon-betts: You wouldn't need to pass in the session, but you'd still need a request to find the service in the first place, so I feel like it's a bit of a wash. Long story short you need the request around either way, and I'm not sure that's much different? Which one is simpler?

```python
event.request.find_service(name="bulk_upsert").upsert(models.File, values=event.values).
# vs.
File.bulk_upsert(session=event.request.db, values=event.values)
```

### Are we concentrating on too few solutions?

If your dog is stuck in the banister and you're trying to workout if a land-mine or a tambourine is the correct solution, the answer might be neither.

I think I can see what you mean about injecting the DB session. It feels like a different kind of method from say `ApplicationInstance.update_lms_data()`, as that method modifies the model in the abstract. The model remains completely unconcerned about the existence of DBs.

This feels more like `db_session.add()` or something. You can't do it without a session, and it only accepts a model and some data. Is this something could tack onto the db session where ever we make it?

### I don't think you can really separate the bulk-upsertiness(tm) from the model

You can obviously move the code outside of the model as we do here with a mixin, or with a service but I don't think you can actually decouple the fact you need co-operation from the model itself when it comes to updating. You might be able to change what you update meaningfully (although I doubt you would) but I don't think you can vary the indexing requirements. It's pretty much the correct answer or broken.

So we could put the code that does the upserting separate from the variables in the model that are required to make it work, and it _appears_ like it _could_ work on anything, but in truth crashes on anything other than the specific models that cater for it. Similarly the models have some random config attached to them that _looks_ like it doesn't do anything but is actually _implicitly_ coupled with the service, rather than more explicitly coupled with the mixin.

So my feeling is a service in this case pretends it's decoupled, whilst actually just concealing the real coupling. Whereas a `BulkUpsertMixin` actively advertises that this model supports that action. Try calling `bulk_upsert` on a model that doesn't support it and the method isn't even there.

Maybe we could split the difference and have `BulkUpsertMixin` that only adds the config, and have something like this in the service?

`if not isinstance(BulkUpsertMixin)...`

### It would help me out if you could nail down some terms a bit for me

I think there might be some slop in my understanding of  how you're using a few of the terms and phrases here.

When you say harmonious with our general approach is this more of a "where we'd like to go" rather "where we are/what we've done" type of thing? Our approach as it appears in the code is clearly pretty varied, so I guess you mean "our future approach" or something instead? For example this is pretty much a slimmed down rip of the same thing from Checkmate: https://github.com/hypothesis/checkmate/blob/main/checkmate/models/db/mixins.py#L44.
Having methods on the models to create / update is also we do in `LtiLaunches.add()`, `GroupInfo.upsert_instructor()`. Lord knows that's up in `h`. 

Normally when people say put the logic in services, it's short hand for put the _**business logic**_ in services, rather than all logic. Seeing as the suggested `BulkUpsert` service wants to be model agnostic, it seems like it can't contain any business type logic. It seems a pure DB concern, which typically would be dealt with in the DB layer. Are you saying here we don't want any kind of logic in our DB layer at all regardless of the type of logic or is there a conflation here?

I wonder if you could also go into more detail about what you mean by thin models in this context. I feel like there's an implication I don't get in "having to inherit a lot of methods". It kind of reads to me like this is an effort for the model classes to load all these methods, or for us as programmers to add them all one by one, rather than something that just happens automatically through inheritance if you add things to either the BASE (as we do in `BaseClass.update_from_dict()`) or in the `BulkUpsertMixin`?

I think there's also a confusion for me about the model become "fat" by virtue of using a mixin. In this case is the "fatness" the fact that the the model has some functionality? Not the fact the model has a lot of actual code inside the class? As both the mixin and the service keep the actual implementation outside of the specific model.

lms/services/canvas_api/client.py
jon-betts: I've given this a shot, but the insert function expects plain dicts, so we have to convert them back again. I tried using a generic method, but as we don't set most of the columns it means weeding out `None`'s. This also means distinguishing between `None`'s that are there by default and `None`'s that we might be trying to set... all of which is to say... maybe just not do this?

lms/services/canvas_api/client.py
jon-betts: We can inject `request.registry.notify`, but we still need `request` around to put into the event. So I've injected the whole request instead.

lms/events/subscriber.py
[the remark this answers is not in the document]
jon-betts: I've had a crack at this and changed the base session object to be a class of our own from which we can add methods like this. I've kept the mixin, but it only adds the required fields. I think this achieves most of what we want:

* It keeps the code in the DB layer (which I think is distinct from the model layer?)
* It keeps the models thin, as long as you don't consider the DB layer to be the model layer
* It provides a way of adding DB functionality which can be shared everywhere (although models still have to have the right fields)
* The mixin still allows us to advertise which models do, and do not support this

What's been lost:

* The link is less obvious as the mixin class doesn't directly relate to `bulk_upsert`
* It does at least give us a location for documentation though, so we can have a pointer to the session object, which makes this discoverable at least
* You can obviously still call this on things that don't support it

lms/models/_mixins.py
```diff
@@ -13,3 +16,12 @@ class CreatedUpdatedMixin:
         onupdate=sa.func.now(),
         nullable=False,
     )
+
+
+class BulkUpsertMixin:
```
marcospri, line 21: Is this mixin here mostly as a documentation / cue of what's going on?

lms/services/canvas_api/client.py
```diff
@@ -205,18 +208,40 @@ def list_files(self, course_id):
...
 
-        return self._client.send(
+        files = self._client.send(
             "GET",
             f"courses/{course_id}/files",
             params={"content_types[]": "application/pdf"},
             schema=self._ListFilesSchema,
         )
 
+        if self._request:
```
marcospri, line 225: Should it take a `store` or similar now? `_request` will always be here.

lms/models/_mixins.py
jon-betts: Pretty much, there's much more detail in the longer comments above. The intention here is to:

 * Cause the parameters to be on the class (helps with auto complete)
 * Provide a positive declaration that this class supports being called with `bulk_upsert()`
 * Provide somewhere to attach documentation to give you a clue about what's going on

Otherwise these just end up being random parameters with no obvious connection to `bulk_upsert()`

lms/services/canvas_api/client.py
jon-betts: I think maybe `request` just shouldn't be optional.

lms/events/subscriber.py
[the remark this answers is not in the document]
seanh: What's the reason for adding methods to a `Session` subclass, rather than using a service? Do we need to invent a new thing here? Or would a service do the job just fine?

```python
class BulkUpsertService:
    def __init__(self, db):
        self._db = db

    def bulk_upsert(self, model_class, values):
        ...

    return BulkUpsertService(request.db)
```

And then:

```python
request.find_service(name="bulk_upsert").bulk_upsert(models.File, event.values)
```

I guess I'm not really seeing the downside to this? And therefore I'm not seeing why we'd introduce a new solution.

Services seem more scalable as you can keep adding more of them, whereas the `Session` class will get too big if you add too many methods to it, and you could end up having to find some way of breaking it up.

If you really don't want to use a service then I think the model mixin that you had originally might be a better approach than the DB session-based one. I think it's simpler to understand. Have a custom SQLAlchemy `Session` subclass seems a bit clever to me, and you still end up needing a mixin anyway.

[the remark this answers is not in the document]
seanh: There's an ADR about the services layer here: https://github.com/hypothesis/h/blob/master/docs/arch/adr-002.rst The Nando Farestan blog post that it links to is very good. 

I don't think we really make a distinction between "business logic" and "DB logic", we just put both in services. I don't really think of us as having a "DB layer" as such, technically there is a `db/` directory but it just contains some necessary SQLAlchemy boiler. Do we have much reason to distinguish between "DB logic" and "business logic"?

There are some slide about thin models in this presentation, starting on slide 50: https://docs.google.com/presentation/d/1O-rl1qr0diNNAwACLAe7StLeZEi4kuIvhfwrX83AMKM/edit#slide=id.g4c957d27dd_0_566 Basically the reasons why we move SQL queries up into the services layer is that multi-model queries don't make sense on any one particular model class; because model objects tend to get all over the whole codebase (and not mocked in tests) so we want them to be close to "value objects" so they don't end up tightly coupling a lot of code; services scale better because you can have many services for one model (whereas if all those methods were on the model class it'd be one too-big class).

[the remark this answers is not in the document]
seanh: > Maybe we could split the difference and have `BulkUpsertMixin` that only adds the config, and have something like this in the service?
> 
>     if not isinstance(BulkUpsertMixin)...

I think it's more Pythonic to avoid the `isinstance()` and just assume that the class has the necessary attrs on it (and go ahead and crash if it doesn't). Just use duck typing, the code will be simpler. You don't need the mixin.

lms/models/_mixins.py
seanh: Don't you end up needing to add the attributes to the class anyway?

```
class File(BulkUpsertMixin):
    BULK_UPSERT_INDEX_ELEMENTS = ...
    BULK_UPSERT_UPDATE_ELEMENTS = ...
```

Won't autocomplete still pick up those class attrs, even without the mixin inheritance?

I think the existence of the `BULK_UPSERT_*` attrs is itself a positive assertion that the class supports bulk upsert too (duck typing).

And the documentation would go on the `CustomSession.bulk_upsert()` method?

I think I preferred the original mixin-only approach over the current custom session + mixin approach just because I'm not a fan of having a custom session object. But it does look to me like the mixin class is unnecessary if using either the custom session or the service approach.

jon-betts: I think this is a good example of me attempting to address feedback and introduce compromise which results in worse outcomes. So don't think this is my preferred situation.

That said, I'll try and explain the difference. Adding a class attribute alone is a positive, but _obscure_ assertion. In the same way that referring to services with strings rather than class-names is a 1:1 correlation, but a much more obscure one. If you have a service called "user", then some _all_ of the places that ask for that service by name mean the service, and _all_ of the services implementing that name are the service, but only _some_ occurrences of "user" in the code base mean either. This means there is a connection, but you have to grep for it, and plenty of what you get back has nothing to do with it.

This is less likely to happen with a longer string like the one above, but the principal is the same.

I've noticed a couple of patterns in some of our discussions about related things:

 * A substitution of _discoverable_ for easily discovered
 * A reluctance to add bread crumbs to aid discovery
 
Whether something is easy to understand or obscure largely doesn't effect you once you understand it. I think a lot of our discussions often start from here rather than assuming you don't know how it works and are trying to find out. Lets assume you don't understand the mechanism and give an example of how you might discover it in each case. Each of the assumes you're starting from looking at the model and trying to work out what it can do to know that `bulk_upsert` is available, but you don't know it exists.

This describes one likely way to find this out, not all, or the only, or the most likely... (bum covering over)

### Solo mixin

* You look at the class
* You see the attributes and mixin
* The names relate, and you know mixins can add functionality _(leap)_
* You click through to the mixin and see the documentation and functionality
* You can use the functionality

### Mixin + functionality elsewhere (session / service)

* You look at the class
* You see the attributes and mixin
* The names relate, and you know mixins can add functionality _(leap)_
* You click through to the mixin and see the documentation about the functionality
* You follow the location
* You see the functionality
* You can use the functionality

### Attributes + functionality elsewhere

* You look at the class
* You see the attributes names and assume it means you can do something related to bulk upsetting _(leap)_
* You grep for the strings in the code base
* You ignore the results that don't relate to it and assume the one outside of the models implements related functionality _(leap)_
* You look at the code and see the documentation and functionality
* (In the case of string indexed services) You know services have a string to call them defined in a different file _(leap)_
* (In the case of string indexed services) Lookup the string name for the service
* You can use the functionality

All of these solutions work of course. It's not impossible, or perhaps even hard to work out any of them. But some of them require more mental leaps and knowledge to get to.

lms/events/subscriber.py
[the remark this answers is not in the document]
jon-betts: I've written a long thing below, but I'd suggest you don't read it. We're at the point where text tennis probably isn't going to give us a good understanding, or be a good use of our time. I'd like to discuss this in person, because I don't think I'm going to be able to understand it without being able to ask questions interactively.

----

I don't think we make any distinction between pure DB functionality and business logic type things, but I'm sort of suggesting here that that's a problem rather and potentially a case of not having gotten into it rather than the solution to the problem.

I think services for us are under specified and sort of a golden hammer type approach. Any code which doesn't fit into one of our predefined buckets ends up lumped into services for lack of any other likely locations to put it.

Seeing as services are basically any code, and any code can be written as a service (including a view if you're mad) then asking if something _can_ be done as a service seems to me to be arguing from the wrong direction. _Everything_ could be a service, but what _should_ be a service? What does making it a service add over it being normal code in another location? What's special about services that isn't true of a plain class? etc.

The reasons to make distinctions between different layers and responsibilities in your services and other parts of the app are largely the same ones as motivate having services in the first place again:

* To have different areas responsible for different things
* To make it easier to understand
* To make it convenient to call

I think it's beneficial for the services to have an identity and set of expectations for what's in there (or not) for the same reason it's a good idea to do that everywhere else.

### I think you said why services are good, but not why this doesn't apply to other things

You've given some reasons you don't want a bunch of specific SQL in all the model objects and I totally agree, but what I think is missing from that is how this is a counter argument to any of the other suggested solutions. Basically you state here that services do these things, and it's implied that other things don't, and I think I need to understand why to internalise this.

The benefits of SQL code in services and out of models you say:

### Multi-model queries don't make sense on any one particular model class 

The session solution fixes this directly. Another way to say this is also that multi-model queries make as much sense on any model as long as they aren't defined there, which the mixin would seem to fix.

### We want them to be close to "value objects" so they don't end up tightly coupling a lot of code

I think I need some explanation on this one. I think this might be a mocking concern? It might be more awkward to mock out a single method on the value objects. That definitely seems like a mark against the mixins. On the session object this seems like a direct reason to do it: we don't mock out basic DB access like adding, deleting and querying. We're just adding one more to that list. It might be a massive error and counter practice to mock out the `BulkInsertService` and the `BulkInsertService` alone because actually it's acting more like a DB functionality. 

We'd make that error because we don't differentiate between lower level DB functionality (which we should not mock) and higher level stuff (which we should).

### You can have many services for one model (whereas if all those methods were on the model class it'd be one too-big class).

The session solution directly addresses this.

With the models I asked above, but I don't think you addressed it. Does a model become "larger" in your mind when it exposes more functionality, even if that functionality is defined in one place and re-used? As in if the code in the file is exactly the same size on the disk, but there's different inheritance which results in more or less methods on the class, is that model "bigger" now?

I tend to define "too-big" as too big to look at when you open code in an editor on the whole. I get it's possible to add so many methods that you can end up with name spacing difficulties or auto complete becomes a giant mess, but I don't think we're talking about getting that far are we? Is this "conceptually" too big, vs. lines of code too big?

jon-betts:
Ok, I've got a new solution which I think checks all of the boxes and should make everyone happy. Given this information I imagine everyone will hate it.

The solution is to have an extension which is added to the session as a descriptor and accessed like this:

```python
class CustomSession(Session):
    bulk = BulkAction()

# ... later

session.bulk.upsert(Model, values)
```

This extension contains an inner `dataclass` which model classes are expected to add to themselves like this:

```python
class MyModel(BASE):
      # Enable bulk actions
      BULK_CONFIG = BulkAction.Config(
          upsert_index_elements=[...],
          upsert_update_elements=[...],
      )
```

Things I think this achieves:

 * The model is "thin" for all meanings of thin. It just attaches a dataclass with some values to itself
 * The configuration object has multiple advantages:
    * This is both a positive statement of functionality _and_ a direct, discoverable and unambiguous one
    * Directly leads you to the code which implements the functionality it provides
    * The configuration and the code that consumes it are in the same file
    * It enforces the names of the fields and should give IDE lint hints for the the contents too
    * It enforces the name of the attribute on the model so you can't mess that up
 * The session extension is completely uncoupled from the session meaning:
    * You can mock it or not as you see fit (usually not I'd imagine)
    * It only has one point of potential name-space conflict with the underlying class (the attribute)
    * The session can't grow too big, because this method starts with a way of breaking up the functionality

lms/db/_session_extension.py
```diff
@@ -0,0 +1,32 @@
...
+    def __get__(self, instance, owner):
+        if not instance:
+            raise TypeError(
+                f"You can only call '{self.name}' on instances not the bare "
+                f"class: '{owner.__name__}'"
+            )
+
+        # Store the session and return ourselves
+        self.session = instance
+        return self
```
jon-betts, line 28: This the only part of this whole thing that actually needs to happen. The rest is just belt and braces stuff of various kinds.

lms/db/_session_extension.py
```diff
@@ -0,0 +1,32 @@
...
+
+    The `name` and `session` attributes will be available to your instance.
+    """
+
+    session = None
+    name = None
+
+    def __set_name__(self, owner, name):
+        # Store the name for more informative errors
+        self.name = name
```
jon-betts, line 17: I doubt any children will need the name, but what the hoo. It means we can give a nice error message in a few lines if you try to access this as a class attribute.

lms/views/api/canvas/files.py
```diff
@@ -19,7 +20,7 @@ def list_files(self):
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
-        return self.canvas_api_client.list_files(self.request.matchdict["course_id"])
+        return self.canvas.api.list_files(self.request.matchdict["course_id"])
```
marcospri, line 23:
```suggestion
        return self.canvas.list_files(self.request.matchdict["course_id"])
```

lms/services/canvas.py
```diff
@@ -0,0 +1,43 @@
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CanvasService:
```
marcospri, line 4: Not sure if this is needed, I think it would would give me pause every time I add something that talks to canvas, is is CanvasService? Or CanvasAPIService?

I think the fact that the only public method returns a value from the api service could mean that the boundary is not that clear yet and this could go into the existing one. "CanvasService" is a better name than the current one thought.

lms/services/canvas.py
```diff
@@ -0,0 +1,43 @@
...
+    """A high level Canvas service."""
+
+    def __init__(self, db_session, canvas_api):
+        self._db_session = db_session
```
marcospri, line 8: Is the DB accessed from this service?

jon-betts: No not yet... This is actually spill over from future PRs. It's a good spot, but I'll leave it in as I'll just have to rewrite the tests to take it out and then put it back later.

lms/views/api/canvas/files.py
jon-betts: ? I'm not sure what's up here

jon-betts:
This has been superceded now

PR #2778, opened 2021-06-03
Add OAuth Authorizaton header support to HTTPService

seanh:
Multiple `BlackboardAPIClient` methods are going to have to do the same duplicate bit of code:

1. Call `OAuth2TokenService` to get the access token for the current user
2. Include the access token in an `Authorizaton: Bearer <ACCESS_TOKEN>` header when calling `HTTPService`

This PR adds that functionality to `HTTPService` so that each time `BlackboardAPIClient` calls `HTTPService` it can just do `http_service.get(..., oauth=True)`.

`HTTPService.request()` is still only 25 lines long with this new feature, and its tests are easy, so I don't think it's worth splitting it into separate services yet.

### DIFF
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -8,7 +8,9 @@ from lms.validation import ValidationError
 class HTTPService:
     """Send HTTP requests with `requests` and receive the responses."""
 
-    def __init__(self, _session=None):
+    def __init__(self, oauth2_token_service, _session=None):
+        self._oauth2_token_service = oauth2_token_service
+
         # A requests session is used so that cookies are persisted across
         # requests and urllib3 connection pooling is used (which means that
         # underlying TCP connections are re-used when making multiple requests
@@ -38,6 +40,7 @@ class HTTPService:
         url,
         timeout=(10, 10),
         schema=None,
+        oauth=False,
         **kwargs,
     ):  # pylint:disable=too-many-arguments
         """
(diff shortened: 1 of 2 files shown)


lms/services/http.py
```diff
@@ -65,6 +68,12 @@ def request(
         :param schema: A schema class to use to validate the response
         :type schema: lms.validation.RequestsResponseSchema
 
+        :param oauth: Include an OAuth 2 access token in the request.
+            If oauth=True the current user's access token will be looked up in
+            the database and included in an Authorization header in the
+            request.
+        :type oauth: bool
```
marcospri, line 75: I find this :type style doc string to be a bit clunky and much prefer the type annotation on the function definition but it's also true that adding for example just this one `oauth: bool = True` annotation would stand out as the only annotation in the fuction/module/maybe even project?

lms/services/http.py
```diff
@@ -93,9 +102,18 @@ def request(
...
             HTTPValidationError.response.
+
+        :raise NoOAuth2Token: If oauth=True was given but we don't have an
+            access token for the current user in our DB
         """
         response = None
 
+        if oauth:
+            kwargs.setdefault("headers", {})
+            assert "Authorization" not in kwargs["headers"]
```
marcospri, line 113: Some python programmers might be triggered by this assert and prefer something like 

```
if "Authorization" in kwargs["headers"]:
   raise ValueError("Can't use custom auth and Oauth simultaneously")
```
I have a preference for something like that, but only a small one.

marcospri (approved):

lms/services/http.py
seanh: Part of the problem is that I haven't gotten used to type annotations yet so I can't easily write them.

Although I did notice that the `:type:`'s keep the type documentation close to the `:param:` documentation, rather than the `:param:` being down in the docstring and the type annotation being way up at the top.

Realistically we should probably start using type annotations as they're clearly where the industry is going

PR #2779, opened 2021-06-03
Show activity indicator when file picker form is submitted

robertknight:
Since the assignment creation form is shown in an iframe, there is no indicator in the browser when the form is being submitted and the browser is waiting for the server's response. This is not a major problem because this typically happens quickly, but in case it is slow show a spinner to indicate that something is happening.

For this purpose I have extracted and re-used the spinner + backdrop that are used when the Canvas/Blackboard file picker dialog is loading. Here it is with an artificially delayed form submission:

https://user-images.githubusercontent.com/2458/120670456-56747080-c488-11eb-91e7-154e9a13eb1c.mov

### DIFF
--- a/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
@@ -117,6 +117,7 @@ describe('FilePickerApp', () => {
         },
         null /* groupSet */
       );
+      assert.isTrue(wrapper.exists('FullScreenSpinner'));
     });
   });
 
@@ -167,6 +168,7 @@ describe('FilePickerApp', () => {
           },
           useGroupSet ? 'groupSet1' : null
         );
+        assert.isTrue(wrapper.exists('FullScreenSpinner'));
       });
     });
   });
--- a/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
@@ -117,6 +117,14 @@ describe('FilePickerApp', () => {
         },
         null /* groupSet */
       );
+    });
+
+    it('shows activity indicator when form is submitted', () => {
+      const wrapper = renderFilePicker();
+      assert.isFalse(wrapper.exists('FullScreenSpinner'));
+
+      selectContent(wrapper, 'https://example.com');
+
       assert.isTrue(wrapper.exists('FullScreenSpinner'));
     });
   });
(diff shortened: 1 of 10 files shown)


lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -117,6 +117,7 @@ describe('FilePickerApp', () => {
         },
         null /* groupSet */
       );
+      assert.isTrue(wrapper.exists('FullScreenSpinner'));
```
LMS007, line 120: It might be useful to add 
`assert.isFalse(wrapper.exists('FullScreenSpinner'));` above selectContent in this test to show its hidden first, then renders. Same for the test below.

Alternately, we could just add new base test that ensures the spinner is not loaded if no action has been taken.

lms/static/scripts/frontend_apps/components/test/FullScreenSpinner-test.js
```diff
@@ -0,0 +1,12 @@
+import { mount } from 'enzyme';
+import { createElement } from 'preact';
+
+import FullScreenSpinner from '../FullScreenSpinner';
+
+describe('FullScreenSpinner', () => {
+  // There is no logic in this component, so this is just a basic "make sure it
```
LMS007, line 7: 👍

lms/static/styles/components/_FullScreenSpinner.scss
```diff
@@ -0,0 +1,18 @@
+@use "sass:math";
+
+@use "../mixins";
+@use "../variables" as var;
+
+.FullScreenSpinner__backdrop {
+  background-color: white;
+  bottom: 0;
```
LMS007, line 8: nit: bottom, left, right and top could be all together
something like this..
```
background-color: white;
opacity: 0.5;  
position: fixed;
left: 0;
top: 0
right: 0
bottom: 0;

```

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
robertknight: Thanks for the feedback.  I have extracted the checks for the `FullScreenSpinner` into separate tests and added checks that the spinner is not displayed when the form is initially rendered.

lms/static/styles/components/_FullScreenSpinner.scss
robertknight: Agreed. The properties were previously sorted alphabetically but it makes sense to group related properties.

Issue #2769, opened 2021-06-02
Store a skeleton module item configuration for Canvas assignments


jon-betts:
For: https://github.com/hypothesis/lms/issues/2764, and a stepping stone to: https://github.com/hypothesis/lms/issues/2527

Currently our Canvas implementation is a bit of an odd duck, as we don't use our "module item configuration" (assignment settings) for Canvas at all. We store all the info in a URL which is stored and controlled by Canvas.

To provide a place to save settings, and to make Canvas more like other implementations we should:

 * Store at least a skeleton record for Canvas assignments
 * If there are any Canvas exceptions to creating these records we should remove them, or alter the Canvas path so it's like the other paths to the extent it can be
 * We might also need to add a Canvas specific mechanism to add the record to allow us to "catch" up as we'll always have lots of assignments in the wild with no record
 
We don't need any particular data in the record other than the fact it is uniquely associated with an assignment in the usual way. Although in any case where it's easy of free to store data in the same way as other implementations we should

## What we're not doing

 * We do not need to read from this data in Canvas at all
 * We do not need to replace the mechanism that reads data from the URL provided by Canvas

PR #2780, opened 2021-06-03
Ensure that Canvas files launches have a module item configuration

jon-betts:
For: https://github.com/hypothesis/lms/issues/2769

We don't do anything with it, and the value we set isn't useful, but it ensures Canvas looks more like a normal launch for things down stream which might look up the details.

### DIFF
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -103,10 +103,25 @@ class BasicLTILaunchViews:
         Via. We have to re-do this file-ID-for-download-URL exchange on every
         single launch because Canvas's download URLs are temporary.
         """
-        self.context.js_config.add_canvas_file_id(
-            self.request.params["custom_canvas_course_id"],
-            self.request.params["file_id"],
+
+        course_id = self.request.params["custom_canvas_course_id"]
+        file_id = self.request.params["file_id"]
+
+        # Normally this would be done during `configure_module_item()` but
+        # Canvas skips that step. We are doing this to ensure that there is a
+        # module item configuration. As a result of this we can rely on this
+        # being around in future code.
+        self.assignment_service.set_document_url(
+            self.request.params["tool_consumer_instance_guid"],
+            self.request.params["resource_link_id"],
+            # This URL is mostly for show. We just want to ensure that a module
+            # configuration exists. If we're going to do that we might as well
+            # make sure this URL is meaningful.
+            document_url=f"canvas_file://course/{course_id}/file_id/{file_id}",
         )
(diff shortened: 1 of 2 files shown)


marcospri:
I did a similar PR a few weeks ago: https://github.com/hypothesis/lms/pull/2678/files only storing the via url instead of the made up schema.

If I remember correctly the main complain was storing this on launch, not on creation but I think this approach helps filling up the DB with data that might be useful down the line.

jon-betts:
Unfortunately for us the creation process in Canvas doesn't come through us I think. We provide them some data, which might, or might not result in the creation of an assignment but they don't call us to make that happen.

This is different from the regular run of things where I think the first launch is the unconfigured one and we go through our normal process. The first time we are aware of an assignment in Canvas is on launch of it for real.

Could be wrong?

lms/views/basic_lti_launch.py
```diff
@@ -103,10 +103,25 @@ def canvas_file_basic_lti_launch(self):
...
+        # Canvas skips that step. We are doing this to ensure that there is a
+        # module item configuration. As a result of this we can rely on this
+        # being around in future code.
+        self.assignment_service.set_document_url(
+            self.request.params["tool_consumer_instance_guid"],
+            self.request.params["resource_link_id"],
+            # This URL is mostly for show. We just want to ensure that a module
+            # configuration exists. If we're going to do that we might as well
+            # make sure this URL is meaningful.
+            document_url=f"canvas_file://course/{course_id}/file_id/{file_id}",
```
seanh, line 120: For Blackboard files we went with `blackboard://content-resource/<ID>`.

The `content-resource` is because Blackboard doesn't really have a "files" feature as such, it has multiple ways to upload files that would need to be handled differently (e.g. they have different APIs for accessing them), and the one we're using is called "content resources". For Canvas there is only one file-uploading feature (as far as I know) and it's called "files".

So I think these ones should start with `canvas://file/` to be consistent with `blackboard://content-resource/`.

So I think that could suggest: `canvas://file/course/{course_id}/file/{file_id}`. Repeating `file` twice seems a bit awkward but the two `file`'s do play different roles.

Alternatively it's actually not too late to change the Blackboard ones to `blackboard_content_resource://...`

seanh (approved):
Looks good. We might want to fiddle with the exact URL format if we want there to be a consistent system to these

lms/views/basic_lti_launch.py
jon-betts: No, I think that's better. I'll swap it over

PR #2781, opened 2021-06-03
Get real blackboard file lists

seanh:
Depends on https://github.com/hypothesis/lms/pull/2778

Get real lists of files from the Blackboard API, instead of returning a fake list.


lms/services/blackboard_api.py
```diff
@@ -45,6 +46,32 @@ def get_token(self, authorization_code):
...
+        )
+
+        return [
+            {
+                "display_name": file_["name"],
+                "id": file_["id"],
+                "updated_at": file_["modified"],
+            }
+            for file_ in response.validated_data
+        ]
```
seanh, line 69: It should be possible for the `list_files()` method to be a one-liner:

```python
return self._http_service.get(
    self._api_url(f"courses/uuid:{course_id}/resources"),
    oauth=True,
    schema=BlackboardListFilesSchema,
)
```

See the comment on the schema below.

lms/validation/blackboard.py
```diff
@@ -0,0 +1,26 @@
...
+
+        id = fields.Str(required=True)
+        name = fields.Str(required=True)
+        modified = fields.Str(required=True)
+
+    results = fields.List(fields.Nested(ResultsSchema), required=True)
+
+    @post_load
+    def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
+        return data["results"]
```
seanh, line 26: This schema is doing a bunch of things for us:

1. It extracts the list of files out of the top-level `"results"` key that Blackboard sends. (The response from Blackboard is `{"results": [ <list_of_file_dicts> ]}`.)
2. It raises an exception with a correctly-formatted error message if the top-level dict has no `"results"` key or of the `"results"` key's value is not a list
3. It raises if any of the file dicts is missing any of the required fields
4. It raises if any of the file dicts has a field with an invalid value

But it should actually be doing a couple more things as well, which would simplify the service:

1. The service has to translate Blackboard API field names to the field names that our frontend expects from our proxy API: `"name"` to `"display_name"` and `"modified"` to `"updated_at"`. The schema could do this so the service doesn't have to
2. The service has to exclude any unknown fields from the results. The schema could do that
3. I'm not sure whether Blackboard's modified times are in the correct format as expected by our frontend. What format does our Canvas list files API return? The schema could translate the format. But I suspect that our frontend merely sorts them alphanumerically and doesn't actually parse the datetimes so it may be fine

tests/unit/lms/services/blackboard_api_test.py
```diff
@@ -57,6 +60,71 @@ def test_if_theres_no_refresh_token_or_expires_in(
...
+                "creatorId": "_26_3",
+                "downloadUrl": "https://blackboard.example.com/downloadUrl3",
+                "id": "_7851_3",
+                "mimeType": "application/pdf",
+                "modified": "2021-04-02T15:39:51.000Z",
+                "name": "File_3.pdf",
+                "parentId": "_5801_1",
+                "size": 13472644,
+                "type": "File",
+            },
```
seanh, line 100: This is what the actual files from the Blackboard API look like. Might be worth adding a local factory to generate these, to reduce test noise

tests/unit/lms/validation/blackboard_test.py
```diff
@@ -0,0 +1,122 @@
...
+                        "name": "File_2.pdf",
+                        "modified": "2021-05-02T15:39:51.000Z",
+                    },
+                    {
+                        "id": "_7851_3",
+                        "name": "File_3.pdf",
+                        "modified": "2021-04-02T15:39:51.000Z",
+                    },
+                ],
+            )
```
seanh, line 53: Could use a test for the valid case when the list of files is empty.

Once the schema is extended to actually mutate the data (excluding unknown fields, changing the names of some fields, ...) this test will have to be improved a bit

tests/unit/lms/services/blackboard_api_test.py
jon-betts: Might be worth slimming this down to the relevant fields + `"noise": "noise` or something. Otherwise it's not clear at a quick look what matters and what doesn't.

seanh: Probably

lms/views/api/blackboard/files.py
```diff
@@ -10,38 +10,14 @@
...
-            },
-            {
-                "id": "blackboard://content-resource/789",
-                "display_name": "Fake Blackboard File 3",
-                "updated_at": "2020-08-03T14:06:00Z",
-            },
-        ]
+        return self.blackboard_api_client.list_files(
+            self.request.matchdict["course_id"]
+        )
```
jon-betts, line 20: The view equivalent of a mic drop :microphone: :boom:

lms/validation/blackboard.py
```diff
@@ -0,0 +1,26 @@
...
+
+from lms.validation._base import RequestsResponseSchema
+
+
+class BlackboardListFilesSchema(RequestsResponseSchema):
+    """Schema for Blackboard API /courses/{courseId}/resources responses."""
+
+    class ResultsSchema(Schema):
+        class Meta:
+            unknown = INCLUDE
```
jon-betts, line 16: The excluding unexpected things is just a case of using `EXCLUDE` here isn't it?

tests/unit/lms/validation/blackboard_test.py
```diff
@@ -0,0 +1,122 @@
...
+                            "id": "_7851_2",
+                            "name": "File_2.pdf",
+                            "modified": "2021-05-02T15:39:51.000Z",
+                        },
+                        {
+                            "id": "_7851_3",
+                            "name": "File_3.pdf",
+                            "modified": "2021-04-02T15:39:51.000Z",
+                        },
+                    ],
```
jon-betts, line 32: Is three files more special than two? I think two proves it's a plural.

tests/unit/lms/validation/blackboard_test.py
jon-betts: I assume you'll probably end up with some extra versions of the test (like empty lists), but if you don't the mark can probably go.

tests/unit/lms/validation/blackboard_test.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    def test_invalid(self, json_data, messages):
+        schema = BlackboardListFilesSchema(
+            factories.requests.Response(json_data=json_data)
+        )
+
+        with pytest.raises(ValidationError) as exc_info:
+            schema.parse()
+
+        assert exc_info.value.messages == messages
```
jon-betts, line 122: ### Maybe don't test the schema messages specifically?

I don't think we actually care _why_ a particular piece of data is invalid (we just raise a blanket exception). So perhaps we should reduce the scope of these tests to include the bits we need (it raises) and exclude the parts our code doesn't actually rely on (it raises this _exact_ set of messages). This would remove the `messages` side of this and really cut this test down.

This is one potential advantage of testing the schema and the end-point together as we don't end up testing the private interface between the two. For example, if the parent function is the only consumer of the schema, and we consume all these error messages, then there's actually no point in testing they are present at all (vary this example to meet the facts etc.).

### Maybe don't check as much about the schema at all

I feel like we've had this chat before, but to what extent do we want to be testing `marshmallow` schema? It seems like we run to the point of testing the behavior of `marshmallow` pretty quick.

I'd be _strongly_ tempted to prove:

 * The schema accepts a full, validly specified document
 * The schema rejects at least a couple of obvious invalid documents

And leave it at that. Few data formats meet a schema, and infinite number don't. We've chosen a small number of the infinite that don't match. Why not check a field rejects a list as well as None, what about a list with None in it, what a list with a list with None in it... 

If every API end-point has a separate schema, and every schema has 100+ lines of test code we're signing ourselves up for thousands of lines of testing code for even a moderately sized API. All of that comes with a maintenance cost, and if it's lower value (proving that marshmallow `Str` does what you'd expect) I'm not sure there's much benefit.

The 122 lines above test 9 lines of schema code. I'd say that's not a good trade off.

tests/unit/lms/validation/blackboard_test.py
```diff
@@ -0,0 +1,122 @@
...
+            (
+                {
+                    "results": [
+                        {
+                            "id": 23,
+                            "name": None,
+                            "modified": False,
+                        },
+                    ]
+                },
```
jon-betts, line 101: It might make a more unit-y test to separate the tests into:

 * Those where the parent structure is wrong
 * Those where elements of a specific file are wrong
 
That would allow you to have a "good" file and then modify each field individually with something like this:

```python
@pytest.mark.parametrize("field", ("id", "modified", "name"))
@pytest.mark.parametrize("bad_value", (MISSING, None, 1234))
def test_invalid_field(self, json_data, messages):        
    bad_file = {"id": "string",  "modified": "string",  "name": "string"}
    if value is MISSING:
        bad_file.pop(field)
    else:
        bad_file[field] = value

    schema = BlackboardListFilesSchema(
        factories.requests.Response(json_data={"results": [bad_file]})
    )        
    with pytest.raises(ValidationError) as exc_info: 
        schema.parse()
```

jon-betts (commented):
This is just an overview of the WIP rather than a proper review:

 * I think the split where you do the mapping in the function as you have it is good, and I'd actually go more that way. Have the schema represent the API, and not do any marshalling
 * To that end I'd suggest removing `make_object()` and just dereferencing in the code (which just results in `response.validated_data["results"] vs. `response.validated_data` in the caller)
 * I don't think we should test the schema as much (possibly at all separately)
 * We're testing a lot of interface between two pieces of code that's functionally private there (imagine the schema was applied in line with our own code, would any testing be removed?) 
 * I'd urge against testing too many counter examples to a schema because it's effectively sampling an infinite space (test all words that aren't "dog")

I think some of this is as a result of getting boxed in by some testing decisions. If it's not Ok to test the parts together we end up policing a private boundary between the two. This increases the total size of the tests by duplicating tests which would cover both in one go, and by tempting us into making assertions about features of that private boundary that don't matter outside the function.

If we don't want to change our minds about testing these things together (because they go through a third party) then I think we're left with changing the code to match our decisions about testing instead. Which means applying the schema as part of the call.

I think it's really worth stopping for a bit and thinking about whether this design can be simpler.

lms/validation/blackboard.py
```diff
@@ -0,0 +1,26 @@
+"""
```
jon-betts, line 1: I don't think this file is in the right place. `lms/validation` is for schema applied before views. This should be next to (or ideally in) the code it's used in.

lms/validation/blackboard.py
seanh: Yep. Which is a lot more convenient than letting the schema include them and then the service having to iterate over the list of dicts from the schema in order to remove the fields that should have been excluded by the schema

lms/services/blackboard_api.py
seanh: I decided to go ahead and move the mutation of the data dict (excluding unknown fields and renaming some of the fields) into `BlackboardListFilesSchema`, for a few reasons:

1. It's what Marshmallow schemas are intended for.

   It's not really accurate to think of these as "validation schemas", what they actually are is serialization/deserialization/validation schemas (the name `lms/validation/` may be unfortunate for this reason). If you read the beginning of the Marshmallow docs they describe it as a <q>library for converting complex datatypes, such as objects, to and from native Python datatypes</q> and they actually cover serialization and deserialization first before getting on to validation.

   It might make more sense if you imagine having a `models.File` object and the schema returning a list of those, rather than using dicts directly as the models. Imagine having multiple schemas (`BlackboardFilesSchema`, `CanvasFilesSchema`, `MoodleFilesSchema`) that parse different responses and all return lists of `models.File` objects.

   Validation is a necessity of deserialization because if the input data is invalid then the schema can't create a valid model object from it.

   We already use Marshmallow schemas in this way in the LMS app, e.g. we have schemas for serializing and deserializing bearer tokens and OAuth 2 state params etc.

   I don't think this is unique to Marshmallow: it's typical for schema libraries to both deserialize and validate at once.

2. It makes the code much shorter and simpler.

   You don't really have to add any code to `BlackboardListFilesSchema` to get it to mutate the file dicts in the ways that we want. Excluding unknown fields is just replacing `unknown = INCLUDE` with `EXCLUDE`. Renaming a field is just using `data_key`: `display_name = fields.Str(data_key="name", ...)`.

   Whereas if `BlackboardListFilesSchema` doesn't do the work then `BlackboardAPIClient` has to call `BlackboardListFilesSchema` and then loop over the list of dicts excluding and renaming fields, which is more code.

3. Sort of the same thing as (2) but it makes the tests much simpler. Instead of having `BlackboardListFilesSchema` tests that test that it returns the dicts from Blackboard unmodified and then `BlackboardAPIClient` tests that feed those unmodified dicts into `BlackboardAPIClient` and assert that it modifies them, instead you just have to test that `BlackboardListFilesSchema` returns the modified dicts. The tests for `BlackboardAPIClient` become just `assert returned == HTTPService.get.return_value`

4. It keeps everything about Blackboard Files API responses in one place.

   `BlackboardListFilesSchema` is the one place that has to deal with Blackboard Files API responses. The `BlackboardListFilesSchema` tests are the only tests where fake Blackboard Files responses are needed. If the responsibility is instead spread between `BlackboardListFilesSchema` and `BlackboardAPIClient` then the tests for both need to know about the Blackboard API response format.

5. It makes `BlackboardAPIClient.list_files()` a one-liner

lms/validation/blackboard.py
seanh: Added a [reply](https://github.com/hypothesis/lms/pull/2781#discussion_r648193154) to the thread above, but: I don't think schemas should just represent the expected data without any other functionality. They validate and _deserialize_ input objects

tests/unit/lms/services/blackboard_api_test.py
seanh: Done, I trimmed these down to just the fields that we care about plus one noise field

lms/validation/blackboard.py
```diff
@@ -0,0 +1,26 @@
...
+Schemas for Blackboard API responses.
+
+See: https://developer.blackboard.com/portal/displayApi
+"""
+from marshmallow import INCLUDE, Schema, fields, post_load
+
+from lms.validation._base import RequestsResponseSchema
+
+
+class BlackboardListFilesSchema(RequestsResponseSchema):
```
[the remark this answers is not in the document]
seanh: I don't think I agree with the approach that the Canvas API client took here.

Our usual architecture style is to separate different types of things into different files (`models/`, `services/`, `validation/`, `views/`, ...). One of the issues I have with Canvas API client is that it's bucking the convention here and using its own architectural style, and that kind of thing does hurt maintenance, productivity and general developer happiness (why is this one lump of code different from everything else?). But I also don't think I'd want to adopt Canvas API client's style consistently across the rest of the codebase:

The `BlackboardAPIClient` class would get too long if it had to contain all the methods (`list_files()`, `public_url()`, ...) and also all the schemas (`BlackboardListFilesSchema`, `BlackboardPublicURLSchema`, ...). I don't find this particularly readable in `CanvasAPIClient` personally. I think it obscures what `CanvasAPIClient` is as when I'm reading through that file I'm seeing lots of methods (`authenticated_users_sections()`, `course_sections()`, ...) which're what I actually care about and then there are also lots of inlined schema classes that kind of get in the way and confuse things visually. The mixture of two different _types_ of things (business logic methods and validation schemas) woven together is also quite visually confusing.

I think the `BlackboardAPIClient` as I have it currently is clearer and easier to read because when you look at the `BlackboardAPIClient` class you can more easily see its public interface (the `get_token()`, `list_files()` and `public_url()` methods) because there's not much else in there (so far just one helper method tucked away at the bottom of the class). Because the code only contains one type of thing (service methods) it's more visually regular and easier to read.

I don't particularly find the abstraction of "`BlackboardAPIClient.list_files()` just returns whatever `BlackboardListFilesSchema` returns" to be particularly problematic. The code is clear enough to me and it's easy enough to look up the schema if I want to see exactly what is returned. But the exact return format is not really something that you'd be wanting to know while working on the `BlackboardAPIClient` code because `BlackboardAPIClient` isn't responsible for the return format.

I think keeping the schemas separate is also more harmonious with the rest of our schemas in a couple of ways:

1. Request schemas can't be inlined into the views or unit tested as one with the views. They go into the `@view_config(..., schema=FooSchema)` and when the view unit tests call the view directly that does not execute the schema. The schemas have to be unit tested separately from the views that use them.

2. Request and response schemas that're used in multiple places can't be inlined into one of the places that they're used.

   For example `OAuthCallbackSchema` (a request schema) and `OAuthTokenResponseSchema` (a response schema) are used by both the Blackboard and Canvas API clients plus any future code for other OAuth APIs. These have to be defined and tested in one shared location (e.g. we're not going to duplicate all the tests).

3. Beyond just schemas I think we generally like to break things up into separate components that go into separate files and get unit tested separately.

Of course many schemas (including `BlackboardListFilesSchema`) aren't used in more than one place. But they might turn out to be in future

tests/unit/lms/validation/blackboard_test.py
seanh: ## Maybe don't test the schema messages specifically?

Done.

The reason for asserting about the schema messages was to make sure that the tests were actually testing what they thought they were testing. If the test merely asserts that `ValidationError` was raised it could be that the test set up is actually wrong and `ValidationError` is actually being raised for some other reason.

_But_ I implemented a suggestion that you made in another comment below that the tests should contain a single valid API response and then the tests for _invalid_ responses should work by taking the valid response and mutating it:

```python
# This is example code, it's not the actual code.

def test_it_raises_if_the_id_isnt_a_string(self, json_data):
    json_data["results"][0]["id"] = 23

    schema = BlackboardListFilesSchema(Response(json_data=json_data))

    with pytest.raises(ValidationError):
        schema.parse()
```

The tests know that the `json_data` fixture is a valid JSON data: there is another test above that passes `json_data` to the schema unmodified and asserts that it raises. And since this test only modified one thing about `json_data`, if `ValidationError` is raised it must have been for that one reason :+1:

seanh: ## Testing the schema and the endpoint together

I think it's better to treat schemas as separately unit-tested components. Some of the reasons are covered in other comments (e.g. consistency with request schemas and with response schemas that're used in multiple places). More generally I think it's better to break things up into small, separate components with small, separate unit tests. In the case of a validation schema unit tests that import the schema and call it directly are "closer to the metal" and as a result are much easier to understand and maintain. There's just much less going on under each test. I can look at the schema file (simple) and its unit tests (simple) and there isn't any other code involved in those unit tests, they just import the schema and call it.

Sorry to be frank here but I find Canvas API client's tests quite hard to understand and maintain. Some quite well-shrouded test bugs have been found in them and I suspect they contain more. And I suspect that if that testing style were used more widely in our codebase we'd get more of those well-camouflaged test bugs.

seanh: ## Don't just test Marshmallow

I'm sympathetic to the idea that we shouldn't write a dozen unit tests for every single schema that're really just "testing Marshmallow" (or at least testing things about the schema that're quite trivial to implement with Marshmallow and unlikely to contain any mistakes). @marcospri has also raised this point and I think he gave a good example: we don't test all the behaviours of SQLAlchemy in the unit tests for every one of our model classes.

The counter-argument is that will some schema tests might _seem_ pointless they actually do have a point: they're protecting us against changes in Marshmallow. A new Marshmallow version is released, Dependabot sends us a PR, CI passes, we assume "Must be fine then" and merge it. This is not a safe assumption if we haven't got tests for our Marshmallow schemas.

I think the solution is probably a middle-ground: do write tests for Marshmallow schemas, but just don't go crazy and write dozens of them. Which of course is what you were suggesting in your comment too.

Where we seem to differ is that I think my tests already achieve this balance. `TestBlackboardListFilesSchema` has:

* One test for a valid response
* Four tests for different kinds of invalid responses:
  1. No `"results"` list
  2. `"results"` list is not a list
  3. One of the files dicts is missing all the required fields
  4. One of the files dicts has invalid values in all the fields
 
Five simple tests total. I don't think that's excessive at all.

The tests are about 70 lines of code. About 10 lines of that is a fixture that looks to me like it might be quite generic and possibly could be shared among all schema tests (or at least all response schema tests). Another 20 lines or so is the fake Blackboard API response dict rather than actually Python code. So it's really more like 40ish lines of actual test code.

tests/unit/lms/validation/blackboard_test.py
seanh: Done.

I didn't end up doing _exactly_ what's suggested here but something very similar. I split up the single `test_invalid()` test (which the big `parametrize()`) into separate tests and I implemented the thing where the tests contain just a single valid API response dict and mutate that dict to make it invalid in different ways.

Here's an example of what one of the tests looks like now:

```python
@pytest.mark.parametrize("missing_field", ["id", "modified", "name"])
def test_it_raises_if_a_required_field_is_missing(self, json_data, assert_raises, missing_field):
    del json_data["results"][0][missing_field]

    assert_raises()
```

I think the change to using a valid `json_data` fixture and mutating it also makes it unnecessary to assert about the actual error messages (see other comment above about that)

lms/validation/blackboard.py
seanh: Done.

I don't agree that it's ideal for the Blackboard API client's schemas to be _in_ the `BlackboardAPIClient` class itself or in the same file for reasons I discussed above: I think this makes the Canvas API client confusing and hard to read because the class/file is too long and because it contains a long mixture of two different types of thing. It won't scale to keep adding methods and schemas to one class/file, it'll get too long. I also think the integrated tests are problematic (also discussed in the other comments). I think schemas are better treated as separate units.

So I ended up moving the service to `services/blackboard_api/service.py` and the schemas to `services/blackboard_api/_schemas.py` where it's clear that they're private to the `services/blackboard_api/` directory.

I think there might be a best-of-both-worlds here where we put the `list_files()` method and `ListFilesSchema` in one file, the `public_url()` method and `PublicURLSchema` in another file, and so no. Allowing each method + schema to be next to each other in the same file but avoiding a really long file that's hard to read. But I'm not sure exactly how that would look or whether it would actually turn out to be an improvement in the end. I think the existing layout is serviceable enough so for practicality (we are on a tight deadline here) I'm gonna vote for not letting the perfect be the enemy of good and going with this for now.

seanh:
I think the code design here is probably simple enough that we can push forward. There are three components to it:

1. `BlackboardListFilesSchema` is a Marshmallow schema that validates and deserializes Blackboard Files API responses, returning a list of dicts in the format that our frontend code expects from our own list files proxy API. This is the component that actually does most of the work of the list files API:

   ```python
   class BlackboardListFilesSchema(RequestsResponseSchema):
       class ResultsSchema(Schema):
           class Meta:
               unknown = EXCLUDE

           display_name = fields.Str(data_key="name", required=True)
           updated_at = fields.Str(data_key="modified", required=True)

       results = fields.List(fields.Nested(ResultsSchema), required=True)

       @post_load
       def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
           return data["results"]
   ```

2. `BlackboardAPIClient.list_files()` is a service method that uses `HTTPService` to call the Blackboard list files API and uses `BlackboardListFilesSchema` to validate and deserialize the response. It's a one-liner:

   ```python
   def list_files(self, course_id):
       return self._http_service.get(
           self._api_url(f"courses/uuid:{course_id}/resources"),
           oauth=True,
           schema=BlackboardListFilesSchema,
       ).validated_data
   ```

3. `BlackboardFilesAPIViews.list_files()` is a one-liner view that just calls the `BlackboardAPIClient.list_files()` service method. The view only exists because something has to hook the service method up to a the route and to the `course_id` in the URL path:

   ```python
   def list_files(self):
       return self.blackboard_api_client.list_files(self.request.matchdict["course_id"])
   ```

One schema and two one-liner methods, so I think it should be simple enough at least for now.

lms/services/blackboard_api/_schemas.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+from marshmallow import EXCLUDE, Schema, fields, post_load
+
+from lms.validation._base import RequestsResponseSchema
+
+
+class BlackboardListFilesSchema(RequestsResponseSchema):
+    """Schema for Blackboard API /courses/{courseId}/resources responses."""
+
+    class ResultsSchema(Schema):
```
marcospri, line 14: I know this is inside `BlackboardListFilesSchema` so it's kinda namespaced already but I think `FileSchema` might be a better name .

tests/unit/lms/services/blackboard_api/_schemas_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+        assert_raises()
+
+    def test_it_raises_if_results_isnt_a_list(self, json_data, assert_raises):
+        json_data["results"] = None
+
+        assert_raises()
+
+    @pytest.mark.parametrize("missing_field", ["id", "modified", "name"])
+    def test_it_raises_if_a_required_field_is_missing(
```
marcospri, line 40: I wonder was the right pytest pattern to generalize this when we have X schemas to test. A base TestSchema class that implements a few of these?

seanh: Dunno, but I suspect there's something like that possible. I can certainly see the value in figuring out a good way to unit test schemas and adding some kind of test helpers for it, as we can then re-use that for all our schema tests. But I dunno exactly what it'd be yet. Sharing test fixtures between tests in `tests/unit/lms/validation/` and tests in `tests/unit/lms/services/blackboard_api/` would unfortunately be made harder by the location of these schemas in the `services/blackboard_api/` dir instead of in the `validation/` dir. Any shared fixtures would have to go in the very top-level `tests/unit/lms/conftest.py` file. I think that might turn out to be a downside of the "localized" layout here

lms/services/blackboard_api/_schemas.py
seanh: Done. Added a small docstring too

marcospri (commented):
Looks good

As for the location/testing/usage of the schemas

- I'm not a fan of in-lining them in the service as it makes reading & navigating the code there more difficult IMO. So if the only change was moving all of them to _schemas.py I'd be happy.

- Testing the schema, if we want to unit test them independently we'll need to figure out a quick non boilerplaty test convention. Tests on this PR are a good starting point.

- I'd stay away of doing to much work in the schema's `make_object`, lifting the `results` seems a reasonable thing to do but I think if the data needs lots of extra processing that should happen on the service.

lms/services/blackboard_api/service.py
```diff
@@ -45,6 +46,23 @@ def get_token(self, authorization_code):
...
+        Return the list of files in the given course.
+
+        :raise ProxyAPIAccessTokenError: if we don't have a Blackboard API
+            access token for the current user
+        """
+        return self._http_service.get(
+            self._api_url(f"courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardListFilesSchema,
+        ).validated_data
```
marcospri, line 60: Would it make sense that `list_files` returned `List[FileSchema]` I'd have thought that would be straightforward thing to do but can figure it out on the docs.

marcospri (approved):

seanh:
> I'm not a fan of in-lining them in the service as it makes reading & navigating the code there more difficult IMO. So if the only
> change was moving all of them to _schemas.py I'd be happy.

Ok. That's how I have it currently. Though I was wondering to myself whether I should rename that to `_validation.py` to match the naming of `lms/validation/`?

> Testing the schema, if we want to unit test them independently we'll need to figure out a quick non boilerplaty test convention. 
> Tests on this PR are a good starting point.

Yeah, I agree with this. I think that's one of the advantages of the regularity here (the app has lots of schemas which are pretty much different takes on the same thing--every schema has the same public interface). We can figure out a good way to test them and then apply it repeatedly.

> I'd stay away of doing to much work in the schema's `make_object`, lifting the results seems a reasonable thing to do but I 
> think if the data needs lots of extra processing that should happen on the service.

I think staying away from _too much_ work makes sense, yes. `make_object()` exists to be used, but it seems fairly clear to me that `make_object()` isn't intended to be and shouldn't be a big complicated method that's doing all kinds of work. Maybe "deserializing" is the guide here. If the data is, for example, a list of user objects in serialized form (e.g. as a list of JSON dicts) then I think it's good for a schema to deserialize those and return a list of `models.User` objects (_if_ that's what you want: just returning dicts is also fine). That seems appropriate to me and I think it's going to help to keep code simpler as it did in this PR. It keeps the knowledge about the JSON format for user dicts in one place in the schema (if we need to actually _serialize_ user objects to dicts as well the same schema can handle both the serialization and the deserialization). This is clearly how Marshmallow is intended to be used which I think is a sign that we're on the right track. And popping the `"results"` also seems like a reasonable part of "deserialization": the top-level `"results"` key is just part of the serialized format.

But as you say you wouldn't want a schema to do any more work beyond what counts as "deserialization". . At some point you've stumbled into "business logic" or "view logic".

We'll see how this pans out as we add more functionality. If it turns out not to work well we'll adjust the design

PR #2783, opened 2021-06-04
Perform Canvas file matching to fix Canvas course copies

jon-betts:
This expands the `CanvasService` to use the data we _may_ have stored about files in the event we can't look them up for one reason or another.

The `public_url` will now attempt to match files if:

 * You request pre-emptive checking with `check_in_course` and the file isn't in the course
 * You get a  `CanvasAPIPermissionError` _and_ you haven't already tried to map (as we'd just get the same result)

If a match is found, then the mapping from the old id to the new is stored. Both of these Ids can be used for future matches to cover obscure situations where something odd happens to the original file.

## Review notes

I'm breaking this PR up into some smaller chunks:

 * [x] https://github.com/hypothesis/lms/pull/2842 - Make the assignment service get() method public

## Matching overview

* The matching is always performed on the **original file id**
* Matching is performed on a **mapped file id** if the assignment has a mapping from the original file id
* We use the file ids to retrieve **our records of when we last saw the files**, to get their names and sizes
* We retrieve the **list of files the current user can see in the current course**
* If any of them **match the names _and_ sizes** of the files we found in our records we have a match
* We will update the **assignment map from the original file id -> the new match** 
* We will proceed with the newly mapped file id

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,4 +1,12 @@
-from lms.services.exceptions import CanvasFileNotFoundInCourse
+from functools import cached_property, lru_cache
+from logging import getLogger
+
+from sqlalchemy import and_
+
+from lms.models import File
+from lms.services.exceptions import CanvasAPIPermissionError, CanvasFileNotFoundInCourse
+
+LOG = getLogger(__name__)
 
 
 class CanvasService:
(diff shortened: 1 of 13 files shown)


lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
-            be in the course but isn't.
+        :raise CanvasFileNotFoundInCourse: if the file is not accessible and
+            cannot be mapped to a file in the course
+        :raise CanvasAPIResourceNotFound: if the file is not valid at all, and
+            cannot be mapped to a file in the course
+        :raise CanvasAPIPermissionError: If we cannot retrieve the file for
+            any other reason
         """
+        # Check to see if we've mapped this file id to something else before
+        file_mapping = self._file_mapping(resource_link_id)
```
jon-betts, line 50: The file mapping is attached to the assignment, which means we now have to pass through the `resource_link_id` which is the id of the assignment. This is the cause of having to thread it through everything else.

The mapping object here is a small shim around interacting with a specific dict inside a key in the extra object below. It's made to look like a dict in the ways we use it.

lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
+        mapped_id = file_mapping.get(file_id, default=file_id)
+
+        # If we are told to check up front, try and fix the file if it's not in
+        # the current course (as far as we can tell from this users point of
+        # view)
+        already_mapped = False
+        if check_in_course and not self._is_file_in_course(mapped_id, course_id):
+            mapped_id = self._match_file_in_course(file_id, course_id, file_mapping)
```
jon-betts, line 58: Much of the confusing complexity of this process is the fact that we perform the mapping in two locations. Here if `check_in_course` is enabled and the file isn't in the course, and later on if we didn't do this.

This means a teacher could get either test, depending on whether the file is in the course or not. A file in the course that somehow (how?) can't be accessed by the teacher might trigger the next search. Although I think this is actually impossible, as if the file is "in the course" from the point of view of the teacher, they can see it. Nevertheless, I don't want to rely on Canvas being sane.

lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
 
-        return self.api.public_url(file_id)
+        try:
+            return self.api.public_url(mapped_id)
+
+        except CanvasAPIPermissionError:
+            # If we've already mapped this once, we aren't going to get a
+            # better answer now. So bail out
+            if already_mapped:
+                raise
```
jon-betts, line 71: Don't map twice, although I suspect this is impossible to reach for real.

lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
+        )
+        return CanvasFileMapping(assignment.extra)
+
+    @cached_property
+    def _application_instance(self):
+        return self._application_instance_service.get()
+
+    @lru_cache
+    def _files_in_course(self, course_id):
+        return self.api.list_files(course_id)
```
jon-betts, line 98: We cache the file results here because we might use them for the "in the course" check and also the mapping. This would save us an API call in that case.

lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
+
+        :param file_id: The original file to match
+        :param course_id: The course to find a matching file in
+        :return: A mapped file, or None
+        """
+
+        file_ids_to_match = [file_id]
+
+        # We match not only the original id, but also the mapped id too. This
+        # allows us to code with some low probability scenarios where we have
```
jon-betts, line 122:
```suggestion
        # allows us to code for some low probability scenarios where we have
```

lms/services/canvas.py
```diff
@@ -1,41 +1,198 @@
...
+                    File.application_instance_id == self._application_instance.id,
+                    File.lms_id.in_(file_ids_to_match),
+                    File.type == "canvas_file",
+                )
+            )
+            .all()
+        )
+
+        if files_to_match:
+            matching_keys = set((file.name, file.size) for file in files_to_match)
```
jon-betts, line 142: In 99.9% of cases these are going to be the same, but it's conceivable they could be different if one or the other file was modified in some way. Can you even rename things in Canvas?

tests/unit/lms/resources/_js_config_test.py
```diff
@@ -187,31 +188,24 @@ def test_it_raises_if_theres_no_ApplicationInstance(
 class TestAddCanvasFileID:
     """Unit tests for JSConfig.add_canvas_file_id()."""
 
-    def test_it_adds_the_viaUrl_api_config(self, js_config):
-        js_config.add_canvas_file_id(
-            "example_canvas_course_id", "example_canvas_file_id"
```
jon-betts, line 192: I don't think we need to prefix strings with "example_" or "test_" in tests, it's kind of obvious by context.

seanh (changes requested):
I think this might be case where storing `{file_id: mapped_file_id}` mappings in a `ModuleItemConfiguration.extra` JSON column seemed like a good idea when we first discussed it, but in practice it turns out to be taking JSON too far and the code ends up being quite hard to follow. I think we might want a `CanvasFileMapping` table instead.

I think breaking a separate `CanvasFileMappingService` out of `CanvasService` might help, where `CanvasFileMappingService` provides a simple `get()` and `set()` interface for working with the file mappings in the DB and then `CanvasService` implements the actual logic (e.g. when do we set a file mapping?) on top of that. Breaking it up into smaller units should make it easier to understand.

I found the JSON stuff on top of the new `ModuleItemConfiguration.extra` column really hard to follow. There's just too many layers of indirection to keep in your head at once. We actually have four nested dicts here: `ModuleItemConfiguration.extra` is a dict containing a `"canvas"` key whose value is a dict with a `"file_mapping"` key whose value is a dict mapping file IDs to file IDs. It uses our `ApplicationSettings` thing which enables two levels of nested dicts with mutation tracking, but then takes it a step further because the `value` in `ApplicationSettings.set(group, key, value)` is itself a dict that needs third-level mutation tracking. I understand what the code is actually doing now but it took me quite a while to figure it out. And that's just the model stuff, I haven't understood the actual course copy / file mapping logic that's implemented on top of the model yet.

 In the service method I see:

```python
file_mapping.get(file_id, default=file_id)
```

To understand that I need to read quite a lot of code and deduce that:

* `file_mapping` came from the `_file_mapping()` helper method, which returns `ModuleItemConfiguration.extra` wrapped in a `CanvasFileMapping` object.
* At the DB level `ModuleItemConfiguration.extra` is a dict (JSON column). Normally mutations to this dict wouldn't be tracked, but we use SQLAlchemy's `MutableDict` so that _top-level_ mutations are tracked
* At the model level `ModuleItemConfiguration.extra` is wrapped in an `ApplicationSettings` so it has the `mic.extra.get(group, key) -> value` and `mic.extra.set(group, key, value)` methods. The `set()` method is implemented in a particular way that adds tracking of _second-level_ dict mutations
* This `ApplicationSettings` is then wrapped in a `CanvasFileMapping` which effectively defaults the `group` and `key` in `extra.get(group, key)` and `extra.set(group, key, value)` to `"canvas"` and `"file_mapping"`, and also changes the `ApplicationSettings` interface by adding a `default` argument to `get()` and by replacing `set` with `__setitem__()`
* Also in this case the `value` in `mic.extra.get(group, key) -> value` and `mic.extra.set(group, key, value)` is itself a dict mapping file IDs to file IDs. This is a third-level dict and mutations to it would not be tracked (even with `MutableDict` and `ApplicationSettings`) but `CanvasFileMapping.__setitem__()` is implemented in a particular way that adds tracking of third-level mutations

I think we may want to step back and ask whether we're over-using JSON in SQLAlchemy here.

I'd like to at least try doing this with `CanvasFileMappingService` that supplies a very simple `get()` and `set()` interface. The logic code in `CanvasService` would then depend on `CanvasFileMappingService`. `CanvasService` and `CanvasFileMappingService` should be unit tested in isolation from each other, with `CanvasService`'s tests using the mock `CanvasFileMappingService`. I think it's quite likely that that will make the code a _lot_ easier to understand:

```python
# Get an assignment's mapped ID or None.
mapped_id = canvas_file_mapping_svc.get(module_item_configuration, file_id)

# Get an assignment's mapped ID or file_id (instead of None).
mapped_id = canvas_file_mapping_svc.get(module_item_configuration, file_id) or file_id

# Set an assignment's mapped ID (overwriting any existing one).
canvas_file_mapping_svc.set(module_item_configuration, file_id, mapped_id)
```

I think `CanvasFileMappingService` could be implemented on top of a `ModuleItemConfiguration.extra` JSON column (no `ApplicationSettings`) but I think it might be better to get rid of the JSON column entirely and have `CanvasFileMappingService` wrap a separate `models.CanvasFileMapping` table instead

seanh:
Closing this in favour of https://github.com/hypothesis/lms/pull/2899/

I'm pretty sure that #2899 works and covers everything that this PR did, but I'm gonna give the code in this PR one last look after closing it in case I missed anything.

I ended up going with the same solution as in this PR, but I simplified the logic a little bit, and I also parcelled it out into a handful of separate components which I sent as separate PRs with their own tests to be kind to @marcospri  reviewing them. Here's a summary of what I did:

* I sent the `ModuleItemConfiguration.extra` column as a separate PR: https://github.com/hypothesis/lms/pull/2876

* I added `ModuleItemConfiguration.{get,set}_canvas_mapped_file_id()` methods to encapsulate the storage of mapped file IDs: https://github.com/hypothesis/lms/pull/2901

  These still use the JSON `extra` column but `ApplicationSettings` isn't involved and the `CanvasFileMapping` class is gone. It ended up being just two one-liner methods on the model!

  (I also tried adding these methods to `AssignmentService`, but it worked out simpler with them on the model.)

* I added `FileService.get()` to encapsulate the storage code for retrieving a `File` from the DB: https://github.com/hypothesis/lms/pull/2892 and https://github.com/hypothesis/lms/pull/2902

* I added `CanvasService.can_see_file_in_course()` as a separate method with its own unit tests and PR. This just separates out this logically separable unit (a boolean function) from the top-level logic: https://github.com/hypothesis/lms/pull/2893

* Similarly I added `CanvasService.find_matching_file_in_course()`: https://github.com/hypothesis/lms/pull/2894

* I added per-request caching to `CanvasAPIClient.list_files()` and `CanvasAPIClient.public_url()`. This allows the higher-level business logic to be simple about calling these methods repeatedly without sending duplicate network requests. https://github.com/hypothesis/lms/pull/2907

* Finally, I added a simplified version of the `public_url()` business logic from this PR based on all the above: https://github.com/hypothesis/lms/pull/2907

* P.S. I also added `ApplicationInstance.files` and `File.application_instance` convenience attributes (SQLAlchemy relations). This is mostly because it makes test factories easier to write because you can use subfactories: https://github.com/hypothesis/lms/pull/2875/

PR #2789, opened 2021-06-07
Add gap between rows in Assignment details form

robertknight:
Fix an issue where the "Assignment content" and "Group assignment" rows of the "Assignment details" form could touch visually if the assignment content row flowed onto multiple lines. The left column had a bottom margin set but the right column did not.

Fix the issue by replacing the left column's bottom margin with a `row-gap` which applies to the whole row. This gap is additive to any cell margins, so the bottom margin for the header row also had to be adjusted.

Strictly speaking `row-gap` requires browsers which are a little newer than our baseline (Safari 12, Chrome 66, Firefox 61), but only slightly. In older browsers the gap will simply be missing, which I think we can live with, as the numbers of such users is very small.

**Before:**

Assignment URL truncation

**After:**

Assignment row gap

### DIFF
--- a/lms/static/styles/components/_FilePickerApp.scss
+++ b/lms/static/styles/components/_FilePickerApp.scss
@@ -3,6 +3,7 @@
 .FilePickerApp__form {
   display: grid;
   grid-template-columns: auto 1fr;
+  row-gap: 1em;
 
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
👍🏻

PR #2790, opened 2021-06-07
Add space between checkbox and group set select

robertknight:
Wrap the contents of the space below the "This is a group assignment" checkbox in a `GroupConfigSelector__select` div with a top margin. The top margin is the same as that of a `<p>` element so the visual margin is the same whether this area is displaying the authorization prompt (which begins with a paragraph) or the group set `<select>` control and its `<label>`.

**Before:**

Group set select no space

**After:**

Group set auth

Group set select

### DIFF
--- a/lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
+++ b/lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
@@ -118,52 +118,56 @@ export default function GroupConfigSelector({
           This is a group assignment
         </LabeledCheckbox>
       </div>
-      {fetchError && useGroupSet && (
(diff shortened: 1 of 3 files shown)


robertknight:
While poking at this I discovered that the vertical misalignment between the "Group set" label's baseline and the text in the `<select>` is caused by a universal selector inherited from the lms.css stylesheet, which shouldn't be used at all in this context.

robertknight:
The alignment of the "Group set" label and `<select>` text is fixed by https://github.com/hypothesis/lms/pull/2793.

lyzadanger (approved):
Let's get this landed, as it takes care of the immediate need, and then I'll step back and look at some of this layout/CSS stuff as a whole.

PR #2792, opened 2021-06-08
Don't create_all when alembic is present on make db

marcospri:
If starting from a empty DB we should create_all and stamp the DB with
the correct alembic version so migrate head does the right thing.

If using `make db` on a non empty DB we should rely on alembic to get us
to the right state, ie don't `create_all` or stamp the DB if alembic is
present.

# Testing

- Drop your DB tables
- Checkout to some commit before a table is added, eg `a2fe3341`
- `make db`. It should create the tables an stamp the DB (but not run alembic!)

```
INFO  [alembic.runtime.migration] Running stamp_revision  -> 5f80062abcf4                                                     
```
- Checkout master and `make db` again. It should run the migrations from the stamped version

```
INFO  [alembic.runtime.migration] Running upgrade 5f80062abcf4 -> bb11b5b06274, Adds new grouping table.  
```

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -88,7 +88,7 @@ BASE = declarative_base(
 )
 
 
-def init(engine, drop=False):
+def init(engine, drop=False, stamp=True):  # pragma: nocover
     """
     Create all the database tables if they don't already exist.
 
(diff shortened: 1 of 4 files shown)


lms/db/__init__.py
```diff
@@ -160,10 +154,9 @@ def includeme(config):
     # Create the SQLAlchemy engine and save a reference in the app registry.
     engine = make_engine(config.registry.settings)
     config.registry["sqlalchemy.engine"] = engine
-    init(engine)
```
marcospri, line 163: This is the main change, don't just init right away, this will create the tables that alembic is expected to create on the second step of `make db` because if alebmic_vesion exist we are not stamping so we were trying to do both, `create_all` and `alembic upgrade head` (without stamping)

jon-betts:
How do we do local development when adding new tables and things?

marcospri:
> How do we do local development when adding new tables and things?

Not sure, I personally use alembic upgrade/downgrade/autogenerate but I don't think this PR makes that easier or harder, the current command on master makes it impossible to get and up to date DB once migrations are committed to the repository.

seanh (approved):

PR #2793, opened 2021-06-08
Align CSS used by static form and frontend apps

robertknight:
This PR incrementally aligns the CSS used by static forms (eg. https://lms.hypothes.is/welcome, https://lms.hypothes.is/flags) with the CSS used by the frontend mini-apps (file picker, assignment launch etc.) by:

- Converting the entry point for the static form's CSS bundle (`lms.css`) to a SASS file so it can use the same mixins, variables, base styles etc. as other bundles
- Using the same CSS reset in lms.scss as frontend_apps.scss
- Changing the font stack and default text color in the static forms to match that used by the frontend apps
- Removing several universal / generic selectors in lms.css

The initial motivation for this was that I uncovered that the frontend apps were loading both the `lms.css` and `frontend_apps.css` bundles and some universal selectors in `lms.css` were causing unexpected effects in the frontend apps (`vertical-align: top`). Aligning the base element styles, typography and CSS reset means that we can combine these bundles in the same page without unexpected problems. For most frontend apps we shouldn't be loading the `lms.css` bundle at all, but I think it makes sense to do this alignment even if we make that change for consistency across the app.

While I was looking at this I notice that we load the Material Icons font on almost every page but don't use these icons anywhere (grep for references to `material-icons` classes). Hence I removed the icons from `base.html.jinja2`.

----

nb. These screenshots are on macOS. The font change will look a little different on other systems.

**Before:**

lms-welcome-before-2

**After:**

lms-welcome-after-2

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -125,7 +125,7 @@ gulp.task(
 );
 
 var cssBundles = [
-  './lms/static/styles/lms.css',
+  './lms/static/styles/lms.scss',
   './lms/static/styles/reports.css',
   './lms/static/styles/frontend_apps.scss',
   './lms/static/styles/ui-playground/ui-playground.scss',
--- a/lms/static/styles/lms.scss
+++ b/lms/static/styles/lms.scss
@@ -1,24 +1,17 @@
+@use 'normalize.css/normalize';
+
+@use 'variables' as var;
+
 * {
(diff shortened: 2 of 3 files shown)


lyzadanger (approved):
👍🏻

PR #2795, opened 2021-06-09
Use class for table styles

robertknight:
This PR changes some table styles, currently used only by the `/flags` page, to use a class rather than element selector. This prevents these styles unexpectedly applying to other `<table>` elements. There should be no visual changes.

I also removed a few unused CSS classes and vendor prefixes. See commits for details.


lyzadanger (approved):
This is fine for now. I'm starting to foment rebellion against any element styles in our CSS (other than a reset). We'll see how that pans out, pitchforks are at the ready.

PR #2797, opened 2021-06-09
Fix regression in LTI launch app height

robertknight:
https://github.com/hypothesis/lms/pull/2793 removed a `height: 100%` rule for the `<body>` and `<html>` elements which caused the assignment content iframe to shrink to a small height because `BasicLtiLaunchApp`'s styles assumed that the parent element filled the page.

### DIFF
--- a/lms/static/styles/frontend_apps.scss
+++ b/lms/static/styles/frontend_apps.scss
@@ -31,6 +31,13 @@
 // Shared styles from frontend-shared package
 @use '@hypothesis/frontend-shared/styles';
 
+// This ensures that root components which set `width: 100%; height: 100%`
+// (eg. the `BasicLtiLaunchApp`) will fill the viewport.
+html,
+body {
+  height: 100%;
+}
+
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):

lms/static/styles/frontend_apps.scss
```diff
@@ -31,6 +31,12 @@
 // Shared styles from frontend-shared package
 @use '@hypothesis/frontend-shared/styles';
 
+// This ensures that root components which set `width: 100%; height: 100%`
+// (eg. the `BasicLtiLaunchApp`) will fill the viewport.
+html, body {
+  height: 100%;
+}
```
robertknight, line 38: This rule used to be in `lms.css` but isn't needed for the static forms. It is not needed for all of the frontend mini-apps either and we might be able to implement a more narrowly scoped change for `BasicLtiLaunchApp` in future. For now I've just documented why it is needed.

PR #2800, opened 2021-06-10
Download real Blackboard files

seanh:
Get real download URLs from the Blackboard API instead of fake ones.

### DIFF
--- a/lms/resources/_js_config.py
+++ b/lms/resources/_js_config.py
@@ -67,7 +67,7 @@ class JSConfig:
                 "path": self._request.route_path(
                     "blackboard_api.files.via_url",
                     course_id=self._request.params["context_id"],
-                    file_id=document_url.split("/")[3],
+                    _query={"document_url": document_url},
                 ),
             }
         else:
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -42,8 +42,7 @@ def includeme(config):
         "blackboard_api.courses.files.list", "/api/blackboard/courses/{course_id}/files"
     )
     config.add_route(
-        "blackboard_api.files.via_url",
-        "/api/blackboard/courses/{course_id}/files/{file_id}/via_url",
+        "blackboard_api.files.via_url", "/api/blackboard/courses/{course_id}/via_url"
     )
 
     config.add_route(
--- a/lms/services/blackboard_api/_schemas.py
+++ b/lms/services/blackboard_api/_schemas.py
@@ -8,21 +8,37 @@ from marshmallow import EXCLUDE, Schema, fields, post_load
 from lms.validation._base import RequestsResponseSchema
 
 
+class _FileSchema(Schema):
+    """Shared base schema for Blackboard file dicts."""
+
+    class Meta:
+        unknown = EXCLUDE
+
+    id = fields.Str(required=True)
+    display_name = fields.Str(data_key="name", required=True)
+    updated_at = fields.Str(data_key="modified", required=True)
+
+    @post_load
+    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
+        data["id"] = f"blackboard://content-resource/{data['id']}/"
+        return data
+
+
 class BlackboardListFilesSchema(RequestsResponseSchema):
     """Schema for Blackboard API /courses/{courseId}/resources responses."""
 
-    class FileSchema(Schema):
-        """Schema for an individual file dict inside the "results" list."""
-
-        class Meta:
-            unknown = EXCLUDE
-
-        id = fields.Str(required=True)
-        display_name = fields.Str(data_key="name", required=True)
-        updated_at = fields.Str(data_key="modified", required=True)
-
-    results = fields.List(fields.Nested(FileSchema), required=True)
+    results = fields.List(fields.Nested(_FileSchema), required=True)
 
     @post_load
-    def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
+    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
         return data["results"]
+
+
+class BlackboardPublicURLSchema(RequestsResponseSchema, _FileSchema):
+    """Schema for Blackboard /courses/{courseId}/resources/{resourceId} responses."""
+
+    downloadUrl = fields.Str(required=True)
+
+    @post_load
+    def post_load(self, data, **_kwargs):
+        return data["downloadUrl"]
--- a/lms/services/blackboard_api/service.py
+++ b/lms/services/blackboard_api/service.py
@@ -1,5 +1,17 @@
-from lms.services.blackboard_api._schemas import BlackboardListFilesSchema
+import re
+
+from lms.services.blackboard_api._schemas import (
+    BlackboardListFilesSchema,
+    BlackboardPublicURLSchema,
+)
+from lms.services.exceptions import BlackboardFileNotFoundInCourse, HTTPError
 from lms.validation.authentication import OAuthTokenResponseSchema
+
+#: A regex for parsing just the file_id part out of one of our custom
+#: blackboard://content-resource/<file_id>/ URLs.
+DOCUMENT_URL_REGEX = re.compile(
+    r"blackboard:\/\/content-resource\/(?P<file_id>[^\/]*)\/"
+)
 
 
 class BlackboardAPIClient:
@@ -58,6 +70,20 @@ class BlackboardAPIClient:
             oauth=True,
             schema=BlackboardListFilesSchema,
         ).validated_data
+
+    def public_url(self, course_id, document_url):
+        file_id = DOCUMENT_URL_REGEX.search(document_url)["file_id"]
+
+        try:
+            return self._http_service.get(
+                self._api_url(f"courses/uuid:{course_id}/resources/{file_id}"),
+                oauth=True,
+                schema=BlackboardPublicURLSchema,
+            ).validated_data
+        except HTTPError as err:
+            if err.response.status_code == 404:
+                raise BlackboardFileNotFoundInCourse(file_id) from err
+            raise
 
     def _api_url(self, path):
         """Return the full Blackboard API URL for the given path."""
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -2,7 +2,6 @@
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
-from lms.services import NoOAuth2Token, ProxyAPIAccessTokenError
 from lms.views import helpers
 
 
@@ -22,23 +21,14 @@ class BlackboardFilesAPIViews:
     @view_config(request_method="GET", route_name="blackboard_api.files.via_url")
     def via_url(self):
         """Return the Via URL for annotating the given Blackboard file."""
-        # Get the user's access token from the DB.
-        # We're not actually going to *use* the access token, since this view
-        # just returns hard-coded URLs. But we want to *get* the access token
-        # so that we can raise an exception if it's missing and trigger the
-        # authentication flow.
-        try:
-            self.request.find_service(name="oauth2_token").get()
-        except NoOAuth2Token as err:
-            raise ProxyAPIAccessTokenError() from err
-
-        # Look up the file_id in a temporary hard-coded list of public URLs.
-        public_url = {
-            "123": "https://h.readthedocs.io/_/downloads/en/latest/pdf/",
-            "456": "https://h.readthedocs.io/_/downloads/client/en/latest/pdf/",
-            "789": "https://packaging.python.org/_/downloads/en/latest/pdf/",
-        }[self.request.matchdict["file_id"]]
 
         return {
-            "via_url": helpers.via_url(self.request, public_url, content_type="pdf"),
+            "via_url": helpers.via_url(
+                self.request,
+                self.blackboard_api_client.public_url(
+                    self.request.matchdict["course_id"],
+                    self.request.params["document_url"],
+                ),
+                content_type="pdf",
+            )
         }
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -1,5 +1,6 @@
 from lms.services.canvas import CanvasService
 from lms.services.exceptions import (
+    BlackboardFileNotFoundInCourse,
     CanvasAPIError,
     CanvasAPIPermissionError,
     CanvasAPIServerError,
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -220,6 +220,18 @@ class CanvasFileNotFoundInCourse(ServiceError):
         super().__init__(self.details)
 
 
+class BlackboardFileNotFoundInCourse(ServiceError):
+    """A Blackboard file ID wasn't found in the current course."""
+
+    explanation = (
+        "Hypothesis couldn't find the assignment's PDF file in the Blackboard course."
+    )
+
+    def __init__(self, document_url):
+        self.details = {"document_url": document_url}
+        super().__init__(self.details)
+
+
 class HTTPError(ServiceError):
     """A problem with an HTTP request to an external service."""
 
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -8,6 +8,7 @@ from pyramid.view import (
 )
 
 from lms.services import (
+    BlackboardFileNotFoundInCourse,
     CanvasAPIPermissionError,
     CanvasFileNotFoundInCourse,
     ProxyAPIAccessTokenError,
@@ -100,6 +101,7 @@ class ExceptionViews:
         )
 
     @exception_view_config(context=ProxyAPIError)
+    @exception_view_config(context=BlackboardFileNotFoundInCourse)
     def proxy_api_error(self):
         return self.error_response(
             message=self.context.explanation, details=self.context.details
--- a/tests/unit/lms/resources/_js_config_test.py
+++ b/tests/unit/lms/resources/_js_config_test.py
@@ -219,7 +219,7 @@ class TestAddDocumentURL:
 
         assert js_config.asdict()["api"]["viaUrl"] == {
             "authUrl": "http://example.com/api/blackboard/oauth/authorize",
-            "path": "/api/blackboard/courses/test_course_id/files/xyz123/via_url",
+            "path": "/api/blackboard/courses/test_course_id/via_url?document_url=blackboard%3A%2F%2Fcontent-resource%2Fxyz123",
         }
 
     def test_it_sets_the_document_url(self, js_config, submission_params):
--- a/tests/unit/lms/services/blackboard_api/_schemas_test.py
+++ b/tests/unit/lms/services/blackboard_api/_schemas_test.py
@@ -1,85 +1,118 @@
 import pytest
 
-from lms.services.blackboard_api._schemas import BlackboardListFilesSchema
+from lms.services.blackboard_api._schemas import (
+    BlackboardListFilesSchema,
+    BlackboardPublicURLSchema,
+)
 from lms.validation._exceptions import ValidationError
 from tests import factories
 
 
 class TestBlackboardListFilesSchema:
-    def test_valid(self, json_data):
+    def test_valid(self, list_files_response):
         schema = BlackboardListFilesSchema(
-            factories.requests.Response(json_data=json_data)
+            factories.requests.Response(json_data=list_files_response)
         )
 
         result = schema.parse()
 
         assert result == [
             {
-                "id": "_7851_0",
+                "id": "blackboard://content-resource/_7851_0/",
                 "updated_at": "2008-05-06T07:26:35.000z",
                 "display_name": "File_0.pdf",
             },
             {
-                "id": "_7851_1",
+                "id": "blackboard://content-resource/_7851_1/",
                 "updated_at": "1983-05-26T02:37:23.000z",
                 "display_name": "File_1.pdf",
             },
         ]
 
-    def test_it_raises_if_results_is_missing(self, json_data, assert_raises):
-        del json_data["results"]
+    def test_it_raises_if_results_is_missing(self, list_files_response):
+        del list_files_response["results"]
 
-        assert_raises()
+        assert_raises(BlackboardListFilesSchema, list_files_response)
 
-    def test_it_raises_if_results_isnt_a_list(self, json_data, assert_raises):
-        json_data["results"] = None
+    def test_it_raises_if_results_isnt_a_list(self, list_files_response):
+        list_files_response["results"] = None
 
-        assert_raises()
+        assert_raises(BlackboardListFilesSchema, list_files_response)
 
     @pytest.mark.parametrize("missing_field", ["id", "modified", "name"])
     def test_it_raises_if_a_required_field_is_missing(
-        self, json_data, assert_raises, missing_field
+        self, list_files_response, missing_field
     ):
-        del json_data["results"][0][missing_field]
+        del list_files_response["results"][0][missing_field]
 
-        assert_raises()
+        assert_raises(BlackboardListFilesSchema, list_files_response)
 
     @pytest.mark.parametrize("invalid_field", ["id", "modified", "name"])
-    def test_it_raises_if_a_field_is_invalid(
-        self, json_data, assert_raises, invalid_field
+    def test_it_raises_if_a_field_is_invalid(self, list_files_response, invalid_field):
+        list_files_response["results"][0][invalid_field] = 23
+
+        assert_raises(BlackboardListFilesSchema, list_files_response)
+
+
+class TestBlackboardPublicURLSchema:
+    def test_valid(self, single_file_response):
+        schema = BlackboardPublicURLSchema(
+            factories.requests.Response(json_data=single_file_response)
+        )
+
+        result = schema.parse()
+
+        assert result == "https://example.com/file0.pdf"
+
+    def test_it_raises_if_a_the_body_isnt_a_dict(self):
+        assert_raises(BlackboardPublicURLSchema, [])
+
+    @pytest.mark.parametrize("missing_field", ["id", "modified", "name"])
+    def test_it_raises_if_a_required_field_is_missing(
+        self, single_file_response, missing_field
     ):
-        json_data["results"][0][invalid_field] = 23
+        del single_file_response[missing_field]
 
-        assert_raises()
+        assert_raises(BlackboardPublicURLSchema, single_file_response)
 
-    @pytest.fixture
-    def json_data(self):
-        """Return the JSON body of a valid Blackboard Files API response."""
-        return {
-            "results": [
-                {
-                    "id": "_7851_0",
-                    "modified": "2008-05-06T07:26:35.000z",
-                    "name": "File_0.pdf",
-                    "unknown_field": "this_should_be_excluded",
-                },
-                {
-                    "id": "_7851_1",
-                    "modified": "1983-05-26T02:37:23.000z",
-                    "name": "File_1.pdf",
-                },
-            ]
-        }
+    @pytest.mark.parametrize("invalid_field", ["id", "modified", "name"])
+    def test_it_raises_if_a_field_is_invalid(self, single_file_response, invalid_field):
+        single_file_response[invalid_field] = 23
+
+        assert_raises(BlackboardPublicURLSchema, single_file_response)
 
     @pytest.fixture
-    def assert_raises(self, json_data):
-        def assert_raises():
-            """Assert that the schema raises when parsing json_data."""
-            schema = BlackboardListFilesSchema(
-                factories.requests.Response(json_data=json_data)
-            )
+    def single_file_response(self, list_files_response):
+        # The response body from the single-file API looks the same as a single
+        # one of the file dicts from the list-files API.
+        return list_files_response["results"][0]
 
-            with pytest.raises(ValidationError):
-                schema.parse()
 
-        return assert_raises
+@pytest.fixture
+def list_files_response():
+    """Return the JSON body of a valid Blackboard Files API response."""
+    return {
+        "results": [
+            {
+                "id": "_7851_0",
+                "modified": "2008-05-06T07:26:35.000z",
+                "name": "File_0.pdf",
+                "downloadUrl": "https://example.com/file0.pdf",
+                "unknown_field": "this_should_be_excluded",
+            },
+            {
+                "id": "_7851_1",
+                "modified": "1983-05-26T02:37:23.000z",
+                "name": "File_1.pdf",
+                "downloadUrl": "https://example.com/file1.pdf",
+            },
+        ]
+    }
+
+
+def assert_raises(schema_class, json_data):
+    """Assert that the schema raises when parsing json_data."""
+    schema = schema_class(factories.requests.Response(json_data=json_data))
+
+    with pytest.raises(ValidationError):
+        schema.parse()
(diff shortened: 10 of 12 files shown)


lms/resources/_js_config.py
```diff
@@ -68,7 +68,7 @@ def add_document_url(self, document_url):
                 "path": self._request.route_path(
                     "blackboard_api.files.via_url",
                     course_id=self._request.params["context_id"],
-                    file_id=document_url.split("/")[3],
+                    _query={"file_id": document_url},
```
seanh, line 71: `JSConfig` just passes the file ID directly from the DB (so in the `blackboard://...` form) to the frontend as a query param on the Via URL API call. The frontend just treats this ID as an opaque string, so while `blackboard://...` do flow through the frontend the frontend code doesn't know anything about them.

lms/routes.py
```diff
@@ -43,7 +43,7 @@ def includeme(config):
     )
     config.add_route(
         "blackboard_api.files.via_url",
-        "/api/blackboard/courses/{course_id}/files/{file_id}/via_url",
+        "/api/blackboard/courses/{course_id}/files/via_url",
```
seanh, line 46: This is a bit unsatisfactory. Botu a course ID and a file ID are needed to get a Via URL for a Blackboard file. So the `/api/blackboard/courses/{course_id}/files/{file_id}/via_url` route was nice because both the course ID and file ID were in the path. But we're using `blackboard://...` URLs as the file IDs and it's a pain to put a URL in the path part of another URL. So I've moved it out into a query param. Another option would be to leave it in the path but encode it

lms/services/blackboard_api/_schemas.py
```diff
@@ -26,3 +26,20 @@ class Meta:
...
+            unknown = EXCLUDE
+
+        id = fields.Str(required=True)
+        downloadUrl = fields.Str(required=True)
+
+    results = fields.List(fields.Nested(FileSchema), required=True)
+
+    @post_load
+    def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
+        return data["results"]
```
seanh, line 45: This is quite similar to the `BlackboardListFilesSchema` above. They're actually both schemas for the Blackboard list files API. But they do different things with the responses. `BlackboardListFilesSchema` returns a list of files each with ID, name and modified date, suitable for returning to our frontend when it wants to show the teacher a list of files. `BlackboardPublicURLSchema` returns a list of files each with ID and download URL, suitable for the `public_url()` method to iterate over to find the download URL of the file with a given ID.

I think it might be better to have just one `BlackboardListFilesSchema` here that's used by both service methods, in which case I think some of the work that the schema currently does may need to move into the service instead.

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +56,34 @@ def list_files(self, course_id):
...
+
+        files = self._http_service.get(
+            self._api_url(f"/courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardPublicURLSchema,
+        ).validated_data
+
+        for file_ in files:
+            if file_["id"] == file_id:
+                return file_["downloadUrl"]
```
seanh, line 83: Something needs to take an ID that we received from the Blackboard Files API and wrap it in a `blackboard://...` URL for storage in our DB. And something needs to take that `blackboard://...` URL and unwrap just the ID from it, because that's what you need when getting the file's public URL from the Blackboard API.

I thought it would be nice if the service's `list_files()` did the wrapping and the service's `public_url()` did the unwrapping, because then the wrapping and unwrapping are next to each other in the same file, and knowledge about the `blackboard://...` URL scheme is encapsulated in this one file:

* When creating an assignment `BlackboardAPIClient.list_files()` returns a list of files with `blackboard://...` IDs to the frontend. These are opaque strings to the frontend. When the user selects a file the frontend posts the selected ID to the `configure_module_item()` view which saves it to the DB. The ID is an opaque string to the `configure_module_item()` view and the DB as well
* When launching an assignment we look up the `document_url` (which is the `blackboard://...` URL) in the DB and the `BlackboardAPIClient.public_url()` view unwraps it to just the ID for use with the Blackboard API for getting a public URL

The only place where the `blackboard://...` format leaks out is in `JSConfig` which has to look at it to know whether we're dealing with a Blackboard Files assignment:

https://github.com/hypothesis/lms/blob/1a0324d8e0e8213bb0f27c46a2561e20206a17af/lms/resources/_js_config.py#L65-L73

This could perhaps be moved into a `BlackboardAPIClient.is_blackboard_files_url(url)` method, and then knowledge of the `blackboard://...` format would be entirely encapsulated within `BlackboardAPIClient`.

`file_id.split("/")[3]` is a little unsatisfactory but we do know for sure that this is a valid URL in that particular format since we created it, so I think it's actually fine. A regex might be more readable though.

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +56,34 @@ def list_files(self, course_id):
...
+    def public_url(self, course_id, file_id):
+        # Extract just the {file_id} part out of the
+        # "blackboard://content-resource/{file_id}/" string.
+        file_id = file_id.split("/")[3]
+
+        files = self._http_service.get(
+            self._api_url(f"/courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardPublicURLSchema,
+        ).validated_data
```
seanh, line 79: To get a public URL we actually have to call the _list files_ API and then iterate over the list of file dicts looking for the one with the ID that we're looking for. The download URLs are in the file dicts. This sucks because on every student launch we might have to paginate through multiple responses to get the full list of files. But I'm not sure if the Blackboard API gives us any other option. I need to take a look at their API docs and see if there's any other way

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +56,34 @@ def list_files(self, course_id):
...
+
+        for file_ in files:
+            if file_["id"] == file_id:
+                return file_["downloadUrl"]
+
+        assert False
```
seanh, line 85: The `assert False` needs to be replaced with raising a suitable exception

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +56,34 @@ def list_files(self, course_id):
...
 
+        for file_ in files:
+            file_["id"] = f"blackboard://content-resource/{file_['id']}/"
+
+        return files
+
+    def public_url(self, course_id, file_id):
+        # Extract just the {file_id} part out of the
+        # "blackboard://content-resource/{file_id}/" string.
+        file_id = file_id.split("/")[3]
```
seanh, line 73: `file_id.split("/")[3]` is a little unsatisfactory but we do know for sure that this is a valid URL in that particular format since we created it, so I think it's actually fine. A regex might be more readable though.

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +56,34 @@ def list_files(self, course_id):
...
         """
-        return self._http_service.get(
+        files = self._http_service.get(
             self._api_url(f"courses/uuid:{course_id}/resources"),
             oauth=True,
             schema=BlackboardListFilesSchema,
         ).validated_data
 
+        for file_ in files:
+            file_["id"] = f"blackboard://content-resource/{file_['id']}/"
```
seanh, line 66: Something needs to take an ID that we received from the Blackboard Files API and wrap it in a `blackboard://...` URL for storage in our DB. And something needs to take that `blackboard://...` URL and unwrap just the ID from it, because that's what you need when getting the file's public URL from the Blackboard API.

I thought it would be nice if the service's `list_files()` did the wrapping and the service's `public_url()` did the unwrapping, because then the wrapping and unwrapping are next to each other in the same file, and knowledge about the `blackboard://...` URL scheme is encapsulated in this one file:

* When creating an assignment `BlackboardAPIClient.list_files()` returns a list of files with `blackboard://...` IDs to the frontend. These are opaque strings to the frontend. When the user selects a file the frontend posts the selected ID to the `configure_module_item()` view which saves it to the DB. The ID is an opaque string to the `configure_module_item()` view and the DB as well
* When launching an assignment we look up the `document_url` (which is the `blackboard://...` URL) in the DB and the `BlackboardAPIClient.public_url()` view unwraps it to just the ID for use with the Blackboard API for getting a public URL

The only place where the `blackboard://...` format leaks out is in `JSConfig` which has to look at it to know whether we're dealing with a Blackboard Files assignment:

https://github.com/hypothesis/lms/blob/1a0324d8e0e8213bb0f27c46a2561e20206a17af/lms/resources/_js_config.py#L65-L73

This could perhaps be moved into a `BlackboardAPIClient.is_blackboard_files_url(url)` method, and then knowledge of the `blackboard://...` format would be entirely encapsulated within `BlackboardAPIClient`.

seanh: When slightly weird thing about this is that when our `list_files()` API returns a list of files to our frontend the `blackboard://...` URL is in the `id` fields of the files, whereas when it's stored in the `ModuleItemConfiguration` table in the DB the `blackboard://...` URL is in the `document_url` column. So sometimes it's an ID, sometimes it's a URL

lms/views/api/blackboard/files.py
```diff
@@ -22,23 +21,12 @@ def list_files(self):
...
-        }[self.request.matchdict["file_id"]]
 
         return {
-            "via_url": helpers.via_url(self.request, public_url, content_type="pdf"),
+            "via_url": helpers.via_url(
+                self.request,
+                self.blackboard_api_client.public_url(
+                    self.request.matchdict["course_id"], self.request.params["file_id"]
+                ),
+            )
```
seanh, line 31: One-liner view again. As commented on the route above, I think it might be nicer if both `course_id` and `file_id` came from `matchdict` instead of one of them coming from `params`

lms/services/blackboard_api/service.py
seanh: Actually I think the code is nicer if we _never_ parse `blackboard://...` URLs. We're given a `blackboard://...` URL for the file that we're looking for. We can wrap all the file IDs from the Blackboard API into `blackboard://...` URLs and then compare _those_ to the `blackboard://...` URL that we were given. That way we only ever need to format `blackboard://...` URLs, we never need to parse them.

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,34 @@ def list_files(self, course_id):
...
         ).validated_data
 
+        return [
+            {
+                "id": file_["id"],
+                "display_name": file_["name"],
+                "updated_at": file_["modified"],
+            }
+            for file_ in files
+        ]
```
seanh, line 69: The `list_files()` method is now doing the work of renaming the `name` and `modified` fields and other excluding fields other than `id`, `display_name` and `updated_at`. This is because the `public_url()` method below now uses the same `BlackboardListFilesSchema` and `public_url()` _doesn't_ want this renaming and excluding, so I moved the renaming and excluding out of `BlackboardListFilesSchema`

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,34 @@ def list_files(self, course_id):
...
+    def public_url(self, course_id, file_id):
+        files = self._http_service.get(
+            self._api_url(f"/courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardListFilesSchema,
+        ).validated_data
+
+        for file_ in files:
+            if file_["id"] == file_id:
+                return file_["downloadUrl"]
```
seanh, line 80: To get a public URL we actually have to call the _list files_ API and then iterate over the list of file dicts looking for the one with the ID that we're looking for. The download URLs are in the file dicts. This sucks because on every student launch we might have to paginate through multiple responses to get the full list of files. But I'm not sure if the Blackboard API gives us any other option. I need to take a look at their API docs and see if there's any other way

lms/services/blackboard_api/_schemas.py
```diff
@@ -18,11 +18,15 @@ class Meta:
...
+        name = fields.Str(required=True)
+        modified = fields.Str(required=True)
+        downloadUrl = fields.Str(required=True)
 
     results = fields.List(fields.Nested(FileSchema), required=True)
 
     @post_load
     def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
+        for file_ in data["results"]:
+            file_["id"] = f"blackboard://content-resource/{file_['id']}/"
```
seanh, line 30: This is where we turn Blackboard file IDs into `blackboard://...` URLs before returning the list of them to the frontend when a teacher is creating an assignment.

They are just opaque strings to the frontend, it knows nothing about the `blackboard://...` format.

When the teacher selects a file the frontend submits the `blackboard://...` URL to the `configure_module_item()` view as the `document_url` and it gets saved in the DB. It's an opaque string to the `configure_module_item()` view and to the DB as well.

One weird thing about this is that `blackboard://...` is the file `id` in the list files response from our proxy API but it's the `document_url` in the `ModuleItemConfiguration` table :man_shrugging: 

The nice thing about doing this in the schema is that both the `list_files()` and `public_url()` methods use this schema. The `public_url()` method receives a `blackboard://...` URL as the `file_id`, gets the list of files from Blackboard (passing them through the schema so their IDs get turned into `blackboard://...` URLs), and compares the `blackboard://...` IDs of the list of files from the schema to the `blackboard://...` ID that it was given in its `file_id` argument. This means:

1. We only every format `blackboard://...` URLs. We never have to parse them
2. This one line in the schema is the only place in the code that knows about the `blackboard://...` format

The only exception to (2) is that `JSConfig` has to look for strings that start with `blackboard://` to know when a Blackboard Files assignment is being launched:

https://github.com/hypothesis/lms/blob/1a0324d8e0e8213bb0f27c46a2561e20206a17af/lms/resources/_js_config.py#L65-L73

These are the only two places in the code where the `"blackboard://"` string exists

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,38 @@ def list_files(self, course_id):
...
         ).validated_data
 
+        return [
+            {
+                "id": self._blackboard_document_url(file_["id"]),
+                "display_name": file_["name"],
+                "updated_at": file_["modified"],
+            }
+            for file_ in files
+        ]
```
seanh, line 69: The `list_files()` method is now doing the work of renaming the `name` and `modified` fields and of excluding the `downloadUrl` field from the `list_files()` response (where that field isn't needed). This is because the `public_url()` method below now uses the same `BlackboardListFilesSchema` and `public_url()` _doesn't_ want this renaming and _does_ want `downloadUrl`.

This does unfortunately mean that knowledge of the Blackboard API's list files response format is now in both the schema and the service (both the `list_files()` and `public_url()` methods) and in the tests for both :man_shrugging:

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,38 @@ def list_files(self, course_id):
...
-        return self._http_service.get(
+        files = self._http_service.get(
             self._api_url(f"courses/uuid:{course_id}/resources"),
             oauth=True,
             schema=BlackboardListFilesSchema,
         ).validated_data
 
+        return [
+            {
+                "id": self._blackboard_document_url(file_["id"]),
```
seanh, line 64: This is where we turn Blackboard file IDs into `blackboard://...` URLs before returning the list of them to the frontend when a teacher is creating an assignment.

They are just opaque strings to the frontend, it knows nothing about the `blackboard://...` format.

When the teacher selects a file the frontend submits the `blackboard://...` URL to the `configure_module_item()` view as the `document_url` and it gets saved in the DB. It's an opaque string to the `configure_module_item()` view and to the DB as well.

One weird thing about this is that `blackboard://...` is the file `id` in the list files response from our proxy API but it's the `document_url` in the `ModuleItemConfiguration` table :man_shrugging: 

Note that we only every format `blackboard://...` URLs. We never have to parse them :)

Also this `_blackboard_document_url()` methodis almost the only place in the code that knows about the `blackboard://...` format. The only exception to (2) is that `JSConfig` has to look for strings that start with `blackboard://` to know when a Blackboard Files assignment is being launched. These are the only two places in the code where the `"blackboard://"` string exists.

https://github.com/hypothesis/lms/blob/1a0324d8e0e8213bb0f27c46a2561e20206a17af/lms/resources/_js_config.py#L65-L73

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,38 @@ def list_files(self, course_id):
...
+    def public_url(self, course_id, file_id):
+        files = self._http_service.get(
+            self._api_url(f"/courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardListFilesSchema,
+        ).validated_data
+
+        for file_ in files:
+            if self._blackboard_document_url(file_["id"]) == file_id:
+                return file_["downloadUrl"]
```
seanh, line 80: To get a public URL we actually have to call the _list files_ API and then iterate over the list of file dicts looking for the one with the ID that we're looking for. The download URLs are in the file dicts. This sucks because on every student launch we might have to paginate through multiple responses to get the full list of files. But I'm not sure if the Blackboard API gives us any other option. I need to take a look at their API docs and see if there's any other way.

Note that we format the IDs from the API into `blackboard://...` URLs. Rather than parsing the `file_id` argument (which is already in `blackboard://...` URL format) into a plain ID. This means that we only ever need for format `blackboard://...` URLs, we never parse them. (Yes, we are wasting some CPU cycles :man_shrugging: )

lms/resources/_js_config.py
seanh: We might want to name the query param (or path component) `document_url` rather than `file_id` since it's a `blackboard://...` URL

lms/services/blackboard_api/service.py
seanh: Mentioned elsewhere but the `file_id` argument should maybe be renamed `document_url` since it's a `blackboard://...` URL

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,38 @@ def list_files(self, course_id):
...
+
+    def public_url(self, course_id, file_id):
+        files = self._http_service.get(
+            self._api_url(f"/courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardListFilesSchema,
+        ).validated_data
+
+        for file_ in files:
+            if self._blackboard_document_url(file_["id"]) == file_id:
```
marcospri, line 79: I know the ship has sailed on the whole `blackboard://...` but it seems like a lot of self inflicted complexity.

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +53,38 @@ def list_files(self, course_id):
...
+                return file_["downloadUrl"]
+
+        assert False
+
+    @staticmethod
+    def _blackboard_document_url(file_id):
+        return f"blackboard://content-resource/{file_id}/"
```
marcospri, line 86: I think this could have a docstring, I'd try to look for this schema on blackboard's docs first time seeing this.

Also `content-resource` seems a bit vague, I'll explicitly include `file` somewhere so they are more self explanation if seen without context.

lms/services/blackboard_api/service.py
seanh: We haven't actually released anything yet, so it's not too late to change our minds. Why do you think it's complex? What would you suggest instead?

marcospri: I think `blackboard://content-resource/{file_id}/` is more complex than just `file_id` I can really find any justification to it other than "the field is the DB is called document_url" and this `blackboard://` makes it, technically, an URL.

There's another PR open at the moment to persist canvas Files, I think we could reevaluate how we store this id once we integrate width that (I don't think there's an immediate need to store BB files but we probably should do it anyway mid-term)

lms/services/blackboard_api/service.py
seanh: Added a docstring

lms/services/blackboard_api/service.py
```diff
@@ -53,12 +54,46 @@ def list_files(self, course_id):
...
+            self._api_url(f"courses/uuid:{course_id}/resources"),
+            oauth=True,
+            schema=BlackboardListFilesSchema,
+        ).validated_data
+
+        for file_ in files:
+            if self._blackboard_document_url(file_["id"]) == document_url:
+                return file_["downloadUrl"]
+
+        raise BlackboardFileNotFoundInCourse(document_url)
```
seanh, line 83: To get a public URL we actually have to call the _list files_ API and then iterate over the list of file dicts looking for the one with the ID that we're looking for. The download URLs are in the file dicts. This sucks because on every student launch we might have to paginate through multiple responses to get the full list of files. But I'm not sure if the Blackboard API gives us any other option. I need to take a look at their API docs and see if there's any other way.

Note that we format the IDs from the API into `blackboard://...` URLs. Rather than parsing the `file_id` argument (which is already in `blackboard://...` URL format) into a plain ID. This means that we only ever need for format `blackboard://...` URLs, we never parse them. (Yes, we are wasting some CPU cycles :man_shrugging: )

lms/resources/_js_config.py
```diff
@@ -68,7 +68,7 @@ def add_document_url(self, document_url):
                 "path": self._request.route_path(
                     "blackboard_api.files.via_url",
                     course_id=self._request.params["context_id"],
-                    file_id=document_url.split("/")[3],
+                    _query={"document_url": document_url},
```
seanh, line 71: `JSConfig` just passes the `document_url` directly from the DB (so in the `blackboard://...` form) to the frontend as a query param on the Via URL API call. The frontend just treats this ID as an opaque string, so while `blackboard://...` URLs do flow through the frontend, the frontend code doesn't know anything about them.

lms/routes.py
```diff
@@ -42,8 +42,7 @@ def includeme(config):
         "blackboard_api.courses.files.list", "/api/blackboard/courses/{course_id}/files"
     )
     config.add_route(
-        "blackboard_api.files.via_url",
-        "/api/blackboard/courses/{course_id}/files/{file_id}/via_url",
+        "blackboard_api.files.via_url", "/api/blackboard/courses/{course_id}/via_url"
```
seanh, line 45: I moved the `{file_id}` path segment out of this URL. It's not actually a file ID (it's a `document_url` in `blackboard://` format) and it's not suitable for including in the path of a URL because it has `/`'s in it. Pyramid's `route_url()` actually URL encode path segments that you pass to it but it doesn't encode `/`'s. Going with the flow here, if you pass a URL to `route_url()` in a query param instead it does get fully encoded. And I think a value like this might be more suited in a query param

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,21 +8,37 @@
...
+        unknown = EXCLUDE
+
+    id = fields.Str(required=True)
+    display_name = fields.Str(data_key="name", required=True)
+    updated_at = fields.Str(data_key="modified", required=True)
+
+    @post_load
+    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
+        data["id"] = f"blackboard://content-resource/{data['id']}/"
+        return data
```
seanh, line 24: The path I ended up following in the end was to stick the the existing design as on master, which is that the schema does all the work of deserializing the data. So `_FileSchema` handles excluding unknown fields (like `downloadUrl`), renaming the `name` and `modified` fields, and even wrapping the `id` in a `blackboard://...` URL. And then `BlackboardListFilesSchema` also handles popping the `"results"` key. The actual service method is then just a one-liner

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,21 +8,37 @@
...
+class BlackboardPublicURLSchema(RequestsResponseSchema, _FileSchema):
+    """Schema for Blackboard /courses/{courseId}/resources/{resourceId} responses."""
+
+    downloadUrl = fields.Str(required=True)
 
     @post_load
-    def make_object(self, data, **_kwargs):  # pylint:disable=no-self-use
-        return data["results"]
+    def post_load(self, data, **_kwargs):
+        return data["downloadUrl"]
```
seanh, line 44: In line with the design of the existing `BlackboardListFilesSchema`, `BlackboardPublicURLSchema` does all the work and just returns the download URL string directly rather than returning a dict

lms/services/blackboard_api/service.py
```diff
@@ -59,6 +65,24 @@ def list_files(self, course_id):
             schema=BlackboardListFilesSchema,
         ).validated_data
 
+    def public_url(self, course_id, document_url):
+        # Get just the file_id part out of the
+        # blackboard://content-resource/file_id/ URL.
+        file_id = re.search(
+            r"blackboard:\/\/content-resource\/(?P<file_id>[^\/]*)\/", document_url
+        )["file_id"]
```
seanh, line 73: This can be done as `file_id = document_url.split("/")[3]`. I thought using a regex would be more readable as you can actually see the `blackboard://content-resource/` in the regex pattern. You can actually see what format is expected. The regex turns out to be pretty ugly when actually written out though.

It seems slightly unbalanced that it's the _schema_ that wraps the IDs from Blackboard in these URLs but it's the _service_ that unwraps them but :man_shrugging: This service + schemas are a pretty small amount of code next to each other in one directory. There are lots of ways to write them that all have pros and cons. Probably not worth too much fiddling

lms/services/blackboard_api/_schemas.py
marcospri: Is it no possible to *not* overwrite ['id'] and add this as a separate field (or keep the original id as a different field, `lms_id` or similar)

If I was using the service I'd expect to use the ID coming from one API method as the ID needed for another API call

seanh: The problem is that this is what happens:

1. Our list files proxy API returns a list of file dicts to our frontend: `[{"id": "blackboard://...", ...}, { ... }, { ... }]`
2. The user selects a file to use for the assignment
3. Our frontend submits the file's `blackboard://...` ID to our backend as the `document_url` field and it gets saved in the DB

The fact that the `document_url` submitted by the frontend starts with `blackboard://...` rather than `https://...` is the only way that the backend knows that it's a Blackboard file.

So we kind of do need to overwrite the ID.

We could do the overwrite in the service method rather than in the schema.

Or we could do the overwrite in the `list_files()` proxy API _view_ rather than in the service or schema.

But without changing the frontend the ID does need to be overwritten before we send it to the frontend.

lms/services/blackboard_api/_schemas.py
marcospri: > In line with the design of the existing `BlackboardListFilesSchema`, `BlackboardPublicURLSchema` does all the work and just returns the download URL string directly rather than returning a dict

I think `all the work` here is accessing one field explicit declared in the schema. If we needed to do "blackboard course copy" kind of hack, adding `size` here would mean changing all the users of the schema or alternatively using a different schema per use case of the same API endpoint which I don't think it's very flexible design.

I reckon the schema should be more coupled to the endpoint it talks to rather than  to the use case we are giving that endpoint at the moment, I think that work for the service.

lms/services/blackboard_api/service.py
```diff
@@ -59,6 +65,24 @@ def list_files(self, course_id):
             schema=BlackboardListFilesSchema,
         ).validated_data
 
+    def public_url(self, course_id, document_url):
+        # Get just the file_id part out of the
+        # blackboard://content-resource/file_id/ URL.
+        file_id = re.search(
```
marcospri, line 71: Should we compile the regexp, move it out of the method and use it here? Some perf benefit but less "code locality". Not strong preference from my side.

lms/services/blackboard_api/service.py
marcospri, line 73: I think the naming here is great, "document_url" vs "file_id" I think we should follow the same or simlar naming convention on `_FileSchema`

lms/services/blackboard_api/_schemas.py
marcospri: If this is massaging the data for the JS it might belong to the view but in the other hand you'll end up spilling the `blackboard://...` knowledge over more places, that's probably a worse outcome.

It's keeping the orignal ID On `lms_id` or similar feasible or marshmallow would make that difficult?

seanh: I actually think it's best for the schema to do all the work. The reason is that it means the schema is the only thing that knows anything about the Blackboard files API's response format, and the schema's tests are the only tests that need to know anything about it. So there's a [fake Blackboard API response in `_schemas_test.py`](https://github.com/hypothesis/lms/blob/9b021fa5832d906548438518abab998881047ce2/tests/unit/lms/services/blackboard_api/_schemas_test.py#L91-L110) and that's the only place in the tests where we need a fake Blackboard API response. The service and view just return what the schema returned directly, and their tests just need to assert that they returned whatever the schema returned.

If we split the responsibility of validating and mutating the Blackboard API responses between the schema, service and view, then knowledge about the API response format has to be duplicated in multiple places in the code and tests. The worse case would be if the schema just returned the Blackboard API response unmodified, and then the service made some changes to it (renaming and omitting fields), and then the view made some more changes (changing the ID's into URLs). In that case the schemas tests would need a fake Blackboard API response, the service's tests would also need a fake Blackboard API response, and the view's tests would need a fake Blackboard API response as it would look after being partially mutated by the service. And the tests would be like this for a reason: because the responsibility for validating and mutating those responses and the knowledge about the response format has actually been split between those three places. If the work is only split between two places (schema and service, or schema and view) you get a lesser version of the same problem.

That's why I think it's actually best for the schema to do "all the work" here: when you try it the code and tests actually work out simpler.

I don't think that will always be the case when you have a schema, service and view. But in this one particular case it happens to work out that way for now.

* * *

We could move the original ID to an `"lms_id"` item (`post_load()` is just a Python function, it can do whatever we want) but it wouldn't change much: the schema just returns these dicts to the service which directly returns them to the view which directly returns them to the frontend, and the frontend isn't expecting an `"lms_id"` field, it would just ignore it (and not send it back to the backend). So nothing would be able to use the `"lms_id"` field.

We could also _add_ a separate `"document_url"` field instead of changing the `"id"` field but again the frontend would ignore a `"document_url"` and just return the `"id"` (in a form field named `document_url`!) so it would break as the backend wouldn't be able to tell that it's a Blackboard file.

I think we might want a frontend change to allow us to include `document_url`'s in the file dicts instead of changing the `id` into a URL, because changing the `id` is just a bit weird. But I'll have to coordinate with the frontend team and see what's possible

lms/services/blackboard_api/_schemas.py
seanh: I think it's potentially true that a schema that's closer to the Blackboard API endpoint is more reusable in cases where we have multiple uses for the same API endpoint or where we're doing some more complicated logic with the responses. But as the Blackboard API client stands right now though, splitting the work between the schema and the service would just complicate the code and tests (discussed in more detail here: <https://github.com/hypothesis/lms/pull/2800#discussion_r652525397>). I think maybe we can be led by the code and change this responsibility split between schema and service in future if it makes the code and tests better for future needs?

lms/services/blackboard_api/_schemas.py
seanh: Started a Slack thread about potential frontend changes: https://hypothes-is.slack.com/archives/C1M8NH76X/p1623836733025000

lms/services/blackboard_api/service.py
seanh: Done

marcospri (approved):
Works nicely on the BB instance

Not 100% convinced about one schema hiding the original ID (well, reformatting it) and the other schema just returning one of its fields, coupling it completely to the service method but following the discussing here and on slack I could see a real possibility of revisiting this when we need to.

PR #2805, opened 2021-06-11
Check presence of `ignoreOtherConfiguration` in `hypothesisConfig`

esanzgar:
Part of https://github.com/hypothesis/viahtml/pull/192

First commit:
h_vialib was modified to include the above option to disregard other
configurations that could be present in the host page.

Although the `ignoreOtherConfiguration` is not set by the code in LMS,
given its importance we test it here too.

Second commit:
Upgrade `h-vialib` using `hdev requirements --package h-vialib` command.

### DIFF
--- a/requirements/bddtests.txt
+++ b/requirements/bddtests.txt
@@ -52,7 +52,7 @@ h-matchers==1.2.11
     # via -r requirements/bddtests.in
 h-pyramid-sentry==1.2.3
     # via -r requirements/requirements.txt
-h-vialib==1.0.15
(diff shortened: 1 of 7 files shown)


robertknight:
@esanzgar - The tests are failing here. I guess the `h_vialib` dependency will need to be updated first?

esanzgar:
Yes, I have just released a new version of h-vialib. After I updated
h-vialib, it should pass the test.

robertknight (approved):
LGTM

PR #2807, opened 2021-06-11
Separate file for service fixtures

jon-betts:
This imports the idea we've had in other places to:

 * Gather service mocks into one file
 * Use a meta fixture to simplify their creation
 
This makes it simpler to find and create service fixtures


tests/unit/services.py
```diff
@@ -0,0 +1,180 @@
...
+    "group_info_service",
+    "grouping_service",
+    "http_service",
+    "launch_verifier",
+    "lti_h_service",
+    "lti_outcomes_client",
+    "oauth1_service",
+    "oauth2_token_service",
+    "h_api",
+    "vitalsource_service",
```
jon-betts, line 45: We could do without this, but it means `conftest.py` ends up re-importing things like `factories` and `pytest` and using them from here and not importing it's own. Which is a bit weird.

tests/unit/services.py
```diff
@@ -0,0 +1,180 @@
...
+)
+
+
+@pytest.fixture
+def mock_service(pyramid_config):
+    def mock_service(service_class, service_name=None):
+        mock_service = mock.create_autospec(service_class, spec_set=True, instance=True)
+        if service_name:
+            pyramid_config.register_service(mock_service, name=service_name)
+        else:  # pragma: no cover
```
jon-betts, line 55: Not yet... but soooon

marcospri (approved):

PR #2809, opened 2021-06-11
Add `make coverage` back as a separate command

jon-betts:
This allows you to see the coverage report again without having to re-run the entire test suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,25 @@ checkformatting: python
 	@tox -qe checkformatting
 
 .PHONY: test
-test: backend-tests frontend-tests
+test: backend-tests coverage frontend-tests
+
+# Backend and frontend tests are split into separate targets because on Jenkins
+# we need to run them with different Docker images, but `make test` runs both.
+.PHONY: backend-tests
+backend-tests: python
+	@tox -qe tests
+
+.PHONY: coverage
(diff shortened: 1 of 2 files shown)


seanh (approved):
Unfortunately `make coverage` will print out the coverage report from the last time any tests were run, which might not have been a complete test run. For example:

```shellsession
$ make test  # Or `tox`
...
(prints out a complete coverage report)

$ make coverage
(prints out the same complete coverage report again)

$ tox tests/unit/lms/app_test.py  # Run a subset of the tests
...

$ make coverage
(complains that you have loads of missing coverage because it prints a coverage report for the entire codebase
but based on the test run of app_test.py only)
```

I was wondering if there was a way to fix that but I'm not sure that there is. The problem is that when you do `tox tests/unit/lms/app_test.py` it's running `coverage run -m pytest ...`. So we might have to make `tox.ini` smart about only prefixing the `pytest` command with `coverage run -m` if no `posargs` are given. Or make it smart about knowing that a `tox tests/unit/lms/app_test.py` should limit the coverage report to just `lms/app.py` and `tests/unit/lms/app_test.py`, but that seems a bit much.

jon-betts:
Such fanciness could be a good fit for `hdev`?

PR #2810, opened 2021-06-11
Split the non Canvas API responsibilities into a new service

jon-betts:
This makes the Canvas API service a pure API interface, and means it does not also have other canvas service responsibilities. This also extracts logic from the view and puts it into the service, thinning the view.

This is being done because:

 * It's probably the correct thing to do
 * The amount of logic in here is about to grow a lot

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -18,16 +18,15 @@ class CanvasService:
 
         :param file_id: The file to look up
         :param course_id: The course the file should be in
-        :param check_in_course: Enable pre-emptive checking for the file in the
-            course regardless of whether the user has permission to see it
+        :param check_in_course: Raise CanvasFileNotFoundInCourse if file_id isn't in
+            course_id
         :return: A URL suitable for public presentation of the file
-        :raises  CanvasFileNotFoundInCourse: If we determine the file should
+        :raise  CanvasFileNotFoundInCourse: if check_in_course=True and file_id isn't in course_id
             be in the course but isn't.
         """
 
         if check_in_course:
             if not self._file_in_course(file_id, course_id):
-                # Looks like a course copy
                 raise CanvasFileNotFoundInCourse(file_id)
 
         return self.api.public_url(file_id)
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,0 +1,42 @@
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CanvasService:
+    """A high level Canvas service."""
+
+    api = None
+
+    def __init__(self, canvas_api):
+        self.api = canvas_api
+
+    def public_url_for_file(self, file_id, course_id, check_in_course=False):
+        """
+        Get a public URL for a Canvas file.
+
+        This will also attempt to check if the file is in the course to detect
+        course copy situations and fix it if we can.
+
+        :param file_id: The file to look up
+        :param course_id: The course the file should be in
+        :param check_in_course: Enable pre-emptive checking for the file in the
+            course regardless of whether the user has permission to see it
+        :return: A URL suitable for public presentation of the file
+        :raises  CanvasFileNotFoundInCourse: If we determine the file should
+            be in the course but isn't.
+        """
+
+        if check_in_course:
+            if not self._file_in_course(file_id, course_id):
+                # Looks like a course copy
+                raise CanvasFileNotFoundInCourse(file_id)
+
+        return self.api.public_url(file_id)
+
+    def _file_in_course(self, file_id, course_id):
+        return any(
+            str(file_["id"]) == str(file_id) for file_ in self.api.list_files(course_id)
+        )
+
+
+def factory(_context, request):
+    return CanvasService(canvas_api=request.find_service(name="canvas_api_client"))
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -33,9 +33,9 @@ class FilesAPIViews:
         public_url = self.canvas.public_url_for_file(
             file_id=self.request.matchdict["file_id"],
             course_id=self.request.matchdict["course_id"],
-            # For teachers we can do a pre-emptive check to see if the file is
-            # in the course or not, to catch course copy scenarios even though
-            # the teacher has permission to see the file
+            # Teachers can have broad permissions and see files that aren't in
+            # the course. So do this slower check (extra API call) to warn the
+            # teacher that their students might not be able to see the file.
             check_in_course=self.request.lti_user.is_instructor,
         )
 
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -2,6 +2,7 @@
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
+from lms.services.canvas import CanvasService
 from lms.views import helpers
 
 
@@ -9,7 +10,7 @@ from lms.views import helpers
 class FilesAPIViews:
     def __init__(self, request):
         self.request = request
-        self.canvas_api_client = request.find_service(name="canvas_api_client")
+        self.canvas = request.find_service(CanvasService)
 
     @view_config(request_method="GET", route_name="canvas_api.courses.files.list")
     def list_files(self):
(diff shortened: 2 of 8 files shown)


tests/unit/lms/views/api/canvas/files_test.py
```diff
@@ -1,113 +1,40 @@
 import pytest
```
jon-betts, line 1: These tests were a bit weird and had quite a few tests to show things _didn't_ happen. Not sure about that as the list of things this code should _not_ do is infinite. In any case, they are much smaller and appropriate to the size of the view now.

lms/services/canvas.py
```diff
@@ -0,0 +1,42 @@
...
+        :raises  CanvasFileNotFoundInCourse: If we determine the file should
+            be in the course but isn't.
+        """
+
+        if check_in_course:
+            if not self._file_in_course(file_id, course_id):
+                # Looks like a course copy
+                raise CanvasFileNotFoundInCourse(file_id)
+
+        return self.api.public_url(file_id)
```
jon-betts, line 33: Not much here yet, but for course copy this is going to get big.

Also honestly, the `file_in_course` function probably was never appropriate to be on the Canvas API object in the first place. I think this is potentially a result of a misapplication of our rules:

 * The views must be thin
 * Therefore this function can't go in there
 * Therefore it should go in the Canvas API

I'd say a few counter points there are:

 * Our views don't need to be so thin they can't have a list comprehension in them
 * Even if the functionality can't go in the view it doesn't imply it should stick to the closest service either

tests/unit/lms/views/api/canvas/files_test.py
seanh: Hmm. I think `test_it_doesnt_catch_CanvasAPIErrors_from_list_files()` might have better been named `test_it_raises_CanvasAPIError_if_list_files_raises()`. It's not really a test for what `FilesAPIView` _doesn't_ do. It's a test for something it positively _does_ do. `FilesAPIView.list_files` should have a `:raise CanvasAPIError: if something goes wrong with the Canvas API request(s)` on it, and this test is effectively a test for that part of the docstring.

_But_ I think that putting `:raise`'s on docstrings is more something for services and other things that get called by our own code. The point is that if I'm writing a new line of code that calls `FooService.foo()` I can look at the docstring and see if there are any raises I need to be aware of. That applies less to views which we don't call ourselves.

I can see some value in it: if we have an exception view for a given exception, `:raise` docstrings could help with grepping for all views that might trigger that exception view. But I wouldn't rate our chances of successfully adding all the right `:raise`'s to all the right views. :man_shrugging:

seanh: I agree about the multiple assertions per unit test

seanh: I can see the point of `test_it_doesnt_check_whether_the_file_is_in_the_course()` too since we have specific reasons why we want to be sure that student launches don't trigger this check

lms/services/canvas.py
```diff
@@ -0,0 +1,42 @@
...
+
+        This will also attempt to check if the file is in the course to detect
+        course copy situations and fix it if we can.
+
+        :param file_id: The file to look up
+        :param course_id: The course the file should be in
+        :param check_in_course: Enable pre-emptive checking for the file in the
+            course regardless of whether the user has permission to see it
+        :return: A URL suitable for public presentation of the file
+        :raises  CanvasFileNotFoundInCourse: If we determine the file should
```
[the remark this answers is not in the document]
seanh: Are there more `:raise`'s to be added here? If `CanvasAPIClient` raises something that `CanvasService` doesn't catch?

tests/unit/lms/services/canvas_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_public_url_for_file_with_unsuccessful_file_check(self, canvas_service):
+        canvas_service.api.list_files.return_value = []
+
+        with pytest.raises(CanvasFileNotFoundInCourse):
+            canvas_service.public_url_for_file(
+                file_id=sentinel.file_id,
+                course_id=sentinel.course_id,
+                check_in_course=True,
+            )
```
seanh, line 29: I think you might want to add a test that `CanvasService` raises if `CanvasAPIClient.public_url()` or `list_files()` raises

seanh (approved):
We need to make sure to test that the "file not found in course" error messages for teachers and students are still working as intended. There's test courses in Canvas that can be used to test this

lms/services/canvas.py
```diff
@@ -0,0 +1,42 @@
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CanvasService:
+    """A high level Canvas service."""
+
+    api = None
```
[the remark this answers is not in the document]
jon-betts: Yep. 'tis a pain otherwise

lms/views/api/canvas/files.py
```diff
@@ -29,14 +30,13 @@ def via_url(self):
...
-
-        public_url = self.canvas_api_client.public_url(
-            self.request.matchdict["file_id"]
+        public_url = self.canvas.public_url_for_file(
+            file_id=self.request.matchdict["file_id"],
+            course_id=self.request.matchdict["course_id"],
+            # For teachers we can do a pre-emptive check to see if the file is
+            # in the course or not, to catch course copy scenarios even though
+            # the teacher has permission to see the file
+            check_in_course=self.request.lti_user.is_instructor,
```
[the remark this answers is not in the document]
jon-betts: I'll have a look over this to see if I can pack in more ideas but keep it short.

tests/unit/lms/services/canvas_test.py
jon-betts: Is there a specific scenario you had in mind? Presumably any code calling any other code and not catching will raise. I don't like testing the negative here (we don't have any code that...) as there's no real limit to it.

seanh: I think there's a finite number of particularly important exceptions that we raise and that get handled by exception views in particular ways, e.g. I think `CanvasAPIClient` raises `CanvasAPIError` or something. It's a pretty finite number: our call stack tends not to be very deep. A view calling a service that calls another service is about as deep as it usually gets.

On other hand perhaps the exception raise just needs to be documented on the one service method that directly raises the exception and not repeated on other service and view methods that call it, and that will be enough to assist developers in ensuring that the exception's always gonna be caught by an exception view that's gonna handle it appropriately

tests/unit/lms/views/api/canvas/files_test.py
jon-betts: I've parameterised the test to pass in both values. This ensures we don't always pass a fixed value, but there's no other behavior at all (we just pass it through) so I think that aspect is covered.

I disagree with the suggested names. This code does not raise any exceptions, so `test_it_raises_CanvasAPIError_if_list_files_raises` is inaccurate unless we were to catch and re-raise. For this reason `test_it_doesnt_catch_CanvasAPIErrors_from_list_files` is a more accurate name. Also phrasing this positively doesn't change the scenario under test, which is testing that we didn't implement any code to catch this. So this is testing a thing we didn't implement, rather than something we did.

The problem withthis is we also don't catch `ValueError`, or `OSError` or any thing form the helper function, or modify the user, or delete the account, or... There's kind of no end to it.

I'm similarly uncomfortable with documenting errors raised by code we call, yet don't raise directly ourselves for a number of reasons:

 * Is this even helpful? Like have you actually read this and found it to be helpful? I could imagine it would mostly make searching for a name more difficult
 * Where do you stop? One layer down, two, all the way until you hit C?
 * If something at the bottom of the stack raises a new error, do we really want coupled documentation that requires us to document that in every class above?
 * Do we want to repeatedly document the reasons for an exception further and further from the code in charge? Will these be accurate over time.
 
I have more sympathy for this in a few cases:

 * The errors are raised by private functions of code and exposed, but not raised directly, in a public interface
 * Service type boundaries as you mention

lms/views/api/canvas/files.py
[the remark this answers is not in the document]
jon-betts: I've rephrased this as:

```
            # Teachers can have broad permissions and see files that aren't in
            # the course. So do this slower check (extra API call) to warn the
            # teacher that their students might not be able to see the file.
```

PR #2811, opened 2021-06-11
Add a file table and update it when we retrieve a list of Canvas files (2)

jon-betts:
This is following on from: https://github.com/hypothesis/lms/pull/2777 where the conversation has gotten a bit too dense to follow.

This is based on https://github.com/hypothesis/lms/pull/2810 and also has https://github.com/hypothesis/lms/pull/2808 merged in. When https://github.com/hypothesis/lms/pull/2808 is merged this can be rebased to skip the first commit.

This can't be merged before: https://github.com/hypothesis/lms/pull/2824

Testing notes:

 * Run `make sql`
 * Run `select * from file` - You shouldn't see anything
 * Open a canvas course as a teacher
 * Run `select * from file` - You should see lots of records
 * Open a canvas course as a teacher
 * It shouldn't crash

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -1,4 +1,5 @@
 import logging
+from functools import cached_property
 
 import alembic.command
 import alembic.config
@@ -6,10 +7,12 @@ import sqlalchemy
 import zope.sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session, sessionmaker
 from sqlalchemy.orm.properties import ColumnProperty
 
-__all__ = ("BASE", "init")
+from lms.db._bulk_action import BulkAction
+
+__all__ = ("BASE", "init", "BulkAction")
 
 
 LOG = logging.getLogger(__name__)
@@ -85,9 +88,6 @@ BASE = declarative_base(
 )
 
 
-SESSION = sessionmaker()
-
-
 def init(engine, drop=False):
     """
     Create all the database tables if they don't already exist.
@@ -121,6 +121,17 @@ def _stamp_db(engine):  # pragma: nocover
 def make_engine(settings):
     """Construct a sqlalchemy engine from the passed ``settings``."""
     return sqlalchemy.create_engine(settings["sqlalchemy.url"])
+
+
+class CustomSession(Session):
+    @cached_property
+    def bulk(self):
+        """Get an object for performing bulk actions."""
+
+        return BulkAction(self)
+
+
+SESSION = sessionmaker(class_=CustomSession)
 
 
 def _session(request):  # pragma: no cover
--- a/lms/db/_bulk_action.py
+++ b/lms/db/_bulk_action.py
@@ -1,0 +1,88 @@
+"""
+Adds a set of low level DB operations for bulk operations.
+
+Why is this here, rather than services?
+
+ * The functionality is very low level and generic (we wish the basic Session
+ object had it)
+ * You shouldn't mock this in your code in general
+ * There's nothing LMS specific about it
+
+For more see: https://stackoverflow.com/c/hypothesis/questions/477
+"""
+
+from dataclasses import dataclass
+from typing import List
+
+from sqlalchemy.dialects.postgresql import insert
+from zope.sqlalchemy import mark_changed
+
+
+class BulkAction:
+    """
+    An extension to the DB session which adds bulk actions.
+
+    These can be accessed via request.db.bulk.<method_name>().
+
+    To enable for a particular model add config to the model like this:
+
+        class MyModel(BASE):
+            # Enable bulk actions
+            BULK_CONFIG = BulkAction.Config(
+                upsert_index_elements=[...],
+                upsert_update_elements=[...],
+            )
+    """
+
+    def __init__(self, session):
+        """
+        Initialize object.
+
+        :session: SQLAlchemy session object
+        """
+        self._session = session
+
+    @dataclass
+    class Config:
+        upsert_index_elements: List[str]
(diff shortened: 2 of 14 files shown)


tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+import sqlalchemy as sa
+from h_matchers import Any
+from sqlalchemy.engine import CursorResult
+
+from lms.db import BASE, BulkAction
+
+
+class TestBulkUpsertMixin:
```
jon-betts, line 9: These were stolen and adapted from Checkmate

tests/unit/lms/events/subscriber_test.py
```diff
@@ -0,0 +1,99 @@
+import pytest
+from h_matchers import Any
+
+from lms.events import FilesDiscoveredEvent
+from lms.events.subscriber import files_discovered
+from lms.models import File
+from tests import factories
+
+
+class TestFilesDiscovered:
```
jon-betts, line 10: It looks like we're making life a lot harder for ourselves than we have to here, as we could just mock out the bulk object and make assertions about that but...

* We don't mock basic DB functionality, which this is now acting as
* This is the only place that actually brings together the bulk upsert, File values etc and makes sure they all work together

That said it's still super long winded for a very short piece of code and I'd appreciate any suggestions on how to cut it down.

lms/services/canvas_api/client.py
```diff
@@ -225,6 +247,7 @@ class _ListFilesSchema(RequestsResponseSchema):
         display_name = fields.Str(required=True)
         id = fields.Integer(required=True)
         updated_at = fields.String(required=True)
```
jon-betts, line 249: This is the change to the schema to mean we get the size too. This forces a change in seemingly unrelated tests

lms/db/_session_extension.py
```diff
@@ -0,0 +1,29 @@
...
+    def __get__(self, instance, owner):
+        if not instance:
+            raise TypeError(
+                f"You can only call '{self.name}' on instances not the bare "
+                f"class: '{owner.__name__}'"
+            )
+
+        # Store the session and return ourselves
+        self.session = instance
+        return self
```
jon-betts, line 29: This indirection layer is here 100% to facilitate having the bulk functionality bolt onto the session object rather than be something directly provided by the session object.

If we don't care about that and are happy to put the bulk action functionality in the session directly there's no point in this.

Another way to achieve the same thing would be:

```python
class BulkAction:
    def __init__(self, session):
        self.session = session

class CustomSession(Session):    
    @cached_property
    def bulk(self):
        return BulkAction(self)
```

Which achieves much the same thing but requires more co-operation from the session object.

jon-betts:
This has been updated now with:

 * Some more and different comments + doc strings
 * Swapped the descriptor for a simpler cached property
 
 Also added a s/o decision record relating to this: https://stackoverflow.com/c/hypothesis/questions/477

lms/db/__init__.py
```diff
@@ -123,6 +123,17 @@ def make_engine(settings):
     return sqlalchemy.create_engine(settings["sqlalchemy.url"])
 
 
+class CustomSession(Session):
+    @cached_property
```
marcospri, line 127: 3.8! yeah

marcospri (approved):
LGTM, what's the merging plan? There's need to be migration PR merged before the model.

Now that we've seen everything working together and if you don't mind some more git chores it might make sense to open a PR with the subscriber & _bulk_action and get that merged.

lms/db/_bulk_action.py
```diff
@@ -0,0 +1,106 @@
...
+
+Why is this here, rather than services?
+
+ * The functionality is very low level and generic (we wish the basic Session
+ object had it)
+ * You shouldn't mock this in your code in general
+ * There's nothing LMS specific about it
+
+For more see: https://stackoverflow.com/c/hypothesis/questions/477
+"""
```
seanh, line 12: :+1:

jon-betts:
Yeah I need to do a migration. I might not split this up, not because I don't want to do the work, but because it doesn't actually do anything without the file table part. So lets say I'm killed my an errant pigeon tomorrow and we never get round to it, it's sort of dead code.

lms/db/__init__.py
jon-betts: We have our next DO right here. Remove all instances of `@lru_cache + @property`

lms/services/canvas_api/client.py
```diff
@@ -210,13 +213,32 @@ def list_files(self, course_id):
...
+                        "type": "canvas_file",
+                        "course_id": course_id,
+                        "lms_id": file["id"],
+                        "name": file["display_name"],
+                        "size": file["size"],
+                    }
+                    for file in files
+                ],
+            )
+        )
```
seanh, line 238: To be honest I kind of think that this should just to `self._request.db.bulk.upsert(File, values=[...])` and skip out the whole event/subscriber thing. I'm not sure the indirection is adding much

seanh (approved):
General approach looks good. I added a couple of comments but nothing that should radically change things. Let's get this split up into separate PRs for detailed review. It looks like there's a DB migration, model class, the generic bulk upsert thing, then finally changing the service to actually upsert the files

lms/db/_bulk_action.py
```diff
@@ -0,0 +1,106 @@
...
+        :rtype: Config
+        :raises NotImplementedError: If the config is missing
+        """
+        try:
+            return model_class.BULK_CONFIG
+        except AttributeError as err:
+            raise NotImplementedError(
+                f"Model class '{model_class}' does not have the expected config "
+                "attribute: 'BULK_CONFIG'"
+            ) from err
```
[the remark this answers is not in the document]
jon-betts: I'll scoop it out

PR #2824, opened 2021-06-15
Store files migration

jon-betts:
Adds the file table for: https://github.com/hypothesis/lms/pull/2811

## Testing notes

```
docker stop lms_postgres_1; docker rm lms_postgres_1; make services
hdev alembic upgrade head
make sql
\d+ files;
```
You should see the file table with:

```
         Column          |            Type             | Collation | Nullable |             Default              
-------------------------+-----------------------------+-----------+----------+----------------------------------
 created                 | timestamp without time zone |           | not null | now()
 updated                 | timestamp without time zone |           | not null | now()
 id                      | integer                     |           | not null | nextval('file_id_seq'::regclass)
 application_instance_id | integer                     |           | not null | 
 type                    | character varying           |           | not null | 
 lms_id                  | character varying           |           | not null | 
 course_id               | character varying           |           |          | 
 name                    | character varying           |           |          | 
 size                    | integer                     |           |          | 
Indexes:
    "pk__file" PRIMARY KEY, btree (id)
    "uq__file__application_instance_id" UNIQUE CONSTRAINT, btree (application_instance_id, lms_id, type, course_id)
Foreign-key constraints:
    "fk__file__application_instance_id__application_instances" FOREIGN KEY (application_instance_id) REFERENCES application_instances(id) ON DELETE CASCADE
```

Then downgrade and update:

* `hdev alembic downgrade -1`
* `make sql; \d+ files` - The table should be gone
* `hdev alembic upgrade +1`
* `make sql; \d+ files` - The table should be back

### DIFF
--- a/lms/migrations/versions/47d738f87eb5_add_file_table.py
+++ b/lms/migrations/versions/47d738f87eb5_add_file_table.py
@@ -1,0 +1,51 @@
+"""
+Add file table.
+
+Revision ID: 47d738f87eb5
+Revises: bb11b5b06274
+Create Date: 2021-06-15 13:44:45.368126
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "47d738f87eb5"
+down_revision = "bb11b5b06274"
+
+
+def upgrade():
+    op.create_table(
+        "file",
+        sa.Column(
+            "created", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column(
+            "updated", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("application_instance_id", sa.Integer(), nullable=False),
+        sa.Column("type", sa.UnicodeText(), nullable=False),
+        sa.Column("lms_id", sa.UnicodeText(), nullable=False),
+        sa.Column("course_id", sa.UnicodeText(), nullable=True),
+        sa.Column("name", sa.UnicodeText(), nullable=True),
+        sa.Column("size", sa.Integer(), nullable=True),
+        sa.ForeignKeyConstraint(
+            ["application_instance_id"],
+            ["application_instances.id"],
+            name=op.f("fk__file__application_instance_id__application_instances"),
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/e69454a18f38_add_file_table.py
```diff
@@ -0,0 +1,50 @@
...
+        sa.Column(
+            "application_instance_id",
+            sa.Integer(),
+            sa.ForeignKey("application_instances.id", ondelete="cascade"),
+            nullable=False,
+        ),
+        sa.Column("type", sa.String(), nullable=False),
+        sa.Column("lms_id", sa.String(), nullable=False),
+        sa.Column("course_id", sa.String()),
+        sa.Column("name", sa.String()),
```
marcospri, line 43: We've been using `sa.Unicode()` or `sa.UnicodeText()`expliclty in other PRs

lms/migrations/versions/e69454a18f38_add_file_table.py
```diff
@@ -0,0 +1,50 @@
...
+        sa.Column("size", sa.Integer()),
+        sa.UniqueConstraint("application_instance_id", "lms_id", "type", "course_id"),
```
marcospri, line 45: I don't have strong opinions about this but if you have handwriting this migration you might have bypassed the naming conventions thingy:

https://github.com/hypothesis/lms/blob/2474c2f61ad58cbe2ead2f90fc11f8886c4eb0cd/lms/db/__init__.py#L76

that would explicitly add a `name=` to the FK and constraint.

lms/migrations/versions/e69454a18f38_add_file_table.py
jon-betts: Does this require the model to be updated to match?

lms/migrations/versions/e69454a18f38_add_file_table.py
jon-betts: This seems to result in the same table, what's the difference?

lms/migrations/versions/e69454a18f38_add_file_table.py
marcospri: Yes, the column has to have the same `sa.Unicode()` type on the model.

marcospri (approved):

PR #2827, opened 2021-06-15
Move ApplicationInstance creation to the service

marcospri:
Looking for where the initial AI.settings were set and found this which I think belongs to the service rather than the view 

## Testing

Behavior shouldn't have changed on http://localhost:8001/welcome

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -22,22 +22,16 @@ class ApplicationInstanceService:
         """
         consumer_key = consumer_key or self._default_consumer_key
 
-        application_instance = self.get_by_consumer_key(consumer_key)
+        application_instance = (
+            self._db.query(ApplicationInstance)
+            .filter_by(consumer_key=consumer_key)
+            .one_or_none()
+        )
 
         if application_instance is None:
             raise ConsumerKeyError()
 
         return application_instance
-
-    def get_by_consumer_key(self, consumer_key):
-        """Return the ApplicationInstance with the given consumer_key or None."""
-
-        # TODO - Inline into get?
-        return (
-            self._db.query(ApplicationInstance)
-            .filter_by(consumer_key=consumer_key)
-            .one_or_none()
-        )
 
     def create(  # pylint:disable=too-many-arguments
         self, lms_url, email, developer_key, developer_secret, aes_secret
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -5,9 +5,9 @@ from lms.services import ConsumerKeyError
 
 
 class ApplicationInstanceService:
-    def __init__(self, db, default_consumer_key):
+    def __init__(self, db, request):
         self._db = db
-        self._default_consumer_key = default_consumer_key
+        self._request = request
 
     @lru_cache
     def get(self, consumer_key=None) -> ApplicationInstance:
@@ -20,7 +20,12 @@ class ApplicationInstanceService:
 
         :raise ConsumerKeyError: if the consumer_key isn't in the database
         """
-        consumer_key = consumer_key or self._default_consumer_key
+        if not consumer_key:
+            consumer_key = (
+                self._request.lti_user.oauth_consumer_key
+                if self._request.lti_user
+                else None
+            )
 
         application_instance = (
             self._db.query(ApplicationInstance)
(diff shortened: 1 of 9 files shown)


lms/services/application_instance.py
```diff
@@ -31,6 +31,27 @@ def get(self, consumer_key=None) -> ApplicationInstance:
 
         return application_instance
 
+    def create(  # pylint:disable=too-many-arguments
```
marcospri, line 34: Not sure if the `from_url` part is need, it just puts the URL value on the URL column

lms/services/application_instance.py
```diff
@@ -31,6 +31,27 @@ def get(self, consumer_key=None) -> ApplicationInstance:
...
+        self, lms_url, email, developer_key, developer_secret, aes_secret
+    ):
+        # If either one of developer_key or developer_secret is missing, then we
+        # don't save the other one either.
+        if not developer_key or not developer_secret:
+            developer_key = None
+            developer_secret = None
+
+        ai = ApplicationInstance.build_from_lms_url(
```
marcospri, line 43: This could be on the service or stay on the model. I left it on the model which I think it makes sense.

lms/services/application_instance.py
```diff
@@ -31,6 +31,27 @@ def get(self, consumer_key=None) -> ApplicationInstance:
...
+            lms_url,
+            email,
+            developer_key,
+            developer_secret,
+            aes_secret,
+            settings={"canvas": {"sections_enabled": bool(developer_key)}},
```
marcospri, line 49: All AI are created thru that view at the moment so I think it's ok to keep this hard coded here.

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -31,6 +33,80 @@ def test_get_raises_ConsumerKeyError_if_consumer_key_is_None(self, pyramid_reque
...
+
+        ai = db_session.query(ApplicationInstance).get(ai.id)
+        assert (
+            bool(ai.settings.get("canvas", "sections_enabled"))
+            == canvas_sections_enabled
+        )
+
     @pytest.fixture
     def svc(self, pyramid_request):
         return factory(mock.sentinel.context, pyramid_request)
```
jon-betts, line 112: I don't think we should be creating the service via the factory in the tests. The factory depends on the service, but the service does not depend on the factory, and it makes seeing the linkage more obscure

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -31,6 +33,80 @@ def test_get_raises_ConsumerKeyError_if_consumer_key_is_None(self, pyramid_reque
...
+        new_application_instance_params["developer_key"] = developer_key
+
+        ai = svc.create(**new_application_instance_params)
+        db_session.flush()
+
+        ai = db_session.query(ApplicationInstance).get(ai.id)
+        assert (
+            bool(ai.settings.get("canvas", "sections_enabled"))
+            == canvas_sections_enabled
+        )
```
jon-betts, line 108: There's really only 4 or so lines of code in the function at the moment, so it's probably worth thinking about how we can test it without quite as much in the tests.

Some suggestions:

 * Parameterise additions to `new_application_instance_params`
 * Pack some more assertions about outcomes into fewer tests
 * Package the call, flush, retrieve into a caller?

jon-betts (changes requested):
I think this removes some code out of the view and pushes it down, but stops short of coalescing all of the behavior into one place. This results in the code for creating an application instance actually being more spread out than before.

This also results in hybrid tests in the service which don't benefit from isolating the layer below and end up quite long as a result, but also don't completely test `build_from_lms_url`.

As far as I can tell `build_from_lms_url` is only called here, so we end up having to test more different paths than our code can actually create, because we have a more general interface to `build_from_lms_url` than it's possible to exercise. I think unifying the two and testing in one place would reduce the number of tests and possible code paths.

I'd suggest:

* Moving the `if not developer_key...` part to the other code (it's basically the other side of an if that's already there) + the settings initialisation
* Either:
    * Move everything about creating into the service
    * Don't have anything there at all (if it's just calling that then putting it in the DB)

This might be a good idea to have a look at the `if` block and private functions in `build_from_lms_url` too (plus the name) as they are needlessly confusing.

marcospri:
Found a service circular dependency while merging some service and model methods. Might clean this up to only do the bits about `create` and left the rest for a different PR.

PR #2829, opened 2021-06-16
Grouping.extra it's not ApplicationSettings

marcospri:
It's also always assigned a full dict so no need for mutability helpers

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -1,5 +1,6 @@
 import sqlalchemy as sa
 from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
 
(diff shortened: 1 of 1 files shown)


seanh (approved):
We might want to use SQLAlchemy's `MutableDict.as_mutable()` thing anyway, there seems no harm in having mutation detection? (Although it doesn't fully work anyway. IIRC it doesn't detect mutations to sub-dicts or sub-lists)

marcospri:
> We might want to use SQLAlchemy's `MutableDict.as_mutable()` thing anyway, there seems no harm in having mutation detection? (Although it doesn't fully work anyway. IIRC it doesn't detect mutations to sub-dicts or sub-lists)

You are right, not harm on having it there :+1:

PR #2831, opened 2021-06-16
Add stub API views for fetching VitalSource book metadata

robertknight:
Replace the API stubs for listing books and fetching book chapters in the LMS frontend with API views in `lms.views.api.vitalsource` in the backend. These views currently return the same hard-coded data that the stubs previously returned, but the sample data is now stored inside a Python module (`lms.views.api.vitalsource_sample_data`).

In future these views will actually call the VitalSource API to get the data.

This is part of https://github.com/hypothesis/product-backlog/issues/1200.

**Changes in detail:**

- Add `/api/vitalsource/books/{book_id}` and `/api/vitalsource/books/{book_id}/toc` routes, implemented in `lms.views.api.vitalsource`
- Convert the VitalSource sample data from a JS module to a Python module in `lms.views.api.vitalsource_sample_data`
- Modify the LMS frontend to invoke the `/api/vitalsource/*` APIs instead of returning stub responses (see `services/vitalsource.js`)
- Change tests for `BookPicker` and `BookSelector` components to use different dummy responses for book and chapter data. Previously they used the hardcoded sample data in the `vitalsource-sample-data.js` module

**Testing:**

Enable the "vitalsource" option at http://localhost:8001/flags and select the VitalSource option in the LMS assignment configuration dialog. There should be no functional changes here except the elimination of a 500ms delay that was previously hardcoded into the book and chapter fetching responses.


lms/views/api/vitalsource.py
```diff
@@ -0,0 +1,27 @@
...
+        return book_list
+
+    @view_config(route_name="vitalsource_api.books.toc")
+    def table_of_contents(self):
+        book_id = self.request.matchdict["book_id"]
+
+        # In future this will fetch the data from VitalSource using the
+        # https://api/vitalsource.com/v4/products/{BOOK_ID}/toc endpoint.
+        if book_id not in toc_data:
+            raise HTTPNotFound("Table of contents not available for book")
```
seanh, line 26: How does the client handle this 404 error? What error dialog is displayed? In general I think `404`'s aren't expected from the API and `400` with a JSON body should be used. See https://github.com/hypothesis/lms/blob/b08d57b2d76d7a76e88afab560e1d848242357c5/lms/views/api/exceptions.py#L25-L91

seanh (approved):

robertknight:
I have updated this PR as follows:

- Rebased on master and adapted to a change in the book selection workflow and frontend code structure. Instead of providing an API to list books available in the current course, there is an API to fetch metadata for a specific book
- When a book is not found a subclass of the existing `ProxyAPIError` is raised by the views. This is then translated to an API error and handled by the frontend using the existing conventions.

In the frontend an invalid book ID will be reported like this:

Invalid book ID

Whereas a valid book ID is handled like this:

Valid book ID

lms/views/api/vitalsource.py
```diff
@@ -0,0 +1,37 @@
+from pyramid.view import view_config, view_defaults
+
+from lms.security import Permissions
+from lms.services.exceptions import ProxyAPIError
+from lms.views.api.vitalsource_sample_data import book_data, toc_data
+
+
+class BookNotFoundError(ProxyAPIError):
+    def __init__(self, book_id):
+        super().__init__(f"Book {book_id} not found")
```
seanh, line 10: This should be in `views/api/exceptions.py` (which is where we keep exception classes). Alternatively you can probably get away with just raising `ProxyAPIError` directly without subclassing it

seanh (approved):
Let's get this merged

lms/views/api/vitalsource.py
robertknight: > This should be in views/api/exceptions.py (which is where we keep exception classes)

Ah, OK. `exceptions.py` defines exception views but all the exception subclasses that I can see in it are re-exported from elsewhere, typically a relevant service.

I have implemented your second suggestion of just using `ProxyAPIError` directly for now, since there is no added functionality in the custom subclass and nothing relies on its instance type being different than `ProxyAPIError`.

PR #2835, opened 2021-06-16
Migration for assigment extra column

jon-betts:
A DB migration adding a JSONB extras column for the assignment for: https://github.com/hypothesis/lms/pull/2783

### DIFF
--- a/lms/migrations/versions/bf67a1e68cea_add_assignment_extras.py
+++ b/lms/migrations/versions/bf67a1e68cea_add_assignment_extras.py
@@ -1,0 +1,31 @@
+"""
+Add assignment extras column.
+
+Revision ID: bf67a1e68cea
+Revises: 47d738f87eb5
+Create Date: 2021-06-16 12:19:17.417022
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "bf67a1e68cea"
+down_revision = "47d738f87eb5"
+
+
+def upgrade():
+    op.add_column(
+        "module_item_configurations",
+        sa.Column(
+            "extra",
+            postgresql.JSONB(astext_type=sa.Text()),
+            server_default=sa.text("'{}'::jsonb"),
+            nullable=False,
+        ),
+    )
+
+
+def downgrade():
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/bf67a1e68cea_add_assignment_extras.py
```diff
@@ -0,0 +1,31 @@
...
+revision = "bf67a1e68cea"
+down_revision = "47d738f87eb5"
+
+
+def upgrade():
+    op.add_column(
+        "module_item_configurations",
+        sa.Column(
+            "extra",
+            postgresql.JSONB(astext_type=sa.Text()),
```
seanh, line 23: Does this produce the same SQL as `MutableDict.as_mutable(JSONB)`? We seem to have started using both `MutableDict.as_mutable(JSONB)` and `postgresql.JSONB(astext_type=sa.Text())` in migrations (e.g. <https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/migrations/versions/37710e6bcb66_add_group_info_info_column.py#L21-L22>, <https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/migrations/versions/f0859cd029fe_add_application_instances_canvas_.py#L20>, <https://github.com/hypothesis/lms/blob/28f158a9cb5edfaca254848581ed35edfbf5d242/lms/migrations/versions/3517eb6254e7_add_course_and_course_groups_exported_from_h_tables.py#L28>, but then <https://github.com/hypothesis/lms/blob/b5a4f40da7d9cc2cd187536f1ae5560da153d08d/lms/migrations/versions/bb11b5b06274_grouping_table.py#L36>

seanh (approved):

lms/migrations/versions/bf67a1e68cea_add_assignment_extras.py
jon-betts: This is the result of the auto migration process, so I'd guess so, but haven't checked yet.

I'd imagine the difference is one has lots of Python semantics that are irrelevant to Posgres.

jon-betts: Yep this results in the same SQL.

Issue #2834, opened 2021-06-16
Handle errors when a Blackboard user's access token is missing or doesn't work


seanh:
If we don't have an access token for the user or if the access token doesn't work (either way: we need to get a new access token) the app is currently showing a **Something went wrong** error dialog. This should be an **Authorize Hypothesis** non-error dialog.

PR #2837, opened 2021-06-16
Fix Blackboard access token errors

seanh:
Part of https://github.com/hypothesis/lms/issues/2834 (but doesn't entirely fix it yet).

How this works for Canvas on master
-----------------------------------

`AuthenticatedClient` calls `OAuth2TokenService` to get an access token
for the user. If we don't have an access token for the user
`OAuth2TokenService` will raise `NoOAuth2Token`. `AuthenticatedClient`
catches `NoOAuth2Token` and raises `ProxyAPIAccessTokenError`:

https://github.com/hypothesis/lms/blob/be46e8c356c0da97d86c53cedab28d4b31570cc4/lms/services/canvas_api/_authenticated.py#L56-L62

An exception view catches `ProxyAPIAccessTokenError` and turns it into a
specific error response. This error response causes the frontend to show
the **Authorize Hypothesis** dialog (which does not look like an error
dialog since this is an expected situation, nothing has gone wrong):

https://github.com/hypothesis/lms/blob/5f958d7d4f8645a9d81c1afb23edb101d9c1fcb9/lms/views/api/exceptions.py#L110-L112

What's going wrong for Blackboard on master

`BlackboardAPIClient` calls `HTTPService` in a couple of places. For
example when launching a Blackboard files assignment:

https://github.com/hypothesis/lms/blob/5f958d7d4f8645a9d81c1afb23edb101d9c1fcb9/lms/services/blackboard_api/service.py#L67-L72

`HTTPService` calls `OAuth2TokenService` to get an access token. This
will raise `NoOAuth2Token` if we don't have an access token for the
user, just as it does when `CanvasAPIClient`'s `AuthenticatedClient`
calls `OAuth2TokenService`:

https://github.com/hypothesis/lms/blob/5f958d7d4f8645a9d81c1afb23edb101d9c1fcb9/lms/services/http.py#L111-L115

Neither `HTTPService` nor `BlackboardAPIClient` nor the view catches the
`NoOAuth2Token` exception so it gets raised back up to Pyramid. There is
no exception view for `NoOAuth2Token` so Pyramid returns a 500 Server
Error which the frontend turns into a **Something went wrong** error
dialog.

Solution
--------

`BlackboardAPIClient` could catch `NoOAuth2Token` and raise
`ProxyAPIAccessTokenError` like `CanvasAPIClient`'s
`AuthenticatedClient` does.

But instead I went for something simpler:

Why use two exception classes when you can use one?

Make the exception view catch `NoOAuth2Token` instead of
`ProxyAPIAccessTokenError`. Delete `ProxyAPIAccessTokenError`.
`CanvasAPIClient` and `BlackboardAPIClient` can just call
`OAuth2TokenService` (directly or indirectly) and lets the
`NoOAuth2Token` error fly up to Pyramid which will direct it to the
`NoOAuth2Token` error exception view.

`NoOAuth2Token` renamed to `OAuth2TokenError`
---------------------------------------------

`ProxyAPIAccessTokenError` was actually used for two purposes:

1. When we don't have an access token for the user.

   This is when `OAuth2TokenService` raises `NoOAuth2Token` and
   `AuthenticatedClient` catches it and raises
   `ProxyAPIAccessTokenError`.

2. When we do have an access token but it doesn't work (e.g. because its
   expired or been revoked).

   `CanvasAPIError.raise_from()` detects access token-related error
   responses from Canvas (`"Invalid access token"`,
   `"refresh_token not found"` or `"Insufficient scopes on access token"`)
   and raises `ProxyAPIAccessTokenError`:

   https://github.com/hypothesis/lms/blob/5f958d7d4f8645a9d81c1afb23edb101d9c1fcb9/lms/services/exceptions.py#L164-L174

`NoOAuth2Token` is now used in both of those cases instead so I've
renamed it to the more general `OAuth2TokenError`.

This works because our behaviour when our OAuth 2 token has expired or
been revoked is the same as our behaviour when we don't have a token: we
tell the frontend to ask the user for a new token.

Testing
-------

Delete all the access tokens from your DB:

Now launch a Canvas assignment and launch a Blackboard files assignment.
In both cases you should get an **Authorize Hypothesis** dialog and then
the assignment should launch.

Break all the access tokens in your DB:

Now launch a Canvas assignment and it should just work: it'll use the
refresh token to get a new access token without user interaction.

**Known issue:** In Blackboard this will show an error because we don't
yet have code for detecting access token-related error responses from
the Blackboard API or for using refresh tokens with Blackboard. But the
**Try again** button on the error dialog will work: it'll ask the user
for a new access token and then launch the assignment successfully.

Break all the access tokens and refresh tokens in your DB:

    tox -qe dockercompose -- exec postgres psql -U postgres -c "UPDATE oauth2_token SET access_token = 'foo', refresh_token = 'bar';"

Now launch a Canvas assignment and you should get the
**Authorize Hypothesis** dialog (no error) and the assignment should
launch successfully.

**Known issue:** In Blackboard this will show an error dialog again, but
again the **Try again** button will work.

### DIFF
--- a/lms/sentry.py
+++ b/lms/sentry.py
@@ -1,16 +1,16 @@
 """Sentry crash reporting integration."""
-from lms.services import ProxyAPIAccessTokenError
+from lms.services import OAuth2TokenError
 
 
-def filter_proxy_api_access_token_error(event):
-    """Filter out all ProxyAPIAccessTokenError's."""
-    return isinstance(event.exception, ProxyAPIAccessTokenError)
+def filter_oauth2_token_error(event):
+    """Filter out all OAuth2TokenError exceptions."""
+    return isinstance(event.exception, OAuth2TokenError)
 
 
 def includeme(config):
     config.add_settings(
         {
-            "h_pyramid_sentry.filters": [filter_proxy_api_access_token_error],
+            "h_pyramid_sentry.filters": [filter_oauth2_token_error],
             "h_pyramid_sentry.retry_support": True,
             "h_pyramid_sentry.sqlalchemy_support": True,
         }
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -13,8 +13,7 @@ from lms.services.exceptions import (
     LTILaunchVerificationError,
     LTIOAuthError,
     LTIOutcomesAPIError,
-    NoOAuth2Token,
-    ProxyAPIAccessTokenError,
+    OAuth2TokenError,
     ProxyAPIError,
     ServiceError,
 )
--- a/lms/services/blackboard_api/service.py
+++ b/lms/services/blackboard_api/service.py
@@ -59,12 +59,7 @@ class BlackboardAPIClient:
         )
 
     def list_files(self, course_id):
-        """
-        Return the list of files in the given course.
-
-        :raise ProxyAPIAccessTokenError: if we don't have a Blackboard API
-            access token for the current user
-        """
+        """Return the list of files in the given course."""
         return self._http_service.get(
             self._api_url(f"courses/uuid:{course_id}/resources"),
             oauth=True,
--- a/lms/services/canvas_api/_authenticated.py
+++ b/lms/services/canvas_api/_authenticated.py
@@ -1,6 +1,6 @@
 """Access to the authenticated parts of the Canvas API."""
 
-from lms.services import NoOAuth2Token, ProxyAPIAccessTokenError
+from lms.services import OAuth2TokenError
 from lms.validation.authentication import OAuthTokenResponseSchema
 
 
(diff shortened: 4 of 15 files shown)


lms/sentry.py
```diff
@@ -1,16 +1,16 @@
 """Sentry crash reporting integration."""
-from lms.services import ProxyAPIAccessTokenError
+from lms.services import OAuth2TokenError
 
 
-def filter_proxy_api_access_token_error(event):
-    """Filter out all ProxyAPIAccessTokenError's."""
-    return isinstance(event.exception, ProxyAPIAccessTokenError)
+def filter_oauth2_token_error(event):
```
marcospri, line 5: This hasn't change on the PR but the name and docstring made me pause for a second, no filtering going on here but rather at h_pyramid_sentry.filters

marcospri (approved):
Works and removes a little bit of complexity :+1:

PR #2838, opened 2021-06-17
Use sa.Unicode over sa.String in DB models

marcospri:
Something mentioned the other day, I think there's some value on being able to copy paste something as simple as "a column with a string" without having to remember String vs Unicode on every PR.

---

Main benefit it's consistency across models.
Behaviour wise main difference is requiring an explicit decoding while
using bytes.

https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Unicode

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -20,12 +20,12 @@ class ApplicationInstance(BASE):
     __tablename__ = "application_instances"
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
-    consumer_key = sa.Column(sa.String, unique=True, nullable=False)
-    shared_secret = sa.Column(sa.String, nullable=False)
-    lms_url = sa.Column(sa.String(2048), nullable=False)
-    requesters_email = sa.Column(sa.String(2048), nullable=False)
+    consumer_key = sa.Column(sa.Unicode, unique=True, nullable=False)
+    shared_secret = sa.Column(sa.Unicode, nullable=False)
+    lms_url = sa.Column(sa.Unicode(2048), nullable=False)
+    requesters_email = sa.Column(sa.Unicode(2048), nullable=False)
     created = sa.Column(sa.TIMESTAMP, default=datetime.utcnow(), nullable=False)
-    developer_key = sa.Column(sa.String)
+    developer_key = sa.Column(sa.Unicode)
     developer_secret = sa.Column(sa.LargeBinary)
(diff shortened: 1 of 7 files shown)


seanh:
We use `UnicodeText` not `Unicode` don't we?

Will this make any difference to the SQL schema that's generated when a new DB is initialized from the models?

marcospri:
> We use `UnicodeText` not `Unicode` don't we?
> 
> Will this make any difference to the SQL schema that's generated when a new DB is initialized from the models?

String & Unicode produce the same SQL type (character without a limit) Unicode & UnicodeText both produce TEXT

```
The TEXT data type can store a string with unlimited length.

If you do not specify the n integer for the VARCHAR data type, it behaves like the TEXT datatype. 
The performance of the VARCHAR (without the size n) and TEXT are the same.
```

No much difference but using UnicodeText where we now use String would need a DB migration to change the column type.

seanh:
So `Text` and `UnicodeText` produce the Postgres type `TEXT`. What type do `String` and `Unicode` produce? `VARCHAR`?

Also, what is the difference between `String` vs `Unicode`? And `Text` vs `UnicodeText`?

marcospri:
> So `Text` and `UnicodeText` produce the Postgres type `TEXT`. What type do `String` and `Unicode` produce? `VARCHAR`?

String and Unicode produce varchar yeah (which is the same as `character varying`). Without a limit they behave like TEXT (as apposed to varchar(25) which imposes that limit).


> 
> Also, what is the difference between `String` vs `Unicode`? And `Text` vs `UnicodeText`?


From postgres point of view none. From sqlalchemy's... I'd say if you use python3.x and current SQLA pretty much none, from the docs: 

```In modern SQLAlchemy, use of the Unicode datatype does not typically imply any encoding/decoding behavior within SQLAlchemy itself. Historically, when DBAPIs did not support Python unicode objects under Python 2, SQLAlchemy handled unicode encoding/decoding services itself which would be controlled by the flag String.convert_unicode; this flag is deprecated as it is no longer needed for Python 3.```

seanh (approved):
@marcospri Feel free to merge this if you like. I'm not sure that it'll enable copy-pasting of these old columns when creating new ones though, because I think the most correct type to use for new columns might be `Unicode` rather than `UnicodeText`

marcospri:
Merged as it goes from String -> Unicode in most cases. As mentioned here going from UnicodeText -> Unicode would involve a sqlalchemy migration (even if the resulting types are functionally equivalent)

PR #2841, opened 2021-06-17
Remove schema validation from HTTPService

seanh:


marcospri (approved):
Looks good, in retrospect it looks like the schema validation was a bit bolted on to `request()`

PR #2843, opened 2021-06-17
Raise a distinct error for missing Canvas files

jon-betts:
For: https://github.com/hypothesis/lms/pull/2783

We can tell the difference between a request to get a specific Canvas file because it's not there verses a general failure. 

Previously asking for a missing file would result in the more generic `CanvasAPIServerError`. In the case of course copy the distinction is important.

 * If the file access results in a permissions error - we should try and map it
 * If the file access results in a 404 - we should try and map it
 * If the file access results in another error we won't try

Previously the last two cases were the same case

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -139,12 +139,10 @@ class CanvasAPIError(ProxyAPIError):
         ) from cause
 
     @staticmethod
-    def _exception_class(response):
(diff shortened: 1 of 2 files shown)


jon-betts:
Err.. due to Canvas weirdness all deleted files as still accessible... so this actually doesn't happen

PR #2844, opened 2021-06-17
Use model defined onupdate on bulk_actions' upsert 

marcospri:
Fixes https://github.com/hypothesis/lms/issues/2909

This should work with all style of onupdate parameters except callables that expect a context parameter:

https://github.com/hypothesis/lms/pull/2844/files#diff-5743e15c634a3617289d65def23d8b87bf8cc430af559d08057d883e23467757R35

### DIFF
--- a/lms/db/_bulk_action.py
+++ b/lms/db/_bulk_action.py
@@ -93,19 +93,23 @@ class BulkAction:
 
         if config.upsert_trigger_onupdate:
             onupdate_columns = self._get_columns_onupdate(model_class)
+
             for column_name, onupdate_value in onupdate_columns:
                 upsert_update_elements.append(column_name)
+
+                # SQL alchemy wraps functions passed to onupdate or default and
+                # could potentially take a "context" argument getting a
+                # suitable context at this point of the execution it's not
+                # possible so we don't support it so we just pass None
+                # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
+                default_value = (
+                    onupdate_value(None) if callable(onupdate_value) else onupdate_value
+                )
+
                 # Copy the values, we don't want to mess with the caller's data
                 values = deepcopy(values)
-                for value_ in values:
-                    value_[column_name] = (
-                        # SQL alchemy wraps functions passed to onupdate or default and could potentially take a "context" argument
-                        # getting a suitable context at this point of the execution it's not possible so we don't support it so we just pass None
-                        # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
-                        onupdate_value(None)
-                        if callable(onupdate_value)
-                        else onupdate_value
-                    )
+                for row in values:
+                    row[column_name] = default_value
 
         stmt = insert(model_class).values(values)
         stmt = stmt.on_conflict_do_update(
--- a/lms/db/_bulk_action.py
+++ b/lms/db/_bulk_action.py
@@ -52,7 +52,7 @@ class BulkAction:
         upsert_update_elements: List[str]
         """Columns to update when a match is found."""
 
-        upsert_trigger_onupdate: bool = True
+        upsert_use_onupdate: bool = True
         """Column to update with the current datetime"""
 
         def __set_name__(self, owner, name):
@@ -91,7 +91,7 @@ class BulkAction:
 
         upsert_update_elements = list(config.upsert_update_elements)
 
-        if config.upsert_trigger_onupdate:
+        if config.upsert_use_onupdate:
             onupdate_columns = self._get_columns_onupdate(model_class)
 
             for column_name, onupdate_value in onupdate_columns:
--- a/lms/db/_bulk_action.py
+++ b/lms/db/_bulk_action.py
@@ -11,9 +11,11 @@ Why is this here, rather than services?
 For more see: https://stackoverflow.com/c/hypothesis/questions/477
 """
 
+from copy import deepcopy
 from dataclasses import dataclass
 from typing import List
 
+from sqlalchemy import inspect
 from sqlalchemy.dialects.postgresql import insert
 from zope.sqlalchemy import mark_changed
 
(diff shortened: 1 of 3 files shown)


lms/db/_bulk_action.py
```diff
@@ -50,6 +51,9 @@ class Config:
         upsert_update_elements: List[str]
         """Columns to update when a match is found."""
 
+        upsert_update_datetime_column: Optional[str] = None
```
marcospri, line 54: Instead of using the config we could do isinstance(model_class, CreateUpdateMixin) which is probably what you want most of the time but it's a bit more magical.

lms/db/_bulk_action.py
```diff
@@ -84,14 +89,32 @@ def upsert(self, model_class, values):
...
+                values = deepcopy(values)
+                for value_ in values:
+                    value_[column_name] = (
+                        # SQL alchemy wraps functions passed to onupdate or default and could potentially take a "context" argument
+                        # getting a suitable context at this point of the execution it's not possible so we don't support it so we just pass None
+                        # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
+                        onupdate_value(None)
+                        if callable(onupdate_value)
+                        else onupdate_value
+                    )
```
jon-betts, line 108:
```suggestion
                # SQL alchemy wraps functions passed to onupdate or default and
                # could potentially take a "context" argument getting a
                # suitable context at this point of the execution it's not
                # possible so we don't support it so we just pass None
                # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
                default_value = (
                    onupdate_value(None) if callable(onupdate_value) else onupdate_value
                )

                # Copy the values, we don't want to mess with the caller's data
                values = deepcopy(values)
                for row in values:
                    row[column_name] = default_value
```

I think the default value is the same for every row, so we don't need to calculate it once per row.

Also, I'm not sure why `value_` has an underscore on the end. Is this because `values` is a list of dicts and it's confusing? I used `row` in the example above which I think makes this clearer?

lms/db/_bulk_action.py
```diff
@@ -50,6 +52,9 @@ class Config:
         upsert_update_elements: List[str]
         """Columns to update when a match is found."""
 
+        upsert_trigger_onupdate: bool = True
+        """Column to update with the current datetime"""
```
jon-betts, line 56:
```suggestion
        """Update any columns with `onupsert` using the appropriate value."""
```

Also, this is probably a me brain thing, but `upsert`, `trigger`, and `onupdate` are all DB terms independently, which makes my brain read this as either:

* `upsert, trigger, onupdate`
* `(upsert trigger), onupdate`

Would something like `upsert_use_onupdate` be clearer?

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -58,6 +77,60 @@ def test_upsert(self, db_session):
...
+            [
+                {"id": 1, "name": "update_existing"},
+            ],
+        )
+
+        rows = list(
+            db_session.query(self.TableWithBulkUpsertOnUpdate).order_by(
+                self.TableWithBulkUpsertOnUpdate.id.asc()
+            )
+        )
```
jon-betts, line 104: I don't think you need to have both a sort and `Any.iterable.containing` as they provide the same benefit of being order independant.

As this little snippet of code is the same between the two tests and might make for a nice helper like:

```python
    def assert_has_rows(self, db_session, *attrs):
        rows = list(db_session.query(self.TableWithBulkUpsert))

        assert (
            rows
            == Any.iterable.containing(
                [
                    Any.instance_of(self.TableWithBulkUpsert).with_attrs(expected)
                    for expected in attrs
                ]
            ).only()
        )
```

Along with the suggested parameterisation above, this would make the end of the test quite snappy:
```python
        self.assert_has_rows(
            db_session,
            # 42 is the onupdate default value
            {"id": 1, "name": "update_existing", column: 42},
            {"id": 2, "name": "pre_existing_2", column: 1},
        )
```

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -58,6 +77,60 @@ def test_upsert(self, db_session):
     def test_upsert_does_nothing_if_given_an_empty_list_of_values(self, db_session):
         assert BulkAction(db_session).upsert(self.TableWithBulkUpsert, []) == []
 
+    def test_upsert_with_onupdate_columns(self, db_session):
+        db_session.add_all(
+            [
+                self.TableWithBulkUpsertOnUpdate(
+                    id=1, name="existing", scalar=0, callable=0, sql=0, default=0
+                ),
```
jon-betts, line 85: I was having a little trouble working out what the focus of this test is with the different columns, until I realised that this is testing all of the different variants at once.

It might help reveal the focus of the test and make it a little shorter as well to parameterise the columns so they are tested one by one:

```suggestion
    @pytest.mark.parametrize("column", ("scalar", "callable", "sql", "default"))
    def test_upsert_with_onupdate_columns(self, db_session, column):
        db_session.add_all(
            [
                self.TableWithBulkUpsertOnUpdate(id=1, name="pre_existing_1", **{column: 0}),
                ...
```

jon-betts (changes requested):
This works and I really like the idea of using the defined `onupdate` mechanism from the ORM, as it's going to work in loads more situations than the more restricted version. It's not significantly more magical and has a lot more possible benefits.

I have a couple of suggestions I think we **should** probably merge in:

 * The doc string for the config option still refers to what it used to do, not what it does now
 * The default value is calculated per row, but I think it's always the same
 
I have some suggestions I think we **could** merge in:

 * I have some suggestions about some ways to make the tests shorter by paramaterising, and a helper function etc.
 * I think it might be a good idea to rename the config option to not use the word "trigger" as it clashes with a RDB term

I've made a ton of suggestions, and I always hate feeling like I'm a remote version of `emacs` for someone, so I've done the work of implementing them (all bar the rename). So if you'd like to see my version and either merge it it, ignore it, or do it your own way it's here: https://github.com/hypothesis/lms/pull/2926

jon-betts (approved):
I'm so sorry this sat for so long. Totally my fault. Looks good!

PR #2854, opened 2021-06-21
Make BlackboardAPIClient low-level

seanh:
Depends on https://github.com/hypothesis/lms/pull/2841

We need a generic `request()` method for making Blackboard API requests
that isn't tied to any particular Blackboard API endpoint (list-files or
public-URL, ...) and that is a separate unit with its own separate unit
tests. We need this because the `request()` method is going to contain
some generic functionality that applies to multiple Blackboard API
endpoints: detecting access token-related error responses from
Blackboard (these need to be handled specially); using refresh tokens
and retrying when requests fail; pagination; etc. We want to implement
these generic functions once in the `request()` method and unit test
them once in the `request()` method's unit tests, and then have
higher-level methods like `list_files()` and `public_url()` get this
functionality for free.

So turn `BlackboardAPIClient` into a low-level API client with a generic
`request()` method rather than high-level `list_files()` and
`public_url()` methods.

Move the knowledge of particular Blackboard API methods like list-files
and public-URL into `BlackboardFilesAPIViews.list_files()` and
`BlackboardFilesAPIViews.public_url()`. This makes the view methods
longer but they're still simple enough: they contain little or no logic.

This is a simpler overall design. We could have a
`BlackboardFilesAPIViews.list_files()` view that calls a higher-level
`BlackboardService.list_files()` method that calls a lower-level
`BlackboardAPIClient.request()`. But the higher-level Blackboard service
and the extra level of indirection are unnecessary. At least for now
it's enough to have a `list_files()` view that calls
`BlackboardAPIClient.request()` and this results in less code and fewer
tests.


lms/services/__init__.py
```diff
@@ -1,6 +1,5 @@
 from lms.services.canvas import CanvasService
 from lms.services.exceptions import (
-    BlackboardFileNotFoundInCourse,
```
seanh, line 3: It's now the view rather than the service that raises `BlackboardFileNotFoundInCourse` so the exception class itself has moved into the views.

lms/services/blackboard_api.py
```diff
@@ -1,18 +1,5 @@
-import re
```
seanh, line 1: The `BlackboardAPIClient` now fits in a single `services/blackboard_api.py` files. It's no longer necessary to have a `services/blackboard_api/` subdir with `service.py` and `_schemas.py` files because the schemas are no longer part of the service: they've moved into the views.

lms/services/blackboard_api.py
```diff
@@ -59,28 +46,8 @@ def get_token(self, authorization_code):
...
-                oauth=True,
-            )
-        except HTTPError as err:
-            if err.response.status_code == 404:
-                raise BlackboardFileNotFoundInCourse(file_id) from err
-            raise
-
-        return BlackboardPublicURLSchema(response).parse()
+    def request(self, method, path):
+        return self._http_service.request(method, self._api_url(path), oauth=True)
```
seanh, line 50: The `request()` method is very simple for now but more functionality will be added to it in future PRs

lms/views/api/blackboard/files.py
```diff
@@ -14,21 +28,32 @@ def __init__(self, request):
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
     def list_files(self):
         """Return the list of files in the given course."""
-        return self.blackboard_api_client.list_files(
-            self.request.matchdict["course_id"]
+        course_id = self.request.matchdict["course_id"]
+        response = self.blackboard_api_client.request(
+            "GET", f"courses/uuid:{course_id}/resources"
         )
+        return BlackboardListFilesSchema(response).parse()
```
seanh, line 35: The `list_files()` view gets a bit longer but it's zero logic: no `if`'s or loops or anything

lms/views/api/blackboard/files.py
```diff
@@ -14,21 +28,32 @@ def __init__(self, request):
...
+        except HTTPError as err:
+            if err.response.status_code == 404:
+                raise BlackboardFileNotFoundInCourse(file_id) from err
+            raise
+
+        public_url = BlackboardPublicURLSchema(response).parse()
+
+        via_url = helpers.via_url(self.request, public_url, content_type="pdf")
+
+        return {"via_url": via_url}
```
seanh, line 59: The `via_url()` view is a bit longer but still the only bit of logic is a single `if`. It only takes three unit tests to fully cover this view

marcospri (commented):
`request` makes a lot of sense.

My only concern is that if we need to add some more beefy API interactions that make more sense in the service we'll either leave some API calls in the views and others in the service or we'll have to refactor and move everything to the service then  -and write the very simple but boilerplaty test for "view calls service", "service calls request()"-

marcospri (approved):

marcospri (approved):

PR #2856, opened 2021-06-21
Handle access token-related errors from Blackboard

seanh:
Depends on https://github.com/hypothesis/lms/pull/2854

Fixes https://github.com/hypothesis/lms/issues/2834

Handle access token-related errors from Blackboard correctly.

It's common for Blackboard API requests to come back with an error
response that means our access token was invalid. For example this
happens when the access token has been expired or has been revoked. When
these happen we need to catch them and raise `OAuth2TokenError` so that
the frontend displays an **Authorize Hypothesis** dialog rather than
displaying a **Something went wrong** error dialog.

Testing
-------

Break all the access tokens in your DB:

This will generate the same response from Blackboard as if the access
tokens had expired. Alternatively you can wait for one hour to let them
actually expire.

Now launch the Blackboard Files test assignment. You should see the
**Authorize Hypothesis** dialog and then the assignment should launch
successfully.

### DIFF
--- a/lms/services/blackboard_api.py
+++ b/lms/services/blackboard_api.py
@@ -1,4 +1,46 @@
+from marshmallow import INCLUDE, fields
+
+from lms.services.exceptions import HTTPError, OAuth2TokenError
+from lms.validation import RequestsResponseSchema, ValidationError
 from lms.validation.authentication import OAuthTokenResponseSchema
+
+
+class BlackboardErrorResponseSchema(RequestsResponseSchema):
+    """
+    Schema for validating error responses from the Blackboard API.
+
+    This schema's parse() method will return the given response's JSON body as
+    a dict (with whatever fields the response body from Blackboard had) and
+    will default to returning an empty dict if the given response's body isn't
+    a JSON dict, or isn't a dict at all, or if the response isn't a
+    requests.Response object (e.g. if it's None).
+
+    """
+
+    # Blackboard API error bodies have a "status" key whose value is the
+    # response's HTTP status code as an int.
+    #
+    # This schema is permissive and doesn't guarantee that the dict returned
+    # from parse() will have a "status" or that "status"'s value will be an
+    # int.
+    status = fields.Raw(required=False, allow_none=True)
+
+    # Blackboard API error bodies have a "message" key whose value is a
+    # human-readable error message string.
+    #
+    # This schema is permissive and doesn't guarantee that the dict returned
+    # from parse() will have a "message" or that "message"'s value will be a
+    # string.
+    message = fields.Raw(required=False, allow_none=True)
+
+    class Meta:
+        unknown = INCLUDE
+
+    def parse(self, *args, **kwargs):
+        try:
+            return super().parse(*args, **kwargs)
+        except ValidationError:
+            return {}
 
 
 class BlackboardAPIClient:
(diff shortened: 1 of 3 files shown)


lms/services/blackboard_api.py
```diff
@@ -1,6 +1,32 @@
...
+    """
+
+    class Meta:
+        unknown = INCLUDE
+
+    def parse(self, *args, **kwargs):
+        try:
+            return super().parse(*args, **kwargs)
+        except ValidationError:
+            return {}
```
seanh, line 27: This schema handles quite a lot for us in very little code. You can just do:

```python
error_dict = BlackboardErrorResponseSchema(err.response).parse()
```

and can then go ahead and assume that `error_dict` is a dict (although you can't assume anything about the contents of the dict) even though `err.response` could have been:

* `None` (or anything else that doesn't have a `.json()` method like `requests.Response` does)
* A `requests.Response` object with a non-JSON body
* A response with a non-dict JSON body (e.g. a list)

And the schema can be re-used wherever we want to parse Blackboard error responses in future.

It could possibly be rewritten without Marshmallow. Something like:

```python
def error_dict(response):
    try:
        error_dict = response.json()
    except (AttributeError, ValueError):
        error_dict = {}

    if not isinstance(error_dict, dict):
        error_dict = {}

    return error_dict
```

... but this would kind of just be duplicating code that Marshmallow and `RequestsResponseSchema` already contains, so I'm not sure.

lms/services/blackboard_api.py
```diff
@@ -47,7 +73,15 @@ def get_token(self, authorization_code):
...
-        return self._http_service.request(method, self._api_url(path), oauth=True)
+        try:
+            return self._http_service.request(method, self._api_url(path), oauth=True)
+        except HTTPError as err:
+            error_dict = BlackboardErrorResponseSchema(err.response).parse()
+
+            if error_dict.get("message") == "Bearer token is invalid":
+                raise OAuth2TokenError() from err
+
+            raise
```
seanh, line 84: When an access token is expired Blackboard sends `{"status": 401, "message": "Bearer token is invalid"}`. I'm not sure there's any way to detect this except by the human-readable error message.

`OAuth2TokenError` is already handled by a special exception view that sends a response that causes the frontend to show the **Authorize Hypothesis** dialog

lms/validation/_base.py
```diff
@@ -149,7 +149,7 @@ def parse(self, *args, **kwargs):
     def _pre_load(self, response, **_kwargs):  # pylint: disable=no-self-use
         try:
             return response.json()
-        except ValueError as err:
+        except (AttributeError, ValueError) as err:
```
seanh, line 152: `AttributeError` is what happens if the given `response` isn't a `requests.Response` object (e.g. if it's `None`)

lms/services/blackboard_api.py
```diff
@@ -1,6 +1,32 @@
...
+    a dict (with whatever fields the response body from Blackboard had) and
+    will default to returning an empty dict if the given response's body isn't
+    a JSON dict, or isn't a dict at all, or if the response isn't a
+    requests.Response object (e.g. if it's None).
+
+    """
+
+    class Meta:
+        unknown = INCLUDE
+
```
marcospri, line 22: There's nothing we could assume will be always there? `message`?

seanh: Hmm, yeah I think I could add `status` and `message` with `required=False` just for documentation purposes. And maybe `fields.Raw` rather than `Int` and `Str`. In trying out this code different ways I found it was simpler if the schema didn't let the calling code make any assumptions about the contents of the dict. If I tried to make the schema enforce that the dict contained a `message` string it just ended up awkward when you try to deal with all the situations (when the JSON body doesn't contain a `message` but does contain other fields, when it does have a `message` but it's not a string, when the JSON body isn't even a dict, ...). But I could add `message = fields.Raw(required=False)` just for documentation purposes

marcospri: Didn't know about fields.Raw, I think that would be useful, the empty schema feels a bit odd to me otherwise (Even If I know what's doing)

Both `required=False` and `allow_none=True` might be needed

seanh: Added these as raw fields

marcospri (approved):

PR #2866, opened 2021-06-22
Don't crash when a Canvas course has no files

seanh:
Test course with no Canvas files but with pre-installed localhost, QA and prod application instances so you can reproduce the crash: https://hypothesis.instructure.com/courses/270 (you'll need to re-run `make devdata`)

Fixes https://github.com/hypothesis/lms/issues/2865

Slack discussion: https://hypothes-is.slack.com/archives/C2BLQDKHA/p1624372670308400

When getting the list of files for a Canvas course that has no files the
bulk upsert code that's trying to upsert the empty list of files into
the DB crashes. The reason is that this:

```python
stmt = insert(MyModel).values([])
```

produces this SQL (you can see the SQL by doing `print(stmt)` in
Python):

```sql
INSERT INTO my_table DEFAULT VALUES RETURNING my_table.id
```

The SQL tells SQLAlchemy to insert rows into the table using the DB
schema's default values for all the rows, which means using `NULL` for
any rows that don't have another default specified in the schema. If any
of these rows have `NOT NULLABLE` constraints this will produce a `null
value in column "foo" violates not-null constraint` error.

### DIFF
--- a/tests/unit/lms/db/_bulk_action_test.py
+++ b/tests/unit/lms/db/_bulk_action_test.py
@@ -6,7 +6,7 @@ from sqlalchemy.engine import CursorResult
 from lms.db import BASE, BulkAction
 
 
-class TestBulkUpsertMixin:
+class TestBulkAction:
     class TableWithBulkUpsert(BASE):
         __tablename__ = "test_table_with_bulk_upsert"
 
@@ -15,7 +15,7 @@ class TestBulkUpsertMixin:
         )
 
         id = sa.Column(sa.Integer, primary_key=True)
-        name = sa.Column(sa.String)
+        name = sa.Column(sa.String, nullable=False)
         other = sa.Column(sa.String)
 
     def test_upsert(self, db_session):
@@ -55,9 +55,14 @@ class TestBulkUpsertMixin:
             ).only()
         )
 
+    def test_upsert_does_nothing_if_given_an_empty_list_of_values(self, db_session):
+        assert BulkAction(db_session).upsert(self.TableWithBulkUpsert, []) == []
+
     def test_it_fails_with_missing_config(self, db_session):
         with pytest.raises(AttributeError):
-            BulkAction(db_session).upsert("object_without_config", [])
+            BulkAction(db_session).upsert(
+                "object_without_config", [{"id": 1, "name": "name", "other": "other"}]
+            )
 
     def test_you_cannot_add_config_with_the_wrong_name(self):
         # Not sure why this isn't ValueError... must be a descriptor thing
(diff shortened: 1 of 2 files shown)


tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -6,7 +6,7 @@
 from lms.db import BASE, BulkAction
 
 
-class TestBulkUpsertMixin:
+class TestBulkAction:
```
seanh, line 9: Noticed that the test class had the wrong name

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -15,7 +15,7 @@ class TableWithBulkUpsert(BASE):
         )
 
         id = sa.Column(sa.Integer, primary_key=True)
-        name = sa.Column(sa.String)
+        name = sa.Column(sa.String, nullable=False)
```
seanh, line 18: We needed a non-nullable column on the test table in order to produce the crash

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -55,9 +55,15 @@ def test_upsert(self, db_session):
...
+    def test_upsert_does_nothing_if_given_an_empty_list_of_values(self, db_session):
+        assert BulkAction(db_session).upsert(self.TableWithBulkUpsert, []) == []
+
     def test_it_fails_with_missing_config(self, db_session):
         with pytest.raises(AttributeError):
-            BulkAction(db_session).upsert("object_without_config", [])
+            BulkAction(db_session).upsert(
+                "object_without_config",
+                [{"id": 1, "name": "name", "other": "other"}],
+            )
```
seanh, line 66: This test was previously passing an empty list to `upsert()` bit wasn't triggering the crash because the test table didn't have any non-nullable columns. I changed it to not pass an empty list so that it doesn't run into the special case

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -55,9 +55,14 @@ def test_upsert(self, db_session):
...
 
+    def test_upsert_does_nothing_if_given_an_empty_list_of_values(self, db_session):
+        assert BulkAction(db_session).upsert(self.TableWithBulkUpsert, []) == []
+
     def test_it_fails_with_missing_config(self, db_session):
         with pytest.raises(AttributeError):
-            BulkAction(db_session).upsert("object_without_config", [])
+            BulkAction(db_session).upsert(
+                "object_without_config", [{"id": 1, "name": "name", "other": "other"}]
+            )
```
seanh, line 65: This test was previously passing an empty list to `upsert()` bit wasn't triggering the crash because the test table didn't have any non-nullable columns. I changed it to not pass an empty list so that it doesn't run into the special case

robertknight (approved):
Looks straightforward to me.

PR #2867, opened 2021-06-23
Update casing of acronyms in identifiers in frontend code

robertknight:
Update the names of several identifiers in frontend code to match the current [conventions for identifier naming](https://github.com/hypothesis/frontend-toolkit/blob/master/docs/js-guide.md#identifier-naming). See commit messages for details.

----

Note for future reference:

The shell command I used involved [fd](https://github.com/sharkdp/fd) and [sd](https://github.com/chmln/sd) (modern alternatives to `find` and `sed`):

```shellsession
$ fd . lms/static | xargs sd ApiCallInfo APICallInfo
```

For more targeted renaming of symbols I used TypeScript's rename refactoring via [`:ALERename`](https://github.com/dense-analysis/ale#2viii-refactoring-rename-actions) (in VS Code this is probably available in the context menu).


esanzgar (approved):
👍🏼 

I haven't tested the changes in the CSS, but I trust it has no effect on the UI.

Issue #2859, opened 2021-06-22
Handle Blackboard Files assignments whose file has been deleted


seanh:
## Problem

Note: There's already an assignment named **localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted** in our [Blackboard test course](https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline) that you can just launch to reproduce this problem (you'll need to re-run `make devdata` in LMS first), so you don't actually need to follow these steps:

1. Log in to our test Blackboard site as a teacher, create a Blackboard files assignment, launch it, and select a file (using the <kbd>Choose PDF from Blackboard</kbd> button)
2. Now go and delete the selected file from Blackboard
3. Go back and launch the assignment

You will see a generic **Something went wrong** error dialog:

![Screenshot from 2021-06-22 11-25-31](https://user-images.githubusercontent.com/22498/122908934-95416c00-d34c-11eb-926f-799354d27a75.png)

You _should_ be seeing a more specific **Hypothesis couldn't find the assignment's PDF file in the Blackboard course** error message. The backend (`/via_url` API) is actually sending an error response (400 Bad Request) with this JSON body:

```json
{
  "message": "Hypothesis couldn't find the assignment's PDF file in the Blackboard course.",
  "details": {"document_url": "_9019_1"}
}
```

But the frontend no longer seems to make any use of the `message` sent by the backend.

There's a docstring documenting how the backend expects this to work here:

https://github.com/hypothesis/lms/blob/f70c0ccbdb2bbc006e93a92013599512c0125698/lms/views/api/exceptions.py#L25-L91

I can't remember whether the `message` string returned by the backend used to be used as the error dialog's title or its body text. I suspect it may have been shown in the body, in which case the error dialog _should_ look like this:

![Screenshot from 2021-06-22 11-30-35](https://user-images.githubusercontent.com/22498/122909670-565fe600-d34d-11eb-86c4-923f859f8a75.png)

Note that there are probably other error situations that are broken in the same way: just saying <q>There was a problem fetching this Hypothesis assignment</q> when the backend has actually supplied a specific `message`.

robertknight:
It looks like this text:

> Something went wrong: there was a problem fetching this Hypothesis assignment

[...]

PR #2872, opened 2021-06-23
Fix ErrorDisplay so it shows error.message

LMS007:
Fix ErrorDisplay to show the provided message in the rare cases where error.message is not empty AND the message prop is missing.

nb. In most cases error.message's text originated from the server.

--------

There are 4 possible states for the messages that ErrorDisplay renders

1. message && error.message = "message: _error.message_"
2. message && !error.message = "message"
3. !message && error.message = "error.message"
4. !message && !error.message = ""

Prior to this fix, case 3 was not working and so the error.message was not showing in the DOM.

run `make devdata`
Load up this assignment https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_80_1&course_id=_19_1

...you can now see the error message in the dialog

_"A problem occurred while handling this request. Hypothesis has been notified."_

Screen Shot 2021-06-23 at 3 06 33 PM

Fixes https://github.com/hypothesis/lms/issues/2859

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
@@ -1,4 +1,4 @@
-import { Fragment, createElement } from 'preact';
+import { createElement } from 'preact';
 
 /**
  * Generate a `mailto:` URL that prompts to send an email with pre-filled fields.
@@ -15,6 +15,8 @@ function emailLink({ address, subject = '', body = '' }) {
 }
 
 /**
+ * Adds punctuation to a string if missing.
+ *
  * @param {string} str
  */
 function toSentence(str) {
@@ -34,8 +36,13 @@ function toSentence(str) {
  * @prop {string|null} [message] -
  *   A short message to display explaining that a problem happened. This is
  *   typically a general message like "There was a problem fetching this assignment".
+ *   In cases where the the error originates from the server, this message may not
+ *   be necessary, but in other cases where the `error.message` is generic and perhaps
+ *   originates from an exception in the client, then this prop can be used to
+ *   provide additional specifics.
  * @prop {ErrorLike} error -
- *   An `Error`-like object with specific details of the problem
+ *   An `Error`-like object with specific details of the problem. If `error` contains
+ *   a `message` property, then that string will be rendered.
  */
 
 /**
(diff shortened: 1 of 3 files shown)


lms/static/scripts/frontend_apps/components/test/LaunchErrorDialog-test.js
```diff
@@ -83,7 +83,7 @@ describe('LaunchErrorDialog', () => {
       retryAction = 'Try again',
     }) => {
       it(`displays expected error for "${errorState}" error state`, () => {
-        const error = new Error('Detailed error info');
```
LMS007, line 86: This now correctly gets printed to the DOM and affected this test. Simple fix is to just remove the error's message

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -84,16 +86,20 @@ export default function ErrorDisplay({ message, error }) {
...
         </p>
       )}
+      {message && !error.message && (
+        // Only the component's message prop is provided
+        <p data-testid="message">{toSentence(message)}</p>
+      )}
+      {!message && error.message && (
+        // Only the error's message property is provided
+        <p data-testid="message">{toSentence(error.message)}</p>
+      )}
```
robertknight, line 102: I wonder if we ought to include a case for `!message && !error.message` here for the sake of completeness. This should almost never be triggered in practice because `Error`-like objects should almost always have a `message` property. This could perhaps use the string "An unknown error occurred."

robertknight (approved):
Thanks for the clear description + truth table of possibilities in the PR. I added some small suggestions but the main changes here look good.

lms/static/scripts/frontend_apps/components/test/ErrorDisplay-test.js
```diff
@@ -151,6 +151,42 @@ describe('ErrorDisplay', () => {
...
+      output: false,
+    },
+  ].forEach(({ message, error, output }, index) => {
+    it(`shows the appropriate error message if provided (${index})`, () => {
+      const wrapper = mount(
+        <ErrorDisplay message={message} error={{ message: error }} />
+      );
+      if (output) {
+        assert.equal(
+          wrapper.find('p[data-testid="message"]').first().text(),
```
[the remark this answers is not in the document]
LMS007: no, I think that was blind copy paste without thinking. Because I added data-testid, we don't the first() call.

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -84,16 +86,20 @@ export default function ErrorDisplay({ message, error }) {
...
     <div className="ErrorDisplay">
-      {message && (
-        <p>
-          {!error.message && toSentence(message)}
-          {error.message && (
-            <Fragment>
-              {message}: <i>{toSentence(error.message)}</i>
-            </Fragment>
-          )}
+      {message && error.message && (
```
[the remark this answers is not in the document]
LMS007: roger

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
LMS007: Sounds good.

LMS007:
- Updated ErrorDisplayProps to explain in more detail how message and error.message work
- Added back exception string in LauchErrorDialog-test
- Added cases when error.message and message are both missing
- Addressed minor fixups in ErrorDisplay-test

PR #2875, opened 2021-06-24
Add File<>ApplicationInstance relation

seanh:
Add `File.application_instance` and `ApplicationInstance.files`. These are convenience attributes that'll be used in future PRs.


marcospri (approved):
This looks like a good thing to have in the models regarless of the course copy feature :+1:

PR #2876, opened 2021-06-24
Add the ModuleItemConfiguration.extra column

seanh:
This is going to be used to store "Canvas file mappings" to fix course-copied Canvas Files assignments.

This will need a DB migration.

PR #2878, opened 2021-06-24
Add models.CanvasModuleItemConfiguration

seanh:
Depends on https://github.com/hypothesis/lms/pull/2876

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -59,6 +59,26 @@ class AssignmentService:
         # just one key from the cache, you have to clear the entire cache.)
         self.get.cache_clear()
 
+    def get_canvas_mapped_file_id(
+        self, tool_consumer_instance_guid, resource_link_id, file_id
+    ):
+        mic = self.get(tool_consumer_instance_guid, resource_link_id)
+
+        if not mic:
+            return None
+
+        return mic.extra.get("canvas_file_mappings", {}).get(file_id)
+
+    def set_canvas_mapped_file_id(
+        self, tool_consumer_instance_guid, resource_link_id, file_id, mapped_file_id
+    ):
+        mic = self.get(tool_consumer_instance_guid, resource_link_id)
+
+        if not mic:
+            raise ValueError("ModuleItemConfiguration not found")
+
+        mic.extra.setdefault("canvas_file_mappings", {})[file_id] = mapped_file_id
+
 
 def factory(_context, request):
     return AssignmentService(db=request.db)
--- a/tests/unit/lms/services/assignment_test.py
+++ b/tests/unit/lms/services/assignment_test.py
@@ -60,6 +60,69 @@ class TestAssignmentService:
 
         assert assignment.document_url == "NEW_DOCUMENT_URL"
 
+    def test_get_canvas_mapped_file_id_returns_None_if_the_assignment_doesnt_exist(
+        self, svc
+    ):
+        assert not svc.get_canvas_mapped_file_id(
+            "unknown_resource_link_id", "unknown_resource_link_id", "file_id"
+        )
+
+    def test_set_canvas_mapped_file_id_creates_a_new_mapping_if_none_exists(
+        self, assignment, svc
+    ):
+        svc.set_canvas_mapped_file_id(
+            assignment.tool_consumer_instance_guid,
+            assignment.resource_link_id,
+            "original_file_id",
+            "mapped_file_id",
+        )
+
+        assert (
+            svc.get_canvas_mapped_file_id(
+                assignment.tool_consumer_instance_guid,
+                assignment.resource_link_id,
+                "original_file_id",
+            )
+            == "mapped_file_id"
+        )
+
+    def test_set_canvas_mapped_file_id_overwrites_an_existing_mapping_if_one_exists(
+        self, assignment, svc
+    ):
+        svc.set_canvas_mapped_file_id(
+            assignment.tool_consumer_instance_guid,
+            assignment.resource_link_id,
+            "original_file_id",
+            "mapped_file_id",
+        )
+
+        svc.set_canvas_mapped_file_id(
+            assignment.tool_consumer_instance_guid,
+            assignment.resource_link_id,
+            "original_file_id",
+            "new_mapped_file_id",
+        )
+
+        assert (
+            svc.get_canvas_mapped_file_id(
+                assignment.tool_consumer_instance_guid,
+                assignment.resource_link_id,
+                "original_file_id",
+            )
+            == "new_mapped_file_id"
+        )
+
+    def test_set_canvas_mapped_file_id_raises_ValueError_if_the_assignment_doesnt_exist(
+        self, svc
+    ):
+        with pytest.raises(ValueError):
+            svc.set_canvas_mapped_file_id(
+                "unknown_tool_consumer_instance_guid",
+                "unknown_resource_link_id",
+                "original_file_id",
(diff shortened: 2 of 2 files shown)


lms/models/module_item_configuration.py
```diff
@@ -50,3 +50,11 @@ class ModuleItemConfiguration(BASE):
...
         nullable=False,
     )
+
+
+class CanvasModuleItemConfiguration(ModuleItemConfiguration):
+    def get_mapped_file_id(self, file_id):
+        return self.extra.get("canvas_file_mappings", {}).get(file_id)
+
+    def set_mapped_file_id(self, file_id, mapped_file_id):
+        self.extra.setdefault("canvas_file_mappings", {})[file_id] = mapped_file_id
```
seanh, line 60: I've put this in a `CanvasModuleItemConfiguration` subclass in order to head off the objection that we can't have Canvas-specific methods on the `ModuleItemConfiguration` class. It means that if you want to use these two methods you have to do `db.query(CanvasModuleItemConfiguration)` not `db.query(ModuleItemConfiguration)`.

I considered just storing the mappings in a separate table `models.CanvasFileMapping`) instead of using a JSON column. But the service method that actually needs to use these mappings alreadys needs to have the `ModuleItemConfiguration` object for other reasons anyway. If the mappings are stored in a different table then the service has to do an extra query to get them, instead of just doing `mic.get_mapped_file_id()` and `mic.set_mapped_file_id()` when it already has `mic`.

These problems can be alleviated. If using a separate table you can use SQLAlchemy relations so that `mic.canvas_file_mappings` is a list of all the `CanvasFileMappings`. And I suspect you might also be able to use eager loading to avoid the additional SQL query.

But it's actually simpler to use JSON, and requires less fancy SQLAlchemy stuff (relations, eager loading).

tests/unit/lms/models/module_item_configuration_test.py
```diff
@@ -0,0 +1,28 @@
...
+        mic.set_mapped_file_id("original_file_id", "mapped_file_id")
+
+        assert mic.get_mapped_file_id("original_file_id") == "mapped_file_id"
+
+    def test_set_mapped_file_id_overwrites_an_existing_mapping_if_one_exists(self, mic):
+        mic.set_mapped_file_id("original_file_id", "mapped_file_id")
+
+        mic.set_mapped_file_id("original_file_id", "new_mapped_file_id")
+
+        assert mic.get_mapped_file_id("original_file_id") == "new_mapped_file_id"
```
seanh, line 17: `set_mapped_file_id()` is written in such a way that SQLAlchemy's mutation tracking does work and changes do get persisted to the DB. This would be easy to get wrong since mutation tracking in nested dicts doesn't normally work even with `MutableDict`. This test is currently not covering that: the test would not fail if the mutation tracking and persistence were broken. But I'm not sure exactly how to write a test that _would_ fail.

tests/unit/lms/models/module_item_configuration_test.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @pytest.fixture
+    def mic(self, db_session):
+        mic = CanvasModuleItemConfiguration(
+            resource_link_id="resource_link_id",
+            tool_consumer_instance_guid="tool_consumer_instance_guid",
+            document_url="document_url",
+        )
+        db_session.add(mic)
+        db_session.flush()
```
seanh, line 27: The flush is needed to let the default value for the `extra` column kick in.

lms/models/module_item_configuration.py
marcospri: Not sure if that's the right place for those two method then, having both `db.query(CanvasModuleItemConfiguration)` and `db.query(ModuleItemConfiguration)` will suggest there's some difference at the storage layer.

I think we should either rename them to "get_canvas_mapped_file_id" and leave them on `ModuleItemConfiguration` (the model is storing canvas specific information, nothing wrong with retrieving that information in the same place IMO) or maybe move them to the service? 


Agree on JSON vs an extra table here.

seanh: Ok, I've changed this to just `ModuleItemConfiguratin.{get,set}_canvas_file_mapping()`

marcospri (approved):
Now that you opened https://github.com/hypothesis/lms/pull/2892#pullrequestreview-694064464 I wonder I you could also place the canvas specific code there? `FileService.get_canvas_mapped_file_id`? 

It will mean passing a model instance to a service which is not something we do atm in that manner. 

They look good to me in the model anyway.

seanh:
> Now that you opened [#2892 (review)](https://github.com/hypothesis/lms/pull/2892#pullrequestreview-694064464) I wonder I you could also place the canvas specific code there? `FileService.get_canvas_mapped_file_id`?
> 
> It will mean passing a model instance to a service which is not something we do atm in that manner.
> 
> They look good to me in the model anyway.

I think the file mappings are stored on assignments (`models.ModuleItemConfiguration`) not on files (`models.File`) so if we were to move the `{get,set}_canvas_mapped_file_id()` methods into the services layer I think they'd go in [`AssignmentService`](https://github.com/hypothesis/lms/blob/7437d13a12e76a572ac34b3ee705e7a5bb0de5df/lms/services/assignment.py#L6-L60). And actually they would fit in quite nicely with the existing methods already on `AssignmentService`:

```
AssignmentService:
    get_document_url(tool_consumer_instance_guid, resource_link_id)
    set_document_url(tool_consumer_instance_guid, resource_link_id, document_url)

    get_canvas_file_mapping(tool_consumer_instance_guid, resource_link_id, file_id)
    set_canvas_file_mapping(tool_consumer_instance_guid, resource_link_id, file_id, mapped_file_od)
```

And `AssignmentService` already has per-request caching of the DB query result for looking up the `ModuleItemConfiguration`.

On the other hand it means that `CanvasService.public_url()` will need two new arguments `tool_consumer_instance_guid` and `resource_link_id`, instead of just one new `module_item_configuration` argument.

seanh:
@marcospri Could you re-review this? I've moved the methods into `AssignmentService`. Looking at the existing code design [`AssignmentService` is actually the only place where `ModuleItemConfiguration` is used currently](https://github.com/hypothesis/lms/search?q=ModuleItemConfiguration&type=code). Currently `AssignmentService` encapsulates `ModuleItemConfiguration` so putting these new `ModuleItemConfiguration`-based methods in `AssignmentService` fits with the existing design. I think it's going to make the business code that finally uses these methods slightly nicer too

lms/services/assignment.py
```diff
@@ -59,6 +59,26 @@ def set_document_url(
...
+
+    def set_canvas_mapped_file_id(
+        self, tool_consumer_instance_guid, resource_link_id, file_id, mapped_file_id
+    ):
+        mic = self.get(tool_consumer_instance_guid, resource_link_id)
+
+        if not mic:
+            raise ValueError("ModuleItemConfiguration not found")
+
+        mic.extra.setdefault("canvas_file_mappings", {})[file_id] = mapped_file_id
```
seanh, line 80: I think it's probably simplest to just mix the Canvas-specific methods in with the non-specific methods in `AssignmentService` for now. But services make it easy to separate them in future if we want. You can have separate services and they can use each other via composition. So for example you could have a separate `CanvasFilesAssignmentService`. The service factory can even return different services depending on the context

marcospri (approved):
That was a lot of moving around these two methods but `lms/services/assignment.py` looks like the best place :+1:

PR #2880, opened 2021-06-25
Explicitly sort by `position` when getting files

marcospri:
Fixes https://github.com/hypothesis/lms/issues/2879

Sorting by name (the default) leads to, depending on number of files/ pagination, weird behavior.

For example, to reproduce the orignal bug, you can rename this file to `AAAA.pdf` (which already exists on the root folder)

https://hypothesis.instructure.com/courses/125/files/folder/Duplicated%20name%20file?

(remember to rename it back to something else to avoid leaving the course in a broken state)

change your pagination setting locally to something smaller in `lms/services/canvas_api/_basic.py` change `PAGINATION_PER_PAGE` to `2`

you'll get an exception if you try to launch the assignment. Notice that is not just a duplicated row, of the two AAAA.pdf files only one is returned but twice so information is lost.

To fix this we do both:

- Remove duplicates and notify in sentry. We don't want to pass the duplicates along in the system but we want to have some track record of that happening.
- We change the ordering on canvas api using the `sort` parameter to avoid the duplicates in the first place

According to the docs https://canvas.instructure.com/doc/api/files.html#method.files.api_index the allowed values for sort are 
**name, size, created_at, updated_at, content_type, user**. All of them allow for duplicates, maybe "created_at" being the less likely to produce them (it uses postgres' datetime so it has a lot of resolution) but could potentially also be problematic. 

There's another, undocumented, option "position", https://github.com/instructure/canvas-lms/blob/master/app/controllers/files_controller.rb#L305 which uses two columns which should be better.

### DIFF
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -209,11 +209,14 @@ class CanvasAPIClient:
             ]
 
     @lru_cache
-    def list_files(self, course_id):
+    def list_files(self, course_id, sort="position"):
         """
         Return the list of files for the given `course_id`.
 
         :param course_id: the Canvas course_id of the course to look in
+        :param sort: field to sort by (on Canvas' API side).
+            Defaults to "position" which is an undocumented option but that it should be the most stable of the options available as it sorts by both "position" and "name" on Canvas' side.
+            https://github.com/instructure/canvas-lms/blob/d43feb92d40d2c69684c4536f74dec37992c557a/app/controllers/files_controller.rb#L305
         :rtype: list(dict)
         """
         # For documentation of this request see:
@@ -222,10 +225,10 @@ class CanvasAPIClient:
         files = self._client.send(
             "GET",
             f"courses/{course_id}/files",
-            params={"content_types[]": "application/pdf"},
+            params={"content_types[]": "application/pdf", "sort": sort},
             schema=self._ListFilesSchema,
         )
-        # Canvas' pagination its broken as it sorts by fields that allow duplicates.
+        # Canvas' pagination is broken as it sorts by fields that allows duplicates.
         # This can lead to objects being skipped or duplicated across pages.
         # We can't detected objects that are not returned but we can detect the duplicates,
         #   remove them and notify sentry to see how often this happens after using a different sort parameter.
@@ -255,8 +258,7 @@ class CanvasAPIClient:
                 ],
             )
         )
-
-        return files
+        return sorted(files, key=lambda file_: file_["display_name"])
 
     class _ListFilesSchema(RequestsResponseSchema):
         """Schema for the list_files response."""
--- a/tests/unit/lms/services/canvas_api/client_test.py
+++ b/tests/unit/lms/services/canvas_api/client_test.py
@@ -353,6 +353,7 @@ class TestCanvasAPIClient:
                     {
                         "content_types[]": "application/pdf",
                         "per_page": Any.string(),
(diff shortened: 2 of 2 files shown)


seanh (changes requested):
This also changes the order of the files in the **Select a file** dialog, since I guess the frontend just displays them in the order returned by the backend. On master they're sorted by name which is a useful ordering for scrolling to find a particular file, and also matches how they seem to be sorted by default in Canvas's files tab. I'm not sure what sorting by "position" and name means?

If I understood correctly by de-duplicating the files returned by Canvas we fix the crash, but this leaves another bug that one of the files returned is missing? So we do need to also change the sort order parameter to get the missing file back.

In that case, I wonder if we could re-sort the files by name ourselves before returning them?

Screenshot from 2021-06-28 10-32-08

seanh:
I filed a bug with Canvas: https://github.com/instructure/canvas-lms/issues/1901

marcospri:
> In that case, I wonder if we could re-sort the files by name ourselves before returning them?
> 

Done. Note that there's an edge case where same named files are render together without any indication of to which folder they belong to. 

This could potentially be solved but would required additional API calls to get the names of the folders separately. Probably not worth the effort until it becomes a problem?

seanh:
> Done. Note that there's an edge case where same named files are render together without any indication of to which folder they 
> belong to.
> 
> This could potentially be solved but would required additional API calls to get the names of the folders separately. Probably not 
> worth the effort until it becomes a problem?

We don't get the folder names, when we get the list of files? Yeah that's difficult, we can't just make loads of API calls to get all the folder names. As far as I know we haven't had any complaints of feature requests about this (our lack of folder support) so I think we're okay to ignore it for now

lms/services/canvas_api/client.py
```diff
@@ -209,11 +209,14 @@ def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
...
     @lru_cache
-    def list_files(self, course_id):
+    def list_files(self, course_id, sort="position"):
         """
         Return the list of files for the given `course_id`.
 
         :param course_id: the Canvas course_id of the course to look in
+        :param sort: field to sort by (on Canvas' API side).
+            Defaults to "position" which is an undocumented option but that it should be the most stable of the options available as it sorts by both "position" and "name" on Canvas' side.
+            https://github.com/instructure/canvas-lms/blob/master/app/controllers/files_controller.rb#L305
```
seanh, line 219: If we're gonna link to a line of the Canvas source code we should probably use a GitHub permalink as line 305 could shift around as the code changes:

```suggestion
            https://github.com/instructure/canvas-lms/blob/d43feb92d40d2c69684c4536f74dec37992c557a/app/controllers/files_controller.rb#L305
```

seanh: What happens if you send an unknown value to Canvas as the sort parameter? Does the API return an error? Or does it just ignore the parameter? I'm wondering if we'd notice if they removed sort. It sounds like they're only keeping it around for backwards compatibility.

seanh (approved):

lms/services/canvas_api/client.py
marcospri: > What happens if you send an unknown value to Canvas as the sort parameter? Does the API return an error? Or does it just ignore the parameter? I'm wondering if we'd notice if they removed sort. It sounds like they're only keeping it around for backwards compatibility.

Currently it defaults to sorting by name, the same default as for no parameter at all.

Issue #2855, opened 2021-06-21
Spike: One Drive integration


marcospri:
Investigate integration options with One Drive (API, authentication) 

- Need to create OneDrive app? Credentials.
- Use OneDrive provided file picker vs building our own.
- Store a permanent URL provided by drive vs storing and file ID or similar an generate the URL on "runtime"

marcospri:
Approach taken on the PoC PR  https://github.com/hypothesis/lms/pull/2881

In summary:

- Assignments are configured using OneDrive's file picker 
- We store a `onedrive://FILE_ID` on the assignment "document_url" 
- When we launch the assignment, we follow the oauth2 flow for the user to get a valid token, with that, we use OneDrive's API to get a download URL for the file based on its ID (we never store that URL as is temporary, we get a new one per launch)


# File picker

OneDrive offerts two different file pickers a "JavaScript" one:

https://docs.microsoft.com/es-es/onedrive/developer/controls/file-pickers/js-v72/?view=odsp-graph-online

and a React one:

https://docs.microsoft.com/es-es/onedrive/developer/controls/file-pickers/react/?view=odsp-graph-online
https://www.npmjs.com/package/@microsoft/file-browser

the react ones is marked as "Preview" with a `This API is provided as a preview for developers and may change based on feedback that we receive. Do not use this API in a production environment.` notice. See https://github.com/OneDrive/onedrive-api-docs/issues/1089  for status updates on progress and some issues.

With that I went for the JavaScript one which integrates very similarly to Google's Drive's. I think we rather use the old one and eventually upgrade rather than pick the unstable one and have to do a downgrade in a hurry if we find issues.


## File picker authentication

The file picker receives the `client_id` via the back end _js_config. It also needs to "land" in a page where its own code is present  (file picker loaded in SPA -> `filepicker.open()` -> file picker pop up with file picker object loaded) so the backend also provides an URL that just renders something like `<script>onedrive.js</script>`

# Backend

We can reuse most of the oauth infrastructure and use it in a very similar manner to canvas and blackboard. The only problem is the need to start storing more that one oauth2 token per user (eg canvas and onedrive ones)

# Proposed tasks moving forward
- Create a onedrive app using a "corporate" hypothesis account (or maybe one drive allows multiple users to access the control panel?)
- Include new app credentials on "dev data"
- Allow multiple oauth keys per user  (new columns, migration, data migration...)
- Frontend work to integrate the file picker
[...]

PR #2881, opened 2021-06-25
onedrive files integration POC

marcospri:
https://github.com/hypothesis/lms/issues/2855

This approach is broken as implemented in this PR:
- Teacher allows our app to access it own files, we get a file ID using the API
- Student has access to the "file ID" but that only exists in the teacher account (from a permissions point of view).

To fix it:

- Using personal accounts, the ones I'm testing with, the student will need to have access to the file (shared with him or a group the student belongs to) and we'd get the file using https://docs.microsoft.com/en-us/onedrive/developer/rest-api/api/drive_sharedwithme?view=odsp-graph-online the "share with me drive.

- Using corporate/organization accounts, the we'd need both a drive_id and file_id to be stored on the assignment and we'd get that from the student with `/drives/{drive-id}/items/{item-id}`


lms/migrations/versions/9ee0679761f0_.py
```diff
@@ -0,0 +1,37 @@
+"""
```
marcospri, line 1: Migrations are not great here, they need to be broken into different steps to deploy

lms/models/oauth2_token.py
```diff
@@ -18,7 +18,7 @@ class OAuth2Token(BASE):
     """
 
     __tablename__ = "oauth2_token"
-    __table_args__ = (sa.UniqueConstraint("user_id", "consumer_key"),)
+    __table_args__ = (sa.UniqueConstraint("user_id", "consumer_key", "api_host"),)
```
marcospri, line 21: Again, migration for this would need to be deployed in stages

lms/config/__init__.py
```diff
@@ -22,6 +22,8 @@ def configure(settings):
         "google_client_id": sg.get("GOOGLE_CLIENT_ID"),
         "google_developer_key": sg.get("GOOGLE_DEVELOPER_KEY"),
         "google_app_id": sg.get("GOOGLE_APP_ID"),
+        "onedrive_client_id": sg.get("ONEDRIVE_CLIENT_ID"),
+        "onedrive_secret": sg.get("ONEDRIVE_CLIENT_SECRET"),
```
seanh, line 26: So this is a single client ID and secret that we'll use for all customers? We don't need different client IDs and secrets for different users?

lms/resources/_js_config.py
```diff
@@ -357,6 +370,14 @@ def _blackboard_files_enabled(self):
         """
         return self._application_instance().settings.get("blackboard", "files_enabled")
 
+    def _onedrive_files_enabled(self):
+        """
+        Return True if the OneDrive Files API is enabled for this request.
+
+        :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
+        """
+        return self._application_instance().settings.get("onedrive", "files_enabled")
```
seanh, line 379: Ok, so we're enabling the OneDrive feature per application instance. That makes sense because we can enable it only for our test application instances during development, and then only for certain customers. Eventually we might want to enable it by default

lms/routes.py
```diff
@@ -45,6 +45,21 @@ def includeme(config):
...
+
+    config.add_route(
+        "onedrive.oauth.authorize",
+        "/api/onedrive/oauth/authorize",
+    )
+    config.add_route(
+        "onedrive.oauth.callback",
+        "/api/onedrive/oauth/callback",
+    )
+    config.add_route("onedrive.files.via_url", "/api/onedrive/via_url")
```
seanh, line 61: Ok, so there are two authorization routes: one for the file picker and one for getting file URLs from the API. And there's no list-files route since we're not using our own file picker

lms/config/__init__.py
marcospri: Yeah it will be one key per "onedrive application" so we could use one for all schools the "Hypothesis One Drive" app or create one app per school if that made sense? I don't see an immediate benefit of multiple apps.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -219,6 +262,16 @@ export default function ContentSelector({
               Select PDF from Google Drive
             </LabeledButton>
           )}
+          {oneDriveFilesEnabled && (
+            <LabeledButton
+              onClick={showOneDrivePicker}
+              variant="primary"
+              data-testid="one-drive-button"
+            >
+              Select PDF from One Drive
```
seanh, line 271: FYI it's OneDrive not One Drive. We might want to actually say Microsoft OneDrive as well

seanh:
One potential issue with this: what happens when a teacher creates and launches an assignment (so the teacher gives Hypothesis access to their OneDrive files), and then later _students_ come along and try to launch the assignment? Presumably we'll show our <kbd>Authorize Hypothesis</kbd> button and when the student clicks it we'll ask OneDrive for access to _the student's_ OneDrive account. Is that going to work? Will the teacher need to have shared the files with the students in OneDrive first, somehow?

seanh:
If you aren't logged in to OneDrive the popup window asks you to log in :+1:

sign-in

seanh:
There are some issues with OneDrive's UI for selecting a file, but I don't know if there's anything we can do about it:

1. You actually have to click on the circle to the top-right of the file to select it. Just clicking on the file doesn't do anything
2. The wording that the UI uses is about "Opening" a file not choosing or selecting

Peek 2021-06-28 13-25

seanh:
OneDrive will ask the user to grant Hypothesis permission to "Read your profile" and "Open OneDrive files". I don't know if there's anything we can do to reduce the permissions that we ask for?

Screenshot from 2021-06-28 13-22-07

seanh:
When a teacher _launches_ an assignment OneDrive asks them to grant Hypothesis permission to "maintain access to data you have given access to ... even when you are not currently using the app". Not sure if I can also ask for this permission in the first place, when creating an assignment? To avoid the second permissions ask for teachers.

Screenshot from 2021-06-28 13-26-01

seanh:
When a _student_ launches an assignment they haven't previously granted Hypothesis access to any OneDrive permissions, so they get this dialog instead. It asks the user to grant Hypothesis permission to "open OneDrive files" and to "maintain access to date you have given access to ... even when you are not currently using the app"

Screenshot from 2021-06-28 13-37-30

seanh:
It looks like every time a teacher creates an assignment, even if they've already granted us access to OneDrive, it will show them this dialog asking "Are you trying to sign in to Hypothesis"

Screenshot from 2021-06-28 13-26-50

seanh:
OneDrive support brings up a new UX issue that we haven't run into before. Currently this will only happen in Canvas, not in other LMS's: OneDrive is the first time that we need access to two different APIs to launch a single assignment. We need access to the Canvas API to use the sections or groups API. And we need access to the OneDrive API to download the file. This causes us to show the exact same dialog asking the user to authorize us twice, which is confusing.

Peek 2021-06-28 13-42

seanh:
Closing this proof-of-concept: its done its job

Issue #2860, opened 2021-06-22
Wrong docs link when a Blackboard courses has no files


seanh:
I haven't created a Blackboard test course with no files so the easiest way to reproduce this issue is to hack the API to pretend that the course has no files:

```diff
diff --git a/lms/views/api/blackboard/files.py b/lms/views/api/blackboard/files.py
index 1bb65bd8..72fe7971 100644
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -14,6 +14,7 @@ class BlackboardFilesAPIViews:
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
     def list_files(self):
         """Return the list of files in the given course."""
+        return []
         return self.blackboard_api_client.list_files(
             self.request.matchdict["course_id"]
         )
```

Log in to our test Blackboard site as a teacher, create a new assignment **in a course that does not have any PDFs uploaded to Blackboard**, launch the assignment, and click the <kbd>Select PDF from Blackboard</kbd> button. You'll see this:

![Screenshot from 2021-06-22 11-45-12](https://user-images.githubusercontent.com/22498/122911545-519c3180-d34f-11eb-95b5-34215bcc7ba8.png)

The **Upload some files to the course** link links to Canvas's docs. In Blackboard's case this needs to link to the appropriate Blackboard docs instead.

PR #2882, opened 2021-06-25
Fix blackboard content link

LMS007:
Make the no file message href dynamic 

Change the LMSFilePicker to render a different documentation help link when no files are returned for Blackboard vs Canvas.

------

Update

I have refactored this PR to no longer use a provider and instead take a `missingFilesHelpLink` prop from ContentSelector.
Secondly, we have decided the 0 way to link to Blackboard docs is to link to an internal alias we maintain that has external links for file upload documentation for both ultra and classic version so blackboard. Currently there is no way to differentiate between these versions from the perspective of the client or config.

Slack discussion here
https://hypothes-is.slack.com/archives/C2BLQDKHA/p1625075688421600

Fixes https://github.com/hypothesis/lms/issues/2860

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -135,6 +135,7 @@ export default function ContentSelector({
           listFilesApi={listFilesApi}
           onCancel={cancelDialog}
           onSelectFile={selectCanvasFile}
+          missingFilesHelpLink="https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-upload-a-file-to-a-course/ta-p/618"
         />
       );
       break;
@@ -145,6 +146,9 @@ export default function ContentSelector({
           listFilesApi={blackboardListFilesApi}
           onCancel={cancelDialog}
           onSelectFile={selectBlackboardFile}
+          // An alias we maintain that provides multiple external documentation links for
+          // different versions of Blackboard (Classic vs. Ultra)
+          missingFilesHelpLink={'https://web.hypothes.is/help/bb-files'}
         />
       );
       break;
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.js
@@ -15,6 +15,31 @@ import FileList from './FileList';
  */
 
 /**
+ * @typedef NoFilesMessageProps
+ * @prop {string} href - Helpful documentation URL to link to
+ */
+
+/**
+ * Renders a helpful message with a link to documentation when there are no
+ * uploaded files.
+ *
+ * @param {NoFilesMessageProps} props
+ */
+function NoFilesMessage({ href }) {
+  return (
+    <div className="FileList__no-files-message">
+      <p>
+        There are no PDFs in this course.{' '}
+        <a href={href} target="_blank" rel="noreferrer">
+          Upload some files to the course
+        </a>{' '}
+        and try again.
+      </p>
+    </div>
+  );
+}
+
+/**
  * @typedef LMSFilePickerProps
  * @prop {string} authToken - Auth token for use in calls to the backend
  * @prop {APICallInfo} listFilesApi -
@@ -22,6 +47,9 @@ import FileList from './FileList';
  * @prop {() => any} onCancel - Callback invoked if the user cancels file selection
  * @prop {(f: File) => any} onSelectFile -
  *   Callback invoked with the metadata of the selected file if the user makes a selection
+ * @prop {string} missingFilesHelpLink - A helpful URL to documentation that explains
+ *   how to upload files to an LMS such as Canvas or Blackboard. This link is only shown
+ *   when the API call to returns available files returns an empty list.
  */
 
 /**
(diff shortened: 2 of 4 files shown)


lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -46,6 +62,9 @@ describe('LMSFilePicker', () => {
         apiCall: fakeApiCall,
       },
       './Dialog': FakeDialog,
+      // Don't mock <FileList> because <NoFiles> requires
+      // it to render for code coverage.
+      './FileList': FileList,
```
LMS007, line 67: Some hackery here, because `<NoFiles>` gets passed as a prop, it won't render in this tests (e.g. not full codecov) unless we don't mock `<FileList>`, so this simply forced FileList to render too. 

I also felt `NoFiles` should **continue** being passed as prop and not a child component because it's not a central child component of `FileLIst`.

robertknight (changes requested):
The design on master is that `LMSFilePicker` is agnostic of LMS-specific details and app configuration, which are handled by the parent `ContentSelector`. The current status on this branch is that `LMSFilePicker` handles some of the LMS-specific details (the help link) whereas the parent `ContentSelector` handles other details (which API method to call to list the files). This is an inconsistency that we should avoid.

If the UI of the dialog needs to diverge significantly depending on the LMS then it will probably make sense to change the responsibilities of the components. However for customizing a single help link I think it would be better to extend the current design where LMS-specific knowledge remains in `ContentSelector`. For example the LMS-specific links can be passed in as a prop to `LMSFilePicker` by `ContentSelector`, which already has different branches that render `LMSFilePicker` with props that depend on the LMS.

LMS007:
Updated PR:

Rob, good suggestions about the provider and just using ContentSelector instead, that simplifies the code a bit. 

I also changed the bb help link to an internal WP alias that we'll maintain. Nate & Matt are working on adding it. I'll probably hold off on merging this PR until that alias is up and working. I also confirmed with Sean that there is no way to differentiate between ultra and classic FWIW.

internal doc alias:
https://web.hypothes.is/help/bb-files

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -339,4 +343,21 @@ describe('LMSFilePicker', () => {
...
+
+  describe('when no files are provided', () => {
+    beforeEach(() => {
+      fakeApiCall = sinon.stub().resolves([]); // no files returned
+    });
+
+    it('renders no file message with a help link', async () => {
+      const wrapper = renderFilePicker();
+      await fakeApiCall.returnValues[0];
+      wrapper.update();
```
robertknight, line 355: This `await ...; wrapper.update();` pair of statements follows the example of other tests, so I think it's OK to use it for the moment. This pattern is a little brittle though in that it will break if `LMSFilePicker` ever does any other additional async steps after the file list API call returns before it re-renders. A better solution is to wait for the `FileList` to be rendered in a non-loading state using the `waitForElement` helper:

```diff
diff --git a/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js b/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
index 55f65b72..4218e475 100644
--- a/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
@@ -7,6 +7,7 @@ import ErrorDisplay from '../ErrorDisplay';
 import FileList from '../FileList';
 import LMSFilePicker, { $imports } from '../LMSFilePicker';
 import mockImportedComponents from '../../../test-util/mock-imported-components';
+import { waitForElement } from '../../../test-util/wait';
 
 describe('LMSFilePicker', () => {
   // eslint-disable-next-line react/prop-types
@@ -351,9 +352,7 @@ describe('LMSFilePicker', () => {
 
     it('renders no file message with a help link', async () => {
       const wrapper = renderFilePicker();
-      await fakeApiCall.returnValues[0];
-      wrapper.update();
-      const fileList = wrapper.find('FileList');
+      const fileList = await waitForElement(wrapper, 'FileList');
       assert.equal(
         fileList.prop('noFilesMessage').props.href,
         'https://fake_help_link'
```

lms/static/scripts/frontend_apps/components/LMSFilePicker.js
```diff
@@ -14,6 +14,31 @@ import FileList from './FileList';
  * @typedef {import("../config").APICallInfo} APICallInfo
  */
 
+/**
+ * @typedef NoFilesProps
+ * @prop {string} href - Helpful documentation url to link to
```
robertknight, line 19:
```suggestion
 * @prop {string} href - Helpful documentation URL to link to
```

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -145,6 +148,9 @@ export default function ContentSelector({
           listFilesApi={blackboardListFilesApi}
           onCancel={cancelDialog}
           onSelectFile={selectBlackboardFile}
+          // An alias we maintain that provides multiple external documentation links for
+          // different versions of Blackboard (classic vs. ultra)
```
robertknight, line 152:
```suggestion
          // different versions of Blackboard (Classic vs. Ultra)
```

This is to make it clear that these terms are Blackboard's official names for these versions rather than just descriptions that we use.

robertknight (approved):
The code looks good. I had some minor suggestions regarding the name of the `NoFiles` component and a few other details. I had a query regarding the content that https://web.hypothes.is/help/bb-files points to. Currently this is a general page about using Hypothesis in Blackboard with no mention of Blackboard files. What is the plan regarding making this link point to something which is more specific to the issue here (eg. a page specifically about Blackboard Files) and is there an issue for that?

LMS007:
> ral page about using Hypothesis in Blackboard with no mention of Blackboard files. What is the pla

https://web.hypothes.is/help/bb-files  page is just an alias that needs to point to a new content page which currently does not exist. I will hold off on merging this PR until that page is done. I'll follow up with support on a timeframe.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
LMS007: I like waitForElement a lot better, much easier  as it takes care of those details which are hard to remember. I'll switch all the tests in this file which could use that call, perhaps in a second commit.

PR #2889, opened 2021-06-28
Jenkinsfile update - Deploy to a second region

indigobravo:
The LMS Jenkinfile has been updated to support deploying to a second
region in parallel with us-west-1.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,14 +49,26 @@ node {
 onlyOnMaster {
     milestone()
     stage("qa deploy") {
-        deployApp(image: img, app: "lms", env: "qa")
+        deployApp(image: img, app: "lms", env: "qa", region: "us-west-1")
     }
 
     milestone()
-    stage("prod deploy") {
-        input(message: "Deploy to prod?")
-        milestone()
-        deployApp(image: img, app: "lms", env: "prod")
+    stage("approval") {
+        input(message: "Proceed to production deploy?")
+    }
+
+    milestone()
+    stage("prod Deploy") {
+        parallel(
+            us: {
+                deployApp(image: img, app: "lms", env: "prod", region: "us-west-1")
+            },
+            ca: {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -49,14 +49,27 @@ node {
 onlyOnMaster {
     milestone()
     stage("qa deploy") {
-        deployApp(image: img, app: "lms", env: "qa")
+        deployApp(image: img, app: "lms", env: "qa", region: "us-west-1")
```
indigobravo, line 52: Pass region value to `deployApp`.

Jenkinsfile
```diff
@@ -49,14 +49,27 @@ node {
...
     }
 
     milestone()
-    stage("prod deploy") {
-        input(message: "Deploy to prod?")
-        milestone()
-        deployApp(image: img, app: "lms", env: "prod")
+    stage("approval") {
+        input(message: "Proceed to production deploy?")
+    }
```
indigobravo, line 58: Introduced dedicated `approval` stage. This has been done for Via already, and I think the change is beneficial. It makes what is going on, and user input requirements more obvious:

Screenshot 2021-06-28 at 11 00 52

Jenkinsfile
```diff
@@ -49,14 +49,27 @@ node {
...
+            parallel(
+                us: {
+                    sleep 2
+                    deployApp(image: img, app: "lms", env: "prod", region: "us-west-1")
+                },
+                ca: {
+                    deployApp(image: img, app: "lms-ca", env: "prod", region: "ca-central-1")
+                }
+            )
+        }
```
indigobravo, line 72: A stage dedicated to production deploys:

1. A `parallel` block has been introduced. When called, it will make two calls to `deployApp`. The first deals with deploying `us-west-1` and the second deals with deploying `ca-central-1`.
2. The `sleep 2` command in the `us` block is required to insert a small pause in the calls to `deployApp`. Without a pause it can end up running the first job twice.

robertknight (approved):

PR #2891, opened 2021-06-28
Onedrive poc share link

marcospri:
Different approach than #2881 

- Here we don't need to authorize the students 
- It will work for any file selected by the teacher 

The process is:
- File picker in "share" mode will automatically create a shared link with our app
- From the shared link (which will render a HTML frame around the document with all onedrive's UI) we generate a download URL
- We store that URL on the assignment 
- Launches use the stored URL verbatim 

As for permissions, "share" implies write (to create the link) so the authorization prompt looks very similar to the one in the other approach but it's only seen once by the teacher (while configuring the assignment) and never while launching it.

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -46,6 +46,11 @@ def includeme(config):
     )
 
     config.add_route(
+        "onedrive.filepicker.authorize",
+        "/api/onedrive/filepicker/authorize",
+    )
+
+    config.add_route(
         "canvas_api.oauth.authorize",
         "/api/canvas/oauth/authorize",
         factory="lms.resources.OAuth2RedirectResource",
--- a/lms/static/scripts/frontend_apps/utils/onedrive-api-client.js
+++ b/lms/static/scripts/frontend_apps/utils/onedrive-api-client.js
@@ -1,0 +1,21 @@
+async function loadOneDriveAPI() {
+  if (window.odapi) {
+    return window.odapi;
+  }
+
+  return new Promise((resolve, reject) => {
+    const oneDriveScript = document.createElement('script');
+    oneDriveScript.src = 'https://js.live.net/v7.2/OneDrive.js';
+    oneDriveScript.onload = () => {
+      resolve(window.odapi);
+    };
+    oneDriveScript.onerror = () => {
+      reject(new Error('Failed to load One Drive API client'));
+    };
+    document.body.appendChild(oneDriveScript);
+  });
+}
+
+// Separate function declaration from export to work around
+// https://github.com/robertknight/babel-plugin-mockable-imports/issues/9.
+export { loadOneDriveAPI };
--- a/lms/templates/onedrive.html.jinja2
+++ b/lms/templates/onedrive.html.jinja2
@@ -1,0 +1,15 @@
+{% extends "lms:templates/base.html.jinja2" %}
+
+{% block content %}
+  <div id="app"></div>
+{% endblock %}
+
+{% block styles %}
+  {% for url in  asset_urls("frontend_apps_css") %}
+    <link rel="stylesheet" href="{{ url }}">
+  {% endfor %}
+{% endblock %}
+
+{% block scripts %}
+  <script type="text/javascript" src="https://js.live.net/v7.2/OneDrive.js"></script>
+{% endblock %}
--- a/lms/templates/onedrive.html.jinja2
+++ b/lms/templates/onedrive.html.jinja2
@@ -1,13 +1,9 @@
 {% extends "lms:templates/base.html.jinja2" %}
 
 {% block content %}
-  <div id="app"></div>
 {% endblock %}
 
 {% block styles %}
-  {% for url in  asset_urls("frontend_apps_css") %}
-    <link rel="stylesheet" href="{{ url }}">
-  {% endfor %}
 {% endblock %}
 
 {% block scripts %}
--- a/lms/views/api/onedrive.py
+++ b/lms/views/api/onedrive.py
@@ -1,0 +1,10 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    request_method="GET",
+    route_name="onedrive.filepicker.authorize",
+    renderer="lms:templates/onedrive.html.jinja2",
+)
+def authorize(request):
+    return {}
--- a/lms/resources/_js_config/file_picker_config.py
+++ b/lms/resources/_js_config/file_picker_config.py
@@ -82,6 +82,7 @@ class FilePickerConfig:
         return {
             "enabled": True,
             "clientId": request.registry.settings["onedrive_client_id"],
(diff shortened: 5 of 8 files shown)


seanh:
As a teacher, the first time I use the <kbd>Select PDF from OneDrive</kbd> button, I'll see this dialog which unfortunately asks for very broad-sounding permissions to <q>read, create, update and delete all OneDrive files</q>. @marcospri Is there nothing we can do to reduce the permissions?

And I'm guessing we'll be able to register our app somehow to get rid of the **unverified**.

Screenshot from 2021-06-28 17-23-14

After they've granted these permissions once, each subsequent time they create or edit an assignment the teacher will see this:

Screenshot from 2021-06-28 17-25-17

**Students** will not see any dialogs from Microsoft / OneDrive, it'll "just work" for them.

seanh:
Users can see in the OneDrive UI that share links have been created. The file on the left is one that a Hypothesis assignment has been created for:

Screenshot from 2021-06-28 17-29-55

If they click on the file the user can get more information about the share link that has been created. @marcospri It looks like we're creating links that allow user to edit files? Is it possible to create read-only links?

Screenshot from 2021-06-28 17-31-20

marcospri:
> If they click on the file the user can get more information about the share link that has been created. @marcospri It looks like we're creating links that allow user to edit files? Is it possible to create read-only links?

I think that's the dialog to create new links, not the view of the existing ones. The ones created thru the picker are set for just viewing. If you right click the file -> details -> manage access you can see the existing ones.

Here I created two links, one for view and one for editing using onesdrive the UI. Also created two different ones thru the picker (one for viewing and changed this line https://github.com/hypothesis/lms/pull/2891/files#diff-b549e14815b20734ec3e4909894de1d58f720ea20f7ba5cfb07414e1b33af6ddR15 to allow for edits in the second one)

The ones created with the UI give the hint about the permissions level. The ones through the API/Picker do not.

Screenshot from 2021-06-29 09-09-53

marcospri:
> As a teacher, the first time I use the Select PDF from OneDrive button, I'll see this dialog which unfortunately asks for very broad-sounding permissions to read, create, update and delete all OneDrive files. @marcospri Is there nothing we can do to reduce the permissions?
> 

No, I'm afraid we need the "Read.Write.All" permission to be able to create links and there's no more granularity than that (ie it would be nice to not display `delete` or `update` there but `Write` involves that) 

As there no scope distinction between been able to create share links for read or write we effectlyy can write / delete all files we can create  a link to.

More information in this issue: 

https://github.com/OneDrive/onedrive-api-docs/issues/630
https://github.com/OneDrive/onedrive-api-docs/issues/630#issuecomment-546091417

marcospri:
> 
> And I'm guessing we'll be able to register our app somehow to get rid of the **unverified**.
> 

Looks like we'll have to verify our domain 

Screenshot from 2021-06-29 09-28-56

seanh:
Ah, okay I see now. The OneDrive UI is a bit confusing which is unfortunate. So here's what I see when I view the files in OneDrive's web interface after a Hypothesis assignment has been created for one of the files, there's an icon that indicates that the file has been shared:

Screenshot from 2021-06-29 09-58-21

To see what share links have been created for a file you need to navigate to that file's **Manage Access** sidebar like this:

Peek 2021-06-29 10-00

Once you get to the **Manage Access** sidebar you can see that a share link has been created "for Hypothesis", but as far as I can tell there's no way to see whether the link is temporary or permanent (it's permanent) or whether it's read-only or read-write (it's read-only):

Screenshot from 2021-06-29 10-01-20

If a user who is **not** logged in to OneDrive gets this "for Hypothesis" share link and opens it in a browser they get this read-only view of the file:

Screenshot from 2021-06-29 10-02-21

A user who **is** logged in to their own OneDrive account gets the same read-only view, only without the banner about signing up to OneDrive.

seanh:
If a teacher creates more than one assignment for the same file it sames to re-use the same share link, it doesn't create multiple share links.

seanh:
@marcospri One thing we'll need to make sure to cover in the production version: the user can delete the share link that Hypothesis created. We'll have to see if we can detect this situation when launching assignments and show a helpful error message

seanh:
Closing this proof-of-concept as I think its done its job

lms/resources/_js_config.py
```diff
@@ -231,6 +231,11 @@ def google_picker_origin():
                     ),
                 },
             },
+            "onedrive": {
+                "enabled": self._onedrive_files_enabled(),
+                "clientId": self._request.registry.settings["onedrive_client_id"],
+                "redirectURI": self._request.route_url("onedrive.filepicker.authorize"),
```
seanh, line 237: @jon-betts I don't think we want the `redirectURI` here

lms/routes.py
```diff
@@ -45,6 +45,11 @@ def includeme(config):
         "blackboard_api.files.via_url", "/api/blackboard/courses/{course_id}/via_url"
     )
 
+    config.add_route(
+        "onedrive.filepicker.authorize",
+        "/api/onedrive/filepicker/authorize",
+    )
```
seanh, line 51: @jon-betts I don't think we need this

lms/views/api/onedrive.py
```diff
@@ -0,0 +1,10 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    request_method="GET",
+    route_name="onedrive.filepicker.authorize",
+    renderer="lms:templates/onedrive.html.jinja2",
+)
+def authorize(request):
+    return {}
```
seanh, line 10: @jon-betts I don't think we need this

lms/static/scripts/frontend_apps/utils/onedrive-api-client.js
```diff
@@ -0,0 +1,21 @@
...
+    };
+    oneDriveScript.onerror = () => {
+      reject(new Error('Failed to load One Drive API client'));
+    };
+    document.body.appendChild(oneDriveScript);
+  });
+}
+
+// Separate function declaration from export to work around
+// https://github.com/robertknight/babel-plugin-mockable-imports/issues/9.
```
robertknight, line 20: This issue shouldn't be applicable any more because we're not using the `transform-async-to-promises` Babel plugin that this issue refers to.

lms/templates/onedrive.html.jinja2
```diff
@@ -0,0 +1,15 @@
...
+
+{% block content %}
+  <div id="app"></div>
+{% endblock %}
+
+{% block styles %}
+  {% for url in  asset_urls("frontend_apps_css") %}
+    <link rel="stylesheet" href="{{ url }}">
+  {% endfor %}
+{% endblock %}
```
robertknight, line 11: Based on a Slack discussion, my understanding is that the only content which is actually needed in this file is the OneDrive `<script>` at the bottom. The `styles` and `content` block can be omitted or made empty.

PR #2892, opened 2021-06-28
Add FileService

seanh:
This will be used by the Canvas service in future to look up matching files in the DB.

While `FileService` might look a little silly right now (it's just a single `get()` method) it actually does encapsulate a non-trivial amount of code and tests, keeping that complexity out of the higher-level business logic that's going to call this service in a future PR. It also provides a place for us to put other file-related methods in future if we ever need them (listing all the files in a course, searching for files by name and size, etc). It'd be the obvious place for code for bulk upserting files too.

### DIFF
--- a/lms/services/file.py
+++ b/lms/services/file.py
@@ -1,0 +1,20 @@
+from lms.models import File
+
+
+class FileService:
+    def __init__(self, db):
+        self._db = db
+
+    def get(self, application_instance, lms_id, type_):
+        """Return the file with application_instance, lms_id and type_ or None."""
+        return (
+            self._db.query(File)
+            .filter_by(
+                application_instance=application_instance, lms_id=lms_id, type=type_
+            )
+            .one_or_none()
+        )
+
+
+def factory(_context, request):
+    return FileService(db=request.db)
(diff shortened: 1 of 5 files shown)


lms/services/file.py
```diff
@@ -0,0 +1,23 @@
...
+    @property
+    def _query(self):
+        return self._db.query(File).filter_by(
+            application_instance=self._application_instance
+        )
+
+
+def factory(_context, request):
+    return FileService(
+        application_instance=request.find_service(name="application_instance").get(),
```
marcospri, line 21: This is a pattern we use in most services now but I think is creates some problems to just avoiding passing one parameter around. 

I much rather do `File.service.get(application_instance, lms_id, type)`  especially if we added `application_instance` to request or context instead coupling each service instance to one AI.

If we were to use this service on for example the admin pages we would need to instantiate it manually once per AI which seems odd. Having a separate "AdminPagesFileService" seeps to defeat the purpose of the service in the first place if they are not reusable. 

In any case, none of this is relevant to this PR which just follows the established pattern so don't mind me.

marcospri (approved):

lms/services/file.py
seanh: Ok, I've moved `application_instance` to a `get()` method argument. I think it might actually make the higher-level code slightly nicer too

marcospri (approved):
I think moving the responsibility of getting the AI out of the service looks nicer :+1:

PR #2893, opened 2021-06-28
Add CanvasService.can_see_file_in_course()

seanh:
Change the existing private method `CanvasService._file_in_course()` into a public `CanvasService.can_see_file_in_course()` method.

Some future business logic in `CanvasService` is going to make use of this method, so I'm tidying it up to make it clearer:

* Changed the name of the method.

  The method doesn't tell you whether a course contains a file. It tells you whether the current user can see the file in the course.

* Added type annotations and a docstring

* Reformatted the code to be longer but more readable.

  Previously it wasn't very obvious that calling this method actually makes a call to the list-files API, because that was hidden away at the end of an `any()`. Now it gets its own line.

  Previously both the `file_id` param and the IDs returned by `list_files()` were passed through `str()`, but the `file_id` param is already a string. Changed it to only stringify the ints from the Canvas API, and did this on its own line with a comment so it's clear why it's being done.

  The reformatting also made it clear that there was no test case for when this method returns `False`. That's partially the unnecessarily-dense code formatting messing up coverage (although I'm kind of surprised that branch coverage didn't catch this) and it's partially due to the method not having its own unit tests and only being tested indirectly.

* Made the method public.

  Nothing outside of `CanvasService` is actually going to call this method. But I think it should be public because it is a logically separate method; because it means that external code _could_ call it in future; it encourages writing a good docstring; and it allows the method to have its own direct unit tests. Breaking code up into smaller, separate units instead of larger units where one top-level public thing calls a private thing.

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -26,15 +26,33 @@ class CanvasService:
         """
 
         if check_in_course:
-            if not self._file_in_course(file_id, course_id):
+            if not self.can_see_file_in_course(file_id, course_id):
                 raise CanvasFileNotFoundInCourse(file_id)
 
         return self.api.public_url(file_id)
 
-    def _file_in_course(self, file_id, course_id):
-        return any(
-            str(file_["id"]) == str(file_id) for file_ in self.api.list_files(course_id)
-        )
+    def can_see_file_in_course(self, file_id: str, course_id: str) -> bool:
+        """
+        Return True if the current user can see file_id in course_id.
+
+        Return False if the current user cannot currently see a file with ID
+        file_id in the course with ID course_id. This could be because the file
+        is in another course, because the file was in course_id but has been
+        deleted, or because the file is in course_id but the current user
+        doesn't have permission to see it (for example files that are marked as
+        "unpublished" in Canvas can only be seen by teachers, not students).
+        """
+        files_in_course = self.api.list_files(course_id)
+
+        for file_ in files_in_course:
+
+            # The Canvas API returns file IDs as ints but the file_id param
+            # that this method receives (from our proxy API) is a string.
+            # Convert ints to strings so that we can compare them.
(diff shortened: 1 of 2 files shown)


lms/services/canvas.py
```diff
@@ -26,15 +26,35 @@ def public_url_for_file(self, file_id, course_id, check_in_course=False):
...
+    def can_see_file_in_course(self, file_id: str, course_id: str) -> bool:
+        """
+        Return True if the current user can see file_id in course_id.
+
+        Return False if the current user cannot currently see a file with ID
+        file_id in the course with ID course_id. This could be because the file
+        is in another course, because the file was in course_id but has been
+        deleted, or because the file is in course_id but the current user
+        doesn't have permission to see it (for example files that are marked as
+        "unpublished" in Canvas can only be seen by teachers, not students).
```
marcospri, line 43: :+1:  Nice, thanks for the explanation here

lms/services/canvas.py
```diff
@@ -26,15 +26,35 @@ def public_url_for_file(self, file_id, course_id, check_in_course=False):
...
+
+        for file_ in files_in_course:
+
+            # The Canvas API returns file IDs as ints but the file_id param
+            # that this method receives (from our proxy API) is a string.
+            # Convert ints to strings so that we can compare them.
+            file_["id"] = str(file_["id"])
+
+            if file_["id"] == file_id:
+                return True
```
marcospri, line 55:
```suggestion

            # The Canvas API returns file IDs as ints but the file_id param
            # that this method receives (from our proxy API) is a string.
            # Convert ints to strings so that we can compare them.
            if str(file_["id"]) == file_id:
                return True
```

I'd have done the `str()` on the conditional itself rather than overriding the dictionary (or creating a `str_file_id` or similar) the comment applies nicer to the conditional I think.

marcospri (approved):

PR #2894, opened 2021-06-28
Add CanvasService.find_matching_file_in_course()

seanh:
A future PR is going to add business logic to `CanvasService.public_url_for_file()` that will call this `CanvasService.find_matching_file_in_course()`.

I've made `find_matching_file_in_course()` a separate public method because it's a logically separate unit that *could* be used by other code outside of `CanvasService` even though it isn't going to be right now, and because having it be a separate public method encourages a good method name, good docstring, and its own separate unit tests. Small separate units instead of one big unit where a top-level public method calls lots of private methods.

When searching for a matching file in a course we do this by calling the Canvas API to get a list of all the files in the course, rather than looking in our `models.File` table in the DB, because:

* It always gets the most up to date list of files from the course, doesn't return false positives or false negatives
* It only returns the files that the current user can see, which is what we want. For example students can't see "unpublished" files but teachers can. Our `models.File` table doesn't know about this

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,3 +1,6 @@
+from typing import Optional
+
+from lms import models
 from lms.services.exceptions import CanvasFileNotFoundInCourse
 
 
@@ -53,6 +56,29 @@ class CanvasService:
                 return True
 
         return False
+
+    def find_matching_file_in_course(
+        self, course_id: str, file_: models.File
+    ) -> Optional[str]:
+        """
+        Return the ID of a file in course_id that matches file_.
+
+        Search for a file that the current Canvas user can see in course_id and
+        that matches the given file_ (same filename and size) and return the
+        matching file's ID.
+
+        Return None if no matching file is found.
+        """
+        file_dicts = self.api.list_files(course_id)
+
+        for file_dict in file_dicts:
+            display_name = file_dict["display_name"]
+            size = file_dict["size"]
+
+            if display_name == file_.name and size == file_.size:
+                return str(file_dict["id"])
+
+        return None
 
 
 def factory(_context, request):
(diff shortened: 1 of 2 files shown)


tests/factories/file.py
```diff
@@ -0,0 +1,16 @@
...
+File = make_factory(
+    models.File,
+    FACTORY_CLASS=SQLAlchemyModelFactory,
+    application_instance=SubFactory(ApplicationInstance),
+    type="canvas_file",
+    lms_id=Sequence(lambda n: f"{n}"),
+    course_id=Faker("random_int"),
+    name=Faker("word"),
+    size=Faker("random_int"),
+)
```
seanh, line 16: Same factory as added by an earlier PR. Once the earlier PR is merged and this PR is rebased, this will disappear

lms/services/canvas.py
```diff
@@ -56,6 +56,27 @@ def can_see_file_in_course(self, file_id: str, course_id: str) -> bool:
 
         return False
 
+    def find_matching_file_in_course(self, course_id: str, file_: "models.File") -> str:
```
seanh, line 59: Is `"models.File"` the right way to do the type annotation here?

marcospri: Not sure, I think you are expected to import `File` and use it here.

I only see this reference to using strings for thing that would be evaluable once the module is loaded: https://www.python.org/dev/peps/pep-0484/#forward-references

This is also possible:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from models import File
```

but ugly, if a module uses a type I think it's fine to import it.

marcospri: You could also take `file_name` and `file_size` instead of the full `File` but I don't think the importing or not importing the type should drive that decision.

marcospri: The function can return None so something like:

```suggestion
    def find_matching_file_in_course(self, course_id: str, file_: "models.File") -> Optional[str]:
```
and from typing import Optional at the top.

seanh: Done, it's now `def find_matching_file_in_course(self, course_id: str, file_: models.File) -> Optional[str]:`

marcospri (approved):

Issue #2799, opened 2021-06-09
Possible race condition with asset serving during deployments


robertknight:
When a new deployment happens we do a rolling update with an additional batch. I need to confirm but if it is possible that during a deployment a page load can be serviced by both the old and new versions of the app then we could have a situation where an old asset gets cached under a new URL. This might explain an issue [seen by a user](https://hypothes-is.slack.com/archives/C2BLQDKHA/p1623242191165700) where we deployed a bug fix but the user still didn't see the fix some hours later.

The problem scenario is:

1. LMS page request hits instance running new version of app. This serves up a response with new asset URLs (ie. that have the latest cache-busting query string at the end)
2. Browser requests asset with new URL. Cloudflare connects to LMS app and hits instance running old version of app. Since the app ignores the query string when serving the asset, it might still serve the request even if the query string is wrong.
3. Cloudflare edge location caches the old version of the asset under the new URL

To re-iterate, the main condition we need to check here is whether the current deployment process makes it possible that during a single page load requests may go to both the old and new versions of the app during the deployment window.

If this problem does exist then it could affect other apps too. Usage is way up over past years, so it is possible that the issue has long existed but simply been rare.

robertknight:
A change we might want to make regardless of whether our current deployment setup has this problem is to actually check the query string in asset requests and serve up an error.

robertknight:
I've started some initial research on this. Some notes on options for avoiding mismatches between the version of the frontend application and the version of the backend:

Our Elastic Beanstalk environments are configured to use an AWS Application Load Balancer to route requests to instances. ALB has support for sticky sessions, which is one way that we could ensure that requests from the frontend to backend after the initial page load always hit the same version of the application: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/sticky-sessions.html. Sticky sessions rely on cookies and adds some complexity to reasoning about how requests are handled.

[...]

PR #2897, opened 2021-06-29
Verify cache buster in static asset requests

robertknight:
Static asset URLs have a cache busting query string at the end (eg. `/assets/scripts/frontend_apps.bundle.js?1234`). The query string will depend on the version of the application. This query string is used when generating asset URLs for inclusion in HTML responses, but not checked when serving requests. As a result, if a client received an HTML response from version X of the application and then requested the asset which was handled by version Y of the application, they could end up receiving version Y of the asset but cache it under the URL for version X.

This PR improves the situation by checking the cache busting query param in asset requests and returning a 404 response if it is wrong. The first commit removes an unnecessary CORS-related decorator which was copied over from h but is not relevant for the lms app. The second commit adds a decorator to static asset requests that checks the cache-buster.

There are various ways that we can reduce the likelihood of an HTML / asset mismatch when deploying the app, but this PR at least makes it so that a request with the wrong asset version is less likely to be cached downstream. For example Cloudflare uses a [3 minute](https://support.cloudflare.com/hc/en-us/articles/200172516-Understanding-Cloudflare-s-CDN#h_51422705-42d0-450d-8eb1-5321dcadb5bc) rather than 120 minute cache expiry for 404 responses. This change will also surface any scenarios where an HTML page is served with outdated asset URLs or asset URLs that do not include the cache-buster for any reason.

Part of https://github.com/hypothesis/lms/issues/2799.

----

TODO: Add/update tests

### DIFF
--- a/lms/assets.py
+++ b/lms/assets.py
@@ -3,6 +3,7 @@ import json
 import os
 
 import importlib_resources
+from pyramid.httpexceptions import HTTPNotFound
 from pyramid.settings import aslist
 from pyramid.static import static_view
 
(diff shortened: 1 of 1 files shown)


robertknight:
The asset serving code is duplicated in h, so this might be a good time to look at extracting it to a shared package. Discussion at https://hypothes-is.slack.com/archives/C1MA4E9B9/p1625141336177500.

robertknight:
This PR has been superseded by https://github.com/hypothesis/lms/pull/2919.

Issue #2764, opened 2021-06-01
Automatically fix course-copied Canvas Files assignments whenever possible


seanh:
## Problem

Go to a Canvas course that contains Canvas Files assignments (e.g. [Developer Test Course with Sections Enabled](https://hypothesis.instructure.com/courses/125/)) and copy the course: go to <kbd>Settings</kbd> &rarr; <kbd>Copy this course</kbd>. Now go to the new course and launch one of the Canvas Files assignments: you should see an error message.

Instead of copying a course you can also just use the pre-created copied course. You first have to log in using either the `eng+canvasteacher@hypothes.is` account (teacher) or the `eng+coursecopy` account (student) (both are in 1Password) then go to [Copy of Developer Test Course with Sections Enabled](https://hypothesis.instructure.com/courses/252/assignments/1334). 

**Note:** If you log in as a student who is also a member of the original course that was copied from you won't run into the error message: the copied assignments actually work fine for students who're members of the original course. You need to log in as a teacher (member of the original course or not) or as a student who is not a member of the original course.

When a new Canvas course is created by copying an existing course any Canvas Files assignments in the new course are broken: students and teachers run into error messages when trying to launch the Canvas Files assignments. **It would be much better if the copied Canvas Files assignments could just work instead of showing an error message.**

Students see this error message:

![Screenshot from 2020-11-23 13-41-55](https://user-images.githubusercontent.com/22498/99968923-b24f9580-2d91-11eb-901d-2fd124b37a5f.png)

Teachers see this error message:

![Screenshot from 2020-11-23 13-44-44](https://user-images.githubusercontent.com/22498/99969208-0fe3e200-2d92-11eb-8258-cfe77286d8f0.png)

### What's actually going wrong

Canvas file IDs are global not per-course. The ["public inline preview url" API](https://canvas.instructure.com/doc/api/files.html#method.files.public_url) that we use to get a Via-compatible URL for a Canvas file takes no course ID only a global file ID: `GET /api/v1/files/:id/public_url`

When you copy a Canvas course it creates new copies of all the original course's files in the new course, with new file IDs.

Canvas also copies the assignments from the original course into the new course but it does not update the file IDs in the assignments. The "external tool URL" that's stored in Canvas for a Canvas Files assignment looks like this: `https://lms.hypothes.is/lti_launches?canvas_file=true&file_id=<ID>`. Where `<ID>` is the ID of the file in the original course. These launch URLs get copied unchanged to the new assignments in the new course.

[...]

PR #2899, opened 2021-06-29
Fix Canvas course copy

seanh:
Fixes https://github.com/hypothesis/lms/issues/2764.

Depends on https://github.com/hypothesis/lms/pull/2907

Testing
=====

All the test cases should work with our [Canvas test site](https://hypothesis.instructure.com/).

You need to run `make devdata` in the LMS project to make sure you have the latest development data.

To get to an SQL shell (to run the various example SQL commands in the test cases below) run `make sql`.

Test data:

* [Developer Test Course with Sections Enabled][] is the original (non-copied) test course
  * [localhost (make devdata) Canvas Files Assignment] is a Canvas Files assignment in this course
  * File **798** (`Françoise_Hardy.pdf`) is a PDF file in this course that the assignment uses
  * `eng+canvasteacher@hypothes.is` is a teacher in this course (all user passwords are in 1Password)
  * `eng+canvasstudent@hypothes.is` is a student in this course

* [Copy of Developer Test Course with Sections Enabled][] is a copied course
  * [localhost (make devdata) Canvas Files Assignment (copy)][] is a copied Canvas Files assignment in this course
  * `eng+canvasteacher@hypothes.is` is a teacher who is in both the copied course and the original course
  * `eng+canvasstudent@hypothes.is` is a student who is in both the copied course and the original course
  * `eng+coursecopyteacher` is a teacher who is in the copied course but *not* the original course
  * `eng+coursecopystudent` is a student who is in the copied course but *not* the original course

Launching a non-course-copied assignment
----------------------------------------

Launch [localhost (make devdata) Canvas Files Assignment][] from [Developer Test Course with Sections Enabled][] as both a teacher and a student.

The assignment should launch successfully.

You should not have any mappings in your DB:

```sql
postgres=# select * from module_item_configurations where extra != '{}';
 id | resource_link_id | tool_consumer_instance_guid | document_url | extra
----+------------------+-----------------------------+--------------+-------
(0 rows)
```

Launching a course-copied assignment: teacher also belongs to original course
-----------------------------------------------------------------------------

Launch [localhost (make devdata) Canvas Files Assignment (copy)][] from [Copy of Developer Test Course with Sections Enabled][] as `eng+canvasteacher@hypothes.is`.

The assignment should launch successfully.

Even though the teacher could access the original file (since they belong to the original course) we detect that the file isn't in the current course and map it to the matching file in the current course. The mapping means that other users who *aren't* in the original course will be able to launch the assignment. You should find that a mapping has been created in the DB:

```sql
postgres=# select extra from module_item_configurations where extra != '{}';
                  extra
------------------------------------------
 {"canvas_file_mappings": {"798": "799"}}
(1 row)
```

Try launching the assignment again and you should find that it just launches using the existing mapping, without creating any new mappings or making extra API requests to find matching files.

Launching a course-copied assignment: student also belongs to original course
-----------------------------------------------------------------------------

First delete any existing file mappings from your DB:

```sql
postgres=# update module_item_configurations set extra = '{}';
UPDATE 18
```

Now launch [localhost (make devdata) Canvas Files Assignment (copy)][] from [Copy of Developer Test Course with Sections Enabled][] as `eng+canvasstudent@hypothes.is`.

The assignment should launch successfully.

You should not have any mappings in your DB:

```sql
postgres=# select * from module_item_configurations where extra != '{}';
 id | resource_link_id | tool_consumer_instance_guid | document_url | extra
----+------------------+-----------------------------+--------------+-------
(0 rows)
```

Since the student belongs to the original course they're able to access the original file. When students launch assignments we **don't** check to make sure that the file is in the current course because that check requires extra network requests, so no mapping gets created.

If you first launch the assignment as `eng+canvasteacher@hypothes.is` (thus creating a mapping) and _then_ launch the assignment as `eng+canvasstudent@hypothes.is` then it should find and use the existing mapping.

Launching a course-copied assignment: teacher does not belong to original course
--------------------------------------------------------------------------------

First delete any existing file mappings from your DB:

```sql
postgres=# update module_item_configurations set extra = '{}';
UPDATE 18
```

Now launch [localhost (make devdata) Canvas Files Assignment (copy)][] from [Copy of Developer Test Course with Sections Enabled][] as `eng+coursecopyteacher@hypothes.is`.

The assignment should launch successfully.

```sql
postgres=# select extra from module_item_configurations where extra != '{}';
                  extra
------------------------------------------
 {"canvas_file_mappings": {"798": "799"}}
(1 row)
```

Try launching the assignment again and you should find that it just launches using the existing mapping, without creating any new mappings or making extra API requests to find matching files.

Launching a course-copied assignment: student does not belong to original course
--------------------------------------------------------------------------------

First delete any existing file mappings from your DB:

```sql
postgres=# update module_item_configurations set extra = '{}';
UPDATE 18
```

Now launch [localhost (make devdata) Canvas Files Assignment (copy)][] from [Copy of Developer Test Course with Sections Enabled][] as `eng+coursecopystudent@hypothes.is`.

The assignment should launch successfully.

```sql
postgres=# select extra from module_item_configurations where extra != '{}';
                  extra
------------------------------------------
 {"canvas_file_mappings": {"798": "799"}}
(1 row)
```

Try launching the assignment again and you should find that it just launches using the existing mapping, without creating any new mappings or making extra API requests to find matching files.

When we don't have a record of the original course's files
----------------------------------------------------------

The creation of "mappings" in order to fix course copied assignments depends on us having information about the assignment's original file in our DB (the file used by the original assignment in the original course, this is the file whose ID comes to us in the `file_id` parameter whenever the original or a copied assignment is launched.

If you have created or launched any assignment in the original course as a teacher then we will have downloaded the info about all of the original course's files (as they were at the time of your launch).

Launching assignments as a student does not download info about the course's files.

It is possible that a user could be launching a course-copied assignment and we don't have info about the assignment's original file in our DB. For example this could happen if the course was copied before we had deployed the file info tracking feature.

First, delete all the file info from your DB:

```sql
postgres=# delete from file;
DELETE 12
```

Also delete any previously created mappings from your DB:

```sql
postgres=# update module_item_configurations set extra = '{}';
UPDATE 18
```

Now launch [localhost (make devdata) Canvas Files Assignment (copy)][] from [Copy of Developer Test Course with Sections Enabled][].

If you launch the assignment as a teacher (member of the original course or not) you'll run into a **Hypothesis couldn't find the file in the course** error message.

If you launch the assignment as a student who isn't a member of the original course you'll get a **Couldn't get the file from Canvas** error.

If you launch the assignment as a student who is also a member of the original course it will launch successfully.

Deleted files
-------------

If you launch [localhost (make devdata) Canvas Files Assignment Whose File Has Been Deleted](https://hypothesis.instructure.com/courses/125/assignments/1370) in [Developer Test Course with Sections Enabled][] as a teacher you'll get the **Hypothesis couldn't find the file in the course** error.

We notice that the file isn't in the course which triggers us to think that the assignment might have been course copied and to look for a matching file that _is_ in the course. We won't find a matching file in the course (since Canvas omits deleted files from its list-files API responses) to the look up fails and we just fall back to the error dialog.

If you launch this assignment as a student it will launch successfully. (This is because Canvas doesn't seem to actually delete files when you delete them. Ever. Students can still download them.)

Deleting and re-uploading a file
--------------------------------

What if you delete an assignment's file from Canvas and then re-upload the same file again? This will create a new file in Canvas with a new file ID. But the new file will have the same filenamena and size as the deleted one.

If you now launch the assignment whose file was deleted as a student it will just launch successfully downloading the original (deleted) file since students can still download deleted files from Canvas.

But if you launch the assignment as a teacher we'll notice that the assignment's file ID isn't in the course and will find the new file in the course with the same name and size and will create a mapping and "fix" the assignment :)

Unpublished files
-----------------

Files can be marked as "unpublished" in Canvas which means that teachers can see them but students can't.

[localhost (make devdata) Canvas Files Assignment Whose File Is Unpublished](https://hypothesis.instructure.com/courses/125/assignments/1371) is an assignment for an unpublished file.

The assignment will launch successfully for teachers but students get a **Couldn't get the file from Canvas** error.

**But** if the course has another file with the same name and size then when a student launched the assignment we would create a mapping to that file and the launch would succeed :)

Editing assignments
-------------------

What if we store a mapping in our DB and then a user edits the assignment and changes its file?

Since the mappings map one file ID to another if the user changes the `file_id` in Canvas the previous mapping will no longer apply. (So editing works as expected.)

Renaming files
----------------------

TODO

[localhost (make devdata) Canvas Files Assignment]: https://hypothesis.instructure.com/courses/125/assignments/875/
[Developer Test Course with Sections Enabled]: https://hypothesis.instructure.com/courses/125
[Copy of Developer Test Course with Sections Enabled]: https://hypothesis.instructure.com/courses/252
[localhost (make devdata) Canvas Files Assignment (copy)]: https://hypothesis.instructure.com/courses/252/assignments/1321

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -55,11 +55,13 @@ class CanvasService:
             if not found_file_id:
                 raise
 
+            # Try again to return a public URL, this time using found_file_id.
+            url = self.api.public_url(found_file_id)
+
             # Store a mapping so we don't have to re-search next time.
             module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
 
-            # Try again to return a public URL, this time using found_file_id.
-            return self.api.public_url(found_file_id)
+            return url
 
 
 class CanvasFileFinder:
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -45,6 +45,10 @@ class CanvasService:
             file_ = self._file_service.get(effective_file_id, type_="canvas_file")
 
             if not file_:
+                # We don't have a record of the assignment's file in our DB.
+                # This can happen, for example, if Hypothesis has not been
+                # launched in the original assignment's course since we
+                # deployed the code that started recording files in the DB.
                 raise
 
             # Look for a copy of the assignment's file that the current user
@@ -52,6 +56,10 @@ class CanvasService:
             found_file_id = self.find_matching_file_in_course(course_id, file_)
 
             if not found_file_id:
+                # We didn't find a matching file in the current course.
+                # This could mean that the file has been deleted, has been
+                # renamed, or the current user doesn't have permission to see
+                # the file.
                 raise
 
             # We found a matching copy of the assignment's file that the
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -42,62 +42,38 @@ class CanvasService:
                 self.assert_file_in_course(effective_file_id, course_id)
             return self.api.public_url(effective_file_id)
         except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
-            # Either the user can't see the file in the current course's list
-            # of files or the user got a permissions error from the Canvas API
-            # when trying to get a public URL for the file.
+            # The user can't see the file in the course. This could be because:
             #
-            # This can happen because the course has been copied and the
-            # assignment's file_id is from the original course. Or it can
-            # happen because the assignment's file has been deleted from
-            # Canvas. Or it can happen because the user doesn't have permission
-            # to see the file ("unpublished" files in Canvas are visible to
-            # instructors but not to students).
+            # * The course has been copied so the assignment's file_id is from
+            #   the original course (and the user isn't in the original course)
+            # * The assignment's file has been deleted from Canvas
+            # * The file *is* in the course but the user can't see it
+            #   (only instructors can see "unpublished" files in Canvas)
             #
             # We'll try to find another copy of the same file that the current
             # user *can* see in the current course and use that instead.
 
-            # Look for a previously saved record of the assignment's file in our DB.
             file = self._file_service.get(effective_file_id, type_="canvas_file")
 
             if not file:
-                # We don't have a record of the assignment's file in our DB.
-                # This can happen, for example, if Hypothesis has not been
-                # launched in the original assignment's course since we
-                # deployed the code that started recording files in the DB.
                 raise
 
-            # Look for a copy of the assignment's file that the current user
-            # *can* see in the current course.
             found_file_id = self.find_matching_file_in_course(course_id, file)
 
             if not found_file_id:
-                # We didn't find a matching file in the current course.
-                # This could mean that the file has been deleted, has been
-                # renamed, or the current user doesn't have permission to see
-                # the file.
                 raise
 
-            # We found a matching copy of the assignment's file that the
-            # current user *can* see in the current course. Store a mapping so
-            # that we don't have to re-do the search the next time the
-            # assignment is launched.
+            # found_file_id is a copy of file_id that the current user *can*
+            # see in the current course.
+
+            # Store a mapping so we don't have to re-search next time.
             module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
 
-            # Try again using the found matching file.
+            # Try again to return a public URL, this time using found_file_id.
             return self.api.public_url(found_file_id)
 
     def assert_file_in_course(self, file_id: str, course_id: str) -> bool:
-        """
-        Raise if the current user can't see file_id in course_id.
-
-        Raise CanvasFileNotFoundInCourse if the current user can't see a file
-        with ID file_id in the course with ID course_id. This could be because
-        the file is in another course, because the file was in course_id but
-        has been deleted, or because the file is in course_id but the current
-        user doesn't have permission to see it (for example files that are
-        marked as "unpublished" in Canvas can only be seen by teachers, not
-        students).
-        """
+        """Raise if the current user can't see file_id in course_id."""
         for file in self.api.list_files(course_id):
             # The Canvas API returns file IDs as ints but the file_id param
             # that this method receives (from our proxy API) is a string.
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,7 +1,7 @@
 from typing import Optional
 
 from lms import models
-from lms.services.exceptions import CanvasFileNotFoundInCourse
+from lms.services.exceptions import CanvasAPIPermissionError, CanvasFileNotFoundInCourse
 
 
 class CanvasService:
@@ -9,28 +9,63 @@ class CanvasService:
 
     api = None
 
-    def __init__(self, canvas_api):
+    def __init__(self, canvas_api, file_service):
         self.api = canvas_api
+        self._file_service = file_service
 
-    def public_url_for_file(self, file_id, course_id, check_in_course=False):
-        """
-        Get a public URL for a Canvas file.
+    def public_url_for_file(
+        self, module_item_configuration, file_id, course_id, check_in_course=False
+    ):
+        mapped_file_id = module_item_configuration.get_canvas_mapped_file_id(file_id)
 
-        This will also attempt to check if the file is in the course to detect
-        course copy situations and fix it if we can.
+        # If there's a previously stored mapping for file_id use that instead.
+        effective_file_id = mapped_file_id or file_id
 
-        :param file_id: The file to look up
-        :param course_id: The course the file should be in
-        :param check_in_course: Raise CanvasFileNotFoundInCourse if file_id isn't in
-            course_id
-        :return: A URL suitable for public presentation of the file
-        :raise  CanvasFileNotFoundInCourse: if check_in_course=True and file_id isn't in course_id
-            be in the course but isn't.
-        """
+        try:
+            return self._public_url(
+                effective_file_id,
+                course_id=course_id if check_in_course else None,
+            )
+        except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
+            # Either the user can't see the file in the current course's list
+            # of files or the user got a permissions error from the Canvas API
+            # when trying to get a public URL for the file.
+            #
+            # This can happen because the course has been copied and the
+            # assignment's file_id is from the original course. Or it can
+            # happen because the assignment's file has been deleted from
+            # Canvas. Or it can happen because the user doesn't have permission
+            # to see the file ("unpublished" files in Canvas are visible to
+            # instructors but not to students).
+            #
+            # We'll try to find another copy of the same file that the current
+            # user *can* see in the current course and use that instead.
 
-        if check_in_course:
-            if not self.can_see_file_in_course(file_id, course_id):
-                raise CanvasFileNotFoundInCourse(file_id)
+            # Look for a previously saved record of the assignment's file in our DB.
+            file_ = self._file_service.get(effective_file_id, type_="canvas_file")
+
+            if not file_:
+                raise
+
+            # Look for a copy of the assignment's file that the current user
+            # *can* see in the current course.
+            found_file_id = self.find_matching_file_in_course(course_id, file_)
+
+            if not found_file_id:
+                raise
+
+            # We found a matching copy of the assignment's file that the
+            # current user *can* see in the current course. Store a mapping so
+            # that we don't have to re-do the search the next time the
+            # assignment is launched.
+            module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
+
+            # Try again using the found matching file.
+            return self._public_url(found_file_id)
+
+    def _public_url(self, file_id, course_id=None):
+        if course_id and not self.can_see_file_in_course(file_id, course_id):
+            raise CanvasFileNotFoundInCourse(file_id)
 
         return self.api.public_url(file_id)
 
@@ -82,4 +117,7 @@ class CanvasService:
 
 
 def factory(_context, request):
-    return CanvasService(canvas_api=request.find_service(name="canvas_api_client"))
+    return CanvasService(
+        canvas_api=request.find_service(name="canvas_api_client"),
+        file_service=request.find_service(name="file"),
+    )
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -38,10 +38,12 @@ class CanvasService:
         effective_file_id = mapped_file_id or file_id
 
         try:
-            return self._public_url(
-                effective_file_id,
-                course_id=course_id if check_in_course else None,
-            )
+            if check_in_course and not self.can_see_file_in_course(
+                effective_file_id, course_id
+            ):
+                raise CanvasFileNotFoundInCourse(effective_file_id)
+
+            return self.api.public_url(effective_file_id)
         except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
             # Either the user can't see the file in the current course's list
             # of files or the user got a permissions error from the Canvas API
@@ -85,13 +87,7 @@ class CanvasService:
             module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
 
             # Try again using the found matching file.
-            return self._public_url(found_file_id)
-
-    def _public_url(self, file_id, course_id=None):
-        if course_id and not self.can_see_file_in_course(file_id, course_id):
-            raise CanvasFileNotFoundInCourse(file_id)
-
-        return self.api.public_url(file_id)
+            return self.api.public_url(found_file_id)
 
     def can_see_file_in_course(self, file_id: str, course_id: str) -> bool:
         """
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -86,7 +86,7 @@ class CanvasFileFinder:
         self, course_id: str, file_id: str
     ) -> Optional[str]:
         """
-        Return the ID of a file in course_id that matches `file`.
+        Return the ID of a file in course_id that matches the file with ID file_id.
 
         Search for a file that the current Canvas user can see in course_id and
         that matches the given `file` (same filename and size) and return the
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -57,9 +57,9 @@ class CanvasService:
             # user *can* see in the current course and use that instead.
 
             # Look for a previously saved record of the assignment's file in our DB.
-            file_ = self._file_service.get(effective_file_id, type_="canvas_file")
+            file = self._file_service.get(effective_file_id, type_="canvas_file")
 
-            if not file_:
+            if not file:
                 # We don't have a record of the assignment's file in our DB.
                 # This can happen, for example, if Hypothesis has not been
                 # launched in the original assignment's course since we
@@ -68,7 +68,7 @@ class CanvasService:
 
             # Look for a copy of the assignment's file that the current user
             # *can* see in the current course.
-            found_file_id = self.find_matching_file_in_course(course_id, file_)
+            found_file_id = self.find_matching_file_in_course(course_id, file)
 
             if not found_file_id:
                 # We didn't find a matching file in the current course.
@@ -98,23 +98,23 @@ class CanvasService:
         marked as "unpublished" in Canvas can only be seen by teachers, not
         students).
         """
-        for file_ in self.api.list_files(course_id):
+        for file in self.api.list_files(course_id):
             # The Canvas API returns file IDs as ints but the file_id param
             # that this method receives (from our proxy API) is a string.
             # Convert ints to strings so that we can compare them.
-            if str(file_["id"]) == file_id:
+            if str(file["id"]) == file_id:
                 return
 
         raise CanvasFileNotFoundInCourse(file_id)
 
     def find_matching_file_in_course(
-        self, course_id: str, file_: models.File
+        self, course_id: str, file: models.File
     ) -> Optional[str]:
         """
-        Return the ID of a file in course_id that matches file_.
+        Return the ID of a file in course_id that matches `file`.
 
         Search for a file that the current Canvas user can see in course_id and
-        that matches the given file_ (same filename and size) and return the
+        that matches the given `file` (same filename and size) and return the
         matching file's ID.
 
         Return None if no matching file is found.
@@ -125,7 +125,7 @@ class CanvasService:
             display_name = file_dict["display_name"]
             size = file_dict["size"]
 
-            if display_name == file_.name and size == file_.size:
+            if display_name == file.name and size == file.size:
                 return str(file_dict["id"])
 
         return None
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -16,6 +16,22 @@ class CanvasService:
     def public_url_for_file(
         self, module_item_configuration, file_id, course_id, check_in_course=False
     ):
+        """
+        Return a public URL for file_id.
+
+        :param module_item_configuration: the ModuleItemConfiguration for the
+            current assignment
+        :param file_id: the Canvas API ID of the file
+        :param course_id: the Canvas API ID of the course that the file is in
+        :param check_in_course: whether to check that file_id is in course_id
+
+        :raise CanvasFileNotFoundInCourse: if check_in_course was True and the
+            current user can't see file_id in course_id's list of files
+
+        :raise CanvasAPIPermissionError: if the user gets a permissions error
+            from the Canvas API when trying to get a public URL for file_id
+        """
+
         mapped_file_id = module_item_configuration.get_canvas_mapped_file_id(file_id)
 
         # If there's a previously stored mapping for file_id use that instead.
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -35,7 +35,7 @@ class CanvasService:
 
         try:
             if check_in_course:
-                self._finder.assert_file_in_course(effective_file_id, course_id)
+                self._finder.assert_file_in_course(course_id, effective_file_id)
             return self.api.public_url(effective_file_id)
         except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
             # The user can't see the file in the course. This could be because:
@@ -71,7 +71,7 @@ class CanvasFileFinder:
         self._api = canvas_api
         self._file_service = file_service
 
-    def assert_file_in_course(self, file_id: str, course_id: str) -> bool:
+    def assert_file_in_course(self, course_id: str, file_id: str) -> bool:
         """Raise if the current user can't see file_id in course_id."""
         for file in self._api.list_files(course_id):
             # The Canvas API returns file IDs as ints but the file_id param
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -38,11 +38,8 @@ class CanvasService:
         effective_file_id = mapped_file_id or file_id
 
         try:
-            if check_in_course and not self.can_see_file_in_course(
-                effective_file_id, course_id
-            ):
-                raise CanvasFileNotFoundInCourse(effective_file_id)
-
+            if check_in_course:
+                self.assert_file_in_course(effective_file_id, course_id)
             return self.api.public_url(effective_file_id)
         except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
             # Either the user can't see the file in the current course's list
@@ -89,28 +86,26 @@ class CanvasService:
             # Try again using the found matching file.
             return self.api.public_url(found_file_id)
 
-    def can_see_file_in_course(self, file_id: str, course_id: str) -> bool:
+    def assert_file_in_course(self, file_id: str, course_id: str) -> bool:
         """
-        Return True if the current user can see file_id in course_id.
+        Raise if the current user can't see file_id in course_id.
 
-        Return False if the current user cannot currently see a file with ID
-        file_id in the course with ID course_id. This could be because the file
-        is in another course, because the file was in course_id but has been
-        deleted, or because the file is in course_id but the current user
-        doesn't have permission to see it (for example files that are marked as
-        "unpublished" in Canvas can only be seen by teachers, not students).
+        Raise CanvasFileNotFoundInCourse if the current user can't see a file
+        with ID file_id in the course with ID course_id. This could be because
+        the file is in another course, because the file was in course_id but
+        has been deleted, or because the file is in course_id but the current
+        user doesn't have permission to see it (for example files that are
+        marked as "unpublished" in Canvas can only be seen by teachers, not
+        students).
         """
-        files_in_course = self.api.list_files(course_id)
-
-        for file_ in files_in_course:
-
+        for file_ in self.api.list_files(course_id):
             # The Canvas API returns file IDs as ints but the file_id param
             # that this method receives (from our proxy API) is a string.
             # Convert ints to strings so that we can compare them.
             if str(file_["id"]) == file_id:
-                return True
+                return
 
-        return False
+        raise CanvasFileNotFoundInCourse(file_id)
 
     def find_matching_file_in_course(
         self, course_id: str, file_: models.File
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -102,8 +102,9 @@ class CanvasFileFinder:
         for file_dict in file_dicts:
             display_name = file_dict["display_name"]
             size = file_dict["size"]
+            id_ = str(file_dict["id"])
 
-            if display_name == file.name and size == file.size:
+            if display_name == file.name and size == file.size and id_ != file.lms_id:
                 return str(file_dict["id"])
 
         return None
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -1,6 +1,5 @@
 from typing import Optional
 
-from lms import models
 from lms.services.exceptions import CanvasAPIPermissionError, CanvasFileNotFoundInCourse
 
 
@@ -11,7 +10,7 @@ class CanvasService:
 
     def __init__(self, canvas_api, file_service):
         self.api = canvas_api
-        self._file_service = file_service
+        self._finder = CanvasFileFinder(canvas_api, file_service)
 
     def public_url_for_file(
         self, module_item_configuration, file_id, course_id, check_in_course=False
@@ -31,15 +30,12 @@ class CanvasService:
         :raise CanvasAPIPermissionError: if the user gets a permissions error
             from the Canvas API when trying to get a public URL for file_id
         """
-
-        mapped_file_id = module_item_configuration.get_canvas_mapped_file_id(file_id)
-
         # If there's a previously stored mapping for file_id use that instead.
-        effective_file_id = mapped_file_id or file_id
+        effective_file_id = module_item_configuration.get_canvas_mapped_file_id(file_id)
 
         try:
             if check_in_course:
-                self.assert_file_in_course(effective_file_id, course_id)
+                self._finder.assert_file_in_course(effective_file_id, course_id)
             return self.api.public_url(effective_file_id)
         except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
             # The user can't see the file in the course. This could be because:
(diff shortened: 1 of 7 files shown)


lms/services/canvas.py
```diff
@@ -1,36 +1,64 @@
...
+                resource_link_id,
+                file_id,
+                found_file_id,
+            )
+
+            return self._public_url(found_file_id)
+
+    def _public_url(self, file_id, course_id=None):
+        if course_id and not self.can_see_file_in_course(file_id, course_id):
+            raise CanvasFileNotFoundInCourse(file_id)
```
seanh, line 61: I think this code could be a bit simpler (and avoid the `# pylint:disable=too-many-arguments`) if:

1. `FileService` didn't require the `application_instance` argument, instead getting the current request's `application_instance` injected by its factory
2. `AssignmentService.{get,set}_canvas_mapped_file_id()` took a single `module_item_configuration` argument and used `module_item_configuration.tool_consumer_instance_guid` and `module_item_configuration.resource_link_id`, instead of taking separate `tool_consumer_instance_guid` and `resource_link_id` arguments. Or alternatively if `{get,set}_canvas_mapped_file_id()` were `ModuleItemConfiguration` methods instead of `AssignmentService` methods

Then `public_url_for_file()` could look like this:

```python
    def public_url_for_file(
        self, module_item_configuration, file_id, course_id, check_in_course=False
    ):
        mapped_file_id = module_item_configuration.get_canvas_mapped_file_id(file_id)

        # If there's a previously stored mapping for file_id use that instead.
        effective_file_id = mapped_file_id or file_id

        try:
            return self._public_url(
                effective_file_id,
                course_id=course_id if check_in_course else None,
            )
        except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
            file_ = self._file_service.get(effective_file_id, type_="canvas_file")

            if not file_:
                raise

            found_file_id = self.find_matching_file_in_course(course_id, file_)

            if not found_file_id:
                raise

            module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)

            return self._public_url(found_file_id)
```

lms/services/canvas.py
```diff
@@ -1,36 +1,71 @@
...
+            # instructors but not to students).
+            #
+            # We'll try to find another copy of the same file that the current
+            # user *can* see in the current course and use that instead.
+
+            # Look for a previously saved record of the assignment's file in our DB.
+            file_ = self._file_service.get(effective_file_id, type_="canvas_file")
+
+            if not file_:
+                raise
```
marcospri, line 48: Maybe the raises here could have a comment too. What's the case in this one? We might not seen this file in the new course?

lms/services/canvas.py
```diff
@@ -1,36 +1,71 @@
...
+
+            # Look for a copy of the assignment's file that the current user
+            # *can* see in the current course.
+            found_file_id = self.find_matching_file_in_course(course_id, file_)
+
+            if not found_file_id:
+                raise
```
marcospri, line 55: And here? We do in fact don't have access to the file or it has been remove?

lms/services/canvas.py
seanh: Added a comment

lms/services/canvas.py
seanh: Done

tests/unit/lms/services/canvas_test.py
```diff
@@ -1,30 +1,370 @@
...
 import pytest
 
-from lms.services import CanvasFileNotFoundInCourse, CanvasService
+from lms.services import (
+    CanvasAPIPermissionError,
+    CanvasFileNotFoundInCourse,
+    CanvasService,
+)
 from lms.services.canvas import factory
 from tests import factories
```
jon-betts, line 11: I think these tests really need some thought into how they can be more concise and easier to read. Many of the tests are longer than my monitor vertically. There's a lot of duplication here and it's not clear to me that we're testing different scenarios, or what is different between scenarios.

I think we could do with unit tests which have a narrower non-overlapping focus and attempt to assert one or two things about the code. I feel like we've ended up with a Cartesian product of all possible combinations of events as scenarios rather than unit tests isolating those different events.

I don't think we need to repeat and assert every call the code makes in each scenario. We should focus in on what's different about this one. For example, with the `check_in_course` tests we should prove we call those functions relating to `check_in_course`. We don't need to re-assert all of the regular calls and return values we've covered in other tests.

In general I would also favour anything that would make this shorter. Some of the fixtures like `file_id_from_current_course` would probably be better being replaced with `file_from_current_course["id"]`. `canvas_api_client` could be replaced with `canvas_service.api` which mirrors the usage in the code. A `functools.partial` around `public_url_for_file` would cut down on the number of fixtures in every test, and allow you to indicate what's important about this scenario by only specifying the important parts of the call that are relevant in this scenario. 

`module_item_configuration` is probably better referred to as `assignment` throughout. It is the return value of the `assignment` service and is more concise and easier to read.

I suspect the level of comment here is an attempt to make this easier to understand, but I'd suggest it's a sign that the tests are too confusing in the first place. Either this level of comment isn't really necessary and should be reduced for readability, or the tests are probably too complicated, and should be simplified so the comments aren't required.

lms/services/canvas.py
```diff
@@ -1,36 +1,79 @@
...
-            be in the course but isn't.
-        """
-
-        if check_in_course:
-            if not self.can_see_file_in_course(file_id, course_id):
-                raise CanvasFileNotFoundInCourse(file_id)
+        self._file_service = file_service
+
+    def public_url_for_file(
+        self, module_item_configuration, file_id, course_id, check_in_course=False
```
jon-betts, line 17: `module_item_configuration` is probably better referred to as `assigment` throughout. It is the return value of the `assignment` service and is more concise and easier to read.

tests/unit/lms/services/canvas_test.py
jon-betts: I think it might be worth going through the exercise of mapping out the planned tests as a skeleton test file (without any content in the test) just to get the names and design down. Something like...

```python
class TestPublicURLForFile:
    def test_it(self):
        # Assert call the Canvas API with the id
        # Assert we return the value
        ...

    def test_it_uses_a_file_mapping_if_present(self):
        # Assert we call the assignment to get a mapping
        # Assert call the Canvas API with the mapped id
        ...

    def test_check_in_course_looks_up_the_file_in_the_course(self):
        # Test we list the files of the correct course
        ...

    def test_check_in_course_raises_if_the_file_is_not_found(self):
        # Assert we raise
        ...

    @pytest.mark.usefixtures("with_failing_lookup")
    def test_it_can_remap_a_file(self):
        # Setup a file record
        # Setup a mapping with a matching file

        # Assert we store the new mapping
        ...

    @pytest.mark.usefixtures("with_failing_lookup")
    def test_it_cannot_map_a_file_without_having_seen_it_before(self):
        # Assert we blow up
        ...

    @pytest.mark.usefixtures("with_failing_lookup")
    @pytest.mark.parametrize("field,value", (("name", "different"), ("size", "99999")))
    def test_it_cannot_map_a_file_if_there_is_no_matching_file(self, field, value):
        # Setup a mapping with a matching file

        # Assert we blow up
        ...

    @pytest.fixture(params=(CanvasFileNotFoundInCourse, CanvasAPIPermissionError))
    def with_failing_lookup(self, param):
        # Configure the situation so we raise the correct error
        ...
```

lms/services/canvas.py
```diff
@@ -1,36 +1,79 @@
...
+                # This could mean that the file has been deleted, has been
+                # renamed, or the current user doesn't have permission to see
+                # the file.
+                raise
+
+            # We found a matching copy of the assignment's file that the
+            # current user *can* see in the current course. Store a mapping so
+            # that we don't have to re-do the search the next time the
+            # assignment is launched.
+            module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
```
jon-betts, line 69:
```suggestion
            module_item_configuration.set_mapped_file_id(file_id, found_file_id, type_="canvas_file")
```

lms/views/api/canvas/files.py
```diff
@@ -30,12 +30,19 @@ def via_url(self):
...
+
         public_url = self.canvas.public_url_for_file(
-            file_id=self.request.matchdict["file_id"],
-            course_id=self.request.matchdict["course_id"],
-            # Teachers can have broad permissions and see files that aren't in
-            # the course. So do this slower check (extra API call) to warn the
-            # teacher that their students might not be able to see the file.
+            module_item_configuration,
+            self.request.matchdict["file_id"],
+            self.request.matchdict["course_id"],
```
jon-betts, line 45: I think kwargs are easier to keep straight once we start to get lots of arguments (or even not) and form some nice natural commentary / documentation. They also facilitate refactoring.

jon-betts (commented):
Some general feedback:

 * I like the solution to raising which skips the ugly double check I had in my code
 * We should use doc-strings
 * Long comments often harm readability and make the code more confusing. Try and find the shortest way to convey the important information
 * I cannot tell from looking at the tests what the individual focus of each of the tests is, or what they are individually responsible for testing. I think this would benefit from a clearer testing plan with non-overlapping tests with a clearer focus
 * The tests are too long in general to be digestible

There are some things which I noticed outside of this code:

 * The `FileService` is probably to broadly named and could really do with being documented. I think this is something we should put back into the linting requirements as it's causing us not to meet our standards
 * The assigment file mapping functions should not be canvas specific. This is pretty easy to achieve in the same way we do in the `FileService` by accepting a file type

lms/services/canvas.py
```diff
@@ -1,36 +1,79 @@
...
-        """
-
-        if check_in_course:
-            if not self.can_see_file_in_course(file_id, course_id):
-                raise CanvasFileNotFoundInCourse(file_id)
+        self._file_service = file_service
+
+    def public_url_for_file(
+        self, module_item_configuration, file_id, course_id, check_in_course=False
+    ):
```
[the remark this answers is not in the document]
seanh: Oh, that's weird. I didn't mean to delete the docstring. I'll put it back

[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/2899/commits/b88be4a4aa8922a638695dcc56ebf6eb0d2cc980

I put the original docstring back and updated it (adding the `module_item_configuration` param) and also fixed it up a bit:

* Added that `CanvasAPIPermissionError` could be raised

* Removed `This will also attempt to check if the file is in the course to detect course copy situations and fix it if we can` from the docstring body because it's not true. On master there is no support for fixing course copy situations yet, so this docstring has been a lie on master since it was added. But also the `check_in_course` doesn't really detect course copy situations, it detects that the current user can't see the file in the course which could be for a number of reasons. Also it's not really true to say "This will also attempt to check if the file is in the course...": by default it won't do that, only if `check_in_course=True` is passed. I think it's better to leave the `:param check_in_course:` and `:raise CanvasFileNotFoundInCourse:` parts of the docstring to explain this, they cover it.

* Removed the `:return:`. If the first line of the docstring is `Return a public URL for file_id` then it doesn't add anything to also have `:return: A URL suitable for public presentation of the file` ("URL suitable for public presentation" is a great way to say "public URL" too :smile: )

* Fixed up some other gems: `:raise CanvasFileNotFoundInCourse: if check_in_course=True and file_id isn't in course_id be in the course but isn't` :)

* General wording tidy up

lms/services/canvas.py
```diff
@@ -1,36 +1,79 @@
...
+            # that we don't have to re-do the search the next time the
+            # assignment is launched.
+            module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
+
+            # Try again using the found matching file.
+            return self._public_url(found_file_id)
+
+    def _public_url(self, file_id, course_id=None):
+        if course_id and not self.can_see_file_in_course(file_id, course_id):
+            raise CanvasFileNotFoundInCourse(file_id)
```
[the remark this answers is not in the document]
seanh: Yep, this is **much** better: https://github.com/hypothesis/lms/pull/2899/commits/cde37f64a539d76076d4eed0299536016cb9f9e6, https://github.com/hypothesis/lms/pull/2899/commits/e01a9bcaf14ddabc295f4da31767644f55eacbe4.

I don't think I like private helper functions like this `self._public_url()` so I wasn't happy with it. And as it turns out it's completely unnecessary and was just obscuring things.

I've slightly bent the common coding guideline about minimising the number of lines inside a `try` here but not by much: three lines (including an `if`) isn't too bad and overall it makes this code much better.

The `can_see_file_in_course()` &rarr; `assert_file_in_course()` change is also a big improvement.

I _didn't_ inline `assert_file_in_course()` into `public_url_for_file()`. I kept it as a separate method for two reasons: 1. It provides a separate place to put a docstring explaining why a file might not be in a course without polluting the higher-level `public_url_for_file()` method with too many comments; 2. I think the `try` part of `public_url_for_file()` becomes too unwieldy if you try to inline `assert_file_in_course()` into it. It becomes less readable. Too much detail at once. Too many levels of indentation. Ugly code formatting. Too many lines of code within a `try`. It's not _too_ bad, but:

```python
try:
    if check_in_course:
        if not any(
            str(file_["id"]) == effective_file_id
            for file_ in self.api.list_files(course_id)
        ):
            raise CanvasFileNotFoundInCourse(effective_file_id)

    return self.api.public_url(effective_file_id)
except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
    ...
```

And that's without any comment or docstring at all to explain why a file might not be in a course or what that could mean. Compared to with the separate method:

```python
try:
    if check_in_course:
        self.assert_file_in_course(effective_file_id, course_id)
    return self.api.public_url(effective_file_id)
except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
    ...
```

There's a minor naming issue now that the argument is called `check_in_course` but the method that it triggers is called `assert_file_in_course()`. Might want to rename one or the other of them.

lms/services/canvas.py
```diff
@@ -1,36 +1,79 @@
...
+            # happen because the assignment's file has been deleted from
+            # Canvas. Or it can happen because the user doesn't have permission
+            # to see the file ("unpublished" files in Canvas are visible to
+            # instructors but not to students).
+            #
+            # We'll try to find another copy of the same file that the current
+            # user *can* see in the current course and use that instead.
+
+            # Look for a previously saved record of the assignment's file in our DB.
+            file_ = self._file_service.get(effective_file_id, type_="canvas_file")
```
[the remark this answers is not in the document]
seanh: Hmm, vim highlights it for some reason. I think it might have been reserved in Python 2 but not in 3. Fixed

lms/services/canvas.py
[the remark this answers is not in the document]
seanh: I agree about the comments. The code without comments is pretty nice and clear:

```python
effective_file_id = module_item_configuration.get_mapped_file_id(file_id) or file_id

try:
    if check_in_course:
        self.assert_file_in_course(effective_file_id, course_id)
    return self.api.public_url(effective_file_id)
except (CanvasFileNotFoundInCourse, CanvasAPIPermissionError):
    file = self._file_service.get(effective_file_id, type_="canvas_file")

    if not file:
        raise

    found_file_id = self.find_matching_file_in_course(course_id, file)

    if not found_file_id:
        raise

    module_item_configuration.set_mapped_file_id(file_id, found_file_id)

    return self.api.public_url(found_file_id)
```

I'll try cutting them down...

lms/views/api/canvas/files.py
seanh: In this case the variable name (`module_item_configuration`) and dict keys (`"file_id"` and `"course_id"`) already make it obvious what each argument is. If kwargs were used they'd just all be repeated:

```python
public_url = self.canvas.public_url_for_file(
    module_item_configuration=module_item_configuration,
    file_id=self.request.matchdict["file_id"],
    course_id=self.request.matchdict["course_id"],
    ...
```

lms/services/canvas.py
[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/2899/commits/f82e77bbfbe03253a01005b51a27afa4c2d25273

lms/services/canvas.py
seanh: I think we should just bite the bullet and rename `ModuleItemConfiguration` to `Assignment` everywhere. I started a separate PR to do that: https://github.com/hypothesis/lms/pull/2912

lms/services/canvas.py
seanh: I'm gonna do this in a separate PR since `ModuleItemConfiguration.{get,set}_canvas_mapped_file_id()` aren't part of this PR.

How about we just drop the `type_` and just call them "file ID mappings"? A single `ModuleItemConfiguration` can't belong to more than one different LMS so I don't think there's a chance of a conflict:

```python
def get_mapped_file_id(self, file_id):
    return self.extra.get("file_mappings", {}).get(file_id)

def set_mapped_file_id(self, file_id, mapped_file_id):
    self.extra.setdefault("file_mappings", {})[file_id] = mapped_file_id
```

seanh: https://github.com/hypothesis/lms/pull/2913

tests/unit/lms/services/canvas_test.py
seanh: I've had a go at reworking these tests: changing test names, removing comments, getting rid of unnecessary tests.

I'm not sure there's much duplication left in these that's worth removing. There is code that superficially looks like duplication because some of these tests do similar-but-different assertions. I think trying to deduplicate that might make things worse rather than better.

This is what I think needs to be covered:

* `check_in_course` can either be `True` or `False`. There needs to be a happy-path test for each case:
  - [x] `check_in_course=True`: `test_if_check_in_course_is_True_it_checks_that_the_file_is_in_the_course()`
  - [x] `check_in_course=False`: `test_without_check_in_course_and_without_a_mapped_file_id()` 
* There can either be a `mapped_file_id` in the DB or not. There needs to be a happy-path test for each case
  - [x] With a mapping in the DB: `test_if_theres_a_mapped_file_id_it_uses_it()`
  - [x] Without a mapping in the DB: This was already covered by earlier tests like `test_without_check_in_course_and_without_a_mapped_file_id()` 
* I don't think we need to cross-product the `check_in_course` happy paths with the `mapped_file_id` ones but we do need tests for when `check_in_course` is `True` and there's no `mapped_file_id` (it should check for the original `file_id` in the course) and for when `check_in_course` is `True` and there _is_ a `mapped_file_id` (it should check for the `mapped_file_id` in the course, not the original `file_id`)
  - [x] `check_in_course=True` and there's no mapping: covered by the previously mentioned `test_if_check_in_course_is_True_it_checks_that_the_file_is_in_the_course()` 
  - [x] `check_in_course=True` and there _is_ a mapping: covered by the previously mentioned `test_if_theres_a_mapped_file_id_it_uses_it()`
* If `check_in_course` is `True` then `assert_file_in_course()` can raise `CanvasFileNotFoundInCourse`. Also `public_url()` can raise `CanvasAPIPermissionsError`, realistically I think this only happens when `check_in_course` is `False`. We need to test that when _either_ `CanvasFileNotFoundInCourse` or `CanvasAPIPermissionsError` is raised:
  - [x] If there's no `File` object in the DB it raises: `test_file_not_found_in_course_but_no_file_info()` and `test_permissions_error_but_no_file_info()`
  - [x] If there is a file in the DB but there's no matching file in the Canvas course, it raises: `test_file_not_found_in_course_but_no_matching_file()` and `test_permissions_error_but_no_matching_file()`
  - [x] If there's both a file in the DB and a matching file in the Canvas course then it saves a mapping and returns the public URL of the matching file: `test_file_not_found_in_course_and_matching_file_found()` and `test_permissions_error_and_matching_file_found()`

All of the tests that're now remaining in the file are needed to cover those above cases. These cases don't literally cover everything that could possibly happen. But I think they give good enough coverage, and cover the most common cases.

* * *

> Some of the fixtures like `file_id_from_current_course` would probably be better being replaced with `file_from_current_course["id"]`

It's actually `str(file_from_current_course["id"])` and `str(file_from_another_course["id"])` every time, which can be a bit of a mouthful, but I've gone ahead and made this change.

> `canvas_api_client` could be replaced with `canvas_service.api`

The downside of this is that it couples the tests more closely to the `CanvasService` code. If `CanvasService.api` changed lots of tests would break. I've gone ahead and made this change anyway.

> A `functools.partial` around `public_url_for_file` would cut down on the number of fixtures in every test, and allow you to indicate what's important about this scenario by only specifying the important parts of the call

This actually doesn't cut down on fixtures much: in most tests it just replaces the `module_item_configuration` fixture with a `public_url_for_file` fixture. In some tests it actually increases the number of fixtures as `module_item_configuration` is still needed anyway. But it does cut down on the arguments that need to be passed to `public_url_for_file()`, allowing each test to pass only the relevant arguments. I've gone ahead and made this change.

* * *

The longest of these tests is now 33 lines, but it's actually only 19 lines of code (including blank lines) if you don't count comments and the argument list. The tests are all quite simple in form too.

* * *

I've cut down the comments somewhat. But I think the comments here are appropriate. The comments are not explaining how the _tests_ work. They're not there because the tests are confusing. They're explaining the complex external scenarios that are being covered. Example:

```python
    def test_file_not_found_in_course_but_no_matching_file(
        self, file_service, file_from_a_different_course, public_url_for_file
    ):
        # This is what happens when a teacher launches an assignment whose
        # file_id is *not* in the current course and even though we do have a
        # record of this file_id in our DB we don't find a matching file in the
        # current course. Since we can't find a matching file we can't fix the
        # assignment so we raise an error.

        # The file record that we find in our DB. Its name doesn't match any
        # file in the current course.
        file_service.get.return_value = factories.File(name="foo")

        with pytest.raises(CanvasFileNotFoundInCourse):
            public_url_for_file(
                file_id=str(file_from_a_different_course["id"]), check_in_course=True
            )
```

jon-betts:
This PR is getting a bit difficult to comment on inline (might be worth making a duplicate?) but I have some feedback about some of the larger functional things.

### We lookup by `effective_id`, but I think it should be the original id

The original PR looked up both the mapped and the original id, but if you're only going to look up one I'd suggest it be the original for a two reasons:

* The cheeky one is it makes it easier to test as what we look up is not dependent on whether it's mapped or not, so the number of different scenarios to check is less
* I think it's actually the better way to go anyway for the user for a few reasons

If we look up the original id instead then:

* In the case of a copy of a copy, you only need to visit the original assignment once to fix both
* The original is older than the copy, so it's been around longer, and therefore has a higher chance of having been visited (and therefore fixable)

Of course you can call with both and fix it that way.

### The splits we have in the code aren't buying us the usual mocking benefits

Hopefully some of the weird questions I was asking, and my rants about hatless-cats will make a bit more sense, but I think this is partially proving difficult to test because the way we've broken up the problem isn't resulting in the usual benefits we see from mocking. And I think it's helped me work out how we can fix it in a way that works for everyone.

If we consider setting up one of the scenarios which involves a mapping we:

* Prime the file service fixture
* Prime the Canvas API fixture with results which causes `assert_file_in_course` to fail

In the case of the file service, in order to get it to return a file object we use the factory to create one. This is actually identical to what we'd do if we were doing end-to-end testing with the addition of the fact we have to update the fixture `return_value` (but maybe skip a DB flush?)

In the case of the Canvas API, we separated out `assert_file_in_course` to test separately, but we aren't mocking it afterwards, so we actually have to recreate the situation that causes it to raise. We could mock out the method on the object with `object.patch` but it's a bit gross.

In the case of the file mappings, as the methods are on the model object, and we don't mock model objects.

The net effect is we've taken all of these elements and split them out, but our testing is still largely forced to recreate and identical scenario to test it as if they were all still private functions. It's pretty much identical to what'd we'd do if it was an integration test (I suspect the tests would work find if you used the real file service for example).

In the case of the file service and the assignment mapping, I actually think that's fine. They are trivial to mock / use directly respectively. The problem is in the local public methods which aren't mocked.

#### Wouldn't it be lovely if...

In my POC for slimmer tests (https://github.com/hypothesis/lms/pull/2917) what I was trying to do was say:

* Prove mapping works
* Prove that we call mapping in these scenarios (without re-proving the mapping works)

The problem is it's quite hard to setup the scenarios to get what we want, and quite hard to prove that mapping has taken place. Instead we have to have indirectly setup cases to cause things to fail and infer by calls our local public functions make what happened.

It would be much nicer to have something like:

```python
def test_we_map_if_the_file_is_not_in_the_course(...):
    assert_file_in_course.side_effect = CanvasFileNotFoundInCourse
    
    public_url_for_file(...)
    
    map_file.assert_called_once_with(...)
```

#### What's the fix?

We _could_ use `object.patch` to mock `assert_file_in_course` and `find_matching_file_in_course` and instead make assumptions about them and setup their return values. This basically fixes the problem, but there's something icky about it to me.

I also have a niggling feeling that there's an aura of ineffable private-ness about these functions that isn't dispelled by making them public. This is my hatless cat right here. You can't see it by looking at the code, but I think these public functions are secretly private in some way anyway.

Yesterday we described how turning these into private functions probably isn't the way as it leads us away from things we'd like to have like doc-strings and independant tests.

So I think the solution is something like a `FileMapper` helper class in this file. The interface of which would be something like:

```python
class FileMapper:
    def __init__(self, canvas_api_service, file_service):
        ...
        
   def remap_file_id(self, file_id, course, assignment) -> Optional[str]:
       ...
```

In there goes all the stuff like this (however you'd like):

```python
            file = self._file_service.get(effective_file_id, type_="canvas_file")

            if not file:
                raise

            found_file_id = self.find_matching_file_in_course(course_id, file)

            if not found_file_id:
                raise

            # found_file_id is a copy of file_id that the current user *can*
            # see in the current course.

            # Store a mapping so we don't have to re-search next time.
            module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
```

Then we test the crap out of that, however we want to, and crucially we mock it out in the wider tests. This means the outer tests are actually not just integration tests in disguise. They should be really easy to write after that.

This also makes writing tests for this `FileMapper` much easier, because we don't have to worry about the cross product of all the situations that can get us to mapping at the same time as testing mapping itself.

jon-betts:
I've had a stab at a (very) rough POC to show the effect of a helper class on the tests: https://github.com/hypothesis/lms/pull/2921

Basically they are miles easier to write and grok. Although there is a bit of fixture proliferation as we are mocking more now.

seanh:
Thanks for the close look.

> ### There's an edge case where we call the list API twice and some other wackiness
> 
> If we have `check_in_course=True` and a file is in the course, but the user has a permissions error when trying to retrieve it, we'll call the API a second time to list the contents of the course during mapping.

This and other situations where `CanvasAPIClient.list_files()` gets called twice won't result in a second network request because [`CanvasAPIClient` has caching](https://github.com/hypothesis/lms/pull/2907).

> So the way this would play out is, we'd do the check and see the file is in the course. Then we'd fail to get it. Then we'd try to map it, and quite likely map it to itself, which would presumably fail again, and is also weird.
> 
> Some solutions:
> 
> * If we get `CanvasAPIPermissionError` and `check_in_course=True` just raise as we know we probably can't fix it
> * Cache the results of `list_files` so at least we fail quickly
> * Ignore it as it's probably fine / won't happen?

We do already cache the results of `CanvasAPIClient.list_files()`. I'm not sure that entirely removes the possibility of mapping a file to itself in theory. But I think in order to map a file to itself we'd have to get a permissions error when trying to get a public URL for the `file_id` (which implies that this user cannot see this `file_id`) and then, somehow, when we get the course's list of files `file_id` **is** included in it. That makes no sense. But I think it may actually be possible if changes happen in Canvas in-between our `public_url()` and `list_files()` method calls. For example if a file changes from unpublished to published.

I wonder if `find_matching_file_in_course()` should account for this and should not return a file whose ID is the same as the file it's searching for? That'd seem to be a simple solution to avoid ever mapping a file to itself. For good measure `ModuleItemConfiguration.set_canvas_mapped_file_id()` could also raise `ValueError` if `file_id` and `mapped_file_id` are the same.

> This does make me wonder if we should be storing the mapping _before_ we succesfully retrieve the file? Perhaps it should be:
> 
> ```python
> url = self.api.public_url(found_file_id)
> 
> module_item_configuration.set_canvas_mapped_file_id(file_id, found_file_id)
> 
> return url
> ```

Yeah, I think that makes sense regardless. I'm not sure whether I can describe a likely edge case where that would actually make a difference in practice. And I'm not sure that change on its own makes it 100% impossible for us to end up mapping a file to itself, at least in theory. But logically I don't think we want to store a mapped file ID that we failed to get a public URL for, so I think we should make this change.

seanh:
> ### We lookup by `effective_id`, but I think it should be the original id

This one is kind of hard to wrap your head around, so I think maybe we should just use them both. I think it only makes a difference in some quite obscure edge cases involving chains of multiple copied courses and renaming files. I'd like to send that as a separate PR though

seanh:
> ### The splits we have in the code aren't buying us the usual mocking benefits
> 
> ... the way we've broken up the problem isn't resulting in the usual benefits we see from mocking.
> 
> ...
> 
> In the case of the Canvas API, we separated out `assert_file_in_course` to test separately, but we aren't mocking it afterwards, so we actually have to recreate the situation that causes it to raise.
> 
> ...
> 
> The net effect is we've taken all of these elements and split them out, but our testing is still largely forced to recreate an identical scenario to test it as if they were all still private functions. It's pretty much identical to what'd we'd do if it was an end-to-end integration test (I suspect the tests would work find if you used the real file service for example).

Yep. I think there is some benefit to `assert_file_in_course()` and `find_matching_file_in_course()` having their own tests: the specific behaviours of those methods can be tested more easily by calling them directly, and then you don't have to test all those specific behaviours more indirectly in the higher-level `public_url_for_file()` tests. But by not mocking  `assert_file_in_course()` and `find_matching_file_in_course()`, `public_url_for_file()`'s tests aren't getting the full benefits.

> We could mock out the method on the object with `object.patch` but it's a bit gross.
> 
> ...
> 
> We _could_ use `object.patch` to mock `assert_file_in_course` and `find_matching_file_in_course` and instead make assumptions about them and setup their return values. This basically fixes the problem, but there's something icky about it to me.

I agree. We absolutely _could_ do that. But I think it's gross too.

> In the case of the file mappings, as the methods are on the model object, and we don't mock model objects ... They are trivial to mock / use directly respectively. 

Yep. I think given that models should only contain simple methods with no side-effects (other than potentially mutating the model instance itself) this is often going to be the case: it's fine not to mock them.

> The problem is in the local public methods which aren't mocked.

Agree, these would be better if mocked.

> I also have a niggling feeling that there's an aura of ineffable private-ness about these functions that isn't dispelled by making them public. This is my hatless cat right here. You can't see it by looking at the code, but I think these public functions are secretly private in some way anyway.

Yeah, I agree. I think there are times when a method is actually only called privately but it "feels like" a public method so you can justify making it public, or it might even be the better choice to make it public. But in this case I think `assert_file_in_course()` and `find_matching_file_in_course()` are a stretch as public methods. They aren't public for the right reasons.

> So I think the solution is something like a `FileMapper` helper class in this file. The interface of which would be something like:
> 
> ```python
> class FileMapper:
>     def __init__(self, canvas_api_service, file_service):
>         ...
>         
>    def remap_file_id(self, file_id, course, assignment) -> Optional[str]:
>        ...
> ```

Yep, I think this is another **huge** improvement.

I actually wanted `assert_file_in_course()` and `find_matching_file_in_course()` to be `CanvasAPIClient` methods for this reason: then I could test them in `CanvasAPIClientTest` and the higher-level `CanvasService`'s tests would get mocks of them. But of course they can't go in `CanvasAPIClient` because they're not the type of low-level method that belongs in there. Hence they became separate methods of `CanvasService` with their own tests which brings some benefits, but `CanvasService.public_url_for_file()`'s tests don't mock them so they don't get the usual benefits of that.

I don't know why I didn't think of moving the methods into a helper, especially since helpers were at one time an explicit design pattern in the LMS app (all those `helpers.py` files). I think that might actually have been a benefit of the `helpers.py` organizational pattern: it keeps the helpers design pattern top of mind.

Anyway, moving these into a helper is **much** better:

* `CanvasService` doesn't have weird public methods that shouldn't really be there
* The tests are improved by mocking

seanh:
@jon-betts This is ready for re-review. Changes:

* I extracted `assert_file_in_course()` and `find_file_in_course()` into a helper and mocked them in `public_url_for_file()`'s tests: https://github.com/hypothesis/lms/pull/2899/commits/170b6850d22bbb4a73435d5979ed12b990892939

  I went for `CanvasFileFinder` rather than `CanvasFileMapper` as the name for the helper. It doesn't actually map anything (the `{get,set}_canvas_mapped_file_id()` calls are still in `public_url_for_file()`). Both of its methods are different takes on finding file IDs:

  ```python
  class CanvasFileFinder:
      """A helper for finding file IDs in the Canvas API."""

      def assert_file_in_course(self, file_id: str, course_id: str) -> bool:
          """Raise if the current user can't see file_id in course_id."""

      def find_matching_file_in_course(self, course_id: str, file_id: str) -> Optional[str]:
        """Return the ID of a file in course_id that matches the file with ID file_id."""
  ```

* Made sure it won't map a file ID to itself: https://github.com/hypothesis/lms/pull/2899/commits/5b4ddf0fa7493f32cd77d97930dcf1e39415365e

* Don't map a file ID until _after_ we successfully get a public URL for it: https://github.com/hypothesis/lms/pull/2899/commits/43da806f699dde7fe8b0898ebebcf4eba60d9b6c

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2927

Issue #2759, opened 2021-05-31
Blackboard API authorization popup window is too small


seanh:
The popup window is too short. You have to scroll down to see the <kbd>Allow</kbd> button:

![Screenshot from 2021-05-26 19-00-59](https://user-images.githubusercontent.com/22498/119709101-bfd6fc80-be54-11eb-9c09-0ff61ad8f080.png)

robertknight:
The requested inner size (ie. ignoring window chrome) of the window is currently hardcoded to 475px in `AuthWindow#authorize`.

seanh:
Note that the size should remain 475px _for Canvas_, but it needs to be bigger for Blackboard

seanh:
Or maybe the Canvas dialog would still look just fine, with the bigger size? Could try it and see

LMS007:
@seanh how exactly does one reproduce this popup window? When I override my auth token for my LMS instance by overriding the columns in the oauth2_token table, I can trigger the error case when trying to select a blackboard file, but the only thing that happens is I see a blank page and it goes to a different url than the one in your screen shot.

http://localhost:8001/api/blackboard/oauth/callback?code=

![bb_auth](https://user-images.githubusercontent.com/3939074/123449942-02454380-d591-11eb-994d-c2830506db46.gif)

This fix here is likely trivial, as Rob mentioned, but I could use a little more context around this use case so I can repro it correctly. Thanks!

seanh:
If the user has previously authorized Hypothesis to access the Blackboard API then Blackboard remembers this and the popup window just closes immediately after opening. And since we use shared Blackboard accounts the access could have already been authorized by someone else.

To see the popup window you need to revoke access. Go to https://aunltd-test.blackboard.com/ultra/tools, click on **Application Authorization**, and revoke Hypothesis Dev.

I think we can probably go ahead and make the popup window bigger in all cases (Canvas and Blackboard), rather than having different size for each.

LMS007:
Great, thanks for detailing that Sean.

PR #2900, opened 2021-06-29
Increase height of AuthWindow popup

LMS007:
The fixed size auth window is a bit too small for Blackboard's content to fit. Increase the height to 560px which is just big enough to fit the content without a vertical scroll bar.

-----------

### Screen shots are from this branch

Screen Shot 2021-06-29 at 9 25 11 AM
Screen Shot 2021-06-29 at 9 25 22 AM

fixes https://github.com/hypothesis/lms/issues/2759


robertknight (approved):
LGTM

PR #2902, opened 2021-06-29
Inject current ApplicationInstance into FileService

seanh:
I'm not sure [what I was thinking](https://github.com/hypothesis/lms/pull/2892#discussion_r660528549) but the higher-level code that's going to call `FileService` ends up simpler if `FileService`'s factory injects the `ApplicationInstance`. Without this change `CanvasService.public_url_for_file()` ends up with too many arguments. See https://github.com/hypothesis/lms/pull/2899/files#r660780163

### DIFF
--- a/lms/services/file.py
+++ b/lms/services/file.py
@@ -2,19 +2,25 @@ from lms.models import File
 
 
 class FileService:
-    def __init__(self, db):
+    def __init__(self, application_instance, db):
+        self._application_instance = application_instance
         self._db = db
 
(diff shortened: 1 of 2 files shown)


marcospri (approved):
I don't think `too-many-arguments` is a good reason too change this, a smarter pylint would complain about `non-reusable-method` :)

But as I said in the other review we use this pattern in every service that filters by `application_instance`, no reason to change one without agreeing the others should change too.

PR #2906, opened 2021-06-30
Enable canvas groups assignments on new installs

marcospri:
Both sections and groups enabled by default for now.

### DIFF
--- a/tests/unit/lms/views/test_application_instance.py
+++ b/tests/unit/lms/views/test_application_instance.py
@@ -68,6 +68,24 @@ class TestCreateApplicationInstance:
             == canvas_sections_enabled
         )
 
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/test_application_instance.py
```diff
@@ -68,6 +68,24 @@ def test_it_sets_canvas_sections_enabled(
             == canvas_sections_enabled
         )
 
+    @pytest.mark.parametrize(
+        "developer_key,canvas_groups_enabled",
+        [("test_developer_key", True), ("", False)],
+    )
+    def test_it_sets_canvas_groups_enabled(
```
marcospri, line 75: Test is copy pasted from the section one. I think we should test both settings independently.

PR #2908, opened 2021-06-30
Search by both original and mapped file ID

seanh:
When launching a Canvas Files assignment and the `CanvasService.public_url_for_file()` method is called:

If there is a `mapped_file_id` in the DB but we nonetheless don't see the `mapped_file_id` in the current course or we get a permissions error when trying to get a public URL for the `mapped_file_id`, use both the `mapped_file_id` and the assignment's original `file_id` to try to find a matching file in the course.

I think this might fix an obscure Canvas course copy edge case involving renaming files in Canvas and possibly requiring chains of multiple course copies. I'm not sure exactly what the edge case that this covers is, yet. But it shouldn't do any harm.

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -100,14 +100,10 @@ class CanvasFileFinder:
                 continue
 
             for file_dict in self._api.list_files(course_id):
-                display_name = file_dict["display_name"]
-                size = file_dict["size"]
-                id_ = str(file_dict["id"])
-
                 if (
-                    display_name == file.name
-                    and size == file.size
-                    and id_ != file.lms_id
+                    file_dict["display_name"] == file.name
+                    and file_dict["size"] == file.size
+                    and str(file_dict["id"]) != file.lms_id
                 ):
                     return str(file_dict["id"])
 
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -47,7 +47,10 @@ class CanvasService:
             # We'll try to find another copy of the same file that the current
             # user *can* see in the current course and use that instead.
             found_file_id = self._finder.find_matching_file_in_course(
-                course_id, effective_file_id
+                course_id,
+                # Use a set to avoid searching for the same ID twice if file_id
+                # and effective_file_id are the same.
+                {file_id, effective_file_id},
             )
 
             if not found_file_id:
@@ -80,30 +83,33 @@ class CanvasFileFinder:
 
         raise CanvasFileNotFoundInCourse(file_id)
 
-    def find_matching_file_in_course(self, course_id, file_id):
+    def find_matching_file_in_course(self, course_id, file_ids):
         """
-        Return the ID of a file in course_id that matches the file with ID file_id.
+        Return the ID of a file in course_id that matches one of the files in file_ids.
 
         Search for a file that the current Canvas user can see in course_id and
-        that matches the given `file` (same filename and size) and return the
-        matching file's ID.
+        that matches one of the files in file_id's (same filename and size) and
+        return the matching file's ID.
 
         Return None if no matching file is found.
         """
-        file = self._file_service.get(file_id, type_="canvas_file")
+        for file_id in file_ids:
+            file = self._file_service.get(file_id, type_="canvas_file")
 
-        if not file:
-            return None
+            if not file:
+                continue
 
-        file_dicts = self._api.list_files(course_id)
+            for file_dict in self._api.list_files(course_id):
+                display_name = file_dict["display_name"]
+                size = file_dict["size"]
+                id_ = str(file_dict["id"])
 
-        for file_dict in file_dicts:
-            display_name = file_dict["display_name"]
-            size = file_dict["size"]
-            id_ = str(file_dict["id"])
-
-            if display_name == file.name and size == file.size and id_ != file.lms_id:
-                return str(file_dict["id"])
+                if (
+                    display_name == file.name
+                    and size == file.size
+                    and id_ != file.lms_id
+                ):
+                    return str(file_dict["id"])
 
         return None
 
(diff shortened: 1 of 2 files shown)


lms/services/canvas.py
```diff
@@ -80,30 +83,33 @@ def assert_file_in_course(self, course_id, file_id):
...
-
-            if display_name == file.name and size == file.size and id_ != file.lms_id:
-                return str(file_dict["id"])
+        for file_id in file_ids:
+            file = self._file_service.get(file_id, type_="canvas_file")
+
+            if not file:
+                continue
+
+            for file_dict in self._api.list_files(course_id):
```
seanh, line 102: We're calling `self._api.list_files(course_id)` once for each `file_id` in `file_ids` even though we actually only need to call it once total. This won't result in unnecessary network requests because there's [caching on `CanvasAPIClient.list_files()`](https://github.com/hypothesis/lms/blob/b08d57b2d76d7a76e88afab560e1d848242357c5/lms/services/canvas_api/client.py#L211).

We could just put `file_dicts = self._api.list_files(course_id)` above this `for` loop, but then it would call `list_files()` once unnecessarily in the case where none of the `file_id`'s are found in the DB and that _would_ be an unnecessary network request.

It is possible of course to call `list_files()` only when necessary and then only once but doing so complicates the logic in this method a little bit, and doesn't save much due to the caching on `list_files()`. We could do it though. It shouldn't break any tests.

lms/services/canvas.py
```diff
@@ -80,30 +83,33 @@ def assert_file_in_course(self, course_id, file_id):
...
+                display_name = file_dict["display_name"]
+                size = file_dict["size"]
+                id_ = str(file_dict["id"])
+
+                if (
+                    display_name == file.name
+                    and size == file.size
+                    and id_ != file.lms_id
+                ):
```
jon-betts, line 111:
```suggestion
                if (
                    file_dict["display_name"] == file.name
                    and file_dict["size"] == file.size
                    and str(file_dict["id"]) != file.lms_id
                ):
```

lms/services/canvas.py
```diff
@@ -47,7 +47,10 @@ def public_url_for_file(
             # We'll try to find another copy of the same file that the current
             # user *can* see in the current course and use that instead.
             found_file_id = self._finder.find_matching_file_in_course(
-                course_id, effective_file_id
+                course_id,
+                # Use a set to avoid searching for the same ID twice if file_id
+                # and effective_file_id are the same.
+                set([file_id, effective_file_id]),
```
jon-betts, line 53:
```suggestion
                {file_id, effective_file_id},
```

lms/services/canvas.py
jon-betts: Yeah I think this is reasonable. In the previous solution I built a set of keys from the size and display name which you could check easily. In this case though we've started additionally asserting that the ids don't match, so I don't think that would be as neat as it was.

You could do a comprehension to get the files first, then do the loop etc... but I think this is fine?

tests/unit/lms/services/canvas_test.py
```diff
@@ -65,7 +65,7 @@ def test_if_the_file_isnt_in_the_course_it_finds_a_matching_file_instead(
         url = public_url_for_file(sentinel.file_id, check_in_course=True)
 
         canvas_file_finder.find_matching_file_in_course.assert_called_once_with(
-            sentinel.course_id, sentinel.mapped_file_id
+            sentinel.course_id, set([sentinel.file_id, sentinel.mapped_file_id])
```
jon-betts, line 68:
```suggestion
            sentinel.course_id, {sentinel.file_id, sentinel.mapped_file_id}
```

tests/unit/lms/services/canvas_test.py
```diff
@@ -105,7 +105,7 @@ def test_if_theres_a_permissions_error_it_finds_a_matching_file_instead(
         url = public_url_for_file(sentinel.file_id)
 
         canvas_file_finder.find_matching_file_in_course.assert_called_once_with(
-            sentinel.course_id, sentinel.mapped_file_id
+            sentinel.course_id, set([sentinel.file_id, sentinel.mapped_file_id])
```
jon-betts, line 108:
```suggestion
            sentinel.course_id, {sentinel.file_id, sentinel.mapped_file_id}
```

tests/unit/lms/services/canvas_test.py
```diff
@@ -209,20 +209,54 @@ def test_find_matching_file_in_course_returns_the_matching_file_id(
...
+            [sentinel.file_id_1, sentinel.file_id_2, sentinel.file_id_3],
+        )
+
+        # It looked up each file_id in the DB in turn.
+        assert file_service.get.call_args_list == [
+            call(sentinel.file_id_1, type_="canvas_file"),
+            call(sentinel.file_id_2, type_="canvas_file"),
+            call(sentinel.file_id_3, type_="canvas_file"),
+        ]
+        assert matching_file_id == str(sentinel.matching_file_id)
```
jon-betts, line 251: There's quite a lot going on in this test at once. I wonder if this is three different parameterised tests being run at once?

jon-betts (approved):
This looks good. I agree about the for loop. I think it's fine. I made a suggestion to make it shorter by removing the `display_name = file_dict["display_name"]` assignments and using the values directly. The names are clear I think about what's going on.

I've also added suggestions to use set literals, instead of creating lists and converting them to sets after.

I found the test a little hard to grok, as there's three separate paths happening at once in there. If there's a way to parameterise that so you only have to think about one case at a time I think that would be good.

lms/services/canvas.py
seanh: :+1:

lms/services/canvas.py
seanh: :+1:

tests/unit/lms/services/canvas_test.py
seanh: :+1:

tests/unit/lms/services/canvas_test.py
seanh: I simplified it a tiny bit but it's not obvious to me how you'd make it simpler with parametrising

PR #2912, opened 2021-07-01
Rename ModuleItemConfiguration to Assignment

seanh:
Rename `models.ModuleItemConfiguration` to `models.Assignment` everywhere, because "module item configuration" is such a terrible name.

Depends on: <https://github.com/hypothesis/devdata/pull/26>

Known limitations:

* The actual DB table is still named `module_item_configurations`.

  Fixing that would require a DB migration and it might not even be possible, I've never done a DB migration to rename a table before.

* "Module item configuration" still appears in the DB migrations.

  We don't go back and change DB migrations.

* <del>The devdata repo still uses `"module_item_configuration"` as a `type` string which means `bin/update_dev_data.py` still has to use it.</del>

  <del>Fixing this would require a corresponding devdata PR.</del>

  (Done. This PR now depends on <https://github.com/hypothesis/devdata/pull/26>)

* <del>The `"configure_assignment"` route's URL is still `/module_item_configurations`.</del>

  <del>I don't think there's any reason not to fix this one.</del>

  (Done)


seanh:
Rebased

lms/routes.py
```diff
@@ -13,8 +13,9 @@ def includeme(config):
 
     config.add_route("config_xml", "/config_xml")
     config.add_route(
-        "module_item_configurations",
-        "/module_item_configurations",
+        "configure_assignment",
+        "/assignment",
+        request_method="POST",
```
seanh, line 18: This is a private URL that our own frontend code posts to when creating assignments (the backend passes the URL to the frontend in the JS config), so we can change it at will. I just changed it to `/assignment` which seems RESTful: `POST`'ing to `/assignment` to create an assignment. Added `request_method="POST"` to clarify that the route is only used for `POST`'ing to (otherwise I'd have to change the route name from `"configure_assignment"` to just `"assignment"`).

seanh: We might even want to rename the route `"create_assignment"` instead of `"configure_assignment"`

seanh: Note that this really is just for creating new assignments. You cannot edit existing assignments in the LMS app. At least not in non-Canvas LMS's. In Canvas you can, but in Canvas this route isn't used at all and we submit assignment creates and updates to Canvas rather than to ourselves.

seanh:
I'm running into conflicts with this PR (and the corresponding devdata one) when working on other issues now, so I'm going to go ahead and merge this. It can be post-merge reviewed

PR #2913, opened 2021-07-01
Rename ModuleItemConfiguration.{get,set}_file_id_mapping()

seanh:
Remove the `canvas` from `ModuleItemConfiguration.{get,set}_canvas_file_id_mapping()`

### DIFF
--- a/lms/models/module_item_configuration.py
+++ b/lms/models/module_item_configuration.py
@@ -51,8 +51,8 @@ class ModuleItemConfiguration(BASE):
         nullable=False,
     )
 
-    def get_canvas_mapped_file_id(self, file_id):
-        return self.extra.get("canvas_file_mappings", {}).get(file_id)
+    def get_mapped_file_id(self, file_id):
+        return self.extra.get("file_mappings", {}).get(file_id)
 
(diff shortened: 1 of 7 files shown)


seanh:
I'm gonna close this one.

I've realised that it's too late to change these to just "file ID mappings" rather than "Canvas file ID mappings" because a bunch of them have already been stored in `extra.canvas_file_mappings` dicts in prod. We'd have to do a DB migration.

I don't think we should change the method to `{get,set}_file_id_mapping(..., type="canvas")` either. That implies that there are types of file ID mappings other than Canvas ones. This obscures what's really going on: in reality Canvas file ID mappings are the only type of file ID mapping, this is a Canvas-only thing that exists for very specific reasons to do with how the Canvas API works and how course copy works in Canvas. So I think <code>{get,set}_<b>canvas</b>_file_id_mapping()</code> is actually clear and appropriate. Adding a `type` argument would also complicate the code slightly by making it able to deal with multiple types and necessitate testing that, while in practice only one type is ever used.

If we ever need non-Canvas file ID mappings for any reason we can make changes then

PR #2917, opened 2021-07-01
A POC for a slimmer test set for course copy

jon-betts:
### The general idea

In the previous tests each possible scenario was separately enumerated, and then multiple assertions were made about each one. In these tests we attempt to isolate individual features of the scenarios/code and test them separately. The hope is single tests can catch one variable, rather than resulting in a cross product of all variations as scenarios.

This means, for example, we can test that the function returns the value of the Canvas API call once, and not re-assert this unless it's relevant to the feature under test.

A good example of this is the first couple of tests in each. In the previous tests we have:

 * `test_without_check_in_course_and_without_a_mapped_file_id`
 * `test_if_check_in_course_is_True_it_checks_that_the_file_is_in_the_course`
 * `test_if_theres_a_mapped_file_id_it_uses_it`
 
The new tests have cover roughly the same ground with:

* `test_it_calls_the_canvas_api`
* `test_it_calls_the_canvas_api_with_a_mapped_file`
* `test_it_without_check_in_course`
* `test_it_with_check_in_course`
* `test_it_raises_with_check_in_course_if_file_not_in_course`

All of the scenarios above can sort of be "built" from these units.

So we have a larger number of shorter, more tightly focused tests. In the example above the lengths are roughly the same, but overall it's much shorter due to:

 * Elimination of duplicate assertions in different scenarios
 * The possibility of using parameterised tests more easily (as we are only focusing on certain things)
 * Some other ideas and tricks which keep the size down
 
### What's gained and lost with this approach

**Lost**

 * At present these tests make fewer assertions about the calls made to services
 * This could be put back in, but there's some wriggling to be done about the calls made with mapped and un-mapped ids
 * I've not done this because I think this actually is about to change (I'm not sure we should be using `effective_id` + we might go to a multi id mode)
 * If you wanted to read through a particular scenario step by step, and follow along everything the code does, this style could be good

**Gained**

 * The tests are shorter and easier to understand. They typically have shorter setup and assertion phases
 * It's easier to convince yourself of what these tests do, and don't cover
 * It's easier to match these tests to the code they test (for example if you had to rewrite the code blind, this is a more explicit spec)

### Some other things which helped keep the size down

 * Turns out it's sort of easier to have a file fixture and drive matching things backwards (matches in canvas etc) than doing it the other way
 * Often we don't need fixtures and can setup the smaller situations we need directly in the tests
 * Having an if in one fixture to raise if passed "MISSING_ID" made this a lot easier

### DIFF
--- a/tests/unit/lms/services/canvas_test.py
+++ b/tests/unit/lms/services/canvas_test.py
@@ -1,3 +1,4 @@
+from functools import partial
 from unittest.mock import DEFAULT, call, sentinel
 
 import pytest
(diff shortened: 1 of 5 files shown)


tests/unit/lms/services/canvas_test.py
```diff
@@ -13,271 +13,165 @@
 
 
 class TestPublicURLForFile:
```
jon-betts, line 15: This diff is a total disaster, I'd recommend just looking at it.

tests/unit/lms/services/canvas_test.py
```diff
@@ -13,271 +13,165 @@
...
+    # A special file id which causes CanvasAPIPermissionError errors
+    MISSING_FILE_ID = "missing_file"
 
-        # check_in_course was False, so it didn't call list_files() to check
-        # whether the file_id was in the course.
-        canvas_service.api.list_files.assert_not_called()
+    def test_it_calls_the_canvas_api(self, canvas_service, public_url_for_file):
+        url = public_url_for_file(file_id="file_id")
 
-        # It calls the Canvas API with the file_id and returns the public URL.
```
jon-betts, line 28: I think the fate of this comment is kind the key to the whole business. To some extent this whole test is now this, and therefore it doesn't require much more commentary beyond the name of the test.

tests/unit/lms/services/canvas_test.py
```diff
@@ -303,18 +197,27 @@ def CanvasService(self, patch):
...
+
+
+@pytest.fixture
+def matching_file_in_course(file_record, canvas_service):
+    # Create a match which is tailored to the file record
+    matching_file_in_course = {
+        "id": "new_file",
+        "display_name": file_record.name,
+        "size": file_record.size,
+    }
```
jon-betts, line 218: Driving the matches backwards this way makes it a bit easier to understand I think. Previously we were pulling items from the return values of `list_files` and working our way backwards.

I think this better reflects how we're thinking of it, which is there is a result in `list_files` that matches our historical record. Rather than our historical record being driven by the results obtained from the list today.

tests/unit/lms/services/canvas_test.py
```diff
@@ -13,271 +13,165 @@
...
-        file_service.get.return_value = None
-
-        with pytest.raises(CanvasAPIPermissionError):
-            public_url_for_file(file_id=str(file_from_a_different_course["id"]))
+    @pytest.fixture(autouse=True)
+    def canvas_api_client(self, canvas_api_client):
+        def public_url(file_id):
+            # Rig public_url to blow if we get MISSING_FILE_ID
+            if file_id == self.MISSING_FILE_ID:
+                raise CanvasAPIPermissionError()
```
jon-betts, line 124: This really helped simplify and unify a lot of scenarios. Now the API just fails if it ever gets called with the bad id.

This means that the `check_in_course` true and false branches do not require different setup (as far as this API client goes).

seanh:
Closing this as I _believe_ its been superseded by https://github.com/hypothesis/lms/pull/2921. @jon-betts correct me if I'm wrong

jon-betts:
Hi @seanh , depends if there's anything interesting here. If you just want to merge some code, then yes, I think the other one is the way to go. However this was intended to demonstrate a difference in style which could still be interesting to talk about.

PR #2919, opened 2021-07-02
Use h-assets package to serve compiled frontend assets

robertknight:
Replace the local code in `lms/assets.py` for serving compiled JS and CSS assets with the [h-assets](https://github.com/hypothesis/h-assets) package. The purpose of this is to reduce duplication between the h and lms projects and to enable future changes to be made in a central place. The initial version of h-assets also fixes a caching hazard by being stricter about cache busting. More details below.

Functionally the views from h-assets work the same as the previous code in `lms/assets.py`, with two differences:

 - Asset responses no longer have an `Access-Control-*` header. This is a hangover from when the Hypothesis client's assets were served from h, from which `lms/assets.py` originated, and was never needed in the lms project.
 - The cache-busting query string in asset URL requests (eg. `/assets/scripts/frontend_apps.bundle.js?1234`) is now verified and a 404 is returned if it is incorrect.

   This prevents a scenario where clients (eg. a browser or Cloudflare) request a version of a static asset ("V2") that is different from what the application is serving ("V1") and then cache V1 of the asset under the URL for V2. This can happen if the request for a subresource (eg. a script or stylesheet) hits a different version of the application than the resource (eg. an HTML page) which referenced the subresource.

Part of https://github.com/hypothesis/lms/issues/2799

----

**Testing:**

- Make sure that the various views (eg. assignment configuration, assignment launch) that reference compiled frontend assets still work
- Get the current URL of one of the assets (eg. from the browser's devtools) and try fetching it with the original query string and different query strings. You should get a 200 with the correct content if the cache buster is valid and a 404 if not


jon-betts (commented):
I'm going to hang fire on this for a little bit, as it's almost identical to: https://github.com/hypothesis/h/pull/6752.

Once that one is out and good I think we can just replicate any changes from there to here.

robertknight:
This PR has been rebased and `lms.assets` has been updated to match the latest version of `h.assets` from https://github.com/hypothesis/h/pull/6752. One difference in the lms app vs h is that I've tried to avoid adding a new setting to control asset reloading in the lms app and have instead relied on the existing "dev" setting which is also used in a few other places. I wasn't sure that it was safe to rely on this setting always being a boolean but that looks OK according to [this Slack thread](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1627573621062800).

jon-betts (approved):
Yep this is working well locally.

It's a little tricky to see many assets in LMS but I can see:

 * Auth dialog
 * File picker 
 * Admin pages
  
So I think this is looking good

robertknight:
I need to confirm but I think this is going to run into the same issue with sourcemap URLs as was found in h. This is addressed in https://github.com/hypothesis/h-assets/pull/7 which will make h-assets initially less strict about cache-busters (only check if present).

robertknight:
I have updated this branch to use a newer version of h-assets which includes https://github.com/hypothesis/h-assets/pull/7. This avoids a problem with browsers failing to fetch sourcemaps when the devtools are open due to sourcemap URLs in bundles not having cache-busting query strings.

PR #2921, opened 2021-07-02
A POC to demonstrate the effect of a helper object on course copy tests

jon-betts:
This POC introduces a helper object with public and tested methods (tests not included here) to show the effect it has on the parent tests of the object when we mock it out.

I think these are the clearest ones yet. Even if you disagree with the precise way these were done, I think it's relatively obvious that these are easier to write and grok.

### What's going on here?

 * This introduces a `CanvasFileMapper` which does the asserting that files are in courses and mapping them
 * It's not a good name, but that's besides the point
 * We then assume this is well tested in a normal way
 * We then mock it out when testing the parent object

I think the result is smaller tests which look much more like our regular tests, and don't have to go to such lengths to setup fixtures and scenarios.

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -45,6 +45,10 @@ class CanvasService:
             file_ = self._file_service.get(effective_file_id, type_="canvas_file")
 
             if not file_:
+                # We don't have a record of the assignment's file in our DB.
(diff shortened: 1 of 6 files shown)


seanh:
Closing this, I've integrated it into https://github.com/hypothesis/lms/pull/2899

PR #2926, opened 2021-07-05
Example of some changes suggested during review

jon-betts:
This is an example of some of the suggestions made reviewing: https://github.com/hypothesis/lms/pull/2844

### DIFF
--- a/lms/db/_bulk_action.py
+++ b/lms/db/_bulk_action.py
@@ -93,19 +93,23 @@ class BulkAction:
 
         if config.upsert_trigger_onupdate:
             onupdate_columns = self._get_columns_onupdate(model_class)
+
             for column_name, onupdate_value in onupdate_columns:
                 upsert_update_elements.append(column_name)
+
+                # SQL alchemy wraps functions passed to onupdate or default and
+                # could potentially take a "context" argument getting a
+                # suitable context at this point of the execution it's not
+                # possible so we don't support it so we just pass None
+                # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
+                default_value = (
+                    onupdate_value(None) if callable(onupdate_value) else onupdate_value
+                )
+
                 # Copy the values, we don't want to mess with the caller's data
                 values = deepcopy(values)
-                for value_ in values:
-                    value_[column_name] = (
-                        # SQL alchemy wraps functions passed to onupdate or default and could potentially take a "context" argument
-                        # getting a suitable context at this point of the execution it's not possible so we don't support it so we just pass None
-                        # https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions
-                        onupdate_value(None)
-                        if callable(onupdate_value)
-                        else onupdate_value
-                    )
+                for row in values:
+                    row[column_name] = default_value
 
         stmt = insert(model_class).values(values)
         stmt = stmt.on_conflict_do_update(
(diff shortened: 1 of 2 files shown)


tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -57,78 +46,37 @@ def test_upsert(self, db_session):
...
         db_session.add_all(
             [
-                self.TableWithBulkUpsertOnUpdate(
-                    id=1, name="existing", scalar=0, callable=0, sql=0, default=0
-                ),
-                self.TableWithBulkUpsertOnUpdate(
-                    id=2, name="existing", scalar=1, callable=1, sql=1, default=1
-                ),
+                self.TableWithBulkUpsert(id=1, name="pre_existing_1", **{column: 0}),
+                self.TableWithBulkUpsert(id=2, name="pre_existing_2", **{column: 1}),
```
jon-betts, line 66: I followed the naming as above here with `pre_...` so the two can be seen apart by name.

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -57,78 +46,37 @@ def test_upsert(self, db_session):
...
-                        },
-                    ]
-                ]
-            ).only()
+        self.assert_has_rows(
+            db_session,
+            # 42 is the onupdate default value
+            {"id": 1, "name": "update_existing", column: 42},
+            {"id": 2, "name": "pre_existing_2", column: 1},
         )
```
jon-betts, line 80: I think this assertion is easier to read now, even if this just means some of the complexity has been shoved elsewhere.

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -145,3 +93,22 @@ class MisconfiguredModel:  # pylint: disable=unused-variable
...
+                    for expected in attrs
+                ]
+            ).only()
+        )
+
+    @pytest.fixture
+    def with_upsert_trigger_onupdate(self):
+        self.TableWithBulkUpsert.BULK_CONFIG.upsert_trigger_onupdate = True
+        yield
+        self.TableWithBulkUpsert.BULK_CONFIG.upsert_trigger_onupdate = False
```
jon-betts, line 114: If we don't like this, it might be possible to get what we want by sub-classing the other one? Not sure. Either way, we might want to pick shorter names for the tables, as that really bloats the line length by causing black to insert a lot of new lines in certain cases.

We can probably get what type of table this is from context. So it could probably be `TestModel` without losing too much.

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -57,78 +46,37 @@ def test_upsert(self, db_session):
...
+            {"id": 2, "name": "pre_existing_2", "other": "pre_2"},
+            {"id": 3, "name": "create_with_id", "other": "post_3"},
+            {"id": 4, "name": "over_block_size", "other": "post_4"},
         )
 
     def test_upsert_does_nothing_if_given_an_empty_list_of_values(self, db_session):
         assert BulkAction(db_session).upsert(self.TableWithBulkUpsert, []) == []
 
-    def test_upsert_with_onupdate_columns(self, db_session):
+    @pytest.mark.parametrize("column", ("scalar", "callable", "sql", "default"))
```
marcospri, line 60: Much nicer :+1:

lms/db/_bulk_action.py
```diff
@@ -93,19 +93,23 @@ def upsert(self, model_class, values):
 
         if config.upsert_trigger_onupdate:
```
marcospri, line 94: I think you are right that this name could lead to confuging (are DB triggers involvced here?)

Not sure which better name we could use `upsert_orm_onupdate`? `upsert_onupdate`?

jon-betts: `upsert_call_onupdate`? `upsert_use_onupdate`?

marcospri: `upsert_use_onupdate` Looks good to me

Issue #2031, opened 2020-08-19
Blackboard Files pagination support


seanh:
Add pagination support to the Blackboard file list API that was created in https://github.com/hypothesis/lms/issues/2030

PR #2928, opened 2021-07-05
Add Blackboard Files pagination support

seanh:
Fixes https://github.com/hypothesis/lms/issues/2031

This adds a little logic to a view but it's only a little: one `for`, one `if` and one `break`. I think it's acceptable. The view is still only  13 lines (including blanks) and I don't think its tests will be too difficult.

**Let's not generalize this:** it's probable that the `paging.nextPage` field is a common field that the Blackboard API sends in responses to different endpoints, and that this code could be put in a generic place where it could be re-used by multiple endpoints.  I haven't investigated that because I don't need to for the Blackboard Files feature. `list_files()` is the only Blackboard API endpoint that we call that returns a list of results. And we have no current plans to add any other Blackboard API-related features. If and when we do add support for another paginated Blackboard API endpoint we can make this code generic then.

### DIFF
--- a/lms/services/blackboard_api.py
+++ b/lms/services/blackboard_api.py
@@ -102,7 +102,13 @@ class BlackboardAPIClient:
 
     def _api_url(self, path):
         """Return the full Blackboard API URL for the given path."""
-        return f"https://{self.blackboard_host}/learn/api/public/v1/{path}"
+
+        if not path.startswith("/"):
+            # Paths that don't start with "/" are treated as relative to this
+            # common Blackboard API path prefix.
+            path = "/learn/api/public/v1/" + path
+
+        return f"https://{self.blackboard_host}{path}"
 
 
 def factory(_context, request):
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -18,6 +18,13 @@ DOCUMENT_URL_REGEX = re.compile(
     r"blackboard:\/\/content-resource\/(?P<file_id>[^\/]*)\/"
 )
 
+# The maxiumum number of paginated requests we'll make before returning.
+PAGINATION_MAX_REQUESTS = 25
+
+# The maximum number of results to request per paginated response.
+# 200 is the highest number that Blackboard will accept here.
+PAGINATION_LIMIT = 200
+
 
 @view_defaults(permission=Permissions.API, renderer="json")
 class BlackboardFilesAPIViews:
@@ -29,10 +36,18 @@ class BlackboardFilesAPIViews:
     def list_files(self):
         """Return the list of files in the given course."""
         course_id = self.request.matchdict["course_id"]
-        response = self.blackboard_api_client.request(
-            "GET", f"courses/uuid:{course_id}/resources"
-        )
-        return BlackboardListFilesSchema(response).parse()
+
+        files = []
+        path = f"courses/uuid:{course_id}/resources?limit={PAGINATION_LIMIT}"
+
+        for _ in range(PAGINATION_MAX_REQUESTS):
+            response = self.blackboard_api_client.request("GET", path)
+            files.extend(BlackboardListFilesSchema(response).parse())
+            path = response.json().get("paging", {}).get("nextPage")
+            if not path:
+                break
+
+        return files
 
     @view_config(request_method="GET", route_name="blackboard_api.files.via_url")
     def via_url(self):
--- a/tests/unit/lms/views/api/blackboard/files_test.py
+++ b/tests/unit/lms/views/api/blackboard/files_test.py
@@ -1,8 +1,13 @@
+from unittest.mock import call, sentinel
+
 import pytest
 
 from lms.services import HTTPError
 from lms.views.api.blackboard.exceptions import BlackboardFileNotFoundInCourse
-from lms.views.api.blackboard.files import BlackboardFilesAPIViews
+from lms.views.api.blackboard.files import (
+    PAGINATION_MAX_REQUESTS,
+    BlackboardFilesAPIViews,
+)
 from tests import factories
 
 pytestmark = pytest.mark.usefixtures("oauth2_token_service", "blackboard_api_client")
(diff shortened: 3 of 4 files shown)


lms/services/blackboard_api.py
```diff
@@ -102,7 +102,13 @@ def request(self, method, path):
 
     def _api_url(self, path):
         """Return the full Blackboard API URL for the given path."""
-        return f"https://{self.blackboard_host}/learn/api/public/v1/{path}"
+
+        if not path.startswith("/"):
+            # Paths that don't start with "/" are treated as relative to this
+            # common Blackboard API path prefix.
+            path = "/learn/api/public/v1/" + path
```
seanh, line 109: This is needed because the `nextPage`'s that Blackboard sends us already have the `"/learn/api/public/v1/"` on them. As a bonus if we even want to use `BlackboardAPIClient` to call an endpoint that doesn't begin with `"/learn/api/public/v1/"` at all we can now do so: just start the path with a `"/"`

lms/views/api/blackboard/files.py
```diff
@@ -18,6 +18,13 @@
     r"blackboard:\/\/content-resource\/(?P<file_id>[^\/]*)\/"
 )
 
+# The maxiumum number of paginated requests we'll make before returning.
+PAGINATION_MAX_REQUESTS = 25
```
jon-betts, line 22: Holy moley... I think this equates to 5000 files. How long does each round take?

lms/services/blackboard_api.py
jon-betts: An alternative to this is to accept a `url` as well as a `path` and only run the path through `_api_url` if `url` is missing.

lms/views/api/blackboard/files.py
seanh: Yeah, it's the same as we do for Canvas: 25 * 200 files. I can't remember why we chose that for Canvas. We might want to reduce it to say 10

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -19,14 +24,99 @@ def test_it(
...
+        ]
+        blackboard_list_files_schema.parse.side_effect = [
+            # The first response from the Blackboard API contains the first
+            # three files.
+            [sentinel.file_1, sentinel.file_2, sentinel.file_3],
+            # The second page's files are in the second response.
+            [sentinel.file_4, sentinel.file_5, sentinel.file_6],
+            # The final page of files.
+            [sentinel.file_7, sentinel.file_8],
+        ]
```
jon-betts, line 66: The way this is laid out makes it a bit hard for me to work out what is being setup here. I think in combination with reducing the number of return values we can slim this down:

```suggestion
        # The first response from the Blackboard API includes the path to
        # the next page in the JSON body.

        # The Blackboard API gives us the whole path to the next
        # page, including limit and offset query params, as a
        # string. For example:
        # "/learn/api/public/v1/courses/uuid:<ID>/resources?limit=200&offset=200"

        blackboard_api_client.request.side_effect = [
            factories.requests.Response(json_data=data)
            for data in [
                {"paging": {"nextPage": "PAGE_2_PATH"}},
                {"paging": {"nextPage": "PAGE_3_PATH"}},
                {},
            ]
        ]

        # Multiple sets of values which are returned one after another
        blackboard_list_files_schema.parse.side_effect = [
            [sentinel.file_1],
            [sentinel.file_2],
            [sentinel.file_3],
        ]
```

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -19,14 +24,99 @@ def test_it(
...
+        return blackboard_api_client
+
+    @pytest.fixture(autouse=True)
+    def blackboard_list_files_schema(self, blackboard_list_files_schema):
+        blackboard_list_files_schema.parse.return_value = [
+            sentinel.file_1,
+            sentinel.file_2,
+            sentinel.file_3,
+        ]
+        return blackboard_list_files_schema
```
jon-betts, line 117: I don't think this is actually required for anything other than the `test_it` test, so this could probably be removed.

```suggestion
```

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -19,14 +24,99 @@ def test_it(
...
+
+    @pytest.fixture(autouse=True)
     def pyramid_request(self, pyramid_request):
         pyramid_request.matchdict["course_id"] = "COURSE_ID"
         return pyramid_request
```
jon-betts, line 122: You can't see it here, but I think every fixture at the bottom of the apart from `views` only applies to one of the test classes or another, and could probably be local to that test class.

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -19,14 +24,99 @@ def test_it(
...
+        assert files == [
+            sentinel.file_1,
+            sentinel.file_2,
+            sentinel.file_3,
+            sentinel.file_4,
+            sentinel.file_5,
+            sentinel.file_6,
+            sentinel.file_7,
+            sentinel.file_8,
+        ]
```
jon-betts, line 86: If you merged the change above this would just be:

```suggestion
        assert files == [sentinel.file_1, sentinel.file_2, sentinel.file_3]
```

jon-betts (approved):
Yep I think this is the right call. If we ever happen to need this on another end-point we can tackle what to do then.

I think this is all good and works but I think the tests can be a little snappier and I've added some suggestions:

 * I think the same fixtures can be expressed in a clearer way by moving the comments above rather than within the fixtures
 * One of the fixtures is only used in one test, but is set up for all tests
 * All but one of the fixtures at the bottom of the file are specific to one or the other test classes and could be moved there. I think this would keep them more local to those they relate to. It's also really handy if you ever come to refactor the tests into different files as you can scoop a class up and get almost all of the fixtures it needs with it.

tests/unit/lms/views/api/blackboard/files_test.py
seanh: :+1: Done

tests/unit/lms/views/api/blackboard/files_test.py
seanh: :+1:

tests/unit/lms/views/api/blackboard/files_test.py
seanh: The `pyramid_request()` one is also needed by all three tests. I localized all the others

Issue #2930, opened 2021-07-06
Only show PDFs from Blackboard


seanh:
When getting the list of the course's files from the Blackboard API (when a teacher is creating a new assignment) only show PDF files, not other types of file.

seanh:
The API docs are here: https://developer.blackboard.com/portal/displayApi

I don't see any way to request only files of a particular type when calling the API. I think we may need to get all the files from the API and then filter the list down to only PDFs ourselves. I think there might be a `mimeType` field (`"application/pdf"`) that we can use in the file dicts that the API returns. If not we may be reduced to filtering by filename extension even though there's no guarantee that a PDF file will have a PDF filename extension (or any filename extension).

PR #2931, opened 2021-07-06
Only show PDFs from Blackboard

seanh:
Fixes https://github.com/hypothesis/lms/issues/2930

When getting the list of a course's files from the Blackboard API when creating an assignment only show PDF files.

As far as I can tell there's no Blackboard API parameter to request only a particular type of file. We just have to retrieve all the files ourselves and then filter them based on their `mimeType` field.

Blackboard API docs:

https://developer.blackboard.com/portal/displayApi

### DIFF
--- a/lms/views/api/blackboard/_schemas.py
+++ b/lms/views/api/blackboard/_schemas.py
@@ -17,6 +17,7 @@ class _FileSchema(Schema):
     id = fields.Str(required=True)
     display_name = fields.Str(data_key="name", required=True)
     updated_at = fields.Str(data_key="modified", required=True)
+    mime_type = fields.Str(data_key="mimeType")
 
     @post_load
     def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
@@ -31,7 +32,15 @@ class BlackboardListFilesSchema(RequestsResponseSchema):
 
     @post_load
     def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        return data["results"]
+        pdf_files = []
+
+        for file in data["results"]:
+            if file.get("mime_type") == "application/pdf":
+                # Delete mime_type: we don't want to send it to the frontend.
+                del file["mime_type"]
+                pdf_files.append(file)
+
+        return pdf_files
 
 
 class BlackboardPublicURLSchema(RequestsResponseSchema, _FileSchema):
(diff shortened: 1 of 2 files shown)


lms/views/api/blackboard/_schemas.py
```diff
@@ -17,6 +17,7 @@ class Meta:
     id = fields.Str(required=True)
     display_name = fields.Str(data_key="name", required=True)
     updated_at = fields.Str(data_key="modified", required=True)
+    mime_type = fields.Str(data_key="mimeType")
```
seanh, line 20: I've made `mime_type` an optional field. If it's missing we'll just assume the file is not a PDF. Folders in Blackboard API responses don't have `mime_type`'s so making the field optional will be useful / necessary in future when we add folder support.

lms/views/api/blackboard/_schemas.py
```diff
@@ -31,7 +32,15 @@ class BlackboardListFilesSchema(RequestsResponseSchema):
...
-        return data["results"]
+        pdf_files = []
+
+        for file in data["results"]:
+            if file.get("mime_type") == "application/pdf":
+                # Delete mime_type: we don't want to send it to the frontend.
+                del file["mime_type"]
+                pdf_files.append(file)
+
+        return pdf_files
```
seanh, line 43: Following the existing design here `BlackboardListFilesSchema` is responsible for converting the files into the format expected by our frontend: renaming fields, omitting fields that we don't use, etc. `BlackboardListFilesSchema` and `TestBlackboardListFilesSchema` are the only places in our code where knowledge of the Blackboard API response format exists. The view just treats the list of dicts returned by `BlackboardListFilesSchema` as opaque and blindly returns them to the frontend.

At least for now this seems to be the best design. In earlier PRs we tried a version where the schema just validates the responses and doesn't mutate them at all, and the view does the mutation. But it just made the code and tests longer and spread knowledge of the Blackboard API response format around.

I think this could likely change in future as requirements change

marcospri: > I think this could likely change in future as the code design changes in response to changing requirements

Yeah I've seen the changes for folders and even If I'm not a huge fan of the schema doing a lot of work this is simple enough to be easily change later on.

marcospri (approved):

PR #2932, opened 2021-07-06
Separate file picker config

jon-betts:
This PR moves the generation of individual config for file pickers into their own collaborator class. This results in more code overall, but this gives us some benefits in return:

* The generation of the config for each file type is completely separate and self contained
* This will make them easy to refactor if we ever chose to re-orient the config around LMS type for example
* They can all be tested independently
* The tests are a bit lower magic
* This goes some way to cutting down the very big `_js_config` class and tests


lms/resources/_js_config/file_picker_config.py
```diff
@@ -0,0 +1,101 @@
+class FilePickerConfig:
+    """A config generator for specific file pickers."""
+
+    def __init__(self):
+        self._config_mapping = {
+            "blackboard": self.blackboard_config,
+            "canvas": self.canvas_config,
+            "google": self.google_files_config,
+            "vitalSource": self.vital_source_config,
+        }
```
seanh, line 10: This can maybe move into `get_config()` below as that's the only place where it's used

seanh (approved):
:+1:, Looks simple enough, and we do need to start to find ways to break up the massive `JSConfig` class

tests/unit/lms/models/application_instance_test.py
```diff
@@ -34,6 +34,9 @@ def test_provisioning_is_not_nullable(self, db_session, application_instance):
 
         assert application_instance.provisioning is not None
 
+    @pytest.mark.skip("This test is wrong, the column is non nullable with a default")
+    # One problem here is we are testing the factory, not the class. But even
+    # if you do get use the class directly it seems to end up with None?
```
seanh, line 39: I think `settings` may default to `None` on a new `ApplicationInstance` that's been created in Python but not flushed to the DB yet because the `server_default` doesn't trigger until DB flush. Not sure how `ApplicationSettings` interacts with that. Perhaps we can just delete this test

jon-betts: I tried flushing and querying this :( still didn't seem to work

lms/resources/_js_config/file_picker_config.py
jon-betts: I actually moved this into the `__init__.py` stuff as it follows the pattern we saw in Course Copy. There's a testing gap in this version, as we don't mock ourselves. So instead of doing that, I've done it in the caller instead.

PR #2933, opened 2021-07-06
Add Blackboard Files folder support

seanh:
Labeled **do not merge** because this PR only has the backend changes for folder support, which breaks the frontend. The corresponding frontend changes are needed before merging this.

This PR:

1. Adds a new API `/api/blackboard/courses/{course_id}/folders/{folder_id}/files` for getting all the files and folders in a given sub-folder of a course

2. Changes both the existing `/api/blackboard/courses/{course_id}/files` API (all top-level files and folders in a course) and the new `/api/blackboard/courses/{course_id}/folders/{folder_id}/files` API to return both files and folders.

   Files look the same as they did before but with a new `type: "File"` field.

   Folders have an `id` that is *not* wrapped in a `blackboard://` URL, they have a `parent_id` (the ID of the parent folder or `None`) and they have a `contents` URL for the API call to get the contents of the folder. Plus they have `type: "Folder"`

## Code design

The layers I've gone for here are:

1. The already-existing `BlackboardAPIClient` service is the low-level Blackboard API client with just a generic `request()` method. Things that are general to all Blackboard API requests go here. For example: prepending the Blackboard host to URL paths; handling access token-related errors; getting and refreshing access tokens.

2. The new `BlackboardService` is the high-level Blackboard service built on top of `BlackboardAPIClient`. `BlackboardService` knows about particular Blackboard API endpoints and provides high-level convenience methods for them. For example `BlackboardService.list_files(course_id, folder_id)` constructs the necessary URL to call to get the files for `course_id` and `folder_id` and calls `BlackboardAPIClient` to make the request.

   The `BlackboardListFilesSchema` that knows about the expected format of list files responses lives in `BlackboardService`, so `BlackboardService.list_files(...)` will either return a list of dicts in the expected format or it'll raise.

   Pagination also lives in `BlackboardService.list_files()`. Pagination is probably actually general and belongs in the lower-level `BlackboardAPIClient`. But for now `list_files()` is the only endpoint that we do pagination for so it's not generalized and just lives in here.

   `BlackboardService` does not know anything about our proxy API and the consistent response format that our proxy API returns (as expected by our frontend) which is different from the particular response formats of APIs like Blackboard's or Canvas's. `BlackboardService` just returns the lists of dicts as returned by Blackboard (but it validates them to make sure Blackboard isn't sending us unexpected responses).

3. `BlackboardFilesAPIViews` implements our Blackboard proxy API. It knows what the path params and or query params of our proxy API are: reading `course_id` and `folder_id` out of `request.matchdict`. It calls `BlackboardService` to get the list of files. And it knows what the response format of our proxy API is: it loops over the list of files from `BlackboardService` and mutates them.

So that's:

1. `BlackboardAPIClient`: low-level stuff that applies to all Blackboard API requests
2. `BlackboardService`: specific high-level Blackboard API methods like `list_files(course_id, folder_id)` and validation
3. `BlackboardFilesAPIViews`: layers our proxy API on top of `BlackboardService`

#### Just put it all in the view

All I'm actually trying to do is add folder support to `BlackboardFilesAPIViews.list_files()`. So I could just skip the `BlackboardService` and put all the code in the view. This has the advantage of keeping everything together in one place, instead of distributing work and knowledge between service and view. The only downside to this is that it might be making the view too long, adding too much logic to the view:

```python
    @view_config(
        request_method="GET", route_name="blackboard_api.courses.files.list_folder"
    )
    def list_files(self):

        folder_id = self.request.matchdict.get("folder_id")

        results = []
        path = f"courses/uuid:{course_id}/resources"

        if folder_id:
            # Get the files and folders in the given folder instead of the
            # course's top-level files and folders.
            path += f"/{folder_id}/children"

        path += f"?limit={PAGINATION_LIMIT}"

        for _ in range(PAGINATION_MAX_REQUESTS):
            response = self.blackboard_api_client.request("GET", path)
            results.extend(BlackboardListFilesSchema(response).parse())
            path = response.json().get("paging", {}).get("nextPage")
            if not path:
                break

        for result in results:
            if result["type"] == "File":
                result["id"] = f"blackboard://content-resource/{result['id']}/"
            elif result["type"] == "Folder":
                result["parent_id"] = folder_id
                result["contents"] = self.request.route_url(
                    "blackboard_api.courses.files.list_folder",
                    course_id=course_id,
                    folder_id=result["id"],
                )

        return results
```

#### Split out local view helpers

Instead of a service just split out a couple of local helpers for the `list_files()` view that can be unit-tested separately:

```python
    @view_config(
        request_method="GET", route_name="blackboard_api.courses.files.list_folder"
    )
    def list_files(self):

        folder_id = self.request.matchdict.get("folder_id")

        results = FilesGetter(self.blackboard_api_client).get(course_id, folder_id)

        return FilesPresenter(self.request.route_url).present(results)
```

### DIFF
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -18,19 +18,11 @@ DOCUMENT_URL_REGEX = re.compile(
     r"blackboard:\/\/content-resource\/(?P<file_id>[^\/]*)\/"
 )
 
-# The maxiumum number of paginated requests we'll make before returning.
-PAGINATION_MAX_REQUESTS = 25
-
-# The maximum number of results to request per paginated response.
-# 200 is the highest number that Blackboard will accept here.
-PAGINATION_LIMIT = 200
-
-
 @view_defaults(permission=Permissions.API, renderer="json")
 class BlackboardFilesAPIViews:
     def __init__(self, request):
         self.request = request
-        self.blackboard_api_client = request.find_service(name="blackboard_api_client")
+        self.blackboard = request.find_service(name="blackboard")
 
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
     @view_config(
@@ -38,28 +30,24 @@ class BlackboardFilesAPIViews:
     )
     def list_files(self):
         """Return the list of files in the given course."""
+
         course_id = self.request.matchdict["course_id"]
         folder_id = self.request.matchdict.get("folder_id")
 
-        files = []
-        path = f"courses/uuid:{course_id}/resources"
+        results = self.blackboard.get_files(course_id, folder_id)
 
-        if folder_id:
-            # Get the files and folders in the given folder instead of the
-            # course's top-level files and folders.
-            path += f"/{folder_id}/children"
+        for result in results:
+            if result["type"] == "File":
+                result["id"] = f"blackboard://content-resource/{result['id']}/"
+            elif result["type"] == "Folder":
+                result["parent_id"] = folder_id
+                result["contents"] = self.request.route_url(
+                    "blackboard_api.courses.files.list_folder",
+                    course_id=course_id,
+                    folder_id=result["id"],
+                )
 
-        path += f"?limit={PAGINATION_LIMIT}"
-
-        for _ in range(PAGINATION_MAX_REQUESTS):
-            import pdb; pdb.set_trace()
-            response = self.blackboard_api_client.request("GET", path)
-            files.extend(BlackboardListFilesSchema(response).parse())
-            path = response.json().get("paging", {}).get("nextPage")
-            if not path:
-                break
-
-        return files
+        return results
 
     @view_config(request_method="GET", route_name="blackboard_api.files.via_url")
     def via_url(self):
@@ -71,7 +59,7 @@ class BlackboardFilesAPIViews:
         file_id = DOCUMENT_URL_REGEX.search(document_url)["file_id"]
 
         try:
-            response = self.blackboard_api_client.request(
+            response = self.blackboard.api.request(
                 "GET", f"courses/uuid:{course_id}/resources/{file_id}"
             )
         except HTTPError as err:
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -33,14 +33,26 @@ class BlackboardFilesAPIViews:
         self.blackboard_api_client = request.find_service(name="blackboard_api_client")
 
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
+    @view_config(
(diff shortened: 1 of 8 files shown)


tests/unit/lms/views/api/blackboard/_schemas_test.py
```diff
@@ -21,11 +21,19 @@ def test_valid(self, list_files_response):
...
+                "type": "File",
             },
             {
                 "id": "blackboard://content-resource/_7851_1/",
                 "updated_at": "1983-05-26T02:37:23.000z",
                 "display_name": "File_1.pdf",
+                "type": "File",
+            },
+            {
+                "id": "blackboard://content-resource/_7851_2/",
```
seanh, line 33: Known issue: I don't think we want to wrap _folder_ IDs in `blackboard://` URLs. Folder IDs aren't going to get stored in our DB so this isn't necessary, we will only have to unwrap them again.

lms/views/api/blackboard/files.py
```diff
@@ -38,7 +38,7 @@ def list_files(self):
         course_id = self.request.matchdict["course_id"]
 
         files = []
-        path = f"courses/uuid:{course_id}/resources?type=file&limit={PAGINATION_LIMIT}"
+        path = f"courses/uuid:{course_id}/resources?limit={PAGINATION_LIMIT}"
```
seanh, line 41: We can probably add a `?folders` query param to our API and, here in the view, toggle the `?type=file= param depending on that param. It would be a hack, but a temporary hack, we could remove it once the frontend changes to understand folders are merged.

marcospri (commented):
Will we need a new param to get the files (and folders) inside a given folder ID?

lms/views/api/blackboard/files.py
```diff
@@ -18,36 +18,36 @@
...
+        for result in results:
+            if result["type"] == "File":
+                result["id"] = f"blackboard://content-resource/{result['id']}/"
+            elif result["type"] == "Folder":
+                result["parent_id"] = folder_id
+                result["contents"] = self.request.route_url(
+                    "blackboard_api.courses.files.list_folder",
+                    course_id=course_id,
+                    folder_id=result["id"],
+                )
```
jon-betts, line 48: In all of the `js_config` stuff we return callbacks like this:

```python
"listFiles": {
                "authUrl": auth_url,
                "path": request.route_path(
                    "blackboard_api.courses.files.list", course_id=course_id
                ),
            },
```

I'm not sure if that's important in general, but it might make it easier if they are the same? Not sure. Might be worth asking the front-end guys because if we know we'll never want an `authURL` and they don't care it doesn't matter.

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/2950/

PR #2937, opened 2021-07-07
Add Microsoft OneDrive backend settings

jon-betts:
Fixes https://github.com/hypothesis/lms/issues/2941

This includes the config for Microsoft OneDrive when doing assignment configuration etc.

### Testing notes

 * Run `make devdata`
 * Go to: https://hypothesis.instructure.com/courses/125/assignments/1257
 * View source on our frame and search for "microsoftOneDrive"
 * You should find: `"microsoftOneDrive": {"enabled": false}`
 * Goto: http://localhost:8001/admin/
 * Search for: `Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a`
 * Check "Microsoft OneDrive enabled"
 * Refresh the view source page
 * You should find `"microsoftOneDrive": {"clientId": "<CLIENT_ID_HERE>", "enabled": true}`


lms/resources/_js_config/file_picker_config.py
```diff
@@ -71,6 +71,19 @@ def google_files_config(cls, context, request, application_instance):
             "origin": origin,
         }
 
+    @classmethod
+    def microsoft_onedrive(cls, _context, request, application_instance):
+        enabled = application_instance.settings.get(
+            "microsoft_onedrive", "files_enabled"
+        )
+        if not enabled:
+            return {"enabled": False}
```
jon-betts, line 80: I think this might be a good pattern to move to for all of these blobs of data. Although I'm not sure if it's safe as the front-end might now rely on all the other fields being present.

In this case the front-end doesn't exist yet, so we can get ahead of the curve.

seanh (approved):

PR #2946, opened 2021-07-08
Add a new high-level BlackboardAPIClient

seanh:
Later more code will be added to `BlackboardAPIClient` and `BasicBlackboardAPIClient` will become a private helper. For now this PR just gets the new high-level `BlackboardAPIClient` in place and gets a couple of simple views using it.


lms/services/__init__.py
```diff
@@ -1,5 +1,6 @@
 from lms.services.canvas import CanvasService
 from lms.services.exceptions import (
+    BlackboardFileNotFoundInCourse,
```
seanh, line 3: This exception moved from the views into the services

lms/services/__init__.py
```diff
@@ -23,6 +24,10 @@ def includeme(config):
     config.register_service_factory(
         "lms.services.basic_blackboard_api.factory", name="basic_blackboard_api_client"
     )
+    config.register_service_factory(
+        "lms.services.blackboard_api.blackboard_api_client_factory",
+        name="blackboard_api_client",
+    )
```
seanh, line 30: The new `"blackboard_api_client"` service. Notice that `"basic_blackboard_api_client"` still exists as a separate service for now

lms/services/blackboard_api/client.py
```diff
@@ -0,0 +1,48 @@
...
+        try:
+            response = self.api.request(
+                "GET", f"courses/uuid:{course_id}/resources/{file_id}"
+            )
+        except HTTPError as err:
+            if err.response.status_code == 404:
+                raise BlackboardFileNotFoundInCourse(file_id) from err
+            raise
+
+        return BlackboardPublicURLSchema(response).parse()
```
seanh, line 48: These methods don't do any mutation of the dicts returned by Blackboard. They just validate Blackboard's responses and return them more or less as-is. It's up to the views to convert those into the response format that our proxy API requires.

For now the _schemas_ are still doing some transformation of the dicts from Blackboard which means that the dicts have already been transformed by the time the service returns them. That'll be changed in a later PR

lms/services/exceptions.py
```diff
@@ -223,3 +223,15 @@ def __init__(self, response=None):
...
+class BlackboardFileNotFoundInCourse(Exception):
+    """A Blackboard file ID wasn't found in the current course."""
+
+    explanation = (
+        "Hypothesis couldn't find the assignment's PDF file in the Blackboard course."
+    )
+
+    def __init__(self, document_url):
+        self.details = {"document_url": document_url}
+        super().__init__(self.details)
```
seanh, line 237: Same exact exception that previously lived in the views

lms/views/api/blackboard/files.py
```diff
@@ -4,72 +4,38 @@
...
-            response = self.blackboard_api_client.request("GET", path)
-            files.extend(BlackboardListFilesSchema(response).parse())
-            path = response.json().get("paging", {}).get("nextPage")
-            if not path:
-                break
-
-        return files
+        return self.blackboard_api_client.list_files(
+            self.request.matchdict["course_id"]
+        )
```
seanh, line 28: There's no transformation for this view to do to the `list_files()` response because the schema is still doing it all. That'll move from the schema into this view later

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -0,0 +1,162 @@
...
+        basic_blackboard_api_client.request.return_value = factories.requests.Response(
+            json_data={"paging": {"nextPage": "NEXT_PAGE"}}
+        )
+
+        files = svc.list_files("COURSE_ID")
+
+        assert basic_blackboard_api_client.request.call_count == PAGINATION_MAX_REQUESTS
+        assert len(files) == PAGINATION_MAX_REQUESTS * len(
+            blackboard_list_files_schema.parse.return_value
+        )
```
seanh, line 98: These are the same exact tests that used to live in the view, now moved into the service

marcospri (approved):

PR #2947, opened 2021-07-08
Move "marshalling" code from schemas to views

seanh:
Make the `BlackboardAPIClient` service (and its schemas) just return Blackboard API dicts in the format given by the Blackboard API, without modifying them.

The service *does* remove unwanted fields that we don't care about from the dicts. In the future we'll actually add Blackboard API query params to request only the fields we want and save network bytes.

The service does also validate that the Blackboard dicts are as we expect.

But it doesn't mutate them any further. Instead it's the view's job to not about our proxy API's response format and transform the dicts into that format.

### DIFF
--- a/lms/services/blackboard_api/_schemas.py
+++ b/lms/services/blackboard_api/_schemas.py
@@ -8,46 +8,32 @@ from marshmallow import EXCLUDE, Schema, fields, post_load
 from lms.validation._base import RequestsResponseSchema
 
 
-class _FileSchema(Schema):
-    """Shared base schema for Blackboard file dicts."""
-
-    class Meta:
-        unknown = EXCLUDE
-
-    id = fields.Str(required=True)
-    display_name = fields.Str(data_key="name", required=True)
-    updated_at = fields.Str(data_key="modified", required=True)
-    mime_type = fields.Str(data_key="mimeType")
-
-    @post_load
-    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        data["id"] = f"blackboard://content-resource/{data['id']}/"
-        return data
-
-
 class BlackboardListFilesSchema(RequestsResponseSchema):
     """Schema for Blackboard API /courses/{courseId}/resources responses."""
 
-    results = fields.List(fields.Nested(_FileSchema), required=True)
+    class FileSchema(Schema):
+        """Schema for individual Blackboard file dicts."""
+
+        class Meta:
+            unknown = EXCLUDE
+
+        id = fields.Str(required=True)
+        name = fields.Str(required=True)
+        modified = fields.Str(required=True)
+        mimeType = fields.Str()
+
+    results = fields.List(fields.Nested(FileSchema), required=True)
 
     @post_load
     def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        pdf_files = []
-
-        for file in data["results"]:
-            if file.get("mime_type") == "application/pdf":
-                # Delete mime_type: we don't want to send it to the frontend.
-                del file["mime_type"]
-                pdf_files.append(file)
-
-        return pdf_files
+        return data["results"]
 
 
-class BlackboardPublicURLSchema(RequestsResponseSchema, _FileSchema):
+class BlackboardPublicURLSchema(RequestsResponseSchema):
     """Schema for Blackboard /courses/{courseId}/resources/{resourceId} responses."""
 
     downloadUrl = fields.Str(required=True)
 
     @post_load
-    def post_load(self, data, **_kwargs):
+    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
         return data["downloadUrl"]
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -23,9 +23,25 @@ class BlackboardFilesAPIViews:
     def list_files(self):
         """Return the list of files in the given course."""
 
-        return self.blackboard_api_client.list_files(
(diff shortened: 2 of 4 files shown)


lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
 
+        class Meta:
+            unknown = EXCLUDE
 
-class BlackboardListFilesSchema(RequestsResponseSchema):
-    """Schema for Blackboard API /courses/{courseId}/resources responses."""
+        id = fields.Str(required=True)
+        name = fields.Str(required=True)
+        modified = fields.Str(required=True)
+        mimeType = fields.Str()
```
seanh, line 23: The diff is a bit confusing because I also inlined `_FileSchema` into `BlackboardListFilesSchema` at the same time. The `FileSchema` class is actually exactly the same as before but without the `data_key`'s and the `@post_load`

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
-class _FileSchema(Schema):
-    """Shared base schema for Blackboard file dicts."""
-
-    class Meta:
-        unknown = EXCLUDE
-
-    id = fields.Str(required=True)
-    display_name = fields.Str(data_key="name", required=True)
-    updated_at = fields.Str(data_key="modified", required=True)
-    mime_type = fields.Str(data_key="mimeType")
```
seanh, line 20: The `data_key`'s here were removed which means that the schema no longer renames the fields from Blackboard

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
 
-    @post_load
-    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        data["id"] = f"blackboard://content-resource/{data['id']}/"
-        return data
+    class FileSchema(Schema):
+        """Schema for individual Blackboard file dicts."""
 
+        class Meta:
+            unknown = EXCLUDE
```
seanh, line 18: The `EXCLUDE` is still here meaning the schema still removes unknown fields from the Blackboard dicts. In the future I think the service will add a query param to the Blackboard API that filters what fields we want to get back to save bytes, at which point this `EXCLUDE` won't be doing anything and could be removed

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
-    display_name = fields.Str(data_key="name", required=True)
-    updated_at = fields.Str(data_key="modified", required=True)
-    mime_type = fields.Str(data_key="mimeType")
+class BlackboardListFilesSchema(RequestsResponseSchema):
+    """Schema for Blackboard API /courses/{courseId}/resources responses."""
 
-    @post_load
-    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        data["id"] = f"blackboard://content-resource/{data['id']}/"
-        return data
```
seanh, line 25: `post_load()` is gone so the schema is no longer wrapping IDs in `blackboard://` URLs. The view does that now

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
     def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
-        pdf_files = []
-
-        for file in data["results"]:
-            if file.get("mime_type") == "application/pdf":
-                # Delete mime_type: we don't want to send it to the frontend.
-                del file["mime_type"]
-                pdf_files.append(file)
-
-        return pdf_files
```
seanh, line 43: This `post_load()` is gone too so the schema no longer removes the `mimeType` field and no longer filters out non-PDF files, the view does that now

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
+        return data["results"]
```
seanh, line 29: It turned out to make everything simpler if the schema _does_ still pop out the results key

lms/services/blackboard_api/_schemas.py
```diff
@@ -8,46 +8,35 @@
...
     """Schema for Blackboard /courses/{courseId}/resources/{resourceId} responses."""
 
+    class Meta:
+        unknown = EXCLUDE
+
     downloadUrl = fields.Str(required=True)
 
     @post_load
-    def post_load(self, data, **_kwargs):
+    def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
```
seanh, line 41: Again it turned out to be simpler to keep a `@post_load` that just pops out the `downloadUrl`

lms/views/api/blackboard/files.py
```diff
@@ -23,10 +23,26 @@ def __init__(self, request):
...
+            if file.get("mimeType") != "application/pdf":
+                continue
+
+            pdf_files.append(
+                {
+                    "id": f"blackboard://content-resource/{file['id']}/",
+                    "display_name": file["name"],
+                    "updated_at": file["modified"],
+                }
+            )
```
seanh, line 42: This loop the view doing what the schema uses to

lms/services/blackboard_api/_schemas.py
seanh: I don't know why but PyLint suddenly decided I needed `no-self-use` here

lms/views/api/blackboard/files.py
marcospri: :+1:  I think I like this approach better

marcospri (approved):

PR #2948, opened 2021-07-08
Move BasicBlackboardAPIClient into BlackboardAPIClient

seanh:
Rename the `BasicBlackboardAPIClient` class to just `BasicClient` and move it into `BlackboardAPIClient`. It's now accessible as `BlackboardAPIClient.api` and is no longer a top-level service.

Note that `BasicClient` is still public and is still used (via `BlackboardAPIClient.api`) by some external code. In a future PR it'll
become completely private.

### DIFF
--- a/lms/services/blackboard_api/basic.py
+++ b/lms/services/blackboard_api/basic.py
@@ -43,7 +43,7 @@ class BlackboardErrorResponseSchema(RequestsResponseSchema):
             return {}
 
 
-class BasicBlackboardAPIClient:
+class BasicClient:
     """A low-level Blackboard API client."""
 
     def __init__(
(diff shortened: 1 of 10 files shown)


lms/services/blackboard_api/basic.py
```diff
@@ -84,10 +84,10 @@ def get_token(self, authorization_code):
         validated_data = OAuthTokenResponseSchema(response).parse()
 
         # Save the access token to the DB.
-        # pylint: disable=no-member
         self._oauth2_token_service.save(
             validated_data["access_token"],
             validated_data.get("refresh_token"),
+            # pylint:disable=no-member
```
seanh, line 90: PyLint forced me to move this :man_shrugging:

marcospri (approved):
Not much to comment, looks good :+1:

PR #2949, opened 2021-07-08
Make BasicClient private

seanh:
Make `BasicClient` a private helper of `BlackboardAPIClient`. It's no longer directly accessible to external code via
`BlackboardAPIClient.api`. Instead `BlackboardAPIClient` exposes its own public methods that delegate privately to `BasicClient`.

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -64,12 +64,6 @@ class BasicClient:
         self._oauth2_token_service = oauth2_token_service
 
     def get_token(self, authorization_code):
-        """
-        Get an access token from Blackboard and save it in the DB.
-
-        :raise services.HTTPError: if something goes wrong with the access
-            token request to Blackboard
-        """
         # Send a request to Blackboard to get an access token.
         response = self._http_service.post(
             self._api_url("oauth2/token"),
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -15,10 +15,17 @@ PAGINATION_LIMIT = 200
 class BlackboardAPIClient:
     """A high-level Blackboard API client."""
 
-    api = None
-
     def __init__(self, basic_client):
-        self.api = basic_client
+        self._api = basic_client
+
+    def get_token(self, authorization_code):
+        """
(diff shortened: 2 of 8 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -64,12 +64,6 @@ def __init__(
         self._oauth2_token_service = oauth2_token_service
 
     def get_token(self, authorization_code):
-        """
-        Get an access token from Blackboard and save it in the DB.
-
-        :raise services.HTTPError: if something goes wrong with the access
-            token request to Blackboard
-        """
```
seanh, line 72: Moved the docstring onto the higher-level public method and didn't want it duplicated here

lms/services/blackboard_api/client.py
```diff
@@ -15,10 +15,17 @@
...
+        self._api = basic_client
+
+    def get_token(self, authorization_code):
+        """
+        Save a new Blackboard access token for the current user to the DB.
+
+        :raise services.HTTPError: if something goes wrong with the access
+            token request to Blackboard
+        """
+        self._api.get_token(authorization_code)
```
seanh, line 28: So view code can just do `blackboard_api_client.get_token()` instead of `blackboard_api_client.api.get_token()`. This also makes mocking simpler as the mock `BlackboardAPIClient` class doesn't need to have an `api` property that's an autospec of `BasicClient`.

Note that I _do_ think we want the actual `get_token()` implementation to live in `BasicClient`. I think that's going to be needed by future PRs that add refresh token support.

tests/unit/services.py
```diff
@@ -97,13 +96,7 @@ def assignment_service(mock_service):
...
 @pytest.fixture
 def blackboard_api_client(mock_service):
-    blackboard_api_client = mock_service(
-        BlackboardAPIClient, service_name="blackboard_api_client"
-    )
-    blackboard_api_client.api = mock.create_autospec(
-        BasicClient, instance=True, spec_set=True
-    )
-    return blackboard_api_client
+    return mock_service(BlackboardAPIClient, service_name="blackboard_api_client")
```
seanh, line 99: Simpler mocking here

marcospri (approved):

PR #2950, opened 2021-07-08
Add folder support to Blackboard Files backend

seanh:
This adds folder support to the backend APIs for Blackboard Files.

Note that without the corresponding frontend changes to understand folders this doesn't work: it breaks the UI. So frontend changes will be needed before merging this.

This commit changes the existing `/api/blackboard/courses/{course_id}/files` API to return a mixed list of files *and* folders.
Files look pretty much like they did before except:

* They now have a new `type: "File"` field
* They now have a new `parent_id` field that is either the `id` of the parent folder or `null` for top-level files

Folders look the same as files except:

* The `id` fields of folders are not wrapped in `blackboard://` URLs like they are for files
* Folders have `type: "Folder"`
* Folders have a `contents` field which is a dict containing the `path` for an API call to get the contents of the folder, plus also the `authUrl` to use to re-authorize in case the request fails with an authorization error

The contents URLs in folders refer to a new `/api/blackboard/courses/{course_id}/folders/{folder_id}/files` API that's also added by this PR that returns the contents of a given folder rather than the course's top-level contents.

Example response, top-level contents:

```json
[
  {
    "display_name": "Dinosaurs Folder",
    "updated_at": "2021-07-08T13:25:59.000Z",
    "type": "Folder",
    "parent_id": null,
    "id": "_9907_1",
    "contents": {
        "authUrl": "...",
        "path": "/api/blackboard/courses/7688d64e8c2d433597d80e9d18a8d373/folders/_9907_1/files"
    }
  },
  {
    "display_name": "Manovich.pdf",
    "updated_at": "2021-06-02T15:39:51.000Z",
    "type": "File",
    "parent_id": null,
    "id": "blackboard://content-resource/_7851_1/"
  }
]
```

Example response, folder `_9907_1`'s contents:

```json
[
  {
    "display_name": "Podokesaurus.pdf",
    "updated_at": "2021-06-16T10:31:50.000Z",
    "type": "File",
    "parent_id": "_9907_1",
    "id": "blackboard://content-resource/_8615_1/"
  },
  {
    "display_name": "Subfolder",
    "updated_at": "2021-07-08T13:26:48.000Z",
    "type": "Folder",
    "parent_id": "_9907_1",
    "id": "_9908_1",
    "contents": {
      "authUrl": "...",
      "path": "/api/blackboard/courses/7688d64e8c2d433597d80e9d18a8d373/folders/_9908_1/files"
    }
  }
]
```

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -43,6 +43,10 @@ def includeme(config):
         "blackboard_api.courses.files.list", "/api/blackboard/courses/{course_id}/files"
     )
     config.add_route(
+        "blackboard_api.courses.folders.files.list",
+        "/api/blackboard/courses/{course_id}/folders/{folder_id}/files",
+    )
+    config.add_route(
         "blackboard_api.files.via_url", "/api/blackboard/courses/{course_id}/via_url"
     )
 
--- a/lms/services/blackboard_api/_schemas.py
+++ b/lms/services/blackboard_api/_schemas.py
@@ -20,6 +20,7 @@ class BlackboardListFilesSchema(RequestsResponseSchema):
         id = fields.Str(required=True)
         name = fields.Str(required=True)
         modified = fields.Str(required=True)
+        type = fields.Str(required=True)
         mimeType = fields.Str()
 
     results = fields.List(fields.Nested(FileSchema), required=True)
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -29,26 +29,37 @@ class BlackboardAPIClient:
         """
         self._api.get_token(authorization_code)
 
-    def list_files(self, course_id):
-        """Return the list of files in the given course."""
+    def list_files(self, course_id, folder_id=None):
+        """Return the list of files in the given course or folder."""
 
-        files = []
-        path = f"courses/uuid:{course_id}/resources?" + urlencode(
-            {
-                "type": "file",
-                "limit": PAGINATION_LIMIT,
-                "fields": "id,name,modified,mimeType",
-            }
+        path = f"courses/uuid:{course_id}/resources"
+
+        if folder_id:
+            # Get the files and folders in the given folder instead of the
+            # course's top-level files and folders.
+            path += f"/{folder_id}/children"
+
+        path = (
+            path
+            + "?"
+            + urlencode(
+                {
+                    "limit": PAGINATION_LIMIT,
+                    "fields": "id,name,type,modified,mimeType",
+                }
+            )
         )
+
+        results = []
 
         for _ in range(PAGINATION_MAX_REQUESTS):
             response = self._api.request("GET", path)
-            files.extend(BlackboardListFilesSchema(response).parse())
+            results.extend(BlackboardListFilesSchema(response).parse())
             path = response.json().get("paging", {}).get("nextPage")
             if not path:
                 break
 
-        return files
+        return results
 
     def public_url(self, course_id, file_id):
         """Return a public URL for the given file."""
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -20,28 +20,45 @@ class BlackboardFilesAPIViews:
         self.blackboard_api_client = request.find_service(name="blackboard_api_client")
 
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
+    @view_config(
+        request_method="GET", route_name="blackboard_api.courses.folders.files.list"
+    )
     def list_files(self):
-        """Return the list of files in the given course."""
+        """Return the list of files in the given course or folder."""
 
-        files = self.blackboard_api_client.list_files(
-            self.request.matchdict["course_id"]
-        )
+        course_id = self.request.matchdict["course_id"]
+        folder_id = self.request.matchdict.get("folder_id")
 
-        pdf_files = []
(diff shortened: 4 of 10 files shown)


lms/services/blackboard_api/_schemas.py
```diff
@@ -20,6 +20,7 @@ class Meta:
         id = fields.Str(required=True)
         name = fields.Str(required=True)
         modified = fields.Str(required=True)
+        type = fields.Str(required=True)
```
seanh, line 23: The view needs the `type` field from the Blackboard API so that it can tell files from folders. This field gets sent through to the frontend too.

lms/views/api/blackboard/files.py
```diff
@@ -20,28 +20,40 @@ def __init__(self, request):
...
+            if result["type"] == "File" and result.get("mimeType") == "application/pdf":
+                response_result["id"] = f"blackboard://content-resource/{result['id']}/"
+                response_results.append(response_result)
+            elif result["type"] == "Folder":
+                response_result["id"] = result["id"]
+                response_result["contents"] = self.request.route_url(
+                    "blackboard_api.courses.files.list_folder",
+                    course_id=course_id,
+                    folder_id=result["id"],
+                )
```
seanh, line 53: This should probably be a dict with `authUrl` and `path` as suggested earlier

lms/views/api/blackboard/files.py
```diff
@@ -20,28 +20,40 @@ def __init__(self, request):
...
+            if result["type"] == "File" and result.get("mimeType") == "application/pdf":
+                response_result["id"] = f"blackboard://content-resource/{result['id']}/"
+                response_results.append(response_result)
+            elif result["type"] == "Folder":
+                response_result["id"] = result["id"]
+                response_result["contents"] = self.request.route_path(
+                    "blackboard_api.courses.files.list_folder",
+                    course_id=course_id,
+                    folder_id=result["id"],
+                )
```
seanh, line 53: This should maybe be a dict with `authUrl` and `path` as suggested earlier

lms/routes.py
```diff
@@ -41,6 +41,10 @@ def includeme(config):
     config.add_route(
         "blackboard_api.courses.files.list", "/api/blackboard/courses/{course_id}/files"
     )
+    config.add_route(
+        "blackboard_api.courses.files.list_folder",
```
marcospri, line 45: "blackboard_api.courses.folder.files"? Not that important we only type this name once

lms/services/blackboard_api/client.py
```diff
@@ -27,20 +27,28 @@ def get_token(self, authorization_code):
         """
         self._api.get_token(authorization_code)
 
-    def list_files(self, course_id):
-        """Return the list of files in the given course."""
+    def list_files(self, course_id, folder_id=None):
```
marcospri, line 30: `list_files` I think is fine as a method name (list_resources doesn't mean much) but maybe include the fact that can return folders in the docstring.

lms/services/blackboard_api/client.py
```diff
@@ -27,20 +27,28 @@ def get_token(self, authorization_code):
...
+        if folder_id:
+            # Get the files and folders in the given folder instead of the
+            # course's top-level files and folders.
+            path += f"/{folder_id}/children"
+
+        path += f"?limit={PAGINATION_LIMIT}"
+
+        results = []
 
         for _ in range(PAGINATION_MAX_REQUESTS):
```
marcospri, line 44: Not necessarily for this PR but I think we should move pagination logic to BasicClient and take a param or similar on `_api.request`

lms/views/api/blackboard/files.py
marcospri: Yeah, probably a good idea for consistency

lms/views/api/blackboard/files.py
```diff
@@ -20,28 +20,40 @@ def __init__(self, request):
...
+
+        for result in results:
+            response_result = {
+                "display_name": result["name"],
+                "updated_at": result["modified"],
+                "type": result["type"],
+                "parent_id": folder_id,
+            }
+
+            if result["type"] == "File" and result.get("mimeType") == "application/pdf":
```
marcospri, line 44: Maybe `File` and `Folder` could be constants on `blackboard_api_client`

lms/services/blackboard_api/client.py
seanh: Yep. I think we'll wait until we actually have a second method that uses pagination, then we'll be able to see what sort of generic design will work

lms/views/api/blackboard/files.py
seanh: Ok, changed this to use the same dict format as used elsewhere: https://github.com/hypothesis/lms/pull/2950/commits/79f6617b842085c9c8ca045fc81d29840c21d78f

lms/routes.py
seanh: https://github.com/hypothesis/lms/pull/2950/commits/16d3d006991ee0a6eae3ee12210c3f656b9d089d

lms/services/blackboard_api/client.py
seanh: https://github.com/hypothesis/lms/pull/2950/commits/e55f288cef33d2da763325636c98be9dc9e7292c

marcospri (approved):

PR #2951, opened 2021-07-08
Remove an obsolete workaround for a Babel plugin conflict

robertknight:
We no longer use the `transform-async-to-promises` Babel plugin, so we don't need to workaround a conflict between it and the `mockable-imports` plugin.


robertknight:
The fact that the frontend build completed means that this worked, so I'm going to merge this PR.

Issue #2033, opened 2020-08-19
Add Blackboard refresh token support


seanh:
The `BlackboardAPIClient` service should support automatically using a refresh token to get a new access token when the user's access token has expired, like the `CanvasAPIClient` does.

PR #2968, opened 2021-07-14
Add Blackboard refresh token support

seanh:
Fixes: https://github.com/hypothesis/lms/issues/2033

Design:

* When an HTTP request fails we need to send a refresh request to get a new access token and then re-try the original request with the new access token, all without bothering the user. The design aim is to implement this "send with refresh and retry" logic in one place where it can be unit tested directly, and then be able to re-use it for each of the Blackboard API methods (list files, get public URL, etc).

* `HTTPService` is split into two:

  1. `HTTPService`, which only knows about HTTP not OAuth
  2. `OAuthHTTPService`. Built on top of `HTTPService` and adds generic OAuth-stuff (not specific to Blackboard)
  
  This change isn't really necessary but it just might be nice to split these into two smaller services

  The `oauth=True` argument that used to be in `HTTPService` has moved into `OAuthHTTPService`. And the new refresh token support is implemented in `OAuthHTTPService` not in `HTTPService`.
  
* Blackboard's `BasicClient` uses `OAuthHTTPService` instead of `HTTPService`

* Blackboard's `get_token()` method gains support for refresh tokens as well as authorization codes

* `OAuthHTTPService` contains the generic "send with refresh and retry" logic where, if a request fails for any reason, we try to use our refresh token to get a new access token and then re-try the original request with the new access token.

  This "send with refresh and retry" is generic (not Blackboard-specific) apart from the using-the-refresh-token part. Sending a request to the Blackboard API to exchange a refresh token for a new access token is Blackboard specific. It's implemented in `BlackboardAPIClient`'s `BasicClient.get_token()`. So `OAuthHTTPService.request()` accepts a `refresh` callable as argument.

  Once we have two or three things using `OAuthHTTPService` we might look to see whether refreshing tokens is actually generic and perhaps users could pass in a couple of parameters instead of a full callable. But for now I think it's probably best to hold off and just do the simplest and most universal thing and just accept a callable. (The same could be said about _getting_ access token in the first place: might be some opportunity for code reuse there one day.)

Known issues:

- [ ] Don't show an error message to the user if the refresh request fails. Just show them the authorization dialog. Refresh tokens expiring or being revoked is normal, not an error
  - [ ] We may want to vary this depending on the specific error message from Blackboard. If the error from Blackboard says "That refresh token is invalid" (or unknown, expired, etc) that's normal and we don't want to show an error to the user. If the refresh token request to Blackboard comes back with some other error then something unexpected _has_ happened and we _do_ want to show an error dialog to the user. As usual this depends on our ability to distinguish one type of third-party error response from another when they probably don't have proper error codes
- [ ] Do show an error message to the user if the second attempt at the API request with a new access token fails. This indicates that something is going wrong and it doesn't seem to be to do with the access token having expired or been revoked, since re-trying with a fresh access token also failed

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -53,6 +53,7 @@ class BasicClient:
         client_secret,
         redirect_uri,
         http_service,
+        oauth_http_service,
         oauth2_token_service,
     ):  # pylint:disable=too-many-arguments
         self.blackboard_host = blackboard_host
@@ -61,20 +62,29 @@ class BasicClient:
         self.redirect_uri = redirect_uri
 
         self._http_service = http_service
+        self._oauth_http_service = oauth_http_service
         self._oauth2_token_service = oauth2_token_service
 
-    def get_token(self, authorization_code):
+    def get_token(self, authorization_code=None, refresh_token=None):
+        data = {"redirect_uri": self.redirect_uri}
+
+        if authorization_code:
+            data["grant_type"] = "authorization_code"
+            data["code"] = authorization_code
+        elif refresh_token:
+            data["grant_type"] = "refresh_token"
+            data["refresh_token"] = refresh_token
+        else:
+            raise AssertionError(
+                "Either one of authorization_code or refresh_token must be given."
+            )
+
         # Send a request to Blackboard to get an access token.
         response = self._http_service.post(
             self._api_url("oauth2/token"),
-            data={
-                "grant_type": "authorization_code",
-                "redirect_uri": self.redirect_uri,
-                "code": authorization_code,
-            },
+            data=data,
             auth=(self.client_id, self.client_secret),
         )
-
         validated_data = OAuthTokenResponseSchema(response).parse()
 
         # Save the access token to the DB.
@@ -85,9 +95,13 @@ class BasicClient:
             validated_data.get("expires_in"),
         )
 
+        return validated_data["access_token"]
+
     def request(self, method, path):
         try:
-            return self._http_service.request(method, self._api_url(path), oauth=True)
+            return self._oauth_http_service.request(
+                method, self._api_url(path), refresh=self.get_token
+            )
         except HTTPError as err:
             error_dict = BlackboardErrorResponseSchema(err.response).parse()
 
--- a/lms/services/oauth_http.py
+++ b/lms/services/oauth_http.py
@@ -1,0 +1,52 @@
+from lms.services import HTTPError
+
+
+class OAuthHTTPService:
+    """Send OAuth 2.0 requests and return the responses."""
+
+    def __init__(self, http_service, oauth2_token_service):
+        self._http_service = http_service
+        self._oauth2_token_service = oauth2_token_service
+
+    def get(self, *args, **kwargs):
+        return self.request("GET", *args, **kwargs)
+
+    def put(self, *args, **kwargs):
+        return self.request("PUT", *args, **kwargs)
+
+    def post(self, *args, **kwargs):
+        return self.request("POST", *args, **kwargs)
+
+    def patch(self, *args, **kwargs):
+        return self.request("PATCH", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        return self.request("DELETE", *args, **kwargs)
+
+    def request(self, method, url, refresh=None, **kwargs):
+        oauth2_token = self._oauth2_token_service.get()
+        access_token = oauth2_token.access_token
+        refresh_token = oauth2_token.refresh_token
+
+        kwargs.setdefault("headers", {})
+        assert "Authorization" not in kwargs["headers"]
+
+        def send_request():
+            kwargs["headers"]["Authorization"] = f"Bearer {access_token}"
+            return self._http_service.request(method, url, **kwargs)
+
+        try:
+            return send_request()
+        except HTTPError:
+            if not refresh_token or not refresh:
+                raise
+
+        access_token = refresh(refresh_token=refresh_token)
+
+        return send_request()
+
+
+def factory(_context, request):
+    return OAuthHTTPService(
+        request.find_service(name="http"), request.find_service(name="oauth2_token")
+    )
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -22,6 +22,9 @@ from lms.services.exceptions import (
 def includeme(config):
     config.register_service_factory("lms.services.http.factory", name="http")
     config.register_service_factory(
+        "lms.services.oauth_http.factory", name="oauth_http"
+    )
+    config.register_service_factory(
         "lms.services.blackboard_api.blackboard_api_client_factory",
         name="blackboard_api_client",
     )
--- a/lms/services/blackboard_api/factory.py
+++ b/lms/services/blackboard_api/factory.py
@@ -13,6 +13,7 @@ def blackboard_api_client_factory(_context, request):
             client_secret=settings["blackboard_api_client_secret"],
             redirect_uri=request.route_url("blackboard_api.oauth.callback"),
             http_service=request.find_service(name="http"),
+            oauth_http_service=request.find_service(name="oauth_http"),
             oauth2_token_service=request.find_service(name="oauth2_token"),
         )
     )
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -7,9 +7,7 @@ from lms.services.exceptions import HTTPError
 class HTTPService:
     """Send HTTP requests with `requests` and receive the responses."""
 
-    def __init__(self, oauth2_token_service, _session=None):
-        self._oauth2_token_service = oauth2_token_service
-
+    def __init__(self, _session=None):
         # A requests session is used so that cookies are persisted across
         # requests and urllib3 connection pooling is used (which means that
         # underlying TCP connections are re-used when making multiple requests
@@ -33,7 +31,7 @@ class HTTPService:
     def delete(self, *args, **kwargs):
         return self.request("DELETE", *args, **kwargs)
 
-    def request(self, method, url, timeout=(10, 10), oauth=False, **kwargs):
+    def request(self, method, url, timeout=(10, 10), **kwargs):
         """
         Send a request with `requests` and return the requests.Response object.
 
@@ -56,12 +54,6 @@ class HTTPService:
             response download. It's a time limit on how long to wait *between
             bytes from the server*. The entire download can take much longer.
 
-        :param oauth: Include an OAuth 2 access token in the request.
-            If oauth=True the current user's access token will be looked up in
-            the database and included in an Authorization header in the
-            request.
-        :type oauth: bool
-
         :param kwargs: Any other keyword arguments will be passed directly to
             requests.Session().request():
             https://docs.python-requests.org/en/latest/api/#requests.Session.request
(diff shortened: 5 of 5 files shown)


lms/services/oauth_http.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def post(self, *args, **kwargs):
+        return self.request("POST", *args, **kwargs)
+
+    def patch(self, *args, **kwargs):
+        return self.request("PATCH", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        return self.request("DELETE", *args, **kwargs)
+
```
jon-betts, line 25:
```suggestion
class OAuthHTTPService(HTTPService):
    """Send OAuth 2.0 requests and return the responses."""

    def __init__(self,  oauth2_token_service):
        super().__init__()
        self._oauth2_token_service = oauth2_token_service
```

In the same way that I think the `HTTPService` is a `Session` rather than having one, I think the `OAuthHTTPService` is an `HTTPService`. We are repeatedly implementing a lot of stub methods with no functionality in them which replicate those in `HTTPService` which replicate those in `Session`. I think this (along with the name) is a strong hint for inheritance instead of composition.

lms/services/oauth_http.py
```diff
@@ -0,0 +1,52 @@
...
+        oauth2_token = self._oauth2_token_service.get()
+        access_token = oauth2_token.access_token
+        refresh_token = oauth2_token.refresh_token
+
+        kwargs.setdefault("headers", {})
+        assert "Authorization" not in kwargs["headers"]
+
+        def send_request():
+            kwargs["headers"]["Authorization"] = f"Bearer {access_token}"
+            return self._http_service.request(method, url, **kwargs)
```
jon-betts, line 36:
```suggestion
            return super().request(method, url, **kwargs)
```

lms/services/oauth_http.py
```diff
@@ -0,0 +1,52 @@
...
+
+        try:
+            return send_request()
+        except HTTPError:
+            if not refresh_token or not refresh:
+                raise
+
+        access_token = refresh(refresh_token=refresh_token)
```
jon-betts, line 44: I think the closure here might be the bad kind of clever. It works, but it's not clear at this point in the code why something is happening. I think it might be better to inline this instead?

lms/services/blackboard_api/_basic.py
```diff
@@ -61,20 +62,29 @@ def __init__(
         self.redirect_uri = redirect_uri
 
         self._http_service = http_service
+        self._oauth_http_service = oauth_http_service
         self._oauth2_token_service = oauth2_token_service
 
-    def get_token(self, authorization_code):
+    def get_token(self, authorization_code=None, refresh_token=None):
```
jon-betts, line 68: I would have expected to see this sort of functionality (or at least quite a bit of it) inside the `OAuthHTTPService`. I think we're on the right track when `oauth2_token_service` is not required in this class. That would show that `OAuthHTTPService` has a nice boundary and is doing an encapsulated task.

Is there a way we can supply the URL or any variable parts of this to a more general `get_token()` function in `OAuthHTTPService`?

jon-betts (commented):
I think there are some changes I'd suggest here:

I think `Session` -> `HTTPService` -> `OAuthHTTPService` is a fairly clear cut case of inheritance over composition. There are 11 methods which basically just proxy up to the parent in the current design which could be deleted. I think this is about a strong a signal for inheritance as you're going to get. The name as well strongly hints what's happening here. For example these could easily be `Session` -> `HTTPSession` -> `OAuthHTTPSession`.

I'm not sure `OAuthHTTPService` is doing all it should. I'd really like to see the non-Blackboard parts `get_token()` coming from the service too. This should be a standardised thing for the most part, and I think you'd only need to pass in the base URL to make this work. I think it's a good sign that `OAuthHTTPService` has a well defined boundary when we don't need the token service outside of it.

lms/services/oauth_http.py
seanh: > I think Session -> HTTPService -> OAuthHTTPService is a fairly clear cut case of inheritance over composition. There are 11 methods which basically just proxy up to the parent in the current design which could be deleted. I think this is about a strong a signal for inheritance as you're going to get. The name as well strongly hints what's happening here. For example these could easily be Session -> HTTPSession -> OAuthHTTPSession.

I think I'd like to stick with the composition-based design here. The intention with `HTTPService` (and now also `OAuthHTTPService`) is to provide a **simpler** interface than the one that `requests` provides, as well as to embody best practices for using `requests` correctly (using a session, handling errors). So I _don't_ think that `HTTPService` "is a" `requests.Session`. It's a service that wraps `requests` and provides a simpler interface.

I don't actually want `HTTPService` to expose all of `requests.Session`'s attributes (`close()`, `mount()`, etc) as its own public interface.

This does have the downside of creating a certain amount of boilerplate because `HTTPService` needs `get()`, `put()`, `post()`, `patch()` and `delete()` methods that just forward to its `request()` method, and `OAuthHTTPService` needs the same (as does the Blackboard's `BasicClient`). (They actually don't really need these. You could just call `request("METHOD", ...)`. But they're nice to have.) If the duplicated boiler is the problem I think we could use a mixin class to reduce that without losing the benefits of composition / taking on the downsides of inheritance:

```python
class HTTPMethodMixin:
    """Add HTTP verb methods (get(), post(), etc) that forward to self.request()."""

    def get(self, *args, **kwargs):
        return self.request("GET", *args, **kwargs)

    def put(self, *args, **kwargs):
        return self.request("PUT", *args, **kwargs)

    def post(self, *args, **kwargs):
        return self.request("POST", *args, **kwargs)

    def patch(self, *args, **kwargs):
        return self.request("PATCH", *args, **kwargs)

    def delete(self, *args, **kwargs):
        return self.request("DELETE", *args, **kwargs)
```

Anyway, this is why I prefer composition over inheritance:

### I don't think the duplicated forwarding methods are a big problem

I think duplicated boilerplate code is a minor sin (and possibly amenable to the mixin-based solution I suggested above) but I don't think it's a big problem.

On the other hand I **do** think that the sins of inheritance are a big problem: tightly coupling `HTTPService` to `requests.Session` (and `OAuthHTTPService` and `BasicClient` to `HTTPService`); getting in the way of isolated unit testing (it will no longer be easy to test `OAuthHTTPService` and `BasicClient` in isolation with a mock `HTTPService`); harming code readability; loss of flexibility. I think [this post](https://hynek.me/articles/python-subclassing-redux/) covers it well (particularly the **Type 1: Code Sharing** section).

### I don't expect the amount of boiler to grow

`HTTPService` originally just had `request()` and Marcos requested adding `get()` and pals for convenience. I don't expect to want to add many other `requests.Session` attributes to `HTTPService`'s public interface so I don't expect the amount of boiler here to grow much. (At some point if you have to write a really large number of forwarding methods this can tip the balance towards inheritance but I don't think we're hitting that point here.)

### This is just Python's way of doing composition

The boiler is just sort of par-for-the-course with composition in Python. Other languages have more concise ways to declare when a class forwards methods to a composed object. Ruby has this for example. There's even a third-party Python lib called [forwardable](https://github.com/5long/forwardable) that implements something like this in Python:

```python
@forwardable()
class HTTPService:
    def_delegators("_session", "get, put, post, patch, delete")

    def __init__(self, ..._):
        self._session = requests.Session()
        ...

    def request(self, ...):
        ...
```

I don't think we should use forwardable. The point I'm trying to make is that boilerplatey forwarding methods are just how you do it in Python so we shouldn't consider them too offensive.

### They're not actually forwarding methods

The `get()`, `put()`, `post()`, `patch()` and `delete()` **don't** actually forward to `requests.Session.get()`, `requests.Session.put()`, `requests.Session.post()`, `requests.Session.patch()` and `requests.Session.delete()`. They forward to `HTTPService.request()` which wraps `requests.Session.request()` and adds custom behaviour. 

It may be that in the `requests` source code `requests.Session.get()` etc do just forward to `self.request()` in which case sub-classing `requests.Session` and just overriding `request()` would work. But this would mean that `HTTPService` now implicitly depends on private implementation details of `requests.Session` and will break if `requests.Session` changes its internal implementation details.

Some classes are explicitly designed for subclassing and provide methods that you're supposed to override and guarantees about how and when those methods will be called. AFAIK `requests.Session.request()` isn't intended to be used in this way, it might just happen to work by accident.

seanh: > I'm not sure OAuthHTTPService is doing all it should. I'd really like to see the non-Blackboard parts get_token() coming from the service too. This should be a standardised thing for the most part, and I think you'd only need to pass in the base URL to make this work. I think it's a good sign that OAuthHTTPService has a well defined boundary when we don't need the token service outside of it.

I think you're very likely right about this one but I'd like to hold off until we have more than one user of `OAuthHTTPService`. Then we'll be able to see what the different implementations of `get_token()` have in common and where they differ. I thought I'd said this somewhere, but I'm just holding off on this one until we have more cases.

lms/services/oauth_http.py
seanh: Yeah, we could inline it. It'd mean a little bit of duplication but only two lines duplicated twice. I don't have a problem with inner functions myself: I think they're perfectly normal and not a clever thing.

lms/services/blackboard_api/_basic.py
seanh: I don't think we should try to answer that question until we have two or three users of `OAuthHTTPService`. At that point it should be clear how to implement a generic `get_token()`

seanh:
Gonna close this and get on with splitting it up into separate PRs

lms/services/oauth_http.py
seanh: Added a PR to split out `OAuthHTTPService`, sticking with the composition-based design: https://github.com/hypothesis/lms/pull/2985

seanh: > > I'm not sure OAuthHTTPService is doing all it should. I'd really like to see the non-Blackboard parts get_token() coming from the service too. This should be a standardised thing for the most part, and I think you'd only need to pass in the base URL to make this work. I think it's a good sign that OAuthHTTPService has a well defined boundary when we don't need the token service outside of it.
> 
> I think you're very likely right about this one but I'd like to hold off until we have more than one user of `OAuthHTTPService`. Then we'll be able to see what the different implementations of `get_token()` have in common and where they differ. I thought I'd said this somewhere, but I'm just holding off on this one until we have more cases.

I ended up writing generic `access_token_request()` and `refresh_token_request()` methods anyway even though they're only used for the Blackboard API so far: https://github.com/hypothesis/lms/pull/2987, https://github.com/hypothesis/lms/pull/2988. These won't necessarily work for other servers besides Blackboard yet but we can generalize them further as we need to when we come to it.

I ended up adding these to `OAuth2TokenService` not `OAuthHTTPService` because `OAuthHTTPService.request()` is going to contain higher-level "send with refresh and retry" logic that I want to unit test with a mock `refresh_token_request()` method.

lms/services/blackboard_api/_basic.py
seanh: I did end up moving `access_token_request()` and `refresh_token_request()` into a generic place in `OAuthHTTPService`: https://github.com/hypothesis/lms/pull/2987 and https://github.com/hypothesis/lms/pull/2988

lms/services/oauth_http.py
seanh: The "final" version of this method ended up looking quite different from this draft version anyway, see https://github.com/hypothesis/lms/pull/2993. I might not be done tweaking it though, so this suggestion may become relevant again

PR #2969, opened 2021-07-14
Remove filter_oauth2_token_error()

seanh:
This doesn't seem to be necessary anymore. These are caught by an exception view which seems to mean that they don't get reported to Sentry.


jon-betts (approved):
Holy moly those two lines of code spawned a lot of test!

Issue #2963, opened 2021-07-12
Only request necessary fields from the Blackboard API


seanh:
The Blackboard API has a `fields` paramter (see [their API docs](https://developer.blackboard.com/portal/displayApi)) that we can use to request only the fields that we need. This should save us some network bandwidth and time.

PR #2970, opened 2021-07-14
Only get needed fields from Blackboard API

seanh:
Save some time and bandwidth by only getting the fields that we actually need from the Blackboard API.

Fixes https://github.com/hypothesis/lms/issues/2963

### DIFF
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlencode
+
 from lms.services.blackboard_api._schemas import (
     BlackboardListFilesSchema,
     BlackboardPublicURLSchema,
@@ -31,7 +33,13 @@ class BlackboardAPIClient:
         """Return the list of files in the given course."""
 
         files = []
-        path = f"courses/uuid:{course_id}/resources?type=file&limit={PAGINATION_LIMIT}"
+        path = f"courses/uuid:{course_id}/resources?" + urlencode(
+            {
+                "type": "file",
+                "limit": PAGINATION_LIMIT,
+                "fields": "id,name,modified,mimeType",
+            }
+        )
 
         for _ in range(PAGINATION_MAX_REQUESTS):
             response = self._api.request("GET", path)
@@ -47,7 +55,8 @@ class BlackboardAPIClient:
 
         try:
             response = self._api.request(
-                "GET", f"courses/uuid:{course_id}/resources/{file_id}"
+                "GET",
+                f"courses/uuid:{course_id}/resources/{file_id}?fields=downloadUrl",
             )
         except HTTPError as err:
             if err.response.status_code == 404:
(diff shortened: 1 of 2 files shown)


lms/services/blackboard_api/client.py
```diff
@@ -31,7 +31,10 @@ def list_files(self, course_id):
         """Return the list of files in the given course."""
 
         files = []
-        path = f"courses/uuid:{course_id}/resources?type=file&limit={PAGINATION_LIMIT}"
+        path = f"courses/uuid:{course_id}/resources"
+        path += "?type=file"
+        path += f"&limit={PAGINATION_LIMIT}"
+        path += "&fields=id,name,modified,mimeType"
```
jon-betts, line 37: I wonder if this is getting on to where we don't want to be doing this with string concatenation but use something like `from urllib.parse import urlencode` instead. I think it means we're more likely to get encoding issues right and also increases the readability.

```suggestion
        path = f"courses/uuid:{course_id}/resources?" + urlencode({
            "type": "file",
            "limit": PAGINATION_LIMIT,
            "fields": "id,name,modified,mimeType"
        })
```

This would result in `type=file&limit=10&fields=id%2Cname%2Cmodified%2CmimeType`, instead of `type=file&limit=10&fields=id,name,modified,mimeType` which might actually be correct, and a good reason to do this.

If Blackboard is weird and doesn't understand it, then you can add `safe=","` to the call.

Also you can do this as well:

```python
print(urlencode({
    "type": "file",
    "limit": PAGINATION_LIMIT,
    "fields": ["id", "name", "modified", "mimeType"]
}, doseq=True))
```

Which results in `type=file&limit=10&fields=id&fields=name&fields=modified&fields=mimeType` which _should_ be the same, but who knows if Blackboard would accept it.

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -70,7 +71,10 @@ def test_it_with_pagination(self, svc, basic_client, blackboard_list_files_schem
 
         # It called the Blackboard API three times getting the three pages.
         assert basic_client.request.call_args_list == [
-            call("GET", "courses/uuid:COURSE_ID/resources?type=file&limit=200"),
+            call(
+                "GET",
+                "courses/uuid:COURSE_ID/resources?type=file&limit=200&fields=id,name,modified,mimeType",
+            ),
```
jon-betts, line 77: Could be a good place for `Any.url` so we don't mind about changes like argument order and encoding.

jon-betts (approved):
Looks like a good efficiency improvement. These can really add up.

I wonder if we've got to the point where we should be use `urlencode` instead of hand combining strings?

lms/services/blackboard_api/client.py
seanh: :+1: Done. I didn't try the `doseq` thing because even if it does work it's not documented in Blackboard's docs. Best to stick to what their docs say

tests/unit/lms/services/blackboard_api/client_test.py
seanh: This does work but it reads a bit misleadingly because the method doesn't actually accept URLs, it takes paths

PR #2974, opened 2021-07-15
Don't format query strings ourselves

seanh:
We don't need to format query strings ourselves: we're using `requests`.


seanh:
I just realised this won't work because Blackboard sends us back the next page URLs with query strings already on them

PR #2976, opened 2021-07-19
Use rpc groups sync for both canvas groups and sections

marcospri:
This is something I missed when we did this, the condition to get the groups via rpc is was only checking sections. Now it checks that either sections or groups are enabled

# Testing

(on `master`, to reproduce the problem)

- Disable sections on http://localhost:8001/admin/instance/Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a/

- Open https://hypothesis.instructure.com/courses/125/assignments/1575 check that the groups the client renders are the group set's ones (Test group, another test group...)

- Force `lms/resources/lti_launch.py::canvas_sections_supported` to return False

- Check  https://hypothesis.instructure.com/courses/125/assignments/1575. Now the group in the client is `Developer Test Course`

(switch to this branch, keep `lms/resources/lti_launch.py::canvas_sections_supported` to returning False)

- Check  https://hypothesis.instructure.com/courses/125/assignments/1575. The client should display the groups set's groups again.

### DIFF
--- a/tests/unit/lms/resources/_js_config/__init___test.py
+++ b/tests/unit/lms/resources/_js_config/__init___test.py
@@ -420,7 +420,7 @@ class TestJSConfigAuthToken:
 class TestJSConfigAPISync:
     """Unit tests for the api.sync sub-dict of JSConfig."""
 
-    @pytest.mark.usefixtures("section_groups_on")
+    @pytest.mark.usefixtures("canvas_sections_on")
     def test_it(self, sync, pyramid_request, GroupInfo):
         assert sync == {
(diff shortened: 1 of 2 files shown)


tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -647,10 +655,15 @@ def context():
 
 
 @pytest.fixture
-def section_groups_on(context):
+def canvas_sections_on(context):
```
marcospri, line 658: Remaded this one, having `groups` in the name was a bit confusing with the new feature

seanh (approved):

PR #2985, opened 2021-07-20
Split out `OAuthHTTPService` from `HTTPService`

seanh:
Split `HTTPService` into two classes `HTTPService` and `OAuthHTTPService`. Two smaller classes instead of one bigger one. This will make more sense when `OAuthHTTPService.request()` is extended with more OAuth-related functionality.

### DIFF
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -7,9 +7,7 @@ from lms.services.exceptions import HTTPError
 class HTTPService:
     """Send HTTP requests with `requests` and receive the responses."""
 
-    def __init__(self, oauth2_token_service, _session=None):
-        self._oauth2_token_service = oauth2_token_service
-
+    def __init__(self, _session=None):
         # A requests session is used so that cookies are persisted across
         # requests and urllib3 connection pooling is used (which means that
         # underlying TCP connections are re-used when making multiple requests
@@ -33,7 +31,7 @@ class HTTPService:
     def delete(self, *args, **kwargs):
         return self.request("DELETE", *args, **kwargs)
 
-    def request(self, method, url, timeout=(10, 10), oauth=False, **kwargs):
+    def request(self, method, url, timeout=(10, 10), **kwargs):
         """
         Send a request with `requests` and return the requests.Response object.
 
@@ -56,12 +54,6 @@ class HTTPService:
             response download. It's a time limit on how long to wait *between
             bytes from the server*. The entire download can take much longer.
 
-        :param oauth: Include an OAuth 2 access token in the request.
-            If oauth=True the current user's access token will be looked up in
-            the database and included in an Authorization header in the
-            request.
-        :type oauth: bool
-
         :param kwargs: Any other keyword arguments will be passed directly to
             requests.Session().request():
             https://docs.python-requests.org/en/latest/api/#requests.Session.request
@@ -81,17 +73,8 @@ class HTTPService:
             HTTPError.__cause__.
 
             The error response will be available as HTTPError.response.
-
-        :raise OAuth2TokenError: If oauth=True was given but we don't have an
-            access token for the current user in our DB
         """
         response = None
-
-        if oauth:
-            kwargs.setdefault("headers", {})
-            assert "Authorization" not in kwargs["headers"]
-            access_token = self._oauth2_token_service.get().access_token
-            kwargs["headers"]["Authorization"] = f"Bearer {access_token}"
 
         try:
             response = self._session.request(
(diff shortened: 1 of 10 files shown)


lms/services/http.py
```diff
@@ -107,5 +90,5 @@ def request(self, method, url, timeout=(10, 10), oauth=False, **kwargs):
         return response
 
 
-def factory(_context, request):
-    return HTTPService(request.find_service(name="oauth2_token"))
+def factory(_context, _request):
+    return HTTPService()
```
jon-betts, line 94: :+1:

lms/services/oauth_http.py
```diff
@@ -0,0 +1,49 @@
...
+        return self.request("PUT", *args, **kwargs)
+
+    def post(self, *args, **kwargs):
+        return self.request("POST", *args, **kwargs)
+
+    def patch(self, *args, **kwargs):
+        return self.request("PATCH", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        return self.request("DELETE", *args, **kwargs)
```
jon-betts, line 21: :sob: - I think my general view on this is well known.

Another tack to take if we don't want to _be_ a session is to stop pretending to be one. Just make callers use `request` instead.

How many of these methods get called? Do we need these yet / at all?

lms/services/oauth_http.py
```diff
@@ -0,0 +1,49 @@
+class OAuthHTTPService:
```
jon-betts, line 1: Something about the name `OAuthHTTPService` is a little odd to me. It's not a service for "doing" OAuth, so much as doing something else which happens to use `OAuth`. I can't think of anything better that isn't long (like `OAuthAuthenticatedHTTPService`), I don't think it's going to cause too much confusion though.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+        http_service.request.assert_called_once_with(
+            sentinel.method,
+            sentinel.url,
+            headers={
+                "Authorization": f"Bearer {oauth2_token_service.get.return_value.access_token}",
+                "Foo": "bar",
+            },
+        )
+        assert response == http_service.request.return_value
```
jon-betts, line 21: This set of assertions is very clear

jon-betts (approved):
These PRs are being reviewed in a slightly different way than normal due to time constraints, so while I have opinions on the general design we aren't going into that in much detail here. Due to the way these PRs are split up it's also hard to know where the appropriate place to suggest changes to certain things might be. So I'll just say things as they occur to me. If the feedback would be more appropriate on another PR, please just mentally translate as feedback for the place you'd think most appropriate.

This looks fine to me as far as achieving what it sets out to. Having the non OAuth HTTP service accepting OAuth related things struck me as very odd before.

I think it's worth thinking about whether the `HTTPService` is actually a private collaborator of the `OAuthHTTPService` at this point. It takes no arguments, and I don't think it's used anywhere else. This would probably make reading the `OAuthHTTPService` a bit easier and maintain any testing boundaries you might want.

We've discussed my ideas about inheritance for some of this before, so I won't go over that again, but instead I'll make a new suggestion. If we don't want to _be_ a session, maybe we shouldn't ape it's interface either. We have a lot of convenience methods added to make ourselves appear to be a session without actually going the extra step of being one. Are they necessary?

lms/services/oauth_http.py
seanh: Originally it only had `request()` and Marcos requested that `get()` and pals be added for convenience. I don't think it's too bad: our services don't want to _be_ a session they just want to forward to _some_ of `requests.Sessions`'s methods (several of them), not everything. It means some duplicated boilerplate but they're just trivial forwarding methods

lms/services/oauth_http.py
seanh: `OAuthRequestsService`a maybe. Or `OAuthClient`

seanh:
> I think it's worth thinking about whether the `HTTPService` is actually a private collaborator of the `OAuthHTTPService` at this point. It takes no arguments, and I don't think it's used anywhere else. This would probably make reading the `OAuthHTTPService` a bit easier and maintain any testing boundaries you might want.

`HTTPService` is actually used directly as a service in some places: <https://github.com/hypothesis/lms/blob/bf044b01858db7630ad038743dc9cc3d14e07281/lms/services/h_api.py#L26> and <https://github.com/hypothesis/lms/blob/bf044b01858db7630ad038743dc9cc3d14e07281/lms/services/lti_outcomes.py#L22>. In general the idea is that if you want to send HTTP requests but don't want OAuth 2 authentication (which we often don't) then use `HTTPService`. But if you want to make requests to an OAuth API then use `OAuthHTTPService`.

PR #2987, opened 2021-07-20
Move `access_token_request()` into `OAuthHTTPService`

seanh:
Move the Blackboard API client's `BasicClient.get_token()` to `OAuthHTTPService.access_token_request()` .

This is mostly because I want a future commit to add `refresh_token_request()` to `OAuthHTTPService` as well, and `access_token_request()` and `refresh_token_request()` go together (they even share code).

The method was renamed from `get_token()` to `access_token_request()` because I think `access_token_request()` is a better name as it sends an "access token request" (what the OAuth 2.0 spec calls this particular request), and the name goes nicely with `refresh_token_request()` which will be added in a future commit (and which will send what the spec calls a "refresh token request") and with the existing `request()`.

I'm not sure that this `access_token_request()` method is fully generic yet. It sends the type of access token requests that the Blackboard API requires. In particular it sends the client ID and secret in a basic HTTP auth header. I think some servers may require these in the body. We can adapt the method as needed as we add more servers.

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -2,7 +2,6 @@ from marshmallow import INCLUDE, fields
 
 from lms.services.exceptions import HTTPError, OAuth2TokenError
 from lms.validation import RequestsResponseSchema, ValidationError
-from lms.validation.authentication import OAuthTokenResponseSchema
 
 
 class BlackboardErrorResponseSchema(RequestsResponseSchema):
@@ -54,7 +53,6 @@ class BasicClient:
         redirect_uri,
         http_service,
         oauth_http_service,
-        oauth2_token_service,
     ):  # pylint:disable=too-many-arguments
         self.blackboard_host = blackboard_host
         self.client_id = client_id
@@ -63,28 +61,13 @@ class BasicClient:
 
         self._http_service = http_service
         self._oauth_http_service = oauth_http_service
-        self._oauth2_token_service = oauth2_token_service
 
     def get_token(self, authorization_code):
-        # Send a request to Blackboard to get an access token.
-        response = self._http_service.post(
-            self._api_url("oauth2/token"),
-            data={
-                "grant_type": "authorization_code",
-                "redirect_uri": self.redirect_uri,
-                "code": authorization_code,
-            },
+        self._oauth_http_service.get_access_token(
+            token_url=self._api_url("oauth2/token"),
+            redirect_uri=self.redirect_uri,
             auth=(self.client_id, self.client_secret),
-        )
-
-        validated_data = OAuthTokenResponseSchema(response).parse()
-
-        # Save the access token to the DB.
-        self._oauth2_token_service.save(
-            validated_data["access_token"],
-            validated_data.get("refresh_token"),
-            # pylint:disable=no-member
-            validated_data.get("expires_in"),
+            authorization_code=authorization_code,
         )
 
     def request(self, method, path):
--- a/lms/services/oauth_http.py
+++ b/lms/services/oauth_http.py
@@ -1,3 +1,6 @@
+from lms.validation.authentication import OAuthTokenResponseSchema
+
+
 class OAuthHTTPService:
     """Send OAuth 2.0 requests and return the responses."""
 
@@ -41,6 +44,41 @@ class OAuthHTTPService:
         headers["Authorization"] = f"Bearer {access_token}"
 
         return self._http_service.request(method, url, headers=headers, **kwargs)
+
+    def get_access_token(self, token_url, redirect_uri, auth, authorization_code):
+        """
+        Make an access token request and save the token in the DB.
+
+        Send an OAuth 2.0 "access token request"
+        (https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3) to get a
+        new access token for the current user and save it to the DB.
+
+        :raise HTTPError: if the HTTP request fails
+        :raise ValidationError: if the server's access token response is invalid
+        """
+        self._token_request(
+            token_url=token_url,
+            auth=auth,
+            data={
+                "redirect_uri": redirect_uri,
+                "grant_type": "authorization_code",
+                "code": authorization_code,
+            },
+        )
+
+    def _token_request(self, token_url, data, auth):
+        response = self._http_service.post(token_url, data=data, auth=auth)
+
+        validated_data = OAuthTokenResponseSchema(response).parse()
+
+        self._oauth2_token_service.save(
+            validated_data["access_token"],
+            validated_data.get("refresh_token"),
+            # pylint:disable=no-member
+            validated_data.get("expires_in"),
+        )
+
+        return validated_data["access_token"]
 
 
 def factory(_context, request):
--- a/tests/unit/lms/services/blackboard_api/_basic_test.py
+++ b/tests/unit/lms/services/blackboard_api/_basic_test.py
@@ -7,7 +7,6 @@ from lms.services.blackboard_api._basic import (
     BlackboardErrorResponseSchema,
 )
 from lms.services.exceptions import HTTPError, OAuth2TokenError
-from lms.validation import ValidationError
 from tests import factories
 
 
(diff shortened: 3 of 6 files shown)


lms/services/oauth2_token.py
```diff
@@ -59,8 +82,24 @@ def get(self):
...
+
+        validated_data = OAuthTokenResponseSchema(response).parse()
+
+        self.save(
+            validated_data["access_token"],
+            validated_data.get("refresh_token"),
+            validated_data.get("expires_in"),  # pylint:disable=no-member
+        )
+
+        return validated_data["access_token"]
```
seanh, line 96: This is split into a separate helper method because it's also going to be called by `refresh_token_request()` in future, see https://github.com/hypothesis/lms/pull/2988

seanh:
I think I probably want to move this method onto `OAuthHTTPService` instead...

lms/services/oauth_http.py
```diff
@@ -42,6 +45,41 @@ def request(self, method, url, headers=None, **kwargs):
...
+        validated_data = OAuthTokenResponseSchema(response).parse()
+
+        self._oauth2_token_service.save(
+            validated_data["access_token"],
+            validated_data.get("refresh_token"),
+            # pylint:disable=no-member
+            validated_data.get("expires_in"),
+        )
+
+        return validated_data["access_token"]
```
seanh, line 81: This is split into a separate helper method because it's also going to be called by `refresh_token_request()` in future, see https://github.com/hypothesis/lms/pull/2988

lms/services/blackboard_api/_basic.py
```diff
@@ -63,28 +61,13 @@ def __init__(
...
-        response = self._http_service.post(
-            self._api_url("oauth2/token"),
-            data={
-                "grant_type": "authorization_code",
-                "redirect_uri": self.redirect_uri,
-                "code": authorization_code,
-            },
+        self._oauth_http_service.access_token_request(
+            token_url=self._api_url("oauth2/token"),
+            redirect_uri=self.redirect_uri,
```
jon-betts, line 68: Passing these two values every time strikes me as a bit odd. If this wasn't a service, I'd expect the interface to be something more like:

```
client = OAuthThing(token_url=..., redirect_uri=..., client_id=..., client_secret=...)
client.access_token_request()
client.refresh_token_request(token)
```

These don't seem like things that change over the life-time of a non-service object. This would mean less state having to be held by the caller which is only of interest to the object. Basically there's a lack of encapsulation here I think.

This suggests to me that perhaps what we want from the service is for it to return an object we can make requests with, rather than be the object we make requests with. So maybe something like:

```
client = self._oauth_http_service.get_authenticated_session(token_url=..., redirect_uri=..., client_id=..., client_secret=...)
```

lms/services/oauth_http.py
jon-betts: Yep good split. It's similar to how it's done in `lms.services.canvas_api._authenticated`

tests/unit/lms/services/blackboard_api/_basic_test.py
```diff
@@ -163,23 +105,12 @@ def test_request_raises_HTTPError_if_the_HTTP_request_fails(
...
-    def basic_client(self, http_service, oauth_http_service, oauth2_token_service):
+    def basic_client(self, http_service, oauth_http_service):
         return BasicClient(
             blackboard_host="blackboard.example.com",
             client_id=sentinel.client_id,
             client_secret=sentinel.client_secret,
             redirect_uri=sentinel.redirect_uri,
             http_service=http_service,
             oauth_http_service=oauth_http_service,
-            oauth2_token_service=oauth2_token_service,
```
jon-betts, line 174: It's good we've tucked away one element, but it still sticks out to me that `client_id`, `client_secret` and `redirect_uri` which are all OAuth related aren't encapsulated inside the service in some way too.

lms/services/oauth_http.py
```diff
@@ -42,6 +45,41 @@ def request(self, method, url, headers=None, **kwargs):
 
         return self._http_service.request(method, url, headers=headers, **kwargs)
 
+    def access_token_request(self, token_url, redirect_uri, auth, authorization_code):
```
jon-betts, line 48:
```suggestion
    def get_access_token(self, token_url, redirect_uri, auth, authorization_code):
```

This is phrased a bit noun-y for my taste. This does not return a request but make one. I think `request_access_token` would work too.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -36,24 +37,104 @@ def test_convenience_methods(self, method, svc, http_service, oauth2_token_servi
...
+        http_service,
+        OAuthTokenResponseSchema,
+        oauth_token_response_schema,
+        oauth2_token_service,
+    ):
+        oauth_token_response_schema.parse.return_value = {
+            "access_token": "access_token",
+            "refresh_token": "refresh_token",
+            "expires_in": 1234,
+        }
```
jon-betts, line 70: This might be better as part of the fixture, then later on, instead of having matching values we'd say:

```python
    response = oauth_token_response_schema.parse.return_value
    oauth2_token_service.save.assert_called_once_with(
        response["access_token"], response["refresh_token"], response["expires_in"]
    )
```

Which would make this test a little shorter and more explicit about where the values some from.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -36,24 +37,104 @@ def test_convenience_methods(self, method, svc, http_service, oauth2_token_servi
...
+    ):
+        http_service.post.side_effect = HTTPError
+
+        with pytest.raises(HTTPError):
+            svc.access_token_request(
+                sentinel.token_url,
+                sentinel.redirect_uri,
+                sentinel.auth,
+                sentinel.authorization_code,
+            )
```
jon-betts, line 104: A caller pattern might be appropriate here to cut down on the waffle? We explicitly don't care what the method was called with in these failure cases.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -36,24 +37,104 @@ def test_convenience_methods(self, method, svc, http_service, oauth2_token_servi
...
+        }
+
+        svc.access_token_request(
+            sentinel.token_url,
+            sentinel.redirect_uri,
+            sentinel.auth,
+            sentinel.authorization_code,
+        )
+
+        oauth2_token_service.save.assert_called_once_with("access_token", None, None)
```
jon-betts, line 136: Small thing, but I'd mentally group the happy cases together and the failure cases after.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -71,3 +152,13 @@ def test_it(
...
+
+
+@pytest.fixture(autouse=True)
+def OAuthTokenResponseSchema(patch):
+    return patch("lms.services.oauth_http.OAuthTokenResponseSchema")
+
+
+@pytest.fixture
+def oauth_token_response_schema(OAuthTokenResponseSchema):
+    return OAuthTokenResponseSchema.return_value
```
jon-betts, line 164: I think both of these fixtures are local to `TestOAuthHTTPService`. Also I believe we normally put our `autouse` patches after other fixtures.

jon-betts (changes requested):
Lots of comment in this ticket, but the areas being suggested for changes are:

* Probably have a think about `access_token_request()` as a name. I think `request_access_token()` or `get_access_token()` imply an action, rather than implying they return or are a `Request`
* The `oauth_http_test.py` has some fixtures at the end which are local to one test. I think it's also our style to have `autouse`'d patch fixtures last in any block of fixtures.

seanh:
Rebased...

lms/services/blackboard_api/_basic.py
seanh: Yeah, I think having a two-step process where you first create an `OAuthClient` object (passing it reusable settings like `client_id`, `client_secret`, etc) and then make a request (passing only request-specific arguments) likely makes sense. I had a scratch version of that at one point but I held off on doing it right now because I wanted to minimize the scope of last-second code changes in order to get the feature shipped on time.

I don't think this is a question of service or not-service though. It's useful for `OAuthHTTPService` to be a service because it relies on two other services (`HTTPService` and `OAuth2TokenService`) and the service factory can inject those, hiding those dependencies from `OAuthHTTPService`'s users. In future the service can also inject any other services or config settings etc that `OAuthHTTPService` comes to depend on or can vary how it constructs the service depending on the context etc, in general hiding service construction from the service's users is useful.

So I think it's a question of how `OAuthHTTPService` is designed.

The scratch version I had was something like this:

```python

Comment (seanh) [reply lms/services/blackboard_api/_basic.py:68]: > This suggests to me that perhaps what we want from the service is for it to return an object we can make requests with, rather than be the object we make requests with. So maybe something like:
> 
> ```
> client = self._oauth_http_service.get_authenticated_session(token_url=..., redirect_uri=..., client_id=..., client_secret=...)
> ```

Ah, yes pretty much exactly what I was thinking. But I'm not totally sure. `get_authenticated_session()` is going to end up with too many arguments.

File: lms/services/oauth_http.py (hunk 5 above)
Comment (seanh) [reply lms/services/oauth_http.py:48]: Yep, I went with `get_access_token()`

**Comment by seanh** [comment]
> Probably have a think about `access_token_request()` as a name. I think `request_access_token()` or `get_access_token()` imply an action, rather than implying they return or are a Request

Yep, I went with `get_access_token()`.

File: tests/unit/lms/services/oauth_http_test.py (hunk 6 above)
Comment (seanh) [reply tests/unit/lms/services/oauth_http_test.py:70]: I didn't move it into the fixture as it's only used by this one test, but I put it in a variable: https://github.com/hypothesis/lms/pull/2987/commits/e1715f11fb313de4da2d31223d2792e9a5d9b4e6

File: tests/unit/lms/services/oauth_http_test.py (hunk 7 above)
Comment (seanh) [reply tests/unit/lms/services/oauth_http_test.py]: Done: https://github.com/hypothesis/lms/pull/2987/commits/1a1064e34fe3d19d642560117aae89ccb020b1f9

File: tests/unit/lms/services/oauth_http_test.py (hunk 8 above)
Comment (seanh) [reply tests/unit/lms/services/oauth_http_test.py]: Done: https://github.com/hypothesis/lms/pull/2987/commits/dee26693a7e973d130ed226811c6831abc2a837c

File: tests/unit/lms/services/oauth_http_test.py (hunk 9 above)
Comment (seanh) [reply tests/unit/lms/services/oauth_http_test.py:158]: Our usual style is to make patch fixtures be top-level functions. So for example if the module under test contains:

```python
from lms.validation.authentication import OAuthTokenResponseSchema
```

then the corresponding test module will contain a module-level function (not a class methods) like this:

```python
@pytest.fixture(autouse=True)
def OAuthTokenResponseSchema(patch):
    return patch("lms.services.oauth_http.OAuthTokenResponseSchema")
```

The reason is that `OAuthTokenResponseSchema` is being imported into the `oauth_http.py` _module_'s namespace. All code in `oauth_http.py` has access to the name `OAuthTokenResponseSchema`. So we want to make sure that all `oauth_http_test.py` tests are patching `OAuthTokenResponseSchema`.

When we put a convenience fixture like `oauth_token_response_schema()` that just returns `OAuthTokenResponseSchema.return_value` we usually put the convenience fixture right after the patch fixture that it depends on.

I didn't know we had a style of putting all of our `autouse` `patch` fixtures after other fixtures. I can see that being useful to group the `patch` fixtures together. But I wonder if convenience fixtures like `oauth_token_response_schema()` should be allowed to go in-between the patch fixtures, since it makes sense for them to be attached to the fixture they depend on.

lms/services/blackboard_api/_basic.py
jon-betts: Yeah, I was just suggesting I suspect you ended up here because you started with a service, rather than you can't do this pattern with one. Yay factory factory?

Also if X is too many arguments on the constructor, better that than X + one or two per call?

jon-betts: Just had a brainwave: You could combine both to get the best of both. So have a `OAuthSettingsBag`, and then pass it through to the factory factory method.

tests/unit/lms/services/oauth_http_test.py
jon-betts: Yep, that works too

tests/unit/lms/services/oauth_http_test.py
jon-betts: I'm pretty sure you've asked me to move stuff like this after all other fixtures. If this is not a thing anymore that would be good as it's one less thing to worry about. I suppose perhaps there's an exception for this particular pattern of adding another fixture to proxy `return_value`?

I'm pretty 50/50 on the utility of things like this where we can use `OAuthTokenResponseSchema.return_value` directly in tests instead.

jon-betts (approved):
Yep looks good.

Is there an extra part to the auto use fixtures after others, a clause like: "...unless you are just making a convenience accessor in which case that goes after the fixture it's a wrapper of"?

lms/services/blackboard_api/_basic.py
seanh: Yes, perhaps.

I didn't end up here because I started with a service. I actually wrote the `OAuthService.make_client()` version but it wasn't a clear win to me for various reasons. It still has the too-many-arguments problem (if you don't add a `SettingsBag` class) and it also makes user code and the tests for that user code more complicated. So I just backed off from it for now. But I think we could revive at some point when there's more time to consider.

seanh:
> Is there an extra part to the auto use fixtures after others, a clause like: "...unless you are just making a convenience accessor in which case that goes after the fixture it's a wrapper of"?

Yep, I think so. Although to be honest I was not aware of the auto use fixtures after others clause, but now that I've heard it I think it's a good idea :) Group like fixtures together. I'm not sure that the rules on either of these need to be strictly policed though

PR #2988, opened 2021-07-20
Add `OAuthHTTPService.refresh_token_request()`

seanh:
Nothing uses this yet but a future commit will extend `BasicClient.request()` to use it.

### DIFF
--- a/lms/services/oauth_http.py
+++ b/lms/services/oauth_http.py
@@ -1,4 +1,14 @@
+from marshmallow import fields
+
+from lms.services import HTTPError, OAuth2TokenError
+from lms.validation import RequestsResponseSchema, ValidationError
 from lms.validation.authentication import OAuthTokenResponseSchema
+
+
+class _OAuthAccessTokenErrorResponseSchema(RequestsResponseSchema):
+    """Schema for parsing OAuth 2 access token error response bodies."""
+
+    error = fields.String(required=True)
 
 
 class OAuthHTTPService:
@@ -66,6 +76,42 @@ class OAuthHTTPService:
             },
         )
 
+    def refresh_access_token(self, token_url, redirect_uri, auth):
+        """
+        Make a refresh token request and save the new token in the DB.
+
+        Send an OAuth 2.0 "refresh token request"
+        (https://datatracker.ietf.org/doc/html/rfc6749#section-6) to get a new
+        access token for the current user and save it to the DB.
+
+        :raise OAuth2TokenError: if we don't have a refresh token for the user
+        :raise HTTPError: if the HTTP request fails
+        :raise ValidationError: if the server's access token response is invalid
+        """
+        refresh_token = self._oauth2_token_service.get().refresh_token
+
+        try:
+            return self._token_request(
+                token_url=token_url,
+                auth=auth,
+                data={
+                    "redirect_uri": redirect_uri,
+                    "grant_type": "refresh_token",
+                    "refresh_token": refresh_token,
+                },
+            )
+        except HTTPError as err:
+            try:
+                error_dict = _OAuthAccessTokenErrorResponseSchema(err.response).parse()
+            except ValidationError:
+                pass
+            else:
+                if error_dict["error"] == "invalid_grant":
+                    # Looks like our refresh token has expired or been revoked.
+                    raise OAuth2TokenError() from err
+
+            raise
+
     def _token_request(self, token_url, data, auth):
         response = self._http_service.post(token_url, data=data, auth=auth)
 
@@ -74,8 +120,7 @@ class OAuthHTTPService:
         self._oauth2_token_service.save(
             validated_data["access_token"],
             validated_data.get("refresh_token"),
-            # pylint:disable=no-member
-            validated_data.get("expires_in"),
+            validated_data.get("expires_in"),  # pylint:disable=no-member
         )
 
         return validated_data["access_token"]
--- a/tests/unit/lms/services/oauth_http_test.py
+++ b/tests/unit/lms/services/oauth_http_test.py
@@ -5,6 +5,7 @@ import pytest
 from lms.services.exceptions import HTTPError, OAuth2TokenError
 from lms.services.oauth_http import OAuthHTTPService, factory
 from lms.validation import ValidationError
+from tests import factories
 
 
 class TestOAuthHTTPService:
@@ -63,12 +64,6 @@ class TestOAuthHTTPService:
         oauth_token_response_schema,
         oauth2_token_service,
     ):
-        validated_data = oauth_token_response_schema.parse.return_value = {
-            "access_token": "access_token",
-            "refresh_token": "refresh_token",
-            "expires_in": 1234,
-        }
-
         call_get_access_token()
 
         http_service.post.assert_called_once_with(
@@ -81,6 +76,7 @@ class TestOAuthHTTPService:
             auth=sentinel.auth,
         )
         OAuthTokenResponseSchema.assert_called_once_with(http_service.post.return_value)
+        validated_data = oauth_token_response_schema.parse.return_value
         oauth2_token_service.save.assert_called_once_with(
             validated_data["access_token"],
             validated_data["refresh_token"],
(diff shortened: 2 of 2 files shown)


lms/services/oauth2_token.py
```diff
@@ -44,6 +44,27 @@ def access_token_request(self, token_url, redirect_uri, auth, authorization_code
             },
         )
 
+    def refresh_token_request(self, token_url, redirect_uri, auth, refresh_token):
```
seanh, line 47: I should remove the `refresh_token` argument from this: the method can call `OAuth2TokenService.get()` itself (with no arguments) to get the `refresh_token` itself. (It'll have to raise something if the `refresh_token` is `None`).

seanh:
I think I probably want to move this method onto `OAuthHTTPService` instead...

lms/services/blackboard_api/_basic.py
```diff
@@ -70,6 +70,31 @@ def get_token(self, authorization_code):
...
+                auth=(self.client_id, self.client_secret),
+            )
+        except HTTPError as err:
+            error_dict = BlackboardErrorResponseSchema(err.response).parse()
+
+            if error_dict.get("error") == "invalid_grant":
+                # Looks like our refresh token has expired or been revoked.
+                raise OAuth2TokenError() from err
+
+            raise
```
seanh, line 96: This error handling could actually move down into `OAuthHTTPService`, it's standard OAuth stuff: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

lms/services/oauth_http.py
```diff
@@ -1,6 +1,39 @@
...
+                "invalid_request",
+                "invalid_client",
+                "invalid_grant",
+                "unauthorized_client",
+                "unsupported_grant_type",
+            ]
+        ),
+    )
+    error_description = fields.String(required=False)
+    error_uri = fields.String(required=False)
```
seanh, line 34: These are the JSON body fields that the error response to an access token or refresh token request should have, according to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

lms/services/oauth_http.py
```diff
@@ -66,6 +99,42 @@ def access_token_request(self, token_url, redirect_uri, auth, authorization_code
...
+            )
+        except HTTPError as err:
+            try:
+                error_dict = _OAuthAccessTokenErrorResponseSchema(err.response).parse()
+            except ValidationError:
+                pass
+            else:
+                if error_dict["error"] == "invalid_grant":
+                    # Looks like our refresh token has expired or been revoked.
+                    raise OAuth2TokenError() from err
```
seanh, line 134: If we _can_ parse the error response body and identify it as an `"invalid_grant"` error then do so because we want to raise `OAuth2TokenError` so that the frontend shows the user the **Authorize** dialog rather than an error dialog. If not just discard the `ValidationError` and re-raise the original `HTTPError` which will show a **Something went wrong** dialog to the user.

lms/services/oauth_http.py
```diff
@@ -74,8 +143,7 @@ def _token_request(self, token_url, data, auth):
         self._oauth2_token_service.save(
             validated_data["access_token"],
             validated_data.get("refresh_token"),
-            # pylint:disable=no-member
-            validated_data.get("expires_in"),
+            validated_data.get("expires_in"),  # pylint:disable=no-member
```
seanh, line 146: PyLint keeps forcing me to move this comment around

lms/services/oauth_http.py
seanh: I think we should probably simplify this down to just what we need: `error = fields.String(required=True)`. The code that uses this schema just needs to be able to do `if error_dict["error"] == "invalid_grant":`. It doesn't need the schema to validate that the error string is one of those from the OAuth 2 spec and it doesn't need the `error_description` and `error_uri` fields. The schema itself is still useful because it'll catch cases like non-JSON responses, invalid JSON, or unexpected JSON (no `"error"` string) and group those all under the `except ValidationError: pass`

jon-betts: I'd definitely trim this down to the absolute minimum we need.

```suggestion
    error = fields.String(required=True)
```

The exclude is the default as inherited from `PlainSchema` so we don't have to have a `Meta` class either.

I think it's a bigger topic, but I think the schema here is essential, but not in a good way. I think we've bound up validation and marshalling together. We don't really want both, but as they are indivisible we can't get away from using it for now. We absolutely should make these be separate things.

It's a bit weird because we want to parse an error, so I think we might be missing some tools on the HTTP service or perhaps a wrapped request object, although that's going a bit far for now.

Another approach is to have something like this:

```python
class JSONResponseSchema(RequestsResponseSchema):
    """Any valid JSON response."""

    class Meta:
        unknown = INCLUDE
```

And then later do `if error_dict.get("error") == "invalid_grant"` as I think this is honest and clear about how this works (and can be used elsewhere when we don't actually care about structure).

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -135,6 +136,96 @@ def test_access_token_request_if_theres_no_refresh_token_or_expires_in(
...
+        http_service,
+        OAuthTokenResponseSchema,
+        oauth_token_response_schema,
+        oauth2_token_service,
+    ):
+        oauth_token_response_schema.parse.return_value = {
+            "access_token": "access_token",
+            "refresh_token": "refresh_token",
+            "expires_in": 1234,
+        }
```
jon-betts, line 151: Feel like I've given this feedback in another PR, but it's probably an even better idea with duplication. I'd suggest putting this in the fixture and referring to the values by reference rather than value.

tests/unit/lms/services/oauth_http_test.py
```diff
@@ -135,6 +136,96 @@ def test_access_token_request_if_theres_no_refresh_token_or_expires_in(
...
+
+    def test_refresh_token_request_raises_HTTPError_if_the_HTTP_request_fails(
+        self, svc, http_service
+    ):
+        http_service.post.side_effect = HTTPError
+
+        with pytest.raises(HTTPError):
+            svc.refresh_token_request(
+                sentinel.token_url, sentinel.redirect_uri, sentinel.auth
+            )
```
jon-betts, line 189: Totally hypocritical of me, but I think this gets a pass without a caller as it crams onto nearly one line

jon-betts (changes requested):
I think the `RequestsResponseSchema` is a very roundabout way of mostly just getting `.json` from a request and catching exceptions. I think I'd really like this _not_ to be the responsibility of the schema, but I think that's a topic for a different day.

In this case as the two things are bound together and we really don't want much from the validation side I'd suggest cutting the schema down to the absolute minimum we need. The rest is really just confusing as it plays no part at all.

I've made a few suggestions to try and fix that:

* Cut the existing schema down to size
* Be honest that we only want the JSON validation part and have a content neutral schema (a bit of a oxymoron for me), but doesn't swim against the grain of our current approach
* Have a `json_from_request(request)` method which we use and is also used by `RequestsResponseSchema` so we can choose when we want one or the other

There are some other comments and stuff, but this is the only part I'm recommending we change for now.

lms/services/oauth_http.py
seanh: For now I've just trimmed this down to one line: https://github.com/hypothesis/lms/pull/2988/commits/cc1513e70f7b8dcd7d03c1fdf8770c9cb8d966b4

I like the idea of a generic schema that just validates that the response is a valid JSON response with a top-level object but doesn't validate anything about the _contents_ of that object. And then the caller is just expected to use `get()`. Something like this:

```python
try:
    foo_dict = JSONObjectResponseSchema(response).parse()
except ValidationError:
    # This could be because:
    # 1. The response isn't a JSON response at all
    # 2. It claims to be a JSON response but the body fails to parse as JSON
    # 3. It is a valid JSON response but not a top-level object (e.g. it's a list)
    ...

# If you get here then you know that foo_dict must be a dict,
# but you don't know anything about the contents of that dict.
if foo_dict.get("bar") == "whatever":
    ...
```

I think this might not be _quite_ as simple as the `JSONResponseSchema` example that you gave because it needs to deal with the cases of valid but non-object JSON responses where trying to call `get()` would raise `AttributeError`. In my example above I've got it raising `ValidationError` if it's not a top-level object.

Elsewhere in the code [`BlackboardErrorResponseSchema`](https://github.com/hypothesis/lms/blob/4b3fb3f2d6fd23f78b87dddfeb8eb27f20a3fa79/lms/services/blackboard_api/_basic.py#L8-L43) does something very similar. The fields (`status` and `message`) aren't really necessary on that (@marcospri just requested that they be added for documentation purposes). If the response is not a JSON response, not a JSON-parseable body, or not a top-level object, then `BlackboardErrorResponseSchema` _returns an empty dict_. It never raises `ValidationError`. This makes the calling code simpler as it can just do `error_dict = BlackboardErrorResponseSchema(err.response).parse()` without catching `ValidationError` and then use `error_dict.get()` without any concern about that raising `AttributeError` because `error_dict` isn't a dict. But I think returning an empty dict when the response body isn't an empty dict might be a little weird. Although it makes the calling code slightly more complicated I think raising `ValidationError` might be more sensible.

I think it's probably useful as a Marshmallow schema rather than a `get_dict()` Python function or something because then other schema classes can subclass it and add particular fields when desired. And also just so that we have a consistent way of parsing and validating requests and responses: by using Marshmallow schemas.

Anyway, for time's sake I'm going to leave introducing a generic `JSONObjectResponseSchema` for a follow-up PR.

* * *

I don't know about adding more to `HTTPService` but maybe. Perhaps you could pass an argument to tell it that you're expecting a JSON object in response and have it raise otherwise, or something like that.

* * *

I'm not sure I agree with your more general point that Marshmallow schemas bounding parsing and validation together is a problem. The pattern seems sensible to me: if you have a request and you're going to read data from it (JSON body, form body, query params, ...) you write a `PyramidRequestSchema` subclass and use that to parse the request. If you have a _response_ you do the same, but with a `RequestsResponseSchema` subclass. In all cases I've seen so far we _do_ want to both parse the request and validate the data. And it's handy to gather all the things that can go wrong (it's not a JSON response; the body isn't valid JSON; the body isn't a top-level object; the body's fields aren't as expected, ...) under a single `ValidationError`.

The world this came from is one where code in views and services just read requests and responses directly. Due to the difficulty of getting this right (remembering to handle all the possible cases) this led to dozens of bugs, all sorts of ways the code could crash because cases were forgotten, Sentry getting flooded with crash reports. So we needed a Right Way to parse and validate external data before using it where we could implement that once (in things like `PyramidRequestSchema` and `RequestsResponseSchema`), get it right once, and re-use it.

I'm not seeing a reason why we should separate parsing and validation? It appears to be serving us well, and I don't have an aversion to using Marshmallow.

The one thing that comes to mind is that if we had an `http_service.request(..., expect_json_object=True)` that would raise if the response didn't have a JSON object body then we'd get the JSON object body back from that as a dict and if we wanted to further validate it we'd just want a schema for validating a dict not a schema for validating a response object. It's not totally clear to me whether or not that'd be an improvement though, I think we'd have to see. But I like that while you _can_ further validate the dict with a schema, you don't _have_ to if you don't need to

seanh:
Rebased...

tests/unit/lms/services/oauth_http_test.py
seanh: Yep, now that there's two of these I think it makes sense to move it into the fixture. Done: https://github.com/hypothesis/lms/pull/2988/commits/dc51b4d49f843c97783eacc400aaca92c1a6bfe5

tests/unit/lms/services/oauth_http_test.py
seanh: :man_shrugging: Sure. We do pass exactly the same args to `refresh_token_request()` every time we call it, so I think we could totally add a caller fixture. But I'll hold off

seanh:
In a previous PR I renamed `access_token_request()` to `get_token()` following a suggestion. In the same vein I've renamed `refresh_token_request()` to `refresh_access_token()` in this PR ("refresh" being a verb here). So we have `get_access_token()` and `refresh_access_token()` which seem to go together nicely. https://github.com/hypothesis/lms/pull/2988/commits/180c64cec5d52b6bec84f6ea653ece84923af1e8

lms/services/oauth_http.py
seanh: Some more discussion of this in Slack: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1627391871074500

PR #2992, opened 2021-07-21
Split TestBasicClient into multiple classes

seanh:
No functional change to the tests, just splitting one test class into multiple. This is going to make these tests easier to extend in future when more tests need to be added.

### DIFF
--- a/tests/unit/lms/services/blackboard_api/_basic_test.py
+++ b/tests/unit/lms/services/blackboard_api/_basic_test.py
@@ -51,8 +51,8 @@ class TestBlackboardErrorResponseSchema:
         assert BlackboardErrorResponseSchema(None).parse() == {}
 
 
-class TestBasicClient:
-    def test_get_token(self, basic_client, oauth_http_service):
+class TestBasicClientGetToken:
+    def test_it(self, basic_client, oauth_http_service):
         basic_client.get_token(sentinel.authorization_code)
 
         oauth_http_service.get_access_token.assert_called_once_with(
@@ -62,6 +62,8 @@ class TestBasicClient:
             authorization_code=sentinel.authorization_code,
         )
 
+
+class TestBasicClientRequest:
     @pytest.mark.parametrize(
         "path,expected_url",
         [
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/blackboard_api/_basic_test.py
```diff
@@ -51,8 +51,8 @@ def test_when_the_response_object_is_None(self):
         assert BlackboardErrorResponseSchema(None).parse() == {}
 
 
-class TestBasicClient:
-    def test_get_token(self, basic_client, oauth_http_service):
+class TestGetToken:
```
jon-betts, line 54:
```suggestion
class TestBasicClientGetToken:
```

As there are tests for different classes in here (`TestBlackboardErrorResponseSchema`) this is kind of confusing, as we are swapping between two different styles.

tests/unit/lms/services/blackboard_api/_basic_test.py
```diff
@@ -62,6 +62,8 @@ def test_get_token(self, basic_client, oauth_http_service):
             authorization_code=sentinel.authorization_code,
         )
 
+
+class TestRequest:
```
jon-betts, line 66:
```suggestion
class TestBasicClientRequest:
```

jon-betts (changes requested):
I'd recommend name spacing the test classes as this is not the only object under test in this module.

Also I'm going to assume this is to make space for future work or something? If this is all there is to it, I'd suggest not doing it. I don't think it adds much as is.

tests/unit/lms/services/blackboard_api/_basic_test.py
seanh: :+1: Done

tests/unit/lms/services/blackboard_api/_basic_test.py
seanh: :+1:

jon-betts (approved):

PR #2993, opened 2021-07-21
Add Blackboard refresh-and-retry support

seanh:
Fixes https://github.com/hypothesis/lms/issues/2033.

When a Blackboard API request fails for any reason try to get a new access token without user interaction by doing a refresh token request and then re-try the original API request once more.

If the refresh token request fails raise an `OAuth2TokenError` which will trigger the **Authorize Hypothesis** dialog to get a new access token with user interaction.

If the refresh token request succeeds but then the second attempt at the API request still fails raise an `HTTPError` which will trigger an error dialog (with **Try again** button).

Testing
=======

Delete all the access tokens from your DB. This will make the tests below easier to do by removing irrelevant access tokens from the DB:

    tox -qe dockercompose -- exec postgres psql -U postgres -c "DELETE FROM oauth2_token;"

Log in to https://aunltd-test.blackboard.com/ as either a teacher or a student, navigate to [Developer Test Course with Original Course View](https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline) and then:

- [x] Launch **localhost (make devdata) Blackboard Files Assignment**. You will have to click through the **Authorize** dialog and then the assignment should launch successfully.

  **The popup window might just open and then immediately close automatically**. This happens because the user has already authorized our app in Blackboard and Blackboard remembers this. If you want to actually see Blackboard's authorization dialog you have to go to [Tools > Application Authorization](https://aunltd-test.blackboard.com/ultra/tools/lti/launchFrame?toolHref=https:~2F~2Faunltd-test.blackboard.com~2Fwebapps~2Fblackboard~2Fexecute~2Fblti~2FlaunchPlacement%3Fblti_placement_id%3D_2_1%26wrapped%3Dtrue%26from_ultra%3Dtrue) and revoke Hypothesis Dev.

- [x] Launch the assignment again. This time it should launch successfully without asking you to authorize because it should re-use your previously saved access token. **The popup window should not appear and then close automatically**. You shouldn't see the popup window at all.

- [x] Break the access token in your DB. This simulates the access token having expired or having been revoked etc:

  ```
  ```

  Launch the assignment again. It should launch successfully **without** asking you to re-authorize because it should automatically use the refresh token to get a new access token. This is the whole point of refresh token support: it changes the user experience by not asking you to re-authorize as often. Again,  **the popup window should not appear and then close automatically**. You shouldn't see the popup window at all.

  You should see that a new access token has been saved to your DB and the access token is no longer `"foo"`:

  ```
  tox -qe dockercompose -- exec postgres psql -U postgres -c "SELECT access_token FROM oauth2_token;"
   access_token
  --------------
   8Pl***Znw
  (1 row)
  ```

- [x] Break the access token **and** refresh token in the DB. This will trigger a refresh request when we get an access token error and then the refresh request will fail:

  ```
  tox -qe dockercompose -- exec postgres psql -U postgres -c "UPDATE oauth2_token SET access_token = 'foo', refresh_token = 'bar';"
  ```

  Launch the assignment again. You should see the **Authorize Hypothesis** dialog and **not** an error dialog.

- [x] Break the code so that the public URL API request fails. This simulates something else going wrong with the API request (rather than our access and/or refresh token being expired or revoked):

  ```diff
  diff --git a/lms/services/blackboard_api/client.py b/lms/services/blackboard_api/client.py
  index caa788ce..4e59b533 100644
  --- a/lms/services/blackboard_api/client.py
  +++ b/lms/services/blackboard_api/client.py
  @@ -56,7 +56,7 @@ class BlackboardAPIClient:
           try:
               response = self._api.request(
                   "GET",
  -                f"courses/uuid:{course_id}/resources/{file_id}?fields=downloadUrl",
  +                f"foo",
               )
           except HTTPError as err:
               if err.response.status_code == 404:
  ```

  Launch the assignment again. You should see a **Something went wrong** error dialog with the details of the failed API request. The specific diff shown above will trigger a _Hypothesis couldn't find the assignment's PDF file in the Blackboard course_ error (because it triggers a 404 from Blackboard which makes our code think that the assignment's file can't be found). A change that triggers a different (non-404) error response from Blackboard error dialog (_A problem occurred while handling this request. Hypothesis has been notified._).

  Note that the code will do a refresh token request and re-try the API request a second time when this happens even though the error message from Blackboard was not related to our access token. We're deliberately being "dumb" here and doing refresh-and-retry in response to *any* error rather than trying to single out the specific access token-related errors that refresh-and-retry might actually fix. This is the simplest and most robust approach, giving us the highest chance of succeeding. Trying to filter out only those errors that refresh-and-retry will fix would mean more complicated code and would introduce the possibility of false negatives where refresh-and-retry would have worked but we didn't try it.

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -64,22 +64,28 @@ class BasicClient:
 
     def get_token(self, authorization_code):
         self._oauth_http_service.get_access_token(
-            token_url=self._api_url("oauth2/token"),
+            token_url=self.token_url,
             redirect_uri=self.redirect_uri,
             auth=(self.client_id, self.client_secret),
             authorization_code=authorization_code,
         )
 
     def request(self, method, path):
+        url = self._api_url(path)
+
         try:
-            return self._oauth_http_service.request(method, self._api_url(path))
-        except HTTPError as err:
-            error_dict = BlackboardErrorResponseSchema(err.response).parse()
+            return self._send(method, url)
+        except (OAuth2TokenError, HTTPError):
+            self._oauth_http_service.refresh_access_token(
+                self.token_url,
+                self.redirect_uri,
+                auth=(self.client_id, self.client_secret),
+            )
+            return self._send(method, url)
 
-            if error_dict.get("message") == "Bearer token is invalid":
-                raise OAuth2TokenError() from err
-
-            raise
+    @property
+    def token_url(self):
+        return self._api_url("oauth2/token")
 
     def _api_url(self, path):
         """Return the full Blackboard API URL for the given path."""
@@ -90,3 +96,14 @@ class BasicClient:
             path = "/learn/api/public/v1/" + path
 
         return f"https://{self.blackboard_host}{path}"
+
+    def _send(self, method, url):
+        try:
+            return self._oauth_http_service.request(method, url)
+        except HTTPError as err:
+            error_dict = BlackboardErrorResponseSchema(err.response).parse()
+
+            if error_dict.get("message") == "Bearer token is invalid":
+                raise OAuth2TokenError() from err
+
+            raise
--- a/lms/services/oauth_http.py
+++ b/lms/services/oauth_http.py
@@ -120,7 +120,7 @@ class OAuthHTTPService:
         self._oauth2_token_service.save(
             validated_data["access_token"],
             validated_data.get("refresh_token"),
-            validated_data.get("expires_in"),  # pylint:disable=no-member
+            validated_data.get("expires_in"),
         )
 
         return validated_data["access_token"]
--- a/tests/unit/lms/services/blackboard_api/_basic_test.py
+++ b/tests/unit/lms/services/blackboard_api/_basic_test.py
@@ -1,4 +1,4 @@
-from unittest.mock import sentinel
+from unittest.mock import DEFAULT, call, sentinel
 
 import pytest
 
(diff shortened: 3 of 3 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -64,22 +64,39 @@ def __init__(
...
+
+        try:
+            return send()
+        except (OAuth2TokenError, HTTPError):
+            self._oauth_http_service.refresh_token_request(
+                self.token_url,
+                self.redirect_uri,
+                auth=(self.client_id, self.client_secret),
+            )
+            return send()
```
seanh, line 95: I originally wanted to add generic "refresh and retry" support to `OAuthHTTPService` but I ended up implementing it here in the Blackboard API client instead. Do we know that we'll always want refresh-and-retry, for every OAuth API? And I don't think a `OAuthHTTPService.request(..., retry=True)` boolean to turn if off would suffice either: the decision about whether or not to refresh-and-retry may depend on API-specific details of the error response, we may want to refresh-and-retry for some errors but not others. You can see in the code above that some Blackboard-specific error parsing is necessarily involved.

It may be worth taking another look at Blackboard's `"Bearer token is invalid"` error response and seeing whether there's any standard OAuth 2 fields in it that could be used to identify it as an access token error. But I don't think there are.

seanh: Hmm, yeah, I think it might be possible to split out generic refresh-and-retry logic into `OAuthHTTPService.request()` and leave the Blackboard-specific error handling here in `BasicClient.request()`.

If we do that then the `OAuthHTTPService.request()` unit tests will have to use `patch.object(...)` to mock `OAuthHTTPService.refresh_token_request()` but I think that's fine.

jon-betts (approved):
Yep this works. Very good and clear testing instructions. Which is just as well because I am a bear of very little brain today.

My only suggestion would be to change the closured `send()` function into a conventional private method. I don't think we make much use of the fact it's a closure and it causes nesting we don't need and obscures the larger logic of the method.

lms/services/blackboard_api/_basic.py
```diff
@@ -64,22 +64,39 @@ def __init__(
...
+            except HTTPError as err:
+                error_dict = BlackboardErrorResponseSchema(err.response).parse()
 
-            if error_dict.get("message") == "Bearer token is invalid":
-                raise OAuth2TokenError() from err
+                if error_dict.get("message") == "Bearer token is invalid":
+                    raise OAuth2TokenError() from err
 
-            raise
+                raise
```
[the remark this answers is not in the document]
seanh: Sure, done

Issue #2990, opened 2021-07-21
Handle errors when users don't have permission to read a Blackboard Files assignment's file


seanh:
With Blackboard Files assignments it's possible that the user doesn't have permission to read the assignment's file. For example it's possible to configure a file in Blackboard so that teachers can read it but students can't. When this happens the Blackboard API for getting the file's public URL sends us a 404 response with a specific error message in the body:

```json
{
  "message": "Target folder or file (xid-9911_1) does not exist or user does not have Read|Write permission.",
  "status": 404
}
```

We currently show a generic *Something went wrong* error dialog when this happens. Instead we should show a specific *You don't have permission to read this assignment's file* dialog along with details about what an instructor needs to do to fix the problem (including a link to Blackboard's documentation of their file permissions UI, or our own documentation if we can't find a suitable Blackboard page). The body of the error response from Blackboard should also be included in the dialog's *Error Details* section.

See https://hypothes-is.slack.com/archives/C2BLQDKHA/p1626813341148100

PR #2996, opened 2021-07-22
Handle Blackboard file permission errors

seanh:
Fixes https://github.com/hypothesis/lms/issues/2990

Add a helpful error dialog for the case when the user doesn't have permission to read a Blackboard Files assignment's file.

### Problem

If the user launching a Blackboard Files assignment (probably a student) doesn't have permission to read the assignment's file then Blackboard's public URL API returns an error message like this:

```json
{
  "message": "Target folder or file (xid-9911_1) does not exist or user does not have Read|Write permission.",
  "status": 404
}
```

We currently show a generic error dialog in response to this, which doesn't help the user to understand what is wrong or what they can do to fix it. The **Error Details** aren't much use either:

Screenshot from 2021-07-22 12-10-32

### Solution

This PR special-cases 404s from Blackboard's public URL API and shows a more helpful error message:

Screenshot from 2021-07-22 11-41-20

**Note:** The weird "**API call failed**" bit in the error dialog is a pre-existing bug (https://github.com/hypothesis/lms/issues/2997), ignore that.

### Testing

1. Re-run `make devdata` to get the latest Blackboard assignments
2. Log in to https://aunltd-test.blackboard.com/ as a student
3. Go to [Developer Test Course with Original Course View](https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline)
4. Launch **localhost (make devdata) Blackboard Files Assignments: Student's Can't Read The File**
6. You should see the new error dialog. On master you'd see the old dialog.
7. Also launch **localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted**. You should see the same error dialog (we can't tell the difference between a file that has been deleted and one that the user lacks read permission for)

### Known limitations

* When the teacher launches the assignment it will probably work because the teacher probably has access to the file. So they'll think the assignment's fine. Then when their students try to launch it they'll get the error dialog. It would be better if we could detect the problem and warn teachers of it when teachers launch the assignment, but I don't know of any way to do that with the Blackboard API.

* The **Error Details** in the error dialog aren't very useful. They just say `{"file_id": "_10704_1"}`. It would be better to include the full error response from Blackboard here. Separate issue: https://github.com/hypothesis/lms/issues/2998

### DIFF
--- a/lms/static/scripts/frontend_apps/components/BasicLTILaunchApp.js
+++ b/lms/static/scripts/frontend_apps/components/BasicLTILaunchApp.js
@@ -26,12 +26,14 @@ import VitalSourceBookViewer from './VitalSourceBookViewer';
  * offered to remedy the problem.
  *
  * Note the two different naming conventions here:
+ *  - "blackboard_*" for the Blackboard specific cases where ErrorState matches the string coming from the backend and are all handled the same way
  *  - "canvas_*" for the canvas specific cases where ErrorState matches the string coming from the backend and are all handled the same way
  *  - "error-*" for the rest
  *
  * @typedef {'error-fetching'|
  *           'error-authorizing'|
  *           'error-reporting-submission'|
+ *           'blackboard_file_not_found_in_course'|
  *           'canvas_api_permission_error'|
  *           'canvas_file_not_found_in_course'|
  *           'canvas_group_set_not_found'|
@@ -123,6 +125,7 @@ export default function BasicLTILaunchApp() {
       e instanceof APIError &&
       e.errorCode !== null &&
       [
+        'blackboard_file_not_found_in_course',
         'canvas_api_permission_error',
         'canvas_file_not_found_in_course',
         'canvas_group_set_not_found',
--- a/lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
+++ b/lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
@@ -102,6 +102,35 @@ export default function LaunchErrorDialog({
           <p>Hypothesis needs your authorization to launch this assignment.</p>
         </BaseDialog>
       );
+    case 'blackboard_file_not_found_in_course':
+      return (
+        <BaseDialog
+          busy={busy}
+          error={error}
+          onRetry={onRetry}
+          title="Hypothesis couldn't find the file in the course"
+        >
+          <p>This might have happened because:</p>
+
+          <ul>
+            <li>
+              The file has been deleted from Blackboard: an instructor needs to
+              re-create the assignment with a new file
+            </li>
+            <li>
+              You don{"'"}t have permission to read the file: an instructor
+              needs to{' '}
+              <a
+                target="_blank"
+                rel="noreferrer"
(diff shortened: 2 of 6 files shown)


lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -102,6 +102,35 @@ export default function LaunchErrorDialog({
...
+              The file has been deleted from Blackboard: an instructor needs to
+              re-create the assignment with a new file
+            </li>
+            <li>
+              You don{"'"}t have permission to read the file: an instructor
+              needs to{' '}
+              <a
+                target="_blank"
+                rel="noreferrer"
+                href="https://web.hypothes.is/help/creating-hypothesis-enabled-readings-in-blackboard/"
```
seanh, line 126: This URL currently doesn't contain any help for this situation. Slack thread about what the URL should be: https://hypothes-is.slack.com/archives/C2BLQDKHA/p1626953417176600?thread_ts=1626813341.148100&cid=C2BLQDKHA

lms/static/scripts/frontend_apps/components/BasicLTILaunchApp.js
```diff
@@ -32,6 +32,7 @@ import VitalSourceBookViewer from './VitalSourceBookViewer';
  * @typedef {'error-fetching'|
  *           'error-authorizing'|
  *           'error-reporting-submission'|
+ *           'blackboard_file_not_found_in_course'|
```
robertknight, line 35: The comment above this `@typedef` needs a slight revision. It currently mentions only `canvas_` and `error-`-prefixed values.

lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -102,6 +102,35 @@ export default function LaunchErrorDialog({
...
+            <li>
+              The file has been deleted from Blackboard: an instructor needs to
+              re-create the assignment with a new file
+            </li>
+            <li>
+              You don{"'"}t have permission to read the file: an instructor
+              needs to{' '}
+              <a
+                target="_blank"
+                rel="noreferrer"
```
robertknight, line 125: The use of `noreferrer` is consistent with the other cases in this module, but it would be good to do away with the need for this attribute in future, as it may be useful for support if they can track how much inbound traffic to knowledgebase articles comes from these help pages. We could perhaps also facilitate that with some suitable query string.

lms/static/scripts/frontend_apps/components/test/BasicLtiLaunchApp-test.js
```diff
@@ -360,6 +360,54 @@ describe('BasicLTILaunchApp', () => {
         'https://via.hypothes.is/123'
       );
     });
+
+    it('shows Blackboard file not found in course error if content URL fetch fails with "blackboard_file_not_found_in_course" error', async () => {
+      // Make the initial URL fetch request reject with a Canvas API permission error.
```
robertknight, line 365:
```suggestion
      // Make the initial URL fetch request reject with a Blackboard API permission error.
```

robertknight (approved):
I haven't done manual testing of this myself, but the code changes look good. I see a couple of comments that need minor updates.

Issue #2760, opened 2021-05-31
Handle errors when getting Blackboard API access tokens


seanh:
If something goes wrong with `BlackboardAPIClient.get_token()`'s request to Blackboard and `HTTPService` raises an `HTTPError` or `HTTPValidationError` you end up with a JSON error message from our proxy API rendered directly to the user in the popup window. The same thing will happen if you click <kbd>Deny</kbd> instead of <kbd>Allow</kbd> on Blackboard's authorization page.

The desired behavior is our usual nice error dialog with <kbd>Try again</kbd> button like we have for the Canvas API.

The Canvas API code handles this with an exception view:

https://github.com/hypothesis/lms/blob/be46e8c356c0da97d86c53cedab28d4b31570cc4/lms/views/api/canvas/authorize.py#L100-L128

But there's a bunch of Canvas-specific stuff in that: the `"canvas_api.oauth.authorize"` route name is passed to `route_url()`, `js_config.enable_canvas_oauth2_redirect_error_mode()` is called, Canvas API scopes are passed to it.

We either need to write something very similar looking (but not exactly the same) for Blackboard or we need to do some refactoring to be able to share code. I'll create a separate issue for this.

Issue #2998, opened 2021-07-22
Show Blackboard API responses in error dialogs


seanh:
When we get an unexpected or error response from the Blackboard API we should show the <del>full</del> Blackboard API response (<del>status and body</del><ins>status and URL minus query params</ins>) in the **Error Details** section in our error dialog.

We don't want to show error response bodies or query params for security reasons but we think we can safely show the status code and the URL (without query params).

The reason for showing the URL is so that you can tell _which_ response failed (which API call, to which external service).

Done when:

- [x] The HTTP status codes and reasons of error responses from the Blackboard API are shown in the **Error Details** section of the frontend's error dialog
- [x] The URL of the failed request is also shown, **with query params stripped**
- [x] <del>Both of the above still apply when we do actually receive a successful response from the Blackboard API (e.g. a 200 OK) but the response doesn't have the form that we expect and our attempt to validate and parse it fails</del> Created a separate issue for this: https://github.com/hypothesis/lms/issues/3319
- [x] <del>The above still apply when an access token request fails and an error dialog is shown in the popup window</del> Created two separate issues: https://github.com/hypothesis/lms/issues/3322 https://github.com/hypothesis/lms/issues/3323

seanh:
We might need to sanitize some exception classes in order to do this. Thread: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1627492028036400

seanh:
Note to self: the details of the third-party error response should be reported to Sentry as well.

seanh:
Slack thread with some initial investigation: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1627492028036400

seanh:
Closing this but I opened some follow-up issues:

* https://github.com/hypothesis/lms/issues/3321
* https://github.com/hypothesis/lms/issues/3319
* https://github.com/hypothesis/lms/issues/3322
* https://github.com/hypothesis/lms/issues/3323

PR #2999, opened 2021-07-22
Handle errors getting blackboard access tokens

seanh:
Fixes https://github.com/hypothesis/lms/issues/2760

## Testing

Some of the tests below are testing existing behaviours that were implemented for Canvas and are just being inherited by Blackboard and aren't really implemented by the code in this PR. But there's no harm in re-testing everything to do with our popup window UX. And some of the tests _are_ testing what's actually in this PR.

* Log in to https://aunltd-test.blackboard.com/ as either a teacher or a student

* Delete all the access tokens from your DB in order to trigger the authorization process when you launch a Blackboard Files assignment:

  ```
  ```

* Revoke your Blackboard API authorization so that Blackboard's authorization popup window doesn't immediately close when it pops up. Go to [Tools > Application Authorization](https://aunltd-test.blackboard.com/ultra/tools/lti/launchFrame?toolHref=https:~2F~2Faunltd-test.blackboard.com~2Fwebapps~2Fblackboard~2Fexecute~2Fblti~2FlaunchPlacement%3Fblti_placement_id%3D_2_1%26wrapped%3Dtrue%26from_ultra%3Dtrue) and revoke Hypothesis Dev

* Go to [Developer Test Course with Original Course View](https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline) and launch **localhost (make devdata) Blackboard Files Assignment**

* Click **Authorize**

- [ ] Close the authorization popup window (by clicking the **x** in the top-right) _without_ clicking **Allow**. Back in the original browser tab it should still be showing the **Authorize** button and clicking it should re-open the popup window.

  https://user-images.githubusercontent.com/22498/126667499-ff418cbb-173c-403e-88b4-65c1b98a8001.mp4

- [ ] If you lose the popup window (don't close it, but just go back to the original tab) then clicking **Authorize** in the original tab should re-raise the popup window

  https://user-images.githubusercontent.com/22498/126667671-168ea427-9922-4246-b2ec-a69274a594ae.mp4

- [ ] In the popup window click **Deny** instead of **Allow**. You should see an **Authorization failed** error message with a **Try again** button that takes you back to Blackboard's authorization page.

  https://user-images.githubusercontent.com/22498/126667819-7b03d0c2-2f47-429a-9784-b77a6b74e0ea.mp4

- [ ] If you click **Deny** and then click **Close** instead of **Try again** then back in the original tab it should still be showing the **Authorize** button which should re-open the window

- [ ] Hack the code so that the server-to-server access token request that happens when you click **Allow** will fail:

  ```diff
  diff --git a/lms/services/blackboard_api/client.py b/lms/services/blackboard_api/client.py
  index caa788ce..bc266d22 100644
  --- a/lms/services/blackboard_api/client.py
  +++ b/lms/services/blackboard_api/client.py
  @@ -27,7 +27,7 @@ class BlackboardAPIClient:
           :raise services.HTTPError: if something goes wrong with the access
               token request to Blackboard
           """
  -        self._api.get_token(authorization_code)
  +        self._api.get_token("foo")
   
       def list_files(self, course_id):
           """Return the list of files in the given course."""
  ```

  Now open the popup window and click **Allow**. You should see an error dialog with a **Try again** button.

  **Known limitation:** It'd be useful if details from the error response from Blackboard were included in this dialog when this happens. There's a separate issue for that: https://github.com/hypothesis/lms/issues/2998

- [ ] Finally, un-hack the code again, re-open the popup window and click **Allow** again. The assignment should launch successfully. If you then re-launch the assignment it should not ask you for authorization again (not even opening a popup window that closes by itself) because it should have saved your access token.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -105,7 +105,7 @@ class JSConfig:
     def enable_oauth2_redirect_error_mode(
         self,
         error_details,
-        auth_url=None,
+        auth_route,
         is_scope_invalid=False,
         canvas_scopes=None,
     ):
@@ -118,14 +118,29 @@ class JSConfig:
 
         :param error_details: Technical details of the error
         :type error_details: str
-        :param auth_url: URL for the "Try again" button in the dialog
-        :type auth_url: str
+        :param auth_route: route for the "Try again" button in the dialog
+        :type auth_route: str
         :param is_scope_invalid: `True` if authorization failed because the
           OAuth client does not have access to all the necessary scopes
         :type is_scope_invalid: bool
         :param canvas_scopes: List of scopes that were requested
         :type canvas_scopes: list(str)
         """
+        if self._lti_user:
+            auth_url = self._request.route_url(
+                auth_route,
+                _query=[
+                    (
+                        "authorization",
+                        BearerTokenSchema(self._request).authorization_param(
+                            self._lti_user
+                        ),
+                    )
+                ],
+            )
+        else:
+            auth_url = None
+
         self._config.update(
             {
                 "mode": "oauth2-redirect-error",
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -104,8 +104,8 @@ class JSConfig:
 
     def enable_oauth2_redirect_error_mode(
         self,
-        error_details,
         auth_route,
+        error_details="",
         is_scope_invalid=False,
         canvas_scopes=None,
     ):
@@ -127,16 +127,12 @@ class JSConfig:
         :type canvas_scopes: list(str)
         """
         if self._lti_user:
+            bearer_token = BearerTokenSchema(self._request).authorization_param(
+                self._lti_user
+            )
             auth_url = self._request.route_url(
                 auth_route,
-                _query=[
-                    (
-                        "authorization",
-                        BearerTokenSchema(self._request).authorization_param(
-                            self._lti_user
-                        ),
-                    )
-                ],
+                _query=[("authorization", bearer_token)],
             )
         else:
             auth_url = None
--- a/lms/views/api/canvas/authorize.py
+++ b/lms/views/api/canvas/authorize.py
@@ -16,7 +16,7 @@ from pyramid.view import exception_view_config, view_config
 
 from lms.security import Permissions
 from lms.services import CanvasAPIServerError
-from lms.validation.authentication import BearerTokenSchema, OAuthCallbackSchema
+from lms.validation.authentication import OAuthCallbackSchema
 
 #: The Canvas API scopes that we need for our Canvas Files feature.
 FILES_SCOPES = (
@@ -112,18 +112,8 @@ def oauth2_redirect(request):
     renderer="lms:templates/api/oauth2/redirect_error.html.jinja2",
 )
 def oauth2_redirect_error(request):
-    auth_url = None
-    if request.lti_user:
-        authorization_param = BearerTokenSchema(request).authorization_param(
-            request.lti_user
-        )
-        auth_url = request.route_url(
-            "canvas_api.oauth.authorize",
-            _query=[("authorization", authorization_param)],
-        )
-
     request.context.js_config.enable_oauth2_redirect_error_mode(
-        auth_url=auth_url,
+        auth_route="canvas_api.oauth.authorize",
         error_details=request.params.get("error_description"),
         is_scope_invalid=request.params.get("error") == "invalid_scope",
         canvas_scopes=FILES_SCOPES + SECTIONS_SCOPES + GROUPS_SCOPES,
--- a/tests/unit/lms/resources/_js_config/__init___test.py
+++ b/tests/unit/lms/resources/_js_config/__init___test.py
@@ -567,46 +567,52 @@ class TestJSConfigRPCServer:
 class TestEnableOAuth2RedirectErrorMode:
     def test_scope_error(self, js_config):
         js_config.enable_oauth2_redirect_error_mode(
-            auth_url=None,
+            auth_route="auth_route",
             error_details="Technical error",
             is_scope_invalid=True,
             canvas_scopes=["scope_a", "scope_b"],
         )
 
         config = js_config.asdict()
-
         assert config["mode"] == "oauth2-redirect-error"
         assert config["OAuth2RedirectError"] == {
-            "authUrl": None,
+            "authUrl": "http://example.com/auth?authorization=xyz123",
             "invalidScope": True,
             "errorDetails": "Technical error",
             "canvasScopes": ["scope_a", "scope_b"],
         }
 
     def test_other_error(self, js_config):
-        auth_url = "https://lms.hypothes.is/auth/url"
         js_config.enable_oauth2_redirect_error_mode(
-            auth_url=auth_url, error_details="Some error"
+            auth_route="auth_route", error_details="Some error"
         )
 
         config = js_config.asdict()
 
         assert config["mode"] == "oauth2-redirect-error"
         assert config["OAuth2RedirectError"] == {
-            "authUrl": auth_url,
+            "authUrl": "http://example.com/auth?authorization=xyz123",
             "invalidScope": False,
             "errorDetails": "Some error",
             "canvasScopes": [],
         }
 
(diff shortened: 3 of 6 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -104,8 +104,8 @@ def asdict(self):
 
     def enable_oauth2_redirect_error_mode(
         self,
-        error_details,
-        auth_url=None,
+        auth_route,
+        error_details="",
```
seanh, line 108: I change this to take an auth _route_ and generate the URL itself. This allows the `route_url()` call and also the `BearerTokenSchema` call to live in this method and be re-used for both Blackboard and Canvas.

Also made `error_details` optional because Blackboard doesn't use it

lms/resources/_js_config/__init__.py
```diff
@@ -118,14 +118,29 @@ def enable_oauth2_redirect_error_mode(
...
+                auth_route,
+                _query=[
+                    (
+                        "authorization",
+                        BearerTokenSchema(self._request).authorization_param(
+                            self._lti_user
+                        ),
+                    )
+                ],
+            )
```
jon-betts, line 140:
```suggestion
            bearer_token = BearerTokenSchema(self._request).authorization_param(
                self._lti_user
            )
            auth_url = self._request.route_url(
                auth_route, _query=[("authorization", bearer_token)]
            )
```

I've noticed sometimes when black really spaces things out like this it can actually help make it more readable and shorter to pull a bit out. It's technically not a one liner any more, but if it gets really stringy like this I think it can help.

I've seen this help with some of the very nested `h_matchers` assertions as well.

lms/resources/_js_config/__init__.py
```diff
@@ -118,14 +118,29 @@ def enable_oauth2_redirect_error_mode(
 
         :param error_details: Technical details of the error
         :type error_details: str
-        :param auth_url: URL for the "Try again" button in the dialog
-        :type auth_url: str
+        :param auth_route: route for the "Try again" button in the dialog
+        :type auth_route: str
```
jon-betts, line 122:
```suggestion
        :param auth_route: route for the "Try again" button in the dialog
```

Are we not doing these?

lms/views/api/canvas/authorize.py
```diff
@@ -112,18 +112,8 @@ def oauth2_redirect(request):
...
-    auth_url = None
-    if request.lti_user:
-        authorization_param = BearerTokenSchema(request).authorization_param(
-            request.lti_user
-        )
-        auth_url = request.route_url(
-            "canvas_api.oauth.authorize",
-            _query=[("authorization", authorization_param)],
-        )
-
```
jon-betts, line 124: :+1: Good re-use here

lms/resources/_js_config/__init__.py
```diff
@@ -118,14 +118,29 @@ def enable_oauth2_redirect_error_mode(
...
         :param is_scope_invalid: `True` if authorization failed because the
           OAuth client does not have access to all the necessary scopes
         :type is_scope_invalid: bool
         :param canvas_scopes: List of scopes that were requested
         :type canvas_scopes: list(str)
```
jon-betts, line 127: Not sure which is the best way to go here, but I'm not sure there's inherently anything "Canvas"y about OAuth scopes (if that's what these are). So this could just be "required_scopes" or something.

On the other hand, if it's always Canvas why hide it? Maybe wait until there's something which isn't Canvas that needs this?

lms/resources/_js_config/__init__.py
jon-betts: I was wondering if the `error_details`, `is_scope_invalid` and `canvas_scopes` are all essentially Canvas specific. Seeing as the caller picks which template this is being rendered in, but looking at the template none of them are used in it, so it seems reasonable to control the values.

Otherwise I was going to suggest this could be `**template_kwargs` or something and they just get passed through.

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -567,57 +567,67 @@ def config(self, config, js_config):
...
+    @pytest.fixture(autouse=True)
+    def auth_route(self, pyramid_config):
+        pyramid_config.add_route("auth_route", "/auth")
+
 
 @pytest.fixture(autouse=True)
 def BearerTokenSchema(patch):
-    return patch("lms.resources._js_config.BearerTokenSchema")
+    BearerTokenSchema = patch("lms.resources._js_config.BearerTokenSchema")
+    BearerTokenSchema.return_value.authorization_param.return_value = "xyz123"
```
jon-betts, line 629:
```suggestion
    BearerTokenSchema.return_value.authorization_param.return_value = "Bearer: token_value"
```
?

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -567,57 +567,67 @@ def config(self, config, js_config):
 class TestEnableOAuth2RedirectErrorMode:
     def test_scope_error(self, js_config):
```
jon-betts, line 568: I know these were here before, but these two tests don't really seem to test much different from each other?

The first one could cover both by calling with `sentinel.error_details`, `sentinel.is_scope_invalid` etc. Maybe have the other to test the defaults?

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -567,57 +567,67 @@ def config(self, config, js_config):
...
+        js_config.enable_oauth2_redirect_error_mode(
+            auth_route="auth_route", error_details="Some error"
+        )
+
+        config = js_config.asdict()
+        assert config["oAuth2RedirectError"]["authUrl"] is None
 
     @pytest.fixture
-    def pyramid_request(self, pyramid_request):
+    def no_user(self, pyramid_request):
```
jon-betts, line 608:
```suggestion
    def with_no_user(self, pyramid_request):
```
Seems to be a pattern we have elsewhere?

jon-betts (commented):
I've not tested this yet, but this is just a code once over. There are lots of cosmetic suggestions, but this would be an approve. This is in case you want go work through those while I do the testing.

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -567,57 +567,67 @@ def config(self, config, js_config):
...
         }
 
-    @pytest.fixture
-    def js_config(self, context, pyramid_request):
-        return JSConfig(context, pyramid_request)
+    @pytest.mark.usefixtures("no_user")
+    def test_auth_url(self, js_config):
+        js_config.enable_oauth2_redirect_error_mode(
+            auth_route="auth_route", error_details="Some error"
+        )
```
jon-betts, line 602:
```suggestion
        js_config.enable_oauth2_redirect_error_mode(auth_route="auth_route")
```

Now this value has a default, I don't _think_ it has any special meaning in this test.

jon-betts (approved):
Yep just tested this now. All works like a charm.

lms/resources/_js_config/__init__.py
seanh: :+1: Done

lms/resources/_js_config/__init__.py
seanh: They were just pre-existing but in this case I think they are actually kind of useful because the method accepts two strings, one bool and one list of strings

lms/resources/_js_config/__init__.py
seanh: In practice the only thing we use OAuth scopes for is for Canvas and what we do with them is Canvas-specific. For example they trigger a Canvas-specific error message about missing developer key scopes. I think if we one day had Blackboard scopes as well we might need a separate `blackboard_scopes` argument. Who knows

tests/unit/lms/resources/_js_config/__init___test.py
seanh: :+1:

tests/unit/lms/resources/_js_config/__init___test.py
seanh: Parametrized this

tests/unit/lms/resources/_js_config/__init___test.py
seanh: :+1: Done

PR #3000, opened 2021-07-22
Generalize `enable_canvas_oauth2_redirect_error_mode()`

seanh:
This is mostly not Canvas-specific (except for the scopes) and we now need to re-use it for Blackboard, so generalize  `enable_canvas_oauth2_redirect_error_mode()` to just  `enable_oauth2_redirect_error_mode()` and remove all unnecessary references to Canvas.

### Testing

* Delete all the access tokens from your DB so that the authorization flow is triggered when you launch an assignment:

  ```
  ```

* Log in to https://hypothesis.instructure.com/ as either a teacher or a student and launch [localhost (make devdata) Canvas Files Assignment](https://hypothesis.instructure.com/courses/125/assignments/875)

* Click **Authorize**

- [ ] In the popup window click **Cancel** instead of **Authorize**. You should see an error dialog with a **Try again** button that takes you back to Canvas's authorization page

- [ ] Hack the code so that the server-to-server request that happens after you click **Authorize** in the popup window fails:

  ```diff
  diff --git a/lms/services/canvas_api/client.py b/lms/services/canvas_api/client.py
  index 0a47abf3..e000d51a 100644
  --- a/lms/services/canvas_api/client.py
  +++ b/lms/services/canvas_api/client.py
  @@ -59,7 +59,7 @@ class CanvasAPIClient:
               to exchange for an access token
           :return: An access token string
           """
  -        return self._client.get_token(authorization_code)
  +        return self._client.get_token("foo")
   
       # Getting authenticated users sections
       # ------------------------------------
  ```

  Open the popup window again and click **Authorize**. You should get a nice error dialog again.

- [ ] Un-hack the code, re-open the popup window, click **Authorize**, and the assignment should launch successfully

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -129,11 +129,11 @@ class JSConfig:
         self._config.update(
             {
                 "mode": "oauth2-redirect-error",
-                "oAuth2RedirectError": {
+                "OAuth2RedirectError": {
                     "authUrl": auth_url,
                     "invalidScope": is_scope_invalid,
                     "errorDetails": error_details,
-                    "canvas_scopes": canvas_scopes or [],
+                    "canvasScopes": canvas_scopes or [],
                 },
             }
         )
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -102,18 +102,19 @@ class JSConfig:
         """
         return self._config
 
-    def enable_canvas_oauth2_redirect_error_mode(
+    def enable_oauth2_redirect_error_mode(
         self,
         error_details,
         auth_url=None,
         is_scope_invalid=False,
-        requested_scopes=None,
+        canvas_scopes=None,
     ):
         """
-        Configure the frontend to show the "Canvas authorization failed" dialog.
+        Configure the frontend to show the "Authorization failed" dialog.
 
-        This is shown when authorizing with the Canvas API fails, following the
-        redirect to Canvas's OAuth 2 authorization endpoint.
+        This is shown when authorizing with a third-party OAuth API like the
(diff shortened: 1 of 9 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -102,18 +102,19 @@ def asdict(self):
...
         return self._config
 
-    def enable_canvas_oauth2_redirect_error_mode(
+    def enable_oauth2_redirect_error_mode(
         self,
         error_details,
         auth_url=None,
         is_scope_invalid=False,
-        requested_scopes=None,
+        canvas_scopes=None,
```
seanh, line 110: Scopes are a general OAuth2 thing but I think as far as this code goes "Canvas scopes" and "Blackboard scopes" might be two different things: the presence of `canvas_scopes` here triggers some very Canvas-specific text explaining that your "developer key" (a Canvas term) is missing required scopes and linking to Canvas-specific docs for fixing it. If we had Blackboard scopes or any other kind of scopes it's unlikely we'd want to trigger the same text. (As it happens we don't yet have any other types of OAuth scopes anyway, only Canvas ones.)

lyzadanger (changes requested):
The test steps worked as documented in the PR description! So, yay. The changes I am requesting here are pedantic and should be quick work for you.

There are a few references here that don't use our current casing conventions, which is camelCase with any leading capitals or acronyms left capitalized. I've added a few comments inline, but the references I noticed (with a name that would match our current conventions) were:

* `canvas_scopes` => `canvasScopes`
* `oAuth*` => `OAuth*`

Tests pass, so that's rad.

I spotted several pre-existing references within the same modules you were modifying that also don't meet casing conventions. Um, do as I say not as I do? :)

lyzadanger (approved):

PR #3008, opened 2021-07-27
Move vitalsource.py to directory to allow other modules in the namespace

marcospri:
Just moved it to allow for example `_schemas.py` inside `vitalsource/`


tests/unit/lms/services/vitalsource_test.py
```diff
@@ -78,7 +78,7 @@ def lti_user(self):
```
seanh, line 78: Don't you also need to move the test file to `tests/unit/lms/services/vitalsource/client_test.py`? And you'll need to add `tests/unit/lms/services/vitalsource/__init__.py` too

seanh (changes requested):

tests/unit/lms/services/vitalsource_test.py
marcospri: You are right, done :+1:

seanh (approved):

PR #3009, opened 2021-07-27
VitalSource API

marcospri:
# Testing notes

- Enable the VS flag on http://localhost:8001/flags
- Get the values from https://github.com/hypothesis/devdata/pull/28 (or `make devdata` once that's merged)
- Configure a vital-source assignment 
- Use `https://bookshelf.vitalsource.com/#/books/9781400847402` on the modal
- You should get the cover image and title and the ToC once you select the book

- Use a different URL with the same format  `https://bookshelf.vitalsource.com/#/books/9781400840000`
- You should get a `Book 9781400840000 not found` message in the UI


lms/services/vitalsource/client.py
```diff
@@ -13,11 +16,44 @@ def __init__(self, lti_launch_key, lti_launch_secret):
...
+            raise ValueError("VitalSource credentials are invalid")
+
+        self._http_service = http_service
 
         self._lti_launch_key = lti_launch_key
         self._lti_launch_secret = lti_launch_secret
+        self._api_key = api_key
+
+
+    def _api_get(self, endpoint):
```
marcospri, line 29: This seems to be all we need as an "API Client"

lms/services/vitalsource/client.py
```diff
@@ -13,11 +16,44 @@ def __init__(self, lti_launch_key, lti_launch_secret):
...
+        return self._http_service.get(
+            url, headers={"X-VitalSource-API-Key": self._api_key}
+        )
+
+    def book_info(self, book_id: str):
+        try:
+            response = self._api_get(f"products/{book_id}")
+        except lms.services.exceptions.HTTPError as exp:
+            if exp.response.status_code == 404:
+                return None
```
marcospri, line 41: The book not existing is something the frontend deals with. I think is nicer for the view to check for None than to handle the exception there.

lms/views/api/vitalsource.py
```diff
@@ -1,32 +1,37 @@
...
-        # In future this will fetch data from VitalSource.
-        if book_id not in book_data:
-            raise ProxyAPIError(f"Book {book_id} not found")
+        book_info = self.vitalsource_service.book_info(book_id)
 
-        return book_data[book_id]
+        if not book_info:
+            raise HTTPNotFound()
+
+        return {
```
marcospri, line 23: I think this is what we settled for on BB, the schema represents more or less the third party responses, we adapt it to our client on the view.

lms/services/vitalsource/client.py
seanh: :+1:

seanh (approved):

lms/services/vitalsource/client.py
```diff
@@ -13,11 +16,43 @@ def __init__(self, lti_launch_key, lti_launch_secret):
         :type lti_launch_secret: str
         :raises ValueError: If credentials are invalid
         """
-        if not lti_launch_key or not lti_launch_secret:
-            raise ValueError("VitalSource LTI launch credentials are invalid")
+        if not all([lti_launch_key, lti_launch_secret, api_key]):
+            raise ValueError("VitalSource credentials are invalid")
```
seanh, line 20: I guess they're "missing" rather than invalid:

```suggestion
        if not all([lti_launch_key, lti_launch_secret, api_key]):
            raise ValueError("VitalSource credentials are missing")
```

We normally wouldn't bother with this to be honest. I think the code just assumes that required settings are present and crashes if they aren't. It's usually obvious enough from the traceback what's wrong. But it's fine to do this as well

lms/services/vitalsource/client.py
seanh: :man_shrugging: That's fine with me. I think we could just raise the `ProxyAPIError` right here and the view wouldn't need to do anything (it wouldn't need to catch it) which would lead to less and simpler code overall. But it's also just fine your way

lms/views/api/vitalsource.py
seanh: Yep, this is the same as what Bb does

tests/unit/lms/services/vitalsource/client_test.py
```diff
@@ -68,9 +71,67 @@ def test_it_signs_lti_launch_form_params(self, svc, lti_user):
             "oauth_version": "1.0",
         }
 
+    def test_api_get(self, svc, http_service):
+        svc._api_get("endpoint/path")  # pylint: disable=protected-access
```
seanh, line 75: I guess you could move the `_api_get()` method into a separate helper class. The helper class could just be inline within the `vitalsource.py` file and could have its own unit tests. Then you could avoid the `protected-access` here.

Or alternatively you could just make `_api_get()` a public method even though nothing outside of `VitalSourceService` actually calls it. In theory it might be useful for a view to call `VitalSourceService.get()` directly one day, so that you don't _have_ to write a new `VitalSourceService` method for every single API call if you don't really need to.

I think the `disable=protected-access` is also a perfectly reasonable solution here though.

tests/unit/lms/services/vitalsource/client_test.py
```diff
@@ -68,9 +71,67 @@ def test_it_signs_lti_launch_form_params(self, svc, lti_user):
...
+
+        http_service.get.assert_called_once_with(
+            "https://api.vitalsource.com/v4/endpoint/path",
+            headers={"X-VitalSource-API-Key": "api_key"},
+        )
+
+    def test_book_info_api(self, svc, book_info_schema):
+        with mock.patch.object(VitalSourceService, "_api_get") as api_get:
```
seanh, line 83: If `_api_get` were moved into a separate helper class then you could patch it with a patch fixture which I think is a bit nicer than using `patch.object()`. Not a big deal though

tests/unit/lms/services/vitalsource/client_test.py
```diff
@@ -110,6 +195,8 @@ def pyramid_config(self, pyramid_config):
         pyramid_config.registry.settings[
             "vitalsource_lti_launch_secret"
         ] = sentinel.vs_lti_launch_secret
+        pyramid_config.registry.settings["vitalsource_api_key"] = sentinel.vs_api_key
```
seanh, line 198: Might just want to put this in the global test settings to that it's always present: https://github.com/hypothesis/lms/blob/98e70824fce1a39e6906002ad7184e8efe20dff8/tests/conftest.py#L16-L45

seanh (approved):
This looks good to me but I haven't tested it. I think it might be a good idea to get @robertknight to test this as he's familiar with VitalSource

lms/views/api/vitalsource.py
```diff
@@ -2,31 +2,35 @@
...
+        book_info = self.vitalsource_service.book_info(book_id)
+
+        if not book_info:
             raise ProxyAPIError(f"Book {book_id} not found")
 
-        return book_data[book_id]
+        return {
+            "id": book_info["vbid"],
+            "title": book_info["title"],
+            "cover_image": book_info["resource_links"]["cover_image"],
```
robertknight, line 25: I'm not sure if a cover image is guaranteed to exist, but we'll get a Sentry error if it turns out that it doesn't.

lms/services/vitalsource/client.py
```diff
@@ -13,11 +16,43 @@ def __init__(self, lti_launch_key, lti_launch_secret):
...
+    def _api_get(self, endpoint):
+        url = f"https://api.vitalsource.com/v4/{endpoint}"
+
+        return self._http_service.get(
+            url, headers={"X-VitalSource-API-Key": self._api_key}
+        )
+
+    def book_info(self, book_id: str):
+        try:
+            response = self._api_get(f"products/{book_id}")
```
robertknight, line 37: Are you validating the format of a book ID anywhere? If not the `book_id` could potentially include characters that require encoding. We do validate it against a regex on the frontend (which IIRC allows only upper-case ASCII letters and dashes) but you shouldn't trust that on the backend.

tests/unit/lms/services/vitalsource/client_test.py
marcospri: Moved there :+1:

tests/unit/lms/services/vitalsource/client_test.py
marcospri: I went for the public method. Using `.get` on the view is something we might need :+1:

lms/services/vitalsource/client.py
marcospri: I somehow thought `ProxyAPIError` was part of the `views` package, I'll raise from the service then

lms/services/vitalsource/client.py
marcospri: Now the book_id format is checked against a regexp on the view @robertknight

lms/views/api/vitalsource.py
```diff
@@ -1,32 +1,39 @@
...
 class VitalSourceAPIViews:
     def __init__(self, request):
         self.request = request
+        self.vitalsource_service = request.find_service(name="vitalsource")
 
     @view_config(route_name="vitalsource_api.books.info")
     def book_info(self):
         book_id = self.request.matchdict["book_id"]
+        if not self._is_valid_book_id(book_id):
+            raise HTTPClientError("Invalid `book_id`. It must only contain [0-9A-Z-].")
```
robertknight, line 19: It looks like this error is resulting in the caller receiving an HTML rather than JSON error response:

```html
<!doctype html>
<html lang=en>
  <head>
    ...
  </head>
  <body>
    <div class="content">
      
  <main class="modal-content u-horizontally-center-children">
    <img src="/static/images/sad-annotation.svg">
    <p>
      Invalid `book_id`. It must only contain [0-9A-Z-].
    </p>
  </main>

    </div>

    
      
      
    
  </body>
</html>
```

lms/views/api/vitalsource.py
```diff
@@ -1,32 +1,39 @@
...
+        return {
+            "id": book_info["vbid"],
+            "title": book_info["title"],
+            "cover_image": book_info["resource_links"]["cover_image"],
+        }
 
     @view_config(route_name="vitalsource_api.books.toc")
     def table_of_contents(self):
         book_id = self.request.matchdict["book_id"]
+        if not self._is_valid_book_id(book_id):
```
robertknight, line 31: Perhaps worth extracting the logic for extract the `book_id` parameter, validating it and throwing an error if invalid into a single helper, to ensure consistency across different endpoints.

robertknight (changes requested):
I found an issue where invalid book IDs cause the client to receive an HTML rather than JSON error response, which the frontend is not able to parse. Otherwise everything works as expected.

lms/views/api/vitalsource.py
marcospri: My bad, switched to json.

Here's is how it looks on the UI if that were to happen (which would only occour if the validation on the backend and frontned diverged)
Screenshot from 2021-07-30 11-46-34

I'd say it's an ok message if we for example got it in a bug report.

lms/views/api/vitalsource.py
marcospri: Yes, good point, added a helper function to do the extracting (and posible raising) but kept the second function that just does the regexp.

robertknight (approved):
Functionally this now works as desired.

`ProxyAPIError` is probably not strictly the right exception class to use for validating the `book_id` parameter. Looking at this again reminded me there is a whole set of machinery that is used for validating Pyramid requests in the form of `PyramidRequestSchema` and the `schema` parameter for `@view_config`. However I don't think that supports validating path as opposed to query string or body parameters. 

I suggest that we get this merged for now. We can revisit whether this is the preferred way to validate path parameters another time.

lms/views/api/vitalsource.py
```diff
@@ -22,7 +25,18 @@ def book_info(self):
...
-        book_id = self.request.matchdict["book_id"]
+        book_id = self._get_book_id()
 
         book_toc = self.vitalsource_service.book_toc(book_id)
         return book_toc["table_of_contents"]
+
+    def _get_book_id(self):
+        book_id = self.request.matchdict["book_id"]
+        if not self._is_valid_book_id(book_id):
+            raise ProxyAPIError("Invalid `book_id`. It must only contain [0-9A-Z-].")
```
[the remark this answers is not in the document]
marcospri: Yeah I tried the marshmallow route but won't work with URL params.

PR #3017, opened 2021-07-29
Raise `ExternalRequestError` instead of `HTTPError`

seanh:
Raise `ExternalRequestError` from `HTTPService` instead of `HTTPError`.

This fixes the reporting of failed HTTP requests by `HTTPService` to Sentry (adding details of the error response to Sentry) and removes an unnecessary class (we don't need both `HTTPError` and `ExternalRequestError`).

When `HTTPService` was written (quite recently) a new `HTTPError` class was added for it to raise:

```python
class HTTPError(ServiceError):
    """A problem with an HTTP request to an external service."""

        super().__init__(response)
        self.response = response
```

I think this was probably a mistake. There was already an existing `ExternalRequestError` class in the code and that class contains code for stringifying the exception nicely for Sentry:

```python
class ExternalRequestError(ServiceError):
        super().__init__()
        self.response = response
        self.details = details

    def __str__(self):
        if self.response is None:
            return self.explanation

        # Log the details of the response. This goes to both Sentry and the
        # application's logs. It's helpful for debugging to know how the
        # external service responded.
        parts = [
            self.explanation + ":",
            str(self.response.status_code or ""),
            self.response.reason,
            self.response.text,
        ]
        return " ".join([part for part in parts if part])
```

`HTTPService` should have used `ExternalRequestError` all along to get this `__str__()` and to avoid adding a second exception class for the same thing.

The name `HTTPError` is kind of nicer than `ExternalRequestError` (it's shorter for one thing) but I think it's probably best to stick with `ExternalRequestError` to avoid confusion with `pyramid.httpexceptions.HTTPError` or requests.HTTPError`.

### Testing

TODO. `HTTPError` was used by `BlackboardAPIClient`, `HAPI` (client), and `LTIOutcomesClient` (the grading feature in non-Canvas LMS's) so all three need to be tested. We need to test what's shown in the frontend's error dialog when an exception happens and also what gets reported to Sentry.

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -1,6 +1,6 @@
 from marshmallow import INCLUDE, fields
 
-from lms.services.exceptions import HTTPError, OAuth2TokenError
+from lms.services.exceptions import ExternalRequestError, OAuth2TokenError
 from lms.validation import RequestsResponseSchema, ValidationError
 
 
@@ -75,7 +75,7 @@ class BasicClient:
 
         try:
             return self._send(method, url)
-        except (OAuth2TokenError, HTTPError):
+        except ExternalRequestError:
             self._oauth_http_service.refresh_access_token(
                 self.token_url,
                 self.redirect_uri,
@@ -100,7 +100,7 @@ class BasicClient:
     def _send(self, method, url):
         try:
             return self._oauth_http_service.request(method, url)
-        except HTTPError as err:
+        except ExternalRequestError as err:
             error_dict = BlackboardErrorResponseSchema(err.response).parse()
 
             if error_dict.get("message") == "Bearer token is invalid":
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -1,7 +1,7 @@
 import requests
 from requests import RequestException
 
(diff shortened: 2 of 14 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -75,7 +75,7 @@ def request(self, method, path):
 
         try:
             return self._send(method, url)
-        except (OAuth2TokenError, HTTPError):
+        except ExternalRequestError:
```
seanh, line 78: `OAuth2TokenError` is an `ExternalRequestError` subclass so there's no need to catch both here. (It might be a little weird for `OAuth2TokenError` to subclass `ExternalRequestError` and maybe it shouldn't but that's out of scope for this PR.

lms/services/http.py
```diff
@@ -85,7 +87,7 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
             )
             response.raise_for_status()
         except RequestException as err:
-            raise HTTPError(response) from err
+            raise ExternalRequestError(response=response) from err
```
seanh, line 90: `ExternalRequestError` accepts multiple kwargs of which `response` is not the first so `(response=response)` is needed rather than just `(response)`

PR #3034, opened 2021-08-03
Request additional fields from BB files API

marcospri:
In preparation to store the files in the DB

More a question than a real PR, but what fields should we request from the API (and which one later store on the DB) for an hypothetical BB course copy.

Available fields on the API are:

```
"id": "string",
"name": "string",
"type": "File",
"size": 0,
"hasLinks": true,
"parentId": "string",
"creatorId": "string",
"created": "2021-08-03T11:26:34.413Z",
"modified": "2021-08-03T11:26:34.413Z",
"mimeType": "string",
"version": 0,
"downloadUrl": "string"
```

should we just do all of them?

### DIFF
--- a/lms/services/blackboard_api/_schemas.py
+++ b/lms/services/blackboard_api/_schemas.py
@@ -22,6 +22,8 @@ class BlackboardListFilesSchema(RequestsResponseSchema):
         modified = fields.Str(required=True)
         type = fields.Str(required=True)
         mimeType = fields.Str()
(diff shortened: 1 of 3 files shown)


jon-betts:
I'd only pull back the fields we have places to store. Otherwise it will bloat the call and slow it down. It might not make much of a difference, but there doesn't seem like a good reason to ask for, or represent, things we don't want.

Issue #3012, opened 2021-07-28
Store Blackboard files in the DB


seanh:
When teachers create a Blackboard Files assignment and we get a list of files from Blackboard we should store the metadata about those files in our DB like we do for Canvas files. This metadata is likely going to be needed in future to fix course copy for Blackboard Files assignments.

Note that we currently [limit what fields we get back from the Blackboard API](https://github.com/hypothesis/lms/blob/78158af41cf19a0c6b1e7d839aa3711766d00eaf/lms/services/blackboard_api/client.py#L40) so save bandwidth. We might have to expand that list of fields to contain any and all fields that we think might be useful or interesting. For example in Canvas we use the filename and size fields to search for matching files after a course copy. See [the Blackboard API docs](https://developer.blackboard.com/portal/displayApi) for the full list of available fields.

PR #3044, opened 2021-08-11
Migration for File.parent_lms_id

marcospri:
For https://github.com/hypothesis/lms/issues/3012

Adds nullable column for tree like relationships.

Autogenerated from https://github.com/hypothesis/lms/pull/3045


jon-betts (approved):
It makes the column, it makes it go. :+1:

Issue #2024, opened 2020-08-19
Handle Blackboard instances that don't have our Application ID


seanh:
A Blackboard instance admin has to add our Blackboard API **Application ID** to the Blackboard instance's **Rest API Integrations** before the Blackboard Files feature can work for that instance, as documented here: https://github.com/hypothesis/lms/wiki/How-to-set-up-Blackboard-API-access-for-the-LMS-app#must-be-done-by-each-blackboard-sites-admin

This has already been done for our localhost, QA and prod rest API integrations on our Blackboard test site (https://aunltd-test.blackboard.com/)

When trying to use the Blackboard Files feature fails because the rest API integration hasn't been created in the Blackboard site we should detect this situation and show a helpful message linking to docs that explain how to get our Application ID and how to add the rest API integration.

To test this you may need to delete the rest API integration from https://aunltd-test.blackboard.com/ to see what happens (e.g. what error response the Blackboard API sends), test it, and then re-add the API integration.

PR #3047, opened 2021-08-11
Bb missing integration dialog

marcospri:
For #2024 

# Testing

- Log in as admin on BB
- https://aunltd-test.blackboard.com/ultra/admin -> Integrations -> Rest API Integrations and here either disable (`Make unavaialble") Hypothesis Dev or remove it.
- Launch an assignment that uses BB API (eg `localhost (make devdata) Blackboard Files Assignment` or/and try to create a new BB files assignment.
- You should get an "authorize" modal followed by another modal explaining the situation
Screenshot from 2021-08-16 15-44-24

- After testing re-enable the Rest integration (or recreate it using application id: `b7a1baf5-3545-4f5d-897f-c036b0d5bd5b` and learn user `administrator`

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -1,4 +1,5 @@
 import functools
+from typing import List
 
 from lms.models import GroupInfo, HUser
 from lms.resources._js_config.file_picker_config import FilePickerConfig
@@ -102,12 +103,13 @@ class JSConfig:
         """
         return self._config
 
-    def enable_oauth2_redirect_error_mode(
+    def enable_oauth2_redirect_error_mode(  # pylint: disable=too-many-arguments
         self,
-        auth_route,
-        error_details="",
-        is_scope_invalid=False,
-        canvas_scopes=None,
+        auth_route: str,
+        error_code=None,
+        error_details: str = "",
+        is_scope_invalid: bool = False,
+        canvas_scopes: List[str] = None,
     ):
         """
         Configure the frontend to show the "Authorization failed" dialog.
@@ -116,15 +118,12 @@ class JSConfig:
         Canvas API or the Blackboard API fails after the redirect to the
         third-party authorization endpoint.
 
+        :param error_code: Code identifying a particular error
         :param error_details: Technical details of the error
-        :type error_details: str
         :param auth_route: route for the "Try again" button in the dialog
-        :type auth_route: str
         :param is_scope_invalid: `True` if authorization failed because the
           OAuth client does not have access to all the necessary scopes
-        :type is_scope_invalid: bool
         :param canvas_scopes: List of scopes that were requested
-        :type canvas_scopes: list(str)
         """
         if self._lti_user:
             bearer_token = BearerTokenSchema(self._request).authorization_param(
@@ -143,6 +142,7 @@ class JSConfig:
                 "OAuth2RedirectError": {
                     "authUrl": auth_url,
                     "invalidScope": is_scope_invalid,
+                    "errorCode": error_code,
                     "errorDetails": error_details,
                     "canvasScopes": canvas_scopes or [],
                 },
--- a/lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
+++ b/lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
@@ -23,20 +23,32 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
     OAuth2RedirectError: {
       authUrl = /** @type {string|null} */ (null),
       invalidScope = false,
+      errorCode = /** @type {string|null} */ (null),
       errorDetails = '',
       canvasScopes = /** @type {string[]} */ ([]),
     },
   } = useContext(Config);
 
-  const title = invalidScope
-    ? 'Developer key scopes missing'
-    : 'Authorization failed';
+  const error = { code: errorCode, details: errorDetails };
 
-  const message = invalidScope
-    ? null
-    : 'Something went wrong when authorizing Hypothesis';
+  const title = (() => {
+    if (invalidScope) {
+      return 'Developer key scopes missing';
+    }
 
-  const error = { details: errorDetails };
+    if (errorCode === 'blackboard_missing_integration') {
+      return 'Missing Blackboard REST API integration';
+    }
+
+    return 'Authorization failed';
+  })();
+
+  const message = (() => {
+    if (invalidScope) {
+      return null;
+    }
+    return 'Something went wrong when authorizing Hypothesis';
+  })();
 
   const retry = () => {
     location.href = /** @type {string} */ (authUrl);
@@ -88,6 +100,26 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
               href="https://github.com/hypothesis/lms/wiki/Canvas-API-Endpoints-Used-by-the-Hypothesis-LMS-App"
             >
               Canvas API Endpoints Used by the Hypothesis LMS App
+            </a>
+            .
+          </p>
+        </Fragment>
+      )}
+
+      {error.code === 'blackboard_missing_integration' && (
+        <Fragment>
+          <p>
+            In order to allow Hypothesis to connect to files in Blackboard, your
+            Blackboard admin needs to add or enable a REST API integration.
+          </p>
+          <p>
+            For more information, please have your Blackboard admin read:{' '}
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href="https://web.hypothes.is/help/enable-the-hypothesis-integration-with-blackboard-files/"
+            >
+              Enable the Hypothesis Integration With Blackboard Files
             </a>
             .
           </p>
--- a/lms/static/scripts/frontend_apps/config.js
+++ b/lms/static/scripts/frontend_apps/config.js
@@ -77,6 +77,10 @@ import { createContext } from 'preact';
  */
 
 /**
+ * @typedef {'blackboard_missing_integration'} OAuthErrorCode
+ */
+
+/**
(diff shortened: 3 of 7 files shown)


lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -23,20 +23,32 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
     OAuth2RedirectError: {
       authUrl = /** @type {string|null} */ (null),
       invalidScope = false,
+      errorCode = '',
```
marcospri, line 26: Adding a new errorCode attribute here instead of a new "missingBBIntegration" or similar. I think we'll have an big proliferation of booleans here otherwise. If you like this approach I'll follow up with another PR to refactor "invalidScope" as an errorCode.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -23,20 +23,32 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
...
 
-  const error = { details: errorDetails };
+    return 'Authorization failed';
+  };
+
+  const error = { code: errorCode, details: errorDetails };
+
+  const message = () => {
+    if (invalidScope || errorCode !== null) {
+      return ' ';
```
marcospri, line 48: This is odd and leads to a random "." to show up in the dialog. Using `null` as before leads to the message `An unknown error occurred.` to show up which is not enterily correct (it's a known error)

This seem to come from here:https://github.com/hypothesis/lms/blob/4c5bad17e096a7e9ea949cbab667efc1856ed63a/lms/static/scripts/frontend_apps/components/ErrorDisplay.js#L109

and I haven't found a way to just not show anything. I could below use a conditional and only use `<ErrorDisplay message={message()} error={error} />`in the `else` part but `ErrorDisplay` is responsible of showing the "contact support" part as well. Maybe a new bool attribute for `ErrorDisplay` to disable this message?

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -93,7 +105,26 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
...
+              rel="noopener noreferrer"
+              href="https://github.com/hypothesis/lms/wiki/How-to-set-up-Blackboard-API-access-for-the-LMS-app#create-a-rest-api-integration-in-the-blackboard-site"
+            >
+              Create a REST API Integration in the Blackboard site
+            </a>
+            .
+          </p>
+        </Fragment>
+      )}
+      <ErrorDisplay message={message()} error={error} />
```
marcospri, line 127: This shows both the concat support and a (I think) superfluous other message which here I temporarly replaced with "."

Screenshot from 2021-08-11 15-53-45

lms/static/scripts/frontend_apps/config.js
```diff
@@ -76,6 +76,10 @@ import { createContext } from 'preact';
  *   @prop {boolean} vitalSource.enabled
  */
 
+/**
+ * @typedef {'blackboard_missing_integration'} OAuthErrorCode
```
marcospri, line 80: No idea if this is useful at all, maybe TS will provide exhaustiveness check if used in a case statement?

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
robertknight: For the purposes of this PR, I suggest we just make this Blackboard error's display match the existing one for Canvas, which currently has a "Something went wrong authorizing Hypothesis" message.

If we wanted to remove the message in future we could make it so that `<ErrorDisplay message={null} .../>` is treated differently from `<ErrorDisplay .../>` without a `message` prop. I forget where, but we do have one other place that uses this pattern - where you set the prop to a value to use, set it to `null` to  not use the feature or omit it to get a default value.

lms/static/scripts/frontend_apps/config.js
robertknight: I don't think TS currently does exhaustiveness checks, but it will complain if you try to test it against a value that it can't have (for example if you made a typo and tried to compare an `OAuthErrorCode` value to `blackbard_missing_integration`). It's also just useful documentation for readers.

robertknight (commented):
I had a look over this. The approach makes sense to me. I suggest following the pattern of the existing Canvas error dialog initially regarding the `message` prop for `ErrorDisplay`. It might make sense to remove this message for both the Canvas and Blackboard cases, in which case I made a suggestion about how that could work.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -93,7 +105,26 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
...
+        <Fragment>
+          <p>
+            A blackboard admin needs to add or enable a REST API integration.
+          </p>
+          <p>
+            For more information see:{' '}
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href="https://github.com/hypothesis/lms/wiki/How-to-set-up-Blackboard-API-access-for-the-LMS-app#create-a-rest-api-integration-in-the-blackboard-site"
```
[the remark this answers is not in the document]
marcospri: Pinged #support to take a look at the modal https://hypothes-is.slack.com/archives/C2BLQDKHA/p1629109244219800

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
jon-betts: :+1: This is _definitely_ better than a boolean per possible thing that could go wrong.

jon-betts (approved):
Yep this works a treat. Having a variable code is much more scalable than specific flags for every eventuality as well.

I've tested this and vetted the Python portion, but you probably want to wait for Rob's +1 on the front-end side, where I don't really have an opinion.

lms/resources/_js_config/__init__.py
```diff
@@ -102,12 +103,13 @@ def asdict(self):
...
         self,
-        auth_route,
-        error_details="",
-        is_scope_invalid=False,
-        canvas_scopes=None,
+        auth_route: str,
+        error_code=None,
+        error_details: str = "",
+        is_scope_invalid: bool = False,
+        canvas_scopes: List[str] = None,
```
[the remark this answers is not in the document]
marcospri: I mostly added the type annotations here because I felt guilt just removing them from the docstring.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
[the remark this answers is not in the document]
marcospri: Updated modal's copy and link :+1:

lms/resources/_js_config/__init__.py
```diff
@@ -143,6 +142,7 @@ def enable_oauth2_redirect_error_mode(
                 "OAuth2RedirectError": {
                     "authUrl": auth_url,
                     "invalidScope": is_scope_invalid,
+                    "errorCode": error_code,
```
robertknight, line 145: I agree that folding `invalidScope` into a new `errorCode` value in a follow-up PR would be a good thing to do.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
robertknight, line 26: The default value of `errorCode` should be strictly `null` according to the definition of the `errorCode` field, and that's what the backend actually uses as the default value. Additionally if you know the backend is always going to set a particular field, as is the case here, you don't need to specify a default.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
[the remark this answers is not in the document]
robertknight: Did you update the link? The one on the branch still points to GitHub.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -84,6 +88,7 @@ import { createContext } from 'preact';
  * @prop {string|null} authUrl
  * @prop {boolean} invalidScope
  * @prop {string} errorDetails
+ * @prop {OAuthErrorCode|null} errorCode
```
robertknight, line 91: As noted above, if the default is `null` then the initialization for `errorCode` in `OAuth2RedirectErrorApp` should either be `null` or just omit it if we know the backend is always going to set it.

robertknight (approved):
Everything worked as expected in manual testing. I noticed two issues:

- Can you confirm that the help link is correct? The current link in the code is https://github.com/hypothesis/lms/wiki/How-to-set-up-Blackboard-API-access-for-the-LMS-app#create-a-rest-api-integration-in-the-blackboard-site but I thought it was going to be from the knowledge base.
- The default value for `errorCode` should either be `null` (to match the backend) or omit it if you know we can rely on the backend to set it

Can I leave it to you to address these and get this merged.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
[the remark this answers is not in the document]
marcospri: My bad, made the commit, take the screenshot but didn't push it.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
marcospri: Well spotted, updated without a default

PR #3048, opened 2021-08-12
Explictly test that canvas.list_file emits an event

marcospri:


tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -383,6 +385,45 @@ def test_list_duplicate_files(self, canvas_api_client, http_session):
...
+                "updated_at": "updated_at_1",
+                "size": 12345,
+            },
+            {
+                "display_name": "display_name_1",
+                "id": 2,
+                "updated_at": "updated_at_1",
+                "size": 12345,
+            },
+        ]
```
jon-betts, line 404:
```suggestion
        files = [
            {
                "id": i,
                "display_name": "display_name_{i}",
                "updated_at": "updated_at_{i}",
                "size": 1000+i,
            } for i in range(2)
        ]
```

It's probably a good idea to vary the things we are asserting against here.

jon-betts (changes requested):
I'm not sure that we want to be asserting a sub-set of the dict here. All elements of it seem important to me?

tests/unit/lms/services/canvas_api/client_test.py
```diff
@@ -398,6 +434,10 @@ def test_public_url(self, canvas_api_client, http_session):
             timeout=Any(),
         )
 
+    @pytest.fixture
+    def registry(self):
+        return create_autospec(Registry, instance=True, spec_set=True)
```
jon-betts, line 439: Another way to go here is to use:

```python
with patch.object(pyramid_request, "registry") as registry:
    yield registry
```

All the same really, but that has the advantage of not having to work out what the class is. This is very clear though (as long as the class is right!).

jon-betts (approved):

Issue #2679, opened 2021-05-12
Show an error message if an ApplicationInstance is launched with a different tool_consumer_instance_guid


seanh:
### Problem

Problems can happen if the same `ApplicationInstance` is installed in multiple different LMS instances, or if the `tool_consumer_instance_guid` of a single LMS instance changes (see https://github.com/hypothesis/support/issues/185). For example course IDs aren't globally unique across LMS's, they're only unique within a single LMS instance. Two different instances of Canvas for example can both have courses with the same ID. So if the unique key on the `Course` table is `(application_instance_id, course_id)` you could get collisions if that `ApplicationInstance` is used in two different instances of Canvas that both have courses with the same ID. (This is not just theoretical either. Canvas course IDs are incrementing integers like `1`, `2`, `3`, etc. So collisions will happen right away.)

This applies to everything: courses, assignments, users, etc.

The existing `Course`, `GradingInfo` and `OAuth2Token` tables in the LMS DB don't include the `tool_consumer_instance_guid` in a unique constraint, they only include the `ApplicationInstance.id` (or the `ApplicationInstance.consumer_key`, an alternative ID, sometimes called `oauth_consumer_key` instead). So they're vulnerable to ID collisions if a single `ApplicationInstance` is used with multiple `tool_consumer_instance_guid`'s.

Fortunately we [don't have to support this use-case](https://hypothes-is.slack.com/archives/CBN3DGW02/p1620825877011700?thread_ts=1620824689.009200&cid=CBN3DGW02). We can show the user an error message and refuse to launch.

### Solution

We'll make it so that you can't install the same ApplicationInstance in multiple LMS's or change your LMS's GUID. We're going to show an error message in that case

1. When an `ApplicationInstance` is first created (when the user submits the https://lms.hypothes.is/welcome form) the `ApplicationInstance` hasn't been installed in an LMS or launched by an LMS yet. At this time the `ApplicationInstance` isn't associated with any `tool_consumer_instance_guid` in the DB.

2. The first time we receive an LTI launch for an `ApplicationInstance` that isn't yet associated with a `tool_consumer_instance_guid` we associate that `ApplicationInstance` with this request's `tool_consumer_instance_guid`.

3. If we receive an LTI launch for an `ApplicationInstance` and the `tool_consumer_instance_guid` in the launch params doesn't match the `ApplicationInstance`'s related `tool_consumer_instance_guid` we refuse to launch and show an error message.

[...]

PR #3053, opened 2021-08-16
Error message if an ApplicationInstance.tool_consumer_guid is reused

marcospri:
For #2679 

# Testing 

- Change a local AI tool_consumer_guid to trigger the error  (in canvas)
   `update application_instances set tool_consumer_instance_guid = 'another' where consumer_key = 'Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a';`

- Launch an assignment, for example: https://hypothesis.instructure.com/courses/125/assignments/873
- You should see an error dialog: 

Screenshot from 2021-08-18 16-33-16

- Same thing should happen when configuring an assignment and in any other LMS (for example in BB use `update application_instances set tool_consumer_instance_guid = 'another' where consumer_key = 'Hypothesc9deb2e461f88be4a8d5364';`) and launch

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -10,6 +10,7 @@ from sqlalchemy.dialects.postgresql import JSONB
 
 from lms.db import BASE
 from lms.models.application_settings import ApplicationSettings
+from lms.models.exceptions import ReusedConsumerKey
 
 LOG = logging.getLogger(__name__)
 
@@ -142,7 +143,10 @@ class ApplicationInstance(BASE):
         ):
             # If we already have a LMS guid linked to the AI
             # and we found a different one report it to sentry
-            raise ValueError("Application Instance launched in a different LMS install")
+            raise ReusedConsumerKey(
+                existing_guid=self.tool_consumer_instance_guid,
+                new_guid=tool_consumer_instance_guid,
+            )
 
         self.tool_consumer_instance_guid = tool_consumer_instance_guid
         for attr in [
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -142,11 +142,7 @@ class ApplicationInstance(BASE):
         ):
             # If we already have a LMS guid linked to the AI
             # and we found a different one report it to sentry
-            LOG.exception(
-                "Application Instance ID:%s launched in a different LMS install",
-                self.id,
-            )
-            return
+            raise ValueError("Application Instance launched in a different LMS install")
 
         self.tool_consumer_instance_guid = tool_consumer_instance_guid
         for attr in [
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -16,10 +16,12 @@ class JSConfig:
         OAUTH2_REDIRECT_ERROR = "oauth2-redirect-error"
         BASIC_LTI_LAUNCH = "basic-lti-launch"
         CONTENT_ITEM_SELECTION = "content-item-selection"
+        ERROR_DIALOG = "error-dialog"
 
     class ErrorCode(str, Enum):
         BLACKBOARD_MISSING_INTEGRATION = "blackboard_missing_integration"
         CANVAS_INVALID_SCOPE = "canvas_invalid_scope"
+        REUSED_TOOL_GUID = "reused_tool_guid"
 
     def __init__(self, context, request):
         self._context = context
@@ -151,6 +153,17 @@ class JSConfig:
                     "errorCode": error_code,
                     "errorDetails": error_details,
                     "canvasScopes": canvas_scopes or [],
+                },
+            }
+        )
+
+    def enable_error_dialog_mode(self, error_code, error_details=""):
+        self._config.update(
+            {
+                "mode": JSConfig.Mode.ERROR_DIALOG,
+                "errorDialog": {
+                    "errorCode": error_code,
+                    "errorDetails": error_details,
                 },
             }
         )
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -1,4 +1,5 @@
 import functools
+from enum import Enum
 from typing import List
 
 from lms.models import GroupInfo, HUser
@@ -10,6 +11,11 @@ from lms.views.helpers import via_url
 
 class JSConfig:
     """The config for the app's JavaScript code."""
+
+    class Mode(str, Enum):
+        OAUTH2_REDIRECT_ERROR = "oauth2-redirect-error"
+        BASIC_LTI_LAUNCH = "basic-lti-launch"
+        CONTENT_ITEM_SELECTION = "content-item-selection"
 
     def __init__(self, context, request):
         self._context = context
@@ -135,7 +141,7 @@ class JSConfig:
 
         self._config.update(
             {
-                "mode": "oauth2-redirect-error",
+                "mode": JSConfig.Mode.OAUTH2_REDIRECT_ERROR,
                 "OAuth2RedirectError": {
                     "authUrl": auth_url,
                     "errorCode": error_code,
@@ -153,7 +159,7 @@ class JSConfig:
 
         :raise ConsumerKeyError: if request.lti_user.oauth_consumer_key isn't in the DB
         """
-        self._config["mode"] = "basic-lti-launch"
+        self._config["mode"] = JSConfig.Mode.BASIC_LTI_LAUNCH
 
         self._config["api"]["sync"] = self._sync_api()
 
@@ -185,7 +191,7 @@ class JSConfig:
 
         self._config.update(
             {
-                "mode": "content-item-selection",
+                "mode": JSConfig.Mode.CONTENT_ITEM_SELECTION,
                 "filePicker": {
                     "formAction": form_action,
                     "formFields": form_fields,
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -17,6 +17,10 @@ class JSConfig:
         BASIC_LTI_LAUNCH = "basic-lti-launch"
         CONTENT_ITEM_SELECTION = "content-item-selection"
 
+    class ErrorCode(str, Enum):
+        BLACKBOARD_MISSING_INTEGRATION = "blackboard_missing_integration"
+        CANVAS_INVALID_SCOPE = "canvas_invalid_scope"
+
     def __init__(self, context, request):
         self._context = context
         self._request = request
--- a/lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
@@ -23,8 +23,7 @@ export default function ErrorDialogApp() {
 
   switch (errorCode) {
     case 'reused_tool_guid':
-      title =
-        'The Hypothesis credentials your install is using may have already been used in a different LMS instance';
+      title = 'Consumer key registered with another site';
       message = 'Reused tool_consumer_instance_guid';
       break;
     default:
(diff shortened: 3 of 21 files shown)


lms/models/application_instance.py
```diff
@@ -142,11 +142,7 @@ def update_lms_data(self, params: dict):
         ):
             # If we already have a LMS guid linked to the AI
             # and we found a different one report it to sentry
-            LOG.exception(
```
marcospri, line 145: Probably better to have a custom exception here but for this PoC this is enough

lms/resources/_js_config/__init__.py
```diff
@@ -125,7 +125,7 @@ def enable_oauth2_redirect_error_mode(  # pylint: disable=too-many-arguments
           OAuth client does not have access to all the necessary scopes
         :param canvas_scopes: List of scopes that were requested
         """
-        if self._lti_user:
+        if auth_route and self._lti_user:
```
marcospri, line 128: Allows to pass `auth_route` None

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -124,6 +131,23 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
           </p>
         </Fragment>
       )}
+      {error.code === 'reused_tool_guid' && (
+        <Fragment>
+          <p>
+          Launched with a different tool_consumer_instance_guid
```
marcospri, line 137: Placeholder messages etc

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -0,0 +1,65 @@
...
+              </a>{' '}
+              of your LMS has changed
+            </li>
+          </ul>
+        </Fragment>
+      )}
+      <ErrorDisplay message={message} error={error} />
+    </Dialog>
+  );
+}
```
marcospri, line 65: I'll update the copy based on the lastest from https://hypothes-is.slack.com/archives/C2BLQDKHA/p1629287584304700

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -0,0 +1,65 @@
...
+
+import ErrorDisplay from './ErrorDisplay';
+import Dialog from './Dialog';
+
+/**
+ * Error dialog displayed when authorization with a third-party API via OAuth
+ * fails.
+ *
+ */
+export default function ErrorDialogApp() {
```
marcospri, line 14: My thinking here is that this could be used for any error, or rather any error that 's not oauth related (covered by `OAuth2RedirectErrorApp`)

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -0,0 +1,65 @@
...
+  } = useContext(Config);
+
+  const error = { code: errorCode, details: errorDetails };
+
+  const title = (() => {
+    return (
+      {
+        reused_tool_guid:
+          'The Hypothesis credentials your install is using may have already been used in a different LMS instance',
+      }[errorCode] || 'An error occurred'
```
marcospri, line 26: The default here are not supposed to happen, but, just in case, not sure if the non descriptive error dialog is better than letting  the exception bubble up to sentry or similar.

lms/models/application_instance.py
```diff
@@ -142,11 +142,7 @@ def update_lms_data(self, params: dict):
...
-                "Application Instance ID:%s launched in a different LMS install",
-                self.id,
-            )
-            return
+            raise ValueError("Application Instance launched in a different LMS install")
```
jon-betts, line 145: I think this does deserve a custom exception as we're catching it with an exception view. There are just so many ways a `ValueError` could be raised that aren't this and would trigger the dialog.

jon-betts (changes requested):
I've tested this and it works. I'm a big fan of the `Enum`s and it appears there's some extra inheritance magic going on with these ones I'd be keen to learn about.

I think on balance a custom error is probably needed. I think `ValueError` is too common an exception. I'd probably be ok with it if we were raising and catching it locally, but we're relying on the exception views, which means it's going all the way up the stack. I'm worried there are other ways this could be raised.

There is a benefit to adding a custom exception class though, which is we can rely on the class alone to determine this is the correct error. This means we'd only need one exception handler instead of two, and this handler could go in the standard place with the others in `h.views.exceptions`.

To summarise:

 * We probably need a custom exception here
 * The exception views could probably one one exception view?
 * I think the correct location for these is `h.views.exceptions`

lms/models/application_instance.py
marcospri: @jon-betts somehow I marked this as resolved myself :shrug: . Yeah, changing it to a new exception type

lms/resources/_js_config/__init__.py
```diff
@@ -11,6 +12,17 @@
 class JSConfig:
     """The config for the app's JavaScript code."""
 
+    class Mode(str, Enum):
```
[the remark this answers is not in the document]
marcospri: There's no need to do `Mode.MODE_NAME.value`, I think when you are using the value of the enum (as opposed of just using auto() and maybe only using it in conditionals) is a bit nicer.

https://docs.python.org/3/library/enum.html#others

tests/unit/lms/views/api/canvas/authorize_test.py
```diff
@@ -195,6 +195,7 @@ def test_it_configures_frontend_app(self, pyramid_request, params, error_code):
     def pyramid_request(self, pyramid_request, OAuth2RedirectResource):
         context = OAuth2RedirectResource(pyramid_request)
         context.js_config = mock.create_autospec(JSConfig, spec_set=True, instance=True)
+        context.js_config.ErrorCode = JSConfig.ErrorCode
```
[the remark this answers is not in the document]
marcospri: Yup, now because the view only lives on the `views/exceptions`  I only need to this once.

lms/models/application_instance.py
marcospri: Absolutely :+1:

lms/models/application_instance.py
```diff
@@ -142,11 +143,11 @@ def update_lms_data(self, params: dict):
         ):
             # If we already have a LMS guid linked to the AI
             # and we found a different one report it to sentry
-            LOG.exception(
-                "Application Instance ID:%s launched in a different LMS install",
-                self.id,
+            raise ReusedToolGuidError(
+                "Application Instance launched in a different LMS install",
```
jon-betts, line 147: I think I'd prefer this text to be inside the exception itself as it acts as good documentation of what the exception represents. Either that or add a docstring?

If there's only one possible value for this it's kind of nice to tuck the string generation into the exception I think.

jon-betts (approved):
Yep this is looking good! Nice to have one error handler instead of two.

lms/models/application_instance.py
marcospri: Yup,that message is a bit pointless if it's always the same, It made a little bit of sense when it was a `ValueError`. Updated.

seanh (changes requested):
I had a go at tweaking the wording:

Screenshot from 2021-08-23 12-51-21

Notes:

* It's nice if the title of the dialog is relatively short (it should fit on one line). The first paragraph after the title can repeat what the title already said but with more words so that you don't have to try to cram all of the information into the title itself

* Say "consumer key" instead of "credentials". Consumer keys are what these things are called and that name is exposed to users both at <https://lms.hypothes.is/welcome> when they register with us, and in the UI of the LMS (Canvas etc) when they install us. So it's best to be direct here rather than making the user figure out that "credentials" really means "consumer key".

* Just referred to "LMS" instead of "LMS instance". "Instance" is programmer terminology and not necessary here

* Used "This could be because", as used in other error dialogs, instead of "This may be due to one of the following".

* Added suggestions for what the user can do (or what their admin must do) to remedy the problem. Always good to have these in an error message.

seanh:
P.S. **Register a new consumer key** should link to <https://lms.hypothes.is/welcome>.

Also I removed the link to the LMS spec, that's not for end users

lms/resources/_js_config/__init__.py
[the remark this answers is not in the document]
seanh: In 3.11 there's gonna be a stdlib in `StrEnum` that I think is equivalent: https://docs.python.org/3.11/library/enum.html#enum.StrEnum

seanh (changes requested):

seanh:
Could @robertknight take a look at the frontend changes here?

lms/static/scripts/frontend_apps/components/test/ErrorDialogApp-test.js
```diff
@@ -0,0 +1,50 @@
...
+    );
+  };
+
+  beforeEach(() => {
+    fakeConfig = {
+      errorCode: null,
+      errorDetails: '',
+    };
+
+    sinon.stub(window, 'close');
```
robertknight, line 29: Stubbing `window.close()` isn't needed in this test. It was needed for `OAuth2RedirectErrorApp` because that is used in a popup window and it has a button that calls `window.close()`.

lms/static/scripts/frontend_apps/index.js
```diff
@@ -60,6 +61,9 @@ switch (config.mode) {
   case 'oauth2-redirect-error':
     app = <OAuth2RedirectErrorApp />;
     break;
+  case 'error-dialog':
```
robertknight, line 64: Can you move this `case` statement above the previous one, so they are in sorted order by `mode`.

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -0,0 +1,65 @@
+import { Fragment, createElement } from 'preact';
+import { useContext } from 'preact/hooks';
+
+import { Config } from '../config';
+
+import ErrorDisplay from './ErrorDisplay';
+import Dialog from './Dialog';
+
+/**
+ * Error dialog displayed when authorization with a third-party API via OAuth
```
robertknight, line 10: This description is still the one for `OAuth2RedirectErrorApp`. Maybe something like "A general-purpose error dialog displayed when a frontend application cannot be launched." Also worth noting in the comments that there are more specific error dialogs for some use cases (eg. `OAuth2RedirectErrorApp`).

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
robertknight: This seems like a reasonable generalization, although we don't know for sure that it is the right one until we have other use cases in future.

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ */
+export default function ErrorDialogApp() {
+  const {
+    ErrorDialog: { errorCode, errorDetails = '' },
+  } = useContext(Config);
+
+  const error = { code: errorCode, details: errorDetails };
+
+  const title = (() => {
```
robertknight, line 21: For initializing a set of related variables based on an enum value we generally prefer a `switch`, like so:

```js
  let title;
  let message;

  switch (errorCode) {
    case 'reused_tool_guid':
      title =
        'The Hypothesis credentials your install is using may have already been used in a different LMS instance';
      message = 'Reused tool_consumer_instance_guid';
      break;
    default:
      title = 'An error occurred';
      message = 'Unknown error occurred';
  }
```

There are pros and cons to this, but it has the benefit of grouping all the initializations for the same enum value. It is also just the most obvious language construct to use for this purpose.

I probably should have recommended this for the recent changes in `OAuth2RedirectErrorApp` as well.

robertknight (commented):
The general approach looks good. I had some minor feedback on code details and the name of the configuration key.

dwhly:
> Just referred to "LMS" instead of "LMS instance".

Maybe "account" instead of instance?  After all it's still the same LMS (Canvas), right?

robertknight (commented):
Latest frontend changes LGTM.

seanh:
> > Just referred to "LMS" instead of "LMS instance".
> 
> Maybe "account" instead of instance? After all it's still the same LMS (Canvas), right?

How about "site"? I think that might be the word-in-common-usage that differentiates one instance of a given web app from another. E.g. people know what a particular "WordPress site" is etc.
 Removing the kind of technical term "LMS" might be good anyway and it avoids the whole "LMS instance" or "LMS account" conundrum.
> ### Consumer key registered with another site
>
> The consumer key that this Hypothesis installation is using appears to have already been used on another site. This could be because:
>
> * This consumer key has already been used on another site. [Register a new consumer key](https://lms.hypothes.is/welcome) for this site and re-install Hypothesis.
>
> * ...

marcospri:
> > > Just referred to "LMS" instead of "LMS instance".
> > 
> > 
> > Maybe "account" instead of instance? After all it's still the same LMS (Canvas), right?
> 
> How about "site"? I think that might be the word-in-common-usage that differentiates one instance of a given web app from another. E.g. people know what a particular "WordPress site" is etc.
> Removing the kind of technical term "LMS" might be good anyway and it avoids the whole "LMS instance" or "LMS account" conundrum.
> 
> > ### Consumer key registered with another site
> > The consumer key that this Hypothesis installation is using appears to have already been used on another site. This could be because:
> > 
> > * This consumer key has already been used on another site. [Register a new consumer key](https://lms.hypothes.is/welcome) for this site and re-install Hypothesis.
> > * ...

Updated the PR with latest tweaks from @seanh

seanh:
Following [more Slack discussion](https://hypothes-is.slack.com/archives/C2BLQDKHA/p1629816901485600?thread_ts=1629287584.304700&cid=C2BLQDKHA) here's a final suggestion:

> ### Consumer key registered with another site
>
> This Hypothesis installation's consumer key appears to have already been used on another site. This could be because:
>
> * The consumer key has already been used on another site. A site admin must [request a new consumer key](https://web.hypothes.is/get-help/) for this site and re-install Hypothesis.
>
> * This site's `tool_consumer_instance_guid` has changed. A site admin must [ask us to update the consumer key](https://web.hypothes.is/get-help/).
>
> If the problem persists...

Both of the links are just links to <https://web.hypothes.is/get-help/> which is the same thing that the **open a support ticket** further down in the dialog links to.

PR #3056, opened 2021-08-17
Replace is_scope_invalid flag with an error code

marcospri:
Small refactor to remove the boolean field and just use an error code

# Testing

- Change the requested scopes at https://github.com/hypothesis/lms/blob/c66700613707002b52309f0e83303e04895346f3/lms/views/api/canvas/authorize.py#L22 (just change one line or modify one to FILES_SCOPES)
- Launch an assignment (eg: https://hypothesis.instructure.com/courses/125/assignments/873) The modal should show up

### DIFF
--- a/lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
+++ b/lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
@@ -22,7 +22,6 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
   const {
     OAuth2RedirectError: {
       authUrl = /** @type {string|null} */ (null),
-      invalidScope = false,
       errorCode = /** @type {string|null} */ (null),
       errorDetails = '',
       canvasScopes = /** @type {string[]} */ ([]),
@@ -32,7 +31,7 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
   const error = { code: errorCode, details: errorDetails };
 
   const title = (() => {
-    if (invalidScope) {
+    if (errorCode === 'canvas_invalid_scope') {
       return 'Developer key scopes missing';
     }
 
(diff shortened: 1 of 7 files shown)


lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -44,9 +43,6 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
   })();
 
   const message = (() => {
-    if (invalidScope) {
-      return null;
-    }
```
marcospri, line 49: There's a couple of thing going on here:

- Now for the `invalidScope` equivalent it shows  `Something went wrong when authorizing Hypothesis` but the alternative, the null value,  displays `An unknown error occurred.` which is arguably worse. 

    I think this should be solved while tackling: https://github.com/hypothesis/lms/issues/2997 and change it to "null" not displaying any message.

- This is now a function that contains only a return statement. I think this way is easier to change/extend later but I'm happy changing it back to just a constant value.

robertknight:
The frontend parts look good to me from a skim through the code.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
robertknight: I would suggest to change this to a constant value, as I think it is pretty obvious how it could be changed to handle other values in future. This isn't one of those scenarios where a lot of effort has gone into the scaffolding that needs to be preserved for the future.

jon-betts (approved):
Yeah this looks good to me. 

I think this is a good example of the correct kind of generality. Having a separate boolean for every situation isn't simpler, easier or in anyway less effort than this type of approach. Worse != simpler.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
marcospri: You are right, I was thinking the next PR based on this one would change it but it actually uses a different file. Changed it to a const value.

PR #3065, opened 2021-08-24
Remove superfluous arg

marcospri:
This gets a bit on the way of doing another bigger refactor.

### DIFF
--- a/lms/services/course.py
+++ b/lms/services/course.py
@@ -5,9 +5,8 @@ from lms.models import Course, CourseGroupsExportedFromH, LegacyCourse
 
 
 class CourseService:
-    def __init__(self, application_instance_service, consumer_key, db):
+    def __init__(self, application_instance_service, db):
         self._application_instance = application_instance_service.get()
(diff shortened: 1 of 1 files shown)


lms/services/course.py
```diff
@@ -121,6 +122,5 @@ def _is_pre_sections(self, authority_provided_id):
 def course_service_factory(_context, request):
     return CourseService(
         request.find_service(name="application_instance"),
-        request.lti_user.oauth_consumer_key,
```
marcospri, line 124: I don't think `request.lti_user.oauth_consumer_key` and request.find_service(name="application_instance").consumer_key should be ever different.

jon-betts (approved):
So long story short, the user consumer key is the one we get called with in LTI params. The application instance service is also instantiated with the same value which is used if you don't supply a consumer key.

So as long as you never need to call this with a custom consumer_key and you don't pass anything to application_instance_service.get() . I think this is fine?

seanh (approved):

PR #3066, opened 2021-08-24
Add coverage back to GitHub actions

seanh:
https://github.com/hypothesis/lms/pull/2809 changed `tox -e lint` to just run the tests and no longer run the coverage check, and added a new separate `tox -e coverage` to run the coverage check. It also updated `make sure` to run the new `tox -e coverage` command (by depending on a new `make coverage` command). But it forgot to add `tox -e coverage` to GitHub Actions (our GitHub Actions config calls `tox` directly, not `make sure`). So coverage hasn't been enforced on CI since 14th June and as a result we now have some missing coverage on master.

Note that the `coverage` environment in tox has to run **after** the `tests` environment. These two can't run in parallel since coverage needs the output of tests as its input. For example this wouldn't normally work because it'd run the test and coverage commands in parallel:

    tox --parallel -e tests,coverage,...

We want to tell tox not to start `coverage` until after `tests` has finished running. But we want to preserve as much parallelism as possible otherwise: `tests` can run at the same time as `functests` or `bddtests` or anything else except `coverage`. And `coverage` can run at the same time as anything except `tests`. To achieve this we use tox's `depends` setting to specify a dependency of `coverage` on `tests`:

    depends =
        coverage: tests

Now if you run `tox --parallel  -e tests,coverage,...` it won't start `coverage` until `tests` has finished.

This also works without parallelism. For example `tox -e coverage,tests,...` will re-order the environments and run `tests` first.

But you can still run `tox -e coverage` (or `make coverage`) to run **just the coverage**. The `depends` setting won't "pull in" the `tests` env and run it if not requested.

See:

https://tox.readthedocs.io/en/latest/example/basic.html#parallel-mode
https://tox.readthedocs.io/en/latest/config.html#conf-depends


.github/workflows/continuous-integration.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       run: yarn build
 
     - name: Run tox
-      run: tox --parallel auto -e checkformatting,lint,tests,functests,bddtests
+      run: tox --parallel -e checkformatting,lint,tests,coverage,functests,bddtests
```
seanh, line 80: Also removed the `auto` from this line in a separate commit. It's not needed: `auto` is the default

tox.ini
```diff
@@ -61,6 +61,8 @@ deps =
     bddtests: -r requirements/bddtests.txt
     dockercompose: -r requirements/dockercompose.txt
     {env:EXTRA_DEPS:}
+depends =
+    coverage: tests
```
seanh, line 65: Tells tox not to start `coverage` until `tests` has finished. Only takes effect if both `coverage` and `tests` are included in the envlist (e.g. `tox --parallel -e coverage,tests,...`)

tox.ini
```diff
@@ -56,11 +56,13 @@ deps =
     {format,checkformatting}: -r requirements/format.txt
     lint: -r requirements/lint.txt
     tests: -r requirements/tests.txt
-    coverage: -r requirements/tests.txt
+    coverage: -r requirements/coverage.txt
```
seanh, line 59: Changed the `coverage` environment to use its own requirements file since it only needs to install `coverage` not all of the test dependencies.

Makefile
```diff
@@ -13,6 +13,7 @@ help:
 	@echo "make format            Correctly format the code"
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
 	@echo "make test              Run all unit tests"
+	@echo "make coverage          Print the unit test coverage report"
```
seanh, line 16: `make coverage` already exists but it was missing from `make help`

Makefile
```diff
@@ -71,7 +72,7 @@ checkformatting: python
 	@tox -qe checkformatting
 
 .PHONY: test
-test: backend-tests coverage frontend-tests
+test: backend-tests frontend-tests
```
seanh, line 75: Changed `make test` so that it doesn't run `coverage`. You can run `make coverage` separately or you can run `make test coverage`. This is how it's done in other projects (e.g. bouncer, h)

marcospri (approved):

PR #3067, opened 2021-08-24
 Store users in the DB POC

marcospri:
- Left HUser intact for now although I think we probably want to store "userid" in the new table.

- LTIUser relates to ApplicationInstance so `LTIUser.oauth_consumer_key` becomes `LTIUser.application_instance.consumer_key`

   this has a number of implications in his PR:
   - The validation of the consumer_key happens earlier, for example in `lms/validation/authentication/_launch_params.py` via `_user_service.upsert_from_lti`
   - I think most uses of `application_instance_service.get()` can be replaced by `lti_user.application_instance` (because the application instance was fetched get upserting the lti_user). I think if that's the case .get() should now always take an argument with the consumer_key.


lms/migrations/versions/7178f5812e85_user_table.py
```diff
@@ -0,0 +1,42 @@
...
+        sa.Column(
+            "updated", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("user_id", sa.UnicodeText(), nullable=False),
+        sa.Column("oauth_consumer_key", sa.UnicodeText(), nullable=False),
+        sa.Column("roles", sa.UnicodeText(), nullable=False),
+        sa.Column("tool_consumer_instance_guid", sa.UnicodeText(), nullable=False),
+        sa.Column("display_name", sa.UnicodeText(), nullable=True),
+        sa.Column("email", sa.UnicodeText(), nullable=True),
```
marcospri, line 33: I'm not sure if the nullability of the fields here is right.

lms/services/application_instance.py
```diff
@@ -33,5 +33,8 @@ def get(self, consumer_key=None) -> ApplicationInstance:
 
 
 def factory(_context, request):
-    consumer_key = request.lti_user.oauth_consumer_key if request.lti_user else None
+    # TODO this is silly, we are picking now the consumer_key from the actual AI
+    consumer_key = (
+        request.lti_user.application_instance.consumer_key if request.lti_user else None
+    )
```
marcospri, line 39: I'd say if nothing is wrong with the approach of this PR this param should be removed.

lms/services/course.py
```diff
@@ -50,15 +49,17 @@ def any_with_setting(self, group, key, value=True):
 
         return bool(
             self._db.query(LegacyCourse)
-            .filter(LegacyCourse.consumer_key == self._consumer_key)
+            .filter(
+                LegacyCourse.consumer_key == self._application_instance.consumer_key
+            )
```
marcospri, line 54: Created a PR with just this https://github.com/hypothesis/lms/pull/3065

lms/services/user.py
```diff
@@ -0,0 +1,67 @@
...
+    ):
+
+        user = self.get(oauth_consumer_key, user_id)
+
+        if not display_name:
+            display_name = create_display_name(
+                lis_person_name_given,
+                lis_person_name_family,
+                lis_person_name_full,
+            )
```
marcospri, line 42: I think this is better here instead of spread around the code base. Let the service own the data it manages.

lms/services/user.py
```diff
@@ -0,0 +1,67 @@
...
+            self._db.query(User)
+            .join(ApplicationInstance)
+            .filter(
+                ApplicationInstance.oauth_consumer_key == oauth_consumer_key,
+                User.user_id == user_id,
+            )
+            .one_or_none()
+        )
+
+    def upsert_from_lti(
```
marcospri, line 22: `upsert_from_lti` as it takes arguments with the names from the lti launches. I haven't found any need for another simpler method.

PR #3070, opened 2021-08-27
Add `resourcekey` parameter to Google Drive PDF URLs

robertknight:
If a user selects a PDF file for an assignment which has a "resource key" associated with it, capture this key from the Google
Picker-provided file metadata and add it to the file download URL as a `resourcekey` query parameter. This mirrors the updated download URL for affected files if you use the "Download" action for the file in the Google Drive UI.

The `resourceKey` field on file objects returned by the picker API is not documented at https://developers.google.com/picker/docs/reference#document but it is present in my testing.

It would still be preferable for us not to generate this URL ourselves, but this commit is an easy incremental change that captures the necessary information for files affected by the security update. I'm going to look into other ways to generate this URL as a follow-up.

----

**Testing:**

1. In Canvas create an assignment by choosing a PDF file from Google Drive which *is affected* by the security update. After configuring the assignment, the URL that is submitted back to Canvas (see the `url` query param in the "URL" field at the bottom of the "Configure external tool" dialog) should have a `resourcekey` parameter.
2. Repeat step 1, but choose a PDF file from Google Drive that *is not affected* by the security update. The URL submitted back to Canvas should not have a `resourcekey` parameter

Additionally the `resourcekey` query parameter can be seen in the client:

Google Drive resourcekey

You can find PDF files in Google Drive which are affected by the security update by searching for `is:security_update_applied`. Example PDF: https://drive.google.com/drive/u/0/search?q=is:security_update_applied. Example: https://drive.google.com/file/d/0Bx7aaYEfzIONNGtseEZVV3JBWEU/view?usp=sharing&resourcekey=0-TMgSyIxJsVVArstX-G4Kog.

The URL that is generated for the assignment should match the one that you get if you click the "Download" button in Google Drive.

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/google-picker-client.js
+++ b/lms/static/scripts/frontend_apps/utils/google-picker-client.js
@@ -36,6 +36,15 @@ export class PickerCanceledError extends Error {
  * @typedef PickerDocument
  * @prop {string} id
  * @prop {string} url
+ * @prop {string} [resourceKey] - A key which is present on a subset of older
+ *   Google Drive files. If present this needs to be provided to various
+ *   Google Drive APIs and share links (via the `resourcekey` query param).
+ *
+ *   As far as we understand, this key was added to old files to guard against
+ *   their IDs being guessed.
+ *
+ *   See https://developers.google.com/drive/api/v3/resource-keys and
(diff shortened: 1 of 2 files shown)


seanh (approved):
:+1:

PR #3091, opened 2021-09-08
Add some extra log output to try and narrow down failures

jon-betts:


lyzadanger (approved):
Thanks! This fixes the ol' glitch.

bin/update_dev_data.py
```diff
@@ -86,11 +87,20 @@ def upsert_application_instance(self, data):
...
+                    resource_link_id=data["resource_link_id"],
+                    tool_consumer_instance_guid=data["tool_consumer_instance_guid"]
+                )
+                .one_or_none()
+            )
+
+        except MultipleResultsFound:
+            print(f"Found more than one row for {data['resource_link_id']} in the DB")
+            print(f"We were going to upsert: {data}")
+            raise
```
seanh, line 103: I think we can remove the `except`. There's a [unique constraint on `(resource_link_id, tool_consumer_instance_guid)`](https://github.com/hypothesis/lms/blob/d62445c2c020eb54c78060c33a943d07938f8a30/lms/models/assignment.py#L26) so the code is gonna work correctly now. Bug fixed

seanh (changes requested):

bin/update_dev_data.py
```diff
@@ -15,6 +15,7 @@
 import tempfile
 
 from pyramid.paster import bootstrap
+from sqlalchemy.exc import MultipleResultsFound
```
seanh, line 18: Unused import. PyLint doesn't seem to be doing its job for some reason

seanh (changes requested):

seanh (approved):

PR #3093, opened 2021-09-09
Fix missing test coverage for lms.assets

marcospri:
Another bit of coverage missing. 

This is a bit light on asserts. Not sure what's important here https://github.com/hypothesis/lms/blob/70f52f90cac40658e74ad0103205a0a2aa012996/lms/assets.py that might be worth explicilty testing for. 

Before most of this file was move to `h_assets` the include had a `no cover`.

### DIFF
--- a/tests/unit/lms/assets_test.py
+++ b/tests/unit/lms/assets_test.py
@@ -1,0 +1,8 @@
+from lms.assets import includeme
+
+
+def test_includeme(pyramid_config):
+    includeme(pyramid_config)
+
+    assets_env = pyramid_config.registry["assets_env"]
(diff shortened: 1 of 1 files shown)


tests/unit/lms/assets_test.py
```diff
@@ -0,0 +1,8 @@
+from lms.assets import includeme
+
+
+def test_includeme(pyramid_config):
+    includeme(pyramid_config)
+
+    assets_env = pyramid_config.registry["assets_env"]
+    assert assets_env.assets_base_url == "/assets"
```
marcospri, line 8: This is the only attribute directly accessible from assets_env

seanh (approved):

PR #3112, opened 2021-09-14
Make it possible to run `make run-docker` and get a functioning system

jon-betts:
Currently in LMS the `make run-docker` command doesn't work and contains instructions saying you should fill out all the environment variables.

Even if you do this, there's no route by default from the container to the host, so it wouldn't work. 

This attempts to automate the process by filling in some values and loading directly from devdata. It turns out (after many blind alleys), that `docker-compose` is completely setup to read directly from our env file already. Which is nice. We then steal some tricks from ViaHTML to get a stable name for the host IP.

## Testing notes

 * `make devdata`
 * `make docker`
 * `make run-docker`
 * Start all the other services
 * Visit some testing assignments

You should now have a functioning LMS instance hopefully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,28 +105,7 @@ docker:
 
 .PHONY: run-docker
 run-docker:
-	# To run the Docker container locally, first build the Docker image using
-	# `make docker` and then set the environment variables below to appropriate
-	# values (see conf/development.ini for non-production quality examples).
-	@docker run \
-		--net lms_default \
-		-e DATABASE_URL=postgresql://postgres@postgres/postgres \
-		-e FEATURE_FLAGS_COOKIE_SECRET \
-		-e H_API_URL_PRIVATE \
-		-e H_API_URL_PUBLIC \
-		-e H_AUTHORITY \
-		-e H_CLIENT_ID \
-		-e H_CLIENT_SECRET  \
-		-e H_JWT_CLIENT_ID \
-		-e H_JWT_CLIENT_SECRET \
-		-e JWT_SECRET \
-		-e LMS_SECRET \
-		-e RPC_ALLOWED_ORIGINS \
-		-e VIA_URL \
-		-e SESSION_COOKIE_SECRET \
-		-e OAUTH2_STATE_SECRET \
-		-p 8001:8001 \
-		hypothesis/lms:$(DOCKER_TAG)
+	@tox -e dockercompose -- up --force-recreate web
 
 .PHONY: backend-lint
 backend-lint: python
--- a/bin/dev_host_bridge.sh
+++ b/bin/dev_host_bridge.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Write the network of the host to the /etc/hosts file if "host.docker.internal"
+# is not supported on the system.
+# From: https://dev.to/bufferings/access-host-from-a-docker-container-4099
+
+# Check if we are on a supported system
+HOST_DOMAIN="host.docker.internal"
+ping -q -c1 $HOST_DOMAIN > /dev/null 2>&1
+
+# If not, then write find our IP and map "host.docker.internal" to it
+if [ $? -ne 0 ]; then
+  HOST_IP=$(ip route | awk 'NR==1 {print $3}')
+  echo -e "$HOST_IP\t$HOST_DOMAIN" >> /etc/hosts
+fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
 version: '3'
+
 services:
   postgres:
     image: postgres:11.5-alpine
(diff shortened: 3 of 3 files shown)


bin/run_docker/dev_host_bridge.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Check if we are on a supported system
+HOST_DOMAIN="host.docker.internal"
+ping -q -c1 $HOST_DOMAIN > /dev/null 2>&1
+
+# If not, then write find our IP and map "host.docker.internal" to it
+if [ $? -ne 0 ]; then
+  HOST_IP=$(ip route | awk 'NR==1 {print $3}')
+  echo -e "$HOST_IP\t$HOST_DOMAIN" >> /etc/hosts
+fi
```
jon-betts, line 15: Nabbed from LMS

Makefile
```diff
@@ -105,28 +105,7 @@ docker:
 
 .PHONY: run-docker
 run-docker:
-	# To run the Docker container locally, first build the Docker image using
-	# `make docker` and then set the environment variables below to appropriate
-	# values (see conf/development.ini for non-production quality examples).
```
jon-betts, line 110: Good luck with that

bin/dev_host_bridge.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Check if we are on a supported system
+HOST_DOMAIN="host.docker.internal"
+ping -q -c1 $HOST_DOMAIN > /dev/null 2>&1
+
+# If not, then write find our IP and map "host.docker.internal" to it
+if [ $? -ne 0 ]; then
+  HOST_IP=$(ip route | awk 'NR==1 {print $3}')
+  echo -e "$HOST_IP\t$HOST_DOMAIN" >> /etc/hosts
+fi
```
jon-betts, line 15: This is lifted from the same solution in ViaHTML to support Mac and linux folk. For a refreshing and annoying change, Linux is the odd one our here which doesn't support "host.docker.internal" by default (Mac and Windows do).

Why? Who knows.

docker-compose.yml
```diff
@@ -1,9 +1,39 @@
...
+    ports:
+      - '127.0.0.1:8001:8001'
+    # Unfortunately if we don't run as root `dev_host_bridge.sh` doesn't work
+    user: root
+    command: /bin/sh -c "/var/lib/lms/bin/dev_host_bridge.sh && bin/init-env supervisord -c conf/supervisord.conf"
+    env_file: .devdata.env
+    environment:
+      - DATABASE_URL=postgresql://postgres@lms_postgres_1/postgres
+      - FEATURE_FLAGS_COOKIE_SECRET=not_a_secret
+      - H_API_URL_PRIVATE=http://host.docker.internal:5000/api/
```
jon-betts, line 32: This is where we use the host we setup

bin/dev_host_bridge.sh
seanh: This writes to the `/etc/hosts` file in the Docker container, right? Not on my host system

docker-compose.yml
seanh: Where do all these envvars come from when the non-Docker `make dev` is run?

I can see that some of these are Docker-specific (like `H_API_URL_PRIVATE=http://host.docker.internal:5000/api/`) but many of them don't look Docker-specific and look like the same values that `make dev` might use

seanh (changes requested):
When I run `make services` I'm getting:

```
ERROR: Network lms_default declared as external, but could not be found. Please create the network manually using `docker network create lms_default` and try again.
```

I think we might want to add this to a separate docker-compose file rather than the default `docker-compose.yml`. It'd be nice to avoid having to list all the services in `Makefile`. But also currently the `make services` command interacts with all the Docker Compose'd services (just Postgres for LMS, Postgres, Elastic and Rabbit for h, etc). `make services` starts all these up `make services args="down"` takes them all down. Etc. By adding the LMS app's own `Dockerfile` to `docker-compose.yml` the effect of `make services args="foobar"` is changed because it's now acting on the new "web" service as well.

Or maybe the two lists of services can be split into two separate groups within the `docker-compose.yml` file

docker-compose.yml
jon-betts: Some come from `tox.ini` most come from `conf/development.ini`.

bin/dev_host_bridge.sh
jon-betts: Yes. This is executed inside the docker container as the first part of the command run. See the `docker-compose.yml` for more details.

docker-compose.yml
```diff
@@ -1,9 +1,39 @@
...
+      lms_default: { }
     ports:
       - '127.0.0.1:5433:5432'
     healthcheck:
         test: ["CMD", "pg_isready", "-U", "postgres"]
         interval: 1s
+
+  web:
+    image: hypothesis/lms:dev
+    profiles: ["web"]
```
jon-betts, line 20: This "profile" should prevent the web service from being started by default unless you specify it by name or select this profile.

jon-betts:
Could you try this again?

I suspect that the earlier network stuff was due to it having been created the once on my machine. I'm not sure why it wasn't on yours as it turns "lms_default" it's one created by docker-compose by default, but... I have no idea why. There's not much in the compose file I can see that should make this happen. It might be based on the directory name? It's all pretty opaque.

I've added a "web" profile to the "web" container, which means it isn't run by default unless you ask for it by name or profile. The profiles are tags which let you group containers and start them together.

docker-compose.yml
seanh: Yeah, I don't see any reason not to move them into `tox.ini`

seanh (approved):
Works for me

PR #3115, opened 2021-09-16
WIP Canvas assignments in the DB

marcospri:


marcospri (commented):
The main question here is: do we keep both `hypothesis_assignment_id` and use `ext_lti_assignment_id` too avoid duplicates while editing? Or just use `ext_lti_assignment_id`?

There's different possibilities when using `ext_lti_assignment_id`, we could for example keep the column, but give it the value from `ext_lti_assignment_id` instead of generating one ourselves. We could just store it extra without adding new columns.

Or we could use both ids independently as I'm doing in this PR but that's a bit of an accident.

I think I'm leaning towards:

-  adding a new column, generically named hypothesis_id or similar
-  Make it unique with tool_consumer_guid (as is not very clear if it's unique globally, probably it is as it looks as UUID, but just in case https://community.canvaslms.com/t5/Canvas-Question-Forum/Is-ext-lti-assignment-id-field-globally-unique/td-p/402981)
- Make the constraint for both `resource_link_id` and the new column not being null at the same  time.
- Explicitly pass the hypothesis_id to assignment_service.create instead of generating one inside.

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -34,9 +35,13 @@ export default function FilePickerFormFields({
   formFields,
   groupSet,
   ltiLaunchURL,
+  hypothesisId,
 }) {
   /** @type {Record<string,string>} */
   const extraParams = groupSet ? { group_set: groupSet } : {};
+  if (hypothesisId) {
```
marcospri, line 42: Should we not send the url of the document to canvas anymore? I think we still should send it even if we don't use it.

lms/services/assignment.py
```diff
@@ -10,29 +12,96 @@ def __init__(self, db):
...
         """
-        assignment = self.get(tool_consumer_instance_guid, resource_link_id)
-
+        assignment = self.get(
+            tool_consumer_instance_guid, resource_link_id, hypothesis_assignment_id
+        )
         return assignment.document_url if assignment else None
 
+    def _generate_hypotehesis_id(self):
+        return base64.urlsafe_b64encode(uuid.uuid4().bytes).decode().rstrip("=")
```
marcospri, line 59: This is a base64 encoded uuid which is 22 chars long.

lms/services/assignment.py
```diff
@@ -10,55 +12,110 @@ def __init__(self, db):
...
+        assignment = (
             self._db.query(Assignment)
-            .filter_by(
-                tool_consumer_instance_guid=tool_consumer_instance_guid,
-                resource_link_id=resource_link_id,
+            .filter(
+                Assignment.tool_consumer_instance_guid == tool_consumer_instance_guid,
+                (
+                    (Assignment.resource_link_id == resource_link_id)
+                    | (Assignment.ext_lti_assignment_id == ext_lti_assignment_id)
```
marcospri, line 34: Not sure if clear from the comment but this can't be an `and` there will be a time when we have stored the ext_lti_assignment_id but not the resource_link_id but we have both !=None in this function (on  first launch).

seanh: And I guess we have pre-existing assignments that have a `resource_link_id` but no `ext_lti_assignment_id`. And also non-Canvas assignments too. So we do need to support looking up by `resource_link_id` only.

I'm wondering about this query and `None` / `NULL` though. For example if the call is `get(tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id=None)` won't that match *all* rows that have `NULL` in the `ext_lti_assignment_id`?

lms/services/assignment.py
```diff
@@ -10,55 +12,110 @@ def __init__(self, db):
...
             .one_or_none()
         )
 
-    def get_document_url(self, tool_consumer_instance_guid, resource_link_id):
+        if not assignment:
+            return None
+
+        if resource_link_id and not assignment.resource_link_id:
+            # First lunch of the assignment, fill the resource_link_id now that we have it.
+            assignment.resource_link_id = resource_link_id
```
seanh, line 45: What about the case where we find an assignment with a matching `ext_lti_assignment_id` and it has a *different* `resource_link_id` than the one we received in the launch params? That should never happen, but maybe we want to make sure that we catch that case and raise an error

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -34,9 +35,13 @@ export default function FilePickerFormFields({
   formFields,
   groupSet,
   ltiLaunchURL,
+  extLTIAssignmentId,
 }) {
   /** @type {Record<string,string>} */
   const extraParams = groupSet ? { group_set: groupSet } : {};
+  if (extLTIAssignmentId) {
+    extraParams['ext_lti_assignment_id'] = extLTIAssignmentId;
```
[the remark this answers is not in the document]
seanh: I think I'm leaning towards just using the LTI name. Why replace it with our own?

seanh (approved):

lms/migrations/versions/64c815401627_assigment_new_id.py
```diff
@@ -0,0 +1,53 @@
...
+    op.alter_column(
+        "module_item_configurations",
+        "resource_link_id",
+        existing_type=sa.VARCHAR(),
+        nullable=True,
+    )
+    op.create_unique_constraint(
+        op.f("uq__module_item_configurations__ext_lti_assignment_id"),
+        "module_item_configurations",
+        ["ext_lti_assignment_id"],
```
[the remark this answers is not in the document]
marcospri: Yes, this was me modifying the model but not regenerating the migration. Fixed :+1:

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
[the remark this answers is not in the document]
marcospri: > I think I'm leaning towards just using the LTI name. Why replace it with our own?

I was just thinking on the length in the URL, that's 36 chars for the value and 21 for the parameter name

Issue #2527, opened 2021-03-29
Move Canvas assignments into the DB


marcospri:
Canva's assignments currently store all their configuration on the LTI launch URL as query params.

Follow the same pattern as other's LMS assignments and create/update/upsert a row in `ModuleItemConfiguration` per assignment.

This might be partially achieved by: https://github.com/hypothesis/lms/issues/2769

seanh:
Unlike in non-Canvas LMS's, in Canvas at the time when the user is creating an assignment and they select the assignment's document there is no LTI assignment ID (`resource_link_id`) yet. We therefore cannot use the `resource_link_id` as the ID for the assignment's config in our DB. For non-Canvas LMS's we currently _do_ use the `resource_link_id` as the ID for looking up the assignment's configuration in the DB: https://github.com/hypothesis/lms/blob/01159673e5f7f10832a2efad5cc91786baee3bb7/lms/models/module_item_configuration.py#L31-L32 but we won't be able to do that for Canvas assignments.

Instead we are going to have to generate our own unique ID for the assignment and save that both in the DB and in the LTI launch params. We will have to submit a launch URL like this to Canvas: `https://lms.hypothes.is/lti_launches?hypothesis_assignment_id=<ID>`. When we receive an LTI launch we read the `hypothesis_assignment_id` from the query params and use it to look up the `ModuleItemConfiguration` in the DB and find the assignment's configuration (e.g. its document URL).

So this means that we have to both:

1. Make an API call to our own backend to create a new `ModuleItemConfiguration` in the DB. This API call should return a new `hypothesis_assignment_id`
2. Still submit an LTI launch URL to Canvas like we currently do. But instead of having the actual assignment configuration in the query params, it's just going to have the `hypothesis_assignment_id`

We better namespace the query param name (`hypothesis_*`) so it doesn't clash with any current or future Canvas launch params.

The first time an assignment is _launched_ (after having been created) we do receive the `resource_link_id` param, as well as our `hypothesis_assignment_id`. At this point we need to record the `resource_link_id` in the `ModuleItemConfiguration` in our DB, forever associating this `resource_link_id` and `hypothesis_assignment_id` with each other.

[...]

PR #3117, opened 2021-09-17
Migration for new ext_lti_assignment_id column

marcospri:
For #2527

Generated from model changes in https://github.com/hypothesis/lms/pull/3118

Test migrating forward and backwards:

- `hdev alembic upgrade head`
- `hdev alembic downgrade d9c9e65c463e`

and check the output of `\d module_item_configurations` on `make sql`'s prompt

### DIFF
--- a/lms/migrations/versions/a58605fdd398_add_ext_lti_assignment_id.py
+++ b/lms/migrations/versions/a58605fdd398_add_ext_lti_assignment_id.py
@@ -1,0 +1,58 @@
+"""
+Adds module_item_configurations.ext_lti_assignment_id and a new CHECK constraint.
+
+Revision ID: a58605fdd398
+Revises: d9c9e65c463e
+Create Date: 2021-09-17 07:59:11.487484
+
+"""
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/a58605fdd398_add_ext_lti_assignment_id.py
```diff
@@ -0,0 +1,58 @@
...
+
+def upgrade():
+    op.add_column(
+        "module_item_configurations",
+        sa.Column("ext_lti_assignment_id", sa.UnicodeText(), nullable=True),
+    )
+    op.alter_column(
+        "module_item_configurations",
+        "resource_link_id",
+        existing_type=sa.VARCHAR(),
```
marcospri, line 25: Added by alermbic autogenerate. I guess it gives hints to generate the SQL.

seanh (approved):
This looks good but I think let's hold off on merging it until the other related PRs are approved, just in case we realise we need to change something

PR #3118, opened 2021-09-17
Model changes for new assignment.ext_lti_assignment_id

marcospri:
Migration on https://github.com/hypothesis/lms/pull/3117


seanh (approved):
Looks good but again I think let's hold off on merging this until we've reviewed the PRs that'll use it, in case we realise we need to change something

PR #3119, opened 2021-09-17
Assignment service uses new ext_lti_assignment_id column

marcospri:
This code should behave exactly as without any changes until further PRs are merged.

More context in https://github.com/hypothesis/lms/pull/3127

## Testing

- No behavior changes, launch and configure URL and canvas file assignments.

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -1,6 +1,9 @@
+import logging
 from functools import lru_cache
 
 from lms.models import Assignment
+
+log = logging.getLogger(__name__)
 
 
 class AssignmentService:
@@ -10,54 +13,120 @@ class AssignmentService:
         self._db = db
 
     @lru_cache(maxsize=128)
-    def get(self, tool_consumer_instance_guid, resource_link_id):
-        return (
-            self._db.query(Assignment)
-            .filter_by(
-                tool_consumer_instance_guid=tool_consumer_instance_guid,
-                resource_link_id=resource_link_id,
-            )
-            .one_or_none()
+    def get(
+        self,
+        tool_consumer_instance_guid,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
+    ):
+        """Get an assignment using using resource_link_id, ext_lti_assignment_id or both."""
+
+        query = self._db.query(Assignment).filter(
+            Assignment.tool_consumer_instance_guid == tool_consumer_instance_guid
         )
 
-    def get_document_url(self, tool_consumer_instance_guid, resource_link_id):
+        if resource_link_id and not ext_lti_assignment_id:
+            # Non canvas assignments
+            query = query.filter(Assignment.resource_link_id == resource_link_id)
+        elif not resource_link_id and ext_lti_assignment_id:
+            # creating/editing a canvas assignment
+            query = query.filter(
+                Assignment.ext_lti_assignment_id == ext_lti_assignment_id
+            )
+        elif resource_link_id and ext_lti_assignment_id:
+            # Canvas launch, potentially the first one where resource_link_id is not yet in the DB
+            query = query.filter(
+                Assignment.ext_lti_assignment_id == ext_lti_assignment_id,
+                (
+                    (Assignment.resource_link_id == resource_link_id)
+                    | (Assignment.resource_link_id.is_(None))
+                ),
+            )
+        else:
+            log.exception(
+                "Can't get an assignment without neither resource_link_id or ext_lti_assignment_id"
+            )
+            return None
+
+        assignment = query.one_or_none()
+        if not assignment:
+            return None
+
+        if resource_link_id and not assignment.resource_link_id:
+            # First lunch of a canvas assignment, fill the resource_link_id now
+            assignment.resource_link_id = resource_link_id
+
+        return assignment
+
+    def get_document_url(
+        self,
+        tool_consumer_instance_guid,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
+    ):
         """
         Return the matching document URL or None.
 
         Return the document URL for the assignment with the given
         tool_consumer_instance_guid and resource_link_id, or None.
         """
-        assignment = self.get(tool_consumer_instance_guid, resource_link_id)
-
+        assignment = self.get(
+            tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
+        )
         return assignment.document_url if assignment else None
 
+    def create(
+        self,
+        tool_consumer_instance_guid,
+        document_url,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
+    ):
+        assignment = Assignment(
+            tool_consumer_instance_guid=tool_consumer_instance_guid,
+            document_url=document_url,
+            resource_link_id=resource_link_id,
+            ext_lti_assignment_id=ext_lti_assignment_id,
+        )
+        self._db.add(assignment)
+        return assignment
+
     def set_document_url(
-        self, tool_consumer_instance_guid, resource_link_id, document_url
+        self,
+        tool_consumer_instance_guid,
+        document_url,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
     ):
         """
-        Save the given document_url.
+        Save the given document_url in an existing assignment or create new one.
 
         Set the document_url for the assignment that matches
-        tool_consumer_instance_guid and resource_link_id. Any existing document
-        URL for this assignment will be overwritten.
+        tool_consumer_instance_guid and resource_link_id/ext_lti_assignment_id
+        or create a new one if none exist on the DB.
+
+        Any existing document URL for this assignment will be overwritten.
         """
-        assignment = self.get(tool_consumer_instance_guid, resource_link_id)
+        assignment = self.get(
+            tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
+        )
 
         if assignment:
             assignment.document_url = document_url
         else:
-            self._db.add(
-                Assignment(
-                    document_url=document_url,
-                    resource_link_id=resource_link_id,
-                    tool_consumer_instance_guid=tool_consumer_instance_guid,
-                )
+            assignment = self.create(
+                document_url=document_url,
+                resource_link_id=resource_link_id,
+                tool_consumer_instance_guid=tool_consumer_instance_guid,
+                ext_lti_assignment_id=ext_lti_assignment_id,
             )
 
         # Clear the cache (@lru_cache) on self.get because we've changed the
         # contents of the DB. (Python's @lru_cache doesn't have a way to remove
         # just one key from the cache, you have to clear the entire cache.)
         self.get.cache_clear()
+
+        return assignment
 
 
 def factory(_context, request):
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -1,11 +1,8 @@
-import logging
 from functools import lru_cache
 
 from sqlalchemy.orm.exc import NoResultFound
 
 from lms.models import Assignment
-
-log = logging.getLogger(__name__)
 
 
 class AssignmentService:
@@ -123,10 +120,18 @@ class AssignmentService:
         assert not old_assignment.ext_lti_assignment_id
         assert not new_assignment.resource_link_id
 
+        old_extra = dict(old_assignment.extra)
+        assert not old_extra or (
+            len(old_extra) == 1 and "canvas_file_mappings" in old_extra
+        )
+
         self._db.delete(old_assignment)
         # Flushing early so the `resource_link_id` constraints doesn't
         # conflict between the deleted record and new_assignment .
         self._db.flush()
+
+        if old_canvas_file_mappings := old_extra.get("canvas_file_mappings"):
+            new_assignment.extra["canvas_file_mappings"] = old_canvas_file_mappings
 
         new_assignment.resource_link_id = old_assignment.resource_link_id
         return new_assignment
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -25,21 +25,24 @@ class AssignmentService:
                 "Can't get an assignment without neither resource_link_id or ext_lti_assignment_id"
             )
 
-        # Non canvas assignments, just use `resource_link_id`
-        if resource_link_id and not ext_lti_assignment_id:
-            return self._get_by_resource_link_id(
-                tool_consumer_instance_guid, resource_link_id
+        if all([resource_link_id, ext_lti_assignment_id]):
+            # When launching an assignment in Canvas there are both resource_link_id and
+            # ext_lti_assignment_id launch params.
+            return self._get_for_canvas_launch(
+                tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
             )
 
-        # Creating/editing a canvas assignment, only `ext_lti_assignment_id` is available
-        if not resource_link_id and ext_lti_assignment_id:
-            return self._get_by_ext_lti_assignment(
+        if not resource_link_id:
+            # When creating or editing an assignment Canvas launches us with an
+            # ext_lti_assignment_id but no resource_link_id.
+            return self._get_for_canvas_assignment_config(
                 tool_consumer_instance_guid, ext_lti_assignment_id
             )
 
-        # We have both ext_lti_assignment_id and resource_link_id, canvas launch.
-        return self._get_for_canvas_launch(
-            tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
+        # Non-Canvas assignment launches always have a resource_link_id and never
+        # have an ext_lti_assignment_id.
+        return self._get_by_resource_link_id(
+            tool_consumer_instance_guid, resource_link_id
         )
 
     def exists(
@@ -56,22 +59,6 @@ class AssignmentService:
             return False
         else:
             return True
-
-    def create(
-        self,
-        tool_consumer_instance_guid,
-        document_url,
-        resource_link_id=None,
-        ext_lti_assignment_id=None,
-    ):
-        assignment = Assignment(
-            tool_consumer_instance_guid=tool_consumer_instance_guid,
-            document_url=document_url,
-            resource_link_id=resource_link_id,
-            ext_lti_assignment_id=ext_lti_assignment_id,
-        )
-        self._db.add(assignment)
-        return assignment
 
     def set_document_url(
         self,
@@ -93,14 +80,16 @@ class AssignmentService:
             assignment = self.get(
                 tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
             )
-            assignment.document_url = document_url
         except NoResultFound:
-            assignment = self.create(
+            assignment = Assignment(
+                tool_consumer_instance_guid=tool_consumer_instance_guid,
                 document_url=document_url,
                 resource_link_id=resource_link_id,
-                tool_consumer_instance_guid=tool_consumer_instance_guid,
                 ext_lti_assignment_id=ext_lti_assignment_id,
             )
+            self._db.add(assignment)
+
+        assignment.document_url = document_url
 
         # Clear the cache (@lru_cache) on self.get because we've changed the
         # contents of the DB. (Python's @lru_cache doesn't have a way to remove
@@ -108,33 +97,6 @@ class AssignmentService:
         self.get.cache_clear()
 
         return assignment
-
-    def _merge_canvas_assignments(self, assignments):
-        # We stored a canvas file assignment before (storing its resource_link_id)
-        # we later configured it (storing its ext_lti_assignment_id) in a new row
-        # and now we are launching it, we want to merge those two assignments
-
-        # order is guaranteed by the query's `order by`
-        old_assignment, new_assignment = assignments
-
-        assert not old_assignment.ext_lti_assignment_id
-        assert not new_assignment.resource_link_id
-
-        old_extra = dict(old_assignment.extra)
-        assert not old_extra or (
-            len(old_extra) == 1 and "canvas_file_mappings" in old_extra
-        )
-
-        self._db.delete(old_assignment)
-        # Flushing early so the `resource_link_id` constraints doesn't
-        # conflict between the deleted record and new_assignment .
-        self._db.flush()
-
-        if old_canvas_file_mappings := old_extra.get("canvas_file_mappings"):
-            new_assignment.extra["canvas_file_mappings"] = old_canvas_file_mappings
-
-        new_assignment.resource_link_id = old_assignment.resource_link_id
-        return new_assignment
 
     def _get_by_resource_link_id(self, tool_consumer_instance_guid, resource_link_id):
         return (
@@ -146,18 +108,6 @@ class AssignmentService:
             .one()
         )
 
-    def _get_by_ext_lti_assignment(
-        self, tool_consumer_instance_guid, ext_lti_assignment_id
-    ):
-        return (
-            self._db.query(Assignment)
-            .filter_by(
-                tool_consumer_instance_guid=tool_consumer_instance_guid,
-                ext_lti_assignment_id=ext_lti_assignment_id,
-            )
-            .one()
-        )
-
     def _get_for_canvas_launch(
         self,
         tool_consumer_instance_guid,
@@ -165,6 +115,8 @@ class AssignmentService:
         ext_lti_assignment_id,
     ):
         """Get a canvas assignment by both resource_link_id and ext_lti_assignment_id."""
+        assert resource_link_id and ext_lti_assignment_id
+
         assignments = (
             self._db.query(Assignment)
             .filter(
@@ -202,11 +154,18 @@ class AssignmentService:
             # ext_lti_assignment_id and no resource_link_id.
             #
             # We need to merge the two assignments into one.
-            assignment = self._merge_canvas_assignments(assignments)
+
+            # order is guaranteed by the query's `order by`
+            old_assignment, new_assignment = assignments
+
+            assert not old_assignment.ext_lti_assignment_id
+            assert not new_assignment.resource_link_id
+
+            assignment = self._merge_canvas_assignments(old_assignment, new_assignment)
         else:
             assignment = assignments[0]
 
-        if resource_link_id and not assignment.resource_link_id:
+        if not assignment.resource_link_id:
             # We found an assignment with an ext_lti_assignment_id but no resource_link_id.
             # This happens the first time a new Canvas assignment is launched: the assignment got created
             # during content-item-selection with an ext_lti_assignment_id but no resource_link_id,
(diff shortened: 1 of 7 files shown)


lms/services/assignment.py
```diff
@@ -10,55 +13,149 @@ def __init__(self, db):
...
+            old_assignment = (
+                assignments[0] if assignments[0].resource_link_id else assignments[1]
+            )
+            assignment = (
+                assignments[0]
+                if assignments[0].ext_lti_assignment_id
+                else assignments[1]
+            )
+            self._db.delete(old_assignment)
+            self._db.flush()
```
marcospri, line 79: We need this because there are already canvas file assignments in the DB which have a `resource_link_id`.

After this, re-configured assignments will be created with `ext_lti_assignment_id` creating a duplicate.

lms/services/assignment.py
```diff
@@ -10,55 +13,149 @@ def __init__(self, db):
...
+                    )
+                    #  First launch of a newly configured assignment
+                    | (
+                        (Assignment.resource_link_id.is_(None))
+                        & (Assignment.ext_lti_assignment_id == ext_lti_assignment_id)
+                    )
+                )
+            )
+        else:
+            log.exception(
```
marcospri, line 57: bddtests expect this behavior, ie not raising here.

lms/services/assignment.py
```diff
@@ -10,55 +13,149 @@ def __init__(self, db):
...
+        tool_consumer_instance_guid,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
+    ):
+        """Get an assignment using using resource_link_id, ext_lti_assignment_id or both."""
+
+        query = self._db.query(Assignment).filter(
+            Assignment.tool_consumer_instance_guid == tool_consumer_instance_guid
         )
+        if resource_link_id and not ext_lti_assignment_id:
```
marcospri, line 27: This conditional and the query are a bit verbose but trying to be clever here makes it a bit opaque.

lms/services/assignment.py
```diff
@@ -1,64 +1,183 @@
...
+                    == tool_consumer_instance_guid,
+                    Assignment.resource_link_id == resource_link_id,
+                )
+                .one_or_none()
             )
-            .one_or_none()
+
+        return self._get_canvas(
+            tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
         )
```
marcospri, line 109: `.get` look less daunting now although all the code is still here (`._get_canvas` and `_merge_canvas_assignments`). I think is a good solution without a more deep refactoring.

seanh (changes requested):
Would you mind trying to split up the `get()` method some more?

I think your `_get_canvas()` is currently dealing with three separate cases (`not resource_link_id and ext_lti_assignment_id` , `resource_link_id and ext_lti_assignment_id`, and when both are `None`) and then the top-level `get()` deals directly with the remaining case (`resource_link_id and not ext_lti_assignment_id`). I think it might split things up more if `get()` began by directing each one of the three different cases to a different method.

I think we might want to consider making `_merge_canvas_assignments()` a public method as well. Partially just to further split up the code into smaller pieces. And partially because I'm not sure that a `get()` method should have a side-effects like merging things in the DB. I'm wondering under exactly what circumstances `_merge_canvas_assignments()` actually happens? But I think it might be better if the service method actually returns two assignments and the view detects that and calls `merge_canvas_assignments()`.

marcospri:
> Would you mind trying to split up the `get()` method some more?

I went a bit on that direction and also moved the private methods to the bottom of the file.

> I think your `_get_canvas()` is currently dealing with three separate cases (`not resource_link_id and ext_lti_assignment_id` , `resource_link_id and ext_lti_assignment_id`, and when both are `None`) and then the top-level `get()` deals directly with the remaining case (`resource_link_id and not ext_lti_assignment_id`). I think it might split things up more if `get()` began by directing each one of the three different cases to a different method.

Yes, I split the public get based on what you are trying to do as a caller and expanded a bit more the comments on what you might find on the db on the canvas method.

> I think we might want to consider making `_merge_canvas_assignments()` a public method as well. Partially just to further split up the code into smaller pieces. And partially because I'm not sure that a `get()` method should have a side-effects like merging things in the DB. I'm wondering under exactly what circumstances `_merge_canvas_assignments()` actually happens? But I think it might be better if the service method actually returns two assignments and the view detects that and calls `merge_canvas_assignments()`.

I don't think having side effects in the get is good but I don't having it return more than one row is better. Get doesn't get called directly, only from set_document/get_document so it happens directly on the predicate (even if  it appears to happen on the view).

I tried a new `is_canvas` predicate and duplicated the db_configured view to have a canvas specific one at that point we could use either a separate service or different methods on the same service but would also have to do the same thing on the db_configure predicate and I didn't like the end result.

For me the fact that `get` hides theses complexity trumps both the ugliness of the side effect query method and mixing canvas an non canvas logic in the service. I could see some clever service dispatch (eg the factory returning one instance of the service or another depending on from which LMS you are calling) could work without having to do a lot if `if is_canvas` all over the place.

lms/services/assignment.py
```diff
@@ -10,55 +13,210 @@ def __init__(self, db):
...
-        self, tool_consumer_instance_guid, resource_link_id, document_url
+        self,
+        tool_consumer_instance_guid,
+        document_url,
+        resource_link_id=None,
+        ext_lti_assignment_id=None,
     ):
         """
-        Save the given document_url.
+        Save the given document_url in an existing assignment or create new one.
```
marcospri, line 88: Same behavior, just added the comment and the `create` method.

lms/services/assignment.py
seanh: Hmm. I think we may have conflicting uses of `AssignmentService.get()`. In the LTI launch predicates the code seems to deliberately call `AssignmentService.get()` with just a `resource_link_id` that might be `None` and handles the fact that `get()` might return `None` by calling `bool` on it. The code does that here:

https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/predicates/_lti_launch.py#L34-L42

and again here:

https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/predicates/_lti_launch.py#L114-L116

In these cases I don't think you want to log an exception since this appears to be normal, expected behaviour. So an exception in the logs would just be confusing. And logging an exception will report the exception to Sentry as well.

In other places where `get()` is called the code seems to assume that it will *not* return `None`. This happens in the Canvas Files API views:

https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/api/canvas/files.py#L37-L47

And it also happens in two places in the basic LTI launch views:

https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/basic_lti_launch.py#L165-L168

https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/basic_lti_launch.py#L209-L217

I *think* these three are assuming that the return value is never `None`?

So there are two conflicting use cases. So I think you might want to split this into two different methods:

```python
class AssignmentService:
    def get(self, ...):
        # This will raise an exception if both resource_link_id and ext_lti_assignment_id are None.
        # This will also raise an exception if there's no matching assignment.
        # I don't *think* we expect these exceptions to ever actually get raised so we can just
        # allow the app to crash with a 500.

    def exists(self, ...):
        # This one always returns either True or False, never raises.
        # Alternatively this could return an assignment or None. But I think True or False might suffice.
        try:
            self.get(*args, **kwargs)
        except ValueError:
            return False
        else:
            return True
```

marcospri: There are two issues here:

- Not passing any of the ids to `.get()`. This is not expected behavior I think, we should detect that early and issue a 400 or similar. The bddtest fail if I raise here but I think only because they are trying weird cases around the lti spec.

    Based on this I think we don't 4xx on those (missing resource_link_id) because we do some sort of lti redirect. https://github.com/hypothesis/lms/blob/onedrive-auth-page/lms/validation/_lti_launch_params.py#L52 

- Not checking for none on get. The views on basic_lti_launch assume there's a document and don't check for None because they come from the `db_configured` predicate that checks exactly that (https://github.com/hypothesis/lms/blob/c86dfaecd820fa44e78c25242dfeed78a0a0a75c/lms/views/basic_lti_launch.py#L160)

     Then use on the predicate itself, here `get_document_url` is checking two things, that there's a row in the db and that it has a document. Refactoring it with `exists()` will mean doing a query to check if there's an assignment and another to get it and check the document. We could return the assignment on `exists` then but then we really wound't have an use for the `get`, will be just renaming it and between `get -> Assignment` and `exist -> Assignment` I prefer the first one.

seanh: Yeah it's doing a redirect instead of 404ing in order to be compliant with the LTI spec. It was necessary in order to get LTI certification.

I think the code that's calling `get(resource_link_id=None)` is doing so deliberately. It looks to me like it's doing `resource_link_id = request.params.get("resource_link_id")` instead of just `request.params["resource_link_id"]` deliberately so that `resource_link_id` will be `None` if it's not in the `request.params`. 

Hmm. Ok, so I think it's actually `get_document_url()` rather than `get()` that has two different use-cases?

1. The predicates are calling `get_document_url()`, possibly passing it `resource_link_id=None`, or just a `resource_link_id` that doesn't exist in the DB, and they expect it to return a document URL or `None`. In this case you don't want to raise if `resource_link_id` is `None` or if there's no matching assignment, you just want to return `None`.

2. The views are calling `get_document_url()` and _assuming_ (because the predicates must have already run) that it's going to return a document URL and not `None`. In this case you could raise if `resource_link_id` is `None` or if there's no matching assignment, because this should never happen.

So you might want separate `get_document_url()` and `has_document_url()` methods?

```python
class AssignmentService:
    @lru_cache
    def get(self, ...):
        # Raise if both resource_link_id and ext_lti_assignment_id are None.
        # Raise if there's no matching assignment.

    def get_document_url(self, ...):
        # Calls get() so it will raise if both resource_link_id and ext_lti_assignment_id are None
        # or if there's no matching assignment.
        # Also raise if there is a matching assignment but it has no document_url.

    def has_document_url(self, *args, **kwargs):
        try:
            self.get_document_url(*args, **kwargs)
        except WhateverError:
            return False
        else:
            return True
```

Note that these won't result in multiple DB queries because both `get_document_url()` and `has_document_url()` work by calling `get()` which has `@lru_cache` on it.

The purpose of all this is to allow `get()` and `get_document_url()` to raise if both `resource_link_id` and `ext_lti_assignment_id` are `None` or if there's no matching assignment, because we never expect that to happen. But in the predicates, where we need it to *not* raise, those can call `has_document_url()` instead.

I don't think `get()` can do `log.exception()` as you have it currently because I think that will report exceptions to Sentry during expected behaviour, the edge cases that those BDD tests are covering.

seanh: The point of all this (in case it's not clear) is to allow the `get()` method to look something like this:

```python
    @lru_cache
    def get(self, tool_consumer_instance_guid, resource_link_id=None, ext_lti_assignment_id=None):
        assert resource_link_id or ext_lti_assignment_id, "Can't get an assignment without a resource_link_id or ext_lti_assignment_id"

        if resource_link_id and ext_lti_assignment_id:
            return self._get_for_canvas_launch(resource_link_id, ext_lti_assignment_id, ...)

        if resource_link_id:
            return self._get_by_resource_link_id(resource_link_id, ...)

        return self._get_by_ext_lti_assignment(ext_lti_assignment_id, ...)
```

You might raise something else rather than an `assert`. But we can make it crash so that we'll get notified if anything is calling `get()` without a `resource_link_id` or an `ext_lti_assignment_id`. At the same time the _predicates_ can use the `has_document_url()` method to avoid any crash

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+            return self._get_by_ext_lti_assignment(
+                tool_consumer_instance_guid, ext_lti_assignment_id
+            )
+
+        # We have both ext_lti_assignment_id and resource_link_id, canvas launch.
+        return self._get_for_canvas_launch(
+            tool_consumer_instance_guid, resource_link_id, ext_lti_assignment_id
+        )
+
+    def exists(
```
marcospri, line 48: Kept it here in the service, it might be useful down the line somewhere other than the predicates.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+                Get it, fill the new value on `resource_link_id` and return it.
+
+            - Two rows, one with each of the ids an the other set to null.
+
+                In this case we stored the assignment in the past on launch with just the `resource_link_id`,
+                later, we re-configured it, creating a new row with just the `ext_lti_assignment_id`.
+
+                We need to get both and merge them into one.
+                We'll keep the one with `ext_lti_assignment_id` as is the newest one.
+        """
```
seanh, line 180:
```suggestion
        """Get a canvas assignment by both resource_link_id and ext_lti_assignment_id."""
```

I think we should document the different cases with code comments rather than in the docstring (see other comment below). The specific cases are sort of an internal implementation detail so probably don't belong in the docstring. And putting them in code comments gets them closer to the code they're documenting.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+        if len(assignments) == 2:
+            assignment = self._merge_canvas_assignments(assignments)
+        else:
+            assignment = assignments[0]
+
+        if resource_link_id and not assignment.resource_link_id:
+            # First lunch of a canvas assignment, fill the resource_link_id now
+            assignment.resource_link_id = resource_link_id
+
+        return assignment
```
seanh, line 216: I think it might be good to use this Python code where each of the different cases is dealt with as the place to put explanatory comments, rather than in the docstring or the SQLAlchemy query above. Putting the explanations of the cases next to the code that handles the cases gives the comments a better chance of being kept up to date. Something like this:

```python
if len(assignments) == 2:
    # We found two assignments: one with the matching resource_link_id and no ext_lti_assignment_id
    # and one with the matching ext_lti_assignment_id and no resource_link_id.
    #
    # This happens because legacy code used to store Canvas assignments in the DB with a
    # resource_link_id and no ext_lti_assignment_id, see https://github.com/hypothesis/lms/pull/2780
    #
    # Whereas current code stores Canvas assignments during content-item-selection with an
    # ext_lti_assignment_id and no resource_link_id.
    #
    # We need to merge the two assignments into one.
    assignment = self._merge_canvas_assignments(assignments)
else:
    assignment = assignments[0]

if resource_link_id and not assignment.resource_link_id:
    # We found an assignment with an ext_lti_assignment_id but no resource_link_id.
    # This happens the first time a new Canvas assignment is launched: the assignment got created
    # during content-item-selection with an ext_lti_assignment_id but no resource_link_id,
    # and then the first time the assignment is launched we add the resource_link_id.
    assignment.resource_link_id = resource_link_id

# We found an assignment with the matching resource_link_id and ext_lti_assignment_id.
return assignment
```

Those comments are pretty long but I think this might be an appropriate time for some long comments. The edge cases we're covering here are pretty damn obscure, especially the merge-assignments one. I'm not sure the wording of my suggested comments is ideal though, could probably be improved.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+                    & (Assignment.ext_lti_assignment_id == ext_lti_assignment_id)
+                )
+            ),
+        )
+
+        assignments = query.all()
+        if not assignments:
+            raise NoResultFound()
+
+        if len(assignments) == 2:
```
seanh, line 207: Ok, so I think `assignments` can _only_ be empty, length 1, or length 2. It's not possible to have three matching assignments. For example you couldn't have this:

1. A row with both the `resource_link_id` and the `ext_lti_assignment_id`
2. A row with the matching `resource_link_id` and no `ext_lti_assignment_id`
3. A row with the matching `ext_lti_assignment_id` but no `resource_link_id`

Because there are unique constraints on both `resource_link_id` and `ext_lti_assignment_id`, so 1 and 2 can't exist at the same time and 1 and 3 also can't exist at the same time. You either get none of them, 1 alone, 2 alone, 3 alone, or 2 and 3 together.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+    def _merge_canvas_assignments(self, assignments):
+        # We stored a canvas file assignment before (storing its resource_link_id)
+        # we later configured it (storing its ext_lti_assignment_id) in a new row
+        # and now we are launching it, we want to merge those two assignments
+        old_assignment = (
+            assignments[0] if assignments[0].resource_link_id else assignments[1]
+        )
+        assignment = (
+            assignments[0] if assignments[0].ext_lti_assignment_id else assignments[1]
+        )
```
seanh, line 124: Here you're implicitly assuming that one of the assignments has a `resource_link_id` and no `ext_lti_assignment_id` and the other has an `ext_lti_assignment_id` and no `resource_link_id`. I wonder if it might be nice to make that assumption explicit? Both because it might make the code more readable to explicitly state the assumption, and because it'll mean that we get notified if something's wrong.

This code also looks like it could assign the _same_ object to `old_assignment` and `assignment` in cases where the above assumption has failed. It might be good to just make it so that it can't possibly assign the two to the same object. Again both for readability and safety.

Maybe something like:

```python
if assignments[0].resource_link_id:
    old_assignment, new_assignment = assignments
else:
    new_assignment, old_assignment = assignments

assert not old_assignment.ext_lti_assignment_id
assert not new_assignment.resource_link_id
```

I think assertions might actually be fine and appropriate for this but could raise some other exception instead if you prefer.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+        self._db.delete(old_assignment)
+        self._db.flush()
```
seanh, line 126: Is the flush really necessary here?

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+        # We stored a canvas file assignment before (storing its resource_link_id)
+        # we later configured it (storing its ext_lti_assignment_id) in a new row
+        # and now we are launching it, we want to merge those two assignments
+        old_assignment = (
+            assignments[0] if assignments[0].resource_link_id else assignments[1]
+        )
+        assignment = (
+            assignments[0] if assignments[0].ext_lti_assignment_id else assignments[1]
+        )
+        self._db.delete(old_assignment)
```
seanh, line 125: Okay, so I guess "merging" the two assignments pretty much means deleting the old one. That might be fine. For example I guess it's fine to delete the old assignment's `document_url` as the new assignment has its own `document_url`.

But what about `assignment.extra`? I'm afraid we may need to copy that over to the new assignment. And I'm afraid that could be quite awkward if the new assignment could also potentially have a non-empty `extra`.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+        self._db.flush()
+        return assignment
```
seanh, line 127: Don't you want to do `assignment.resource_link_id = old_assignment.resource_link_id`?

lms/services/assignment.py
marcospri: I don't think we have to worry about extra. Existing canvas DB assignments don't have anything store there and after all we only care about those in all this merge bossiness.

lms/services/assignment.py
marcospri: Yes good point, much more readable. Assertions are :+1:

lms/services/assignment.py
marcospri: That already happens on the calling function but added it here to have a more independent function

lms/services/assignment.py
marcospri: One thing we could do for canvas assignments is store a list of all the `document_urls` a assignment had, might be useful? But I rather do that after all of this is done.

lms/services/assignment.py
marcospri: Yes, otherwise the old and new rows conflict on having the same `resource_link_id` before the delete takes effect 

This could be solved in other ways I think (deferrable constraint? CTEs?) but it all looks more complicated than a flush.

lms/services/assignment.py
```diff
@@ -10,55 +15,206 @@ def __init__(self, db):
...
+                    (Assignment.resource_link_id == resource_link_id)
+                    & (Assignment.ext_lti_assignment_id.is_(None))
+                )
+                #  First launch of a newly configured assignment
+                | (
+                    (Assignment.resource_link_id.is_(None))
+                    & (Assignment.ext_lti_assignment_id == ext_lti_assignment_id)
+                )
+            ),
+        )
```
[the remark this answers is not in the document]
marcospri: The `in` behaves weirdly with null as anything that compares against it (`select 1 where null = null;`) so I think we need to use `is null` (sqlalcehmy converts `x == null` to `x is null` so that will work too) explicitly.

lms/services/assignment.py
seanh: Weird. I would think that if you're both deleting an object and updating another object in the same unit of work SQLAlchemy or Postgres would be smart enough to process the deletes first, but I guess not

lms/services/assignment.py
```diff
@@ -10,55 +12,210 @@ def __init__(self, db):
...
-                    document_url=document_url,
-                    resource_link_id=resource_link_id,
-                    tool_consumer_instance_guid=tool_consumer_instance_guid,
-                )
+        except NoResultFound:
+            assignment = self.create(
+                document_url=document_url,
+                resource_link_id=resource_link_id,
+                tool_consumer_instance_guid=tool_consumer_instance_guid,
+                ext_lti_assignment_id=ext_lti_assignment_id,
```
seanh, line 102: I think we can refactor this a little so that there's only one line in the `try` and so that it's clear that the intention is to set the `document_url` on either a pre-existing or a newly-created assignment:

```python
try:
    assignment = self.get(...)  # Get the existing assignment if there is one.
except NoResultFound:
    # If there's no existing assignment create a new one.
    assignment = Assignment(...)  # Here pass everything *except* document_url to Assignment().
    self._db.add(assignment)

assignment.document_url = document_url
```

This `except` is taking advantage of the fact that SQLAlchemy's default interface lets you call `Assignment(...)` with any kwargs, or even no kwargs, so you can call it without `document_url`. I think you actually probably don't want the `AssignmentService.create()` method: it's taking two lines of code in the `except` and replacing them with one line in the `except` plus a 14-line `create()` method and it's less flexible because when calling `create()` you can't omit the `tool_consumer_instance_guid` or `document_url` and then set them later like you can when calling `Assignment()` directly.

tests/factories/assignment.py
```diff
@@ -10,4 +10,5 @@
     resource_link_id=RESOURCE_LINK_ID,
     tool_consumer_instance_guid=TOOL_CONSUMER_INSTANCE_GUID,
     document_url=Faker("uri"),
+    ext_lti_assignment_id=None,
```
seanh, line 13: I think this would default to `None` FWIW

tests/unit/lms/services/assignment_test.py
```diff
@@ -1,51 +1,138 @@
...
 
+from lms.models import Assignment
 from lms.services.assignment import AssignmentService, factory
 from tests import factories
 
 
 class TestAssignmentService:
-    def test_get(self, svc, assignment):
+    def test_get_without_ids(self, svc, assignment):
+
```
seanh, line 14:
```suggestion
```

tests/unit/lms/services/assignment_test.py
```diff
@@ -1,51 +1,138 @@
...
+        with pytest.raises(ValueError):
+            assignment = svc.get(
+                assignment.tool_consumer_instance_guid,
+                resource_link_id=None,
+                ext_lti_assignment_id=None,
+            )
+
+    def test_get_returns_None_if_theres_no_matching_assignment(self, svc):
```
seanh, line 22:
```suggestion
    def test_get_raises_NoResultFound_if_theres_no_matching_assignment(self, svc):
```

tests/unit/lms/services/assignment_test.py
```diff
@@ -1,51 +1,138 @@
...
+        with pytest.raises(NoResultFound):
+            svc.get("TOOL_CONSUMER_INSTANCE_GUID", "RESOURCE_LINK_ID")
+
+    def test_get_returns_None_if_theres_no_matching_assignment_with_two_ids(self, svc):
```
seanh, line 26:
```suggestion
    def test_get_raises_NoResultFound_if_theres_no_matching_assignment_with_two_ids(self, svc):
```

tests/unit/lms/services/assignment_test.py
```diff
@@ -1,51 +1,138 @@
...
+    @pytest.mark.parametrize(
+        "old_extra,new_extra",
+        [
+            ({}, {}),
+            ({"canvas_file_mappings": {1: 2}}, {"canvas_file_mappings": {1: 2}}),
+        ],
+    )
+    def test_get_merge_existing(
+        self, old_extra, new_extra, svc, assignment, assignment_canvas_not_launched
+    ):  # pylint:disable=protected-access
```
seanh, line 78: You can use the `db_session` fixture instead of `svc._db`, then you won't need the `# pylint:disable=protected-access`

lms/services/assignment.py
```diff
@@ -10,55 +12,210 @@ def __init__(self, db):
...
+            return self._get_by_resource_link_id(
+                tool_consumer_instance_guid, resource_link_id
+            )
 
-        Return the document URL for the assignment with the given
-        tool_consumer_instance_guid and resource_link_id, or None.
-        """
-        assignment = self.get(tool_consumer_instance_guid, resource_link_id)
+        # Creating/editing a canvas assignment, only `ext_lti_assignment_id` is available
+        if not resource_link_id and ext_lti_assignment_id:
```
seanh, line 35:
```suggestion
        if not resource_link_id and ext_lti_assignment_id:
            # When creating or editing an assignment Canvas launches us with an
            # ext_lti_assignment_id but no resource_link_id.
```

seanh (changes requested):

seanh (changes requested):
I think we're going to have to change this so that the method that the _predicates_ call (currently `AssignmentService.exists()`) can never have any side effects. The reason is this:

1. Create a new assignment in Canvas. Canvas launches us with an `ext_lti_assignment_id` but no `resource_key`. We call `_get_for_canvas_assignment_config()` and it finds no matching assignment in the DB so we save a new assignment with the `ext_lti_assignment_id` and no `resource_key`

2. Now launch the newly-created Canvas assignment for the first time. This time Canvas launches us with both a `resource_link_id` and an `ext_lti_assignment_id`. We call `_get_for_canvas_launch()`. It finds the assignment that we created in step 1 above and sets the existing assignment's `resource_link_id`.

   But look at the call stack when the `assignment.resource_link_id = resource_link_id` line gets executed: `AssignmentService` is being called by one of the view predicates. At this point in the request processing the request **has not been authorized**. The `_permits()` function in `security.py` hasn't been called yet. Could an unauthorized request cause the `resource_link_id`'s of assignments to be set? Or cause two existing assignments to get merged?

In fact I _think_ it turns out that unauthorized requests _cannot_ modify the DB: when authorization fails we fail the request (with a `404` or something) and that causes the DB transaction to not get committed. But I don't think we want to rely on that as the one thing preventing us from making potentially unauthorized changes to the DB.

Nonetheless, I don't think we want to do this: view predicates get called by Pyramid _before_ the view gets selected, before authorization happens, I think we're asking for trouble by letting them make changes to the DB.

marcospri:
Refactored again in another PR https://github.com/hypothesis/lms/pull/3158

PR #3122, opened 2021-09-17
Basic API view to create canvas assignments

marcospri:


lms/views/api/canvas/assignments.py
```diff
@@ -0,0 +1,50 @@
...
+
+        url = None
+        extra = {}
+        if content_type == "url":
+            url = content["url"]
+        elif content_type == "file":
+            url = f"canvas://file/course/{params['course_id']}/file_id/{params['content']['file']['id']}"
+            extra = {"canvas_file": params["content"]["file"]}
+        else:
+            raise ValueError("Unhandled content type on assignment")
```
marcospri, line 39: I don't think this needs to be handled + custom exception, seeing it on sentry should be enough.

lms/validation/_api.py
```diff
@@ -69,3 +70,21 @@ class APIRecordResultSchema(JSONPyramidRequestSchema):
...
+            size = fields.Integer(required=True)
+
+        type = fields.Str(required=True)
+        url = fields.Str(required=False)
+        file = fields.Nested(File, required=False)
+
+    ext_lti_assignment_id = fields.Str(required=True)
+    course_id = fields.Integer(required=True)
+    content = fields.Nested(Content, required=True)
+    groupset = fields.Integer(required=False, allow_none=True)
```
marcospri, line 90: Not sure if this is the right place for the schema.

lms/services/assignment.py
```diff
@@ -149,6 +150,9 @@ def set_document_url(
                 ext_lti_assignment_id=ext_lti_assignment_id,
             )
 
+        if extra:
+            assignment.extra = extra
+
```
marcospri, line 155: We want to store some info we see even if we don't query it directly, more details about canva's file, group set id ...

seanh (changes requested):
Although it's only going to be used for Canvas at first I think we'd like the new create-assignment API to be used in all LMS's in future.

Could we move the `create()` API to a non-Canvas-specific location, and could we enable it to be extensible to create non-Canvas assignments in future? Probably just rename the `file` parameter to `canvas_file` so that we can add ones like `blackboard_file` etc in future.

In non-Canvas LMS's when a teacher launches an unconfigured assignment the frontend shows the configure-assignment dialog and when the user selects a document the frontend does this weird form submission to the backend to create the assignment in the DB. See [`unconfigured_basic_lti_launch()`](https://github.com/hypothesis/lms/blob/d6807c6103288ecf4ac8173cbc990c9094ca2d8d/lms/views/basic_lti_launch.py#L233-L271) (the `form_fields` variable is the fields for the HTML form that the frontend will submit) and [`configure_assignment()`](https://github.com/hypothesis/lms/blob/d6807c6103288ecf4ac8173cbc990c9094ca2d8d/lms/views/basic_lti_launch.py#L290-L324) (the view that receives the form submission from the frontend). This is pretty horrible and I think we'd one day like to replace it with the frontend calling this new create-assignment API.

lms/validation/_api.py
```diff
@@ -69,3 +70,36 @@ class APIRecordResultSchema(JSONPyramidRequestSchema):
...
+
+    groupset = fields.Integer(required=False, allow_none=True)
+    """Groupset the assignment belongs to if created as a smalls groups assignment"""
+
+    content = fields.Nested(Content, required=True)
+
+    @validates_schema
+    def validate(self, data, **_kwargs):
+        if data["content"]["type"] == "url" and not data["content"].get("url"):
+            raise ValidationError("url is mandatory with type is 'url'")
```
seanh, line 105: And `data["content"]["file"]` is mandatory if `data["content"]["type"] == "file"` right?

seanh (approved):

PR #3123, opened 2021-09-17
Send the client info about the new create assignment endpoint

marcospri:
Send the frontend details about the new create assignment endpoint.

# Testing

The fronted currently ignores this so there's should be not behavior changes.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -189,6 +189,18 @@ class JSConfig:
             "allowedOrigins": self._request.registry.settings["rpc_allowed_origins"]
         }
 
+    def _create_assignment_api(self):
+        if not self._context.is_canvas:
+            return None
+
+        return {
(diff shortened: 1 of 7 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -189,6 +189,18 @@ def enable_lti_launch_mode(self):
...
 
+    def _create_assignment_api(self):
+        if not self._context.is_canvas:
+            return None
+
+        return {
+            "path": self._request.route_path("canvas_api.assignments.create"),
+            "data": {
+                "ext_lti_assignment_id": self._request.params["ext_lti_assignment_id"],
+                "course_id": self._request.params["custom_canvas_course_id"],
```
marcospri, line 200: We also send `custom_canvas_course_id` to the FE, which in turn will send it back to the backed on create_assignment so we can generate a `canvas://`  url, which includes the course.

PR #3125, opened 2021-09-17
Add redirect page for MS OneDrive authentication (OAuth)

esanzgar:
By default the OneDrive picker uses the page it was launched from as the redirect
URL for authentication. In the LMS context, the OneDrive picker opens a
window with the URL '/content_item_selection' and some AOuth specific parameters (`?oauth=....`). That causes the LMS to return a 404 `Page not found`.

OneDrive has a `redirectUri` option to launch the picker in a different
URL (see
https://docs.microsoft.com/en-us/onedrive/developer/controls/file-pickers/js-v72/open-file?view=odsp-graph-online).
This URL must be registered on the app's registration portal as a
redirect URL.

The redirect page for authentication must content the OneDrive JS
bundle.

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -51,6 +51,10 @@ def includeme(config):
     )
 
     config.add_route(
+        "onedrive.filepicker.redirect_uri", "/onedrive/filepicker/redirect"
+    )
+
+    config.add_route(
         "canvas_api.oauth.authorize",
         "/api/canvas/oauth/authorize",
         factory="lms.resources.OAuth2RedirectResource",
--- a/lms/templates/onedrive.html.jinja2
+++ b/lms/templates/onedrive.html.jinja2
@@ -1,0 +1,9 @@
+<!doctype html>
+<html lang=en>
+  <head>
+    <meta charset=utf-8>
+  </head>
+  <body>
+      <script type="text/javascript" src="https://js.live.net/v7.2/OneDrive.js"></script>
+  </body>
+</html>
--- a/lms/views/onedrive.py
+++ b/lms/views/onedrive.py
@@ -1,0 +1,15 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    request_method="GET",
+    route_name="onedrive.filepicker.redirect_uri",
+    renderer="lms:templates/onedrive.html.jinja2",
+)
+def redirect_uri(_request):
+    """
+    Return a basic HTML page with One Drive's JS SDK loaded.
+
+    This view's URL is provided to One Drive's frontend config as target to open the filepicker in.
+    """
+    return {}
--- a/lms/resources/_js_config/file_picker_config.py
+++ b/lms/resources/_js_config/file_picker_config.py
@@ -82,6 +82,7 @@ class FilePickerConfig:
         return {
             "enabled": True,
             "clientId": request.registry.settings["onedrive_client_id"],
+            "redirectURI": request.route_url("onedrive.filepicker.redirect_uri"),
         }
 
     @classmethod
--- a/tests/unit/lms/resources/_js_config/file_picker_config_test.py
+++ b/tests/unit/lms/resources/_js_config/file_picker_config_test.py
@@ -105,6 +105,9 @@ class TestFilePickerConfig:
         expected = {"enabled": enabled}
         if enabled:
             expected["clientId"] = sentinel.client_id
+            expected["redirectURI"] = pyramid_request.route_url(
+                "onedrive.filepicker.redirect_uri"
+            )
 
         assert config == expected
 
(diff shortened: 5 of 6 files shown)


marcospri (approved):
@esanzgar Added a small test so CI passes.

Looks good but I think I wrote most of this on the PoC PR, maybe @seanh can take a quick look.

lms/routes.py
```diff
@@ -50,6 +50,11 @@ def includeme(config):
         "blackboard_api.files.via_url", "/api/blackboard/courses/{course_id}/via_url"
     )
 
+    config.add_route(
+        "onedrive.filepicker.authorize",
+        "/api/onedrive/filepicker/authorize",
+    )
```
seanh, line 56:
```suggestion
    config.add_route(
        "onedrive.filepicker.authorize", "/api/onedrive/filepicker/authorize"
    )
```

The trailing comma forces Black to wrap each item onto its own line. If you remove the trailing comma it'll fit them both on one line

lms/templates/onedrive.html.jinja2
```diff
@@ -0,0 +1,5 @@
+{% extends "lms:templates/base.html.jinja2" %}
+
+{% block scripts %}
+  <script type="text/javascript" src="https://js.live.net/v7.2/OneDrive.js"></script>
+{% endblock %}
```
seanh, line 5: Ok, so we've got our backend rendering an HTML page with `OneDrive.js` in it.

Should this page be extending `base.html.jinja2`? The base template contains a bunch of stuff that I'm not sure whether we want from this page.

lms/routes.py
seanh: I think `"onedrive.filepicker.redirect"` might be a better name for this route, so it's the OneDrive redirect URL rather than an authorize URL.

For the path I think maybe just `"/onedrive/filepicker/redirect"` since it's an HTML page not an API endpoint.

lms/views/api/onedrive.py
```diff
@@ -0,0 +1,10 @@
+from pyramid.view import view_config
+
+
+@view_config(
+    request_method="GET",
+    route_name="onedrive.filepicker.authorize",
+    renderer="lms:templates/onedrive.html.jinja2",
+)
+def authorize(_request):
```
seanh, line 9:
```suggestion
def redirect_uri(_request):
```

I think the file should be `lms/views/onedrive.py` as I don't think this is an API endpoint

tests/unit/lms/views/api/onedrive_test.py
```diff
@@ -0,0 +1,8 @@
+from lms.views.api.onedrive import authorize
+
+
+class TestOneDriveAuthorizeView:
+    def test_it(self, pyramid_request):
+        template_variables = authorize(pyramid_request)
+
+        assert template_variables == {}
```
seanh, line 8: You could make this a function (rather than a class method) and cram it onto one line:

```suggestion
def test_redirect_uri(pyramid_request):
    assert authorize(pyramid_request) == {}
```

lms/views/api/onedrive.py
seanh: I wonder if we could add a docstring explaining what this view is for? My understanding is that this is the URL that the OneDrive popup window redirects to and it must contain the `OneDrive.js` script that receives the redirect and shows the OneDrive file picker

seanh (changes requested):
A few tweaks...

lms/templates/onedrive.html.jinja2
marcospri: Good point, changed to a minimal html template, no base.

lms/views/api/onedrive.py
marcospri: Added a bit of context in the docstring

lms/views/onedrive.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@view_config(
+    request_method="GET",
+    route_name="onedrive.filepicker.redirect_uri",
+    renderer="lms:templates/onedrive.html.jinja2",
+)
+def redirect_uri(_request):
+    """
+    Return a basic HTML page with One Drive's JS sdk loaded.
```
seanh, line 11:
```suggestion
    Return a basic HTML page with One Drive's JS SDK loaded.
```

lms/views/onedrive.py
```diff
@@ -0,0 +1,15 @@
...
+
+    This view's URL is provided to One Drive's front end config as target to open the filepicker in.
```
seanh, line 13:
```suggestion
    This view's URL is provided to One Drive's frontend config as target to open the filepicker in.
```

seanh (approved):

lms/views/onedrive.py
esanzgar: OneDrive should be a single word (in contrast to Google Drive).

PR #3126, opened 2021-09-20
Enable canvas assignment creation API

marcospri:
This is the final piece needed for storing all canvas assignments on the DB using the assigments.create endpoint merged already.

Here, the frontend will use this API if the backend provided the details only when creating/editing assignments.

# Testing notes 

## Check already configured (in canvas) assignments can get launched

- `delete from module_item_configurations ;` unless you want to keep some of your db assignments
- HTML assignment, https://hypothesis.instructure.com/courses/125/assignments/1513. It won't store any record on the DB
- Canvas file assigment, https://hypothesis.instructure.com/courses/125/assignments/875, It will store a new record in the DB, but not using the new code.

## Edit existing assignments

- Edit https://hypothesis.instructure.com/courses/125/assignments/1513 ( https://en.wikipedia.org/wiki/Margaret_(singer))
- New record will show up on the DB ( no `resource_link_id`)
- Launch it, `resource_link_id` will be filled now.

- Edit https://hypothesis.instructure.com/courses/125/assignments/875 with a canvas file
- New record will show up on the DB ( no `resource_link_id`) (note that this is a duplicate of another already exiting assignment. One will have a `resource_link_id`, the newest will have only `ext_lti_assignment_id`
- Launch it. The record with only `resource_link_id` will get deleted and the new one will have all information completed.

## Create new assignments

- Create new assignments and launch  them for both URL and canvas file. New row on creating, filled with `resource_link_id` on launch.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -8,6 +8,7 @@ import {
 } from 'preact/hooks';
 
 import { Config } from '../config';
+import { apiCall } from '../utils/api';
 import { truncateURL } from '../utils/format';
 
 import ContentSelector from './ContentSelector';
@@ -64,14 +65,17 @@ function contentDescription(content) {
 export default function FilePickerApp({ onSubmit }) {
   const submitButton = useRef(/** @type {HTMLInputElement|null} */ (null));
   const {
+    api: { authToken },
     filePicker: {
       formAction,
       formFields,
+      createAssignmentAPI: createAssignmentAPI,
       canvas: { groupsEnabled: enableGroupConfig, ltiLaunchUrl },
     },
   } = useContext(Config);
 
   const [content, setContent] = useState(/** @type {Content|null} */ (null));
+  const [extLTIAssignmentId, setExtLTIAssignmentId] = useState(null);
 
   const [groupConfig, setGroupConfig] = useState(
     /** @type {GroupConfig} */ ({
@@ -89,7 +93,36 @@ export default function FilePickerApp({ onSubmit }) {
    * render.
    */
   const [shouldSubmit, setShouldSubmit] = useState(false);
-  const submit = useCallback(() => setShouldSubmit(true), []);
+  const submit = useCallback(
+    async (/** @type {Content} */ content) => {
+      async function createAssignment() {
+        const data = {
+          ...createAssignmentAPI.data,
+          content,
+          groupset: groupConfig.groupSet,
+        };
+        const assignment = await apiCall({
+          authToken,
+          path: createAssignmentAPI.path,
+          data,
+        });
+        setExtLTIAssignmentId(assignment.ext_lti_assignment_id);
+      }
+      if (content && createAssignmentAPI && !extLTIAssignmentId) {
+        try {
+          await createAssignment();
+        } catch (error) {
+          setErrorInfo({
+            message: 'Creating or editing an assignment',
+            error: error,
+          });
+          return;
+        }
+      }
+      setShouldSubmit(true);
+    },
+    [authToken, createAssignmentAPI, extLTIAssignmentId, groupConfig.groupSet]
+  );
 
   // Submit the form after a selection is made via one of the available
   // methods.
@@ -107,7 +140,7 @@ export default function FilePickerApp({ onSubmit }) {
     content => {
       setContent(content);
       if (!enableGroupConfig) {
-        submit();
+        submit(content);
       }
     },
     [enableGroupConfig, submit]
@@ -152,7 +185,7 @@ export default function FilePickerApp({ onSubmit }) {
               <LabeledButton
                 disabled={groupConfig.useGroupSet && !groupConfig.groupSet}
                 variant="primary"
-                onClick={submit}
+                onClick={() => submit(content)}
               >
                 Continue
               </LabeledButton>
@@ -164,6 +197,7 @@ export default function FilePickerApp({ onSubmit }) {
             ltiLaunchURL={ltiLaunchUrl}
             content={content}
             formFields={formFields}
+            extLTIAssignmentId={extLTIAssignmentId}
             groupSet={groupConfig.useGroupSet ? groupConfig.groupSet : null}
           />
         )}
--- a/lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
@@ -13,6 +13,7 @@ import { contentItemForContent } from '../utils/content-item';
  * @prop {Record<string,string>} formFields - Form fields provided by the backend
  *   that should be included in the response without any changes
  * @prop {string|null} groupSet
+ * @prop {string|null} extLTIAssignmentId
  */
 
 /**
@@ -34,9 +35,13 @@ export default function FilePickerFormFields({
   formFields,
   groupSet,
   ltiLaunchURL,
+  extLTIAssignmentId,
 }) {
   /** @type {Record<string,string>} */
   const extraParams = groupSet ? { group_set: groupSet } : {};
+  if (extLTIAssignmentId) {
+    extraParams.ext_lti_assignment_id = extLTIAssignmentId;
+  }
   const contentItem = JSON.stringify(
     contentItemForContent(ltiLaunchURL, content, extraParams)
   );
--- a/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
@@ -7,6 +7,7 @@ import { Config } from '../../config';
 import FilePickerApp, { $imports } from '../FilePickerApp';
 import { checkAccessibility } from '../../../test-util/accessibility';
 import mockImportedComponents from '../../../test-util/mock-imported-components';
+import { waitFor, waitForElement } from '../../../test-util/wait';
 
 function interact(wrapper, callback) {
   act(callback);
@@ -31,6 +32,9 @@ describe('FilePickerApp', () => {
 
   beforeEach(() => {
     fakeConfig = {
+      api: {
+        authToken: 'dummyAuthToken',
+      },
       filePicker: {
         formAction: 'https://www.shinylms.com/',
         formFields: { hidden_field: 'hidden_value' },
@@ -55,12 +59,18 @@ describe('FilePickerApp', () => {
   /**
    * Check that the expected hidden form fields were set.
    */
-  function checkFormFields(wrapper, expectedContent, expectedGroupSet) {
+  function checkFormFields(
+    wrapper,
+    expectedContent,
+    expectedGroupSet,
+    expectedExtLTIAssignmentId
+  ) {
     const formFields = wrapper.find('FilePickerFormFields');
     assert.deepEqual(formFields.props(), {
       children: [],
       content: expectedContent,
       formFields: fakeConfig.filePicker.formFields,
+      extLTIAssignmentId: expectedExtLTIAssignmentId,
       groupSet: expectedGroupSet,
       ltiLaunchURL: fakeConfig.filePicker.canvas.ltiLaunchUrl,
     });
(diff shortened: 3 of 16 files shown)


lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,37 @@ export default function FilePickerApp({ onSubmit }) {
...
+      data.groupset = groupConfig.groupSet;
+      const assignment = await apiCall({
+        authToken,
+        path: createAssignmentAPI.path,
+        data: data,
+      });
+      setExtLTIAssignmentId(assignment.ext_lti_assignment_id);
+    }
+
+    if (content && createAssignmentAPI && !extLTIAssignmentId) {
```
marcospri, line 113: Ony call the API if:
- Content already selected
- API enabled the config
- And we don't have already called it

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,37 @@ export default function FilePickerApp({ onSubmit }) {
...
       submitButton.current.click();
     }
-  }, [shouldSubmit]);
+  }, [
+    shouldSubmit,
+    extLTIAssignmentId,
+    authToken,
+    content,
+    createAssignmentAPI,
+    groupConfig.groupSet,
```
marcospri, line 130: This seems like a long list but eslit suggested that.

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -34,9 +35,13 @@ export default function FilePickerFormFields({
   formFields,
   groupSet,
   ltiLaunchURL,
+  extLTIAssignmentId,
 }) {
   /** @type {Record<string,string>} */
   const extraParams = groupSet ? { group_set: groupSet } : {};
+  if (extLTIAssignmentId) {
+    extraParams.ext_lti_assignment_id = extLTIAssignmentId;
+  }
```
marcospri, line 44: `extraParams` end up in the URL on `lms/static/scripts/frontend_apps/utils/content-item.js`

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,37 @@ export default function FilePickerApp({ onSubmit }) {
   // Submit the form after a selection is made via one of the available
   // methods.
   useEffect(() => {
+    async function createAssignment() {
+      const data = { ...createAssignmentAPI.data };
```
marcospri, line 103: I get here:

```lms/static/scripts/frontend_apps/components/FilePickerApp.js:103:13 - error TS2322: Type '{ constructor?: Function | undefined; toString?: (() => string) | undefined; toLocaleString?: (() => string) | undefined; valueOf?: (() => Object) | undefined; hasOwnProperty?: ((v: PropertyKey) => boolean) | undefined; isPrototypeOf?: ((v: Object) => boolean) | undefined; propertyIsEnumerable?: ((v: PropertyKey) =>...' is not assignable to type 'Record<string, string | Content | null>'.
  Property 'constructor' is incompatible with index signature.
    Type 'Function' is not assignable to type 'string | Content | null'.
      Type 'Function' is not assignable to type 'string'.
```

I can figure out where the function come from, shouldn't createAssignmentAPI be of type `APICallInfo`

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,39 @@ export default function FilePickerApp({ onSubmit }) {
   // Submit the form after a selection is made via one of the available
   // methods.
   useEffect(() => {
+    async function createAssignment() {
+      /** @type Record<string, string|Content|null> */
```
marcospri, line 102: Maybe just `{Object}`? I was confused but the error in the next line, see comment.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,39 @@ export default function FilePickerApp({ onSubmit }) {
   // Submit the form after a selection is made via one of the available
   // methods.
   useEffect(() => {
+    async function createAssignment() {
```
marcospri, line 101: Should this be defined outside as a const?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,39 @@ export default function FilePickerApp({ onSubmit }) {
...
+      /** @type Record<string, string|Content|null> */
+      const data = { ...createAssignmentAPI.data };
+
+      data.content = content;
+      data.groupset = groupConfig.groupSet;
+      const assignment = await apiCall({
```
marcospri, line 107: This doesn't have an expected error response, eg a re-authorization.

Should I then try/catch it? If that's the case how can I show a generic, something went wrong message? (or something more usefull, but this case would likely be a bug).

Should I have boolean `creatingAssignment` or similar between this an getting the answer?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: Huh. It looks like this is a quirk to do with usage of the legacy `Object` type. The fix is:

```diff
diff --git a/lms/static/scripts/frontend_apps/config.js b/lms/static/scripts/frontend_apps/config.js
index e738df27..f5cc7d65 100644
--- a/lms/static/scripts/frontend_apps/config.js
+++ b/lms/static/scripts/frontend_apps/config.js
@@ -13,7 +13,7 @@ import { createContext } from 'preact';
  *   URL to display in a popup window if the call fails due to an authorization
  *   error. This is used when the API call requires an OAuth 2 authorization
  *   flow to be completed with eg. an external LMS's API
- * @prop {Object} [data]
+ * @prop {object} [data]
  */
```

Here `object` means "some value that is not a primitive (like a string or number), but we don't know what fields it might have. `Object` is a thing that exists for compatibility with existing JSDoc comments. @esanzgar has been busy changing it to `object` or `Record` recently.

`Object` is a thing that exists for compatibility with old JSDoc comments.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: > Should I then try/catch it? If that's the case how can I show a generic, something went wrong message? (or something more usefull, but this case would likely be a bug).

Yes. There is an existing `errorInfo` state in this component that shows a dialog when set. We can probably re-use that here.

> Should I have boolean creatingAssignment or similar between this an getting the answer?

Does this call involve any calls to potentially slow third-party APIs or does all the work happen in our app. If the former, we probably should add a loading state. If the latter, it may be less important.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: If you create the `data` object in one step, you can avoid needing any type hints:

```diff
diff --git a/lms/static/scripts/frontend_apps/components/FilePickerApp.js b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
index d1a94aae..8282b067 100644
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -99,15 +99,16 @@ export default function FilePickerApp({ onSubmit }) {
   // methods.
   useEffect(() => {
     async function createAssignment() {
-      /** @type Record<string, string|Content|null> */
-      const data = { ...createAssignmentAPI.data };
+      const data = {
+        ...createAssignmentAPI.data,
+        content,
+        groupset: groupConfig.groupSet,
+      };

-      data.content = content;
-      data.groupset = groupConfig.groupSet;
       const assignment = await apiCall({
         authToken,
         path: createAssignmentAPI.path,
-        data: data,
+        data,
       });
       setExtLTIAssignmentId(assignment.ext_lti_assignment_id);
     }
```

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: The idea is that effects are informed of their dependencies so they re-run when any of them change. In this case, can any of the values change after all the requirements are met to trigger the `createAssignment` call (the `content && createAssignmentAPI && !extLTIAssignmentId` check above)?

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -13,6 +13,7 @@ import { contentItemForContent } from '../utils/content-item';
  * @prop {Record<string,string>} formFields - Form fields provided by the backend
  *   that should be included in the response without any changes
  * @prop {string|null} groupSet
+ * @prop {string|null} extLTIAssignmentId
```
robertknight, line 16: Is there a good piece documentation somewhere in the backend code about what this parameter means?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: Ah - I see that the `!extLTIAssignmentId` check at the end ensures that we only call this once.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: You can do, although you'd need to use [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) to help Preact keep track of the effect's dependencies. This is fine for the moment.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -94,13 +98,39 @@ export default function FilePickerApp({ onSubmit }) {
...
+        authToken,
+        path: createAssignmentAPI.path,
+        data: data,
+      });
+      setExtLTIAssignmentId(assignment.ext_lti_assignment_id);
+    }
+
+    if (content && createAssignmentAPI && !extLTIAssignmentId) {
+      createAssignment();
+      return;
```
robertknight, line 117: What is the relation between calling this new API and submitting the form?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: Added a try and used `errorInfo`. This is how the UI looks:

Screenshot from 2021-09-23 15-07-43

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -113,6 +114,55 @@ describe('FilePickerApp', () => {
...
   }
 
+  context('when create assignment configuration is enabled', () => {
+    const authURL = 'https://testlms.hypothes.is/authorize';
+    const path = '/api/canvas/assignments';
+    let fakeAPICall;
+    let fakeNewAssignment;
+
+    beforeEach(() => {
+      fakeConfig.filePicker.createAssignmentAPI = {
```
marcospri, line 124: I'm trying to set the config value for this:

https://github.com/hypothesis/lms/blob/0fb537ecdf423d414a5be470cf76a1d55879bd9e/lms/static/scripts/frontend_apps/components/FilePickerApp.js#L66

to latter assert an api call is made based on this conditional:

https://github.com/hypothesis/lms/blob/0fb537ecdf423d414a5be470cf76a1d55879bd9e/lms/static/scripts/frontend_apps/components/FilePickerApp.js#L117

but I don't think I'm setting it right, I might be typo I can't spot or something more fundamentally. 

The rest of the test might be wrong as well but I haven't got as far as I'm doing something wrong on the setup part.

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -104,6 +114,81 @@ describe('FilePickerApp', () => {
...
+    });
+
+    it('calls backend api when content is selected', async () => {
+      const onSubmit = sinon.stub().callsFake(e => e.preventDefault());
+      const wrapper = renderFilePicker({ onSubmit });
+
+      selectContent(wrapper, 'https://example.com');
+
+      await waitFor(() => fakeAPICall.called);
+      await delay(100);
```
marcospri, line 148: This is the last change since you last looked at it @robertknight 

Not sure if necessary of a symptom of something else that's missing/wrong.

robertknight (approved):

PR #3127, opened 2021-09-20
Use DB configured canvas assignments in launches

marcospri:


lms/resources/_js_config/__init__.py
```diff
@@ -87,6 +87,14 @@ def add_document_url(self, document_url):
                     _query={"document_url": document_url},
                 ),
             }
+
+        elif document_url.startswith("canvas://"):
+            self.add_canvas_file_id(
+                self._request.params["custom_canvas_course_id"],
+                self._request.params["resource_link_id"],
+                self._request.params["file_id"],
```
marcospri, line 95: Note that after all of this we are still using the param from the URL and not from document_url. 

I think we should do that on a separate URL and refactor `canvas_api.files.via_url` view to take the full URL instead of the separate parts, as the blackboard one does.

lms/views/basic_lti_launch.py
```diff
@@ -54,8 +54,7 @@ def basic_lti_launch(self, document_url=None, grading_supported=True):
         if grading_supported:
             self.context.js_config.maybe_enable_grading()
 
-        if document_url is not None:
```
marcospri, line 57: Was only called with none from `canvas_file_basic_lti_launch`. Not we pass the URL always.

lms/views/basic_lti_launch.py
```diff
@@ -94,38 +93,40 @@ def store_lti_data(self):
...
             # make sure this URL is meaningful.
-            document_url=f"canvas://file/course/{course_id}/file_id/{file_id}",
+            document_url=document_url,
             tool_consumer_instance_guid=self.request.params[
                 "tool_consumer_instance_guid"
             ],
             resource_link_id=resource_link_id,
         )
-
-        self.context.js_config.add_canvas_file_id(course_id, resource_link_id, file_id)
```
seanh, line 126: I think there's a refactoring here where you've removed the call to `add_canvas_file_id()` from this view, and moved it into `JSConfig` instead (the new `elif document_url.startswith("canvas://")` case). I can see that potentially making sense because `JSConfig` already contains a case for `if document_url.startswith("blackboard://")`. But would you mind sending this as a separate PR first? I think it's a logically separate refactoring. This stuff is quite hard to wrap your head around so I think one change at a time is going to make it easier for me to review

seanh (changes requested):

lms/views/basic_lti_launch.py
```diff
@@ -166,6 +168,50 @@ def db_configured_basic_lti_launch(self):
...
+            #
+            # This happens because legacy code used to store Canvas assignments in the DB with a
+            # resource_link_id and no ext_lti_assignment_id, see https://github.com/hypothesis/lms/pull/2780
+            #
+            # Whereas current code stores Canvas assignments during content-item-selection with an
+            # ext_lti_assignment_id and no resource_link_id.
+            #
+            # We need to merge the two assignments into one.
+
+            # order is guaranteed by the query's `order by`
```
seanh, line 194:
```suggestion
```

I think we should remove the comment as it's unclear that _the query_ refers to (since it refers to a SQLAlchemy query in another file). It's part of `get_for_canvas_launch()`'s contract/docstring that if it returns two assignments it'll be the old one first: https://github.com/hypothesis/lms/blob/a6b7e551289181a9acb862b29aa9b4eef055f2ee/lms/services/assignment.py#L74

seanh (approved):

PR #3132, opened 2021-09-20
Rename Object to object

esanzgar:
For the past few months, we have started to use the `object` type instead of `Object`. Both are almost interchangeable, but we prefer `object` for simplicity.

This PR substitutes all `Object` to `object` and `Object.<...>` to `Record<...>` and makes the code more consistent.

* First commit is an automatic transformation
* Second commit is a manual fix of a few types

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
@@ -1,7 +1,7 @@
 /**
  * Generate a `mailto:` URL that prompts to send an email with pre-filled fields.
  *
- * @param {Object} args
+ * @param {object} args
  * @param {string} args.address - Email address to sent to
  * @param {string} [args.subject] - Pre-filled subject line
  * @param {string} [args.body] - Pre-filled body
@@ -26,7 +26,7 @@ function toSentence(str) {
  *
  * @typedef ErrorLike
  * @prop {string} [message]
- * @prop {Object} [details] - Optional JSON-serializable details of the error
+ * @prop {object} [details] - Optional JSON-serializable details of the error
  */
 
 /**
--- a/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDisplay.js
@@ -26,7 +26,7 @@ function toSentence(str) {
  *
  * @typedef ErrorLike
  * @prop {string} [message]
- * @prop {object} [details] - Optional JSON-serializable details of the error
+ * @prop {object|string} [details] - Optional JSON-serializable details of the error
  */
 
 /**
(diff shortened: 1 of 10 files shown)


lms/static/scripts/frontend_apps/components/ErrorDisplay.js
```diff
@@ -26,7 +26,7 @@ function toSentence(str) {
  *
  * @typedef ErrorLike
  * @prop {string} [message]
- * @prop {Object} [details] - Optional JSON-serializable details of the error
+ * @prop {object} [details] - Optional JSON-serializable details of the error
```
robertknight, line 29: From reading the code below it looks to me like this should be `object|string`.

lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -12,7 +12,7 @@ import Dialog from './Dialog';
  */
 export default function ErrorDialogApp() {
   const {
-    errorDialog: { errorCode, errorDetails = '' },
+    errorDialog: { errorCode, errorDetails = {} },
```
robertknight, line 15: Looking at the `ErrorDisplay` implementation, it looks like the previous default was correct (empty string) but the type definition of `ErrorLike` was incomplete.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -22,7 +22,7 @@ export default function OAuth2RedirectErrorApp({ location = window.location }) {
     OAuth2RedirectError: {
       authUrl = /** @type {string|null} */ (null),
       errorCode = /** @type {string|null} */ (null),
-      errorDetails = '',
+      errorDetails = {},
```
robertknight, line 25: The previous value was correct. `ErrorLike` had an incomplete type.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -112,19 +112,19 @@ import { createContext } from 'preact';
  * The documentation for these properties lives in the `JSConfig` class in
  * the backend code.
  *
- * @typedef ConfigObject
+ * @typedef Config
```
robertknight, line 115: Can you revert this change. This type intentionally has an `Object` suffix to disambiguate it from the `Config` variable that this module exports.

robertknight (changes requested):
Most of this looks good but I found a couple of issues related to the `details` property of `ErrorLike` objects and the renaming of `ConfigObject` => `Config`.

lms/static/scripts/frontend_apps/components/ErrorDisplay.js
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/config.js
esanzgar: In my opinion, having a variable and  type with the same name is not a bad thing. Nevertheless, I have reverted.

robertknight (approved):

lms/static/scripts/frontend_apps/config.js
robertknight: > In my opinion, having a variable and type with the same name is not a bad thing. 

I suppose this does happen with classes already, as you get both the class constructor value and the type derived from it. In that the case though the relationship between the type (constructor) and the value (instance type) is very obvious. In this case its a bit different than normal (the `Config` value is React context with a type of `ConfigObject`).

PR #3133, opened 2021-09-21
Load MS OneDrive client

esanzgar:
Added a function that creates the `<script>` element to load the
OneDrive client.

The `OneDrive` type has been globally added to the `window` similarly to
what happens with `window.gapi`.

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/test/onedrive-api-client-test.js
+++ b/lms/static/scripts/frontend_apps/utils/test/onedrive-api-client-test.js
@@ -1,0 +1,68 @@
+import { loadOneDriveAPI } from '../onedrive-api-client';
+
+describe('loadOneDriveAPI', () => {
+  let fakeScriptElement;
+
+  beforeEach(() => {
+    // Prevent the tests from making a real network request to load the real
+    // script.
+    fakeScriptElement = document.createElement('fake-script');
+    sinon.stub(document, 'createElement').returns(fakeScriptElement);
+    delete window.OneDrive;
+  });
+
+  afterEach(() => {
+    document.createElement.restore();
+    fakeScriptElement.remove();
+    delete window.OneDrive;
+  });
+
+  it('resolves if OneDrive script loads', async () => {
+    const onLoad = loadOneDriveAPI();
+    const options = { option: 'dummy' };
+
+    // Simulate successful script loading.
+    window.OneDrive = {
+      open: sinon.stub(),
+    };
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/index.d.ts
```diff
@@ -0,0 +1,9 @@
+export {};
```
robertknight, line 1: Is this needed for a `.d.ts` file?

Also, was there a particular reason for the `index.d.ts` file name? I'm not certain but I think this might override the type of `index.js`, which we don't want. I'd prefer to give it a name like say `types.d.ts` or `onedrive.d.ts` to make it obvious that the types here are not related to the `index.js` file.

robertknight (commented):
I had a few queries, mainly about the `index.d.ts` file name and whether there are official types for it. Otherwise just a few minor comments.

lms/static/scripts/frontend_apps/utils/test/onedrive-api-client-test.js
```diff
@@ -0,0 +1,67 @@
...
+    delete window.OneDrive;
+  });
+
+  afterEach(() => {
+    document.createElement.restore();
+    fakeScriptElement.remove();
+    delete window.OneDrive;
+  });
+
+  it('successfully loads OneDrive', async () => {
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/index.d.ts
esanzgar: Yes, I agree.

esanzgar:
I included your feedback on the last commit.

lms/static/scripts/frontend_apps/utils/test/onedrive-api-client-test.js
```diff
@@ -0,0 +1,68 @@
...
+    fakeScriptElement.dispatchEvent(new Event('load'));
+    const oneDrive = await onLoad;
+    oneDrive.open(options);
+
+    assert.exists(fakeScriptElement.src);
+    assert.calledOnce(window.OneDrive.open);
+    assert.calledWith(window.OneDrive.open, options);
+  });
+
+  it('throws an error if OneDrive script fail to load', async () => {
```
robertknight, line 37:
```suggestion
  it('throws an error if OneDrive script fails to load', async () => {
```

robertknight (approved):
LGTM.

PR #3134, opened 2021-09-22
Integrate OneDrive file picker

esanzgar:
I created the `OneDrivePickerClient` class that uses the
`loadOneDriveAPI` to load the OneDrive JS bundle asynchronously.

`OneDrivePcikerClient` has a single method, `showPicker` to launch the
OneDrive picker. It is a callback based method.

To test this PR:

* Login into http://localhost:8001/admin
* Find  the following consumer key: `Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a`
* Click on the `Microsoft OneDrive enabled` checkbox
* Click Save

Then, you should be able to see the `Select PDF from OneDrive` option when setting an assignment in LMS:

Screenshot 2021-09-22 at 14 52 03

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -185,8 +185,8 @@ export default function ContentSelector({
   }
 
   const showGooglePicker = async () => {
+    setLoadingIndicatorVisible(true);
     try {
-      setLoadingIndicatorVisible(true);
       const picker = /** @type {GooglePickerClient} */ (googlePicker);
       const { id, url } = await picker.showPicker();
       await picker.enablePublicViewing(id);
@@ -203,28 +203,22 @@ export default function ContentSelector({
     }
   };
 
-  const showOneDrivePicker = () => {
+  const showOneDrivePicker = async () => {
     setLoadingIndicatorVisible(true);
-    const picker = /** @type {OneDrivePickerClient} */ (oneDrivePicker);
-    /** @param {any} file */
-    const success = file => {
-      const sharingURL = file.value[0].permissions[0].link.webUrl;
-      const url = OneDrivePickerClient.encodeSharingURL(sharingURL);
+    try {
+      const picker = /** @type {OneDrivePickerClient} */ (oneDrivePicker);
+      const { url } = await picker.showPicker();
       onSelectContent({ type: 'url', url });
-    };
-    const cancel = () => {
+    } catch (error) {
       setLoadingIndicatorVisible(false);
-    };
-    /** @param {Error} error */
-    const error = error => {
-      setLoadingIndicatorVisible(false);
-      console.error(error);
-      onError({
-        title: 'There was a problem choosing a file from OneDrive',
-        error,
-      });
-    };
-    picker.showPicker({ success, cancel, error });
+      if (!(error instanceof PickerCanceledError)) {
+        console.error(error);
+        onError({
+          title: 'There was a problem choosing a file from OneDrive',
+          error,
+        });
+      }
+    }
   };
 
   return (
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -6,6 +6,8 @@ import {
   GooglePickerClient,
   PickerCanceledError,
 } from '../utils/google-picker-client';
+import { OneDrivePickerClient } from '../utils/onedrive-picker-client';
+
 import BookPicker from './BookPicker';
 import FullScreenSpinner from './FullScreenSpinner';
 import LMSFilePicker from './LMSFilePicker';
@@ -55,6 +57,11 @@ export default function ContentSelector({
         developerKey: googleDeveloperKey,
         origin: googleOrigin,
       },
+      microsoftOneDrive: {
+        enabled: oneDriveFilesEnabled,
+        clientId: oneDriveClientId,
+        redirectURI: oneDriveRedirectURI,
+      },
       vitalSource: { enabled: vitalSourceEnabled },
     },
   } = useContext(Config);
@@ -92,6 +99,20 @@ export default function ContentSelector({
       origin: window === window.top ? window.location.href : googleOrigin,
     });
   }, [googleDeveloperKey, googleClientId, googleOrigin]);
+
+  // Initialize the OneDrive client if credentials have been provided.
+  // We do this eagerly to make the picker load faster if the user does click
+  // on the "Select from OneDrive" button.
+  const oneDrivePicker = useMemo(() => {
+    if (!oneDriveClientId || !oneDriveRedirectURI) {
+      return null;
+    }
+
+    return new OneDrivePickerClient({
+      clientId: oneDriveClientId,
+      redirectURI: oneDriveRedirectURI,
+    });
+  }, [oneDriveClientId, oneDriveRedirectURI]);
 
   /** @param {string} url */
   const selectURL = url => {
@@ -182,6 +203,30 @@ export default function ContentSelector({
     }
   };
 
+  const showOneDrivePicker = () => {
+    setLoadingIndicatorVisible(true);
+    const picker = /** @type {OneDrivePickerClient} */ (oneDrivePicker);
+    /** @param {any} file */
+    const success = file => {
+      const sharingURL = file.value[0].permissions[0].link.webUrl;
+      const url = OneDrivePickerClient.encodeSharingURL(sharingURL);
+      onSelectContent({ type: 'url', url });
+    };
+    const cancel = () => {
+      setLoadingIndicatorVisible(false);
+    };
+    /** @param {Error} error */
+    const error = error => {
+      setLoadingIndicatorVisible(false);
+      console.error(error);
+      onError({
+        title: 'There was a problem choosing a file from OneDrive',
+        error,
+      });
+    };
+    picker.showPicker({ success, cancel, error });
+  };
+
   return (
     <>
       {isLoadingIndicatorVisible && <FullScreenSpinner />}
@@ -220,6 +265,15 @@ export default function ContentSelector({
               data-testid="google-drive-button"
             >
               Select PDF from Google Drive
+            </LabeledButton>
+          )}
+          {oneDriveFilesEnabled && (
+            <LabeledButton
+              onClick={showOneDrivePicker}
+              variant="primary"
+              data-testid="onedrive-button"
+            >
+              Select PDF from OneDrive
             </LabeledButton>
           )}
           {vitalSourceEnabled && (
--- a/lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
@@ -5,6 +5,7 @@ import { mount } from 'enzyme';
 import { act } from 'preact/test-utils';
 
 import mockImportedComponents from '../../../test-util/mock-imported-components';
+import { delay, waitFor } from '../../../test-util/wait';
 import { Config } from '../../config';
 import { PickerCanceledError } from '../../utils/google-picker-client';
 import ContentSelector, { $imports } from '../ContentSelector';
@@ -28,9 +29,6 @@ describe('ContentSelector', () => {
     FakeOneDrivePickerClient = sinon.stub().returns({
       showPicker: sinon.stub(),
     });
-    FakeOneDrivePickerClient.encodeSharingURL = sinon
-      .stub()
-      .callsFake(url => `https://api.onedrive.com/v1.0/shares/u!${url}`);
 
     fakeConfig = {
       api: { authToken: 'dummy-auth-token' },
@@ -79,11 +77,7 @@ describe('ContentSelector', () => {
     const noop = () => {};
     return mount(
       <Config.Provider value={fakeConfig}>
-        <ContentSelector
-          setErrorInfo={noop}
-          onSelectContent={noop}
-          {...props}
-        />
+        <ContentSelector onError={noop} onSelectContent={noop} {...props} />
       </Config.Provider>
     );
   };
@@ -374,6 +368,8 @@ describe('ContentSelector', () => {
   });
 
   describe('OneDrive picker', () => {
+    let picker;
+
     function clickOneDriveButton(wrapper) {
       const btn = wrapper.find('LabeledButton[data-testid="onedrive-button"]');
       interact(wrapper, () => {
@@ -382,14 +378,7 @@ describe('ContentSelector', () => {
     }
 
     beforeEach(() => {
-      const picker = FakeGooglePickerClient();
-      picker.showPicker.resolves({
-        id: 'doc1',
-        url: 'https://files.google.com/doc1',
-      });
-      picker.enablePublicViewing.resolves();
-      FakeGooglePickerClient.resetHistory();
-
+      picker = FakeOneDrivePickerClient();
       // Silence errors logged if showing OneDrive Picker fails.
       sinon.stub(console, 'error');
     });
@@ -399,8 +388,9 @@ describe('ContentSelector', () => {
     });
 
     it('skips initialization of OneDrive Picker client if parameters not provided', () => {
+      FakeOneDrivePickerClient.resetHistory();
       fakeConfig.filePicker.microsoftOneDrive = {
-        enabled: true,
+        enabled: true, // clientId and redirectURI are undefined
       };
       renderContentSelector();
 
@@ -408,9 +398,7 @@ describe('ContentSelector', () => {
     });
 
     it("doesn't show the OneDrive button if option is disabled", () => {
-      fakeConfig.filePicker.microsoftOneDrive = {
-        enabled: false,
-      };
+      fakeConfig.filePicker.microsoftOneDrive.enabled = false; // clientId and redirectURI are defined
       const wrapper = renderContentSelector();
 
       assert.isFalse(
@@ -418,7 +406,7 @@ describe('ContentSelector', () => {
       );
     });
 
-    it('initializes of OneDrive Picker client if parameters are provided', () => {
+    it('initializes OneDrive Picker client if parameters are provided', () => {
       renderContentSelector();
 
       const { clientId, redirectURI } = fakeConfig.filePicker.microsoftOneDrive;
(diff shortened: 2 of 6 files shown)


lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -79,11 +79,7 @@ describe('ContentSelector', () => {
     const noop = () => {};
     return mount(
       <Config.Provider value={fakeConfig}>
-        <ContentSelector
-          setErrorInfo={noop}
-          onSelectContent={noop}
-          {...props}
-        />
+        <ContentSelector onError={noop} onSelectContent={noop} {...props} />
```
esanzgar, line 82: I found that the the `setErrorInfo` prop was incorrect. Currently, the prop is called `onError`.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -63,18 +76,17 @@ describe('ContentSelector', () => {
...
       </Config.Provider>
     );
   };
 
   const isLoadingIndicatorVisible = wrapper =>
     wrapper.exists('FullScreenSpinner');
 
+  const onNextExecutionCycle = () =>
```
robertknight, line 87: We have a function called `delay` which is used in quite a few tests in the client which is basically the same as this, but with a configurable delay. `delay(0)` is quite a common idiom to wait one macro-task.

The term "execution cycle" is a not completely clear. Something like "event loop term" would be more specific, as it differentiates this from something like `Promise.resolve()` which waits only for the next microtask execution, whereas this waits for the next macrotask execution.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -352,6 +369,137 @@ describe('ContentSelector', () => {
...
+    });
+
+    afterEach(() => {
+      console.error.restore();
+    });
+
+    it('skips initialization of OneDrive Picker client if parameters not provided', () => {
+      fakeConfig.filePicker.microsoftOneDrive = {
+        enabled: true,
+      };
```
robertknight, line 399: To clarify, this doesn't initialize the picker because the parameters _other than enabled_ are missing?

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -352,6 +369,137 @@ describe('ContentSelector', () => {
...
+      renderContentSelector();
+
+      assert.notCalled(FakeOneDrivePickerClient);
+    });
+
+    it("doesn't show the OneDrive button if option is disabled", () => {
+      fakeConfig.filePicker.microsoftOneDrive = {
+        enabled: false,
```
robertknight, line 407: You might want to include the parameters here to prove that the option is not shown because of the `enabled` value, rather than because the other parameters are missing. You could also consider combining this with the previous test as part of a parametrized test.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,70 @@
...
+export class OneDrivePickerClient {
+  /**
+   * @param {object} options
+   * @param {string} options.clientId
+   * @param {string} options.redirectURI - the URL that is used to launch the picker
+   */
+  constructor({ clientId, redirectURI }) {
+    loadOneDriveAPI();
+
+    this._filePickerOptions = {
```
robertknight, line 13: This value is only used in one method and its structure is very specific to that use case. I would prefer this to be declared as a local there, with only the minimal amount of state (the `clientId` and `redirectURI` properties) kept as instance fields on the class.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,70 @@
...
+    };
+  }
+
+  /**
+   * Opens the OneDrive picker
+   *
+   * @returns Promise<{url: string}> - the URL of the file
+   * @throws {Error} - there are two types of errors:
+   *   - PickerCancelledError: raise when the user cancels the OneDrive dialog picker
+   *   - Other errors: raise when (1) the OneDrive client fails to load (`window.OneDrive` is
```
robertknight, line 32:
```suggestion
   *   - Other errors: raised when (1) the OneDrive client fails to load (`window.OneDrive` is
```

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,70 @@
...
+  async showPicker() {
+    return new Promise((resolve, reject) => {
+      const success = (/** @type {any} */ file) => {
+        const sharingURL = file.value[0].permissions[0].link.webUrl;
+        const url = OneDrivePickerClient.encodeSharingURL(sharingURL);
+        resolve({ url });
+      };
+      const cancel = () => reject(new PickerCanceledError());
+      const error = (/** @type {Error} */ error) => reject(error);
+      window.OneDrive.open({
```
robertknight, line 44: This call could fail if the OneDrive script has not completed loading before `showPicker` is called. Also I think it makes sense to consider the `window.OneDrive` global an implementation detail of the `loadOneDriveAPI` helper.

What I would suggest to do instead is to save the promise from `loadOneDriveAPI` in the constructor and then await the result here before calling the `open` method on it.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,70 @@
...
+   * 2. Convert the base64 encoded result to unpadded base64url format by
+   *   removing '=' characters from the end of the value, replacing '/'
+   *   with '_' and '+' with '-'
+   * 3. Append u! to be beginning of the string.
+   *
+   * @param {string} sharingURL
+   */
+  static encodeSharingURL(sharingURL) {
+    const b64SharedLink = btoa(sharingURL)
+      .replace(/=/g, '')
```
robertknight, line 64: The comment above says "(remove) '=' characters from the end of the value". Strictly speaking this removes "=" chars from anywhere in the value.  That's OK because the "=" padding char can't occur anywhere within the rest of a valid base64 value, assuming the standard base64 alphabet.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
+import { PickerCanceledError } from '../google-picker-client';
```
robertknight, line 1: If we're going to reuse this exception for both the Google and OneDrive picker we probably want to move it into a shared location in future.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
...
+      },
+    });
+    delete window.OneDrive;
+  });
+
+  afterEach(() => {
+    $imports.$restore();
+    delete window.OneDrive;
+  });
+  describe('#constructor', () => {
```
robertknight, line 30:
```suggestion

  describe('#constructor', () => {
```

Added a new line.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
...
+  });
+
+  describe('#showPicker', () => {
+    it('fails to invoke `window.OneDrive.open` if OneDrive client failed to load', async () => {
+      const oneDrive = createClient();
+      let expectedError;
+
+      // Emulate failure to load the OneDrive client:
+      // 1. an unhandled rejection raised in the constructor
+      // 2. window.OneDrive is undefined
```
robertknight, line 48: The connection between `loadOneDriveAPI` failing and `window.OneDrive` not being defined is implicit. This hazard would be avoided if `showPicker` used the result of `loadOneDriveAPI`. Additionally it would allow `showPicker` to return a more obvious error (eg. "Failed to load OneDrive" or whatever `loadOneDriveAPI` rejects with).

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
...
+
+  beforeEach(() => {
+    fakeLoadOneDriveAPI = sinon.stub();
+
+    $imports.$mock({
+      './onedrive-api-client': {
+        loadOneDriveAPI: fakeLoadOneDriveAPI,
+      },
+    });
+    delete window.OneDrive;
```
robertknight, line 23: If you implement my suggestion in in `onedrive-picker-client.js` to have this class not directly reference `window.OneDrive` then you can more easily mock the `window.OneDrive` interface in the tests by making `fakeLoadOneDriveAPI` return a stub.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
...
+      let expectedError;
+
+      // Emulate the successful loading of the OneDrive client and the picker
+      // erroring while the user interacting with it.
+      window.OneDrive = {
+        open: sinon.stub(),
+      };
+      const onError = oneDrive.showPicker();
+      const { error } = window.OneDrive.open.getCall(0).args[0];
+      error(oneDriveError);
```
robertknight, line 122: Is the argument to the `error` callback guaranteed to be an `Error` or is it some Error-like object?

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      assert.equal(oneDriveError, expectedError);
+    });
+  });
+
+  describe('#encodeSharingURL', () => {
+    it('calls `loadOneDriveAPI`', () => {
+      const url = OneDrivePickerClient.encodeSharingURL(
+        'https://1drv.ms/b/s!AmH'
```
robertknight, line 136: This needs some additional test cases to exercise the transformations that `encodeSharingURL` applies to the URL.

robertknight (changes requested):
The changes in `ContentSelector.js` look fine and straightforward. I found a race condition in the way that the OneDrive picker script is loaded and used in the `OneDrivePickerClient` class.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -352,6 +369,137 @@ describe('ContentSelector', () => {
...
+  describe('OneDrive picker', () => {
+    function clickOneDriveButton(wrapper) {
+      const btn = wrapper.find('LabeledButton[data-testid="onedrive-button"]');
+      interact(wrapper, () => {
+        btn.props().onClick();
+      });
+    }
+
+    beforeEach(() => {
+      const picker = FakeGooglePickerClient();
```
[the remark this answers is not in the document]
esanzgar: It is not needed. I copied and pasted from the Google Picker tests and I forgot to remove these lines.

Corrected, thanks.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
esanzgar: Correct.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
esanzgar: I have set `enable` false while leaving the `clientId` and `redirectURI` defined.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
esanzgar: OK, let me try it.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
esanzgar: As far as a I tested it is an `Error`.

lyzadanger (commented):
I wanted to leave some notes about some functional testing. The flow works as I'd generally expect it to. I was able during testing to get myself into a state where Canvas wouldn't display anything in the assignment-configuration modal dialog (it was an empty screen), with no errors in the console. After I did a bit of resetting of branch and dev server, I cannot reproduce this again, so I don't know that it was meaningful. If I see it again, I will definitely collect more information. There were a number of factors at play so it may have been entirely a fluke.

I was able to test this in Blackboard, also, once I tracked down the right application_instance.

A couple of notes:

* The picker re-auths every single time. I don't know if there's anything to be done about that; the user always sees this as a first step:

Screen Shot 2021-09-28 at 9 59 02 AM

* Selecting a document always results in the same (kind of confusing) URL shown to the user

Screen Shot 2021-09-28 at 9 32 55 AM

Can we do anything about that?

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
esanzgar: This is a great suggestion, I have implemented in an added commit.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -352,6 +369,137 @@ describe('ContentSelector', () => {
...
+    it('hides loading indicator if user cancels picker', async () => {
+      const onError = sinon.stub();
+      const wrapper = renderContentSelector({ onError });
+      const picker = FakeOneDrivePickerClient();
+      // Emulates the cancellation of the picker.
+      picker.showPicker.rejects(new PickerCanceledError());
+
+      clickOneDriveButton(wrapper);
+      await onNextExecutionCycle();
+      wrapper.update();
```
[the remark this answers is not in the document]
esanzgar: That's very useful, thanks! I implemented it on the second commit.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -72,7 +72,11 @@ import { createContext } from 'preact';
  *   @prop {string} google.clientId
  *   @prop {string} google.developerKey
  *   @prop {string} google.origin
- * @prop {object} vitalSource
+ * @prop {object} microsoftOneDrive
+ *   @prop {boolean} microsoftOneDrive.enabled
```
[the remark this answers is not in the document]
esanzgar: I agree.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
esanzgar: Implemented in the last commit.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
esanzgar: Yes, I realised that too, but I thought the same.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
esanzgar: Agree.

robertknight (commented):
This looks good overall, just a few minor comments.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -182,6 +203,24 @@ export default function ContentSelector({
     }
   };
 
+  const showOneDrivePicker = async () => {
+    setLoadingIndicatorVisible(true);
```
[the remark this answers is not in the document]
esanzgar: Sure, I will change that in a follow up PR.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -352,6 +367,130 @@ describe('ContentSelector', () => {
...
+      // Emulates the cancellation of the picker.
+      picker.showPicker.rejects(new PickerCanceledError());
+
+      clickOneDriveButton(wrapper);
+      await waitFor(() => {
+        wrapper.update();
+        return isLoadingIndicatorVisible(wrapper) === false;
+      });
+
+      assert.isFalse(isLoadingIndicatorVisible(wrapper));
```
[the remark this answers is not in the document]
esanzgar: Removed.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,73 @@
+import { PickerCanceledError } from './google-picker-client';
+import { loadOneDriveAPI } from './onedrive-api-client';
+
+export class OneDrivePickerClient {
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,144 @@
...
+
+      assert.equal(expectedError, errorLoading);
+    });
+
+    it('resolves with a URL when a file is selected', async () => {
+      fakeLoadOneDriveAPI.resolves(fakeOneDrive);
+      const oneDrive = createClient();
+
+      // Emulates the user selecting a file.
+      const onSuccess = oneDrive.showPicker();
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,144 @@
...
+        expectedResult:
+          'https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvP3g9JUQxJTg4JUQwJUI1/root/content',
+      },
+      {
+        sharingURL: 'https://1drv.ms/b/s!a%20a',
+        expectedResult:
+          'https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvcyFhJTIwYQ/root/content',
+      },
+    ].forEach(({ sharingURL, expectedResult }) => {
+      it('calls `loadOneDriveAPI`', () => {
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -0,0 +1,144 @@
...
+        expectedError = error;
+      }
+
+      assert.equal(expectedError, oneDriveError);
+    });
+  });
+
+  describe('#encodeSharingURL', () => {
+    [
+      {
```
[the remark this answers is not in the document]
esanzgar: 👍🏼

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -0,0 +1,73 @@
...
+   * @throws {Error} - there are two types of errors:
+   *   - PickerCancelledError: raised when the user cancels the OneDrive dialog picker
+   *   - Error: raised when (1) the OneDrive client fails to load, or when (2)
+   *       the picker fails with internal or backend problem.
+   */
+  async showPicker() {
+    const oneDrive = await this._onLoadOneDriveAPI;
+    return new Promise((resolve, reject) => {
+      const success = (/** @type {any} */ file) => {
+        const sharingURL = file.value[0].permissions[0].link.webUrl;
```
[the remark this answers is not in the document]
esanzgar: So far, this is what I have concluded:

* The OneDrive picker or REST API doesn't provide a built-in method to encode the URL
* The `permission` array returned by the picker doesn't fully conform with the `permission` returned by the REST API (https://docs.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0)
* Because there is not documentation about the `file` structure returned as argument to the `success` callback , I can't guaranty `webUrl` be always available.  Under normal circumstances, `webUrl` seems to be there.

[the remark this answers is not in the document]
robertknight: > Because there is not documentation about the file structure returned as argument to the success callback 

I see documentation at the URL linked in the JSDoc comment for the class. https://docs.microsoft.com/en-us/onedrive/developer/controls/file-pickers/js-v72/open-file?view=odsp-graph-online says:

> When the user is done picking file(s), the success callback receives response object. This object contains properties, include value property which is a collection of Item resource with a subset of the item's properties.

Where _item resource_ is a link to https://docs.microsoft.com/en-us/onedrive/developer/rest-api/resources/driveitem?view=odsp-graph-online. That page's description of the `webUrl` property implies that it is always present as, unlike other properties, it does not mention conditions under which it is present.

lms/static/scripts/frontend_apps/utils/test/onedrive-picker-client-test.js
```diff
@@ -116,25 +116,28 @@ describe('OneDrivePickerClient', () => {
...
-  describe('#encodeSharingURL', () => {
+  describe('.encodeSharingURL', () => {
     [
       {
-        sharingURL: 'https://1drv.ms/b/s!AmH',
+        sharingURL: atob('aHR0cHM6Ly8xZHJ2Lm1zL2IvaQ=='),
         expectedResult:
-          'https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvcyFBbUg/root/content',
+          'https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvaQ/root/content',
+        reason: 'removing the trailing `=` characters ',
```
robertknight, line 125: I'd suggest adding "in the base64 representation" to each of these reason strings, just to make it more obvious at a glance that this is not talking about a transformation of the input URL directly.

robertknight (approved):
Code changes LGTM. I'll leave it with you to agree any functional changes with Lyza.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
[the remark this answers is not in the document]
esanzgar: About the `permission` collection it states that:

> The set of permissions for the item. Read-only. Nullable.

So I understand it can be an `Array` or `null`.

PR #3138, opened 2021-09-23
Minor improvements in google-*-client-test.js

esanzgar:
While working on the OneDrive support I modelled my tests on the google
drive tests and I noticed a few minor issues:

* google-api-client-test.js leaves some HTML elements in the body

* google-picker-client-test.js is not restoring the mocked imports

* an unnecessary async

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/test/google-api-client-test.js
+++ b/lms/static/scripts/frontend_apps/utils/test/google-api-client-test.js
@@ -13,6 +13,7 @@ describe('loadLibraries', () => {
 
   afterEach(() => {
     document.createElement.restore();
+    dummyScript.remove();
   });
 
   it('loads API script', async () => {
--- a/lms/static/scripts/frontend_apps/utils/test/google-picker-client-test.js
+++ b/lms/static/scripts/frontend_apps/utils/test/google-picker-client-test.js
@@ -113,8 +113,12 @@ describe('GooglePickerClient', () => {
     });
   });
 
+  afterEach(() => {
+    $imports.$restore();
+  });
+
   describe('#constructor', () => {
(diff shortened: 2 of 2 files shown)


lms/static/scripts/frontend_apps/utils/test/google-api-client-test.js
```diff
@@ -13,6 +13,7 @@ describe('loadLibraries', () => {
 
   afterEach(() => {
     document.createElement.restore();
+    dummyScript.remove();
```
robertknight, line 16: This script doesn't get appended to the document, so calling `remove` has no effect. The browser just garbage-collects the DOM element when it is no longer referenced. Calling `remove` doesn't do any harm though.

lms/static/scripts/frontend_apps/utils/test/google-picker-client-test.js
```diff
@@ -113,8 +113,12 @@ describe('GooglePickerClient', () => {
     });
   });
 
+  afterEach(() => {
+    $imports.$restore();
```
robertknight, line 117: Good spot. A flaw in the current `$imports` API is that this mistake is easy to make.

robertknight (approved):
LGTM. The `dummyScript.remove()` call is not necessary currently, but it doesn't hurt either.

lms/static/scripts/frontend_apps/utils/test/google-api-client-test.js
esanzgar: It does get appended:
https://github.com/hypothesis/lms/blob/bd75cd47b757557a887db796da55d969fd48a5e8/lms/static/scripts/frontend_apps/utils/google-api-client.js#L18

PR #3141, opened 2021-09-27
Simplify tests for `ContentSelector-test.js`

esanzgar:
I use the `delay` function to simplify a few tests. Previously,
execution of `await FakeGooglePickerClient().showPicker()` had the same
effect (waiting for the macrotask execution) but it had other
consequences.

I removed the `FakeGooglePickerClient.resetHistory` line because it was
unnecessary: on every test, a new `sinon.stub` is assigned to the
FakeGooglePickerClient, and FakeGooglePickerClient is instantiated only
once per test.


lyzadanger (approved):
This is certainly more readable. For each of these `await delay(0)`, I presume the intent is essentially "next tick"?

esanzgar:
> This is certainly more readable. For each of these `await delay(0)`, I presume the intent is essentially "next tick"?

Correct.

PR #3149, opened 2021-09-27
Log refresh token changes

marcospri:
We seem to be prompting the users to complete an oauth more often than before in canvas: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1632415274068500

This should only happen if the refresh token is no longer valid which seem to happen from time to time: https://my.papertrailapp.com/groups/14104911/events?q=refresh_token+not+found

A refresh token could be no longer valid if canvas decides that's the case or, if canvas sent us a new one and we still used the old one.

The refresh token changing should to not occur according to Canvas's docs: `The response to this request will not contain a new refresh token; the same refresh token is to be reused.`  https://canvas.instructure.com/doc/api/file.oauth.html but I'd like to log that to confirm it.

If refresh tokens did in fact change we could be hitting a bug where two almost simultaneous API request request and new access_token, one of them getting a new refresh_token while the other one uses the one not yet updated refresh token present in the DB.

### DIFF
--- a/tests/unit/lms/services/oauth2_token_test.py
+++ b/tests/unit/lms/services/oauth2_token_test.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import datetime
 from unittest import mock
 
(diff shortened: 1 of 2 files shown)


tests/unit/lms/services/oauth2_token_test.py
```diff
@@ -79,6 +105,10 @@ def svc(self, pyramid_request):
             pyramid_request.lti_user.user_id,
         )
 
+    @pytest.fixture
+    def logger(self, patch):
+        return patch("lms.services.oauth2_token.logger")
```
seanh, line 110: You could use pytest's builtin `caplog` fixture instead of patching the logger: https://docs.pytest.org/en/stable/logging.html?highlight=caplog#caplog-fixture

seanh (approved):

PR #3151, opened 2021-09-27
Split `services.ConsumerKeyError` into two

seanh:
**Problem:** `services.ConsumerKeyError` is being abused for two different purposes.  `ConsumerKeyError` is a subclass of `LTILaunchVerificationError` and as such it's *supposed* to be raised by the `LTILaunchVerifier` service when LTI launch verification fails. But `ConsumerKeyError` is *also* being raised by `ApplicationInstanceService.get(consumer_key)` when there's no `ApplicationInstance` for the given `consumer_key` in the DB. This second case
is not an LTI launch verification error and is not an appropriate use of an `LTILaunchVerificationError` subclass.

**Solution:** split `services.ConsumerKeyError` into two separate classes:

1. `services.ConsumerKeyLaunchVerificationError(LTILaunchVerificationError)`:

   This is raised by the `LTILaunchVerifier` service when LTI launch verification fails because the consumer key given in the request was not found in our database.  It's a subclass of `LTILaunchVerificationError` which is the base class for all LTI launch verification failures raised by `LTILaunchVerifier`.

2. `services.ApplicationInstanceNotFound(Exception)`:

   This is raised by `ApplicationInstanceService.get(consumer_key)` when there's no `ApplicationInstance` with the requested `consumer_key` in the DB.

   Lots of methods throughout the codebase indirectly raise `ApplicationInstanceNotFound` (because they call `ApplicationInstanceService.get()` and don't catch it) and lots of methods catch `ApplicationInstanceNotFound`.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -1,21 +1,3 @@
-class LTILaunchVerificationError(Exception):
-    """
-    Raised when LTI launch request verification fails.
-
-    This is the base class for all LTI launch request verification errors.
-    Different subclasses of this exception class are raised for specific
-    failure types.
-    """
-
-
-class ConsumerKeyError(LTILaunchVerificationError):
-    """Raised when a given ``consumer_key`` doesn't exist in the DB."""
-
-
-class LTIOAuthError(LTILaunchVerificationError):
-    """Raised when OAuth signature verification of a launch request fails."""
-
-
 class ExternalRequestError(Exception):
     """
     A problem with a network request to an external service.
--- a/lms/services/launch_verifier.py
+++ b/lms/services/launch_verifier.py
@@ -15,7 +15,7 @@ class LTILaunchVerificationError(Exception):
 
 
 class ConsumerKeyLaunchVerificationError(LTILaunchVerificationError):
-    """Raised when a given ``consumer_key`` doesn't exist in the DB."""
+    """Raised when the request's consumer_key doesn't exist in the DB."""
 
 
 class LTIOAuthError(LTILaunchVerificationError):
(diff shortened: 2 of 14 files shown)


lms/services/exceptions.py
```diff
@@ -12,7 +16,7 @@ class LTILaunchVerificationError(ServiceError):
     """
 
 
-class ConsumerKeyError(LTILaunchVerificationError):
+class ConsumerKeyLaunchVerificationError(LTILaunchVerificationError):
     """Raised when a given ``consumer_key`` doesn't exist in the DB."""
```
jon-betts, line 20: Should this mention that the `consumer_key` in this case is from LTI params to differentiate it from the new exception?

lms/services/launch_verifier.py
```diff
@@ -30,9 +34,9 @@ def verify(self):
         :raise NoConsumerKey: If the request has no ``oauth_consumer_key``
           parameter (maybe it's not an LTI launch request at all?)
 
-        :raise ConsumerKeyError: If the request's ``oauth_consumer_key``
-          parameter isn't found in our database (this appears to be an invalid
-          LTI launch request).
+        :raise ConsumerKeyLaunchVerificationError: If the request's
+          ``oauth_consumer_key`` parameter isn't found in our database (this
```
jon-betts, line 38:
```suggestion
          `oauth_consumer_key` parameter isn't found in our database (this
```

I think we're moving away from the restructured text stuff?

jon-betts (approved):
Yep this looks good. The new names for the exceptions are both clearer than the one they replaced and distinguish between the two.

I think the docstring is still as it was before though, and could benefit from being a bit clearer.

PR #3152, opened 2021-09-27
Move exceptions into the files that raise them

seanh:
These exception classes are actually only ever raised (and only intended to be raised) by these particular services, so move them into the same files with the service classes that they belong to. This discourages reused but that's the intention here, and doing this makes the intention clearer.

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -1,21 +1,23 @@
+from lms.services.application_instance import ApplicationInstanceNotFound
 from lms.services.canvas import CanvasService
 from lms.services.exceptions import (
-    ApplicationInstanceNotFound,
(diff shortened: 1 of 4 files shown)


jon-betts (approved):
I'm a big fan of this. If we don't (and don't want to) raise these from anywhere else, it seems counter intuitive to keep them anywhere else other than the code that raises it.

In general I think one of the bigger issues in understanding our code is due to non-localism like this and anything that can cut down on that is a good move :+1:.

It can also lead to exactly the confusion fixed on one of the previous PRs. Without this all the exceptions get lumped together you can't tell what is a general exception you _should_ re-use, and which are specific to certain places and you shouldn't.

PR #3153, opened 2021-09-27
Move `HAPIError` into the file that raises it

seanh:
`HAPIError` is only ever raised (and only ever intended to be raised) by the `HAPI` service so move it into the same file with `HAPI` service. This discourages reuse but that's the intention here, and doing this makes the intention clearer.

A downside of this change is that it's now harder to see all the `ExternalRequestError` subclasses (they're no longer all in one file).


jon-betts (approved):
More localism. More goodism.

PR #3155, opened 2021-09-27
Don't wrap `HAPIError` in `HTTPInternalServerError`

seanh:
I'm not sure why the code does this: `LTIHService` doesn't need to raise an `HAPIError` wrapped in an `HTTPInternalServerError`. It can just raise the `HAPIError` directly and the exception view can catch `HAPIError`.

This is easier to follow because you can just see that the `HAPI` service raises `HAPIError` and the exception view catches `HAPIError` and you can see what the exception view does with that error. Rather than having to know that the `HAPI` service raises `HAPIError` which `LTIHService` catches and turns into an `HTTPInternalServerError` which
is then caught by the exception view for `HTTPServerError` (a parent class of `HTTPInternalServerError`).

Note that the exception view previously caught `HTTPServerError` and all its subclasses, which is all the Pyramid 5xx `httpexception`'s. But I think the `HAPIError` wrapped in an `HTTPInternalServerError` was the only case where any `HTTPServerError` was actually being raised. So I don't think there's any unintended change in behaviour here.

### DIFF
--- a/lms/views/exceptions.py
+++ b/lms/views/exceptions.py
@@ -1,6 +1,7 @@
 import h_pyramid_sentry
-from pyramid import httpexceptions, i18n
+from pyramid import i18n
 from pyramid.config import not_
+from pyramid.httpexceptions import HTTPClientError
 from pyramid.view import (
     exception_view_config,
     forbidden_view_config,
@@ -8,6 +9,7 @@ from pyramid.view import (
 )
 
 from lms.models import ReusedConsumerKey
+from lms.services import HAPIError
 from lms.validation import ValidationError
 
 _ = i18n.TranslationStringFactory(__package__)
@@ -33,21 +35,17 @@ def forbidden(_exc, request):
     return {"message": _("You're not authorized to view this page")}
 
 
-@exception_view_config(
-    context=httpexceptions.HTTPClientError, renderer=DEFAULT_RENDERER
-)
+@exception_view_config(context=HTTPClientError, renderer=DEFAULT_RENDERER)
 def http_client_error(exc, request):
     """Handle an HTTP 4xx (client error) exception."""
     return _http_error(exc, request)
 
 
-@exception_view_config(
-    context=httpexceptions.HTTPServerError, renderer=DEFAULT_RENDERER
-)
-def http_server_error(exc, request):
-    """Handle an HTTP 5xx (server error) exception."""
+@exception_view_config(context=HAPIError, renderer=DEFAULT_RENDERER)
+def hapi_error(exc, request):
     h_pyramid_sentry.report_exception()
-    return _http_error(exc, request)
+    request.response.status_int = 500
+    return {"message": str(exc)}
 
 
 @exception_view_config(
--- a/lms/services/lti_h.py
+++ b/lms/services/lti_h.py
@@ -1,7 +1,4 @@
 from h_api.bulk_api import CommandBuilder
-from pyramid.httpexceptions import HTTPInternalServerError
-
-from lms.services import HAPIError
 
 
 class LTIHService:
@@ -47,11 +44,7 @@ class LTIHService:
         if not self._application_instance_service.get().provisioning:
             return
 
-        try:
-            self._h_api.execute_bulk(commands=self._yield_commands(h_groups))
-
-        except HAPIError as err:
-            raise HTTPInternalServerError(explanation=err.explanation) from err
+        self._h_api.execute_bulk(commands=self._yield_commands(h_groups))
 
         # Keep a note of the groups locally for reporting purposes.
         for h_group in h_groups:
--- a/tests/unit/lms/services/lti_h_test.py
+++ b/tests/unit/lms/services/lti_h_test.py
@@ -2,7 +2,6 @@ from unittest.mock import create_autospec, sentinel
 
 import pytest
 from h_api.bulk_api import CommandBuilder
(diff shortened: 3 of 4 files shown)


lms/views/exceptions.py
```diff
@@ -33,21 +35,17 @@ def forbidden(_exc, request):
     return {"message": _("You're not authorized to view this page")}
 
 
-@exception_view_config(
-    context=httpexceptions.HTTPClientError, renderer=DEFAULT_RENDERER
-)
+@exception_view_config(context=HTTPClientError, renderer=DEFAULT_RENDERER)
```
jon-betts, line 38: :+1: I don't know why I find this so aggravating, but I do.

jon-betts (commented):
If I had to guess what's going on here I'd say it's something along the lines of:

* We already catch `HTTPClientError` errors en masse, so catching `HTTPInternalServerError` seems like the symmetrical partner to this
* We _probably_ didn't want to raise `HTTPInternalServerError` or a derivative in the service as we had the `ServiceError` hierarchy

Although this is the only route by which _our_ code will issue a `HTTPInternalServerError` is this true of Pyramid? This PR removes handling of Pyramid 500 derivatives as well. This seems like a change of behavior, as these will now presumably fall through to the `error()` handler and have the message suppressed.

I'm not sure about catching the `HAPIError` exception specifically, as it looks like this could easily be folded into the `error()` general handler at the bottom that catches and returns 500's in general. Would a duck typing approach like the one here work for this? https://github.com/hypothesis/lms/pull/3154

The other question is, do we particularly need to go into the details? If we suppress the message for all other 500's maybe we don't need to differentiate between all these at all? And the solution is to just remove the specific handler entirely?

seanh:
> Although this is the only route by which our code will issue a `HTTPInternalServerError` is this true of Pyramid? This PR removes handling of Pyramid 500 derivatives as well. This seems like a change of behavior, as these will now presumably fall through to the `error()` handler and have the message suppressed.

I don't _think_ Pyramid itself ever raises `HTTPServerError` so I don't think there's a change of behaviour here. I'm fairly sure that `HAPIError` is the only exception that this exception view ever catches. Though it'd be difficult to comprehensively prove this.

Even if Pyramid does raise `HTTPServerError` itself I don't think it was our intention that `HTTPServerError`'s raised by Pyramid should be stringified and shown to the user, rather than showing the fixed error message that we use for general unexpected `Exception`'s. I think showing the `str()` of the exception was intended for `HTTPServerError`'s raised by our own code, of which the `HAPIError`-wrapped-in-an-`HTTPInternalServerError` was the only remaining example. So I don't think we'd be removing an intentional behaviour by removing the stringification of `HTTPServerError`'s.

> I'm not sure about catching the `HAPIError` exception specifically, as it looks like this could easily be folded into the `error()` general handler at the bottom that catches and returns 500's in general. Would a duck typing approach like the one here work for this? #3154

As far as I can tell the `http_server_error()` view is already an exception view that only catches `HAPIError`'s specifically. This PR is just simplifying things by removing the unnecessary wrapping of `HAPIError` in `HTTPInternalServerError` and changing the exception view to explicitly catch `HAPIError` only. I don't think this exception view was ever catching any other exceptions. So I'm gonna scope this PR to just that simplification and clarification.

I'm not sure about folding the `hapi_error()` exception view into the general `error()` one in a later PR. After some forthcoming further refactorings the two exception views will look like this:

```python
@view_defaults(renderer="lms:templates/error.html.jinja2")
class ExceptionViews:
    ...

    @exception_view_config(context=HAPIError)
    def hapi_error(self):
        return self.internal_server_error(str(self.exception))

    @exception_view_config(context=Exception)
    def error(self):
        return self.internal_server_error(_( "Sorry, but something went wrong... "))

    def internal_server_error(self, message):
        self.request.response.status_int = 500
        return { "message": message}
```

What would be the motivation for combining these into a single view?

The two separate exception views have the advantage of making it explicit that `HAPIError` in particular is being singled out for special treatment, which is what's actually intended in this case.

`HAPIError` differs from the rest of our `ExternalRequestError`'s in that it can be raised during both API and non-API requests so it needs exception views in both `views/api/exceptions.py` and `views/exceptions.py`. It's good for `views/exceptions.py` to contain explicit mention of `HAPIError` to clarify this.

A more duck-typing protocol-based approach might just hide these facts by removing explicit mention of `HAPIError`:

```python
@exception_view_config(context=Exception)
def error(self):
    self.request.response.status_int = 500
    return { "message": getattr(self.exception, "message", _( "Sorry, but something went wrong... "))}
```

I think this code implies that we have multiple different types of exception that implement the same protocol (a `message` attribute) that might pass through this exception view. But in fact we have only the `HAPIError` class specifically. So I wonder if applying duck-typing here might just be hiding things.

seanh:
> The other question is, do we particularly need to go into the details? If we suppress the message for all other 500's maybe we don't need to differentiate between all these at all? And the solution is to just remove the specific handler entirely?

:man_shrugging: Maybe. That'd be a change of behaviour though. Currently we show "Connecting to Hypothesis failed" for `HAPIError`'s whereas we show "Sorry, but something went wrong..." for general unexpected exceptions. The custom error message for the case when communicating with h fails doesn't seem _super_ important but I don't feel particularly against it either. It adds very little complexity once the code is tidied up. I'm inclined to just leave it as-is

jon-betts (approved):

PR #3156, opened 2021-09-27
Don't unnecessarily manually report things to Sentry

seanh:
If an exception is handled by an exception view then Sentry's Pyramid integration won't automatically report the exception *unless* the exception view results in a 500 response. Since this exception view does always result in a 500 there's no need for us to manually report the exception.

### DIFF
--- a/tests/unit/lms/views/exceptions_test.py
+++ b/tests/unit/lms/views/exceptions_test.py
@@ -16,7 +16,6 @@ class ExceptionViewTest:
 
     expected_result = None
     response_status = None
-    report_to_sentry = None
 
     def handle(self, pyramid_request):
         if self.exception is None:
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/exceptions_test.py
```diff
@@ -99,11 +83,6 @@ class TestError(ExceptionViewTest):
     exception = None
 
     response_status = 500
-    # Although I don't think it would do any harm (sentry_sdk seems smart
-    # enough to not double report the exception to Sentry) we don't need to
-    # call report_exception() in the case of a non-HTTPError exception
-    # because Sentry's Pyramid integration does it for us automatically.
```
seanh, line 105: This comment was incorrect. It's the `500` status that causes the exception to be automatically reported, not the exception class.

jon-betts:
There are some suggestions on https://github.com/hypothesis/lms/pull/3155 which might change the solution here.

jon-betts (approved):
So I think the changes in the tests are that all the cases have become the same case (no manual reporting).

Looks good!

PR #3158, opened 2021-09-28
Assignment service re-refactor

marcospri:
Refactor of the service to avoid doing the merge on get/exists/set_document and just doing it explicitly on the view.

It might look like too much ceremony over just returning 2 rows on get but I really dislike get returning possibly two elements (does it always return a list? or sometimes an element, sometimes a list?) and the possibility to even just do `assignment = svc.get()[0]` in the future and potentially run into the merging problem unknowingly.

I think this way `exists` and `get` behave like you would expect and we only deal with the legacy merging etc on the view. The view could be split in any way but I went for a predicate `is_canvas`.

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -140,11 +140,7 @@ class AssignmentService:
         assignment.document_url = document_url
         assignment.extra = self._update_extra(assignment.extra, extra)
 
-        # Clear the cache (@lru_cache) on self.get because we've changed the
-        # contents of the DB. (Python's @lru_cache doesn't have a way to remove
-        # just one key from the cache, you have to clear the entire cache.)
-        self.get.cache_clear()
-
+        self._clear_cache()
         return assignment
 
     def _get_by_resource_link_id(self, tool_consumer_instance_guid, resource_link_id):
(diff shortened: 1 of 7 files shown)


lms/services/assignment.py
```diff
@@ -91,18 +138,7 @@ def set_document_url(  # pylint:disable=too-many-arguments
...
-            assignment.extra = extra
-
-        if extra:
-            assignment.extra = extra
-
-        if extra:
-            assignment.extra = extra
-
-        if extra:
-            assignment.extra = extra
```
marcospri, line 105: Making really sure extra was set (comes from the rabases I assume)

lms/views/basic_lti_launch.py
```diff
@@ -166,6 +166,50 @@ def db_configured_basic_lti_launch(self):
         ).document_url
         return self.basic_lti_launch(document_url)
 
+    @view_config(db_configured=True, is_canvas=True)
```
marcospri, line 169: Did the split using a predicate here. Other alternatives work as well but I think predicates work nicely to route based on url params and similar.

seanh (approved):
I think the approach looks overall good :+1: I'll do some more thorough testing once it's finished and merged into the other PR.

`get()` is still setting `resource_link_id`'s and I don't think it should be.

I think you can replace the new predicate with just Pyramid's built-in `request_param` predicate and that would be better.

I think we should keep `get()` split up into one small(er) method for each individual case

lms/services/assignment.py
```diff
@@ -28,9 +28,22 @@ def get(
...
+                )
+
+            assignment = assignments[0]
+            if not assignment.resource_link_id:
+                # We found an assignment with an ext_lti_assignment_id but no resource_link_id.
+                # This happens the first time a new Canvas assignment is launched: the assignment got created
+                # during content-item-selection with an ext_lti_assignment_id but no resource_link_id,
+                # and then the first time the assignment is launched we add the resource_link_id.
+                assignment.resource_link_id = resource_link_id
+            return assignment
```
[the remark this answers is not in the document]
marcospri: This is a bit sorter now that the is not assigned here.

PR #3159, opened 2021-09-28
Remove an unnecessary helper function

seanh:
Because of recent changes (in https://github.com/hypothesis/lms/pull/3155/) this helper function is now only called in one place and its docstring is inaccurate.


jon-betts (approved):
Yap! Not much to say there.

PR #3162, opened 2021-09-29
Remove a pointless docstring

seanh:
It's pretty obvious that this method handles `HTTPClientError`'s:

```python
@exception_view_config(context=HTTPClientError)
def http_client_error(self):
    ...
```

So a docstring that says "Handle an HTTP 4xx (client error) exception" really isn't adding much (except possibly explaining that `HTTPClientError` = 4xx?).


marcospri (approved):

jon-betts (approved):
I think this needs some re-basing, but the actual removed comment seems fine.

PR #3163, opened 2021-09-29
Remove a pointless docstring

seanh:
It's clear that this method handles `ValidationError`'s:

```python
@exception_view_config(context=ValidationError, ...)
def validation_error(self):
    ...
```

A docstring saying "Handle a ValidationError" doesn't add anything.


marcospri (approved):
No testing notes but I'll take the risk.

PR #3164, opened 2021-09-29
Remove an unnecessary docstring

seanh:
I don't think this docstring is actually adding anything useful.


marcospri (approved):
Maybe close to `message` we could explicit say that we don't want to leak data to the clients and hence we avoid having anything about the exception on the message.

PR #3167, opened 2021-09-29
Rename "reused tool GUID" to "reused consumer key"

seanh:
The thing being reused here is called a "consumer key" (or "OAuth consumer key") in LTI not a "tool GUID". The LTI spec actually contains something else called a "tool GUID" (or "tool consumer instance GUID") so it's probably best not to use "tool GUID" in place of "consumer key".

There's a sense in which "tool GUID" is actually correct here. The LTI app (us) is called the "tool provider" and the LMS (Canvas etc) is the "tool consumer" and the thing that is shared between the two (the "application instance" in our own code's terminology) is the "tool". Since the OAuth consumer key is the thing that uniquely identifies a particular application instance I can see why you might call that the "tool GUID". Nonetheless, I think LTI always calls this the "consumer key", introducing our own terms is just going to introduce confusion, especially because of the similarity to "tool consumer instance GUID".

### DIFF
--- a/lms/views/exceptions.py
+++ b/lms/views/exceptions.py
@@ -47,14 +47,14 @@ class ExceptionViews:
     @exception_view_config(
         ReusedConsumerKey, renderer="lms:templates/error_dialog.html.jinja2"
     )
-    def reused_tool_guid_error(self):
+    def reused_consumer_key(self):
         self.request.response.status_int = 400
 
         self.request.context.js_config.enable_error_dialog_mode(
-            self.request.context.js_config.ErrorCode.REUSED_TOOL_GUID,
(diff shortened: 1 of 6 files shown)


lms/views/exceptions.py
```diff
@@ -53,9 +53,9 @@ def validation_error(self):
     @exception_view_config(
         ReusedConsumerKey, renderer="lms:templates/error_dialog.html.jinja2"
```
seanh, line 54: Note that the exception associated with this view is already called `ReusedConsumerKey` but the view was called `reused_tool_guid_error()`

marcospri (approved):

PR #3175, opened 2021-09-30
Add more thorough tests for assignment service

seanh:
Note that there are a couple of `@pytest.mark.xfail`'s in these tests that should be removed once the service itself has been fixed.

### DIFF
--- a/tests/unit/lms/services/assignment_test.py
+++ b/tests/unit/lms/services/assignment_test.py
@@ -1,184 +1,593 @@
-import uuid
 from unittest.mock import sentinel
 
 import pytest
+from sqlalchemy.orm.exc import MultipleResultsFound
 
-from lms.models import Assignment
 from lms.services.assignment import AssignmentService, factory
 from tests import factories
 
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/assignment_test.py
```diff
@@ -1,188 +1,604 @@
...
+    # resource_link_id and the matching ext_lti_assignment_id then get() will
+    # find that assignment if called with:
+    #
+    # 1. The matching resource_link_id and no ext_lti_assignment_id
+    # 2. The matching ext_lti_assignment_id and no resource_link_id
+    # 3. Both the matching resource_link_id and the matching ext_lti_assignment_id.
+    @pytest.mark.parametrize(
+        "resource_link_id,ext_lti_assignment_id",
+        [
+            pytest.param(
```
marcospri, line 55: Never seen `pytest.param` before. Here it will use the id as the test name instead of just the combination of values? :+1:

seanh: Yeah it goes into the test name in pytest's output. And also just helps to document each case for anyone reading the code (similar to a code comment)

seanh:
Rebased...

marcospri (approved):

PR #3176, opened 2021-09-30
Remove extra from set_document_url

seanh:
The `extra` argument doesn't appear to be used.

I'm not sure whether we should add an `extra` argument to `set_document_url()`: it's `set_document_url()` not
`set_document_url_and_merge_extras()`. Let's at least hold off until we have some code that uses the new argument.

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -119,13 +119,12 @@ class AssignmentService:
         except ValueError:
             return False
 
-    def set_document_url(  # pylint:disable=too-many-arguments
+    def set_document_url(
         self,
         document_url,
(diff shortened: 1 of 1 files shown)


marcospri (changes requested):
Fair enough, it's used on the next PR https://github.com/hypothesis/lms/pull/3122 but it could be added there.

You have a  good point on the naming even without the extra `set_document_url` also does the whole upsert thing so it could be renamed to `save` or similar.

There's seems to be a coverage problem on CI after removing this.

seanh:
Rebased...

marcospri (approved):

PR #3178, opened 2021-10-01
Extract common `PickerCanceledError` to a common path

esanzgar:
Both `GooglePickerClient` and `OneDrivePickerClient` use
`PickerCanceledError`. @robertknight suggested to extract this error and
move it to a common location:
https://github.com/hypothesis/lms/pull/3134#discussion_r717529172


lyzadanger (approved):
This seems entirely sensible as the Error subclass is shared by both external file pickers.

Any planned follow-up from this? We might consider moving the contents of `utils/api.js` (which only contains one thing: the `APIError` class) here?

PR #3179, opened 2021-10-01
Call `setLoadingIndicatorVisible(false)`

esanzgar:
Currently, `setLoadingIndicatorVisible(false)` is only call when the
pickers encounter an error. However, it would be more correct to call
`setLoadingIndicatorVisible(false)` always, either on success or on
error. Thus I move the line inside a `finally` clause. Suggested by
@robertknight here:
https://github.com/hypothesis/lms/pull/3134#discussion_r718334171


lyzadanger (approved):
This is a good quality improvement, yep.

PR #3184, opened 2021-10-04
Always call basic_lti_launch with a document_url

marcospri:


lms/resources/_js_config/__init__.py
```diff
@@ -87,6 +87,12 @@ def add_document_url(self, document_url):
                     _query={"document_url": document_url},
                 ),
             }
+        elif document_url.startswith("canvas://"):
+            course_id = self._request.params["custom_canvas_course_id"]
+            file_id = self._request.params["file_id"]
+            resource_link_id = self._request.params["resource_link_id"]
```
marcospri, line 93: The next PR changes the via url to take https://github.com/hypothesis/lms/pull/3186 document URL directly.

lms/views/basic_lti_launch.py
```diff
@@ -112,18 +111,16 @@ def canvas_file_basic_lti_launch(self):
...
             # configuration exists. If we're going to do that we might as well
             # make sure this URL is meaningful.
-            document_url=f"canvas://file/course/{course_id}/file_id/{file_id}",
+            document_url=document_url,
         )
-
-        self.context.js_config.add_canvas_file_id(course_id, resource_link_id, file_id)
-
-        return self.basic_lti_launch(grading_supported=False)
+        return self.basic_lti_launch(document_url=document_url, grading_supported=False)
```
seanh, line 123: Ok, so previously `canvas_file_basic_lti_launch()` called `js_config.add_canvas_file_id()` itself then called `self.basic_lti_launch()` with `document_url=None` which meant that `self.basic_lti_launch()` *did not* call `js_config.add_document_url()`. So `js_config.add_canvas_file_id()` was called and `js_config.add_document_url()` was not called.

Now `canvas_file_basic_lti_launch()` does not call `js_config.add_canvas_file_id()` anymore but instead calls `self.basic_lti_launch()` with a `document_url` so `self.basic_lti_launch()` *will* now call `js_config.add_document_url()`. So it has swapped: now `js_config.add_canvas_file_id()` is no longer called but `js_config.add_document_url()` *is* called.

This sounds like it breaks things: does `canvas_file_basic_lti_launch()` now add a document URL instead of a Canvas file ID? But I see that you've added `canvas://file/` URL special-casing to `JSConfig.add_document_url()` so that calling `add_document_url()` with a `canvas://file/` does the same as what calling `add_canvas_file_id()` would do. So actually there should not be any functional change here.

I really don't think the `basic_lti_launch()` method is helping to make the code easy to understand here. I get that it's trying to deduplicate code. But I think once it contains optional behaviour like `document_url` and `grading_supported` it just makes things hard to follow. It's kind of classic procedural spaghetti code to be honest. So I'm glad to see `document_url` becoming no longer optional.

lms/resources/_js_config/__init__.py
seanh: Ok, so calling `JSConfig.add_document_url(document_url)` with a `canvas://file/` URL does the same thing as calling `JSConfig.add_canvas_file_id(course_id, resource_link_id, file_id)` would do. And I can see that this fits with the already-existing special-casing of `blackboard://` URLs in this method :+1:

It's weird that `add_document_url(document_url)` actually ignores the given `document_url` (other than doing `document_url.startswith("canvas://file/")`) and uses `request.params` instead. But I see that https://github.com/hypothesis/lms/pull/3186/ changes it to use `document_url` instead of `request.params`

lms/resources/_js_config/__init__.py
```diff
@@ -87,6 +87,12 @@ def add_document_url(self, document_url):
                     _query={"document_url": document_url},
                 ),
             }
+        elif document_url.startswith("canvas://"):
```
seanh, line 90:
```suggestion
        elif document_url.startswith("canvas://file/"):
```

I think the Canvas file URLs style was chosen so that `canvas://file/*` can be used for Canvas files and `canvas://foo/*` could be used for other things in future. (I think the existing `startswith("blackboard://")` in this same method is also wrong for the same reason: should be `startswith("blackboard://content-resource/")`

lms/resources/_js_config/__init__.py
```diff
@@ -87,6 +87,12 @@ def add_document_url(self, document_url):
...
+            course_id = self._request.params["custom_canvas_course_id"]
+            file_id = self._request.params["file_id"]
+            resource_link_id = self._request.params["resource_link_id"]
+
+            self.add_canvas_file_id(course_id, resource_link_id, file_id)
```
seanh, line 95: This is now the only place where `add_canvas_file_id()` is called so you could inline it, but I see that in https://github.com/hypothesis/lms/pull/3186/ you've deleted it

seanh (approved):

PR #3185, opened 2021-10-04
Move `APIError` to `errors.js`

esanzgar:
Suggested here:
https://github.com/hypothesis/lms/pull/3178#pullrequestreview-768896614.

Additionally:

- improved the type of one of the arguments in the constructor, and
- create a minimal test to check the assignment of properties


lyzadanger (commented):
One thing I noticed here is that there are no tests for the `errors` module. I presume coverage is passing because the error classes are getting exercised in other tests (they're not mocked). What's your thought on a strategy here for testing?

lyzadanger (approved):
Thanks!

PR #3186, opened 2021-10-04
Via url endpoint for canvas files takes a document_url

marcospri:
Changes the parameters the via url endpoint takes, from course and file ids as separate arguments to one `canvas://` like URL with both params encoded there.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -46,27 +46,6 @@ class JSConfig:
         """
         return self._request.find_service(name="application_instance").get()
 
-    def add_canvas_file_id(self, course_id, resource_link_id, canvas_file_id):
-        """
-        Set the document to the Canvas file with the given canvas_file_id.
-
-        This configures the frontend to make an API call to get the URL to use
-        as the src for the Via iframe.
-
-        :raise HTTPBadRequest: if a request param needed to generate the config
-            is missing
-        """
-        self._config["api"]["viaUrl"] = {
-            "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
-            "path": self._request.route_path(
-                "canvas_api.files.via_url",
-                course_id=course_id,
-                resource_link_id=resource_link_id,
-                file_id=canvas_file_id,
-            ),
-        }
-        self._add_canvas_speedgrader_settings(canvas_file_id=canvas_file_id)
-
     def add_document_url(self, document_url):
         """
         Set the document to the document at the given document_url.
@@ -88,14 +67,17 @@ class JSConfig:
                 ),
             }
         elif document_url.startswith("canvas://"):
-            course_id = self._request.params["custom_canvas_course_id"]
-            file_id = self._request.params["file_id"]
-            resource_link_id = self._request.params["resource_link_id"]
-
-            self.add_canvas_file_id(course_id, resource_link_id, file_id)
+            self._config["api"]["viaUrl"] = {
+                "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
+                "path": self._request.route_path(
+                    "canvas_api.files.via_url",
+                    resource_link_id=self._request.params["resource_link_id"],
+                ),
+            }
         else:
             self._config["viaUrl"] = via_url(self._request, document_url)
-            self._add_canvas_speedgrader_settings(document_url=document_url)
+
+        self._add_canvas_speedgrader_settings(document_url=document_url)
 
     def add_vitalsource_launch_config(self, book_id, cfi=None):
         vitalsource_svc = self._request.find_service(name="vitalsource")
--- a/lms/views/api/canvas/files.py
+++ b/lms/views/api/canvas/files.py
@@ -1,9 +1,17 @@
 """Proxy API views for files-related Canvas API endpoints."""
+import re
+
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
 from lms.services.canvas import CanvasService
 from lms.views import helpers
+
+#: A regex for parsing the COURSE_ID and FILE_ID parts out of one of our custom
+#: canvas://file/course/COURSE_ID/file_id/FILE_ID URLs.
+DOCUMENT_URL_REGEX = re.compile(
+    r"canvas:\/\/file\/course\/(?P<course_id>[^\/]*)\/file_id\/(?P<file_id>[^\/]*)"
+)
 
 
 @view_defaults(permission=Permissions.API, renderer="json")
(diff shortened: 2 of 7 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -88,11 +67,15 @@ def add_document_url(self, document_url):
...
-            self.add_canvas_file_id(course_id, resource_link_id, file_id)
+            self._config["api"]["viaUrl"] = {
+                "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
+                "path": self._request.route_path(
+                    "canvas_api.files.via_url",
+                    resource_link_id=self._request.params["resource_link_id"],
+                    _query={"document_url": document_url},
+                ),
+            }
+            self._add_canvas_speedgrader_settings(document_url=document_url)
```
marcospri, line 78: This make speed grader launches to be "url configured". I think this is ok.

lms/views/api/canvas/files.py
```diff
@@ -33,16 +41,19 @@ def via_url(self):
...
             name="application_instance"
         ).get()
-
         assignment = self.request.find_service(name="assignment").get(
             application_instance.tool_consumer_instance_guid,
             self.request.matchdict["resource_link_id"],
         )
 
+        # We don't use the assignment document_url directly as it might be out of sync with what's stored in canvas
+        document_url = self.request.params["document_url"]
```
marcospri, line 50: This view always takes the document_url param.

It will up to the callers, the basic_lti_views to decide where that URL comes from, directly from the DB row or built form the query params for legacy ones.

lms/views/api/canvas/files.py
```diff
@@ -1,10 +1,18 @@
...
+import re
+
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
 from lms.services.canvas import CanvasService
 from lms.views import helpers
 
+#: A regex for parsing just the file_id part out of one of our custom
+#: blackboard://content-resource/<file_id>/ URLs.
```
marcospri, line 11: Ops, copied the whole thing from blackboard

lms/views/api/canvas/files.py
seanh: This does kind of leave me wondering why we don't use the `document_url` from the DB? If the initial LTI launch  would always update the `ModuleItemConfiguration` in the DB with the latest `document_url` then by the time we get to this `via_url` API call we would know that the DB is up to date.

In the case of Blackboard files the `via_url` API doesn't need to query the DB at all. It just receives the `document_url` in the query params and uses it. But for Canvas files, because of course copy, the `via_url` has to look up the assignment in the DB anyway. At which point it may as well use the `document_url` from the assignment. Then no `document_url` query param would be needed.

seanh (changes requested):
I think we can make the `_add_canvas_speedgrader_settings()` call in `JSConfig` unconditional.

I'm also wondering why the `via_url` API doesn't use the `document_url` from the DB? Then it wouldn't need a `document_url` query param. This would require the initial LTI launch request to create/update the assignment in the DB with the latest `document_url` from the launch params before generating the `via_url` URL. But I think that should be possible? (In fact I think it already does that?)

marcospri:
> I'm also wondering why the `via_url` API doesn't use the `document_url` from the DB? Then it wouldn't need a `document_url` query param. This would require the initial LTI launch request to create/update the assignment in the DB with the latest `document_url` from the launch params before generating the `via_url` URL. But I think that should be possible? (In fact I think it already does that?)

I think this is only true if we deploy the code to call the new endpoint on editing. Otherwise the edits will happen on canvas but not on our DB and we'll be launching the wrong URL.

seanh:
> > I'm also wondering why the `via_url` API doesn't use the `document_url` from the DB? Then it wouldn't need a `document_url` query param. This would require the initial LTI launch request to create/update the assignment in the DB with the latest `document_url` from the launch params before generating the `via_url` URL. But I think that should be possible? (In fact I think it already does that?)
> 
> I think this is only true if we deploy the code to call the new endpoint on editing. Otherwise the edits will happen on canvas but not on our DB and we'll be launching the wrong URL.

Can you double-check this? I might be confused but it looks to me like the `canvas_file_basic_lti_launch()` view updates the DB:

https://github.com/hypothesis/lms/blob/aaef5e4363ab6e1d8bce49e275e572c4f79eab49/lms/views/basic_lti_launch.py#L111-L122

So that would mean that by the time the frontend makes the subsequent `via_url` API request the DB has already been updated with the latest `document_url`.

marcospri:
> So that would mean that by the time the frontend makes the subsequent `via_url` API request the DB has already been updated with the latest `document_url`.

You are absolutely right, I was thinking in "canvas assignments" vs "canvas files assignments" but of course the non canvas files ones don't go thru this endpoint, they just use `via_url` :+1:

lms/views/api/canvas/files.py
```diff
@@ -1,10 +1,18 @@
 """Proxy API views for files-related Canvas API endpoints."""
+import re
+
 from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
 from lms.services.canvas import CanvasService
 from lms.views import helpers
 
+#: A regex for parsing just the file_id part out of one of our custom
```
seanh, line 10:
```suggestion
#: A regex for parsing just the FILE_ID part out of one of our custom
```

lms/views/api/canvas/files.py
```diff
@@ -33,16 +41,16 @@ def via_url(self):
         application_instance = self.request.find_service(
             name="application_instance"
         ).get()
-
         assignment = self.request.find_service(name="assignment").get(
             application_instance.tool_consumer_instance_guid,
             self.request.matchdict["resource_link_id"],
         )
 
+        url_match = DOCUMENT_URL_REGEX.search(assignment.document_url)
```
seanh, line 49: Could we call this `document_url_match`? Just to be clearer. I wasn't sure when I saw `url_match` being used below whether it referred to the `/via_url` API's URL or to the `document_url`

lms/resources/_js_config/__init__.py
```diff
@@ -88,14 +67,18 @@ def add_document_url(self, document_url):
...
-            file_id = self._request.params["file_id"]
-            resource_link_id = self._request.params["resource_link_id"]
-
-            self.add_canvas_file_id(course_id, resource_link_id, file_id)
+            self._config["api"]["viaUrl"] = {
+                "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
+                "path": self._request.route_path(
+                    "canvas_api.files.via_url",
+                    resource_link_id=self._request.params["resource_link_id"],
+                    _query={"document_url": document_url},
```
seanh, line 75: I don't think the `_query` is needed anymore?

lms/views/api/canvas/files.py
marcospri: Reworded, we get both FILE_ID and COURSE_ID.

PR #3193, opened 2021-10-04
Migration to change file.size to BigInt

marcospri:
## Test 

- Edit a local canvas file assigment and get the list of canvas files.
- `select * from file; ` Check the file there all have size
- Migrate with `make db`
-  `select * from file; ` again, same sizes
- `\d file;` shows size as a bigint


lms/migrations/versions/490952a3fbe8_big_int_for_file_size.py
```diff
@@ -0,0 +1,34 @@
+"""
+Use a wider integer type for file.type .
```
jon-betts, line 2:
```suggestion
Use a wider integer type for file.size to account for very large file size.
```

jon-betts, line 2:
```suggestion
Use a wider integer type for file.size to account for files over 2GB.
```

PR #3194, opened 2021-10-04
Update file model to use bigint for size

marcospri:


lms/models/file.py
```diff
@@ -53,5 +53,5 @@ class File(CreatedUpdatedMixin, BASE):
     name = sa.Column(sa.UnicodeText())
     """The user facing file name."""
 
-    size = sa.Column(sa.Integer())
+    size = sa.Column(sa.BigInteger())
```
jon-betts, line 56:
```suggestion
    # Use a big int here to allow files over 2GB in size.
    size = sa.Column(sa.BigInteger())
```

jon-betts (approved):
Yep. Not much to say on this one!

It might be nice to have a test case? Not sure if that's required though. It does give us a point to write "Canvas has these massive files sometimes and although we don't proxy them we do list them..." or something to that effect

PR #3196, opened 2021-10-05
One drive verify size

marcospri:
Set the exact same content and answer with a content-length header.

## Testing 
Set application_id to `f49d0253-f947-43f1-9735-ce73fde94af3` and compare the output with https://web.hypothes.is/.well-known/microsoft-identity-association.json

```
curl  https://web.hypothes.is/.well-known/microsoft-identity-association.json | md5sum 
1814193bf1f9c27500f60a05077b43b2                         
                     
curl  http://localhost:8001/.well-known/microsoft-identity-association.json | md5sum
1814193bf1f9c27500f60a05077b43b2 
````

### DIFF
--- a/lms/views/onedrive.py
+++ b/lms/views/onedrive.py
@@ -1,3 +1,6 @@
+import textwrap
+
+from pyramid.response import Response
 from pyramid.view import view_config
 
 
@@ -21,8 +24,15 @@ def redirect_uri(_request):
     renderer="json",
 )
 def verify_domain(request):
-    return {
-        "associatedApplications": [
-            {"applicationId": request.registry.settings["onedrive_client_id"]}
-        ]
-    }
+    application_id = request.registry.settings["onedrive_client_id"]
+    response_text = textwrap.dedent(
+        f"""\
+    {{
+      "associatedApplications": [
+        {{
+          "applicationId": "{application_id}"
+        }}
+      ]
+    }}"""
+    )
+    return Response(text=response_text, content_type="application/json")
--- a/tests/unit/lms/views/onedrive_test.py
+++ b/tests/unit/lms/views/onedrive_test.py
@@ -1,3 +1,4 @@
+import json
 from unittest.mock import sentinel
 
 from lms.views.onedrive import redirect_uri, verify_domain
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lms/views/onedrive.py [hunk 1]
```diff
@@ -21,8 +24,19 @@ def redirect_uri(_request):
     renderer="json",
 )
 def verify_domain(request):
-    return {
-        "associatedApplications": [
-            {"applicationId": request.registry.settings["onedrive_client_id"]}
-        ]
-    }
+    application_id = request.registry.settings["onedrive_client_id"]
+    response_text = textwrap.dedent(
```
marcospri, line 28: TIL about texwrap.dedent but I think doing this on a template and getting the size back for `content_length` would have been more verbose.

lms/views/onedrive.py
```diff
@@ -21,8 +24,17 @@ def redirect_uri(_request):
...
+        f"""\
+    {{
+      "associatedApplications": [
+        {{
+          "applicationId": "{application_id}"
+        }}
+      ]
+    }}"""
+    )
+    request.response.content_length = len(response_text)
```
marcospri, line 38: I was wrong before, we are not sending a content type header. Now we both set it and include exactly the same json as renderer in the verification site.

tests/unit/lms/views/onedrive_test.py
```diff
@@ -10,6 +11,14 @@ def test_redirect_uri(pyramid_request):
 def test_verify_domain(pyramid_request):
     pyramid_request.registry.settings["onedrive_client_id"] = sentinel.client_id
 
-    assert verify_domain(pyramid_request) == {
-        "associatedApplications": [{"applicationId": sentinel.client_id}]
+    assert json.loads(verify_domain(pyramid_request).text) == {
+        "associatedApplications": [{"applicationId": f"{sentinel.client_id}"}]
```
marcospri, line 15: forces str() on sentinel.client_id

lms/views/onedrive.py
seanh: Does this line do anything? I don't think setting `request.response.*` has any effect if the view then returns a _different_ `Response` object. It'd have to be `return Response(..., content_length=len(response_text))`. But I don't think you need to do either: I think you're seeing the correct `Content-Length` because Pyramid (or WebOb) is computing it for you automatically.

The purpose of `request.response` is for when views are using a renderer like the JSON renderer or the Jinja2 renderer where the view just returns a dict (or whatever type of return value the renderer needs). In those cases the view can set `request.response.*` properties (usually headers) and the renderer will pick those up.

lms/views/onedrive.py
seanh: Did you consider doing it manually?

```python
def verify_domain(request):
    application_id = request.registry.settings["onedrive_client_id"]
    response_text = f"""{{
  "associatedApplications": [
    {{
      "applicationId": "{application_id}"
    }}
  ]
}}"""
    request.response.content_length = len(response_text)

```

Too ugly?

seanh (changes requested):

lms/views/onedrive.py
marcospri: Yup, I did that first, still using the json rendered and didn't look back. `Response` sets the content-length on it's own.

lms/views/onedrive.py
marcospri: Yeah, tried that but it was not very nice.

seanh (approved):

PR #3198, opened 2021-10-05
Update supported browsers list

esanzgar:
All frontend Hypothesis app are going to have the same base line as
this:

https://github.com/hypothesis/bouncer/pull/469/commits/6a01199bfc196a8932be271f61c07029dbf886c5

In LMS, the `browser_check` bundle is transpiled to an older version of the
browsers.


robertknight (approved):
👍

PR #3199, opened 2021-10-05
Explicitly set Content-Length header

marcospri:
Length header is not present when deployed 

````
curl -I --http1.1  https://qa-lms.hypothes.is/.well-known/microsoft-identity-association.json
````
(same result without the I --http1.1 flag)

I'm not sure if this will fix/change anything it but adding the header explicitly to the Response here.

### DIFF
--- a/lms/views/onedrive.py
+++ b/lms/views/onedrive.py
@@ -35,4 +35,8 @@ def verify_domain(request):
       ]
     }}"""
     )
-    return Response(text=response_text, content_type="application/json")
+    return Response(
+        text=response_text,
+        content_type="application/json",
+        headers={"Content-Length": str(len(response_text))},
+    )

### CODE REVIEW

File: lms/views/onedrive.py [hunk 1]
```diff
@@ -35,4 +35,8 @@ def verify_domain(request):
       ]
     }}"""
     )
-    return Response(text=response_text, content_type="application/json")
+    return Response(
+        text=response_text,
+        content_type="application/json",
+        headers={"Content-Length": str(len(response_text))},
```
marcospri, line 41: You can lie here (ie set it to some random value) and that's what you get, so at least this take preference over any other mechanism at play.

seanh (approved):
I really don't think this should be necessary / should make any difference but let's see

PR #3200, opened 2021-10-05
Explicitly include content type header

marcospri:
https://github.com/hypothesis/lms/pull/3199 broke the inclusion of `content-type`

On this PR:

```
 curl -I  http://localhost:8001/.well-known/microsoft-identity-association.json 
HTTP/1.1 200 OK
Server: gunicorn
Date: Tue, 05 Oct 2021 15:28:00 GMT
Connection: close
Content-Length: 111
Content-Type: application/json
```

### DIFF
--- a/lms/views/onedrive.py
+++ b/lms/views/onedrive.py
@@ -37,6 +37,6 @@ def verify_domain(request):
     )
     return Response(
         text=response_text,
+        content_length=str(len(response_text)),
         content_type="application/json",
-        headers={"Content-Length": str(len(response_text))},
     )


lms/views/onedrive.py
```diff
@@ -37,6 +37,8 @@ def verify_domain(request):
     )
     return Response(
         text=response_text,
-        content_type="application/json",
-        headers={"Content-Length": str(len(response_text))},
+        headers={
+            "Content-Length": str(len(response_text)),
+            "Content-Type": "application/json",
+        },
```
seanh, line 43: Ah, I see what we did here: if you pass `headers=` then that overrides all the headers. But there are also convenience properties for various common headers that _don't_ remove any other headers, e.g. `content_length=`. See https://docs.pylonsproject.org/projects/pyramid/en/latest/api/response.html.

So if you set the `Content-Length` like this then it doesn't remove the `Content-Type`:

```python
return Response(text=response_text, content_length=str(len(response_text)))
```

_But_ you haven't told it that it's JSON so it returns `Content-Type: text/html`. So you still have to set the `Content-Type` like this as well:

```python
return Response(
    text=response_text,
    content_length=str(len(response_text)),
    content_type="application/json",
)
```

I think using the `content_length` and `content_type` convenience properties is the right thing to do, rather than `headers`, because it doesn't remove any _other_ headers that might have been present. Although in practice I don't think it makes a difference, I don't think there are any other headers in this case. I think there could also be some extra behaviour attached to these convenience properties that setting `headers` wouldn't give you.

seanh (changes requested):

seanh (approved):

PR #3209, opened 2021-10-11
Disable NR on onedrive template

marcospri:


marcospri:
This is reproducible on localhost has well with no NR agent.

PR #3221, opened 2021-10-12
Replace Browserify with Rollup

robertknight:
Change the LMS frontend to use Rollup for bundling JS. This change was done for the same reasons and using a similar approach to the corresponding change in the Hypothesis client. See https://github.com/hypothesis/client/pull/3810.

As part of the bundling change, I found it necessary to wrap the initialization code for the frontend app in `frontend_apps/index.js` in a function. See the first commit for details.

Unlike in client https://github.com/hypothesis/client/pull/3810, the bundles are still generated as classic scripts using an IIFE rather than native ES modules. This is because I think it will be easier to make that change in a separate step in this project.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM node:16.9.1-alpine as frontend-build
 
 ENV NODE_ENV production
-COPY .babelrc rollup.config.js gulpfile.js package.json yarn.lock ./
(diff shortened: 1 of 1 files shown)


robertknight:
The backend tests are failing due to an issue with pip. That should be fixed with https://github.com/hypothesis/tox-pip-sync/pull/20.

PR #3227, opened 2021-10-14
Take resource_link_id from query params when speed grading

marcospri:
The `resource_link_id` provided by canvas on launches done while using
speedgrading is incorrectly set to the value of "context_id` (the course
ID). This is not causing any problems (apart from some orphan
assignments
on the DB) because we are using the `url` and `file_id` URL params to
get the assignment details.

This would become a problem once we switch to reading the assignments
info from the DB instead of the URL like we do for other LMS so it needs
to be fixed.

# Testing

On master:

- Delete your assignments:

`truncate module_item_configurations ;`

- Launch a canvas assignment as a student: https://hypothesis.instructure.com/courses/125/assignments/1967
 
- Check `module_item_configurations` new row on database with resource_link_id `ec05d6beddad805005f98816f7e1bf26707a346f`.

- Launch now the assigment as teacher speedgrading: https://hypothesis.instructure.com/courses/125/gradebook/speed_grader?assignment_id=1967&student_id=35

- Check `module_item_configurations` again, there will be now two row, the newest with a resource_link_id of `f3cd019017839c4630358662a05540f2f6ec5f93` which happens to be the course ID (so repeating this process on another assignment won't keep adding rows but modifing that one).

Repeat the process on this branch, only one row will be created.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -79,8 +79,26 @@ class LTILaunchResource:
         return course
 
     @property
+    def _is_speed_grader(self):
+        return bool(self._request.GET.get("learner_canvas_user_id"))
+
+    @property
     def resource_link_id(self):
-        return self._request.params.get("resource_link_id")
+        # Canvas SpeedGrader launches LTI apps with the wrong resource_link_id,
+        # see:
+        #
+        # * https://github.com/instructure/canvas-lms/issues/1952
+        # * https://github.com/hypothesis/lms/issues/3228
+        #
+        # We add the correct resource_link_id as a query param on the launch
+        # URL that we submit to Canvas and use that instead of the incorrect
+        # resource_link_id that Canvas puts in the request's body.
+        if self._is_speed_grader and (
+            resource_link_id := self._request.GET.get("resource_link_id")
+        ):
+            return resource_link_id
+
+        return self._request.POST.get("resource_link_id")
 
     @property
     def is_canvas(self):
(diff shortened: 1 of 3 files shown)


lms/views/api/lti.py
```diff
@@ -115,6 +115,8 @@ def get_speedgrader_launch_url(self):
             "focused_user": parsed_params["h_username"],
             "learner_canvas_user_id": parsed_params["learner_canvas_user_id"],
             "group_set": parsed_params.get("group_set"),
+            "ext_lti_assignment_id": parsed_params.get("ext_lti_assignment_id"),
```
jon-betts, line 118: What is this one for?

jon-betts (commented):
A smattering of random suggestions, which doesn't attempt to get at the heart of whether this works or not.

lms/views/api/lti.py
marcospri: Both `ext_lti_assignment_id` and `resource_link_id` identify an assignment on canvas.

lms/resources/lti_launch.py
```diff
@@ -78,9 +78,20 @@ def h_group(self):
...
+    def _is_speed_grader(self):
+        return bool(self._request.GET.get("learner_canvas_user_id"))
+
     @property
     def resource_link_id(self):
-        return self._request.params.get("resource_link_id")
+        # If we are speed grading take the value from the query params that we injected
+        # ourselves instead of the wrong value provided by canvas in the body.
+        if self._is_speed_grader and (
+            resource_link_id := self._request.GET.get("resource_link_id")
```
seanh, line 90: Just to make sure I checked the WebOb docs and it looks like `request.GET` _is_ an officially supported way to get only the query params:

https://docs.pylonsproject.org/projects/webob/en/stable/api/request.html#webob.request.BaseRequest.GET  
https://docs.pylonsproject.org/projects/webob/en/stable/reference.html#query-post-variables

PR #3242, opened 2021-10-18
Remove ProxyAPIError

seanh:
Just use its parent class `ExternalRequestError` instead.

### `ProxyAPIError` is unnecessary

There's an [`ExternalRequestError` class](https://github.com/hypothesis/lms/blob/a6b7e551289181a9acb862b29aa9b4eef055f2ee/lms/services/exceptions.py#L1-L34) that adds `explanation`, `response` and `details` attributes (all optional) and a `__str__()` method based on these attributes (the output of `__str__()` shows up in Sentry and Papertrail):

```python
        super().__init__()
        self.response = response
        self.details = details

    def __str__(self):
        if self.response is None:
            return self.explanation

        parts = [
            self.explanation + ":",
            str(self.response.status_code or ""),
            self.response.reason,
            self.response.text,
        ]
        return " ".join([part for part in parts if part])
```

`ExternalRequestError` is **never raised directly**. Only its various subclasses are raised.

[`ProxyAPIError`](https://github.com/hypothesis/lms/blob/a6b7e551289181a9acb862b29aa9b4eef055f2ee/lms/services/exceptions.py#L47-L54) is then an `ExternalRequestError` subclass that doesn't add any functionality, exists for semantic reasons only:

```python
class ProxyAPIError(ExternalRequestError):
    """
    A problem with a third-party API request.

    Raised whenever a third-party API request times out or when an
    unsuccessful, invalid or unexpected response is received from a third-party
    API.
    """
```

The only special handling of `ProxyAPIError` is that there's a [special exception view for `ProxyAPIError`](https://github.com/hypothesis/lms/blob/a6b7e551289181a9acb862b29aa9b4eef055f2ee/lms/views/api/exceptions.py#L96-L100):

```python
@exception_view_config(context=ProxyAPIError)
def proxy_api_error(self):
    return self.error_response(
        message=self.context.explanation, details=self.context.details
    )
```

Nothing else catches `ProxyAPIError` specifically.

### What this PR does

The exception class inheritance hierarchy, _before this PR_, looks like this:

```
Exception
-> ExternalRequestError
   -> ProxyAPIError
      -> CanvasAPIError
         -> CanvasAPIPermissionError
         -> CanvasAPIServerError
   -> OAuth2TokenError
   -> LTIOutcomesAPIError
   -> HAPIError
```

This PR:

1. Changes the `proxy_api_error()` exception view into an `external_request_error()` exception views that catches `ExternalRequestError` instead
2. Changes code that was raising `ProxyAPIError` to raise `ExternalRequestError` instead
3. Changes `CanvasAPIError` to subclass `ExternalRequestError` directly instead of `ProxyAPIError`
4. Deletes `ProxyAPIError`

After:

```
Exception
-> ExternalRequestError
    -> CanvasAPIError
       -> CanvasAPIPermissionError
       -> CanvasAPIServerError
   -> OAuth2TokenError
   -> LTIOutcomesAPIError
   -> HAPIError
```

#### Simulate a `CanvasAPIError`

`CanvasAPIError`'s aren't affected by this PR. `CanvasAPIError` is now just an `ExternalRequestError` subclass, no longer a `ProxyAPIError` subclass, but the same exact exception view that used to catch `ProxyAPIError`'s now catches `ExternalRequestError`'s so it still handles `CanvasAPIError`'s.

There are many different reasons why `CanvasAPIError` might be raised and many ways to simulate it. One way is to break the `via_url` API:

```diff
diff --git a/lms/services/canvas_api/client.py b/lms/services/canvas_api/client.py
index 14556e09..64c3e41b 100644
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -281,7 +281,7 @@ class CanvasAPIClient:
         # https://canvas.instructure.com/doc/api/files.html#method.files.public_url
 
         return self._client.send(
-            "GET", f"files/{file_id}/public_url", schema=self._PublicURLSchema
+            "GET", f"files/foo/public_url", schema=self._PublicURLSchema
         )["public_url"]
 
     class _PublicURLSchema(RequestsResponseSchema):
```

Then launch a Canvas Files assignment. You'll see that the error dialog is the same on this branch as on master.

On master `CanvasAPIError`'s would not be reported to Sentry. On this branch they _are_ reported to Sentry due to the `report_exception()` call in the exception view.

#### Simulate an `OAuth2TokenError`

`OAuth2TokenError`'s aren't affected by this PR: there's a [separate exception view for `OAuth2TokenError`](https://github.com/hypothesis/lms/blob/a6b7e551289181a9acb862b29aa9b4eef055f2ee/lms/views/api/exceptions.py#L102-L104) and since `OAuth2TokenError` is more specific than `ExternalRequestError` (since it's a subclass) `OAuth2TokenError`'s will still continue to be caught by their own exception view.

To simulate this you can:

* Delete all the access tokens from your DB and launch an assignment. It should ask you to re-authorize and then launch the assignment successfully
* Change all the access tokens to "foo" and launch an assignment. The assignment should launch successfully (it will use the refresh token to get a new access token)
* Change all the access tokens _and_ refresh tokens to "foo" and launch an assignment. It should ask you to re-authorize and then launch the assignment successfully

#### Simulate an `LTIOutcomesAPIError`

Here's one way to create an `LTIOutcomesAPIError`:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 907d808a..65cb5da6 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -103,7 +103,7 @@ class LTIOutcomesClient:
 
         try:
             response = self.http_service.post(
-                url=self.service_url,
+                url="https://httpbin.org/status/400",
                 data=xml_body,
                 headers={"Content-Type": "application/xml"},
                 auth=self.oauth1_service.get_client(),
```

Then launch a Canvas assignment **as a student** to trigger the error.

On master the `/submissions` request will fail with a 500 and this body:

```json
{"message": "A problem occurred while handling this request. Hypothesis has been notified."}
```

The `LTIOutcomesAPIError` is handled by the final fallback case in the fallback `api_error()` exception view.

On this branch the `/submissions` request will fail with a 400 and this body:

```json
{"message": "Error calling LTI Outcomes service"}
```

The `LTIOutcomesAPIError` is handled by the new/modified `external_request_error()` exception view.

The resulting frontend error dialog is exactly the same.

#### Simulate an `HAPIError`

```diff
diff --git a/lms/services/h_api.py b/lms/services/h_api.py
index a31dfeea..a18137be 100644
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -86,10 +86,15 @@ class HAPI:
         if body is not None:
             request_args["data"] = body
 
+        url = self._base_url + path.lstrip("/")
+
+        if self._request.url.endswith("/sync"):
+            url = "http://httpbin.org/status/400"
+
         try:
             response = self._http_service.request(
                 method=method,
-                url=self._base_url + path.lstrip("/"),
+                url=url,
                 auth=self._http_auth,
                 headers=headers,
                 **request_args,
```

On master this results in a 500 response from the `/sync` API with this body:

```json
{"message": "A problem occurred while handling this request. Hypothesis has been notified."}
```

On this branch it's now a 400 with:

```json
{"message": "Connecting to Hypothesis failed"}
```

The resulting frontend dialog looks exactly the same.

### DIFF
--- a/tests/unit/lms/views/api/exceptions_test.py
+++ b/tests/unit/lms/views/api/exceptions_test.py
@@ -1,6 +1,6 @@
 import pytest
 
-from lms.services import CanvasAPIError, CanvasAPIPermissionError, LTIOutcomesAPIError
+from lms.services import CanvasAPIError, CanvasAPIPermissionError
 from lms.validation import ValidationError
 from lms.views.api.exceptions import APIExceptionViews
 
@@ -42,23 +42,6 @@ class TestExternalRequestError:
     @pytest.fixture
     def context(self):
         return CanvasAPIError(explanation="test_explanation", details={"foo": "bar"})
-
-
-class TestLTIOutcomesAPIError:
-    def test_it(self, pyramid_request, views):
-        json_data = views.external_request_error()
-
-        assert pyramid_request.response.status_code == 400
-        assert json_data == {
-            "message": "test_explanation",
-            "details": {"foo": "bar"},
-        }
-
-    @pytest.fixture
-    def context(self):
-        return LTIOutcomesAPIError(
-            explanation="test_explanation", details={"foo": "bar"}
-        )
 
 
 class TestNotFound:


tests/unit/lms/views/api/exceptions_test.py
```diff
@@ -43,23 +44,6 @@ def context(self):
...
-        assert json_data == {
-            "message": "test_explanation",
-            "details": {"foo": "bar"},
-        }
-
-    @pytest.fixture
-    def context(self):
-        return LTIOutcomesAPIError(
-            explanation="test_explanation", details={"foo": "bar"}
-        )
```
seanh, line 60: Deleted this test in a separate commit. The test was misnamed: `TestLTIOutcomesAPIError.test_it()`, but it's actually a test for the `proxy_api_error()` exception view (now `external_request_error()`). The test was also unnecessary: the `TestExternalRequestError` class above has it covered.

tests/unit/lms/views/api/exceptions_test.py
```diff
@@ -118,3 +102,8 @@ def context():
 @pytest.fixture
 def views(context, pyramid_request):
     return APIExceptionViews(context, pyramid_request)
+
+
+@pytest.fixture(autouse=True)
+def report_exception(patch):
+    return patch("lms.views.api.exceptions.report_exception")
```
seanh, line 109: `lms/views/api/exceptions.py` not imports `h_pyramid_sentry.report_exception()`. We don't want any tests to start accidentally making network requests to Sentry so we add this module-scoped autouse text fixture to ensure that `report_exception()` is always patched for all tests in this module. This is why patch fixtures should always be module-scoped and have `autouse`.

PR #3244, opened 2021-10-18
Fix test watch mode

robertknight:
Fix `gulp test --watch`. See https://github.com/hypothesis/frontend-shared/pull/214.


robertknight:
This is obsoleted by https://github.com/hypothesis/lms/pull/3245, which includes this fix via the @hypothesis/frontend-build package.

PR #3246, opened 2021-10-19
Take ext_lti_assignment_id from query params when speed grading

marcospri:
Same as the work previously done for `resource_link
t_id` https://github.com/hypothesis/lms/pull/3227 but now applying to `ext_lti_assignment_id`.

Note that ext_lti_assignment_id is never sent on SpeedGrader so `resource_link
t_id` should be enough to identify the assignment but getting both will make regular and speed grading launches to behave exactly the same.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -101,6 +101,17 @@ class LTILaunchResource:
         return self._request.POST.get("resource_link_id")
 
     @property
+    def ext_lti_assignment_id(self):
+        # Canvas SpeedGrader launches don't provide ext_lti_assignment_id
+        # but include it on the SpeedGrader URL we submit to canvas.
+        if self._is_speed_grader and (
+            ext_lti_assignment_id := self._request.GET.get("ext_lti_assignment_id")
+        ):
+            return ext_lti_assignment_id
+
+        return self._request.POST.get("ext_lti_assignment_id")
+
+    @property
     def is_canvas(self):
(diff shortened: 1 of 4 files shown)


lms/resources/lti_launch.py
```diff
@@ -100,6 +100,22 @@ def resource_link_id(self):
...
 
+    @property
+    def ext_lti_assignment_id(self):
+        # Canvas SpeedGrader launches don't provide `ext_lti_assignment_id`
+        #
+        # But in order to fix:
+        # * https://github.com/instructure/canvas-lms/issues/1952
+        # * https://github.com/hypothesis/lms/issues/3228
+        #
+        # We include it on the SpeedGrader URL we submit to canvas.
```
seanh, line 111: Is this code really related to those two issues? Those issues are about the incorrect `resource_link_id` in SpeedGrader launches. I think this `ext_lti_assignment_id` property exists for another reason.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -38,6 +38,25 @@ def test_it(self, pyramid_request, learner_id, get_id, post_id, expected):
         assert LTILaunchResource(pyramid_request).resource_link_id == expected
 
 
+class TestExtLtiAssignmentId:
```
seanh, line 41:
```suggestion
class TestExtLTIAssignmentID:
```

PR #3248, opened 2021-10-19
Remove `LTIOutcomesAPIError`

seanh:
Depends on https://github.com/hypothesis/lms/pull/3242

This exception subclass isn't providing any functionality, nothing catches it or handles it specially. Just use the base class `ExternalRequestError` directly instead.

### Testing

Here's one way to trigger an `LTIOutcomesAPIError`:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 8757ef98..dab9d56c 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -133,6 +133,7 @@ class LTIOutcomesClient:
             status = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
                 "imsx_codeMajor"
             ]
+            foo = data["foo"]
 
         except KeyError as err:
             raise ExternalRequestError("Malformed LTI outcome response") from err
```

Then launch an assignment _as a student_. The error response from the `/submissions` API should look the same and the error dialog shown by the frontend should be the same.

One difference that this PR makes compared to master is that LTI Outcomes API errors will not appears as `ExternalRequestError` in Papertrail and Sentry which does make them less easy to distinguish. You can still see which route the exception came from (`lti_api.submissions.record`) and the `explanation` string that was passed though, so you can still distinguish them.

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -8,7 +8,6 @@ from lms.services.exceptions import (
     CanvasFileNotFoundInCourse,
     ExternalRequestError,
     HTTPError,
-    LTIOutcomesAPIError,
     OAuth2TokenError,
 )
 from lms.services.h_api import HAPIError
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -153,15 +153,6 @@ class CanvasAPIServerError(CanvasAPIError):
     """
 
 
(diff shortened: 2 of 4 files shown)


marcospri (approved):
Looks good, nothing more to add :+1:

PR #3251, opened 2021-10-19
Remove `HTTPError` (use `ExternalRequestError` instead)

seanh:
`HTTPError` is the exception class raised by `HTTPService` whenever anything goes wrong with a server-to-server HTTP request.

`HTTPService` is almost always used when sending HTTP requests: `HAPI`, `LTIOutcomesAPIClient`, `VitalSourceService` and `BlackboardAPIClient` all use `HTTPService`. `CanvasAPIClient` is the one thing that sends external HTTP requests without using `HTTPService` and that's only because `CanvasAPIClient`'s code is unusual and can't be easily refactored to work like the others do.

`HTTPError` is kind of the same thing as `ExternalRequestError` but with fewer attributes and less functionality:

```python
    """
    A problem with a network request to an external service.

    ...
    """
    
        super().__init__()
        self.response = response
        self.details = details

    def __str__(self):
        ...

class HTTPError(Exception):
    """A problem with an HTTP request to an external service."""

        super().__init__(response)
        self.response = response
```

This PR deletes `HTTPError` and changes `HTTPService` to raise `ExternalRequestError` instead.

As a result all the services that use `HTTPService` will now get better error reporting: details of error responses from third-party services will be sent to Papertrail and Sentry.

Testing
=======

Blackboard API Errors
---------------------

Hack the code to simulate an error from the Blackboard API:

```diff
diff --git a/lms/services/blackboard_api/_basic.py b/lms/services/blackboard_api/_basic.py
index 54b78336..8235b76f 100644
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -72,6 +72,7 @@ class BasicClient:
 
     def request(self, method, path):
         url = self._api_url(path)
+        url = "http://httpbin.org/status/403"
 
         try:
             return self._send(method, url)
```

Then launch a test Blackboard Files assignment from https://aunltd-test.blackboard.com/.

* The `/via_url` API request gets a 400 error response with `{"message": "External request failed"}`
* A generic "Something went wrong: There was a problem fetching this Hypothesis assignment" error dialog is shown.
  * Note that the `"External request failed"` error message that the backend returned is not displayed by the frontend. This is an existing issue: https://github.com/hypothesis/lms/issues/3259

You can also try launching the **localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted** assignment and see that the `{"error_code": "blackboard_file_not_found_in_course", "details": {"file_id": "_9019_1"}}` response is sent and triggers the **Hypothesis couldn't find the file in the course** error dialog from the frontend.

h API Errors
------------

Hack the code to simulate an error from h's sync API:

```diff
diff --git a/lms/services/h_api.py b/lms/services/h_api.py
index 9a9204b8..806866b3 100644
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -86,10 +86,15 @@ class HAPI:
         if body is not None:
             request_args["data"] = body
 
+        url = self._base_url + path.lstrip("/")
+
+        if self._request.url.endswith("/sync"):
+            url = "http://httpbin.org/status/500"
+
         try:
             response = self._http_service.request(
                 method=method,
-                url=self._base_url + path.lstrip("/"),
+                url=url,
                 auth=self._http_auth,
                 headers=headers,
                 **request_args,
```

Then launch any Canvas assignment. The `/sync` API request will get a 400 response with this JSON body:

```json
{"message": "Connecting to Hypothesis failed"}
```

This will trigger an error dialog from the frontend.

LTI Outcomes API Errors
-----------------------

Hack the code to simulate an error from the LTI Outcomes API:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 15c476e2..c9088042 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -99,7 +99,7 @@ class LTIOutcomesClient:
         response = None
 
         response = self.http_service.post(
-            url=self.service_url,
+            url="http://httpbin.org/status/500",
             data=xml_body,
             headers={"Content-Type": "application/xml"},
             auth=self.oauth1_service.get_client(),
```

Then launch any Canvas assignment as a student. The `/submissions` API request will get a 400 response with this body:

```json
{"message": "Error calling LTI Outcomes service"}
```

Which will trigger an error dialog from the frontend.

Here's another way to break the LTI Outcomes API that will get the same result:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 1792e164..885e2587 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -110,7 +110,7 @@ class LTIOutcomesClient:
             raise
 
         try:
-            data = xmltodict.parse(response.text)
+            data = xmltodict.parse("blah blah")
         except ExpatError as err:
             raise ExternalRequestError(
                 "Unable to parse XML response from LTI Outcomes service", response
```

And another:

```diff
diff --git a/lms/services/lti_outcomes.py b/lms/services/lti_outcomes.py
index 1792e164..fe9d0893 100644
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -132,6 +132,7 @@ class LTIOutcomesClient:
             status = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
                 "imsx_codeMajor"
             ]
+            data["foo"]
 
         except KeyError as err:
             raise ExternalRequestError("Malformed LTI outcome response") from err
```

### DIFF
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -5,7 +5,7 @@ from lms.services.blackboard_api._schemas import (
     BlackboardListFilesSchema,
     BlackboardPublicURLSchema,
 )
-from lms.services.exceptions import BlackboardFileNotFoundInCourse, HTTPError
+from lms.services.exceptions import BlackboardFileNotFoundInCourse, ExternalRequestError
 
 # The maxiumum number of paginated requests we'll make before returning.
 PAGINATION_MAX_REQUESTS = 25
@@ -26,8 +26,8 @@ class BlackboardAPIClient:
         """
         Save a new Blackboard access token for the current user to the DB.
 
-        :raise services.HTTPError: if something goes wrong with the access
-            token request to Blackboard
+        :raise services.ExternalRequestError: if something goes wrong with the
+            access token request to Blackboard
         """
         self._api.get_token(authorization_code)
 
@@ -91,8 +91,8 @@ class BlackboardAPIClient:
                 "GET",
                 f"courses/uuid:{course_id}/resources/{file_id}?fields=downloadUrl",
             )
-        except HTTPError as err:
-            if err.response.status_code == 404:
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
                 raise BlackboardFileNotFoundInCourse(file_id) from err
             raise
 
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -163,14 +163,6 @@ class CanvasFileNotFoundInCourse(Exception):
         super().__init__(self.details)
 
 
-class HTTPError(Exception):
-    """A problem with an HTTP request to an external service."""
-
-    def __init__(self, response=None):
-        super().__init__(response)
-        self.response = response
-
-
 class BlackboardFileNotFoundInCourse(Exception):
     """A Blackboard file ID wasn't found in the current course."""
 
--- a/lms/services/h_api.py
+++ b/lms/services/h_api.py
@@ -3,7 +3,7 @@
 from h_api.bulk_api import BulkAPI, CommandBuilder
 
 from lms.models import HUser
-from lms.services import ExternalRequestError, HTTPError
+from lms.services import ExternalRequestError
 
 
 class HAPIError(ExternalRequestError):
@@ -94,9 +94,7 @@ class HAPI:
                 headers=headers,
                 **request_args,
             )
-        except HTTPError as err:
-            response = getattr(err, "response", None)
-
-            raise HAPIError("Connecting to Hypothesis failed", response) from err
+        except ExternalRequestError as err:
+            raise HAPIError("Connecting to Hypothesis failed", err.response) from err
 
         return response
--- a/lms/services/lti_outcomes.py
+++ b/lms/services/lti_outcomes.py
@@ -2,7 +2,7 @@ from xml.parsers.expat import ExpatError
 
 import xmltodict
 
-from lms.services.exceptions import ExternalRequestError, HTTPError
+from lms.services.exceptions import ExternalRequestError
 
 __all__ = ["LTIOutcomesClient"]
 
@@ -105,10 +105,9 @@ class LTIOutcomesClient:
                 headers={"Content-Type": "application/xml"},
                 auth=self.oauth1_service.get_client(),
             )
-        except HTTPError as err:
-            raise ExternalRequestError(
-                "Error calling LTI Outcomes service", response
-            ) from err
+        except ExternalRequestError as err:
+            err.message = "Error calling LTI Outcomes service"
+            raise
 
         try:
             data = xmltodict.parse(response.text)
--- a/lms/services/vitalsource/client.py
+++ b/lms/services/vitalsource/client.py
@@ -1,7 +1,7 @@
 import oauthlib
 from oauthlib.oauth1 import SIGNATURE_HMAC_SHA1, SIGNATURE_TYPE_BODY
 
-from lms.services.exceptions import ExternalRequestError, HTTPError
+from lms.services.exceptions import ExternalRequestError
 from lms.services.vitalsource._schemas import BookInfoSchema, BookTOCSchema
 
 
@@ -35,9 +35,9 @@ class VitalSourceService:
     def book_info(self, book_id: str):
         try:
             response = self.get(f"products/{book_id}")
-        except HTTPError as exp:
-            if exp.response.status_code == 404:
-                raise ExternalRequestError(f"Book {book_id} not found") from exp
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
+                err.message = f"Book {book_id} not found"
 
             raise
 
@@ -46,9 +46,9 @@ class VitalSourceService:
     def book_toc(self, book_id: str):
         try:
             response = self.get(f"products/{book_id}/toc")
-        except HTTPError as exp:
-            if exp.response.status_code == 404:
-                raise ExternalRequestError(f"Book {book_id} not found") from exp
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
+                err.message = f"Book {book_id} not found"
 
             raise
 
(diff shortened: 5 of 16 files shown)


lms/services/exceptions.py
```diff
@@ -12,7 +12,7 @@ class ExternalRequestError(Exception):
     :type details: JSON-serializable dict or ``None``
     """
 
-    def __init__(self, explanation=None, response=None, details=None):
+    def __init__(self, explanation="", response=None, details=None):
```
seanh, line 15: The `__str__()` method below would crash if `explanation` was `None`

lms/services/h_api.py
```diff
@@ -94,9 +94,7 @@ def _api_request(self, method, path, body=None, headers=None):
                 headers=headers,
                 **request_args,
             )
-        except HTTPError as err:
-            response = getattr(err, "response", None)
```
seanh, line 98: `ExternalRequestError` always has a `response` attr (which may be `None`) so it's not necessary to use `getattr()` here

lms/services/lti_outcomes.py
```diff
@@ -98,17 +98,12 @@ def _send_request(self, request_body):
...
-            response = self.http_service.post(
-                url=self.service_url,
-                data=xml_body,
-                headers={"Content-Type": "application/xml"},
-                auth=self.oauth1_service.get_client(),
-            )
-        except HTTPError as err:
-            raise ExternalRequestError(
-                "Error calling LTI Outcomes service", response
-            ) from err
```
seanh, line 111: `LTIOutcomesClient` was catchin `HTTPError` and re-raising it as `ExternalRequestError`. Now that `HTTPService` raises `ExternalRequestError` directly `LTIOutcomesClient` no longer needs a `try` / `except` at all here. (This does mean that the `"Error calling LTI Outcomes service"` string has been lost but I don't think that was adding anything.)

(`LTIOutcomesClient` previously caught `HTTPError` and raised `LTIOutcomesAPIError`, a pointless `ExternalRequestError` subclass, but that was recently replaced with just raising `ExternalRequestError` directly.)

lms/services/vitalsource/client.py
```diff
@@ -35,9 +35,9 @@ def get(self, endpoint):
     def book_info(self, book_id: str):
         try:
             response = self.get(f"products/{book_id}")
-        except HTTPError as exp:
-            if exp.response.status_code == 404:
-                raise ExternalRequestError(f"Book {book_id} not found") from exp
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
```
seanh, line 39: Again `err.response` might be `None` so you have to use `getattr()` here

lms/services/vitalsource/client.py
```diff
@@ -46,9 +46,9 @@ def book_info(self, book_id: str):
     def book_toc(self, book_id: str):
         try:
             response = self.get(f"products/{book_id}/toc")
-        except HTTPError as exp:
-            if exp.response.status_code == 404:
-                raise ExternalRequestError(f"Book {book_id} not found") from exp
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
+                err.explanation = f"Book {book_id} not found"
```
seanh, line 51: `err.response` might be `None` so you have to use `getattr()` here

lms/services/lti_outcomes.py
```diff
@@ -105,10 +105,9 @@ def _send_request(self, request_body):
                 headers={"Content-Type": "application/xml"},
                 auth=self.oauth1_service.get_client(),
             )
-        except HTTPError as err:
-            raise ExternalRequestError(
-                "Error calling LTI Outcomes service", response
-            ) from err
+        except ExternalRequestError as err:
+            err.message = "Error calling LTI Outcomes service"
+            raise
```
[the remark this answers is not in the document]
seanh: I've used this pattern in a few places

tests/unit/lms/services/vitalsource/client_test.py
```diff
@@ -90,18 +90,24 @@ def test_book_info_not_found(self, svc):
...
-            side_effect=HTTPError(factories.requests.Response(status_code=404)),
+            side_effect=ExternalRequestError(
+                response=factories.requests.Response(status_code=404)
+            ),
         ):
-            with pytest.raises(ExternalRequestError):
+            with pytest.raises(ExternalRequestError) as exc_info:
                 svc.book_info("BOOK_ID")
 
+            assert exc_info.value.message == "Book BOOK_ID not found"
```
seanh, line 100: This test wasn't actually testing that the right path was taken through the code. As a result I actually broke the code and this test was still passing. Only some missing coverage alerted me to this problem. So I added this extra assert to fix the test.

tests/unit/lms/services/vitalsource/client_test.py
```diff
@@ -115,18 +121,24 @@ def test_book_toc_not_found(self, svc):
...
-            side_effect=HTTPError(factories.requests.Response(status_code=404)),
+            side_effect=ExternalRequestError(
+                response=factories.requests.Response(status_code=404)
+            ),
         ):
-            with pytest.raises(ExternalRequestError):
+            with pytest.raises(ExternalRequestError) as exc_info:
                 svc.book_toc("BOOK_ID")
 
+            assert exc_info.value.message == "Book BOOK_ID not found"
```
seanh, line 131: This test wasn't actually testing that the right path was taken through the code. As a result I actually broke the code and this test was still passing. Only some missing coverage alerted me to this problem. So I added this extra assert to fix the test.

lms/services/blackboard_api/client.py
```diff
@@ -91,8 +91,8 @@ def public_url(self, course_id, file_id):
                 "GET",
                 f"courses/uuid:{course_id}/resources/{file_id}?fields=downloadUrl",
             )
-        except HTTPError as err:
-            if err.response.status_code == 404:
+        except ExternalRequestError as err:
+            if getattr(err.response, "status_code", None) == 404:
```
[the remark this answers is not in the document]
seanh: Sure! I think that could remove some duplication. I'll do that in a follow-up PR

marcospri (approved):

PR #3254, opened 2021-10-20
New is_legacy_speed_grader property

marcospri:
Detects if we are currently speedgrading without the fix for

https://github.com/instructure/canvas-lms/issues/1952

yet applied.


lms/resources/lti_launch.py
```diff
@@ -82,6 +82,19 @@ def h_group(self):
...
         return bool(self._request.GET.get("learner_canvas_user_id"))
 
+    @property
+    def is_legacy_speed_grader(self):
+        # Detect speed grader launches without the fix for
+        #
+        # * https://github.com/instructure/canvas-lms/issues/1952
+        # * https://github.com/hypothesis/lms/issues/3228
+        #
+        # applied which includes the correct value for `resource_link_id` in the query params.
```
seanh, line 92: I think this is a docstring rather than a comment. Maybe something like:

```python
        """
        Return True if the current request is a legacy SpeedGrader launch.

        To work around a Canvas bug we add the assignment's resource_link_id as
        a query param on the LTI launch URLs that we submit to SpeedGrader (see
        https://github.com/instructure/canvas-lms/issues/1952 and
        https://github.com/hypothesis/lms/issues/3228).

        "Legacy" SpeedGrader submissions are ones from before we implemented
        this work-around, so they don't have the resource_link_id query param.
        """
```

Always good to be explicit in the first line of a docstring ("Return True if..." rather then "Detect...").

lms/resources/lti_launch.py
```diff
@@ -82,6 +82,19 @@ def h_group(self):
...
+        return (
+            self._is_speed_grader
+            and self.resource_link_id == self._request.POST["context_id"]
```
seanh, line 95: I wonder if we want something like this instead?

```suggestion
            and "resource_link_id" not in self._request.GET
```

I think a legacy SpeedGrader URL is one that we submitted to Canvas before we started adding `resource_link_id` query params to SpeedGrader URLs. If we instead rely on `self.resource_link_id == self._request.POST["context_id"]` then this might start returning `False` instead of `True` if Instructure fix https://github.com/instructure/canvas-lms/issues/1952.

seanh (changes requested):

PR #3257, opened 2021-10-20
Fix `ExternalRequestError.__str__()` crashes

seanh:
`ExternalRequestError.__str__()` is based on `self.explanation` and `self.response`. Both of these are optional arguments to `ExternalRequestError.__init__()` so there are four possible cases:
    
1. Both `response` and `explanation` are `None`: `ExternalRequestError()`
2. `explanation` is a string but `response` is `None`: `ExternalRequestError("Something went wrong")`
3. `response` is a `Response` but `explanation` is `None`: `ExternalRequestError(response=response)`
4. `explanation` is a string and `response` is a `Response`: `ExternalRequestError(explanation, response)`
    
Two of these cases were missing from the tests and were actually resulting in `__str__()` crashing. Add tests and fix the cases.
    
I'd like to also change the format of `ExternalRequestError`'s string representations (it's a bit of a mess) but that's for another PR.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -19,14 +19,16 @@ class ExternalRequestError(Exception):
         self.details = details
 
     def __str__(self):
+        explanation = self.explanation or "ExternalRequestError"
+
         if self.response is None:
-            return self.explanation
+            return explanation
 
         # Log the details of the response. This goes to both Sentry and the
         # application's logs. It's helpful for debugging to know how the
         # external service responded.
         parts = [
-            self.explanation + ":",
+            explanation + ":",
             str(self.response.status_code or ""),
             self.response.reason,
             self.response.text,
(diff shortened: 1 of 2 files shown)


lms/services/exceptions.py
```diff
@@ -19,14 +19,16 @@ def __init__(self, explanation=None, response=None, details=None):
         self.details = details
 
     def __str__(self):
+        explanation = self.explanation or "ExternalRequestError"
```
seanh, line 22: I think the string representation should probably be something like:

    "ExternalRequestError(explanation='...', response=Response(status=xxx, reason='...', text='...'))"

That's a lot clearer and more explicit, and can represent cases when either the `explanation` or `Response` is missing:

* `"ExternalRequestError(explanation=None, response=Response(...))"`
* `"ExternalRequestError(explanation='...', response=None)"`
* `"ExternalRequestError(explanation=None, response=None)"`

Might also want to include `self.details` in `__str__()`.

But that's for another PR. For now I've handled the case where `self.explanation` is `None` by defaulting it to `ExternalRequestError` :man_shrugging:

marcospri (approved):

PR #3264, opened 2021-10-21
Improve the string representation of `ExternalRequestErrors`

seanh:
Improve the string representation of `ExternalRequestError`'s (which appears in New Relic, Sentry, Papertrail, etc) to be more detailed and explicit.

This changes the string representation to:

    "ExternalRequestError(message='...', response=Response(status_code=xxx, reason='...', text='...')"

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -15,7 +15,7 @@ class ExternalRequestError(Exception):
     :type details: JSON-serializable dict or ``None``
     """
 
-    def __init__(self, message="External request failed", response=None, details=None):
+    def __init__(self, message=None, response=None, details=None):
         super().__init__()
         self.message = message
         self.response = response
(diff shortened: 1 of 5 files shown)


lms/services/exceptions.py
```diff
@@ -12,26 +12,31 @@ class ExternalRequestError(Exception):
...
+                f"status_code={repr(self.response.status_code)}, "
+                f"reason={repr(self.response.reason)}, "
+                f"text={repr(self.response.text)})"
+            )
+        else:
+            response = "None"
+
+        return (
+            f"ExternalRequestError(message={repr(self.message)}, response={response})"
+        )
```
seanh, line 36: Changed this from `__str__()` to `__repr__()` because `__repr__()` is used by the debugger (so now we get to see the nice string representation in the debugger as well).

tests/unit/lms/services/exceptions_test.py
```diff
@@ -16,71 +16,55 @@
...
+                requests.Response,
+                instance=True,
+                status_code=response_attrs["status"],
+                reason=response_attrs["reason"],
+                text=response_attrs["text"],
+            )
+        else:
+            response = None
+
+        err = ExternalRequestError(message=message, response=response)
```
seanh, line 67: Combined the four separate tests into a single parametrized one

tests/unit/lms/views/exceptions_test.py
```diff
@@ -44,7 +44,11 @@ def test_hapi_error(self, assert_response, pyramid_request):
 
         template_data = ExceptionViews(exception, pyramid_request).hapi_error()
 
-        assert_response(template_data, 500, message="hapi_error_test_message")
+        assert_response(
+            template_data,
+            500,
+            message="ExternalRequestError(message='hapi_error_test_message', response=None)",
```
seanh, line 50: This is because [the exception view for `HAPIError` calls `str()` on the exceptoin](https://github.com/hypothesis/lms/blob/2e5f76ab1edb8c82a0e4d24c8ba41470e1c62283/lms/views/exceptions.py#L36-L38). I don't think we want that as these `"ExternalRequestError(...)"` strings are meant for our monitoring systems not for end users. I should fix that exception view

PR #3307, opened 2021-11-01
Misc lti launch functional test fixes 

marcospri:


tests/functional/views/basic_lti_launch_test.py
```diff
@@ -137,45 +124,56 @@ def test_basic_lti_launch(
...
-        lti_params = request.getfixturevalue(lti_params_fixture_name)
+        get_params, post_params = request.getfixturevalue(launch_params_fixture_name)
 
-        canvas_file_id = lti_params["file_id"]
-        canvas_course_id = lti_params["custom_canvas_course_id"]
+        canvas_file_id = post_params["file_id"]
+        canvas_course_id = post_params["custom_canvas_course_id"]
 
-        response = app.post(
-            "/lti_launches?learner_canvas_user_id=USER_ID",
```
marcospri, line 173: This was wrong and changing the launch type (always speedgrader) from the fixture.

lms/resources/lti_launch.py
```diff
@@ -79,11 +79,11 @@ def h_group(self):
         return course
 
     @property
-    def _is_speed_grader(self):
+    def _is_speedgrader(self):
```
marcospri, line 82: Following naming convention.

tests/functional/views/basic_lti_launch_test.py
```diff
@@ -16,30 +16,22 @@
...
         del lti_params["oauth_signature"]
 
-        response = app.post(
-            "/lti_launches",
-            params=lti_params,
-            headers={
-                "Accept": "text/html",
-                "Content-Type": "application/x-www-form-urlencoded",
-            },
+        response = post_launch(
```
marcospri, line 24: Switch to a fixture to call the app as needs now to include the get_params.

seanh: :+1: Saves a bunch of duplication too, and I like to see `response = post_launch(post_params=lti_params, status=403)` because that contains only the parameters that this test actually cares about, and hides away URL and headers that are the same in every test

tests/functional/views/basic_lti_launch_test.py
```diff
@@ -16,30 +16,22 @@
...
+            post_params=lti_params,
             status=403,
```
seanh, line 26: If you delete the trailing comma from this line then Black will format the call onto one line for you. I sometimes wonder whether we should just disable Black's [magic trailing comma](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma) feature

tests/functional/views/basic_lti_launch_test.py
```diff
@@ -137,45 +124,56 @@ def test_basic_lti_launch(
...
+            param(
+                "canvas_file_fixed_speedgrader_launch_params",
+                None,
+                id="fixed SpeedGrader launch with no existing DB rows",
+            ),
+            param(
+                "canvas_file_fixed_speedgrader_launch_params",
+                "legacy_speedgrader_assignment",
+                id="fixed SpeedGrader launch with legacy SpeedGrader DB row",
+            ),
```
seanh, line 155: What's a "fixed" SpeedGrader launch?

tests/functional/views/basic_lti_launch_test.py
```diff
@@ -329,15 +330,27 @@ def canvas_url_lti_params(self, lti_params, sign_lti_params):
...
+                resource_link_id=post_params["context_id"],
+            )
+        )
+
+    @pytest.fixture
+    def canvas_file_fixed_speedgrader_launch_params(self, canvas_file_launch_params):
+        _, post_params = canvas_file_launch_params
+        return {
+            "learner_canvas_user_id": "USER_ID",
+            # Fixed (non-legacy) SpeedGrader launches include resource_link_id in the URL params
```
seanh, line 351: Ok so a "fixed" SpeedGrader launch means a non-legacy one. I think maybe we should call them "SpeedGrader launches" and "legacy SpeedGrader launches"

seanh (approved):

tests/functional/views/basic_lti_launch_test.py
```diff
@@ -351,6 +364,25 @@ def _sign(params):
 
         return _sign
 
+    @pytest.fixture
+    def post_launch(self, app):
+        def _post_launch(post_params, get_params=None, **kwargs):
+            url = "/lti_launches"
+            if get_params:
+                url += f"?{urlencode(get_params)}"
```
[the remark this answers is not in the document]
marcospri: Yup, `DummyRequest` ha some wilderness around this on the unit tests as well where everything goes into the same dictionary.

tests/functional/views/basic_lti_launch_test.py
marcospri: This is mostly me not re-visiting the commas after adding/remmoving params. 

I think we should keep black's default, it's very handy to control a bit the formatting when for example mixing with with `pylint: disable...` hints.

Issue #3319, opened 2021-11-01
Log and monitor external requests and responses that fail validation


seanh:
### Background

If there's an error while calling the Blackboard API (for example a timeout, or a 4xx or 5xx response) then we show the details of the failed request and response in the error dialog:

![Screenshot from 2021-11-01 12-19-54](https://user-images.githubusercontent.com/22498/139670491-b7fe382f-6233-4126-a380-aea586f768d4.png)

These details also get reported to Sentry:

![Screenshot from 2021-11-01 12-21-44](https://user-images.githubusercontent.com/22498/139670645-e4351b32-42d0-43d8-81f8-b3b8f657bd94.png)

And they also get logged to Papertrail:

```
lms.services.exceptions.ExternalRequestError: ExternalRequestError(message=None, extra_details=None, request=Request(method='GET', url='http://httpbin.org/status/418', body=None), response=Response(status_code=418, reason="I'M A TEAPOT", body='\n    -=[ teapot ]=-\n\n       _...._\n     .\'  _ _ `.\n    | ."` ^ `". _,\n    \\_;`"---"`|//\n      |       ;/\n      \\_     _/\n        `"""`\n'))
```

### Problem

This error reporting doesn't work if there's a **validation error**: when we receive a successful response from the external service but that response isn't what we expected and doesn't pass our validation schema. In this case only the validation error is shown in the error dialog, not the request and response. The validation error alone isn't very helpful for debugging because it doesn't tell you _which_ external request failed:

![Screenshot from 2021-11-01 12-25-42](https://user-images.githubusercontent.com/22498/139671110-bd2220e6-1de1-4d5a-8a5f-9fed9faa0313.png)

Also, nothing gets logged to Papertrail or Sentry to so we won't even know that these errors are happening unless users report them to us.

### Not just Blackboard

I'm using Blackboard as an example here but we want a single solution for request and response reporting for validation errors from all external APIs that we talk to (h, Blackboard, LTI Outcomes, Canvas, ...)

### What's Happening (Technical Details)

[...]

PR #3324, opened 2021-11-02
Log and monitor external requests and responses that fail validation

seanh:
Depends on https://github.com/hypothesis/lms/pull/3326. Fixes https://github.com/hypothesis/lms/issues/3319

This works by changing `RequestsResponseSchema` to raise `ExternalRequestError`'s which will be handled by the same `external_request_error()` exception view that handles `ExternalRequestError`'s from `HTTPService`. A new `validation_errors` attr is added to `ExternalRequestError` so that it can record any validation errors along with the request and response.

### Testing

Let's break some of the Blackboard and Canvas schemas to trigger some validation errors:

```diff
diff --git a/lms/services/blackboard_api/_schemas.py b/lms/services/blackboard_api/_schemas.py
index 3f609e14..1c9e8300 100644
--- a/lms/services/blackboard_api/_schemas.py
+++ b/lms/services/blackboard_api/_schemas.py
@@ -36,6 +36,7 @@ class BlackboardPublicURLSchema(RequestsResponseSchema):
     """Schema for Blackboard /courses/{courseId}/resources/{resourceId} responses."""
 
     downloadUrl = fields.Str(required=True)
+    foo = fields.Str(required=True)
 
     @post_load
     def post_load(self, data, **_kwargs):  # pylint:disable=no-self-use
diff --git a/lms/services/canvas_api/client.py b/lms/services/canvas_api/client.py
index 14556e09..985a09db 100644
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -288,6 +288,7 @@ class CanvasAPIClient:
         """Schema for the public_url response."""
 
         public_url = fields.Str(required=True)
+        foo = fields.Str(required=True)
 
     def course_group_categories(self, course_id):
         return self._client.send(
```

Now:

- [x] Launch a Blackboard Files assignment
- [x] Launch any Canvas assignment

In all cases you should see the request, response and validation errors in the error dialog, logs and Sentry. 

Also test:

- [x] Launching Blackboard Files and Canvas assignments with and without an access token in the DB. (_Without_ breaking the code)

### Result

Screenshot from 2021-11-02 13-26-42

Sentry:

Screenshot from 2021-11-02 13-27-47

Papertrail:

```
lms.services.exceptions.ExternalRequestError: ExternalRequestError(message=None, extra_details=None, request=Request(method='GET', url='https://hypothesis.instructure.com/api/v1/files/652/public_url', body=None), response=Response(status_code=200, reason='OK', body='{"public_url":"https://inst-fs-iad-prod.inscloudgate.net/files/46a48d16-c57c-4d22-97f6-d1fd11aba4e2/A%20Third%20File.pdf?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.eyJpYXQiOjE2MzU4NDE5NzIsInVzZXJfaWQiOm51bGwsInJlc291cmNlIjoiL2ZpbGVzLzQ2YTQ4ZDE2LWM1N2MtNGQyMi05N2Y2LWQxZmQxMWFiYTRlMi9BJTIwVGhpcmQlMjBGaWxlLnBkZiIsImp0aSI6ImMwOTA4NDM5LTNlZDctNDYxMC04ODRjLTk5MjBmZWI3ZDM5OCIsImhvc3QiOm51bGwsImV4cCI6MTYzNTkyODM3Mn0.OzXMbBpxSQnrUhz48nH_NkK8_eZbQL0yy2MZL6LM1IMa3HLEKO02gz1HqoaR4vkCVURmulhSlFp8kYqLbw8mQw"}'), validation_errors={'foo': ['Missing data for required field.']})
```

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -1,7 +1,7 @@
 from marshmallow import INCLUDE, fields
 
 from lms.services.exceptions import ExternalRequestError, OAuth2TokenError
-from lms.validation import RequestsResponseSchema, ValidationError
+from lms.validation import RequestsResponseSchema
 
 
 class BlackboardErrorResponseSchema(RequestsResponseSchema):
@@ -38,7 +38,7 @@ class BlackboardErrorResponseSchema(RequestsResponseSchema):
     def parse(self, *args, **kwargs):
         try:
             return super().parse(*args, **kwargs)
-        except ValidationError:
+        except ExternalRequestError:
             return {}
 
 
--- a/lms/services/canvas_api/_basic.py
+++ b/lms/services/canvas_api/_basic.py
@@ -6,8 +6,7 @@ from urllib.parse import urlencode
 import requests
 from requests import RequestException, Session
 
-from lms.services import CanvasAPIError
-from lms.validation import ValidationError
+from lms.services import CanvasAPIError, ExternalRequestError
 
 
 class BasicClient:
@@ -95,8 +94,8 @@ class BasicClient:
         result = None
         try:
             result = schema(response).parse()
-        except ValidationError as err:
-            CanvasAPIError.raise_from(err, request, response)
+        except ExternalRequestError as err:
+            CanvasAPIError.raise_from(err, request, response, err.validation_errors)
 
         # Handle pagination links. See:
         # https://canvas.instructure.com/doc/api/file.pagination.html
--- a/lms/services/oauth_http.py
+++ b/lms/services/oauth_http.py
@@ -1,7 +1,7 @@
 from marshmallow import fields
 
 from lms.services import ExternalRequestError, OAuth2TokenError
-from lms.validation import RequestsResponseSchema, ValidationError
+from lms.validation import RequestsResponseSchema
 from lms.validation.authentication import OAuthTokenResponseSchema
 
 
@@ -104,7 +104,7 @@ class OAuthHTTPService:
         except ExternalRequestError as err:
             try:
                 error_dict = _OAuthAccessTokenErrorResponseSchema(err.response).parse()
-            except ValidationError:
+            except ExternalRequestError:
                 pass
             else:
                 if error_dict["error"] == "invalid_grant":
--- a/lms/validation/_base.py
+++ b/lms/validation/_base.py
@@ -4,6 +4,7 @@ from marshmallow import pre_load
 from pyramid.httpexceptions import HTTPUnsupportedMediaType
 from webargs import pyramidparser
 
+from lms.services import ExternalRequestError
 from lms.validation._exceptions import ValidationError
 
 
@@ -134,14 +135,18 @@ class RequestsResponseSchema(PlainSchema):
         either return the successfully parsed params or raise a validation
         error.
 
-        :raise lms.validation.ValidationError: if the response isn't valid
+        :raise ExternalRequestError: if the response isn't valid
         """
         try:
-            result = self.load(self.context["response"], *args, **kwargs)
+            return self.load(self.context["response"], *args, **kwargs)
         except marshmallow.ValidationError as err:
-            raise ValidationError(messages=err.messages) from err
+            response = self.context.get("response")
+            request = response.request if response is not None else None
+            validation_errors = err.messages
 
-        return result
+            raise ExternalRequestError(
+                request=request, response=response, validation_errors=validation_errors
+            ) from err
 
     @marshmallow.pre_load(pass_many=True)
     def _pre_load(self, response, **_kwargs):  # pylint: disable=no-self-use
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -120,6 +120,7 @@ class APIExceptionViews:
                 "body": self.context.response_body,
             },
         )
+        sentry_sdk.set_context("validation_errors", self.context.validation_errors)
 
         report_exception()
 
@@ -141,6 +142,7 @@ class APIExceptionViews:
                     "status_code": self.context.status_code,
                     "reason": self.context.reason,
                 },
+                "validation_errors": self.context.validation_errors,
             },
         )
 
--- a/tests/unit/lms/validation/_base_test.py
+++ b/tests/unit/lms/validation/_base_test.py
@@ -1,11 +1,14 @@
 import json
+from unittest.mock import sentinel
 
 import pytest
 from marshmallow import fields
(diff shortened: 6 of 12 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -38,7 +38,7 @@ class Meta:
     def parse(self, *args, **kwargs):
         try:
             return super().parse(*args, **kwargs)
-        except ValidationError:
+        except ExternalRequestError:
```
seanh, line 41: The `RequestsResponseSchema` subclass now raises `ExternalRequestError` rather than `ValidationError`, so `BlackboardErrorResponseSchema` has to catch `ExternalRequestError` instead

lms/services/canvas_api/_basic.py
```diff
@@ -95,7 +94,7 @@ def _send_prepared(self, request, schema, request_depth=1):
         result = None
         try:
             result = schema(response).parse()
-        except ValidationError as err:
+        except ExternalRequestError as err:
```
seanh, line 97: `schema` here is an instance of one of the `RequestsResponseSchema` subclass defined in [`services/canvas_api/client.py`](https://github.com/hypothesis/lms/blob/master/lms/services/canvas_api/client.py), or I think it can also be `OAuthTokenResponseSchema` which is also a `RequestsResponseSchema` base class. `RequestsResponseSchema` now raises `ExternalRequestError` rather than `ValidationError` so this just needs to catch that instead

lms/services/oauth_http.py
```diff
@@ -104,7 +104,7 @@ def refresh_access_token(self, token_url, redirect_uri, auth):
         except ExternalRequestError as err:
             try:
                 error_dict = _OAuthAccessTokenErrorResponseSchema(err.response).parse()
-            except ValidationError:
+            except ExternalRequestError:
```
seanh, line 107: `_OAuthAccessTokenErrorResponseSchema` (which is private to this file) is a `RequestsResponseSchema` subclass so it now raises `ExternalRequestError` rather than `ValidationError`

lms/validation/_base.py
```diff
@@ -136,14 +137,18 @@ def parse(self, *args, **kwargs):
         either return the successfully parsed params or raise a validation
         error.
 
-        :raise lms.validation.ValidationError: if the response isn't valid
+        :raise ExternalRequestError: if the response isn't valid
         """
         try:
-            result = self.load(self.context["response"], *args, **kwargs)
+            return self.load(self.context["response"], *args, **kwargs)
```
seanh, line 143: Took the opportunity to stick a `return` here rather than the `return result` that was previously at the bottom of this method (after the whole `try ... except`). Saves a couple of lines of code

lms/validation/_base.py
```diff
@@ -136,14 +137,18 @@ def parse(self, *args, **kwargs):
...
         except marshmallow.ValidationError as err:
-            raise ValidationError(messages=err.messages) from err
+            response = self.context.get("response")
+            request = response.request if response is not None else None
+            validation_errors = err.messages
```
seanh, line 147: I'm not sure if it's strictly necessary but I'm being careful to make sure this code doesn't crash if `response` is None, hence the `request = response.request if response is not None else None`. Some tests fail if you don't do this. I suspect this may be the fault of the tests rather than an actual bug in the production code. But there's no harm in being careful here I guess

lms/validation/_base.py
```diff
@@ -136,14 +137,18 @@ def parse(self, *args, **kwargs):
...
 
-        return result
+            raise ExternalRequestError(
+                request=request, response=response, validation_errors=validation_errors
+            ) from err
```
seanh, line 151: This is where I changed `RequestsResponseSchema` to raise an `ExternalRequestError` with the `request`, `response` and `validation_errors` instead of raising a `ValidationError`. The `ExternalRequestError` will get caught by the `external_request_error()` exception view that reports the `request`, `response` and `validation_errors` to Sentry and the frontend (if nothing in the services or views catches and handles the exception).

lms/validation/authentication/_oauth.py
```diff
@@ -138,9 +138,4 @@ class OAuthTokenResponseSchema(RequestsResponseSchema):
 
     access_token = fields.Str(required=True)
     refresh_token = fields.Str()
-    expires_in = fields.Integer()
-
-    @marshmallow.validates("expires_in")
-    def validate_quantity(self, expires_in):  # pylint:disable=no-self-use
-        if expires_in <= 0:
-            raise marshmallow.ValidationError("expires_in must be greater than 0")
```
seanh, line 146: This `@marshmallow.validates` method would need updating to raise `ExternalRequestError` instead of `ValidationError`, and it'd have to pass the `request`, `response` and `validation_errors` to `ExternalRequestError`. But I realised that there's no need for a `@marshmallow.validates` method here: you can just use Marshmallow's builtin `Range` validator in the `expires_in` field

lms/views/api/exceptions.py
```diff
@@ -120,6 +120,7 @@ def external_request_error(self):
                 "body": self.context.response_body,
             },
         )
+        sentry_sdk.set_context("validation_errors", self.context.validation_errors)
```
seanh, line 123: Send any validation errors to Sentry

lms/views/api/exceptions.py
```diff
@@ -141,6 +142,7 @@ def external_request_error(self):
                     "status_code": self.context.status_code,
                     "reason": self.context.reason,
                 },
+                "validation_errors": self.context.validation_errors,
```
seanh, line 145: Send any validation errors to the frontend for the **Error Details** box

tests/unit/lms/validation/_base_test.py
```diff
@@ -30,3 +33,48 @@ def test_it_fails_without_json_content_type_header(self, content_type):
...
+        schema = self.MySchema(response)
+
+        with pytest.raises(ExternalRequestError) as exc_info:
+            schema.parse()
+
+        exc = exc_info.value
+        assert isinstance(exc, ExternalRequestError)
+        assert exc.request == sentinel.request
+        assert exc.response == response
+        assert exc.validation_errors == validation_errors
```
seanh, line 80: `RequestsResponseSchema` had no unit tests :fearful: It actually does quite a few things since it's responsible for raising if the JSON-parsing fails as well as if the schema validation fails, and it's now responsible for setting `request`, `response` and `validation_errors`. So I added a couple of tests

marcospri (approved):

PR #3326, opened 2021-11-02
Remove `ExternalRequestError.extra_details`

seanh:
`ExternalRequestError` is starting to have too many attributes, especially since I want to add a new `ExternalRequestError.validation_errors` attr in a future PR.

The `ExternalRequestError.extra_details` attr isn't really necessary and isn't used for anything that we really need, so this commit removes it.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -13,17 +13,13 @@ class ExternalRequestError(Exception):
 
     :arg response: The response that was received
     :type response: requests.Response
-
-    :arg extra_details: Extra details about what went wrong, for debugging
-    :type extra_details: JSON-serializable dict
     """
 
-    def __init__(self, message=None, request=None, response=None, extra_details=None):
+    def __init__(self, message=None, request=None, response=None):
         super().__init__()
         self.message = message
         self.request = request
         self.response = response
-        self.extra_details = extra_details
 
     @property
     def url(self) -> Optional[str]:
@@ -77,7 +73,7 @@ class ExternalRequestError(Exception):
         # The name of this class or of a subclass if one inherits this method.
         class_name = self.__class__.__name__
 
-        return f"{class_name}(message={self.message!r}, extra_details={self.extra_details!r}, request={request}, response={response})"
+        return f"{class_name}(message={self.message!r}, request={request}, response={response})"
 
     def __str__(self):
         return repr(self)
@@ -125,12 +121,7 @@ class CanvasAPIError(ExternalRequestError):
         exception_class = cls._exception_class(response)
 
         raise exception_class(
-            message="Calling the Canvas API failed",
-            request=request,
-            response=response,
-            extra_details={
-                "validation_errors": getattr(cause, "messages", None),
-            },
+            message="Calling the Canvas API failed", request=request, response=response
         ) from cause
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


lms/services/exceptions.py
```diff
@@ -125,12 +121,7 @@ def raise_from(cls, cause, request, response):
         exception_class = cls._exception_class(response)
 
         raise exception_class(
-            message="Calling the Canvas API failed",
-            request=request,
-            response=response,
-            extra_details={
-                "validation_errors": getattr(cause, "messages", None),
-            },
```
seanh, line 133: This is some Canvas-specific code that causes the validation errors (but not the request or response) to be sent to the frontend if validating a Canvas API response fails. I'm working on a PR that will send the request, response and validation errors to the frontend, Sentry and Papertrail for all external request validation errors (https://github.com/hypothesis/lms/pull/3324) and that will obsolete this Canvas-specific code

lms/services/exceptions.py
```diff
@@ -199,8 +190,8 @@ class CanvasFileNotFoundInCourse(Exception):
     error_code = "canvas_file_not_found_in_course"
 
     def __init__(self, file_id):
-        self.extra_details = {"file_id": file_id}
-        super().__init__(self.extra_details)
+        self.details = {"file_id": file_id}
+        super().__init__(self.details)
```
seanh, line 194: There are some exception classes where `details` recently got erroneously renamed to `extra_details`: this `CanvasFileNotFoundInCourse` class, the `BlackboardFileNotFoundInCourse` below, and `CanvasGroupError` in another file. These all go to the [catch-all `api_error()` exception view](https://github.com/hypothesis/lms/blob/566cf7084773265314ebe46d51f17bd4510fa4e3/lms/views/api/exceptions.py#L159-L200) which calls [`self.error_response(..., details=context.extra_details)`](https://github.com/hypothesis/lms/blob/566cf7084773265314ebe46d51f17bd4510fa4e3/lms/views/api/exceptions.py#L191) and `error_response()` [uses the given details as the `details` dict in the error response](https://github.com/hypothesis/lms/blob/566cf7084773265314ebe46d51f17bd4510fa4e3/lms/views/api/exceptions.py#L223-L224). So what we're dealing with in this case is called `details`, not `extra_details`

lms/services/lti_outcomes.py
```diff
@@ -137,19 +137,7 @@ def _get_body(cls, data):
...
-                # may not exist.
-                description = header["imsx_POXResponseHeaderInfo"]["imsx_statusInfo"][
-                    "imsx_description"
-                ]
-            except KeyError:
-                pass
-
-            raise ExternalRequestError(
-                message="LTI outcome request failed", extra_details=description
-            )
```
seanh, line 152: This bit of code was putting part of the LTI outcomes API response body into `extra_details` as an attempt to debug something. See <https://github.com/hypothesis/lms/pull/2575>. Since we now report the entire response body to Sentry and Papertrail for all external requests, this is no longer needed.

lms/views/api/exceptions.py
```diff
@@ -188,7 +186,7 @@ def api_error(self):
         if hasattr(self.context, "error_code"):
             return self.error_response(
                 error_code=self.context.error_code,
-                details=getattr(self.context, "extra_details", None),
+                details=getattr(self.context, "details", None),
```
seanh, line 189: `details=details`, rather than `details=extra_details` (this is part of the erroneous renaming of `details` to `extra_details` mentioned above)

marcospri (approved):

PR #3327, opened 2021-11-03
Handle VitalSource assignments in the creation endpoint schema

marcospri:
# Testing

- On https://github.com/hypothesis/lms/pull/3126, try to create a VS assignment. Fail.
- Rebase that branch onto this one, the creation will still fail, but the message won't be marshmallow's validation one.

### DIFF
--- a/lms/validation/_api.py
+++ b/lms/validation/_api.py
@@ -90,10 +90,15 @@ class APICreateAssignmentSchema(PyramidRequestSchema):
             size = fields.Int(required=True)
 
         type = fields.Str(
-            required=True, validate=marshmallow.validate.OneOf(["url", "file"])
+            required=True,
+            validate=marshmallow.validate.OneOf(["url", "file", "vitalsource"]),
         )
         url = fields.Url(required=False, allow_none=True)
+
         file = fields.Nested(File, required=False, allow_none=True)
+        bookID = fields.Str(required=False, allow_none=True)
+        cfi = fields.Str(required=False, allow_none=True)
+        """VitalSource IDs"""
 
     ext_lti_assignment_id = fields.Str(required=True)
     """Canvas only assignment unique identifier"""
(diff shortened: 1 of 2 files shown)


lms/validation/_api.py
```diff
@@ -90,10 +90,15 @@ class File(Schema):
...
-            required=True, validate=marshmallow.validate.OneOf(["url", "file"])
+            required=True,
+            validate=marshmallow.validate.OneOf(["url", "file", "vitalsource"]),
         )
         url = fields.Url(required=False, allow_none=True)
+
         file = fields.Nested(File, required=False, allow_none=True)
+        bookID = fields.Str(required=False, allow_none=True)
+        cfi = fields.Str(required=False, allow_none=True)
+        """VitalSource IDs"""
```
seanh, line 101: I think the `bookID` and `cfi` fields _must_ be present if `type` is `"vitalsource"` and _must not_ be present otherwise. We'd need to add that to the `@validates_schema()` method below I think. Let's make sure we have good unit tests for this as it'd be easy to mess up.

It's getting to the point where I feel we have three separate schemas here (one for URLs, one for files, and one for VitalSource) and we should have a meta-schema that tries to validate the request using each of the sub-schemas in turn. I'm not exactly sure how we'd implement that though.

lms/views/api/assignments.py
```diff
@@ -36,6 +36,11 @@ def create(self):
         elif content_type == "file":
             url = f"canvas://file/course/{params['course_id']}/file_id/{params['content']['file']['id']}"
             extra = {"canvas_file": params["content"]["file"]}
+        elif content_type == "vitalsource":
+            # We still don't store details of VitalSource assignments.
+            # Don't send a real ID back to the frontend to avoid confusion.
+            return {"ext_lti_assignment_id": None}
```
seanh, line 42: So what's the deal here? We're _not_ storing VitalSource assignments in the DB? But we'll have to start storing them before we can actually start using this new API?

lms/validation/_api.py
marcospri: I don't think a few conditionals here are to bad, at least for now where different types don't interact with each other.


>  and must not be present otherwise
Not sure about that one I think if the type says "vitalsource" and we have enough information we should accept it even it also sends something in the URL, same for the other types. I'd `be liberal in what you accept` here.

lms/validation/_api.py
```diff
@@ -108,7 +113,16 @@ class File(Schema):
 
     @validates_schema
     def validate(self, data, **_kwargs):
-        if data["content"]["type"] == "url" and not data["content"].get("url"):
+        content = data["content"]
+        type_ = data["content"]["type"]
```
marcospri, line 117: I'd say this makes the rest easier to read, not sure.

tests/unit/lms/validation/_api_test.py
```diff
@@ -153,17 +153,35 @@ def all_fields(self):
...
-    def test_it_doesnt_raise_valid_url_assignment(self, json_request, url_assignment):
-        request = json_request(url_assignment)
+    @pytest.mark.parametrize(
+        "assignment_fixture_name",
+        ["url_assignment", "file_assignment", "vitalsource_assignment"],
+    )
+    def test_it_doesnt_raise_valid__assignment(
+        self, json_request, assignment_fixture_name, request
+    ):
+        assignment_body = request.getfixturevalue(assignment_fixture_name)
```
marcospri, line 163: Went for this style on the happy path ones

seanh (approved):

PR #3334, opened 2021-11-03
Add some missing test coverage

seanh:
It's important for monitoring to pass the request and response when raising, but these tests weren't covering it.

### DIFF
--- a/tests/unit/lms/services/canvas_api/_basic_test.py
+++ b/tests/unit/lms/services/canvas_api/_basic_test.py
@@ -127,7 +127,7 @@ class TestBasicClient:
         assert exc.request == request
         assert exc.response == response
 
-    @pytest.mark.usefixtures("paginated_results")
+    @pytest.mark.usefixtures("with_paginated_results")
     def test_send_follows_pagination_links_for_many_schema(
         self, basic_client, PaginatedSchema, http_session
     ):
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/canvas_api/_basic_test.py
```diff
@@ -79,19 +79,27 @@ def test_send_raises_CanvasAPIError_for_error_responses(
     ):
         http_session.send.return_value = factories.requests.Response(status_code=501)
 
-        with pytest.raises(CanvasAPIError):
+        with pytest.raises(CanvasAPIError) as exc_info:
             basic_client.send("METHOD", "path/", schema=Schema)
 
+        exc = exc_info.value
+        assert exc.request == http_session.send.call_args[0][0]
```
jon-betts, line 86: This is true and compact, butI found it a bit hard to grok at first pass. Not sure if something like this might help?

```suggestion
        assert exc.request == http_session.send.call_args[0][0]  # First positional arg is request
```

tests/unit/lms/services/canvas_api/_basic_test.py
```diff
@@ -172,6 +186,10 @@ def paginated_results(self, http_session):
             factories.requests.Response(status_code=200),
         ]
 
+    @pytest.fixture
+    def paginated_results(self, http_session, paginated_responses):
```
jon-betts, line 190: I think we normally name these side effect only fixtures something like 'with_paginated_results` to help distinguish them.

jon-betts (approved):
Yep, this looks all fine and dandy. 

The fixturing and assertions here are a bit dense, so some comments might help to explain where things come from.

PR #3343, opened 2021-11-05
POC for using the assignment API in non Canvas LMSs

marcospri:


lms/views/api/canvas/files.py
```diff
@@ -28,7 +28,12 @@ def list_files(self):
         :raise lms.services.CanvasAPIError: if the Canvas API request fails.
             This exception is caught and handled by an exception view.
         """
-        return self.canvas.api.list_files(self.request.matchdict["course_id"])
+        course_id = self.request.matchdict["course_id"]
+        files = self.canvas.api.list_files(course_id)
+        for file in files:
+            file["id"] = f"canvas://file/course/{course_id}/file_id/{file['id']}"
```
marcospri, line 34: I think this probably will be a safer change if we updated put this into "lmsID"/"lmsURI" or similar while keeping the current id field. Same thing on BB ones.

lms/views/basic_lti_launch.py
```diff
@@ -285,7 +285,7 @@ def unconfigured_basic_lti_launch(self):
         ).authorization_param(self.request.lti_user)
 
         self.context.js_config.enable_content_item_selection_mode(
-            form_action=self.request.route_url("configure_assignment"),
+            form_action=self.request.route_url("lti_launches"),
```
marcospri, line 288: Because the creation API has already created the assignment no need to configure it. Not sure if it has any other implications.

lms/validation/_api.py
```diff
@@ -84,10 +84,11 @@ class APICreateAssignmentSchema(PyramidRequestSchema):
 
     class Content(Schema):
```
marcospri, line 85: I think the validation here is a bit lacking with this changes. Some fields are required only in one LMS but not other. This could be  enforced on `@validates_schema`but not sure what the better alternatives are.

- Split the view per LMS? (canvas vs the rest) and use one schema in each view

- Use some sort of polymorphism with marshemllow. This doesn't seem to be supported by the core library  https://github.com/marshmallow-code/apispec/pull/182 and will need to use some other dependency like https://github.com/marshmallow-code/marshmallow-oneofschema

seanh: Yeah, I think splitting the views might be best.

We could have it post to one URL in Canvas and to another URL in non-Canvas, since the backend uses the `JSConfig` to control what URL the frontend calls. Then use different views for different routes. Alternatively we could keep the URL the same but use the `request_param` predicate in `@view_config` to make it use one schema or another depending on a request param.

Either way you could either use two different views or one view with two `@view_config`'s on it:

```python
@view_config(..., schema=APICreateAssignmentSchema)
@view_config(..., schema=APICanvasCreateAssignmentSchema)
def create_assignment(request):
    ...
```

Whatever works best

lms/validation/_api.py
```diff
@@ -84,10 +84,11 @@ class APICreateAssignmentSchema(PyramidRequestSchema):
...
         class File(Schema):
-            display_name = fields.Str(required=True)
-            id = fields.Int(required=True)
-            updated_at = fields.Str(required=True)
-            size = fields.Int(required=True)
+            class Meta:
+                unknown = EXCLUDE
+
+            # VALIDATE STARTS WITH our schemas
+            id = fields.Str(required=True)
```
seanh, line 91: There doesn't seem to be a builtin `startswith()` validator (see the [builtin validators](https://marshmallow.readthedocs.io/en/stable/marshmallow.validate.html)) so I think you'd just add a `@validates` method: https://marshmallow.readthedocs.io/en/stable/quickstart.html#field-validators-as-methods. You can also write your own reusable validator functions but I doubt that'd be worthwhile for something this simple

lms/views/basic_lti_launch.py
seanh: I think this is where more frontend work would be needed here. On this branch when I select a file the frontend both calls the create assignment API and then also makes a form submission to the `/lti_launches` endpoint. This form submission is unnecessary. The frontend should just call the create assignment API, that's all that's necessary to configure the assignment. But _then_ the frontend needs to go ahead and launch the assignment (show the document and the client etc), but without making any form submissions or doing any page reloads. The frontend should have all the information it needs to launch the assignment (e.g. the document URL) from the create assignment API's response

seanh (approved):

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -47,6 +47,9 @@ function contentDescription(content) {
     case 'url':
       return truncateURL(content.url, 65 /* maxLength */);
     case 'file':
+      if (content.file.id.startsWith("blackboard://")){
+            return 'PDF file in BlackBoard';
```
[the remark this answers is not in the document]
marcospri: They show up after you picked the file. On the screen where the group assignment configuration happens.

PR #3351, opened 2021-11-08
BB Groups PoC

marcospri:
3800.6.0
```
/learn/api/public/v2/courses/uuid:{course_id}/groups/sets"
/learn/api/public/v2/courses/uuid:{course_id}/groups/sets/{group_set_id}/groups
/learn/api/public/v2/courses/uuid:{course_id}/groups
```

3600.0.0
```/learn/api/public/v2/courses/uuid:{course_id}/groups/{group['id']}/users/uuid:{user_id}```

`select * from "user" where application_instance_id in (select id from application_instances where tool_consumer_info_version in ('3800.0.0-rel.32+f2f0da9', '3800.0.2-rel.6+f954336'))`

Returns one user created October 30, 2021, 12:02 AM

### CODE REVIEW

File: lms/views/api/blackboard/sync.py [hunk 1]
```diff
@@ -0,0 +1,90 @@
+from lms.security import Permissions
+from lms.services import CanvasAPIError
+from pyramid.view import view_config
+from lms.views import (
+    CanvasGroupSetEmpty,
```
marcospri, line 5: These all say `Canvas` but I think we probably wan't to remove that part and reuse them.

lms/resources/_js_config/__init__.py
```diff
@@ -213,6 +213,12 @@ def enable_content_item_selection_mode(self, form_action, form_fields):
                     "formAction": form_action,
                     "formFields": form_fields,
                     "createAssignmentAPI": self._create_assignment_api(),
+                    # The "content item selection" that we submit to Canvas's
+                    # content_item_return_url is actually an LTI launch URL with
+                    # the selected document URL or file_id as a query parameter. To
+                    # construct these launch URLs our JavaScript code needs the
+                    # base URL of our LTI launch endpoint.
+                    "ltiLaunchUrl": self._request.route_url("lti_launches"),
```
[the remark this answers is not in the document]
seanh: The `ltiLaunchUrl` setting shouldn't be needed outside of Canvas, and yet it seems that the JavaScript code crashes when creating an assignment in a non-Canvas LMS if this setting is missing: `Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL`. @marcospri Can you send a separate PR to make `ltiLaunchUrl` a Canvas-only setting and demonstrate the issue? Then we can ask the frontend team to fix it

lms/resources/_js_config/__init__.py
```diff
@@ -411,6 +417,7 @@ def _get_students(self):
             )
             yield {
                 "userid": h_user.userid(self._authority),
+                "lms_id": grading_info.user_id,
```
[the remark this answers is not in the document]
seanh: :+1:

lms/resources/lti_launch.py
```diff
@@ -126,6 +127,14 @@ def ext_lti_assignment_id(self):
 
         return self._request.POST.get("ext_lti_assignment_id")
 
+    @property
+    def is_blackboard(self):
+
+        return (
```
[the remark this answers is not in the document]
seanh: Blackboard seems to use the exact string `"BlackboardLearn"` for the `tool_consumer_info_product_family_code`, so I think we can do `return product_family_code == "BlackboardLearn"`. I think we could make this an `@property` `ApplicationInstance.is_blackboard() -> bool`. Or possibly a `@property` `ApplicationInstance.product()` that returns one of a fixed set of values like `"canvas"`, `"blackboard"`, etc from an enum.

lms/resources/lti_launch.py
```diff
@@ -210,6 +228,26 @@ def canvas_is_group_launch(self):
...
+
+    @property
+    def is_blackboard_group_launch(self):
+        if self.is_blackboard and not self.blackboard_groups_enabled:
+            return False
+
+        return self.assignment and self.assignment.extra.get("group_set")
+
+    @property
+    def assignment(self):
```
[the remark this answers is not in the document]
seanh: :+1: `get_current()` sounds like a good idea

lms/services/blackboard_api/client.py
```diff
@@ -97,3 +99,54 @@ def public_url(self, course_id, file_id):
...
+                if group["groupSetId"] == group_category_id
+            ]
+
+        return groups
+
+    def user_groups(self, course_id, user_id, group_category_id):
+
+        groups = []
+
+        for group in self.course_groups(course_id, group_category_id):
```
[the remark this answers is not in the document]
seanh: I think the loop that Canvas uses doesn't send an extra HTTP request each time round the loop, though. It's only one HTTP request total (plus possible pagination).

Can you do another version of the POC where we try to update a record of each student's groups in the DB each time they launch an assignment?

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -123,7 +126,7 @@ export default function GroupConfigSelector({
             // Currently all fetch errors are handled by attempting to re-authorize
             // and then re-fetch group sets.
             <>
-              <p>Canvas needs your permission to fetch group sets</p>
+              <p>GENERIC LMS needs your permission to fetch group sets</p>
```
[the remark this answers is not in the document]
seanh: This should say "**Hypothesis** needs your permission to fetch group sets". It isn't Canvas or Blackboard that's trying to fetch the group sets, it's us.

lms/views/api/blackboard/sync.py
seanh: :+1:

seanh (changes requested):

lms/services/blackboard_api/client.py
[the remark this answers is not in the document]
marcospri: Oh yeah I don't mean the line in the for itself but the whole loop will do n-groups + 1 request total (plus pagination)

I'll look into getting the info to/from the DB :+1:  in any case.

lms/resources/_js_config/__init__.py
```diff
@@ -463,48 +470,70 @@ def _hypothesis_client(self):
...
-                    "context_id": req.params["context_id"],
-                    "custom_canvas_course_id": req.params["custom_canvas_course_id"],
+            }
+
+            if "learner_canvas_user_id" in req.params:
+                sync_api_config["data"]["learner"] = {
+                    "canvas_user_id": req.params["learner_canvas_user_id"],
                     "group_set": req.params.get("group_set"),
+                }
+        if self._context.is_blackboard:
```
seanh, line 517: I think we need to only include the sync API config **for Blackboard groups assignments**. Most Blackboard installations won't have the groups feature enabled because they won't have given us access to the Blackboard API, so we don't want to make any requests to the Blackboard API by default.

marcospri: Good point, I think the sync api itself also should check if it's a groups assignment too.

lms/services/blackboard_api/client.py
```diff
@@ -113,4 +114,35 @@ def group_set_groups(self, course_id, group_set_id):
...
+            groups_by_id = {group["id"]: group for group in groups}
+            responses = self._api.async_requests(
+                "GET",
+                [
+                    f"/learn/api/public/v2/courses/uuid:{course_id}/groups/{group['id']}"
+                    for group in groups
+                ],
+            )
+            groups = [
+                groups_by_id[str(response.url).split("/")[-1]]
```
marcospri, line 143: response.url has the full URL (including schema, domain...) instead of that the path we use here.

lms/services/blackboard_api/client.py
```diff
@@ -113,4 +114,35 @@ def group_set_groups(self, course_id, group_set_id):
...
+
+        if group_category_id:
+            groups = [
+                group
+                for group in self.course_groups(course_id)
+                if group["groupSetId"] == group_category_id
+            ]
+
+        if user_id:
+            groups_by_id = {group["id"]: group for group in groups}
```
marcospri, line 134: Create a dict of the groups by ID to retrieve them later as the responses will be out of order.

lms/services/blackboard_api/_basic.py
```diff
@@ -83,6 +83,42 @@ def request(self, method, path):
...
+                    url = self._api_url(url)
+                    task = asyncio.ensure_future(
+                        async_request(session, method, url, headers)
+                    )
+                    tasks.append(task)
+                try:
+                    return await asyncio.gather(*tasks, return_exceptions=False)
+                except ExternalRequestError:
+                    for task in tasks:
+                        task.cancel()
```
marcospri, line 117: Cancel the rest of the request as soon as we fail in one.

lms/services/blackboard_api/_basic.py
```diff
@@ -83,6 +83,42 @@ def request(self, method, path):
...
+        headers["Authorization"] = f"Bearer {access_token}"
+
+        async def async_request(aio_session, method, url, headers=None):
+            try:
+                async with aio_session.request(
+                    method, url, headers=headers
+                ) as response:
+                    return response
+            except aiohttp.ClientError:
+                raise ExternalRequestError("Blackboard async request failed")
```
marcospri, line 102: This won't have as many details as the rest of `ExternalRequestError` (request and responses objects are different form the ones from `requests` so it's not trivial to attach them to the exception) but I hope the fact that we only use this in one place ATM and stack trace in celery would provide enough details.

seanh: I think you probably want an explicit `raise ExternalRequestError(...) from err)`

PR #3360, opened 2021-11-10
Store grouping membership

marcospri:


lms/services/blackboard_api/_schemas.py
```diff
@@ -54,7 +54,7 @@ class Meta:
 
         id = fields.Str(required=True)
         name = fields.Str(required=True)
-        groupSetId = fields.Str(required=True)
+        groupSetId = fields.Str(required=False, allow_none=True)
```
marcospri, line 57: @seanh You mention yesterday that this was a possibility, we don't care about groups with no groupset but the API will return them and we have to filter them out, good find :+1:

lms/services/grouping.py
```diff
@@ -97,6 +100,23 @@ def upsert_canvas_group(  # pylint: disable=too-many-arguments
...
 
+    def upsert_grouping_memberships(self, user: User, groups: List[Grouping]):
+        memberships = []
+        for group in groups:
+            if membership := (
+                self._db.query(GroupingMembership)
+                .filter_by(grouping=group, user=user)
+                .one_or_none()
+            ):
+                memberships.append(membership)
```
marcospri, line 111: I'm not sure this will update the "updated" field which would be interesting to do.

lms/services/user.py
```diff
@@ -36,6 +36,8 @@ def store_lti_user(self, lti_user: LTIUser):
         else:
             self.db_session.add(new_user)
 
+        return existing_user or new_user
```
marcospri, line 39: I'm reusing "store_lti_user" but I might make more sense in a real version to use a `.get_current` or similar that it's lru_cache'ed.

lms/views/api/blackboard/sync.py
```diff
@@ -52,12 +61,27 @@ def _get_blackboard_groups(self):
             if not learner_groups:
                 raise CanvasStudentNotInGroup(group_set=group_set_id)
 
-            return learner_groups
+            groups = self._to_groups_groupings(learner_groups)
+            self._grouping_service.upsert_grouping_memberships(
```
marcospri, line 65: Store the memberships on student launches.

lms/views/api/blackboard/sync.py
```diff
@@ -52,12 +61,27 @@ def _get_blackboard_groups(self):
...
+                self._user_service.store_lti_user(lti_user), groups
+            )
+            return groups
 
         if grading_student_id := self._request.json.get("gradingStudentId"):
-            return self._blackboard_api.user_groups(
-                course_id, grading_student_id, group_set_id
-            )
+            db = self._grouping_service._db
```
marcospri, line 71: All of this belongs to the service but did it here for the PoC

lms/views/api/blackboard/sync.py
```diff
@@ -52,12 +61,27 @@ def _get_blackboard_groups(self):
...
+            course = self._course_service.get(authority_provided_id)
+            return [
+                gm.grouping
+                for gm in db.query(GroupingMembership)
+                .join(User)
+                .filter(
+                    GroupingMembership.grouping_id.in_([g.id for g in course.children]),
+                    User.user_id == grading_student_id,
+                )
+                .options(joinedload("grouping"))
```
marcospri, line 83: Avoids one extra query per membership to get the group.

lms/views/api/blackboard/sync.py
```diff
@@ -52,12 +61,27 @@ def _get_blackboard_groups(self):
...
+            db = self._grouping_service._db
+
+            authority_provided_id = hashed_id(tool_guid, course_id)
+            course = self._course_service.get(authority_provided_id)
+            return [
+                gm.grouping
+                for gm in db.query(GroupingMembership)
+                .join(User)
+                .filter(
+                    GroupingMembership.grouping_id.in_([g.id for g in course.children]),
```
marcospri, line 80: Not sure what would be the right approach but we also want to preload those children in the real version. Maybe a `with_children` bool on `.get()`?

lms/views/api/blackboard/sync.py
```diff
@@ -68,7 +92,7 @@ def _get_blackboard_groups(self):
         if not groups:
             raise CanvasGroupSetEmpty(group_set=group_set_id)
 
-        return groups
+        return self._to_groups_groupings(groups)
```
marcospri, line 95: We are are already returning `Grouping` now.

marcospri: Note that `_to_groups_groupings` uses `upsert_canvas_group`. That's wrong, needs to be adapted for BB.

lms/services/grouping.py
seanh: Hmm. Yeah we might have to add an extra line to update `updated` manually

lms/views/basic_lti_launch.py
```diff
@@ -368,4 +369,9 @@ def configure_assignment(self):
 
         self.context.js_config.maybe_enable_grading()
 
+        # Re-do enable_lti_launch_mode to account for any changes after the
+        # document has been selected
+        JSConfig._hypothesis_client.fget.cache_clear()
```
marcospri, line 374: Need to clear this @lru_cache to get the right "groups" value to the client.

lms/views/basic_lti_launch.py
```diff
@@ -368,4 +369,9 @@ def configure_assignment(self):
...
+        self.context.js_config.enable_lti_launch_mode()
```
marcospri, line 375: Re doing this to get the right value for syncAPI.

PR #3372, opened 2021-11-12
BB groups 

marcospri:


lms/resources/lti_launch.py
```diff
@@ -197,6 +198,30 @@ def canvas_groups_enabled(self):
...
+    @property
+    def is_blackboard_group_launch(self):
+        """Return True if the current assignment uses BlackBoard groups."""
+        if not self.blackboard_groups_enabled:
+            return False
+
+        tool_consumer_instance_guid = self._request.params[
+            "tool_consumer_instance_guid"
+        ]
+        assignment = self._assignment_service.get(
```
marcospri, line 220: I had this as an `assignment` property here and also proposed having a `get_current` on the service itself but at the end inlined it as it depends on `self.resource_link_id` which it made it adding it on the service too coupled to not only the request but an specific context.

lms/resources/_js_config/file_picker_config.py
```diff
@@ -12,12 +12,19 @@ def blackboard_config(cls, _context, request, application_instance):
 
         return {
             "enabled": enabled,
+            "groupsEnabled": context.blackboard_groups_enabled,
```
marcospri, line 15: TODO: missing change on config.js

PR #3375, opened 2021-11-15
Increase default timeouts on HTTPService

marcospri:
Should reduce the occasional timeout:

https://sentry.io/organizations/hypothesis/issues/2771404256/?project=259908&query=is%3Aunresolved#exception

### DIFF
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -31,7 +31,7 @@ class HTTPService:
     def delete(self, *args, **kwargs):
         return self.request("DELETE", *args, **kwargs)
 
-    def request(self, method, url, timeout=(10, 10), **kwargs):
+    def request(self, method, url, timeout=(22, 22), **kwargs):
         """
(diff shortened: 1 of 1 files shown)


seanh (changes requested):
Canvas is unfortunately the one thing that doesn't use `HTTPService`, it still needs to be refactored. It calls requests itself with a slightly shorter timeout:

https://github.com/hypothesis/lms/blob/e005f99cd761d87a2d16aaeb3a4b9b666917f70b/lms/services/canvas_api/_basic.py#L88-L92

I'm not sure about increasing the timeout to 22 seconds anyway. We could try it but that's a long time to wait. I wonder if another timeout somewhere higher up the chain would kick in before then anyway. I also wonder if occupying a web worker for up to 22 seconds my cause problems for other requests that might get queued behind this one

seanh:
New attempt: https://github.com/hypothesis/lms/pull/3393

PR #3383, opened 2021-11-15
Rename the `canvas_file` predicate to `legacy_canvas_file`

seanh:
This predicate is only used for "legacy Canvas File assignments" now.


seanh:
Closing this for now as we want to revisit the whole "store Canvas assignments in the DB" and "make Canvas assignments the same as other LMS's" project in January

PR #3388, opened 2021-11-15
Enable MS OneDrive

marcospri:


seanh (approved):
I hadn't realised that we'd have to duplicate the default value in two different places. What do you think about adding a `onedrive_files_enabled` property to `ApplicationInstance` to remove the duplication?

PR #3393, opened 2021-11-16
Use a longer timeout for "user's groups" requests

seanh:
This is an attempt to fix https://github.com/hypothesis/lms/issues/3392.

Use `(20, 20)` (instead of the default `(10, 10)`) as the `requests` timeout only when making `courses/{course_id}/groups?include[]=users` requests. Keep the default timeout for all other requests.

This requires refactoring `BasicClient` and `AuthenticatedClient` to allow `CanvasAPIClient` to pass in a custom timeout only for the `courses/{course_id}/groups?include[]=users` calls. `CanvasAPIClient` needs to pass the custom timeout to `AuthenticatedClient.send()` which needs to pass it to `BasicClient.send()` which needs to pass it to `BasicClient._send_prepared()` which passes it to `requests.Session.send()`.

### DIFF
--- a/lms/services/canvas_api/_authenticated.py
+++ b/lms/services/canvas_api/_authenticated.py
@@ -3,6 +3,8 @@
 from lms.services import OAuth2TokenError
 from lms.validation.authentication import OAuthTokenResponseSchema
 
+DEFAULT_TIMEOUT = (10, 10)
+
 
 class AuthenticatedClient:
     """
@@ -35,19 +37,19 @@ class AuthenticatedClient:
         self._client_secret = client_secret
         self._redirect_uri = redirect_uri
 
-    def send(self, method, path, schema, params=None):
+    def send(
+        self, method, path, schema, timeout=DEFAULT_TIMEOUT, params=None
+    ):  # pylint:disable=too-many-arguments
         """
         Send a Canvas API request, and retry it if there are OAuth problems.
 
-        :param method: HTTP method to use (e.g. "POST")
-        :param path: The path in the API to make a request to
-        :param schema: Schema to apply to the return values
-        :param params: Any query parameters to add to the request
+        See BasicClient.send() for documentation of parameters, return value
+        and exceptions raised.
+
         :raise OAuth2TokenError: if the request fails because our Canvas API
             access token for the user is missing, expired, or has been deleted
-        :return: JSON deserialised object
         """
-        call_args = (method, path, schema, params)
+        call_args = (method, path, schema, timeout, params)
 
         oauth2_token = self._oauth2_token_service.get()
         access_token = oauth2_token.access_token
@@ -55,8 +57,7 @@ class AuthenticatedClient:
 
         try:
             return self._client.send(
-                *call_args,
-                headers={"Authorization": f"Bearer {access_token}"},
+                *call_args, headers={"Authorization": f"Bearer {access_token}"}
             )
         except OAuth2TokenError:
             if not refresh_token:
@@ -65,8 +66,7 @@ class AuthenticatedClient:
         new_access_token = self.get_refreshed_token(refresh_token)
 
         return self._client.send(
-            *call_args,
-            headers={"Authorization": f"Bearer {new_access_token}"},
+            *call_args, headers={"Authorization": f"Bearer {new_access_token}"}
         )
 
     def get_token(self, authorization_code):
@@ -115,6 +115,7 @@ class AuthenticatedClient:
             url_stub="",
             params=params,
             schema=OAuthTokenResponseSchema,
+            timeout=DEFAULT_TIMEOUT,
         )
 
         self._oauth2_token_service.save(
--- a/lms/services/canvas_api/_basic.py
+++ b/lms/services/canvas_api/_basic.py
@@ -47,6 +47,7 @@ class BasicClient:
         method,
         path,
         schema,
+        timeout,
         params=None,
         headers=None,
         url_stub="/api/v1",
@@ -55,13 +56,28 @@ class BasicClient:
         Make a request to the Canvas API and apply a schema to the response.
 
         :param method: HTTP method to use
+
         :param path: Path fragment to add to the end of the URL
+
         :param schema: A schema object which this request must
+
         :param params: URL parameters to include
+
+        :param timeout: The timeout to pass to requests
+        :type timeout: Either a (connect_timeout, read_timeout) 2-tuple or a
+            single number that will be used for both the connect timeout and
+            the read timeout. The values are in seconds and can be either ints
+            or floats. See:
+            https://docs.python-requests.org/en/latest/user/quickstart/#timeouts
+            https://docs.python-requests.org/en/latest/user/advanced/#timeouts
+
         :param headers: Headers to include
+
         :param url_stub: Path prefix
-        :return: The result of applying the schema to the response
+
         :raise CanvasAPIError: For any validation or request errors
+
+        :return: The result of applying the schema to the response
         """
         # Always request the maximum items per page for requests which return
         # more than one thing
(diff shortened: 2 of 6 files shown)


lms/services/canvas_api/_authenticated.py
```diff
@@ -35,16 +35,31 @@ def __init__(  # pylint: disable=too-many-arguments
         self._client_secret = client_secret
         self._redirect_uri = redirect_uri
 
-    def send(self, method, path, schema, params=None):
+    def send(
+        self, method, path, schema, params=None, timeout=(10, 10)
+    ):  # pylint:disable=too-many-arguments
```
seanh, line 40: I'm not sure that we can avoid the `# pylint:disable=too-many-arguments` here without spending more time refactoring this code than I want to right now. (`HTTPService` avoids this problem by using `**kwargs` to allow arbitary kwargs to be passed through to requests. But it generally has a much "thinner" design where requests is wrapped in fewer layers.)

lms/services/canvas_api/_basic.py
```diff
@@ -50,18 +50,34 @@ def send(
         params=None,
         headers=None,
         url_stub="/api/v1",
+        timeout=(10, 10),
```
seanh, line 53: Hmm. I _think_ `BasicClient` might actually only ever be called by `AuthenticatedClient` which does pass the `timeout` arg, so possibly I can omit the default value here...

lms/services/canvas_api/_basic.py
```diff
@@ -50,18 +50,34 @@ def send(
...
         :param url_stub: Path prefix
-        :return: The result of applying the schema to the response
+
+        :param timeout: The timeout to pass to requests
+        :type timeout: Either a (connect_timeout, read_timeout) 2-tuple or a
+            single number that will be used for both the connect timeout and
+            the read timeout. The values are in seconds and can be either ints
+            or floats. See:
+            https://docs.python-requests.org/en/latest/user/quickstart/#timeouts
+            https://docs.python-requests.org/en/latest/user/advanced/#timeouts
```
seanh, line 76: :man_shrugging: I think the other option would be to remove the whole docstring from `AuthenticatedClient` and have it just refer the reader to `BasicClient`'s docstring. (That's what `OAuthHTTPService` does with `HTTPService`.)

lms/services/canvas_api/client.py
```diff
@@ -319,14 +319,23 @@ def course_groups(self, course_id, only_own_groups=True, include_users=False):
...
         """
         params = {"only_own_groups": only_own_groups}
+        send_kwargs = {}
+
         if include_users:
             params["include[]"] = "users"
 
+            # It looks like Canvas's course groups API may sometimes be very
+            # slow when called with ?include[]=users (possibly for courses that
+            # have many users) so user a larger timeout for these particular
```
seanh, line 329: Typo

lms/services/canvas_api/_authenticated.py
```diff
@@ -35,28 +37,27 @@ def __init__(  # pylint: disable=too-many-arguments
...
         """
         Send a Canvas API request, and retry it if there are OAuth problems.
 
-        :param method: HTTP method to use (e.g. "POST")
-        :param path: The path in the API to make a request to
-        :param schema: Schema to apply to the return values
-        :param params: Any query parameters to add to the request
+        See BasicClient.send() for documentation of parameters, return value
+        and exceptions raised.
```
seanh, line 47: This docstring was mostly a duplication of `BasicClient.send()`'s docstring so I removed it

lms/services/canvas_api/_authenticated.py
seanh: Note that I added the `timeout` argument _before_ the `params` argument which changes the order of the arguments and could possibly break some callers. The only thing that calls this is `CanvasAPIClient` (in `client.py`) and it always passes `params` as a keyword argument, so nothing is broken

lms/services/canvas_api/_basic.py
```diff
@@ -55,13 +56,28 @@ def send(
...
+
         :param headers: Headers to include
+
         :param url_stub: Path prefix
-        :return: The result of applying the schema to the response
+
         :raise CanvasAPIError: For any validation or request errors
+
+        :return: The result of applying the schema to the response
```
seanh, line 80: Moved the `:return:` to after the `:raise:`, which is the correct order

lms/services/canvas_api/_basic.py
```diff
@@ -47,6 +47,7 @@ def send(
         method,
         path,
         schema,
+        timeout,
```
seanh, line 50: I added `timeout` here as a non-optional argument with no default value. `AuthenticatedClient` (above) is the only thing that calls `BasicClient` so it makes sense for the default `timeout` to live in `AuthenticatedClient` (otherwise I'd have to duplicate it). So I changed `AuthenticatedClient` to always pass `timeout` when calling `BasicClient.send()`

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -46,30 +47,32 @@ def test_send_refreshes_and_retries_for_ProxyAPIAccessTokenError(
...
+        call_args = ("METHOD", "/path", sentinel.schema, (10, 10), sentinel.params)
         result = authenticated_client.send(*call_args)
 
-        basic_client.send.assert_has_calls(
-            (
-                # Initial call with access token
-                call(
-                    *call_args,
-                    headers={"Authorization": f"Bearer {oauth_token.access_token}"},
+        assert basic_client.send.call_args_list == [
```
seanh, line 53: I'm not sure why this test was using `assert_has_calls()`, `call_args_list == ...` is the correct thing to use here

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -94,6 +97,7 @@ def test_get_token(
         basic_client.send.assert_called_once_with(
             "POST",
             "login/oauth2/token",
+            url_stub="",
```
seanh, line 100: The arguments were in the wrong order in this `assert_called_once_with()`. I put them in the right order.

tests/unit/lms/services/canvas_api/_authenticated_test.py
```diff
@@ -122,14 +126,15 @@ def test_get_refreshed_token(
...
```
seanh, line 129: Again putting the arguments in the right order

tests/unit/lms/services/canvas_api/_basic_test.py
```diff
@@ -37,37 +37,42 @@ def test_sends_builds_the_right_url(
...
 
     @pytest.mark.parametrize("method", ("GET", "POST", "PATCH"))
     def test_send_uses_the_request_method(
         self, basic_client, Schema, method, http_session
     ):
-        basic_client.send(method, "path/", schema=Schema)
+        basic_client.send(method, "path/", schema=Schema, timeout=sentinel.timeout)
 
-        http_session.send.assert_called_once_with(Any.request(method), timeout=Any())
+        assert http_session.send.call_args[0][0] == Any.request(method)
```
seanh, line 56: `assert_called_once_with(Any.request(method), timeout=Any())` was the wrong thing to use here. This test is only trying to test that the right HTTP method is passed to `http_session.send()` so it only cares that the first positional argument (`call_args[0][0]`) is any request object with the right `method`

tests/unit/lms/services/canvas_api/_basic_test.py
```diff
@@ -131,27 +140,35 @@ def test_send_raises_CanvasAPIError_for_validation_errors(
...
         # mutate the object in place, so the information is destroyed
 
-        http_session.send.assert_has_calls(
-            [
-                call(Any.request(), timeout=Any()),
-                call(Any.request(url="http://example.com/next/0"), timeout=Any()),
-                call(Any.request(url="http://example.com/next/1"), timeout=Any()),
-            ]
-        )
+        assert http_session.send.call_args_list == [
```
seanh, line 153: Fixed the same incorrect use of `assert_has_calls()` again

marcospri (approved):
Look good, tested both the regular timeout and the increased one code paths but we can't really test if 20 is the number that will make it not timeout, it's twice the default so let's hope it's enough :+1:

PR #3394, opened 2021-11-16
Add product property to ApplicationInstance

marcospri:


lms/models/application_instance.py
```diff
@@ -18,6 +19,16 @@
...
 
+    class Product(str, Enum):
+        BLACKBOARD = "BlackboardLearn"
+        CANVAS = "canvas"
+        MOODLE = "moodle"
+        D2L = "desire2learn"
+        BLACKBAUD = "BlackbaudK12"
+        ILIAS = "ilias"
+        SAKAI = "sakai"
+        UNKNOWN = "unknown"
```
marcospri, line 30: All values from the DB

lms/models/application_instance.py
```diff
@@ -18,6 +19,16 @@
 class ApplicationInstance(BASE):
     """Class to represent a single lms install."""
 
+    class Product(str, Enum):
```
marcospri, line 22: I prefer this to be nested as you can do check against the Enum without needing to import anything new

marcospri: (str,Enum) makes this a str too so: 

Product("canvas") == Product.CANVAS == "canvas"

lms/models/application_instance.py
seanh: Can we remove Ilias? I have no idea what that is! And there's only one instance of it in the DB.

There's also "schoology" (seven instances in the DB). We may want to add that

seanh (approved):

PR #3395, opened 2021-11-16
New get method on UserService

marcospri:
Note that the service is no longer a dataclass as it was not possible to use `lru_cache` without extra modification.

Between having to either have to play with different dataclass option and/or having to add a __hash__ method I prefer switching to a traditional class, there's no really data here. Renamed a db_session to db and added `_` for all fields to match other services.

### DIFF
--- a/lms/services/user.py
+++ b/lms/services/user.py
@@ -1,14 +1,15 @@
-from dataclasses import dataclass
+from functools import lru_cache
 from typing import Optional
 
-from sqlalchemy.orm import Session
+from sqlalchemy.exc import NoResultFound
 
-from lms.models import LTIUser
-from lms.models.user import User
-from lms.services.application_instance import ApplicationInstanceService
+from lms.models import LTIUser, User
 
 
-@dataclass
+class UserNotFound(Exception):
+    """The requested User wasn't found in the database."""
+
+
 class UserService:
     """
     A service for working with users.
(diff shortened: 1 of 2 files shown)


lms/services/user.py
```diff
@@ -34,11 +34,33 @@ def store_lti_user(self, lti_user: LTIUser):
             existing_user.roles = new_user.roles
 
         else:
-            self.db_session.add(new_user)
+            self._db.add(new_user)
+
+    @lru_cache
+    def get(self, application_instance, user_id: str) -> User:
```
marcospri, line 40: `user_id` here is `User.user_id`, we pass this instead of a `lti_user` as we will use this with both a `lti_user` (where we can do `lti_user.user_id`) an also while grading and in that case the `user_id` is all we have (and we don't really want to construct a fake LTIUser for this)

lms/services/user.py
```diff
@@ -1,24 +1,24 @@
...
 
-from lms.models import LTIUser
-from lms.models.user import User
-from lms.services.application_instance import ApplicationInstanceService
+
+class UserNotFound(Exception):
+    """The requested User wasn't found in the database."""
 
 
-@dataclass
```
seanh, line 11: Yep. Dataclasses are great but lets not abuse or overuse them. A service really probably shouldn't be a dataclass

lms/services/user.py
```diff
@@ -34,11 +34,33 @@ def store_lti_user(self, lti_user: LTIUser):
...
+            .filter_by(
+                application_instance=application_instance,
+                user_id=user_id,
+            )
+            .one_or_none()
+        )
+        if not existing_user:
+            raise UserNotFound()
+
+        return existing_user
```
seanh, line 59: I think you might be able to use [`one()`](https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=one_or_none#sqlalchemy.engine.CursorResult.one) here instead for slightly simpler code:

```python
try:
    return self._db.query(User).filter_by(...).one()
except NoResultFound as err:
    raise UserNotFound() from err
```

seanh (approved):

PR #3405, opened 2021-11-17
Add CourseService.generate_authority_provided_id

marcospri:
This was a bit scattered around. 

We don't usually receive an authority_provided_id so calling the course service it's a bit awkward, construct the authority_provided_id then call get with it so maybe another version of `get` could be useful but I reckon this goes in the right direction anyway.

### DIFF
--- a/tests/unit/lms/services/course_test.py
+++ b/tests/unit/lms/services/course_test.py
@@ -112,6 +112,12 @@ class TestCourseService:
         assert db_session.query(Course).count() == 1
         assert course.authority_provided_id == "new authority_provided_id"
 
+    def test_generate_authority_provided_id(self, svc):
+        assert (
+            svc.generate_authority_provided_id("tool", "context_id")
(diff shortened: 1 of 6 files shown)


tests/unit/lms/services/course_test.py
```diff
@@ -112,6 +112,12 @@ def test_upsert_creates_new(self, svc, db_session):
         assert db_session.query(Course).count() == 1
         assert course.authority_provided_id == "new authority_provided_id"
 
+    def test_generate_authority_provided_id(self, svc):
```
marcospri, line 115: You could argue this is not a real unit test and I'm testing hashed_id instead of `generate_authority_provided_id`... and that's probably true but I think the alternative  wouldn't really test much I could change it to use

```
 @pytest.fixture
    def hashed_id(self, patch):
        return patch("lms.resources.lti_launch.hashed_id")
```

if we prefer that

seanh (approved):

PR #3406, opened 2021-11-17
Merges upsert_canvas_section and upsert_canvas_group methods

marcospri:
Merges `upsert_canvas_section` and `upsert_canvas_group` methods of the grouping service into one.

This is in anticipation for what would become `upsert_blackboard_group`

- Renamed the existing upsert to `_upsert`
- New method is named `upsert_with_parent` and takes a non optional parent grouping.
- `upsert_with_parent` generates the authority_provided_id based on a `type_` param. It doesn't handle `type_` as in the current data model course don't have a parent grouping.

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -1,64 +1,13 @@
 from typing import Optional
 
-from lms.models import CanvasGroup, CanvasSection, Grouping
 from lms.models._hashed_id import hashed_id
+from lms.models.grouping import Grouping
 
 
 class GroupingService:
-    def __init__(self, db, application_instance_service, course_service):
+    def __init__(self, db, application_instance_service):
         self._db = db
         self._application_instance = application_instance_service.get_current()
-        self._course_service = course_service
-
-    def upsert(self, grouping):
-        db_grouping = (
-            self._db.query(Grouping)
-            .filter_by(
-                application_instance=grouping.application_instance,
-                authority_provided_id=grouping.authority_provided_id,
-            )
-            .one_or_none()
-        )
-        if not db_grouping:
-            self._db.add(grouping)
-        else:
-            # Update any fields that might have changed
-            db_grouping.lms_name = grouping.lms_name
-            db_grouping.extra = grouping.extra
-
-        return db_grouping or grouping
-
-    def upsert_canvas_section(
-        self, tool_consumer_instance_guid, context_id, section_id, section_name
-    ):
-        """
-        Upsert a Grouping for a course section.
-
-        :param tool_consumer_instance_guid: Tool consumer GUID
-        :param context_id: Course id the section is a part of
-        :param section_id: A section id for a section group
-        :param section_name: The name of the section
-        """
-
-        section_authority_provided_id = hashed_id(
-            tool_consumer_instance_guid, context_id, section_id
-        )
-
-        course = self._course_service.get(
-            self._course_service.generate_authority_provided_id(
-                tool_consumer_instance_guid, context_id
-            )
-        )
-
-        return self.upsert(
-            CanvasSection(
-                application_instance=self._application_instance,
-                authority_provided_id=section_authority_provided_id,
-                lms_id=section_id,
-                lms_name=section_name,
-                parent_id=course.id,
-            )
-        )
 
     @staticmethod
     def generate_authority_provided_id(
@@ -80,49 +29,56 @@ class GroupingService:
             tool_consumer_instance_guid, parent.lms_id, type_.value, lms_id
         )
 
-    def upsert_canvas_group(  # pylint: disable=too-many-arguments
+    def upsert_with_parent(  # pylint: disable=too-many-arguments
         self,
         tool_consumer_instance_guid,
-        context_id,
-        group_id,
-        group_name,
-        group_set_id,
+        lms_id,
+        lms_name,
+        parent: Grouping,
+        type_: Grouping.Type,
+        extra=None,
     ):
         """
-        Upserst a Grouping for a canvas group.
+        Upsert a Grouping generating the authority_provided_id based on its parent.
 
         :param tool_consumer_instance_guid: Tool consumer GUID
-        :param context_id: Course id the group is a part of
-        :param group_id: Canvas group id
-        :param group_name: The name of the group
-        :param group_set_id: Id of the canvas group set this group belongs to
+        :param lms_id: ID of this grouping on the LMS
+        :param lms_name: Name of the grouping on the LMS
+        :param parent: Parent of grouping
+        :param type_: Type of the grouping
+        :param extra: Any extra information to store linked to this grouping
         """
-
-        group_authority_provided_id = hashed_id(
-            tool_consumer_instance_guid, context_id, "canvas_group", group_id
+        authority_provided_id = self.generate_authority_provided_id(
+            tool_consumer_instance_guid, lms_id, parent, type_
         )
 
-        course_authority_provided_id = hashed_id(
-            tool_consumer_instance_guid, context_id
-        )
-
-        course = self._course_service.get(course_authority_provided_id)
-
-        return self.upsert(
-            CanvasGroup(
+        grouping = (
+            self._db.query(Grouping)
+            .filter_by(
                 application_instance=self._application_instance,
-                authority_provided_id=group_authority_provided_id,
-                lms_id=group_id,
-                lms_name=group_name,
-                parent_id=course.id,
-                extra={"group_set_id": group_set_id},
+                authority_provided_id=authority_provided_id,
             )
+            .one_or_none()
         )
+
+        if not grouping:
+            grouping = Grouping(
+                application_instance=self._application_instance,
+                authority_provided_id=authority_provided_id,
+                lms_id=lms_id,
+                parent_id=parent.id,
+                type=type_,
+            )
+            self._db.add(grouping)
+
+        grouping.lms_name = lms_name
+        grouping.extra = extra
+
+        return grouping
 
 
 def factory(_context, request):
     return GroupingService(
         request.db,
         request.find_service(name="application_instance"),
-        request.find_service(name="course"),
     )
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -1,10 +1,8 @@
 from unittest.mock import sentinel
 
 import pytest
-from sqlalchemy.orm import make_transient
 
-from lms.models import Grouping
-from lms.models._hashed_id import hashed_id
+from lms.models import CanvasGroup, Grouping
 from lms.services.grouping import GroupingService, factory
 from tests import factories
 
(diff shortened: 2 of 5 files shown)


lms/services/grouping.py
```diff
@@ -26,74 +26,48 @@ def upsert(self, grouping):
...
-
-        section_authority_provided_id = hashed_id(
-            tool_consumer_instance_guid, context_id, section_id
-        )
-
-        course = self._course_service.get(
-            self._course_service.generate_authority_provided_id(
-                tool_consumer_instance_guid, context_id
+        if type_ == Grouping.Type.CANVAS_SECTION:
+            return hashed_id(tool_consumer_instance_guid, parent.lms_id, lms_id)
```
marcospri, line 34: Sections and courses generated the authority IDs before groupings were a thing so they use the specific order of params (and don't include the type)

lms/services/grouping.py
```diff
@@ -26,74 +26,48 @@ def upsert(self, grouping):
...
-        return self.upsert(
-            CanvasSection(
-                application_instance=self._application_instance,
-                authority_provided_id=section_authority_provided_id,
-                lms_id=section_id,
-                lms_name=section_name,
-                parent_id=course.id,
-            )
-        )
+        return hashed_id(tool_consumer_instance_guid, parent.lms_id, type_, lms_id)
```
marcospri, line 40: This is the schema CANVAS_GROUPS (and BLACKBOARD_GROUPS) use.

lms/services/grouping.py
```diff
@@ -26,74 +26,48 @@ def upsert(self, grouping):
...
 
-    def upsert_canvas_group(  # pylint: disable=too-many-arguments
+    def upsert_with_parent(  # pylint: disable=too-many-arguments
```
marcospri, line 42: Not sure about this name. 

I don't want to merge this with `upsert` as I prefer the service clients don't have to bother constructing Grouping objects themselves.

lms/services/grouping.py
```diff
@@ -1,12 +1,12 @@
-from lms.models import CanvasGroup, CanvasSection, Grouping
 from lms.models._hashed_id import hashed_id
+from lms.models.grouping import CanvasGroup, Grouping
+from lms.services.course import CourseService
 
 
 class GroupingService:
-    def __init__(self, db, application_instance_service, course_service):
+    def __init__(self, db, application_instance_service):
```
seanh, line 7: So `GroupingService` no longer needs `course_service`, ...

lms/services/grouping.py
seanh: It looks to me like nothing outside of this service actually calls the `upsert()` method so I think you could make it a private `_upsert()` method and move it to the bottom of the class, then you could name the public method `upsert()`:

```python
class GroupingService:
    ...

    def upsert(  # pylint: disable=too-many-arguments
        self,
        tool_consumer_instance_guid,
        lms_id,
        lms_name,
        parent: Grouping,
        type_: Grouping.Type,
        extra=None,
    ):
        return self._upsert(...)

    def _upsert(self, grouping):
        ...
``` 

This would mean that the `_upsert()` method would have to be unit-tested indirectly via `upsert()` (to avoid the tests calling a private method) but that's probably a good thing since `_upsert()` is really just an implementation details of `upsert()`

seanh: I wonder about the `# pylint: disable=too-many-arguments` though. Why don't you want the caller to have to construct `Grouping` objects? Instead of:

```python
grouping_service.upsert(
    tool_consumer_instance_guid=tool_consumer_instance_guid,
    lms_id=lms_id,
    lms_name=lms_name,
    parent=parent_grouping,
    type_=Grouping.Type.*,
    extra={whatever},
)
```

It'd be:

```python
grouping_service.upsert(
    Grouping(
        tool_consumer_instance_guid=tool_consumer_instance_guid,
        lms_id=lms_id,
        lms_name=lms_name,
        parent=parent_grouping,
        type_=Grouping.Type.*,
        extra={whatever},
    )
)
```

It shouldn't be _too_ much harder from the point of view of the caller and it avoids the `too-many-arguments`.

`too-many-arguments` is a common issue with services when the caller passes in all the arguments directly and the service constructs the model object. And I think having the caller pass in the model object is probably the solution.

Having said that, it'd probably take some more work to make passing-in-the-model work as nicely as it does in my example above, so I think just leaving things as they are is probably the most pragmatic solution right now.

lms/services/grouping.py
```diff
@@ -26,74 +26,48 @@ def upsert(self, grouping):
...
-        )
-
-        course_authority_provided_id = hashed_id(
-            tool_consumer_instance_guid, context_id
-        )
-
-        course = self._course_service.get(course_authority_provided_id)
-
         return self.upsert(
             CanvasGroup(
```
seanh, line 62: It looks like `upsert_with_parent()` _always_ creates a `CanvasGroup` object?

seanh (changes requested):
I noticed that `GroupingService.upsert_with_parent()` always seems to create `CanvasGroup` models. Doesn't it need to use `CanvasSection` or `BlackboardGroup` depending on the type?

lms/services/grouping.py
marcospri: > It looks like `upsert_with_parent()` _always_ creates a `CanvasGroup` object?

Oh this is very wrong, I'll check why a bunch of tests haven't exploded.

lms/services/grouping.py
marcospri: Made _upsert private :+1: 

I don't think it has many practical implications here but I rather the rest of the code not construct ORM objects outside the service so that becomes the responsibility of the service, if the views construct the object they might as well `db_session.add()` them.

To be fair, in this case the service it's not doing that much  (calling _upsert, generating the authority provided id)

seanh (changes requested):
Can we add something to the PR description explaining what the motiviation for this refactoring is?

Some specific suggestions in the comments:

1. Can `GroupingService.generate_authority_provided_id()` be private?

2. Can the `if type_ == Grouping.Type.COURSE` branch be removed from `GroupingService.generate_authority_provided_id()`? I don't think it's ever used

3. Can the two `parent.lms_id if parent else None` branches (in `generate_authority_provided_id()` and `upsert_with_parent()`) be removed? I don't think the `else None` branches are ever taken

4. Can `upsert_with_parent()` be renamed to `upsert()`?

### Notes

The diff is quite hard to follow so I'm trying to summarize what changed here:

1. `GroupingService.upsert()` was only being called by `GroupingService` itself, so it has become a private method `_upsert()` :+1: (Could be moved to a separate PR to make review easier)

2. Instead of `GroupingService.upsert_canvas_section()` and `GroupingService.upsert_canvas_group()` we now have a single `GroupingService.upsert_with_parent()` with a `type_` parameter.
   
   The way that `GroupingService.upsert_with_parent()` seems to work is that upserting a grouping is the same for any type of grouping except that:
   
   1. A grouping can either have a parent or not: `parent_id=parent.id if parent else None`. Is this necessary? Is `upsert_with_parent()` ever actually called with `parent=None`?

   2. `upsert_with_parent()` calls the new `GroupingService.generate_authority_provided_id()` which behaves differently for different types of group:
   
      1. If `type_` is `CANVAS_SECTION` it does `hashed_id(tool_consumer_instance_guid, parent.lms_id, lms_id)`

      2. If `type_` is `COURSE` it does `CourseService.generate_authority_provided_id(tool_consumer_instance_guid, lms_id)`. I don't think this branch is ever used though? Does anything call `upsert_with_parent()` with `type_=COURSE`?
      
      3. Otherwise it does `hashed_id(tool_consumer_instance_guid, parent.lms_id if parent else None, type_, lms_id)`. This will happen if `type_` is `CANVAS_GROUP` or presumably in future for Blackboard groups and other future types of grouping.
      
         There's again a branch where where `parent` can be `None`: `parent.lms_id if parent else None`. But I don't think this branch is ever used? Does anything call `upsert_with_parent()` with `parent=None`?

I can see some benefits to this refactoring:

1. Benefit: not having to add a new `GroupingService.upsert_blackboard_group()` method and more generally to avoid having to add a new `GroupingService.upsert_*()` method for each new type of grouping.

2. Benefit: having all the different ways of generating `authority_provided_id`'s for different types of grouping collected together into one place (in `GroupingService.generate_authority_provided_id()`) rather than spread around. This benefit doesn't seem to be fully achieved though, because there's still a separate `CourseService.generate_authority_provided_id()` for course groupings.

lms/services/grouping.py
marcospri: > I think we can rename upsert_with_parent() to just upsert() now?

I  think if the public method was called `upsert` I would expect it to work with groupings that don't have parents (courses). I took the route of keeping the `_with_parent` naming but removing support for when parent==None which added a few branches in the code.

lms/services/grouping.py
```diff
@@ -1,106 +1,90 @@
...
-            # Update any fields that might have changed
-            db_grouping.lms_name = grouping.lms_name
-            db_grouping.extra = grouping.extra
-
-        return db_grouping or grouping
-
-    def upsert_canvas_section(
-        self, tool_consumer_instance_guid, context_id, section_id, section_name
+    @staticmethod
+    def generate_authority_provided_id(
```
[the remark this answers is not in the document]
marcospri: I inlined it for now.

lms/services/grouping.py
```diff
@@ -1,106 +1,90 @@
...
-            CanvasSection(
-                application_instance=self._application_instance,
-                authority_provided_id=section_authority_provided_id,
-                lms_id=section_id,
-                lms_name=section_name,
-                parent_id=course.id,
-            )
+        return hashed_id(
+            tool_consumer_instance_guid,
+            parent.lms_id if parent else None,
```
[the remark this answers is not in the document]
marcospri: No currently, removed.

seanh (changes requested):
Testing this branch versus master I think this branch has changed how `authority_provided_id`'s are computed for Canvas groups: the code results in different `authority_provided_id`'s for the same groups on this branch versus on master.

I think the reason might be that master uses the string `"canvas_group"` when computing the `authority_provided_id` for a group: https://github.com/hypothesis/lms/blob/3cc7de6778d3b95e8fd26cadc477b66e072b1b05/lms/services/grouping.py#L79-L81

Whereas on this branch it uses `Grouping.Type.CANVAS_GROUP` which is an enum member not a string so it has a different value.

We need to add a failing test that is failing with the code currently on this branch and would pass if the `authority_provided_id` generation were fixed

tests/unit/lms/services/grouping_test.py
```diff
@@ -15,100 +14,80 @@ class TestGroupingService:
     CONTEXT_ID = "context_id"
     TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
 
-    def test_upsert_inserts(self, svc, db_session):
-        # Start with no groupings
-        assert not db_session.query(Grouping).count()
+    def test_upsert_with_parents_inserts(self, svc, db_session):
+        course = factories.Course()
+        application_instance = factories.ApplicationInstance()
```
seanh, line 19:
```suggestion
```

tests/unit/lms/services/grouping_test.py
```diff
@@ -15,100 +14,80 @@ class TestGroupingService:
...
 
-        test_grouping = Grouping(
-            application_instance=factories.ApplicationInstance(),
-            authority_provided_id="ID",
+        # Start with no CanvasGroup
+        assert not db_session.query(CanvasGroup).count()
+
+        test_grouping = svc.upsert_with_parent(
+            tool_consumer_instance_guid=application_instance.tool_consumer_instance_guid,
```
seanh, line 25: You can use the course's application instance (which I guess is what would happen in reality as well). Then you don't need the `application_instance` variable above:

```suggestion
            tool_consumer_instance_guid=course.application_instance.tool_consumer_instance_guid,
```

tests/unit/lms/services/grouping_test.py
```diff
@@ -15,100 +14,80 @@ class TestGroupingService:
...
         )
 
-        svc.upsert(test_grouping)
-
-        assert db_session.query(Grouping).one() == test_grouping
+        assert db_session.query(CanvasGroup).one() == test_grouping
 
-    def test_upsert_updates(self, svc, db_session):
-        grouping = factories.Grouping()
+    def test_upsert_with_parents_updates(self, svc, db_session, application_instance):
```
seanh, line 34: This test doesn't seem right. If I break the code so that it actually doesn't update anything the test still passes:

```diff
diff --git a/lms/services/grouping.py b/lms/services/grouping.py
index 48d26486..fc647c03 100644
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -73,10 +73,6 @@ class GroupingService:
         )
         if not db_grouping:
             self._db.add(grouping)
-        else:
-            # Update any fields that might have changed
-            db_grouping.lms_name = grouping.lms_name
-            db_grouping.extra = grouping.extra
 
         return db_grouping or grouping
 
```

It looks like the test code itself actually updates the grouping:

https://github.com/hypothesis/lms/blob/fa8ee0f917db775b9989c05d6d0c8c652067d751/tests/unit/lms/services/grouping_test.py#L49-L50

Here's a replacement test that works. It works by calling `upsert_with_parent()` twice which in a way isn't ideal, but it leads to a simple test that does fail if the code is wrong:

```python
    def test_upsert_with_parents_updates(self, svc, db_session, application_instance):
        course = factories.Course()
        tool_consumer_instance_guid = course.application_instance.tool_consumer_instance_guid
        lms_id = "lms_id"
        old_name = "old_name"
        old_extra = {"extra": "old"}
        new_name = "new_name"
        new_extra = {"extra": "new"}

        # Insert a new grouping.
        svc.upsert_with_parent(
            tool_consumer_instance_guid=tool_consumer_instance_guid,
            lms_id=lms_id,
            lms_name=old_name,
            parent=course,
            type_=Grouping.Type.CANVAS_GROUP,
            extra=old_extra,
        )

        # Update the previously-inserted grouping with new values.
        grouping = svc.upsert_with_parent(
            tool_consumer_instance_guid=tool_consumer_instance_guid,
            lms_id=lms_id,
            lms_name=new_name,
            parent=course,
            type_=Grouping.Type.CANVAS_GROUP,
            extra=new_extra,
        )

        # upsert_with_parent() should return a Grouping that has the updated values.
        assert grouping.lms_name == new_name
        assert grouping.extra == new_extra

        # The Grouping's values should have been updated in the DB as well.
        db_grouping = db_session.query(CanvasGroup).one()
        assert db_grouping.lms_name == new_name
        assert db_grouping.extra == new_extra
```

lms/services/grouping.py
```diff
@@ -80,49 +29,60 @@ def generate_authority_provided_id(
...
+            .one_or_none()
         )
+        if not db_grouping:
+            self._db.add(grouping)
+        else:
+            # Update any fields that might have changed
+            db_grouping.lms_name = grouping.lms_name
+            db_grouping.extra = grouping.extra
+
+        return db_grouping or grouping
```
seanh, line 81: I'm not sure that `_upsert()` really needs to be a separate method anymore since it's only ever called by the `upsert_with_parent()` method right above it, and if you inline it the code can be simpler. This passes all the tests, and it's particularly nice that there's only one `Grouping` object involved:

```python
class GroupingService:
    ...

    def upsert_with_parent(  # pylint: disable=too-many-arguments
        self,
        tool_consumer_instance_guid,
        lms_id,
        lms_name,
        parent: Grouping,
        type_: Grouping.Type,
        extra=None,
    ):
        authority_provided_id = self.generate_authority_provided_id(
            tool_consumer_instance_guid, lms_id, parent, type_
        )

        grouping = (
            self._db.query(Grouping)
            .filter_by(
                application_instance=self._application_instance,
                authority_provided_id=authority_provided_id,
            )
            .one_or_none()
        )

        if not grouping:
            grouping = Grouping(
                application_instance=self._application_instance,
                authority_provided_id=authority_provided_id,
                lms_id=lms_id,
                parent_id=parent.id,
                type=type_,
            )
            self._db.add(grouping)

        grouping.lms_name = lms_name
        grouping.extra = extra

        return grouping
```

seanh (changes requested):
Re-tested and it works for me. Some suggestions for the code and tests

tests/unit/lms/services/grouping_test.py
marcospri: Great thanks for the new test version.

The one I had broke (as in it wasn't testing the real  code) after removing `make_transient(grouping)`

lms/services/grouping.py
marcospri: Inlined upsert, we might have to recover it later down the line but not for blackboard groups and we might need a slightly different method so this is better and cleaner for now.

seanh (approved):

PR #3408, opened 2021-11-17
Support VS launches in SpeedGrader

marcospri:
# Testing

- Create a new VitalSource assignment (for example 0077383966) so there's no SG submissions
- Launch in SpeedGrader, there won't be any submissions.
- Launch as a student
- Back at SpeedGrader, refresh, there will be one submission now.


seanh (approved):
I've tested this locally and it works for me, and the backend changes look good :+1:

PR #3414, opened 2021-11-19
Reorg application_instances fixtures to work together

marcospri:
application_instance_service returns application_instance on get_current
otherwise two different, conflicting application_instances rows are
created on the DB.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -7,6 +7,7 @@ from pyramid import testing
 from pyramid.request import apply_request_extensions
 
 from lms.db import SESSION
+from lms.models import ApplicationSettings
 from tests import factories
 from tests.conftest import TEST_SETTINGS, get_test_database_url
 from tests.unit.services import *  # pylint: disable=wildcard-import,unused-wildcard-import
@@ -179,6 +180,9 @@ def httpretty_():
 def application_instance(pyramid_request):
     return factories.ApplicationInstance(
         consumer_key=pyramid_request.lti_user.oauth_consumer_key,
+        developer_key="TEST_DEVELOPER_KEY",
+        provisioning=True,
+        settings=ApplicationSettings({}),
     )
 
 
--- a/tests/unit/lms/resources/_js_config/__init___test.py
+++ b/tests/unit/lms/resources/_js_config/__init___test.py
@@ -9,7 +9,6 @@ from lms.resources._js_config import JSConfig
 from lms.services import ApplicationInstanceNotFound, HAPIError
 
 pytestmark = pytest.mark.usefixtures(
-    "application_instance_service",
     "grading_info_service",
     "grant_token_service",
     "h_api",
@@ -17,6 +16,7 @@ pytestmark = pytest.mark.usefixtures(
 )
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestEnableContentItemSelectionMode:
     def test_it(self, js_config):
         js_config.enable_content_item_selection_mode(
@@ -114,6 +114,7 @@ class TestEnableContentItemSelectionMode:
         return patch("lms.resources._js_config.FilePickerConfig")
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestEnableLTILaunchMode:
     def test_it(self, bearer_token_schema, context, grant_token_service, js_config):
         js_config.enable_lti_launch_mode()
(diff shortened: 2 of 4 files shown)


tests/unit/conftest.py
```diff
@@ -178,7 +179,12 @@ def httpretty_():
 @pytest.fixture
 def application_instance(pyramid_request):
     return factories.ApplicationInstance(
-        consumer_key=pyramid_request.lti_user.oauth_consumer_key,
+        consumer_key=pyramid_request.lti_user.oauth_consumer_key
+        if pyramid_request.lti_user
+        else None,
```
seanh, line 184: I don't think the `if ... else None` should be necessary here: the production code should never be creating an `ApplicationInstance` with no `consumer_key`, so the tests shouldn't need to either. `consumer_key` is non-nullable on `ApplicationInstance`.

The problem seems to be that some of the `JSConfig` tests are activating the `application_instance_service` fixture when that fixture isn't actually used by the code under test:

```diff
diff --git a/tests/unit/conftest.py b/tests/unit/conftest.py
index 275097b7..ff3a5d0f 100644
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -179,9 +179,7 @@ def httpretty_():
 @pytest.fixture
 def application_instance(pyramid_request):
     return factories.ApplicationInstance(
-        consumer_key=pyramid_request.lti_user.oauth_consumer_key
-        if pyramid_request.lti_user
-        else None,
+        consumer_key=pyramid_request.lti_user.oauth_consumer_key,
         developer_key="TEST_DEVELOPER_KEY",
         provisioning=True,
         settings=ApplicationSettings({}),
diff --git a/tests/unit/lms/resources/_js_config/__init___test.py b/tests/unit/lms/resources/_js_config/__init___test.py
index caf97589..5954d857 100644
--- a/tests/unit/lms/resources/_js_config/__init___test.py
+++ b/tests/unit/lms/resources/_js_config/__init___test.py
@@ -9,7 +9,6 @@ from lms.resources._js_config import JSConfig
 from lms.services import ApplicationInstanceNotFound, HAPIError
 
 pytestmark = pytest.mark.usefixtures(
-    "application_instance_service",
     "grading_info_service",
     "grant_token_service",
     "h_api",
@@ -17,6 +16,7 @@ pytestmark = pytest.mark.usefixtures(
 )
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestEnableContentItemSelectionMode:
     def test_it(self, js_config):
         js_config.enable_content_item_selection_mode(
@@ -114,6 +114,7 @@ class TestEnableContentItemSelectionMode:
         return patch("lms.resources._js_config.FilePickerConfig")
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestEnableLTILaunchMode:
     def test_it(self, bearer_token_schema, context, grant_token_service, js_config):
         js_config.enable_lti_launch_mode()
@@ -361,6 +362,7 @@ class TestMaybeEnableGrading:
         return pyramid_request
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestMaybeSetFocusedUser:
     def test_it_does_nothing_if_theres_no_focused_user_param(
         self, js_config, pyramid_request
@@ -435,6 +437,7 @@ class TestJSConfigAuthToken:
         return config["api"]["authToken"]
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestJSConfigAPISync:
     """Unit tests for the api.sync sub-dict of JSConfig."""
 
@@ -512,6 +515,7 @@ class TestJSConfigDebug:
         return config["debug"]
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestJSConfigHypothesisClient:
     """Unit tests for the "hypothesisClient" sub-dict of JSConfig."""
 
@@ -584,6 +588,7 @@ class TestJSConfigHypothesisClient:
         context.canvas_groups_enabled = True
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestJSConfigRPCServer:
     """Unit tests for the "rpcServer" sub-dict of JSConfig."""
 
```

seanh (changes requested):

tests/unit/conftest.py
marcospri: Yes, good point better to fix those test rather than accommodate their brokenness into the fixture, updated :+1:

seanh (approved):

PR #3419, opened 2021-11-22
Model for storing grouping membership

marcospri:
Needs migration before merging

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -82,6 +82,8 @@ class Grouping(CreatedUpdatedMixin, BASE):
         nullable=False,
     )
 
+    memberships = sa.orm.relationship("GroupingMembership", back_populates="grouping")
+
     @property
     def name(self):
         """Return an h-compatible group name."""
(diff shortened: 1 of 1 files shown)


lms/models/grouping.py
```diff
@@ -104,9 +106,23 @@ class CanvasGroup(Grouping):
...
+class BlackboardGroup(Grouping):
     __mapper_args__ = {"polymorphic_identity": Grouping.Type.BLACKBOARD_GROUP}
 
 
 class Course(Grouping):
     __mapper_args__ = {"polymorphic_identity": Grouping.Type.COURSE}
+
+
+class GroupingMembership(CreatedUpdatedMixin, BASE):
+    __tablename__ = "grouping_membership"
```
marcospri, line 118: I think you mentioned something about the table name @seanh

No strong opinions from my side on this "grouping_membership" works, "grouping_users" or something similar works as well.

lms/models/grouping.py
```diff
@@ -104,9 +106,23 @@ class CanvasGroup(Grouping):
...
-class BlackBoardGroup(Grouping):
+class BlackboardGroup(Grouping):
     __mapper_args__ = {"polymorphic_identity": Grouping.Type.BLACKBOARD_GROUP}
 
 
 class Course(Grouping):
     __mapper_args__ = {"polymorphic_identity": Grouping.Type.COURSE}
+
+
+class GroupingMembership(CreatedUpdatedMixin, BASE):
```
marcospri, line 117: No `id` primary_key here, user_id & grouping_id together are the primary key, which I think is nicer here that adding `id` as primary key + adding a unique constraint for those together.

seanh (approved):

PR #3425, opened 2021-11-22
New service methods to handle grouping memberships

marcospri:


lms/services/grouping.py
```diff
@@ -97,6 +109,59 @@ def upsert_canvas_group(  # pylint: disable=too-many-arguments
...
+                self._db.query(GroupingMembership)
+                .filter_by(grouping_id=group.id, user_id=user.id)
+                .one_or_none()
+            ):
+                membership.updated = func.now()
+                continue
+
+            group.memberships.append(GroupingMembership(grouping=group, user=user))
+
+    def get_course_groupings_for_user(
```
marcospri, line 130: I had another version of this method that was more general (with an optional parent_id param so it worked to find memberships to top-level courses) but that mean that

- It needed and extra `application_instance` param 
- Having a `group_set_id` made even less sense than in this version

lms/services/grouping.py
```diff
@@ -97,6 +109,59 @@ def upsert_canvas_group(  # pylint: disable=too-many-arguments
...
+    ):
+        """
+        Get the groupings a user belongs to in a given course.
+
+        :param course:  The course the users belongs to
+        :param user_id: User.user_id of the user we are filtering by
+        :param type_: Type of subdivision within the course
+        :param group_set_id: Optionally filter by `group_set_id` stored in the Course.extra
+        """
+        course_grouping = aliased(Course)
```
marcospri, line 145: We are querying `Grouping` but also joining with `Grouping` to filter by parent

lms/services/grouping.py
```diff
@@ -97,6 +109,59 @@ def upsert_canvas_group(  # pylint: disable=too-many-arguments
...
+            ):
+                membership.updated = func.now()
+                continue
+
+            group.memberships.append(GroupingMembership(grouping=group, user=user))
+
+    def get_course_groupings_for_user(
+        self,
+        course: Course,
+        user_id: str,
```
marcospri, line 133: user_id: str (and not User) as this will be used when grading when we only have the ID. Still possible to do user.user_id when calling if you do have a `User` object.

seanh (approved):

PR #3430, opened 2021-11-23
Blackboard sync API configuration

marcospri:
Config for sync API talking into account Blackboard.

Added an empty sync API endpoint for blackboard.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -2,7 +2,7 @@ import functools
 from enum import Enum
 from typing import List, Optional
 
-from lms.models import GroupInfo, HUser
+from lms.models import ApplicationInstance, GroupInfo, HUser
 from lms.resources._js_config.file_picker_config import FilePickerConfig
 from lms.services import HAPIError
 from lms.validation.authentication import BearerTokenSchema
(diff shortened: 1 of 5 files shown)


lms/views/api/blackboard/sync.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, request):
+        self._request = request
+
+    @view_config(
+        route_name="blackboard_api.sync",
+        request_method="POST",
+        renderer="json",
+        permission=Permissions.API,
+    )  # pylint: disable=no-self-use
+    def sync(self):
```
marcospri, line 16: I'm adding here a useless view to allow for `req.route_path("blackboard_api.sync")` 

I think adding all the boilerplate for it here it's not a bad idea.

lms/routes.py
```diff
@@ -55,6 +55,9 @@ def includeme(config):
         "blackboard_api.courses.group_sets.list",
         "/api/blackboard/courses/{course_id}/group_sets",
     )
+    config.add_route(
+        "blackboard_api.sync", "/api/blackboard/sync", request_method="POST"
```
seanh, line 59: We probably don't need the `request_method` here since it's on the view as well

lms/resources/_js_config/__init__.py
```diff
@@ -510,3 +503,43 @@ def _sync_api(self):
...
+        if self._context.is_canvas and (
+            self._context.canvas_sections_enabled
+            or self._context.canvas_is_group_launch
+        ):
+            return self._canvas_sync_api()
+
+        if (
+            self._application_instance().product
+            == ApplicationInstance.Product.BLACKBOARD
+            and self._context.is_blackboard_group_launch
```
seanh, line 540: I noticed a naming inconsistency between `is_blackboard_group_launch` and `canvas_is_group_launch`. It's not from this PR though

seanh (approved):

PR #3431, opened 2021-11-23
Add a `CHECK` constraint to the grouping table

seanh:
Course groupings cannot have parents, whilst all non-course groupings must have parents. This is true of the production data currently. This commit adds a DB-level constraint to ensure it.


seanh:
This would need a DB migration

marcospri (approved):
Looks good.

There's to PRs open with migrations, the second one will need to be rebased and adjust the `down_revision` to avoid having two branches in alembic

PR #3432, opened 2021-11-23
Constrain the `grouping.type` column to valid values

seanh:
This means that each time we add a new type of grouping we'll need to do a simple DB migration to widen this constraint. For that cost we get the benefit of ensuring at the DB-level that the `type` column can only contain valid values.

The code actually contains a [`Grouping` model](https://github.com/hypothesis/lms/blob/67d2266f919a63adaf14e4274bd838ced9aba053/lms/models/grouping.py#L14-L96) which is the base class for the [`CanvasSection`, `CanvasGroup`, `BlackboardGroup` and `Course` models](https://github.com/hypothesis/lms/blob/67d2266f919a63adaf14e4274bd838ced9aba053/lms/models/grouping.py#L99-L112). But it's never valid to actually instantiate the `Grouping` model: all groupings must be a Canvas section, Canvas group, Blackboard group or course, never just a "grouping".

By adding a DB-level constraint that `type` must be `'course'`, `'canvas_section'`, `'canvas_group'` or `'blackboard_group'` this PR prevents code from instantiating a `Grouping` and adding it to the DB.

Maybe the `Grouping` model should also be an abstract class? And maybe it should be private, renamed to `_Grouping` and not exposed as a top-level `lms.models.*` name?

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -27,6 +27,10 @@ class Grouping(CreatedUpdatedMixin, BASE):
             "(type='course' AND parent_id IS NULL) OR (type!='course' AND parent_id IS NOT NULL)",
             name="courses_must_NOT_have_parents_and_other_groupings_MUST_have_parents",
         ),
+        sa.CheckConstraint(
+            "type in ('course', 'canvas_section', 'canvas_group', 'blackboard_group')",
+            name="grouping_type_must_be_a_valid_value",
+        ),
     )
 
     id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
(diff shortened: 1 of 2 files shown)


marcospri:
>Maybe the Grouping model should also be an abstract class? 

I think the service will still want to instantiate Grouping but passing a explicit `type`.

> And maybe it should be private, renamed to _Grouping and not exposed as a top-level lms.models.* name?

This is a good idea.

lms/models/grouping.py
```diff
@@ -23,6 +23,10 @@ class Type(str, Enum):
     __table_args__ = (
         sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
         sa.UniqueConstraint("lms_id", "application_instance_id", "parent_id", "type"),
+        sa.CheckConstraint(
```
marcospri, line 26: There's also https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#sqlalchemy.dialects.postgresql.ENUM

I would pick which ever plays nicer with alembic (as far as I remember alembic didn't detect changes in the type so the explicit constraint might be simpler)

PR #3433, opened 2021-11-24
Simplify a test

seanh:
Just remove a little code duplication from this test, hopefully makes the test a little more readable.

This also rearranges the test to have the standard three sections Arrange, Act and Assert (separated by two blank lines).

### DIFF
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -29,42 +29,28 @@ class TestGroupingService:
 
         assert db_session.query(CanvasGroup).one() == test_grouping
 
-    def test_upsert_with_parents_updates(self, svc, db_session):
+    def test_upsert_with_parent_updates_existing_groupings(self, svc, db_session):
         course = factories.Course()
-        tool_consumer_instance_guid = (
-            course.application_instance.tool_consumer_instance_guid
-        )
-        lms_id = "lms_id"
+        kwargs = {
+            "tool_consumer_instance_guid": course.application_instance.tool_consumer_instance_guid,
+            "lms_id": "lms_id",
+            "parent": course,
+            "type_": Grouping.Type.CANVAS_GROUP,
+        }
         old_name = "old_name"
         old_extra = {"extra": "old"}
         new_name = "new_name"
         new_extra = {"extra": "new"}
+        # Insert an existing grouping into the DB.
+        svc.upsert_with_parent(lms_name=old_name, extra=old_extra, **kwargs)
 
-        # Insert a new grouping.
-        svc.upsert_with_parent(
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/grouping_test.py
```diff
@@ -29,42 +29,28 @@ def test_upsert_with_parents_inserts(self, svc, db_session):
 
         assert db_session.query(CanvasGroup).one() == test_grouping
 
-    def test_upsert_with_parents_updates(self, svc, db_session):
+    def test_upsert_with_parent_updates_existing_groupings(self, svc, db_session):
```
seanh, line 32: Made the test name a little longer and more explicit, this should be clearer I think (and still fits on one line)

tests/unit/lms/services/grouping_test.py
```diff
@@ -29,42 +29,28 @@ def test_upsert_with_parents_inserts(self, svc, db_session):
...
-        tool_consumer_instance_guid = (
-            course.application_instance.tool_consumer_instance_guid
-        )
-        lms_id = "lms_id"
+        kwargs = {
+            "tool_consumer_instance_guid": course.application_instance.tool_consumer_instance_guid,
+            "lms_id": "lms_id",
+            "parent": course,
+            "type_": Grouping.Type.CANVAS_GROUP,
+        }
```
seanh, line 39: These same kwargs are used in both calls to `upsert_with_parent()`. Gathering them into a variable removes some noise and duplication and in the call to `upsert_with_parent()` the values that actually matter (`new_name` and `new_extra`) are now easier to see:

<pre><code>svc.upsert_with_parent(<b>lms_name=old_name, extra=old_extra</b>, **kwargs)</code></pre>

tests/unit/lms/services/grouping_test.py
```diff
@@ -29,42 +29,28 @@ def test_upsert_with_parents_inserts(self, svc, db_session):
...
         old_name = "old_name"
         old_extra = {"extra": "old"}
         new_name = "new_name"
         new_extra = {"extra": "new"}
+        # Insert an existing grouping into the DB.
+        svc.upsert_with_parent(lms_name=old_name, extra=old_extra, **kwargs)
```
seanh, line 45: I think the reason why this test has to call `upsert_with_parent()` as part of its arrange block (rather than just inserting the grouping into the DB itself) is that groupings are difficult to construct because you have to generate the `authority_provided_id`. You can't just do something like `db_session.add(Grouping(lms_name=old_name, extra=old_extra, **kwargs))`, it ends up being much more awkward than that. It might be possible to fix this with some refactoring.

marcospri (approved):

PR #3436, opened 2021-11-24
Remove `Grouping.Type.GROUPING`

seanh:
We only have four types of group so far: courses, Canvas sections, Canvas groups, and Blackboard groups. It's not actually valid to have a row in the `grouping` table with `"grouping"` in the `type` column. Every row should have either `"course"`, `"canvas_section"`, `"canvas_group"` or `"blackboard_group"` in the `type` column. SQLAlchemy should be constructing `Course`, `CanvasSection`, `CanvasGroup` or `BlackboardGroup` models for every row that a query returns, it should never be constructing `Grouping`'s directly.


tests/unit/lms/models/grouping_test.py
```diff
@@ -1,14 +1,14 @@
 import pytest
 
-from lms.models import Grouping
```
marcospri, line 3: I didn't remember we had Course and LegacyCourse factories, :+1:

PR #3437, opened 2021-11-24
BB sync API returns all groups in the group set for all users

marcospri:
This takes the current sync API that returns just [] and change it to return all the groups in the group set.

- No error handling for non existing group set or no groups in it.
- No error handling for the student not having access to that particular endpoint

This doesn't add much value on itself but I think it easier to review here the private methods and the consecutive PRs that add the conditionals (if student/teacher/grading) and error handling will be more focused.

# Testing

- Switch to `bb-poc-testing`
- Pick this branch there `git cherry-pick bb-sync-teacher`
- Create a new assignment on BB, select group assignment and pick a group set.
- Launching the assignment as a teacher should diplay all the groups in the group set.

### DIFF
--- a/lms/views/api/blackboard/sync.py
+++ b/lms/views/api/blackboard/sync.py
@@ -1,17 +1,72 @@
 from pyramid.view import view_config
 
+from lms.models import Grouping
 from lms.security import Permissions
 
 
 class Sync:
     def __init__(self, request):
         self._request = request
+        self._assignment_service = self._request.find_service(name="assignment")
+        self._grouping_service = self._request.find_service(name="grouping")
+        self._course_service = self._request.find_service(name="course")
+        self._blackboard_api = self._request.find_service(name="blackboard_api_client")
+
+        self._tool_consumer_instance_guid = self._request.json["lms"][
+            "tool_consumer_instance_guid"
+        ]
 
     @view_config(
         route_name="blackboard_api.sync",
         request_method="POST",
         renderer="json",
         permission=Permissions.API,
-    )  # pylint: disable=no-self-use
+    )
     def sync(self):
-        return []
+        groups = self._get_blackboard_groups()
+
+        self._sync_to_h(groups)
+
+        authority = self._request.registry.settings["h_authority"]
+        return [group.groupid(authority) for group in groups]
+
+    def _get_blackboard_groups(self):
+        group_set_id = self._group_set()
+        course_id = self._request.json["course"]["context_id"]
+        groups = self._blackboard_api.group_set_groups(course_id, group_set_id)
+
+        return self._to_groups_groupings(groups)
+
+    def _group_set(self):
+        return self._assignment_service.get(
+            self._tool_consumer_instance_guid,
+            self._request.json["assignment"]["resource_link_id"],
+        ).extra["group_set_id"]
+
+    def _to_groups_groupings(self, groups):
+        course = self._get_course()
+
+        return [
+            self._grouping_service.upsert_with_parent(
+                tool_consumer_instance_guid=self._tool_consumer_instance_guid,
+                lms_id=group["id"],
+                lms_name=group["name"],
+                parent=course,
+                type_=Grouping.Type.BLACKBOARD_GROUP,
+                extra={"group_set_id": group["groupSetId"]},
+            )
+            for group in groups
+        ]
+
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/api/blackboard/sync_test.py
```diff
@@ -1,7 +1,70 @@
+import pytest
```
marcospri, line 1: This test is mostly copied from the canvas version using the same `assert_sync_and_return_groups` fixture adapted for black board groups.

tests/unit/lms/views/api/blackboard/sync_test.py
```diff
@@ -1,7 +1,70 @@
...
+    "assignment_service",
+    "lti_h_service",
+    "grouping_service",
+    "course_service",
+    "blackboard_api_client",
+)
 
 
-def test_it(pyramid_request):
+def test_it(pyramid_request, assert_sync_and_return_groups, blackboard_api_client):
```
marcospri, line 16: Only one test as there's no really conditionals or error handling in the view for this version.

seanh:
> * No error handling for non existing group set or no groups in it.
> * No error handling for the student not having access to that particular endpoint

Can you add issues or notes for these to the [Blackboard Groups project board](https://github.com/orgs/hypothesis/projects/53) so that we don't forget about them?

lms/views/api/blackboard/sync.py
```diff
@@ -1,17 +1,72 @@
...
     def __init__(self, request):
         self._request = request
+        self._assignment_service = self._request.find_service(name="assignment")
+        self._grouping_service = self._request.find_service(name="grouping")
+        self._course_service = self._request.find_service(name="course")
+        self._blackboard_api = self._request.find_service(name="blackboard_api_client")
+
+        self._tool_consumer_instance_guid = self._request.json["lms"][
+            "tool_consumer_instance_guid"
+        ]
```
seanh, line 17: I think this'll crash if you call the API with no `lms.tool_consumer_instance_guid` param (and I noticed that there's no guard schema on the view either). It's not super important because this API is only ever called by our own frontend code (and only _can_ be called by our own frontend code, due to the necessary permissions). But can we add an issue or note to the Blackboard Groups project board to add validation to this endpoint?

If someone does make an invalid and/or unauthenticated request the code would not get as far as executing the code in the view method since the permission or view guard schema (if it had one) would fail. But I'm not sure whether the code in the view class's `__init__()` method would actually still get executed and crash? (I'm guessing not, but I'm not sure off the top of my head.)

(If there _is_ a crash due to this code being in the `__init__()` then you can move it into a lazy `@property` instead, so you can still have the deduplication without the potentially crashing `__init__()`.)

lms/views/api/blackboard/sync.py
```diff
@@ -1,17 +1,72 @@
...
+        group_set_id = self._group_set()
+        course_id = self._request.json["course"]["context_id"]
+        groups = self._blackboard_api.group_set_groups(course_id, group_set_id)
+
+        return self._to_groups_groupings(groups)
+
+    def _group_set(self):
+        return self._assignment_service.get(
+            self._tool_consumer_instance_guid,
+            self._request.json["course"]["resource_link_id"],
```
seanh, line 43: I think we might want to do all the extracting params from the request at the top of the view method itself. I think it's just nicer to have all the param-reading in one place so you can see which params are actually needed, and this can be near where the schema is defined (once there is a schema):

```python
class Sync:
    ...

    @view_config(..., schema=BlackboardSyncAPISchema)
    def sync(self):
        lms = request.parsed_params["lms"]
        tool_consumer_instance_guid = lms["tool_consumer_instance_guid"]

        course = request.parsed_params["course"]
        context_id = course["context_id"]

        assignment = request.parsed_params["assignment"]
        resource_link_id = assignment["resource_link_id"]

        ...
```

seanh: By the way, separate PR but I just noticed that the `resource_link_id` is not really part of the course. It's the _assignment_ ID. I think it should be `request.params["assignment"]["resource_link_id"]`. Here's a PR to fix it: <https://github.com/hypothesis/lms/pull/3438> (_this_ PR will need updating after merging that).

seanh (approved):

PR #3439, opened 2021-11-24
Remove the Grouping.Type enum

seanh:
Now that we have a DB-level constraint to ensure that the `grouping` table can't contain invalid types (https://github.com/hypothesis/lms/pull/3432/) is it easier to just type `"course"` rather than having to do `from lms.models import Grouping` and then use `Grouping.Type.COURSE`?

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -1,5 +1,3 @@
-from enum import Enum
-
 import sqlalchemy as sa
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.mutable import MutableDict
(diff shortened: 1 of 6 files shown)


marcospri:
Not sure if I agree with the premise here 

> is it easier to just type "course" rather than having to do from lms.models import Grouping

It might be easier for "course" but is it `canvas_group`? `group_canvas`? You still need to go and take a look at the model file probably. 

I think the interface of the grouping methods is clearer with the enum and the contraints it's just there as a safety mechanism

seanh:
Fair enough, let's leave the enum in for now. I think it may be over-design personally. If we add a new type of grouping we'll need to add that string to the `CHECK` constraint in the model code, add a DB migration to add the string to the constraint on QA and prod, _and_ add the string to the enum. And the enum has almost gotten us in trouble once when we didn't realise that we had to pass `type_.value` rather than `type_` to `hashed_id()`.

We can remove the duplication from the `CHECK` constraint by having it use the enum:

```python
sa.CheckConstraint(
    f"type in {tuple(type_.value for type_ in Type)}",
    name="grouping_type_must_be_a_valid_value",
),
```

but I think this is probably getting way too clever.

PR #3442, opened 2021-11-25
Store group set on the assignment while configuring it

marcospri:
Store the group set of the assignment on extra for DB configured ones.

Note the change from "group_set" -> "group_set_id". We are storing currently the same value for canvas on the groupings themselves and it's probably  nicer if the names match between different extras even if they are in different tables.

```
select extra from grouping where type = 'canvas_group' and extra is not null  order by id desc limit 5;
{"group_set_id": 2631}
{"group_set_id": 2631}
{"group_set_id": 2631}
{"group_set_id": 2631}
{"group_set_id": 2631}
```

As a follow up this values in devdata will need to change: https://github.com/hypothesis/devdata/blob/1e24cf8f088c51cfd13439de4f4420fd763a3f7e/lms/devdata.json#L290

### DIFF
--- a/tests/unit/lms/views/basic_lti_launch_test.py
+++ b/tests/unit/lms/views/basic_lti_launch_test.py
@@ -129,7 +129,7 @@ def unconfigured_basic_lti_launch_caller(context, pyramid_request):
     return views.unconfigured_basic_lti_launch()
 
 
-def configure_assignment_caller(context, pyramid_request):
+def configure_assignment_caller(context, pyramid_request, parsed_params=None):
     """
     Call BasicLTILaunchViews.configure_assignment().
 
@@ -145,6 +145,8 @@ def configure_assignment_caller(context, pyramid_request):
         "resource_link_id": "TEST_RESOURCE_LINK_ID",
         "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
     }
+    if parsed_params:
+        pyramid_request.parsed_params.update(parsed_params)
 
     views = BasicLTILaunchViews(context, pyramid_request)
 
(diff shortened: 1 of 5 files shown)


tests/unit/lms/views/basic_lti_launch_test.py
```diff
@@ -430,6 +433,20 @@ def test_it_adds_the_document_url(self, context, pyramid_request):
...
+        configure_assignment_caller(
+            context, pyramid_request, parsed_params={"group_set": "ID"}
+        )
+
+        assignment_service.upsert.assert_called_once_with(
+            pyramid_request.parsed_params["document_url"],
+            pyramid_request.parsed_params["tool_consumer_instance_guid"],
+            pyramid_request.parsed_params["resource_link_id"],
+            extra={"group_set_id": "ID"},
+        )
```
seanh, line 448: I think it might be better to integrate this test into the [`test_it_saves_the_assignments_document_url_to_the_db()`](https://github.com/hypothesis/lms/blob/13cc730be29007f8de041499242e114ef1899b12/tests/unit/lms/views/basic_lti_launch_test.py#L412-L422) test above. And to be honest the two other tests in this class could also just be two assertions added to a single test:

```python
class TestConfigureAssignment:
    @pytest.mark.parametrize(
        "parsed_params,expected_extras",
        [
            (None, {}),
            ({"group_set": 42}, {"group_set_id": 42}),
        ],
    )
    def test_it(
        self,
        assignment_service,
        context,
        pyramid_request,
        parsed_params,
        expected_extras,
    ):
        configure_assignment_caller(context, pyramid_request, parsed_params)

        assignment_service.upsert.assert_called_once_with(
            pyramid_request.parsed_params["document_url"],
            pyramid_request.parsed_params["tool_consumer_instance_guid"],
            pyramid_request.parsed_params["resource_link_id"],
            extra=expected_extras,
        )
        context.js_config.add_document_url.assert_called_once_with(
            pyramid_request.parsed_params["document_url"]
        )
        context.js_config.maybe_enable_grading.assert_called_once_with()
```

seanh (changes requested):
:+1: Looks good, just a couple of small suggestions

PR #3443, opened 2021-11-25
Add `AsyncOAuthHTTPService`

marcospri:


lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,60 @@
+import asyncio
+from typing import List
+
+import aiohttp
+
+from lms.services import ExternalRequestError
+
+
+class AsyncOAuthHTTPService:
```
marcospri, line 9: Just one service, in theory we could have both a `AsyncHTTPService` and build `AsyncOAuthHTTPService` on top. In practice we only have one usage for this so that might be too much for now.

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self, oauth2_token_service):
+        self._oauth2_token_service = oauth2_token_service
+
+    def get(
+        self, urls, concurrency=10, headers=None, timeout=10
+    ) -> List[aiohttp.ClientResponse]:
+        """
+        Sends async requests to all urls.
+
```
marcospri, line 18: TODO some docs on the params

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,60 @@
...
+        :raise ExternalRequestError: if something goes wrong with the HTTP
+            request
+        """
+
+        headers = headers or {}
+
+        access_token = self._oauth2_token_service.get().access_token
+        headers["Authorization"] = f"Bearer {access_token}"
+
+        async def _async_request(semaphore, aio_session, url, headers=None):
```
marcospri, line 29: I kept this as private an even inside the main `get` method to hide all the `async` bits away. They probably could live as private top level methods

tests/unit/lms/services/async_oauth_http_test.py
```diff
@@ -0,0 +1,81 @@
...
+import pytest
+from aiohttp import TooManyRedirects
+from aioresponses import CallbackResult, aioresponses
+
+from lms.services import ExternalRequestError
+from lms.services.async_oauth_http import AsyncOAuthHTTPService, factory
+
+
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+class TestAsyncOAuthHTTPService:
```
marcospri, line 14: These two test are not asserting much or anything about the async details of the service.

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,60 @@
...
+                    )
+                    tasks.append(task)
+                try:
+                    return await asyncio.gather(*tasks, return_exceptions=False)
+                except ExternalRequestError:
+                    for task in tasks:
+                        task.cancel()
+                    raise
+
+        return asyncio.run(_prepare_requests(urls, headers))
```
marcospri, line 56: I was worrying about the ordering here but it's guaranteed by asyncio.gather:

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

```The order of result values corresponds to the order of awaitables in aws.```

File: lms/services/async_oauth_http.py (hunk 3 above)
Comment (seanh) [reply lms/services/async_oauth_http.py:29]: They look like they could even be private helper functions outside the class, they don't use `self`. Or private `@staticmethod`'s at the bottom of the class

File: tests/unit/lms/services/async_oauth_http_test.py [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+import httpretty
+import pytest
+from aiohttp import TooManyRedirects
+from aioresponses import CallbackResult, aioresponses
+
+from lms.services import ExternalRequestError
+from lms.services.async_oauth_http import AsyncOAuthHTTPService, factory
+
+
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
seanh, line 13: ?

tests/unit/lms/services/async_oauth_http_test.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_get_success(
+        self, svc, urls, with_successful_responses, oauth2_token_service
+    ):
+        responses = svc.get(urls)
+
+        for i, url in enumerate(urls):
+            assert responses[i].status == 200
+
+            # Order is kept from the URLs param
+            assert responses[i].headers["url"] == url
```
seanh, line 24: You can use `zip()`:

```suggestion
        for url, response in zip(urls, responses):
            assert response.status == 200
            assert response.headers["url"] == url
```

tests/unit/lms/services/async_oauth_http_test.py
```diff
@@ -0,0 +1,81 @@
...
+    def urls(self):
+        """Return the URLs that we'll be sending test requests to."""
+        return ["https://example.com/example", "https://example.com/another"]
+
+    @pytest.fixture
+    def svc(self, oauth2_token_service):
+        return AsyncOAuthHTTPService(oauth2_token_service)
+
+    @pytest.fixture(autouse=True)
+    def unhttpretty(self):
```
seanh, line 71:
```suggestion
    def httpunpretty(self):
```

:)

Although I think you can just override the `httpretty_` fixture to disable it for tests in this class:

```python
class TestAsyncOAuthHTTPService:
    ...

    @pytest.fixture(autouse=True)
    def httpretty_(self):
        pass
```

Is this necessary though? The tests seem to pass without it

marcospri: :thinking: This was causing an error when I tested it locally but now it works with it, removed.

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,60 @@
...
+                    task = asyncio.create_task(
+                        _async_request(semaphore, session, url, headers)
+                    )
+                    tasks.append(task)
+                try:
+                    return await asyncio.gather(*tasks, return_exceptions=False)
+                except ExternalRequestError:
+                    for task in tasks:
+                        task.cancel()
+                    raise
```
seanh, line 54: I think you might be able to simplify the code slightly:

1. Move the `timeout` argument from `get()` to `ClientSession()`, just because this ends up formatting nicer (less line wrapping)
2. Don't catch `aiohttp.ClientError` in `_async_request()`, just catch it in `_prepare_requests()`

I think this ends up working the same?

```python
        async def _async_request(semaphore, aio_session, url, headers=None):
            async with semaphore:
                async with aio_session.get(url, headers=headers) as response:
                    return response

        async def _prepare_requests(urls, headers=None):
            semaphore = asyncio.Semaphore(concurrency)

            async with aiohttp.ClientSession(timeout=timeout) as session:
                tasks = []

                for url in urls:
                    tasks.append(
                        asyncio.create_task(
                            _async_request(semaphore, session, url, headers)
                        )
                    )

                try:
                    return await asyncio.gather(*tasks, return_exceptions=False)
                except aiohttp.ClientError as err:
                    for task in tasks:
                        task.cancel()

                    raise ExternalRequestError(
                        "Blackboard async request failed"
                    ) from err
```

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,69 @@
...
+    try:
+        async with semaphore:
+            async with aio_session.request(method, url, **kwargs) as response:
+                return response
+    except aiohttp.ClientError as err:
+        raise ExternalRequestError("Blackboard async request failed") from err
+
+
+async def _prepare_requests(method, urls, concurrency=10, **kwargs):
+    semaphore = asyncio.Semaphore(concurrency)
```
seanh, line 42: Definitely test this to make sure but I don't think the semaphore is needed. `aiohttp.ClientSession()` has a `connector` argument which I think is a [`TCPConnector`](https://docs.aiohttp.org/en/stable/client_reference.html#tcpconnector) object by default. `TCPConnector.__init__()` has a `limit` argument which is the "total number simultaneous connections" and is `100` by default.

I don't think we need to customise this to `10`, the default of `100` is probably fine (?)

lms/services/async_oauth_http.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+
+async def _async_request(aio_session, method, url, **kwargs):
+    async with aio_session.request(method, url, **kwargs) as response:
+        return response
+
+
+async def _prepare_requests(method, urls, **kwargs):
+    async with aiohttp.ClientSession(connector=aiohttp.TCPConnector()) as session:
```
seanh, line 51:
```suggestion
    async with aiohttp.ClientSession() as session:
```

I think `TCPConnector()` is the default? https://github.com/aio-libs/aiohttp/blob/b446af77ce83182b1c17aa5ba5758a2cd224d4be/aiohttp/client.py#L231-L232

seanh (approved):

PR #3444, opened 2021-11-25
Ensure that child groupings must belong to the same application instance as their parents

seanh:
### Problem

On master it's possible to create a grouping (e.g. a Blackboard group) whose parent grouping (a Blackboard course) belongs to a different application instance, which makes no sense.

For example this first creates `PARENT_GROUPING` with `application_instance_id` `1` and then creates `CHILD_GROUPING` with `application_instance_id` `2` and with `PARENT_GROUPING` (`parent_id` `1`) as its parent:

```sql
postgres=# INSERT INTO grouping (application_instance_id, authority_provided_id, lms_id, lms_name, type) VALUES (1, 'authority_provided_id', 'lms_id_1', 'PARENT_GROUPING', 'course');
INSERT 0 1
postgres=# INSERT INTO grouping (application_instance_id, authority_provided_id, parent_id, lms_id, lms_name, type) VALUES (2, 'authority_provided_id', 1, 'lms_id_2', 'CHILD_GROUPING', 'blackboard_group');
INSERT 0 1
postgres=# select * from grouping;
-[ RECORD 1 ]-----------+---------------------------
created                 | 2021-12-03 18:00:03.412114
updated                 | 2021-12-03 18:00:03.412114
id                      | 1
application_instance_id | 1
authority_provided_id   | authority_provided_id
parent_id               | 
lms_id                  | lms_id_1
lms_name                | PARENT_GROUPING
settings                | {}
extra                   | {}
-[ RECORD 2 ]-----------+---------------------------
created                 | 2021-12-03 18:00:11.305157
updated                 | 2021-12-03 18:00:11.305157
id                      | 2
application_instance_id | 2
authority_provided_id   | authority_provided_id
parent_id               | 1
lms_id                  | lms_id_2
lms_name                | CHILD_GROUPING
type                    | blackboard_group
settings                | {}
extra                   | {}
```

Or the same thing in Python, on master:

```python
>>> db.add(models.Grouping(application_instance_id=1, authority_provided_id="authority_provided_id", lms_id="lms_id_1", lms_name="PARENT_GROUPING", type="course"))
>>> db.add(models.Grouping(application_instance_id=2, authority_provided_id="authority_provided_id", lms_id="lms_id_2", lms_name="CHILD_GROUPING", type="blackboard_group", parent_id=1))
>>> tm.commit()
>>> tm.begin()
<transaction._transaction.Transaction at 0x7f2254e8bdc0>
>>> db.query(models.Grouping).all()
[Course(created=datetime.datetime(2021, 12, 3, 18, 12, 24, 386936), updated=datetime.datetime(2021, 12, 3, 18, 12, 24, 386936), id=1, application_instance_id=1, authority_provided_id='authority_provided_id', parent_id=None, lms_id='lms_id_1', lms_name='PARENT_GROUPING', type='course', settings=ApplicationSettings({}), extra={}),
 BlackboardGroup(created=datetime.datetime(2021, 12, 3, 18, 12, 24, 386936), updated=datetime.datetime(2021, 12, 3, 18, 12, 24, 386936), id=2, application_instance_id=2, authority_provided_id='authority_provided_id', parent_id=1, lms_id='lms_id_2', lms_name='CHILD_GROUPING', type='blackboard_group', settings=ApplicationSettings({}), extra={})]
```

### Solution

This branch changes the the `grouping.parent_id -> grouping.id` foreign key constraint to a compound foreign key constraint `(grouping.parent_id, grouping.application_instance_id) -> (grouping.id, grouping.application_instance_id)`. Matching up the `application_instance_id`'s isn't necessary to find/identify the parent grouping, but it ensures that the parent grouping's `application_instance_id` must be the same as the child's.

On **this branch** the second SQL query will fail with a foreign key constraint violation:

```sql
postgres=# INSERT INTO grouping (application_instance_id, authority_provided_id, lms_id, lms_name, type) VALUES (1, 'authority_provided_id', 'lms_id_1', 'PARENT_GROUPING', 'course');
INSERT 0 1
postgres=# INSERT INTO grouping (application_instance_id, authority_provided_id, parent_id, lms_id, lms_name, type) VALUES (2, 'authority_provided_id', 1, 'lms_id_2', 'CHILD_GROUPING', 'blackboard_group');
ERROR:  insert or update on table "grouping" violates foreign key constraint "fk__grouping__parent_id__grouping"
DETAIL:  Key (parent_id, application_instance_id)=(1, 2) is not present in table "grouping".
```

On **this branch**:

```python
>>> db.add(models.Grouping(application_instance_id=1, authority_provided_id="authority_provided_id", lms_id="lms_id_1", lms_name="PARENT_GROUPING", type="course"))
>>> db.add(models.Grouping(application_instance_id=2, authority_provided_id="authority_provided_id", lms_id="lms_id_2", lms_name="CHILD_GROUPING", type="blackboard_group", parent_id=1))
>>> tm.commit()
---------------------------------------------------------------------------
ForeignKeyViolation                       Traceback (most recent call last)
...
ForeignKeyViolation: insert or update on table "grouping" violates foreign key constraint "fk__grouping__parent_id__grouping"
DETAIL:  Key (parent_id, application_instance_id)=(1, 2) is not present in table "grouping".


The above exception was the direct cause of the following exception:

IntegrityError                            Traceback (most recent call last)
...
IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "grouping" violates foreign key constraint "fk__grouping__parent_id__grouping"
DETAIL:  Key (parent_id, application_instance_id)=(1, 2) is not present in table "grouping".

[SQL: INSERT INTO grouping (application_instance_id, authority_provided_id, parent_id, lms_id, lms_name, type) VALUES (%(application_instance_id)s, %(authority_provided_id)s, %(parent_id)s, %(lms_id)s, %(lms_name)s, %(type)s) RETURNING grouping.id]
[parameters: ({'application_instance_id': 1, 'authority_provided_id': 'authority_provided_id', 'parent_id': None, 'lms_id': 'lms_id_1', 'lms_name': 'PARENT_GROUPING', 'type': 'course'}, {'application_instance_id': 2, 'authority_provided_id': 'authority_provided_id', 'parent_id': 1, 'lms_id': 'lms_id_2', 'lms_name': 'CHILD_GROUPING', 'type': 'blackboard_group'})]
...
```

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -23,6 +23,12 @@ class Grouping(CreatedUpdatedMixin, BASE):
     __table_args__ = (
         sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
         sa.UniqueConstraint("lms_id", "application_instance_id", "parent_id", "type"),
+        sa.UniqueConstraint("id", "application_instance_id"),
+        sa.ForeignKeyConstraint(
+            ["parent_id", "application_instance_id"],
+            ["grouping.id", "grouping.application_instance_id"],
+            ondelete="cascade",
+        ),
         sa.CheckConstraint(
             "(type='course' AND parent_id IS NULL) OR (type!='course' AND parent_id IS NOT NULL)",
             name="courses_must_NOT_have_parents_and_other_groupings_MUST_have_parents",
@@ -49,13 +55,15 @@ class Grouping(CreatedUpdatedMixin, BASE):
     #:
     #: For example if the grouping represents a Canvas section or group then parent_id
     #: will reference the grouping for the course that the section or group belongs to.
-    parent_id = sa.Column(
-        sa.Integer(),
-        sa.ForeignKey("grouping.id", ondelete="cascade"),
-        nullable=True,
-    )
+    parent_id = sa.Column(sa.Integer(), nullable=True)
     children = sa.orm.relationship(
-        "Grouping", backref=sa.orm.backref("parent", remote_side=[id])
+        "Grouping",
+        backref=sa.orm.backref(
+            "parent",
+            remote_side=[id, application_instance_id],
+            overlaps="application_instance",
+        ),
+        overlaps="application_instance",
     )
 
     #: The LMS's ID for the grouping.
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -10,7 +10,7 @@ from lms.models._hashed_id import hashed_id
 class GroupingService:
     def __init__(self, db, application_instance_service):
         self._db = db
-        self._application_instance = application_instance_service.get_current()
+        self.application_instance = application_instance_service.get_current()
 
     @staticmethod
     def generate_authority_provided_id(
@@ -58,7 +58,7 @@ class GroupingService:
         grouping = (
             self._db.query(Grouping)
             .filter_by(
-                application_instance=self._application_instance,
+                application_instance=self.application_instance,
                 authority_provided_id=authority_provided_id,
             )
             .one_or_none()
@@ -66,7 +66,7 @@ class GroupingService:
 
         if not grouping:
             grouping = Grouping(
-                application_instance=self._application_instance,
+                application_instance=self.application_instance,
                 authority_provided_id=authority_provided_id,
                 lms_id=lms_id,
                 parent_id=parent.id,
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -1,6 +1,7 @@
 from unittest.mock import sentinel
 
 import pytest
+from sqlalchemy.exc import IntegrityError
 
 from lms.models import CanvasGroup, Grouping, GroupingMembership
 from lms.services.grouping import GroupingService, factory
@@ -14,7 +15,7 @@ class TestGroupingService:
     TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
 
     def test_upsert_with_parents_inserts(self, svc, db_session):
-        course = factories.Course()
+        course = factories.Course(application_instance=svc.application_instance)
 
         # Start with no CanvasGroup
         assert not db_session.query(CanvasGroup).count()
@@ -29,8 +30,35 @@ class TestGroupingService:
 
         assert db_session.query(CanvasGroup).one() == test_grouping
 
+    def test_you_cant_upsert_a_grouping_whose_parent_has_a_different_application_instance(
+        self, svc, application_instance_service, db_session
+    ):
+        parent = factories.Course()
+        # GroupingService uses ApplicationInstanceService.get_current() as the
+        # application_instance for the new grouping it inserts.
+        # Here the parent course will have a different application_instance,
+        # which will trigger an IntegrityError when we flush the DB.
+        assert (
+            parent.application_instance
+            != application_instance_service.get_current.return_value
+        )
+
+        svc.upsert_with_parent(
+            tool_consumer_instance_guid="tool_consumer_instance_guid",
+            lms_id="lms_id",
+            lms_name="lms_name",
+            type_=Grouping.Type.CANVAS_GROUP,
+            parent=parent,
+        )
+
+        with pytest.raises(
+            IntegrityError,
+            match='insert or update on table "grouping" violates foreign key constraint "fk__grouping__parent_id__grouping"',
+        ):
+            db_session.flush()
+
     def test_upsert_with_parent_updates_existing_groupings(self, svc, db_session):
-        course = factories.Course()
+        course = factories.Course(application_instance=svc.application_instance)
         kwargs = {
             "tool_consumer_instance_guid": course.application_instance.tool_consumer_instance_guid,
             "lms_id": "lms_id",
@@ -56,7 +84,9 @@ class TestGroupingService:
         assert db_grouping.extra == new_extra
 
     def test_canvas_group_and_sections_dont_conflict(self, svc, db_session):
-        course = factories.Course(lms_id=self.CONTEXT_ID)
+        course = factories.Course(
+            lms_id=self.CONTEXT_ID, application_instance=svc.application_instance
+        )
         db_session.flush()
 
         group = svc.upsert_with_parent(
--- a/lms/migrations/versions/f581d8a273cc_constrain_child_groupings_to_parents_.py
+++ b/lms/migrations/versions/f581d8a273cc_constrain_child_groupings_to_parents_.py
@@ -1,0 +1,45 @@
+"""
+Constrain child groupings to belong to their parent's application instance.
+
+Revision ID: f581d8a273cc
+Revises: 6882c201b56e
+Create Date: 2022-01-25 17:57:55.813338
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+revision = "f581d8a273cc"
+down_revision = "6882c201b56e"
+
+
+def upgrade():
+    op.drop_constraint(
+        "fk__grouping__parent_id__grouping", "grouping", type_="foreignkey"
+    )
+    op.create_unique_constraint(
+        op.f("uq__grouping__id"), "grouping", ["id", "application_instance_id"]
+    )
+    op.create_foreign_key(
+        op.f("fk__grouping__parent_id__grouping"),
+        "grouping",
+        "grouping",
+        ["parent_id", "application_instance_id"],
+        ["id", "application_instance_id"],
+        ondelete="cascade",
+    )
+
+
+def downgrade():
+    op.drop_constraint(
+        op.f("fk__grouping__parent_id__grouping"), "grouping", type_="foreignkey"
+    )
+    op.drop_constraint(op.f("uq__grouping__id"), "grouping", type_="unique")
+    op.create_foreign_key(
+        "fk__grouping__parent_id__grouping",
+        "grouping",
+        "grouping",
+        ["parent_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )


lms/models/grouping.py
```diff
@@ -23,6 +23,12 @@ class Type(str, Enum):
     __table_args__ = (
         sa.UniqueConstraint("application_instance_id", "authority_provided_id"),
         sa.UniqueConstraint("lms_id", "application_instance_id", "parent_id", "type"),
+        sa.UniqueConstraint("id", "application_instance_id"),
```
seanh, line 26: It was necessary to add this `(id, application_instance_id)` unique constraint to get around this error:

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.InvalidForeignKey) there is no unique constraint matching given keys for referenced table "grouping"
```

It seems a bit odd because `id` is already unique (it's the primary key) so `(id, application_instance_id)` must be unique anyway, but apparently a separate unique constraint is needed.

lms/models/grouping.py
```diff
@@ -23,6 +23,12 @@ class Type(str, Enum):
...
+        sa.ForeignKeyConstraint(
+            ["parent_id", "application_instance_id"],
+            ["grouping.id", "grouping.application_instance_id"],
+            ondelete="cascade",
+        ),
```
seanh, line 31: Moved this `ForeignKey` from the `parent_id` column below into `__table_args__`. When you want a compound foreign key you have to put it in `__table_args__`.

lms/models/grouping.py
```diff
@@ -49,13 +55,15 @@ class Type(str, Enum):
...
+    parent_id = sa.Column(sa.Integer(), nullable=True)
     children = sa.orm.relationship(
-        "Grouping", backref=sa.orm.backref("parent", remote_side=[id])
+        "Grouping",
+        backref=sa.orm.backref(
+            "parent",
+            remote_side=[id, application_instance_id],
+            overlaps="application_instance",
+        ),
+        overlaps="application_instance",
```
seanh, line 66: The two `overlaps` are necessary to avoid this error from SQLAlchemy:

```
sqlalchemy.exc.InvalidRequestError: One or more mappers failed to initialize - can't proceed with initialization of other mappers. Triggering mapper: 'mapped class Grouping->grouping'. Original exception was: relationship 'Grouping.parent' will copy column grouping.application_instance_id to column grouping.application_instance_id, which conflicts with relationship(s): 'Grouping.application_instance' (copies application_instances.id to grouping.application_instance_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="application_instance"' to the 'Grouping.parent' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)
```

The link <https://sqlalche.me/e/14/qzyx> contains some useful explanation.

tests/unit/lms/services/grouping_test.py
```diff
@@ -13,8 +14,8 @@ class TestGroupingService:
     CONTEXT_ID = "context_id"
     TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
 
-    def test_upsert_with_parents_inserts(self, svc, db_session):
-        course = factories.Course()
+    def test_upsert_with_parents_inserts(self, svc, application_instance, db_session):
+        course = factories.Course(application_instance=application_instance)
```
seanh, line 18: The test was creating a child grouping whose parent grouping has a different application instance: `factories.Course()` generates a new application instance for `course` (which is going to be used as the parent grouping) and then when the test calls `GroupingService.upsert_with_parent()`, `GroupingService` uses `ApplicationInstanceService.get_current()` as the application instance for the new grouping that it inserts.

The mock `application_instance_service`'s `get_current()` returns the value of the `application_instance` fixture. I might want to do something to remove this quite implicit dependency of this test on these connections between the three fixtures involved.

lms/services/grouping.py
```diff
@@ -10,7 +10,7 @@
 class GroupingService:
     def __init__(self, db, application_instance_service):
         self._db = db
-        self._application_instance = application_instance_service.get_current()
+        self.application_instance = application_instance_service.get_current()
```
seanh, line 13: I made `GroupingService.application_instance` a public attribute so that the unit tests can access it. A hack, perhaps? But I think it's acceptable: there's no real reason for this attribute to be public (only the tests access it). But there's no really strong reason why it needs to be private either. With `GroupingService.application_instance` public if anyone should want to know which `application_instance` `GroupingService` is going to use for any new groupings it creates, now they can. `GroupingService.application_instance` is arguably part of the "public interface" anyway even if nothing actually access it: it's the `application_instance` that it's going to use for new groupings that it creates for you. Any `parent` groupings that you pass to `GroupingService.upsert_with_parent()` must have `GroupingService.application_instance` as their `application_instance` or you'll get an `IntegrityError`

tests/unit/lms/services/grouping_test.py
```diff
@@ -14,7 +15,7 @@ class TestGroupingService:
     TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
 
     def test_upsert_with_parents_inserts(self, svc, db_session):
-        course = factories.Course()
+        course = factories.Course(application_instance=svc.application_instance)
```
seanh, line 18: This test and a couple of others below were creating groupings with different `application_instance`'s than their parents and so now crashing, so I fixed them.

The way it works is:

1. If you don't pass any `application_instance` to `factories.Course()` it'll generate a new one, and so `course` (which is going to be passed as the `parent` to `GroupingService.upsert_with_parent()` below) would end up with a different `application_instance` than the one that `GroupingService` uses for the new child grouping that it creates

2. `GroupingService.upsert_with_parent()` uses the current `ApplicationInstance` (from `ApplicationInstanceService.get_current()` as the `Grouping.application_instance` for any new child groupings that it creates.

3. In the tests the real `GroupingService` uses the mock `ApplicationInstanceService` from the `application_instance_service` fixture. So it uses `application_instance_service.get_current.return_value` as the `application_instance` for any new child groupings it creates

4. The mock `application_instance_service.get_current.return_value` also happens to be the same object as the `application_instance` fixture

Rather than having the tests implicitly rely on these connections between the `GroupingService` under test and the mock `application_instance_service` and `application_instance` fixtures, I made `GroupingService.application_instance` public so the tests can access `svc.application_instance`.

tests/unit/lms/services/grouping_test.py
```diff
@@ -29,8 +30,35 @@ def test_upsert_with_parents_inserts(self, svc, db_session):
...
+            tool_consumer_instance_guid="tool_consumer_instance_guid",
+            lms_id="lms_id",
+            lms_name="lms_name",
+            type_=Grouping.Type.CANVAS_GROUP,
+            parent=parent,
+        )
+
+        with pytest.raises(
+            IntegrityError,
+            match='insert or update on table "grouping" violates foreign key constraint "fk__grouping__parent_id__grouping"',
```
seanh, line 56: The `match` here is necessary to ensure that we're catching the `IntegrityError` that we think we're catching. Otherwise something else could be going wrong (there's all sorts of ways to trigger an `IntegrityError`), the code or test could be broken but still passing.

seanh:
Marking this as **do not merge** because I want to manually test that the read-writeable `application_instance`, `parent` and `children` relationships are still working. But I believe the code is final and can be reviewed

seanh:
Reminder to self: this also needs a DB migration

lms/services/grouping.py
jon-betts: I think "there's no real reason for this attribute to be public" might be the reason for it to be private we're looking for. Our tests and the unit under test are always tightly bound, so I wouldn't worry about accessing a private variable in the unit tests for an item. If it's expedient I'd just do it.

On the other hand, I don't actually have a problem with in this instance. I don't think any other code is going to start accessing the application instance via this service, and even if it did it's not hard to change.

tests/unit/lms/services/grouping_test.py
```diff
@@ -56,7 +84,9 @@ def test_upsert_with_parent_updates_existing_groupings(self, svc, db_session):
         assert db_grouping.extra == new_extra
 
     def test_canvas_group_and_sections_dont_conflict(self, svc, db_session):
-        course = factories.Course(lms_id=self.CONTEXT_ID)
+        course = factories.Course(
+            lms_id=self.CONTEXT_ID, application_instance=svc.application_instance
+        )
```
jon-betts, line 89: Might this be better as a course fixture which you then set `lms_id` on? It would replace three uses if so, which I think is a good threshold to make a fixture.

lms/services/grouping.py
jon-betts: Duh, just noticed this is an instance not the service, but I think it's still fine.

To be clear, I don't think this would be fine on a model object, or something else we'd expect to be passed about a lot and consulted, as all attributes could be part of the interface, but our services are typically instantiated when you need them and then have a few methods called on them.

tests/unit/lms/services/grouping_test.py
jon-betts: I think this good. Explicit is better than implicit. I mentioned below that this is something we now do three times, so I think a fixture would be in order.

It might be good to add a comment like `# Ensure the course is attached to the same application instance the service uses` or something.

tests/unit/lms/services/grouping_test.py
```diff
@@ -29,8 +30,35 @@ def test_upsert_with_parents_inserts(self, svc, db_session):
...
+        assert (
+            parent.application_instance
+            != application_instance_service.get_current.return_value
+        )
+
+        svc.upsert_with_parent(
+            tool_consumer_instance_guid="tool_consumer_instance_guid",
+            lms_id="lms_id",
+            lms_name="lms_name",
+            type_=Grouping.Type.CANVAS_GROUP,
```
jon-betts, line 50: Do any of these four values matter? It might be nice to try and explain they don't (if they don't) so the important ones stand out. Maybe with values like "any_string" or `Any` objects or something. Whatever we can get away with.

lms/models/grouping.py
jon-betts: I'm going to freely admit I don't know what this is about, and so can't really comment. I assume it just blew up when you tried it and this was the solution?

jon-betts (approved):
Yep this looks good to me. It seems like Postgres / SQLAlchemy has forced our hands in a few places which has complicated the solution. I think those lines would benefit from some short comments explaining why they are there (Postgres made me because...) so we know to remove them if we change our mind. Otherwise they are quite mysterious.

I think there's a wider question of which checks should be done in the DB and which should be done in the code. I think this is simple enough that it can be done at the DB level, which is nice in the sense it will kick in even if you insert rows manually. I think much more complicated and you'd have to do it in code. It does seem more complicated, or at least indirect, than a quick check and exception. I'd be interested in any background ideas / thinking about when DB constraints are the right way to go.

I'd suggest a fixture for the course as we repeat the same construction three times, but it's not essential.

lms/models/grouping.py
seanh: Bah, it's just a basic part of how SQLAlchemy works: all multi-column constraints need to be done in the `__table_args__` (things like `ForeignKey(...)` or `unique=True` etc inside a `Column()` are convenience shortcuts that only work for single-column constraints). So I don't think it's worth a comment to explain it as it's just normal SQLAlchemy business

lms/models/grouping.py
seanh: Yeah, it's something to do with how SQLAlchemy figures out the relationships between tables (or between different rows of the same table, in this instance) in order to hook-up the convenience relationship attributes on the ORM objects like `.parent` and `.children`. For some internal SQLAlchemy reason we need to give it an extra clue in this case

tests/unit/lms/services/grouping_test.py
seanh: They don't matter but I can't think of a great solution:
 
* You can't use something like `Any` because it actually has to be a string because it's actually going to go into the DB
* I don't think hard-coded strings like `"tool_consumer_instance_guid"` or even `"any_string"` really get across the message that these values don't matter to this test. And they're visually noisy: they obliterate the `parent=parent` argument that actually does matter
* I thought of maybe a `random_string()` function but it's still visually noisy and you don't want actual randomness in tests
* There's also the `type_=Grouping.Type.CANVAS_GROUP` which is not a string. This can be any `Grouping.Type.*` except `Grouping.Type.CANVAS_GROUP` because course's don't have parents

factory_boy does this so well when creating model objects for us: allowing each test to specify only the values that it cares about, and having appropriate values be generated for all the other attributes.

We _could_ actually write a factory for generating the args to the `upsert_with_parent()` method:

```diff
diff --git a/tests/unit/lms/services/grouping_test.py b/tests/unit/lms/services/grouping_test.py
index 72e10a35..4c2c35fe 100644
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -1,5 +1,6 @@
 from unittest.mock import sentinel
 
+import factory as factory_boy
 import pytest
 from sqlalchemy.exc import IntegrityError
 
@@ -10,6 +11,16 @@ from tests import factories
 pytestmark = pytest.mark.usefixtures("course_service", "application_instance_service")
 
 
+upsert_with_parent_args = factory_boy.make_factory(
+    dict,
+    # In the real version you'd randomise these hard-coded strings.
+    tool_consumer_instance_guid="tool_consumer_instance_guid",
+    lms_id="lms_id",
+    lms_name="lms_name",
+    parent=None,
+    type_=Grouping.Type.CANVAS_GROUP,
+)
+
+
 class TestGroupingService:
     CONTEXT_ID = "context_id"
     TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
@@ -43,13 +54,7 @@ class TestGroupingService:
             != application_instance_service.get_current.return_value
         )
 
-        svc.upsert_with_parent(
-            tool_consumer_instance_guid="tool_consumer_instance_guid",
-            lms_id="lms_id",
-            lms_name="lms_name",
-            type_=Grouping.Type.CANVAS_GROUP,
-            parent=parent,
-        )
+        svc.upsert_with_parent(**upsert_with_parent_args(parent=parent))
 
         with pytest.raises(
             IntegrityError,
```

tests/unit/lms/services/grouping_test.py
seanh: I think these `Course` objects are only being created by the tests because `svc.upsert_with_parent()` requires the `parent` argument to be a `Course` whose `application_instance` is the same as `svc.application_instance`. If we add a factory or caller method or something (as mentioned in <https://github.com/hypothesis/lms/pull/3444/files#r792004256>) to prevent each test from having to pass every single argument to `upsert_with_parent()` then I think these will go away

tests/unit/lms/services/grouping_test.py
seanh: I'm gonna merge the PR as-is and explore this in a separate PR

lms/services/grouping.py
seanh: > Our tests and the unit under test are always tightly bound, so I wouldn't worry about accessing a private variable in the unit tests for an item.

I guess I prefer that the unit tests work only through the public interface of the unit under test so that I can freely refactor the unit under test and as long as I haven't changed its public contract the unit tests should still pass.

That said I think you're right about the unit tests being very close to the unit under test, such that it's acceptable for them to access a private attribute if pragmatism demands it.

In this case I felt that `GroupingService.application_instance` _was_ arguably actually part of `GroupingService`'s public interface: it's the `application_instance` that `GroupingService` will use for any groupings that it creates for you when you call methods like `GroupingService.upsert()` etc. It doesn't actually _have_ to be public in the sense that no production code outside of `GroupingService` actually reads `GroupingService.application_instance` currently. But I think you can make some sort of argument that semantically it is part of the public interface.

You can also make a pretty strong argument that it doesn't matter a whole lot.

Anyway I left it public

tests/unit/lms/services/grouping_test.py
seanh: Mentioned elsewhere but these tests later got refactored (introducing a caller pattern) in a way that removed this duplication so that a fixture was no longer needed

seanh:
> It seems like Postgres / SQLAlchemy has forced our hands in a few places which has complicated the solution. I think those lines would benefit from some short comments explaining why they are there (Postgres made me because...) so we know to remove them if we change our mind. Otherwise they are quite mysterious.

These commends got added in another PR: https://github.com/hypothesis/lms/pull/3602

PR #3446, opened 2021-11-26
BlackboardListGroups schema exposes enrollment.type

marcospri:
It will be necessary to distinguish between groups can see but don't
necessary belong to (SelfEnrollment) and groups the student can only see if the
are member (InstructorOnly)

PR #3447, opened 2021-11-26
Blackboard course groups

marcospri:
~~This PR includes the commits from https://github.com/hypothesis/lms/pull/3446~~

Method to get groups of a course it needs more complexity that initially thought:
- Using AsyncOauthService to make multiple request the the API 
- Make those request only for self-enrollment groups.

### DIFF
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -114,3 +114,55 @@ class BlackboardAPIClient:
             f"/learn/api/public/v2/courses/uuid:{course_id}/groups/sets/{group_set_id}/groups",
         )
         return BlackboardListGroups(response).parse()
+
+    def course_groups(
+        self, course_id, group_set_id=None, current_student_own_groups_only=True
+    ):
+        """
+        Return the groups in a course.
+
+        :param course_id: ID of the course
+        :param group_set_id: Only return groups that belong to this group set
+        :param current_student_own_groups_only: Only return groups the current user (a student) belong to.
+        """
+        response = self._api.request(
+            "GET",
+            f"/learn/api/public/v2/courses/uuid:{course_id}/groups",
+        )
+        groups = BlackboardListGroups(response).parse()
+
+        if group_set_id:
+            groups = [group for group in groups if group["groupSetId"] == group_set_id]
+
+        if not current_student_own_groups_only:
+            return groups
+
+        instructor_only_groups = []
+        self_enrollment_groups = []
+        self_enrollment_check_urls = []
+
+        for group in groups:
+            if group["enrollment"]["type"] == "InstructorOnly":
+                # We don't need to check "InstructorOnly" groups, we can only see those if we are member
+                instructor_only_groups.append(group)
+            else:
+                # For the rest will have to make a HTTP request per group
+                self_enrollment_groups.append(group)
+                self_enrollment_check_urls.append(
+                    self._api._api_url(  # pylint: disable=protected-access
+                        f"/learn/api/public/v2/courses/uuid:{course_id}/groups/{group['id']}"
+                    )
+                )
+
+        if self_enrollment_groups:
+            responses = self._request.find_service(name="async_oauth_http").request(
+                "GET", self_enrollment_check_urls
+            )
+            # If we are a member of any of the SelfEnrollment groups
+            # we'll get a 200 response from the endpoint
+            self_enrollment_groups = [
+                group
+                for group, response in zip(self_enrollment_groups, responses)
+                if response.status == 200
+            ]
+        return self_enrollment_groups + instructor_only_groups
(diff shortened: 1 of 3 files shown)


lms/services/blackboard_api/client.py
```diff
@@ -113,4 +113,56 @@ def group_set_groups(self, course_id, group_set_id):
...
+        if group_set_id:
+            groups = [group for group in groups if group["groupSetId"] == group_set_id]
+
+        if not current_student_own_groups_only:
+            return groups
+
+        instructor_only_groups = []
+        self_enrollment_groups = []
+        self_enrollment_check_urls = []
+
```
marcospri, line 143: Trying to fit this into a line didn't look any better

lms/services/blackboard_api/client.py
```diff
@@ -113,4 +113,56 @@ def group_set_groups(self, course_id, group_set_id):
...
+            "GET",
+            f"/learn/api/public/v2/courses/uuid:{course_id}/groups",
+        )
+        groups = BlackboardListGroups(response).parse()
+
+        if group_set_id:
+            groups = [group for group in groups if group["groupSetId"] == group_set_id]
+
+        if not current_student_own_groups_only:
+            return groups
```
marcospri, line 138: Returning early here on the `not` instead of nesting the rest of the code

lms/services/blackboard_api/client.py
```diff
@@ -113,4 +113,56 @@ def group_set_groups(self, course_id, group_set_id):
...
+
+        instructor_only_groups = []
+        self_enrollment_groups = []
+        self_enrollment_check_urls = []
+
+        for group in groups:
```
marcospri, line 144: We need to split the groups between InstructorOnly/SelfEnrollment and generate the URLs so one pass in a for loop ended up being nicer than multiple list comprehensions or something else.

tests/unit/services.py
```diff
@@ -21,37 +22,11 @@
...
-    "h_api",
-    "http_service",
-    "launch_verifier",
-    "lti_h_service",
-    "lti_outcomes_client",
-    "oauth1_service",
-    "oauth2_token_service",
-    "oauth_http_service",
-    "user_service",
-    "vitalsource_service",
```
marcospri, line 52: I needed to add `async_oauth_http_service` here but I don't see why we should keep this list, happy to add it back if this would break something.

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -237,19 +237,110 @@ def test_it(
...
+
+    def test_it_with_group_set_id(self, svc, blackboard_list_groups, groups):
+        blackboard_list_groups.parse.return_value = groups
+
+        groups = svc.course_groups(
+            "COURSE_ID", "OTHER_GROUP_SET", current_student_own_groups_only=False
+        )
+
+        assert len(groups) == 1
+        assert groups[0]["groupSetId"] == "OTHER_GROUP_SET"
```
seanh, line 278: Could do this in a single assert:

```suggestion
        assert [group["groupSetId"] for group in groups] == ["OTHER_GROUP_SET"]
```

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -237,19 +237,110 @@ def test_it(
...
+    ):
+        blackboard_list_groups.parse.return_value = groups
+
+        groups = svc.course_groups(
+            "COURSE_ID", "GROUP_SET", current_student_own_groups_only=True
         )
-        assert group_sets == blackboard_list_group_set_groups.parse.return_value
+
+        assert len(groups) == 1
+        assert groups[0]["enrollment"]["type"] == "InstructorOnly"
```
seanh, line 290: Could be:

```suggestion
        assert [group["enrollment"]["type"] for group in groups] == ["InstructorOnly"]
```

tests/unit/services.py
seanh: This is actually need because we do `from tests.unit.services import *`:

https://github.com/hypothesis/lms/blob/f677d836efb068f7365a961fbadc0836c0c10bac/tests/unit/conftest.py#L13

It's necessary to import all the fixtures from `services.py` into `conftest.py` otherwise pytest won't discover those fixtures.

Without the `__all__` `conftest.py` is now importing a bunch more stuff (e.g. everything that `services.py` imports) which is wrong even if it doesn't actually cause any tests to fail.

I think this is over-organization and we should just inline the contents of `tests/unit/services.py` into `tests/unit/conftest.py`. I think it's fine to just keep all pytest fixtures in the standard location in the `conftest.py` files where people (and pytest_ expect to find them. Some of the `conftest.py` files might get quite big but that's probably fine.

seanh (approved):

tests/unit/services.py
marcospri: Oh I forgot about fixture discovery, fair enough, added it back :+1:

PR #3451, opened 2021-11-29
Expose the user.user_id while grading

marcospri:
Will be later needed to get the value back on the sync API.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -413,6 +413,7 @@ class JSConfig:
             )
             yield {
                 "userid": h_user.userid(self._authority),
+                "lmsId": grading_info.user_id,
                 "displayName": h_user.display_name,
                 "LISResultSourcedId": grading_info.lis_result_sourcedid,
                 "LISOutcomeServiceUrl": grading_info.lis_outcome_service_url,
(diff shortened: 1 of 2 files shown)


tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -306,6 +306,7 @@ def test_it_adds_the_grading_settings(
                     "LISResultSourcedId": f"test_lis_result_sourcedid_{i}",
                     "displayName": f"test_h_display_name_{i}",
                     "userid": f"acct:test_h_username_{i}@TEST_AUTHORITY",
+                    "lmsId": f"test_user_id_{i}",
```
marcospri, line 309: I think here `userid` is just a word/concept vs `lmsId` which is more like `LISResultSourcedId`

lms/resources/_js_config/__init__.py
```diff
@@ -413,6 +413,7 @@ def _get_students(self):
             )
             yield {
                 "userid": h_user.userid(self._authority),
+                "lmsId": grading_info.user_id,
```
marcospri, line 416: I think here `userid` is just a word/concept vs `lmsId` which is more like `LISResultSourcedId`

Also I think having `userId` and `userid` would be extra confusing so even if the attribute in GradingInfo is called "user_id" I'm exposing it as lmsId which is a name we use in other parts of the modeling.

seanh (approved):

PR #3459, opened 2021-12-01
Fix missing setting on AI admin view

marcospri:


lms/views/admin.py
```diff
@@ -85,6 +85,7 @@ def update_instance(self):
             ("canvas", "sections_enabled"),
             ("canvas", "groups_enabled"),
             ("blackboard", "files_enabled"),
+            ("blackboard", "groups_enabled"),
```
marcospri, line 88: Without this the DB changes weren't made. I didn't notice because I hard coded the value while testing the UI.

seanh (approved):

PR #3462, opened 2021-12-01
Simplify JSConfig._sync_api()

seanh:
This doesn't need to check whether `application_instance.product` is `BLACKBOARD`: it wouldn't have been possible to create a Blackboard groups assignment for this application instance in the first place if the LMS wasn't Blackboard.

Also, if the assignment's settings say that it's a Blackboard Groups assignment then we should obey those and launch it as a Blackboard Groups assignment.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -540,12 +540,7 @@ class JSConfig:
         ):
             return self._canvas_sync_api()
 
-        if (
-            self._application_instance().product
-            == ApplicationInstance.Product.BLACKBOARD
-            and self._context.is_blackboard_group_launch
-        ):
-
+        if self._context.is_blackboard_group_launch:
             return self._blackboard_sync_api()
 
         return None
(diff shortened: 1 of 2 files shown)


marcospri:
I think this change + what we are also discussing, making existing groups assignments work after feature has been disabled which we could achieve with something like:


``` diff
diff --git a/lms/resources/lti_launch.py b/lms/resources/lti_launch.py
index f16d4149..abe4658c 100644
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -208,9 +208,6 @@ class LTILaunchResource:
     @property
     def is_blackboard_group_launch(self):
         """Return True if the current assignment uses Blackboard groups."""
-        if not self.blackboard_groups_enabled:
-            return False
-
         tool_consumer_instance_guid = self._request.params[
             "tool_consumer_instance_guid"
         ]
```

would mean that `is_blackboard_group_launch` is no longer about blackboard groups, it's just `is_db_configured_group_launch` together with this change could lead to a hard to spot bug while interacting with different LMS (eg the n-th bug while making Canvas store their assignment on the DB, would make this also True). I think we should keep the check here

PR #3463, opened 2021-12-01
Remove `ApplicationInstance.product`

seanh:
Depends on https://github.com/hypothesis/lms/pull/3461 and https://github.com/hypothesis/lms/pull/3462. It we merge both of those then `ApplicationInstance.product` will no longer be used.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -1,7 +1,6 @@
 import logging
 import secrets
 from datetime import datetime
-from enum import Enum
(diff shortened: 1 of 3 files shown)


marcospri:
This is not used in  master (if it were removed by those two PRs) but It's needed for this which is still not merged: https://github.com/hypothesis/lms/pull/3458/files#diff-0c43f1136df847df408667fea0b43e71320531d5ddb5a7c601d1728f3179ce0eR7

PR #3464, opened 2021-12-01
Keep Blackboard groups assignments if feature is disabled

seanh:
If the Blackboard Groups feature is enabled for an application instance and users create some groups assignments and then the feature gets disabled for the application instance those pre-existing groups assignments should remain groups assignments, theys shouldn't become course assignments.

Slack thread:

https://hypothes-is.slack.com/archives/C4K6M7P5E/p1638379099478700

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -205,9 +205,6 @@ class LTILaunchResource:
     @property
     def is_blackboard_group_launch(self):
         """Return True if the current assignment uses Blackboard groups."""
-        if not self.blackboard_groups_enabled:
-            return False
-
         tool_consumer_instance_guid = self._request.params[
             "tool_consumer_instance_guid"
         ]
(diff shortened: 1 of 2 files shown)


marcospri:
Makes sense but I think based on what I commented here https://github.com/hypothesis/lms/pull/3462#issuecomment-983934474 that we should:

- merge `canvas_is_group_launch` and `is_blackboard_group_launch` in one method, `is_group_launch` 
- keep the `Product.BLACKBOARD` check in `_sync_api()` (ie, don't merge https://github.com/hypothesis/lms/pull/3462) 
- Change `_sync_api()` to just use the new `is_group_launch`

seanh:
@marcospri Can you look again? I think the idea of merging `canvas_is_group_launch()` and `is_blackboard_group_launch()` may no longer apply as the two are quite different: `canvas_is_group_launch()` looks for `"group_set"` in the request params whereas `is_blackboard_group_launch()` looks at the `Assignment` extras in the DB

marcospri (approved):

PR #3467, opened 2021-12-02
Fixes test_it_when_grading flaky grading explicitly setting a lti user

marcospri:


tests/unit/lms/views/api/blackboard/sync_test.py
```diff
@@ -52,7 +52,7 @@ def test_it_when_student(
     )
 
 
-@pytest.mark.usefixtures("user_service")
+@pytest.mark.usefixtures("user_service", "user_is_instructor")
```
marcospri, line 55: Not 100% sure of the reason but this sometimes fails and sometimes passes.

In any case while grading, you are an instructor so maybe this should have been here anyway.

seanh (approved):

PR #3468, opened 2021-12-02
Remove underscores from Blackboard sync API view

seanh:
The purpose of leading underscores is to distinguish a package/class/module's public interface from its private parts.

View classes don't have a public interface in the sense of attributes that're used by other parts of our code: we don't write code to call our views, they just get called by Pyramid.

The leading underscores could be useful to visually identify which things are going to be called by Pyramid (e.g. the actual view functions/methods) and which things are not called directly by Pyramid (e.g. helpers). But the actual view functions/methods are already identifiable by their `@view_config` decorators.

So I don't think there's any purpose to using leading underscores in view modules/classes (?), therefore there's no reason to ugly-up the code with them.

Note that we absolutely do use leading underscores in cases where they're necessary to distinguish things that're meant to be accessed by other parts of our code or by Pyramid versus things that are private. I'm just saying that underscores aren't needed for that purpose in view modules/classes.

### DIFF
--- a/lms/views/api/blackboard/sync.py
+++ b/lms/views/api/blackboard/sync.py
@@ -7,11 +7,11 @@ from lms.services import UserService
 
 class Sync:
     def __init__(self, request):
-        self._request = request
-        self._grouping_service = self._request.find_service(name="grouping")
-        self._blackboard_api = self._request.find_service(name="blackboard_api_client")
+        self.request = request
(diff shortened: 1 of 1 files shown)


marcospri (approved):
:+1: Makes total sense
I think we should do the same on `lms/views/api/canvas/sync.py` then.

PR #3470, opened 2021-12-03
Handle Blackboard  groups edge cases

marcospri:
Handles `blackboard_student_not_in_group` and `blackboard_group_set_empty`

Sync API returns

```
{
    "error_code": "blackboard_student_not_in_group",
    "details": {
        "group_set": "_55_1"
    }
}
```

## Testing

- Use `localhost (make devdata) Groups Assignment With No Groups` and `localhost (make devdata) Groups Assignment with No Members` to  trigger the error conditions.

### DIFF
--- a/lms/views/__init__.py
+++ b/lms/views/__init__.py
@@ -1,5 +1,4 @@
 from lms.views.api.canvas.exceptions import (
-    CanvasGroupError,
     CanvasGroupSetEmpty,
     CanvasGroupSetNotFound,
     CanvasStudentNotInGroup,
--- a/lms/views/api/blackboard/sync.py
+++ b/lms/views/api/blackboard/sync.py
@@ -4,6 +4,19 @@ from lms.models import Grouping
 from lms.security import Permissions
 from lms.services import UserService
 from lms.validation import APIBlackboardSyncSchema
+from lms.views.api.exceptions import GroupError
+
+
+class BlackboardStudentNotInGroup(GroupError):
+    """Student doesn't belong to any of the groups in a group set assignment."""
+
+    error_code = "blackboard_student_not_in_group"
+
+
+class BlackboardGroupSetEmpty(GroupError):
+    """Canvas GroupSet doesn't contain any groups."""
+
+    error_code = "blackboard_group_set_empty"
 
 
 class Sync:
@@ -46,6 +59,9 @@ class Sync:
             learner_groups = self.blackboard_api.course_groups(
                 course_id, group_set_id, current_student_own_groups_only=True
             )
+            if not learner_groups:
+                raise BlackboardStudentNotInGroup(group_set=group_set_id)
+
             groups = self.to_groups_groupings(course, learner_groups)
             self.grouping_service.upsert_grouping_memberships(user, groups)
             return groups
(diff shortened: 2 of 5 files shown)


lms/views/__init__.py
```diff
@@ -5,6 +5,8 @@
     CanvasStudentNotInGroup,
 )
 
+from lms.views.api.blackboard.exceptions import BlackboardStudentNotInGroup
+
```
seanh, line 9: I think this import is unnecessary since `BlackboardStudentNotInGroup` isn't used outside of the `lms.views.api.blackboard` package. The reason why we do this in packages like `models`, `validation`, `services`, etc is because it makes it easier for other packages to import things. Instead of:

```python
from lms.models.user import User
from lms.models.grouping import CanvasGroup
from lms.models.application_instance import ApplicationInstance
```

You can just do:

```python
from lms.models import User, CanvasGroup, ApplicationInstance
```

It also means that the internal structure of a package like `lms.models` is effectively a private implementation detail of that package which is nice because other packages don't need to know about it (hence imports are easier to write) and because we can reorganise the code inside `lms.models` without breaking any other packages. There's good reason to split a package into separate modules (we don't want one really big file) but also good reason not to tightly couple other packages to that sub-organization.

In this case `BlackboardStudentNotInGroup` isn't being exported for other packages to import so there's no need.

(The imports of the Canvas exceptions above should probably be removed for the same reason.)

lms/views/api/blackboard/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class BlackboardGroupError(Exception):
+    def __init__(self, group_set):
+        self.details = {"group_set": group_set}
+        super().__init__(self.details)
+
+
+class BlackboardStudentNotInGroup(BlackboardGroupError):
+    """Student doesn't belong to any of the groups in a group set assignment."""
+
+    error_code = "blackboard_student_not_in_group"
```
seanh, line 10: :+1: Looks good. There's an informal protocol being implemented here (the `error_code` and `details` attributes) that isn't documented anywhere or defined by an abstract class or anything, but this is used by the `api_error()` exception view that will catch this exception: https://github.com/hypothesis/lms/blob/72b7dd9e0eb60570936311c0cc6b1f8f224d2be4/lms/views/api/exceptions.py#L159-L200

(And actually I think the protocol is actually documented by the exception view's docstring.)

It's worth checking to make sure that `api_error()` actually *is* the exception view that handles this exception, I think it will be.

The `BlackboardGroupError` base class looks unnecessary to me, but I'm guessing you're expecting to add another subclass to it in future.

lms/views/api/blackboard/sync.py
```diff
@@ -5,6 +5,9 @@
 from lms.services import UserService
 
 
+from lms.views import BlackboardStudentNotInGroup
```
seanh, line 8:
```suggestion
from lms.views.api.blackboard.exceptions import BlackboardStudentNotInGroup
```

There's no need to import the exception into `views/__init__.py`, as mentioned in another comment above.

If this exception is only every imported into `sync.py` then I think you could also just move the exception class directly into `sync.py`. But putting it in a separate `exceptions.py` file is also fine.

lms/views/api/canvas/exceptions.py
```diff
@@ -1,22 +1,19 @@
-class CanvasGroupError(Exception):
-    def __init__(self, group_set):
-        self.details = {"group_set": group_set}
-        super().__init__(self.details)
```
seanh, line 4: Looks like this has just been moved to `GroupError` below :+1:

seanh (approved):
Looks good

seanh (approved):
Looks good

PR #3493, opened 2021-12-06
Make `factories.User.h_userid` unique

seanh:
`factories.User.h_userid` is using a single random hex digit to try to make each user ID different from the previous:

```python
Faker("hexify", text="acct:^@example.com")
```

This quickly results in duplicate user IDs being generated:

```
$ make shell
>>> factories.User()
User(..., h_userid='acct:f@example.com')

>>> factories.User()
User(..., h_userid='acct:4@example.com')

>>> factories.User()
User(..., h_userid='acct:4@example.com')
```

Change it to use a sequence number instead, which is guaranteed to never generate the same ID twice:

```python
$ make shell
>>> factories.User()
User(..., h_userid='acct:user_1@example.com')

>>> factories.User()
User(..., h_userid='acct:user_2@example.com')

>>> factories.User()
User(..., h_userid='acct:user_3@example.com')
```

### DIFF
--- a/tests/factories/attributes.py
+++ b/tests/factories/attributes.py
@@ -1,6 +1,6 @@
 """Shared attributes used by test factories."""
 
-from factory import Faker
+from factory import Faker, Sequence
 
 OAUTH_CONSUMER_KEY = Faker("hexify", text="Hypothesis" + "^" * 32)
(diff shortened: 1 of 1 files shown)


marcospri (approved):

seanh:
Rebased

PR #3503, opened 2021-12-09
Handle all async errors

marcospri:
## Testing

- Make a change similar to:

```diff
diff --git a/lms/services/blackboard_api/client.py b/lms/services/blackboard_api/client.py
index 838dfd85..03ba6b97 100644
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -156,7 +156,7 @@ class BlackboardAPIClient:
                 self_enrollment_groups.append(group)
                 self_enrollment_check_urls.append(
                     self._api._api_url(  # pylint: disable=protected-access
-                        f"/learn/api/public/v2/courses/uuid:{course_id}/groups/{group['id']}"
+                        f"/learn/api/public/v25/courses/uuid:{course_id}/groups/{group['id']}"
                     )
                 )

```

- Launch `localhost (make devdata) Groups Assignment` has a student.
- You should get a dialog with some details about the failed request.

Screenshot from 2022-01-20 16-15-23

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -65,14 +65,14 @@ class ExternalRequestError(Exception):
 
 
 class ExternalAsyncRequestError(Exception):
-    def __init__(
-        self, message=None, request=None, response=None, validation_errors=None
-    ):
+    def __init__(self, response=None):
         super().__init__()
-        self.message = message
-        self.request = request
         self.response = response
-        self.validation_errors = validation_errors
+
+        # For compatibility with ExternalRequestError.
+        self.message = None
+        self.request = None
+        self.validation_errors = None
 
     @property
     def _request_info(self):
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -108,7 +108,7 @@ class ExternalAsyncRequestError(Exception):
     @property
     def response_body(self) -> None:
         """Return the response body."""
-        return None
+        return getattr(self.response, "sync_text", None)
 
     def __repr__(self) -> str:
         return _repr_external_request_exception(self)
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -92,6 +92,49 @@ class ExternalRequestError(Exception):
         return repr(self)
 
 
+class ExternalAsyncRequestError(ExternalRequestError):
+    def __init__(
+        self,
+        message=None,
+        response=None,
+        exception: Optional[Exception] = None,
+    ):
+        if exception and message:
+            message = f"{message}: {exception.__class__.__name__}"
+
+        super().__init__(message=message, response=response)
+        self.exception = exception
+
+        self._request_info = getattr(response, "request_info", None) or getattr(
+            exception, "request_info", None
+        )
+
+    @property
+    def url(self) -> Optional[str]:
+        """Return the request's URL."""
+        return str(url) if (url := getattr(self._request_info, "url", None)) else None
+
+    @property
+    def method(self) -> Optional[str]:
+        """Return the HTTP request method."""
+        return getattr(self._request_info, "method", None)
+
+    @property
+    def status_code(self) -> Optional[int]:
+        """Return the response's status code."""
+        return getattr(self.response, "status", None)
+
+    @property
+    def request_body(self) -> None:
+        """Return the request body."""
+        return None
+
+    @property
+    def response_body(self) -> None:
+        """Return the response body."""
+        return None
+
+
 class OAuth2TokenError(ExternalRequestError):
     """
     A problem with an OAuth 2 token for an external API.
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -57,56 +57,27 @@ class ExternalRequestError(Exception):
         """Return the response body."""
         return getattr(self.response, "text", None)
 
-    def __repr__(self):
-        # Include the details of the request and response for debugging. This
-        # appears in the logs and in tools like Sentry and Papertrail.
-        request = (
-            "Request("
-            f"method={self.method!r}, "
-            f"url={self.url!r}, "
-            f"body={self.request_body!r}"
-            ")"
-        )
-
-        response = (
-            "Response("
-            f"status_code={self.status_code!r}, "
-            f"reason={self.reason!r}, "
-            f"body={self.response_body!r}"
-            ")"
-        )
-
-        # The name of this class or of a subclass if one inherits this method.
-        class_name = self.__class__.__name__
-
-        return (
-            f"{class_name}("
-            f"message={self.message!r}, "
-            f"cause={self.__cause__!r}, "
-            f"request={request}, "
-            f"response={response}, "
-            f"validation_errors={self.validation_errors!r})"
-        )
+    def __repr__(self) -> str:
+        return _repr_external_request_exception(self)
 
     def __str__(self):
         return repr(self)
 
 
-class ExternalAsyncRequestError(ExternalRequestError):
+class ExternalAsyncRequestError(Exception):
     def __init__(
-        self,
-        message=None,
-        response=None,
-        exception: Optional[Exception] = None,
+        self, message=None, request=None, response=None, validation_errors=None
     ):
-        if exception and message:
-            message = f"{message}: {exception.__class__.__name__}"
+        super().__init__()
+        self.message = message
+        self.request = request
+        self.response = response
+        self.validation_errors = validation_errors
 
-        super().__init__(message=message, response=response)
-        self.exception = exception
-
-        self._request_info = getattr(response, "request_info", None) or getattr(
-            exception, "request_info", None
+    @property
+    def _request_info(self):
+        return getattr(self.response, "request_info", None) or getattr(
+            self.__cause__, "request_info", None
         )
 
     @property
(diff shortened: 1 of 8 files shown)


lms/services/exceptions.py
```diff
@@ -91,6 +91,49 @@ def __str__(self):
         return repr(self)
 
 
+class ExternalAsyncRequestError(ExternalRequestError):
```
seanh, line 94: I wonder whether the subclassing is worthwhile here (potentially coupling the two classes together), or should they just be two completely separate classes that happen to provide the same interface?

seanh (approved):
I guess dealing with async external request errors was different enough that we need a separate exception class, rather than extending the one `ExternalRequestError` class to be able to handle both at once?

lms/services/exceptions.py
marcospri: Sub-classing gives us the `__repr__`  of `ExternalRequestError`.

We could try to cram everything on one class but I reckon it could end up being more confusing, it will have 2 or 3 conditionals in each method.

lms/services/exceptions.py
```diff
@@ -91,6 +91,49 @@ def __str__(self):
...
+    def __init__(
+        self,
+        message=None,
+        response=None,
+        exception: Optional[Exception] = None,
+    ):
+        if exception and message:
+            message = f"{message}: {exception.__class__.__name__}"
```
seanh, line 102: I like the idea of reporting the original exception that caused the `ExternalAsyncRequestError`. But:

1. I think we can already access the original exception as `self.__cause__` so we don't need the `exception` argument
2. I think it's good to keep `__cause__` and `message` as two separate attributes, rather than combining them into one string
3. I think I'd like to add `__cause__` support to `ExternalRequestError` so that `ExternalRequestError` and `ExternalAsyncRequestError` remain the same and so that all request errors have `__cause__` reporting not just the async ones

It think we could add `__cause__` to `ExternalRequestError.__repr__()` (which appears in Papertrail) and to [the `external_request_error()` exception view](https://github.com/hypothesis/lms/blob/594cb2793597578df82222f88b62ab777441c4a6/lms/views/api/exceptions.py#L111-L153) (which controls what appears in Sentry and in the error dialog).

For example currently if a request to the Canvas API times out `ExternalRequestError` doesn't make it clear that the problem was a timeout. If you hack the code like this:

```diff
diff --git a/lms/services/canvas_api/_basic.py b/lms/services/canvas_api/_basic.py
index ec641864..3eebf12b 100644
--- a/lms/services/canvas_api/_basic.py
+++ b/lms/services/canvas_api/_basic.py
@@ -99,6 +99,7 @@ class BasicClient:
         )
 
     def _send_prepared(self, request, schema, timeout, request_depth=1):
+        request.url = "https://www.doesntexist.com/"
         response = None
 
         try:
```

You'll get an error dialog like this:

Screenshot from 2022-01-20 17-58-48

If we add `__cause__` handling like this:

```diff
diff --git a/lms/views/api/exceptions.py b/lms/views/api/exceptions.py
index f6d78591..184d0004 100644
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -140,6 +140,7 @@ class APIExceptionViews:
         return self.error_response(
             message=message,
             details={
+                "cause": str(self.context.__cause__),
                 "request": {
                     "method": self.context.method,
                     "url": strip_queryparams(self.context.url),
```

then the cause will be shown in the error dialog:

Screenshot from 2022-01-20 18-11-57

We should also add `sentry_sdk.set_context()` calls for `__cause__` like we have for `request`, `response` and `validation_errors` so that it appears in Sentry, and we should add `__cause__` to `ExternalRequestError.__repr__()` so that it appears in Papertrail.

lms/services/exceptions.py
seanh: I think we should avoid `ExternalAsyncRequestError` subclassing `ExternalRequestError` for a couple of reasons:

First, semantically it means that an `ExternalAsyncRequestError` **is a** `ExternalRequestError` and I'm not sure that we want that. For example `isinstance(async_external_request_error, ExternalRequestError)` will be `True`, `catch ExternalRequestError` will catch `AsyncExternalRequestError`'s, etc.

Second, it means that the `ExternalRequestError` class is now part of the `ExternalAsyncRequestError`. You can no longer change `ExternalRequestError` without changing `ExternalAsyncRequestError` as well. For example if I add a new method to `ExternalRequestError` (which a PR that I'm working on does) `ExternalAsyncRequestError` will inherit that method but the inherited method probably won't work if called on an `ExternalAsyncRequestError` instance because `request` and `response` will be different types of object than with `ExternalRequestError`.

I think there's a high chance of making a change to `ExternalRequestError` and not realising that `ExternalAsyncRequestError` is a subclass and you've affected or broken `ExternalAsyncRequestError`.

I think there may already be some unintended side-effects of the inheritance. For example an `ExternalAsyncRequestError` object has `request` and `validation_errors` attributes whose values are `None`.

I think we want to keep these two classes decoupled from each other but providing the same interface. The simplest way for them to share code might be for them to use shared helper functions?

```python
    ...

    def __repr__(self):
        return _repr(self)

    def __str__(self):
        return repr(self)


class ExternalAsyncRequestError(Exception):
    ...

    def __repr__(self):
        return _repr(self)

    def __str__(self):
        return repr(self)


def _repr(exception):
    ...
```

lms/services/exceptions.py
```diff
@@ -91,6 +91,49 @@ def __str__(self):
...
+
+    @property
+    def request_body(self) -> None:
+        """Return the request body."""
+        return None
+
+    @property
+    def response_body(self) -> None:
+        """Return the response body."""
+        return None
```
seanh, line 134: Is there no way to get the failed request and response bodies? They're useful for debugging

seanh (changes requested):

lms/services/exceptions.py
marcospri: `__cause__` handled in a different PR

lms/services/exceptions.py
marcospri: I think `ExternalAsyncRequestError` being a `ExternalRequestError` from OOP PoV was fine but I agree that that mostly ties them together implementation wise which could be a problem.

Separated them and added `ExternalAsyncRequestError` to the same exception view from now.

lms/services/exceptions.py
marcospri: I think is possible but I rather do that in a separate PR as it might be a bit complex.

lms/services/exceptions.py
```diff
@@ -57,37 +57,62 @@ def response_body(self) -> Optional[str]:
...
+
+    @property
+    def status_code(self) -> Optional[int]:
+        """Return the response's status code."""
+        return getattr(self.response, "status", None)
+
+    @property
+    def request_body(self) -> None:
+        """Return the request body."""
+        return None
```
seanh, line 101: No support for request bodies?

lms/services/exceptions.py
```diff
@@ -57,37 +57,62 @@ def response_body(self) -> Optional[str]:
...
+
+    @property
+    def reason(self) -> Optional[str]:
+        """Return the response's HTTP reason string, e.g. 'Bad Request'."""
+        return getattr(self.response, "reason", None)
+
+    @property
+    def response_body(self) -> None:
+        """Return the response body."""
+        return None
```
seanh, line 111: No support for response bodies?

seanh:
Testing Notes
=============

There's two different ways that `ExternalAsyncRequestError` can be raised:

1. `AsyncOAuthHTTPService._prepare_requests()` [raises `ExternalAsyncRequestError` if it catches an `aiohttp.ClientError`](https://github.com/hypothesis/lms/blob/d26f6a1dccf178343776af9bd8a18e71265ba5fa/lms/services/async_oauth_http.py#L65-L71). To simulate:

   ```diff
   diff --git a/lms/services/async_oauth_http.py b/lms/services/async_oauth_http.py
   index d00a1152..3e5dd9fe 100644
   --- a/lms/services/async_oauth_http.py
   +++ b/lms/services/async_oauth_http.py
   @@ -55,7 +55,7 @@ async def _prepare_requests(method, urls, **kwargs):
                    _async_request(
                        session,
                        method,
   -                    url,
   +                    "https://68767431846138.com/",
                        **kwargs,
                    )
                )
   ```

   When this happens we're not currently getting details of the failed request in the error. We're also not getting the failed response but I'm not sure we can: I think this `except` block might only be triggered in cases where no response was received?

   Error dialog:

   Screenshot from 2022-02-01 18-07-47

   The request and response are empty in Sentry too:

   Screenshot from 2022-02-01 18-10-34

2. `BlackboardAPIClient.course_groups()` [raises `ExternalAsyncRequestError` if any of the responses aren't a 200 or a 403](https://github.com/hypothesis/lms/blob/d26f6a1dccf178343776af9bd8a18e71265ba5fa/lms/services/blackboard_api/client.py#L174-L177). To simulate:

   ```diff
   diff --git a/lms/services/async_oauth_http.py b/lms/services/async_oauth_http.py
   index d00a1152..a042cd30 100644
   --- a/lms/services/async_oauth_http.py
   +++ b/lms/services/async_oauth_http.py
   @@ -55,7 +55,7 @@ async def _prepare_requests(method, urls, **kwargs):
                    _async_request(
                        session,
                        method,
   -                    url,
   +                    "http://httpbin.org/status/418",
                        **kwargs,
                    )
                )
   ```

   In this case we do seem to get the request and response. Error dialog:

   Screenshot from 2022-02-01 18-14-34

   But the request and response _bodies_ are missing from Sentry because `ExternalAsyncRequestError` doesn't support the `request_body` and `response_body` attributes:

   Screenshot from 2022-02-01 18-15-17

seanh (approved):
This looks like a solid foundation for me.

Do you think we should remove the `message`, `request` and `validation_error` arguments from `ExternalAsyncRequestError.__init__()` since they're never used? Only the `response` argument is ever passed to it. (`message` is actually passed in this PR currently but in both cases I think it should be removed.)

In the case of the `except aiohttp.ClientError` block in `AsyncOAuthHTTPService` do we need a follow-up issue for passing `request` and if possible also `response` to `ExternalAsyncRequestError()`?

Do we need a follow-up issue for implementing the `request_body` and `response_body` attributes in `ExternalAsyncRequestError`?

PR #3514, opened 2021-12-14
Validate blackboard API params

marcospri:


lms/views/api/blackboard/sync.py
```diff
@@ -11,68 +12,68 @@ def __init__(self, request):
...
+    def get_blackboard_groups(self, course, group_set_id):
         lti_user = self.request.lti_user
-        group_set_id = self.group_set()
-        course_id = self.request.json["course"]["context_id"]
-        course = self.get_course(course_id)
 
         if lti_user.is_learner:
             user = self.request.find_service(UserService).get(
-                self.request.find_service(name="application_instance").get_current(),
+                course.application_instance,
```
marcospri, line 45: Using the relationship here instead of another separate call to `ai_service.get_current()`

lms/validation/_api.py
```diff
@@ -132,3 +137,31 @@ def validate(self, data, **_kwargs):
...
+        class GroupInfo(Schema):
+            class Meta:
+                unknown = INCLUDE
+
+        lms = fields.Nested(LMS, required=True)
+        course = fields.Nested(Course, required=True)
+        assignment = fields.Nested(Assignment, required=True)
+        group_info = fields.Nested(GroupInfo, required=True)
+
+        gradingStudentId = fields.Str(required=False, allow_none=True)
```
marcospri, line 162: This one is not sent by use, it's injected by the frontend so the naming has JS conventions.

lms/views/api/blackboard/sync.py
```diff
@@ -11,68 +12,69 @@ def __init__(self, request):
...
-            )
+            .get(tool_consumer_instance_guid, resource_link_id)
             .extra["group_set_id"]
         )
 
     def to_groups_groupings(self, course, groups):
         return [
             self.grouping_service.upsert_with_parent(
-                tool_consumer_instance_guid=self.tool_consumer_instance_guid,
+                tool_consumer_instance_guid=course.application_instance.tool_consumer_instance_guid,
```
marcospri, line 77: `upsert_with_parent` might not need to take `tool_consumer_instance_guid` at all and do this instead but I reckon that would be better on a separate PR.

lms/validation/_api.py
```diff
@@ -132,3 +137,31 @@ def validate(self, data, **_kwargs):
             raise ValidationError(
                 "both bookID and cfi are mandatory with type is 'vitalsource'"
             )
+
+
+class APISyncBlackboardSchema(PyramidRequestSchema):
```
seanh, line 142:
```suggestion
class APIBlackboardSyncSchema(PyramidRequestSchema):
```

Ideally I think it should be named `BlackboardSyncAPISchema`. In speaking we call it the "Blackboard sync API" and we'd call the schema the "Blackboard sync API schema", so `BlackboardSyncAPISchema`

But there's something weird going on with the naming of the schemas in this file: `APIRecordSpeedgraderSchema`, `APIReadResultSchema`, `APIRecordResultSchema`, `APICreateAssignmentSchema`, etc. If it's the "create assignment API" then I'd think it'd be called `CreateAssignmentAPISchema`. I'm not sure why we keep putting `API` at the start.

But if we want to be consistent with those existing names we should put `API` at the start as well

lms/validation/_api.py
```diff
@@ -1,12 +1,17 @@
...
 
 from lms.validation._base import JSONPyramidRequestSchema, PyramidRequestSchema
 
-__all__ = ["APIRecordSpeedgraderSchema", "APIReadResultSchema", "APIRecordResultSchema"]
+__all__ = [
+    "APIRecordSpeedgraderSchema",
+    "APIReadResultSchema",
+    "APIRecordResultSchema",
+    "APISyncBlackboardSchema",
+]
```
seanh, line 14:
```suggestion
```

We should just get rid of the `__all__`, I don't think it's needed

lms/validation/_api.py
```diff
@@ -132,3 +137,31 @@ def validate(self, data, **_kwargs):
...
+        course = fields.Nested(Course, required=True)
+        assignment = fields.Nested(Assignment, required=True)
+        group_info = fields.Nested(GroupInfo, required=True)
+
+        gradingStudentId = fields.Str(required=False, allow_none=True)
+
+    authUrl = fields.URL(required=True)
+    path = fields.Str(required=True)
+
+    data = fields.Nested(Data, required=True)
```
seanh, line 167: The schema looks odd to me: I don't think that `authUrl` and `path` are part of the request body that the frontend posts to the backend. And the contents of `data` are just the top-level contents of the request body, they're not in a `data` subdict. Won't this result in a validation error when the frontend posts to the sync API? As I don't think the validation schema matches the shape of the request

lms/resources/_js_config/__init__.py
```diff
@@ -510,28 +511,30 @@ def _canvas_sync_api(self):
...
-                    "context_id": req.params["context_id"],
-                },
-                "assignment": {
-                    "resource_link_id": req.params["resource_link_id"],
-                },
-                "group_info": {
-                    key: value
-                    for key, value in req.params.items()
-                    if key in GroupInfo.columns()
+        return APISyncBlackboardSchema(req).load(
```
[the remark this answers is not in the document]
seanh: I like the idea. I think Marshmallow schemas have two methods `dump()` and `load()`:

1. `dump()` serializes a model object (possibly just a dict) into a dict. See <https://marshmallow.readthedocs.io/en/stable/quickstart.html#serializing-objects-dumping>.

2. `load()` deserializes a dict into a model object (possibly just a dict): https://marshmallow.readthedocs.io/en/stable/quickstart.html#deserializing-objects-loading

So I think if we want to use `APISyncBlackboardSchema` to serialize the data dict to send to the frontend we want to use `APISyncBlackboardSchema.load()` for that.

I think the schema is only about the request body (the `data` part) so I think it'd be:

```python
return {
    "authUrl": ...,
    "path": ...,
    "data": APISyncBlackboardSchema(req).dump(...),
}
```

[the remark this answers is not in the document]
seanh: I'm not sure that dumping actually requires the `request` object, so I think you should be able to do `APISyncBlackboardSchema().dump({...})`. But this might be an existing issue with `PyramidRequestSchema`, the `request` should be an argument to `parse()` rather than to `__init__()`

seanh (changes requested):
I think this puts Blackboard one step ahead of the Canvas sync API which I don't think has validation :+1:

I think there might be some refactoring of the sync views that could be split into a separate PR?

I'm not sure that the schema in this PR works: it needs to validate just the `data` part, the frontend doesn't send `authUrl`, `path` and `data` params in the request body.

lms/resources/_js_config/__init__.py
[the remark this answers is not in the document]
seanh: Actually, rather than having `APISyncBlackboardSchema` just validate the data to be serialized, maybe it should generate it? Something like:

```python
>>> APISyncBlackboardSchema.dump(request)
{
    "lms": {
        "tool_consumer_instance_guid": "...",
        ...,
    },
    ...
}
```

This would mean that if we want to change the sync API params (e.g. to add a new param) we only need to change it in one place: in the `APISyncBlackboardSchema` class, since the schema class is responsible for both generating the params that we send to the frontend and validating the params that the frontend sends back to us.

Some of the authentication-related schema classes do something similar. For example `BearerTokenSchema` both serializes a bearer token from an `LTIUser` and deserializes a bearer token into an `LTIUser`. `OAuthCallbackSchema` also does something similar

marcospri:
A more focused PR on https://github.com/hypothesis/lms/pull/3574

PR #3515, opened 2021-12-14
Debounce fetching students grades

esanzgar:
Every time the next or previous arrow is pressed in the LMS grader it
makes a network request. This could cause a big number of meaningless
requests. In order to avoid this, we have debounced the requests of
student grades, so that only the last request made within the last 500
milliseconds is executed, the rest are ignored.

On the other hand, this causes a delay of 500 milliseconds also when a
specific user is selected via the dropdown.


esanzgar:
We discussed this PR as a team. We don't want to add more that 200ms debouncing time because it makes the application to feel sluggish. There is quite a fair amount of clicks that can be caught in 200ms so I think it is worth the effort.

I will substitute this solution for `lodash.debounce` in a follow up PR.

esanzgar:
Closed in favour of the `lodash.debounce` version: https://github.com/hypothesis/lms/pull/3519

PR #3518, opened 2021-12-17
Bulk upsert helper

marcospri:
I initially had this a branch against your upsert-helper one but had to rebase here so now the diff is a mess. Pointing to master for now.

I wrote a couple of version of this: 

- https://github.com/hypothesis/lms/pull/3518/commits/f34ac448181ad5f94a89fb9d2e8ad49ebe0c9548 using only the ORM. 

There's a lot of bookkeeping there which make the code not that clear and it's also less performatn and it does have a race condition (between the getting and the creation, the one row version in master also has that problem).

- https://github.com/hypothesis/lms/pull/3518/commits/6c2cda0dd6fef8c61c2df7df5bf85794ac5e46a4 using both the core api and the ORM

This is the version you'll see also if you look at the full diff. This is 90% of what we already have on "bulk_actions" but just as a function.

Maybe this makes the upsert code even more scattered through tthe code base but all the DB logic is now on the function and the existing "bulk_action" just calls that with the config in the model.

I don't think we necessary want to configure the upsert functionally at the model level, seems to me more like a query you might want to use with different parameters in different situations rather than something more intrinsic to the model.

### DIFF
--- a/lms/services/_upsert.py
+++ b/lms/services/_upsert.py
@@ -1,6 +1,7 @@
 """A helper for upserting into DB tables."""
 
 from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy import or_, and_
 
 
 def upsert(db, model_class, query_kwargs, update_kwargs):
@@ -14,3 +15,52 @@ def upsert(db, model_class, query_kwargs, update_kwargs):
         setattr(model, key, value)
 
     return model
+
+
+def bulk_upsert(db, models, search_columns, update_columns):
+    def _search_criteria_values(model, search_columns):
+        return {column: getattr(model, column.key) for column in search_columns}
+
+    if not models:
+        return []
+
+    ModelsClass = models[0].__class__
+    assert all((model.__class__ == ModelsClass for model in models))
+
+    search_criteria = []
+    existing_models_by_search = {}
+    for model in models:
+        model_search_criteria = _search_criteria_values(model, search_columns)
+        model_search_values = tuple(model_search_criteria.values())
+
+        # Build a lookup dict from the value of the search column to the model
+        existing_models_by_search[model_search_values] = model
+
+        # For every model we have to "and" the search columns
+        search_criteria.append(
+            and_(*[column == value for column, value in model_search_criteria.items()])
+        )
+
+    # Aggregate all search criteria in a big "or" to find all existing columns
+    db_models = db.query(ModelsClass).filter(or_(*search_criteria)).all()
+
+    for db_model in db_models:
+        model_search_criteria = _search_criteria_values(db_model, search_columns)
+        model_search_values = tuple(model_search_criteria.values())
+
+        # For every existing model, update the neccessary columns finding the original model in the lookup dict
+        update_model = existing_models_by_search[model_search_values]
+        for update_column in update_columns:
+            setattr(
+                db_model,
+                update_column.key,
+                getattr(update_model, update_column.key),
+            )
+
+        del existing_models_by_search[model_search_values]
+
+    # We deleted models from the lookup dict as we went through the existing ones.
+    # The only ones left are the ones that need an insert.
+    (db.add(model) for model in existing_models_by_search)
+
+    return list(existing_models_by_search.values()) + db_models
--- a/lms/services/_upsert.py
+++ b/lms/services/_upsert.py
@@ -1,90 +1,22 @@
 """A helper for upserting into DB tables."""
 
 from copy import deepcopy
+from typing import List
+
+from sqlalchemy import column, inspect, tuple_
 from sqlalchemy.dialects.postgresql import insert
-from sqlalchemy import inspect, column
-from sqlalchemy.orm.exc import NoResultFound
-from sqlalchemy import or_, and_, tuple_
 from zope.sqlalchemy import mark_changed
 
 
-def upsert(db, model_class, query_kwargs, update_kwargs):
-    try:
-        model = db.query(model_class).filter_by(**query_kwargs).one()
-    except NoResultFound:
-        model = model_class(**query_kwargs)
-        db.add(model)
-
-    for key, value in update_kwargs.items():
-        setattr(model, key, value)
-
-    return model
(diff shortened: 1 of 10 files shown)


lms/services/_upsert.py
```diff
@@ -0,0 +1,86 @@
...
+
+    index_elements_columns = [column(c) for c in index_elements]
+
+    stmt = insert(model_class).values(values)
+    stmt = stmt.on_conflict_do_update(
+        # The columns to use to find matching rows.
+        index_elements=index_elements,
+        # The columns to update.
+        set_={element: getattr(stmt.excluded, element) for element in update_columns},
+    ).returning(*index_elements_columns)
```
marcospri, line 67: If the index_elements_columns where enough to identify the rows for the upsert it will work to get the rows back.

lms/services/_upsert.py
```diff
@@ -0,0 +1,86 @@
...
+    result = db.execute(stmt).all()
+
+    # Let SQLAlchemy know that something has changed, otherwise it will
+    # never commit the transaction we are working on and it will get rolled
+    # back
+    mark_changed(db)
+
+    # Return ORM objects based on index_elements
+    return (
+        db.query(model_class).filter(tuple_(*index_elements_columns).in_(result)).all()
```
marcospri, line 78: This makes an extra query but it mean that we can return full ORM objects

lms/services/grouping.py
```diff
@@ -32,52 +33,49 @@ def generate_authority_provided_id(
...
-    def upsert_with_parent(  # pylint: disable=too-many-arguments
-        self,
-        tool_consumer_instance_guid,
-        lms_id,
-        lms_name,
-        parent: Grouping,
-        type_: Grouping.Type,
-        extra=None,
-    ):
+    class _GroupingData(TypedDict):
```
marcospri, line 36: This might be a bit meh I still prefer it over just a list of random dicts, it's the same in practice but IMO this more explicit than a long comment.

lms/services/grouping.py
```diff
@@ -32,52 +33,49 @@ def generate_authority_provided_id(
...
-            )
-            self._db.add(grouping)
+            grouping_data["authority_provided_id"] = authority_provided_id
+            grouping_data["application_instance_id"] = self._application_instance.id
+            grouping_data["parent_id"] = grouping_data["parent"].id
 
-        grouping.lms_name = lms_name
-        grouping.extra = extra
+            del grouping_data["parent"]
+            del grouping_data["tool_consumer_instance_guid"]
```
marcospri, line 70: This is not great, instead of modifying the original dict I think it would be better to crate a new dict with just the required keys.

tests/unit/conftest.py
```diff
@@ -177,13 +177,15 @@ def httpretty_():
...
-    return factories.ApplicationInstance(
+def application_instance(db_session, pyramid_request):
+    ai = factories.ApplicationInstance(
         consumer_key=pyramid_request.lti_user.oauth_consumer_key,
         developer_key="TEST_DEVELOPER_KEY",
         provisioning=True,
         settings=ApplicationSettings({}),
     )
+    db_session.flush()
+    return ai
```
marcospri, line 188: We need to flush the AI to get the PK populated.

tests/unit/lms/db/_bulk_action_test.py
```diff
@@ -35,16 +34,17 @@ def test_upsert(self, db_session):
...
 
-        result = BulkAction(db_session).upsert(
-            self.TableWithBulkUpsert,
-            [
-                {"id": 1, "name": "update_old", "other": "post_1"},
-                {"id": 3, "name": "create_with_id", "other": "post_3"},
-                {"id": 4, "name": "over_block_size", "other": "post_4"},
-            ],
-        )
+        upsert_rows = [
```
marcospri, line 37: I didn't move these tests, just tweaked them to make sure that the changes didn't break it.

lms/services/_upsert.py
```diff
@@ -0,0 +1,86 @@
...
+
+    result = db.execute(stmt).all()
+
+    # Let SQLAlchemy know that something has changed, otherwise it will
+    # never commit the transaction we are working on and it will get rolled
+    # back
+    mark_changed(db)
+
+    # Return ORM objects based on index_elements
+    return (
```
marcospri, line 77: This function already takes a few parameters but maybe this part could also be under a boolean flag `return_orm_objects` or similar.

If you just want to upsert and forget like we do with files you don't want to pay the price for this query + serializing all the objects.

lms/services/_upsert.py
```diff
@@ -0,0 +1,86 @@
...
+        db.query(model_class).filter(tuple_(*index_elements_columns).in_(result)).all()
+    )
+
+
+def _get_columns_onupdate(model_class):
+    """Get which columns which have an onupdate clause and its value."""
+    model_details = inspect(model_class)
+
+    return [(c.name, c.onupdate.arg) for c in model_details.c if c.onupdate]
```
seanh, line 86: I wonder if we can avoid the cleverness here? I think the only column that has `onupdate` in the LMS app is the `updated` column on various models and its value is always `sa.func.now()`. I wonder if we could just have hard-coded code for that particular column

seanh (changes requested):
Is `db/_bulk_action.py` really needed anymore? It looks like `services/_upsert.py::bulk_upsert()` is pretty much a standalone function, or could be, and then `GroupingService` seems to just call `services/_upsert.py::bulk_upsert()` directly?

I wonder if we want to do a refactor first to change `db/_bulk_action.py` into just a services helper function and change its existing users to use it that way?

I think that may result in more straightforward and easier to understand code: just a function with parameters and a docstring is quite easy to follow. The existing bulk upsert stuff is quite unique and unusual: you have this `BulkAction` class that's attached to SQLAlchemy via a custom session, and also bulk action configs on the models, etc.

marcospri:
Cleaner version on https://github.com/hypothesis/lms/pull/3518

PR #3519, opened 2021-12-17
Debounce fetching students grades

esanzgar:
In the LMS grader, every time that the next or previous arrow is pressed it
makes a network request. This could cause a big number of meaningless
requests. In order to avoid this, we have debounced the requests of
student grades, so that only the last request made within the last 200
milliseconds is executed, the rest are ignored.

This causes a delay of 200 milliseconds also when a specific user is
selected via the dropdown.

I have also fixed the following issue that I noticed: if you give a zero
to an student, navigate to another student and come back to the terrible
student the zero grade was not showing. I corrected that.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
+++ b/lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
@@ -3,9 +3,15 @@ import {
   LabeledButton,
   Spinner,
 } from '@hypothesis/frontend-shared';
-
 import classnames from 'classnames';
-import { useEffect, useLayoutEffect, useState, useRef } from 'preact/hooks';
+import {
+  useEffect,
+  useLayoutEffect,
+  useState,
+  useRef,
+  useCallback,
+} from 'preact/hooks';
+import debounce from 'lodash.debounce';
 
 import ErrorDialog from './ErrorDialog';
 import { useService, GradingService } from '../services';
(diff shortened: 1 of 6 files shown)


esanzgar:
Before:

https://user-images.githubusercontent.com/8555781/146576558-b75ba538-2823-4d0e-b660-d974fd03c79c.mov

After:


https://user-images.githubusercontent.com/8555781/146576607-0af340fe-570d-429d-a3e6-32ac7a6d5712.mov

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -35,39 +41,54 @@ const useFetchGrade = (student, setFetchGradeError) => {
...
+  const debouncedFetchGrade = useCallback(
+    debounce(
+      /**
+       * @param {StudentInfo} student
+       * @param {[boolean]} keepResult
+       */
+      async (/** @type {StudentInfo} */ student, keepResult) => {
         try {
           const { currentScore } = await gradingService.fetchGrade({ student });
-          if (!ignoreResults && currentScore) {
```
esanzgar, line 50: If `currentScore` was `null` it was treated the same as `0` (falsy value)

lms/static/scripts/frontend_apps/services/grading.js
```diff
@@ -10,7 +10,7 @@ import { apiCall } from '../utils/api';
  * The fetched grade success result.
  *
  * @typedef FetchGradeResult
- * @prop {number} currentScore - The fetched grade
+ * @prop {number|null} currentScore - The fetched grade
```
esanzgar, line 13: I added `null` because I confirmed that value empirically in Blackboard. Is it ever `undefined`?

lms/static/scripts/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "noEmit": true,
     "strict": true,
     "target": "ES2020",
-    "useUnknownInCatchVariables": false
+    "useUnknownInCatchVariables": false,
+    "esModuleInterop": true
```
esanzgar, line 13: `esModuleInterop` is necessary for TS to understand the import of a default export: `import debounce from 'loadash/debounce'`

lms/static/scripts/frontend_apps/components/SubmitGradeForm.js
```diff
@@ -35,39 +41,54 @@ const useFetchGrade = (student, setFetchGradeError) => {
...
-      fetchData();
+      },
+      200
+    ),
+    []
+  );
+
+  useEffect(() => {
+    /** @type [boolean] */
+    const keepResult = [true];
```
esanzgar, line 75: keepResult needs to be an Array or an Object, because the value of this variable could change during the lifetime of the debounced function execution. We pass by reference to the debounce function, so the reference leads to the current value, when it is needed.

esanzgar:
We have chatted about this PR. It appears that the approach suggested here is a bit too complicated ([without `lodash.debounce` was simpler](https://github.com/hypothesis/lms/pull/3515/files)) for the limited benefit. [@lyzadanger has suggested](https://github.com/hypothesis/lms/pull/3515/files) instead to streamline/refactor the fetching of the student grades. In the past, [I suggested](https://github.com/hypothesis/lms/pull/2313/files) to use the hook `useReducer` to coordinate the different states that are interdependent.

I will close this PR and I will open a new one to tackle just the issue we discovered about the `zero` grade not showing.

PR #3529, opened 2021-12-20
Temporary fix around distutils deprecation

marcospri:
Currently master test are broken as this warning is produced:

```DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.```

I'm not 100% sure of everything at play here but the latest setuptools release (from yesterday) has this in the change log:

```#2896: Setuptools once again makes its local copy of distutils the default. To override, set SETUPTOOLS_USE_DISTUTILS=stdlib.```
https://github.com/pypa/setuptools/blob/main/CHANGES.rst

Applying that fixed/hides the warning and makes the build green again.


esanzgar (approved):
We are not in a hurry to fix this, so if you want to postpone this until the rest of the backend team is available it's OK. Otherwise, the PR feels sensible.

PR #3533, opened 2021-12-24
Fix issue displaying a `zero` grade

esanzgar:
If a student was given a zero grade, the grade was not showing when
navigating to another student and coming back to the `zero` graded
student.

Before:

https://user-images.githubusercontent.com/8555781/146576558-b75ba538-2823-4d0e-b660-d974fd03c79c.mov

After:

https://user-images.githubusercontent.com/8555781/146576607-0af340fe-570d-429d-a3e6-32ac7a6d5712.improvement

### DIFF
--- a/lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
@@ -100,7 +100,7 @@ describe('SubmitGradeForm', () => {
     assert.strictEqual(wrapper.find(inputSelector).prop('defaultValue'), '');
   });
 
-  it('clears out the previous grade when next students does not have a grade (`null` value)', async () => {
+  it('clears the displayed grade value if the currently-focused student has an empty grade', async () => {
     const wrapper = renderForm();
 
     await waitForGradeFetch(wrapper);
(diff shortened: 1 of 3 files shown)


lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
```diff
@@ -95,9 +95,46 @@ describe('SubmitGradeForm', () => {
...
+
+    await waitForGradeFetch(wrapper);
+    assert.strictEqual(wrapper.find(inputSelector).prop('defaultValue'), '10');
+
+    fakeGradingService.fetchGrade.resolves({});
+    wrapper.setProps({ student: fakeStudentAlt });
+    await waitForGradeFetch(wrapper);
+
+    assert.strictEqual(wrapper.find(inputSelector).prop('defaultValue'), '');
+  });
```
esanzgar, line 127: I am not sure about this situation (see below).

lms/static/scripts/frontend_apps/services/grading.js
```diff
@@ -10,7 +10,7 @@ import { apiCall } from '../utils/api';
  * The fetched grade success result.
  *
  * @typedef FetchGradeResult
- * @prop {number} currentScore - The fetched grade
+ * @prop {number|null} [currentScore] - The fetched grade
```
esanzgar, line 13: This type was incomplete. I have verified the possibility of `null` value in Blackboard.

Regarding adding also `undefined`, I did it as a defensive programming design just in case we encounter the situation the under unforeseen circumstances.

lyzadanger (approved):
Thanks for looking into this and solving it. Wrangling with nullish or non-numeric types here is a little messy, but I can't think of a better way than what you've done. Thank you!

In the future it may be worth taking a closer look at when and how the `GRADE_MULTIPLIER` is applied to see if it could be abstracted more. It would be nice to have a setter function that could take any value for `grade` and decide what to do with it instead of that logic happening scattershot, externally.

lms/static/scripts/frontend_apps/components/test/SubmitGradeForm-test.js
esanzgar: I will remove this test.

PR #3563, opened 2022-01-11
Ignore the deprecation warnings from Marshmallow after setuptools release

marcospri:
Apply the same fix as https://github.com/hypothesis/checkmate/pull/449/commits/bdcc4e7920be0b81a28faacc1bbae1dab4295e92

When it's time to remove it, it will be the same across projects.


seanh:
Is this necessary? `make sure` seems to be passing for me without any warnings

jon-betts (approved):

jon-betts:
Try wiping out your tox directories. You probably have an older and still compatible version of `distutils` or whatever is the underlying culprit.

seanh:
I tried that and didn't get any warnings :shrug: I think this is harmless to merge anyway

seanh (approved):

PR #3574, opened 2022-01-19
Validate params on BB sync API

marcospri:
A more focused version of https://github.com/hypothesis/lms/pull/3514 with only the param validation.

Left on that PR are:

- A refactor or the sync view. Fetch most params on the main view function and pass them along to helper functions.

- Validate also the params the backed sends to the frontend so they are in sync in both directions.

for which I made a note to revisit once the rest of the BB sprint issues are closed.

# Testing 

No new feature or behavior changes. 

Open a BB group assignment as both a teacher and student.


seanh (approved):
Looks good

PR #3579, opened 2022-01-20
Upsert as function

marcospri:
Trying to see how https://github.com/hypothesis/lms/pull/3518#pullrequestreview-856735220 will look.

If we merged something similar to this the next step will be some changes to the bulk_upsert function to be usable with groupings.

## Testing

- `truncate file` on make sql
- Configure canvas and blackboard file assignments.
- `select * from file` 
File will be present with two different application_instance_id.


lms/services/blackboard_api/client.py
```diff
@@ -63,24 +64,29 @@ def list_files(self, course_id, folder_id=None):
...
-                        "name": file["name"],
-                        "size": file["size"],
-                        "parent_lms_id": file["parentId"],
-                    }
-                    for file in results
-                ],
-            )
+        application_instance = self._request.find_service(
+            name="application_instance"
+        ).get_current()
```
marcospri, line 69: Not sure if this is right. should the service be injected on the constructor here?

lms/services/upsert.py
```diff
@@ -0,0 +1,94 @@
...
+
+from copy import deepcopy
+from typing import List
+
+from sqlalchemy import inspect
+from sqlalchemy.dialects.postgresql import insert
+from zope.sqlalchemy import mark_changed
+
+
+def bulk_upsert(  # pylint:disable=too-many-arguments
```
marcospri, line 11: No changes from the existing code here except the necessary to take the parameters from the function ones instead of the model config.

tests/unit/conftest.py
```diff
@@ -177,13 +177,15 @@ def httpretty_():
...
-def application_instance(pyramid_request):
-    return factories.ApplicationInstance(
+def application_instance(db_session, pyramid_request):
+    ai = factories.ApplicationInstance(
         consumer_key=pyramid_request.lti_user.oauth_consumer_key,
         developer_key="TEST_DEVELOPER_KEY",
         provisioning=True,
         settings=ApplicationSettings({}),
     )
+    db_session.flush()
```
marcospri, line 187: Needed so ai has a non null id.

lms/services/canvas_api/client.py
```diff
@@ -241,23 +242,28 @@ def list_files(self, course_id, sort="position"):
...
+                    "type": "canvas_file",
+                    "course_id": course_id,
+                    "lms_id": file["id"],
+                    "name": file["display_name"],
+                    "size": file["size"],
+                }
+                for file in files
+            ],
+            ["application_instance_id", "lms_id", "type", "course_id"],
+            ["name", "size"],
```
seanh, line 264: I wonder if we need a `FileService.upsert()` method that both `BlackboardAPIClient` and `CanvasAPIClient` can call? Might remove some duplication here

lms/services/blackboard_api/client.py
seanh: I guess we usually inject these into the constructors, yes. Though I don't think there's a huge difference.

If we had a `FileService.upsert()` for `BlackboardAPIClient` to call (as suggested in another comment) then I don't think `BlackboardAPIClient` would need the `application_instance` here and it looks like `FileService` already has the `application_instance` injected

seanh (changes requested):
This looks good to me: it looks like it removes a couple of hundred lines of code, and it also makes things conceptually lighter by removing the custom SQLAlchemy session, the `BulkAction.Config` dataclass, the use of Pyramid events, etc.

I think it might be simpler and remove some duplication if there was a `FileService.upsert()` that both `CanvasAPIClient` and `BlackboardAPIClient` called.

It's a shame that the `bulk_upsert()` function has `# pylint:disable=too-many-arguments`. I wonder if wrapping it in a `BulkUpsertService` and having the service factory inject the `db` argument would allow the PyLint disable to be removed?

But I also wonder whether the `BlackboardAPIClient` and `CanvasAPIClient` tests should **not** mock the `bulk_upsert()` function and should actually test that the `File`'s got upserted into the DB. And this might conflict with wrapping `bulk_upsert()` in a service.

If `FileService.upsert()` is added as suggested in the comments then I think the tests for `FileService` could **not** mock `bulk_upsert()` but the tests for `BlackboardAPIClient` and `CanvasAPIClient` could probably just use a mock `FileService`.

tests/unit/conftest.py
seanh: Is this really needed? The tests seem to pass without it. I wonder if it might slow things down by adding an extra `flush()` to every test that uses this fixture

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -130,33 +128,34 @@ def test_it_doesnt_send_paginated_requests_forever(
...
-                ],
-            )
+        file_service.upsert.assert_called_once_with(
+            [
+                Any.dict.containing(
+                    {
+                        "lms_id": id_ + 1,
+                        "type": "blackboard_file"
+                        if type_ == "File"
+                        else "blackboard_folder",
```
seanh, line 154: I wonder if the `type_` argument to the test could actually be `"blackboard_file"` or `"blackboard_folder"`, then you'd only need `"type": type_` here?

tests/unit/lms/services/file_test.py
```diff
@@ -24,6 +25,33 @@ def test_it_doesnt_return_matching_files_from_other_application_instances(
...
+                for i, file in enumerate(files)
+            ]
+        )
+
+        assert db_session.query(File).count() == total_files
+        # Refresh the model instances to gather changed data from the DB
+        _ = [db_session.refresh(file) for file in files]
+        for i, file in enumerate(files):
+            assert file.size == i
+            assert file.name == f"file_{i}"
```
seanh, line 53: Do we need a bit more testing here? Maybe some files that already exist (so they should get updated), some files that don't exist yet (should get created) (`factories.File.build_batch()` can create `models.File` objects without adding them to the DB), some files that already exist and already have the up-to-date values. It can probably still be all in one test

lms/services/canvas_api/client.py
```diff
@@ -50,6 +49,7 @@ def __init__(self, authenticated_client, request):
         """
         self._client = authenticated_client
         self._request = request
```
seanh, line 51: I think `self._request` and the `request` argument to `__init__()` are now unused and can be removed

seanh (changes requested):
Looks good to me. Some minor requests. I'll give it a test locally too

tests/unit/conftest.py
marcospri: Odd, I added it when the tests were setup differently and it was required. Not needed anymore :+1:

lms/services/canvas_api/client.py
marcospri: True, nice :+1:

tests/unit/lms/services/file_test.py
```diff
@@ -24,6 +26,58 @@ def test_it_doesnt_return_matching_files_from_other_application_instances(
...
+    def test_upsert(self, db_session, svc, application_instance):
+        existing_files_count = db_session.query(File).count()
+
+        # This files will be created on the DB
+        update_files = factories.File.create_batch(
+            5, application_instance=application_instance
+        )
+        db_session.flush()
+        # These won't exist on the DB
+        insert_files = factory_boy.build_batch(dict, 5, FACTORY_CLASS=factories.File)
```
marcospri, line 38: `File.build_batch` objects ended up in the DB.

seanh: Were they really ending up in the DB? `factories.File.build_batch()` definitely isn't supposed to add things to the DB, and when I tested it didn't seem to, this seems to work:

```diff
diff --git a/tests/unit/lms/services/file_test.py b/tests/unit/lms/services/file_test.py
index f020defe..bf73b2e5 100644
--- a/tests/unit/lms/services/file_test.py
+++ b/tests/unit/lms/services/file_test.py
@@ -35,7 +35,7 @@ class TestGet:
         )
         db_session.flush()
         # These won't exist on the DB
-        insert_files = factory_boy.build_batch(dict, 5, FACTORY_CLASS=factories.File)
+        insert_files = factories.File.build_batch(5)
         assert db_session.query(File).count() == existing_files_count + len(
             update_files
         )
@@ -53,9 +53,9 @@ class TestGet:
             ]
             + [
                 {
-                    "type": file["type"],
-                    "course_id": file["course_id"],
-                    "lms_id": file["lms_id"],
+                    "type": file.type,
+                    "course_id": file.course_id,
+                    "lms_id": file.lms_id,
                     "name": f"insert_file_{i}",
                     "size": i * 100,
                 }
@@ -73,8 +73,8 @@ class TestGet:
             assert file.size == i * 10
             assert file.name == f"update_file_{i}"
 
-        for i, file_dict in enumerate(insert_files):
-            file = db_session.query(File).filter_by(lms_id=file_dict["lms_id"]).one()
+        for i, file_ in enumerate(insert_files):
+            file = db_session.query(File).filter_by(lms_id=file_.lms_id).one()
             assert file.size == i * 100
             assert file.name == f"insert_file_{i}"
  
```

tests/unit/lms/services/file_test.py
```diff
@@ -24,6 +26,58 @@ def test_it_doesnt_return_matching_files_from_other_application_instances(
...
         assert not svc.get(file_.lms_id, file_.type)
 
+    def test_upsert(self, db_session, svc, application_instance):
+        existing_files_count = db_session.query(File).count()
+
+        # This files will be created on the DB
+        update_files = factories.File.create_batch(
+            5, application_instance=application_instance
+        )
+        db_session.flush()
```
seanh, line 36: I don't think the `flush()` is needed?

```suggestion
```

tests/unit/lms/services/file_test.py
```diff
@@ -24,6 +26,58 @@ def test_it_doesnt_return_matching_files_from_other_application_instances(
...
+                for i, file in enumerate(insert_files)
+            ],
+        )
+
+        assert db_session.query(File).count() == existing_files_count + len(
+            update_files
+        ) + len(insert_files)
+
+        # Refresh the model instances to gather changed data from the DB
+        _ = [db_session.refresh(file) for file in update_files]
```
seanh, line 71:
```suggestion
        for file in update_files:
            db_session.refresh(file)
```

seanh (approved):

tests/unit/lms/services/file_test.py
marcospri: > Were they really ending up in the DB? 

They were at some point, it took me a while to find the `factory_boy.build_batch(dict, 5, FACTORY_CLASS=factories.File)` workaround. But seems to work now? :shrug:

PR #3583, opened 2022-01-21
Expose cause on ExternalRequestError

marcospri:
Based on comments from https://github.com/hypothesis/lms/pull/3503.
Screenshot from 2022-01-25 10-06-25

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -82,6 +82,7 @@ class ExternalRequestError(Exception):
         return (
             f"{class_name}("
             f"message={self.message!r}, "
+            f"cause={self.__cause__!r}, "
             f"request={request}, "
             f"response={response}, "
             f"validation_errors={self.validation_errors!r})"
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -140,6 +140,7 @@ class APIExceptionViews:
         return self.error_response(
             message=message,
             details={
+                "cause": repr(self.context.__cause__),
                 "request": {
                     "method": self.context.method,
                     "url": strip_queryparams(self.context.url),
--- a/tests/unit/lms/views/api/exceptions_test.py
+++ b/tests/unit/lms/views/api/exceptions_test.py
@@ -62,6 +62,7 @@ class TestExternalRequestError:
         assert json_data == {
             "message": context.message,
             "details": {
+                "cause": "ValueError('foo')",
                 "request": {
                     "method": context.method,
                     "url": "https://example.com/",  # The URL without query string.
@@ -84,7 +85,7 @@ class TestExternalRequestError:
 
     @pytest.fixture
     def context(self):
-        return ExternalRequestError(
+        context = ExternalRequestError(
             message="test_message",
             request=requests.Request(
                 "GET", "https://example.com?foo=bar", data="request_body"
@@ -94,6 +95,8 @@ class TestExternalRequestError:
             ),
             validation_errors=sentinel.validation_errors,
         )
+        context.__cause__ = ValueError("foo")
+        return context
 
 
 class TestNotFound:
--- a/tests/unit/lms/services/exceptions_test.py
+++ b/tests/unit/lms/services/exceptions_test.py
@@ -36,14 +36,15 @@ class TestExternalRequestError:
         assert err.response_body is None
 
     @pytest.mark.parametrize(
-        "message,request_,response,validation_errors,expected",
+        "message,request_,response,validation_errors,cause,expected",
         [
             (
                 None,
                 None,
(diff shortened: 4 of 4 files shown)


lms/views/api/exceptions.py
```diff
@@ -140,6 +141,7 @@ def external_request_error(self):
         return self.error_response(
             message=message,
             details={
+                "cause": str(self.context.__cause__),
```
marcospri, line 144: @seanh Should this be `repr` instead?

lms/services/exceptions.py
```diff
@@ -82,6 +82,7 @@ def __repr__(self):
         return (
             f"{class_name}("
             f"message={self.message!r}, "
+            f"cause={self.__cause__}, "
```
seanh, line 85: Since this is `ExternalRequestError`'s `repr()` we should probably use `!r` to call `repr()` rather than `str()` on `self.__cause__`:

```suggestion
            f"cause={self.__cause__!r}, "
```

Then it'll print `cause=KeyError('foo')` instead of just `cause='foo'`, for example.

lms/views/api/exceptions.py
```diff
@@ -127,6 +127,7 @@ def external_request_error(self):
             },
         )
         sentry_sdk.set_context("validation_errors", self.context.validation_errors)
+        sentry_sdk.set_context("cause", self.context.__cause__)
```
seanh, line 130:
```suggestion
```

I tried this out in Sentry and it doesn't seem to work: you need to pass a dict to `set_context()` not an exception object, so it'd have to be something like  `sentry_sdk.set_context("cause", {"exc": repr(self.context.__cause__)})`. But I also noticed that Sentry seems to pull out the `__cause__` and the whole chain of exceptions automatically, so I actually don't think we need to do anything.

seanh: For testing Sentry by the way: get your own free Sentry account, create a project, get the project's DSN, then do `export SENTRY_DSN='https://e7c5d3f2958a4c4781b11782931a8f86@o31866.ingest.sentry.io/1309830'` before running `make dev`

lms/views/api/exceptions.py
seanh: I think it should probably be `repr`, yes. The result depends on the individual exception class (and what it has for its `__str__()` and `__repr__()` methods) but in general `__repr__()` is the one that's supposed to contain more info so that's probably what we want for debugging. If you compare `str()` and `repr()` for common standard library exception classes `str()` will usually return just `'foo'` whereas `repr()` will give `ValueError('foo')`. I think we probably want to know the exception class, otherwise we may not be able to tell what went wrong

seanh (changes requested):

lms/views/api/exceptions.py
marcospri: Yes I think sentry get's the cause of the exception automatically:

https://sentry.io/organizations/hypothesis/issues/2959750441/?project=5952630&query=is%3Aunresolved

tests/unit/lms/views/api/exceptions_test.py
```diff
@@ -62,6 +62,7 @@ def test_it(self, context, pyramid_request, views, report_exception, sentry_sdk)
         assert json_data == {
             "message": context.message,
             "details": {
+                "cause": "None",
```
seanh, line 65: Could we add a `__cause__` to the test exception so that we test the exception view's cause rendering?

```diff
diff --git a/tests/unit/lms/views/api/exceptions_test.py b/tests/unit/lms/views/api/exceptions_test.py
index 64def955..3063b96f 100644
--- a/tests/unit/lms/views/api/exceptions_test.py
+++ b/tests/unit/lms/views/api/exceptions_test.py
@@ -62,7 +62,7 @@ class TestExternalRequestError:
         assert json_data == {
             "message": context.message,
             "details": {
-                "cause": "None",
+                "cause": "ValueError('foo')",
                 "request": {
                     "method": context.method,
                     "url": "https://example.com/",  # The URL without query string.
@@ -85,7 +85,7 @@ class TestExternalRequestError:
 
     @pytest.fixture
     def context(self):
-        return ExternalRequestError(
+        context = ExternalRequestError(
             message="test_message",
             request=requests.Request(
                 "GET", "https://example.com?foo=bar", data="request_body"
@@ -95,6 +95,8 @@ class TestExternalRequestError:
             ),
             validation_errors=sentinel.validation_errors,
         )
+        context.__cause__ = ValueError("foo")
+        return context
 
 
 class TestNotFound:
```

seanh (changes requested):
Looks good, one small tweak to the tests

seanh (approved):

PR #3586, opened 2022-01-24
Don't convert groups assignments to sections in sync API

marcospri:
This PR could be replaced with the much simpler: https://github.com/hypothesis/lms/pull/3586

---

f355d8d060b1f824886bfe73f73c0ec1239793c7 had a couple of problems:

- Left the same AI setting check in `lms/views/api/canvas/sync.py`.
- The scopes used are still based on the AI setting. (But enabling and disabling seems to work until the token is deleted).

There's some duplication between `lms/views/api/canvas/sync.py` and `lms/resources/lti_launch.py`

Not sure where we could put some canvas shared code functions. Directly on `CanvasService`? There's only  `public_url_for_file` there at the moment.

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -63,6 +63,38 @@ class CanvasService:
 
             return url
 
+    @staticmethod
+    def is_speedgrader(request) -> bool:
+        if "learner_canvas_user_id" in request.GET:
+            # Location while launching an assignment
+            return bool(request.GET["learner_canvas_user_id"])
+
+        # Location when using the Sync API
+        try:
+            return "learner" in request.json
+        except:  # pylint:disable=bare-except
+            return False
+
+    @staticmethod
+    def group_set(request) -> int:
+        if "group_set" in request.params:
+            # Location while launching an assignment
+            value = request.params["group_set"]
+        else:
+            # Location when using the Sync API
+            value = request.json["learner"].get("group_set")
+
+        return int(value)
+
+    @staticmethod
+    def is_group_launch(request) -> bool:
+        try:
+            CanvasService.group_set(request)
+        except (KeyError, ValueError, TypeError):
+            return False
+        else:
+            return True
+
 
 class CanvasFileFinder:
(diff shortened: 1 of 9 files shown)


lms/services/canvas.py
```diff
@@ -1,14 +1,47 @@
...
+
+
+def is_canvas_group_launch(request):
+    try:
+        group_set(request)
+    except (KeyError, ValueError, TypeError):
+        return False
+    else:
+        return True
+
```
marcospri, line 34: I having trouble finding the right place for these.

- Adding them to the existing `CanvasService` means that we'd need `CanvasService` in `resource/lti_resource` and JSConfig. That also brings `canvas_api_client` which needs a applications insntace (via `get_current`). 

This causes some functests to fail as just using the resources requires a application instance which was not the case before.

- Using a different service class. I think that will mean we have CanvasService, CanvasAPIClient + a new one. It can work but not sure about the naming for that one. `CanvasHelpers`, `CanvasRequest`, meh.

- Using plain function as in this diff. Works as well. It needs a bit more importing in general and individual patching in the tests.

lms/services/canvas.py
```diff
@@ -63,6 +63,38 @@ def public_url_for_file(
 
             return url
 
+    @staticmethod
```
marcospri, line 66: I moved this all around the place now and while I don't think static methods in the service is the absolute perfect solution I think is better than some of the alternatives:

- Common functions somewhere in `resources/` and new resource for the sync api. 

This will work but we'll need a new class + methods for every new  caller. I think this approach will add to much friction to reuse.

- Helper module `helpers/canvas.py` 

Helper functions/modules could work. LTIResource is very much a helper class already. But I think adding the helper concept now in the code base has the problem that it's not clear why some canvas functions might be in the helper module, some in LTIResource and some in any of the canvas related services.



As `staticmethods` in the service is not a surprising location. Having them as `staticmethods` allows to use them directly in the service or as standalone functions when you don't want to pull the whole service.

PR #3587, opened 2022-01-24
VitalSource assignments in blackboard

marcospri:
This was done before as part of the "canvas assignment in the DB" effort but that had to be rebased.

There's a couple of assumptions in that work that don't apply anymore:

- Canvas assignments are stored in the DB.
- Non-canvas LMS use the new "assignment API" to create/configure assignments.

Now those are not true so a few changes are needed but if we were to go in that direction again this new changes won't get in the way.

The summary of the changes is that the back-end will provide a new `url` field while configuring the assignment and both client and server will check for `vitalsource://` on document_url when they need specific VS behavior.

### On Canvas

- Configure a VS assignment on `master` (using the book ID `9781319367077` for example)

- An URL like `http://localhost:8001/lti_launches?vitalsource_book=true&book_id=9781319367077&cfi=%2F6%2F20%5B%3Bvnd.vst.idref%3Dpart01%5D%21%2F4%2F2%5BPT01_sec_1%5D` would be called on LMS

- Change to this branch, launch the same assignment. Same `http://localhost:8001/lti_launches?vitalsource_book=true&book_id=9781319367077&cfi=%2F6%2F20%5B%3Bvnd.vst.idref%3Dpart01%5D%21%2F4%2F2%5BPT01_sec_1%5D` URL will be used.

- Reconfigure the assignment and launch it. The launch URL will like: `http://localhost:8001/lti_launches?url=vitalsource%3A%2F%2Fbook%2FbookID%2F9781319367077%2Fcfi%2F%2F6%2F14%5B%3Bvnd.vst.idref%3Dfm6%5D`

### On Blackboard

- Enable VS on http://localhost:8001/admin/instance/Hypothesis14af0fe87c9deb2e461f88be4a8d5364/

- Configure a VS assignment and launch it.

This doesn't inject the client but I think this is expected in local: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1637231685257800?thread_ts=1635347477.018300&cid=C4K6M7P5E (same thing happens on Canvas).

### DIFF
--- a/lms/services/vitalsource/_schemas.py
+++ b/lms/services/vitalsource/_schemas.py
@@ -3,7 +3,7 @@ Schemas for VitalSource API responses.
 
 See: https://developer.vitalsource.com/hc/en-us/categories/360001974433
 """
-from marshmallow import EXCLUDE, Schema, fields
+from marshmallow import EXCLUDE, Schema, fields, post_load
 
 from lms.validation._base import RequestsResponseSchema
 
@@ -32,5 +32,15 @@ class BookTOCSchema(RequestsResponseSchema):
 
         url = fields.Str(required=False)
         """vitalsource:// like url identifying the book and chapter"""
+
+        @post_load
+        def set_url(self, in_data, **_kwargs):
+            # pylint:disable=import-outside-toplevel
+            from lms.services.vitalsource.client import VitalSourceService
+
+            in_data["url"] = VitalSourceService.generate_document_url(
+                self.context["book_id"], in_data["cfi"]
+            )
+            return in_data
 
     table_of_contents = fields.List(fields.Nested(Chapter), required=True)
--- a/lms/services/vitalsource/_schemas.py
+++ b/lms/services/vitalsource/_schemas.py
@@ -30,4 +30,7 @@ class BookTOCSchema(RequestsResponseSchema):
         cfi = fields.Str(required=True)
         page = fields.Str(required=True)
 
+        url = fields.Str(required=False)
+        """vitalsource:// like url identifying the book and chapter"""
+
     table_of_contents = fields.List(fields.Nested(Chapter), required=True)
--- a/lms/services/vitalsource/client.py
+++ b/lms/services/vitalsource/client.py
@@ -60,11 +60,9 @@ class VitalSourceService:
 
             raise
 
-        toc = BookTOCSchema(response).parse()
-        for chapter in toc["table_of_contents"]:
-            chapter["url"] = self.generate_document_url(book_id, chapter["cfi"])
-
-        return toc
+        schema = BookTOCSchema(response)
+        schema.context["book_id"] = book_id
+        return schema.parse()
 
     @staticmethod
     def parse_document_url(document_url):
--- a/lms/services/vitalsource/client.py
+++ b/lms/services/vitalsource/client.py
@@ -60,11 +60,19 @@ class VitalSourceService:
 
             raise
 
-        return BookTOCSchema(response).parse()
+        toc = BookTOCSchema(response).parse()
+        for chapter in toc["table_of_contents"]:
+            chapter["url"] = self.generate_document_url(book_id, chapter["cfi"])
+
+        return toc
 
     @staticmethod
     def parse_document_url(document_url):
         return DOCUMENT_URL_REGEX.search(document_url).groupdict()
+
+    @staticmethod
+    def generate_document_url(book_id, cfi):
+        return f"vitalsource://book/bookID/{book_id}/cfi/{cfi}"
 
     def get_launch_params(self, document_url, lti_user):
         """
--- a/lms/services/vitalsource/client.py
+++ b/lms/services/vitalsource/client.py
@@ -1,8 +1,16 @@
+import re
+
 import oauthlib
 from oauthlib.oauth1 import SIGNATURE_HMAC_SHA1, SIGNATURE_TYPE_BODY
 
 from lms.services.exceptions import ExternalRequestError
 from lms.services.vitalsource._schemas import BookInfoSchema, BookTOCSchema
+
+#: A regex for parsing the BOOK_ID and CFI parts out of one of our custom
+#: vitalsource://book/bookID/BOOK_ID/cfi/CFI URLs.
+DOCUMENT_URL_REGEX = re.compile(
+    r"vitalsource:\/\/book\/bookID\/(?P<book_id>[^\/]*)\/cfi\/(?P<cfi>.*)"
+)
 
 
 class VitalSourceService:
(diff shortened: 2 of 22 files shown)


lms/services/vitalsource/client.py
```diff
@@ -52,9 +60,21 @@ def book_toc(self, book_id: str):
 
             raise
 
-        return BookTOCSchema(response).parse()
+        toc = BookTOCSchema(response).parse()
+        for chapter in toc["table_of_contents"]:
+            chapter["url"] = self.generate_document_url(book_id, chapter["cfi"])
```
marcospri, line 65: Here we inject the URL directly into the dict. Not sure if this is the more marshmellow-y way of doing this. 

The URL can't be generated from the data of `BookTOCSchema` itself. It needs its parent `book_id`.

lms/views/basic_lti_launch.py
```diff
@@ -91,6 +93,23 @@ def store_lti_data(self):
                 request, h_user=lti_user.h_user, lti_user=lti_user
             )
 
+    @view_config(request_param="vitalsource_book=true", configured=False)
```
marcospri, line 96: This is necessary only if we need backwards compatibility with existing assignments.

As this is a "legacy view" that could be eventually removed I think having the `request_param` on the view_config is much more clear and independent that the existing predicate.

lms/services/vitalsource/client.py
seanh: I think it's probably fine. If you want all the work to be done in the schema perhaps it can be `toc = BookTOCSchema(response).parse(book_id)`. Then the schema would have all the info it needs to inject the chapter URLs itself

lms/views/basic_lti_launch.py
seanh: Let's question whether we need that backwards compatibility with existing VitalSource assignments. I suspect we don't (it's not being used by real users yet) so the code can be simpler

seanh (approved):

lms/services/vitalsource/_schemas.py
```diff
@@ -30,4 +30,17 @@ class Meta:
         cfi = fields.Str(required=True)
         page = fields.Str(required=True)
 
+        url = fields.Str(required=False)
+        """vitalsource:// like url identifying the book and chapter"""
+
+        @post_load
+        def set_url(self, in_data, **_kwargs):
+            # pylint:disable=import-outside-toplevel
```
marcospri, line 38: Avoiding circular import here.

I do think `generate_document_url` belongs to the service so this annotations is probably better that moving the function to some other namespace.

seanh (approved):

robertknight:
> This doesn't inject the client but I think this is expected in local: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1637231685257800?thread_ts=1635347477.018300&cid=C4K6M7P5E (same thing happens on Canvas).

The VitalSource book viewer always tries to load the production Hypothesis client, which is currently unable to talk to the local LMS app. We might be able to remove that restriction, which would make testing more convenient.

lms/services/vitalsource/client.py
```diff
@@ -52,9 +60,19 @@ def book_toc(self, book_id: str):
...
+        schema.context["book_id"] = book_id
+        return schema.parse()
+
+    @staticmethod
+    def parse_document_url(document_url):
+        return DOCUMENT_URL_REGEX.search(document_url).groupdict()
+
+    @staticmethod
+    def generate_document_url(book_id, cfi):
+        return f"vitalsource://book/bookID/{book_id}/cfi/{cfi}"
```
robertknight, line 73: The CFI may contain characters like `;` and `[`. For example: `vitalsource://book/bookID/9781400847402/cfi//6/28[;vnd.vst.idref=ch5]`.

I wonder if we should be URI-encoding the whole CFI. I'm just reading up on the relevant specs to figure out what we "should" be doing.

lms/views/basic_lti_launch.py
```diff
@@ -91,6 +92,23 @@ def store_lti_data(self):
...
 
+    @view_config(vitalsource_book=True)
+    def legacy_vitalsource_lti_launch(self):
+        """
+        Respond to a legacy configured VitalSource assignment.
+
+        Legacy VitalSource assignments use `vitalsource_book=true` as opposed to
+        a `vitalsource://` URL as the document_url.
+
+        The assignment shouldn't be "configured" in any other way to match this view."
```
robertknight, line 103:
```suggestion
        The assignment shouldn't be "configured" in any other way to match this view.
```

tests/unit/lms/services/vitalsource/_schemas_test.py
```diff
@@ -0,0 +1,50 @@
...
+                    "page": "2",
+                    "url": "vitalsource://book/bookID/BOOK_ID/cfi//cif_02",
+                },
+            ]
+        }
+
+
+@pytest.fixture
+def list_chapters_response():
+    """Return the JSON body of a valid Blackboard Files API response."""
```
robertknight, line 36: Presumably this should say "VitalSource Book TOC" response instead of "Blackboard Files API"?

robertknight (approved):
The frontend changes look good overall. I added a few minor corrections and comments. I noted in Slack that we might want to look at standardizing the parsing/generating of custom URL schemes to avoid inconsistencies creeping in, now that we have at least three such schemes.

lms/static/scripts/frontend_apps/components/test/ContentSelector-test.js
```diff
@@ -511,13 +511,14 @@ describe('ContentSelector', () => {
 
     const picker = wrapper.find('BookPicker');
     interact(wrapper, () => {
-      picker.props().onSelectBook({ id: 'test-book' }, { cfi: '/1/2' });
+      picker
+        .props()
+        .onSelectBook({ id: 'test-book' }, { url: 'vitalsource://' });
```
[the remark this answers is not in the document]
marcospri: Yes, it was a bit lazy only using the schema there.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -52,6 +52,11 @@ function formatContentURL(content) {
   if (content.url.includes('blackboard:')) {
     return 'PDF file in Blackboard';
   }
+  // All VitalSource file URLs start with the literal string `vitalsource://`
+  if (content.url.includes('vitalsource://')) {
+    return 'Book from VitalSource';
```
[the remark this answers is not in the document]
marcospri: I think that will need a bigger refactor now as the URL type only includes the URL itself but we could probably include some common fields there for canvas/blackboard/vitalsource.

lms/services/vitalsource/client.py
marcospri: We are URL encoding it when it's going to be on an URL for example on SpeedGrader (well, we have to go around a Canvas bug there).

I think if we urlencoded it on the API we might double-encode it ourselves unless we do a bigger refactor.

We can revisit this of course if it turns out we should encode it.

PR #3596, opened 2022-01-26
Use bulk_upsert for gruping memberships

marcospri:
## Testing

- `truncate grouping_membership` on `make sql`
- Launch a groups assignment in blackboard.
- `select * from grouping_membership;` should return rows.
- Launch the assignment again.
- `select * from grouping_membership;` created` should be different than `updated`.

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -5,6 +5,7 @@ from sqlalchemy.orm import aliased
 
 from lms.models import Course, Grouping, GroupingMembership, User
 from lms.models._hashed_id import hashed_id
+from lms.services.upsert import bulk_upsert
 
 
 class GroupingService:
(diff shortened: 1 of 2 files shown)


tests/unit/lms/services/grouping_test.py
```diff
@@ -205,9 +205,10 @@ def test_get_course_grouping_for_user_with_different_user(
         assert not groupings
 
     @pytest.fixture
-    def with_course_memberships(self, svc, application_instance):
+    def with_course_memberships(self, svc, db_session, application_instance):
         courses = factories.Course.create_batch(5)
         user = factories.User(application_instance=application_instance)
+        db_session.flush()
```
marcospri, line 211: This seems to be a staple of my PR now.

I'd say this worked before because adding the course and user to the session (via  `GroupingMembership(grouping=group, user=user)` populated the IDs. Now using the core API we need the IDs beforehand.

lms/services/grouping.py
```diff
@@ -86,16 +87,21 @@ def upsert_grouping_memberships(self, user: User, groups: List[Grouping]):
...
-            group.memberships.append(GroupingMembership(grouping=group, user=user))
+
+        bulk_upsert(
+            self._db,
+            GroupingMembership,
+            [
+                {
+                    "grouping_id": group.id,
+                    "user_id": user.id,
+                    "updated": func.now(),
```
marcospri, line 98: As this is the only column being updated we explicitly pass the column and value.

seanh (approved):

PR #3598, opened 2022-01-26
Add unique index for grouping.authority_provided_id

marcospri:
The index part is not strictly necessary but will make upserts (with bulk_upsert) more efficient. 

I think the uniqueness should have been there form the start.


marcospri:
authority_provided_id could be duplicated across application instances.

Issue #3500, opened 2021-12-08
Bulk upsert of groupings and grouping memberships


chrisshaw:
(possibly important when group sets have many groups)

PR #3599, opened 2022-01-26
Bulk upsert for groupings

marcospri:
Last one for https://github.com/hypothesis/lms/issues/3500

Using bulk_upsert for groupings. Needs to return the ORM objects in this one.

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -33,52 +33,52 @@ class GroupingService:
             tool_consumer_instance_guid, parent.lms_id, type_.value, lms_id
         )
 
-    def upsert_with_parent(  # pylint: disable=too-many-arguments
+    def upsert_with_parent(
         self,
-        tool_consumer_instance_guid,
-        lms_id,
-        lms_name,
-        parent: Grouping,
+        grouping_dicts: List[dict],
         type_: Grouping.Type,
-        extra=None,
+        parent: Grouping,
     ):
         """
         Upsert a Grouping generating the authority_provided_id based on its parent.
 
-        :param tool_consumer_instance_guid: Tool consumer GUID
-        :param lms_id: ID of this grouping on the LMS
-        :param lms_name: Name of the grouping on the LMS
-        :param parent: Parent of grouping
-        :param type_: Type of the grouping
-        :param extra: Any extra information to store linked to this grouping
+        :param grouping_dicts: A list of dicts containing:
+            lms_id: ID of this grouping on the LMS
+            lms_name: Name of the grouping on the LMS
+            extra: Any extra information to store linked to this grouping
+
+        :param parent: Parent grouping for all upserted groups
+        :param type_: Type of the groupings
         """
-        authority_provided_id = self.generate_authority_provided_id(
-            tool_consumer_instance_guid, lms_id, parent, type_
+
+        if not parent.id:
+            # Make sure we have a PK for the parent before upserting
+            self._db.flush()
+
+        return bulk_upsert(
+            self._db,
+            Grouping,
+            [
+                {
+                    "application_instance_id": self.application_instance.id,
+                    "authority_provided_id": self.generate_authority_provided_id(
+                        self.application_instance.tool_consumer_instance_guid,
+                        grouping["lms_id"],
+                        parent,
+                        type_,
+                    ),
+                    "lms_id": grouping["lms_id"],
+                    "parent_id": parent.id,
+                    "type": type_,
+                    "lms_name": grouping["lms_name"],
+                    "extra": grouping.get("extra"),
+                    "updated": func.now(),
+                }
+                for grouping in grouping_dicts
+            ],
+            index_elements=["application_instance_id", "authority_provided_id"],
+            update_columns=["lms_name", "extra", "updated"],
         )
-
-        grouping = (
-            self._db.query(Grouping)
-            .filter_by(
-                application_instance=self.application_instance,
-                authority_provided_id=authority_provided_id,
-            )
-            .one_or_none()
-        )
-
-        if not grouping:
-            grouping = Grouping(
-                application_instance=self.application_instance,
-                authority_provided_id=authority_provided_id,
-                lms_id=lms_id,
-                parent_id=parent.id,
-                type=type_,
-            )
-            self._db.add(grouping)
-
-        grouping.lms_name = lms_name
-        grouping.extra = extra
-
-        return grouping
 
     def upsert_grouping_memberships(self, user: User, groups: List[Grouping]):
         """
--- a/lms/services/upsert.py
+++ b/lms/services/upsert.py
@@ -2,6 +2,7 @@
 
 from typing import List
 
+from sqlalchemy import column, tuple_
 from sqlalchemy.dialects.postgresql import insert
 from zope.sqlalchemy import mark_changed
 
@@ -21,6 +22,7 @@ def bulk_upsert(
     :param values: Dicts of values to upsert
     :param index_elements: Columns to match when upserting. This must match an index.
     :param update_columns: Columns to update when a match is found.
+    :return: A lazy query of the affected `model_class` rows.
     """
     if not values:
         # Don't attempt to upsert an empty list of values into the DB.
@@ -40,13 +42,15 @@ def bulk_upsert(
         # cause a "null value violates not-null constraint" crash.
         return []
 
+    index_elements_columns = [column(c) for c in index_elements]
+
     stmt = insert(model_class).values(values)
     stmt = stmt.on_conflict_do_update(
         # The columns to use to find matching rows.
         index_elements=index_elements,
         # The columns to update.
         set_={element: getattr(stmt.excluded, element) for element in update_columns},
-    )
+    ).returning(*index_elements_columns)
 
     result = db.execute(stmt)
 
@@ -55,4 +59,6 @@ def bulk_upsert(
     # back
     mark_changed(db)
 
-    return result
+    return db.query(model_class).filter(
+        tuple_(*index_elements_columns).in_(result.all())
+    )
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -3,9 +3,8 @@ from unittest.mock import sentinel
 
 import pytest
 from h_matchers import Any
-from sqlalchemy.exc import IntegrityError
 
-from lms.models import CanvasGroup, Grouping, GroupingMembership
+from lms.models import Grouping, GroupingMembership
 from lms.services.grouping import GroupingService, factory
 from tests import factories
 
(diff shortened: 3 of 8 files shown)


lms/services/upsert.py
```diff
@@ -2,16 +2,18 @@
 
 from typing import List
 
+from sqlalchemy import column, tuple_
 from sqlalchemy.dialects.postgresql import insert
 from zope.sqlalchemy import mark_changed
 
 
-def bulk_upsert(
+def bulk_upsert(  # pylint:disable=too-many-arguments,
```
marcospri, line 10: Instead of a parameter here we could a second `bulk_upsert_returning` function that uses the existing one.

lms/services/grouping.py
```diff
@@ -33,53 +33,43 @@ def generate_authority_provided_id(
...
-        lms_name,
-        parent: Grouping,
-        type_: Grouping.Type,
-        extra=None,
-    ):
+    def upsert_with_parent(self, grouping_dicts: List[dict]):
         """
         Upsert a Grouping generating the authority_provided_id based on its parent.
 
-        :param tool_consumer_instance_guid: Tool consumer GUID
```
marcospri, line 48: No longer taking `tool_consumer_instance_guid`. Using the one from `self.application_instance`.

tests/unit/lms/services/grouping_test.py
```diff
@@ -84,14 +84,20 @@ def test_if_no_grouping_already_exists_it_inserts_a_new_one(
         assert db_session.query(CanvasGroup).one() == test_grouping
 
     def test_if_a_grouping_already_exists_it_updates_it(
-        self, db_session, upsert_with_parent
+        self, db_session, svc, upsert_with_parent
     ):
+
+        parent = factories.Course(application_instance=svc.application_instance)
```
marcospri, line 90: Both need the same parent. 

I'd say this only worked before because both `parent.id`'s was None without a flush.

tests/unit/lms/services/grouping_test.py
```diff
@@ -166,15 +171,20 @@ def test_you_can_have_a_group_and_a_section_with_the_same_id(
...
-    def upsert_with_parent(self, svc):
-        return partial(
-            svc.upsert_with_parent,
-            tool_consumer_instance_guid="t_c_i_guid",
-            lms_id="lms_id",
-            lms_name="lms_name",
-            parent=factories.Course(application_instance=svc.application_instance),
-            type_=Grouping.Type.CANVAS_GROUP,
-        )
+    def upsert_with_parent(self, svc, db_session):
```
marcospri, line 174: Maybe the fixutre itself should be named `upsert_with_parent_single_item`?

lms/services/grouping.py
```diff
@@ -33,53 +33,43 @@ def generate_authority_provided_id(
...
-        :param lms_name: Name of the grouping on the LMS
-        :param parent: Parent of grouping
-        :param type_: Type of the grouping
-        :param extra: Any extra information to store linked to this grouping
+        :param grouping_dicts: A list of dicts containing:
+            lms_id: ID of this grouping on the LMS
+            lms_name: Name of the grouping on the LMS
+            parent: Parent of grouping
+            type: Type of the grouping
+            extra: Any extra information to store linked to this grouping
```
seanh, line 45: So this now upserts a list of groupings at once, not just one

lms/services/grouping.py
```diff
@@ -33,53 +33,43 @@ def generate_authority_provided_id(
...
+                    "type": grouping["type"],
+                    "lms_name": grouping["lms_name"],
+                    "extra": grouping.get("extra"),
+                    "updated": func.now(),
+                }
+                for grouping in grouping_dicts
+            ],
+            index_elements=["application_instance_id", "authority_provided_id"],
+            update_columns=["lms_name", "extra", "updated"],
+            returning=True,
```
seanh, line 70: :+1: Nice, this seems pretty simple

lms/services/grouping.py
seanh: I wonder about the consequences of this change

lms/services/upsert.py
```diff
@@ -55,4 +60,11 @@ def bulk_upsert(
...
 
-    return result
+    if not returning:
+        return result
+
+    return (
+        db.query(model_class)
+        .filter(tuple_(*index_elements_columns).in_(result.all()))
+        .all()
+    )
```
seanh, line 70: What's the difference between these two returns? Would it be possible to make it do `returning=True` all the time and avoid the extra parameter?

lms/services/grouping.py
marcospri: I think it should only have consequences if we where doing the wrong thing before, if `application_instance.tool_consumer_instance_guid != tool_consumer_instance_guid`.

Can't tell if that has happen in the past. I doubt it as you will have to come up with a tool_consumer_instance_guid from somewhere. But of course we've seen lots of weird edge cases.

lms/services/upsert.py
marcospri: We don't use the current return value of this function (and in fact, this PR changes it even when returning=False)

My only problem with always returning the last part is that it will add and extra query + object serialization to for example the current `File.upsert`.

tests/unit/lms/services/grouping_test.py
seanh: I think we may need to re-design the tests a little now that `upsert_with_parent()` accepts a list of dicts:

* I think we need at least one test that passes it a list of more than one dict

* `upsert_with_parent()` can now insert some new groupings and update some existing groupings at the same time. I think we need a test that does that. We probably want to combine `test_if_no_grouping_already_exists_it_inserts_a_new_one()` and `test_if_a_grouping_already_exists_it_updates_it()` into a single test that both inserts a new grouping and updates an existing one in one call

* I'm not sure that the partial caller fixture for `upsert_with_parent()` makes much sense anymore. That fixture made sense when `upsert_with_parent()` took several separate arguments because the fixture was simple and it could inject defaults for all the arguments. But now that `upsert_with_parent()` takes a list of dicts I think the fixture might just be getting in the way

I had a go at trying to write a test for inserting one row and updating another at the same time. I can't for the life of me get this test to actually work. But here's where I got to:

```python
class TestUpsertWithParent:
    def test_it(self, db_session, grouping_dict, svc):
        course = factories.Course(application_instance=svc.application_instance)

        # Insert an existing grouping into the DB.
        existing_grouping = factories.CanvasSection(parent=course)

        # Upsert one grouping that already exists and one that doesn't exist yet.
        groupings = svc.upsert_with_parent(
            [
                {
                    "lms_id": existing_grouping.lms_id,
                    "lms_name": "updated_name",
                    "parent": existing_grouping.parent,
                    "type": existing_grouping.type,
                    "extra": {"updated": "extra"},
                },
                {
                    "lms_id": "new_grouping_id",
                    "lms_name": "new_grouping_name",
                    "parent": course,
                    "type": Grouping.Type.CANVAS_GROUP,
                    "extra": {"new": "extra"},
                },
            ]
        )

        # It should have updated the existing grouping.
        existing_grouping = db_session.query(Grouping).filter_by(lms_id=existing_grouping.lms_id).one()
        assert existing_grouping.lms_name == "updated_name"
        assert existing_grouping.extra == {"updated": "extra"}

        # It should have created the new grouping.
        new_grouping = db_session.query(Grouping).filter_by(lms_id="new_grouping_id").one()
        assert new_grouping.lms_name == "new_grouping_name"
        assert new_grouping.parent == course
        assert new_grouping.type == Grouping.Type.CANVAS_GROUP
        assert new_grouping.extra == {"updated": "extra"}
```

seanh (changes requested):

tests/unit/lms/services/grouping_test.py
seanh: Here's another attempt that avoids using the ORM to insert the `existing_grouping`. This gets past the constraint error but it seems that `upsert_with_parent()` doesn't actually update `existing_grouping`:

```python
    def test_it(self, db_session, grouping_dict, svc):
        course = factories.Course(application_instance=svc.application_instance)
        # Insert an existing grouping into the DB.
        svc.upsert_with_parent(
            [
                {
                    "lms_id": "existing_id",
                    "lms_name": "existing_name",
                    "parent": course,
                    "type": Grouping.Type.CANVAS_SECTION,
                    "extra": {"existing": "extra"},
                }
            ]
        )

        # Upsert one grouping that already exists and one that doesn't exist yet.
        groupings = svc.upsert_with_parent(
            [
                {
                    "lms_id": "existing_id",
                    "lms_name": "updated_name",
                    "parent": course,
                    "type": Grouping.Type.CANVAS_SECTION,
                    "extra": {"updated": "extra"},
                },
                {
                    "lms_id": "new_grouping_id",
                    "lms_name": "new_grouping_name",
                    "parent": course,
                    "type": Grouping.Type.CANVAS_GROUP,
                    "extra": {"new": "extra"},
                },
            ]
        )

        # It should have updated the existing grouping.
        existing_grouping = (
            db_session.query(Grouping).filter_by(lms_id="existing_id").one()
        )
        assert existing_grouping.lms_name == "updated_name"
        assert existing_grouping.extra == {"updated": "extra"}

        # It should have created the new grouping.
        new_grouping = (
            db_session.query(Grouping).filter_by(lms_id="new_grouping_id").one()
        )
        assert new_grouping.lms_name == "new_grouping_name"
        assert new_grouping.parent == course
        assert new_grouping.type == Grouping.Type.CANVAS_GROUP
        assert new_grouping.extra == {"updated": "extra"}

```

lms/services/upsert.py
marcospri: Changed to return the query in all cases (with out the call to `.all()`) :+1:

lms/services/grouping.py
```diff
@@ -33,52 +33,52 @@ def generate_authority_provided_id(
...
+    def upsert_with_parent(
         self,
-        tool_consumer_instance_guid,
-        lms_id,
-        lms_name,
-        parent: Grouping,
+        grouping_dicts: List[dict],
         type_: Grouping.Type,
-        extra=None,
+        parent: Optional[Grouping],
```
marcospri, line 40: Keep type_ and parent as top level params. It means they are common to all grouping_dicts but that behavior we support now. 

This makes the grouping_dicts thinner which I think is a nice side effect.

seanh: Is `parent` really optional? I think we always pass it

lms/services/grouping.py
```diff
@@ -33,52 +33,52 @@ def generate_authority_provided_id(
...
-                type=type_,
-            )
-            self._db.add(grouping)
+        :param parent: Parent grouping for all upserted groups
+        :param type_: Type of the groupings
+        """
 
-        grouping.lms_name = lms_name
-        grouping.extra = extra
+        if parent and not parent.id:
```
seanh, line 54: If `parent` is made non-optional this can be simpler:
```suggestion
        if not parent.id:
```

lms/services/grouping.py
```diff
@@ -33,52 +33,52 @@ def generate_authority_provided_id(
...
+                {
+                    "application_instance_id": self.application_instance.id,
+                    "authority_provided_id": self.generate_authority_provided_id(
+                        self.application_instance.tool_consumer_instance_guid,
+                        grouping["lms_id"],
+                        parent,
+                        type_,
+                    ),
+                    "lms_id": grouping["lms_id"],
+                    "parent_id": parent.id if parent else None,
```
seanh, line 71: Could also be simplified is `parent` was non-optional:

```suggestion
                    "parent_id": parent.id,
```

lms/services/upsert.py
```diff
@@ -16,6 +17,8 @@ def bulk_upsert(
     """
     Create or update the specified values in a table.
 
+    Returns a lazy query of the affected `model_class` rows.
```
seanh, line 20: Might as well but this in a `:return:` at the bottom of the docstring, since we're using `:param:`

seanh (approved):
I'm wondering about whether the `parent` argument to `upsert_with_parent()` is really optional or not?

If it's not optional then the code can be slightly simplified in a few places.

If it _is_ optional then I think `upsert_with_parent()` should be renamed to `upsert()`

marcospri:
> If it's not optional then the code can be slightly simplified in a few places.

Changed to being not optional. The grouping that doesn't have a parent, Course, has it's own service so this one is likely to be always used with a parent.

PR #3601, opened 2022-01-26
Remove autodetecting of columns defined in the ORM on bulk_upsert

marcospri:
Remove some complexity to try to auto detect columns defined with "onupdate" on the ORM model.

This might not even be complex enough as SQLA has lots of edge cases for every feature.

Doing this only for some hardcoded condition (the name of the column is "updated" or that the class uses `CreatedUpdatedMixin) might achieve the same thing now but I think it will hide the fact that it's not doing it in general.

Just removing all of this and remembering to pass all the columns that need updating explicitly it's cleaner.

### DIFF
--- a/lms/services/file.py
+++ b/lms/services/file.py
@@ -1,3 +1,5 @@
+from sqlalchemy import func
+
 from lms.models import File
 from lms.services.upsert import bulk_upsert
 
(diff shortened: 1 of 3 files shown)


seanh (approved):
:+1: I think this is hard to argue with: we seem to have used 72 lines (including tests) of very tricky code for something that we could have done with 6 very straightforward lines.

This does require each caller of `bulk_upsert()` to pass in `func.now()` for the `updated` column if they want it to be updated and there will be some duplication there once we have multiple callers of `bulk_upsert()`. But I think the duplication is probably worthwhile for the straightforwardness that it gains us

PR #3605, opened 2022-01-26
Don't crash if `tool_consumer_instance_guid` changes

seanh:
Fixes https://github.com/hypothesis/lms/issues/3604

If an LMS changes its `tool_consumer_instance_guid` (or if a `consumer_key` gets installed in a second LMS with a different `tool_consumer_instance_guid`) we're supposed to refuse to launch and show an error dialog: https://github.com/hypothesis/lms/pull/3053

But since we added a new DB constraint (https://github.com/hypothesis/lms/pull/3494) we're now crashing with an `IntegrityError` instead.

The reason is that the changed `tool_consumer_instance_guid` causes us to generate a different `h_userid` and then when looking for the user in the DB we unnecessarily include the `h_userid` in the query. Fix it by removing the `h_userid` from the query so that we do find the existing user with the same `user_id` and `application_instance` even though the `h_userid` is different.

Slack threads:

1. https://hypothes-is.slack.com/archives/C1MA4E9B9/p1643209011047900
2. https://hypothes-is.slack.com/archives/C01RAT0PE72/p1643214772002300
3. https://hypothes-is.slack.com/archives/C2BLQDKHA/p1643220371904379

### DIFF
--- a/lms/services/user.py
+++ b/lms/services/user.py
@@ -67,7 +67,6 @@ class UserService:
             .filter_by(
                 application_instance=model_user.application_instance,
                 user_id=model_user.user_id,
-                h_userid=model_user.h_userid,
             )
             .one_or_none()
         )


lms/services/user.py
```diff
@@ -67,7 +67,6 @@ def _find_existing_user(self, model_user: User) -> Optional[User]:
             .filter_by(
                 application_instance=model_user.application_instance,
                 user_id=model_user.user_id,
-                h_userid=model_user.h_userid,
```
seanh, line 70: I don't see any reason why we need to include `h_userid` in this query: just `(application_instance, user_id)` is unique.

The query originates from https://github.com/hypothesis/lms/pull/3282 but there's no discussion of why `h_userid` was included.

Note that before https://github.com/hypothesis/lms/pull/3494 (merged earlier today) there was no unique constraint on `(application_instance, user_id)`, only on `(application_instance, user_id, h_userid)`. So `(application_instance, user_id)` wasn't guaranteed to be unique. In practice though `(application_instance, user_id)` was actually unique apart from one duplicate row that somehow got into the DB (see <https://hypothes-is.slack.com/archives/C1MA4E9B9/p1643209011047900>).

Also, [`UserService.get()`](https://github.com/hypothesis/lms/blob/c2979ab714bd229e6454adada093a4d58c4a02af/lms/services/user.py#L41-L62) assumes that `(application_instance, user_id)` is unique and will crash otherwise.

seanh: I think we could do some refactoring here because `get()` and `_find_existing_user()` are almost the same thing and don't both need to exist. Maybe for a follow-up PR though, this one fixes the bug

jon-betts (approved):
Yep. Seeing as this is now guaranteed unique, I can't see why you'd need this.

PR #3610, opened 2022-01-27
Raise if a course grouping has a parent

seanh:
`generate_authority_provided_id()` raises if you try to generate an `authority_provided_id` for a non-course grouping with no parent. For good measure it should also raise if you try to generate one for a _course_ grouping _with_ a parent: if `type_ == Grouping.Type.COURSE` then the `parent` argument would not be used to it makes no sense to call `generate_authority_provided_id()` in this way.

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -21,10 +21,10 @@ class GroupingService:
         type_: Grouping.Type,
     ):
         if type_ == Grouping.Type.COURSE:
+            assert parent is None, "Course groupings can't have a parent"
             return hashed_id(tool_consumer_instance_guid, lms_id)
 
-        # For the rest of types, parent is mandatory
-        assert parent is not None
+        assert parent is not None, "Non-course groupings must have a parent"
 
         if type_ == Grouping.Type.CANVAS_SECTION:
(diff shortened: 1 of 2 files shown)


lms/services/grouping.py
```diff
@@ -21,10 +21,10 @@ def generate_authority_provided_id(
         type_: Grouping.Type,
     ):
         if type_ == Grouping.Type.COURSE:
+            assert parent is None, "Course groupings can't have a parent"
```
seanh, line 24: I added strings to the exceptions. This means that you see ``AssertionError: Course groupings can't have a parent` in the traceback instead of just `AssertionError` and the tests can use `match` to make sure they're catching the `AssertionError` that they expected and not another one.

lms/services/grouping.py
```diff
@@ -21,10 +21,10 @@ def generate_authority_provided_id(
...
             return hashed_id(tool_consumer_instance_guid, lms_id)
 
-        # For the rest of types, parent is mandatory
-        assert parent is not None
+        assert parent is not None, "Non-course groupings must have a parent"
```
seanh, line 27: The assertion string is better than a comment

PR #3611, opened 2022-01-27
Canvas grouping memberships

marcospri:
We should be storing this information at least to have the same data for blackboard and canvas.

I don't think we can start using that information right away to avoid a query to the API when speed grading but we could after this builds up the table for a while.

### DIFF
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -2,7 +2,7 @@ from pyramid.view import view_config
 
 from lms.models import Grouping
 from lms.security import Permissions
-from lms.services import CanvasAPIError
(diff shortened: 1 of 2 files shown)


lms/views/api/canvas/sync.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, request):
     )
     def sync(self):
         if self._is_group_launch:
-            groups = self._to_groups_groupings(self._get_canvas_groups())
+            groups = self._get_canvas_groups()
```
marcospri, line 28: `_get_canvas_groups` Now returns Groupings directly.

seanh (approved):

PR #3618, opened 2022-01-27
Tidy `upsert_with_parent()` tests

seanh:
Mostly by using a partial fixture to reduce noise and duplication.

### DIFF
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -73,42 +73,22 @@ class TestGenerateAuthorityProvidedID:
 
 
 class TestUpsertWithParent:
-    CONTEXT_ID = "context_id"
-    TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
-
-    def test_if_no_grouping_already_exists_it_inserts_a_new_one(self, svc, db_session):
-        course = factories.Course(application_instance=svc.application_instance)
-
-        # Start with no CanvasGroup
-        assert not db_session.query(CanvasGroup).count()
-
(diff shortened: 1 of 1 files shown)


tests/unit/lms/services/grouping_test.py
```diff
@@ -73,42 +73,22 @@ def generate_authority_provided_id(self, svc):
 
 
 class TestUpsertWithParent:
-    CONTEXT_ID = "context_id"
-    TOOL_CONSUMER_INSTANCE_GUID = "t_c_i_guid"
```
seanh, line 77: It's nice to be rid of these (they moved into the `partial` below): I think these kinds of `CONSTANTS` on test classes are ugly and distracting

tests/unit/lms/services/grouping_test.py
```diff
@@ -73,42 +73,22 @@ def generate_authority_provided_id(self, svc):
...
-            tool_consumer_instance_guid=course.application_instance.tool_consumer_instance_guid,
-            lms_id="lms_id",
-            lms_name="lms_name",
-            parent=course,
-            type_=Grouping.Type.CANVAS_GROUP,
-        )
+    def test_if_no_grouping_already_exists_it_inserts_a_new_one(
+        self, db_session, upsert_with_parent
+    ):
+        test_grouping = upsert_with_parent()
```
seanh, line 79: The `upsert_with_parent()` fixture is a partial that calls `svc.upsert_with_parent()` with default args so the tests only need to pass the args that matter for each test

marcospri (approved):

PR #3619, opened 2022-01-28
Add partial types for OneDrive file picker responses

robertknight:
We've seen some situations where certain properties may be missing or
arrays may be empty [1]. We haven't clarified the details yet, but as a step
towards using types to enforce checking of potentially missing fields,
add some partial types for OneDrive picker responses.

The OneDrive picker is loaded from a remote source and doesn't have its
own types. Microsoft do publish a package,
@microsoft/microsoft-graph-types, for the Microsoft Graph API, but it is
huge, which will slow down tsc, and also doesn't provide good
information on the nullability of properties. Everything is simply listed as
nullable. Therefore I think it may be sensible for us to define our own
partial types based on what we use and observed behavior of the API.

[1] https://hypothes-is.slack.com/archives/C2BLQDKHA/p1643309746775569

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
+++ b/lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
@@ -2,10 +2,48 @@ import { PickerCanceledError } from '../errors';
 import { loadOneDriveAPI } from './onedrive-api-client';
 
 /**
+ * Partial type for OneDrive share links.
+ *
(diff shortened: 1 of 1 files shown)


robertknight:
Hi @esanzgar - I'm not sure how this will fit in with the changes you are making to improve OneDrive error handling. Perhaps it would make sense for you to take this over?

esanzgar (approved):
Super!

I was starting to address the nullability of certain properties, so this is useful. Thanks!

PR #3633, opened 2022-02-01
upsert_grouping_memberships flushes when necessary

marcospri:
For newly created users or groupings the ORM objects will be present in the session but might not be yet flushed and their `.id` fields will be `None` causing upsert to fail.

Fixes https://sentry.io/organizations/hypothesis/issues/2978797202/?project=259908&referrer=slack

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -87,6 +87,9 @@ class GroupingService:
         :param user:  User the that belongs to the groups
         :param groups: List of groups the `user` belongs to
         """
+        if not user.id or any((group.id is None for group in groups)):
+            # Ensure all ORM objects have their PK populated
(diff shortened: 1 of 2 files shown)


lms/services/grouping.py
```diff
@@ -87,6 +87,9 @@ def upsert_grouping_memberships(self, user: User, groups: List[Grouping]):
         :param user:  User the that belongs to the groups
         :param groups: List of groups the `user` belongs to
         """
+        if not user.id or any((group.id is None for group in groups)):
+            # Unsure all ORM objects have their PK populated
```
seanh, line 91:
```suggestion
            # Ensure all ORM objects have their PK populated.
```

seanh (approved):

PR #3641, opened 2022-02-02
Bulk upsert returning

marcospri:
For https://sentry.io/organizations/hypothesis/issues/2982856751/?referrer=slack

This involves a hacky workaround in bulk_upsert to always return the same type.

Also handled the same case directly in Grouping.upsert_with_parent.

I don't think the change in Grouping.upsert_with_parent is enough as it would leave this surprising behaviour in bulk_upsert.

An alternative would be to always return a list in `bulk_upsert` but that will cause a query and serializing the results in File.upsert but it's definitely less hacky.

# Testing 
- Switch to master
- Apply a diff similar to:

```diff --git a/lms/views/api/canvas/sync.py b/lms/views/api/canvas/sync.py
index 80af13c1..580e915c 100644
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -83,12 +83,12 @@ class Sync:
 
         if self._is_speedgrader:
             # SpeedGrader requests are made by the teacher, get the student we are grading
-            return self._to_groups_groupings(
-                self._canvas_api.user_groups(
-                    self._request.json["course"]["custom_canvas_course_id"],
-                    int(self._request.json["learner"]["canvas_user_id"]),
-                    self.group_set(),
-                )
+            return self._to_groups_groupings([]
+                #self._canvas_api.user_groups(
+                #    self._request.json["course"]["custom_canvas_course_id"],
+                #    int(self._request.json["learner"]["canvas_user_id"]),
+                #    self.group_set(),
+                #)
             )
 
         try:
```

- Speedgrade any canvas assignment. You get a crash like the sentry one.
- Keeping the diff switch to this branch. No crash. No groups either, this is a edge case on speedgrader.
```

### DIFF
--- a/lms/services/grouping.py
+++ b/lms/services/grouping.py
@@ -50,6 +50,8 @@ class GroupingService:
         :param parent: Parent grouping for all upserted groups
         :param type_: Type of the groupings
         """
+        if not grouping_dicts:
+            return []
 
         if not parent.id:
             # Make sure we have a PK for the parent before upserting
--- a/lms/services/upsert.py
+++ b/lms/services/upsert.py
@@ -40,7 +40,10 @@ def bulk_upsert(
         # default values for all of the columns. If my_table has a column
         # with a NOT NULLABLE constraint and no default value this will
         # cause a "null value violates not-null constraint" crash.
-        return []
+        #
+        # We do a wasteful query here to maintain
+        # the same return type in all branches.
+        return db.query(model_class).filter(False)
 
     index_elements_columns = [column(c) for c in index_elements]
 
--- a/tests/unit/lms/services/grouping_test.py
+++ b/tests/unit/lms/services/grouping_test.py
@@ -75,6 +75,15 @@ class TestGenerateAuthorityProvidedID:
 
 
 class TestUpsertWithParent:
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: lms/services/upsert.py [hunk 1]
```diff
@@ -40,7 +40,10 @@ def bulk_upsert(
         # default values for all of the columns. If my_table has a column
         # with a NOT NULLABLE constraint and no default value this will
         # cause a "null value violates not-null constraint" crash.
-        return []
+        #
+        # We do a wasteful query here to maintain
+        # the same return type in all branches.
+        return db.query(model_class).filter(False)
```
marcospri, line 46: This is a hacky workaround to have the same value. There's no empty-query set or similar concept in SQLA that I could find.

lms/services/grouping.py
```diff
@@ -50,6 +50,8 @@ def upsert_with_parent(
         :param parent: Parent grouping for all upserted groups
         :param type_: Type of the groupings
         """
+        if not grouping_dicts:
```
marcospri, line 53: Not really necessary but we already know about this case so we might as well handle it.

seanh (approved):

PR #3642, opened 2022-02-03
Each ExternalException type gets its own __repr__ method

marcospri:
No changes made to the method, just  the same method in both classes for now.

This will make addressing https://github.com/hypothesis/lms/pull/3503#discussion_r796851988 easier than adding conditionals here.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -58,7 +58,35 @@ class ExternalRequestError(Exception):
         return getattr(self.response, "text", None)
 
     def __repr__(self) -> str:
-        return _repr_external_request_exception(self)
+        # Include the details of the request and response for debugging. This
+        # appears in the logs and in tools like Sentry and Papertrail.
+        request = (
+            "Request("
+            f"method={self.method!r}, "
+            f"url={self.url!r}, "
+            f"body={self.request_body!r}"
+            ")"
+        )
+
+        response = (
+            "Response("
+            f"status_code={self.status_code!r}, "
+            f"reason={self.reason!r}, "
+            f"body={self.response_body!r}"
+            ")"
+        )
+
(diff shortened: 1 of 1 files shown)


seanh (changes requested):
Hmm, I think we want to avoid having two very similar versions of the `__repr__()` method to maintain because that method is actually quite non-trivial to write and test correctly.

I think we may want `ExternalRequestError` and `ExternalAsyncRequestError` instances to implement the same public interface so that `_repr_external_request_exception()` and the exception view can treat them the same, rather than either having to duplicate those methods or complicate them with conditionals.

Some suggestions:

* Just leave the `message`, `request` and `validation_errors` arguments in `ExternalAsyncRequestError.__init__()`. Possibly a little odd since those arguments are never actually used when calling `ExternalAsyncRequestError()`? On the other hand it is consistent with `ExternalRequestError`.

* Remove the `__init__()` arguments but keep the attributes for compatibility:

  ```python
  class ExternalAsyncRequestError(Exception):
          super().__init__()

          self.response = response

          # For compatibility with ExternalRequestError.
          self.message = None
          self.request = None
          self.validation_errors = None
  ```

* Remove the `__init__()` arguments but replace them with `@property`'s for compatibility:

  ```python
  class ExternalAsyncRequestError(Exception):
          super().__init__()
          self.response = response

      ...

      @property
      def message(self):
          """For compatibility with ExternalRequestError."""
          return None

      @property
      def request(self):
          """For compatibility with ExternalRequestError."""
          return None

      @property
      def validation_errors(self):
          """For compatibility with ExternalRequestError."""
          return None
  ```

marcospri:
Fair enough, this is not necessary then. Closing this one and adapting the next one in the PR chain.

PR #3643, opened 2022-02-03
Remove unused attributes from ExternalAsyncRequestError

marcospri:
# Testing 

Nothing really changed but as a sanity check:

- Open a BB group assignment as a student with these diffs:

- Requests without response:

   ```diff
   diff --git a/lms/services/async_oauth_http.py b/lms/services/async_oauth_http.py
   index d00a1152..3e5dd9fe 100644
   --- a/lms/services/async_oauth_http.py
   +++ b/lms/services/async_oauth_http.py
   @@ -55,7 +55,7 @@ async def _prepare_requests(method, urls, **kwargs):
                    _async_request(
                        session,
                        method,
   -                    url,
   +                    "https://notreallyadomainthatworks.com/",
                        **kwargs,
                    )
                )
   ```

   Error dialog:

   
Screenshot from 2022-02-03 13-52-01

   Sentry event: 

   https://sentry.io/organizations/hypothesis/issues/2985295889/?project=5952630&query=is%3Aunresolved

- Failed request with a response:

   ```diff
   diff --git a/lms/services/async_oauth_http.py b/lms/services/async_oauth_http.py
   index d00a1152..a042cd30 100644
   --- a/lms/services/async_oauth_http.py
   +++ b/lms/services/async_oauth_http.py
   @@ -55,7 +55,7 @@ async def _prepare_requests(method, urls, **kwargs):
                    _async_request(
                        session,
                        method,
   -                    url,
   +                    "http://httpbin.org/status/418",
                        **kwargs,
                    )
                )
   ```

   Error dialog:
   
Screenshot from 2022-02-03 13-57-33   

   

   Sentry event:  https://sentry.io/organizations/hypothesis/issues/2985309778/events/c5d593cbe0544891975434e6e027f16d/?project=5952630&query=is%3Aunresolved

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -65,14 +65,14 @@ class ExternalRequestError(Exception):
 
 
 class ExternalAsyncRequestError(Exception):
-    def __init__(
-        self, message=None, request=None, response=None, validation_errors=None
-    ):
+    def __init__(self, response=None):
         super().__init__()
-        self.message = message
-        self.request = request
         self.response = response
-        self.validation_errors = validation_errors
+
+        # For compatibility with ExternalRequestError.
+        self.message = None
+        self.request = None
+        self.validation_errors = None
 
     @property
     def _request_info(self):
--- a/tests/unit/lms/services/exceptions_test.py
+++ b/tests/unit/lms/services/exceptions_test.py
@@ -141,30 +141,24 @@ class TestExternalAsyncRequestError:
         assert err.status_code == expected
 
     @pytest.mark.parametrize(
-        "message,request_,response,cause,expected",
+        "response,cause,expected",
         [
             (
-                None,
-                None,
                 None,
                 None,
                 "ExternalAsyncRequestError(message=None, cause=None, request=Request(method=None, url=None, body=None), response=Response(status_code=None, reason=None, body=None), validation_errors=None)",
             ),
             (
-                "Some message",
-                None,
                 Mock(
                     status=400, reason="OK", request_info=Mock(method="POST", url="URL")
                 ),
                 KeyError("cause"),
-                "ExternalAsyncRequestError(message='Some message', cause=KeyError('cause'), request=Request(method='POST', url='URL', body=None), response=Response(status_code=400, reason='OK', body=None), validation_errors=None)",
+                "ExternalAsyncRequestError(message=None, cause=KeyError('cause'), request=Request(method='POST', url='URL', body=None), response=Response(status_code=400, reason='OK', body=None), validation_errors=None)",
             ),
         ],
     )
-    def test_str(self, message, request_, response, cause, expected):
+    def test_str(self, response, cause, expected):
         err = ExternalAsyncRequestError(
-            message=message,
-            request=request_,
             response=response,
         )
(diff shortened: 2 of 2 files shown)


tests/unit/lms/services/exceptions_test.py
```diff
@@ -141,32 +141,26 @@ def test_status_code(self, response, expected):
...
-                None,
                 Mock(
-                    status=400, reason="OK", request_info=Mock(method="POST", url="URL")
+                    status=400,
+                    reason="OK",
+                    request_info=Mock(method="POST", url="URL"),
                 ),
                 KeyError("cause"),
-                "ExternalAsyncRequestError(message='Some message', cause=KeyError('cause'), request=Request(method='POST', url='URL', body=None), response=Response(status_code=400, reason='OK', body=None), validation_errors=None)",
+                "ExternalAsyncRequestError(cause=KeyError('cause'), request=Request(method='POST', url='URL', body=None), response=Response(status_code=400, reason='OK', body=None))",
```
marcospri, line 158: This is the async version of the test. Removed the deleted attributes.

The sync version code/test didn't change.

lms/views/api/exceptions.py
```diff
@@ -128,7 +128,9 @@ def external_request_error(self):
                 "body": self.context.response_body,
             },
         )
-        sentry_sdk.set_context("validation_errors", self.context.validation_errors)
+
+        validation_errors = getattr(self.context, "validation_errors", None)
```
marcospri, line 132: Need some conditions here as now 
ExternalRequestError and ExternalAsyncRequestError differ. No message or validation errors in the latter.

Here went for some conditions in the same method instead of two different ones. Could always split it the diverge more.

lms/services/exceptions.py
```diff
@@ -157,16 +152,11 @@ def __repr__(self) -> str:
             ")"
         )
 
-        # The name of this class or of a subclass if one inherits this method.
-        class_name = self.__class__.__name__
```
marcospri, line 161: Inline directly on the return block.

seanh (changes requested):
Mentioned in <https://github.com/hypothesis/lms/pull/3642>: I think we probably want instances of the two exception classes to have the same API so that the `__repr__()` method and the exception view can treat them the same.

This does mean that `validation_errors` and the other unused attributes will appear in error dialogs, Sentry etc whenever there's an `ExternalAsyncRequestError` even though `ExternalAsyncRequestError`'s never actually have any `validation_errors`. It'll always just be `validation_errors=None`. But that's probably fine

marcospri:
> Mentioned in #3642: I think we probably want instances of the two exception classes to have the same API so that the `__repr__()` method and the exception view can treat them the same.
> 
> This does mean that `validation_errors` and the other unused attributes will appear in error dialogs, Sentry etc whenever there's an `ExternalAsyncRequestError` even though `ExternalAsyncRequestError`'s never actually have any `validation_errors`. It'll always just be `validation_errors=None`. But that's probably fine

Adeapted this PR :+1: 

I think is still a good idea to remove them from __init__ even if keep the attributes for compatibility.

seanh (approved):

PR #3644, opened 2022-02-03
Expose ExternalAsyncRequestError.response_body

marcospri:
# Testing 

- Same as https://github.com/hypothesis/lms/pull/3643

- Requests without response, no changes, there's no body.
- Requests with response:

   Error dialog: Same error dialog as the response is not rendered here

Screenshot from 2022-02-03 14-08-24

   Sentry event: The event includes the response text.

   https://sentry.io/organizations/hypothesis/issues/2985309778/events/4b6d0dd6ef7c46bfb6671d9713a2a7b9/?project=5952630&query=is%3Aunresolved

### DIFF
--- a/lms/services/async_oauth_http.py
+++ b/lms/services/async_oauth_http.py
@@ -44,6 +44,10 @@ class AsyncOAuthHTTPService:
 
 async def _async_request(aio_session, method, url, **kwargs):
     async with aio_session.request(method, url, **kwargs) as response:
+        # Calling `.text()` here caches the result in `response` but is still behind a coroutine.
+        # We assign it to another response attribute for it to be
+        # available in a sync context without needing to start coroutine.
+        response.sync_text = await response.text()
         return response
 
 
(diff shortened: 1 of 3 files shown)


lms/services/async_oauth_http.py
```diff
@@ -44,6 +44,10 @@ def request(
 
 async def _async_request(aio_session, method, url, **kwargs):
     async with aio_session.request(method, url, **kwargs) as response:
+        # Calling `.text()` here caches the result in `response` but is still behind a coroutine.
+        # We assign it to another response attribute for it to be
+        # available in a sync context without needing to start coroutine.
+        response.sync_text = await response.text()
```
marcospri, line 50: The alternative to this would be to to so something like 

https://github.com/hypothesis/lms/blob/38a08c221b0e40d669d3d4db10cb8e6c22667749/lms/services/async_oauth_http.py#L40

In the exception class. Having a corutine to get the text and run it in another thread with async.run.

I think the method ins this PR, get the body when we are already in a async context, might be useful if we ever needed the responses text in for example the BB service while having all the async code only in `async_oauth_service`.

For now it works well as we already `getattr(response...` in the exception for other properties.

seanh: :man_shrugging: Whatever works I guess :)

seanh (approved):

Issue #3640, opened 2022-02-02
Record `GroupingMembership`'s for all groupings


seanh:
Apparently [we only record `GroupingMembership`'s in the DB for Blackboard groups](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1643811404185479?thread_ts=1643808229.962529&cid=C1MA4E9B9). We should record these for courses, Canvas groups and Canvas sections as well.

marcospri:
Depending on https://github.com/hypothesis/private-issues/issues/65#issuecomment-1028506873 we have to decide if for instructors we want to store only strict membership (ie a teacher himself belongs to a section). 

We don't query this information at the moment. We query membership for students but we try to get all the relevant groups/sections the teachers can see.

For canvas groups for example it doesn't seem possible to add instructors to group sets.

marcospri:
We store memberships for:

- Students in canvas groups
- Students in canvas sections
- Students in blackboard groups

We don't currently store memberships for teachers in those and for anybody when there's no group/sections involved, ie for top level courses.

PR #3661, opened 2022-02-07
Store student group memberships for canvas sections

marcospri:
Part of https://github.com/hypothesis/lms/issues/3640

Store memberships of students/sections

# Testing

- Start with an empty table `truncate grouping_membership`

- Launch https://hypothesis.instructure.com/courses/125/assignments/873 **as a student**

- `select * from grouping join grouping_membership on grouping.id = grouping_membership.grouping_id;`

```
-[ RECORD 1 ]-----------+-----------------------------------------
created                 | 2022-01-27 10:49:53.646926
updated                 | 2022-02-07 12:00:35.725194
id                      | 116
application_instance_id | 8
authority_provided_id   | d2871e5e1c36b58f207d9ef5b22dcb7829f52e61
parent_id               | 115
lms_id                  | 144
lms_name                | Section 1
type                    | canvas_section
settings                | {}
extra                   | null
created                 | 2022-02-07 12:00:35.725194
updated                 | 2022-02-07 12:00:35.725194
grouping_id             | 116
user_id                 | 7
```

After this we are still missing membership information for the course group.

### DIFF
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from pyramid.view import view_config
 
 from lms.models import Grouping
@@ -27,7 +29,7 @@ class Sync:
         if self._is_group_launch:
             groups = self._get_canvas_groups()
         else:
-            groups = self._to_section_groupings(self._get_sections())
+            groups = self._get_sections()
 
         self._sync_to_h(groups)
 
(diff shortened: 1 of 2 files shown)


lms/views/api/canvas/sync.py
```diff
@@ -166,6 +173,12 @@ def _get_course(self):
             self._course_service.generate_authority_provided_id(tool_guid, context_id)
         )
 
+    def _get_user(self, user_id):
+        return self._request.find_service(UserService).get(
+            self._request.find_service(name="application_instance").get_current(),
+            user_id,
+        )
```
seanh, line 180: I wonder if this could be:

```python
@property
def user(self):
    ...
```

with no `user_id` argument, since the same value is always passed.

Was also wondering whether we should have `request.user` like we have `request.lti_user`?

seanh (approved):

lms/views/api/canvas/sync.py
marcospri: > Was also wondering whether we should have request.user like we have request.lti_user

Yes I think this is the right call. I have a half baked branch locally where lti_user has a user field. I'll clean that up and push it.

PR #3665, opened 2022-02-08
Fix some test file names

seanh:
Our naming convention is `foo_test.py` not `test_foo.py` because it plays better with autocompletion.


jon-betts (approved):
:+1: These always catch me out. If I start out in a directory like this I end up spreading the poison elsewhere as my brain gets scrambled.

PR #3671, opened 2022-02-11
WIP course.application_instance_id

marcospri:
# Relate `course` and `application_instances` using application_instances primary key

course relates to `application_instances` using `consumer_key` instead of application_instances's primary key.

In lti1.1 consumer_key doesn't exist anymore.

I'm sure we could find approaches to have a fake consumer_key value in the table if we need it for backwards compatibility but we should fix these relationships regardless.

The same problem exists in `group_info` and `oauth2_token`.

In this PR the migration does the change in one go and the model and code change are in the same commit.

A mergable/deployable version of this needs to do it in small steps to avoid downtime, probably something like:

- New nullable `application_id_column` (migration)
- Fill `application_id_column` on creation (code change)
- Fill the existing rows and make the column not nullable (migration)
- Change the code to query by application_instance_id instead of consumer_key, not using .get(). (code change)
- Swap the primary keys on course.


lms/services/course.py
```diff
@@ -75,7 +75,7 @@ def get(self, authority_provided_id):
 
     def _create_legacy(self, authority_provided_id):
         course = LegacyCourse(
-            consumer_key=self._application_instance.consumer_key,
```
marcospri, line 78: This could stay and will'll just populate the value for LTI1.1 if needed for backwards compatibility from outside LMS (analytics? metabase?).

PR #3676, opened 2022-02-14
Improve OneDrive permission error

esanzgar:
Currently, if MS SharePoint is setup in such a way that doesn't allow to
create public links, the OneDrive error message caused is not very
helpful.

We have created a new error type, `PickerPermissionError`, that is throw
when such permission restrictions are encountered. The new error message
contains more information about the error and a link to additional
resources.

### Before

image

### After

error-message-one-drive

Closes https://github.com/hypothesis/product-backlog/issues/1245

### DIFF
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -16,6 +16,8 @@ import FilePickerFormFields from './FilePickerFormFields';
 import GroupConfigSelector from './GroupConfigSelector';
 
 /**
+ * @typedef {import("preact").ComponentChildren} Children
+ *
  * @typedef {import('../api-types').File} File
  * @typedef {import('../utils/content-item').Content} Content
  * @typedef {import('../utils/content-item').URLContent} URLContent
@@ -31,8 +33,9 @@ import GroupConfigSelector from './GroupConfigSelector';
 
 /**
  * @typedef ErrorInfo
- * @prop {string} message
+ * @prop {string} description
  * @prop {Error} error
+ * @prop {Children} [children]
  */
 
 /**
@@ -195,11 +198,7 @@ export default function FilePickerApp({ onSubmit }) {
       </form>
       {shouldSubmit && <FullScreenSpinner />}
       {errorInfo && (
-        <ErrorModal
-          description={errorInfo.message}
-          error={errorInfo.error}
-          onCancel={() => setErrorInfo(null)}
-        />
+        <ErrorModal {...errorInfo} onCancel={() => setErrorInfo(null)} />
       )}
     </main>
   );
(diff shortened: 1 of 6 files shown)


lms/static/scripts/frontend_apps/errors.js
```diff
@@ -46,6 +46,15 @@ export class PickerCanceledError extends Error {
   }
 }
 
+/**
+ * Error thrown when the user cancels file selection.
+ */
+export class PickerPermissionError extends Error {
+  constructor() {
+    super('Unable to share publicly the file');
```
robertknight, line 54: This message needs at least a grammatical correction, although I would be inclined to reword it to be slightly more generic. I suggest something like: "Unable to update file permissions".

A detailed explanation of _why_ the file permissions need to be updated and what changes are made can be left to the Knowledge Base article.

esanzgar: I have updated the wording in the comment and in the error.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
```diff
@@ -71,16 +75,20 @@ export class OneDrivePickerClient {
...
-        //
-        // See https://hypothes-is.slack.com/archives/C2BLQDKHA/p1643309746775569.
-        const driveItem = result.value[0];
-        const name = driveItem.name;
-        const sharingURL = driveItem.permissions[0].link.webUrl;
-        const url = OneDrivePickerClient.downloadURL(sharingURL);
-        resolve({ name, url });
+        try {
+          // @ts-expect-error
+          const driveItem = result.value[0];
```
esanzgar, line 80: I only capture this very specific error. Other errors are assumed not to be permission errors and are caught and re-thrown on the below try/catch.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -31,8 +33,9 @@ import GroupConfigSelector from './GroupConfigSelector';
 
 /**
  * @typedef ErrorInfo
- * @prop {string} message
+ * @prop {string} description
```
esanzgar, line 36: This nomenclature change aligns better with the `ErrorModal` prop and simplifies the usage.

esanzgar:
@lyzadanger, unfortunately this is a little bit difficult to test in development.

In QA, one can test it in this way:

1. Go to the SharePoint admin center: https://4wfbj8-admin.sharepoint.com/_layouts/15/online/AdminHome.aspx#/sharing

2. Login as `eng-at-hypothes.is@4wfbj8.onmicrosoft.com` (credentials in 1Password)

3. Go to the Policies > Sharing

4. Lower the OneDrive permission slider one level down from the top (so the `Anyone, Users can share files and folders using links that don't require sign-in` is not selected).

The result of not having the necessary permission is that the respond object doesn't contain the property `value`, which is typically an array of one or more items.

The change in the type definition reflects this possibility and the added test checks this situation.

lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
esanzgar: There is a test to catch these other possible errors.

PR #3693, opened 2022-02-15
Return a dataclass from API exception views

seanh:
Return a dataclass from the API exception views.
    
This is to enable a future addition. Currently the exception views all call a `response_body()` helper method that returns a JSON-serializable dict:
    
```python
return self.error_response(
    message=...,
    details=...,
    ...
)
```

I now want to add refresh API information to some error responses but not others so I'd have to add a `refresh=True` argument to `error_response()` but that's going to trigger PyLint's "too-many-arguments" warning.
    
So this commit replaces the `error_response()` helper method with a JSON-serializable dataclass called `ErrorResponseBody`. Views can return instances of this dataclass directly and Pyramid's JSON renderer will serialize them to JSON for the response body:

```python
return ErrorResponseBody(
    message=...,
    details=...,
    ...
)
```
    
The `ErrorResponseBody.__json__()` method is basically the same method as the `error_response()` helper used to be but using `self` attrs instead of arguments.
    
A future commit can now add a new `refresh` attribute to `ErrorResponseBody` without triggering any PyLint warnings.

Slack thread: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1644937953064909

### DIFF
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -1,4 +1,5 @@
 """Error views for the API."""
+from dataclasses import dataclass
 from urllib.parse import urlparse, urlunparse
 
 import sentry_sdk
@@ -103,10 +104,15 @@ class APIExceptionViews:
         self.context = context
         self.request = request
 
+        # 400 is the default status for error responses.
+        # Some of the views below override this.
+        self.request.response.status_int = 400
+
     @exception_view_config(context=ValidationError)
     def validation_error(self):
-        return self.error_response(
-            422, message=self.context.explanation, details=self.context.messages
+        self.request.response.status_int = 422
+        return ErrorBody(
+            message=self.context.explanation, details=self.context.messages
         )
 
     @exception_view_config(context=ExternalRequestError)
@@ -139,7 +145,7 @@ class APIExceptionViews:
         # error dialog.
         message = self.context.message or "External request failed"
 
-        return self.error_response(
+        return ErrorBody(
             message=message,
             details={
                 "request": {
@@ -155,14 +161,13 @@ class APIExceptionViews:
         )
 
     @exception_view_config(context=OAuth2TokenError)
-    def oauth2_token_error(self):
-        return self.error_response()
+    def oauth2_token_error(self):  # pylint:disable=no-self-use
+        return ErrorBody()
 
     @exception_view_config(context=HTTPBadRequest)
     def http_bad_request(self):
-        return self.error_response(
-            status=self.context.code, message=self.context.detail
-        )
+        self.request.response.status_int = self.context.code
+        return ErrorBody(message=self.context.detail)
 
     @exception_view_config(
         # It's unfortunately necessary to mention CanvasAPIPermissionError
@@ -194,14 +199,14 @@ class APIExceptionViews:
         """
 
         if hasattr(self.context, "error_code"):
-            return self.error_response(
+            return ErrorBody(
                 error_code=self.context.error_code,
                 details=getattr(self.context, "details", None),
             )
 
         # Exception details are not reported here to avoid leaking internal information.
-        return self.error_response(
-            500,
+        self.request.response.status_int = 500
+        return ErrorBody(
             message=_(
                 "A problem occurred while handling this request. Hypothesis has been notified."
             ),
@@ -209,29 +214,43 @@ class APIExceptionViews:
 
     @forbidden_view_config()
     def forbidden(self):
-        return self.error_response(
-            403, message=_("You're not authorized to view this page.")
-        )
+        self.request.response.status_int = 403
+        return ErrorBody(message=_("You're not authorized to view this page."))
 
     @notfound_view_config()
     def notfound(self):
-        return self.error_response(404, message=_("Endpoint not found."))
+        self.request.response.status_int = 404
+        return ErrorBody(message=_("Endpoint not found."))
 
-    def error_response(self, status=400, error_code=None, message=None, details=None):
-        self.request.response.status_int = status
 
-        response = {}
+@dataclass
+class ErrorBody:
+    error_code: str = None
+    message: str = None
+    details: dict = None
 
-        if error_code is not None:
-            response["error_code"] = error_code
+    def __json__(self, _request):
+        """
+        Return a JSON-serializable representation of this error response body.
 
-        if message is not None:
-            response["message"] = message
+        If a view with renderer="json" returns an ErrorBody object then
+        Pyramid's JSON renderer will call this method to get an object to
+        JSON-serialize for the response body. See:
+        https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#using-a-custom-json-method
+        """
 
-        if details is not None:
-            response["details"] = details
+        body = {}
 
-        return response
+        if self.error_code is not None:
+            body["error_code"] = self.error_code
+
+        if self.message is not None:
+            body["message"] = self.message
+
+        if self.details is not None:
+            body["details"] = self.details
+
+        return body
 
 
 def strip_queryparams(url):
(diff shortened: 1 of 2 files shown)


lms/views/api/exceptions.py
```diff
@@ -103,10 +104,15 @@ def __init__(self, context, request):
         self.context = context
         self.request = request
 
+        # 400 is the default status for error responses.
+        # Some of the views below override this.
+        self.request.response.status_int = 400
```
seanh, line 109: The `error_response()` helper method used to set `self.request.response.status_int` _and_ return the error response body dict. So it was doing two things. It would be weird if the new `ErrorResponseBody` dataclass would set `self.request.response.status_int`: the class just represents the the response _body_, it has nothing to do with the status code. So I just moved status code setting into the views.

(I could write a class to represent the whole error response, not just the body. Then views could just return an instance. But it would be more work because Pyramid's JSON renderer represents a class that represents only the response body, and we get a lot for free by using the JSON renderer.)

lms/views/api/exceptions.py
```diff
@@ -103,10 +104,15 @@ def __init__(self, context, request):
...
+
     @exception_view_config(context=ValidationError)
     def validation_error(self):
-        return self.error_response(
-            422, message=self.context.explanation, details=self.context.messages
+        self.request.response.status_int = 422
```
seanh, line 113: If a view doesn't want the default status code of `400` set by the `__init__()` method above then it just overwrites it.

lms/views/api/exceptions.py
```diff
@@ -155,14 +161,13 @@ def external_request_error(self):
         )
 
     @exception_view_config(context=OAuth2TokenError)
-    def oauth2_token_error(self):
-        return self.error_response()
+    def oauth2_token_error(self):  # pylint:disable=no-self-use
```
seanh, line 164: The PyLint disable is annoying but its legit: the `oauth2_token_error()` view is in the view class because it inherits the class's `@view_defaults`, not because it uses `self`

tests/unit/lms/views/api/exceptions_test.py
```diff
@@ -138,23 +142,61 @@ class TestAPIError:
...
+                {
+                    "error_code": sentinel.error_code,
+                    "message": sentinel.message,
+                    "details": sentinel.details,
+                },
+            ),
+        ],
+    )
+    def test_json(self, pyramid_request, error_body, expected):
+        assert error_body.__json__(pyramid_request) == expected
```
jon-betts, line 199:
```suggestion
class TestErrorResponseBody:
    @pytest.mark.parametrize(
        "values",
        (
            {},
            {"error_code": sentinel.error_code},
            {"message": sentinel.message},
            {"details": sentinel.details},
            {
                "error_code": sentinel.error_code,
                "message": sentinel.message,
                "details": sentinel.details,
            }
        )
    )
    def test_json(self, pyramid_request, values):
        assert ErrorResponseBody(**values).__json__(pyramid_request) == values
```

I think we can use the fact the inputs and outputs are kind of the same to pack this down a bit.

I think you could probably get away without the individual cases as it's kind of boxed in on both sides by the first and last case too.

lms/views/api/exceptions.py
```diff
@@ -194,44 +199,58 @@ def api_error(self):
...
+        if self.message is not None:
+            body["message"] = self.message
 
-        if details is not None:
-            response["details"] = details
+        if self.details is not None:
+            body["details"] = self.details
 
-        return response
+        return body
```
jon-betts, line 253: Gitlab won't let me make a code suggested ('cos lines have been deleted)... but what about:

```python
return {key: value for key, value in asdict(self).items() if value is not None}
```

This is kind of nice because it's shorter, but it's also nice because it doesn't directly depend on the fields the `dataclass` has, so if they changed this would continue to work.

lms/views/api/exceptions.py
```diff
@@ -139,7 +145,7 @@ def external_request_error(self):
         # error dialog.
         message = self.context.message or "External request failed"
 
-        return self.error_response(
+        return ErrorResponseBody(
```
jon-betts, line 148: Is this maybe just an `ErrorBody`?

lms/views/api/exceptions.py
jon-betts: By the looks of it if you return a `Response` object it skips the `json` renderer entirely:

 * https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#writing-view-callables-which-use-a-renderer
 
```python
from pyramid.response import Response

@view_config(renderer='json')
def view(request):
    return Response('OK') # json renderer avoided
```

So unless we wanted to replicate the whole shebang, this seems like a real pain to do. Also... why this seems fine to me.

lms/views/api/exceptions.py
```diff
@@ -155,14 +161,13 @@ def external_request_error(self):
         )
 
     @exception_view_config(context=OAuth2TokenError)
-    def oauth2_token_error(self):
-        return self.error_response()
+    def oauth2_token_error(self):  # pylint:disable=no-self-use
+        return ErrorResponseBody()
```
jon-betts, line 165: Should the error have some kind of default for error code at least? It seems strange to return an entirely empty dict?

jon-betts (approved):
This all seems fine to me. I have some suggestions about making a few things shorter.

It would be a change of behavior, but it seems strange to not return anything at all in the dict. I was thinking you'd want a default code at least so you could visually see it was an error?

seanh:
> It would be a change of behavior, but it seems strange to not return anything at all in the dict. I was thinking you'd want a default code at least so you could visually see it was an error?

This is actually the API design:

* A `400` response with `{}` as the body tells the frontend that we don't have an access token for this user yet (or our access token has expired or been revoked and can't be refreshed). It causes the frontend to show the authorization dialog (not an error)
* A `400` with `{message: "..."}` or `{error_code: "..."}` tells the frontend to show an error dialog (the `error_code`'s refer to canned HTML error messages that the frontend has, whereas the `message` one is just a string to be rendered in a generic error dialog)

We could change this but then we'd have to update the frontend code as well

tests/unit/lms/views/api/exceptions_test.py
seanh: This is a nice idea but the very next PR that I have coming will mean that the inputs and outputs are no longer the same

lms/views/api/exceptions.py
seanh: Same issue: it's a nice suggestion for saving space but the next PR would break it. There's gonna be a dataclass field added that doesn't get translated directly into the output dict in the same way

lms/views/api/exceptions.py
seanh: Yeah, that's nicer

lms/views/api/exceptions.py
seanh: See other comment: <https://github.com/hypothesis/lms/pull/3693#issuecomment-1040654000>. The API design is quite brilliant, I'm sure you'll agree

lms/views/api/exceptions.py
seanh: Done

seanh:
> By the looks of it if you return a `Response` object it skips the `json` renderer entirely:

Yeah, that's how Pyramid works. So whenever a view is using a renderer it can always just to bypass that by returning a response directly.

I'm sure we could do some kind of `JSONResponse` class but I agree that working with the builtin renderer seems to be going with the flow here, and individual views doing `request.response.status_int = <whatever>` is also very Pyramid

lms/views/api/exceptions.py
seanh: Yeah, that's how Pyramid works. So whenever a view is using a renderer (even, say, the Jinja2 template renderer) it can always just bypass that by returning a response directly. For example you can return an `HTTPRedirect` (since all the HTTP exceptions are `Request` subclasses).

I'm sure we could do some kind of `JSONResponse` class but I agree that working with the builtin JSON renderer seems to be going with the flow here, and individual views doing `request.response.status_int = <whatever>` is also very Pyramid

PR #3694, opened 2022-02-16
Add refresh API info to error responses

seanh:
Add info to certain API error responses (but not others) instructing the frontend how to call the new refresh API before re-trying the original API request.

PR #3697, opened 2022-02-16
Add a separate `CanvasAPIInsufficientScopesError`

seanh:
So far this does exactly the same thing as `OAuth2TokenError` but in the future the two exception views will behave differently: `OAuth2TokenError`'s can happen because the user's access token has expired, which can potentially be fixed by refreshing the access token. But scopes errors can't be fixed by refreshing because the refreshed access token would have the same scopes as the original.

# Testing

1. Delete any access tokens from your DB:

   ```
   tox -qe dockercompose -- exec postgres psql -U postgres -c "DELETE FROM oauth2_token;"
   ```

2. Hack the code so that, when we get a new access token, we don't ask for all the scopes that we need:

   ```diff
   diff --git a/lms/views/api/canvas/authorize.py b/lms/views/api/canvas/authorize.py
   index 4290f5b0..53da83e8 100644
   --- a/lms/views/api/canvas/authorize.py
   +++ b/lms/views/api/canvas/authorize.py
   @@ -21,7 +21,6 @@ from lms.validation.authentication import OAuthCallbackSchema
    #: The Canvas API scopes that we need for our Canvas Files feature.
    FILES_SCOPES = (
        "url:GET|/api/v1/courses/:course_id/files",
   -    "url:GET|/api/v1/files/:id/public_url",
    )
   ```

3. Launch [localhost (make devdata) Canvas Files Assignment](https://hypothesis.instructure.com/courses/125/assignments/875) and click through the authorization dialog so that you get an access token with the missing scope.

   (Due to the bug created by the diff above you'll be in a re-authorization loop but that's fine. We're just trying to get an invalid access token into your DB.)

4. Undo the diff from step 2 and re-launch the assignment. A `CanvasAPIInsufficientScopesError` will be raised (you'll see it in the logs) and the authorization dialog (not an error dialog) will be shown. Click through the authorization process and you'll get a new access token _with_ the missing scope and the assignment will launch successfully

### DIFF
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -3,6 +3,7 @@ from lms.services.canvas import CanvasService
 from lms.services.exceptions import (
     BlackboardFileNotFoundInCourse,
     CanvasAPIError,
+    CanvasAPIInsufficientScopesError,
     CanvasAPIPermissionError,
     CanvasAPIServerError,
     CanvasFileNotFoundInCourse,
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -195,12 +195,16 @@ class CanvasAPIError(ExternalRequestError):
             status_code == 401
             and {"message": "Insufficient scopes on access token."} in errors
         ):
-            return OAuth2TokenError
+            return CanvasAPIInsufficientScopesError
 
         if status_code == 401:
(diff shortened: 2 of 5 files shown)


seanh:
I've merged this into https://github.com/hypothesis/lms/pull/3694

PR #3698, opened 2022-02-17
Take consumer_key directly from AI, not LTIUSer in JSConfig

marcospri:
A small refactor showing which I think should be the first of a series of PRs to remove the usage of `LTIUser.oauth_consumer_key` in favor of using it directly from the application instance.

Currently `get_current` itself gets the consumer_key value from LTIUSer. This is fine now but it will be a problem when `get_current` needs to use either LTI1.1 or LTI1.3 parameters.

Note that **get_current shouldn't get the value directly from the request params** as in some cases, when using BearerToken, the right value comes securely in the token, then added into the LTIUser.

Independently of the LTI update having consumer_key directly on LTIUser seems to leak one detail of application instance and "join" with application instance in the wrong field like we do in some DB tables.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -35,9 +35,6 @@ class JSConfig:
         return self._lti_user.h_user
 
     @property
-    def _consumer_key(self):
-        return self._lti_user.oauth_consumer_key
-
     def _application_instance(self):
         """
         Return the current request's ApplicationInstance.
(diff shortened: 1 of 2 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -35,9 +35,6 @@ def _h_user(self):
         return self._lti_user.h_user
 
     @property
-    def _consumer_key(self):
```
marcospri, line 38: Removed `_consumer_key`. 

Note that I carry `property` to `_application_instance` which I think it makes sense.

lms/resources/_js_config/__init__.py
```diff
@@ -537,8 +534,7 @@ def _sync_api(self):
             return self._canvas_sync_api()
 
         if (
-            self._application_instance().product
-            == ApplicationInstance.Product.BLACKBOARD
+            self._application_instance.product == ApplicationInstance.Product.BLACKBOARD
```
marcospri, line 537: Two less characters give us one line back.

PR #3701, opened 2022-02-17
Add refresh API info to Canvas error responses

seanh:
Add information to some of our API's error responses instructing the frontend to make a refresh request and try again.

This covers Canvas only, Blackboard is to come later.

Testing
=======

[localhost (make devdata) Canvas Files Assignment](https://hypothesis.instructure.com/courses/125/assignments/875) is a suitable assignment for testing this.

Enable the `frontend_refresh` feature flag
------------------------------------------

You need the [`frontend_refresh` feature flag](https://github.com/hypothesis/lms/pull/3695) on to test this PR:

```
```

Note that the feature flag doesn't actually enable frontend refreshing code (the frontend code hasn't been written yet) it just _disables_ backend refreshing code.

Refreshable error from Canvas
-----------------------------

1. Launch the assignment to make sure that you have an access token in your DB (click through the authorization dialog if you get it)

2. Invalidate all the access tokens in your DB:

   ```
   ```

3. Re-launch the assignment.

   No refresh request will happen because the feature flag has disabled backend refreshing and the frontend refreshing code hasn't been written yet. Instead you'll see the authorization dialog.
   
   If you open dev tools and look at the failed `via_url` and `sync` API response bodies you'll see that they contain refresh info for the frontend:
   
   ```json
   {"refresh": {"method": "POST", "path": "/api/canvas/oauth/refresh"}}
   ```

Non-refreshable error from Canvas
---------------------------------

There's all kinds of ways to produce a non-refreshable error from the Canvas API. You could pick any. Let's go with forcing an unexpected error response from Canvas:

1. Apply this diff to cause Canvas to send us unexpected errors:

   ```diff
   diff --git a/lms/services/canvas_api/client.py b/lms/services/canvas_api/client.py
   index e88d5016..f1ff8529 100644
   --- a/lms/services/canvas_api/client.py
   +++ b/lms/services/canvas_api/client.py
   @@ -279,7 +279,7 @@ class CanvasAPIClient:
            # https://canvas.instructure.com/doc/api/files.html#method.files.public_url
    
            return self._client.send(
   -            "GET", f"files/{file_id}/public_url", schema=self._PublicURLSchema
   +            "GET", f"files/foo/public_url", schema=self._PublicURLSchema
            )["public_url"]
 
        class _PublicURLSchema(RequestsResponseSchema):
   ```

2. Launch the assignment

   You'll see an error dialog and if you look at the response body of the failed `via_url` request it won't contain any refresh info.

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -19,13 +19,19 @@ class ExternalRequestError(Exception):
     """
 
     def __init__(
-        self, message=None, request=None, response=None, validation_errors=None
-    ):
+        self,
+        message=None,
+        request=None,
+        response=None,
+        validation_errors=None,
+        refreshable=False,
+    ):  # pylint: disable=too-many-arguments
         super().__init__()
         self.message = message
         self.request = request
         self.response = response
         self.validation_errors = validation_errors
+        self.refreshable = refreshable
 
     @property
     def url(self) -> Optional[str]:
@@ -181,7 +187,7 @@ class CanvasAPIError(ExternalRequestError):
         error_description = response_json.get("error_description", "")
 
         if {"message": "Invalid access token."} in errors:
-            raise OAuth2TokenError(**kwargs) from cause
+            raise OAuth2TokenError(refreshable=True, **kwargs) from cause
 
         if error_description == "refresh_token not found":
             raise OAuth2TokenError(**kwargs) from cause
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -229,7 +229,7 @@ class ErrorBody:
     message: str = None
     details: dict = None
 
-    def __json__(self, _request):
+    def __json__(self, request):
         """
         Return a JSON-serializable representation of this error response body.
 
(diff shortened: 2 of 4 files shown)


lms/services/exceptions.py
```diff
@@ -19,13 +19,19 @@ class ExternalRequestError(Exception):
...
 
     def __init__(
-        self, message=None, request=None, response=None, validation_errors=None
-    ):
+        self,
+        message=None,
+        request=None,
+        response=None,
+        validation_errors=None,
+        refreshable=False,
```
seanh, line 27: `ExternalRequestError.refreshable` has to be `False` by default because `ExternalRequestError` and its many subclasses (including `OAuth2TokenError`, `CanvasAPIError`, `CanvasAPIPermissionError`, `CanvasAPIServerError` and `HAPIError`) are raised all over the place, even in places like the `HAPI` client and the `LTIOutcomesService`. If `refreshable` were `True` by default we'd end up marking all sorts of things as refreshable

lms/services/exceptions.py
```diff
@@ -181,7 +187,7 @@ def raise_from(cls, cause, request, response, validation_errors=None):
         error_description = response_json.get("error_description", "")
 
         if {"message": "Invalid access token."} in errors:
-            raise OAuth2TokenError(**kwargs) from cause
+            raise OAuth2TokenError(refreshable=True, **kwargs) from cause
```
seanh, line 190: This only marks errors as refreshable when we get this specific `{"message": "Invalid access token."}` response body from Canvas.

We don't want `refreshable=T

lms/views/api/exceptions.py
```diff
@@ -238,7 +238,28 @@ def __json__(self, _request):
...
         """
-        return {key: value for key, value in asdict(self).items() if value is not None}
+        body = {key: value for key, value in asdict(self).items() if value is not None}
+
+        # If the exception is refreshable add refresh API info for the frontend.
+        # This tells the frontend to try making a refresh request then
+        # re-trying the original request.
+        if request.feature("frontend_refresh") and getattr(
+            request.exception, "refreshable", False
+        ):
```
seanh, line 248: This `if` will look nicer (and fit on one line) when the feature flag is removed

lms/views/api/exceptions.py
```diff
@@ -238,7 +238,28 @@ def __json__(self, _request):
...
+            oauth2_token_service = request.find_service(name="oauth2_token")
+
+            try:
+                oauth2_token_service.get()
```
seanh, line 252: TODO: Add `@lru_cache` to `OAuth2TokenService.get()` so we're not doing a duplicate DB query here (`get()` will already have been called earlier in the request)

marcospri: Separate PR for that?

marcospri (approved):

lms/views/api/exceptions.py
seanh: Yep: https://github.com/hypothesis/lms/pull/3714

PR #3704, opened 2022-02-21
Add `user` to the request

marcospri:
This adds `user` alongside `lti_user` to the request object for convince.

This PR doesn't yet use the new request property.

# Testing

No functional changes here. Just launch any assignment as a sanity check. 

A diff like 

```diff
diff --git a/lms/views/basic_lti_launch.py b/lms/views/basic_lti_launch.py
index 7eefdde1..c88078b4 100644
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -58,6 +58,8 @@ class BasicLTILaunchViews:
 
         self.context.js_config.add_document_url(document_url)
 
+        print("USER" * 10, self.request.user)
+
         return {}
 
     def sync_lti_data_to_h(self):

```

can force access to the new property on launching for example: https://hypothesis.instructure.com/courses/125/assignments/875

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -190,6 +190,14 @@ def _get_lti_user(request):
     return lti_user
 
 
+def _get_user(request):
+    return request.find_service(UserService).get(
+        request.find_service(name="application_instance").get_current(),
+        request.lti_user.user_id,
+    )
+
+
 def includeme(config):
     config.set_security_policy(SecurityPolicy(config.registry.settings["lms_secret"]))
     config.add_request_method(_get_lti_user, name="lti_user", property=True, reify=True)
+    config.add_request_method(_get_user, name="user", property=True, reify=True)
--- a/tests/unit/lms/security_test.py
+++ b/tests/unit/lms/security_test.py
@@ -13,6 +13,7 @@ from lms.security import (
     Permissions,
     SecurityPolicy,
     _get_lti_user,
+    _get_user,
     includeme,
 )
 from lms.validation import ValidationError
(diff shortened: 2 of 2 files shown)


lms/services/user.py
```diff
@@ -17,10 +17,17 @@ class UserService:
...
     """
 
-    def __init__(self, application_instance_service, db, h_authority: str):
+    def __init__(self, application_instance_service, db, h_authority: str, request):
         self._application_instance_service = application_instance_service
         self._db = db
         self._h_authority = h_authority
+        self._request = request
+
+    def get_current(self):
```
marcospri, line 26: Matches the naming and functionality of `application_instance_service.get_current`.

As the result would be available directly on request we won't see much usage of this but I think is nice to have matching APIs in the services.

lms/services/user.py
```diff
@@ -17,10 +17,17 @@ class UserService:
     At the moment this is purely used for recording/reporting purposes.
     """
 
-    def __init__(self, application_instance_service, db, h_authority: str):
+    def __init__(self, application_instance_service, db, h_authority: str, request):
         self._application_instance_service = application_instance_service
         self._db = db
         self._h_authority = h_authority
+        self._request = request
```
marcospri, line 24: Including here `request.lti_user` instead of the whole request resulted in a circular dependency y between the two `config.add_request_method` added properties, lti_user and user.

seanh (approved):
I think it's a shame to pass the whole `request` into `UserService.__init__()` as we usually try to avoid services having the request. Alternatives:

## Do pass the whole request into `UserService.__init__()`

As you have in this PR currently. It works and I think it's a reasonable option :man_shrugging: 

## Don't have a service method

The `_get_user()` method in `security.py` can just do it directly (note: still calling `UserService.get()`):

```python
def _get_user(request):
    application_instance_service = request.find_service(name="application_instance")
    user_service = request.find_service(UserService)

    return user_service.get(
        application_instance_service.get_current(),
        request.lti_user.user_id,
    )
```

## Take an `lti_user` argument

`UserService.get_current()` could take `request`, just `request.lti_user` or just `request.lti_user.user_id` as an argument. The name `get_current()` might not be exactly right anymore

marcospri:
> Take an `lti_user` argument

This created a circular dependency between the two request attribtes.


>  Don't have a service method

Yeah I think is a  better compromise than passing the whole request, updated.

PR #3718, opened 2022-02-22
Feature flag Blackboard refresh tokens

seanh:
Same as we [already did for Canvas](https://github.com/hypothesis/lms/pull/3695).

Testing:

1. Make sure you have a Blackboard access token in your DB: launch any Blackboard files or groups assignment and click through the authorization flow if you get it

2. Invalidate the access tokens in your DB:

3. Re-launch the assignmnent.

   With the `frontend_refresh` feature flag enabled (`export FEATURE_FLAG_FRONTEND_REFRESH=true`) no refresh should be attempted and you should get an authorization dialog.

   With the feature flag _disabled_ your access token should be automatically refreshed and the assignment should launch without asking you for authorization.

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -53,11 +53,13 @@ class BasicClient:
         redirect_uri,
         http_service,
         oauth_http_service,
+        refresh_enabled,
     ):  # pylint:disable=too-many-arguments
         self.blackboard_host = blackboard_host
         self.client_id = client_id
         self.client_secret = client_secret
         self.redirect_uri = redirect_uri
+        self.refresh_enabled = refresh_enabled
 
         self._http_service = http_service
         self._oauth_http_service = oauth_http_service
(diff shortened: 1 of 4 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -53,11 +53,13 @@ def __init__(
         redirect_uri,
         http_service,
         oauth_http_service,
+        refresh_enabled,
     ):  # pylint:disable=too-many-arguments
         self.blackboard_host = blackboard_host
         self.client_id = client_id
         self.client_secret = client_secret
         self.redirect_uri = redirect_uri
+        self.refresh_enabled = refresh_enabled
```
seanh, line 62: I made this public to make it easy to test. It's not important because the feature flag will soon be removed anyway

marcospri (approved):
Same as Canvas, no surprises.

Most of this can be deleted when we remove the feature flag :+1:

PR #3726, opened 2022-02-22
Add `BlackboardAPIClient.refresh_access_token()`

seanh:
Expose `BlackboardAPIClient.refresh_access_token()` as a public service method so that the refresh API view can call it in a future PR.

Same as `BlackboardAPIClient.get_token()`, just forwards to `OAuthHTTPService` (injecting Blackboard-specific details like the token URL).

### DIFF
--- a/lms/services/blackboard_api/_basic.py
+++ b/lms/services/blackboard_api/_basic.py
@@ -70,6 +70,13 @@ class BasicClient:
             authorization_code=authorization_code,
         )
 
+    def refresh_access_token(self):
+        self._oauth_http_service.refresh_access_token(
+            self.token_url,
+            self.redirect_uri,
+            auth=(self.client_id, self.client_secret),
+        )
+
     def request(self, method, path):
         url = self._api_url(path)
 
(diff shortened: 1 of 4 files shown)


lms/services/blackboard_api/_basic.py
```diff
@@ -70,6 +70,13 @@ def get_token(self, authorization_code):
             authorization_code=authorization_code,
         )
 
+    def refresh_access_token(self):
+        self._oauth_http_service.refresh_access_token(
+            self.token_url,
+            self.redirect_uri,
+            auth=(self.client_id, self.client_secret),
+        )
```
seanh, line 78: This is exactly the same as the existing code for doing a Blackboard refresh:

https://github.com/hypothesis/lms/blob/92eba8703f450a999f3f01ed055588a8c32a578e/lms/services/blackboard_api/_basic.py#L79-L83

This existing code will be removed soon, when the `frontend_refresh` feature flag is removed

marcospri: Why not call the new method on `request()`?

To make the deletion part later on clearer?

marcospri (approved):

PR #3728, opened 2022-02-22
Add Blackboard support to the refresh API using `ApplicationInstance.product`

seanh:
`ApplicationInstance.product`-based approach to adding Blackboard support to the refresh API.


lms/views/api/refresh.py
```diff
@@ -0,0 +1,40 @@
...
+        assert application_instance.product == ApplicationInstance.Product.BLACKBOARD
+        refresh_blackboard_access_token(request)
+
+
+def refresh_canvas_access_token(request):
+    canvas_api_client = request.find_service(name="canvas_api_client")
+    oauth2_token_service = request.find_service(name="oauth2_token")
+    refresh_token = oauth2_token_service.get().refresh_token
+
+    canvas_api_client.get_refreshed_token(refresh_token)
```
marcospri, line 35: Not the point of this PR but I think we should refactor the canvas side so this two functions are basically the same.

seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/3730

PR #3729, opened 2022-02-22
Make the refresh API not Canvas-specific

seanh:
Move the refresh API into a non-Canvas-specific location, rename the route to be non-Canvas-specific, etc.

The actual body code of the view still only supports Canvas, but a future PR will extend it to add Blackboard support.


seanh:
I've merged this into https://github.com/hypothesis/lms/pull/3728

PR #3730, opened 2022-02-22
Add Blackboard support to the refresh API using `request.matched_route`

seanh:
`request.matched_route`-based approach to adding Blackboard support to the refresh API.

# Testing

## Testing a request from a Canvas user

1. Log in to <https://hypothesis.instructure.com/> as the __Canvas: Instructor (Testing Account)__ user (in 1Password) and launch an assignment so that you have an access token for this user in your DB

2. Get a bearer token for the user to use in test requests. The `h_userid` in the code below is that of the __Canvas: Instructor (Testing Account)__ user:

   ```python
   $ make shell
   >>> from lms.validation.authentication import BearerTokenSchema
   >>> bearer_token_schema = BearerTokenSchema(request)
   >>> user = db.query(models.User).filter_by(h_userid="acct:3a022b6c146dfd9df4ea8662178eac@lms.hypothes.is").one()
   >>> lti_user = factories.LTIUser(user_id=user.user_id, oauth_consumer_key=user.application_instance.consumer_key, tool_consumer_instance_guid=user.application_instance.tool_consumer_instance_guid)
   >>> authorization_param = bearer_token_schema.authorization_param(lti_user)
   >>> authorization_param
   'Bearer eyJ***t-Y'
   ```

3. Check the value of the user's access token in your DB:

   ```
   tox -qe dockercompose -- exec postgres psql -U postgres -c "SELECT oauth2_token.access_token FROM oauth2_token, public.user WHERE oauth2_token.user_id = public.user.user_id AND public.user.h_userid = 'acct:3a022b6c146dfd9df4ea8662178eac@lms.hypothes.is';"
   ```

4. Use the bearer token to make a test request to the new refresh API. You should get a 200 OK response with no body. In the example below I'm using [httpie](https://httpie.io/) to make requests:

   ```terminal
   $ http POST 'http://localhost:8001/api/canvas/oauth/refresh' Authorization:'Bearer eyJ***t-Y'
   HTTP/1.1 200 OK
   Connection: close
   Content-Length: 4
   Content-Type: application/json
   Date: Mon, 14 Feb 2022 14:14:01 GMT
   Server: gunicorn

   null
   ```

5. Check the user's access token in your DB again: it should have changed

## Testing a request from a Blackboard user

1. Log in to <https://aunltd-test.blackboard.com/ultra> as the `blackboardteacher` user (in 1Password) and launch a Blackboard groups or files assignment so that you have an access token for this user in your DB

2. Get a bearer token for the user to use in test requests. The `h_userid` in the code below is that of the `blackboardteacher` user:

   ```python
   $ make shell
   >>> from lms.validation.authentication import BearerTokenSchema
   >>> bearer_token_schema = BearerTokenSchema(request)
   >>> user = db.query(models.User).filter_by(h_userid="acct:52140fb5154b791a0e1f4c5f917f0d@lms.hypothes.is").one()
   >>> lti_user = factories.LTIUser(user_id=user.user_id, oauth_consumer_key=user.application_instance.consumer_key, tool_consumer_instance_guid=user.application_instance.tool_consumer_instance_guid)
   >>> authorization_param = bearer_token_schema.authorization_param(lti_user)
   >>> authorization_param
   'Bearer eyJ***t-Y'
   ```

3. Invalidate the access token in your DB:

   ```
   ```

4. Use the bearer token to make a test request to the new refresh API. You should get a 200 OK response with no body. In the example below I'm using [httpie](https://httpie.io/) to make requests:

   ```terminal
   $ http POST 'http://localhost:8001/api/canvas/oauth/refresh' Authorization:'Bearer eyJ***t-Y'
   HTTP/1.1 200 OK
   Connection: close
   Content-Length: 4
   Content-Type: application/json
   Date: Mon, 14 Feb 2022 14:14:01 GMT
   Server: gunicorn

   null
   ```

5. Check the user's access token in your DB again: it should have changed

### DIFF
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -254,10 +254,13 @@ class ErrorBody:
                 # If we don't have an access token we can't refresh it.
                 pass
             else:
-                body["refresh"] = {
-                    "method": "POST",
-                    "path": request.route_path("canvas_api.oauth.refresh"),
-                }
+                if request.matched_route.name.startswith("canvas_api."):
+                    path = request.route_path("canvas_api.oauth.refresh")
+                else:
+                    assert request.matched_route.name.startswith("blackboard_api.")
+                    path = request.route_path("blackboard_api.oauth.refresh")
+
+                body["refresh"] = {"method": "POST", "path": path}
 
         return body
 
--- a/lms/views/api/refresh.py
+++ b/lms/views/api/refresh.py
@@ -13,8 +13,8 @@ def get_refreshed_token(request):
     """
     Refresh the user's access token.
 
-    Send a request to Canvas to get a refreshed access token for the
-    authenticated user and save it to the DB.
+    Send a request to get a refreshed access token for the authenticated user
+    and save it to the DB.
     """
     canvas_api = request.find_service(name="canvas_api_client")
     oauth2_token_service = request.find_service(name="oauth2_token")
--- a/lms/views/api/refresh.py
+++ b/lms/views/api/refresh.py
@@ -1,24 +1,25 @@
-from pyramid.view import view_config
+from pyramid.view import view_config, view_defaults
 
 from lms.security import Permissions
 
 
-@view_config(
-    request_method="POST",
-    route_name="canvas_api.oauth.refresh",
-    permission=Permissions.API,
-    renderer="json",
-)
-def get_refreshed_token(request):
-    """
-    Refresh the user's access token.
+@view_defaults(request_method="POST", permission=Permissions.API, renderer="json")
+class RefreshViews:
+    """Get a refreshed access token for the authenticated user and save it to the DB."""
 
-    Send a request to get a refreshed access token for the authenticated user
-    and save it to the DB.
-    """
-    canvas_api = request.find_service(name="canvas_api_client")
-    oauth2_token_service = request.find_service(name="oauth2_token")
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
 
-    refresh_token = oauth2_token_service.get().refresh_token
+    @view_config(route_name="canvas_api.oauth.refresh")
+    def get_refreshed_token_from_canvas(self):
+        canvas_api_client = self.request.find_service(name="canvas_api_client")
+        oauth2_token_service = self.request.find_service(name="oauth2_token")
+        refresh_token = oauth2_token_service.get().refresh_token
 
-    canvas_api.get_refreshed_token(refresh_token)
+        canvas_api_client.get_refreshed_token(refresh_token)
+
+    @view_config(route_name="blackboard_api.oauth.refresh")
+    def get_refreshed_token_from_blackboard(self):
+        blackboard_api_client = self.request.find_service(name="blackboard_api_client")
+        blackboard_api_client.refresh_access_token()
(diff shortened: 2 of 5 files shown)


lms/views/api/refresh.py
```diff
@@ -0,0 +1,38 @@
...
+        Refresh the user's access token.
+
+        Send a request to get a refreshed access token for the authenticated
+        user and save it to the DB.
+        """
+        if self.request.matched_route.name == "canvas_api.oauth.refresh":
+            self.refresh_canvas_access_token(self.request)
+        else:
+            assert self.request.matched_route.name == "blackboard_api.oauth.refresh"
+            self.refresh_blackboard_access_token(self.request)
```
seanh, line 25: It turns out that there's very little in common between what the view does for a Canvas refresh versus a Blackboard refresh. It might be nicer to just have two separate views instead of a single view that switches based on `request.matched_route`

lms/views/api/refresh.py
```diff
@@ -0,0 +1,25 @@
...
+        canvas_api_client = self.request.find_service(name="canvas_api_client")
+        oauth2_token_service = self.request.find_service(name="oauth2_token")
+        refresh_token = oauth2_token_service.get().refresh_token
+
+        canvas_api_client.get_refreshed_token(refresh_token)
+
+    @view_config(route_name="blackboard_api.oauth.refresh")
+    def get_refreshed_token_from_blackboard(self):
+        blackboard_api_client = self.request.find_service(name="blackboard_api_client")
+        blackboard_api_client.refresh_access_token()
```
seanh, line 25: The only thing that these two views have in common with each other is the `@view_defaults` and docstring on the class

tests/unit/lms/views/api/exceptions_test.py
```diff
@@ -232,6 +247,9 @@ def pyramid_request(self, pyramid_request):
         # exception that was raised by the original view:
         # https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.exception
         pyramid_request.exception = ValueError()
+
+        pyramid_request.matched_route = Mock(spec_set=["name"])
```
seanh, line 251: Pyramid's `DummyRequest` doesn't have a `matched_route` attribute, even though real requests do. I don't think there's any better way to mock a Pyramid `Route` object than this: it's not `create_autospec()`-friendly

marcospri: Yeah it's a bit too dummy, but this works.

lms/views/api/exceptions.py
```diff
@@ -254,10 +254,13 @@ def __json__(self, request):
                 # If we don't have an access token we can't refresh it.
                 pass
             else:
-                body["refresh"] = {
-                    "method": "POST",
-                    "path": request.route_path("canvas_api.oauth.refresh"),
-                }
+                if request.matched_route.name.startswith("canvas_api."):
```
marcospri, line 257: We could eventually use prefixes to organize the routes and maybe it's possible to check that (instead of name.startswith) 

https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#using-a-route-prefix-to-compose-applications


I don't think `name.startswith` is problematic here at the moment in any case.

lms/views/api/refresh.py
marcospri: Not the point of this PR but I think we should refactor the canvas side so these two views are basically the same.

marcospri (approved):

Issue #3639, opened 2022-02-02
Move `roles` off the `user` table


seanh:
The `user` table includes a `rows` column (instructor, learner, admin, etc):

https://github.com/hypothesis/lms/blob/0ae07d059298b7b42c72614f1dd04e39078660fb/lms/models/user.py#L37-L38

This isn't right: the same user can be an instructor in one course and a learner in another.

We should move the `roles` column off the `user` table, probably onto the `group_membership` table.

Questions:

* Can users also have different roles in different Blackboard or Canvas groups within the same course?

seanh:
Slack thread: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1643808229962529

Issue #3664, opened 2022-02-08
Track who "signed in" on every "sign-in"


jon-betts:
**Easily report on who and when users "sign into" the LMS app.**

## Why

For our updated definition of billable active, we need to be able to know when someone has signed in. This way we can charge for:

- Teachers who never annotate
- Students who never annotate, either because
  - their teacher has created reading assignments and added footnotes, but not required students to actually annotate, or
  - because they _should_ be using the app but aren't motivated enough to actually do the assignment.

And **not** charge for students who enrolled in a class but then dropped out.

## Constraints

- This needs to work only for sign-ins post-implementation.

## Notes

In the LMS context, we are currently happy using the app launch as sign-in.

jon-betts:
### What resolution?

What resolution do we need to know this at? We could log every single login, but alternatively we could group logins by a user by day or week etc.

If we did this it would bound the length of the table to `time_periods * users` regardless of the number of individual logins.

## What extra detail?

What do we need to know other than the user about the login? Do we need to know which group they logged into, or is the application instance enough?

## Might this be better in the context of a concrete requirement?

Could we get the exact report people would like, and then see what's required to produce it? My feeling is we want to replace the annotated in a group as the mark of active for logged into a group.

If this was the case it would immediately answer some of the questions above:

 * Resolution: Minimum 6 monthly
 * Additional details: Group, Application instance

chrisshaw:
I need to rewrite this around the reports we'll use it for a la https://github.com/hypothesis/private-issues/issues/122 and its siblings.

PR #3732, opened 2022-02-23
Always update `updated` on User

marcospri:
In practice this will serve as a `last_login` field and it's also in line with `File.updated` and `GroupMembership.updated` which also get updated every time they are seen instead of only when something changed.

Note that once that https://github.com/hypothesis/lms/issues/3639 gets done nothing on the current User table will ever change, so without this change, created == updated always.

## Testing

- Remove your users: 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate \"user\", grouping_membership"```

- Launch any assignment, for example: https://hypothesis.instructure.com/courses/125/assignments/873

- Check the updated field:

```tox -qe dockercompose -- exec postgres psql -U postgres -c "select updated from \"user\""```

- Wait at least a millisecond and launch the assignment again, re-query the updated field. It will be the same.

- Switch to this branch, re-launch and query. The date will update on every launch.

### DIFF
--- a/lms/services/user.py
+++ b/lms/services/user.py
@@ -1,6 +1,7 @@
 from functools import lru_cache
 from typing import Optional
 
+from sqlalchemy import func
 from sqlalchemy.exc import NoResultFound
 
 from lms.models import LTIUser, User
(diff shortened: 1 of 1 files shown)


marcospri:
This could be useful for the "when" side of https://github.com/hypothesis/lms/issues/3664

PR #3748, opened 2022-03-03
Add application_instance_id to GroupInfo, Oauth2Token and GradingInfo

marcospri:
Adding the three columns in one migration to avoid extra bookkeeping and PR noise.

All columns added as null able in this first step.

The migration is auto generated from the model changes in: https://github.com/hypothesis/lms/pull/3749/files

# Testing
Run the migration:

- ```hdev alembic upgrade head```

```
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 7a4812876915 -> 407ff423b9e3, Add missing application_instances FK.
```

- Check the new table structure, new column and FK

```tox -qe dockercompose -- exec postgres psql -U postgres -c "\d group_info" ```
```tox -qe dockercompose -- exec postgres psql -U postgres -c "\d lis_result_sourcedid" ```
```tox -qe dockercompose -- exec postgres psql -U postgres -c "\d oauth2_token" ```

PR #3749, opened 2022-03-03
Model changes to add application_instance_id

marcospri:
Model changes for the migration in https://github.com/hypothesis/lms/pull/3748. 

Needs to the deployed after the migration is deployed.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -87,7 +87,9 @@ class ApplicationInstance(BASE):
     #: (each token belongs to a different user of this application
     #: instance's LMS).
     access_tokens = sa.orm.relationship(
-        "OAuth2Token", back_populates="application_instance"
+        "OAuth2Token",
+        back_populates="application_instance",
+        foreign_keys="OAuth2Token.consumer_key",
     )
 
     #: A list of all the courses for this application instance.
@@ -95,7 +97,9 @@ class ApplicationInstance(BASE):
 
     #: A list of all the GroupInfo's for this application instance.
     group_infos = sa.orm.relationship(
-        "GroupInfo", back_populates="application_instance"
+        "GroupInfo",
+        back_populates="application_instance",
+        foreign_keys="GroupInfo.consumer_key",
     )
 
     #: A list of all the files for this application instance.
--- a/lms/models/grading_info.py
+++ b/lms/models/grading_info.py
@@ -48,6 +48,14 @@ class GradingInfo(CreatedUpdatedMixin, BASE):
     lis_result_sourcedid = sa.Column(sa.UnicodeText(), nullable=False)
     lis_outcome_service_url = sa.Column(sa.UnicodeText(), nullable=False)
     oauth_consumer_key = sa.Column(sa.UnicodeText(), nullable=False)
+
+    #: The ApplicationInstance that this group belongs to foreign key
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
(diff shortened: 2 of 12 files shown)


tests/unit/lms/services/group_info_test.py
```diff
@@ -119,7 +119,7 @@ def params(self):
         return {
             column: f"TEST_{column.upper()}"
             for column in GroupInfo.columns()
-            if column not in ("consumer_key", "_info")
+            if column not in ("consumer_key", "_info", "application_instance_id")
```
marcospri, line 122: This automatic param generation based on the name only it's not great. 

Noted to revisit the service and test for a refactor when this settles.

lms/models/application_instance.py
```diff
@@ -87,15 +87,19 @@ class Product(str, Enum):
     #: (each token belongs to a different user of this application
     #: instance's LMS).
     access_tokens = sa.orm.relationship(
-        "OAuth2Token", back_populates="application_instance"
+        "OAuth2Token",
+        back_populates="application_instance",
+        foreign_keys="OAuth2Token.consumer_key",
```
marcospri, line 92: This `foreign_keys` is needed on the SQLA side to know what FK should the relationship follow in the backwards relationship to access_tokens and group_infos now that there are two FK pointing to `application_instances`

lms/models/group_info.py
```diff
@@ -40,9 +40,16 @@ class GroupInfo(BASE):
...
+    application_instance_id = sa.Column(
+        sa.Integer(),
+        sa.ForeignKey("application_instances.id", ondelete="cascade"),
+        nullable=True,
+    )
+
     #: The ApplicationInstance that this group belongs to.
     application_instance = sa.orm.relationship(
-        "ApplicationInstance", back_populates="group_infos"
+        "ApplicationInstance", back_populates="group_infos", foreign_keys=[consumer_key]
```
marcospri, line 52: This is the other side of the relationship, `foreign_keys` also needed here to know which FK to follow.

In this PR, as application_instance_id is null-able everything still uses `consumer_key`

lms/models/application_instance.py
```diff
@@ -87,15 +87,19 @@ class Product(str, Enum):
...
 
     #: A list of all the courses for this application instance.
     courses = sa.orm.relationship("LegacyCourse", back_populates="application_instance")
 
     #: A list of all the GroupInfo's for this application instance.
     group_infos = sa.orm.relationship(
-        "GroupInfo", back_populates="application_instance"
+        "GroupInfo",
+        back_populates="application_instance",
+        foreign_keys="GroupInfo.consumer_key",
```
marcospri, line 102: Same, we point application_instances.group_infos to make the queries to use GroupInfo.consumer_key.

lms/models/grading_info.py
```diff
@@ -48,6 +48,14 @@ class GradingInfo(CreatedUpdatedMixin, BASE):
     lis_result_sourcedid = sa.Column(sa.UnicodeText(), nullable=False)
     lis_outcome_service_url = sa.Column(sa.UnicodeText(), nullable=False)
     oauth_consumer_key = sa.Column(sa.UnicodeText(), nullable=False)
+
+    #: The ApplicationInstance that this group belongs to foreign key
```
jon-betts, line 52: What's up with this weird comment style in all these files?

tests/unit/lms/services/group_info_test.py
jon-betts: Eh. :shrug: Maybe this should just be hard-coded?

PR #3752, opened 2022-03-04
Fill application_instance_id in GroupInfo

marcospri:
Fills the new column for new and "in fly" rows. Next step will be a migration to fill up the rest and mark the column as non-nullable.

### Inserting

-  Remove existing group_info

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate group_info;"```

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from group_info;"
```

you'll get one row per section and one for the course. All with the `application_instance_id`

`tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from  lis_result_sourcedid;"`

### Updating

- Set application_instance_id to null

```tox -qe dockercompose -- exec postgres psql -U postgres -c "update group_info set application_instance_id = NULL;"```

- Relaunch - https://hypothesis.instructure.com/courses/125/assignments/875

No more nulls in application_instance_id.

### DIFF
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -26,7 +26,7 @@ class GroupInfoService:
         self._db = request.db
         self._lti_user = request.lti_user
 
-    def upsert(self, h_group, consumer_key, params):
+    def upsert(self, h_group, application_instance, params):
         """
         Upsert a row into the `group_info` DB table.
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lms/services/group_info.py [hunk 1]
```diff
@@ -43,7 +43,7 @@ def upsert(self, h_group, consumer_key, params):
         :param h_group: the group to upsert
         :type h_group: models.HGroup
 
-        :param consumer_key: the GroupInfo.consumer_key value to set
+        :param application_instance: the ApplicationInstance this group belongs to
 
         :param params: the other GroupInfo columns to set
         :type params: dict
```
jon-betts, line 49: It would be good to tidy these type declarations into annotations. Also the reference to `models.GroupInfo` is weird, as it's `GroupInfo` locally.

PR #3751, opened 2022-03-04
Fill application_instance_id in GradingInfo

marcospri:
Fills the new column for new and "in fly" rows. Next step will be a migration to fill up the rest and mark the column as non-nullable.

### Inserting

-  Remove existing grading info

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate lis_result_sourcedid;"```

- Launch the blackboard assignment `localhost (make devdata) HTML Assignment` as the student user `blackboardstudent2`
- https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_36_1&course_id=_19_1

The new columns is filled in the new row:

`tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from  lis_result_sourcedid;"`

### Updating

- Set application_instance_id to null

`tox -qe dockercompose -- exec postgres psql -U postgres -c "update lis_result_sourcedid set application_instance_id = null;"`

- Launch the assignment again

`tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from  lis_result_sourcedid;"`

The column will have a value again.

PR #3750, opened 2022-03-04
Fill application_instance_id in OAuth2TokenService

marcospri:
Fills the new column for new and "in fly" rows. Next step will be a migration to fill up the rest and mark the column as non-nullable.

### Updating

-  Remove existing tokens

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"```

- On the branch `ai-fk-model`, launch https://hypothesis.instructure.com/courses/125/assignments/875

You'll get a new token with a null application_instance_id

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from oauth2_token;"


 application_instance_id 
-------------------------
                        
(1 row)
```

- Switch to this branch and invalidate the token with:

```tox -qe dockercompose -- exec postgres psql -U postgres -c "update oauth2_token set access_token = 'nope';"```

- Launch the assignment again, application_instance_id will get the correct value

```

tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from oauth2_token;"                        


 application_instance_id 
-------------------------
                       8
(1 row)
```

### Inserting

-  Remove existing tokens

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"```

- Launch the assignment again, application_instance_id will get the correct value

```

tox -qe dockercompose -- exec postgres psql -U postgres -c "select application_instance_id from oauth2_token;"

Issue #3719, opened 2022-02-22
Add an "application_instance_id" FK to tables that currently use `consumer_key`


marcospri:
## Action plan

- https://github.com/hypothesis/lms/issues/3719
  - [x] https://github.com/hypothesis/lms/pull/3748
  - [x]  https://github.com/hypothesis/lms/pull/3749
  - [x]  https://github.com/hypothesis/lms/pull/3750
  - [x]  https://github.com/hypothesis/lms/pull/3751
  - [x]  https://github.com/hypothesis/lms/pull/3752
  - [x]  https://github.com/hypothesis/lms/pull/3753
  - [x]  https://github.com/hypothesis/lms/pull/3767
  - [x]  https://github.com/hypothesis/lms/pull/3768
  - [x] https://github.com/hypothesis/lms/pull/3779
  - [x] https://github.com/hypothesis/lms/pull/3780
  - [x] https://github.com/hypothesis/lms/pull/3781

PR #3753, opened 2022-03-04
Migration to backfill application_instance_id

marcospri:
For: https://github.com/hypothesis/lms/issues/3719

Model changes in: https://github.com/hypothesis/lms/pull/3754

As:

- https://github.com/hypothesis/lms/pull/3752
- https://github.com/hypothesis/lms/pull/3751
- https://github.com/hypothesis/lms/pull/3750

are already merged all new rows should have `application_instance_id`

A quick query on metabase to check that:

```
select 'group_info', application_instance_id from group_info where application_instance_id is not null
union
select 'token', application_instance_id from oauth2_token where application_instance_id is not null
union
select 'grading', application_instance_id from lis_result_sourcedid where application_instance_id is not null
```

### Testing

Ideally we could test this on a LMS's DB dump to get an ballpark figure of how long is going to take.

In any case the migration can be run with:

```hdev alembic upgrade head```

```                                                      
dev installed: *** listing modules disabled by tox-pip-sync in pyproject.toml ***
dev run-test-pre: PYTHONHASHSEED='230568549'
dev run-test: commands[0] | alembic -c /home/marcos/hypo/lms/conf/alembic.ini upgrade head
INFO  [alembic.runtime.migration] Running upgrade 407ff423b9e3 -> 0a3bfcc14133, Bakfill application_instance_id in group_info, lis_result_sourcedid and oauth2_token.
```

### DIFF
--- a/lms/migrations/versions/0a3bfcc14133_backfill_application_instance_id.py
+++ b/lms/migrations/versions/0a3bfcc14133_backfill_application_instance_id.py
@@ -1,0 +1,85 @@
+"""
+Bakfill application_instance_id in group_info, lis_result_sourcedid and oauth2_token.
+
+Revision ID: 0a3bfcc14133
+Revises: 407ff423b9e3
+Create Date: 2022-03-04 11:39:06.573269
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0a3bfcc14133"
+down_revision = "407ff423b9e3"
+
+
+def upgrade():
+    conn = op.get_bind()
+    # Fist backfill the new application_instance_id columns based on the existing consumer_key columns
+    conn.execute(
+        """
+            UPDATE group_info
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lms/migrations/versions/0a3bfcc14133_backfill_application_instance_id.py [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+revision = "0a3bfcc14133"
+down_revision = "407ff423b9e3"
+
+
+def upgrade():
+    conn = op.get_bind()
+    # Fist backfill the new application_instance_id columns based on the existing consumer_key columns
+    conn.execute(
+        """
+            UPDATE group_info
```
marcospri, line 22: The three queries are pretty similar:
- Set application_instance_id from the  application_instances.id
- When the consumer_key on the table matches the one in application_instance
- Only for rows that don't have it already.

PR #3763, opened 2022-03-08
JSTOR schema support in the backend

marcospri:
See https://github.com/hypothesis/via/pull/701


lms/config/__init__.py
```diff
@@ -72,6 +72,7 @@ def configure(settings):
         "admin_auth_google_client_secret": sg.get("ADMIN_AUTH_GOOGLE_CLIENT_SECRET"),
         "blackboard_api_client_id": sg.get("BLACKBOARD_API_CLIENT_ID"),
         "blackboard_api_client_secret": sg.get("BLACKBOARD_API_CLIENT_SECRET"),
+        "jstor_ip": sg.get("JSTOR_IP"),
```
robertknight, line 75: Ultimately the IP (or other credentials) will need to be associated either with the LMS app installation directly or indirectly (eg. if we make it a property of an entity of the institution/school associated with the installation). For an initial internal demo that is only usable by the Hypothesis test LMS installations, we could use a fixed IP in an env var.

robertknight (commented):
This worked as expected. I had a question about checking feature flags when launching the assignment.

lms/resources/_js_config/__init__.py
```diff
@@ -83,6 +84,10 @@ def add_document_url(self, document_url):
                 "launchUrl": launch_url,
                 "launchParams": launch_params,
             }
+        elif self._jstor_service.enabled and document_url.startswith("jstor://"):
```
marcospri, line 87: Added the `self._jstor_service.enabled` check here.

`document_url.startswith("jstor://")` could be in `_jstor_service.is_jstor` or similar but is simple enough that is easier to read it inline.

lms/services/jstor.py
```diff
@@ -0,0 +1,28 @@
...
+    def enabled(self):
+        return self._settings.get("enabled", False)
+
+    @staticmethod
+    def via_url(request, document_url):
+        return via_url(
+            request,
+            document_url,
+            content_type="pdf",
+            options={"jstor.ip": request.registry.settings["jstor_ip"]},
```
marcospri, line 18: It's now a trivial change to do `{"jstor.ip": settings["jstor_ip"]}` to use a per application instance value.

Left the value from the env variable now as is probably easier to handle while we only have one.

lms/services/jstor.py
```diff
@@ -0,0 +1,28 @@
...
+        )
+
+
+def factory(_context, request):
+    # application_instance.settings is an ApplicationSettings which overwrites `get`
+    # convert to a dict to be able to access ["jstor"] securely
```
marcospri, line 24: `ApplicationSettings.get` takes a group and key, expecting always nested values.

lms/views/helpers/_via.py
```diff
@@ -16,17 +16,25 @@ def via_url(request, document_url, content_type=None):
     :param request: Request object
     :param document_url: Document URL to present in Via
     :param content_type: Either "pdf" or "html" if known, None if not
+    :param options: Any extra options for the url
```
marcospri, line 19: `ViaClient.url_for` includes an `options` value already. This just exposes in on the helper.

robertknight:
If JSTOR is disabled for a Hypothesis LMS installation then launching a JSTOR assignment fails, as expected. The error for anyone interested is:

JSTOR disabled error

lms/resources/_js_config/__init__.py
```diff
@@ -83,6 +85,8 @@ def add_document_url(self, document_url):
                 "launchUrl": launch_url,
                 "launchParams": launch_params,
             }
+        elif jstor_service.enabled and document_url.startswith("jstor://"):
+            self._config["viaUrl"] = jstor_service.via_url(self._request, document_url)
```
robertknight, line 89: Currently if JSTOR is disabled the user will see an exception from Via. That's OK, although it may make more sense to have the LMS app report the error.

lms/services/jstor.py
robertknight, line 24: I couldn't parse the part of this comment after "overwrites `get`". Can you clarify why wrapping `ApplicationSettings.get` is needed here?

robertknight (commented):
This all works as expected. I didn't understand the comment about why the `ApplicationSettings.get` access is wrapped (ie. why the settings is converted to a dict first and then `get` called on that). Can you clarify?

lms/services/jstor.py
marcospri: As `ApplicationSettings.get` is implemented it only works to access two nested levels.

"get("jstor", "enabled", False)" - Returns False when either "jstor" or "jstor->enable" don't exist.

Here we want the full `jstor` dict. But ApplicationSettings doesn't allow for just one level, ie you can't do get("jstor", default=False).

Having "key" be optional here https://github.com/hypothesis/lms/blob/1b6049dc81d9287ab34fd78efe1eac56ed844f96/lms/models/application_settings.py#L35 might be the right solution but I rather tackle that on it's own.

robertknight: Ah - I see why I didn't understand the comment! I didn't realize these were two separate sentences (ending after `get` and beginning with `Convert`). Can you update this and then it is good to go.

PR #3764, opened 2022-03-09
Remove useless settings set in test

marcospri:
The settings are set a bit below, that's the line that sets them. The removed one doesn't have any effect as far as I can tell:

https://github.com/hypothesis/lms/blob/753b9c0fbcf7fe1634992228f2fd6d0b141c4524/tests/unit/lms/validation/authentication/_oauth_test.py#L165


jon-betts (approved):
If the test still pass, this is good!

PR #3765, opened 2022-03-11
GroupInfoService test fixtures simplification

marcospri:
- Use the global `application_instance` fixture
- Use a factory instead of creating models directly to be align to our testing style.

I few test here could use a deeper refactor but this is a start and gets in the way of another PR.


jon-betts (approved):
This is much snapperi

PR #3780, opened 2022-03-14
Stop filling OAuth2Token consumer_key

marcospri:
For https://github.com/hypothesis/lms/issues/3719

Stop filling the old column now that it can hold nulls.

#### Oauth2Token
- `tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"`

- Check the rows again:

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select consumer_key, application_instance_id from oauth2_token;"
 consumer_key | application_instance_id 
--------------+-------------------------
              |                       8
(1 row)
```

### DIFF
--- a/tests/unit/lms/services/oauth2_token_test.py
+++ b/tests/unit/lms/services/oauth2_token_test.py
@@ -22,7 +22,6 @@ class TestOAuth2TokenService:
         oauth2_token = db_session.query(OAuth2Token).one()
         assert oauth2_token == Any.object(OAuth2Token).with_attrs(
             {
-                "consumer_key": application_instance.consumer_key,
                 "application_instance_id": application_instance.id,
                 "user_id": lti_user.user_id,
                 "access_token": "access_token",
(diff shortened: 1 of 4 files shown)


tests/unit/lms/services/oauth2_token_test.py
```diff
@@ -43,7 +42,7 @@ def test_get_raises_OAuth2TokenError_with_no_token(
     ):
         wrong_value = "WRONG"
         if wrong_param == "application_instance":
-            application_instance.consumer_key = "WRONG"
+            application_instance.id = 100
             wrong_value = application_instance
```
jon-betts, line 46: I think this test might be a case of superficial similarity. The code used to try and account for the difference between the two cases is about the same as just having two cases. e.g.:

```python
    def test_get_raises_OAuth2TokenError_with_mismatching_application_instance(
        self, db_session, application_instance, lti_user
    ):
        service = OAuth2TokenService(
            db_session,
            application_instance=factories.ApplicationInstance(),
            user_id=lti_user.user_id,
        )

        with pytest.raises(OAuth2TokenError):
            service.get()

    def test_get_raises_OAuth2TokenError_with_mismatching_user(
        self, db_session, application_instance, lti_user
    ):
        service = OAuth2TokenService(
            db_session, application_instance=application_instance, user_id="WRONG"
        )

        with pytest.raises(OAuth2TokenError):
            service.get()
```

If we want to get very fancy, we can use a partial to create the service. This has the advantage of only including the field that matters in the definition of the test, which makes it clearer what's changing / important:

```python
    from functools import partial

   ...

    def test_get_raises_OAuth2TokenError_with_bad_ai(self, get_service):
        service = get_service(application_instance=factories.ApplicationInstance())

        with pytest.raises(OAuth2TokenError):
            service.get()

    def test_get_raises_OAuth2TokenError_with_bad_user(self, get_service):
        service = get_service(user_id="WRONG")

        with pytest.raises(OAuth2TokenError):
            service.get()

    ...

    @pytest.fixture
    def svc(self, get_service):
        return get_service()

    @pytest.fixture
    def get_service(self, pyramid_request, application_instance):
        return partial(
            OAuth2TokenService,
            db=pyramid_request.db,
            application_instance=application_instance,
            user_id=pyramid_request.lti_user.user_id,
        )
```

You could also skip the `svc` fixture here and use `get_service` in every test without much loss for even shorter tests:

```python
    def test_get_returns_token_when_present(self, get_service, oauth_token):
        assert get_service().get() == oauth_token

    def test_get_raises_OAuth2TokenError_with_bad_ai(self, get_service):
        with pytest.raises(OAuth2TokenError):
            get_service(application_instance=factories.ApplicationInstance()).get()

    def test_get_raises_OAuth2TokenError_with_bad_user(self, get_service):
        with pytest.raises(OAuth2TokenError):
            get_service(user_id="WRONG").get()
```

Where your appetite falls off for compactness may vary. I think keeping the arrange / act / assert setup is probably worth it.

jon-betts (approved):
Yep I think this works, but there is some scope to tighten up the tests a bit.

I've made some suggestions about using a `functools.partial` to create a fixture which allows you to only be forced to specify some fields with defaults taking over for the test. I think this is  a pretty powerful technique to allow you to highlight what's changing and therefore important in a given test.

It also happens to make them quite a bit shorter.

PR #3781, opened 2022-03-14
Stop filling GroupInfo consumer_key

marcospri:
Before merging this changes need to be make on the sparky side to join on application_instance_id instead of consumer_key. 

The analytics side should be already be prepared as we recreate the consumer_key column there using the new FK.
---

For https://github.com/hypothesis/lms/issues/3719

Stop filling the old column now that it can hold nulls.

- `tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"`

- Check the rows:

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select consumer_key, application_instance_id from group_info;"
```
```
 consumer_key | application_instance_id 
--------------+-------------------------
              |                       8
              |                       8
              |                       8
              |                       8
(4 rows)
```

### DIFF
--- a/lms/services/group_info.py
+++ b/lms/services/group_info.py
@@ -12,7 +12,7 @@ class GroupInfoService:
     Usage::
 
         group_info = request.find_service(name="group_info")
-        group_info.upsert(authority_provided_id, consumer_key, request.params)
+        group_info.upsert(h_group, application_instance, request.params)
     """
 
     GROUPING_TYPES = {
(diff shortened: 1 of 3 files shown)


lms/services/group_info.py
```diff
@@ -26,13 +26,13 @@ def __init__(self, _context, request):
...
-    def upsert(self, h_group, application_instance, params):
+    def upsert(self, h_group, application_instance, params: dict):
         """
         Upsert a row into the `group_info` DB table.
 
         Find the models.GroupInfo matching the given h_group or create it if
-        none exists. Then update the GroupInfo's consumer_key to the given
-        consumer_key, and update its other columns from the items in `params`.
+        none exists. Then update the GroupInfo's application_instance to the given
+        application_instance, and update its other columns from the items in `params`.
```
jon-betts, line 35: :+1:

jon-betts (approved):
Yep, this look good. Very to the point.

We do need to do something on the Sparky stuff to make it compatible, but I don't think you need to wait. We aren't live yet, so we can try and catch up.

PR #3785, opened 2022-03-15
AES Service 

marcospri:


lms/models/application_instance.py
```diff
@@ -179,52 +174,3 @@ def product(self):
             product = self.Product.UNKNOWN
 
         return product
-
-    @classmethod
-    def build_from_lms_url(  # pylint:disable=too-many-arguments
```
marcospri, line 184: Moved to ApplicationInstanceService.

lms/services/application_instance.py
```diff
@@ -1,18 +1,22 @@
...
+from lms.services.aes import AESService
 
 
 class ApplicationInstanceNotFound(Exception):
     """The requested ApplicationInstance wasn't found in the database."""
 
 
 class ApplicationInstanceService:
-    def __init__(self, db, request):
+    def __init__(self, db, request, aes_service):
```
marcospri, line 16: Injecting the aes_service to use it on `build_from_lms_url`

lms/services/aes.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def encrypt(self, aes_iv, plain_text: Union[str, bytes]):
+        if isinstance(plain_text, str):
+            plain_text = plain_text.encode("utf-8")
+
+        return AES.new(self.secret, AES.MODE_CFB, aes_iv).encrypt(plain_text)
+
+    @staticmethod
+    def build_iv():
+        return Random.new().read(AES.block_size)
```
jon-betts, line 23: I think this would definitely benefit from some doc strings, especially given the domain specific language etc.

lms/services/application_instance.py
```diff
@@ -54,6 +58,40 @@ def get_by_consumer_key(self, consumer_key) -> ApplicationInstance:
...
+        self._db.add(application_instance)
+        return application_instance
+
 
 def factory(_context, request):
-    return ApplicationInstanceService(request.db, request)
+    return ApplicationInstanceService(
+        request.db,
+        request,
+        request.find_service(AESService),
```
jon-betts, line 96: It's happening again! I think it's a bit odd to pass a mixture of things from the request, and the request. Does this still work if you pass the `lti_user` as I think that's all the request is used for internally?

lms/models/application_instance.py
jon-betts: Yep, this is a good fix. This doesn't look like model code.

tests/unit/lms/models/application_instance_test.py
```diff
@@ -83,50 +83,22 @@ def test_lms_host_raises_ValueError(self, application_instance, lms_url):
...
+        assert developer_secret == aes_service.decrypt.return_value
 
-        assert application_instance.consumer_key
-        assert application_instance.shared_secret
-        assert application_instance.lms_url == "https://example.com/"
-        assert application_instance.requesters_email == "example@example.com"
-        assert application_instance.developer_key == "TEST_DEVELOPER_KEY"
-        assert application_instance.developer_secret
-        assert application_instance.settings == {}
+    def test_decryped_developer_secret_returns_None_if_ApplicationInstance_has_no_developer_secret(
```
jon-betts, line 98:
```suggestion
    def test_decrypted_developer_secret_returns_None_if_ApplicationInstance_has_no_developer_secret(
```

tests/unit/lms/models/application_instance_test.py
```diff
@@ -83,50 +83,22 @@ def test_lms_host_raises_ValueError(self, application_instance, lms_url):
...
+        self, application_instance, aes_service
+    ):
+        assert application_instance.decrypted_developer_secret(aes_service) is None
```
jon-betts, line 101: It might be nice to set the developer key to `None` here to show the point of the test, even if it's not required.

tests/unit/lms/services/aes_service_test.py
```diff
@@ -0,0 +1,25 @@
+from unittest.mock import sentinel
+
+import pytest
+
+from lms.services.aes import AESService, factory
+
+
+class TestAESService:
+    @pytest.mark.parametrize("text, expected", [(b"TEXT", b"TEXT"), ("TEXT", b"TEXT")])
+    def test_it(self, svc, text, expected):
```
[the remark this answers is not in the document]
jon-betts: Hmm... :thinking: I think I'd call this a "round-trip" test, and I think it's valuable to have, but what's missing here is that you could replace the encrypt and decrypt functions with `lamdba value: value` and it would still pass.

It would be good (and quite easy) to capture some canned responses for the encrypt and decrypt functions. You could also assert the `build_iv` gives you something of the expected length at least.

tests/unit/lms/services/canvas_api/factory_test.py
```diff
@@ -10,8 +10,13 @@
...
 
 class TestCanvasAPIClientFactory:
+    @pytest.mark.usefixtures("aes_service")
     def test_building_the_CanvasAPIClient(
-        self, pyramid_request, CanvasAPIClient, AuthenticatedClient, file_service
+        self,
+        pyramid_request,
+        CanvasAPIClient,
+        AuthenticatedClient,
+        file_service,
```
jon-betts, line 19:
```suggestion
        self, pyramid_request, CanvasAPIClient, AuthenticatedClient, file_service
```

If you snip the last comma off, Black will back this onto one line

tests/unit/lms/views/reports_test.py
```diff
@@ -24,13 +25,9 @@ def test_build_launches_rows(self, pyramid_request):
...
 
-        def build_ai_from_pair(pair):
-            return ApplicationInstance.build_from_lms_url(
-                pair[0], pair[1], None, None, None, None
-            )
-
         app_instances = [
-            build_ai_from_pair(pair) for pair in zip(test_urls, test_emails)
+            factories.ApplicationInstance(lms_url=pair[0], requesters_email=pair[1])
+            for pair in zip(test_urls, test_emails)
```
jon-betts, line 30:
```suggestion
            factories.ApplicationInstance(lms_url=url, requesters_email=email)
            for url, email in zip(test_urls, test_emails)
```

jon-betts (changes requested):
Overall I think this is a big improvement. There was a lot of stuff in the model here.

I do think the new service needs documentation. I think the tests could do with at least a few pre baked versions which prove the the AES stuff does something AES related. It would be ideal if these values could be grabbed from something external as a check that we agree with an external source on this. At a minimum just baking some of our own would still be better.

lms/services/application_instance.py
marcospri: No, unfortunately there's a circular dependency between lti_user <-> application instance if we bring the LTIUser into the factory.

I think the right fix is a refactor/merge of security.Identity/LTIUser but I rather do that once we have the LTI1.3 stuff in so it's clearer it will work for both.

jon-betts (approved):
The tests here look really good. I know it seems like a pain, but these kinds of tests are actually really helpful. Recently I swapped out the underlying crypto library, and it's very nerve wracking without some exact values to prove we can still encrypt and decrypt as before.

Issue #3699, opened 2022-02-17
Remove LegacyCourse


marcospri:
To adapt existing tables for LTI1.3 we need to relate to application instances using a FK instead of consumer key.

`course` (model name LegacyCourse) is one of the tables with this problem but instead of adding a new key we are deprecating the table and switch to use only `groupings`.

There's also a product desire to consolidate all courses in a single table: https://github.com/hypothesis/private-issues/issues/95

- [x] https://github.com/hypothesis/lms/pull/3788
- [x] https://github.com/hypothesis/lms/pull/3793
- [x] https://github.com/hypothesis/lms/pull/3808
- [x] https://github.com/hypothesis/lms/pull/3787

----

The `grouping` table is meant to replace the existing `course` and we already renamed its model to `LegacyCourse`.

There is still a fair amount of code dealing with maintaining both tables:

https://github.com/hypothesis/lms/blob/master/lms/services/course.py
https://github.com/hypothesis/lms/blob/master/lms/resources/lti_launch.py#L29

The information we need for grouping is both in `course` and `group_info` but `group_info`doesn't have rows for each `course`.

Of those missing rows, around 5k, 4k don't have a row in H's group table. 

I think we could do the migration, stop reading and writing from course and forget about those missing rows.

Some more details about the existing data with SQL:

```sql
--  Total missing groupings ~ 22k
select (select count(*) from course) - (select count(*) from grouping where type ='course')


-- We have lots of courses from AI with no GUID. That tell us that they've not been launched in a while and if they were they might get a new authority_provided_id depending of the GUID.
-- ~17k
select (select count(*) from course join application_instances on course.consumer_key = application_instances.consumer_key where application_instances.tool_consumer_instance_guid is not null ) - (select count(*) from grouping where type ='course')



-- Those course records
select * from course join application_instances on course.consumer_key = application_instances.consumer_key
where (course.consumer_key, authority_provided_id ) not in 
    (
        select consumer_key, authority_provided_id from grouping join application_instances on application_instances.id = grouping.application_instance_id
    ) 
    and   application_instances.tool_consumer_instance_guid is not null
    


-- For those, find the ones we DON'T have record in group_info
--~ 5k
select course.authority_provided_id from course
left outer JOIN group_info  ON course.authority_provided_id = group_info.authority_provided_id and group_info.consumer_key = course.consumer_key
join application_instances on course.consumer_key = application_instances.consumer_key

where (course.consumer_key, course.authority_provided_id ) not in 
[...]

PR #3787, opened 2022-03-16
Stop using LegacyCourse

marcospri:
For #3699 

After merging https://github.com/hypothesis/lms/pull/3808 and with the logging introduced in https://github.com/hypothesis/lms/pull/3788 we can check that we haven't found any occurrence where we would've used legacy course: https://my.papertrailapp.com/groups/14104911/events?q=Created+course+from+existing+legacy+course.+context_id since the back-fill was deployed.

This PR removes all code usage of LegacyCourse but leaves the table there in case we have to revert any of this.

## Testing

- Remove existing grouping/courses 

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate grouping,grouping_membership,course"
```

- Check groupings has a row for the course

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from grouping where type = 'course'"
          created           |          updated           | id | application_instance_id |          authority_provided_id           | parent_id |                  lms_id                  |                  lms_name                   |  type  |                                                                                                                            settings                                                                                                                             |                     extra                      
----------------------------+----------------------------+----+-------------------------+------------------------------------------+-----------+------------------------------------------+---------------------------------------------+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------------------------------------
 2022-05-05 08:06:59.005543 | 2022-05-05 08:06:59.005543 |  8 |                       8 | 653e9620a656a954c684942f6443fa3e3410c03a |           | f3cd019017839c4630358662a05540f2f6ec5f93 | Developer Test Course with Sections Enabled | course | {"jstor": {"enabled": true, "site_code": null}, "canvas": {"groups_enabled": true, "sections_enabled": true}, "blackboard": {"files_enabled": false, "groups_enabled": false}, "vitalsource": {"enabled": true}, "microsoft_onedrive": {"files_enabled": true}} | {"canvas": {"custom_canvas_course_id": "125"}}
(1 row)
```

- Course remains empty

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from course"                        
 consumer_key | authority_provided_id | settings 
--------------+-----------------------+----------
(0 rows)

```

- Relaunch https://hypothesis.instructure.com/courses/125/assignments/873
- The same grouping row is re-used.


lms/resources/lti_launch.py
```diff
@@ -38,31 +38,14 @@ def get_or_create_course(self):
...
-        # Capture the state here, before any other SQLA query does an implicit flush and then
-        # removing legacy_course from `db.new`
-        is_new_legacy_course = legacy_course in self._request.db.new
-
-        course = course_service.upsert(
-            authority_provided_id,
+        return course_service.upsert(
+            course_service.generate_authority_provided_id(
+                tool_consumer_instance_guid, context_id
+            ),
```
jon-betts, line 44: I think this refactor has tidied this up to the extent it's pointing out there's something weird about `generate_authority_provided_id`. As far as I can tell it's never called for anything other than just to feed-back to the course service.

This makes me think it's not really an external method. The two patterns I can spot are:

```
    course_service.upsert(
        course_service.generate_authority_provided_id(
            tool_consumer_instance_guid, context_id
        ),
        ...
    )
```

```
    course_service.get(
        course_service.generate_authority_provided_id(
            self.tool_consumer_instance_guid, course_id
        )
    )
```

I can't spot `upsert` or `get` ever being called by anything else, which makes me think this should be an internal method and exposed through the interface.

lms/resources/lti_launch.py
```diff
@@ -38,31 +38,14 @@ def get_or_create_course(self):
...
             context_id,
             params["context_title"],
             self._course_extra(),
-            legacy_course.settings,
```
jon-betts, line 55: This call has a lot of arguments. I think it might be easier to piece together with key word args rather than positional.

This all passes the nod test if everything is in the right place, but it's very hard to spot ordering errors.

lms/services/course.py
```diff
@@ -73,17 +60,6 @@ def get(self, authority_provided_id):
...
-        course = LegacyCourse(
-            consumer_key=self._application_instance.consumer_key,
-            authority_provided_id=authority_provided_id,
-            settings=self._new_course_settings(authority_provided_id),
-        )
-
-        self._db.add(course)
-
-        return course
-
```
jon-betts, line 86: I think `_create` could be in-lined now. I'm not sure why it's taken out of the upsert method?

As a nice bonus if we alphabetically sort the methods all these internal methods which are for `upsert` will be next to it for easier reading.

I've also noticed `generate_authority_provided_id` is after a bunch of private methods which makes it easy to miss.

jon-betts (approved):
Yeah, this is a really nice clean up.

I think this is starting the virtual circle of refactoring and pointing the way to some next steps we could take. None of these were introduced in this PR, they are just being highlighed by it.

### Things you might want to do in this PR

 * Using key word args on the call to `upsert` might make the code easier to follow
 * Some re-ordering and in-lining in the `CourseService` would aid readbility and fit our standards a bit better

### A good follow up PR

I have the strong impression that `generate_authority_provided_id()` should be private and the `CourseService` interface would be the better for it. It's generally a bad sign when you read from an object just to feed the result back into it and I think this PR makes it clearer that's what we're doing.

I've not performed an exhaustive search, but I'm fairly sure it's never used independently, and we never use the methods we pass it to without it. So for example instead of the interface being:

```python
course_service.get(
    course_service.generate_authority_provided_id(tool_consumer_instance_guid, context_id)
)
```

It should probably be:

```python
course_service.get_by_guid(tool_consumer_instance_guid, context_id)
```

..or something with a better name.

PR #3793, opened 2022-03-18
Fix temporally course logging logic for courses

marcospri:
- Use temp bool variables to capture the state in the session correctly
- Log the course's context_id instead of the consumer_key

## Testing

First, what we are trying to find here is cases where we do have an existing "LegacyCourse" but not a "Course". This means that the course was created before the new "Course" model and not launched since, until now.

We are not interested in cases where both LegacyCourse and Course are created at the same time. That would be the very common case of a new course using Hypothesis.

(LegacyCourse's table name is "course", Course's is "grouping". I'll use the model names here as is less confusing)

## No Course / No LegacyCourse

tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from grouping where type ='course'"
tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from course"

Launch  https://hypothesis.instructure.com/courses/125/assignments/873

Nothing gets logged. This course is new as far as LMS is concern.

## Both Course / Legacy course exist

Launch  https://hypothesis.instructure.com/courses/125/assignments/873 again

Nothing gets logged, this is an existing course, nothing interesting.

## Course exists / LegacyCourse doesn't

tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from course"

Launch  https://hypothesis.instructure.com/courses/125/assignments/873 again

No logging. This case shouldn't ever happen, but just in case.

## LegacyCourse exists / Course doesn't

tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from grouping where type ='course'"

Launch  https://hypothesis.instructure.com/courses/125/assignments/873 again

Here we do get the log. The LegacyCourse was in the database and we created and new Course from it.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -42,6 +42,10 @@ class LTILaunchResource:
         )
 
         legacy_course = course_service.get_or_create(authority_provided_id)
+        # Capture the state here, before any other SQLA query does an implicit flush and then
+        # removing legacy_course from `db.new`
+        is_new_legacy_course = legacy_course in self._request.db.new
+
         course = course_service.upsert(
             authority_provided_id,
             context_id,
@@ -49,11 +53,12 @@ class LTILaunchResource:
             self._course_extra(),
             legacy_course.settings,
         )
+        new_course = course.id is None
 
-        if not self._request.db.is_modified(legacy_course) and not course.id:
+        if not is_new_legacy_course and new_course:
             LOG.warning(
-                "Created course from existing legacy course. consumer_key: %s",
-                legacy_course.consumer_key,
+                "Created course from existing legacy course. context_id: %s",
+                context_id,
             )
 
         return legacy_course, course


lms/resources/lti_launch.py
```diff
@@ -42,18 +42,21 @@ def get_or_create_course(self):
...
         )
+        new_course = course.id is None
 
-        if not self._request.db.is_modified(legacy_course) and not course.id:
+        if not new_legacy_course and new_course:
             LOG.warning(
-                "Created course from existing legacy course. consumer_key: %s",
-                legacy_course.consumer_key,
+                "Created course from existing legacy course. context_id: %s",
+                context_id,
```
marcospri, line 59: Logging the consumer_key was not the interesting bit. The course id is more interesting if we need to manually check something about it later.

lms/resources/lti_launch.py
```diff
@@ -42,18 +42,21 @@ def get_or_create_course(self):
         )
 
         legacy_course = course_service.get_or_create(authority_provided_id)
+        new_legacy_course = legacy_course in self._request.db.new
```
marcospri, line 45: We can check the state down below, accessing the service again in line 47 does an implicit flush marking this as not new. 

Capturing the state here and checking it later.

jon-betts: If this is important I'd definitely put a comment here. Not a chance in a million I'd spot that if I came to refactor.

jon-betts: If this is important I'd definitely put a comment here. Not a chance in a million I'd spot that if I came to refactor.

jon-betts: Github is being weird again. Not sure if this is coming through.

If we have an object for `legacy_course` then in generally I expect `new_legacy_course` to also be an object rather than a boolean.

I think a nice pattern for this is just to prefix booleans with `is_` to the extent it makes gramatical sense (it doesn't always). So we'd have:

* `legacy_course` - object
* `is_new_legacy_course` - bool

jon-betts: Github is being weird again. Not sure if this is coming through.

If we have an object for `legacy_course` then in generally I expect `new_legacy_course` to also be an object rather than a boolean.

I think a nice pattern for this is just to prefix booleans with `is_` to the extent it makes gramatical sense (it doesn't always). So we'd have:

* `legacy_course` - object
* `is_new_legacy_course` - bool

jon-betts (approved):
There's some GH mess still going on here. But this works as described.

It's small, but I think it would benefit from a bit of name change / commenting.

Issue #3723, opened 2022-02-22
Hoist LTI1.3 parameters into existing marshmallow schemas


marcospri:
My proposed approach will be to validate and decode the JWT token at the `security` level
and the each existing marshmallow schema that reads LTI1.1 parameters should map the LTI1.3 values when present.

Something similar to what was done on the PoC: 

https://github.com/hypothesis/lms/pull/3672/files#diff-ac438dfdb882a1f2620de809aedfe2ed92fa45e214f1983a5ddd27592f388f86R93
https://github.com/hypothesis/lms/pull/3672/files#diff-13eb6a4c86a8780b4d9fa6fd70059edd9394a9b017a06f6be9140bde8e8e97b5R32

Issue #3720, opened 2022-02-22
Provide access to the LTI1.3 parameters outside a marshmallow schema


marcospri:
Currently we read LTI parameters mostly from schemas defined per view using `parsed_params`.

We could, and I think should, push the LTI1.3 values into those same schemas so the code using `parsed_parms` doesn't need to worry about being LTI1.3 or LTI1.1.

There are however places where we access `request.params` directly, mostly on predicates and the likes where we don't yet have chosen a view and we don't have a schema.

We should have a way to expose the values from the JWT token so for example predicates don't need to worry about decoding it. 

Ideally we could have something like `request.lti_params` which is agnostic to the LTI version. We should keep `request.params` as it is, providing raw access to pyramid when needed.

PR #3804, opened 2022-03-23
Provide access to LTI1.3 parameters included in the secure JWT

marcospri:
This PR solves both:

https://github.com/hypothesis/lms/issues/3723
https://github.com/hypothesis/lms/issues/3720

### TODO

- [x] Sqash commits before merge.

## Changes summary

- A dictionary  that translates LTI1.1 to LTI1.3 names

https://github.com/hypothesis/lms/pull/3804/files#diff-51c8c114a41a6e7fbb11b0b864264eda079d85b577d359fc7886bebdef201bf4

- Changed `PyramidRequestSchema` to use the translation dictionary when a `id_token` parameters is present.

https://github.com/hypothesis/lms/pull/3804/files#diff-4022dcd9ee6d82833d767b89bd52e9821864fda7f81eed537aa06b67b61b9cc9

- A new method in `ApplicationService` to find instances using LTI1.3 parameters

https://github.com/hypothesis/lms/pull/3804/files#diff-17a0e48c5e3354637d9ad94190c94643da3ae0c072fc8f63307c1e3b904d6064

- A new "authentication source" for LTI1.3 https://github.com/hypothesis/lms/pull/3804/files#diff-b0028bbc04c7f0ef9f8940267f79bfd5f20e991d09c3abc3d57e327d762f0d1d

This reads the dict provided by the schema and produces a LTIUser

### LTI1.1

- Nothing has changed on the LTI1.1 but I'm double checking launching a few assignment

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_398_1&course_id=_19_1

Also trying to reconfigure an assignment using  
`localhost (make devdata) with Sections Enabled` in canvas.

#### Create the registration if not already in your DB

- `make shell`
```python
db.add(
    models.LTIRegistration(
        issuer="https://canvas.instructure.com", 
        client_id="93820000000000200", 
        auth_login_url="https://canvas.instructure.com/api/lti/authorize_redirect",                 
        key_set_url="https://canvas.instructure.com/api/lti/security/jwks", 
        token_url="ttps://canvas.instructure.com/api/lti/authorize"
    )
)
tm.commit()
```

- Now launch the LTI1.3 assignment: https://hypothesis.instructure.com/courses/319/assignments/2781
- You don't yet have an application_instance using the lti registration so you'll get `You're not authorized to view this page` response.

This 403 response ins the `http://localhost:8001/lti_launches` endpoint which is receiving now the id_token JWT payload.

- Link one existing ApplicationInstance to the new registration:

In `make shell`

```python
ai = db.query(models.ApplicationInstance).order_by(models.ApplicationInstance.id.desc()).first()

ai.lti_registration_id = db.query(models.LTIRegistration).one().id
ai.deployment_id = "609:6ed5aff92b5257ab607ec9c629844ea8586fde68"

tm.commit()
```

- Launch https://hypothesis.instructure.com/courses/319/assignments/2781. The launch is successful and `http://localhost:8001/lti_launches` returns a 200 response.


lms/lti.py
```diff
@@ -0,0 +1,86 @@
+from dataclasses import dataclass
+from typing import Callable, Optional
+
+
```
marcospri, line 4: This is in a new top level `lti.py` file. It could go in a different place but not fit any of them perfectly.

This being just plain python (no pyramid, no marshmallow, etc) I think is fine to leave it at the top level.

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+@dataclass
+class _ParamMapping:
```
marcospri, line 6: This class might be a bit of an overkill but it looks nicer than just plain tuples and if only as a documentation mechanism I think is worth it.

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+    key: str
+    """Name of the parameter in LTI1.3"""
+
+    sub: Optional[str] = None
+    """For nested values, name in the second object"""
+
+    function: Optional[Callable] = None
```
marcospri, line 13: I had this as "method" before but the resulting method needed to be a static one for pylint and having a function is a bit nicer as it doesn't need a `gettattr` as the method approach.

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+    """If additional processing is needed, function to further process the value"""
+
+
+def _stringify_roles(lti_13_value):
```
marcospri, line 17: I think roles is the only parameter that needs further processing.

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Roles are a list in LTI1.3 but comma separated in LTI1.1.
+    """
+    return ",".join(lti_13_value)
+
+
+class LTI13Params(dict):
+    """Dictionary subclass that translates LTI1.1 parameter names to the LTI1.3 ones."""
+
+    lti_param_mapping = {
```
marcospri, line 29: This is surprisingly small, not 100% of the parameters are here as I'm not covering the ones needed for configuring assignments but it's let's say 75% of the real size.

I think this approach has a few things going for it:

-  It's just python, all it's doing its translating one name to another so it doesn't need to be more complicated.
- It's all in one place and not scattered around different files.
- Works for marshmallow schema and more important, works with out them. 

Marshmallow schema expect a dictionary to `Schema().load` data through them so having this translation exposed as a dict works transparently as marshmallow is concern.

We have a problem where sometimes we need to access the LTI parameters without access to a schema. For example in pyramid predicates. Having this as a dict allows for that without any mayor app refactors.

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+            "https://purl.imsglobal.org/spec/lti/claim/resource_link", "id"
+        ),
+        "issuer": _ParamMapping("iss"),
+        "client_id": _ParamMapping("aud"),
+        "deployment_id": _ParamMapping(
+            "https://purl.imsglobal.org/spec/lti/claim/deployment_id"
+        ),
+    }
+
+    def get(self, key, default=None):
```
marcospri, line 66: Need to override this one to use the class's `__getitem__`

lms/lti.py
```diff
@@ -0,0 +1,86 @@
...
+            return self[key] or default
+        except KeyError:
+            return default
+
+    def __getitem__(self, key):
+        """Access the LTI1.3 parameters using their corresponding LTI1.1 names."""
+        if key not in self.lti_param_mapping:
+            return super().__getitem__(key)
+
+        mapping = self.lti_param_mapping[key]
```
marcospri, line 77: If not there we are fine rising KeyError as a regular dict.

lms/models/lti_user.py
```diff
@@ -42,6 +42,21 @@ def is_learner(self):
         """Whether this user is a learner."""
         return "learner" in self.roles.lower()
 
+    @staticmethod
+    def from_auth_params(application_instance, auth_params):
```
marcospri, line 46: auth_params is not just some random dict, is of the type LTIAuthParamsSchema.

I had to do a lot of ceremony to annotate it here. I'll revisit it later.
https://github.com/hypothesis/lms/pull/3804/commits/ab216ec02302bda69c401dca1a5544bd5737fd18

lms/resources/lti_launch.py
```diff
@@ -122,7 +122,7 @@ def resource_link_id(self):
         ):
             return resource_link_id
 
-        return self._request.POST.get("resource_link_id")
+        return self._request.lti_params.get("resource_link_id")
```
marcospri, line 125: This is a big change even if it looks a bit silly.

This runs from a predicate which run before "view derivers" (we add `parsed_params` to the request in a web deriver).

That means that self.request.params will contain `id_token` but we don't have access to it's containing information. Exposing `lti_params` directly on the request allows to fix this.

lms/validation/_base.py
```diff
@@ -94,6 +95,18 @@ def parse(self, *args, **kwargs):
     def _handle_error(error, _req, _schema, *, error_status_code, error_headers):
         raise ValidationError(messages=error.messages) from error
 
+    @marshmallow.pre_load
+    def _decode_jwt(self, data, **_kwargs):
```
marcospri, line 99: This makes all schemas used to get "parsed_params" LTI1.3 compatible using the params from the id_token.

lms/validation/_base.py
```diff
@@ -1,12 +1,25 @@
...
 from pyramid.httpexceptions import HTTPUnsupportedMediaType
 from webargs import pyramidparser
 
 from lms.services import ExternalRequestError
 from lms.validation._exceptions import ValidationError
 
+from marshmallow.utils import get_value
+
+
+class Reach(fields.Field):
```
marcospri: Marshmallow makes this a bit difficult.

It looks like you could use Pluck or Nested or some other combination but you have to do something like this (from: https://github.com/marshmallow-code/marshmallow/issues/1315#issuecomment-514266136)

lms/validation/_lti.py
```diff
@@ -1,6 +1,134 @@
...
+    user_id = marshmallow.fields.Str(required=True)
+    roles = marshmallow.fields.Str(required=True)
+    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
+    lis_person_name_given = marshmallow.fields.Str(load_default="")
+    lis_person_name_family = marshmallow.fields.Str(load_default="")
+    lis_person_name_full = marshmallow.fields.Str(load_default="")
+    lis_person_contact_email_primary = marshmallow.fields.Str(load_default="")
+
+
+class LTI13AuthParamsSchema(LTIAuthParamsSchema):
```
marcospri: This could be tidier up but the marshmallow approach as a few problems IMO:

- Marshmallow doesn't make it nicer that just a map of keys, arguably 

```
   tool_consumer_instance_guid = Reach(
        fields.Str(),
   data_key="https://purl.imsglobal.org/spec/lti/claim/tool_platform",
        path="guid",
        required=True,
)
```

is harder to read that the other solution.

- We need to do the mapping in each Schema with spreads the "LTI1.3 mapping problem" across multiple files.


-As is tied to the schema is not possible to use the mappings without one. This is a problem when we don't have a "parsed_params" and we still need to access the LTI parameters as we do in the predicates.

lms/validation/_lti.py
```diff
@@ -1,6 +1,134 @@
...
+    context_title = fields.Str(required=True)
+    lti_version = fields.Str(validate=OneOf(["LTI-1p0", "1.3.0"]), required=True)
+
+    custom_canvas_api_domain = fields.Str()
+    custom_canvas_course_id = fields.Str()
+    launch_presentation_return_url = fields.Str()
+    tool_consumer_info_product_family_code = fields.Str()
+
+
+class LTI13CommonLTILaunchSchema(LTI13AuthParamsSchema):
```
marcospri: There's a lot of classes here. I think another issue with this approach is that we end up duplicating the information.

lms/validation/authentication/_launch_params.py
```diff
@@ -5,11 +5,12 @@
 from lms.models import LTIUser, display_name
 from lms.services import ApplicationInstanceNotFound, LTILaunchVerificationError
 from lms.validation._base import PyramidRequestSchema
+from lms.validation._lti import LTIAuthParamsSchema
 
 __all__ = ("LaunchParamsAuthSchema",)
 
 
-class LaunchParamsAuthSchema(PyramidRequestSchema):
+class LaunchParamsAuthSchema(LTIAuthParamsSchema, PyramidRequestSchema):
```
marcospri, line 13: I think this could be moved to `lms/validation/authentication/_lti.py` and renamed to `LTI11AuthSchema` along side the LTI13 contuerpart.

Left is as it for now.

tests/unit/lms/lti_test.py
```diff
@@ -0,0 +1,129 @@
...
+
+from lms.lti import LTI13Params, _ParamMapping
+
+
+def sample_function(value):
+    return value * 2
+
+
+class TestLTI13Params:
+    @pytest.mark.parametrize(
```
marcospri, line 12: Testing the basic mechanism of the class here, independently of any LTI mapping.

tests/unit/lms/lti_test.py
```diff
@@ -0,0 +1,129 @@
...
+            param("key", {"notkey": "value"}, "default", "default", id="default"),
+        ],
+    )
+    def test_get(self, key, data, default, expected):
+        assert LTI13Params(data).get(key, default) == expected
+
+    @pytest.mark.parametrize(
+        "lti_11_key,value",
+        [
+            ("user_id", "USER_ID"),
```
marcospri, line 62: And testing all the LTI mappings here.

tests/unit/lms/lti_test.py
```diff
@@ -0,0 +1,129 @@
...
+            ("issuer", "ISSUER"),
+            ("client_id", "CLIENT_ID"),
+            ("deployment_id", "DEPLOYMENT_ID"),
+        ],
+    )
+    def test_13_mappings(self, lti_13_launch_params, lti_11_key, value):
+        assert LTI13Params(lti_13_launch_params)[lti_11_key] == value
+
+    @pytest.fixture
+    def lti_13_launch_params(self):
```
marcospri, line 83: All over this PR there's a lot of "lti_13", not sure if "lti13" no underscore it's better.

tests/unit/conftest.py
```diff
@@ -190,6 +190,22 @@ def application_instance(db_session):
     return application_instance
 
 
+@pytest.fixture
+def lti13_application_instance(db_session):
+    lti_registration = factories.LTIRegistration()
+
+    application_instance = factories.ApplicationInstance(
```
marcospri, line 197: not reusing `application_instance`, we want two different rows in the DB.

lms/services/application_instance.py
```diff
@@ -58,6 +58,27 @@ def get_by_consumer_key(self, consumer_key) -> ApplicationInstance:
         except NoResultFound as err:
             raise ApplicationInstanceNotFound() from err
 
+    @lru_cache
+    def get_by_deployment_id(
```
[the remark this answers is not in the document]
jon-betts: I think this is closer to being right as it is actually. I think 1.3 is super vague.

What I'm less sure is if this is looking it up by deployment any more than it's looking it up by any of the other things. I think:

* If there's a collective noun of some kind for `issuer`, `client_id` and `deployment_id` - then that's best
* If not, then I think the main thing this name does for us is disinguish itself from other things and make it greppable... which this totally does

lms/lti.py
jon-betts: We've talked about this at great length, but long story short:

 * It's ideal if we don't mix the 1.1 and 1.3 things together in the same dict
 * There's a slightly simpler way to do this

jon-betts (commented):
I'm just going to post my random thoughts at the moment. Nothing much to say yet as I think this is going to change a bit

lms/services/application_instance.py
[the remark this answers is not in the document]
marcospri: I think the name for `issuer + client_id + deployment_id` is "install" or application_instance itself.

I leave it as get_by_deployment_id.

lms/validation/_base.py
```diff
@@ -94,6 +95,20 @@ def parse(self, *args, **kwargs):
...
+    def _decode_jwt(self, data, **_kwargs):
+        """Use the values encoded in the `id_token` JWT if present."""
+        if "id_token" not in self.context["request"].POST:
+            return data
+
+        params = LTIParams.from_v13(self.context["request"].lti_jwt)
+        # Make the rest of params also accessible to marshmallow in case any are not coming from the JWT
+        # eg query parameters
+        # This is to make it backwards compatible with schemas that mix LTI parameters with others that belong to the LMS app (eg the `url` parameter).
+        params.update(self.context["request"].params)
```
marcospri, line 108: This extra parameters will be feed to the schema but ignored if not used, ie they won't appear to be present at all if not explicitly declared.

lms/validation/authentication/_lti.py
```diff
@@ -0,0 +1,56 @@
...
+
+    location = "form"
+
+    iss = marshmallow.fields.Str(required=True)
+    aud = marshmallow.fields.Str(required=True)
+    deployment_id = marshmallow.fields.Str(required=True)
+
+    @marshmallow.pre_load
+    def _lti_v3_fields(self, data, **_kwargs):  # pylint:disable=no-self-use
+        data["iss"] = data.v13["iss"]
```
marcospri, line 26: This, using `pre_load` is the clearest way I've found to reference `v13`. 

We could use a custom marshmallow field but as this is the only place this needs to happen that would be overkill and not at all easier to read.

lms/security.py
```diff
@@ -173,6 +179,10 @@ def _get_lti_user(request):
         partial(bearer_token_schema.lti_user, location="form"),
         OAuthCallbackSchema(request).lti_user,
     ]
+    if "id_token" in request.params:
```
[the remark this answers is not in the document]
jon-betts: I'm a bit confused over the main motivation here. Avoiding extra work seems like a good reason to do this, but the comment says it's because we might have a bogus token?

I'm not quite sure what this means in the current (quite strange) design where we just try everything until something works. In that design presumably we'd expect it to be broken and just move onto the next thing?

lms/validation/authentication/_lti.py
jon-betts: There's `Pluck`? Not sure if you tried this:

* https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html?highlight=Pluck#marshmallow.fields.Pluck

This seems to allow you to surface a sub key into a parent object, but... why not just let it be nested? It seems like we've avoided mixing the params in the lower object just to mix them here? I think this doesn't achieve the aim of keeping them separate which is to highlight which version we are using in the code.

Is the most natural thing here to have a `v13` sub-schema? This would mean the code would look like:

```python
request.parsed_params['v13']['iss']
```

Which is more verbose, but also completely explicit.

lms/models/lti_user.py
jon-betts: This would be fine content for a doc-string. I think we should discuss turning on the linting requirement again to document things as I feel our default has changed from documenting most things with a few exceptions, to not documenting things by default.

It's annoying to write things like `Intialize the object` for init methods, but I think it's probably on balance better than not having docstrings for our methods.

lms/security.py
```diff
@@ -197,7 +207,34 @@ def _get_user(request):
...
+
+
 def includeme(config):
     config.set_security_policy(SecurityPolicy(config.registry.settings["lms_secret"]))
     config.add_request_method(_get_lti_user, name="lti_user", property=True, reify=True)
+    config.add_request_method(_get_lti_jwt, name="lti_jwt", property=True, reify=True)
     config.add_request_method(_get_user, name="user", property=True, reify=True)
+    config.add_request_method(
+        _get_lti_params, name="lti_params", property=True, reify=True
+    )
```
jon-betts, line 240: I'm pretty skeptical of request methods, as they are very non-local by nature and it's quite difficult to put together how they are used. As they appear on the request object it can be difficult to immediately even notice they aren't part of `pyramid`. In this case they are prefixed with `lti_` which should tip the reader off, which you can then try and grep for the source (which can be anywhere).

In this case `lti_jwt` only seems to be used in the `_get_lti_params` method. So it's defined for the whole app, but is it, practically speaking, a private variable for this function? `lti_params` also seems to only be used in the context object in `lms/resources/lti_launch.py` so seems to be functionally private to that too?

My current understanding is this is defined in `security` and added on all requests, even those which don't have any LTI params, to be used in a context which is only available in one set of views? Should this be in the context object?

I could see how `lti_params` could be more generally useful. Is this the start of moving away from `parsed_params`? Otherwise this seems to tread the exact same ground?

lms/services/application_instance.py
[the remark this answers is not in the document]
jon-betts: Maybe this should be ```get_blah(lti_v13_params: LTI13Params)``` then? Not sure how to name that, or if that's much worse.

lms/security.py
jon-betts: I've just spotted that these are used as `self.context['request'].lti_jwt`. I was looking for `request.lti_jwt` for usages.

I'm 100% unsure if this means everything I said is wrong, or kind of proves my point.

lms/validation/_base.py
jon-betts: This kind of really seems like it's in the wrong place to me, or I don't understand, or both? This seems like it would apply way too broadly?

I'm also very confused about the relationship between the code in `security` adding this stuff, the eventual schemas that use it etc...

I suspect honestly I'm really missing the point here. Might be time to call the review and have a chat.

jon-betts (commented):
Eh.. I have a bunch of random comments on here, but I think the jist is actually that I don't get what's going on Which probably means non of my comments apply.

So I can't really give a review yet, although this is probably feedback of a kind.

lms/security.py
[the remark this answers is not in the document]
marcospri: It's probably not worded very well.

What I mean is that I just don't outright do `schemas = [LTI13AuthSchema]` just in case id_token is there but somehow doesn't work and we'd like to continue checking other schemas. Not sure how probable that scenario is.

lms/security.py
marcospri: I think `lti_params` makes sense in  the request, it's  going to be used in views,resource and predicates.

`lti_jwt` not so much. It belongs to a service JWT service with a `@cache` decorator but as I mention somewhere here that's a big TODO, that service it bigger that what is present here and will required IMO moving other JWT usages to it so I left it for another PR if that makes sense.

lms/services/application_instance.py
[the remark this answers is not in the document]
marcospri: Not sure, the only moment we use this is the callback of the auth flow, before an actual LTI launch is made so that's why I preferred the separated params.

I can't find a good name for them that is not a synonym of `application_instance`.

lms/security.py
```diff
@@ -197,7 +207,35 @@ def _get_user(request):
     )
 
 
+def _get_lti_jwt(request):
```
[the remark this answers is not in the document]
jon-betts: We had a chat about this and yep, lets park it for now and come back later.

lms/security.py
jon-betts: We had a chat about this, and I think we agreed it only makes sense in the same places that the `resources/lti_launch.py` does, rather than all views. This means this probably can be sucked up into the `LTILaunchResource`.

lms/services/application_instance.py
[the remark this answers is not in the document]
jon-betts: Ignore me! This isn't used in an LTI "launch" context at all, and is called with a custom schema elsewhere.

lms/validation/_base.py
jon-betts: We had a chat and I think the conclusions are:

* Although all `request.parsed_params` are `PyramidRequestSchema`
* ...and all `LTILaunchBlahSchema` are `PyramidRequestSchema`
* ... not all `PyramidRequestSchema` have anything to do with `LTILaunch...` - plenty are used for internal stuff, or Canvas etc. So this is more specific.
* `lms.validation._lti.LTIAuthParamsSchema` looks like a good home, but it's not a child of `PyramidRequestSchema`
* ... but every child of it is and it's not used independently

So:

 * `LTIAuthParamsSchema` should probably be a child of `PyramidRequestSchema`
 * All its children can then stop having multiple inheritance
 * And this method can move there

lms/models/lti_user.py
jon-betts: This seems to be duplicate code from: 

* https://github.com/hypothesis/lms/blob/jwt-auth/lms/validation/authentication/_launch_params.py#L67

We should try and collapse these two paths to be the same path. Presumably we can remove the testing from one or the other too? (If there's any)

lms/validation/_lti.py
```diff
@@ -1,8 +1,22 @@
+import marshmallow
 from webargs import fields
 
 from lms.validation._base import PyramidRequestSchema
 
 
+class LTIAuthParamsSchema(marshmallow.Schema):
```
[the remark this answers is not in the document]
jon-betts: Or as this seems to be at the bottom of everything, is this maybe an `LTIV11CoreSchema`. We might be converting 1.3 to 1.1 before using it, but it is using 1.1 names.

tests/unit/conftest.py
jon-betts: Might be worth keeping an eye on this to see if we end up using it in more than one place. Or alternatively we could move it where it's used and ship it outwards as it gets more widely used?

tests/unit/lms/lti_test.py
```diff
@@ -0,0 +1,70 @@
...
+        assert not LTIParams.from_v13({}).v11
+
+    def test_v11(self):
+        sample_dict = {"test": "key"}
+
+        params = LTIParams(sample_dict)
+        assert params == params.v11 == sample_dict
+
+    @pytest.fixture
+    def lti_v13_launch_params(self):
```
jon-betts, line 37: This seems to be overlapping with `lms.tests.conftest.py:lti_v13_params`?

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -24,16 +24,16 @@ def test_it(self, lti_launch, course_service):
...
     @pytest.mark.parametrize(
-        "learner_id,get_id,post_id,expected",
+        "learner_id,get_id,lti_id,expected",
         [
-            param(None, None, "POST_ID", "POST_ID", id="regular"),
-            param("USER_ID", "GET_ID", "POST_ID", "GET_ID", id="new_speedgrader"),
-            param("USER_ID", None, "POST_ID", "POST_ID", id="old_speedgrader"),
+            param(None, None, "LTI_ID", "LTI_ID", id="regular"),
+            param("USER_ID", "GET_ID", "LTI_ID", "GET_ID", id="new_speedgrader"),
+            param("USER_ID", None, "LTI_ID", "LTI_ID", id="old_speedgrader"),
```
jon-betts, line 31: :+1: Good improvement to specificity here. Took me a couple of reads to try and work out why though!

tests/unit/lms/security_test.py
```diff
@@ -506,6 +510,20 @@ def test_it_returns_LTIUsers_from_OAuth2_state_params(
         canvas_oauth_callback_schema.lti_user.assert_called_once_with()
         assert lti_user == canvas_oauth_callback_schema.lti_user.return_value
 
+    def test_it_returns_LTIUser_from_openid_auth_schema(
+        self,
+        LTI13AuthSchema,
+        lti13_auth_schema,
+        pyramid_request,
```
jon-betts, line 517: Worth trying deleting the trailing comma and running black. It might back it up onto one line?

jon-betts (changes requested):
We had a long and rambling chat about this and plenty of what I'm writing here is actually about that discussion... anywho:

Things I actually suggest we do
-------------------------------

Things we should probably look at (before general non-actionable musings):

 * Should `lms.lti.LTIParams` be in models? Also helps to point out this isn't a new concept.
 * `LTIUser.from_auth_params()` looks like duplicate code and could be re-used in `lms.validation.authentication._launch_params.py`. If so there might be some tests we can clean up too.
 * Try `lti.security._get_lti_params` out in `lms.resources.lti_launch` for size
 * Simplify the inheritance in the schema to be single inheritance with `PyramidRequestSchema` at the bottom
 * Move `_decode_jwt` out of the `PyramidRequestSchema` and into `lms.validation._lti.LTIAuthParamsSchema`
 * Have a think about naming to see if we can reduce confusion a bit - Maybe by introducing V11 in some places to make it obvious when it's a counterpart to a V13 thing etc.

Begin rambling: Too much stuff!
---------------

I think my general bouncing off this PR is actually more to do with bouncing off of LMS itself. There's a lot of the same sorts of pathologies I saw in the refactor of `h` where we have multiple overlapping systems with intertwined responsibilities. The collection of "resources", "predicates", "schema", "services", "request methods" and "view derivers" is quite a lot.

Also, in this context the "schema" are actually in some cases more like services or security policies, and more like resources in others. The request methods substitute for services in some cases and resources in others. The views and resources and predicates all have a hand in similar types of behavior etc.

I think we'd do well to mark some of these concepts as deprecated and see what the minimum span of competing systems is to get things done.

Security related stuff
----------------------

The `lms.validation.authentication.*` family I think are inside out. They are all called something along the lines of `BlahSchema` and happen to have a method called `lti_user()`, but I think the truth is the other way around. They are all `get_lti_user_from_X()` methods, which happen to privately use a schema to get that done. The immutable thing that unites them all is _not_ their use of `marshmallow`, but their purpose to get a user.

This is part of the wider weirdness of the security system, and would probably be tidied up in any serious attempt to refactor that.

This doesn't add to it much, but the following heirarchy is confusing:

* `_lti.LTIAuthParamsSchema`
* `authentication._launch_params.LaunchParamsAuthSchema`
* `authentication._lti.LTI13AuthSchema`

The `lti` / `launch_params` distinction is confusing, but I think this is a case of starting a namespace that makes more sense (LTI) and not having moved everything there yet?

If there's a way of reducing the confusion there though it would be good. Maybe move `authentication._launch_params.LaunchParamsAuthSchema` to `authentication._lti.LTI11AuthSchema`? 

It would also be good to think about what defines `_lti.LTIAuthParamsSchema` and pick a name. Is this `_lti.LTICoreSchema`, what makes it a core schema? All the values are V11 as well. Should we say so?

lms/validation/_lti_launch_params.py
```diff
@@ -1,39 +1,79 @@
...
+)
 from marshmallow.validate import OneOf
 
+from lms.models import LTIParams
 from lms.validation._base import PyramidRequestSchema
 from lms.validation._exceptions import LTIToolRedirect
 
 
-class _CommonLTILaunchSchema(PyramidRequestSchema):
+class LTIV11CoreSchema(PyramidRequestSchema):
```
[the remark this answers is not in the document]
marcospri: Also merged it with  _CommonLTILaunchSchema and now there's only `LTIV11CoreSchema` as a base  class for these

tests/unit/conftest.py
marcospri: As is relatively big for a fixture I think it makes sense to be close to the v11 version but it's also true that it's only used once now.

jon-betts (approved):
Sorry for all the back and forth, but I think there are some decent improvements over what was here before.

 * Having the V11, V13 auth stuff side by side makes much more sense
 * The hierarchy of the schema now makes more sense to me too

There's still larger mess at play, but we can only fix so much while still delivering anything

Issue #3667, opened 2022-02-09
Don't send simultaneous OAuth refresh requests for the same user


seanh:
# Problem

See https://github.com/hypothesis/product-backlog/issues/1230 for the original bug report and how to reproduce the issue locally.

Slack threads:

* <https://hypothes-is.slack.com/archives/C1MA4E9B9/p1644342948547379>
* <https://hypothes-is.slack.com/archives/C4K6M7P5E/p1644338894692669>
* <https://hypothes-is.slack.com/archives/C1MA4E9B9/p1644405136981779>
* <https://hypothes-is.slack.com/archives/C4K6M7P5E/p1644416393843629>

Our backend's API handles OAuth 2 refresh requests in a way that is transparent to the frontend:

1. The frontend sends a request to the backend's API. For example this might be a `/api/canvas/assignments/{resource_link_id}/via_url` request, which requests a public Via URL for the identified Canvas Files assignment.
2. The backend receives this request and sends a request on to the Canvas API's public URL endpoint
3. The backend receives an error response from Canvas indicating that our access token for this user has expired
4. The backend sends a refresh token request to Canvas and gets a new access token
5. The backend re-sends the original request to Canvas's public URL endpoint and this time gets a successful response
6. The backend returns the public Via URL to the frontend

So the frontend just received a successful response from the backend (after some waiting) and is blissfully unaware that a token refresh had to happen.

This solution worked great until the frontend started sending multiple requests to the backend in parallel. For example one request for a public Via URL for the assignment's PDF file and another request to the "sync API" that returns the list of groups to show for this user. The two requests from the frontend cause the backend to send _two_ refresh requests to Canvas at the same time:

1. Request A sends a refresh request to Canvas
2. Request B sends a refresh request to Canvas
3. Request A receives a refreshed access token from Canvas
4. Canvas receives request B's refresh request and invalidates the refreshed access token that it just send to request A
5. Request A re-tries its original Canvas API request using its now-invalided refreshed access token, and gets an error
6. Request A sends an authorization error back to the frontend
7. The frontend shows the "You need to re-authorize" dialog to the user

Request B will have succeeded and will most likely have stored a refreshed, working access token in the DB (although there are race conditions here) but it doesn't matter: if _either_ of the requests fails then the frontend shows the authorization dialog.

# Potential Solutions

One way or another we need to coordinate only sending one refresh request at a time for a given user.

## Solution 1: Have the frontend coordinate refresh requests

[...]

PR #3806, opened 2022-03-25
Remove the frontend_refresh feature flag

robertknight:
Now that it has been tested, enable this behavior by default.

Fixes https://github.com/hypothesis/lms/issues/3667


jon-betts (approved):
Not much to say here. It all looks good. There are no traces of the `frontend_refresh` left as far as I can spot.

PR #3818, opened 2022-03-28
[WIP]Configure assignments without deep-linking

marcospri:


marcospri:
Close in favor of the much simpler approach in https://github.com/hypothesis/lms/pull/3820

PR #3819, opened 2022-03-28
Use tool_consumer_instance_guid from the DB if not present in the request params

marcospri:
`tool_consumer_instance_guid` is an optional parameter in the LTI spec but
required in our app.

In general all LMS do send it as a request parameters so ApplicationInstance.update_lms_data will set the right value on the DB.

For LMS that don't send it (only the LTI1.3 certification tool as of today) we'll allow the request params to be absent but we'll rely on the application_instance row to have the tool_consumer_instance_guid value pre-filled.

### On the removed error handling for accessing application_instance

Having an application instance is fundamental to any LTI launch so `resources/lti_launch` should assume it's already there.

If for any reason it's not there, the exception should bubble up and produce an error instead of disabling some features, and then produce an error.

We are also picking the tool_consumer_guid from the DB instead of the request param.

#### Testing the case where the GUID is not sent.

We'll rely on the functional tests to test this case as no other LMS behaves like this.

#### Testing existing behavior hasn't changed

- Checkout `master`
- https://hypothesis.instructure.com/courses/125/assignments/873, create an annotation there.
- Checkout `simplify-error-handling`
- Reload - https://hypothesis.instructure.com/courses/125/assignments/873, the annotation should still be there.

#### Current behavior in master

- Checkout `master`
- Log in as a teacher in Canvas and edit an assignment, try to configure an assignment and mark it as a groups one selecting any group set.

https://hypothesis.instructure.com/courses/125/assignments/2754/edit?name=TEMP+-+marcos&due_at=null&points_possible=0

- Now remove your application_instances

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from application_instances;"
```

- Try to edit the assignment again
- You'll get an 404 from the backend and a sad H page as a response.
- This is because the line at:

https://github.com/hypothesis/lms/blob/948d2fc8135c077e18a1c7018ac80cedf4629bfd/lms/services/launch_verifier.py#L153

#### New behavior in this branch

- Switch to `simplify-error-handling`
- `make devdata` to get your application instances back
- Again try to configure an assignment as a groups one

https://hypothesis.instructure.com/courses/125/assignments/2754/edit?name=TEMP+-+marcos&due_at=null&points_possible=0

- It should work as expected 
- Delete your application instances

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "delete from application_instances;"
```

- And try configure the assignment again

https://hypothesis.instructure.com/courses/125/assignments/2754/edit?name=TEMP+-+marcos&due_at=null&points_possible=0

- Same behavior as before, a 404 and an error page.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -157,27 +157,30 @@ class ApplicationInstance(BASE):
         """
         Update all the LMS-related attributes present in `params`.
 
-        If the current instance already has a `tool_consumer_instance_guid`
-        report it on logging and don't update any of the columns.
+        :raise ValueError: If tool_consumer_instance_guid is not present in params or the model itself.
+        :raise ReusedConsumerKey: If the tool_consumer_instance_guid present in params
+            is different than the one already recorded in the DB.
         """
 
         tool_consumer_instance_guid = params.get("tool_consumer_instance_guid")
-        if not tool_consumer_instance_guid:
-            # guid identifies the rest of the LMS data, if not there skip any updates
-            return
+        if not tool_consumer_instance_guid and not self.tool_consumer_instance_guid:
+            # We don't have a guid in either the request params or stored on the DB
+            raise ValueError("No GUID present in either request params or the DB")
 
         if (
-            self.tool_consumer_instance_guid
+            tool_consumer_instance_guid
+            and self.tool_consumer_instance_guid
             and self.tool_consumer_instance_guid != tool_consumer_instance_guid
         ):
             # If we already have a LMS guid linked to the AI
-            # and we found a different one report it to sentry
+            # and we've been sent a different one, report it to sentry
             raise ReusedConsumerKey(
                 existing_guid=self.tool_consumer_instance_guid,
                 new_guid=tool_consumer_instance_guid,
             )
 
-        self.tool_consumer_instance_guid = tool_consumer_instance_guid
+        if tool_consumer_instance_guid:
+            self.tool_consumer_instance_guid = tool_consumer_instance_guid
         for attr in [
             "tool_consumer_info_product_family_code",
             "tool_consumer_instance_description",
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -34,11 +34,6 @@ class JSConfig:
     def _h_user(self):
         return self._lti_user.h_user
 
-    @property
-    def _application_instance(self):
-        """Return the current request's ApplicationInstance."""
-        return self._request.find_service(name="application_instance").get_current()
-
     def add_document_url(self, document_url):
         """
         Set the document to the document at the given document_url.
@@ -195,7 +190,7 @@ class JSConfig:
             HTML form that we submit
         """
 
-        args = self._context, self._request, self._application_instance
+        args = self._context, self._request, self._context.application_instance
 
         self._config.update(
             {
@@ -394,7 +389,7 @@ class JSConfig:
         and had grading info recorded for them.
         """
         grading_infos = self._grading_info_service.get_by_assignment(
-            application_instance=self._application_instance,
+            application_instance=self._context.application_instance,
             context_id=self._request.params.get("context_id"),
             resource_link_id=self._request.params.get("resource_link_id"),
         )
@@ -432,7 +427,7 @@ class JSConfig:
         # mutable. You can do self._hypothesis_client["foo"] = "bar" and the
         # mutation will be preserved.
 
-        if not self._application_instance.provisioning:
+        if not self._context.application_instance.provisioning:
             return {}
 
         api_url = self._request.registry.settings["h_api_url_public"]
@@ -527,7 +522,8 @@ class JSConfig:
             return self._canvas_sync_api()
 
         if (
-            self._application_instance.product == ApplicationInstance.Product.BLACKBOARD
+            self._context.application_instance.product
+            == ApplicationInstance.Product.BLACKBOARD
             and self._context.is_blackboard_group_launch
         ):
 
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -4,7 +4,6 @@ import logging
 
 from lms.models import LTIParams
 from lms.resources._js_config import JSConfig
-from lms.services import ApplicationInstanceNotFound
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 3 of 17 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -35,7 +35,7 @@ def _h_user(self):
         return self._lti_user.h_user
 
     @property
-    def _application_instance(self):
+    def application_instance(self):
```
marcospri, line 38: Removing the underscore as this is now used from `lti_launch.py`

lms/resources/lti_launch.py
```diff
@@ -24,21 +23,17 @@ def __init__(self, request):
         """Return the context resource for an LTI launch request."""
         self._request = request
         self._authority = self._request.registry.settings["h_authority"]
-        self._application_instance_service = self._request.find_service(
```
marcospri, line 27: No longer needed.

lms/resources/lti_launch.py
```diff
@@ -24,21 +23,17 @@ def __init__(self, request):
...
         """Get the course this LTI launch based on the request's params."""
         course_service = self._request.find_service(name="course")
         params = self._request.parsed_params
 
-        tool_consumer_instance_guid = params["tool_consumer_instance_guid"]
         context_id = params["context_id"]
 
         authority_provided_id = course_service.generate_authority_provided_id(
-            tool_consumer_instance_guid, context_id
+            self.js_config.application_instance.tool_consumer_instance_guid, context_id
```
marcospri, line 36: This change is in a separate commit.

Instead of using the parameter from the request use it directly form the application instance which is now guaranteed to have the same one:


https://github.com/hypothesis/lms/blob/5140598c77655437695d09166fc7d7947043434e/lms/models/application_instance.py#L175


This simplifiies things a bit for LTI1.3

jon-betts: I know that the `js_config` happens to have an application instance, but what has the application instance got to do with `js_config`? Are we just using this because it happens to have one?

Seems a bit weird.

lms/resources/lti_launch.py
```diff
@@ -178,12 +173,7 @@ def canvas_sections_supported(self):
             # Canvas course sections feature was released.
             return False
 
-        try:
-            application_instance = self._application_instance_service.get_current()
-        except ApplicationInstanceNotFound:
-            return False
-
-        return bool(application_instance.developer_key)
+        return bool(self.js_config.application_instance.developer_key)
```
jon-betts, line 176: Does this explode with an attribute error now if `application_instance` is None? How is this handled?

jon-betts (commented):
I have a couple of questions about this before I review it. Which is mostly:

 * Why do we access the application instance via `js_config`? At first blush, it doesn't seem to be particularly related to it?
 * Do we just get attribute errors now?

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -146,11 +144,9 @@ def test_it(self, bearer_token_schema, context, grant_token_service, js_config):
             "rpcServer": {"allowedOrigins": ["http://localhost:5000"]},
         }
 
-    def test_it_raises_if_theres_no_ApplicationInstance(
-        self, application_instance_service, js_config
-    ):
-        application_instance_service.get_current.side_effect = (
-            ApplicationInstanceNotFound
+    def test_it_raises_if_theres_no_ApplicationInstance(self, context, js_config):
+        type(context).application_instance = mock.PropertyMock(
```
marcospri, line 148: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock

```Because of the way mock attributes are stored you can’t directly attach a PropertyMock to a mock object. Instead you can attach it to the mock type object:

```

marcospri:
> I have a couple of questions about this before I review it. Which is mostly:
> 
> * Why do we access the application instance via `js_config`? At first blush, it doesn't seem to be particularly related to it?


No, you are right, it was just there. I moved it to resources/lti_launch where I think it does belong.

> * Do we just get attribute errors now?

Not having an application_instance is not something we should expect in an LTI launch. That's checked first check in for example: https://github.com/hypothesis/lms/blob/2ad49c405fd62091a758d4274c4dc0c8123a6e5a/lms/validation/authentication/_launch_params.py#L64

I we went that far without failing I think we should fail loudly and seeing an "ApplicationInstanceNotFound" from `.get_current()` in sentry would be a good starting point for debugging. 

I haven't been able to trigger the exception in any of those on `master` without first failing somewhere else.

lms/resources/_js_config/__init__.py
```diff
@@ -34,11 +34,6 @@ def __init__(self, context, request):
     def _h_user(self):
         return self._lti_user.h_user
 
-    @property
```
marcospri, line 37: Moved to `restouces/lti_launch`

lms/resources/lti_launch.py
```diff
@@ -25,21 +24,22 @@ def __init__(self, request):
...
+    def application_instance(self):
+        """Return the current request's ApplicationInstance."""
+        return self._request.find_service(name="application_instance").get_current()
+
     def get_or_create_course(self):
         """Get the course this LTI launch based on the request's params."""
         course_service = self._request.find_service(name="course")
         params = self._request.parsed_params
 
-        tool_consumer_instance_guid = params["tool_consumer_instance_guid"]
```
marcospri, line 38: Prefer the value from the DB instead of the request params

lms/resources/lti_launch.py
```diff
@@ -179,12 +179,7 @@ def canvas_sections_supported(self):
             # Canvas course sections feature was released.
             return False
 
-        try:
```
marcospri, line 182: No need to check for application instance existence here. We'd fail much sooner and, if for reason, we found that there's no AI at this point we rather raise an exception (from the AI service) that hide the error.

lms/models/application_instance.py
```diff
@@ -157,21 +157,23 @@ def update_lms_data(self, params: dict):
...
+            is different than the one already recorded in the DB.
         """
 
         tool_consumer_instance_guid = params.get("tool_consumer_instance_guid")
-        if not tool_consumer_instance_guid:
-            # guid identifies the rest of the LMS data, if not there skip any updates
-            return
+        if not tool_consumer_instance_guid and not self.tool_consumer_instance_guid:
+            # We don't have a guid in either the request params or stored on the DB
+            raise ValueError("No GUID present in either request params or the DB")
```
marcospri, line 168: This is not a situation we've seen in any LMS yet so raising an unhanded exception will point us in the right direction if it ever happens.

PR #3828, opened 2022-04-01
JWT helper functions moved to JWTService

marcospri:
Moving existing `jwt` usage to its own service in preparation for including the LTI1.3 new bits in the same service.

### Bearer token

  The authorization header used for `http://localhost:8001/api/canvas/sync` includes a jwt encoded/decoded with the new service.

#### OAuthCallbackSchema

- Delete all the tokens to trigger an authorization 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"```

The state param sent to `https://hypothesis.instructure.com/login/oauth2/auth?client_id` is encoded/decoded with the existing service.

### Test migration for existing session/tokens

This might be overkill testing but just to prove we decode the tokens created in `master`.

#### BearerToken

- Switch to `master`, and start configuring one assignment. Stop when seeing the different buttons.  

https://hypothesis.instructure.com/courses/125/assignments/2754/edit?name=TEMP+-+marcos&due_at=null&points_possible=0

- Switch to this branch and click the "Select PDF from canvas"
- The authorization JWT sent to http://localhost:8001/api/canvas/courses/125/files created in master has been decoded in this branch.

#### OAuthCallbackSchema

- Switch to `master`, and start configuring one assignment. Stop when seeing the different buttons.

- Delete all the tokens to trigger an authorization 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token;"```

- Launch https://hypothesis.instructure.com/courses/125/assignments/873, click authorize and leave the popup window open, don't authorize there yet.

Check the query params  for `https://hypothesis.instructure.com/login/oauth2/auth`, the `state` param is an encoded JWT.

- Switch to this branch; `jwt-service`

Click "authorize". The authorization will be successful using the encoded JWT from master.

### CODE REVIEW

File: lms/services/jwt.py [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+        payload = copy.deepcopy(payload)
+        payload["exp"] = datetime.datetime.utcnow() + lifetime
+
+        jwt_str = jwt.encode(payload, secret, algorithm="HS256")
+
+        return jwt_str
+
+
+def factory(_context, _request):
+    return JWTService()
```
marcospri, line 59: This is a silly factory but I think we should keep the pattern still. If we need to make it more complex we'll just have to change the factory like in any other service.

jon-betts: If we don't need a service factory then it's more efficient not to do it. The default way you are supposed to use `pyramid-services` is to just register an instance the once, which is then re-used on all requests. 

Our default is backwards and we've adopted the per request kind without usually thinking about it. I don't _think_ this was a conscious choice, and we don't have to try and justify it. I think the first few were done this way, and then the rest were copied.

We should probably be moving the other way for those services that don't require it, and then only move towards per-request services when we really have to. If all we need is to read some values from the config, it's not required to read it from the request. We can see an example of this in Checkmate:

 * https://github.com/hypothesis/checkmate/blob/main/checkmate/services/__init__.py
 
My gut feel is a lot of the services which we repeatedly re-create, actually only need to be created once at worker startup.

tests/unit/lms/services/jwt_test.py
```diff
@@ -107,3 +109,20 @@ def test_it_uses_custom_lifetime(self):
...
+class TestFactory:
+    def test_it(self, pyramid_request, JWTService):
+        jwt_service = factory(sentinel.context, pyramid_request)
+
+        JWTService.assert_called_once()
+        assert jwt_service == JWTService.return_value
+
+    @pytest.fixture
+    def JWTService(self, patch):
+        return patch("lms.services.jwt.JWTService")
```
jon-betts, line 128: This all seems like a lot of test for not very much code. If you count out the comments there's not very much going on in that file. I've had a go at rewriting the first set of tests to trim down on the volume.

Some rules of thumbs I've noticed:

 * You ideally want your tests to be ~2x the length of the code
 * Above 3x gets really hard to maintain
 * For example, in this code it's hard to spot that we have the same test twice (take a look!)

Here's my attempt:

```python
class TestDecodeWithSecret:
    def test_it_can_decode_a_jwt_from_encode_jwt(self, svc):
        original_payload = {"key": "value"}
        jwt_str = svc.encode_with_secret(original_payload, "test_secret")

        decoded_payload = svc.decode_with_secret(jwt_str, "test_secret")

        assert decoded_payload == original_payload

    @pytest.mark.parametrize(
        "error,kwargs",
        (
            param(InvalidJWTError, {"secret": "WRONG"}, id="wrong_secret"),
            param(InvalidJWTError, {"algorithm": "HS384"}, id="wrong_algorithm"),
            param(InvalidJWTError, {"lifetime": None}, id="no_expiry"),
            param(
                ExpiredJWTError,
                {"lifetime": timedelta(hours=-1)},
                id="already_expired",
            ),
        ),
    )
    def test_it_raises_for_bad_jwts(self, svc, error, kwargs):
        jwt_str = self.encode_jwt({"key": "value"}, **kwargs)

        with pytest.raises(error):
            svc.decode_with_secret(jwt_str, "test_secret")

    def encode_jwt(
        self,
        payload,
        lifetime=timedelta(hours=1),
        secret="test_secret",
        algorithm="HS256",
    ):
        """Return payload encoded to a jwt with secret and algorithm."""
        payload = copy.deepcopy(payload)

        if lifetime:
            payload["exp"] = datetime.utcnow() + lifetime

        return jwt.encode(payload, secret, algorithm=algorithm)
```

We can also see the bigness is begetting more bigness. As the tests had run so long, they were split in two which (very slightly) increases the size again. If we can get the length under-control they could probably be one test class with `test_decode_with_secret_*` instead of `test_it_` again.

The above is a very straight refactor of the existing tests, but there's also an element to which we seem to be testing the underlying `jwt` library more than our own code. I think if some of the tests used a patched `jwt` library they could be cut down further to a few intergrated tests (like these) and probably one `assert_called_once_with` etc. You'd have to suck it an see though; it could make it worse.

tests/unit/lms/validation/authentication/_bearer_token_test.py
```diff
@@ -1,36 +1,38 @@
...
     MissingSessionTokenError,
 )
-from tests import factories
 
 
 class TestBearerTokenSchema:
-    def test_it_serializes_lti_users_into_bearer_tokens(self, lti_user, schema, _jwt):
+    def test_it_serializes_lti_users_into_bearer_tokens(
+        self, lti_user, schema, jwt_service
+    ):
```
jon-betts, line 19: The use of `test_it` is a bit confusing on an object with multiple methods and ways of using it

tests/unit/lms/validation/authentication/_bearer_token_test.py
```diff
@@ -103,23 +107,14 @@ def schema(self, pyramid_request):
...
-
+def jwt_service(jwt_service, lti_user):
+    jwt_service.decode_with_secret.return_value = lti_user._asdict()
+    jwt_service.encode_with_secret.return_value = "ENCODED_JWT"
 
-@pytest.fixture
-def lti_user():
-    """Return the original LTIUser that was encoded as a JWT in the request."""
-    return factories.LTIUser()
+    return jwt_service
```
jon-betts, line 120: Some of the tests in this file don't seem like they are testing this class / are duplicates.

* `test_serialize_and_deserialize_via_marshmallow_api` - Seems like a test of our base class
* `test_parse_via_webargs_api` - Seems like a test of marshmallow / a duplicate of the first test

There's also a chunk of code which very nearly the same in 4 tests which is:

```python
        with pytest.raises(error) as exc_info:
            schema.lti_user("headers")

        assert exc_info.value.messages == {
            "headers": {field: [message]},
        }
```

A small helper there + some parameterisation might go a long way to shortening the test.

tests/unit/lms/validation/authentication/_bearer_token_test.py
```diff
@@ -103,23 +107,14 @@ def schema(self, pyramid_request):
...
 @pytest.fixture(autouse=True)
-def _jwt(patch, lti_user):
-    _jwt = patch("lms.validation.authentication._bearer_token._jwt")
-    _jwt.encode_jwt.return_value = "ENCODED_JWT"
-    _jwt.decode_jwt.return_value = lti_user._asdict()
-    return _jwt
-
+def jwt_service(jwt_service, lti_user):
+    jwt_service.decode_with_secret.return_value = lti_user._asdict()
+    jwt_service.encode_with_secret.return_value = "ENCODED_JWT"
```
jon-betts, line 118: I know it's a point of contention, but when a fixture is only used in one place I prefer it to be scoped as close to that place as possible. This is partly a general preference for local vs. global, but also for the nuts and bolts reason that it makes refactoring in future much easier.

If you have 4 test classes and a soup of fixtures at the bottom, it's more difficult to work out what's shared and what's not to break them up. If the fixtures are part of the classes they are used in, then it's easy to snip out the whole test class.

jon-betts (commented):
I think there's a bit of a decision point to be made here. The tests and some of the code is sort of inherited from before, but it's also not as good as it could be.

The tests are quite long winded, and in some cases seem to test classes and functionality below the class in question. In general I think these tests could do with a tidy and shortening pass. As a pretty good sign they are too long, quite a few include basically the same test twice, but it's hard to spot.

A very different concern I have is that we have a JWT service which is reasonably generic, but we're still doing JWT things elsewhere in the code. e.g.

* `lms.extensions.feature_flags._helpers`
* `lms.services.grant_token`

The grant token stuff in particular seems like it would be very easy to fold in.

## Decision time!

It seems like there are a few possible futures here:

 * Tackle test refactoring in this PR (the code has been changed a bit anyway, so why not just go the whole way?)
 * Tackle the test refactoring in another PR
 * Tackle unification in another PR (one for each service?)
 * Don't bother tackling unification - but also lose the benefit of a central service

lms/services/jwt.py
```diff
@@ -0,0 +1,59 @@
+import copy
+import datetime
+
+import jwt
+
+from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
+
+DEFAULT_LIFETIME = datetime.timedelta(hours=1)
```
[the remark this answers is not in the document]
marcospri: Yeah, one of the usage relies of the default but it's probably better to explicit pass it in both.

Having the default maybe implies that is the right value for every occasion.

tests/unit/lms/services/jwt_test.py
```diff
@@ -0,0 +1,80 @@
...
+    ):
+        """Return payload encoded to a jwt with secret and algorithm."""
+        payload = copy.deepcopy(payload)
+
+        payload["exp"] = datetime.datetime.utcnow() + lifetime
+
+        return jwt.encode(payload, secret, algorithm=algorithm)
+
+    @freeze_time("2022-04-04")
+    def test_encode_with_secret(self, svc, jwt):
```
marcospri, line 60: Here I replaced all of the encoding jwt test with this one:

- `test_it_can_decode_a_jwt_from_encoded_jwt` already tests the encode <-> decode full flow.
- Here only asserting how we use the `jwt` library.
- Using this `freeze_time` library. Maybe a very small usage to warrant a new dependency?
 
I found that when time is involved what freeze gun does is what I want to do.

marcospri:
> * Tackle test refactoring in this PR (the code has been changed a bit anyway, so why not just go the whole way?)

I tried to refactor the JWTService ones (as is the new service added in the PR following your example. I originally just mostly copy pasted the original ones.

For others I just lightly touched (eg `_bearer_token_test.py`). I've been much lazier.


> * Tackle unification in another PR (one for each service?)

I think I'd prefer to go this route for other usages (that I missed before). In retrospect I should have started the service with the LTI1.3 bits and do a unification PR later.

tests/unit/lms/validation/authentication/_bearer_token_test.py
```diff
@@ -86,40 +89,19 @@ def test_it_raises_if_the_user_id_param_is_missing(self, schema, _jwt, param):
             "headers": {param: ["Missing data for required field."]},
         }
 
-    def test_serialize_and_deserialize_via_marshmallow_api(self, lti_user, schema):
```
marcospri, line 89: In this file I haven't done any effort to refactor the tests.

I could not resist to remove these tests you spotted where we seem to be testing something else. We seem to have quite a few of these.

tests/unit/lms/services/jwt_test.py
jon-betts: I think we should give this a shot, and maybe separately look for tests which could be simplified by using this. I can't think of any off the top of my head, but I know we've ended up with all sorts of nonsense with fixtures for `now` and patching out `datetime` and things like that.

tests/unit/lms/services/jwt_test.py
```diff
@@ -0,0 +1,80 @@
...
+        lifetime,
+        secret="test_secret",
+        algorithm="HS256",
+    ):
+        """Return payload encoded to a jwt with secret and algorithm."""
+        payload = copy.deepcopy(payload)
+
+        payload["exp"] = datetime.datetime.utcnow() + lifetime
+
+        return jwt.encode(payload, secret, algorithm=algorithm)
```
jon-betts, line 57: Maybe pop this after all the tests. My brain tends to see the first private function or fixture and think "tests over".

jon-betts (approved):
Yeah I think this is looking really good. The JWT tests ended up _so_ much better than they started and kicked 60 lines out at the same time.

I'd be interested at taking a run at simplifying the other tests here too.

PR #3836, opened 2022-04-06
LTI1.3 core certification func tests

marcospri:
Functional tests for the LTI1.3 core certification.

I kept a 1 to 1 relation with the cases in the certification tool.

- For the the first ones, the bad payloads ones, the tools runs the same tests with two different payloads.
- For the student/teacher the tests are very similar but the tool has two separate suites, ones for teachers and one for students. Kept the same split here.

The same thing it's true about some individual tests that could be parameterized version of another one but kept them separated as the certification tool does.

This PR marks every new tests with and `xfail` until other related PRs are merged but we could focus here on the testing/fixtures etc.


tests/functional/conftest.py
```diff
@@ -68,3 +69,23 @@ def db_session(db_engine):
         factories.clear_sqlalchemy_session()
         session.close()
         conn.close()
+
+
+@pytest.fixture
+def do_lti_launch(app):
```
marcospri, line 75: Moved from `tests/functional/views/basic_lti_launch_test.py`

tests/functional/lti_certification/v13/conftest.py
```diff
@@ -0,0 +1,173 @@
...
+        return json.load(key)
+
+
+@pytest.fixture
+def jwt_headers():
+    return {"kid": JWT_KID}
+
+
+@pytest.fixture
+def make_jwt(jwt_headers, jwt_private_key):
```
marcospri, line 33: `make_jwt` uses our `jwt_private_key` to sign the JWTs

We have `jwk_endpoint_intercept` that would use the public key to decode them.

tests/functional/lti_certification/v13/conftest.py
```diff
@@ -0,0 +1,173 @@
...
+        },
+        "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint": {
+            "scope": [
+                "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem",
+                "https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly",
+                "https://purl.imsglobal.org/spec/lti-ags/scope/score",
+            ],
+            "lineitems": "https://ltiadvantagevalidator.imsglobal.org/ltitool/rest/assignmentsgrades/10843/lineitems",
+        },
+    }
```
jon-betts, line 173:
```suggestion
@pytest.fixture
def student_payload(common_payload):
    return {
        "sub": "STUDENT_ID",
        "nonce": "d45a2497-4389-45ea-8d61-b1fa7e407447",
        "name": "STUDENT_FIRST_NAME STUDENT_MIDDLE_NAME STUDENT_LAST_NAME",
        "given_name": "STUDENT_FIRST_NAME",
        "family_name": "STUDENT_LAST_NAME",
        "middle_name": "STUDENT_MIDDLE_NAME",
        "email": "student@email.com",
        "https://purl.imsglobal.org/spec/lti/claim/roles": [
            "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner"
        ],
        **common_payload,
    }


@pytest.fixture
def teacher_payload(common_payload):
    return {
        "sub": "TEACHER_ID",
        "nonce": "6f889151-c368-4e32-91af-5fe831ee266b",
        "name": "TEACHER_FIRST_NAME TEACHER_MIDDLE_NAME TEACHER_LAST_NAME",
        "given_name": "TEACHER_FIRST_NAME",
        "family_name": "TEACHER_LAST_NAME",
        "middle_name": "TEACHER_MIDDLE_NAME",
        "email": "teacher@email.com",
        "https://purl.imsglobal.org/spec/lti/claim/roles": [
            "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor"
        ],
        **common_payload,
    }


@pytest.fixture
def common_payload():
    # This is the parts of the payload common to both teacher and student
    now = int(datetime.timestamp(datetime.now()))

    return {
        "exp": now + 60,
        "iat": now,
        "iss": "https://ltiadvantagevalidator.imsglobal.org",
        "aud": "imstester_4ba76ab",
        "locale": "en-US",
        "https://purl.imsglobal.org/spec/lti/claim/target_link_uri": "https://localhost/lti_launches",
        "https://purl.imsglobal.org/spec/lti/claim/deployment_id": "testdeploy",
        "https://purl.imsglobal.org/spec/lti/claim/message_type": "LtiResourceLinkRequest",
        "https://purl.imsglobal.org/spec/lti/claim/version": "1.3.0",
        "https://purl.imsglobal.org/spec/lti/claim/context": {
            "id": "COURSE_ID",
            "label": "COURSE_LABEL",
            "title": "COURSE_TITLE",
            "type": ["http://purl.imsglobal.org/vocab/lis/v2/course#CourseSection"],
        },
        "https://purl.imsglobal.org/spec/lti/claim/resource_link": {
            "id": "RESOURCE_ID",
            "title": "Introduction Assignment",
            "description": "This is the introduction assignment",
        },
        "https://purl.imsglobal.org/spec/lti-nrps/claim/namesroleservice": {
            "context_memberships_url": "https://ltiadvantagevalidator.imsglobal.org/ltitool/namesandroles.html?memberships=10843",
            "service_versions": ["2.0"],
        },
        "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint": {
            "scope": [
                "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem",
                "https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly",
                "https://purl.imsglobal.org/spec/lti-ags/scope/score",
            ],
            "lineitems": "https://ltiadvantagevalidator.imsglobal.org/ltitool/rest/assignmentsgrades/10843/lineitems",
        },
    }
```

Kind of giant suggestion, but nearly everything about these two fixtures is the same. Not only does this mean we have a lot of duplication, but it's also hard to spot what is different about them or not.

tests/functional/lti_certification/v13/core/test_bad_payloads.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        del jwt_headers["kid"]
+
+        response = do_lti_launch(
+            {"id_token": make_jwt(test_payload, jwt_headers)}, status=403
+        )
+
+        assert response.html
+
+    @pytest.mark.xfail(reason="JWT verification disabled")
```
jon-betts, line 26: Double fail. Fail harder!

tests/functional/conftest.py
```diff
@@ -68,3 +69,23 @@ def db_session(db_engine):
...
+            url,
+            params=post_params,
+            headers={
+                "Accept": "text/html",
+                "Content-Type": "application/x-www-form-urlencoded",
+            },
+            **kwargs,
+        )
+
+    return _do_lti_launch
```
jon-betts, line 91:
```suggestion
    def do_lti_launch(post_params, get_params=None, **kwargs):
        url = "/lti_launches"
        if get_params:
            url += f"?{urlencode(get_params)}"

        return app.post(
            url,
            params=post_params,
            headers={
                "Accept": "text/html",
                "Content-Type": "application/x-www-form-urlencoded",
            },
            **kwargs,
        )

    return do_lti_launch
```

I think our "standard" is to have the inner be the same name as the outer. I couldn't fully tell you why. I think this is to prevent people from "choosing any name they like". I'm not sure what the consequence would be if they did though.

tests/functional/lti_certification/v13/conftest.py
```diff
@@ -0,0 +1,173 @@
...
+        tool_consumer_instance_guid="TEST_CONSUMER_INSTANCE_GUID",
+        lti_registration=lti_registration,
+        deployment_id="testdeploy",
+    )
+
+
+@pytest.fixture(autouse=True)
+def jwk_endpoint_intercept(lti_registration, jwt_public_key):
+    jwt_public_key["kid"] = JWT_KID
+    jwk = {"keys": [jwt_public_key]}
```
jon-betts, line 65: This part is a little confusing to me. If we want `jwt_public_key` to have a fixed `kid` why don't we do it in the `jwt_public_key` fixture? Does it sometimes need to not have this value?

Also, if we want it to match the value in the headers it might be nice to read one from the other. For example, the headers could be:

```python
@pytest.fixture
def jwt_headers(jwt_public_key):
    return {"kid": jwt_public_key["kid"]}
```

Which would:

 * Explain that these two are supposed to line up explicitly
 * Remove the need for the `JWT_KID` variable, as it would be explicitly set inside `jwt_public_key`

jon-betts (commented):
This is very hard to review for a few reasons:

 * Everything is skipped, so I've no idea if it works or not
 * I'm not sure what the tool does
 
I'm not really sure how I'd check that these represent the LTI tests?

tests/functional/lti_certification/v13/core/test_bad_payloads.py
```diff
@@ -0,0 +1,149 @@
...
+        ]
+        for missing_claim in missing_claims:
+            del test_payload[missing_claim]
+
+        response = do_lti_launch({"id_token": make_jwt(test_payload)}, status=403)
+
+        assert response.status_code == 403
+        assert response.html
+
+    @pytest.mark.xfail(reason="JWT verification disabled")
```
jon-betts, line 81:
```suggestion
    @pytest.mark.xfail(reason="JWT verification disabled", strict=True)
```
?

tests/functional/lti_certification/v13/core/test_valid_student_launches.py
```diff
@@ -0,0 +1,104 @@
...
+        del student_payload["name"]
+        del student_payload["email"]
+        del student_payload["given_name"]
+        del student_payload["family_name"]
+        del student_payload["middle_name"]
+
+        js_config = assert_launch_get_config(student_payload)
+        assert js_config["mode"] == JSConfig.Mode.BASIC_LTI_LAUNCH
+
+    @pytest.mark.xfail(reason="Pending. Context is required in our schemas")
```
jon-betts, line 84:
```suggestion
    @pytest.mark.xfail(reason="Pending. Context is required in our schemas", strict=True)
```
?

tests/functional/lti_certification/v13/core/test_bad_payloads.py
```diff
@@ -0,0 +1,149 @@
...
+
+        response = do_lti_launch(
+            {"id_token": make_jwt(test_payload, jwt_headers)}, status=403
+        )
+
+        assert response.html
+
+    @pytest.mark.xfail(reason="JWT verification disabled")
+    def test_incorrect_kid_in_jwt_header(self):
+        """The KID provided is incorrect (and signing verification is impossible)"""
```
jon-betts, line 28:
```suggestion
    @pytest.mark.xfail(reason="JWT verification disabled", strict=True)
    def test_incorrect_kid_in_jwt_header(self):
        """The KID provided is incorrect (and signing verification is impossible)"""
        
        raise NotImplementedError()
```

Why doesn't this test have a body?

jon-betts (approved):
There are a couple of bits here I'm not sure about in some suggestions, but I think this is a good enough base to proceed on and then start fixing it up.

It's not really logically possible to say if these are correct until we can see them work at least once. So the main value here is in breaking up the PRs a bit I think.

tests/functional/lti_certification/v13/core/test_bad_payloads.py
marcospri: Oh I was missing the payload from the tool. It also blew up when I added `strict=True`

PR #3837, opened 2022-04-06
Support for LTI1.3 single tenant flow

marcospri:
The single tenant vs multi tenant distinction in the LTI spec is supported in our data model where:
 - single tenant: 1 to 1 relationship between LTIRegistration.issuer <-> ApplicationInstance
 - multi tenant: 1 to many relationship between them.

All LMS I've tried use the multi tenant setup (and the spec recommends it) but the certification tool uses the single tenant model and, as there's only one registration -> application instance per "issuer" it doesn't seen the `client_id` value on the OIDC request.

For that case we have to make `client_id` optional and take it form the DB.

### Setup

- `make shell`
```python
db.add(
    models.LTIRegistration(
        issuer="https://canvas.instructure.com", 
        client_id="93820000000000200", 
        auth_login_url="https://canvas.instructure.com/api/lti/authorize_redirect",                 
        key_set_url="https://canvas.instructure.com/api/lti/security/jwks", 
        token_url="ttps://canvas.instructure.com/api/lti/authorize"
    )
)

tm.commit()
```

```python
ai = db.query(models.ApplicationInstance).order_by(models.ApplicationInstance.id.desc()).first()

ai.lti_registration_id = db.query(models.LTIRegistration).one().id
ai.deployment_id = "609:6ed5aff92b5257ab607ec9c629844ea8586fde68"

tm.commit()
```

You can launch https://hypothesis.instructure.com/courses/319/assignments/2781 and the OIDC flow should work correctly.

Copy the `http://localhost:8001/lti/1.3/oidc` request made by canvas as curl and remove the `client_id` parameter to end up with something like:

```
curl 'http://localhost:8001/lti/1.3/oidc' \
  -H 'Connection: keep-alive' \
  -H 'Cache-Control: max-age=0' \
  -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="99", "Google Chrome";v="99"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Linux"' \
  -H 'Upgrade-Insecure-Requests: 1' \
  -H 'Origin: null' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.84 Safari/537.36' \
  -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9' \
  -H 'Sec-Fetch-Site: cross-site' \
  -H 'Sec-Fetch-Mode: navigate' \
  -H 'Sec-Fetch-Dest: iframe' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  --data-raw 'iss=https%3A%2F%2Fcanvas.instructure.com&login_hint=969fc294ee26e3f1d0c9714af3351dc4eacfd6df&target_link_uri=http%3A%2F%2Flocalhost%3A8001%2Flti_launches%3Furl%3Dhttps%253A%252F%252Felpais.es&lti_message_hint=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ2ZXJpZmllciI6IjhhZGU2OGIzMzE0ZTVjYjQwNzk4ZGFkNjgzYTAwM2I0OGI3OWM4OGY1MWEzYTUyNWFlODQ0YTU3MTZiMmY4ODgwOTc5MmRiMjhiNjBkMmZlNDkyODZmMzdjMmIwMjc1OWQ3ZWI0NjYyODRkZjQ5MGZmY2RjOGIyNGUxYTRjZDE4IiwiY2FudmFzX2RvbWFpbiI6Imh5cG90aGVzaXMuaW5zdHJ1Y3R1cmUuY29tIiwiY29udGV4dF90eXBlIjoiQ291cnNlIiwiY29udGV4dF9pZCI6OTM4MjAwMDAwMDAwMDAzMTksImNhbnZhc19sb2NhbGUiOiJlbiIsImV4cCI6MTY0OTI0MzAxMX0.xUBYax3t2Via3mftAHi9Wh4DLjJLRJYe3xZIjCJwev4&canvas_region=us-east-1' \
  --compressed
```

which will return a 302 in this branch but a 403 in `master`.

No LMS use this model so it's no possible to test it direclty.

### DIFF
--- a/lms/services/lti_registration.py
+++ b/lms/services/lti_registration.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from lms.models import LTIRegistration
 
 
@@ -5,18 +7,25 @@ class LTIRegistrationService:
     def __init__(self, db):
         self._db = db
 
-    def get(self, issuer, client_id):
+    def get(self, issuer: str, client_id: Optional[str] = None):
         """
-        Get a LTIRegistration based on the uniqueness of issuer + client_id.
+        Get an LTIRegistration based on issuer and client_id.
 
-        - issuer is provided by the platform (ie the LMS) and is generally a URL identifying the LMS.
-        - client_id is also provided by the LMS and it should be unique within the issuer.
+        For LMSs that support single tenant we'll have only one (issuer,
+        client_id) pair and the OIDC request might not provide the client_id so
+        the query will be based on issuer only.
+
+        :param issuer: provided by the platform (ie the LMS) and is generally
+            a URL identifying the LMS
+        :param client_id: provided by the LMS, should be unique within the
+            issuer
         """
-        return (
-            self._db.query(LTIRegistration)
-            .filter_by(issuer=issuer, client_id=client_id)
-            .one_or_none()
-        )
+        query = self._db.query(LTIRegistration).filter_by(issuer=issuer)
+
+        if client_id:
+            query = query.filter_by(client_id=client_id)
+
+        return query.one_or_none()
 
 
 def factory(_context, request):
--- a/lms/views/lti/oidc.py
+++ b/lms/views/lti/oidc.py
@@ -40,7 +40,7 @@ class OIDCRequestSchema(PyramidRequestSchema):
     location = "query_and_form"
 
     iss = fields.Str(required=True)
-    client_id = fields.Str(required=True)
+    client_id = fields.Str(required=False)
 
     target_link_uri = fields.Str(required=True)
     login_hint = fields.Str(required=True)
@@ -54,13 +54,12 @@ class OIDCRequestSchema(PyramidRequestSchema):
 )
 def oidc_view(request):
     params = request.parsed_params
+    issuer, client_id = params["iss"], params.get("client_id")
 
-    registration = request.find_service(LTIRegistrationService).get(
-        issuer=params["iss"], client_id=params["client_id"]
-    )
+    registration = request.find_service(LTIRegistrationService).get(issuer, client_id)
     if not registration:
         raise HTTPForbidden(
-            f"""Registration not found. iss:'{params["iss"]}', client_id:'{params["client_id"]}'"""
+            f"""Registration not found. iss:'{issuer}', client_id:'{client_id}'"""
         )
 
     params = {
@@ -72,7 +71,9 @@ def oidc_view(request):
         "prompt": "none",
         "login_hint": params["login_hint"],
         "lti_message_hint": params["lti_message_hint"],
-        "client_id": params["client_id"],
+        # We might not have been provided the client id in the request, so use
+        # the client id from the registration
+        "client_id": registration.client_id,
         "state": uuid.uuid4().hex,
         "nonce": uuid.uuid4().hex,
         # The value will be configured when registering the tool in the LMS and
--- a/tests/unit/lms/services/lti_registration_test.py
+++ b/tests/unit/lms/services/lti_registration_test.py
@@ -10,6 +10,9 @@ class TestLTIRegistrationService:
     def test_get(self, svc, registration):
         assert svc.get(registration.issuer, registration.client_id) == registration
 
+    def test_without_client_id(self, svc, registration):
+        assert svc.get(registration.issuer) == registration
+
     @pytest.fixture
     def registration(self):
         return factories.LTIRegistration()
--- a/tests/unit/lms/views/lti/oidc_test.py
+++ b/tests/unit/lms/views/lti/oidc_test.py
@@ -16,6 +16,7 @@ class TestOIDC:
         lti_registration.get.return_value.auth_login_url = (
             "https://lms.com/auth_login_url"
         )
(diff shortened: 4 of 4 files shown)


lms/services/lti_registration.py
```diff
@@ -1,22 +1,28 @@
...
-        Get a LTIRegistration based on the uniqueness of issuer + client_id.
+        Get a LTIRegistration based on issuer and client_id.
+
+        - `issuer` is provided by the platform (ie the LMS) and is generally a URL identifying the LMS.
+        - `client_id` is also provided by the LMS and it should be unique within the issuer.
 
-        - issuer is provided by the platform (ie the LMS) and is generally a URL identifying the LMS.
-        - client_id is also provided by the LMS and it should be unique within the issuer.
+        For LMS that support single tenant we'll have only one (issuer, client_id) pair
+          and the OIDC request might not provide the client_id so the query will be based on issuer only.
```
jon-betts, line 18: Any reason not to use standard docstring formatting for the params here?

```python
        """
        Get an LTIRegistration based on issuer and client_id.

        For LMSs that support single tenant we'll have only one (issuer,
        client_id) pair and the OIDC request might not provide the client_id so
        the query will be based on issuer only.

        :param issuer: provided by the platform (ie the LMS) and is generally
            a URL identifying the LMS
        :param client_id: provided by the LMS, should be unique within the 
            issuer
        """
```

lms/views/lti/oidc.py
```diff
@@ -56,11 +56,11 @@ def oidc_view(request):
     params = request.parsed_params
 
     registration = request.find_service(LTIRegistrationService).get(
-        issuer=params["iss"], client_id=params["client_id"]
+        issuer=params["iss"], client_id=params.get("client_id")
     )
     if not registration:
         raise HTTPForbidden(
-            f"""Registration not found. iss:'{params["iss"]}', client_id:'{params["client_id"]}'"""
+            f"""Registration not found. iss:'{params["iss"]}', client_id:'{params.get("client_id")}'"""
```
jon-betts, line 63: Doing something like:

```issuer, client_id = params["iss"], params.get("client_id")```

Would prevent you from having to repeat the access here.

File: lms/views/lti/oidc.py [hunk 3]
```diff
@@ -72,7 +72,7 @@ def oidc_view(request):
         "prompt": "none",
         "login_hint": params["login_hint"],
         "lti_message_hint": params["lti_message_hint"],
-        "client_id": params["client_id"],
+        "client_id": registration.client_id,
```
jon-betts, line 75: :+1: Seems like the better way to do it anyway.

I wonder if it's worth saying something like:

```suggestion
        # We might not have been provided the client id in the request, so use
        # the client id from the registration
        "client_id": registration.client_id,
```

tests/unit/lms/services/lti_registration_test.py
```diff
@@ -10,6 +10,9 @@ class TestLTIRegistrationService:
     def test_get(self, svc, registration):
         assert svc.get(registration.issuer, registration.client_id) == registration
```
jon-betts, line 11: If we can look it up by `issuer` alone in the test below, do we actually prove that the client id is used at all? Does this test require a noise element with the same issuer and a different client?

tests/unit/lms/views/lti/oidc_test.py
```diff
@@ -34,7 +35,7 @@ def test_it(self, lti_registration, pyramid_request):
                 "prompt": "none",
                 "login_hint": pyramid_request.parsed_params["login_hint"],
                 "lti_message_hint": pyramid_request.parsed_params["lti_message_hint"],
-                "client_id": pyramid_request.parsed_params["client_id"],
+                "client_id": "REGISTRATION_CLIENT_ID",
```
jon-betts, line 38:
```suggestion
                "client_id": lti_registration.get.return_value.client_id,
```
?

If this works, I don't think you need the line above

tests/unit/lms/services/lti_registration_test.py
```diff
@@ -10,6 +10,9 @@ class TestLTIRegistrationService:
...
 
+    def test_without_client_id(self, svc, registration):
+        assert svc.get(registration.issuer) == registration
```
jon-betts, line 14: ~Does this fail if more than one matches? Should it?~ Should we test for it?

Yes. `one_or_none` should do this for us.

jon-betts (approved):
This looks good. The testing is a bit of a pain, but good instructions.

I think the tests might be a bit under-constrained. I suspect we could break the code and not have it show up in a few different ways. Like ignoring the `client_id` entirely, or replacing `one_or_none()` with `first()`.

PR #3838, opened 2022-04-06
Injection of LTI1.3 params doesn't fail when missing

marcospri:
Any missing parameters should make their way to the schema and fail the
validation step, not cause an exception.

These conditions don't occur in normal operation but when they happen, for example in the LTI certification, this PR makes the schema handle the missing data instead of raising some `KeyError` or `AttributeError`

### DIFF
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -38,7 +38,7 @@ class LTIV11CoreSchema(PyramidRequestSchema):
     @pre_load
     def _decode_jwt(self, data, **_kwargs):
         """Use the values encoded in the `id_token` JWT if present."""
-        if "id_token" not in self.context["request"].POST:
+        if not self.context["request"].lti_jwt:
             return data
 
         params = LTIParams.from_v13(self.context["request"].lti_jwt)
--- a/lms/validation/authentication/_lti.py
+++ b/lms/validation/authentication/_lti.py
@@ -90,10 +90,13 @@ class LTI13AuthSchema(LTIV11CoreSchema):
     deployment_id = marshmallow.fields.Str(required=True)
 
     @marshmallow.pre_load
-    def _lti_v3_fields(self, data, **_kwargs):  # pylint:disable=no-self-use
-        data["iss"] = data.v13["iss"]
-        data["aud"] = data.v13["aud"]
-        data["deployment_id"] = data.v13[f"{CLAIM_PREFIX}/deployment_id"]
+    def _lti_v13_fields(self, data, **_kwargs):  # pylint:disable=no-self-use
+        if not self.context["request"].lti_jwt:
+            return data
+
+        data["iss"] = data.v13.get("iss")
+        data["aud"] = data.v13.get("aud")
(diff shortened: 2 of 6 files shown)


lms/validation/_lti_launch_params.py
```diff
@@ -38,7 +38,10 @@ class Meta:
     @pre_load
     def _decode_jwt(self, data, **_kwargs):
         """Use the values encoded in the `id_token` JWT if present."""
-        if "id_token" not in self.context["request"].POST:
+        if (
+            "id_token" not in self.context["request"].POST
+            or not self.context["request"].lti_jwt
```
marcospri, line 43: If the jwt validation failed we shouldn't try to read it.

lms/validation/authentication/_lti.py
```diff
@@ -90,10 +90,13 @@ class LTI13AuthSchema(LTIV11CoreSchema):
     deployment_id = marshmallow.fields.Str(required=True)
 
     @marshmallow.pre_load
-    def _lti_v3_fields(self, data, **_kwargs):  # pylint:disable=no-self-use
-        data["iss"] = data.v13["iss"]
-        data["aud"] = data.v13["aud"]
-        data["deployment_id"] = data.v13[f"{CLAIM_PREFIX}/deployment_id"]
+    def _lti_v13_fields(self, data, **_kwargs):  # pylint:disable=no-self-use
+        if not getattr(data, "v13", None):
```
marcospri, line 94: Similarly, if we don't have the v13 params just stop here.

tests/functional/lti_certification/v13/core/test_bad_payloads.py
```diff
@@ -122,9 +113,7 @@ def test_deployment_id_claim_missing(self, test_payload, assert_missing_claim):
             status=403,
         )
 
-    @pytest.mark.xfail(
-        reason="Missing error handling for missing JWT fields", strict=True
-    )
+    @pytest.mark.xfail(reason="Missing tool_guid", strict=True)
```
marcospri, line 116: Just updated the reason here, not handled by this PR.

tests/unit/lms/validation/authentication/_lti_test.py
```diff
@@ -120,6 +120,16 @@ def test_it_raises_missing_application_instance(
         with pytest.raises(ValidationError):
             schema.lti_user()
 
+    def test_it_missing_v13(self, pyramid_request):
+        pyramid_request.lti_jwt = None
```
jon-betts, line 124:
```suggestion
        pyramid_request.lti_jwt = {}
```

I think the jwt method returns empty dict if this isn't set

jon-betts (approved):
This looks good, but I think some of the conditions can be simplified. I'm not sure why we're using `getattr` instead of getting the attribute directly. My understanding is it will always be there?

The tests still pass with the simpler test.

lms/validation/authentication/_lti.py
```diff
@@ -90,10 +90,13 @@ class LTI13AuthSchema(LTIV11CoreSchema):
...
+            return data
```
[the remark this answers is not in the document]
marcospri: At the end the vest thing to do is use the same condition, just check for `request.lti_jwt`.

PR #3864, opened 2022-04-19
Accept items that define a model in devdata

marcospri:


jon-betts (commented):
This works, but I think it's more complicated than it needs to be for a couple of reasons:

* We have tow locations for the class type "type" and "model" and I can't see why we can't re-use one
* A pre-existing decision to merge data and metadata which means we spend some time pulling it apart again

Also, apropos of nothing, it's weird that we start the transaction in `init` and then stop it in  `create_all()`. I feel like it makes more sense for the transaction to be opened and closed in the same scope.

bin/update_dev_data.py
```diff
@@ -42,21 +42,30 @@ def create_all(self):
...
-                self.upsert_assignment(data_dict)
-            else:
-                raise RuntimeError(f"Unrecognized type: {type_}")
+    def upsert_model(self, **item):
+        data = item["data"]
 
-        self.tm.commit()
+        for key, value in data.items():
+            # Decode any base64 fields
+            if key.startswith("base64_"):
```
marcospri, line 56: Sort of generalized this base64 behavior.

jon-betts: I like this, but I do wish it was a bit more obvious it's not part of the key. The keys here are `"base64_encoded_developer_secret_bytes` which really reads like:

 * `base64_encoded` / `developer_secret_bytes`

We could use some chars which can't be the attribute of python object like:

 * `base64:encoded_developer_secret_bytes`
 * `<base64>encoded_developer_secret_bytes`

I also wonder if value is the place to put it? The down side of this is whatever identifier you pick could be something that shows up in real data. For example we could use the [URL data format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) in the key for this, as it's somewhat standard:

`data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==`

jon-betts (approved):
This looks good, but I think the following key is quite confusing:

`base64_encoded_developer_secret_bytes`

As there are two ways to read it:

 * `base64_encoded` -  `developer_secret_bytes` _(Wrong!)_
 * `base64` -  `encoded_developer_secret_bytes`
 
The second one is right, but I'd say the first one is more intuitive. There's a long ramble about how you might fix this, but I'd say the two best solutions are probably:


```python
{
    "base64:encoded_developer_secret_bytes": "SGVsbG8sIFdvcmxkIQ=="
}
```

Or

```python
{
    "encoded_developer_secret_bytes": {"__base64__": "SGVsbG8sIFdvcmxkIQ=="}
}
```

marcospri:
> 
> ```python
> {
>     "encoded_developer_secret_bytes": {"__base64__": "SGVsbG8sIFdvcmxkIQ=="}
> }
> ```

I like the dict approach more :+1:

PR #3870, opened 2022-04-20
LTI Advantage service and LTI1.3 grading

marcospri:
## Testing

- Start the server with more than one worker so the LMS can make a request to the `/lti/1.3/jwks` endpoint while the server handles another request.

```diff
diff --git a/conf/supervisord-dev.conf b/conf/supervisord-dev.conf
index 63063eb8..aec0ab0d 100644
--- a/conf/supervisord-dev.conf
+++ b/conf/supervisord-dev.conf
@@ -3,7 +3,7 @@ nodaemon = true
 silent = true
 
 [program:web]
-command=gunicorn --workers 1 --paste conf/development.ini
+command=gunicorn --workers 2 --paste conf/development.ini
 stdout_events_enabled=true
 stderr_events_enabled=true
 stopsignal = KILL
```

- `make devdata` 

- Start a tunnel to `blackboard-jon`

`loophole http 8001 --hostname blackboard-jon`

- Launch https://aunltd-test.blackboard.com/webapps/blackboard/content/contentWrapper.jsp?content_id=_664_1&displayName=Test+with+grading&course_id=_19_1&navItem=content&href=%2Fwebapps%2Fblackboard%2Fexecute%2Fblti%2FlaunchPlacement%3Fblti_placement_id%3D_88_1%26content_id%3D_664_1%26course_id%3D_19_1 as a teacher (you might need to configure it)

(Or if you need to create a new assignment use `Hypothesis LTI1.3` from the build content dropdown)

 
- Launch it as `Blackboard Student 2`
- Launch the assignment again as a teacher. You should get the grading bar with one student in the dropdown now.
- Switch to the student on the drop-down. You'll get a empty grade.
- Grade the student
- Relaunch the assignment and select the student again. You'll get the submitted grade.

### DIFF
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -44,11 +44,13 @@ _V11_TO_V13 = (
     ("tool_consumer_instance_guid", [f"{CLAIM_PREFIX}/tool_platform", "guid"]),
     (
         "tool_consumer_info_product_family_code",
-        [
-            f"{CLAIM_PREFIX}/tool_platform",
-            "product_family_code",
-        ],
+        [f"{CLAIM_PREFIX}/tool_platform", "product_family_code"],
     ),
+    (
+        "lis_outcome_service_url",
+        ["https://purl.imsglobal.org/spec/lti-ags/claim/endpoint", "lineitem"],
+    ),
+    ("lis_result_sourcedid", ["sub"]),
 )
 
 
--- a/lms/services/jwt.py
+++ b/lms/services/jwt.py
@@ -95,6 +95,15 @@ class JWTService:
             LOG.debug("Invalid JWT. %s", str(err))
             return {}
 
+    @classmethod
+    def encode_with_private_key(cls, private_key, payload: dict, headers=None):
+        return jwt.encode(
+            payload,
+            private_key,
+            algorithm="RS256",
+            headers=headers,
+        )
+
     @staticmethod
     @lru_cache
     def _get_jwk_client(jwk_url: str):
--- a/lms/services/jwt.py
+++ b/lms/services/jwt.py
@@ -8,13 +8,15 @@ from jwt.exceptions import ExpiredSignatureError, InvalidTokenError, PyJWTError
 
 from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
 from lms.services.lti_registration import LTIRegistrationService
+from lms.services.rsa_key import RSAKeyService
 
 LOG = logging.getLogger(__name__)
 
 
 class JWTService:
-    def __init__(self, registration_service):
+    def __init__(self, registration_service, rsa_key_service):
         self._registration_service = registration_service
+        self._rsa_key_service = rsa_key_service
 
     @classmethod
     def decode_with_secret(cls, jwt_str, secret) -> dict:
@@ -95,13 +97,13 @@ class JWTService:
             LOG.debug("Invalid JWT. %s", str(err))
             return {}
 
-    @classmethod
-    def encode_with_private_key(cls, private_key, payload: dict, headers=None):
+    def encode_with_private_key(self, payload: dict):
+        key = self._rsa_key_service.get_random_key()
         return jwt.encode(
             payload,
-            private_key,
+            self._rsa_key_service.private_key(key),
             algorithm="RS256",
-            headers=headers,
+            headers={"kid": key.kid},
         )
 
     @staticmethod
@@ -117,7 +119,10 @@ class JWTService:
 
 
 def factory(_context, request):
-    return JWTService(registration_service=request.find_service(LTIRegistrationService))
+    return JWTService(
+        registration_service=request.find_service(LTIRegistrationService),
+        rsa_key_service=request.find_service(RSAKeyService),
+    )
 
 
 def _get_lti_jwt(request):
--- a/lms/services/lti_grading/factory.py
+++ b/lms/services/lti_grading/factory.py
@@ -1,7 +1,19 @@
 from lms.services.lti_grading._v11 import LTI11GradingService
+from lms.services.lti_grading._v13 import LTI13GradingService
+from lms.services.ltia_http import LTIAHTTPService
 
 
 def service_factory(_context, request):
+    application_instance = request.find_service(
+        name="application_instance"
+    ).get_current()
+
+    if application_instance.lti_version == "1.3.0":
+        return LTI13GradingService(
+            grading_url=request.parsed_params["lis_outcome_service_url"],
+            ltia_service=request.find_service(LTIAHTTPService),
+        )
+
     return LTI11GradingService(
         grading_url=request.parsed_params["lis_outcome_service_url"],
         http_service=request.find_service(name="http"),
--- a/lms/services/ltia_http.py
+++ b/lms/services/ltia_http.py
@@ -1,0 +1,68 @@
+import uuid
+from datetime import datetime, timedelta
+
+from lms.models import LTIRegistration
+from lms.services import JWTService
+
+
+class LTIAHTTPService:
+    """Send LTI Advantage requests and return the responses."""
+
+    def __init__(
+        self, lti_registration: LTIRegistration, jwt_service: JWTService, http
+    ):
+        self._lti_registration = lti_registration
+        self._jwt_service = jwt_service
+        self._http = http
+
+    def request(self, method, url, scopes, headers=None, **kwargs):
+        headers = headers or {}
+
+        assert "Authorization" not in headers
+
+        access_token = self._get_access_token(scopes)
+        headers["Authorization"] = f"Bearer {access_token}"
+
+        return self._http.request(method, url, headers=headers, **kwargs)
+
+    def _get_access_token(self, scopes):
+        """
+        Get an access token from the LMS to use in LTA services.
+
+        https://datatracker.ietf.org/doc/html/rfc7523
+        https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token
+        """
+        now = datetime.utcnow()
+        signed_jwt = self._jwt_service.encode_with_private_key(
+            {
+                "exp": now + timedelta(hours=1),
+                "iat": now,
+                "nonce": uuid.uuid4().hex,
+                "jti": uuid.uuid4().hex,
+                "iss": self._lti_registration.client_id,
+                "sub": self._lti_registration.client_id,
+                "aud": self._lti_registration.token_url,
+            }
+        )
+
+        response = self._http.post(
+            self._lti_registration.token_url,
+            data={
+                "grant_type": "client_credentials",
+                "client_assertion_type": "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+                "client_assertion": signed_jwt,
+                "scope": " ".join(scopes),
+            },
+        )
+
+        return response.json()["access_token"]
+
+
+def factory(_context, request):
+    return LTIAHTTPService(
+        request.find_service(name="application_instance")
+        .get_current()
+        .lti_registration,
+        request.find_service(JWTService),
+        request.find_service(name="http"),
+    )
--- a/tests/functional/lti_certification/v13/grading/test_assignment_and_grading.py
+++ b/tests/functional/lti_certification/v13/grading/test_assignment_and_grading.py
@@ -1,0 +1,74 @@
+class TestAssignmentAndGradingSerices:
+    """
+    From: https://www.imsglobal.org/spec/lti/v1p3/cert/#assignment-and-grade-services-testing
+
+    Assignments and Grade Services (AGS) is tested as pure service (without any UI).
+    The Tool is required to acquire the OAuth2 tokens from the IMS Global
+    testing OAuth2 server necessary to interact with the AGS system.
+    Testing for AGS for the Tool is very different from all other testing
+    in the Certification Suite.
+
+    Since it is possible to jump directly to testing for AGS, the Certification
+    Suite provides the place to launch a standard, Learner-based LTI 1.3
+    launch into your Tool. However, that is the only prescribed test in the
+    Certification Suite.
+
+    After that launch it is the responsibility of the Tool alone to work with
+    the AGS API to create lineitems and scores in the Certification Suite.
+    All interaction with the Gradebook simulated by the Certification Suite
+    can be viewed on the Results page
+    """
+
+    def test_it(self):
+        """
+        Due to the nature of this test in the certification no actual testing
+        happens here.
+
+        The test in the certification tool expects the tool to initiate
+        communication with the service but it's not tied to any launch or real
+        flow in the app.
+
+        We include here the script used for the certification for completeness.
+        ```
+        # These are included in the JWT sent by the testing tool
+        # https://purl.imsglobal.org/spec/lti/claim/resource_link -> id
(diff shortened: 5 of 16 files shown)


conf/supervisord-dev.conf
```diff
@@ -3,7 +3,7 @@ nodaemon = true
 silent = true
 
 [program:web]
-command=gunicorn --paste conf/development.ini
+command=gunicorn --workers 1 --paste conf/development.ini
```
marcospri, line 6: Made the default explicit so it's easier to change.

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -65,6 +65,11 @@ def test_get_all_public_jwks(self, svc, valid_keys):
         ]
         assert keys == Any.list.containing(expected_keys).only()
 
+    def test_get_random_key(self, svc, valid_keys):
+        key = svc.get_random_key()
```
marcospri, line 69: I'm not testing the randomness element of this

conf/supervisord-dev.conf
jon-betts: You _should_ be able to set this in the `conf/development.ini`, but I have a recollection that certain fields like the port and stuff just don't work:

 * https://docs.gunicorn.org/en/stable/configure.html#paster-applications
 
Honestly thought I'm not sure if that's just more confusing than having it here

lms/resources/_js_config/__init__.py
```diff
@@ -228,7 +228,7 @@ def maybe_enable_grading(self):
             # Only instructors can grade assignments.
             return
 
-        if "lis_outcome_service_url" not in self._request.params:
+        if "lis_outcome_service_url" not in self._context.lti_params:
```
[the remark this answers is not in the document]
jon-betts: Just for my own notes, the difference here is that `lti_params` has been through the wringer of conversions, so this makes using the old LTI 1.1 names 1.3 compatible.

lms/services/lti_outcomes.py
```diff
@@ -1,26 +1,74 @@
...
 
 import xmltodict
 
+from lms.services import LTIAHTTPService
 from lms.services.exceptions import ExternalRequestError
 
 __all__ = ["LTIOutcomesClient"]
 
 
 class LTIOutcomesClient:
```
jon-betts, line 12: I have a suggestion here which is pretty disruptive. I don't think this is one class, I think it's two which have been put together.

I think the correct solution here might be do use a feature of the services we don't make use of normally, which is the ability to send back a difference class from the interface. Something like:

```python
class LTIGradingClient:
    def read_result(self, grading_id):
        ...

    def record_result(self, grading_id, score=None, pre_record_hook=None):
        ...


class LTI13GradingClient(LTIGradingClient):
    ...


class LTI11GradingClient(LTIGradingClient):
    ...
```

Then in the factory:

```python
    application_instance = request.find_service(name="application_instance").get_current()
    if application_instance.lti_version == "1.3.0":
        return LTI13GradingClient(...)
    else:
        return LTI11GradingClient(...)
```

I think this would have a few positive outcomes:

 * Each class becomes about one thing only and can be split into sub-files in the dir
 * It's clear what items are for supporting 1.1, 1.3 or are shared (probably very little)
 * All the switching behavior is removed
 * The LTI 1.3 class will probably get to a size where `LTIAHTTPService` can either become private to it, or inlined

jon-betts (commented):
I've not gone through this in detail yet, because I have a suggestion which I think might make this quite different.

The idea is:

 * Have two separate classes for V11, and V13 grading
 * Have the factory pick the right one and return it
 * Inline the new HTTP service into the V13 (or at least make it private)
 
I think this would reduce the complexity of each item and keep the V11 and V13 behavior distinct.

lms/services/lti_outcomes.py
```diff
@@ -1,26 +1,74 @@
...
+            if err.status_code == 404:
+                return None
+            raise
+
+        results = response.json()
+        if not results:
+            return None
+
+        return results[-1]["resultScore"] / results[-1]["resultMaximum"]
+
```
jon-betts, line 70: This looks very easy to make explode? In the other class there's some try catching around to prevent this from going mad. Also fun opportunity for a divide by zero here, which you don't see every day.

lms/services/ltia_http.py
```diff
@@ -0,0 +1,72 @@
...
+            "client_assertion": jwt,
+            "scope": " ".join(scopes),
+        }
+        response = self._http.post(
+            self._lti_registration.token_url,
+            data=auth_request,
+        )
+        return response.json()["access_token"]
+
+    def request(self, scopes, method, url, headers=None, **kwargs):
```
jon-betts, line 54:
```suggestion
    def request(self, method, url, headers=None, scopes=None, **kwargs):
        if not scopes:
            raise TypeError("Scopes are required")
```

I realise the defaulting here is awkward, but I feel like this quite strongly breaks the interface you'd expect by having scopes first.

If we did this it would also nicely mix with the suggestion to sub-class above.

lms/services/ltia_http.py
```diff
@@ -0,0 +1,72 @@
...
+from lms.services import JWTService
+
+
+class LTIAHTTPService:
+    """Send LTI Advantage requests and return the responses."""
+
+    def __init__(self, lti_registration, jwt_service, http):
+        self._lti_registration = lti_registration
+        self._jwt_service = jwt_service
+        self._http = http
```
jon-betts, line 13:
```suggestion
class LTIAHTTPService(HTTPService):
    """Send LTI Advantage requests and return the responses."""

    def __init__(self, lti_registration, jwt_service):
        self._lti_registration = lti_registration
        self._jwt_service = jwt_service
```

I know it is _forbidden_, but if we are an HTTP service, then we would get the benefit of all the extra `get()`, `post()` convenience methods added to the base class.

tests/functional/lti_certification/v13/grading/test_assignment_and_grading.py
```diff
@@ -0,0 +1,71 @@
...
+                "label": "Introduction Assignment",
+                "resourceId": resource_id,
+                "tag": "certification_grade",
+            },
+        )
+        out.service_url = response.json()["id"]
+        out.record_result("40899", 0.8)
+        out.read_result("40899")
+        ```
+        """
```
jon-betts, line 71: :shrug:

conf/supervisord-dev.conf
marcospri: I tried a few of those sections in the .ini and didn't work. At the end I was happy with anywhere I could grep for `workers`.

lms/models/lti_params.py
```diff
@@ -44,11 +44,13 @@ def from_v13(cls, v13_params):
     ("tool_consumer_instance_guid", [f"{CLAIM_PREFIX}/tool_platform", "guid"]),
     (
         "tool_consumer_info_product_family_code",
-        [
-            f"{CLAIM_PREFIX}/tool_platform",
-            "product_family_code",
-        ],
+        [f"{CLAIM_PREFIX}/tool_platform", "product_family_code"],
     ),
+    (
```
marcospri, line 49: New param mapping. Grading related.

lms/resources/_js_config/__init__.py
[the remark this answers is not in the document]
marcospri: Yeah

`request.params` the raw parameters of the request.
`request.lti_params` the params with the translation layer for the different versions.
`request.lti_jwt` the raw 1.3 params inside the jwt.

lms/services/lti_grading/factory.py
```diff
@@ -0,0 +1,13 @@
+from lms.services.lti_grading._v11 import factory as factory_v11
+from lms.services.lti_grading._v13 import factory as factory_v13
+
+
+def factory(context, request):
```
marcospri, line 5: I was having circular import problems around so now each version has it's own factory function and this one just decides which one to use.

lms/services/lti_grading/interface.py
```diff
@@ -0,0 +1,33 @@
+class LTIGradingService:  # pragma: no cover
```
marcospri, line 1: I'm cheating here not sure if there was anything worth testing here.

marcospri, line 1: I ended up not using a dataclass for these

- Being a bit pedantic this is not "data". This is very debatable but it's just my preference here.
- More concretely we've seen before problems when the dataclass magic interfired with other decorators. For example for `lru_cache` you need to make the dataclass hasheable. Not the end of the world but having the manual `__init__` is not horrible either.

lms/services/ltia_http.py
marcospri: Not a bad idea but I think it will take a bit of work to adapt the tests (that now use the mocked http_service. I rather revisit it later.

Also have friend that once used inheritance over composition once and this computer exploded...

lms/views/api/lti.py
```diff
@@ -18,13 +19,15 @@ class LTIOutcomesViews:
     def __init__(self, request):
         self.request = request
         self.parsed_params = self.request.parsed_params
-        self.lti_outcomes_client = self.request.find_service(name="lti_outcomes_client")
+        self.lti_grading_service: LTIGradingService = self.request.find_service(
```
marcospri, line 22: Not sure we've been doing this attribute typing thing but we have to import the service class anyway for `find_service`. 

With this I get auto-completion for self.lti_grading_service

tests/unit/lms/services/lti_grading/factory_test.py
```diff
@@ -0,0 +1,56 @@
+from unittest.mock import Mock, sentinel
+
+import pytest
+
+from lms.services.lti_grading.factory import factory
+
+
+class TestFactory:
+    def test_v11(
```
marcospri, line 9: Again, cheating here a bit. I'm testing

`lms.services.lti_grading.factory` but asserting the behavior of `lms.services.lti_grading._v11|v13.factory`.

I personally think mocking `_v11.factory` here and having an independent test for it on the `_v11` file wound't add much value to the test. I'm using the fact that the other modules start with an underscore as a justification.

tests/unit/lms/validation/authentication/_lti_test.py
```diff
@@ -30,7 +30,7 @@ def test_it_returns_the_lti_user_info(
                 "lis_person_name_given": "TEST_GIVEN_NAME",
                 "oauth_signature": "TEST_OAUTH_SIGNATURE",
                 "roles": "Instructor",
-                "tool_consumer_instance_guid": "TEST_GUID",
+                "tool_consumer_instance_guid": "TEST_TOOL_CONSUMER_INSTANCE_GUID",
```
marcospri, line 33: This aligns more with the reset of the values.

lms/services/lti_outcomes.py
marcospri: Not sure about catching KeyError and friends. I think if some LMS breaks the spec we rather find out in sentry/logs before. If it does happen often and we have to handle we'll see.

I see division by zero more plausible. But not sure what to do, return None? return the absolute `resultScore`?

lms/services/ltia_http.py
marcospri: I agree it doesn't make much sense as the first param, move it after `url`. 

Not a fan of doing the `scopes=None` and then having to check against it being none. I guess with my proposal you could still pass None as the scope but the same is true of method and url.

tests/unit/lms/services/lti_grading/factory_test.py
```diff
@@ -0,0 +1,56 @@
...
+    ):
+        application_instance_service.get_current.return_value = Mock(
+            lti_version="1.3.0"
+        )
+        svc = factory(sentinel.context, pyramid_request)
+
+        LTI13GradingService.assert_called_once_with(
+            sentinel.grading_url, ltia_http_service
+        )
+        assert svc == LTI13GradingService.return_value
```
jon-betts, line 41: Yeah. These are kind of long winded, but only because we have long names for our services. I don't think you can do this any better.

tests/unit/lms/services/rsa_key_service_test.py
jon-betts: Yeah, kind of logically impossible to do really. You could call it 1 million times and technically get the same one everytime.

tests/unit/lms/views/api/lti_test.py
```diff
@@ -7,36 +7,36 @@
 
 from lms.views.api.lti import CanvasPreRecordHook, LTIOutcomesViews
 
-pytestmark = pytest.mark.usefixtures("lti_outcomes_client")
+pytestmark = pytest.mark.usefixtures("lti_grading_service")
 
 
 class TestRecordCanvasSpeedgraderSubmission:
     GRADING_ID = "lis_result_sourcedid"
```
jon-betts, line 14: I think a sentinel would be better here again. The string `lis_result_sourcedid` is a bit misleading

tests/unit/services.py
```diff
@@ -62,6 +63,7 @@
     "aes_service",
     "jwt_service",
     "rsa_key_service",
+    "ltia_http_service",
```
jon-betts, line 66: I think if you notice, all the services above `jstor` (which I think was added recently) are in alphabetical order here, which makes it a bit easier to scan when we have so many.

tests/unit/services.py
```diff
@@ -246,3 +248,8 @@ def lti_registration_service(mock_service):
 @pytest.fixture
 def rsa_key_service(mock_service):
     return mock_service(RSAKeyService)
+
+
+@pytest.fixture
+def ltia_http_service(mock_service):
+    return mock_service(LTIAHTTPService)
```
jon-betts, line 255: I think the functions are "roughly" in alphabetical order too.

jon-betts (commented):
This is still just a kind of linting review. I've made many suggestions here, but my biggest one is that we should look for some seams to break up this PR so it's not so huge.

This would also get around the fact there's a wasteland of comments on here (which is kind of inevitable with 900 lines of changes).

Some ideas for splits:

* Moving the LTIOutcomes to the new shape, but minus the V13 part
* Any random tidies which aren't essential to the code (some test name changes, grading info changes, there were some more I think)
* Maybe introducing the LTIA + V13 part in one go? It feels like this might jank in most of the rest of the code?

A random observation I had is the grading V11 tests are in the very decomposed style as well. That's not something to address here, but I bet they could be a lot shorter.

tests/unit/services.py
marcospri: Sorted this and the fixtures below :+1:

jon-betts (changes requested):
I'm not sure how this happened, but the factory setup doesn't work as it's got the wrong name:

```
        "lms.services.lti_grading.service_factory", iface=LTIGradingService
    )
```

Which gives a failure to grab the grade. If that's fixed, and ViaHTML is bodged, then this does seem to work.

lms/services/jwt.py
```diff
@@ -95,6 +97,17 @@ def decode_lti_token(self, id_token: str) -> dict:
             LOG.debug("Invalid JWT. %s", str(err))
             return {}
 
+    def encode_with_private_key(self, payload: dict):
+        key = self._rsa_key_service.get_random_key()
```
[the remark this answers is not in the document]
marcospri: We use both the private key and the `kid`. I think returning a tuple or similar would be more awkward than returning the whole RSAKey.

jon-betts (approved):
Yep. Looks good!

PR #3877, opened 2022-04-25
Expose `submitted_at` in canvas grading API 

marcospri:
For #3647 

Added as an optional parameter first to make the deployment/rollout easier. As of now there's no code that that sets the date in the frontend so all submissions still should use the default 2001 date.

# Testing

- Create a new assignment in canvas as a techer https://hypothesis.instructure.com/courses/125/assignments/new?submission_types=none&name=&due_at=null&points_possible=0&assignment_group_id=166

- Launch the assignment as `eng+canvasstudent@hypothes.is`

- Back to the teacher account, open the assignment in speed-grader, the date of the submission will be `Submitted:
Dec 31, 2000 at 5pm`

### DIFF
--- a/lms/validation/_api.py
+++ b/lms/validation/_api.py
@@ -47,6 +47,9 @@ class APIRecordSpeedgraderSchema(JSONPyramidRequestSchema):
     ext_lti_assignment_id = fields.Str(required=False, allow_none=True)
     """Canvas doesn't send this value on speed grader launches"""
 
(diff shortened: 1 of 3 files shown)


jon-betts (approved):
Yep I've given this a shot and it all works.

I think this ticket could have done with a bit more context to know why certain things were happening and what it meant.

I've added some notes about the tests where I think we now have two tests without clear responsibilities, and can probably get away with one, or two simpler ones.

PR #3882, opened 2022-04-26
Enable `reportActivity` setting on the client config on canvas instances

marcospri:
Following the conversation here:

https://hypothes-is.slack.com/archives/C4K6M7P5E/p1650458302131809?thread_ts=1650415818.372909&cid=C4K6M7P5E

Added the new config option only for canvas. Leaving it out for other LMSs

## Testing 

- Launch https://hypothesis.instructure.com/courses/125/assignments/2947, check the configuration at `http://localhost:8001/lti_launches?url=https://example.com`.
- Ctrl+F for `hypothesisClient` in the response. Nothing has changed, the config doesn't include the new property:

    ```hypothesisClient": {"services"...}```

- Enable `submit_on_annotation` on  http://localhost:8001/flags

- Refresh the launch and check the response again for `http://localhost:8001/lti_launches?url=https://example.com`, Ctrl+F for `hypothesisClient`. 

    The new `{"reportActivity": {"events": ["create", "update"], "method": "reportActivity"}` should be there.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -338,6 +338,10 @@ class JSConfig:
         # Enable the LMS frontend to receive notifications on annotation activity
         # We'll use this information to only send the submission to canvas on first annotation.
         if self._request.feature("submit_on_annotation"):
+            # The `reportActivity` setting causes the front-end to make a call
+            # back to the parent iframe for the specified events. The method in
+            # the iframe happens to be called `reportActivity` too, but this is
+            # a co-incidence. It could have any name.
             self._hypothesis_client["reportActivity"] = {
                 "method": "reportActivity",
                 "events": ["create", "update"],
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -337,10 +337,11 @@ class JSConfig:
         }
         # Enable the LMS frontend to receive notifications on annotation activity
         # We'll use this information to only send the submission to canvas on first annotation.
-        self._hypothesis_client["reportActivity"] = {
-            "method": "reportActivity",
-            "events": ["create", "update"],
-        }
+        if self._request.feature("submit_on_annotation"):
+            self._hypothesis_client["reportActivity"] = {
+                "method": "reportActivity",
+                "events": ["create", "update"],
+            }
 
     def _auth_token(self):
         """Return the authToken setting."""
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -335,6 +335,12 @@ class JSConfig:
                 **kwargs,
             },
         }
+        # Enable the LMS frontend to receive notifications on annotation activity
+        # We'll use this information to only send the submission to canvas on first annotation.
+        self._hypothesis_client["reportActivity"] = {
+            "method": "reportActivity",
+            "events": ["create", "update"],
+        }
 
     def _auth_token(self):
         """Return the authToken setting."""
--- a/tests/unit/lms/resources/_js_config/__init___test.py
+++ b/tests/unit/lms/resources/_js_config/__init___test.py
@@ -238,9 +238,19 @@ class TestAddDocumentURL:
 
         assert submission_params()["document_url"] == "example_document_url"
 
+    @pytest.mark.parametrize("submit_on_annotation_enabled", [True, False])
     def test_it_sets_the_canvas_submission_params(
-        self, pyramid_request, js_config, submission_params
+        self,
+        pyramid_request,
+        js_config,
+        submission_params,
+        submit_on_annotation_enabled,
     ):
+        if submit_on_annotation_enabled:
+            pyramid_request.feature.side_effect = (
+                lambda feature: feature == "submit_on_annotation"
+            )
+
         js_config.add_document_url("canvas://file/course_id/COURSE_ID/file_id/FILE_ID")
 
         assert submission_params() == Any.dict.containing(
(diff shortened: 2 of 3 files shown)


tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -546,31 +548,21 @@ def config(self, config):
...
         )
-        grant_token = config["services"][0]["grantToken"]
-        assert grant_token == grant_token_service.generate_token.return_value
 
-    def test_it_includes_the_group(self, config, context):
-        groups = config["services"][0]["groups"]
-
-        assert groups == [context.h_group.groupid.return_value]
+        assert len(config["services"]) == 1
+        assert config["services"][0] == {
```
marcospri, line 557: I think we prefer this style of assertion instead of one per test. 

Of the boolean properties we were only asserting about `enableShareLinks`, added the reset.

lyzadanger (commented):
This does indeed appear to set the correct configuration. I determined this by logging in the registered `reportActivity` callback https://github.com/hypothesis/lms/blob/58620cce2e1dabce6743a70e276235bfc22bd482/lms/static/scripts/frontend_apps/services/client-rpc.js#L90-L96 and creating/editing annotations in an assignment to see if  that callback got invoked.

Can't speak to the tests!

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -599,6 +591,14 @@ def test_it_is_mutable(self, config):
 
         assert config["a_key"] == "a_value"
 
+    def test_it_includes_report_activity_in_canvas(self, config, context):
+        context.is_canvas = True
+
+        assert config["reportActivity"] == {
+            "method": "reportActivity",
+            "events": ["create", "update"],
+        }
```
jon-betts, line 600: If you comment out `context.is_canvas = True` this still works, as I think it's the default.

You could parameterise it, or perhaps box it in by setting `canvas=False` as the default and making a complete assertion about the response in `test_it` which would not include `reportActivity`?

lms/resources/_js_config/__init__.py
```diff
@@ -456,6 +456,13 @@ def _hypothesis_client(self):
                 }
             ]
         }
+        if self._context.is_canvas:
```
jon-betts, line 459: Why is this Canvas only? Also this could really do with a comment explaining what this and what it does.

jon-betts (commented):
I don't think I can really review this yet, as I don't really know what it does or why. This ticket says we are adding the new option for Canvas, but it doesn't say what the new option does, or why only Canvas gets it.

This ticket links to a lengthy chat in Slack, but it's a long thread and I don't know how much is relevant. I think it would benefit me reviewing it now and also us coming back should we ever read the ticket for it to be self contained and explain what this is and why it's being done.

I'm a bit concerned this doesn't seem to follow the patterns we have for other call back type things, but not exactly knowing what this supposed to do I don't know if this is relevant feedback or not?

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -546,31 +548,21 @@ def config(self, config):
...
+        assert len(config["services"]) == 1
+        assert config["services"][0] == {
+            "allowFlagging": False,
+            "allowLeavingGroups": False,
+            "apiUrl": "https://example.com/api/",
+            "authority": "TEST_AUTHORITY",
+            "enableShareLinks": False,
+            "grantToken": grant_token_service.generate_token.return_value,
+            "groups": [context.h_group.groupid.return_value],
+        }
```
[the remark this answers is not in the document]
marcospri: Yup :+1:

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -222,34 +223,40 @@ def test_it_sets_the_document_url(self, js_config, submission_params):
...
-            submission_params()["resource_link_id"]
-            == pyramid_request.params["resource_link_id"]
-        )
-        assert (
-            submission_params()["ext_lti_assignment_id"]
-            == pyramid_request.params["ext_lti_assignment_id"]
-        )
+        if submit_on_annotation_enabled:
+            # pylint:disable=protected-access
+            assert js_config._hypothesis_client["reportActivity"] == {
```
marcospri, line 256: The fixture `config` won't have `hypothesisClient` unless `enable_launch_mode` is called. I think accessing the private property directly is less confusing than calling enable_launch_mode for its side effects.

lms/resources/_js_config/__init__.py
```diff
@@ -329,6 +329,13 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
                 **kwargs,
             },
         }
+        # Enable the LMS frontend to receive notifications on annotation activity
+        # We'll use this information to only send the submission to canvas on first annotation.
+        if self._request.feature("submit_on_annotation"):
```
jon-betts, line 334:
```suggestion
        if self._request.feature("submit_on_annotation"):
            # The `reportActivity` setting causes the front-end to make a call 
            # back to the parent iframe for the specified events. The method in 
            # the iframe happens to be called `reportActivity` too, but this is
            # a co-incidence. It could have any name.
```

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -223,33 +224,27 @@ def test_it_sets_the_document_url(self, js_config, submission_params):
...
-        assert submission_params()["learner_canvas_user_id"] == "test_user_id"
-
-        assert (
-            submission_params()["resource_link_id"]
-            == pyramid_request.params["resource_link_id"]
-        )
-        assert (
-            submission_params()["ext_lti_assignment_id"]
-            == pyramid_request.params["ext_lti_assignment_id"]
+        assert submission_params() == Any.dict.containing(
```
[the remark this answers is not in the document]
jon-betts: Yep, much clearer.

jon-betts (approved):
I think the code here is good.

The name of the configuration variable could be improved. Names which include words like 'report' which can be a noun or a verb are a bit unclear, and I think the name could be improved to more easily give the impression that you are registering an event call back.

I understand this is intended for more general use. I'd suggest a likely future requirement is to register more than one callback for different events. Now would probably be the cheapest time to make this a list rather than a single value, while we don't have any backwards compatibility to worry about.

In any case, none of that was introduced in this PR so doesn't effect the review here.

lyzadanger (approved):
Tested: configuration values show up as expected per feature flag.

PR #3884, opened 2022-04-26
New method in VS source to get anonymous document URL

marcospri:


lms/services/vitalsource/client.py
```diff
@@ -77,32 +77,39 @@ def parse_document_url(document_url):
     def generate_document_url(book_id, cfi):
         return f"vitalsource://book/bookID/{book_id}/cfi/{cfi}"
 
-    def get_launch_params(self, document_url, lti_user: LTIUser) -> Tuple[str, Optional[Dict[str, str]]]:
+    def get_iframe_url(self, document_url: str) -> str:
```
marcospri, line 80: iframe_url is not a great name

lms/resources/_js_config/__init__.py
```diff
@@ -71,13 +71,20 @@ def add_document_url(self, document_url):
...
-            launch_url, launch_params = vitalsource_svc.get_launch_params(
-                document_url, self._request.lti_user
-            )
-            self._config["vitalSource"] = {
-                "launchUrl": launch_url,
-                "launchParams": launch_params,
-            }
+            if self._request.feature("vitalsource_anon_launch"):
+                self._config["vitalSource"] = {
+                    "launchUrl": vitalsource_svc.get_iframe_url(document_url),
```
robertknight, line 76: Maybe `get_launch_url` for consistency with the `launchUrl` field?

lms/services/vitalsource/client.py
```diff
@@ -77,32 +80,39 @@ def parse_document_url(document_url):
...
+        """
+        Return an URL like https://hypothesis.vitalsource.com/books/{book_id}/cfi/{cfi}.
```
robertknight, line 85: The exact URL is an implementation detail that will probably change in future, we're still discussing that with them. I would say something like "Return a URL to load the VitalSource book viewer at a particular book and location."

lms/services/vitalsource/client.py
```diff
@@ -77,32 +80,39 @@ def parse_document_url(document_url):
...
+        Note that this method is an alternative to `get_launch_params` below.
+
+        :param document_url: `vitalsource://` type URL identifying the document
+        """
+        url_params = self.parse_document_url(document_url)
+        return f"https://hypothesis.vitalsource.com/books/{url_params['book_id']}/cfi/{url_params['cfi']}"
+
+    def get_launch_params(
+        self, document_url, lti_user: LTIUser
+    ) -> Tuple[str, Dict[str, str]]:
```
robertknight, line 98: Might be worth noting that the LTI launch method is deprecated and we plan to drop it soon.

lms/resources/_js_config/__init__.py
```diff
@@ -71,13 +71,20 @@ def add_document_url(self, document_url):
...
+                    "launchParams": None,
```
robertknight, line 77: You can just omit `launchParams` entirely here. It was present in my earlier code to avoid an extra branch. The frontend only cares whether it is nullish or not.

robertknight (approved):
LGTM. After merging into the other branch, can you rebase and squash this change into the commit with the Python changes from that branch, to simplify the history.

PR #3890, opened 2022-04-28
Expose the LTI version on model.ApplicationInstance

marcospri:
Not used anywhere yet but needed for https://github.com/hypothesis/lms/pull/3870

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -198,3 +198,16 @@ class ApplicationInstance(BASE):
             product = self.Product.UNKNOWN
 
         return product
+
+    @property
+    def lti_version(self) -> str:
+        """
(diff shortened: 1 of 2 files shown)


lms/models/application_instance.py
```diff
@@ -198,3 +198,10 @@ def product(self):
             product = self.Product.UNKNOWN
 
         return product
+
+    @property
+    def lti_version(self):
```
jon-betts, line 203: I'd probably add a docstring here. Without any extra context, i'd expect this to return "1.1.0" rather than "LTI-1p0". So it's probably worth mentioning why this is the way it is.

jon-betts (approved):
Yep very simple. Could do with a docstring.

PR #3891, opened 2022-04-28
Rename LTIOutcomes to LTIGrading and prepare for LTI1.3 support

marcospri:
Nothing should have change as the code is mostly just moved around, but just in case I'm checking the grading in BB

- Launch https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_36_1&course_id=_19_1 as `Blackboard Student 2`

 
- Launch the assignment again as a teacher. You should get the grading bar with one student in the dropdown now.
- Switch to the student on the drop-down. You'll get a empty grade.
- Grade the student
- Relaunch the assignment and select the student again. You'll get the submitted grade.

### DIFF
--- a/lms/services/lti_grading/_v11.py
+++ b/lms/services/lti_grading/_v11.py
@@ -3,36 +3,25 @@ from xml.parsers.expat import ExpatError
 import xmltodict
 
 from lms.services.exceptions import ExternalRequestError
+from lms.services.http import HTTPService
+from lms.services.lti_grading.interface import LTIGradingService
+from lms.services.oauth1 import OAuth1Service
 
-__all__ = ["LTIOutcomesClient"]
 
+class LTI11GradingService(LTIGradingService):
+    #  See: LTI1.1 Outcomes https://www.imsglobal.org/specs/ltiomv1p0/specification
+    def __init__(
+        self, grading_url, http_service: HTTPService, oauth1_service: OAuth1Service
+    ):
+        super().__init__(grading_url)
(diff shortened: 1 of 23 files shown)


tests/unit/lms/services/__init___test.py
```diff
@@ -19,7 +18,6 @@ class TestIncludeme:
             ("h_api", HAPI),
             ("launch_verifier", LaunchVerifier),
             ("grading_info", GradingInfoService),
-            ("lti_outcomes_client", LTIOutcomesClient),
```
marcospri, line 22: This only tests the services added by name. I made a note to extend this to include every service, include the ones that are registered by class only.

lms/services/lti_grading/_v11.py
```diff
@@ -152,8 +118,6 @@ def _pox_envelope(body):
 
         The returned dict renders to an ``<imsx_POXEnvelopeRequest>`` document
         with the given ``body`` as the contents of the ``<imsx_POXBody>`` tag.
-
-        :rtype: dict
         """
```
jon-betts, line 121: Similar:

```python
        """Return ``body`` wrapped in an imsx_POXEnvelopeRequest envelope."""
        
        # Dict in xmltodict-format for turning into XML
        return {

```

jon-betts (approved):
Yep this looks good. 

I think some of the existing documentation on the internal methods is a bit much. I think this would be a nice time to cut this down.

PR #3905, opened 2022-05-03
Remove unused createAssignmentAPI

marcospri:
All the backend pieces for this are in place but when the frontend
changes were merged (and the API was actually used) we found different
unknown behaviours / bugs in canvas around the IDs identifying
assignments that make the API fail under some circumstances.

The frontendcode revert happen here: https://github.com/hypothesis/lms/pull/3390/files#diff-703382780bfa43006217d915354dd9a1dfa974d6e391dba1554cd8d8132082faL96

If this code is useful we can always revert it back.


jon-betts (approved):
:shrug: I'm going to assume it's not needed. Git is good. We can always get it back.

jon-betts (approved):
:shrug: I'm going to assume it's not needed. Git is good. We can always get it back.

jon-betts:
Why are you so weird github?

PR #3908, opened 2022-05-03
Use the deepLinkingAPI on LTI1.1 

marcospri:
Follow up for https://github.com/hypothesis/lms/pull/3917

In the backend:

- We enable the deepLinkingAPI for LTI1.1 too (ie for all deeplinking launches)
- A new view generates the form form fields for LTI1.1. 

In the frontend:
- The equivalent of `content-item` is now done in the backend so it can be removed.

## Testing 

Log into canvas as a teacher

### LTI1.3

(No changes from last PR, but worth double checking)

- `make devdata` 
- Edit https://hypothesis.instructure.com/courses/319/assignments/2781  using the `Hypothesis LTI1.3 (localhost)` install

There will be now a call to `http://localhost:8001/lti/1.3/deep_linking/form_fields` showing up on the network tab of your browser.  This happens right after the last "continue" button in the configuring assignment screen.

- Saving the assignment will store the new document.

### LTI1.1

- Edit https://hypothesis.instructure.com/courses/125/assignments/873

There will be now a call to `http://localhost:8001/lti/1.1/deep_linking/form_fields` showing up on the network tab of your browser.  This happens right after the last "continue" button in the configuring assignment screen.

- Saving the assignment will store the new document.
- Testing it with
  - An URL
  - A URL with groups enabled
  - A canvas file
  - A canvas file with groups enabled.


lms/services/ltia_http.py
```diff
@@ -15,6 +15,21 @@ def __init__(
         self._jwt_service = jwt_service
         self._http = http
 
+    def sign(self, payload: dict):
```
marcospri, line 18: This gets its own method again.

lms/services/ltia_http.py
```diff
@@ -32,18 +47,7 @@ def _get_access_token(self, scopes):
...
-                "exp": now + timedelta(hours=1),
-                "iat": now,
-                "nonce": uuid.uuid4().hex,
-                "jti": uuid.uuid4().hex,
-                "iss": self._lti_registration.client_id,
-                "sub": self._lti_registration.client_id,
-                "aud": self._lti_registration.token_url,
-            }
-        )
+        signed_jwt = self.sign({"aud": self._lti_registration.token_url})
```
marcospri, line 50: Calling the new method using an explicit different `aud`.

lms/static/scripts/frontend_apps/utils/content-item.js
```diff
@@ -18,44 +18,3 @@
  *
  * @typedef {FileContent|URLContent} Content
  */
-
-/**
```
marcospri, line 22: This all happens all in the backend now.

lms/resources/_js_config/__init__.py
```diff
@@ -226,6 +226,7 @@ def enable_content_item_selection_mode(self, form_action, form_fields):
                 },
             }
         )
+        self._add_deeplinking_api()
```
marcospri, line 229: Done as separate call instead of directly assign to `_config['deeplinkingAPI']` to allow `_add_deeplinking_api` to leave the config untouched (when I split this and do it only for 1.3)

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,49 @@ export default function FilePickerApp({ onSubmit }) {
    * render.
    */
   const [shouldSubmit, setShouldSubmit] = useState(false);
-  const submit = useCallback(() => setShouldSubmit(true), []);
+
+  const [deeplinkingFields, setDeeplinkingFields] = useState(
+    /** @type {Record<string,string>} */ null
+  );
+
+  const submit = useCallback(
```
marcospri, line 121: This is most of the new frontend code, the rest is mostly removing other code.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,49 @@ export default function FilePickerApp({ onSubmit }) {
...
+    async (/** @type {Content} */ content) => {
+      async function getDeeplinkingFields() {
+        const data = {
+          ...deeplinkingAPI.data,
+          content,
+          extra_params: {
```
marcospri, line 127: Mimics the API of the contentItemSelection function.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -187,9 +231,8 @@ export default function FilePickerApp({ onSubmit }) {
         )}
         {content && (
           <FilePickerFormFields
-            ltiLaunchURL={ltiLaunchUrl}
             content={content}
-            formFields={formFields}
+            formFields={{ ...formFields, ...deeplinkingFields }}
```
marcospri, line 235: `formFields` could be merged into deeplinkingFields, I can't see a reason to have both.

lms/static/scripts/frontend_apps/components/FilePickerFormFields.js
```diff
@@ -33,20 +29,13 @@ export default function FilePickerFormFields({
   content,
   formFields,
   groupSet,
-  ltiLaunchURL,
 }) {
-  /** @type {Record<string,string>} */
-  const extraParams = groupSet ? { group_set: groupSet } : {};
```
marcospri, line 39: formFields does now include everything needed for contentitemselection/deeplinking.

lms/static/scripts/frontend_apps/utils/content-item.js
```diff
@@ -18,44 +18,3 @@
  *
  * @typedef {FileContent|URLContent} Content
```
marcospri, line 19: I left the file in for the type definitions but I don't think the compiler likes that

```
TS2306: File '/home/marcos/hypo/lms/lms/static/scripts/frontend_apps/utils/content-item.js' is not a module.
```

lms/views/lti/deeplinking.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @view_config(route_name="lti.v11.deeplinking.formfields")
+    def deeplinking_payload_v11(self):
+        url = self._content_to_url(
+            self.request.parsed_params["content"],
+            self.request.parsed_params["extra_params"],
+        )
+
+        return {
+            "content_items": json.dumps(
```
marcospri, line 65: Build the content item selection payload here.

lms/views/lti/deeplinking.py
```diff
@@ -0,0 +1,91 @@
...
+                            "@type": "LtiLinkItem",
+                            "mediaType": "application/vnd.ims.lti.v1.ltilink",
+                            "url": url,
+                        },
+                    ],
+                }
+            )
+        }
+
+    def _content_to_url(self, content, extra_params):
```
marcospri, line 79: For deeplinking we might produce an URL back to our LTI launch endpoint including extra info about the document.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,51 @@ export default function FilePickerApp({ onSubmit }) {
    * render.
    */
   const [shouldSubmit, setShouldSubmit] = useState(false);
-  const submit = useCallback(() => setShouldSubmit(true), []);
+
+  const [deeplinkingFields, setDeeplinkingFields] = useState(
+    /** @type {Record<string,string>} */ null
```
robertknight, line 118:
```suggestion
    /** @type {Record<string,string>|null} */ (null)
```

For technical reasons, the expressions `@type` annotates have to be wrapped in parens. Also the type should include the `null` possibility.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,51 @@ export default function FilePickerApp({ onSubmit }) {
...
+  );
+
+  const submit = useCallback(
+    async (/** @type {Content} */ content) => {
```
robertknight, line 122: Use:

```js
const submit = useCallback(
  /** @param {Content} content */
  async content => { ... }
, ...);
```

lms/static/scripts/frontend_apps/utils/content-item.js
robertknight: In the client we add this to the bottom of some definition-only files as a workaround:

```js
// Make TypeScript treat this file as a module.
export const unused = {};
```

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,51 @@ export default function FilePickerApp({ onSubmit }) {
...
+            path: deeplinkingAPI.path,
+            data,
+          })
+        );
+      }
+      // When deeplinkingAPI is present we want to call the backend to return the form
+      // fields we'll forward to the LMS to complete the DeepLinking request
+      if (content && deeplinkingAPI && !deeplinkingFields) {
+        try {
+          await getDeeplinkingFields();
```
robertknight, line 143: What does the user see while this request is in-flight? ie. Do they see a loading state and are they prevented from changing anything or re-submitting while this request is in-flight? The `shouldSubmit` state causes a spinner to show, but that is not currently set until after this returns. One way to handle this might be to set the `shouldSubmit` state earlier, but change the `useEffect` that submits the form to check both for `shouldSubmit` being `true` and `deeplinkingFields` being set (if needed).

robertknight: Follow-up question: Does this new API modify any state on the backend? If not we could potentially call it eagerly while waiting for the user to confirm their content (that's out of scope for this PR).

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,7 +113,51 @@ export default function FilePickerApp({ onSubmit }) {
...
+          await apiCall({
+            authToken: deeplinkingAPI.authToken,
+            path: deeplinkingAPI.path,
+            data,
+          })
+        );
+      }
+      // When deeplinkingAPI is present we want to call the backend to return the form
+      // fields we'll forward to the LMS to complete the DeepLinking request
+      if (content && deeplinkingAPI && !deeplinkingFields) {
```
robertknight, line 141: `deeplinkingAPI` is a mandatory field according to the updated `FilePickerConfig` type. Is that correct? If so it doesn't need to be checked for here.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
robertknight: To check my understanding here, `formFields` are static fields which don't depend on the selected content, whereas `deeplinkingFields` are the dynamic fields that do depend on the content?

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -88,7 +89,8 @@ function contentDescription(content) {
 export default function FilePickerApp({ onSubmit }) {
   const submitButton = /** @type {{ current: HTMLInputElement }} */ (useRef());
   const {
-    filePicker: { formAction, formFields, ltiLaunchUrl, blackboard, canvas },
+    deeplinkingAPI,
+    filePicker: { formAction, formFields, blackboard, canvas },
```
robertknight, line 93: According to the types, `deeplinkingAPI` is part of the `filePicker` property rather than the top-level object. Which is correct?

robertknight (commented):
I had an initial look over. This generally makes sense. I left some notes and questions.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: I think it makes sense inside `filePicker`, move it there everywhere.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: I think that distincit might be true currently but at the end they are all (in this PR) transparent values that the frontend forwards back to the backend. If the backend want to separate them in any way I think it should do that on the "data" property of the new API .

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: It won't be present for LTI1.1 launches on the first PR.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: Yeah that might be one distinction between the two.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: Updated following your suggestion (`useEffect that submits the form to check both for shouldSubmit being true and deeplinkingFields being set`)

> Follow-up question: Does this new API modify any state on the backend? If not we could potentially call it eagerly while waiting for the user to confirm their content (that's out of scope for this PR).

It doesn't change any state  at the moment but it might in the near future (ie record the assignment configuration in the database).

Not sure about eagerly doing this, it might complicate things around the JWT, double submissions, it might not be worth it.

lms/resources/_js_config/__init__.py
```diff
@@ -219,21 +219,27 @@ def add_deep_linking_api(self):
         Add the details of the "DeepLinking API" in LMS where we support deep linking.
 
         This API will be used by the frontend to retrieve the form fields required
-        for the deep linking submission while on CONTENT_ITEM_SELECTION mode to store the
+        for the deep linking submission while on FILE_PICKER mode to store the
```
jon-betts, line 222: :+1:

lms/resources/_js_config/__init__.py
```diff
@@ -219,21 +219,27 @@ def add_deep_linking_api(self):
...
         selected content on the LMS.
         """
-        self._config.setdefault("filePicker", {})
-        self._config["filePicker"]["deepLinkingAPI"] = {
-            "path": self._request.route_path("lti.deep_linking.form_fields"),
+        config = {
```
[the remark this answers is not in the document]
jon-betts: Yep. I assumed this was maybe what was going on in the first formulation with the break? It looked like there was a missing other side to the if that I think is here now.

lms/views/lti/deep_linking.py
```diff
@@ -18,14 +18,9 @@
     on this parameter: instead we use a separate URL
     to distinguish deep linking launches.
 
- - In the case of LTI1.3:
+  - We add configuration to enable a call back to `file_picker_to_form_fields_v13`
```
jon-betts, line 21: Should we vague this up a bit to include 1.1?

lms/routes.py
```diff
@@ -109,5 +109,8 @@ def includeme(config):
     config.add_route("lti.oidc", "/lti/1.3/oidc")
     config.add_route("lti.jwks", "/lti/1.3/jwks")
     config.add_route(
-        "lti.deep_linking.form_fields", "/lti/1.3/deep_linking/form_fields"
+        "lti.deep_linking.form_fields.v13", "/lti/1.3/deep_linking/form_fields"
+    )
+    config.add_route(
+        "lti.deep_linking.form_fields.v11", "/lti/1.1/deep_linking/form_fields"
```
jon-betts, line 115: I like the URL here it reads very clearly. Why is the route name different?

lms/views/lti/deep_linking.py
```diff
@@ -137,6 +132,34 @@ def file_picker_to_form_fields_v13(self):
...
+        """
+        Return a JSON-LD `ContentItem` representation of the LTI content.
+
+        See https://www.imsglobal.org/specs/lticiv1p0/specification.
+        """
+        url = self._content_to_url(
+            self.request.route_url("lti_launches"),
+            self.request.parsed_params["content"],
+            self.request.parsed_params.get("extra_params"),
+        )
```
jon-betts, line 146: This is a repeated chunk from above. Maybe all this could move inside `_content_to_url` and this could be `_get_content_url(self.request)` or something?

I think this actually makes it easier to understand as where these values come from is right next to the code that uses them rather than separated as it is now:

```python
    @classmethod
    def _get_content_url(cls, request):
        """
        Translate content information from the frontend to a launch URL.

        We submit the content information to the LMS as an URL pointing to our
        `lti_launches` endpoint with any information required to identity
        the content as query parameters.
        """

        content = request.parsed_params["content"]
        params = dict(request.parsed_params.get("extra_params") or {})

        if content["type"] == "file":
            params["canvas_file"] = "true"
            params["file_id"] = content["file"]["id"]
        elif content["type"] == "url":
            params["url"] = content["url"]
        else:
            raise ValueError(f"Unknown content type: '{content['type']}'")

        return (
            urlparse(request.route_url("lti_launches"))
            ._replace(query=urlencode(params))
            .geturl()
        )
```

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -664,6 +646,43 @@ def routes(self, pyramid_config):
...
+        config = js_config.asdict()
+        assert config["filePicker"]["deepLinkingAPI"] == {
+            "path": "/lti/1.1/deep_linking/form_fields",
+            "data": {
+                "content_item_return_url": sentinel.content_item_return_url,
+            },
+        }
+
+    def test_it_adds_deep_linking_v13(self, js_config, context, application_instance):
+        application_instance.lti_registration_id = 100
```
jon-betts, line 667: This works, but how it works isn't very clear. On first inspection this didn't look like it was related to the test. 

We could use a returnless fixture to make this more obvious? Maybe a comment is all that's required?

```python
    @pytest.mark.usefixtures("with_lti_13")
    def test_it_adds_deep_linking_v13(self, js_config, context):
        ...
        
    @pytest.fixture
    def with_lti_13(self, application_instance):
        # Make the application instance `lti_version` return "1.3.0"
        application_instance.lti_registration_id = 100    
```

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -97,24 +111,29 @@ def test_it(self, ltia_http_service, application_instance, pyramid_request):
...
+        extra_params,
+        ltia_http_service,
+        pyramid_request,
     ):
         pyramid_request.parsed_params.update(
             {"content": content, "extra_params": extra_params}
         )
 
-        DeepLinkingFieldsViews(pyramid_request).file_picker_to_form_fields_v13()
+        fields = getattr(DeepLinkingFieldsViews(pyramid_request), view_name)()
```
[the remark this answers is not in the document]
jon-betts: Yeah sometimes you can't win. I wonder if we are taking a long dance around just testing the private method directly?

If it's convenient, we should do it. If we really want to ensure it links up, we could mock the return value in the previous tests. As a static / class method it should be dirt simple to test.

lms/views/lti/deep_linking.py
```diff
@@ -137,6 +132,34 @@ def file_picker_to_form_fields_v13(self):
             )
         }
 
+    @view_config(route_name="lti.deep_linking.form_fields.v11")
+    def file_picker_to_form_fields_v11(self):
```
jon-betts, line 136: The LTI 1.1 path doesn't use the application instance we look-up in the init, so that query probably isn't getting us anything. This means it's should probably move into the 1.3 method.

jon-betts (commented):
I can't really do more than comment on this at the moment as I'm hitting issues with the 1.3 side of the branch. I'm sure it's something dumb I'm not doing, but the result is this response when I save:

```json
{"errors":{"jwt":[{"attribute":"jwt","type":"JWT verification failure","message":"JWT verification failure"}]}}
```

As for the PR, I think it makes some stuff that was there from the 1.3 branch make a bit more sense, and generally looks sensible to me. There are a couple of observations and ideas:

* We lookup an application instance for both 1.1 and 1.3, but I think it's only used in 1.3, so it's probably a bit wasteful to do it all the time
* There's a little repetition around the content URL method. This might be better if all the request decombobulating happened in the method
* There's a bit of fiddling around a parameterised test to get over the differences between 1.1 and 1.3. I think we might want to mock the inner method in the other two tests and then test it independently. It's a perfectly valid technique, and I think it will make the tests clearer and simpler
```

File: lms/routes.py [hunk 25]
```diff
@@ -109,5 +109,8 @@ def includeme(config):
     config.add_route("lti.oidc", "/lti/1.3/oidc")
     config.add_route("lti.jwks", "/lti/1.3/jwks")
     config.add_route(
-        "lti.deep_linking.form_fields", "/lti/1.3/deep_linking/form_fields"
+        "lti.v13.deep_linking.form_fields", "/lti/1.3/deep_linking/form_fields"
```
marcospri, line 112: Changed the route name to match the URL

jon-betts:
Just wanted to pop a note on here to say we resolved the issue I was having. Looks like I had some old RSA keys around from a previous PR which was causing the trouble.

jon-betts (approved):
Yeah I think this is looking really good.

I wonder if we are shooting ourselves in the foot by not using this technique more often. I think we've had an unofficial rule of not testing private methods, but I think it might be causing us to miss opportunities as:

* Tests are already intimately aware of the internals of our classes due to our mocking style
* We often follow line for line, call for call, how the function works
* Therefore there is no real private/public divide between a class and it's tests (it's white box all the way)
* Pulling out a common function is a good way to reduce code repetition, but we aren't making use of the same in the tests

I think the crucial part to hook this up is the mocking part which marries the two tests together. Without this we would not know that the private tests relate to the public ones. This linkage gives us the ability to do this.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -142,7 +135,7 @@ export default function FilePickerApp({ onSubmit }) {
           ...deepLinkingAPI.data,
           content,
           extra_params: {
-            groupSet: groupConfig.useGroupSet ? groupConfig.groupSet : null,
+            group_set: groupConfig.useGroupSet ? groupConfig.groupSet : null,
```
marcospri, line 138: This was a bug in the merged PR, the name the backend is expecting is `group_set`

robertknight (approved):

marcospri:
Added https://github.com/hypothesis/lms/pull/3908/commits/0bbda441fdf4d15ec68012a45e8041c0ef8e0cfd 

since last reviewed


Without this commit non-groups assignments would produce a lti launch URL containing `?group_set=None`. 

This could arguably done in the frontend but I think is a potentially common error to make (send a null value to the backend) and we guard against it in the future doing it on the backend.

Issue #3907, opened 2022-05-03
Investigate assisting support with mapping JSTOR ids in LMS


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/110
 
We have a list of JSTOR ids. It would be nice to see how we can help support find the correct values for JSTOR ids. Some questions to see if we can answer:

 * Can we effectively map values?
 * Are JSTOR ok with us keeping this data in any form?
 * Would we be able to update it?
 * Could we provide a one off mapping / script / integration with our admin UI cheaply?

PR #3910, opened 2022-05-03
Example code for matching JSTOR ids

jon-betts:
For: 

 * https://github.com/hypothesis/lms/issues/3907

### Effectiveness

The current code spits out the following assessment of it's own behavior:

```
'perfect id': 2711, 
'perfect name+id': 802, 
'perfect name': 488, 
'excellent': 83, 
'good': 209, 
'ok': 738, 
'dubious': 60,
'fail': 300, 
```

This breaks down to roughly:

 * 4001/5391 - 74% - A perfect id or name match
 * 1090/5391 - 20% - A suggestion of some sort
 * 300/5391 - 6% - No match

Not that a "perfect" match here doesn't imply it's correct. For example one record has a perfect match against a record with the name "OLD VALUE DO NOT USE". There is no way around this until we have some human verified records to compare against.

This is likely over matching, as quite a lot of our installations probably have no JSTOR id to match against. This suggests the effectiveness on real Unis where a match is there to be had could be higher.


robertknight:
Closed per https://hypothes-is.slack.com/archives/C4K6M7P5E/p1654008023282139.

PR #3912, opened 2022-05-03
Jstor passthru

jon-betts:


lms/services/jstor.py
```diff
@@ -2,27 +2,43 @@
 
 
 class JSTORService:
-    def __init__(self, settings):
-        self._settings = settings
+    def __init__(self, enabled, site_code):
```
marcospri, line 5: I don't remember why I was so set on passing the whole settings here. It event took some fighting with ApplicationSettings.

In any case, having the params explicitly declared is nicer than taking a random dict without known shape :+1:

lms/services/jstor.py
```diff
@@ -2,27 +2,43 @@
...
+        :param enabled: Whether JSTOR is enabled on this instance
+        :param site_code: The site code to use to identify the organization
+        """
+        self._enabled = enabled
+        self._site_code = site_code
 
     @property
-    def enabled(self):
-        return self._settings.get("enabled", False)
+    def enabled(self) -> bool:
```
marcospri, line 16: Should we  call this from:

https://github.com/hypothesis/lms/blob/jstor-save-site-code/lms/resources/_js_config/file_picker_config.py#L108

jon-betts: Yes. Yes we should

marcospri (approved):

Issue #3909, opened 2022-05-03
Validate the LTI1.3 implementation on Moodle


marcospri:
So far we tested different parts of the implementation in Canvas, Blackboard and the IMS certification tool.

Test the the full LTI1.3 integration, fix any bugs that may arise and produce a document to replicate the installation and tests.

marcospri:
## Adding a LTI1.3 tool in moodle

- Login as "admin" at https://hypothesisuniversity.moodlecloud.com/login/index.php
- Go to Site Administration -> Plugins -> Activity Modules -> External Tool -> Manage Tools.

(or directly to: https://hypothesisuniversity.moodlecloud.com/mod/lti/toolconfigure.php)

- Click "Configure a tool manually"
- Enter the configuration details:

![Screenshot from 2022-05-04 12-28-47](https://user-images.githubusercontent.com/1433832/166664945-854a2b26-f91a-47c2-9f22-0d34d7071db0.png)


- Using the details after setup, create a LTIRegistration & ApplicationInstance (already done in QA)

![Screenshot from 2022-05-04 12-27-45](https://user-images.githubusercontent.com/1433832/166665021-6bacce1d-e0e5-41a0-98d7-de4d95a68af1.png)


- Create tests assignments following https://web.hypothes.is/help/creating-hypothesis-enabled-readings-in-moodle/

marcospri:
As far as we know LTI1.3 works fine in Moodle.

PR #3913, opened 2022-05-04
Build correct grading service URLs in Moodle

marcospri:
For https://github.com/hypothesis/lms/issues/3909

A small fix to correctly build URLs, just appending `/something` won't work in Moodle.

### DIFF
--- a/lms/services/lti_grading/_v13.py
+++ b/lms/services/lti_grading/_v13.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timezone
+from urllib.parse import urlparse
 
 from lms.services.exceptions import ExternalRequestError
 from lms.services.lti_grading.interface import LTIGradingService
@@ -22,7 +23,7 @@ class LTI13GradingService(LTIGradingService):
         try:
             response = self._ltia_service.request(
                 "GET",
-                self.grading_url + "/results",
+                self._service_url(self.grading_url, "/results"),
                 scopes=self.LTIA_SCOPES,
                 params={"user_id": grading_id},
                 headers={"Accept": "application/vnd.ims.lis.v2.resultcontainer+json"},
@@ -44,7 +45,7 @@ class LTI13GradingService(LTIGradingService):
     def record_result(self, grading_id, score=None):  # pylint:disable=arguments-differ
         return self._ltia_service.request(
             "POST",
-            self.grading_url + "/scores",
+            self._service_url(self.grading_url, "/scores"),
             scopes=self.LTIA_SCOPES,
             json={
                 "scoreMaximum": 1,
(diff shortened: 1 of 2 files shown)


lms/services/lti_grading/_v13.py
```diff
@@ -54,5 +55,20 @@ def record_result(self, grading_id, score=None):  # pylint:disable=arguments-dif
                 "activityProgress": "Completed",
                 "gradingProgress": "FullyGraded",
             },
-            headers={"Content-Type": "application/vnd.ims.lis.v2.lineitem+json"},
+            headers={"Content-Type": "application/vnd.ims.lis.v1.score+json"},
```
marcospri, line 58: This values was wrong: https://www.imsglobal.org/spec/lti-ags/v2p0#example-posting-a-final-score-update

I guess both blackboard and IMS didn't care about that. Moodle does. I retested in the other ones and still works with the right value.

tests/unit/lms/services/lti_grading/_v13_test.py
```diff
@@ -94,4 +118,4 @@ def response(self):
 
     @pytest.fixture
     def svc(self, ltia_http_service):
-        return LTI13GradingService(self.SERVICE_URL, ltia_http_service)
+        return LTI13GradingService("http://example.com/service_url", ltia_http_service)
```
marcospri, line 121: Can't be a `sentinel`. Needs to be a string. But we'd only use `SERVICE_URL` here

lms/services/lti_grading/_v13.py
```diff
@@ -22,7 +23,7 @@ def read_result(self, grading_id):
         try:
             response = self._ltia_service.request(
                 "GET",
-                self.grading_url + "/results",
+                self._service_url(self.grading_url, "/results"),
                 scopes=self.LTIA_SCOPES,
                 params={"user_id": grading_id},
```
jon-betts, line 28: Note to self: I've given this a quick check and requests will take care of merging the params here into those on the provided URL

tests/unit/lms/services/lti_grading/_v13_test.py
```diff
@@ -8,16 +8,28 @@
...
+    @pytest.mark.parametrize(
+        "service_url,results_url",
+        [
+            ("https://lms.com/lineitems", "https://lms.com/lineitems/results"),
+            (
+                "https://lms.com/lineitems?param=1",
+                "https://lms.com/lineitems/results?param=1",
+            ),
+        ],
+    )
```
jon-betts, line 21: I think you could just use the more difficult example here, as it's a superset of everything going on in the other example.

tests/unit/lms/services/lti_grading/_v13_test.py
```diff
@@ -60,12 +72,24 @@ def test_read_bad_response_lti_result(self, svc, ltia_http_service, bad_response
...
+    @pytest.mark.parametrize(
+        "service_url,scores_url",
+        [
+            ("https://lms.com/lineitems", "https://lms.com/lineitems/scores"),
+            (
+                "https://lms.com/lineitems?param=1",
+                "https://lms.com/lineitems/scores?param=1",
+            ),
+        ],
+    )
```
jon-betts, line 84: Same here. The second case covers the first as well.

jon-betts (approved):
Yep this looks good.

I think the tests could be slimmed down to the example with the param instead of parameterising, as it covers everything about the previous case as well.

Issue #3911, opened 2022-05-03
Deep-linking support in LTI1.3


marcospri:
Implement deep linking support for LTI1.3 using the corresponding LTIA API.

I think this can be achieved in two steps:

- Include a new API that return the deeplinking form fields that frontend need to submit to Canvas for deeplinking for LTI1.3

In LTI1.3 this happends to be only one parameter an "JWT" with all the required information.

- Extend the same mechanisms to work for LTI1.1

This will be mostly a refactoring effort. It will consist in moving most of what https://github.com/hypothesis/lms/blob/master/lms/static/scripts/frontend_apps/utils/content-item.js does to the backend.

- Having both version in place will make tracking canvas assignments plausible as there will be now a backend part involved in assignment creation. This will required further work not tied to this issue.

PR #3917, opened 2022-05-04
Deep Linking in LTI1.3 launches

marcospri:
For: #3911 

First half of https://github.com/hypothesis/lms/pull/3908

Here we only make changes for LTI1.3 launches:

- We set a new "deeplinkingAPI" field in the config the backend sends to the frontend.
- When that's present the frontend calls the new API with the content selected by the user and the backend answers with the form fields required for the deeplinking form submission. For LT1.3 that's only a `JWT` field with all the encoded fields in it.
- The frontend submits the form to canvas as usual.

## Testing 

Log into canvas as a teacher

### LTI1.3

- `make devdata` 
- Edit https://hypothesis.instructure.com/courses/319/assignments/2781  using the `Hypothesis LTI1.3 (localhost)` install

There will be now a call to `http://localhost:8001/lti/1.3/deeplinking/formfields` showing up on the network tab of your browser.  This happens right after the last "continue" button in the configuring assignment screen.

- Saving the assignment will store the new document.

### LTI1.1

- Nothing should have change here, edit a LTI1.1 assignment, https://hypothesis.instructure.com/courses/125/assignments/873

- Not call to http://localhost:8001/lti/1.3/deeplinking/formfields

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -214,6 +214,27 @@ class JSConfig:
             }
         )
 
+    def add_deep_linking_api(self):
+        """
+        Add the details of the "DeepLinking API" in LMS where we support deep linking.
+
+        This API will be used by the frontend to retrieve the form fields required
+        for the deep linking submission while on CONTENT_ITEM_SELECTION mode to store the
+        selected content on the LMS.
+        """
+        self._config.setdefault("filePicker", {})
+        self._config["filePicker"]["deepLinkingAPI"] = {
+            "path": self._request.route_path("lti.deep_linking.form_fields"),
+            "data": {
+                "content_item_return_url": self._context.lti_params[
+                    "content_item_return_url"
+                ],
+                "deep_linking_settings": self._context.lti_params.get(
+                    "deep_linking_settings"
+                ),
+            },
+        }
+
     def maybe_enable_grading(self):
         """Enable our LMS app's built-in assignment grading UI, if appropriate."""
 
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -108,3 +108,6 @@ def includeme(config):
 
     config.add_route("lti.oidc", "/lti/1.3/oidc")
     config.add_route("lti.jwks", "/lti/1.3/jwks")
+    config.add_route(
+        "lti.deep_linking.form_fields", "/lti/1.3/deep_linking/form_fields"
+    )
--- a/lms/services/ltia_http.py
+++ b/lms/services/ltia_http.py
@@ -15,6 +15,27 @@ class LTIAHTTPService:
         self._jwt_service = jwt_service
         self._http = http
 
+    def sign(self, payload: dict):
+        """
+        Sign a payload with one of our private keys.
+
+        We include the default values needed for LTIA APIs
+        Those default claims are defined here:
+
+            https://www.imsglobal.org/spec/security/v1p0/#id-token
+        """
+        now = datetime.utcnow()
+        payload = {
+            "exp": now + timedelta(hours=1),
+            "iat": now,
+            "nonce": uuid.uuid4().hex,
+            "iss": self._lti_registration.client_id,
+            "sub": self._lti_registration.client_id,
+            "aud": self._lti_registration.issuer,
+            **payload,
+        }
+        return self._jwt_service.encode_with_private_key(payload)
+
     def request(self, method, url, scopes, headers=None, **kwargs):
         headers = headers or {}
 
@@ -32,16 +53,10 @@ class LTIAHTTPService:
         https://datatracker.ietf.org/doc/html/rfc7523
         https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token
         """
-        now = datetime.utcnow()
-        signed_jwt = self._jwt_service.encode_with_private_key(
+        signed_jwt = self.sign(
             {
-                "exp": now + timedelta(hours=1),
-                "iat": now,
-                "nonce": uuid.uuid4().hex,
-                "jti": uuid.uuid4().hex,
-                "iss": self._lti_registration.client_id,
-                "sub": self._lti_registration.client_id,
                 "aud": self._lti_registration.token_url,
+                "jti": uuid.uuid4().hex,
             }
         )
 
--- a/lms/static/scripts/frontend_apps/components/FilePickerApp.js
+++ b/lms/static/scripts/frontend_apps/components/FilePickerApp.js
@@ -8,6 +8,7 @@ import {
 } from 'preact/hooks';
 
 import { Config } from '../config';
+import { apiCall } from '../utils/api';
 import { truncateURL } from '../utils/format';
 
 import ContentSelector from './ContentSelector';
@@ -88,7 +89,15 @@ function contentDescription(content) {
 export default function FilePickerApp({ onSubmit }) {
   const submitButton = /** @type {{ current: HTMLInputElement }} */ (useRef());
   const {
-    filePicker: { formAction, formFields, ltiLaunchUrl, blackboard, canvas },
+    api: { authToken },
+    filePicker: {
+      deepLinkingAPI,
+      formAction,
+      formFields,
+      ltiLaunchUrl,
+      blackboard,
+      canvas,
+    },
   } = useContext(Config);
 
   const [content, setContent] = useState(/** @type {Content|null} */ (null));
@@ -111,25 +120,79 @@ export default function FilePickerApp({ onSubmit }) {
    * render.
    */
   const [shouldSubmit, setShouldSubmit] = useState(false);
-  const submit = useCallback(() => setShouldSubmit(true), []);
+
+  const [deepLinkingFields, setDeepLinkingFields] = useState(
+    /** @type {Record<string,string>|null} */ (null)
+  );
+
+  const submit = useCallback(
+    /** @param {Content} content */
+    async content => {
+      // Set shouldSubmit to true early to show the spinner while fetching form fields
+      setShouldSubmit(true);
+
+      if (!deepLinkingAPI || deepLinkingFields) {
+        return;
+      }
+
+      // When deepLinkingAPI is present we want to call the backend to return the form
+      // fields we'll forward to the LMS to complete the Deep Linking request
+      try {
+        const data = {
+          ...deepLinkingAPI.data,
+          content,
+          extra_params: {
+            groupSet: groupConfig.useGroupSet ? groupConfig.groupSet : null,
+          },
+        };
+        setDeepLinkingFields(
+          await apiCall({
+            authToken: authToken,
+            path: deepLinkingAPI.path,
+            data,
+          })
+        );
+      } catch (error) {
+        setErrorInfo({
+          message: 'Unable to configure assignment',
+          error: error,
+        });
+        // Reset the state in case of an error allowing to start over
+        setShouldSubmit(false);
+        setContent(null);
+      }
+    },
+
+    [
+      authToken,
+      deepLinkingFields,
+      deepLinkingAPI,
+      groupConfig.groupSet,
+      groupConfig.useGroupSet,
+    ]
+  );
 
   // Submit the form after a selection is made via one of the available
   // methods.
   useEffect(() => {
-    if (shouldSubmit) {
+    if (
+      shouldSubmit &&
+      // We either are not using the deepLinkingAPI, or if we are, wait for deepLinkingFields to be available
+      (!deepLinkingAPI || (deepLinkingAPI && deepLinkingFields))
+    ) {
       // Submit form using a hidden button rather than calling `form.submit()`
       // to facilitate observing the submission in tests and suppressing the
       // actual submit.
       submitButton.current.click();
     }
-  }, [shouldSubmit]);
+  }, [shouldSubmit, deepLinkingAPI, deepLinkingFields]);
 
   /** @type {(c: Content) => void} */
   const selectContent = useCallback(
     content => {
       setContent(content);
       if (!enableGroupConfig) {
-        submit();
+        submit(content);
       }
     },
     [enableGroupConfig, submit]
@@ -178,7 +241,7 @@ export default function FilePickerApp({ onSubmit }) {
               <LabeledButton
                 disabled={groupConfig.useGroupSet && !groupConfig.groupSet}
                 variant="primary"
-                onClick={submit}
+                onClick={() => submit(content)}
               >
                 Continue
               </LabeledButton>
@@ -189,7 +252,7 @@ export default function FilePickerApp({ onSubmit }) {
           <FilePickerFormFields
             ltiLaunchURL={ltiLaunchUrl}
             content={content}
-            formFields={formFields}
+            formFields={{ ...formFields, ...deepLinkingFields }}
             groupSet={groupConfig.useGroupSet ? groupConfig.groupSet : null}
           />
         )}
--- a/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
@@ -2,6 +2,7 @@
 
 import { act } from 'preact/test-utils';
 import { mount } from 'enzyme';
+import { waitFor, waitForElement } from '../../../test-util/wait';
 
 import { Config } from '../../config';
 import FilePickerApp, { $imports } from '../FilePickerApp';
@@ -31,6 +32,7 @@ describe('FilePickerApp', () => {
 
   beforeEach(() => {
     fakeConfig = {
+      api: { authToken: 'DUMMY_AUTH_TOKEN' },
       filePicker: {
         formAction: 'https://www.shinylms.com/',
         formFields: { hidden_field: 'hidden_value' },
@@ -58,12 +60,17 @@ describe('FilePickerApp', () => {
   /**
    * Check that the expected hidden form fields were set.
    */
-  function checkFormFields(wrapper, expectedContent, expectedGroupSet) {
+  function checkFormFields(
+    wrapper,
+    expectedContent,
+    expectedGroupSet,
+    extraFormFields = {}
+  ) {
     const formFields = wrapper.find('FilePickerFormFields');
     assert.deepEqual(formFields.props(), {
       children: [],
       content: expectedContent,
-      formFields: fakeConfig.filePicker.formFields,
+      formFields: { ...fakeConfig.filePicker.formFields, ...extraFormFields },
       groupSet: expectedGroupSet,
       ltiLaunchURL: fakeConfig.filePicker.ltiLaunchUrl,
     });
@@ -132,6 +139,80 @@ describe('FilePickerApp', () => {
       selectContent(wrapper, 'https://example.com');
 
       assert.isTrue(wrapper.exists('FullScreenSpinner'));
+    });
+  });
+
+  context('when deepLinkingAPI is present', () => {
+    const deepLinkingAPIPath = '/lti/1.3/deep_linking/form_fields';
+    const deepLinkingAPIData = { some: 'data' };
+    let fakeAPICall;
+    let fakeFormFields;
+
+    beforeEach(() => {
+      fakeConfig.filePicker.deepLinkingAPI = {
+        path: deepLinkingAPIPath,
+        data: deepLinkingAPIData,
+      };
+
+      fakeAPICall = sinon.stub();
+      fakeFormFields = { JWT: 'JWT VALUE' };
+
+      fakeAPICall
+        .withArgs(sinon.match({ path: deepLinkingAPIPath }))
+        .resolves(fakeFormFields);
+
+      $imports.$mock({
+        '../utils/api': { apiCall: fakeAPICall },
+      });
+    });
+
+    it('fetches form field values via deep linking API when content is selected', async () => {
+      const onSubmit = sinon.stub().callsFake(e => e.preventDefault());
+      const wrapper = renderFilePicker({ onSubmit });
+
+      selectContent(wrapper, 'https://example.com');
+
+      await waitFor(() => fakeAPICall.called);
+      assert.calledWith(fakeAPICall, {
+        authToken: 'DUMMY_AUTH_TOKEN',
+        path: deepLinkingAPIPath,
+        data: {
+          ...deepLinkingAPIData,
+          content: { type: 'url', url: 'https://example.com' },
+          extra_params: { groupSet: null },
+        },
+      });
+
+      await waitFor(() => onSubmit.called, 100);
+
+      wrapper.update();
+      checkFormFields(
+        wrapper,
+        {
+          type: 'url',
+          url: 'https://example.com',
+        },
+        null /* groupSet */,
+        fakeFormFields
+      );
+    });
+
+    it('shows an error if the deepLinkingAPI call fails', async () => {
+      const error = new Error('Something happened');
+      const onSubmit = sinon.stub().callsFake(e => e.preventDefault());
+      fakeAPICall
+        .withArgs(sinon.match({ path: deepLinkingAPIPath }))
+        .rejects(error);
+
+      const wrapper = renderFilePicker({ onSubmit });
+
+      selectContent(wrapper, 'https://example.com');
+
+      await waitForElement(wrapper, 'ErrorModal');
+
+      const errDialog = wrapper.find('ErrorModal');
+      assert.equal(errDialog.length, 1);
+      assert.equal(errDialog.prop('error'), error);
     });
   });
 
--- a/lms/static/scripts/frontend_apps/config.js
+++ b/lms/static/scripts/frontend_apps/config.js
@@ -71,6 +71,7 @@ import { createContext } from 'preact';
  * @typedef FilePickerConfig
  * @prop {string} formAction
  * @prop {Record<string,string>} formFields
+ * @prop {APICallInfo} [deepLinkingAPI]
  * @prop {string} ltiLaunchUrl
  * @prop {object} blackboard
  *   @prop {boolean} blackboard.enabled
--- a/lms/views/lti/deep_linking.py
+++ b/lms/views/lti/deep_linking.py
@@ -1,43 +1,57 @@
 """
-Views for handling deep linking launches.
+Deep Linking related views.
 
-A deep linking request is an LTI launch request (so it's a form
-submission containing all the usual launch params, including the OAuth 1
-signature) that's used by LMS's during the assignment creation process.
+If an LMS (like Canvas) is configured for "deep linking" then we need to send
+the results of the file picker to the LMS instead of storing it ourselves. This
+is done by the front-end.
 
-When an instructor creates a Hypothesis assignment in an LMS that supports
-deep linking the LMS launches our app in order for us to present an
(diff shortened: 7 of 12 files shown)


lms/services/ltia_http.py
```diff
@@ -15,6 +15,21 @@ def __init__(
         self._jwt_service = jwt_service
         self._http = http
 
+    def sign(self, payload: dict):
```
marcospri, line 18: This becomes its own method.  The default for "aud" here is the issuer.

lms/services/ltia_http.py
```diff
@@ -32,18 +47,7 @@ def _get_access_token(self, scopes):
...
-                "exp": now + timedelta(hours=1),
-                "iat": now,
-                "nonce": uuid.uuid4().hex,
-                "jti": uuid.uuid4().hex,
-                "iss": self._lti_registration.client_id,
-                "sub": self._lti_registration.client_id,
-                "aud": self._lti_registration.token_url,
-            }
-        )
+        signed_jwt = self.sign({"aud": self._lti_registration.token_url})
```
marcospri, line 50: For new tokens, the `aud` is the URL of the endpoint instead.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -1,4 +1,5 @@
 import { FullScreenSpinner, LabeledButton } from '@hypothesis/frontend-shared';
+import { apiCall } from '../utils/api';
```
robertknight, line 2: Can you move this to above the `../utils/format` import below. Imports are grouped into third party, then local. Within a group, imports are sorted by path.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,25 +120,77 @@ export default function FilePickerApp({ onSubmit }) {
...
+        };
+        setDeeplinkingFields(
+          await apiCall({
+            authToken: authToken,
+            path: deeplinkingAPI.path,
+            data,
+          })
+        );
+      }
+      // Set shouldSubmit to true early to get the spinner before any potential apiCall
```
robertknight, line 147:
```suggestion
      // Set shouldSubmit to true early to show the spinner while fetching form fields
```

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,25 +120,77 @@ export default function FilePickerApp({ onSubmit }) {
...
+      setShouldSubmit(true);
+
+      // When deeplinkingAPI is present we want to call the backend to return the form
+      // fields we'll forward to the LMS to complete the DeepLinking request
+      if (content && deeplinkingAPI && !deeplinkingFields) {
+        try {
+          await getDeeplinkingFields();
+        } catch (error) {
+          setErrorInfo({
```
robertknight, line 156: What does the current error UI look like? I'm wondering if we need to reset the `shouldSubmit` state here.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -71,6 +71,7 @@ import { createContext } from 'preact';
  * @typedef FilePickerConfig
  * @prop {string} formAction
  * @prop {Record<string,string>} formFields
+ * @prop {APICallInfo} deeplinkingAPI
```
robertknight, line 74: It looks like this is optional in the backend? If so this should be `[deeplinkingAPI]`, which is equivalent to setting the type to `APICalInfo|undefined`. This will result in an error if you try to use it without checking if it is set.

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -135,6 +142,80 @@ describe('FilePickerApp', () => {
...
+      fakeAPICall
+        .withArgs(sinon.match({ path: deeplinkingAPIPath }))
+        .resolves(fakeFormFields);
+
+      $imports.$mock({
+        '../utils/api': { apiCall: fakeAPICall },
+      });
+    });
+
+    it('calls deeplinkingAPI when content is selected', async () => {
```
robertknight, line 169:
```suggestion
    it('fetches form field values via deeplinking API when content is selected', async () => {
```

This test covers both calling the API and populating the form with the response, so I think it is helpful for the description to mention that.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,25 +120,77 @@ export default function FilePickerApp({ onSubmit }) {
...
+            message: 'Creating deeplinking form fields',
```
robertknight, line 157: The message for an error should be a brief description of what failed in the past tense, and something that is short but meaningful to users. For example: "Unable to configure assignment". The details of the error should be included elsewhere in the error modal since you're passing error data.

robertknight (commented):
I've done a first pass over this and requested some minor changes. Overall it looks good though. I will do some more manual testing in case I spot any other issues.

lms/resources/_js_config/__init__.py
```diff
@@ -538,3 +539,23 @@ def _sync_api(self):
...
+            },
+        }
+
+        if self._application_instance.lti_version == "1.3.0":
+            config["path"] = self._request.route_path("lti.deeplinking.formfields")
+            config["data"]["deep_linking_settings"] = self._context.lti_params[
+                "deep_linking_settings"
+            ]
+
+            self._config["filePicker"]["deeplinkingAPI"] = config
```
jon-betts, line 561: Is this a bug? Why doesn't this start with:

```
if not self._context.is_canvas or self._application_instance.lti_version != "1.3.0":
    return
```

Should this line be out-dented? Did the tests miss this?

lms/services/ltia_http.py
jon-betts: Doc strings again

lms/resources/_js_config/__init__.py
```diff
@@ -538,3 +539,23 @@ def _sync_api(self):
             return self._blackboard_sync_api()
 
         return None
+
+    def _add_deeplinking_api(self):
```
jon-betts, line 543: What is deeplinking? Why is it added as part of `enable_content_item_selection_mode`? Comments and doc-strings would help.

Is this always canvas only, or are we just enabling it for canvas for the moment? What part does it play in a Canvas launch?

jon-betts (changes requested):
I think this might need a bit more attention. There are no comments or docstrings at all in the new code on the Python side. I do not really understand what deeplinking is, why we need it, what it does for us or why it's only used for Canvas. This is all information I'd hope to be told by reading through the code, rather than trying to guess.

We cannot expect everyone who looks at the code to be an expert in the LTI spec, what all the terminology means as well as our decisions about when we apply it and why. Given how looping and confusing LMS is, I don't think documenting our code is optional.

There's also a piece of code in the config portion which is either factored very strangely or might contain a bug. I can't quite tell.

lms/resources/_js_config/__init__.py
marcospri: No, it's doing what's is supposed to do but it's organized weirdly as I split this from a PR that did this for both 1.1 and 1.3.

Moved the 1.3 conditional to the early return for this  PR :+1:

lms/resources/_js_config/__init__.py
marcospri: I do know why we need to support deeplinking on LTI1.3 (to behave the same in both LTI versions) but I don't have a good answer to why we only use it canvas in LTI1.1.

I think the real reason might be that the first implementation supporting only canvas did and that defined what's expected of the integration there.

lms/resources/_js_config/__init__.py
```diff
@@ -538,3 +539,32 @@ def _sync_api(self):
...
+        """
+        Add the details of the "Deeplinking API" in LMS where we support deeplinking.
+
+        This API will be used by the frontend to retrieve the form fields required
+        for the deeplinking submission while on CONTENT_ITEM_SELECTION mode to store the
+        selected content on the LMS.
+        """
+        if (
+            not self._context.is_canvas
+            or self._application_instance.lti_version != "1.3.0"
```
marcospri, line 553: Only 1.3 to split the PRs, this one adding the new API and FE behavior.

The next one using it on LTI1.1 (which will remove some FE code)

lms/services/ltia_http.py
```diff
@@ -32,16 +53,10 @@ def _get_access_token(self, scopes):
...
+        signed_jwt = self.sign(
             {
-                "exp": now + timedelta(hours=1),
-                "iat": now,
-                "nonce": uuid.uuid4().hex,
-                "jti": uuid.uuid4().hex,
-                "iss": self._lti_registration.client_id,
-                "sub": self._lti_registration.client_id,
                 "aud": self._lti_registration.token_url,
+                "jti": uuid.uuid4().hex,
```
marcospri, line 59: These key/values are specific to the token request.

jon-betts (commented):
I think one of the things coming out of the discussion around this is there's a difference between the concepts of:

 * "showing the file picker" 
 * "storing the return result of the picker ourselves"
*  "LTI content item selection / deeplinking storing the return result of the picker in the LMS"

And our terminology is kind of "content item selection" for all of these processes which leads to some confusion.

### I don't think we should add the config in the file picker launching function

Summarising our chat basically, but as a result of this, it seems like the only place that should add the deeplinking config is the deeplinking view. I think that adding it to the show the file picker function kind of accidentally adds the config in places it's needed while also adding them where it's not.

This should hopefully cut down on the conditionality, as we'll know we should add it by context.

I'm 55% of the mind that setting this config would be better if it was done inline in the view, as it tells a good story about what the flow is.

* Call this
* I configure you to call "that"
* "that" is the next thing in the file

### Other thoughts

 * Still some strange formatting around the config setup. Looks like it's setup for some conditionality that isn't there
 * The big comment at the top of the file is kind of a general chat on the theme of deeplinking, rather than focussed on what it means in our app, how it works and what you need to know
 * I don't really understand why we have a chunk of code to re-translate a bunch of data coming back from our own front-end. Why doesn't it sent the format we want instead?
 * The LTI1.3 flow is more obvious now, but what happens to LTI1.1. Why doesn't it have to call us back?

lms/resources/_js_config/__init__.py
```diff
@@ -538,3 +539,32 @@ def _sync_api(self):
...
+                ],
+            },
+        }
+
+        config["path"] = self._request.route_path("lti.deeplinking.formfields")
+        config["data"]["deep_linking_settings"] = self._context.lti_params[
+            "deep_linking_settings"
+        ]
+
+        self._config["filePicker"]["deeplinkingAPI"] = config
```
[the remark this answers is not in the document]
marcospri: It's strange because in the follow up PR it changes some bits only for LTI1.1. 

I'll commit your version as it makes more sense standalone and it's not a big change later.

lms/views/lti/deeplinking.py
```diff
@@ -58,11 +68,83 @@ def content_item_selection(context, request):
...
+                "https://purl.imsglobal.org/spec/lti/claim/message_type": "LtiDeepLinkingResponse",
+                "https://purl.imsglobal.org/spec/lti/claim/version": "1.3.0",
+                "https://purl.imsglobal.org/spec/lti-dl/claim/content_items": [
+                    {"type": "ltiResourceLink", "url": url}
+                ],
+                "https://purl.imsglobal.org/spec/lti-dl/claim/data": self.request.parsed_params[
+                    "deep_linking_settings"
+                ],
+            }
+        )
```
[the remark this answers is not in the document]
marcospri: Not really because `lti_signature` is already a jwt token. Inlined it.

lms/views/lti/deeplinking.py
```diff
@@ -58,11 +68,83 @@ def content_item_selection(context, request):
...
+
+    def __init__(self, request):
+        self.request = request
+
+        self.application_instance = self.request.find_service(
+            name="application_instance"
+        ).get_current()
+
+    @view_config(route_name="lti.deeplinking.formfields")
+    def formfields_v13(self):
```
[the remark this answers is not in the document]
marcospri: >  the front-end uses this to convert our standard file-picker output back into a set of data it can pass on to LMS to complete the deeplinking request?

Yes, exactly

I went for `file_picker_to_formfields_v13` (we are already on deeplinking.py).

lms/views/lti/deeplinking.py
```diff
@@ -2,42 +2,51 @@
...
+  https://www.imsglobal.org/specs/lticiv1p0/specification-3
 
-Especially this page:
+- LTI 1.3 https://www.imsglobal.org/spec/lti-dl/v2p0
 
-https://www.imsglobal.org/specs/lticiv1p0/specification-3
 
 Canvas LMS's Content Item docs are also useful:
 
 https://canvas.instructure.com/doc/api/file.content_item.html
```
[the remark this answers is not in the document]
marcospri: Used a combination or yours and the existing one :+1:

lms/views/lti/deeplinking.py
```diff
@@ -58,11 +68,83 @@ def content_item_selection(context, request):
...
 
     context.js_config.enable_content_item_selection_mode(
-        form_action=request.params["content_item_return_url"],
+        form_action=request.parsed_params["content_item_return_url"],
         form_fields={
             "lti_message_type": "ContentItemSelection",
-            "lti_version": request.params["lti_version"],
+            "lti_version": request.parsed_params["lti_version"],
         },
     )
```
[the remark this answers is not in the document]
marcospri: I now prefer the "set-the-config-in-the-view" approach. It simplifies the logic (no need to check for is canvas).

lms/views/lti/deeplinking.py
```diff
@@ -58,11 +68,83 @@ def content_item_selection(context, request):
...
+    After the user picks a document to be annotated the frontend will call these views
+    to get all the necessary fields to configure the assignment submitting them in a form to the LMS.
+    """
+
+    def __init__(self, request):
+        self.request = request
+
+        self.application_instance = self.request.find_service(
+            name="application_instance"
+        ).get_current()
```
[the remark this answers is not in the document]
marcospri: Yes I vaguely remember having that in one of the first LTI PRs but it was mixed with something else.

Made a note to revisit it.

lms/views/lti/deeplinking.py
```diff
@@ -58,11 +68,83 @@ def content_item_selection(context, request):
...
+
+        params = dict(extra_params or {})
+
+        content = self.request.parsed_params["content"]
+        if content["type"] == "file":
+            params["canvas_file"] = "true"
+            params["file_id"] = content["file"]["id"]
+        else:
+            assert content["type"] == "url", "Unknown content type"
+            params["url"] = content["url"]
```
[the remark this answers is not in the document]
marcospri: That would be ideal and made a note in my post-LTI cleanup for it but I rather to leave it out of this PR as it might involve some changes around all different file pickers in the frontend side.

lms/views/lti/deeplinking.py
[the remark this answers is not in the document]
marcospri: > * In the case of LTI1.1:
    * ??? TBA - I'm not really sure what happens here?

There's another follow up PR (mentioned on the PR description but it needs heavy rebasing) to follow the same approach in LTI1.1.

In the backend it mostly adding the v11 view which instead of a JWT returns a few json fields. Currently the frontend constructs that message itself.

On the frontend it will involve mostly removing code but I wanted to have this PR first setting the machinery in both the backend/frontend (and having low risk as is LTI1.3 only) and in the follow up just adapt the existing code and focus more on testing the LTI1.1 changes.

lms/views/lti/deep_linking.py
```diff
@@ -49,20 +62,96 @@
...
-        form_action=request.params["content_item_return_url"],
+        form_action=request.parsed_params["content_item_return_url"],
         form_fields={
             "lti_message_type": "ContentItemSelection",
-            "lti_version": request.params["lti_version"],
+            "lti_version": request.parsed_params["lti_version"],
         },
     )
 
+    if application_instance.lti_version == "1.3.0":
```
marcospri, line 83: This conditional will disappear in the next PR where we enable the same flow for v11.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -41,3 +61,64 @@ def context(self):
...
+            }
+        )
+        assert fields["JWT"] == ltia_http_service.sign.return_value
+
+    @pytest.fixture
+    def pyramid_request(self, pyramid_request):
+        pyramid_request.parsed_params = {
+            "deep_linking_settings": sentinel.deep_linking_settings
+        }
+        return pyramid_request
```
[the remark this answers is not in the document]
jon-betts: It just occurred to me it would be really sweet if matchers captured the last thing they matched:

```python
expected_url = Any.url()
assert {"some_big_thing" : expected_url}
assert expected_url.last_match == ...
```

That way you could make kind of vague statements about the whole, then make more specific ones about parts you snip out after.

lms/views/lti/deep_linking.py
```diff
@@ -1,43 +1,56 @@
...
+
+  https://www.imsglobal.org/specs/lticiv1p0/specification-3
+
+- LTI 1.3 https://www.imsglobal.org/spec/lti-dl/v2p0
 
-https://www.imsglobal.org/specs/lticiv1p0/specification-3
 
 Canvas LMS's Content Item docs are also useful:
 
 https://canvas.instructure.com/doc/api/file.content_item.html
```
jon-betts, line 43: I think this is more relevant for the reader than before. Formatting here is a bit inconsistent

lms/views/lti/deep_linking.py
```diff
@@ -1,43 +1,56 @@
 """
-Views for handling deep linking launches.
+Deeplinking related views.
```
jon-betts, line 2:
```suggestion
Deep Linking related views.
```

lms/static/scripts/frontend_apps/components/test/FilePickerApp-test.js
```diff
@@ -135,6 +142,80 @@ describe('FilePickerApp', () => {
     });
   });
 
+  context('when deeplinkingAPI is present', () => {
+    const deeplinkingAPIPath = '/lti/1.3/deeplinking/formfields';
+    const deeplinkingAPIData = { some: 'data' };
```
jon-betts, line 147:
```suggestion
  context('when deepLinkingAPI is present', () => {
    const deepLinkingAPIPath = '/lti/1.3/deep_linking/form_fields';
    const deepLinkingAPIData = { some: 'data' };
```
There's a lot of this in the JS side. I'm not going to point them all out as I'm sure grep is better at this than me!

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,25 +120,79 @@ export default function FilePickerApp({ onSubmit }) {
...
+              data,
+            })
+          );
+        }
+      }
+      // Set shouldSubmit to true early to show the spinner while fetching form fields
+      setShouldSubmit(true);
+
+      // When deeplinkingAPI is present we want to call the backend to return the form
+      // fields we'll forward to the LMS to complete the DeepLinking request
```
jon-betts, line 153:
```suggestion
      // fields we'll forward to the LMS to complete the Deep Linking request
```

lms/views/lti/deep_linking.py
```diff
@@ -49,20 +62,96 @@
...
+        the content as query parameters.
+        """
+        params = dict(extra_params or {})
+
+        if content["type"] == "file":
+            params["canvas_file"] = "true"
+            params["file_id"] = content["file"]["id"]
+        else:
+            assert content["type"] == "url", "Unknown content type"
+            params["url"] = content["url"]
```
jon-betts, line 155: I think the usage of assert here is hiding a coverage issue. We don't test the failure case here.

lms/routes.py
```diff
@@ -108,3 +108,4 @@ def includeme(config):
 
     config.add_route("lti.oidc", "/lti/1.3/oidc")
     config.add_route("lti.jwks", "/lti/1.3/jwks")
+    config.add_route("lti.deeplinking.formfields", "/lti/1.3/deeplinking/formfields")
```
jon-betts, line 111:
```suggestion
    config.add_route("lti.deep_linking.form_fields", "/lti/1.3/deep_linking/form_fields")
```
I don't think "formfields" is a thing either?

jon-betts (changes requested):
I think the tests here could do with another look over. I've given an example of one of the views with a lot of lines commented out which the tests still pass on. There's a more subtle coverage miss in the form fields stuff as well with an assert which I don't think we test.

I've put an example of how you might change the second set of tests, but that's just an example. Pick and choose if you see anything nice.

I'm not sure where we are with the renaming change, but there are a lot of `deeplinking` and `formfields` still about which I think should be "deep_linking" or "deepLinking" depending on where they are etc. I'm assuming the whole changes haven't been made yet?

lms/views/lti/deep_linking.py
marcospri: Raising in that case now :+1:

lms/views/lti/deep_linking.py
marcospri: Moved things around a little bit, I think it looks a bit better now.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -1,17 +1,18 @@
...
 import pytest
 
 from lms.resources import LTILaunchResource
 from lms.resources._js_config import JSConfig
-from lms.views.lti.deep_linking import deep_linking_launch
+from lms.views.lti.deep_linking import DeepLinkingFieldsViews, deep_linking_launch
 
 pytestmark = pytest.mark.usefixtures(
     "application_instance_service", "course_service", "lti_h_service"
 )
```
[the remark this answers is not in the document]
marcospri: Yes :+1: . "course_service" can also be removed.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -22,6 +23,7 @@ def test_it_enables_content_item_selection_mode(self, context, pyramid_request):
                 "lti_version": "TEST_LTI_VERSION",
             },
         )
+        context.js_config.add_deeplinking_api.assert_called_once()
 
     def test_it_records_the_course_in_the_DB(self, context, pyramid_request):
         deep_linking_launch(context, pyramid_request)
```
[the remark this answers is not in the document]
marcospri: Moved everything to the same test + the one for v13.

On the v13 one I just assert the extra call to js_config. On the next PR that test can be removed.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -1,36 +1,54 @@
...
-pytestmark = pytest.mark.usefixtures(
-    "application_instance_service", "course_service", "lti_h_service"
-)
 
+@pytest.mark.usefixtures("application_instance_service", "lti_h_service")
+class TestDeepLinkingLaunch:
+    def test_it(
+        self, context, pyramid_request, lti_h_service, application_instance_service
+    ):
+        application_instance_service.get_current.return_value = create_autospec(
```
marcospri, line 18: The application_instance_service is wired up to return a real `application_instance` on `get_current`. 

This makes sense in a lot of tests but needed this here to be able to make assertions on how we could the result.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -1,36 +1,54 @@
...
+    def test_it_enables_content_item_selection_mode_lti_v13(
+        self, context, pyramid_request, application_instance
+    ):
+        application_instance.lti_registration_id = 100
 
-    def test_it_records_the_course_in_the_DB(self, context, pyramid_request):
         deep_linking_launch(context, pyramid_request)
 
-        context.get_or_create_course.assert_called_once_with()
+        context.js_config.add_deep_linking_api.assert_called_once()
```
marcospri, line 47: Just asserting what's different from the v11 call.

tests/unit/lms/views/lti/deep_linking_test.py
```diff
@@ -41,3 +59,74 @@ def context(self):
...
+            Any.dict.containing(
+                {
+                    "https://purl.imsglobal.org/spec/lti-dl/claim/content_items": [
+                        Any.dict.containing({"url": Any.url.with_query(output_params)})
+                    ]
+                }
+            )
+        )
+
+    def test_it_with_uknown_file_type(self, pyramid_request):
```
jon-betts, line 119:
```suggestion
    def test_it_with_unknown_file_type(self, pyramid_request):
```

jon-betts (approved):
Yep this is looking really good. Sorry about going on about the names, I don't think it's exactly crucial to fix, but it soothes my soul to try and get it right.

There's a small typo in the new test.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
marcospri: I added the reset for shouldSubmit and content. Without setting content to null the modal was left in an in-between state.

Some video demo:

https://user-images.githubusercontent.com/1433832/168266338-2fa36c8f-bcec-46ad-b97e-8fe8ad3b9166.mp4

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -111,25 +120,81 @@ export default function FilePickerApp({ onSubmit }) {
...
-  const submit = useCallback(() => setShouldSubmit(true), []);
+
+  const [deepLinkingFields, setDeepLinkingFields] = useState(
+    /** @type {Record<string,string>|null} */ (null)
+  );
+
+  const submit = useCallback(
+    /** @param {Content} content */
+    async content => {
+      async function getDeepLinkingFields() {
```
robertknight, line 131: This callback can potentially streamlined a bit:

- The `content` argument is non-nullable, so you don't need to check if is truthy
- If the content of `getDeepLinkingFields` is inlined where it is used, that avoids needing two checks for whether `deepLinkingAPI` is set, since the surrounding `if` block has a check
- Since the `if (deepLinkingAPI ...)` is at the end of the function with no code after it, you can invert the logic and make it an early bail-out, reducing the number of levels of indentation and avoiding the `return` in the `catch` clause

A sketch:

```js
async content => {
  // Set shouldSubmit to true early to show the spinner while fetching form fields
  setShouldSubmit(true);
                                                                                   
  if (!deepLinkingAPI || deepLinkingFields) {
    return;
  }
                                                                                   
  // When deepLinkingAPI is present we want to call the backend to return the form
  // fields we'll forward to the LMS to complete the Deep Linking request
  try {
    const data = {
      ...deepLinkingAPI.data,
      content,
      extra_params: {
        groupSet: groupConfig.useGroupSet ? groupConfig.groupSet : null,
      },
    };
    setDeepLinkingFields(
      await apiCall({
        authToken: authToken,
        path: deepLinkingAPI.path,
        data,
      })
    );
  } catch (error) {
    setErrorInfo({
      message: 'Unable to configure assignment',
      error: error,
    });
    setShouldSubmit(false);
    setContent(null);
  }
},
```

robertknight (approved):
I added some suggestions to streamline the `submit` callback in `FilePickerApp.js`. Otherwise this looks good 👍

Issue #3906, opened 2022-05-03
Store a JSTOR id against the application instance and send it to Via


jon-betts:
For: 

 * https://github.com/hypothesis/private-issues/issues/110

What we need to do:

 * Nice idea to find out what JSTOR call this id, so we don't invent our own terminology for it which doesn't match our partners
    * It's called a "site code"
 * Add a JSTOR id setting to the application instance (no migration required!)
 * Make it editable in the UI
 * Pass that value along to Via in the settings when it's present

## Tickets

 * https://github.com/hypothesis/devdata/pull/49
 * https://github.com/hypothesis/lms/pull/3918
 * https://github.com/hypothesis/lms/pull/3912

jon-betts:
Done!

PR #3918, opened 2022-05-04
Update the admin page to allow text field input + some refactors

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/3906

The names have changed a bit to reflect a read only field and a settings text field which you can change.

### Testing notes

 * Start _everything_
 * Go to http://localhost:8001/admin/
 * Enter `Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a`
 * Enable JSTOR and add a value to the site code and save
 * The values should be saved
 * Remove and clear the values and save
 * They should be removed
 * Add them back again

### DIFF
--- a/lms/views/admin.py
+++ b/lms/views/admin.py
@@ -81,17 +81,24 @@ class AdminViews:
     def update_instance(self):
         ai = self._get_ai_or_404(self.request.matchdict["consumer_key"])
 
-        for setting, sub_setting in (
-            ("canvas", "sections_enabled"),
-            ("canvas", "groups_enabled"),
-            ("blackboard", "files_enabled"),
-            ("blackboard", "groups_enabled"),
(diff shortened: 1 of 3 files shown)


lms/views/admin.py
```diff
@@ -81,17 +81,24 @@ def show_instance(self):
...
+            ("jstor", "site_code", str),
         ):
-            enabled = self.request.params.get(f"{setting}.{sub_setting}") == "on"
-            ai.settings.set(setting, sub_setting, enabled)
+            value = self.request.params.get(f"{setting}.{sub_setting}")
+            if setting_type == bool:
+                value = value == "on"
+            else:
+                assert setting_type == str
+                value = value if value else None
```
marcospri, line 99: `value.strip()`? 

It's not like it's an open form. We can live without that.

PR #3920, opened 2022-05-05
Replace uses of `any` type with `void` or `unknown`

robertknight:
Replace various uses of the `any` type with the either `void`, for callbacks whose result is unused or `unknown` for values where the type should be checked or asserted before use. No changes to the calling code were required.


lyzadanger (approved):
Oooh, yes, get this landed plz so I can merge my changes (RPC `server` and `client` modules) on this...

PR #3929, opened 2022-05-06
[WIP] Split LTIResources in classes per LMS

marcospri:


lms/resources/lti_launch/factory.py
```diff
@@ -0,0 +1,37 @@
+from lms.resources.lti_launch.base import LTILaunchResource
+from lms.resources.lti_launch.canvas import CanvasLTILaunchResource
+from lms.resources.lti_launch.blackboard import BlackboardLTILaunchResource
+
+from lms.models import ApplicationInstance
+
+
+def _is_canvas(request):
```
marcospri, line 8: We'll eventually might have to use this in multiple places.

I think the less ugly way of doing this having 

`def ApplicationInstanceService.product(self, request)`

Sometimes we can decide (or we thinkg we can) solely with the model's product but the request might be needed like in canvas.

lms/resources/lti_launch/factory.py
```diff
@@ -0,0 +1,37 @@
...
+    """Return True if Canvas is the LMS that launched us."""
+    if (
+        request.find_service(name="application_instance").get_current().product
+        == ApplicationInstance.Product.CANVAS
+    ):
+        return True
+
+    # TODO params only works for LTI1.1, but lti_params is part of c
+    # context and that's what we are building here.
+    if "custom_canvas_course_id" in request.params:
```
marcospri, line 18: Might need to move `lti_params` generation here and inject it into the resource.

lms/resources/lti_launch/base.py
```diff
@@ -137,17 +115,7 @@ def ext_lti_assignment_id(self):
...
-        if (
-            self._request.parsed_params.get("tool_consumer_info_product_family_code")
-            == "canvas"
-        ):
-            return True
-
-        if "custom_canvas_course_id" in self._request.parsed_params:
-            return True
-
+    def is_canvas(self) -> bool:
```
marcospri, line 118: Again, if we continue with this pattern I don't see a reason to do `is_canvas` very often if we use a common interface.

jon-betts (commented):
I think there are a few big themes that jump out at me from having a look through this:

### Keeping product specific code together and separate

I think both sides of this are equally important:

 * Gathering together product specific code in one place
 * Getting it out of the standard path

I think there's quite a bit missing from the second one here in this PR. There are a lot of speedgrader or other canvas specific concepts right at the bottom of the stack here. I don't know if that's a case of this being a first step, or whether we might eventually run out of road with this approach. I think we'd want to prove we can move to something with these wholly separate, as a solution which is mostly separate (apart from these X things) might stand in the way of a different solution which does allow complete separation.

We don't have to do it all at once, but I think I'd like to be see it could be done.

### Using parameterisation and context instead of repetition

I think this is a big theme in LMS code in general, but we have a lot of `is_canvas` `is_moodle` everywhere which I think points to a couple of different issues

 * We should be using parameterisation `product == Product.MOODLE` instead of `_is_moodle`
 * Ideally we shouldn't really be checking as much, because the context should ideally make it irrelevant - The Moodle context doesn't need to check if it's moodle to do moodle things, and therefore the more conditional code we have that relies on this, the more likely we've failed in the previous theme of collecting together product code.

I think there might be some missing collective concepts around concepts like `sections` and `groups` here. To give them their full names I think they are probably `canvas_sections` and `blackboard_groups` etc. I've been out of the loop too long to know the intricacies, but I think we should imagine a future where there are `moodle_conglomerates`, `blackbaud_huddles`, `eduflow_covens` etc. and see if we can group them into like behaviors in our product and use parameterisation. So we have `is_grouping_enabled(grouping_type)` rather than a large proliferation of `is_...` methods.

## How could we proceed?

I think this PR does a good job of showing we don't have a central mechanism that returns the product we are in, and it's kind of crazy when you think about it. I think this is a good place to start. I'd recommend a small wedge that might get us on this path might be to:

* Create a `get_product()` which returns a single `Enum` value
    * If this is the one in `ApplicationInstance` it should probably pulled out somewhere more prominent
 * Find the right place to put this that would make it available ideally as widely as possible (the context could be the right place, if everything that want's it can see it?)
 * Add the product type to the context
 * Look to eliminate places where we have `is_canvas` etc. and replace them with checks and remove those functions

### Why start here?

 * My gut feel is this gets at fixing the crux of the problem (what product are we inside?) and is well worth concentrating on all by itself to find the right place to make this accessible
 * This also starts dropping easy to find markers throughout the code of product specific code paths which we should snip out
 * I think this is the system that would power returning different context or services depending on the product type that underpins these things

I think this would lead quite naturally to a next PR where we enable the switching method and return a context class which just has the product type hard coded to prove the mechanism. Then after that we could start picking concepts one by one to migrate out.

For an incremental approach like this, I think you might want to first prove that we can get to the end though, even if this meant a quite horrible huge POC to flush out the problems.

## A random rant about resources / contexts

I'm sure there's some hugely significant difference between a resource and a context, and if you look through the Pyramid docs they are called "resources" however...

 * `resource` is a horribly vague word that can mean just about anything
 * `context` gives you a little more
 * It might be called `resource` in the docs but...
 * ... it shows up as `context` absolutely everywhere in the code

I don't know why they didn't pick one side of the fence, but I'm tempted to say we should. To muddy the waters even further we refer to them as `traversals` in `h`.

So my 2c would be that it's better to pick a name which immediately gets you in the right headspace ("this is where the variable context comes from") rather than another more technical and potentially accurate term which leads to confusion.

PR #3931, opened 2022-05-06
Rename `content_item_selection` module to `deep_linking`

marcospri:
In preparation for LTI1.3 / Deeplinking 2.0 using the same namespace.

We use deeplinking and content-item-selection as equivalent terms and deeplinking seems to be the preferred name in the spec in LTI1.0 and the only one in LTI1.3

https://www.imsglobal.org/specs/lticiv1p0
https://canvas.instructure.com/doc/api/file.content_item.html

# Testing 

Just as a sanity check try to configure/edit an assignment in Canvas.

### DIFF
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -174,8 +174,8 @@ class URLConfiguredBasicLTILaunchSchema(BasicLTILaunchSchema):
         return _data
 
 
-class ContentItemSelectionLTILaunchSchema(_CommonLTILaunchSchema):
-    """Schema for content item selection LTI launches."""
+class DeepLinkingLTILaunchSchema(_CommonLTILaunchSchema):
+    """Schema for deep linking LTI launches."""
 
     lti_message_type = fields.Str(
         validate=OneOf(["ContentItemSelectionRequest"]), required=True
--- a/lms/validation/__init__.py
+++ b/lms/validation/__init__.py
@@ -59,7 +59,7 @@ from lms.validation._exceptions import LTIToolRedirect, ValidationError
 from lms.validation._lti_launch_params import (
     BasicLTILaunchSchema,
     ConfigureAssignmentSchema,
-    ContentItemSelectionLTILaunchSchema,
+    DeepLinkingLTILaunchSchema,
     LTIV11CoreSchema,
     URLConfiguredBasicLTILaunchSchema,
 )
--- a/lms/views/basic_lti_launch.py
+++ b/lms/views/basic_lti_launch.py
@@ -3,7 +3,7 @@ Views for handling what the LTI spec calls "Basic LTI Launches".
 
 A Basic LTI Launch is the form submission POST request that an LMS sends us
 when it wants our app to launch an assignment, as opposed to other kinds of
-LTI launches such as the Content Item Selection launches that some LMS's
+LTI launches such as the deep linking launches that some LMS's
 send us while *creating* a new assignment.
 
 The spec requires Basic LTI Launch requests to have an ``lti_message_type``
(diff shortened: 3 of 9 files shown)


lms/views/lti/deeplinking.py
```diff
@@ -1,5 +1,5 @@
 """
-Views for handling content item selection launches.
+Views for handling content item selection / deep linking launches.
```
marcospri, line 2: I think both terms are equivalent, maybe content-item was the first name but it looks that deeplinking is the prefered name since 1.0:

https://www.imsglobal.org/specs/lticiv1p0

```
Learning Tools Interoperability® Deep Linking
(formerly named Content-Item Message)
```

jon-betts:
```suggestion
Views for handling deep linking launches.
```

Maybe just go the whole way and pick a side?

jon-betts: Also that link would be a killer addition to the docstring here.

jon-betts (changes requested):
My gut feeling is that if we're going to change the terminology (which I think we should) then we should make an effort to try and find this everywhere and make it consistent. Grepping for "content item" or "content_item" still gets lots of hits.

To the uninformed (i.e. me) it's not clear that `enable_content_item_selection_mode()` has anything to do with "deeplinking" for example.

lms/views/lti/deeplinking.py
```diff
@@ -1,19 +1,19 @@
...
+Views for handling deep linking launches.
 
-A content item selection request is an LTI launch request (so it's a form
+A deeplinking request is an LTI launch request (so it's a form
 submission containing all the usual launch params, including the OAuth 1
 signature) that's used by LMS's during the assignment creation process.
 
 When an instructor creates a Hypothesis assignment in an LMS that supports
-content item selection the LMS launches our app in order for us to present an
+deeplinking the LMS launches our app in order for us to present an
```
marcospri, line 9: content item is the right wording in the next line. We send "content items" using the deeplinking API.

lms/validation/_lti_launch_params.py
```diff
@@ -174,8 +174,8 @@ def _decode_url(self, _data, **_kwargs):  # pylint:disable=no-self-use
         return _data
 
 
-class ContentItemSelectionLTILaunchSchema(_CommonLTILaunchSchema):
-    """Schema for content item selection LTI launches."""
+class DeeplinkingLTILaunchSchema(_CommonLTILaunchSchema):
```
marcospri, line 177: I used the same convetion in the other PR, `Deeplinking`/`deeplinking_` instead of `DeepLinking`/`deep_linking_`. 

I reckon the convetion is to not split "concepts" even if they are two words. We do the same for for example "blackboard_` instead of `black_board_`.

jon-betts: I don't think it's anything that fancy. It's just about replacing spaces with underscores or upper case letters. 

"Blackboard" is a product name, so that's why it's all one word. If we were talking about an actual "black board" we would do it the other way, or if their product was called "Bla Ck Boar D" we might do something else.

In the spec https://www.imsglobal.org/specs/lticiv1p0-intro it's always referred to as "deep linking" or "Deep Linking". "deeplinking" doesn't appear in that document and would be something we invented. So I think the correct thing is "deep_linking" or "DeepLinking".

jon-betts (commented):
I don't think "deeplinking" is a word and it doesn't appear in the LTI docs. I think the correct phrase is "deep linking", which would effect the names chosen in this PR.

I don't think it's crucially important, but the question was raised so I've added comment.

I think camel casing and snake casing is a good example of an area that benefits from a simple rule that does not require subjective decisions about readability or what does or doesn't form a concept. This should be something you can do independently without effort or thought and come to the right conclusion. Like other linting things, it's not important enough to warrant discussion, but does benefit from some uniformity.

I think that rule is / should be (in the case of snake case):

 * Write out the phrase as you would in English
 * Replace spaces with underscores

I'm sure there are edge cases when companies do weird things, or if this would result in something totally unreadable.

I think talking about choosing the right words is actually a worthy thing to do when picking classes and variables, and I appreciate this feedback in my PRs, but how to render those names into our format should be automatic.

jon-betts (approved):
Yep, this looks good to me!

PR #3939, opened 2022-05-10
Rename `content_item_selection` mode to `file_picker`

marcospri:
Naming was confusing as it was mixing to different concepts:

- Our app displaying the filepicker
- Enabling or not the use of LTI's content_item_selection / deeplinking
API to communicate the selected document to the LMS.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -15,7 +15,7 @@ class JSConfig:
     class Mode(str, Enum):
         OAUTH2_REDIRECT_ERROR = "oauth2-redirect-error"
         BASIC_LTI_LAUNCH = "basic-lti-launch"
-        CONTENT_ITEM_SELECTION = "content-item-selection"
+        FILE_PICKER = "content-item-selection"
         ERROR_DIALOG = "error-dialog"
 
     class ErrorCode(str, Enum):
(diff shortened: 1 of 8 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -15,7 +15,7 @@ class JSConfig:
     class Mode(str, Enum):
         OAUTH2_REDIRECT_ERROR = "oauth2-redirect-error"
         BASIC_LTI_LAUNCH = "basic-lti-launch"
-        CONTENT_ITEM_SELECTION = "content-item-selection"
+        FILE_PICKER = "content-item-selection"
```
marcospri, line 18: I'll change the FE side of this in other PR, there it seems this is the only wrinkle and the "FilePicker" naming is the preferred one.

I rather do this after tackling LTI1.1 which involves a lot (for my FE ability) FE changes, we do use a enum after all.

tests/unit/lms/views/content_item_selection_test.py
```diff
@@ -15,7 +15,7 @@ class TestContentItemSelection:
     def test_it_enables_content_item_selection_mode(self, context, pyramid_request):
         content_item_selection(context, pyramid_request)
 
-        context.js_config.enable_content_item_selection_mode.assert_called_once_with(
+        context.js_config.enable_file_picker_mode.assert_called_once_with(
             form_action="TEST_CONTENT_ITEM_RETURN_URL",
```
jon-betts, line 19:
```suggestion
            form_action="TEST_FILE_PICKER_RETURN_URL",
```
?

jon-betts (approved):
Yeah, the legacy value of the enum is unfortunate, but we can tackle that later. I've not grepped through the code for instances of this, but the swaps here seem pretty uncontroversial.

PR #3941, opened 2022-05-10
Handle all JSTOR work on the LMS side

jon-betts:
For: 

 * https://github.com/hypothesis/private-issues/issues/110

Requires:

 * https://github.com/hypothesis/h-vialib/pull/35
 * An update in Via to add the above

This completely removes the need to have any JSTOR specific code in Via.

## Testing notes

 * Get the devdata as of (https://github.com/hypothesis/devdata/pull/53)
 * Start all the services
 * Open the dev tools with caching disabled
 * Visit https://hypothesis.instructure.com/courses/125/assignments/2968
 * You should see a request go to Via with a normal URL in it
 * You should see the JSTOR content banner in Via

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -1,9 +1,8 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
 
-from jose import jwt
-
-from lms.services import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError
 from lms.services.http import HTTPService
+from lms.services.jwt import JWTService
 from lms.views.helpers import via_url
 
 
@@ -63,7 +62,12 @@ class JSTORService:
         if "/" not in doi:
             doi = f"{self.DEFAULT_DOI_PREFIX}/{doi}"
 
-        token = self._get_access_token(self._site_code)
+        token = JWTService.encode_with_secret(
+            {"site_code": self._site_code},
+            secret=self._secret,
+            lifetime=timedelta(hours=1),
+        )
+
         s3_url = self._http.request(
             method="GET",
             url=f"{self._api_url}/pdf-url/{doi}",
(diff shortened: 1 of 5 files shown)


lms/services/jstor.py
```diff
@@ -28,17 +47,58 @@ def via_url(self, request, document_url):
...
+
+        if not s3_url.startswith("https://"):
+            raise ExternalRequestError(
+                f"Expected to get an S3 URL but got: '{s3_url}' instead"
+            )
+
+        return s3_url
+
+    def _get_access_token(self, site_code):
+        return jwt.encode(
```
marcospri, line 81: This could be done using our `JWTService.encode_with_secret`
```python
encode_with_secret({"site_code": site_code},  self._secret, timedelta(hours=1))


Is not abstracting much but it will make the transaction to jwt-jose easier (just switching it on the service).
```

jon-betts: Yep. This wasn't around in Via, I'll fix that up

lms/services/jstor.py
```diff
@@ -29,16 +47,56 @@ def via_url(self, request, document_url):
 
         return via_url(
             request,
-            document_url,
+            document_url=self._get_public_url(document_url),
             content_type="pdf",
-            options={"via.jstor.site_code": self._site_code},
+            # Show content partner banner in client for JSTOR.
+            options={"via.client.contentPartner": "jstor"},
```
jon-betts, line 53: To me this makes a whole lot more sense than the code we had in Via catching this all over the place

lms/services/jstor.py
```diff
@@ -1,22 +1,40 @@
...
         self._site_code = site_code
 
+        self._http = http_service
+
     @property
     def enabled(self) -> bool:
         """Get whether this instance is configured for JSTOR."""
 
-        return bool(self._enabled and self._site_code)
+        return bool(self._enabled and self._api_url and self._site_code)
```
marcospri, line 37: As this type of service/sources proliferate we might want to also make these checks on the admin pages, ie "you enabled this but it's missing the site code and the env var is not set."

marcospri (approved):
I tested this with and without the changes in Via. All looks good.

This is marked as a WIP, not sure if there's anything left to add/change.

jon-betts:
This won't show the banner yet until the change in `h-vialib` and Via to accept the parameter. I'd also like to merge that change into this PR.

PR #3942, opened 2022-05-10
PyLint: fail on useless-suppression

seanh:
I'm not sure exactly which messages do and don't cause PyLint to fail (non-zero exit status) by default but apparently `useless-suppression` doesn't (maybe no "informational" category messages do).

Fix our pylintrc files so that any messages from PyLint will cause it to fail.

Unfortunately the only way I can find to do this is to list all six of PyLint's message categories individually, and to list them using cryptic single-letter identifiers. I can't find any way to say `fail-on=*` or `fail-on=all` (`all` does work with some PyLint settings but doesn't seem to work with `fail-on`).

New versions of PyLint frequently add new messages but hopefully they rarely add new message _categories_ so this shouldn't need to be updated often.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -12,6 +12,11 @@ load-plugins=pylint.extensions.bad_builtin,
              pylint.extensions.overlapping_exceptions,
              pylint.extensions.redefined_variable_type,
 
+# Fail if there are *any* messages from PyLint.
+# The letters refer to PyLint's message categories, see
+# https://pylint.pycqa.org/en/latest/messages/messages_introduction.html
+fail-on=C,E,F,I,R,W
+
 [MESSAGES CONTROL]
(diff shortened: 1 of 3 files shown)


jon-betts (approved):
Yep. Looks good.

I would watch out for some of these though, as they are often the source of linting Heisenbugs. I've seen a few of these that seem to fail intermittently if added _or_ removed.

seanh:
> I would watch out for some of these though, as they are often the source of linting Heisenbugs. I've seen a few of these that seem to fail intermittently if added or removed.

Yes I've seen that kind of behaviour from PyLint as well. This doesn't add any new messages though, it just makes PyLint fail on messages that it was already printing but not failing on, so perhaps it won't trigger any such changes.

PR #3943, opened 2022-05-10
Remove "refresh-and-retry" logic from the backend

seanh:
We now implement OAuth2 refresh-and-retry in the frontend:

https://github.com/hypothesis/lms/pull/3782
https://github.com/hypothesis/lms/pull/3806

And this seems to have been working well:

https://hypothes-is.slack.com/archives/C2BLQDKHA/p1652180496018999

Remove the backend refresh-and-retry logic and tests: it's no longer used.


lms/services/blackboard_api/_basic.py
```diff
@@ -80,20 +78,7 @@ def refresh_access_token(self):
...
-            if self.refresh_enabled:
-                self._oauth_http_service.refresh_access_token(
-                    self.token_url,
-                    self.redirect_uri,
-                    auth=(self.client_id, self.client_secret),
-                )
-                return self._send(method, url)
-
-            raise
+        return self._send(method, self._api_url(path))
```
jon-betts, line 81: I don't think `_send` is call anywhere else now. It can probably be inlined now.

lms/services/canvas_api/_authenticated.py
```diff
@@ -58,23 +54,10 @@ def send(
             access token for the user is missing, expired, or has been deleted
         """
         call_args = (method, path, schema, timeout, params)
```
jon-betts, line 56: Packing the `call_args` here was useful before, but I think it's a bit weird on it's own now. This can be folded into the `send()` call now I think

tests/unit/lms/services/blackboard_api/_basic_test.py
```diff
@@ -178,5 +140,4 @@ def basic_client(http_service, oauth_http_service):
         redirect_uri=sentinel.redirect_uri,
         http_service=http_service,
         oauth_http_service=oauth_http_service,
-        refresh_enabled=False,
     )
```
jon-betts, line 143: Now that the tests are significantly shorter I wonder if we can move to the more standard `Test<class>.test_<method>` layout rather than `Test<method>.test_it` layout like the Canvas API tests?

I suspect it would make the tests a bit shorter and easier to look through. At the moment there are a lot of tests like `test_it_raises_ExternalRequestError_if_the_request_fails` which require you to scroll up to the class name to work out what "it" is in this context.

jon-betts (approved):
Yeah this looks good, but I think the code removal has left some artefacts that made more sense before. There are a few places where some code was required before but can now be inlined. 

The Blackboard tests are in an older style which broke out test classes per method and might be clearer if they matched the class for class style, like we see in the Canvas ones.

To test this I logged into a few things in Canvas and Blackboard and then wiped the `oauth2_token` table. I could see refreshing happening as expected in Canvas, but I couldn't get it to trigger in Blackboard. Am I missing something?

lms/services/blackboard_api/_basic.py
seanh: Done

lms/services/canvas_api/_authenticated.py
seanh: Done

tests/unit/lms/services/blackboard_api/_basic_test.py
seanh: Done

Issue #3690, opened 2022-02-15
How should registration process work for LTI1.3


marcospri:
- Should we keep having a open form?
- Should we move registration to the admin pages?

Independently of the solution we opt for, this is roughly the information that needs to be exchanged:

- We must provide the school with some configuration parameters. Unless we opt for the last option on https://github.com/hypothesis/lms/issues/3688 this will be the same values for every school (ie this could be a kb article per LMS)

If we use one JWK per school there would be an extra step of generating those and sending an unique URL to the school.

- The previous step will generate a `client_id`
- The tool then needs to be added to the account/course using that `client_id`. That will generate a `deployment_id`.

- The school needs to send hypothesis both `client_id` & `deployment_id` along with some auth related URL (JWK endpoint, oidc endpoint, issuer name). 

This could happen over email, using an open form, using a form behind a login, form using an unique token....

marcospri:
Some other tool's install instructions:

- https://networkmanagerguide.pressbooks.com/chapter/configuring-the-lti-1-3-plugin/#canvas
- https://help.vocareum.com/en/articles/5085436-lti-1-3-setup-guide-for-canvas
- https://zoomappdocs.docs.stoplight.io/lti-pro-v2/guides/canvas/LTI-Pro-1-3-canvas/lti-1.3

chrisshaw:
Here's another we generally look at, not that they're a good reference, just another one.


- https://support.piazza.com/support/solutions/articles/48001191921--lti-1-3-configure-piazza-with-canvas
  - https://support.piazza.com/support/solutions/articles/48001191922--lti-1-3-configure-piazza-with-moodle
  - https://support.piazza.com/support/solutions/articles/48001191926--lti-1-3-configure-piazza-with-d2l-brightspace

chrisshaw:
We decided to gate this. We will, however, likely need to provide a way for customers to provide their Canvas API keys to us somehow.

We could still just have support send the keys and then shrink the online form to just dev keys. Then again, we want to close down the 1.1 form anyway. So perhaps we develope a public facing solution, such as a one-time code or something.

marcospri:
A rough flow of how registrations could work could be:

- All the information the LMS needs from us can be in a KB article. No need to generate keys or anything.

- Once they follow the KB article instructions they need to send us:
    - client_id
    - deployment_id
    - key_set_url
    - auth_login_url 
   
[...]

PR #3945, opened 2022-05-11
Allow to search for application instances in the admin pages

marcospri:
https://user-images.githubusercontent.com/1433832/168567305-adc225e5-a9d1-46d6-9f5a-6085f4b23776.mp4

Main motivation here is to make the admin pages useful in development while working on LTI1.3. This PR doesn't try to tackle https://github.com/hypothesis/lms/issues/3690 (ie how to get new registration/application instances) into the DB.

In LTI1.3 install are identified by three fields: issuer, client_id and deployment ID. Typing all of them would be cumbersome so opted for a search form instead and merged the existing `consumer_key` query into that.

Added GUID to the search fields as it might be useful. 

## Testing

- Go to http://localhost:8001/admin/instances/
- Test the search functionality, some test data to make some queries
- Test editing an application instance works as expected.

```
Consumer key: Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b
Issuer: https://blackboard.com
Client ID: 4aa65f52-1d75-48d2-8a32-42b5e4ae995e
Deployment ID: 609:6ed5aff92b5257ab607ec9c629844ea8586fde68
GUID: VCSy8DAKCFIQYaItpSGyXhOXIlk5A6NWduVVG1u3:canvas-lms
```

### DIFF
--- a/lms/templates/admin/instances.results.html.jinja2
+++ b/lms/templates/admin/instances.results.html.jinja2
@@ -1,0 +1,32 @@
+{% extends "lms:templates/admin/base.html.jinja2" %}
+{% block header %}
+Application instances
+{% endblock %}
+
+{% block content %}
+
+<table class="table">
+  <thead>
+    <tr>
+      <th>ID</th>
+      <th>Consumer Key</th>
+      <th>GUID</th>
+      <th>Issuer</th>
+      <th>Deployment ID</th>
+      <th></th>
+    </tr>
+  </thead>
+  <tbody>
+{% for instance in instances %}
+    <tr>
+      <td>{{instance.id}}</td>
+      <td>{{instance.consumer_key}}</th>
+      <td>{{instance.tool_consumer_instance_guid}}</td>
+      <td>{{instance.lti_registration.issuer}}</td>
+      <td>{{instance.deployment_id}}</td>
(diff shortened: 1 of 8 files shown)


lms/templates/admin/instances.results.html.jinja2
```diff
@@ -0,0 +1,32 @@
+{% extends "lms:templates/admin/base.html.jinja2" %}
+{% block header %}
+Application instances
+{% endblock %}
+
+{% block content %}
+
+<table class="table">
```
marcospri, line 8: Little table view for multiple results

lms/views/admin.py
```diff
@@ -34,52 +35,83 @@ def index(self):
...
+            tool_consumer_instance_guid=self.request.params.get(
+                "tool_consumer_instance_guid"
             ),
         )
 
+        if not instances:
+            self.request.session.flash("No instances found", "errors")
+            return HTTPFound(location=self.request.route_url("admin.instances"))
+
+        if len(instances) == 1:
```
marcospri, line 82: I think I prefer this behavior, I'm searching for something that identifies only one AI I rather be taken directly to it.

Some fields do indeed always identify an AI (consumer_key) while other might do "often" (the LTI1.3 ones, deployment_id).

seanh (approved):
:+1: Works for me

PR #3954, opened 2022-05-16
Fixes for canvas files and canvas groups in LTI1.3

marcospri:
Main thing is using `lti_params` instead of params. I rather do this feature by feature instead of a good old search and replace as is easier to test individually. 

When we are "done" I think it will be easier to review the ones that still use `request.params` one by one.

There was also a couple of things getting on the way of these features working on LTI1.3:

- A bug on LTIParams. Fixed and added a test for that.
- A difference between versions in one parameter canvas sends. `str` vs `int` depending on the version.

## Testing

(you can test a broken version of this in `master` first)

- `make devdata`

- Edit an LTI1.3 assignment https://hypothesis.instructure.com/courses/319/assignments/2781/edit?return_to=https%3A%2F%2Fhypothesis.instructure.com%2Fcourses%2F319%2Fassignments%2F2781

- Select Canvas files and pick the existing file. Also make it a groups assignment picking the existing group set. Save it.

- Launching the assignment will pick the right canvas PDF and the group selector will show on the sidebar.

### DIFF
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -62,6 +62,14 @@ _V11_TO_V13 = (
         "deep_linking_settings",
         ["https://purl.imsglobal.org/spec/lti-dl/claim/deep_linking_settings"],
     ),
+    (
+        "custom_canvas_course_id",
+        [f"{CLAIM_PREFIX}/custom", "canvas_course_id"],
+    ),
+    (
+        "custom_canvas_api_domain",
+        [f"{CLAIM_PREFIX}/custom", "canvas_api_domain"],
+    ),
 )
 
 
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -85,6 +85,8 @@ def _to_lti_v11(v13_params):
             if path_item not in value:
                 break
             value = value[path_item]
+        else:
+            # We only found the key if we exhausted all of `v13_path`
             found = True
 
         # We don't want to add partial values along v13_path
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -66,6 +66,22 @@ class _CommonLTILaunchSchema(LTIV11CoreSchema):
     launch_presentation_return_url = fields.Str()
     tool_consumer_info_product_family_code = fields.Str()
 
+    @pre_load
+    def _load_canvas_course_id(self, data, **_kwargs):  # pylint: disable=no-self-use
+        canvas_course_id = data.get("custom_canvas_course_id")
+        if canvas_course_id and isinstance(canvas_course_id, int):
+            # In LTI1.3 course_id is sent as an integer
+            # instead of as a string in LTI1.1.
+            # Coercing both version to integers could be handled
+            # with fields.Int(strict=False)
+            #
+            # Doing a manual conversion to str here instead so both versions
+            # behave the same as LMS is concern
+            # and to allow for future changes on canvas' side
+            data["custom_canvas_course_id"] = str(canvas_course_id)
+
+        return data
+
     @validates_schema
     def validate_consumer_key(self, data, **_kwargs):  # pylint: disable=no-self-use
         if (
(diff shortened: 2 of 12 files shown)


tests/functional/lti_certification/v13/conftest.py
```diff
@@ -148,6 +148,6 @@ def common_payload():
         # Mirror the custom param we set in the LTI compliance testing tool
         # This gets around the fact it doesn't send a GUID by default
         "https://purl.imsglobal.org/spec/lti/claim/custom": {
-            "certification_guid": "GUID"
+            "certification_guid": "TEST_CONSUMER_INSTANCE_GUID"
```
marcospri, line 151: Now that we are doing `application_instance.update_lms_data(context.lti_params)` is was easy to detect the the value here didn't match the one stored in the DB in the fixture further up.

lms/models/lti_params.py
```diff
@@ -77,6 +85,8 @@ def _to_lti_v11(v13_params):
             if path_item not in value:
                 break
             value = value[path_item]
+        else:
```
marcospri, line 88: Added a test for this. 

We only want to translate a key if we found the whole path. The code was trying to do that but always taking the first match anyway.

lms/validation/_lti_launch_params.py
```diff
@@ -66,6 +66,22 @@ class _CommonLTILaunchSchema(LTIV11CoreSchema):
     launch_presentation_return_url = fields.Str()
     tool_consumer_info_product_family_code = fields.Str()
 
+    @pre_load
+    def _load_canvas_course_id(self, data, **_kwargs):  # pylint: disable=no-self-use
+        canvas_course_id = data.get("custom_canvas_course_id")
+        if canvas_course_id and not isinstance(canvas_course_id, str):
+            # In LTI1.3 course_id is sent as an integer
```
marcospri, line 73: A little annoying canvas detail that breaks the validation.

lms/validation/_lti_launch_params.py
```diff
@@ -66,6 +66,22 @@ class _CommonLTILaunchSchema(LTIV11CoreSchema):
...
+            # instead of as a string in LTI1.1.
+            # Coercing both version to integers could be handled
+            # with fields.Int(strict=False)
+            #
+            # Doing a manual conversion to str here instead so both versions
+            # behave the same as LMS is concern
+            # and to allow for future changes on canvas' side
```
marcospri, line 80: What I'm trying to say here at the end is that we rather treat the value as just some opaque value that we forward to the Canvas API.

A string is more opaque than an integer and if they change to let's say "uuids" we are covered.

jon-betts (approved):
This works (now I've tested it correctly).

The `for ... else` makes me a little queasy. I had to go look it up to ensure I understood it correctly. I've got a suggestion that avoids the need, or another one to make it a little simpler if we want to keep it.

I'm not sure if the logic around the string conversion there is correct or not. I feel like this might stringify anything, which kind of voids any checking of the value at all. If we got a dict, I think we might stringify it then validate it as a good string?

Not introduced here, but looking at the schema tests they've ended up in a weird spot. I think we might have split off a base-class but not moved the tests to match. So we have the base class being tested through a specific sub-class, but it's not clear why that should be the one where the tests go vs. any other test. I think it would be good to have a simpler structure where the test class matches the class under test etc.

tests/unit/lms/validation/_lti_launch_params_test.py
```diff
@@ -85,6 +85,13 @@ def test_ValidationError_raised_when_res_link_missing_with_bad_return_url(
...
+    def test_allows_int_custom_canvas_course_id(self, pyramid_request):
+        pyramid_request.params["custom_canvas_course_id"] = 1
+
+        params = BasicLTILaunchSchema(pyramid_request).parse()
+
+        assert params["custom_canvas_course_id"] == str(1)
+
     @pytest.mark.parametrize(
         "required_param",
         ["oauth_consumer_key", "tool_consumer_instance_guid", "user_id"],
```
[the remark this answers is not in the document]
marcospri: I think `_CommonLTILaunchSchema` has been there "forever" but it didn't really have any code.

Moved that test to where it belongs.

PR #3957, opened 2022-05-16
LTI 1.1 `/config_xml` view internal rename

marcospri:
Moving the existing /config_xml view to the canvas namespace. Including the version in the route to accommodate the v13 in a follow up PR.

### Testing

Just open http://localhost:8001/config_xml

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -11,7 +11,7 @@ def includeme(config):
     config.add_route("logout", "/logout")
     config.add_route("reports", "/reports")
 
-    config.add_route("config_xml", "/config_xml")
+    config.add_route("canvas.v11.config", "/config_xml")
     config.add_route(
         "configure_assignment",
         "/assignment",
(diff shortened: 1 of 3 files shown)


lms/routes.py
```diff
@@ -114,3 +113,5 @@ def includeme(config):
     config.add_route(
         "lti.v11.deep_linking.form_fields", "/lti/1.1/deep_linking/form_fields"
     )
+
+    config.add_route("lti.v11.config", "/config_xml")
```
marcospri, line 117: Moving it down here with the rest of LTI related routes.

lms/views/lti/config.py
```diff
@@ -2,12 +2,12 @@
 
 
 @view_config(
-    route_name="config_xml",
+    route_name="lti.v11.config",
```
marcospri, line 5: Following the convention of other LTI names.

jon-betts (approved):
This only changes the internal name of the view and it's location. The external URL will remain the same. This is a bit of a shame, but I think might be unavoidable, as this URL has already been passed out to customers when they integrated us.

I'm not sure if it's ever read again, but it's certainly safe to leave it alone.

jon-betts (approved):
Yep this looks good.

PR #3961, opened 2022-05-17
Add LTI1.3 config endpoint for canvas

marcospri:
Add the same configuration endpoint concept for LTI1.3. 

In LTI1.3 the endpoint must return json instead of xml but the information is basically the same.

## Testing

Head to: http://localhost:8001/canvas/1.3/config

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -12,6 +12,7 @@ def includeme(config):
     config.add_route("reports", "/reports")
 
     config.add_route("canvas.v11.config", "/config_xml")
+    config.add_route("canvas.v13.config", "/canvas/1.3/config")
     config.add_route(
         "configure_assignment",
         "/assignment",
--- a/lms/views/canvas/config.py
+++ b/lms/views/canvas/config.py
@@ -1,3 +1,9 @@
+"""
+Configuration views for Canvas.
+
+These views allow an admin user to paste in a URL and have Hypothesis automatically
+configured in Canvas.
+"""
 from pyramid.view import view_config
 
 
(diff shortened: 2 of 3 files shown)


lms/views/lti/config.py
```diff
@@ -2,7 +2,7 @@
 
 
 @view_config(
-    route_name="lti.v11.config",
+    route_name="lti.v11.canvas.config",
```
marcospri, line 5: The v11 endpoint it's also canvas specific so using the same name conventions in both.

lms/views/lti/config.py
```diff
@@ -14,3 +14,63 @@ def config_xml(request):
         "launch_url": request.route_url("lti_launches"),
         "content_item_url": request.route_url("content_item_selection"),
     }
+
+
+@view_config(
+    route_name="lti.v13.canvas.config",
```
marcospri, line 20: This is canvas only for now. 

We could try to do a generic view and add to that the canvas specific confic. We don't know exactly what will be different in other LMSs so I rather wait until we need it and keep this is simple for now.

lms/views/lti/config.py
```diff
@@ -14,3 +14,63 @@ def config_xml(request):
...
+def config_json(request):
+    """
+    Return the LTI1.3 configuration for Canvas installs as a json document.
+
+    https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#anatomy-of-a-json-configuration
+    """
+    return {
+        "title": "Hypothesis",
+        "description": "Hypothesis",
+        "oidc_initiation_url": request.route_url("lti.oidc"),
```
marcospri, line 33: Mostly a 1 to 1  translation of the XML document. 

More details here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#anatomy-of-a-json-configuration

jon-betts (commented):
Most everything I've written in this ticket can be ignored I think in light of the fact...

I don't think this has anything to do with LTI (the interoperability standard) as this is a completely Canvas specific concept. I think they invented the concept, format and contents.

This is specifically a set of end-points for configuring a Canvas instance, rather than anything you even _could_ use in Moodle. Moodle might have their own equivalent, it might even have the same format (I don't think this is true), but it wouldn't be part of the LTI spec even if it was.

Long story short:

 * This can probably go in `views.api.canvas`
 * As such any suggestions I've made about URLs or splitting out Canvas specific portions don't really make any sense
 
The 1.1 edition of this URL probably can't change (or can it? I think we might just hand it out to Canvas admins to plug in). But the new one is probably something along the lines of `/api/canvas/config/1.3`.

lms/routes.py
```diff
@@ -12,6 +12,7 @@ def includeme(config):
     config.add_route("reports", "/reports")
 
     config.add_route("canvas.v11.config", "/config_xml")
+    config.add_route("canvas.v13.config", "/canvas/1.3/config")
```
jon-betts, line 15: If we followed the file layout then the two variations (1.1 and 1.3) are in the same file, which might suggest the hierarchy is `canvas/config/1.3`.

On the other hand I've seen conventions where the it's something like `<api_version>/<path>` before which is a bit more like this. Is that what you were going for?

lms/views/canvas/config.py
```diff
@@ -7,10 +7,74 @@
     request_method="GET",
 )
 def config_xml(request):
-    """Render the XML configuration."""
+    """
+    Render the XML configuration.
```
jon-betts, line 11:
```suggestion
    Generate XML configuration for the admin pages in Canvas.
    
    This allows an admin user to paste in a URL and have Hypothesis automatically
    configured in Canvas.
```

Maybe something like this? Also seeing as this extra detail applies to both, maybe this would good as a file level doc string?

jon-betts (approved):
Yep this totally works. 

I think working through this has helped us clarify that this is a Canvas only thing, and this seems the right home for it. I think it would be good to expand the docstrings to explain our new found understanding.

For example the current docstring for `config_xml` is `Render the XML configuration` which I think misses an opportunity to add more meaning.

PR #3964, opened 2022-05-18
CanvasPreRecordHook support for LTI1.3

marcospri:
Canvas grading extensions for LTI1.3. 

Very similar to the LTI1.1 just on the JSON body instead of the XML one.

## Testing

Instructions mostly copied from https://github.com/hypothesis/lms/pull/3956.

Note that `submit_on_annotation` is not really necessary for this but it makes the testing much simpler IMO.

---

- Ensure that the `submit_on_annotation` feature flag is enabled on http://localhost:8001/flags
or start the server with `export FEATURE_FLAG_SUBMIT_ON_ANNOTATION=true`
- Launch a grade-able assignment as a student

LTI1.1 https://hypothesis.instructure.com/courses/125/assignments/2956
LTI1.3 https://hypothesis.instructure.com/courses/319/assignments/2988

- Add an annotation

- Log in as an instructor (incognito window / different browser) and launch the same assignment; go into Speed Grader
- You should  see a recent submission from the student who created the annotation.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -302,8 +302,10 @@ class JSConfig:
         :raise HTTPBadRequest: if a request param needed to generate the config
             is missing
         """
-        lis_result_sourcedid = self._request.params.get("lis_result_sourcedid")
-        lis_outcome_service_url = self._request.params.get("lis_outcome_service_url")
+        lis_result_sourcedid = self._context.lti_params.get("lis_result_sourcedid")
+        lis_outcome_service_url = self._context.lti_params.get(
+            "lis_outcome_service_url"
+        )
 
         # Don't set the Canvas submission params in non-Canvas LMS's.
         if not self._context.is_canvas:
@@ -322,18 +324,23 @@ class JSConfig:
         if not lis_outcome_service_url:
             return
 
+        custom_canvas_user_id = self._context.lti_params["custom_canvas_user_id"]
+
         self._config["canvas"]["speedGrader"] = {
             "submissionParams": {
                 "h_username": self._h_user.username,
                 "lis_result_sourcedid": lis_result_sourcedid,
                 "lis_outcome_service_url": lis_outcome_service_url,
-                "learner_canvas_user_id": self._request.params["custom_canvas_user_id"],
+                # Canvas sends custom_canvas_user_id as in integer in LTI1.3 and as a string in LT1.1
+                "learner_canvas_user_id": str(custom_canvas_user_id)
+                if isinstance(custom_canvas_user_id, int)
+                else custom_canvas_user_id,
                 "group_set": self._request.params.get("group_set"),
                 # Canvas doesn't send the right value for this on speed grader launches
                 # sending instead the same value as for "context_id"
-                "resource_link_id": self._request.params.get("resource_link_id"),
+                "resource_link_id": self._context.lti_params.get("resource_link_id"),
                 # Canvas doesn't send this value at all on speed grader submissions
-                "ext_lti_assignment_id": self._request.params.get(
+                "ext_lti_assignment_id": self._context.lti_params.get(
                     "ext_lti_assignment_id"
                 ),
                 **kwargs,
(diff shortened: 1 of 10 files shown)


lms/views/api/lti.py
```diff
@@ -96,22 +96,23 @@ class CanvasPreRecordHook:
     def __init__(self, request):
         self.request = request
 
-    def __call__(self, score=None, request_body=None):
-        request_body["resultRecord"].setdefault("result", {})["resultData"] = {
-            "ltiLaunchUrl": self.get_speedgrader_launch_url()
-        }
+    def __call__(self, score=None, request_body=None) -> dict:
```
marcospri, line 99: This whole file seem to be about grading. I think it could be renamed. 

That been said I'd like to tackle that in a separate PR to avoid some rebase-hell.

lms/services/lti_grading/_v11.py
```diff
@@ -44,11 +44,6 @@ def record_result(self, grading_id, score=None, pre_record_hook=None):
         if pre_record_hook:
             request = pre_record_hook(score=score, request_body=request)
 
-            if not isinstance(request, dict):
```
marcospri, line 47: I don't see a reason to type check this and not any other function in our codebase.

tests/unit/lms/views/api/lti_test.py
```diff
@@ -65,47 +65,79 @@ class TestCanvasPreRecordHook:
...
+        self, parsed_params, url_params, pyramid_request
     ):
         pyramid_request.parsed_params.update(parsed_params)
-        pyramid_request.parsed_params["submitted_at"] = submitted_at
 
-        result = CanvasPreRecordHook(pyramid_request)(
-            score=None, request_body={"resultRecord": {}}
+        launch_url = CanvasPreRecordHook(pyramid_request).get_speedgrader_launch_url()
+
+        assert launch_url == Any.url.with_path("/lti_launches").with_query(
```
marcospri, line 86: Testing `get_speedgrader_launch_url` directly here with mulitple url params.

tests/unit/lms/views/api/lti_test.py
```diff
@@ -65,47 +65,79 @@ class TestCanvasPreRecordHook:
...
+            )
         )
 
+    @pytest.mark.parametrize(
+        "submitted_at", (datetime.datetime(2022, 2, 3, tzinfo=timezone.utc), None)
+    )
+    def test_it_v11(self, hook, submitted_at, get_speedgrader_launch_url_mock):
+        hook.request.parsed_params["submitted_at"] = submitted_at
+
+        result = hook(score=None, request_body={"resultRecord": {}})
```
marcospri, line 100: Now testing each LTI version with a mock of get_speedgrader_launch_url

lms/resources/_js_config/__init__.py
```diff
@@ -327,13 +330,16 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
                 "h_username": self._h_user.username,
                 "lis_result_sourcedid": lis_result_sourcedid,
                 "lis_outcome_service_url": lis_outcome_service_url,
-                "learner_canvas_user_id": self._request.params["custom_canvas_user_id"],
+                # Canvas sends custom_canvas_user_id as in integer in LTI1.3 and as a string in LT1.1
+                "learner_canvas_user_id": str(custom_canvas_user_id)
```
marcospri, line 334: I'm tempted to move this (and the same thing that we did in marshmallow in the previous PR) to LTIParams.

We can return the right value there the same way we tweak the value of `roles`.

https://github.com/hypothesis/lms/blob/master/lms/models/lti_params.py#L98


It might be better to have a couple more conditionals there, close to the params themselves than scattered around.

jon-betts (commented):
Just dropping a comment review so the in progress review can be read while I finish up

jon-betts (approved):
This is working now. I'd forgotten about the alternative way of setting the feature flags, and this works much better.

You had a couple of suggestions for changes which I agree with:

* `api/lti.py` -> `api/lti/grading.py`
* Moving stringification of the params where params are converted

I think I'd add putting a constant in the hooks class for the default date would be good too:

```python
class CanvasPreRecordHook:
    DEFAULT_SUBMISSION_DATE=datetime.datetime(2001, 1, 1, tzinfo=timezone.utc)
```

We repeat this in a few places, and a constant is a good way of adding meaning.

There are a few little suggestions about styling and naming and such.

PR #3974, opened 2022-05-19
Refactor CourseService to take only GUID and context ID.

marcospri:
Callers usually have those two pieces of information but had to generate
an authority_provided_id instead before calling get/upsert.


lms/views/api/blackboard/sync.py
```diff
@@ -124,9 +124,6 @@ def sync_to_h(self, groups):
         lti_h_svc.sync(groups, group_info)
 
     def get_course(self, course_id):
-        course_service = self.request.find_service(name="course")
-        return course_service.get(
-            course_service.generate_authority_provided_id(
-                self.tool_consumer_instance_guid, course_id
-            )
+        return self.request.find_service(name="course").get(
```
marcospri, line 127: Using the service has a better API with this change IMO.

lms/views/api/canvas/sync.py
```diff
@@ -17,7 +17,6 @@ def __init__(self, request):
         self._request = request
         self._grouping_service = self._request.find_service(name="grouping")
         self._canvas_api = self._request.find_service(name="canvas_api_client")
-        self._course_service = self._request.find_service(name="course")
```
marcospri, line 20: We can inline this now.

jon-betts (commented):
I've not gone through this in crazy detail as it's only a draft but... YES. This looks like a perfect example of exposing private guts of the service out to the callers. I love refactors like this.

I'm not sure if this is what were were talking about before but feeding output from a class back in to itself is a strong code smell.

tests/unit/lms/services/course_test.py
```diff
@@ -44,60 +23,121 @@ def test_it_inserts_False_if_theres_a_matching_row_in_course_groups_exported_fro
...
 
         assert svc.any_with_setting("group", "key", value) is expected
 
-    def test_upsert_returns_existing(self, svc, application_instance, db_session):
-        existing_course = factories.Course(application_instance=application_instance)
+    def test_get_by_context_id(self, svc, course):
+        assert svc.get_by_context_id(course.lms_id) == course
+
+    def test_get_by_context_id_with_no_match(self, svc):
+        assert svc.get_by_context_id("NO MATCH") is None
```
jon-betts, line 45: Added this case and moved it above upsert to match the service ordering

PR #3975, opened 2022-05-19
Remove code to support LTI-authenticated LMS launches

robertknight:
**Depends on https://github.com/hypothesis/lms/pull/3973**

VitalSource's book reader is currently loaded without any pass-through
of LMS credentials. We do expect to eventually have some kind of
automatic sign-in based on a user's LMS/LTI identity. However this will
very likely not involve a single LMS/LTI keypair for our lms app, but
instead we'll use some kind of per-institution credentials, and will
probably use VitalSource APIs instead of their LTI launch support.

Part of https://github.com/hypothesis/private-issues/issues/102


marcospri (approved):

robertknight:
https://github.com/hypothesis/lms/pull/3973 is landed to remove the feature flag. I'll land these changes next Monday (2020-05-23).

PR #3998, opened 2022-05-27
Use IDs from the lti1p1 claim when available in LTIParams

marcospri:


tests/unit/lms/models/lti_params_test.py
```diff
@@ -46,3 +46,20 @@ def test_it_doesnt_set_partial_keys(self):
...
+                },
+                "https://purl.imsglobal.org/spec/lti/claim/lti1p1": {
+                    "user_id": "user_id-v11",
+                    "resource_link_id": "resource_link_id-v11",
+                },
+            }
+        )
+
+        assert params["user_id"] == "user_id-v11"
+        assert params["resource_link_id"] == "resource_link_id-v11"
```
jon-betts, line 65: It's a little counter intuitive that the items which come later in the list are higher priority, but it's simple enough if you spot it (we overwrite). I don't think it's an issue if you test for it though as you'll get told off it you change it wrong.

How short could you get this test with parameterisation? If it's not too much different, then it would be nice to not mix the two.

jon-betts: I've tried this a couple of ways, and I'm not really sure. Some observations:

 * The way it gets spaced out in a `parameterize` is obnoxious
 * Importing and interpolating `CLAIM_PREFIX` helps quite a bit
 * If you only do one case at a time, you can get away with just `v11` / `v13` as the value
 * If there were more cases, it might make more sense to parameterize
 
The parameterized version looks like this:

```python
    @pytest.mark.parametrize(
        "key,params",
        (
            (
                "user",
                {
                    "sub": "v13",
                    f"{CLAIM_PREFIX}/lti1p1": {"user_id": "v11"},
                },
            ),
            (
                "resource_link_id",
                {
                    f"{CLAIM_PREFIX}/resource_link_id": {"id": "v13"},
                    f"{CLAIM_PREFIX}/lti1p1": {"resource_link_id": "v11"},
                },
            ),
        ),
    )
    def test_prefers_lti1p1_idas(self, key, params):
        params = LTIParams.from_v13(params)

        assert params[key] == "v11"
```

This might be a case where reptition is actuallty better:

```python
    def test_prefers_lti1p1_resource_link_id(self):
        params = LTIParams.from_v13(
            {
                f"{CLAIM_PREFIX}/resource_link_id": {"id": "v13"},
                f"{CLAIM_PREFIX}/lti1p1": {"resource_link_id": "v11"},
            }
        )

        assert params["resource_link_id"] == "v11"

  # ... etc.
```

jon-betts (approved):
I've tried this and it works. The code is good.

I think I am missing a bit of context about _why_ we would do this though? Is this to ensure that assignments created by a 1.1 application instance can be read from a 1.3 one? I'm not sure on the exact scenario this helps with.

PR #4007, opened 2022-05-30
Admin pages to create LTI1.3 installs

marcospri:
- Move common used macros to their own file
- Rename build_from_lms_url to create


lms/routes.py
```diff
@@ -1,4 +1,4 @@
-def includeme(config):
+def includeme(config):  # pylint:disable=too-many-statements
```
marcospri, line 1: We could make `lms/routes/admin.py`. 

I rather split it in a follow up PR.

lms/services/application_instance.py
```diff
@@ -116,29 +116,44 @@ def search(
 
         return query.all()
 
-    def build_from_lms_url(  # pylint:disable=too-many-arguments
-        self, lms_url, email, developer_key, developer_secret, settings
+    def create(  # pylint:disable=too-many-arguments
```
marcospri, line 119: `lms_url` doesn't seem like the main thing here. Using just `create`.

lms/templates/admin/instances.html.jinja2
```diff
@@ -5,6 +5,11 @@ Application instances
 
 {% block content %}
 
+
+<div class="block has-text-right">
```
marcospri, line 9: Create button in the main page.

lms/views/application_instances.py
```diff
@@ -10,19 +10,10 @@
 )
 def create_application_instance(request):
     """Create application instance in the database and respond with key and secret."""
-
-    # Default developer_key and developer_secret to None rather than letting
-    # them be empty strings.
     developer_key = request.params["developer_key"].strip()
     developer_secret = request.params["developer_secret"].strip()
 
-    # If either one of developer_key or developer_secret is missing, then we
```
marcospri, line 19: Moved to the service itself.

lms/models/lti_registration.py
```diff
@@ -33,3 +33,7 @@ class LTIRegistration(CreatedUpdatedMixin, BASE):
     """Location of the public keys provied by the issuer for this registration"""
     token_url = sa.Column(sa.UnicodeText, nullable=False)
     """Endpoint to request new oauth tokens to use with LTIA APIs"""
+
+    application_instances = sa.orm.relationship(
```
marcospri, line 37: The other side of of `back_populate` in application_instance.

Useful avoid writing a ORM query on the admin pages.

lms/templates/admin/instance.html.jinja2
```diff
@@ -50,43 +31,69 @@
...
-    {{ read_only_field("Tool consumer instance description", instance.tool_consumer_instance_description) }}
-    {{ read_only_field("Tool consumer instance URL", instance.tool_consumer_instance_url) }}
-    {{ read_only_field("Tool consumer instance name", instance.tool_consumer_instance_name) }}
-    {{ read_only_field("Tool consumer instance contact email", instance.tool_consumer_instance_contact_email) }}
-    {{ read_only_field("Tool consumer info version", instance.tool_consumer_info_version) }}
+    <fieldset class="box">
+        <legend class="label has-text-centered">Tool Consumer</legend>
+        {{ macros.disabled_text_field("GUID", instance.tool_consumer_instance_guid) }}
+        <div class="columns">
+            <div class="column is-half">
```
marcospri, line 48: Side by side of some Tool consumer attributes.

lms/templates/admin/instance.new.html.jinja2
```diff
@@ -0,0 +1,33 @@
...
+{% endblock %}
+
+{% block content %}
+
+<form method="POST" action="{{ request.route_url("admin.registration.new.instance", id_=lti_registration.id) }}">
+    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+    <fieldset class="box mt-6">
+        <legend class="label has-text-centered">Registration</legend>
+
+        {{ macros.registration_preview(request, lti_registration) }}
```
marcospri, line 17: Reference to the registration

lms/templates/admin/registration.html.jinja2
```diff
@@ -0,0 +1,21 @@
+{% import "macros.jinja" as macros %}
+
+{% extends "lms:templates/admin/base.html.jinja2" %}
+{% block header %}
+    Registration {{registration.issuer}} / {{ registration.client_id }}
+{% endblock %}
+
+{% block content %}
+    {{ macros.registrations_fields(request, registration) }}
```
marcospri, line 9: Details of the registration.

lms/templates/admin/registration.html.jinja2
```diff
@@ -0,0 +1,21 @@
...
+
+    <fieldset class="box">
+    <legend class="label has-text-centered">Registration application instances</legend>
+        <div class="block has-text-right">
+            <a class="button is-primary" href="{{ request.route_url("admin.registration.new.instance", id_=registration.id) }}">New instance</a>
+        </div>
+
+        {% if registration.application_instances %}
```
marcospri, line 17: AIs that belong to this registration.

lms/views/admin/__init__.py
```diff
@@ -0,0 +1,23 @@
+from pyramid.httpexceptions import HTTPFound, HTTPNotFound
+from pyramid.view import forbidden_view_config, notfound_view_config, view_config
+
+
+@forbidden_view_config(path_info="/admin/*")
```
marcospri, line 5: Moved these from the main admin.py

lms/views/admin/application_instance.py
```diff
@@ -1,38 +1,21 @@
...
-
-@notfound_view_config(path_info="/admin/*", append_slash=True)
-def notfound(_request):
-    return HTTPNotFound()
+from lms.views.admin import check_required_and_redirect
 
 
 @view_defaults(request_method="GET", permission=Permissions.ADMIN)
-class AdminViews:
+class AdminApplicationInstanceViews:
```
marcospri, line 12: Renamed and moved to a new file dealing only with AIs

lms/templates/admin/instance.html.jinja2
```diff
@@ -50,43 +31,69 @@
...
 
-    {{ read_only_field("Custom Canvas API domain", instance.custom_canvas_api_domain) }}
+    </fieldset>
 
-    <fieldset class="box mt-6">
+    <fieldset class="box">
         <legend class="label has-text-centered">Canvas settings</legend>
+        {{ macros.disabled_text_field("Custom Canvas API domain", instance.custom_canvas_api_domain) }}
+        {{ macros.disabled_text_field("Developer key", instance.developer_key) }}
         {{ settings_checkbox("Canvas sections enabled", "canvas", "sections_enabled") }}
```
marcospri, line 78: I haven't been able to set the settings side by side. I'll have another go in a different PR. I won't be defeated by CSS.

lms/routes.py
jon-betts: :shrug: This has the advantage of being a single place to look?

lms/services/application_instance.py
```diff
@@ -116,29 +116,52 @@ def search(
...
+                created=datetime.utcnow(),
+                settings=settings or {},
+                deployment_id=deployment_id,
+                lti_registration_id=lti_registration_id,
+            )
+            self._db.add(application_instance)
+            self._db.flush()  # Force the returned AI to have an ID
+        except IntegrityError:
+            # Rollback the failed transaction.
+            # Otherwise all following DB queries will fail
```
jon-betts, line 161: Why is this something we want to carry on after? What's the situation that causes this that we are ok with?

lms/services/lti_registration.py
```diff
@@ -27,6 +27,40 @@ def get(self, issuer: str, client_id: Optional[str] = None):
...
+
+        return query.all()
+
+    def create_registration(  # pylint:disable=too-many-arguments
+        self,
+        issuer: str,
+        client_id: str,
+        auth_login_url: str,
+        key_set_url: str,
+        token_url: str,
```
jon-betts, line 51: I'm not sure basic types add a lot. They don't cause much trouble either, but I think it might harm readability vs.

```python
def create_registration(  # pylint:disable=too-many-arguments
        self, issuer, client_id, auth_login_url, key_set_url, token_url
    ):
```

lms/services/lti_registration.py
```diff
@@ -27,6 +27,40 @@ def get(self, issuer: str, client_id: Optional[str] = None):
 
         return query.one_or_none()
 
+    def get_by_id(self, id_) -> LTIRegistration:
+        return self._db.query(LTIRegistration).get(id_)
+
+    def search_registrations(self, *, issuer=None, client_id=None) -> LTIRegistration:
```
jon-betts, line 33: Is the `*` here forcing `kwargs`? Do we want to force kwargs?

lms/templates/admin/macros.jinja
```diff
@@ -0,0 +1,131 @@
+{% macro field_body(label) %}
```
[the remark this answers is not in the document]
jon-betts: I don't think you have to put all the macros in here. You could keep the common ones and put things which are only used in one file like `registrations_fields` in that file.

Otherwise we introduce distance between the definition and the usage without any other benefit (like re-use).

lms/views/admin/application_instance.py
```diff
@@ -41,11 +28,65 @@ def index(self):
     def instances(self):  # pylint: disable=no-self-use
         return {}
 
+    @view_config(
+        route_name="admin.registration.new.instance",
+        request_method="POST",
+        renderer="lms:templates/admin/instance.new.html.jinja2",
+    )
+    def instance_new(self):
```
jon-betts, line 36:
```suggestion
    def new_instance(self):
```
?

lms/views/admin/lti_registration.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    def registrations(self):  # pylint: disable=no-self-use
+        return {}
+
+    @view_config(
+        route_name="admin.registration.new",
+        request_method="GET",
+        renderer="lms:templates/admin/registration.new.html.jinja2",
+    )
+    def registration_new(self):  # pylint: disable=no-self-use
```
jon-betts, line 32:
```suggestion
    def new_registration(self):  # pylint: disable=no-self-use
```

lms/views/admin/lti_registration.py
```diff
@@ -0,0 +1,131 @@
...
+        return {}
+
+    @view_config(
+        route_name="admin.registration.new",
+        request_method="POST",
+    )
+    def registration_new_post(self):
```
jon-betts, line 39:
```suggestion
    def new_registration_callback(self):
```
Not sure about the name, but I feel like "POST" doesn't give a lot away about what's going on?

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -97,11 +101,36 @@ def test_build_from_lms_url(self, developer_secret, developer_key, service):
...
+
+        with pytest.raises(IntegrityError):
+            service.create_application_instance(
+                "https://example.com/",
+                "example@example.com",
+                None,
+                None,
+                {},
+                deployment_id=lti_v13_application_instance.deployment_id,
+                lti_registration_id=lti_v13_application_instance.lti_registration.id,
```
jon-betts, line 128: Some `kwargs` here would be very helpful

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -84,8 +85,11 @@ def test_get_by_deployment_id_raises_on_missing(
             (None, None),
         ],
     )
-    def test_build_from_lms_url(self, developer_secret, developer_key, service):
-        application_instance = service.build_from_lms_url(
+    def test_create(self, developer_secret, developer_key, service, aes_service):
```
jon-betts, line 88:
```suggestion
    def test_create_application_instance(self, developer_secret, developer_key, service, aes_service):
```

tests/unit/lms/services/application_instance_service_test.py
```diff
@@ -97,11 +101,36 @@ def test_build_from_lms_url(self, developer_secret, developer_key, service):
...
+        if developer_secret and developer_key:
+            aes_service.build_iv.assert_called_once()
+            aes_service.encrypt.assert_called_once_with(
+                aes_service.build_iv.return_value, developer_secret
+            )
+
+            assert application_instance.developer_key == developer_key
             assert application_instance.developer_secret
 
+    def test_create_duplicate_raises(
```
jon-betts, line 114:
```suggestion
    def test_create_application_instance_duplicate_raises(
```

tests/unit/lms/services/lti_registration_test.py
```diff
@@ -10,9 +10,45 @@ class TestLTIRegistrationService:
     def test_get(self, svc, registration):
         assert svc.get(registration.issuer, registration.client_id) == registration
 
-    def test_without_client_id(self, svc, registration):
+    def test_get_without_client_id(self, svc, registration):
         assert svc.get(registration.issuer) == registration
 
+    def test_create(self, svc):
```
jon-betts, line 16:
```suggestion
    def test_get(self, svc):
```
?

tests/unit/lms/services/lti_registration_test.py
```diff
@@ -10,9 +10,45 @@ class TestLTIRegistrationService:
...
+
+        assert svc.get("ISSUER", "CLIENT_ID") == registration
+
+    def test_get_by_id(self, svc, registration, db_session):
+        # Force the registration to have an ID
+        db_session.flush()
+
+        assert svc.get_by_id(registration.id) == registration
+
+    def test_search(self, svc):
```
jon-betts, line 29:
```suggestion
    def test_search_registrations(self, svc):
```

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -2,28 +2,68 @@
...
+@pytest.mark.usefixtures(
+    "pyramid_config", "application_instance_service", "lti_registration_service"
+)
+class TestAdminApplicationInstanceViews:
+    def test_instances(self, views):
+        assert views.instances() == {}
+
+    def test_instance_new(
+        self, pyramid_request_new_instance, application_instance_service
+    ):
```
jon-betts, line 21: If you decided to change the names above, this should change with it etc.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -2,28 +2,68 @@
...
+
+        response = AdminApplicationInstanceViews(
+            pyramid_request_new_instance
+        ).instance_new()
+
+        assert response.status_code == 400
 
     def test_search_not_query(self, pyramid_request):
-        response = AdminViews(pyramid_request).search()
+        response = AdminApplicationInstanceViews(pyramid_request).search()
```
jon-betts, line 66: There is a `views` fixture which looks like this. It looks like it would remove this repetition to use it.

tests/unit/lms/views/admin/lti_registration_test.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def test_registration_new(self, views):
+        assert views.registration_new() == {}
+
+    def test_registration_new_post(
+        self, pyramid_request_new_registration, lti_registration_service
+    ):
+        response = AdminLTIRegistrationViews(
+            pyramid_request_new_registration
+        ).registration_new_post()
```
jon-betts, line 26:
```suggestion
    @pytest.mark.usefixture('with_form_submission')
    def test_registration_new_post(self, views, lti_registration_service):
        response = views.registration_new_post()
```

If we use `pyramid_request_new_registration` to modify the request instead of relying on it to return it, we can use the same fixture in each test. This will cut down on some repetition and make the tests easier to read.

I think this is true for all the tests here which modify the request before constucting their own view instance. I think you can construct the standard one, and then modify the request after. It will still reference it.

tests/unit/lms/views/application_instance_test.py
```diff
@@ -70,6 +39,10 @@ def test_it_sets_settings(
                 }
             ),
         )
+        assert result == {
+            "consumer_key": application_instance_service.create_application_instance.return_value.consumer_key,
+            "shared_secret": application_instance_service.create_application_instance.return_value.shared_secret,
+        }
```
jon-betts, line 45: Popping out `application_instance_service.create_application_instance.return_value` as `created_ai` or something might make this clearer.

tests/unit/lms/views/admin/lti_registration_test.py
```diff
@@ -0,0 +1,156 @@
...
+        del pyramid_request_new_registration.params[missing]
+
+        response = AdminLTIRegistrationViews(
+            pyramid_request_new_registration
+        ).registration_new_post()
+
+        assert response.status_code == 400
+
+    def test_search_not_query(self, pyramid_request):
+        response = AdminLTIRegistrationViews(pyramid_request).search()
```
jon-betts, line 71:
```suggestion
    def test_search_not_query(self, views, pyramid_request):
        response = views.search()
```

The same seems true for all the rest of the views which don't use the new stuff either.

jon-betts (changes requested):
From a users point of view I think this is everything it needs to be right now. I'm sure we could UX this to death and make it wonderful, but I think all the major surprises have been ironed out. In any case, this is as much due diligence I think it makes sense to do before we hand it over to users to get real feedback.

I've not gone over the templating code in any particular detail. As it works and looks as it should I'm not going to review the HTML here.

Due to the fact it's admin pages I'm similarly not going to go over the backend view stuff either. The proof is in the pudding and I think admin pages have a lower burden of goodness.

### What could we improve?

There are also quite a few places where the name of the test doesn't reflect the thing being tested. Either because it looks like the name was refactored, or in a few cases it's just different.

I think the testing could be tightened up. Some of the fixturing seems a bit off. There are some fixtures around that we aren't using for the views in a couple of places and I can't see why. I mean it works, but I think it could definitely be improved.

### Anything else?

If the `*,` is for forcing people to use `kwargs` I'm not sure how I feel about it. I feel like it's a good practice to do, so I'm not sure  why I'm uncomfortable going out of our way to force it. At a minimum it adds some noise to the method signature. If we added it everywhere I don't think I'd like it, and I don't see why we _wouldn't_ add it everywhere? It just seems a bit heavy handed.

I quite like type hints for objects, and maybe for lists and dicts sometimes. Adding `str` to every param in a method stuck me as odd though. I couldn't tell you exactly why, but I think to some extent `str` is kind of the default, and it adds a lot of noise and bumps a single line onto many.

I don't think I've worked out my own opinion on this stuff, and it doesn't do much harm either way.

So far I think I'm leaning towards:

 * Objects: yes!
 * Dicts and lists, or compounds of: where handy?
 * Str, int - don't like 'em, but why?

lms/services/application_instance.py
marcospri: All access to the DB thru the same session won't work after the integrity error without the rollback.

The specific case here is accessing the DB after the error creating the instance to take you to the registration page, we do a LTIRegistration.get_by_id that fails otherwise.

I could get the registration here at the beginning of the method in all cases https://github.com/hypothesis/lms/blob/7291206ac4398fd7ef3216d00a94297b2a35734e/lms/views/admin/application_instance.py#L37

it will be a wasted query in the happy path but it's not like the admin pages are the bastion of performance. With that we can forget about all this business for now.

marcospri: I ended up doing this. No need for the rollback now.

lms/views/admin/application_instance.py
```diff
@@ -33,32 +33,30 @@ def instances(self):  # pylint: disable=no-self-use
         request_method="POST",
         renderer="lms:templates/admin/instance.new.html.jinja2",
     )
-    def instance_new(self):
-        lti_registration_id = self.request.matchdict["id_"]
+    def new_instance(self):
+        lti_registration = self.lti_registration_service.get_by_id(
+            self.request.matchdict["id_"]
+        )
```
jon-betts, line 39: It's a little confusing that the `id_` field to new application instance is the id of a registration?

Issue #3990, opened 2022-05-26
Clear up tech-debt before assignment work begins


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122

Currently there is some mess around the assignment table (which is currently called `module_item_configurations`) specifically:

### Un-used Canvas tracking code

 * There's a lot of complexity we don't use in the assignment service and LTI launches about handling Canvas assignments
 * It's not used and it can go
 * The column `ext_lti_assignment_id` should be removed (we might bundle this with other changes for convenience)

### Renaming the table

The table `module_item_configuration` is extremely confusing - we refer to it as assignments internally this would be a good name

This is part of our "sane" pillar for reporting, to make SQL more intuitive. This isn't required, but it would go a long way to making things easier to conceptualise.

See:

 * https://petegraham.co.uk/rename-postgres-table-with-alembic/
 * https://www.postgresql.org/docs/current/sql-createview.html - Details here about requirements to make a view updatable

### Fixing some complexity around the course service

There's a draft PR for this linked below. We're likely to be messing about with this for upcoming assignment tracking, so we might as well make it easier to interact with first.

## Tasks

 * https://github.com/hypothesis/lms/pull/3997
   * https://github.com/hypothesis/lms/pull/4012
   * https://github.com/hypothesis/lms/pull/4022
   * https://github.com/hypothesis/lms/pull/4023
   * https://github.com/hypothesis/lms/pull/4024
 * https://github.com/hypothesis/lms/pull/3974
 * https://github.com/hypothesis/lms/pull/3999
 * https://github.com/hypothesis/lms/pull/4000
 * https://github.com/hypothesis/lms/pull/4009
 * https://github.com/hypothesis/lms/pull/4008
 * https://github.com/hypothesis/lms/pull/4010
 * https://github.com/hypothesis/lms/pull/4011
 * https://github.com/hypothesis/lms/pull/4013
 * https://github.com/hypothesis/lms/pull/4014
 * https://github.com/hypothesis/lms/pull/4015
 * https://github.com/hypothesis/lms/pull/3996
    * https://github.com/hypothesis/lms/pull/4019
 * https://github.com/hypothesis/lms/pull/4030
 * https://github.com/hypothesis/lms/pull/4029
 * https://github.com/hypothesis/lms/pull/4048
 * https://github.com/hypothesis/lms/pull/4031
 * https://github.com/hypothesis/lms/pull/4036
   *  https://github.com/hypothesis/lms/pull/4037

[...]

PR #4008, opened 2022-05-30
Unify (a bit) the course and grouping services

jon-betts:
For: 

Originally I was going to try and get this to pass a mapper along or something, but honestly I think it's too much fuss. There's still quite a bit going on without that:

 * `generate_authority_provided_id` -> `get_authority_provided_id`
    * No longer requires `tool_consumer_instance_guid`
    * As `parent` is optional, it's now the last param and defaults to `None`
 * `upsert_with_parent` -> `upsert_groupings`
    * No longer requires a parent (can be used for courses)
    * Sets `settings` on create
 * `upsert` -> `upsert_course`
    * Uses `upsert_groupings` internally instead of re-implementing

Other changes:

 * The course service is generally `context_id` centric now, rather than `authority_provided_id` centric
 * More test coverage than before

### Testing notes

 * Start all the things
 * Do `make sql`
 * `truncate grouping, grouping_membership;`
 * Open any assignment (e.g. https://hypothesis.instructure.com/courses/125/assignments/873)
 * `select * from grouping` - You should see rows
 * `update grouping set lms_name='SOMEVALUE'`
 * Reload the assignment
 * `select * from grouping` - The name should have been updated


lms/services/course.py
```diff
@@ -51,61 +57,43 @@ def upsert(self, context_id, name, extra, settings=None) -> Course:
...
-            authority_provided_id
+            context_id
         ):
             course_settings.set("canvas", "sections_enabled", False)
 
         return course_settings
 
-    def _is_pre_sections(self, authority_provided_id):
+    def _is_pre_sections(self, context_id):
         return bool(
```
marcospri, line 93: I wonder how often this is true. Probably something to look at and probably remove in another pass.

(It's not a lot of code though, doesn't get in the way).

marcospri (approved):

PR #4009, opened 2022-05-30
Rewrite the tests to use factories and fixtures

jon-betts:
Also removed a test which mostly seemed a repeat of the
authority id tests.


lms/services/grouping.py
```diff
@@ -57,27 +57,32 @@ def upsert_with_parent(
...
+                # From params
+                "parent_id": parent.id,
+                "type": type_,
+                # Things the caller provides
+                "lms_id": grouping["lms_id"],
+                "lms_name": grouping["lms_name"],
+                "extra": grouping.get("extra"),
+            }
+            for grouping in grouping_dicts
+        ]
```
jon-betts, line 80: Moving this out makes debugging a lot easier, as you can drop a break point after all this is done, instead of getting lead into all the stuff that happens here.

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
-    def test_when_empty(self, svc):
-        course = factories.Course()
-
+    def test_it_with_empty_list(self, svc, parent_course):
         assert not svc.upsert_with_parent(
             [],
             type_=Grouping.Type.CANVAS_GROUP,
-            parent=course,
+            parent=parent_course,
         )
```
jon-betts, line 84: Used a fixture for the parent course

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
 
-    def test_it(self, db_session, svc):
-        # Create a parent course for the groupings to belong to.
-        # The course has to belong to svc.application_instance.
-        course = factories.Course(application_instance=svc.application_instance)
-        # A pre-existing grouping that we'll update.
-        existing_grouping = {
-            "lms_id": "existing_id",
-            "lms_name": "existing_name",
-            "extra": {"existing": "extra"},
```
jon-betts, line 95: Split out the create and update parts

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
+            # We'll update existing_grouping's lms_name and extra.
+            "lms_name": "new_name",
+            "extra": {"updated": "extra"},
         }
-        # Unfortunately we have to use upsert_with_parent() itself to insert the pre-existing grouping.
-        svc.upsert_with_parent(
-            [existing_grouping],
-            type_=Grouping.Type.CANVAS_GROUP,
-            parent=course,
-        )
```
jon-betts, line 102: We no longer rely on the thing we are testing to test itself. We still use `generate_authority_provided_id` though, so it's not perfect.

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
+        assert groupings == [created_grouping]
+        assert created_grouping == Any.object.with_attrs(attrs)
+        assert created_grouping == Any.object.with_attrs(
+            {
+                "application_instance": svc.application_instance,
+                "authority_provided_id": Any.string(),
+                "updated": Any(),
+                "parent": parent_course,
+                "type": Grouping.Type.CANVAS_GROUP,
+            }
```
jon-betts, line 111: Bit more detail added to the tests here

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
-        assert blackboard_group.type != canvas_section.type
-        assert (
-            canvas_group.authority_provided_id != canvas_section.authority_provided_id
-        )
-        assert (
-            canvas_group.authority_provided_id != blackboard_group.authority_provided_id
-        )
-        assert (
-            blackboard_group.authority_provided_id
-            != canvas_section.authority_provided_id
```
jon-betts, line 182: This was a really long winded test which duplicates testing for the `generated_authority_provided_id` mechanism, but doing it through upsert.

tests/unit/lms/services/grouping_test.py
```diff
@@ -75,111 +76,85 @@ def generate_authority_provided_id(self):
...
+            extra={"existing": "extra"},
+            parent=parent_course,
+            # Construct the authority_provided_id correctly, otherwise we get
+            # a fake one generated by factory boy that doesn't line up
+            authority_provided_id=GroupingService.generate_authority_provided_id(
+                parent_course.application_instance.tool_consumer_instance_guid,
+                lms_id,
+                parent_course,
+                type_=Grouping.Type.CANVAS_GROUP,
+            ),
```
jon-betts, line 156: I think this awkwardness here is why the previous author used the function to insert this. It's certainly less bother.

PR #4010, opened 2022-05-30
Remove `h_group` property from the context

jon-betts:
For: 

This was just calling get or create course. This has also pointed out that we were clearly calling it more than once in lots of places.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -40,28 +40,6 @@ class LTILaunchResource:
         )
 
     @property
-    def h_group(self):
-        """
-        Return the h group for the current request.
-
-        The group's name is generated from the LTI course's title and is
-        usually a valid Hypothesis group name.  For example if the course's
-        title is too long for a Hypothesis group name it'll be truncated. But
-        this doesn't currently handle course titles that are *too short* to be
-        Hypothesis group names (shorter than 3 chars) - in that case if you try
-        to create a Hypothesis group using the generated name you'll get back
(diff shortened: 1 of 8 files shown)


lms/views/basic_lti_launch.py
```diff
@@ -70,7 +69,7 @@ def sync_lti_data_to_h(self):
         """
 
         self.request.find_service(name="lti_h").sync(
-            [self.context.h_group], self.request.params
```
marcospri, line 73: We should change the signature and docs here (they were wrong before this PR):

https://github.com/hypothesis/lms/blob/refactor-h-group/lms/services/lti_h.py#L28

Probably `groupings` instead of h_groups?

lms/resources/lti_launch.py
```diff
@@ -39,28 +39,6 @@ def get_or_create_course(self):
             extra=self._course_extra(),
         )
 
-    @property
```
marcospri, line 42: Can't comment on `def get_or_create_course` line but 

- Should it be a `@property def course()` I don't dislike the `get_or_create_name` but a lof of things in this resource seem to be nouns.

- Should it be `lru_cached?`

lms/views/basic_lti_launch.py
jon-betts: I think this is true, but it's also quite a big change on it's own (as it passes all the way through that class. I think I'll do it in a separate PR.

marcospri (approved):

PR #4012, opened 2022-05-31
Remove unused `AssignmentService` methods

marcospri:
For: 

All of this was only used in the view removed in https://github.com/hypothesis/lms/pull/4000

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -1,7 +1,5 @@
 from functools import lru_cache
 
-from sqlalchemy.orm.exc import MultipleResultsFound
-
 from lms.models import Assignment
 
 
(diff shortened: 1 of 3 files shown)


lms/services/assignment.py
```diff
@@ -11,225 +9,48 @@ class AssignmentService:
...
-
-           i.  One with the matching resource_link_id and no ext_lti_assignment_id
-           ii. And one with the matching ext_lti_assignment_id and no resource_link_id
-
-           The assignment with the resource_link_id will always be first in the sequence.
-
-        :rtype: sequence of either 0, 1 or 2 models.Assignment objects
-        """
-        assert resource_link_id and ext_lti_assignment_id
+    def get(self, tool_consumer_instance_guid, resource_link_id):
```
marcospri, line 12: All of the code here (in this file) deals mostly with:

- Finding things by either of two different IDS.
- Merging assignments when they were created at different times with different IDs.

lms/services/assignment.py
```diff
@@ -11,225 +9,48 @@ class AssignmentService:
...
-        self._get_for_canvas_assignment_config.cache_clear()
-
-        self.get.cache_clear()
-        # Private methods are cached so different but equivalent args to get
-        # (default values, args vs kwargs) still don't hit the database.
-        self._get_by_resource_link_id.cache_clear()
-        self._get_for_canvas_assignment_config.cache_clear()
-        self.get_for_canvas_launch.cache_clear()
-        self.exists.cache_clear()
-
```
jon-betts, line 232: :wave:

tests/unit/lms/services/assignment_test.py
```diff
@@ -600,12 +129,10 @@ def noise():
     factories.Assignment(
         tool_consumer_instance_guid=MATCHING_TOOL_CONSUMER_INSTANCE_GUID,
         resource_link_id="noise_resource_link_id",
-        ext_lti_assignment_id="noise_ext_lti_assignment_id",
     )
     factories.Assignment(
         tool_consumer_instance_guid="noise_tool_consumer_instance_guid",
         resource_link_id=MATCHING_RESOURCE_LINK_ID,
-        ext_lti_assignment_id=MATCHING_EXT_LTI_ASSIGNMENT_ID,
     )
```
jon-betts, line 136: Good lord these tests were horribly long.

There's something bugging me about these still looking at them, but I think that's a job for another PR. The `MATCHING_TOOL_CONSUMER_INSTANCE_GUID` stuff at the top seems a bit extra. I suspect we could create a fixture and then refer to the values from it.

jon-betts (approved):
Good lord this is a lot of complexity that's been ripped out. I think this might spur us to refactor again now we can see what's left more clearly, but that's absolutely a job for follow up PRs.

I think in particular I'd like to start a discussion about how to name service methods so we don't have 1000 hits for `get(` in our code etc. I don't think we need an iron clad rule from which we never deviate, but it might be nice to have a guideline for when we are being lazy.

There's one typo in the doc string, but that's really all I have to say about it.

PR #4018, opened 2022-06-01
Upgrade pyjwt

marcospri:


jon-betts (approved):
The tests all pass. None of the content in the change log looks like it's changing the interface at all, so this looks like a good fix.

PR #4022, opened 2022-06-02
Remove references to `ext_lti_assignment`

marcospri:
For:

Remove references to the column before another PR to remove it from the table.

Discovered a unused schema that wasn't deleted in a previous cleanup.


jon-betts (approved):
The PR as written looks good, but is this missing the constraint in `lms.models.assignment`:

```python
    sa.CheckConstraint(
        "NOT(resource_link_id IS NULL AND ext_lti_assignment_id IS NULL)",
        name="nullable_resource_link_id_ext_lti_assignment_id",
    ),
```

Or should that come later?

PR #4026, opened 2022-06-02
Add article metadata and thumbnail to JSTOR picker

robertknight:
**Depends on ~~https://github.com/hypothesis/lms/pull/4025~~, ~~https://github.com/hypothesis/lms/pull/4027~~, https://github.com/hypothesis/lms/pull/4028, ~~https://github.com/hypothesis/lms/pull/4039~~**

This PR adds article metadata and thumbnails to the JSTOR article picker form:

JSTOR assignment picker

The thumbnail we currently display is of the article itself. There is ongoing discussion [here](https://hypothes-is.slack.com/archives/C07NXBDNW/p1654182208320269) of whether we want to use something else (eg. the issue/book cover) instead, as that may look better.

The current implementation does not check whether the user has access to the content, the plan is to add that in future we'd show some appropriate indication.

The content of this dialog overlaps somewhat with the form used for picking VitalSource books. It isn't completely clear what a useful abstraction might be that could be used for both (eg. would it be just the layout or also some of the interaction/data fetching logic?) so that has been left for subsequent work.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/JSTORPicker.js
+++ b/lms/static/scripts/frontend_apps/components/JSTORPicker.js
@@ -10,7 +10,28 @@ import {
 import classnames from 'classnames';
 import { useRef, useState } from 'preact/hooks';
 
-import { toJSTORUrl } from '../utils/jstor';
+import { formatErrorMessage } from '../errors';
+import { urlPath, useAPIFetch } from '../utils/api';
+import { articleIdFromURL, jstorURLFromArticleId } from '../utils/jstor';
+
+/**
+ * @template T
+ * @typedef {import('../utils/fetch').FetchResult<T>} FetchResult
+ */
+
+/**
+ * Response for an `/api/jstor/articles/{article_id}` call.
+ *
+ * @typedef Metadata
+ * @prop {string} title
+ */
+
+/**
+ * Response for an `/api/jstor/articles/{article_id}/thumbnail` call.
+ *
+ * @typedef Thumbnail
+ * @prop {string} image
+ */
 
 /**
  * @typedef JSTORPickerProps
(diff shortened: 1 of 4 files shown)


robertknight:
Articles that do not have a title are displayed without any useful metadata:

JSTOR picker no title

There is a follow-up ticket to improve this: https://github.com/hypothesis/lms/issues/4049

lyzadanger:
I note that the commit message includes:

> Cancel metadata/thumbnail fetch if the user types in the URL field
   while the fetch is in-flight. This avoids potential confusion due to
   the thumbnail/metadata not matching the current field value.

This is _true_, but can be hard to spot in the associated diff because it happens magically behind the scenes. I wonder if there is another way to word this that is more explicit about what causes the fetch cancelation?

lms/static/scripts/frontend_apps/components/test/JSTORPicker-test.js
```diff
@@ -1,26 +1,70 @@
...
+  }
+
   const renderJSTORPicker = (props = {}) => mount(<JSTORPicker {...props} />);
 
   beforeEach(() => {
-    fakeToJSTORUrl = sinon.stub().returns('jstor://1234');
+    fakeArticleIdFromURL = sinon.stub().returns('1234');
+    setAPIFetchResult = {};
+
+    function useAPIFetchFake(path) {
```
lyzadanger, line 46: The mocking in this test module is wee bit complex...

lyzadanger (commented):
This is working as designed in my basic testing.

A few high-level notes:

* The test mocking for `JSTORPicker` is somewhat complex
* It would be neato if this interface would accept IDs, and/or be more accepting of different JSTOR domains
* If a metadata fetch fails because of a transient network glitch, the user is stuck: clicking the "arrow button" next to the URL field will not re-attempt the fetch as the path/`key` is the same as the last fetch attempt and the underlying hook won't re-fetch. i.e. there is no "try again" possible with the hook's current design.
* As a follow-up, I'd like to adjust icon alignment to look better when titles wrap:
Screen Shot 2022-06-07 at 9 40 16 AM

robertknight:
> If a metadata fetch fails because of a transient network glitch, the user is stuck: clicking the "arrow button" next to the URL field will not re-attempt the fetch as the path/key is the same as the last fetch attempt and the underlying hook won't re-fetch. i.e. there is no "try again" possible with the hook's current design.

I just so happened to be working on this! - See https://github.com/hypothesis/lms/pull/4055.

lms/static/scripts/frontend_apps/components/JSTORPicker.js
```diff
@@ -26,24 +47,46 @@ import { toJSTORUrl } from '../utils/jstor';
...
+    articleId ? urlPath`/api/jstor/articles/${articleId}/thumbnail` : null
+  );
+
+  let renderedError;
+  if (error) {
+    renderedError = error;
+  } else if (metadata.error) {
+    renderedError = formatErrorMessage(
+      metadata.error,
+      'Unable to fetch article details'
```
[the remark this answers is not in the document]
robertknight: > Is this "game over" for now? Is the expected human behavior that a user would try a different article ID, or cancel out of the picker,

Yes, those are the available options at present.

Some additional things we could add in future:

- A retry button, to handle transient network issues (relies on the PR that adds a `retry` callback to the `useFetch` result)
- Support the proxied forms of JSTOR URLs (eg. https://www-jstor-org.someproxy.myedu.org/stable/12234) and article IDs as inputs, as I think these are likely to be the most common kind of invalid URL here
- A link to the knowledge base explaining how to get JSTOR URLs

lms/static/scripts/frontend_apps/components/test/JSTORPicker-test.js
robertknight: If we're going to mock stateful hooks, that is unavoidably a bit more fiddly than functions. We may be able to create or find a useful abstraction for this though.

lyzadanger (approved):
Let's get this landed!

PR #4028, opened 2022-06-02
Add API views to fetch JSTOR article metadata and thumbnails

robertknight:
These will be used by the frontend when configuring a JSTOR assignment, to help the user check they are using the right content. To see how these views get used, see https://github.com/hypothesis/lms/pull/4026.

 - Add `/api/jstor/articles/{article_id}` route which returns article
   metadata
 - Add `/api/jstor/articles/{article_id}/thumbnail` route which returns
   the article image as a `data:image...` URI
 - Ensure that DOIs are escaped when generating API URLs. The JSTOR API
   accepts DOIs with or without encoding "/", but since these article
   IDs come from user input they could include other characters that
   ought to be encoded.

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -1,9 +1,21 @@
 from datetime import timedelta
+from urllib.parse import quote
+
+import requests
+from marshmallow import fields
 
 from lms.services.exceptions import ExternalRequestError
 from lms.services.http import HTTPService
 from lms.services.jwt import JWTService
+from lms.validation._base import RequestsResponseSchema
 from lms.views.helpers import via_url
+
+
+class JSTORMetadataSchema(RequestsResponseSchema):
+    """Response schema for `/metadata/{doi}` endpoint in JSTOR API."""
+
+    # The response has other fields, but we currently only use the "title" field.
+    title = fields.List(fields.Str(), required=True)
 
 
 class JSTORService:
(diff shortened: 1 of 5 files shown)


lms/services/jstor.py
```diff
@@ -62,28 +115,46 @@ def _get_public_url(self, url):
...
-        return s3_url
+def format_uri(template: str, *params: str):
+    """
+    Format a URI string.
+
+    This is like `template.format(*params)` except that the params are
+    percent-encoded.
+    """
+    encoded = [quote(param, safe="") for param in params]
+    return template.format(*encoded)
```
robertknight, line 157: I discussed last week using a library like [uritemplate](https://uritemplate.readthedocs.io/en/latest/) to handle safely escaping dynamic path components. However I realized that for our very limited needs, this two-liner is fine for the time being and saves a dependency. On the frontend we have a `urlPath` utility which does the same thing (example: ``urlPath`/api/jstor/articles/${doi}` ``).

lms/services/jstor.py
```diff
@@ -62,28 +115,46 @@ def _get_public_url(self, url):
...
+        s3_url = self._api_request(format_uri("/pdf-url/{}", doi)).text
 
+        if not s3_url.startswith("https://"):
+            raise ExternalRequestError(
+                f"Expected to get an S3 URL but got: '{s3_url}' instead"
+            )
+
+        return s3_url
+
+    def _api_request(self, endpoint: str) -> requests.Response:
```
marcospri, line 129: I was going to suggest doing the `format_uri` here instead than from the caller but that will only work nicely as long as what's true about this two endpoints is true for any new ones.

lms/services/jstor.py
```diff
@@ -62,28 +115,46 @@ def _get_public_url(self, url):
...
+            return f"{self.DEFAULT_DOI_PREFIX}/{article_id}"
+        return article_id
 
-        if not s3_url.startswith("https://"):
-            raise ExternalRequestError(
-                f"Expected to get an S3 URL but got: '{s3_url}' instead"
-            )
 
-        return s3_url
+def format_uri(template: str, *params: str):
```
marcospri, line 149: TIL about the typing of args

https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

robertknight:
Thanks for the review Marcos!

PR #4029, opened 2022-06-02
Refactor LTI launch tests to remove caller pattern and make them easier to maintain

jon-betts:
For:

This PR rewrites the unit tests for the Basic LTI launch views to:

 * Reduce the length from 510 lines to 342
 * Remove the caller pattern
 * Test commonly used internal methods once, and then use fixtures to assert that other things call the right things
 * Increase test coverage in a number of places
 * Make the test layout closely follow the file layout, to make it obvious to spot coverage issues

The end result is it should be much easier to make changes to the class, and know where to put tests for them, and have confidence you've actually covered all the cases.

### DIFF
--- a/tests/unit/lms/views/lti/basic_launch_test.py
+++ b/tests/unit/lms/views/lti/basic_launch_test.py
@@ -1,4 +1,5 @@
 from unittest import mock
+from unittest.mock import Mock, sentinel
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


marcospri (approved):
Nice, we have a few of these, some module that is tested with a  different style. One less now :+1:

jon-betts:
Jenkins is a bit broken at the moment, but this code is actually fine. This is related to the move over to `main` as the main branch.

PR #4031, opened 2022-06-02
Move items out of the constructor of the LTI launches into the correct places

jon-betts:
For:


lms/views/lti/basic_launch.py
```diff
@@ -291,15 +275,20 @@ def _do_launch(self, document_url, grading_supported=True):
         if grading_supported:
             self.context.js_config.maybe_enable_grading()
 
+        self.context.js_config.enable_lti_launch_mode()
+        self.context.js_config.maybe_set_focused_user()
```
jon-betts, line 279: `maybe_set_focused_user` appears to be grading related, but it's actually Speedgrader related. Which in our world is `grading_supported=False`. This is only for Canvas.

tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -252,9 +235,16 @@ def test__store_lti_data(
         lti_h_service,
         LtiLaunches,
         grading_info_service,
+        application_instance_service,
     ):
+        application_instance_service.get_current.return_value = Mock()
```
jon-betts, line 240: Our standard application instance service returns a real application instance, which is kind of a pain here

lms/views/lti/basic_launch.py
marcospri: This is a bit of a mess but I wouldn't get into refactoring it now.

We very explicitly say here `grading_supported` but the we call all of this `maybe_` methods that do check themselves if that's the case.

`js_config.add_document_url` calls, unconditionally, self._add_canvas_speedgrader_settings (which should be renamed to have the `maybe_?).

This means that for canvas_files, the only place we pass `grading_supported=False` we don't call `maybe_enable_grading()` (that checks itself if we are in canvas) but for the rest, for all non-canvas launches we call `add_canvas_speedgrader_settings` that also check if we are in canvas.

I think once we try to tackle recording if assignment are gradable etc, we should unify "maybe_enable_grading", call it all the time and have it deal with this differences.


For this PR I think what we are doing is fine.

lms/views/lti/basic_launch.py
```diff
@@ -184,6 +167,10 @@ def unconfigured_launch_not_authorized(self):
         document (for example a learner rather than a teacher). We just show an
         error page.
         """
+        # We should still update all the data we can, even though the user
+        # can't make use of it right away
+        self._store_lti_data()
```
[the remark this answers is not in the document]
jon-betts: All the things that happen in `store_lti_data()` apart from the assignment and grading. Which I think are at the moment:

 * `application_instance`
 * `course` and `user` (sent to `h`)
 * Adding a row to `LtiLaunches`

lms/views/lti/basic_launch.py
```diff
@@ -291,15 +278,20 @@ def _do_launch(self, document_url, grading_supported=True):
...
         self.context.js_config.add_document_url(document_url)
 
         return {}
 
     def _store_lti_data(self):
         """Store LTI launch data in our LMS database."""
-
         request = self.request
 
+        request.find_service(name="application_instance").get_current().update_lms_data(
```
marcospri, line 291: This _could_ make sense in the constructor.

The main thing update_lms_data is doing is potentially raising `ReusedConsumerKey`, yeah it might update some data like it says in the tin but the raising part is what matters most.

marcospri (approved):

lms/views/lti/basic_launch.py
jon-betts: There's a separate PR to try and deal with the `maybe_` world: https://github.com/hypothesis/lms/pull/4050

Issue #3995, opened 2022-05-26
Ensure assignments are updated whenever we see them


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122

Depends on:

 * https://github.com/hypothesis/lms/issues/3990 - We really want the Canvas assignment simplifications before we start this

Currently there's quite a patch work of different ways of dealing with assignments in the different variations of LTI launches in `basic_lti_launch.py`.

Some goals:

 * Ensure we always attempt to upsert the assignment whenever we interact with it
 * Try and get this to happen in the fewest places in the code (currently it's all over the place)

This is because we want to be tracking more things which might change (like the name, and definitely the updated date) and we don't want it to happen in loads of places which will make accomplishing this hard.

## Tasks

* https://github.com/hypothesis/lms/pull/4040
* https://github.com/hypothesis/lms/pull/4031
* https://github.com/hypothesis/lms/pull/4032

PR #4032, opened 2022-06-02
Upsert data about assignments on every launch

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/3995

### What's changed?

Previously every launch either did, or didn't update the assignment:

* `configure_assignment` - Yes (unchanged)
* `blackboard_copied_launch` - Yes (unchanged)
* `brightspace_copied_launch` - Yes (unchanged)
* `canvas_file_launch` - Yes (unchanged)
* `db_configured_launch` - No (now **Yes**)
* `url_configured_launch` - No (now **Yes**)
* `legacy_vitalsource_launch` - No (now **Yes**)
* `unconfigured_launch` - No (unchanged)
* `unconfigured_launch_not_authorized` - No (unchanged)

In addition to doing this for all configured launches:

 * We also only do it in one place
 * `_do_launch` has been renamed `_show_document` to reflect the fact it's only document based launches, not all launches

### DIFF
--- a/lms/views/lti/basic_launch.py
+++ b/lms/views/lti/basic_launch.py
@@ -75,7 +75,7 @@ class BasicLaunchViews:
         if group_set := self.request.parsed_params.get("group_set"):
             extra["group_set_id"] = group_set
 
-        return self._do_launch(
+        return self._show_document(
             document_url=self.request.parsed_params["document_url"],
             assignment_extra=extra,
         )
@@ -95,7 +95,7 @@ class BasicLaunchViews:
         # The ``db_configured=True`` view predicate ensures that this view
         # won't be called if there isn't a matching document_url in the DB. So
         # here we can safely assume that the document_url exists.
-        return self._do_launch(
+        return self._show_document(
             document_url=self.assignment_service.get_assignment(
                 self.context.lti_params["tool_consumer_instance_guid"],
                 self.context.resource_link_id,
@@ -113,7 +113,7 @@ class BasicLaunchViews:
         and saved in the LMS, which passes it back to us in each launch request.
         All we have to do is pass the URL to Via.
         """
-        return self._do_launch(document_url=self.request.parsed_params["url"])
+        return self._show_document(document_url=self.request.parsed_params["url"])
 
     @view_config(
         authorized_to_configure_assignments=True,
@@ -214,7 +214,7 @@ class BasicLaunchViews:
         course_id = self.context.lti_params["custom_canvas_course_id"]
         file_id = self.request.params["file_id"]
 
-        return self._do_launch(
+        return self._show_document(
             document_url=f"canvas://file/course/{course_id}/file_id/{file_id}",
             grading_supported=False,
         )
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -93,38 +92,33 @@ def test_configure_assignment(
...
-            pyramid_request.parsed_params["document_url"],
-            pyramid_request.parsed_params["tool_consumer_instance_guid"],
-            pyramid_request.parsed_params["resource_link_id"],
-            extra=expected_extras,
-        )
-
-        _do_launch.assert_called_once_with(
+        _show_document.assert_called_once_with(
             document_url=pyramid_request.parsed_params["document_url"],
-            grading_supported=True,
```
jon-betts, line 105: This is the default

lms/views/lti/basic_launch.py
```diff
@@ -101,12 +95,12 @@ def db_configured_launch(self):
...
-        return self._do_launch(document_url=document_url)
+        return self._do_launch(
+            document_url=self.assignment_service.get_assignment(
+                self.context.lti_params["tool_consumer_instance_guid"],
+                self.context.resource_link_id,
+            ).document_url
+        )
 
     @view_config(url_configured=True, schema=URLConfiguredBasicLTILaunchSchema)
     def url_configured_launch(self):
```
marcospri, line 106: There's a subtle going on here left from the last attempt:

- `db_configured_launch` has both `db_configured=True` and `url_configured=False`
- `url_configured_launch` has just `url_configured=True`


So before this change `db_configured` and `url_configured` never were true at the same time but now the can and `url_configured_launch` is taking precedence.

I think every test would pass if you remove the `url_configured=False` from `db_configured_launch` but assignment editing will stop working on Canvas.

We talk about how the predicates hide this things so that's something to to have in mind for an eventual refactor. 

Looks like a for a lot of these, what's different about them is how we construct/where we pick from the `document_url` maybe that could help in predicate-less refactor.

jon-betts: Yeah... they are hugely confusing. I think this is a "hiding" rather than "revealing" abstraction. The details are tucked away in such a way as it's difficult to think about them.

I don't believe this PR changes anything about these though?

marcospri: No, not related to this PR

I was trying to find some edge  case where having the row in the DB would confuse the existing code but I haven't found any.

marcospri (approved):
I'v been testing this and can't find any problems. I'm scared from the last attempt when interaction of a few things caused bugs (speedgrader + canvas bug + row in the DB + ... = bug) but I think we are good :+1: now.

lms/views/lti/basic_launch.py
jon-betts: I was working through the logic of all of this, and I think there are basically two main cases for the configured launches:

 * Those that read from params to get the URL
 * Those that read from the assignment (course copied + configured)
 
In the case of the param based ones, the presence of the assignment is irrelevant. It's upsert either way. The other types do require it to exist. 

So I think the work this might be trying to do is:

 * `url_configured_launch` 
    * `url_configured=True` - This is correct, we don't care about db configured or not
 * `db_configured_launch`
    * `db_configured=True` - This is correct, it's required
    * `url_configured=False` - This feels like an attempt to avoid precidence issues. This is "true", but unnecessary if url configured kicks in first. It's also not complete, as really this should be "course_copied=False" as well etc. So I'm not sure if this is required or not

PR #4034, opened 2022-06-03
LTIParams changes

marcospri:
- Refactor the `_to_lti_v11`. Stop using `for...else` and use a more explicit `KeyError` except block.

- Move handling of int -> str conversion for custom canvas parameters to LTIParams .


jon-betts (approved):
Yep, these both look really good. 

Pulling the integer conversion into the method definitely looks like the right thing to do. If we find ourselves getting more and more of these we might consider adding a more general solution, but it's not required for two fields.

PR #4036, opened 2022-06-03
Remove the `ext_lti_assignment_id` column from the model

marcospri:
For:

Remove the code first, migration will follow on https://github.com/hypothesis/lms/pull/4037.

The order is reversed from other DB changes as here we are deleting a column, we want to the code to stop using it before we delete it for real in the DB.

#  Testing 

Launch a few canvas assignments as a sanity check:

- https://hypothesis.instructure.com/courses/125/assignments/873 
- https://hypothesis.instructure.com/courses/125/assignments/875
- Also SpeedGrading https://hypothesis.instructure.com/courses/125/gradebook/speed_grader?assignment_id=873&student_id=35

### DIFF
--- a/lms/models/assignment.py
+++ b/lms/models/assignment.py
@@ -24,23 +24,11 @@ class Assignment(BASE):
     __tablename__ = "assignment"
     __table_args__ = (
         sa.UniqueConstraint("resource_link_id", "tool_consumer_instance_guid"),
-        sa.UniqueConstraint("tool_consumer_instance_guid", "ext_lti_assignment_id"),
-        sa.CheckConstraint(
-            "NOT(resource_link_id IS NULL AND ext_lti_assignment_id IS NULL)",
-            name="nullable_resource_link_id_ext_lti_assignment_id",
-        ),
     )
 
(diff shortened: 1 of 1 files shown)


lms/models/assignment.py
```diff
@@ -24,23 +24,11 @@ class Assignment(BASE):
...
 
-    ext_lti_assignment_id = sa.Column(sa.UnicodeText, nullable=True)
-    """
-    ID given for the assignment by Canvas.
-
-    Null for non canvas assignments and older Canvas ones.
-    """
-
-    resource_link_id = sa.Column(sa.Unicode, nullable=True)
+    resource_link_id = sa.Column(sa.Unicode, nullable=False)
```
jon-betts, line 31: I think you mentioned we would have to delete some rows to make this true? What happens if you change a constraint that invalidates existing rows anyway?

jon-betts (approved):
Aside from migrations there are no more references to `ext_lti_assignment_id` left. 

I'd be interested to know how a migration would interact with existing rows which don't meet the criteria.

PR #4039, opened 2022-06-03
Add `useAPIFetch` hook to simplify API fetches

robertknight:
This is a hook that wraps `useFetch` and deals with the details of making an authenticated API call for a given path + params. Fetching a given data item can then be done with a one-liner. Example from the upcoming JSTORPicker changes:

```js
  const [articleId, setArticleId] = useState(/** @type {string|null} */ (null));

  /** @type {FetchResult<Metadata>} */
  const metadata = useAPIFetch(
    articleId ? urlPath`/api/jstor/articles/${articleId}` : null
  );

  /** @type {FetchResult<Thumbnail>} */
  const thumbnail = useAPIFetch(
    articleId ? urlPath`/api/jstor/articles/${articleId}/thumbnail` : null
  );
```

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/api.js
+++ b/lms/static/scripts/frontend_apps/utils/api.js
@@ -1,4 +1,8 @@
+import { useContext } from 'preact/hooks';
+
+import { Config } from '../config';
 import { APIError } from '../errors';
+import { useFetch } from './fetch';
 
 /**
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
For follow-up: extract `useAPIFetch` and its tests from `utils` as it doesn't meet criteria for being a utility. 

Looks good!

Issue #3947, opened 2022-05-13
Keep track of  lti launches for LTI1.3


marcospri:
The plan is now settled on having a new table:

- https://github.com/hypothesis/lms/pull/4282
- https://github.com/hypothesis/lms/pull/4283
- https://github.com/hypothesis/lms/pull/4144
- https://github.com/hypothesis/lms/pull/4334
- https://github.com/hypothesis/lms/pull/4347
- https://github.com/hypothesis/lms/pull/4354

---
Instead of adapting the existing LtiLaunches table we could create a new, more flexible table to store this kind of event.

https://hypothes-is.slack.com/archives/C1MA4E9B9/p1652458221825969

If reusing the existing table we need to:

Similar to https://github.com/hypothesis/lms/issues/3719

Missed then as the column here is named `lti_key` and it's nullable.

The plan is to, in separate PRs:

- Add `application_instance_id` to the table as nullable column.
- Change the code to fill the new column
- Make a DB migration to back fill the new column and make it null able
- Review the impact in the analytics DB
- Stop filling the old column
- Remove the old column.

chrisshaw:
I would consider this need "maintain continuity for the current reporting using the table" (which is how I think you were already thinking, I saw you checking various reports in Metabase 🙂).

We will need it primarily to drive [this question](https://metabase.hypothes.is/question/1082-top-20-schools-by-lti-launches-last-16-weeks), which does break down launches by school and gets used regularly.

(It did strike me that, as long as we're still capturing `context_id` in the table, we shoudl be able to backfill at any time.)

PR #4041, opened 2022-06-03
Use `lti_params` for 1.3 compatibility in the LTI launches

jon-betts:
There were quite a few places where we didn't use the 1.1 -> 1.3 compat
layer. I think this might mean these were not working correctly for those
situations?


marcospri:
Good catch I was no paying much attention to LtiLaunes until I tackled https://github.com/hypothesis/lms/issues/3947 but there's not reason to insert non-sense there from the start.

PR #4042, opened 2022-06-06
Explicitly check for the user roles to send canvas submissions

marcospri:
In LTI1.1 we used the presence of `lis_result_sourcedid` as proxy for
being a student but that's not true in LTI1.3.

We'll check for the role in both LTI versions and also still require
`lis_result_sourcedid` to be present as is mandatory to talk to the
grading API.

## Testing

- On `master` launch https://hypothesis.instructure.com/courses/319/assignments/2988

Get an error ( the error being `User not found in course or is not a student`)

- Switch to `submision-teacher`, launch again, no error.

### DIFF
--- a/lms/views/lti/basic_launch.py
+++ b/lms/views/lti/basic_launch.py
@@ -298,9 +298,11 @@ class BasicLaunchViews:
             # For students in Canvas with grades to submit we need to enable
             # Speedgrader settings for gradable assignments
             # `lis_result_sourcedid` associates a specific user with an
-            # assignment. This is evidence that a student is launching us
-            if assignment_gradable and self.context.lti_params.get(
-                "lis_result_sourcedid"
+            # assignment.
+            if (
+                assignment_gradable
+                and self.request.lti_user.is_learner
+                and self.context.lti_params.get("lis_result_sourcedid")
             ):
                 self.context.js_config.add_canvas_speedgrader_settings(document_url)
 
--- a/tests/unit/lms/views/lti/basic_launch_test.py
+++ b/tests/unit/lms/views/lti/basic_launch_test.py
@@ -297,15 +297,31 @@ class TestBasicLaunchViews:
         context.js_config.enable_grading_bar.assert_not_called()
 
     @pytest.mark.usefixtures(
-        "with_gradable_assignment", "is_canvas", "with_student_grading_id"
+        "with_gradable_assignment",
+        "is_canvas",
+        "with_student_grading_id",
+        "user_is_learner",
     )
     def test__show_document_enables_speedgrader_settings(self, svc, context):
         svc._show_document(sentinel.document_url)  # pylint: disable=protected-access
 
-        context.js_config.add_canvas_speedgrader_settings(sentinel.document_url)
+        context.js_config.add_canvas_speedgrader_settings.assert_called_once_with(
+            sentinel.document_url
+        )
 
     @pytest.mark.usefixtures("with_gradable_assignment", "with_student_grading_id")
     def test__show_document_no_speedgrader_without_canvas(self, svc, context):
+        svc._show_document(sentinel.document_url)  # pylint: disable=protected-access
+
+        context.js_config.add_canvas_speedgrader_settings.assert_not_called()
+
+    @pytest.mark.usefixtures(
+        "is_canvas",
+        "with_gradable_assignment",
+        "with_student_grading_id",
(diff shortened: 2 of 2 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -307,21 +307,15 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
...
-        # Don't post submission to Canvas for these cases.
-        if not lis_outcome_service_url:
+        if (
+            # Don't set the Canvas submission params in non-Canvas LMS's.
+            not self._context.is_canvas
+            # Don't set the Canvas submission for non-gradable users
+            or not self._lti_user.is_instructor
+            # We need lis_result_sourcedid to talk to the API.
+            # Also on LTI1.1 lis_result_sourcedid == None for non-gradeable users
+            or not lis_result_sourcedid
```
marcospri, line 317: There's a change of behavior here. It could that canvas is sometimes sending us the `lis_result_sourcedid` for cases where "is_instructor" would be false.

If that's the case we'll stop sending submissions there but I think is either a bug (ours? canvas?) or something very rare.

lms/resources/_js_config/__init__.py
```diff
@@ -307,21 +307,15 @@ def _add_canvas_speedgrader_settings(self, **kwargs):
...
 
-        # Don't set the Canvas submission params in non-Canvas LMS's.
-        if not self._context.is_canvas:
-            return
-
-        # When a Canvas assignment is launched by a teacher or other
-        # non-gradeable user there's no lis_result_sourcedid in the LTI
-        # launch params.
-        # Don't post submission to Canvas for these cases.
-        if not lis_result_sourcedid:
```
marcospri, line 318: This was true for teachers in LTI1.3

lms/views/lti/basic_launch.py
```diff
@@ -298,9 +298,11 @@ def _configure_js_to_show_document(self, document_url, assignment_gradable):
             # For students in Canvas with grades to submit we need to enable
```
marcospri, line 298: We already mention here "student" we are just making it explicit in the code (and fixing a bug in the process).

tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -310,6 +310,14 @@ def test__show_document_no_speedgrader_without_canvas(self, svc, context):
 
         context.js_config.add_canvas_speedgrader_settings.assert_not_called()
 
+    @pytest.mark.usefixtures(
+        "with_gradable_assignment", "with_student_grading_id", "user_is_instructor"
+    )
+    def test__show_document_no_speedgrader_with_instructor(self, svc, context):
```
marcospri, line 316: The positive case was already covered.

lms/views/lti/basic_launch.py
```diff
@@ -298,9 +298,11 @@ def _configure_js_to_show_document(self, document_url, assignment_gradable):
...
             # Speedgrader settings for gradable assignments
             # `lis_result_sourcedid` associates a specific user with an
-            # assignment. This is evidence that a student is launching us
```
marcospri, line 301: > This is evidence that a student is launching us

Not anymore (not in LTI1.3)

tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -297,19 +297,35 @@ def test__show_document_does_not_enable_grading_for_canvas(self, svc, context):
...
+        "with_gradable_assignment",
+        "is_canvas",
+        "with_student_grading_id",
+        "user_is_learner",
     )
     def test__show_document_enables_speedgrader_settings(self, svc, context):
         svc._show_document(sentinel.document_url)  # pylint: disable=protected-access
 
-        context.js_config.add_canvas_speedgrader_settings(sentinel.document_url)
+        context.js_config.add_canvas_speedgrader_settings.assert_called_once_with(
```
marcospri, line 308: I've been yelling at the coverage report for a while until I notice this.

My personal favorite is asserting `assert_called_` themselves all the time.

jon-betts: Oh FFS... yes good catch. I'm thinking it might be worth replacing this whole mess of fixtures by just setting the correct params to make each of these things happen. This way it would also document what we actually recieve, and should make it easy to parameterize.

PR #4048, opened 2022-06-06
Separate checking for the application instance GUID and updating in the launch view

jon-betts:
For:

This is mostly "for show" to allow us to see what's happening. This allows a clean division between things we need for the view  to work (checking the GUID) and data storage for our persisted model.

The code here is the same / a little worse, as it calls the check GUID twice effectively. But it makes the intent clearer.

This also avoids a potential problem where something crucial is happening as a side effect of updating the data, but it's not clear. So you could easily think you are doing something safe by moving the update, where you are actually creating a huge problem.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -166,18 +166,12 @@ class ApplicationInstance(BASE):
             # guid identifies the rest of the LMS data, if not there skip any updates
             return
 
-        if (
-            self.tool_consumer_instance_guid
-            and self.tool_consumer_instance_guid != tool_consumer_instance_guid
-        ):
-            # If we already have a LMS guid linked to the AI
-            # and we found a different one report it to sentry
-            raise ReusedConsumerKey(
-                existing_guid=self.tool_consumer_instance_guid,
-                new_guid=tool_consumer_instance_guid,
-            )
+        self.check_guid_aligns(tool_consumer_instance_guid)
 
+        # This looks a bit weird after the statement above, but our GUID might
+        # be null and still pass the check above
         self.tool_consumer_instance_guid = tool_consumer_instance_guid
+
         for attr in [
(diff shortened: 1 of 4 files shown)


lms/models/application_instance.py
```diff
@@ -190,6 +184,25 @@ def update_lms_data(self, params: dict):
 
             setattr(self, attr, params.get(attr))
 
+    def check_guid_aligns(self, tool_consumer_instance_guid):
```
marcospri, line 187: Yeah this is more explicit :+1: 

We could eventually move this to the constructor of `LTILaunchResource` and remove it from `update_lms_data`. We avoid the double call there and we'll have deeplinking (which also calls `update_lms_data` and basic_lti_lauches do the same thing.

I can live without that, not sure if it's better.

lms/views/lti/basic_launch.py
```diff
@@ -298,19 +300,23 @@ def _do_launch(self, document_url, grading_supported=True):
...
         # user and group corresponding to the LTI user and course.
-        request.find_service(name="lti_h").sync([self.context.course], request.params)
+        self.request.find_service(name="lti_h").sync([self.context.course], lti_params)
 
         # Report all LTI assignment launches to the /reports page.
         LtiLaunches.add(
-            request.db,
-            request.params.get("context_id"),
-            request.params.get("oauth_consumer_key"),
+            self.request.db,
```
marcospri, line 313: I've seen this in another PR. It's fine to merge it with this if you want to avoid some chain-rebasing.

marcospri (approved):

lms/views/lti/basic_launch.py
jon-betts: Oh well. This should have already gone out and needed a rebase...

Issue #3991, opened 2022-05-26
Store more information about assignments


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122

Depends on:

 * https://github.com/hypothesis/lms/issues/3990 - We want the assignment updating to be simpler before we crack on with this
 * https://github.com/hypothesis/lms/issues/3995 - We want to know we've got the updating code everywhere before we add to it

There are some simple bits of metadata that are missing from our assignments table which make it hard to make use of:

 * We can't answer any questions about time
 * We don't know what assignments are called
 * We'd also like to know if assignments are graded or not
 
### Updates to assignment table

| Action | Field | Type | Reason |
|----------|--------|--------|------------|
| Add | `created` | timestamp | Assignment per time-period questions |
| Add | `updated` | timestamp | Assignment per time-period questions |
| Add | `graded` | boolean | Counting graded vs not in reports |
| Add | `title` | text | _[optional]_ Most reporting would benefit from being able to see what things are |
| Add | `description` | text | _[optional]_ Most reporting would benefit from being able to see what things are |

### Tasks

 * https://github.com/hypothesis/lms/pull/4050
 * https://github.com/hypothesis/lms/pull/4059
 * https://github.com/hypothesis/lms/pull/4056

jon-betts:
Done!

PR #4050, opened 2022-06-06
Move decision part of `maybe_` JS config to the LTI view

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/3991

We had a number of methods in the JSConfig class which detected whether they should perform an action and then performed it. These were:

 * `maybe_enable_grading()`
* `maybe_set_focused_user()`
* `add_canvas_speedgrader_settings()`

This is bad for a number of reasons:

* When you call them you don't know if anything happened or not or why
* It's not clear from the caller whether you should be calling them (similar, but different)
* JSConfig is bloated with lots of decisions, rather than concentrating on doing

This resulted in a few things:

* The basic launch view was ignorant of whether something actually is graded
* Worse, the basic launch view had a spurious notion of `grading_enabled` which was basically irrelevant / wrong
* A lack of clarity around what happens and why when reading the code

This work moves the real "is_gradable" value into the view, and enables the work we want to do to store that value, and have it line up.

### Review notes

* No behavior should have changed
* Each method is tackled in turn in different commits for easy reviewing

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -229,7 +229,7 @@ class JSConfig:
         self._config.setdefault("filePicker", {})
         self._config["filePicker"]["deepLinkingAPI"] = config
 
-    def enable_grading(self):
+    def enable_grading_bar(self):
         """Enable our LMS app's built-in assignment grading UI."""
 
         # Get one student dict for each student who has launched the assignment
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -231,29 +231,39 @@ class JSConfig:
         self._config.setdefault("filePicker", {})
         self._config["filePicker"]["deepLinkingAPI"] = config
 
-    def maybe_enable_grading(self):
-        """Enable our LMS app's built-in assignment grading UI, if appropriate."""
+    def enable_grading(self):
+        """Enable our LMS app's built-in assignment grading UI."""
 
-        if not self._lti_user.is_instructor:
-            # Only instructors can grade assignments.
-            return
+        # Get one student dict for each student who has launched the assignment
+        # and had grading info recorded for them.
+        students = []
 
-        if "lis_outcome_service_url" not in self._context.lti_params:
-            # Only "gradeable" assignments can be graded.
-            # Assignments that don't have the lis_outcome_service_url param
-            # aren't set as gradeable in the LMS.
-            return
+        grading_infos = self._grading_info_service.get_by_assignment(
+            application_instance=self._application_instance,
+            context_id=self._context.lti_params.get("context_id"),
+            resource_link_id=self._context.lti_params.get("resource_link_id"),
+        )
 
-        if self._context.is_canvas:
-            # Don't show our built-in grader in Canvas because it has its own
-            # "SpeedGrader" and we support that instead.
-            return
+        for grading_info in grading_infos:
+            h_user = HUser(
+                username=grading_info.h_username,
+                display_name=grading_info.h_display_name,
+            )
+            students.append(
+                {
+                    "userid": h_user.userid(self._authority),
+                    "displayName": h_user.display_name,
+                    "lmsId": grading_info.user_id,
+                    "LISResultSourcedId": grading_info.lis_result_sourcedid,
+                    "LISOutcomeServiceUrl": grading_info.lis_outcome_service_url,
+                }
+            )
 
         self._config["grading"] = {
             "enabled": True,
             "courseName": self._context.lti_params.get("context_title"),
             "assignmentName": self._context.lti_params.get("resource_link_title"),
-            "students": list(self._get_students()),
+            "students": students,
         }
 
     def maybe_set_focused_user(self):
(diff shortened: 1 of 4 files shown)


lms/views/lti/basic_launch.py
```diff
@@ -294,8 +282,35 @@ def _show_document(
             extra=assignment_extra,
         )
 
+        # An assignment has been configured in the LMS as "gradable" if it has
+        # the `lis_outcome_service_url` param
+        gradable = bool(self.context.lti_params.get("lis_outcome_service_url"))
```
jon-betts, line 287: It looks goofy leaving the `gradable` value out here, but we want to store this against the assignment.

The gradability of an assignment does not depend on whether we show you the interface or not.

lms/resources/_js_config/__init__.py
```diff
@@ -231,54 +231,43 @@ def add_deep_linking_api(self):
         self._config.setdefault("filePicker", {})
         self._config["filePicker"]["deepLinkingAPI"] = config
 
-    def maybe_enable_grading(self):
-        """Enable our LMS app's built-in assignment grading UI, if appropriate."""
+    def enable_grading(self):
```
marcospri, line 234: I'd rename this to "enable_grading_bar"

lms/views/lti/basic_launch.py
marcospri, line 287: I think this conditional could be:

`def is_gradable` in LTITResource

I can't point to the specific use case but if we split this view in multiple ones or we wanted to use this in deep-linking (which also uses LTIResoruce, it's also an LTI launch of a different type) we could re-use it.

It just feels like the right place IMO.

Also, not sure about naming but we might want to have both:

`is_assignment_gradable` (does the assignment it self support grading)
`is_user_gradable` (can the user himself have a grade)


I think having those two would help using 

 `_add_canvas_speedgrader_settings` here. 
 Currently that  happens all the time with ` self.context.js_config.add_document_url(document_url)` and  `_add_canvas_speedgrader_settings` has a very similar, duplicated conditional soup.

lms/views/lti/basic_launch.py
```diff
@@ -294,8 +282,48 @@ def _show_document(
...
+        # For students in Canvas with grades to submit we need to enable
+        # Speedgrader settings for gradable assignments
+        if (
+            assignment_gradable
+            and self.context.is_canvas
+            # `lis_result_sourcedid` associates a specific user with an
+            # assignment. This is evidence that a student is launching us
+            and self.context.lti_params.get("lis_result_sourcedid")
+        ):
+            self.context.js_config.add_canvas_speedgrader_settings(document_url)
```
jon-betts, line 315: It looks like there's some way of usefully combo-ing the if statements here, but I don't think there actually is. I _think_ it's best to just separate these out and have them both state what they need.

lms/views/lti/basic_launch.py
```diff
@@ -294,8 +282,48 @@ def _show_document(
...
+
+        # We add a `focused_user` query param to the SpeedGrader LTI launch
+        # URLs we submit to Canvas for each student when the student launches
+        # an assignment. Later, Canvas uses these URLs to launch us when a
+        # teacher grades the assignment in SpeedGrader.
+        if focused_user := self.request.params.get("focused_user"):
+            self.context.js_config.set_focused_user(focused_user)
```
jon-betts, line 322: TODO - This is only speedgrader, should this go above?

lms/views/lti/basic_launch.py
jon-betts: I've pulled the speed grader stuff in to, but I'd like to leave off moving things back out of the view again for is gradable until there's some clear demand for it.

lms/views/lti/basic_launch.py
marcospri: Yeah, it's tempting to shuffle these around trying to find conditional enlightenment but it's what's it is.

Nice that this happens here instead of in various `maybe_` methods far away.

lms/views/lti/basic_launch.py
marcospri: Yeah I think there's an implicit `is_canvas` here. For now I'd say it pretty clear from the coments.

tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -255,11 +249,81 @@ def test__show_document(self, svc, context, lti_h_service, assignment_service):
...
+    ):
+        svc._show_document(sentinel.document_url)  # pylint: disable=protected-access
+
+        context.js_config.add_canvas_speedgrader_settings.assert_not_called()
+
+    @pytest.mark.usefixtures("with_gradable_assignment", "is_canvas")
+    def test__show_document_no_speedgrader_without_grading_id(self, svc, context):
+        svc._show_document(sentinel.document_url)  # pylint: disable=protected-access
+
+        context.js_config.add_canvas_speedgrader_settings.assert_not_called()
```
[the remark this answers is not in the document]
marcospri: I had to do something similar in my grading branch. Hopefully they'd line up.

PR #4052, opened 2022-06-07
Product service and model

marcospri:
Adds a `.product` property to the request object.

This doesn't look like much of an improvement over the current situation but enables many other refactors.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -1,6 +1,5 @@
 import logging
 from datetime import datetime
-from enum import Enum
 from urllib.parse import urlparse
 
 import sqlalchemy as sa
@@ -15,16 +14,6 @@ LOG = logging.getLogger(__name__)
 
 class ApplicationInstance(BASE):
     """Class to represent a single lms install."""
-
-    class Product(str, Enum):
-        BLACKBOARD = "BlackboardLearn"
-        CANVAS = "canvas"
-        MOODLE = "moodle"
-        D2L = "desire2learn"
-        BLACKBAUD = "BlackbaudK12"
-        SCHOOLOGY = "schoology"
-        SAKAI = "sakai"
-        UNKNOWN = "unknown"
 
     __tablename__ = "application_instances"
     __table_args__ = (
@@ -204,15 +193,6 @@ class ApplicationInstance(BASE):
                 existing_guid=self.tool_consumer_instance_guid,
                 new_guid=tool_consumer_instance_guid,
             )
-
-    @property
-    def product(self):
-        try:
-            product = self.Product(self.tool_consumer_info_product_family_code)
-        except ValueError:
-            product = self.Product.UNKNOWN
-
-        return product
 
     @property
     def lti_version(self) -> str:
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -123,3 +123,18 @@ def _get_key(data: dict, data_path: List[str]):
         value = value[path_item]
 
     return value
+
+
+def _get_lti_params(request):
+    if not request.lti_jwt:
+        # If we don't have a LTI1.3 JWT there's no params to map.
+        # Use the request original params.
+        return request.params
+
+    return LTIParams.from_v13(request.lti_jwt)
+
+
+def includeme(config):
+    config.add_request_method(
+        _get_lti_params, name="lti_params", property=True, reify=True
+    )
--- a/lms/models/product.py
+++ b/lms/models/product.py
@@ -1,0 +1,49 @@
+from dataclasses import dataclass
+from enum import Enum
+
+
+@dataclass
+class Product:
+    class Family(str, Enum):
+        BLACKBAUD = "BlackbaudK12"
+        BLACKBOARD = "BlackboardLearn"
+        CANVAS = "canvas"
+        D2L = "desire2learn"
+        MOODLE = "moodle"
+        SAKAI = "sakai"
+        SCHOOLOGY = "schoology"
+        UNKNOWN = "unkown"
+
+        @classmethod
+        def _missing_(cls, _value):
+            return cls.UNKNOWN
+
+    family: Family
+
+    @classmethod
+    def from_request(cls, request):
+        return Product(family=cls._get_family(request))
+
+    @classmethod
+    def _get_family(cls, request):
+        # If we are in an API request, where we are forwarding the product type ourselves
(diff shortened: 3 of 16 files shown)


lms/models/application_instance.py
```diff
@@ -205,15 +194,6 @@ def check_guid_aligns(self, tool_consumer_instance_guid):
                 new_guid=tool_consumer_instance_guid,
             )
 
-    @property
-    def product(self):
```
marcospri, line 209: This property didn't correctly get "CANVAS" right in all circumstances.

lms/models/lti_params.py
```diff
@@ -123,3 +123,18 @@ def _get_key(data: dict, data_path: List[str]):
         value = value[path_item]
 
     return value
+
+
+def _get_lti_params(request):
```
marcospri, line 128: We want to use LTIParams to get the product across LTI versions.

Currently LTIParams is tied to context which makes it tricky to use outside a view.

lms/resources/lti_launch.py
```diff
@@ -80,17 +80,7 @@ def resource_link_id(self):
 
     @property
     def is_canvas(self):
-        """Return True if Canvas is the LMS that launched us."""
-        if (
-            self._request.parsed_params.get("tool_consumer_info_product_family_code")
-            == "canvas"
```
marcospri, line 86: Moved to the product service.

lms/services/product.py
```diff
@@ -0,0 +1,24 @@
+from lms.models import Product
+
+
+def _get_product(request):
+    if request.content_type == "application/json":
```
marcospri, line 5: In the API we don't need to guess, we could tell ourselves (ie send the data to the frontend and get it back when it calls us).

This will help having generic version of some api endpoints, /groups/ vs /canvas/groups /blackboard/groups

lms/services/product.py
```diff
@@ -0,0 +1,24 @@
...
+        # We are in an API request, where we are forwarding the product type ourselves
+        product_name = request.json["lms"]["product"]
+
+    else:
+        # In an LTI launch we'll use the parameters available to guess
+        product_name = request.lti_params.get("tool_consumer_info_product_family_code")
```
marcospri, line 11: Using the new `request.lti_params` here. 

We could build LTIParams here but I feel we'd be duplicating a simple but potentially easy to get wrong code.

lms/resources/_js_config/__init__.py
```diff
@@ -55,6 +55,7 @@ def add_document_url(self, document_url):
         if document_url.startswith("blackboard://"):
             self._config["api"]["viaUrl"] = {
                 "authUrl": self._request.route_url("blackboard_api.oauth.authorize"),
+                "product": self._request.product.value,
```
marcospri, line 58: Send the product to the frontend so we'll receive it back and we avoid the need to guess the product without the context of an LTI launch.

lms/services/product.py
```diff
@@ -0,0 +1,24 @@
...
+
+    try:
+        return Product(product_name)
+    except ValueError:
+        # If that fails we can check to detect canvas specific parameter
+        if "custom_canvas_course_id" in request.lti_params:
+            return Product.CANVAS
```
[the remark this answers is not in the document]
marcospri: We should always know what product we are in but raising when we don't could be a potentially big change.

I don't think the difference between we don't know and we didn't get the info is real in practice, if we didn't get it from LTI we should use custom variables or some other heuristic.

I changed the logic a bit.

jon-betts (commented):
I really like where this is going, but I think we can tweak the UI and implementation to make it a bit snappier. Here are some ideas I had:

 * You can move the default value behavior into the Enum makes factoring a bit easier
 * If we move the Enum around a bit and refer to it we can have:
     * It still be called `ProductFamily` and string-ify that way
     * Be referred to in the code as `Product.Family` instead of `Product.ProductFamily`
     
I wonder if it would be better to be explicit and have the front end value for product family be called `productFamily` instead of just `product` as well? I think this is a hold over from the previous solution?

tests/unit/lms/models/product_test.py
```diff
@@ -0,0 +1,54 @@
...
+    ("moodle", Product.ProductFamily.MOODLE),
+    ("schoology", Product.ProductFamily.SCHOOLOGY),
+    ("sakai", Product.ProductFamily.SAKAI),
+    ("wut", Product.ProductFamily.UNKNOWN),
+    ("", Product.ProductFamily.UNKNOWN),
+    (None, Product.ProductFamily.UNKNOWN),
+]
+
+
+class TestFromRequest:
```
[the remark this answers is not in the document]
marcospri: Yeah, makes sense.

lms/resources/_js_config/__init__.py
```diff
@@ -64,6 +65,7 @@ def add_document_url(self, document_url):
         elif document_url.startswith("canvas://"):
             self._config["api"]["viaUrl"] = {
                 "authUrl": self._request.route_url("canvas_api.oauth.authorize"),
+                "product": self._request.product.family,
```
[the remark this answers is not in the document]
marcospri: I think it could be:

`product: {family: XXX}`


but maybe overkill as of now.

lms/models/product.py
```diff
@@ -0,0 +1,46 @@
...
+        BLACKBAUD = "BlackbaudK12"
+        BLACKBOARD = "BlackboardLearn"
+        CANVAS = "canvas"
+        D2L = "desire2learn"
+        MOODLE = "moodle"
+        SAKAI = "sakai"
+        SCHOOLOGY = "schoology"
+        UNKNOWN = "unkown"
+
+    family: ProductFamily
```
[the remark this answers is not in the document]
marcospri: I just renamed ProductFamily and keep it nested in the product. I think that's what's we are after.

lms/models/product.py
```diff
@@ -0,0 +1,46 @@
...
+    class ProductFamily(str, Enum):
+        BLACKBAUD = "BlackbaudK12"
+        BLACKBOARD = "BlackboardLearn"
+        CANVAS = "canvas"
+        D2L = "desire2learn"
+        MOODLE = "moodle"
+        SAKAI = "sakai"
+        SCHOOLOGY = "schoology"
+        UNKNOWN = "unkown"
+
```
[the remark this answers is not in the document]
marcospri: Yeah this simplifies things :+1:

Issue #3993, opened 2022-05-26
Associate users with assignments


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122

Currently we associate users with their groupings, and groupings with assignments, but it doesn't necessarily follow that you can get from one to the other.

Every user in a course will be in the course grouping, but might not be given every assignment for example.

Tracking users separately will allow us to more accurately answer questions about which teachers and students are in a given assignment.

## Tasks

 * https://github.com/hypothesis/lms/pull/4063
 * https://github.com/hypothesis/lms/pull/4064
 * https://github.com/hypothesis/lms/pull/4062

## Design

The idea here is we want:

 * A nice small table for the relationship
 * To capture all the different types of role
 * Map those roles to something we understand
 * Reserve the right to remap those roles at a later date if we change our mind

To support this we are going for a fully normalized table structure where roles are stored in their own table with one row per unique value we see along with our interpretation. That way we can re-grade the table at a later date if we change our mind.

All the enums are being kept on the Python side to make for easy migrations.

### Role table (`lti_role`)

| Field | Type | Reason |
|--------|--------|------------|
| `id` | int4 | Primary key | 
| `value` | text |The raw string value as provided by the LMS
| `type` | text | Our intepretation of the value

For the type we want:

 * "learner" - For reports
 * "teacher" - For reports
 * "admin" - For future functionality

This allows us to keep all the details we ever see about roles and to be able to reinterpret them in the future if we so chose to. We don't bloat the relation table with long raw text strings, but we capture all the detail for later.

### Assignment to user table (`assignment_membership`)

| Field | Type | Reason |
|--------|--------|------------|
| `assignment_id` | int4 | FK to assignment
| `user_id` | int4 | FK to `lti_users`
| `role_id` | int4 | FK to `lti_roles`. Answer teacher / student based questions
| `created` | timestamp | Facilitate per time period questions
| `updated` | timestamp | Facilitate per time period questions

Unique across:

 * PK - `assignment_id, user_id, user_role`

See also:

 * https://www.postgresql.org/docs/9.1/datatype-enum.html

jon-betts:
Done

PR #4062, opened 2022-06-08
Track what roles users have within assignments

jon-betts:
For:

Needs:

 * https://github.com/hypothesis/lms/pull/4063

## Testing notes

 * Nuke your DB
 * Start everything in the world and the visit a lot of assignments as different users
 * Observe `lti_role` and `assignment_membership` filling up

### DIFF
--- a/lms/models/lti_role.py
+++ b/lms/models/lti_role.py
@@ -1,0 +1,117 @@
+from enum import Enum, unique
+
+
+@unique
+class RoleType(str, Enum):
+    """Enum for the different types of role a user can have."""
+
+    INSTRUCTOR = "instructor"
+    LEARNER = "learner"
+    ADMIN = "admin"
+
+    @classmethod
+    def parse_lti_role(cls, role):
+        """Parse an LTI role string into one of our role types."""
+
+        # We have to do this work around because Enums can't have private
+        # attributes. They are just interpreted as extra values for the enum.
+        return _RoleParser.parse_role(role)
+
+
+class _RoleParser:
+    """Close collaborator class for parsing roles."""
+
+    @classmethod
+    def parse_role(cls, role) -> RoleType:
+        return (
+            cls._parse_v13_role(role) or cls._parse_v11_role(role) or RoleType.LEARNER
+        )
+
+    _V11_INSTRUCTOR_STRINGS = (
+        "instructor",
+        "teachingassistant",
+        "contentdeveloper",
+        "faculty",
+        "mentor",
+        "staff",
+    )
+
+    @classmethod
+    def _parse_v11_role(cls, role):
+        role = role.lower()
+
+        for string in cls._V11_INSTRUCTOR_STRINGS:
+            if string in role:
+                return RoleType.INSTRUCTOR
+
+        if "learner" in role:
+            return RoleType.LEARNER
+
+        if "admin" in role:
+            return RoleType.ADMIN
+
+        return None
+
+    _V13_PREFIX = "http://purl.imsglobal.org/vocab/lis/v2"
+    _V13_ROLE_MAPPINGS = {
+        # https://www.imsglobal.org/spec/lti/v1p3/#lis-vocabulary-for-context-roles
+        # A.2.1 LIS vocabulary for system roles
+        # Core system roles
+        f"{_V13_PREFIX}/system/person#Administrator": RoleType.ADMIN,
+        f"{_V13_PREFIX}/system/person#None": RoleType.LEARNER,
+        # Non‑core system roles
+        f"{_V13_PREFIX}/system/person#AccountAdmin": RoleType.ADMIN,
+        f"{_V13_PREFIX}/system/person#Creator": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/system/person#SysAdmin": RoleType.ADMIN,
+        f"{_V13_PREFIX}/system/person#SysSupport": RoleType.ADMIN,
+        f"{_V13_PREFIX}/system/person#User": RoleType.LEARNER,
+        # A.2.2 LIS vocabulary for institution roles
+        # Core institution roles
+        f"{_V13_PREFIX}/institution/person#Administrator": RoleType.ADMIN,
+        f"{_V13_PREFIX}/institution/person#Faculty": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/institution/person#Guest": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#None": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#Other": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#Staff": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/institution/person#Student": RoleType.LEARNER,
+        # Non‑core institution roles
+        f"{_V13_PREFIX}/institution/person#Alumni": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#Instructor": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/institution/person#Learner": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#Member": RoleType.LEARNER,
+        f"{_V13_PREFIX}/institution/person#Mentor": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/institution/person#Observer": RoleType.LEARNER,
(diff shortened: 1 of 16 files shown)


lms/models/lti_role.py
```diff
@@ -0,0 +1,145 @@
...
+    # Export for others convenience
+    Type = RoleType
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
+
+    _value = sa.Column("value", sa.UnicodeText(), nullable=False, unique=True)
+    """The raw string from LTI params."""
+
+    type = sa.Column("type", sa.Enum(RoleType), nullable=False, unique=False)
+    """Our interpretation of the value."""
```
jon-betts, line 37: This totally works, but it causes the type of the column in SQL to actually be "roletype" which suggests SQLAlchemy is creating some custom thing in Postgres rather than storing a string as per the Stackoverflow answer I cribbed this from...

lms/models/lti_role.py
```diff
@@ -0,0 +1,147 @@
...
+            cls._parse_v13_role(role) or cls._parse_v11_role(role) or RoleType.LEARNER
+        )
+
+    _V11_INSTRUCTOR_STRINGS = (
+        "instructor",
+        "teachingassistant",
+        "contentdeveloper",
+        "faculty",
+        "mentor",
+        "staff",
```
jon-betts, line 66: This is **not** the same list as we use in other places. We count more things here to be instructors.

This probably needs to be vetted by some outside folk?

In any case currently this does not interact with the existing role system at all and shouldn't cause any trouble. We may want to unify them in the future though.

lms/models/lti_role.py
```diff
@@ -0,0 +1,147 @@
...
+        f"{_V13_PREFIX}/membership/ContentDeveloper": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/membership/Instructor": RoleType.INSTRUCTOR,
+        # Look out for this weirdo!
+        f"{_V13_PREFIX}/membership/Learner#Instructor": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/membership/Learner": RoleType.LEARNER,
+        f"{_V13_PREFIX}/membership/Mentor": RoleType.INSTRUCTOR,
+        # Non‑core context roles
+        f"{_V13_PREFIX}/membership/Manager": RoleType.INSTRUCTOR,
+        f"{_V13_PREFIX}/membership/Member": RoleType.LEARNER,
+        f"{_V13_PREFIX}/membership/Office": RoleType.INSTRUCTOR,
```
jon-betts, line 127: You could probably condense this down into something which happened to pass the tests etc, but this is very explicit, and very easy to change.

lms/services/lti_role_service.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class LTIRoleService:
+    """A service for dealing with LTIRole objects."""
+
+    def __init__(self, db_session: Session):
+        self._db = db_session
+
+    def get_roles(self, role_strings: List[str]) -> List[LTIRole]:
+        """Get a list of role objects for the provided strings."""
```
jon-betts, line 15: This could accept an LTIUser as well, but it seems overly coupled to me. It feels right in the assignment service though.

tests/factories/lti_role.py
```diff
@@ -0,0 +1,28 @@
...
+            "http://purl.imsglobal.org/vocab/lis/v2/membership/Learner",
+            "http://purl.imsglobal.org/vocab/lis/v2/membership/Mentor",
+            "http://purl.imsglobal.org/vocab/lis/v2/membership/Manager",
+            "http://purl.imsglobal.org/vocab/lis/v2/membership/Office",
+        ],
+    )
+
+    sub_role = Faker("hexify", text="SubRole-^^^^^^^")
+
+    value = factory.LazyAttribute(lambda o: f"{o.primary_role}#{o.sub_role}")
```
jon-betts, line 28: We need to be careful we don't accidentally generate two role strings the same, or it will break the uniqueness requirement. This is _unlikely_ to happen here, but is still possible.

Should this be an incrementing value instead?

tests/unit/lms/models/lti_role_test.py
```diff
@@ -0,0 +1,102 @@
...
+            ("urn:lti:role:ims/lis/ContentDeveloper", RoleType.INSTRUCTOR),
+            ("urn:lti:role:ims/lis/Instructor", RoleType.INSTRUCTOR),
+            ("urn:lti:role:ims/lis/Learner", RoleType.LEARNER),
+            ("urn:lti:role:ims/lis/Mentor", RoleType.INSTRUCTOR),
+            ("urn:lti:role:ims/lis/TeachingAssistant", RoleType.INSTRUCTOR),
+            ("urn:lti:role:ims/lis/TeachingAssistant/Grader", RoleType.INSTRUCTOR),
+            ("urn:lti:sysrole:ims/lis/Administrator", RoleType.ADMIN),
+            ("urn:lti:sysrole:ims/lis/None", RoleType.LEARNER),
+            ("urn:lti:sysrole:ims/lis/SysAdmin", RoleType.ADMIN),
+        ),
```
jon-betts, line 50: This is very extra, but I did this in a TDD way and worked backwards from what I though should be true

lms/services/lti_role_service.py
```diff
@@ -0,0 +1,38 @@
...
+from lms.models import LTIRole
+
+
+class LTIRoleService:
+    """A service for dealing with LTIRole objects."""
+
+    def __init__(self, db_session: Session):
+        self._db = db_session
+
+    def get_roles(self, role_string) -> List[LTIRole]:
```
marcospri, line 14: `role_string` / `roles :str` or `lti_roles:str`? 

I'll probably would have written that but not a strong preference

lms/views/lti/basic_launch.py
```diff
@@ -273,14 +273,17 @@ def _show_document(self, document_url, assignment_extra=None):
...
             document_url=document_url,
             tool_consumer_instance_guid=self.context.lti_params[
                 "tool_consumer_instance_guid"
             ],
             resource_link_id=self.context.resource_link_id,
             extra=assignment_extra,
         )
+        self.assignment_service.upsert_assignment_membership(
+            assignment=assignment, lti_user=self.request.lti_user
+        )
```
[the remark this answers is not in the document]
marcospri: I think makes sense as two separate functions. We only use it here in any case so it's not something we have to remember to do all over the place.

lms/services/assignment.py
```diff
@@ -62,6 +71,40 @@ def upsert_assignment(
 
         return assignment
 
+    def upsert_assignment_membership(
+        self, assignment: Assignment, lti_user: LTIUser
```
marcospri, line 75: If we take `user` here we could do `user.roles` now but as we remove that we should take `roles` separaptly.

I think that would make sense, the user is not tied to a role so it makes sense it's two separate concepts arguments.

lms/services/lti_role_service.py
jon-betts: Yeah I'll have a think. Just trying to differentiate it from the `LTIRole` object, which it absolutely isn't, and to imply that it's singular, not a list.

The type annotation does this of course, but only here, rather than in the caller.

lms/services/assignment.py
jon-betts: Eugh this is a bit of a mess. Presumably `lti_user` just gets roles from the params anyway, so it might be clearer to totally ignore it.

We don't want to get mixed up in `user.roles` if we are going to get rid of them, so I think separate args might be the way.

lms/services/lti_role_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+        :param role_description: A comma delimited set of role strings
+        """
+        role_strings = [role.strip() for role in role_description.split(",")]
+
+        # pylint: disable=no-member
+        # Pylint is confused about the `in_` for some reason
+        roles = self._db.query(LTIRole).filter(LTIRole.value.in_(role_strings)).all()
+
+        if missing := (set(role_strings) - set(role.value for role in roles)):
```
marcospri, line 26: I wonder if an upsert would be faster/shorter

This is not a problem in any case.

marcospri (approved):
Nice. 

I think having this data in the DB, apart from any value it has for analytics, BI... might be useful to understand the roles our users have in the system. 

We'll make any further changes with more data to back them up.

lms/services/lti_role_service.py
jon-betts: Yes it could be. I was thinking that this will end up being basically read only, and this gets it to only one query (in that case), but an upsert would probably be the same regardless...

I suspect this would be faster than an upsert, but I didn't actually bench it.

PR #4063, opened 2022-06-08
Change `store_lti_user` to be `upsert_user` and to return the stored user

jon-betts:
For:

 * https://github.com/hypothesis/lms/pull/4062

This allows us to use the returned value


marcospri (approved):
Yeah, makes sense.

PR #4064, opened 2022-06-08
DB migration for LTI roles and assignment membership tables

jon-betts:
For:

Migration for:

* https://github.com/hypothesis/lms/pull/4062

### DIFF
--- a/lms/migrations/versions/fe80746cf145_add_lti_roles_and_assignment_membership.py
+++ b/lms/migrations/versions/fe80746cf145_add_lti_roles_and_assignment_membership.py
@@ -1,0 +1,78 @@
+"""
+Add lti_roles and assignment_membership.
+
+Revision ID: fe80746cf145
+Revises: 83ec5ef85d44
+Create Date: 2022-06-08 20:37:02.158624
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "fe80746cf145"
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/fe80746cf145_add_lti_roles_and_assignment_membership.py
```diff
@@ -0,0 +1,78 @@
...
+        sa.Column("value", sa.UnicodeText(), nullable=False),
+        sa.Column(
+            "type",
+            sa.Enum(
+                "instructor",
+                "learner",
+                "admin",
+                name="roletype",
+                native_enum=False,
+                length=64,
```
jon-betts, line 30: As we have `native_enum=False` I don't think the values here matter at all

marcospri: From the docs  https://docs.sqlalchemy.org/en/14/core/type_basics.html

> When False, uses VARCHAR + check constraint for all backends. 

I haven't seen the check constraint when using alembic or SQLA's create_all but in any case we mentioned before, having this being checked in python only seems enough so it's probably for the better.

jon-betts: I think I separately disabled the checking as well to try and get this. It took a few runs at before it didn't add anything as a type etc.

PR #4065, opened 2022-06-09
[WIP]Sync api refactor

marcospri:


lms/views/api/sync.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self._sync_to_h(groupings)
+
+        authority = self.request.registry.settings["h_authority"]
+        return [group.groupid(authority) for group in groupings]
+
+    def _sync_to_h(self, groups):
+        lti_h_svc = self.request.find_service(name="lti_h")
+        group_info = self.request.json["group_info"]
+        lti_h_svc.sync(groups, group_info)
```
marcospri, line 72: This is probably duplicated in basic_lti_launch

lms/resources/_js_config/__init__.py
```diff
@@ -431,75 +431,49 @@ def _groups(self):
...
+        if self._context.grouping_type == Grouping.Type.COURSE:
+            return None
 
-    def _blackboard_sync_api(self):
         req = self._request
         return {
-            "authUrl": req.route_url("blackboard_api.oauth.authorize"),
-            "product": self._request.product.family,
-            "path": req.route_path("blackboard_api.sync"),
+            "authUrl": self._lms_auth_url,
```
marcospri, line 451: Common config for both

lms/services/grouping.py
```diff
@@ -8,6 +9,13 @@
 from lms.services.upsert import bulk_upsert
 
 
+class GroupingInfo(TypedDict):
```
marcospri, line 12: If you are  thinking what's up with this guy and "TypeDict"...


It could be dataclass but then it will need to be imported and we'll go from dict -> dataclass -> dict. With this we only go from dict with one set of keys to another dict.

In any case is for documentation mostly, pyright also complains if I pass the wrong key (or not all of them).

lms/services/product/grouping/interface.py
```diff
@@ -0,0 +1,25 @@
...
+        self._lti_user = lti_user
+        self._grouping_service = grouping_service
+
+    def get_groups(self, _course, _group_set_id, _grading_student_id) -> List[Grouping]:
+        raise NotImplementedError
+
+    def get_sections(self, course, grading_student_id) -> List[Grouping]:
+        raise NotImplementedError
+
+    def _to_groupings(
```
marcospri, line 18: This could be called `store_..` and maybe moved to the main GroupingService.

lms/views/api/sync.py
```diff
@@ -0,0 +1,72 @@
...
+        request_method="POST",
+        renderer="json",
+        permission=Permissions.API,
+        schema=APISyncSchema,
+    )
+    def sync(self):
+        course_id = self.request.parsed_params["course"]["context_id"]
+        course = self.request.find_service(name="course").get_by_context_id(course_id)
+
+        if group_set_id := self.request.parsed_params("group_set_id"):
```
marcospri, line 57: We could send us the "Grouping.Type" and decide bused on that.

Issue #3992, opened 2022-05-26
Associate assignments with groupings


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122

We currently don't know which courses and sections etc. that assignments are associated with. This makes it very difficult to present any useful information about them. We should associate assignments with groupings, and keep this up to date.

### Assignment to grouping table (`assignment_groupings`?)

| Field | Type | Reason |
|--------|--------|------------|
| `assignment_id` | int4 | FK to assignment
| `grouping_id` | int4 | FK to grouping
| `created` | timestamp | Facilitate per time period questions
| `updated` | timestamp | Facilitate per time period questions

Unique across:

 * PK - `assignment_id, grouping_id`

jon-betts:
Done

PR #4066, opened 2022-06-09
Track assignment course relationships

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/3992

Requires:

 * https://github.com/hypothesis/lms/pull/4067

We want to know all the other relations between an assignment and sections and groups etc, but it's more complicated as these values aren't available at launch.

## Testing notes

 * Nuke your DB
 * Start everything in the world and the visit a lot of assignments as different users
 * Observe `assignment_grouping` filling up


marcospri (approved):
Looks good, another win of the refactor that got us here.

PR #4068, opened 2022-06-10
Refactor lti_launch group properties

marcospri:
Use request.product in the methods themselves to provide a more
concise API instead of having methods per LMS.

Again, not a massive improvement but having the same "API" in the LTIResource class allow for further refactors.

- canvas_sections_supported -> sections_enabled

It doesn't have any practical implications but it makes for a cleaner API

-  canvas_sections_supported inlined in sections_enabled

We only used the supported concept in `sections_enabled`. It could make sense as a separate concept if it was used elsewhere.

- canvas_groups_enabled and blackboard_groups_enabled moved/inlined in the filepicker.

### DIFF
--- a/lms/models/grouping.py
+++ b/lms/models/grouping.py
@@ -18,6 +18,11 @@ class Grouping(CreatedUpdatedMixin, BASE):
         CANVAS_GROUP = "canvas_group"
         BLACKBOARD_GROUP = "blackboard_group"
 
+        # These are the LMS agnostic versions of the ones avobe.
+        # They don't get stored in the DB but are meaningful in the codebase
+        SECTION = "section"
+        GROUP = "group"
+
     __tablename__ = "grouping"
     __mapper_args__ = {"polymorphic_on": "type"}
     __table_args__ = (
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -2,7 +2,7 @@ import functools
 from enum import Enum
 from typing import List, Optional
 
-from lms.models import GroupInfo, HUser, Product
+from lms.models import GroupInfo, Grouping, HUser, Product
 from lms.resources._js_config.file_picker_config import FilePickerConfig
 from lms.services import HAPIError, JSTORService
 from lms.validation.authentication import BearerTokenSchema
@@ -424,9 +424,12 @@ class JSConfig:
         }
 
     def _groups(self):
-        if self._context.canvas_sections_enabled or self._context.is_group_launch:
-            return "$rpc:requestGroups"
-        return [self._context.course.groupid(self._authority)]
+        if self._context.grouping_type == Grouping.Type.COURSE:
+            return [self._context.course.groupid(self._authority)]
+
+        # If not using the default COURSE grouping point the FE
+        # to the sync API to dynamically get the relevant groupings.
+        return "$rpc:requestGroups"
 
     def _canvas_sync_api(self):
         req = self._request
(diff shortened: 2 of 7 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -491,14 +491,13 @@ def _blackboard_sync_api(self):
 
     def _sync_api(self):
```
jon-betts, line 492: Maybe `_get_sync_api` or something? This sounds like a verb not a noun

jon-betts: Looks like we might want a:

```python
if not self._context.is_group_launch:
    return None
```

At the top here?

It would also be nice if this was before the functions it calls as the way you read through the file is kind of 1, 3, 2 at the moment.

jon-betts (commented):
I think this is showing us that multiple concepts we had before (`blackboard_x`, canvas_x`) are really `product + x` rather than their own concept.

I wonder if the true zen moment here is to realise that in the same way that groups are all one mode, that sections are too? Just a different named kind?

Can you have Canvas groups and sections on at the same time, or is this a single parameterized feature?

lms/resources/_js_config/file_picker_config.py
```diff
@@ -8,7 +8,9 @@ class FilePickerConfig:
     def blackboard_config(cls, context, request, application_instance):
         """Get Blackboard files config."""
         files_enabled = application_instance.settings.get("blackboard", "files_enabled")
-        groups_enabled = context.blackboard_groups_enabled
+        groups_enabled = application_instance.settings.get(
```
marcospri, line 11: These replace canvas_groups_enabled and blackboard_groups_enabled.

They are only relevant in the filepicker.

lms/resources/lti_launch.py
```diff
@@ -128,49 +128,41 @@ def canvas_sections_enabled(self):
...
+                tool_consumer_instance_guid, self.resource_link_id
+            )
+            return assignment.extra.get("group_set_id") if assignment else None
+
+        return None
 
     @property
     def is_group_launch(self):
-        return self.canvas_is_group_launch or self.is_blackboard_group_launch
+        return bool(self.group_set_id)
```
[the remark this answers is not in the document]
marcospri: Removed this in favor of `grouping_type == GROUP`

lms/models/grouping.py
```diff
@@ -18,6 +18,11 @@ class Type(str, Enum):
         CANVAS_GROUP = "canvas_group"
         BLACKBOARD_GROUP = "blackboard_group"
 
+        # These are the LMS agnostic versions of the ones avobe.
```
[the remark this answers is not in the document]
jon-betts: A dangeous game your planin' here son. Mighty dangerous.

Yeah, this seems like a way to start introducing this without jumping all the way in one go. Might want to group "COURSE" down here with the LMS agnostic versions.

lms/resources/_js_config/__init__.py
```diff
@@ -490,16 +493,13 @@ def _blackboard_sync_api(self):
...
+        if self._request.product.family == Product.Family.CANVAS:
             return self._canvas_sync_api()
 
-        if (
-            self._request.product.family == Product.Family.BLACKBOARD
-            and self._context.is_blackboard_group_launch
-        ):
+        if self._request.product.family == Product.Family.BLACKBOARD:
 
             return self._blackboard_sync_api()
```
jon-betts, line 504:
```suggestion
        if self._request.product.family == Product.Family.BLACKBOARD:
            return self._blackboard_sync_api()
```

Gah... my OCD

Issue #4061, opened 2022-06-08
Make JSTOR picker thumbnail show first rather than last page of article


robertknight:
The JSTOR picker currently appears to use the last rather than first page of the article as the thumbnail. See https://hypothes-is.slack.com/archives/C02T75RKBTK/p1654704176844539?thread_ts=1654704066.828039&cid=C02T75RKBTK. Adding an `offset=1` query param appears to resolve the issue, but we need to confirm that this is the right fix.

PR #4076, opened 2022-06-13
Make JSTOR article thumbnail show first page instead of last page.

robertknight:
Fix an issue where the JSTOR article picker would show a thumbnail of the last page of the article instead of the first page.

The resolution of the image has also been increased to make it sharper. The size of the UI in the thumbnail is still too small to really make it possible to check if it is the correct thumbnail, but fixing this requires layout changes in a separate PR.

Fixes https://github.com/hypothesis/lms/issues/4061

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from typing import Optional
 from urllib.parse import quote
 
 import requests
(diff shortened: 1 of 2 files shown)


lms/services/jstor.py
```diff
@@ -97,7 +98,19 @@ def thumbnail(self, article_id: str):
         :raise ExternalRequestError: If the response doesn't look like a valid `data:` URI
         """
         doi = self._doi_from_article_id(article_id)
-        data_uri = self._api_request(format_uri("/thumbnail/{}", doi)).text
+        params = {
+            # `offset` specifies the page number. The default value of 0 returns
```
marcospri, line 102: This works here and it's well documented but I wonder if something like "width" would make more sense to be forwarded from the frontend?

Os is there a concern it could be manipulated and request huge images?

marcospri (approved):

lms/services/jstor.py
robertknight: > This works here and it's well documented but I wonder if something like "width" would make more sense to be forwarded from the frontend?

Yes, logically it would make sense for the frontend to specify the image size it wants. It has been avoided so far just to reduce the amount of plumbing involved.

> Os is there a concern it could be manipulated and request huge images?

The LMS API is currently only available to our frontend, so API abuse isn't a major concern, but it probably would be a good idea to put min/max constraints on the width value on the backend.

PR #4078, opened 2022-06-13
Fix handling of JSTOR metadata/thumbnail requests that include a DOI prefix

robertknight:
Metadata/thumbnail requests for JSTOR article IDs with a DOI prefix (eg. 10.5325/jafrideve.18.2.0019) did not match the expected route. It seems this is because the path is URL-decoded before being matched against the route's path pattern [1], and the default regex for path parameters is `[^/]+`. Fix this by specifying a custom pattern for the path parameter.

**Testing:**

1. Configure an assignment and choose JSTOR as the content type
2. Enter https://www.jstor.org/stable/10.5325/jafrideve.18.2.0019 as the article link.
3. Press Enter and the metadata/thumbnail should be fetched correctly

----

[1] Per https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#route-pattern-syntax: "Literal strings in the path segment should represent the decoded value of the PATH_INFO provided to Pyramid"

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -98,9 +98,16 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route("lti_api.result.read", "/api/lti/result", request_method="GET")
     config.add_route("lti_api.result.record", "/api/lti/result", request_method="POST")
 
-    config.add_route("jstor_api.articles.metadata", "/api/jstor/articles/{article_id}")
+    # JSTOR article IDs need a custom pattern because they may contain a slash,
+    # after URL-decoding of the path.
+    jstor_article_id_pat = r"(10\.[0-9]+/)?[^/]+"
     config.add_route(
-        "jstor_api.articles.thumbnail", "/api/jstor/articles/{article_id}/thumbnail"
+        "jstor_api.articles.metadata",
+        f"/api/jstor/articles/{{article_id:{jstor_article_id_pat}}}",
+    )
+    config.add_route(
+        "jstor_api.articles.thumbnail",
+        f"/api/jstor/articles/{{article_id:{jstor_article_id_pat}}}/thumbnail",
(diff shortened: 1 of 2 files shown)


lms/routes.py
```diff
@@ -98,9 +98,16 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route("lti_api.result.read", "/api/lti/result", request_method="GET")
     config.add_route("lti_api.result.record", "/api/lti/result", request_method="POST")
 
-    config.add_route("jstor_api.articles.metadata", "/api/jstor/articles/{article_id}")
+    # JSTOR article IDs need a custom pattern because they may contain a slash,
+    # after URL-decoding of the path.
+    jstor_article_id_pat = r"(10\.[0-9]+/)?[^/]+"
     config.add_route(
-        "jstor_api.articles.thumbnail", "/api/jstor/articles/{article_id}/thumbnail"
+        "jstor_api.articles.metadata",
```
marcospri, line 105: We might also get something similar using "remainder match" from https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#route-pattern-syntax

But the regex is more explicit :+1:

marcospri (approved):
Looks good. In hindsight using *not* path parameters could have simplify things.

We can always go that direction if things get out of control here.

lms/routes.py
robertknight: Ah, but a remainder match must come at the end of the path according to those docs. Here it is the "root" endpoint for a JSTOR article and then there are endpoints under that for different kinds of information. One way around that would be to re-organize the path so that the article ID does come at the end, which is how JSTOR's own API is organized. That would be simpler,  but then also different from our typical endpoint structure for other resources.

PR #4079, opened 2022-06-13
Update handling of slashes in JSTOR API endpoint paths

robertknight:
The JSTOR API endpoints are being moved to new infrastructure ([Related Slack thread](https://hypothes-is.slack.com/archives/C02T75RKBTK/p1655126049981469)). The old endpoints accepted DOIs with or without percent-encoding of the slash between the DOI prefix and suffix. The new endpoints require that the slash is not percent-encoded.

**Testing:**

Update the `JSTOR_API_URL` env var in your `.devdata.env` file to:

```
export JSTOR_API_URL="https://labs.jstor.org/api/anno"
```

And check that the metadata/thumbnail still load correctly when configuring a JSTOR assignment (eg. using this article link: https://www.jstor.org/stable/26158829)

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -153,7 +153,7 @@ def format_uri(template: str, *params: str):
     This is like `template.format(*params)` except that the params are
     percent-encoded.
     """
-    encoded = [quote(param, safe="") for param in params]
+    encoded = [quote(param, safe="/") for param in params]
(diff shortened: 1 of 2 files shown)


marcospri (approved):
Works as expected :+1: 

Maybe do a quick PR on devdata to update the URL once everything it's merged.

PR #4090, opened 2022-06-15
Remove `canvas_api_custom_domain` from the launch context

jon-betts:
This reverts commit fb2a12046ce9be63b963a1e2c98312ee3a6fb52e.

### DIFF
--- a/lms/resources/_js_config/file_picker_config.py
+++ b/lms/resources/_js_config/file_picker_config.py
@@ -73,16 +73,15 @@ class FilePickerConfig:
     def google_files_config(cls, context, request, application_instance):
         """Get Google file picker config."""
 
-        # Get the URL of the top-most page that the LMS app is running in.
-        #
-        # The frontend has to pass this to Google Picker, otherwise Google
-        # Picker refuses to launch in an iframe.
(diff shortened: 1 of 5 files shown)


marcospri (approved):
Nice find, we don't need a method that doesn't do anything extra to access one parameter. 

I wonder if we could remove this also from one of the schemas:


```diff
diff --git a/lms/validation/_lti_launch_params.py b/lms/validation/_lti_launch_params.py
index c1fe3871..3d2e21e2 100644
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -61,7 +61,6 @@ class _CommonLTILaunchSchema(LTIV11CoreSchema):
     lti_version = fields.Str(validate=OneOf(["LTI-1p0", "1.3.0"]), required=True)
     oauth_consumer_key = fields.Str(required=False)
 
-    custom_canvas_api_domain = fields.Str()
     custom_canvas_course_id = fields.Str()
     launch_presentation_return_url = fields.Str()
     tool_consumer_info_product_family_code = fields.Str()
```

we don't validate anything about it and I think having it there is what made it available in `parsed_params` (vs request.params).

Issue #3968, opened 2022-05-18
Celery task to rotate private keys


marcospri:
Currently the data on the `RSAKey` table doesn't have any logic to add/delete/rotate keys.

- Define a policy for key rotation
- Implement the policy on a periodic celery task.

PR #4092, opened 2022-06-16
Add RSAKeyService.rotate

marcospri:
For #3968 

Not yet executed in production as is meant to be called from a Celery task. 

## Testing

- Truncate rsa_key and `make devdata` before starting.

- `tox -qe dockercompose -- exec postgres psql -U postgres -c "select created, kid, expired from  rsa_key;"`

```
         created          |               kid                | expired 
---------------------------+----------------------------------+---------
 2022-06-21 07:41:31.52763 | f1a394d0268e4fd8b45391ce4c1ee901 | f
(1 row)
```

(this one comes from devdata)

- Rotate the keys, on `make shell`

```
from datetime import timedelta
from lms.services import RSAKeyService
rsa_service = request.find_service(RSAKeyService)
rsa_service.rotate(5, timedelta(days=1), timedelta(days=1))
tm.commit()
```

- `tox -qe dockercompose -- exec postgres psql -U postgres -c "select kid, expired from  rsa_key;"`

```          created           |               kid                | expired 
 2022-06-24 15:29:51.547215 | f1a394d0268e4fd8b45391ce4c1ee901 | f
 2022-06-24 15:30:33.958874 | 53f1be3f52b44672b377b87c01ef1124 | f
 2022-06-24 15:30:33.958874 | 0ccd632409554b328b313aabf5d3b9a9 | f
 2022-06-24 15:30:33.958874 | 9859b8e1ce8b4457b7f74d5f9dda98be | f
 2022-06-24 15:30:33.958874 | f8081ab08fa84bf1ba9f9373f22d92d4 | f
(5 rows)
```

Nothing got expired, but 4 new keys were created to match the target

- Rotate again

```
from datetime import timedelta
from lms.services import RSAKeyService
rsa_service = request.find_service(RSAKeyService)
rsa_service.rotate(5, timedelta(days=1), timedelta(days=1))
tm.commit()
```

```
          created           |               kid                | expired 
----------------------------+----------------------------------+---------
 2022-06-24 15:29:51.547215 | f1a394d0268e4fd8b45391ce4c1ee901 | f
 2022-06-24 15:30:33.958874 | 53f1be3f52b44672b377b87c01ef1124 | f
 2022-06-24 15:30:33.958874 | 0ccd632409554b328b313aabf5d3b9a9 | f
 2022-06-24 15:30:33.958874 | 9859b8e1ce8b4457b7f74d5f9dda98be | f
 2022-06-24 15:30:33.958874 | f8081ab08fa84bf1ba9f9373f22d92d4 | f
(5 rows)

```

Nothing has been expired or deleted

- Now with short ages:

```
rsa_service.rotate(5, timedelta(minutes=1), timedelta(minutes=1))
```

```
          created           |               kid                | expired 
----------------------------+----------------------------------+---------
 2022-06-24 15:29:51.547215 | f1a394d0268e4fd8b45391ce4c1ee901 | t
 2022-06-24 15:30:33.958874 | 53f1be3f52b44672b377b87c01ef1124 | t
 2022-06-24 15:30:33.958874 | 0ccd632409554b328b313aabf5d3b9a9 | t
 2022-06-24 15:30:33.958874 | 9859b8e1ce8b4457b7f74d5f9dda98be | t
 2022-06-24 15:30:33.958874 | f8081ab08fa84bf1ba9f9373f22d92d4 | t
 2022-06-24 15:31:16.356098 | a0d03c0de65e455985bace53da618d18 | f
 2022-06-24 15:31:16.356098 | 72b2cd9d226b4d5dab161994fa858117 | f
 2022-06-24 15:31:16.356098 | 4ac434c089f14463beceb6b54c154db3 | f
 2022-06-24 15:31:16.356098 | 4fe21a6daf1f45279fb14742ae8aec1e | f
 2022-06-24 15:31:16.356098 | 828fa4dd2cdb462bb6fc20934c64e5fe | f
(10 rows)

```

existing keys got expired.

- Running it again 

```
rsa_service.rotate(5, timedelta(minutes=1), timedelta(minutes=1))
```

```
          created           |               kid                | expired 
----------------------------+----------------------------------+---------
 2022-06-24 15:31:16.356098 | a0d03c0de65e455985bace53da618d18 | t
 2022-06-24 15:31:16.356098 | 72b2cd9d226b4d5dab161994fa858117 | t
 2022-06-24 15:31:16.356098 | 4ac434c089f14463beceb6b54c154db3 | t
 2022-06-24 15:31:16.356098 | 4fe21a6daf1f45279fb14742ae8aec1e | t
 2022-06-24 15:31:16.356098 | 828fa4dd2cdb462bb6fc20934c64e5fe | t
 2022-06-24 15:37:03.85588  | 62ffe00dabed42a2adbc7a674872df3b | f
 2022-06-24 15:37:03.85588  | 096ab3ef6c0944598be714e1b54e61a5 | f
 2022-06-24 15:37:03.85588  | f0b3455bf7b94115bd168f7b688aea1a | f
 2022-06-24 15:37:03.85588  | 5e7c31c2814f49a3ba7be411bb7d18a0 | f
 2022-06-24 15:37:03.85588  | 569572d2c6e540e7934f01c35cdaec53 | f
(10 rows)
```

Expired ones got deleted, active ones got expired and new ones got created.

### DIFF
--- a/lms/services/rsa_key.py
+++ b/lms/services/rsa_key.py
@@ -1,4 +1,5 @@
 import json
+from datetime import datetime, timedelta
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
@@ -16,6 +17,51 @@ class RSAKeyService:
     def __init__(self, db, aes_service):
         self._db = db
         self._aes_service = aes_service
+
+    def rotate(
+        self,
+        target_keys: int,
+        max_age=timedelta(days=7),
+        max_expired_age=timedelta(days=14),
+    ):
+        """
+        Rotate RSA Keys keeping `target_keys` active keys at all times.
+
+        Expires active keys older than `max_age`
+        and deletes expired ones older than `max_expired_age`.
+
+        Keys have a live-cycle of:
+            active -> expired -> deleted
+
+        In flight request we signed with an active key will still be
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task.
+        """
+        now = datetime.now()
+        new_keys = target_keys
+
+        for key in self._db.query(RSAKey).all():
+            key_age = now - key.created
+            # Delete expired keys that have been around for `max_expired_age`
+            if key.expired and key_age >= max_expired_age:
+                self._db.delete(key)
+            # Expire valid keys after max_age
+            elif not key.expired and key_age >= max_age:
+                key.expired = True
+            # Every valid key we see means we need to create one less
+            else:
+                new_keys -= 1
+
+        for _ in range(new_keys):
+            self.generate()
+
+        # Assert the expected result, if we find something different the transition will be aborted
+        # and we'd be back at the initial state
+        assert (
+            self._db.query(RSAKey).filter_by(expired=False).count() == target_keys
+        ), "The number of active RSAKey doesn't match the target"
 
     def generate(self) -> RSAKey:
         """Generate a new random RSA key pair."""
@@ -39,10 +85,15 @@ class RSAKeyService:
         return self._aes_service.decrypt(key.aes_cipher_iv, key.private_key)
 
     def get_all_public_jwks(self):
-        """Get all non-expired keys."""
+        """
+        Get all keys.
+
+        We don't filter out expired ones here to allow in-flight request to
+        continue while keys are rotating.
+        """
         return [
             dict(json.loads(key.public_key), use="sig", kid=key.kid)
-            for key in self._db.query(RSAKey).filter_by(expired=False).all()
+            for key in self._db.query(RSAKey).filter_by().all()
         ]
 
     def get_random_key(self) -> RSAKey:
--- a/tests/unit/lms/services/rsa_key_service_test.py
+++ b/tests/unit/lms/services/rsa_key_service_test.py
@@ -1,8 +1,10 @@
+from datetime import datetime, timedelta
 from unittest.mock import sentinel
 
 import pytest
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
+from freezegun import freeze_time
 from h_matchers import Any
 from jose import constants
 
@@ -12,6 +14,37 @@ from tests import factories
 
 
 class TestAESService:
+    @freeze_time("2022-1-15")
+    def test_rotate(self, svc, valid_keys, expired_keys, db_session):
+        # Keys created the 10th, max ages of 2 and 4 both expire valid and
+        # deleted expired on the 15th.
+        target_keys = 5
+
+        svc.rotate(
+            target_keys, max_age=timedelta(days=2), max_expired_age=timedelta(days=4)
+        )
+
+        assert all((key.expired for key in valid_keys))
+        assert not any((key in db_session for key in expired_keys))
+        assert db_session.query(RSAKey).filter_by(expired=False).count() == target_keys
+
+    @freeze_time("2022-1-11")
+    @pytest.mark.usefixtures("expired_keys")
+    def test_rotate_doesnt_expire_valid(self, svc, valid_keys):
+        # Keys created the 10th, max age of 2, doesn't expire the keys on the
+        # 11th.
+        svc.rotate(3, max_age=timedelta(days=2), max_expired_age=timedelta(days=5))
+
+        assert not any((key.expired for key in valid_keys))
+
+    @freeze_time("2022-1-11")
+    @pytest.mark.usefixtures("valid_keys")
+    def test_rotate_doesnt_delete_recent_expires(self, svc, expired_keys, db_session):
+        # Keys created the 10th, max expired age of 5, doesn't delete the keys
+        # on the 11th.
+        svc.rotate(3, max_age=timedelta(days=2), max_expired_age=timedelta(days=5))
+        assert all((key in db_session for key in expired_keys))
+
     def test_generate(self, svc, aes_service, rsa, jwk):
         jwk.RSAKey.return_value.to_dict.return_value = {}
 
@@ -45,7 +78,6 @@ class TestAESService:
             algorithm=constants.Algorithms.RS256,
             key=public_bytes.return_value.decode("utf-8"),
         )
-
         assert key.private_key == aes_service.encrypt.return_value
         assert key.aes_cipher_iv == aes_service.build_iv.return_value
 
@@ -57,11 +89,12 @@ class TestAESService:
         aes_service.decrypt.assert_called_once_with(sentinel.iv, sentinel.private_key)
         assert private_key == aes_service.decrypt.return_value
 
-    def test_get_all_public_jwks(self, svc, valid_keys):
+    def test_get_all_public_jwks(self, svc, valid_keys, expired_keys):
         keys = svc.get_all_public_jwks()
 
         expected_keys = [
-            {"use": "sig", "kid": key.kid, "key": "value"} for key in valid_keys
+            {"use": "sig", "kid": key.kid, "key": "value"}
+            for key in (valid_keys + expired_keys)
         ]
         assert keys == Any.list.containing(expected_keys).only()
 
@@ -72,10 +105,19 @@ class TestAESService:
 
     @pytest.fixture
     def valid_keys(self):
-        # Create some noise with some expired keys.
-        factories.RSAKey.create_batch(size=3, expired=True)
         return factories.RSAKey.create_batch(
-            size=3, expired=False, public_key='{"key": "value"}'
+            size=3,
+            public_key='{"key": "value"}',
+            created=datetime(2022, 1, 10),
+        )
+
+    @pytest.fixture
+    def expired_keys(self):
+        return factories.RSAKey.create_batch(
+            size=3,
+            expired=True,
+            public_key='{"key": "value"}',
+            created=datetime(2022, 1, 10),
         )
 
     @pytest.fixture
--- a/bin/update_dev_data.py
+++ b/bin/update_dev_data.py
@@ -59,6 +59,7 @@ class DevDataFactory:
 
         model = getattr(models, item["type"])
         assert "id" in data, "id key needed when using 'model'"
+        self._update_id_sequence(model, data["id"])
         index_elements = {"id"}
         update_columns = set(data.keys()) - index_elements
         bulk_upsert(self.db, model, [data], index_elements, update_columns)
@@ -116,6 +117,14 @@ class DevDataFactory:
     def setattrs(object_, attrs):
         for name, value in attrs.items():
             setattr(object_, name, value)
+
+    def _update_id_sequence(self, model, value):
+        """Update the sequence of the PK so devdata rows don't overlap with the new ones."""
+        sequence = f"{model.__tablename__}_id_seq"
+        self.db.execute(
+            f"""SELECT SETVAL('{sequence}', GREATEST(:value+1, (SELECT last_value FROM "{sequence}")))""",
+            {"value": value},
+        )
 
 
 def devdata():
--- a/tests/factories/rsa_key.py
+++ b/tests/factories/rsa_key.py
@@ -4,5 +4,5 @@ from factory.alchemy import SQLAlchemyModelFactory
 from lms import models
 
 RSAKey = make_factory(
-    models.RSAKey, FACTORY_CLASS=SQLAlchemyModelFactory, public_key="{}"
+    models.RSAKey, FACTORY_CLASS=SQLAlchemyModelFactory, public_key="{}", expired=False
 )
--- a/tests/unit/services.py
+++ b/tests/unit/services.py
@@ -99,7 +99,12 @@ def application_instance_service(mock_service, application_instance):
 
 @pytest.fixture
 def aes_service(mock_service):
-    return mock_service(AESService)
+    aes_service = mock_service(AESService)
+    # Don't return MagicMock objects from these functions so the output can
+    # be tested against things which store them in the DB
+    aes_service.encrypt.return_value = b"fake_ciphertext"
+    aes_service.build_iv.return_value = b"fake_aes_iv"
+    return aes_service
 
 
 @pytest.fixture


lms/services/rsa_key.py
```diff
@@ -17,6 +17,35 @@ def __init__(self, db, aes_service):
...
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task or similar.
+        """
+        # Delete any already expired keys
+        self._db.execute(delete(RSAKey).where(RSAKey.expired.is_(True)))
+
+        # Mark existing keys as expired
+        self._db.execute(
```
marcospri, line 37: https://docs.sqlalchemy.org/en/14/tutorial/orm_data_manipulation.html#orm-enabled-update-statements

lms/services/rsa_key.py
```diff
@@ -17,6 +17,35 @@ def __init__(self, db, aes_service):
...
+        Keys have a live-cycle of:
+            active -> expired -> deleted
+
+        In flight request we signed with an active key will still be
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task or similar.
+        """
+        # Delete any already expired keys
```
marcospri, line 33: This is not very sophisticated process but I think it more than enough.

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +12,20 @@
...
+        target_keys = 5
+
+        # Assert what we think is the initial setup
+        assert not any((key.expired for key in valid_keys))
+        assert all((key.expired for key in expired_keys))
+        assert all((key in db_session for key in expired_keys))
+
+        svc_generates.rotate(target_keys)
+
+        assert all((key.expired for key in valid_keys))
```
marcospri, line 25: Not asserting anything about the keys themselves (other than the expiration status)

lms/services/rsa_key.py
```diff
@@ -17,6 +17,35 @@ def __init__(self, db, aes_service):
...
+            update(RSAKey).where(RSAKey.expired.is_(False)).values(expired=True)
+        )
+
+        # Create new keys
+        for _ in range(target_keys):
+            self.generate()
+
+        assert (
+            self._db.query(RSAKey).filter_by(expired=False).count() == target_keys
+        ), "The number of active RSAKey doesn't match the target"
```
jon-betts, line 47: This is a bit odd. If we allow people to control it, why only have one correct answer? Why do we enforce it after we've done the work? Is this to get an error, but not actually do anything?

It seems like we should either:

* Allow the caller to control the number independently of how many there are
* Read the number from DB and use it directly to create new keys

jon-betts (commented):
I think I might be missing something, but what is the assertion about key count for?

Also, as a general rule, it's ideal if actions called by Celery are idempotent. The rotation here, if called twice, would immediately rotate out the keys it just created, meaning that consumers live keys would move straight to deleted.

An expiry date based system would be more forgiving and allow you to call it multiple times. For example, if there was an expiry date, we could:

* Mark all keys past the date as expired
* Delete all expired keys past the date + some grace period
* Ensure we top back up to X active keys

You wouldn't necessarily need a separate expired column, as it could be based on the date entirely.

I'm wondering how these two things might come together to cause a problem:

* We want to change the number of keys in circulation
* The task runs and raises an assertion error
* Celery retry is enabled and causes the task to run again
* All customers keys are immediately rendered deleted

Quite a chain of co-incidences there, but you see what I mean.

lms/services/rsa_key.py
marcospri: Added a comment, I just want stop everything if that assumption is not true.

lms/services/rsa_key.py
```diff
@@ -17,6 +18,44 @@ def __init__(self, db, aes_service):
         self._db = db
         self._aes_service = aes_service
 
+    def rotate(self, target_keys: int, expires_at: datetime, delete_after: datetime):
```
marcospri, line 21: I prefer absolute dates here and only depend on `now` in the celery task.

lms/services/rsa_key.py
```diff
@@ -17,6 +18,44 @@ def __init__(self, db, aes_service):
...
+        In flight request we signed with an active key will still be
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task or similar.
+        """
+        # Delete any already expired keys that hasn't ben updated (expired) recently
+        self._db.execute(
+            delete(RSAKey).where(
+                RSAKey.expired.is_(True), RSAKey.updated <= delete_after
```
marcospri, line 37: Also include a date here to avoid deleting them in consecutive calls.

lms/services/rsa_key.py
```diff
@@ -17,6 +18,44 @@ def __init__(self, db, aes_service):
...
+        Keys have a live-cycle of:
+            active -> expired -> deleted
+
+        In flight request we signed with an active key will still be
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task or similar.
+        """
+        # Delete any already expired keys that hasn't ben updated (expired) recently
```
jon-betts, line 34:
```suggestion
        This method is meant to be called periodically by a celery task or
        similar.
        """
        # Delete any already expired keys that haven't been updated (expired)
        # recently
```

lms/services/rsa_key.py
jon-betts: I think I'm confused about the use of updated column here, this all seems a bit complicated. This doesn't pass the tests, but couldn't the underlying logic be something like this?

```python
        for key in self._db.query(RSAKey).all():
            if key.created >= delete_after:
                self._db.delete(key)
            elif key.created >= expires_at:
                key.expired = True
            else:
                target_keys -= 1

        for _ in range(target_keys):
            self.generate()
```

jon-betts (commented):
I suspect I might be missing the point here, but it seems to me this could be a lot simpler than it is by iterating through the keys and basing our decisions on the create date.

I don't think we need to worry about bulk operations as:

* There will never be many keys
* SQLAlchemy might batch up our requests and do them all at once? Not sure about this

lms/services/rsa_key.py
```diff
@@ -17,6 +18,44 @@ def __init__(self, db, aes_service):
...
+
+        Keys have a live-cycle of:
+            active -> expired -> deleted
+
+        In flight request we signed with an active key will still be
+        verifiable as the public key endpoint query (get_all_public_jwks)
+        includes expired keys.
+
+        This method is meant to be called periodically by a celery task or similar.
+        """
```
jon-betts, line 33: Depending on how belt and braces we want to be:

```python
assert expires_at < delete_after, "Expires date is before delete date"
```
?

lms/services/rsa_key.py
```diff
@@ -17,6 +18,51 @@ def __init__(self, db, aes_service):
...
+        now = datetime.now()
+        new_keys = target_keys
+
+        for key in self._db.query(RSAKey).all():
+            key_age = now - key.created
+            # Expire valid keys after max_age
+            if not key.expired and key_age >= max_age:
+                key.expired = True
+            # Delete expired keys that have been around for `max_expired_age`
+            if key.expired and key_age >= max_expired_age:
```
marcospri, line 51: Comparing against the same date in both cases but only deleting ones that are already expired.

jon-betts: Yeah. I think this is a best of both!

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -72,10 +108,20 @@ def test_get_random_key(self, svc, valid_keys):
...
     @pytest.fixture
     def valid_keys(self):
-        # Create some noise with some expired keys.
-        factories.RSAKey.create_batch(size=3, expired=True)
         return factories.RSAKey.create_batch(
-            size=3, expired=False, public_key='{"key": "value"}'
+            size=3,
+            expired=False,
+            public_key='{"key": "value"}',
+            created=datetime(2022, 1, 10),
```
jon-betts, line 115: Some of these could move to the factory, so we don't have to set as many values.

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -90,6 +136,12 @@ def jwk(self, patch):
     def rsa(self, patch):
         return patch("lms.services.rsa_key.rsa")
 
+    @pytest.fixture
+    def svc_generates(self, svc):
+        with patch.object(svc, "generate") as generate:
+            generate.side_effect = lambda: factories.RSAKey(expired=False)
+            yield svc
```
jon-betts, line 143: :thinking: There's something fishy about this. In general we don't need to mock out things like this.

When I remove this fixture (by making it return `svc`), everything bar one test still passes, complaining about converting `MagicMock`s to binary, but I can't see where a mock is getting in.

What's this all about?

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +14,39 @@
...
 class TestAESService:
+    @freeze_time("2022-1-15")
+    def test_rotate(self, svc_generates, valid_keys, expired_keys, db_session):
+        """Keys created the 10th, max ages of 2 and 4 both expire valid and deleted expired on the 15th."""
+        target_keys = 5
+
+        # Assert what we think is the initial setup
+        assert not any((key.expired for key in valid_keys))
+        assert all((key.expired for key in expired_keys))
+        assert all((key in db_session for key in expired_keys))
```
jon-betts, line 25: This feels more like debugging than something we'd need to keep around. If we set this up this way, do we need to assert it?

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +14,39 @@
 
 
 class TestAESService:
+    @freeze_time("2022-1-15")
+    def test_rotate(self, svc_generates, valid_keys, expired_keys, db_session):
+        """Keys created the 10th, max ages of 2 and 4 both expire valid and deleted expired on the 15th."""
```
jon-betts, line 19:
```suggestion
        # Keys created the 10th, max ages of 2 and 4 both expire valid and 
        # deleted expired on the 15th.
```

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +14,39 @@
...
+        svc_generates.rotate(target_keys, timedelta(days=2), timedelta(days=4))
+
+        assert all((key.expired for key in valid_keys))
+        assert not any((key in db_session for key in expired_keys))
+        assert db_session.query(RSAKey).filter_by(expired=False).count() == target_keys
+
+    @freeze_time("2022-1-11")
+    @pytest.mark.usefixtures("expired_keys")
+    def test_rotate_doesnt_expire_valid(self, svc_generates, valid_keys):
+        """Keys created the 10th, max age of 2,  doesn't expire the keys on the 11th."""
```
jon-betts, line 36:
```suggestion
        # Keys created the 10th, max age of 2, doesn't expire the keys on the
        # 11th.
```

I'm not clear on what this is saying

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +14,39 @@
...
+        svc_generates.rotate(len(valid_keys), timedelta(days=2), timedelta(days=5))
+
+        assert not any((key.expired for key in valid_keys))
+
+    @freeze_time("2022-1-11")
+    @pytest.mark.usefixtures("valid_keys")
+    def test_rotate_doesnt_delete_recent_expires(
+        self, svc_generates, expired_keys, db_session
+    ):
+        """Keys created the 10th, max expired age of 5,  doesn't delete the keys on the 11th."""
```
jon-betts, line 46:
```suggestion
        # Keys created the 10th, max expired age of 5, doesn't delete the keys
        # on the 11th.
```

tests/unit/lms/services/rsa_key_service_test.py
```diff
@@ -12,6 +14,39 @@
...
+
+        assert all((key.expired for key in valid_keys))
+        assert not any((key in db_session for key in expired_keys))
+        assert db_session.query(RSAKey).filter_by(expired=False).count() == target_keys
+
+    @freeze_time("2022-1-11")
+    @pytest.mark.usefixtures("expired_keys")
+    def test_rotate_doesnt_expire_valid(self, svc_generates, valid_keys):
+        """Keys created the 10th, max age of 2,  doesn't expire the keys on the 11th."""
+        svc_generates.rotate(len(valid_keys), timedelta(days=2), timedelta(days=5))
```
jon-betts, line 37:
```suggestion
        svc_generates.rotate(
            3, max_age=timedelta(days=2), max_expired_age=timedelta(days=5)
        )
```

Would these all be clearer with kwargs?

tests/unit/lms/services/rsa_key_service_test.py
jon-betts, line 25:
```suggestion
        # Assert what we think is the initial setup
        assert not any((key.expired for key in valid_keys))
        assert all((key.expired for key in expired_keys))
        assert all((key in db_session for key in expired_keys))
        
        target_keys = 5
```

Keeps the var nearer where it's used

tests/unit/lms/services/rsa_key_service_test.py
jon-betts: I think I've worked it out. Our `AESService` mock isn't returning strings, so we can't use it with the DB, which means we disable this.

If we have something like this:

```python
    @pytest.fixture
    def aes_service(self, aes_service):
        # Don't return MagicMock objects from these functions so the output can
        # be tested against things which store them in the DB
        aes_service.encrypt.side_effect = lambda aes_iv, plain_text: plain_text
        aes_service.build_iv.return_value = b"fake_aes_iv"
        return aes_service
```

It's no longer required. I wonder if this should be in the base service fixture?

jon-betts (commented):
I think this is looking good, but there's some oddities in the fixturing which I think is to get around a problem with the `aes_service` fixture not returning strings and bytes.

I think we should fix that, because otherwise it's not clear we are really doing what we should.

tests/unit/lms/services/rsa_key_service_test.py
marcospri: I just moved "expired", moving "public_key" seemed too far from the assertion.

jon-betts (approved):
Yep this works well. Approved!

PR #4093, opened 2022-06-16
Celery initial setup

marcospri:
Configuration has been taken mostly verbatim from checkmate.

# Testing 

- `make services dev`

Now it show the output of a celery worker

### DIFF
--- a/lms/tasks/celery.py
+++ b/lms/tasks/celery.py
@@ -1,0 +1,105 @@
+"""Celery app and configuration."""
+
+import logging
+import os
+import sys
+from contextlib import contextmanager
+
+import celery.signals
+from celery import Celery
+from kombu import Exchange, Queue
+from pyramid.scripting import prepare
+
+from lms.app import create_app
+
+LOG = logging.getLogger(__name__)
+
+
+app = Celery("lms")
+app.conf.update(
+    broker_url=os.environ.get("CELERY_BROKER_URL"),
+    # What options should we have when sending messages to the queue?
+    broker_transport_options={
+        "max_retries": 2,
+        # The delay until the first retry
+        "interval_start": 0.2,
+        # How many seconds added to the interval for each retry
+        "interval_step": 0.2,
+        # Maximum number of seconds to sleep between each retry
+        "interval_max": 0.6,
+    },
+    # Tell celery where our tasks are defined
+    imports=("lms.tasks",),
+    # Acknowledge tasks after the task has executed, rather than just before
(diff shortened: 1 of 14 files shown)


lms/tasks/celery.py
```diff
@@ -0,0 +1,109 @@
+"""Celery app and configuration."""
```
marcospri, line 1: This file is pretty much copy pasted from checkmate.

I haven't seen any value that would need to be different in LMS from the start.

conf/development.ini
```diff
@@ -9,8 +9,6 @@ pyramid.debug_notfound = false
 pyramid.debug_routematch = false
 pyramid.default_locale_name = en
 
-sqlalchemy.url = postgresql://postgres@localhost:5433/postgres
```
marcospri, line 12: Moved to tox.ini

jon-betts: This does look like it might be worth doing in a separate PR (also given recent config changes).

lms/tasks/celery.py
```diff
@@ -0,0 +1,109 @@
...
+LOG = logging.getLogger(__name__)
+
+
+app = Celery("lms")
+app.conf.update(
+    # Without some kind of default here we can't even import the module
+    # in the tests
+    broker_url=os.environ.get(
+        "CELERY_BROKER_URL", "amqp://guest:guest@localhost:5674//"
+    ),
```
jon-betts, line 24: In the spirit of having a single place to look for settings, should this move to `tox.ini`?

lms/tasks/rsa_key.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime, timedelta
+
+from lms.services import RSAKeyService
+from lms.tasks.celery import app
+
+
+@app.task
+def rotate_keys():
```
jon-betts, line 8: Docstring here would be good explaining what this does without having to travel and read around.

jon-betts (approved):
This looks good to me and introduces Celery to LMS (which we've wanted for a while). So I suspect now this hard work has been done we'll take advantage of this.

Is the plan to add more to `h-periodic` to trigger this?

### About the config

I think there are a few mixed things in here which might be worth thinking about:

* If we have defaults for these values, are they really config? Are we ever going to want to change them. Maybe constants in the code is good enough. LMS already has a _lot_ of config options to fill in.
* I don't think the dev value for the Celery broker URL should be in the code when everything else is in `tox.ini`

PR #4095, opened 2022-06-16
Example of using simpler predicates

jon-betts:
A POC for using simpler boolean predicate functions, and then wrapping them, rather than creating objects.

This makes the functions massively easier to test, and it reveals some previously hidden similarities between them.


lms/views/predicates/_predicates.py
```diff
@@ -0,0 +1,101 @@
+def is_canvas_file(_context, request):
```
marcospri, line 1: We could use pyramid's `request_param="canvas_file"`

Having it named probably has some documentation benefit.

lms/views/predicates/_predicates.py
```diff
@@ -0,0 +1,101 @@
...
+        return False
+
+    roles = request.lti_user.roles.lower()
+
+    return any(
+        role in roles for role in ["administrator", "instructor", "teachingassistant"]
+    )
+
+
+def _is_assignment_based(context, request, resource_link_id):
```
marcospri, line 82: Not sure about this name, I'd be more explicit about the DB part.

lms/views/lti/basic_launch.py
```diff
@@ -240,26 +228,20 @@ def legacy_vitalsource_launch(self):
             )
         )
 
-    def _course_copied_launch(self, original_resource_link_id):
+    def _assignment_based_launch(self, resource_link_id):
```
marcospri, line 231: Same thing here, everything is "assignment based" here I reckon. What's speceial here is the DB (stored_assignment? ) part.

lms/views/predicates/_predicates.py
```diff
@@ -0,0 +1,101 @@
...
+    return "url" in request.params
+
+
+def is_db_configured(context, request):
+    """Get if this launch has assignment configuration in the DB."""
+
+    return _is_assignment_based(context, request, context.resource_link_id)
+
+
+class ResourceLinkParams:
```
marcospri, line 31: `HistoryResourceLinkParams`?

lms/views/lti/basic_launch.py
jon-betts: Yeah that's true. I was meaning it as in the DB model, but as a word it applies to everything

lms/views/predicates/_predicates.py
jon-betts: ... _wait for it!_

I'm intending to add `resource_link_id` to this in an upcoming PR

lms/views/predicates/_predicates.py
jon-betts: Not yet, as it's used in `is_configured` below...

jon-betts: If you look at the way this has turned out, we have sort of gone this way, but with our own. I'm sure there's an extra step here to get the rest of the way, but I think it might be to do away with the predicates for the most part.

lms/views/predicates/_predicates.py
```diff
@@ -0,0 +1,63 @@
...
+        # Horrible work around
+        if param == ResourceLinkParam.LTI:
+            resource_link_id = context.resource_link_id
+        else:
+            resource_link_id = context.lti_params.get(param)
+
+        if not resource_link_id:
+            continue
+
+        assigment = request.find_service(name="assignment").get_assignment(
```
marcospri, line 41: We might be making some extra queries with the same augments depending of the type of launch.

That's fine I reckon, simply slapping a `lru_cache` might create some bugs as it's something we might create during the launch (as opposed to an application instance for example)

PR #4097, opened 2022-06-17
Sync api refactor plugin POC

jon-betts:
This doesn't work (there's some snaggles about user / lti_user), but it shows a direction of travel.

The aim here is to make the product specific part:

 * Completely external to the service
 * This is to allow it to be moved to a product directory one day
 * To keep as much conditionality and shared code as possible in the service
 * This makes for longer service tests, but not repeating ourselves in the plugins

The general theme is to view these as a product specific extension of the service code itself, so it's ok that they have a tightly related interface, but they should be independently instantiate-able.


lms/services/grouping/_plugin/blackboard.py
```diff
@@ -0,0 +1,57 @@
...
+            return learner_groups
+
+        raise BlackboardStudentNotInGroup(group_set=group_set_id)
+
+    def get_groups_for_grading(
+        self, svc, course, group_set_id, grading_student_id=None
+    ):
+        return svc.get_course_groupings_for_user(
+            course, grading_student_id, type_=self.group_type, group_set_id=group_set_id
+        )
```
jon-betts, line 43: This could be in the default implementation. Nothing Blackboardy here.

lms/services/grouping/service.py
```diff
@@ -150,11 +174,46 @@ def get_course_groupings_for_user(
...
+            groupings = self.plugin.get_groups_for_grading(
+                self, course, group_set_id, grading_student_id
+            )
+
+        else:
+            groupings = self.plugin.get_groups_for_instructor(
+                self, course, group_set_id
+            )
+
+        return self._to_groupings(groupings, course, self.plugin.group_type)
```
jon-betts, line 211: These two functions are annoyingly similar, but until we can do something about `group_set_id` this split looks here to stay.

lms/services/grouping/service.py
```diff
@@ -150,11 +174,46 @@ def get_course_groupings_for_user(
 
         return query.all()
 
+    def get_sections(self, lti_user, course, grading_student_id=None):
+        if not self.plugin.sections_type:
+            return None
```
jon-betts, line 179: This wouldn't be required if we had standard non-product types. But seeing as it's here, it's a handy way to skip the entire behavior if it's not supported.

lms/services/grouping/factory.py
```diff
@@ -0,0 +1,23 @@
...
+
+def factory(_context, request):
+    if request.product.family == Product.Family.BLACKBOARD:
+        plugin = BlackboardGroupingPlugin(
+            request.find_service(name="blackboard_api_client")
+        )
+    elif request.product.family == Product.Family.CANVAS:
+        plugin = CanvasGroupingPlugin(request.find_service(name="canvas_api_client"))
+    else:
+        plugin = GroupingServicePlugin()
```
jon-betts, line 15: The split between the plugin and the service is what I'm after. This will allow this to be setup when the Product object is setup one day. At that point the code would be:

```python
  return GroupingService(
        db=request.db,
        application_instance=request.find_service(
            name="application_instance"
        ).get_current(),
        plugin=request.product.grouping_plugin,
   )
```

lms/services/grouping/_plugin/blackboard.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class BlackboardGroupSetEmpty(GroupError):
+    """Canvas GroupSet doesn't contain any groups."""
+
+    error_code = "blackboard_group_set_empty"
+
+
+class BlackboardGroupSetNotFound(GroupError):
+    error_code = "blackboard_group_set_not_found"
```
jon-betts, line 20: Looks like we could sort this out with something like:

`raise GroupError("blackboard_group_set_not_found")`

?

PR #4107, opened 2022-06-20
Use a document service to replace duplication in predicates and launch view

jon-betts:
### Before this PR

Previously 65% of the job of working out which document to use was done in predicates, and then another 65% was done in the view as the flow was usually something like:

 * Look up an assignment to determine we are a copied course
 * Set a flag
 * The view looks at the flag
 * Then starts the business of finding the application again

This had all sorts of non-obvious linkages and duplication.

### After this PR

 * There is now a document URL service which will tell you the launch URL
 * If it's there you are "configured"
 * If it's not, you're not
 * There is now only one predicate for this "has_document_url"


lms/validation/_base.py
```diff
@@ -4,7 +4,7 @@
 from pyramid.httpexceptions import HTTPUnsupportedMediaType
 from webargs import pyramidparser
 
-from lms.services import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError
```
jon-betts, line 7: Breaking circular imports

lms/validation/_lti_launch_params.py
```diff
@@ -146,32 +143,6 @@ def handle_error(self, error, data, *, many, **kwargs):
...
-        if (
-            url.lower().startswith("http%3a")
-            or url.lower().startswith("https%3a")
-            or url.lower().startswith("canvas%3a")
-            or url.lower().startswith("vitalsource%3a")
-        ):
-            url = unquote(url)
-            _data["url"] = url
-
-        return _data
```
jon-betts, line 172: This is a prime example of non-helpful use of marshmallow. We actually don't use this to check that a URL is there. We already know it is by the time we get here.

This is purely just hiding the logic around URL decoding away from ourselves where it's used.

lms/views/lti/basic_launch.py
```diff
@@ -16,24 +16,37 @@
...
+    return any(
+        role in roles for role in ["administrator", "instructor", "teachingassistant"]
+    )
+
+
+# This is imported in `lms.views.predicates`
+LTI_LAUNCH_PREDICATES = {
+    "has_document_url": has_document_url,
+    "authorized_to_configure_assignments": is_authorized_to_configure_assignments,
+}
```
jon-betts, line 49: These are the predicates now, and they are here because it's the only place they are used

lms/views/lti/basic_launch.py
```diff
@@ -16,24 +16,37 @@
...
+    """Get if the current user allowed to configured assignments."""
+
+    if not request.lti_user:
+        return False
+
+    roles = request.lti_user.roles.lower()
+
+    return any(
+        role in roles for role in ["administrator", "instructor", "teachingassistant"]
+    )
```
jon-betts, line 42: This should really be a permission

lms/services/document_url.py
```diff
@@ -0,0 +1,107 @@
...
+            if document_url := document_url_source(context, request):
+                return document_url
+
+        return None
+
+    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource)%3a", re.IGNORECASE)
+
+    @classmethod
+    def _from_deep_linking_provided_url(cls, _context, request):
+        """Get the URL from our own configuration set during deep linking."""
```
marcospri, line 38: `our own configuration in the launch URL`?

lms/services/document_url.py
```diff
@@ -0,0 +1,107 @@
...
+            # SpeedGrader launches where query params get double-encoded.
+            # See https://github.com/instructure/canvas-lms/issues/1486
+            if cls._ENCODED_URL.match(url):
+                url = unquote(url)
+
+        return url
+
+    @classmethod
+    def _from_canvas_file(cls, context, request):
+        """Get a Canvas file URL based on temporary data passed from Canvas."""
```
marcospri, line 51: This is the same as URL, you mean temporary here because we just use it to construct the `canvas://` URL?

marcospri (approved):
I was going to suggest to merge the new service with the assignments one but it's true that in many cases (all Canvas) you do get the URL for an assignment but not from a DB assignment so it's probably clearer to keep them separated as well :+1:

lms/services/document_url.py
```diff
@@ -0,0 +1,107 @@
...
+        """Get a document URL from an assignment in the DB matching a param."""
+
+        for param in (
+            "resource_link_id",  # A normal LTI (non-deep linked) launch
+            "ext_d2l_resource_link_id_history",  # A Brightspace course we can copy
+            "resource_link_id_history",  # A Blackboard course we can copy
+        ):
+            # Horrible work around
+            if param == "resource_link_id":
+                resource_link_id = context.resource_link_id
```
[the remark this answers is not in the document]
jon-betts: Yeah I started looking at a PR to fix this, but it's absolutely everywhere. I think it's a case of pushing this down into the `LTIParams` and then never worrying about this again, but it's actually going to be a lot of faff, because we fixture this out so much.

lms/services/document_url.py
jon-betts: The data is temporary in the sense that Canvas changes it every call potentially. It's not something we can store. 

I was trying to distil down the essence of the long comment on this from before. Looks like this is too far the other way and could do with some extra clarification.

lms/services/document_url.py
jon-betts: I was trying to say that we set this value, it's not something the LMS decides to do on it's own.

Maybe this would read better as 

```suggestion
        """Get the URL from the configuration we send to the LMS during deep linking."""
```

... or something that fits on the line etc.

PR #4109, opened 2022-06-20
Use permissions to tell if someone can edit an assignment

jon-betts:
This PR swaps the `authorized_to_edit_assignment` for a permission as:

 * It's weird that it's a predicate when it sounds very like a security thing
 * This keeps all the logic for permissions in one place (`security.py`)
 * This untangles some sideways linkage between deep linking and LTI launch views

Although I'm not suggesting we do, this also shows how we could have a completely unified launch view without any predicates as well.

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -23,6 +23,7 @@ class Identity(NamedTuple):
 
 class Permissions(Enum):
     LTI_LAUNCH_ASSIGNMENT = "lti_launch_assignment"
+    LTI_CONFIGURE_ASSIGNMENT = "lti_configure_assignment"
     API = "api"
     ADMIN = "admin"
 
@@ -78,9 +79,15 @@ class LTISecurityPolicy:
         userid = self.authenticated_userid(request)
 
         if userid:
-            return Identity(
-                userid, [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
-            )
+            permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
+
+            if any(
+                role in request.lti_user.roles.lower()
+                for role in ["administrator", "instructor", "teachingassistant"]
+            ):
+                permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
+
+            return Identity(userid, permissions)
 
         return Identity("", [])
 
--- a/lms/views/lti/basic_launch.py
+++ b/lms/views/lti/basic_launch.py
@@ -23,37 +23,20 @@ from lms.validation.authentication import BearerTokenSchema
 
 
 def has_document_url(context, request):
-    """Get if the current launch has a resolvable document URL."""
+    """
+    Get if the current launch has a resolvable document URL.
+
+    This is imported into `lms.views.predicates` to provide the
+    `has_document_url` predicate.
+    """
     return bool(
         request.find_service(DocumentURLService).get_document_url(context, request)
     )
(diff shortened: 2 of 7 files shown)


lms/views/lti/basic_launch.py
```diff
@@ -23,37 +23,20 @@
 
 
 def has_document_url(context, request):
-    """Get if the current launch has a resolvable document URL."""
+    """
+    Get if the current launch has a resolvable document URL.
+
+    This is imported into `lms.views.predicates` to provide the
```
marcospri, line 29: This dependency between this function and the predicates is a bit confusing so nice comment here.

I guess we eventually moved to having the "one true launch view" this would be used locally.

lms/views/lti/basic_launch.py
```diff
@@ -94,6 +81,15 @@ def unconfigured_launch(self):
 
         This happens if we cannot resolve a document URL for any reason.
         """
+
+        if not self.request.has_permission(Permissions.LTI_CONFIGURE_ASSIGNMENT):
```
marcospri, line 85: I reckon this is much nicer to read and understand in one go rather than two separate view + predicates.

lms/security.py
```diff
@@ -116,9 +117,15 @@ def identity(self, request):
...
-            return Identity(
-                userid, [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
-            )
+            permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
+
+            if any(
+                role in request.lti_user.roles.lower()
+                for role in ["administrator", "instructor", "teachingassistant"]
+            ):
+                permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
```
marcospri, line 126: Nice, we under utilize the security part of security.py/security policies and abuse it for other things.

marcospri (approved):

lms/views/lti/basic_launch.py
```diff
@@ -94,6 +81,15 @@ def unconfigured_launch(self):
...
+            # Looks like the user is not an instructor, so show an error page
+
+            # https://docs.pylonsproject.org/projects/pyramid/en/latest/narr
+            # /renderers.html?highlight=override_renderer#overriding-a-renderer-at-runtime
+            self.request.override_renderer = "lms:templates/lti/basic_launch/unconfigured_launch_not_authorized.html.jinja2"
```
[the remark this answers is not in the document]
jon-betts: I was trying to find that. We could do that, but I think it executes the render right away. As mentioned here:

https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/introduction.html?highlight=render_to_response#write-testable-views

> When you use a [renderer](https://docs.pylonsproject.org/projects/pyramid/en/latest/glossary.html#term-renderer) with your view callable, you are freed from needing to return a "webby" Response object. Instead your views can return a simple Python dictionary. Pyramid will take care of rendering the information in that dictionary to a Response on your behalf. **As a result, your views are more easily tested, since you don't need to parse HTML to evaluate the results.**

I think we would have to parse the results of the HTML to check what we got. I don't think we want to go that far, certainly not now?

lms/views/lti/basic_launch.py
jon-betts: The other option was to register the predicate here, but seeing as they are global whether you like it not, it doesn't seem like a good idea.

PR #4114, opened 2022-06-21
Use a method for registering predicates attached to the object

jon-betts:
The idea here is to make the point where we actually add predicates more obvious.

_edit_: I'm not sure about this one. Seems like it would be worth it if we did this a lot, but we only do it once.


jon-betts:
Overall I think this is neat, but makes things longer and isn't required

Issue #4112, opened 2022-06-21
Remove the `/reports` endpoint


robertknight:
After a [discussion](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1655731930320239) in Slack, we think this endpoint can be removed as it has been replaced by other tools/services.

This route is also the only one that is protected by the username/password login at `/login` and `/logout` as far as I can tell. These login forms accept a fixed username and password defined by the `HASHED_PW` and `USERNAME` settings, unlike the main admin pages which rely on Google as an identity provider. If the reports endpoint is removed, everything that supports the static username/password login could potentially also be removed.

Related issues: https://github.com/hypothesis/lms/issues/394

robertknight:
This has been completed.

PR #4115, opened 2022-06-21
Remove reports

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4112

Requires:

 * https://github.com/hypothesis/lms/pull/4116

This PR pulls out the reports page, and everything associated with it. It's a lot of stuff, and we don't use it. This has some nice benefits:

 * Security and auth is much simpler
 * Lots of code not to maintain
 * Some tests we won't spend time waiting on any more
 * CSS we don't build


robertknight:
This will fix https://github.com/hypothesis/lms/issues/4112.

marcospri (approved):
I'll always remember the "USERNAME" environment variable. 


Not much more to add, good riddance reports endpoint!

PR #4116, opened 2022-06-21
Refactor config loading in LMS

jon-betts:
This was horribly more confusing that it needed to be. It still is to some extent, but it's more under control.

This should be the same behaviour as before.

### DIFF
--- a/lms/config.py
+++ b/lms/config.py
@@ -41,7 +41,7 @@ class _Setting:
 
 
 SETTINGS = (
-    _Setting("sqlalchemy.url", read_from="database_url"),
+    _Setting("database_url"),
     # Whether we're in "dev" mode (as opposed to QA, production or tests).
     _Setting("dev", value_mapper=asbool),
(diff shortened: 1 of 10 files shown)


lms/config.py
```diff
@@ -0,0 +1,121 @@
...
+        raise SettingError("LMS_SECRET must contain only ASCII characters") from err
+
+
+@dataclass(frozen=True)
+class _Setting:
+    """The properties of a setting and how to read it."""
+
+    name: str
+    read_from: str = None
+    mapping: Callable = None
```
marcospri, line 36: Not sure about this name (`mapping`), but I can't come up with a better one to suggest.

marcospri (approved):
This is much nicer :+1: 

And all those seemly random tests made any attempt to reorganize the file a nightmare

lms/config.py
jon-betts: `value_mapper`?

PR #4117, opened 2022-06-21
Use plugin to encapsulate per-LMS grouping differences and use it on the sync API

marcospri:
This doesn't significantly reduce number of lines (in fact adds a few) but I think that the per product split is an improvement.

The split in the plugins using methods by type/role makes the tests easier to follow.

Now the two sync views are in a much better position to be refactored and merged into one.

# Testing

## Canvas

- Course group  https://hypothesis.instructure.com/courses/122/assignments/856
- Sections https://hypothesis.instructure.com/courses/125/assignments/873
- Groups https://hypothesis.instructure.com/courses/125/assignments/1833

## Blackboard

- Course https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_36_1&course_id=_19_1
- Groups https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_398_1&course_id=_19_1


lms/services/grouping/service.py
```diff
@@ -146,11 +145,59 @@ def get_course_groupings_for_user(
...
+        return self._to_groupings(user, groupings, course, self.plugin.group_type)
+
+    def _to_groupings(self, user, groupings, course, type_):
+        if groupings and not isinstance(groupings[0], Grouping):
+            groupings = [
+                {
+                    "lms_id": grouping["id"],
+                    "lms_name": grouping["name"],
+                    "extra": {
+                        "group_set_id": grouping.get("group_category_id")
```
marcospri, line 192: This param mapping belongs probably to the individual plugins? Not sure.

lms/services/grouping/service.py
```diff
@@ -146,11 +145,59 @@ def get_course_groupings_for_user(
...
+            )
+
+        else:
+            groupings = self.plugin.get_groups_for_instructor(
+                self, course, group_set_id
+            )
+
+        return self._to_groupings(user, groupings, course, self.plugin.group_type)
+
+    def _to_groupings(self, user, groupings, course, type_):
```
marcospri, line 185: This could become "store_launch_groupings" or similar and add any other relationships here.

If it was public we could call it also for the course grouping.

jon-betts: I've mentioned it in slack, but I think a lot of this is kind of cruft as a result of the various APIs sending back different things.

It's expected that Blackboard and Canvas will have different return values, but it's not clear why we don't map them into something we expect before we emit them from the clients on our side.

Ideally these values would require no mapping at all.

tests/unit/lms/services/grouping/factory_test.py
```diff
@@ -0,0 +1,77 @@
...
+        pyramid_request.product.family = Product.Family.BLACKBOARD
+
+        service_factory(sentinel.context, pyramid_request)
+
+        BlackboardGroupingPlugin.assert_called_once_with(blackboard_api_client)
+        GroupingService.assert_called_once_with(
+            db=Any(),
+            application_instance=Any(),
+            plugin=BlackboardGroupingPlugin.return_value,
+        )
```
jon-betts, line 61: This is kind of regrettable, but should go away soon.

Issue #4082, opened 2022-06-14
Show sensible error if PDF is not available for item in JSTOR picker


robertknight:
PDFs are not available for some JSTOR items. This can happen for two reasons:

1. The item does not have a PDF (or other content). For example collections (eg. books, journal issues), for which only the items within the collection have associated PDFs. (Example: https://www.jstor.org/stable/i274509, which is a journal _issue_, rather than an article within the issue).
2. The item has a PDF, but the user's institution does not have access.

In the response from the [JSTOR `/metadata/{doi}` endpoint](https://labs.jstor.org/api/anno/docs), the first case is indicated by `has_pdf: false`. The second case is indicated by the `requestor_access_level` field having a value other than `full_access`.

robertknight:
My current plan is to show metadata for container/collection items, but also show an error message indicating that the article cannot be used for an assignment:

<img width="716" alt="JSTOR container notice" src="https://user-images.githubusercontent.com/2458/174061084-f3b8aca6-388f-4bff-8d04-ed6177f69f63.png">

Notice there is no green tick next to the title.

robertknight:
This will need to be done after migrating to the new JSTOR metadata API. See https://github.com/hypothesis/lms/issues/4165. A list of test cases where compiled when working on the initial version of this, which will still be useful. See https://github.com/hypothesis/lms/pull/4118#issuecomment-1164325952.

robertknight:
I revised the description to make it more generic (that is, to show a sensible error regardless of the reason why an item does not have a PDF), and also mention the case where the user's institution lacks access. I also added notes about what the lms app needs to look for in the JSTOR `/metadata/{doi}` API response.

PR #4118, opened 2022-06-22
Populate `is_collection` field in JSTOR article metadata response

robertknight:
Populate the `is_collection` field of the JSTOR metadata API response, so that the frontend can show a notice to the user informing them that they must instead pick a specific article within the collection.

This PR previously contained the frontend changes to use this field. That aspect has since been extracted into a separate PR: https://github.com/hypothesis/lms/pull/4124

**TODO:**

- The current code in this PR was written before I'd tested all the different article types. It needs to either be updated so that it passes all  the cases in [this comment](https://github.com/hypothesis/lms/pull/4118#issuecomment-1164325952) or equivalent logic will be needed on JSTOR's side

**Testing:**

In the JSTOR assignment picker:

1. Enter one of the article IDs from the "Collection IDs" list in [this comment](https://github.com/hypothesis/lms/pull/4118#issuecomment-1164325952) and press Enter. You should see an error message indicating that the item is from a collection, and the submit button should be disabled even if the T&Cs checkbox is checked.
2. Enter one of the article IDs from the "Non-collection IDs" list in the same comment and press Enter. You should not see the error message and the submit button should become enabled once the T&Cs checkbox is checked.

Fixes https://github.com/hypothesis/lms/issues/4082

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -100,7 +100,11 @@ class JSTORService:
 
         title = self._get_title_from_metadata(metadata)
 
-        return {"title": title}
+        # Collections (eg. books) have a different schema than regular articles.
+        # Test for a field that only appears in collections.
+        is_collection = "tb" in metadata
+
+        return {"title": title, "is_collection": is_collection}
 
     def thumbnail(self, article_id: str):
         """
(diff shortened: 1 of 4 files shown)


robertknight:
This branch is obsolete as the structure of the metadata API response is changing. See https://github.com/hypothesis/lms/issues/4165. The test cases listed in https://github.com/hypothesis/lms/pull/4118#issuecomment-1164325952 are still useful.

PR #4120, opened 2022-06-22
Remove reading from the paster file and use ENV vars exclusively

jon-betts:
Requires:

 * https://github.com/hypothesis/lms/pull/4119

This requires a move to all config to be in `tox.ini` like it would be in live.


jon-betts:
Nope. These are used a lot in the functests and the BDD tests. Lets not do this right now

PR #4126, opened 2022-06-23
Product centric plugins

jon-betts:


tests/unit/lms/product/factory_test.py
```diff
@@ -0,0 +1,69 @@
...
+        assert product == CanvasProduct.from_request.return_value
+        assert product.family == Product.Family.CANVAS
+
+    @pytest.fixture(autouse=True)
+    def with_mocked_from_request(self):
+        with ExitStack() as stack:
+            from_requests = [
+                stack.enter_context(patch.object(product_class, "from_request"))
+                for product_class in self.PRODUCTS
+            ]
```
jon-betts, line 67: This looks like a lot, but it allows us to avoid having to load specific services and things that the individual products might require when calling `from_request` and means this shouldn't need to change much as new products are added.

lms/product/factory.py
```diff
@@ -0,0 +1,39 @@
...
+from lms.product.product import Product
+
+
+def get_product_from_request(request) -> Product:
+    """Get the correct product object from the provided request."""
+
+    family = _get_family(request)
+    product_class = _get_implementing_class(family)
+    product = product_class.from_request(request)
+    product.family = family
```
jon-betts, line 14: Doing this outside makes the `from_request` stuff much easier, as only the `GenericProduct` should really have a variable value and it meant passing it through unnecessarily.

lms/product/product.py
```diff
@@ -0,0 +1,47 @@
...
+    """A collection of plugins used to separate LMS specific functionality."""
+
+    grouping_service: GroupingServicePlugin
+
+
+@dataclass
+class Product:
+    """The main product object which is passed around the app."""
+
+    plugin: Plugins
```
jon-betts, line 37: I think `product.plugin.<name>` reads better than `product.plugins.<name>`?

lms/services/grouping/factory.py
```diff
@@ -1,24 +1,11 @@
...
 def service_factory(_context, request):
-    # We plan to put an interface around this soon. So this won't happen here
-    if request.product.family == Product.Family.BLACKBOARD:
-        plugin = BlackboardGroupingPlugin(
-            request.find_service(name="blackboard_api_client")
-        )
-    elif request.product.family == Product.Family.CANVAS:
-        plugin = CanvasGroupingPlugin(request.find_service(name="canvas_api_client"))
-    else:
-        plugin = GroupingServicePlugin()
```
jon-betts, line 16: Lot of code to remove this... but it's a big picture move.

tests/unit/lms/services/grouping/factory_test.py
```diff
@@ -1,77 +1,25 @@
...
-    def BlackboardGroupingPlugin(self, patch):
-        return patch("lms.services.grouping.factory.BlackboardGroupingPlugin")
-
-    @pytest.fixture
-    def CanvasGroupingPlugin(self, patch):
-        return patch("lms.services.grouping.factory.CanvasGroupingPlugin")
-
-    @pytest.fixture
-    def GroupingServicePlugin(self, patch):
-        return patch("lms.services.grouping.factory.GroupingServicePlugin")
```
jon-betts, line 77: As we added new products this would have grown and grown, and so would every other thing we added plugins to.

Now we should be able to add new things like Moodle sections without altering the grouping service at all, which sounds like a win.

lms/product/blackboard/product.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+
+from lms.product.blackboard._plugin.grouping import BlackboardGroupingPlugin
+from lms.product.product import Plugins, Product
+
+
+@dataclass
+class BlackboardProduct(Product):
```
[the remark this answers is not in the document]
marcospri: The simpler names might be better, they are after all already in 

`lms.product.blackboard.product`.

lms/product/blackboard/product.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+
+from lms.product.blackboard._plugin.grouping import BlackboardGroupingPlugin
+from lms.product.product import Plugins, Product
+
+
+@dataclass
```
marcospri, line 7: Not a fan of data classes here, not holding that much data.

In any case if we regret the decision it not a big change.

lms/product/factory.py
```diff
@@ -0,0 +1,39 @@
...
+    return product
+
+
+def _get_implementing_class(family):
+    for lms_class in Product.__subclasses__():
```
marcospri, line 19: If we have to import them anyway I'd would have writing this just a dictionary {family -> class}.get(family, GenericProduct)

Nothing wrong with this approach in any case.

lms/product/canvas/product.py
```diff
@@ -0,0 +1,17 @@
...
+class CanvasProduct(Product):
+    """A product for Canvas specific settings and tweaks."""
+
+    family = Product.Family.CANVAS
+
+    @classmethod
+    def from_request(cls, request):
+        api_client = request.find_service(name="canvas_api_client")
+
+        return cls(plugin=Plugins(grouping_service=CanvasGroupingPlugin(api_client)))
```
marcospri, line 17: I think this is nice, all plugins are neatly just instantiated here :+1:

lms/product/product.py
```diff
@@ -0,0 +1,47 @@
...
+
+@dataclass
+class Plugins:
+    """A collection of plugins used to separate LMS specific functionality."""
+
+    grouping_service: GroupingServicePlugin
+
+
+@dataclass
+class Product:
```
marcospri, line 34: I'd personally collapse `Product` and `GenericProduct` into just one class. I'm not sure the extra layer is giving us much.

marcospri (approved):
Looks good. 

It might feel a bit extra now as we only have one plugin that we use in  exactly one service but it's a good starting point to add new ones.

lms/product/factory.py
jon-betts: I think that's significantly less fancy, in a good way. The auto discovery is prone to error or at least confusion.

lms/product/blackboard/product.py
jon-betts: It's just because it means we don't have to write a lot of methods, and the types will be more strongly enforced by linting etc.

lms/product/blackboard/product.py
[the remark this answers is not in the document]
jon-betts: I've renamed the main class, but I might keep the plugins separate so we can tell when we are using the generic or specialised version.

lms/product/product.py
jon-betts: Yeah... it could be. My heart is telling me not to though. Not sure why.

As it is, the `GenericProduct` almost exactly mirrors the kind of thing you have to implement. And I like the total separation between the interface and the implementation.

tests/unit/lms/product/canvas/_plugin/grouping_test.py
```diff
@@ -126,6 +126,12 @@ def test_get_groups_for_grading(
         )
         assert canvas_api_client.user_groups.return_value == api_groups
 
+    def test_from_request(self, pyramid_request, canvas_api_client):
+        plugin = CanvasGroupingPlugin.from_request(pyramid_request)
+        assert isinstance(plugin, CanvasGroupingPlugin)
+        # pylint: disable=protected-access
+        assert plugin._canvas_api == canvas_api_client
```
jon-betts, line 133: Yeah this isn't great, but ... it's fine

tests/unit/lms/product/product_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class FakePlugin:
+    request: None
+
+    @classmethod
+    def from_request(cls, request):
+        instance = cls()
+        instance.request = request
+        return instance
```
jon-betts, line 16: I absolutely could not get this to work with regular mocks of any kind. All sorts of crazy stuff happened. This works and I'm going to take a step back because I burned a lot of time on this, but if you can do better please do.

tests/unit/lms/product/product_test.py
```diff
@@ -0,0 +1,37 @@
...
+        for field in fields(PluginConfig):
+            assert hasattr(product.plugin, field.name)
+            plugin = getattr(product.plugin, field.name)
+            assert isinstance(plugin, FakePlugin)
+            assert plugin.request == sentinel.request
+
+    @pytest.fixture
+    def plugin_conf(self):
+        return PluginConfig(
+            **{field.name: FakePlugin for field in fields(PluginConfig)}
```
marcospri, line 36: I didn't know about `dataclassses.fields` :+1:

tests/unit/lms/product/product_test.py
marcospri: This doesn't look horrible in any case. If you already have fought with it we can keep it until the next product refactor.

lms/product/product.py
```diff
@@ -0,0 +1,61 @@
...
+
+    grouping_service: GroupingServicePlugin
+
+
+@dataclass
+class PluginConfig:
+    """A collection of plugin class definitions."""
+
+    # These also provide the default implementations
+    grouping_service: type = GroupingServicePlugin
```
marcospri, line 38: I might have inlined this directly in `Product`, mostly because we only have one now.  I see  the appeal of having it in a separated container, probably more future proof.

marcospri (approved):
Looks good. I think you mention you already had a go at creating a second plugin and that affected the design.

We can always revisit this when we actually have a few plugins in case it gets in the way but it looks flexible enough.

lms/product/product.py
jon-betts: I think it makes it easier to see the structure of the product without loads of indentation

PR #4127, opened 2022-06-24
Sync view group set id refactor

marcospri:
After #4133  been merged for a while it would be safer to switch to this approach.

It might not be evident in the diff but with this both sync APIs (canvas/bb) are mostly exactly the same code now. 

The next step is to merge them into one view.

# Testing

Sanity check with:

Canvas sections https://hypothesis.instructure.com/courses/125/assignments/873
Canvas groups https://hypothesis.instructure.com/courses/125/assignments/1833
Canvas course https://hypothesis.instructure.com/courses/122/assignments/856

BB course https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_36_1&course_id=_19_1
BB groups https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_398_1&course_id=_19_1

### DIFF
--- a/lms/views/api/blackboard/sync.py
+++ b/lms/views/api/blackboard/sync.py
@@ -9,10 +9,6 @@ class Sync:
         self.request = request
         self.grouping_service = self.request.find_service(name="grouping")
 
-        self.tool_consumer_instance_guid = self.request.parsed_params["lms"][
-            "tool_consumer_instance_guid"
-        ]
-
     @view_config(
         route_name="blackboard_api.sync",
         request_method="POST",
@@ -25,28 +21,15 @@ class Sync:
             self.request.user,
             self.request.lti_user,
             self.get_course(self.request.parsed_params["course"]["context_id"]),
-            self.group_set(),
+            self.request.parsed_params["assignment"]["group_set_id"],
             self.request.parsed_params.get("gradingStudentId"),
         )
 
-        self.sync_to_h(groups)
+        self.request.find_service(name="lti_h").sync(
+            groups, self.request.parsed_params["group_info"]
+        )
         authority = self.request.registry.settings["h_authority"]
         return [group.groupid(authority) for group in groups]
-
-    def group_set(self):
-        return (
-            self.request.find_service(name="assignment")
-            .get_assignment(
-                self.tool_consumer_instance_guid,
-                self.request.parsed_params["assignment"]["resource_link_id"],
-            )
-            .extra["group_set_id"]
-        )
-
-    def sync_to_h(self, groups):
-        lti_h_svc = self.request.find_service(name="lti_h")
-        group_info = self.request.parsed_params["group_info"]
-        lti_h_svc.sync(groups, group_info)
 
     def get_course(self, course_id):
         return self.request.find_service(name="course").get_by_context_id(course_id)
--- a/lms/views/api/canvas/sync.py
+++ b/lms/views/api/canvas/sync.py
@@ -15,13 +15,15 @@ class Sync:
         permission=Permissions.API,
     )
     def sync(self):
+        course = self.get_course(self._request.json["course"]["context_id"])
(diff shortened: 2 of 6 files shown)


lms/views/api/canvas/sync.py
```diff
@@ -39,32 +41,5 @@ def sync(self):
...
-        return "learner" in self._request.json
-
-    def group_set(self):
-        if self._is_speedgrader:
-            return int(self._request.json["learner"].get("group_set"))
-
-        return int(self._request.json["course"].get("group_set"))
-
-    @property
-    def _is_group_launch(self):
```
marcospri, line 53: If we have a group_set_id, it's a group launch.

lms/views/api/blackboard/sync.py
```diff
@@ -25,28 +21,15 @@ def sync(self):
             self.request.user,
             self.request.lti_user,
             self.get_course(self.request.parsed_params["course"]["context_id"]),
```
jon-betts, line 23: I'd probably inline the get course here too, it's a one liner.

lms/views/api/blackboard/sync.py
```diff
@@ -25,28 +21,15 @@ def sync(self):
...
-            self.group_set(),
+            self.request.parsed_params["assignment"]["group_set_id"],
             self.request.parsed_params.get("gradingStudentId"),
```
jon-betts, line 25: There are a lot of args here, using `kwargs` would be helpful.

lms/views/api/blackboard/sync.py
```diff
@@ -9,10 +9,6 @@ def __init__(self, request):
         self.request = request
         self.grouping_service = self.request.find_service(name="grouping")
```
jon-betts, line 10: Does this need to be an object any more? This looks a lot like it could be a functional style view and just instantiate this service inline?

lms/views/api/canvas/sync.py
jon-betts: There's a bit of weirdness around having `group_set_id` be referred to as `group_set`.

lms/views/api/canvas/sync.py
```diff
@@ -39,32 +41,5 @@ def sync(self):
...
-            return False
-        else:
-            return True
-
-    def _get_course(self):
-        return self._request.find_service(name="course").get_by_context_id(
-            self._request.json["course"]["context_id"],
-        )
+    def get_course(self, course_id):
+        return self._request.find_service(name="course").get_by_context_id(course_id)
```
jon-betts, line 45: As these two views are so similar, basically everything I said for the other goes for this

lms/views/api/blackboard/sync.py
jon-betts: Also, where the hell does `gradingStudentId` come from and why is it camel case?

lms/views/api/canvas/sync.py
```diff
@@ -15,21 +15,23 @@ def __init__(self, request):
         permission=Permissions.API,
     )
     def sync(self):
+        course = self.get_course(self._request.json["course"]["context_id"])
         grading_user_id = self._request.json.get("learner", {}).get("canvas_user_id")
```
jon-betts, line 19: I'm not sure how we intend to unify these things, but it feels like we are making life hard for ourselves by keeping `canvas_user_id` in the name.

I'm not sure of the full picture here, but it feels like we should have a standard field name and then modify how it works in the Canvas situtation?

jon-betts (approved):
This is much better than before, and is really starting to cut down on the cruft.

I think there might have to be some more config sliding shenanigans to get to where we need to go (having "canvas_X" in the config settings feels like a bit of a dead end).

I've made some suggestions for small tidies here and there, but I don't think it's necessarily worth bothering with, as the plan is to immediately rewrite this anyway.

PR #4129, opened 2022-06-24
Simplify the JSTOR service using inlining and refactoring

jon-betts:
While reviewing https://github.com/hypothesis/lms/pull/4098 I noticed quite a few things, but they weren't introduced in that PR.

This is a quick follow up to remove a lot of the extra private methods with two observations:

 * All of the methods that call the API went through the same sequence of steps before calling it
   * This has all moved into the API call
   * We may need to break this up if future calls which don't fit this pattern are found. We'll see
 * Inlined all methods which are only used in one place
 * Removed a bit of un-used generality in some of the methods

The aims are to:

 * Reduce repetition
 * Keep code closer to where it's needed and used
 * Make the code easier to read

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -78,11 +78,23 @@ class JSTORService:
         :param request: Pyramid request
         :param document_url: The URL to annotate
         :return: A URL for Via configured to launch the requested document
+        :raises ExternalRequestError: If we get a value which doesn't look like
+            a public URL from JSTOR
         """
+
+        # Get a signed S3 URL for the given JSTOR URL.
+        s3_url = self._api_request(
+            "/pdf-url/{doi}", doi=document_url.replace("jstor://", "")
+        ).text
+
+        if not s3_url.startswith("https://"):
+            raise ExternalRequestError(
+                f"Expected to get an S3 URL but got: '{s3_url}' instead"
+            )
 
         return via_url(
             request,
-            document_url=self._get_public_url(document_url),
+            document_url=s3_url,
             content_type="pdf",
             # Show content partner banner in client for JSTOR.
             options={"via.client.contentPartner": "jstor"},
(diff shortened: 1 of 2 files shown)


lms/services/jstor.py
```diff
@@ -213,37 +202,22 @@ def _get_title_from_metadata(cls, metadata: dict) -> Optional[str]:
...
+        # Extract text nodes using HTMLParser. We rely on it being tolerant of
+        # invalid markup.
+        chunks = []
+        parser = HTMLParser()
+        parser.handle_data = chunks.append
+        parser.feed(html)
+        parser.close()
+
+        # Strip leading/trailing whitespace and duplicate spaces
+        return " ".join("".join(chunks).split())
```
jon-betts, line 220: Inlined white space removal here

lms/services/jstor.py
```diff
@@ -134,49 +144,28 @@ def thumbnail(self, article_id: str):
...
-
-        if not s3_url.startswith("https://"):
-            raise ExternalRequestError(
-                f"Expected to get an S3 URL but got: '{s3_url}' instead"
-            )
-
-        return s3_url
+        doi = article_id
+        if "/" not in article_id:
+            doi = f"{self.DEFAULT_DOI_PREFIX}/{article_id}"
```
jon-betts, line 156: Inlined `_doi_from_article_id` here

lms/services/jstor.py
```diff
@@ -134,49 +144,28 @@ def thumbnail(self, article_id: str):
...
 
-    def _api_request(
-        self, endpoint: str, params: Optional[dict] = None
-    ) -> requests.Response:
-        """
-        Make an authenticated request to the JSTOR API.
+        url = self._api_url + path_template.format(doi=quote(doi, safe="/"))
```
jon-betts, line 158: Inlined `_format_uri` here

lms/services/jstor.py
```diff
@@ -94,13 +104,11 @@ def metadata(self, article_id: str):
 
         :param article_id: A JSTOR article ID or DOI
         """
-        doi = self._doi_from_article_id(article_id)
-        response = self._api_request(_format_uri("/metadata/{}", doi))
```
jon-betts, line 98: This pattern of:

* `_doi_from_article_id`
* `_format_uri` (passed only `doi`)

Was repeated before every call to `_api_request`, and each of these functions were only used in this way.

Generally this is a good sign that you should move things inside the target function.

marcospri (approved):
Looks good. I don't see any code that would imply behavior changes but I haven't tested this with some esoteric DOI formats either.

I don't know if the current API is in state of flux or things are more or less settled now.

In any case, the assumptions made here while inlining some bits make sense and can be easy adapted if some API changes break them.

PR #4130, opened 2022-06-24
Move `resource_link_id` to the `LTIParams`

jon-betts:
We had a quirk where you read all the params from `context.lti_params` except for `resource_link_id` which was read from the context. 

This fixes that weirdness. It also shows a likely place for another plugin to get all the Canvas behavior out of there.

### DIFF
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -45,16 +45,15 @@ class LTIParams(dict):
 def _apply_canvas_quirks(lti_params, request):
     # Canvas SpeedGrader launches LTI apps with the wrong resource_link_id,
     # see:
-    #
     # * https://github.com/instructure/canvas-lms/issues/1952
     # * https://github.com/hypothesis/lms/issues/3228
     #
     # We add the correct resource_link_id as a query param on the launch
     # URL that we submit to Canvas and use that instead of the incorrect
     # resource_link_id that Canvas puts in the request's body.
-    is_speedgrader = request.params.get("learner_canvas_user_id")
+    is_speedgrader = request.GET.get("learner_canvas_user_id")
 
-    if is_speedgrader and (resource_link_id := request.params.get("resource_link_id")):
+    if is_speedgrader and (resource_link_id := request.GET.get("resource_link_id")):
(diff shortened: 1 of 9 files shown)


lms/models/lti_params.py
```diff
@@ -22,8 +24,47 @@ def v11(self):
...
+    #
+    # * https://github.com/instructure/canvas-lms/issues/1952
+    # * https://github.com/hypothesis/lms/issues/3228
+    #
+    # We add the correct resource_link_id as a query param on the launch
+    # URL that we submit to Canvas and use that instead of the incorrect
+    # resource_link_id that Canvas puts in the request's body.
+    is_speedgrader = request.params.get("learner_canvas_user_id")
+
+    if is_speedgrader and (resource_link_id := request.params.get("resource_link_id")):
```
marcospri, line 57: I think we want to be more explicit about where `resource_link_id` is coming from here. 

`request.GET` vs `request.params` (which squashes together GET and POST and I think in the case of LTI1.1 will always be there but it's not very clear which one we are picking/takes precedence for pyramid.).

I'd replace `request.params.get("resource_link_id")` by `request.GET.get("resource_link_id")`

lms/services/document_url.py
```diff
@@ -92,13 +92,7 @@ def _from_assignment_in_db(self, context, _request):
...
             "ext_d2l_resource_link_id_history",  # Ditto for Brightspace
         ):
-            # Horrible work around
-            if param == "resource_link_id":
-                resource_link_id = context.resource_link_id
-            else:
-                resource_link_id = context.lti_params.get(param)
-
-            if resource_link_id and (
+            if (resource_link_id := context.lti_params.get(param)) and (
```
marcospri, line 95: We need to make a grepeable PR replacing all contex.lti_params by request.lti_params

lms/models/lti_params.py
marcospri: The same is true for `learner_canvas_user_id` but there's is less important as there's no clash between GET and POST params.

marcospri (approved):

PR #4131, opened 2022-06-24
Attempting to create an `LTIParam` plugin

jon-betts:
Turns out this is going to need a bit of thought, as:

* We use the product to provide the plugin
* We use the plugin to make the LTI params
* We use the LTI params to determine the product...

Which causes an infinite loop


jon-betts:
Basically I don't think this is going to work. We can try again later.

PR #4133, opened 2022-06-27
Add extra configuration to blackboard and canvas sync API.

marcospri:
Adding this ahead of any changes on the API to populate the client's
config and avoid backwards incompatible changes in the next steps.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -450,6 +450,10 @@ class JSConfig:
                     ],
                     "group_set": req.params.get("group_set"),
                 },
+                "assignment": {
+                    "resource_link_id": self._context.lti_params["resource_link_id"],
+                    "group_set_id": self._context.group_set_id,
+                },
                 "group_info": {
                     key: value
                     for key, value in self._context.lti_params.items()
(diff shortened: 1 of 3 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -483,6 +487,7 @@ def _blackboard_sync_api(self):
                 },
                 "assignment": {
                     "resource_link_id": self._context.lti_params["resource_link_id"],
+                    "group_set_id": self._context.group_set_id,
```
jon-betts, line 490: This whole block and the one above look very similar. It might be a good idea to have a function to produce the standard body, and then edit in the changes for the products. This has the nice effect of pointing out where they are similar or different.

lms/resources/_js_config/__init__.py
```diff
@@ -483,6 +487,7 @@ def _blackboard_sync_api(self):
...
                 },
                 "group_info": {
```
jon-betts, line 492: It would be lovely to have a comment for each of these sections saying why we pass them through. I can't see why we'd need the group info? Or is this just a weird way of passing a bunch of`LTI launch params?

jon-betts (approved):
Not much to say. I'm missing the context for this, but it looks like it can't cause any problems.

I think it would be good to have a private method to generate this big block of config, as it looks like almost the same config both times. This would make the differences stand out and make updating both in tandem in future the default.

PR #4143, opened 2022-06-28
Make application instance part of the LTI context

marcospri:
I reckon we are aiming to reduce the size of `LTILaunchResource` removing LMS specific bits from it.

application_instance is a fundamental part of a launch and fits better here than in JSConfig. 

Removed some, IMO, superfluous exception handling. The application instance is suppose to be there by the time we are handling the launch, if it's not there we rather let any exceptions bubble up to sentry instead of hiding it. We do the same thing for user/lti_user.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -35,11 +35,6 @@ class JSConfig:
     def _h_user(self):
         return self._lti_user.h_user
 
-    @property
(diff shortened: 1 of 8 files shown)


jon-betts (approved):
Yeah, this looks really good. I'd be surprised if this is the only places we do this, as I feel like I've seen it everywhere.

Not much to say, other than it's removed a lot of annoying code.

PR #4144, opened 2022-06-28
Emit pyramid events to store them in the DB

marcospri:
For #3947 

The previous PRs this depends on are not yet merged.

https://github.com/hypothesis/lms/pull/4282
https://github.com/hypothesis/lms/pull/4283

Do not merge yet.

This should allow us to keep track of launches in LTI1.3 while also making the table flexible enough to track other events.

## Testing 

- Start with a fresh DB and in `main` branch

```git checkout main; docker stop lms_postgres_1; docker rm lms_postgres_1; make services db devdata```

- Launch a few assignment (still on main):

https://hypothesis.instructure.com/courses/125/assignments/1513

and from another course

https://hypothesis.instructure.com/courses/252/assignments/1323

- Check the launches table

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from lti_launches"

 id |          created           |                context_id                |                  lti_key                   
----+----------------------------+------------------------------------------+--------------------------------------------
  6 | 2022-08-03 15:56:59.875628 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  7 | 2022-08-03 15:57:04.053202 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  8 | 2022-08-03 15:57:05.850989 | 3216ae11f12d910d3e8feb12d1b2e2a219185a35 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
(3 rows)
```

- Switch to this branch and apply the migrations

```
git checkout lti-events; hdev alembic upgrade head
INFO  [alembic.runtime.migration] Running upgrade 1e146996cca6 -> 42b39684836f, Add events, event_type, event_user and event_data tables.
INFO  [alembic.runtime.migration] Running upgrade 42b39684836f -> 48955a83f040, Backfills the events table using the data from  `lti_launches`.
```

### Check the data migration

- Check that `lti_lauches` is now a view

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "\dv lti_launches"
            List of relations
 Schema |     Name     | Type |  Owner   
--------+--------------+------+----------
 public | lti_launches | view | postgres
(1 row)
```

- And the data should be accessible there:

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from lti_launches"

 id |          created           |                context_id                |                  lti_key                   
----+----------------------------+------------------------------------------+--------------------------------------------
  2 | 2022-08-03 15:57:04.053202 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  1 | 2022-08-03 15:56:59.875628 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  3 | 2022-08-03 15:57:05.850989 | 3216ae11f12d910d3e8feb12d1b2e2a219185a35 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
(3 rows)
```

The IDs would be different (now from the event table). The rest should match.

### Check the new table functionality

- Launch a few assignments again

https://hypothesis.instructure.com/courses/125/assignments/1513
https://hypothesis.instructure.com/courses/252/assignments/1323

- Also configure an assignment, for example 

https://hypothesis.instructure.com/courses/125/assignments/3118/edit?name=Testing+upgrade+-+marcos&due_at=null&points_possible=10

- Query the new events tables

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select event.*, user_id, lti_role_id, extra from event left join event_user on event_id = event.id left join event_data on event_data.event_id = event.id"
 id |         timestamp          | type_id | application_instance_id | course_id | assignment_id | grouping_id | user_id | lti_role_id |                                        extra                                         
----+----------------------------+---------+-------------------------+-----------+---------------+-------------+---------+-------------+--------------------------------------------------------------------------------------
  4 | 2022-08-03 15:59:46.364437 |       1 |                       8 |         1 |            25 |             |       1 |           1 | 
  5 | 2022-08-03 15:59:50.792776 |       1 |                       8 |         1 |            26 |             |       1 |           1 | 
  6 | 2022-08-03 15:59:53.647374 |       1 |                       8 |         3 |            27 |             |       1 |           1 | 
  7 | 2022-08-03 16:00:13.291383 |       2 |                       8 |         1 |               |             |       1 |           1 | {"document_url": "http://localhost:8001/lti_launches?url=https%3A%2F%2Fexample.com"}
  2 | 2022-08-03 15:57:04.053202 |       1 |                       8 |         1 |               |             |         |             | 
  1 | 2022-08-03 15:56:59.875628 |       1 |                       8 |         1 |               |             |         |             | 
  3 | 2022-08-03 15:57:05.850989 |       1 |                       8 |         3 |               |             |         |             | 
(7 rows)
```

- Note the three rows migrated from `lti_launches` before, those don't have user or assignment
- The new launches do have user and assignment
- There's also a row for the deep linking event. No assignment but extra data.

- Check the lti_launches view 

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from lti_launches"

id |          created           |                context_id                |                  lti_key                   
----+----------------------------+------------------------------------------+--------------------------------------------
  6 | 2022-08-03 15:59:53.647374 | 3216ae11f12d910d3e8feb12d1b2e2a219185a35 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  3 | 2022-08-03 15:57:05.850989 | 3216ae11f12d910d3e8feb12d1b2e2a219185a35 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  5 | 2022-08-03 15:59:50.792776 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  4 | 2022-08-03 15:59:46.364437 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  2 | 2022-08-03 15:57:04.053202 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
  1 | 2022-08-03 15:56:59.875628 | f3cd019017839c4630358662a05540f2f6ec5f93 | Hypothesisb3be0b33d0b5e4b1cf3aaf04e0e1819a
(6 rows)
```

The row for deep linking is not present. But the new ones, created after the migration are here.

### CODE REVIEW

File: lms/models/_mixins.py [hunk 1]
```diff
@@ -2,11 +2,7 @@
 
 
 class CreatedUpdatedMixin:
-    created = sa.Column(
-        sa.DateTime(),
-        server_default=sa.func.now(),
-        nullable=False,
-    )
```
marcospri, line 9: Removing a `,`

lms/models/lti_event.py
```diff
@@ -0,0 +1,94 @@
...
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
+from sqlalchemy import insert
+
+from lms.db import BASE
+
+
+class LTIEventType(BASE):
```
marcospri, line 12: Following the roles table, saving a bit of data for type in the main events table. Not that big but it will add up.

lms/models/lti_event.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class LTIEvent(BASE):
+    """Track each LTI launch."""
+
+    __tablename__ = "lti_event"
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+
+    created = sa.Column(
```
marcospri, line 48: Not using the mixin, we only care about `created`.

lms/models/lti_event.py
```diff
@@ -0,0 +1,94 @@
...
+        nullable=False,
+    )
+
+    @staticmethod
+    @lru_cache
+    def _get_type_row(db, type_: LTIEventType.Type):
+        return db.query(LTIEventType).filter_by(type=type_).one()
+
+    @classmethod
+    def add(cls, db, type_, **kwargs):
```
marcospri, line 91: This will obvs go to the service. Looking at getting the model right first.

lms/resources/_js_config/__init__.py
```diff
@@ -214,6 +214,10 @@ def add_deep_linking_api(self):
                 "content_item_return_url": self._context.lti_params[
                     "content_item_return_url"
                 ],
+                "context_id": self._context.lti_params["context_id"],
+                "lms": {
```
marcospri, line 218: Neccesary to use the grouping service (get_course) in the Deep Linking API.

lms/resources/_js_config/__init__.py
```diff
@@ -214,6 +214,10 @@ def add_deep_linking_api(self):
                 "content_item_return_url": self._context.lti_params[
                     "content_item_return_url"
                 ],
+                "context_id": self._context.lti_params["context_id"],
```
marcospri, line 217: Send the context id back to ourselves. We could use `course.id` but I think that's a bit more confusing.

lms/views/lti/basic_launch.py
```diff
@@ -186,6 +186,16 @@ def _record_assignment(self, document_url, extra, is_gradable):
             assignment=assignment, groupings=[self.context.course]
         )
 
+        # Store the launch itself
+        LTIEvent.add(
```
marcospri, line 190: Calling this thing like the old one. I'll replace it by a service.

lms/views/lti/deep_linking.py
```diff
@@ -183,3 +187,15 @@ def _get_content_url(request):
...
             .geturl()
         )
+
+    def store_lti_event(self):
+        course = self.request.find_service(name="course").get_by_context_id(
+            self.request.parsed_params["context_id"]
+        )
+        LTIEvent.add(
+            self.request.db,
+            type_=LTIEventType.Type.DEEP_LINKING,
```
marcospri, line 197: Store deeplinking launches too.

lms/models/lti_event.py
```diff
@@ -0,0 +1,94 @@
...
+            length=64,
+            # Use the string values, not the keys to persist the values
+            values_callable=lambda obj: [item.value for item in obj],
+        ),
+        nullable=False,
+        unique=False,
+    )
+
+
+class LTIEvent(BASE):
```
marcospri, line 41: Explicitly calling this `LTIEvent` and not just event.

The main reason to do this is to continue supporting counting launches with LTI1.3. 

Fixing the new table would involve a fair amount of extra work while here we can just do a one query backfill from the old to the new.

lms/models/_mixins.py
jon-betts: I can't explain why this always makes me so happy.

lms/models/lti_event.py
```diff
@@ -0,0 +1,94 @@
...
+            # type on the Python side, and leave the Postgres side open as a plain
+            # VARCHAR
+            native_enum=False,
+            create_constraint=False,
+            validate_strings=True,
+            # Without a length SQLAlchemy will constrain it to the longest value
+            # we happen to have right now, which could limit us in future
+            length=64,
+            # Use the string values, not the keys to persist the values
+            values_callable=lambda obj: [item.value for item in obj],
```
jon-betts, line 34: This is an annoying pile 'o' code where if you get one part wrong it behaves strangely. If we see this for a third time, I think it's definitely time for a convenience wrapper in `db.py` or something.

lms/models/lti_event.py
jon-betts: I think leaving the old table behind is definitely the way to go.

I'm wondering if the above sentence should read:

> Explicitly calling this LTIEvent and not just _launches_.

Is that what you meant? I wonder by _not_ calling it `Event` what we are trying to rule out? I think there might be events we want to keep track of that aren't strictly speaking anything to do with LTI. 

For example groups reaching a certain level of annotation activity, or users returning after an absence or something.

lms/models/lti_event.py
jon-betts: This is such a nit picky thing, but when I was thinking about this I thought this might want to be something like `timestamp` instead.

There's a bit of confusion when we use `created` and `updated` in our DB. For example in `assignment` table we have a `created` row, but it doesn't record when the assignment was created, but when the row in our DB was created. The assignment could have been created a long time before, and then deep linking setup etc. If we  were table to track the actual create date too, we'd have a bit of a naming problem on our hands.

It's conceivable (for the sake of argument, rather than because I think it'll happen) that we might infer an event happened at a different time and then store that time. For example we scan some activity and detect events and save them with different timestamps (other than now). So the DB row might get created now, but the event might have happened yesterday because we detected it during a schedule task or something.

This is only a semantic difference, but I think our `created` and `updated` columns are basically _metadata_ about when we stored information, but I think the event `timestamp` is _primary data_ about the event.

lms/views/lti/basic_launch.py
jon-betts: Is there any merit in using the Pyramid events for this sort of thing? Or is that just superficial name matching?

The reason I think we might want to do that is it makes integrating event raising into code very cheap. We just emit an event without a service lookup. This means our code and tests don't get festooned with `event_service` stuff.

I'm also imagining that the event service that listens to the events might not even record all the events it's told about. So you could sprinkle these all through the code, like logging, and we'd store the ones that mattered, which might change over time.

No idea if any of that makes any sense.

lms/migrations/versions/a012766eb62e_events.py
```diff
@@ -0,0 +1,92 @@
...
+    op.create_table(
+        "lti_event",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column(
+            "created", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column("type", sa.Integer(), nullable=True),
+        sa.Column("user_id", sa.Integer(), nullable=True),
+        sa.Column("application_instance_id", sa.Integer(), nullable=False),
+        sa.Column("course_id", sa.Integer(), nullable=False),
```
jon-betts, line 38: `grouping_id` ?

Not sure if this would be an either or situation. If we had the grouping we could work out the course if we wanted to. On the other hand, this is kind of a de-normalised table on purpose to make querying it super easy. So maybe both?

lms/migrations/versions/a012766eb62e_events.py
```diff
@@ -0,0 +1,92 @@
...
+        sa.Column("assignment_id", sa.Integer(), nullable=True),
+        sa.Column(
+            "extra",
+            postgresql.JSONB(astext_type=sa.Text()),
+            server_default=sa.text("'{}'::jsonb"),
+            nullable=False,
+        ),
```
jon-betts, line 45: This is the one part of the table which has the possibility of tanking performance. We should be pretty restrained about what we put in here.

jon-betts (commented):
The event table as presented is very close to what I had in mind with a couple of exceptions. 

I hadn't thought of normalising the type out, but it makes sense from a speed point of view. This might make it a bit harder for the casuals to query, but not hugely, and is decent practice regardless.

I've made some other scattered comments and thoughts throughout.

lms/events/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+from pyramid.events import subscriber
+from pyramid.request import Request
+
+from lms.models import Event, EventType, LtiLaunches
+from lms.services import EventService, LTIRoleService
+
+
+@dataclass
+class LTILaunchEvent:
```
marcospri, line 11: I think we should keep the event type close to the subscriber function.

This could be lms/events/lti.py or something along those lines but this can be moved/split any way we want as it grows.

lms/events/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Record Launch in the legacy LtiLaunches table
+    LtiLaunches.add(
+        request.db,
+        request.lti_params.get("context_id"),
+        request.lti_params.get("oauth_consumer_key"),
+    )
+
+    # Record the launch as en event in the Event model's table
+    request.find_service(EventService).insert_event(
```
marcospri, line 33: Now that we have celery we could use a celery task to do the EventService.insert_event (and the rest of the queries, ie send only some IDs in the celery task)

We could do that as a follow up.

lms/migrations/versions/3273c40d33a8_events.py
```diff
@@ -0,0 +1,137 @@
...
+        ),
+        sa.PrimaryKeyConstraint("id", "event_id", name=op.f("pk__event_user")),
+        sa.UniqueConstraint(
+            "user_id", "lti_role_id", name=op.f("uq__event_user__user_id")
+        ),
+    )
+
+    # Manually insert the event_types
+    conn = op.get_bind()
+    conn.execute("""INSERT INTO event_type (type) values ('configured_launch')""")
```
marcospri, line 128: The rest of this is auto generated. 

I could have done this in a loop but I don't think is worth the effort.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
+from enum import Enum
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
+
+from lms.db import BASE, varchar_enum
+
```
marcospri, line 8: Everything this going to be very normalized here

The thinking is to keep the base table very small, allowing the time filter queries to be relatively fast and then pull any extra data as needed.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+
+class EventType(BASE):
```
marcospri, line 10: Event types get their own table.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+    """Time the event occurred, defaults to now() if not specified"""
+
+    type_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("event_type.id", ondelete="cascade")
+    )
+    application_instance = sa.orm.relationship("EventType")
+
+    """One of EventType"""
+
+    application_instance_id = sa.Column(
```
marcospri, line 40: All of these are nullable to accommodate depending of the even type.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+    course = sa.orm.relationship("Course", foreign_keys=[course_id])
+    """If this event relates to one course only, which one"""
+
+    assignment_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("assignment.id", ondelete="cascade"), nullable=True
+    )
+    assignment = sa.orm.relationship("Assignment")
+    """Which assignment this event relates to"""
+
+    grouping_id = sa.Column(
```
marcospri, line 60: Course is technically a grouping but a more common one.

From more general to less ai -> course -> assignment -> grouping. Some event might have all of them, some none.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+        sa.Integer(), sa.ForeignKey("grouping.id", ondelete="cascade"), nullable=True
+    )
+    grouping = sa.orm.relationship("Grouping", foreign_keys=[grouping_id])
+    """If this event belongs to one particular grouping, which one"""
+
+
+class EventUser(BASE):
+    __tablename__ = "event_user"
```
marcospri, line 68: User is meaningless with a role IMO so we need the table in case we have multiple roles.

We can also have multi-user events for free.

lms/services/event.py
```diff
@@ -0,0 +1,33 @@
...
+    def insert_event(
+        self, type_: EventType, user: None, lti_roles=None, data=None, **kwargs
+    ):
+
+        event = Event(type_id=self._get_type_pk(type_), **kwargs)
+        self._db.add(event)
+
+        if user:
+            for lti_role in lti_roles or [None]:
+                self._db.add(EventUser(event=event, user=user, lti_role=lti_role))
```
marcospri, line 21: We are not upserting here, just inserting. SQLA will generate one INSERT statement per role but it will send them all at once to the DB.

We can use the ORM, no need to optimize here.

lms/services/event.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if data:
+            print(data)
+            self._db.add(EventData(event=event, data=data))
+
+    @lru_cache
+    def _get_type_pk(self, type_: EventType.Type) -> int:
```
marcospri, line 27: Need to return the ID, otherwise the models from other sessions would get cached and break.

lms/services/event.py
```diff
@@ -0,0 +1,33 @@
+from functools import lru_cache
+
+from sqlalchemy.orm import Session
+
+from lms.models import Event, EventData, EventType, EventUser
+
+
+class EventService:
```
marcospri, line 8: TODO comments/docstrings galore

lms/views/lti/basic_launch.py
```diff
@@ -61,11 +60,13 @@ def __init__(self, context, request):
     def configured_launch(self):
         """Display a document if we can resolve one to show."""
 
-        return self._show_document(
+        self._show_document(
             document_url=self.request.find_service(DocumentURLService).get_document_url(
                 self.request
             )
         )
+        self.request.registry.notify(LTILaunchEvent(self.request))
```
marcospri, line 68: On the event type:

I think LTILaunch is a very broad concept. Deep Linking is also a type of launch but we don't merge them together for example.

Now that we can be more granular I think it makes sense to keep track of:

- Configured launches (ie a user sees a document)
- Deep linking (ie a teacher configures a document in Canvas)
- Configure assignment? (ie a teacher configures an assignment in non-canvas?) Not in this PR.

lms/views/lti/basic_launch.py
```diff
@@ -215,18 +225,11 @@ def _configure_js_to_show_document(self, document_url, assignment_gradable):
...
+        """Persist application instance info to the DB."""
 
         self.context.application_instance.update_lms_data(self.context.lti_params)
 
-        # Record all LTILaunches for future reporting
-        LtiLaunches.add(
-            self.request.db,
-            self.context.lti_params.get("context_id"),
-            self.context.lti_params.get("oauth_consumer_key"),
-        )
```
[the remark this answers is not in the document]
marcospri: Moved to the event handler. We'll be doing both tables simultaneously

lms/migrations/versions/3273c40d33a8_events.py
```diff
@@ -0,0 +1,137 @@
...
+            ["user.id"],
+            name=op.f("fk__event_user__user_id__user"),
+            ondelete="cascade",
+        ),
+        sa.PrimaryKeyConstraint("id", "event_id", name=op.f("pk__event_user")),
+        sa.UniqueConstraint(
+            "user_id", "lti_role_id", name=op.f("uq__event_user__user_id")
+        ),
+    )
+
```
marcospri, line 125: Should I backfill the new table with the old one here?

If we want to do it I reckon now is the best time.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+    """Which user this events relates to"""
+
+    lti_role_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("lti_role.id", ondelete="cascade"), nullable=True
+    )
+    lti_role = sa.orm.relationship("LTIRole")
+    """What role the user plays in event."""
+
+
+class EventData(BASE):
```
marcospri, line 95: Extra, potentially larger data goes also on its own table.

lms/services/event.py
marcospri, line 8: The service handles the DB side of things for now only

It could also have a wrapper around pyramid events but I reckon, unless we start doing something more complex, having it as vanilla pyramid on the views is much easier to understand/ search for docs etc.

lms/views/lti/basic_launch.py
```diff
@@ -186,6 +186,16 @@ def _record_assignment(self, document_url, extra, is_gradable):
             assignment=assignment, groupings=[self.context.course]
         )
 
+        # Store the launch itself
+        LTIEvent.add(
+            self.request.db,
+            type_=LTIEventType.Type.LAUNCH,
```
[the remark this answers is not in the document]
marcospri: Now that there's a EventType table that would a bit awkward in the code I rekcon.

lms/migrations/versions/a012766eb62e_events.py
marcospri: Yeah, why not both?

lms/events/__init__.py
```diff
@@ -0,0 +1,70 @@
+from dataclasses import dataclass
+
+from pyramid.events import subscriber
+from pyramid.request import Request
+
+from lms.models import Event, EventType, LtiLaunches
+from lms.services import EventService, LTIRoleService
+
```
marcospri, line 8: As discussed before, yeah I think it makes sense to use pyramid events for this.

We should keep events tied to things that don't seem fundamental to the linear executing of the code.

It's a fine line but I think, emitting and event and storing some metadata in the DB is fine, but doing the same thing for for example storing the assignment would be too much.

lms/events/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    LtiLaunches.add(
+        request.db,
+        request.lti_params.get("context_id"),
+        request.lti_params.get("oauth_consumer_key"),
+    )
+
+    # Record the launch as en event in the Event model's table
+    request.find_service(EventService).insert_event(
+        type_=EventType.Type.CONFIGURED_LAUNCH,
+        user=request.user,
```
marcospri, line 35: We could potentially load the event up with some of this information but keeping the event very thin also works.

lms/events/__init__.py
```diff
@@ -1,70 +1,77 @@
...
-class LTIDeepLinkingEvent:
-    request: Request
-    document_url: str
+class LTIEvent(BaseEvent):
+    def __post_init__(self):
+        # pylint:disable=no-member
+        if not self.user_id:
+            self.user_id = self.request.user.id
+
+        if not self.role_ids:
```
marcospri, line 32: This nesting is not very satisfying too look at but it's not complicated it does the trick.

lms/events/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+from pyramid.events import subscriber
+from pyramid.request import Request
+
+from lms.models import EventType, LtiLaunches
+from lms.services import EventService, LTIRoleService
+
+
+@dataclass
+class BaseEvent:  # pylint:disable=too-many-instance-attributes
```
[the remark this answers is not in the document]
jon-betts: Yeah, what can you do?

lms/events/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    request: Request
+    type: EventType.Type
+
+    user_id: Optional[int] = None
```
jon-betts, line 16: I don't know whether the `Option` here has any kind of effect or not? In the case of `dataclass` setting this to `None` means it's optional.

I genuinely don't really know. I suppose it certainly doesn't hurt?

lms/events/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+                    self.request.lti_params.get("resource_link_id"),
+                )
+                .id
+            )
+
+
+@subscriber(LTIEvent)
+def handle_lti_event(event: LTIEvent):
+    # Record Launch in the legacy LtiLaunches table
+    LtiLaunches.add(
```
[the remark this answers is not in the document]
jon-betts: I kind of thing this shouldn't be here at all. It's a legacy bump, and I'd leave it out of the nice new system and leave it in the launch view. Also, I don't think every event we raise indicates a separate launch.We might fire two distinct events, during one login, but the user didn't login twice.

If we did this then we could have a single `handle_event` subscriber which wasn't `LTI` specific at all, meaning we'd support all events with a single subscriber.

lms/services/event.py
jon-betts: I don't really understand this. Can't you put in the string and let SQLAlchemy work it out?

lms/views/lti/basic_launch.py
```diff
@@ -61,11 +61,15 @@ def __init__(self, context, request):
...
         """Display a document if we can resolve one to show."""
 
-        return self._show_document(
+        self._show_document(
             document_url=self.request.find_service(DocumentURLService).get_document_url(
                 self.request
             )
         )
+        self.request.registry.notify(
+            LTIEvent(self.request, EventType.Type.CONFIGURED_LAUNCH)
```
jon-betts, line 70: It might be nice to duplicate the `Type` on the `LTIEvent` object so you can do:

```LTIEvent(self.request, LTIEvent.Type...)```

And not have to worry about a random import.

File: lms/events/__init__.py (hunk 33 above)
Comment (marcospri) [reply lms/events/__init__.py:16]: I have some typing checking tool installed ([pyright](https://github.com/microsoft/pyright)) and  it doesn't like int = None

Everything seems to work and this is more pedantically right, the best version of right!

File: lms/services/event.py [hunk 36]
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import TYPE_CHECKING
+
+from sqlalchemy.orm import Session
+
+from lms.models import Event, EventData, EventType, EventUser
+
+if TYPE_CHECKING:  # pragma: no cover
```
marcospri, line 10: Used for breaking the circular dependency.

lms/services/event.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
```
marcospri, line 1: This makes the `if TYPE_CHECKING` work as expected. 

Allow a type annotation reference a non yet imported name.

lms/services/event.py
marcospri: The table main table is using `event_id` and then we have a table with a PK and the type name. 

The idea is to save a bit space there in the long run storing only the FK on the main table.

To do this without this trick we'll either make a query manually to get the PK of the type every time of use some join on the insert. I reckon this is small enough to be worth it.

tests/unit/conftest.py
```diff
@@ -110,6 +110,11 @@ def pyramid_request(db_session, application_instance, lti_v11_params):
         application_instance_id=application_instance.id,
         user_id=lti_v11_params["user_id"],
     )
+    pyramid_request.user = factories.User(
```
marcospri, line 113: Our real request has a `request.user` prop.

tests/unit/conftest.py
```diff
@@ -166,6 +171,8 @@ def pyramid_config(pyramid_request):
 
         config.action(None, configure_jinja2_assets, args=(config,))
 
+        config.registry.notify = mock.Mock()
```
marcospri, line 174: Mock just the notify part of registry.

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -480,9 +480,11 @@ def with_no_user(self, pyramid_request):
 
 class TestAddDeepLinkingAPI:
     def test_it_adds_deep_linking_v11(self, js_config, pyramid_request):
-        pyramid_request.lti_params = {
-            "content_item_return_url": sentinel.content_item_return_url,
-        }
+        pyramid_request.lti_params.update(
```
marcospri, line 483: This assignment was replacing the whole lti_params for the test. We just want to add on top of the general fixture.

tests/unit/lms/services/user_test.py
```diff
@@ -14,7 +14,7 @@ class TestUserService:
     def test_upsert_user(self, service, lti_user, db_session):
         user = service.upsert_user(lti_user)
 
-        saved_user = db_session.query(User).one()
+        saved_user = db_session.query(User).order_by(User.id.desc()).first()
```
marcospri, line 17: pyramid_request now creates an user (and pyramid_request is bring in by `lti_user`.

lms/db/__init__.py
```diff
@@ -113,6 +114,22 @@ def init(engine, drop=False, stamp=True):  # pragma: nocover
...
+
+
+def create_data(engine):
+    """
+    Create any data needed for the application.
+
+    This type of data is usually crated by migrations which are not executed
+    while using `init` above
+
+    Creating it here allows the data to available in the development environment and tests.
```
marcospri, line 128: This is the main reason to do this. In general we don't run the migration to start a new environment.

An alternative could be to move this to devdata but then will have to deal with this in unit/func/bdd tests anyway. 

I think there's an argument that this is fundamental data for the app that happends to live in the DB while devdata is just useful for testing in the local env.

lms/events/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+    data: Optional[dict] = None
+    """Extra data to associate with this event"""
+
+    def __post_init__(self):
+        for event_field in fields(self):
+            getter_name = f"_get_{event_field.name}"
+
+            # If we don't have a value for `field` and we have implemented
+            # a _get_`field.name` method, use that to get a default
```
marcospri, line 41: This is a bit of a soup of gettattr/hasattr/setattr but avoids a lot of `if not self.field` below.

All those conditionals were making pyling angry and it was getting a bit hard to read.

lms/events/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+@dataclass
+class LTIEvent(BaseEvent):
+    """Class to represent LTI-related events."""
+
+    # pylint:disable=no-member
+
+    Type = EventType.Type
+    """Expose the type here for the callers convenience"""
+
+    def _get_user_id(self):
```
marcospri, line 55: Split by field now.

lms/migrations/versions/48955a83f040_events_backfill.py
```diff
@@ -0,0 +1,74 @@
...
+        JOIN application_instances
+            ON lti_launches.lti_key = application_instances.consumer_key
+        JOIN grouping
+            ON context_id = grouping.lms_id and grouping.application_instance_id = application_instances.id and grouping.type = 'course'
+    """
+    )
+
+    # Rename the existing table to leave the name free for the view.
+    # We'll delete it after a while.
+    op.execute("ALTER TABLE lti_launches RENAME TO old_lti_launches")
```
marcospri, line 52: I reckon better safe than sorry here. We can delete it after a while.

lms/migrations/versions/48955a83f040_events_backfill.py
```diff
@@ -0,0 +1,74 @@
...
+                ON event.application_instance_id = application_instances.id
+            JOIN grouping
+                ON event.course_id = grouping.id
+            WHERE event.type_id = {launch_event_type_id}
+        )"""
+    )
+
+
+def downgrade():
+    pass
```
marcospri, line 74: TODO missing migration.

tests/bdd/steps/lms_db.py
```diff
@@ -20,13 +20,11 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.engine = sqlalchemy.create_engine(TEST_DATABASE_URL)
         self.session_maker = sessionmaker(bind=self.engine.connect())
-        db.init(self.engine, stamp=False)
+        db.init(self.engine, stamp=False, drop=True)
```
marcospri, line 23: This I thinks it's just easier, now we do
`db.init(drop=True)` which drops any tables that might have existed, creates them again.

We don't need `_initial_wipe` anymore.

lms/migrations/versions/48955a83f040_events_backfill.py
jon-betts: Could we split this out as a separate PR? I'm trying to work out what the order we'd deploy these is and I can't quite work it out.

lms/migrations/versions/48955a83f040_events_backfill.py
```diff
@@ -0,0 +1,74 @@
...
+                lti_launches.created,
+                {launch_event_type_id},
+                application_instances.id,
+                grouping.id
+        FROM lti_launches
+        JOIN application_instances
+            ON lti_launches.lti_key = application_instances.consumer_key
+        JOIN grouping
+            ON context_id = grouping.lms_id and grouping.application_instance_id = application_instances.id and grouping.type = 'course'
+    """
```
jon-betts, line 47: I think there's a benefit to querying performance if the physical layout of the rows matches the logical one. So it's ideal if the rows are inserted in timestamp order. (Something about that here https://www.cybertec-postgresql.com/en/column-correlation-explained/).

So we might want to ensure we sort the rows by `created` date here before we insert them.

Also we'd ideally migrate all the old rows and start capturing the new ones at exactly the same moment after, but I'm not sure how we'd do that.

lms/services/event.py
jon-betts: :thinking: I wonder if there's a way around this. Usually you can avoid this by having a third file when two import from each other.

So if the events live somewhere separate from the subscriber, I bet we could fix this.

So maybe:

 * `lms/events/event.py`
 * `lms/services/event.py` - Imports from above
 * `lms/events/subscribers.py` - Imports from both of the above

lms/services/event.py
jon-betts: Right gotcha. 

I think you might want to put a fixed number on the `lru_cache` here, as otherwise it can end up causing memory leaks when used in a method with `self`. I'm not going to pretend to understand why, but lint told me so.

lms/services/event.py
```diff
@@ -0,0 +1,48 @@
...
+
+if TYPE_CHECKING:  # pragma: no cover
+    from lms.events import BaseEvent
+
+
+class EventService:
+    def __init__(self, db: Session):
+        self._db = db
+
+    def insert_event(self, event: BaseEvent):
```
jon-betts, line 18: Doc string?

lms/views/lti/basic_launch.py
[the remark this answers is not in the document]
jon-betts: I think this has been removed from the event handler now, so should this come back or are we trying to do the migration in one go?

lms/migrations/versions/48955a83f040_events_backfill.py
jon-betts: I'm wondering if using the launch event type is going to complicate our lives here, as it means we'll have the same type, but two very different types of rows when it comes to pre-and-post data collection.

I'm not sure how this will look, but I'm thinking of trying to write SQL and then having to remember that most of the rows will have NULLs for various fields. I'm not sure if that will cause more or less problems.

Maybe we can save ourselves a headache by either using a special "legacy_login" type or not migrating at all? I don't know which is best.

jon-betts (commented):
I've had a look through, and I'm getting the feeling this PR isn't quite done yet? There's a migration mixed in here which has the backwards half missing?

I've put some ideas and notes in. I'd appreciate if there was any way to chop this up into some more bite sized peices.

I feel like that might have to happen for deployment anyway?

PR #4149, opened 2022-06-29
Remove old LMS specific sync views

marcospri:
Remove old LMS versions of the sync api in favor of the LMS agnostic one.


jon-betts (approved):
There was a missed bit with the `APIBlackboardSyncSchema`, but I thought I'd just get rid of it and push it.

This looks good to me!

PR #4150, opened 2022-06-29
Move grouping sync config to the plugin

jon-betts:
This moves the Canvas and Blackboard specific parts of grouping config to the grouping service plugin.

I think at this point there's a genuine question about whether this is just the "GroupingPlugin" and should maybe move into the product dir? With this it's no longer exclusively used by the service, and instead is becoming the one stop shop for everything you need to implement to setup grouping.

### DIFF
--- a/lms/product/blackboard/_plugin/grouping.py
+++ b/lms/product/blackboard/_plugin/grouping.py
@@ -18,6 +18,8 @@ class BlackboardGroupingPlugin(GroupingServicePlugin):
 
     group_type = Grouping.Type.BLACKBOARD_GROUP
     sections_type = None  # We don't support sections in Blackboard
+    auth_route = "blackboard_api.oauth.authorize"
+    sync_route = "blackboard_api.sync"
 
     @classmethod
     def from_request(cls, request):
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -431,23 +431,14 @@ class JSConfig:
         if self._context.grouping_type == Grouping.Type.COURSE:
             return None
 
-        if self._request.product.family == Product.Family.CANVAS:
-            config_class = CanvasConfig
-
-        elif self._request.product.family == Product.Family.BLACKBOARD:
-            config_class = BlackboardConfig
-
-        else:
-            return None
-
-        if not config_class.sync_route:
+        plugin = self._request.product.plugin.grouping_service
+        if not plugin.sync_route:
             return None
 
         return {
-            "authUrl": self._request.route_url(config_class.auth_route),
-            "path": self._request.route_path(config_class.sync_route),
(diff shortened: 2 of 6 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -427,78 +427,40 @@ def _groups(self):
...
+                        "group_set_id": self._context.group_set_id,
+                    },
+                    "group_info": {
+                        key: value
+                        for key, value in self._context.lti_params.items()
+                        if key in GroupInfo.columns()
+                    },
+                },
+            ),
+        }
```
jon-betts, line 466: This whole business could be in the default method in the plugin, however it's a bit awkward while we rely on the context object for `group_set_id`.

Also all the `plugin.*` becomes `self.*`.

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -286,98 +287,81 @@ def authToken(self, config):
...
-                "foo": "bar",  # This item should be missing from group_info.
             }
         )
-        return pyramid_request
+        pyramid_request.product = Product.from_request(pyramid_request)
+        pyramid_request.product.plugin.grouping_service = grouping_plugin
+
+        with patch.object(pyramid_request, "route_url") as route_url:
+            route_url.side_effect = lambda route_name, **kwargs: f"route:{route_name}"
+            yield pyramid_request
```
jon-betts, line 364: If this was all in the plugin it would be much simpler to test. It would also cut down the giagnormous file here.

lms/resources/_js_config/__init__.py
```diff
@@ -427,78 +427,40 @@ def _groups(self):
...
             return None
 
-        if self._request.product.family == Product.Family.CANVAS:
-            return self._canvas_sync_api()
-
-        if self._request.product.family == Product.Family.BLACKBOARD:
-            return self._blackboard_sync_api()
+        plugin = self._request.product.plugin.grouping_service
+        if not plugin.sync_route:
+            return None
```
jon-betts, line 436: I can't help but think this isn't quite the right thing here. By the time we get here presumably we know we should be ready to go?

lms/product/blackboard/_plugin/grouping.py
```diff
@@ -18,6 +18,8 @@ class BlackboardGroupingPlugin(GroupingServicePlugin):
 
     group_type = Grouping.Type.BLACKBOARD_GROUP
     sections_type = None  # We don't support sections in Blackboard
+    auth_route = "blackboard_api.oauth.authorize"
+    sync_route = "blackboard_api.sync"
```
jon-betts, line 22: Presumably this is going to go away soon?

jon-betts:
Too soon for this. Whatever happens in:

 * https://github.com/hypothesis/lms/pull/4148
 
Blows this out of the water

PR #4151, opened 2022-06-29
VitalSource automatic sign-in prototype

robertknight:
For:

This is a prototype/spike of a set of changes that enables users to be automatically logged in to VitalSource based on their LTI identity.

Currently, before users launch a VitalSource assignment, they are required to first create their own VitalSource account, and then add the books used in the assignment to their account. When they launch the assignment, they then need to enter their VS credentials before they get taken to the book. This is undesirable friction:

VitalSource login screen

This PR changes the process so that we use user identity information from the LTI launch to make API calls to VitalSource to test whether the current user has access to the book, and then generate a temporary URL that will automatically log them in. If no corresponding user account could be found, or it does not have access to the book, we show an error.

**Summary of changes:**

- Add two new VitalSource-related settings per LMS installation: an API key and the name of a user ID field from an LTI launch request, which is passed to VitalSource to identify the LTI user. Which ID field is used depends on the institution. I believe we'll need to add a third new setting in future, for a link which opens VitalSource's own book reader in the LMS. This is needed for certain error handling scenarios.
- Add methods to `VitalSourceService` to get a user ID from the LTI launch parameters and generate a temporary URL that can be used to launch the VitalSource book reader. This temporary URL will automatically log the user in and put them in the correct book/chapter. Before generating a launch URL we check if the VS user has access to the book, and present an error if not.
- Change how VitalSource launches work, such that the launch URL is fetched by the LMS frontend via an API call, instead of being generated by the initial LMS launch. This is needed because VS assignment launches now involve calls to a third-party API which may take time, and may also result in errors that we need to translate into an appropriate error screen in the frontend. This follows the pattern of how various Canvas and Blackboard files launches work.
- Handle errors that can occur when launching VitalSource assignments in `LaunchErrorDialog`. These will in future need to include an installation-specific link to VitalSource's own book reader.

**Setting up credentials:**

1. In your local LMS admin pages (http://localhost:8001/admin) find the LMS install that you want to test with ( `Hypothesis6b31df6897cac5c46efdca29cb086eeb` is "Localhost (devdata) with Sections Enabled", which I used)
2. For the `vitalsource.api_key` setting, enter the value from the "API key" field of the "VitalSource Test School" entry in 1PW.
3. For the `vitalsource.user_reference` setting, enter the value `user_id`

**Error shown if there is no VitalSource account that corresponds to the current LTI user:**

VitalSource SSO - Account not found

**Error shown if the VitalSource account does not have a license to the book:**

VitalSource SSO - Book not available

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -68,9 +68,20 @@ class JSConfig:
                 ),
             }
         elif document_url.startswith("vitalsource://"):
+            lti_params = self._context.lti_params
             vitalsource_svc = self._request.find_service(name="vitalsource")
-            self._config["vitalSource"] = {
-                "launchUrl": vitalsource_svc.get_launch_url(document_url)
+            book_location = vitalsource_svc.parse_document_url(document_url)
+            self._config["api"]["viaUrl"] = {
+                "path": self._request.route_url(
+                    "vitalsource_api.launch_url",
+                    _query={
+                        "book_id": book_location["book_id"],
+                        "cfi": book_location["cfi"],
+                        "user_reference": vitalsource_svc.get_user_reference(
+                            lti_params
+                        ),
+                    },
+                )
             }
         elif jstor_service.enabled and document_url.startswith("jstor://"):
             self._config["viaUrl"] = jstor_service.via_url(self._request, document_url)
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -110,6 +110,7 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route(
         "vitalsource_api.books.toc", "/api/vitalsource/books/{book_id}/toc"
     )
+    config.add_route("vitalsource_api.launch_url", "/api/vitalsource/launch_url")
 
     config.add_route("admin.index", "/admin/")
     config.add_route("admin.instances", "/admin/instances/")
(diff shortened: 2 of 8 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -68,9 +68,20 @@ def add_document_url(self, document_url):
                 ),
             }
         elif document_url.startswith("vitalsource://"):
+            lti_params = self._context.lti_params
             vitalsource_svc = self._request.find_service(name="vitalsource")
-            self._config["vitalSource"] = {
-                "launchUrl": vitalsource_svc.get_launch_url(document_url)
+            book_location = vitalsource_svc.parse_document_url(document_url)
+            self._config["api"]["viaUrl"] = {
```
robertknight, line 74: We should changing the `viaUrl` naming here so we can reuse it for assignments where the launch steps are the same, but Via is not involved.

robertknight: This is handled in https://github.com/hypothesis/lms/pull/4160.

robertknight:
I think we've learned what we needed to from this spike. The next steps are:

1. Discuss the new errors scenarios (VS account not found, no book license) with VitalSource and find suitable messaging for them and/or ways to eliminate them or reduce how often they can appear.
2. Write up a ticket with a list of sub-tasks needed for a production implementation of this

jon-betts:
Closed in favour of: https://github.com/hypothesis/lms/pull/4205

PR #4152, opened 2022-06-29
Parameterize auth urls

jon-betts:
This is a small change to parameterize the authentication URLs we use in various places.

This doesn't go as far as replacing it in the views and routes definitions, but it does:

 * Remove some conditionality around products
 * Means that if we need to add this in future none of the code will need to change to accommodate it

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -4,6 +4,8 @@ from typing import List, Optional
 
 from lms.models import GroupInfo, Grouping, HUser
 from lms.product import Product
+from lms.product.blackboard import Blackboard
+from lms.product.canvas import Canvas
 from lms.resources._js_config.file_picker_config import FilePickerConfig
 from lms.services import HAPIError, JSTORService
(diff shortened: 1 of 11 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -433,13 +433,8 @@ def _sync_api(self):
 
         req = self._request
```
jon-betts, line 435: This isn't the only place in this file these values are used. They are also used in the file picker sections above, however I feel like it isn't correct to use the values there.

I think we want a similar abstract to `Product` for things like `FilePickerSource` which might fill out these values. The _might_ happen to match those for Blackboard in Blackboard files, but that feels like co-incidence.

For example if we used a 3rd party file picker like Google and integrated OAuth it would be a mistake to pass in the Blackboard OAuth settings just because you happen to be in Blackboard.

marcospri (approved):
Yeah, nicer and progresses in the "no if `Product ==` statements around the code base" goal :+1:

PR #4157, opened 2022-06-30
Increase the timeout for LTIA requests

marcospri:
These requests could involve the LMS doing a HTTP request back to us to
get corresponding JWK while we wait for it to finish.


jon-betts (commented):
I'd like to know a bit more about how many calls like this we expect to make during one call to us as Gunicorn / NGINX timeouts are over the whole transaction, rather than per item.

Two 28 second calls could easily get us over that line by themselves, without any other work being done by the service.

marcospri:
Closing this. The problem might be about the number of workers we use rather than the timeouts. When we see timeouts in production (instead of QA) we can think about the timeouts.

Issue #4080, opened 2022-06-13
Associate assignments with groups and sections


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/122
 
In:

 * https://github.com/hypothesis/lms/issues/3992
 
We began associating courses with assignments. We could (should?) also track other grouping types like sections and groups with them.

### Questions?

 * Does this help us answer questions we care about?
 * Might it be worth doing it anyway, as we can't go back in time and capture data we missed?
 * How would we do it?
 * How does it prioritise with other types of data collection? 

# Tasks

 * [x] https://github.com/hypothesis/lms/pull/4066
 * [ ] https://github.com/hypothesis/lms/pull/4158

jon-betts:
There's some tidy up work here (https://github.com/hypothesis/lms/pull/4182), but this task is done.

PR #4158, opened 2022-06-30
Track assignment group / section relationships

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4080

Based on:

 * https://github.com/hypothesis/lms/pull/4148

We were already tracking the association between assignments and courses, and this adds tracking between assignments and other types of grouping.

### Review notes

The PR is in two commits. The first adds the required data and the second consumes it. If we want to we can split this into two PRs and merge the first one only to allow clients in the wild to get the new config before it's required.

### Testing notes

 * Start everything
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/873
 * Look for assignment grouping relations which aren't to do with course:
```sql
SELECT assignment_id, grouping.type
FROM assignment_grouping
JOIN grouping ON grouping.id = grouping_id
WHERE type != 'course'
```
* In `main` you shouldn't find any, and in this PR you should

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -20,6 +20,11 @@ class AssignmentService:
     def __init__(self, db: Session):
         self._db = db
 
+    def get_assignment_by_id(self, assignment_id) -> Assignment:
+        """Get an assignment by id."""
+
+        return self._db.query(Assignment).get(assignment_id)
+
     def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
         """Get an assignment by resource_link_id."""
 
--- a/lms/views/api/sync.py
+++ b/lms/views/api/sync.py
@@ -11,6 +11,7 @@ class APISyncSchema(PyramidRequestSchema):
         product = fields.Str(required=True)
 
     lms = fields.Nested(LMS, required=True)
+    assignment_id = fields.Int(required=True)
     context_id = fields.Str(required=True)
     group_set_id = fields.Str(required=False, allow_none=True)
     group_info = fields.Dict(required=True)
@@ -48,8 +49,19 @@ def sync(request):
             grading_student_id=grading_student_id,
         )
 
+    # Sync the groups over to H so they are ready to be annotated against
     request.find_service(name="lti_h").sync(
         groupings, request.parsed_params["group_info"]
     )
+
+    # Store the relationship between the assignment and the groupings
+    assignment_service = request.find_service(name="assignment")
+    assignment_service.upsert_assignment_groupings(
+        assignment=assignment_service.get_assignment_by_id(
+            request.parsed_params["assignment_id"]
+        ),
+        groupings=groupings,
+    )
+
     authority = request.registry.settings["h_authority"]
     return [group.groupid(authority) for group in groupings]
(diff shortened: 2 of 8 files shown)


tests/unit/lms/views/lti/basic_launch_test.py
```diff
@@ -182,12 +182,6 @@ def test__show_document(
             sentinel.document_url, assignment_extra=sentinel.assignment_extra
         )
 
-        context.js_config.enable_lti_launch_mode.assert_called_once_with()
-        context.js_config.set_focused_user.assert_not_called()
-        context.js_config.add_document_url.assert_called_once_with(
-            sentinel.document_url
-        )
```
jon-betts, line 189: Moved this down to better follow the layout of the tested code and to make getting the assignment easier

lms/services/assignment.py
```diff
@@ -20,6 +20,11 @@ class AssignmentService:
     def __init__(self, db: Session):
         self._db = db
 
+    def get_assignment_by_id(self, assignment_id) -> Assignment:
+        """Get an assignment by id."""
+
+        return self._db.query(Assignment).get(assignment_id)
```
jon-betts, line 26: Once we know the assignment we want, we don't have to keep passing indirect identifiers about.

lms/views/api/sync.py
```diff
@@ -48,8 +49,19 @@ def sync(request):
...
 
+    # Sync the groups over to H so they are ready to be annotated against
     request.find_service(name="lti_h").sync(
         groupings, request.parsed_params["group_info"]
     )
+
+    # Store the relationship between the assignment and the groupings
+    assignment_service = request.find_service(name="assignment")
+    assignment_service.upsert_assignment_groupings(
+        assignment=assignment_service.get_assignment_by_id(
```
marcospri, line 60: `upsert_assignment_groupings` uses the non ORM bulk upsert so it only uses `assignmen.id` 

We could simplify the PR based on that and save one precious DB query. I'm also happy with the approach of using the whole object shown here.

lms/views/api/sync.py
```diff
@@ -11,6 +11,7 @@ class LMS(Schema):
         product = fields.Str(required=True)
 
     lms = fields.Nested(LMS, required=True)
+    assignment_id = fields.Int(required=True)
```
marcospri, line 14: This being required means all in-flight request will fail. It's probably not a lot but we'll have to schedule the deploy early morning.

Alternately we could make this optional and have a conditional further down to do the data collection based on that.

marcospri (approved):
Added a couple of comments/questions but this good looks good in it's current shape.

lms/views/api/sync.py
jon-betts: Yep. The commits are made so we can split them up if required and we will start sending it first. Although I'm wondering how how we handle present but not expected values. So more work might need to be done.

I wanted to have a chat with you about how much this is worth working around vs. a yolo approach.

lms/views/api/sync.py
jon-betts: Yeah. I think this is worth doing. We've chatted about this and decided this can be merged to get the stats, and then we can follow up with a fix.

lms/views/api/sync.py
jon-betts: We've had a chat and decided that today is a good day to deploy this, as most users are on holiday.

PR #4160, opened 2022-06-30
Rename `viaUrl` JS configuration / API to `contentUrl`

robertknight:
~~**Depends on https://github.com/hypothesis/lms/pull/4159**~~

This configuration refers to the URL of the assignment content iframe.
This used to always be Via, but that is no longer the case. VitalSource
uses it as well for example.

In this PR I kept to the existing naming convention for object properties in the JS config. We might want to revise that in future since it differs from both the `snek_case` of Python and our flavor of `camelCaseWithTLAs` in JS.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -49,7 +49,7 @@ class JSConfig:
         jstor_service = self._request.find_service(iface=JSTORService)
 
         if document_url.startswith("blackboard://"):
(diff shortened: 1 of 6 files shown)


robertknight:
I realized there are still other reference to Via that need to be adjusted a well. eg. the `via_url` property on the result of the content URL response.

robertknight:
I'm going to close this for now and revisit when I come back from holiday.

PR #4162, opened 2022-06-30
Remove duplicate `TEST_DATABASE_URL` default values

seanh:
`TEST_DATABASE_URL` is always set in `tox.ini` so there's no need to have default values for it in the Python code as well.

### DIFF
--- a/tests/bdd/steps/lms_db.py
+++ b/tests/bdd/steps/lms_db.py
@@ -8,9 +8,7 @@ from lms import db, models
 from tests.bdd.step_context import StepContext
 from tests.conftest import get_test_database_url
 
-TEST_DATABASE_URL = get_test_database_url(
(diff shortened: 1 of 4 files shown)


jon-betts (approved):

seanh:
This actually breaks Test Pilot since the tests now fail if they're not run through tox (since `TEST_DATABASE_URL` won't be set). We could replace this PR with one that removes the `TEST_DATABASE_URL`'s from `tox.ini` and leaves them in the Python code instead, that wouldn't break Test Pilot. Or we could [fix Test Pilot](https://github.com/hypothesis/lms/pull/new/add-envvars-for-which-files-to-check)

seanh:
Closing in favor of https://github.com/hypothesis/lms/pull/4175

PR #4166, opened 2022-07-01
Fix product family for GET endpoints

marcospri:
# Testing 

- Without this fix

```update oauth2_token set access_token = 'nope', refresh_token ='nope';```

and launch https://hypothesis.instructure.com/courses/125/assignments/875

There's this error message that flashes for a while.

That doesn't happen in this branch.

### DIFF
--- a/lms/product/factory.py
+++ b/lms/product/factory.py
@@ -19,7 +19,7 @@ def _get_family(request):
     if request.matched_route.name.startswith("canvas_api."):
         return Product.Family.CANVAS
 
-    if request.matched_route.name.startswith("blackboard_api.."):
+    if request.matched_route.name.startswith("blackboard_api."):
         return Product.Family.BLACKBOARD
 
     # If we are in an API request, where we are forwarding the product type ourselves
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lms/resources/_js_config/__init__.py [hunk 1]
```diff
@@ -51,7 +51,6 @@ def add_document_url(self, document_url):
         if document_url.startswith("blackboard://"):
             self._config["api"]["viaUrl"] = {
                 "authUrl": self._request.route_url("blackboard_api.oauth.authorize"),
-                "product": self._request.product.family,
```
marcospri, line 54: This was completely wrong. It was not in the right place it and, in any case this are `GET` request so the mechanism was wrong.

lms/product/factory.py
```diff
@@ -15,6 +15,13 @@ def get_product_from_request(request) -> Product:
 
 
 def _get_family(request):
+    # First, if we are in an LMS specific route, return that
+    if request.matched_route.name.startswith("canvas_api."):
```
marcospri, line 19: This works for GET requests, authentication ones...

jon-betts (approved):
I've given this a test locally. Seems to solve the issue

PR #4168, opened 2022-07-01
Add hashes to requirements files

seanh:
We can't actually do this until we get rid of `tox-pip-sync` because adding hashes to our requirements files breaks `tox-pip-sync`. So this PR is just a reminder to myself for now that we should do this.


seanh:
Closing in favour or <https://github.com/hypothesis/lms/pull/4273>

PR #4170, opened 2022-07-01
Don't send the unnecessary nonce value in LTIA service

marcospri:
Fixing a couple of awkward spots here:

- LTIAHTTP service was using a set of defaults
- Those defaults were used in both the LTIHTTP own request method to sign token request and also using the `sign` method to sign any payload.
- Those defaults look very similar but are not actually the same.

Sending `nonce` on the token request was causing an error as D2L doesn't allow the unexpected parameters: https://sentry.io/organizations/hypothesis/issues/3394604678/?referrer=slack

The fix here is passing all the parameters in needed in both cases directly to JWTService and removing the public `sign` method on LTIA service.

# Testing

- Deep link an assignment: 

https://hypothesis.instructure.com/courses/125/assignments/2956/edit?name=Testing+grading+(marcos)&due_at=null&points_possible=10

- Submitting a submission:
 
make sure to `export FEATURE_FLAG_SUBMIT_ON_ANNOTATION=true` before `make dev`. 

Launch the same assignment as a student and make an annotation. Double check the submission checking the last submission date as a teacher in speed grader)

### DIFF
--- a/lms/services/ltia_http.py
+++ b/lms/services/ltia_http.py
@@ -15,27 +15,6 @@ class LTIAHTTPService:
         self._jwt_service = jwt_service
         self._http = http
 
-    def sign(self, payload: dict):
-        """
-        Sign a payload with one of our private keys.
-
-        We include the default values needed for LTIA APIs
-        Those default claims are defined here:
-
-            https://www.imsglobal.org/spec/security/v1p0/#id-token
-        """
-        now = datetime.utcnow()
-        payload = {
-            "exp": now + timedelta(hours=1),
-            "iat": now,
-            "nonce": uuid.uuid4().hex,
-            "iss": self._lti_registration.client_id,
-            "sub": self._lti_registration.client_id,
-            "aud": self._lti_registration.issuer,
-            **payload,
-        }
-        return self._jwt_service.encode_with_private_key(payload)
-
     def request(self, method, url, scopes, headers=None, **kwargs):
         headers = headers or {}
 
@@ -53,8 +32,13 @@ class LTIAHTTPService:
         https://datatracker.ietf.org/doc/html/rfc7523
         https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token
         """
-        signed_jwt = self.sign(
+        now = datetime.utcnow()
+        signed_jwt = self._jwt_service.encode_with_private_key(
             {
+                "exp": now + timedelta(hours=1),
+                "iat": now,
+                "iss": self._lti_registration.client_id,
+                "sub": self._lti_registration.client_id,
                 "aud": self._lti_registration.token_url,
                 "jti": uuid.uuid4().hex,
             }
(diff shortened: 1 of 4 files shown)


lms/services/ltia_http.py
```diff
@@ -15,27 +15,6 @@ def __init__(
         self._jwt_service = jwt_service
         self._http = http
 
-    def sign(self, payload: dict):
```
marcospri, line 18: This method was not unit tested individually.

lms/services/ltia_http.py
```diff
@@ -15,27 +15,6 @@ def __init__(
...
-        """
-        Sign a payload with one of our private keys.
-
-        We include the default values needed for LTIA APIs
-        Those default claims are defined here:
-
-            https://www.imsglobal.org/spec/security/v1p0/#id-token
```
marcospri, line 25: The comment is correct, it's using those defaults but that's not what we want all the time.

lms/services/ltia_http.py
```diff
@@ -53,8 +32,13 @@ def _get_access_token(self, scopes):
...
         """
-        signed_jwt = self.sign(
+        now = datetime.utcnow()
+        signed_jwt = self._jwt_service.encode_with_private_key(
             {
+                "exp": now + timedelta(hours=1),
+                "iat": now,
+                "iss": self._lti_registration.client_id,
+                "sub": self._lti_registration.client_id,
                 "aud": self._lti_registration.token_url,
```
marcospri, line 42: Before we were already using a differnt `aud` here.

tests/unit/lms/services/ltia_http_test.py
```diff
@@ -18,7 +18,6 @@ def test_request(self, svc, jwt_service, application_instance, uuid, http_servic
                 "aud": application_instance.lti_registration.token_url,
                 "exp": datetime(2022, 4, 4, 1, 0),
                 "iat": datetime(2022, 4, 4, 0, 0),
-                "nonce": uuid.uuid4.return_value.hex,
```
marcospri, line 21: Nonce was not part of of the `client_credentials` parameters but we were sending it because it was in the defaults.

lms/services/ltia_http.py
```diff
@@ -53,8 +32,13 @@ def _get_access_token(self, scopes):
...
         https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token
         """
-        signed_jwt = self.sign(
+        now = datetime.utcnow()
+        signed_jwt = self._jwt_service.encode_with_private_key(
             {
+                "exp": now + timedelta(hours=1),
+                "iat": now,
+                "iss": self._lti_registration.client_id,
+                "sub": self._lti_registration.client_id,
```
marcospri, line 41: Explicit added all keys here, except `nonce` which is not part of this request.

jon-betts (approved):
I've tested this with the provided instructions and it seems to work fine.

I'm not super keen on the duplication here in two different dimensions:

* The `exp` stuff already has logic for it in the JWT service, so it seems like a duplication of effort to have it again. If you call `encode_with_secret` this is available. Why is `encode_with_private_key` any different? One seems like it could be a call to the other
* Some of the fields are different, but the majority are the same. Seems like duplication of effort, but I don't have strong feelings on it.

PR #4172, opened 2022-07-01
Use a plugin setting to move `group_set_id` to the grouping service from the context

jon-betts:


lms/product/canvas/_plugin/grouping.py
```diff
@@ -18,10 +18,36 @@ class CanvasGroupingPlugin(GroupingServicePlugin):
 
     group_type = Grouping.Type.CANVAS_GROUP
     sections_type = Grouping.Type.CANVAS_SECTION
+    deep_linking = True
```
marcospri, line 21: I think this belongs to the top level product object

lms/product/plugin/grouping_service.py
```diff
@@ -24,6 +24,21 @@ class GroupingServicePlugin:  # pragma: nocover
...
+
+    If so we will read group set info from the URL rather than the DB.
+    """
+
+    def get_grouping_type(self, svc) -> Grouping.Type:
+        """Get the type of grouping used in this launch."""
+
+        if svc.get_group_set_id():
+            return Grouping.Type.GROUP
+
```
marcospri, line 39: I'll add a `if self.sections_enabled: return SECTIONS`

And move most of the current overriden `get_grouping_type` to that sections_enabled plugin (and have the default be `return False`

lms/product/canvas/_plugin/grouping.py
```diff
@@ -18,10 +18,36 @@ class CanvasGroupingPlugin(GroupingServicePlugin):
...
 
+    def get_grouping_type(self, svc) -> Grouping.Type:
+        """
+        Return the type of grouping used in this launch.
+
+        Grouping types describe how the course members are divided.
+        If neither of the LMS grouping features are used "COURSE" is the default.
+        """
+
+        if (
```
marcospri, line 36: I'd move this conditional to `sections_enabled` and this method can go away as I mention below.

I reckon that way what grouping type is being selected is more clear in all LMSs.

lms/product/canvas/_plugin/grouping.py
```diff
@@ -18,10 +18,36 @@ class CanvasGroupingPlugin(GroupingServicePlugin):
...
+            # This is a legacy SpeedGrader URL, submitted to Canvas before our
+            # Canvas course sections feature was released.
+            not (
+                "focused_user" in self._request.params
+                and "learner_canvas_user_id" not in self._request.params
+            )
+            # We need a developer key to talk to the API
+            and self._context.application_instance.developer_key
+            and self._context.course.settings.get("canvas", "sections_enabled")
```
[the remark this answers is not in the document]
marcospri: Sending the wrong grouping here won't be evident as we check for `!= course` and then the sync API takes the decision again based on the presence or not of `grouup_set_id`.

PR #4173, opened 2022-07-01
Use pyramid services as the initialisation mechanism

jon-betts:
I didn't want to go this way initially as I didn't want confusion between plugins and services. Here plugins are registered using the same mechanism as services but they aren't intended to be used stand alone like services, and are a different thing.

The plugins were starting to look quite service like, but attempting to solve up-coming problems was creating an awkward situation where we probably want access to the context sometimes. At that point the interface really is the same as services so :shrug: why not just embrace it?

### Let's just use pyramid services

This is a well understood mechanism. Which has some nice features:

* Naturally lazy loading and caching
* Plugins get full access to the request and context if required, meaning we should never have to pass it in to specific calls - as any plugin that needs it can keep it in the init.
* Plugins with no special requirements can be registered as global services and never re-created again

We are just putting a layer on top to get some features I think are useful:

* The plugin object has well defined fields and types
   * This allows very easy auto complete and documentation, making it hard to partially initialise one etc.
* Consuming a plugin is very easy and doesn't involve a service look up (that you are aware of)
* The configuration when specifying products is absolutely minimal with no duplication of defaults or instantiation
* The plugin object can easily be overriden by setting a value to anything you like, which is nice for testing (the default fixture setup might be a bit crap, but once that's done the usage should be ok).

### DIFF
--- a/lms/product/__init__.py
+++ b/lms/product/__init__.py
@@ -5,6 +5,14 @@ from lms.product.product import Product
 
 
 def includeme(config):
+    # Register the default plugins
+    config.include("lms.product.plugin")
+
+    # Give everyone a chance to register their plugins etc.
+    config.include("lms.product.canvas")
+    config.include("lms.product.blackboard")
+
+    # Add the `request.product` method
     config.add_request_method(
         get_product_from_request, name="product", property=True, reify=True
     )
--- a/lms/product/blackboard/__init__.py
+++ b/lms/product/blackboard/__init__.py
@@ -1,1 +1,10 @@
+from lms.product.blackboard._plugin.grouping import BlackboardGroupingPlugin
 from lms.product.blackboard.product import Blackboard
+
+
+def includeme(config):  # pragma: nocover
+    """Register all of our plugins."""
+
+    config.register_service_factory(
+        BlackboardGroupingPlugin.factory, iface=BlackboardGroupingPlugin
+    )
--- a/lms/product/plugin/grouping_service.py
+++ b/lms/product/plugin/grouping_service.py
@@ -62,10 +62,6 @@ class GroupingServicePlugin:  # pragma: nocover
 
         return None
 
-    @classmethod
-    def from_request(cls, request):
-        return cls()
-
 
 class GroupError(Exception):
     """Exceptions raised by plugins."""
--- a/lms/product/plugin/plugin.py
+++ b/lms/product/plugin/plugin.py
@@ -1,0 +1,34 @@
+from dataclasses import dataclass
+
+from lms.product.plugin.grouping_service import GroupingServicePlugin
+
+
+@dataclass
+class PluginConfig:
+    """A collection of plugin class definitions."""
+
+    # These also provide the default implementations
+    grouping_service: type = GroupingServicePlugin
+
+
+class Plugins:
+    """A collection of plugins used to separate LMS specific functionality."""
+
+    class _LazyPlugin:
+        # Lazy load plugins based on the plugin config from pyramid services
+        plugin_name = None
+
+        def __set_name__(self, owner, name):
+            self.plugin_name = name
+
+        def __get__(self, instance, owner):
+            plugin_class = getattr(instance._plugin_config, self.plugin_name)
+            plugin = instance._request.find_service(iface=plugin_class)
+            setattr(instance, self.plugin_name, plugin)  # Overwrite the attr
+            return plugin
+
+    grouping_service: GroupingServicePlugin = _LazyPlugin()
+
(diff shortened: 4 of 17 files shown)


lms/product/plugin/plugin.py
```diff
@@ -0,0 +1,34 @@
...
+        plugin_name = None
+
+        def __set_name__(self, owner, name):
+            self.plugin_name = name
+
+        def __get__(self, instance, owner):
+            plugin_class = getattr(instance._plugin_config, self.plugin_name)
+            plugin = instance._request.find_service(iface=plugin_class)
+            setattr(instance, self.plugin_name, plugin)  # Overwrite the attr
+            return plugin
```
jon-betts, line 28: This is the only bit of complexity here really. This is a descriptor. When Python adds this to the class `__set_name__` is called and we are told the name of the attr we were called with. We use this to find out the correct plugin class from the config and look it up and return it.

We also overwrite the attr which means subsequent access is as fast as it can be.

lms/product/product.py
```diff
@@ -65,9 +50,4 @@ class Product:
     def from_request(cls, request):
         """Create a populated product object from the provided request."""
 
-        plugins = {
-            name: plugin_class.from_request(request)
-            for name, plugin_class in asdict(cls.plugin_config).items()
-        }
-
-        return cls(plugin=Plugins(**plugins))
+        return cls(plugin=Plugins(request, cls.plugin_config))
```
jon-betts, line 53: Hopefully this will be easier to test than the previous version

lms/product/plugin/plugin.py
```diff
@@ -0,0 +1,34 @@
...
+
+from lms.product.plugin.grouping_service import GroupingServicePlugin
+
+
+@dataclass
+class PluginConfig:
+    """A collection of plugin class definitions."""
+
+    # These also provide the default implementations
+    grouping_service: type = GroupingServicePlugin
```
jon-betts, line 11: This is the same class as before. You'll also notice that the products haven't changed at all through this, and the interface is exactly the same.

lms/product/blackboard/__init__.py
```diff
@@ -1 +1,8 @@
+from lms.product.blackboard._plugin.grouping import BlackboardGroupingPlugin
 from lms.product.blackboard.product import Blackboard
+
+
+def includeme(config):
+    config.register_service_factory(
+        BlackboardGroupingPlugin.factory, iface=BlackboardGroupingPlugin
+    )
```
jon-betts, line 8: Going for high localism here. We should be able to alter Canvas stuff without editing anything outside this directory ideally.

lms/product/__init__.py
```diff
@@ -5,6 +5,14 @@
 
 
 def includeme(config):
+    # Register the default plugins
+    config.include("lms.product.plugin")
+
+    # Give everyone a chance to register their plugins etc.
+    config.include("lms.product.canvas")
+    config.include("lms.product.blackboard")
```
jon-betts, line 13: Little bit of make work here to register each new product, but literally a one liner

lms/product/plugin/grouping_service.py
```diff
@@ -62,10 +62,6 @@ def get_groups_for_grading(
 
         return None
 
-    @classmethod
-    def from_request(cls, request):
-        return cls()
```
jon-betts, line 67: If a class requires no special init, we can now create it as a singleton and register it once for the whole app.

lms/product/plugin/grouping_service.py
```diff
@@ -62,10 +62,6 @@ def get_groups_for_grading(
```
jon-betts, line 62: I've moved this from the original location, as we now have to register all our plugins. It's also quite often required to refer to them when you are working on implementing child plugins. The idea here is that this directory will list all the plugins available in a way which is more similar to the product specific dirs.

An object like this obviously has affinity with both the implementers / product side and the consumer / service side. In this case there are more ties now on this side.

tests/unit/conftest.py
```diff
@@ -83,6 +84,14 @@ def lti_v13_params():
     }
 
 
+@pytest.fixture
+def with_plugins(pyramid_request, mock_service):
+    # Ensure that when plugins are read from the current product, all of its
+    # plugins are mocked and ready to be looked up
+    for plugin_class in asdict(pyramid_request.product.plugin_config).values():
+        mock_service(plugin_class)
```
jon-betts, line 92: You can't automatically do this as a part of the `pyramid_request` as `mock_service` requires `pyramid_config` which requires `pyramid_request`...

Due to our lazy loading mechanism, we don't need to do this up front. So we can create the product object, then register the mocks before they are called.

PR #4175, opened 2022-07-03
Remove unnecessary test envvars from `tox.ini`

seanh:
These apparently aren't needed: the tests still pass without them.

We actually don't set _any_ test envvars in `tox.ini` in any of our projects AFAIK which is a nice feature to keep if we can because it allows you to run `.tox/tests/bin/pytest <FILES>` directly if you want to, which is faster (and you can assign a fish or bash alias to make it easier to type).

[Screencast from 03-07-22 12:18:00.webm](https://user-images.githubusercontent.com/22498/177037098-87c1fc3a-d518-4eb5-a763-2051a793eee4.webm)


jon-betts (approved):
Yep, this is one of those nice PRs. If the CI passes, review complete!

PR #4185, opened 2022-07-05
Pass next parameter to pyramid_googleauth to redirect to the target URL

marcospri:
See https://github.com/hypothesis/pyramid-googleauth/pull/5#pullrequestreview-1028730759


jon-betts (approved):
This seems very straight forward. I've not actually tested this, as it requires the lib which is in-flight. But if it didn't work, it wouldn't be subtle.

PR #4186, opened 2022-07-05
Remove `is_canvas` in favour of `product.family`

jon-betts:


marcospri (commented):
Yeah, we don't need this. We manage to get rid of it before we had is_blackboard/is_moodle... :+1:

marcospri (approved):

PR #4187, opened 2022-07-05
Course plugin

jon-betts:


lms/resources/lti_launch.py
```diff
@@ -1,20 +1,16 @@
...
 
-LOG = logging.getLogger(__name__)
-
 
 class LTILaunchResource:
     """
     Context resource for LTI launch requests.
 
-    Many methods and properties of this class are only meant to be called when
+    Some methods and properties of this class are only meant to be called when
```
marcospri, line 13: Goal is to say `All`?

lms/product/canvas/_plugin/course_service.py
```diff
@@ -0,0 +1,31 @@
+from lms.models import ApplicationSettings, CourseGroupsExportedFromH
+from lms.product.plugin.course_service import CourseServicePlugin
+
+
+class CanvasCoursePlugin(CourseServicePlugin):
+    def __init__(self, db_session, parsed_params):
+        self._db_session = db_session
+        self._parsed_params = parsed_params
```
marcospri, line 8: Not a fan of having parsed_params here which are a bit of an unknown shape.

With parsed_params this plugin will only work on the LTILaunch view (which it might be true forever, but it's not very clear).

I'd vote for either:
- Passing the param we care about in the factory. Then the factory only it's `parsed_params` dependant.

- Passing the whole lti_params. This seems like the same situation but lti_params means something everywhere you see it. It's a known shape. In practice this has the same limitation, only will work in LTILaunches but I much more explicit. I'd say using parsed_params outside a view it's a bit confusing in general. More here where both parsed_params and LTIParams happens to be the same (or very similar).


- Passing `custom_canvas_course_id` but take it from LTIParams in the factory. Best of both worlds?

jon-betts: If all these things are in `LTIParam`s then yep! This was a direct lift and shift of the previous functionality.

lms/product/canvas/_plugin/course_service.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def get_new_course_settings(
+        self, settings: ApplicationSettings, authority_provided_id
+    ) -> ApplicationSettings:
+        # Disable sections for courses which existed before sections but we
+        # haven't seen since
+        if self._db_session.query(CourseGroupsExportedFromH).get(authority_provided_id):
+            settings.set("canvas", "sections_enabled", False)
```
jon-betts, line 16: This is actually a change in behavior. Previously this did not only apply in the case of Canvas, but now it will. This doesn't seem like it should logically be a problem, but who knows?

Also over time, this seems less and less likely that it matters at all. As courses this old we've never seen since they were created, and are still being re-used should get rarer over time.

PR #4191, opened 2022-07-06
Remove tox-pip-sync

seanh:
See also: a [previous version of this PR in h](https://github.com/hypothesis/h/pull/7475) (that was not merged).

This PR:

* Removes the `tox-pip-sync` plugin
* Adds the new [tox-faster](https://github.com/hypothesis/tox-faster) plugin, which is just a tiny plugin containing a [one-line tox optimisation](https://github.com/hypothesis/tox-faster/blob/15b8b02bdd0f0a5927123ce4a0e99898c1bdaf95/src/tox_faster/main.py#L6-L8) that was extracted from `tox-pip-sync`. This shaves a few hundred ms off tox's startup time (I think it depends on the size of the venv)
* Adds the new [pip-sync-faster](https://github.com/hypothesis/pip-sync-faster) to our `requirements/*.{in,txt}` files
* Calls `pip-sync-faster` in `tox.ini`'s `commands_pre` section to re-sync the virtualenv if the `requirements/*.txt` files have changed. This saves another few hundred ms compared to calling `pip-sync` directly

Testing
-------

If you see `could not find executable pip-sync-faster` you may need to `rm -rf .tox`. And unfortunately if we merge this all developers may need to run that. I'm not sure why this seems to have passed using the cached `.tox` directory on CI :man_shrugging: 

You'll want to make sure that it correctly updates the venvs when the requirements files change, and that it does not change the venvs when they don't. You'll also want to test that it can create a venv that doesn't already exist.

An easy way to test the speed of this is to run:

```terminal
$ time tox -e format
```

This seems to be about as fast as `main` to me, but please do test it yourself.


tox.ini
```diff
@@ -37,14 +37,7 @@ passenv =
...
 deps =
-    dev: -r requirements/dev.txt
-    {format,checkformatting}: -r requirements/format.txt
-    lint: -r requirements/lint.txt
-    tests: -r requirements/tests.txt
-    coverage: -r requirements/coverage.txt
-    functests: -r requirements/functests.txt
-    bddtests: -r requirements/bddtests.txt
-    dockercompose: -r requirements/dockercompose.txt
+    -r requirements/{env:TOX_ENV_NAME}.txt
```
seanh, line 40: No functional change here I just collapsed the several separate `deps` lines into one by using the `TOX_ENV_NAME` envvar (and relying on the `checkformatting.txt` -> `format.txt` symlink that this PR adds)

Makefile
```diff
@@ -154,6 +154,10 @@ requirements/%.txt: requirements/%.in
 	@touch -a $(subst requirements.txt,dev.txt,$@)
 	@tox -qe $(subst requirements,dev,$(basename $(notdir $@))) --run-command 'pip-compile --allow-unsafe --quiet $(args) $<'
 
+# checkformatting.txt is a symlink not an actual requirements file so it needs its own recipe.
+requirements/checkformatting.txt:
+	@ln -frs requirements/format.txt requirements/checkformatting.txt
```
seanh, line 159: Try `rm requirements/checkformatting.txt` and then `make requirements`, it'll re-create it

jon-betts (changes requested):
I can't get this to work at the moment. This is what I'm seeing:

```shell
> make requirements
make: Nothing to be done for 'requirements'.

> rm requirements/coverage.txt
> make requirements
ERROR: InvocationError for command could not find executable pip-sync-faster
___________________________________________________________ summary ___________________________________________________________
ERROR:   coverage: commands failed
make: *** [Makefile:155: requirements/coverage.txt] Error 1

> rm -rf .tox
> make requirements
make: Nothing to be done for 'requirements'.

> cat requirements/coverage.txt
# Nothing here, it's empty

> rm requirements/coverage.txt
> make requirements
ERROR: InvocationError for command could not find executable pip-sync-faster
___________________________________________________________ summary ___________________________________________________________
ERROR:   coverage: commands failed
make: *** [Makefile:155: requirements/coverage.txt] Error 1
```

I think there's a boot straping problem?
 If `pip-sync-faster` is needed to build the environment, and it's read from `env.txt` and you don't have `env.txt` then you can't build it.

seanh:
> It also seems like we can fail, but in such a way as to create an empty file which then appears to `make` to be correct.

This is because `Makefile` [creates the `*.txt` file empty if it doesn't exist](https://github.com/hypothesis/lms/blob/6eda1d8f748958ff2a045b544d6c1566fed453f8/Makefile#L154), it has to otherwise tox will crash because `tox.ini` lists the `*.txt` file in tox's `deps`. If `make requirements` then crashes or is killed before the `pip-compile` command completes then the empty `*.txt` file is left behind. I'm not sure if there's anything we can do about this one. To force it to re-compile the empty `*.txt` file you can delete it, run `touch requirements/foo.in` or run `make` with the `--always-make` option.

seanh:
> ```terminal
> > make requirements
> make: Nothing to be done for 'requirements'.
> ```

Just a note that this is to be expected: it happens when all your `requirements/*.txt` files are present and their mtimes are the same as or newer than the `*.in`'s. To force it to re-compile the `*.txt` files you can do `touch requirements/*.in` or run `make` with the `--always-make` option.

seanh:
> > rm requirements/coverage.txt
> > make requirements
> ERROR: InvocationError for command could not find executable pip-sync-faster
>
> I think there's a bootstrapping problem? If `pip-sync-faster` is needed to build the environment, and it's read from `env.txt` and you don't have `env.txt` then you can't build it.

Yep, it turns out there are two parts to the cold-start problem:

1. `pip-tools` (for `pip-compile`): since the `pip-tools` dependency is listed in the `requirements/*.txt` files `make requirements` won't work if the venv and `requirements/*.txt` files have both been deleted: `could not find executable pip-compile`. You're not seeing this in your testing because it's masked by the `pip-sync-faster` problem (see below) which happens first. This `pip-compile` problem is a problem with the implementation of `make requirements` that already exists on `main`. The reason we didn't notice it before is that `main` also uses `tox-pip-sync` which [installs an unpinned version of `pip-tools`](https://github.com/hypothesis/tox-pip-sync/blob/74ff009ccd3b66ed6025876f986ed9475d8761de/src/tox_pip_sync/_pip_sync.py#L50). So `make requirements` on `main` is actually broken but is being accidentally saved by `tox-pip-sync`.

   I fixed this by adding a `pip install pip-tools` to the `make requirements` recipe in `Makefile`. This will install the latest version of `pip-tools` if it isn't already installed in the venv which gets us out of the bootstrap problem (and by the time the `make requirements` command has finished that same version of `pip-tools` should now be pinned in the requirements file), but it won't reinstall `pip-tools` or change the installed version if it already exists in the venv.

2. `pip-sync-faster`: again, the dependency is listed in the `requirements/*.txt` files so if both the venvs and the `*.txt` files have been deleted then when you run `make requirements` the `pip-sync-faster` command in `commands_pre` is gonna fail: `could not find executable pip-sync-faster`.

   I've fixed this by moving the `pip-sync-faster` command from `commands_pre` to `commands`. When `make requirements` runs `tox -e <ENV> --run-command 'pip-compile ...'` that overrides the `commands` from `tox.ini` so the `pip-sync-faster` command will not be attempted for this tox run (and by the time this tox run finishes `pip-sync-faster` will have been installed in the venv).

Here's my testing of the fixed version:

```terminal
# This happens when the mtimes of your requirements.txt files
# are all newer than the requirements.in's. To force it to
# remake the files you can either do `touch requirements/*.in`
# or run make with the --always-make option.
$ make requirements
make: Nothing to be done for 'requirements'.

# Let's delete .tox and force it to recompile all the
# requirements files (this will also force it to create all
# the venvs from scratch since we deleted .tox).
# This should not result in any changes since we haven't 
# changed any of the .in files and we're not passing
# args=--upgrade.
$ rm -rf .tox
$ make --always-make requirements

# Let's delete .tox again and this time delete all the
# requirements.txt files as well and force it to remake both
# the venvs the txt files from scratch.
# This *will* result in changes since it will upgrade all
# packages to their latest versions.
$ rm -rf .tox requirements/*.txt
$ make requirements

# And of course other make commands like `make sure` should
# be capable of creating the venvs if they don't exist as 
# well.
$ rm -rf .tox
$ make sure
```

Makefile
```diff
@@ -174,7 +178,7 @@ requirements/lint.txt: requirements/tests.txt requirements/functests.txt require
 # requirements/*.in files from disk ($(wildcard requirements/*.in)) and replace
 # the .in's with .txt's.
 .PHONY: requirements requirements/
-requirements requirements/: $(foreach file,$(wildcard requirements/*.in),$(basename $(file)).txt)
+requirements requirements/: $(foreach file,$(wildcard requirements/*.in),$(basename $(file)).txt) requirements/checkformatting.txt
```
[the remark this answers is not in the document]
seanh: https://github.com/hypothesis/lms/pull/4191/commits/4374503903ce33f26c3875242d1092c747a047b6

seanh:
@jon-betts Ok, this is ready for re-review. Changes:

* Removed the special-casing of `checkformatting.txt`. Made it a normal file instead of a symlink and gave it a `checkformatting.in` that just does `-r format.txt` (which now means that this new dependency needs to be added to `Makefile` as well): https://github.com/hypothesis/lms/pull/4191/commits/4374503903ce33f26c3875242d1092c747a047b6

* Made a couple of tweaks to fix the cold-start problems: https://github.com/hypothesis/lms/pull/4191/commits/236beb510fcf239932d145c48eb29a157fc6a08f

  This should now work as expected:

  ```terminal
   $ rm -rf .tox requirements/*.txt
   $ make requirements
   ```

tox.ini
jon-betts: I think just biting the bullet and having a matching name is one of the things we should have been doing all along to reduce edge cases and complexity.

jon-betts (commented):
I've tested this, and it does seem to work. There's definitely some rough edges here, some I suspect we'll be able to file off, and others which we seem to have to resign ourselves to as a result of using make.

Having to list the dependency arrangement of the files seems like a shame as the files already do this for themselves. This gives another area for template divergence and the possibility of drift between the files and the Makefile. But we don't rely on this very often and when you spotted it it would be a momentary annoyance if you could figure it out quickly then you'd fix it.

I'm conscious that this requires two new tools and repos which haven't been reviewed yet. A quick scan shows they don't have tests either. I'd be happier to give a more enthusiastic yes if we could get those other items up to our usual standards as well.

tox.ini
seanh: Yeah, avoiding special cases is a good idea.

This is something I've realised over time. The way that tox naturally wants to work is one requirements file pair and corresponding venv per "command": `tox -e lint` has `requirements/lint.{in,txt}` and `.tox/lint/`, etc.

More than once we've attempted to deviate from this. For example it has often been pointed out as "silly" that there's a `.tox/coverage/` venv with just `coverage` installed in it (for running the `coverage report` command) when `coverage` already has to be installed in the `tests` env in order to run the tests with coverage collection. But this is not silly:

* It means that to print out the coverage report you just do `tox -e coverage` instead of some painful long command like `tox -e tests --run-command 'coverage report'`
* It means that we can inform tox of the dependency of the `coverage` env on the `tests` env so that if you run `tox --parallel -e checkformatting,lint,tests-py{38,39,310},coverage,functests` tox will run everything in parallel but will know not to begin the coverage report until the unit tests for all versions of Python are finished (`make sure` does this)

* So there's a "pointless" `coverage.{in,txt}` file pair and `.tox/coverage/` venv. It's not in fact pointless at all. And what harm does it cause? Are you running out of hard drive space? (The `coverage` venv is 21MB and does not take a significant amount of time to create.)

Similarly we wanted to avoid having both `format.{in,txt}` and `checkformatting.{in,txt}` even though we did have separate `.tox/format/` and `.tox/checkformatting/` venvs. All that did was make things awkward.

Let tox be tox

seanh:
> Having to list the dependency arrangement of the files seems like a shame as the files already do this for themselves. This gives another area for template divergence and the possibility of drift between the files and the Makefile. But we don't rely on this very often and when you spotted it it would be a momentary annoyance if you could figure it out quickly then you'd fix it.

Worth noting that this is to do with the `make requirements` that's on `main` not to do with what this PR adds.

Repeating your dependencies in `Makefile` is pretty typical I think (in typical `Makefile`'s for C or C++ projects etc). I think this is going to be okay: the addition of `checkformatting.{in,txt}` is almost the only time we've changed the set of requirements files or the dependencies between them in the several years since this stuff was first moved into tox. The only other example I can think of is the addition of `bddtests.{in,txt}` in LMS.

Nonetheless I think it could be solved if we wanted to, though I don't propose doing that in this PR. I think we could get `make` to read the list of dependencies for each requirements file from the output of a script. Something like this in `Makefile` (not tested!):

```make
requirements/%.txt: $(shell bin/get_rqmnts_file_prereqs $@)
```

`bin/get_rqmnts_file_prereqs` would be a script that takes a requirements file name as argument and returns the list of other requirements files that it directly depends on (i.e. finds all the `-r`'s in a given requirements file). It would **not** need to find and read those dependency requirements files: `make` will figure out the transitive dependencies itself.

I don't think it'd be worth doing though because I don't think there's a standard library for parsing requirements files and I'd rather not write our own code to parse them. Although the script should be fairly simple I guess.

Alternatively we could forget about `make requirements` only re-compiling the file(s) that need to be re-compiled and have it just re-compile all the files in the right order every time. This would be slower in those cases where not all the files need to be re-compiled, but it'd be much simpler to implement. The entire implementation of such a `make requirements` would be something like:

```make
.PHONY: requirements
requirements: python
  @touch -a dev.txt
  @tox -qe dev --run-command 'pip --quiet --disable-pip-version-check install pip-tools'
  @tox -qe dev --run-command 'pip-compile --allow-unsafe --quiet requirements.txt'
  @tox -qe dev --run-command 'pip-compile --allow-unsafe --quiet dev.txt'
  @touch -a format.txt
  @tox -qe format --run-command 'pip --quiet --disable-pip-version-check install pip-tools'
  @tox -qe format --run-command 'pip-compile --allow-unsafe --quiet format.txt'
  ...
```

(You could remove the repetition from that at the cost of some cleverness: variables or a function or loop.)

> I'm conscious that this requires two new tools and repos which haven't been reviewed yet. A quick scan shows they don't have tests either. I'd be happier to give a more enthusiastic yes if we could get those other items up to our usual standards as well.

Discussed on Slack: I'll arrange PRs to get these reviewed...

seanh:
Here's tox-faster ready for review: https://github.com/hypothesis/tox-faster/pull/7

I'll do the same with pip-sync-faster...

seanh:
> Nonetheless I think it could be solved if we wanted to, though I don't propose doing that in this PR. I think we could get `make` to read the list of dependencies for each requirements file from the output of a script. Something like this in `Makefile` (not tested!):
> 
> ```makefile
> requirements/%.txt: $(shell bin/get_rqmnts_file_prereqs $@)
> ```
> 
> `bin/get_rqmnts_file_prereqs` would be a script that takes a requirements file name as argument and returns the list of other requirements files that it directly depends on (i.e. finds all the `-r`'s in a given requirements file). It would **not** need to find and read those dependency requirements files: `make` will figure out the transitive dependencies itself.

I realised that all you need for this `get_rqmnts_file_prereqs` command is a `grep`. Here's a PR to implement it: https://github.com/hypothesis/lms/pull/4231

jon-betts (approved):
The other items are being reviewed / approved. We aren't going to not do this, but those tools can change independently. So I don't see any reason not to go ahead with this.

We'll no doubt suck it and see any issues pretty quick.

seanh:
Rebased and squashed

seanh:
Merging this now that pip-sync-faster has been reviewed as well

PR #4192, opened 2022-07-07
Suggest URLs for new registrations

marcospri:
This add a "suggest" button for the URL for new registrations. This keeps the option to add anything, avoid the need for any JS and/or extra steps.

# Testing 

- Go to http://localhost:8001/admin/registration
- Use issuer `https://blackboard.com`, get suggestions.
- Submit the form, the suggestions get used to create the registration.


lms/models/lti_registration.py
```diff
@@ -37,3 +39,62 @@ class LTIRegistration(CreatedUpdatedMixin, BASE):
     application_instances = sa.orm.relationship(
         "ApplicationInstance", back_populates="lti_registration"
     )
+
+    @staticmethod
+    def urls(issuer, client_id):
```
marcospri, line 44: Does this belong to each product?

Not sure, there's not request.product here in the admin pages.

lms/models/lti_registration.py
```diff
@@ -37,3 +39,62 @@ class LTIRegistration(CreatedUpdatedMixin, BASE):
...
+            key_set_url = (
+                urlparse("https://brightspace.com/d2l/.well-known/jwks")
+                ._replace(netloc=issuer_host)
+                .geturl()
+            )
+            token_url = (
+                urlparse("https://brightspace.com/core/connect/token")
+                ._replace(netloc=issuer_host)
+                .geturl()
+            )
```
jon-betts, line 76:
```suggestion
            auth_login_url = f"https://{issuer_host}/d2l/lti/authenticate"
            key_set_url = f"https://{issuer_host}/d2l/.well-known/jwks"
            token_url = f"https://{issuer_host}/core/connect/token"
```

What's the reason for using `urlparse` here?

lms/models/lti_registration.py
```diff
@@ -37,3 +39,62 @@ class LTIRegistration(CreatedUpdatedMixin, BASE):
...
+                urlparse("https://moodlecloud.com/mod/lti/certs.php")
+                ._replace(netloc=issuer_host)
+                .geturl()
+            )
+
+        return {
+            "auth_login_url": auth_login_url,
+            "key_set_url": key_set_url,
+            "token_url": token_url,
+        }
```
jon-betts, line 100: I don't think the model is the place for this. This really seems like something that should be in the admin view as it's only used there, not in the normal run of things. 

In general I think we're moving away from adding code to the models, especially as here it's a static method which doesn't rely on any details for the instance.

jon-betts: The amount of config in the code here makes me a bit uncomfortable. This might be a personal style thing, but I'd prefer a configuration that could be updated and some code which will then use that config.

I wanted to try this myself to make sure it's possible and came up with the following (just a suggestion):

```python
URL_SUGGESTIONS = {
    "https://blackboard.com": {
        "auth_login_url": "https://developer.blackboard.com/api/v1/gateway/oidcauth",
        "key_set_url": "https://developer.blackboard.com/api/v1/management/applications/{client_id}/jwks.json",
        "token_url": "https://developer.blackboard.com/api/v1/gateway/oauth2/jwttoken",
    },
    ".instructure.com": {
        "auth_login_url": "https://canvas.instructure.com/api/lti/authorize_redirect",
        "key_set_url": "https://canvas.instructure.com/api/lti/security/jwks",
        "token_url": "https://canvas.instructure.com/login/oauth2/token",
    },
    ".brightspace.com": {
        "auth_login_url": "https://{issuer_host}/d2l/lti/authenticate",
        "key_set_url": "https://{issuer_host}/d2l/.well-known/jwks",
        "token_url": "https://{issuer_host}/core/connect/token",
    },
    ".moodlecloud.com": {
        "auth_login_url": "https://{issuer_host}/mod/lti/auth.php",
        "key_set_url": "https://{issuer_host}/mod/lti/certs.php",
        "token_url": "https://{issuer_host}/mod/lti/certs.php",
    },
}

def urls(issuer, client_id):
    for suffix, url_templates in URL_SUGGESTIONS.items():
        if issuer.endswith(suffix):
            issuer_host = urlparse(issuer).netloc

            return {
                key: template.format(issuer_host=issuer_host, client_id=client_id)
                for key, template in url_templates.items()
            }

    return {"auth_login_url": None, "key_set_url": None, "token_url": None}
```

This config could then live in a JSON file if we wanted to keep the code slim.

lms/models/lti_registration.py
jon-betts: There wouldn't be a product in play as you say, but it could be something you get from each one as an array. I think that might be overly obscure though. I've made some suggestions below to how we could get this out of the code though.

lms/views/admin/lti_registration.py
```diff
@@ -32,17 +80,35 @@ def registrations(self):
...
+            response.status_code = 400
+            return response
+
+        return render_to_response(
+            "lms:templates/admin/registration.new.html.jinja2",
+            self._get_url_suggestions(
+                self.request.params["issuer"], self.request.params["client_id"]
+            ),
+            request=self.request,
+        )
```
jon-betts, line 103: I could be missing something, but it looks like we're using `render_to_response` because we are modifying the status code? I think we can work around this using `self.request.response` which will allow us to modify the status code:

I think this means we can do this:

```suggestion
    @view_config(
        route_name="admin.registration.suggest_urls",
        request_method="POST",
        renderer="lms:templates/admin/registration.new.html.jinja2",
    )
    def suggest_lms_urls(self):
        if flash_validation(self.request, LTIRegistrationBaseSchema):
            self.request.response.status_code = 400
            return {}

        return self._get_url_suggestions(
            self.request.params["issuer"], self.request.params["client_id"]
        )
```

If we also do that for the method below, then I think every view here has the same renderer, so we could move it to the `view_defaults` for the class.

jon-betts (approved):
This is looking good. This kind of stuff is a godsend for users. I also like the schema based validation, rather than implementing our own much worse version of a validator.

I initially thought maybe we should lean in and make this a two page wizard type of thing, GDS style, but while we are here to help our users, they don't need that much hand holding and seeing all the data side by side might be helpful. Also in cases where there is no suggestion, this style will be much better, as the user can input the values into all the fields in one go. 

We should prompt people (just in slack, not the product) to let us know if there are missing suggestions we could be making, as the system should be quite easy to extend with new suggestions.

There are a couple of spots where I think we could tighten things up potentially:

 * We have an inheritance stack of schema, but I'm not 100% clear and whether we need the middle layer
 * We use `render_to_response` quite a bit, but unless I'm missing something I think we only use it to change the stats code we return, which can be done with `request.response`. This could simplify things quite a bit, but would require a test change.

lms/views/admin/__init__.py
```diff
@@ -17,14 +19,11 @@ def index(request):
...
-            "errors",
-        )
+def flash_validation(request, schema):
+    try:
+        schema(request).parse()
+    except ValidationError as err:
+        request.session.flash(err.messages["form"], "validation")
         return True
 
     return False
```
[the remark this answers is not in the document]
marcospri: It's us wrapping marshmallow error handling

https://github.com/hypothesis/lms/blob/dfbaeea67272fb442a10b924934a41e5437722e8/lms/validation/_exceptions.py#L11

https://github.com/hypothesis/lms/blob/dfbaeea67272fb442a10b924934a41e5437722e8/lms/validation/_base.py#L94

lms/views/admin/lti_registration.py
```diff
@@ -1,11 +1,59 @@
...
+class LTIRegistrationURLSSchema(PyramidRequestSchema):
+    location = "form"
+
+    auth_login_url = fields.URL(required=True)
+    key_set_url = fields.URL(required=True)
+    token_url = fields.URL(required=True)
+
+
+class LTIRegistrationSchema(LTIRegistrationBaseSchema, LTIRegistrationURLSSchema):
+    location = "form"
```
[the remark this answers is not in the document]
marcospri: I think I refactored the usage out and left that in :facepalm:. Removed.

PR #4194, opened 2022-07-08
Explicilty disable autoexpiring sessions

marcospri:
This achieves the desired result meant in #4184 

# Testing

- Go to http://localhost:8001/admin and login, close the browser, open again, still logged in (not changed in this PR)

- Leave the browser session open now for over 20 minutes. The session survives.

### DIFF
--- a/lms/session.py
+++ b/lms/session.py
@@ -14,17 +14,18 @@ def includeme(config):
 
     config.set_session_factory(
         SignedCookieSessionFactory(
+            # https://docs.pylonsproject.org/projects/pyramid/en/latest/api/session.html#pyramid.session.SignedCookieSessionFactory
             secret=config.registry.settings["session_cookie_secret"],
             secure=secure,
             # ``httponly=True`` is recommended by the Pyramid docs to protect
-            # the cookie from cross-site scripting vulnerabilities, see:
-            # https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/sessions.html
+            # the cookie from cross-site scripting vulnerabilities.
             httponly=True,
             # 30 days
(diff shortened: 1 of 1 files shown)


lms/session.py
```diff
@@ -14,17 +14,18 @@ def includeme(config):
 
     config.set_session_factory(
         SignedCookieSessionFactory(
+            # https://docs.pylonsproject.org/projects/pyramid/en/latest/api/session.html#pyramid.session.SignedCookieSessionFactory
             secret=config.registry.settings["session_cookie_secret"],
             secure=secure,
             # ``httponly=True`` is recommended by the Pyramid docs to protect
-            # the cookie from cross-site scripting vulnerabilities, see:
-            # https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/sessions.html
```
marcospri, line 21: These links all point to the same place. Added a link to where all this params are declared at the top.

lms/session.py
```diff
@@ -14,17 +14,18 @@ def includeme(config):
...
+            # the cookie from cross-site scripting vulnerabilities.
             httponly=True,
             # 30 days
             max_age=60 * 60 * 24 * 30,
+            # Disable autoexpiring sessions. Rely only on max_age
+            timeout=None,
```
marcospri, line 26: In a previous PR:

- Added a value for max_age

That could be tested checking the session survived a browser (chrome) restart.

- Removed the autoexpiring parameters.

That didn't have any effect because the values removed were actually the defatuls.


Now passing None has the desired effect:

```A number of seconds of inactivity before a session times out. If None then the cookie never expires. ```

PR #4195, opened 2022-07-08
Improved LTI version detection

marcospri:
Some niceties around LTI version.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -202,7 +202,7 @@ class ApplicationInstance(BASE):
         The return values (LTI-1p0, "1.3.0) are the same the spec defines
         and will match the version parameter on lti launches.
         """
-        if self.lti_registration_id:
+        if self.lti_registration_id and self.deployment_id:
             return "1.3.0"
 
         return "LTI-1p0"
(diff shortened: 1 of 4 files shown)


lms/models/application_instance.py
```diff
@@ -202,7 +202,7 @@ def lti_version(self) -> str:
         The return values (LTI-1p0, "1.3.0) are the same the spec defines
         and will match the version parameter on lti launches.
         """
-        if self.lti_registration_id:
+        if self.lti_registration_id and self.deployment_id:
```
marcospri, line 205: We should never find this one without the other in our DB but this is more accurate.

lms/templates/admin/instance.html.jinja2
```diff
@@ -3,6 +3,11 @@
 {% extends "lms:templates/admin/base.html.jinja2" %}
 {% block header %}
     Application instance {{instance.id}}
+    {% if instance.lti_version == "1.3.0" %}
```
marcospri, line 6: Display a hint of the version of an instance

jon-betts (approved):
I've tried this out locally and it works as intended. I think "LTI 1.1" is easier to read than "LTI1.1".

Issue #4197, opened 2022-07-08
Provide enough data to call H from a modified LTI launch view


jon-betts:

jon-betts:
I think Noodle have used the provided tools to try and exchange the token, but it doesn't seem to be working so far. I'm waiting on some more details to see if I can help.

PR #4205, opened 2022-07-08
WIP - View providing context and connection details

jon-betts:
For:


jon-betts:
This has probably served it's purpose now, as there is enough context in the merged code to work out how to extend it

Issue #3923, opened 2022-05-05
Remove unused consumer_key columns


marcospri:
After 

https://github.com/hypothesis/lms/pull/3779
https://github.com/hypothesis/lms/pull/3780
https://github.com/hypothesis/lms/pull/3781

we don't use consumer_key on GroupInfo, GradingInfo and OAuth2Token. Remove those columns.

PR #4210, opened 2022-07-11
Remove unused columns

marcospri:
For https://github.com/hypothesis/lms/issues/3923

After the relationship with application_instance was added back we can now remove the leftover columns.

For group_info we added code in the loading part of anaylics that would regenerate this column: https://github.com/hypothesis/analytics/blob/main/refresh/load.sh#L46

### DIFF
--- a/lms/migrations/versions/a9990b0efb3f_remove_consumer_key.py
+++ b/lms/migrations/versions/a9990b0efb3f_remove_consumer_key.py
@@ -1,0 +1,35 @@
+"""
+Remove unused consumer_key columns.
+
+Revision ID: a9990b0efb3f
+Revises: 1e146996cca6
+Create Date: 2022-07-11 12:28:13.684023
+
+"""
+import sqlalchemy as sa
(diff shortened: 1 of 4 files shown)


lms/migrations/versions/a9990b0efb3f_remove_consumer_key.py
```diff
@@ -0,0 +1,35 @@
...
+        sa.Column("consumer_key", sa.VARCHAR(), autoincrement=False, nullable=True),
+    )
+    op.add_column(
+        "lis_result_sourcedid",
+        sa.Column("oauth_consumer_key", sa.TEXT(), autoincrement=False, nullable=True),
+    )
+    op.add_column(
+        "group_info",
+        sa.Column("consumer_key", sa.VARCHAR(), autoincrement=False, nullable=True),
+    )
```
jon-betts, line 35: We might need some due diligance here to check if any reporting is relying on this as a join condition.

Issue #4088, opened 2022-06-15
Title is incomplete for some JSTOR articles


robertknight:
The title displayed in the JSTOR picker for some articles is incomplete:

<img width="734" alt="JSTOR article partial title" src="https://user-images.githubusercontent.com/2458/173816899-6b1b4db1-fd48-4f46-9c74-aba38f4aff76.png">

The article link here is https://www.jstor.org/stable/10.3138/j.ctt1whm96v.10.

The metadata returned from the JSTOR API at https://labs.jstor.org/api/anno/metadata/10.3138/j.ctt1whm96v.10 is:

```json
{
  "id": "898a889a-4069-3678-a9ad-b4c832533738",
  "doi": "10.3138/j.ctt1whm96v.10",
  "score": 63.22056,
  "raw_type": "chapter",
  "type": "chapter",
  "title": [
    "Talking about History:"
  ],
  "subtitle": [
    "The Encomium Emmae reginae and the Court of Harthacnut"
  ],
  "translated_title": [],
  "author": [
    "ELIZABETH M. TYLER"
  ],
  "citation_line": "2017",
  "is_review": null,
  "reviewed_works": [],
  "serial_numbers": [],
  "journal": [],
  "jid": [],
  "iid": [],
  "book_volume": null,
  "edition": [],
  "editor": [],
  "publisher": [
    {
      "code": "utp",
      "publisher": "University of Toronto Press",
      "role": "PUBLISHER"
    },
    {
      "code": "utp",
      "publisher": "University of Toronto Press",
      "role": "LICENSOR"
    }
  ],
  "tokens": [
    "123959834688505",
    "123959837174993",
    "123959837460917",
    "123959837464124",
    "123959837464482",
    "16124",
    "17000000004",
    "17000000006",
    "17000000011",
    "17000000013",
    "17000000017",
    "17000000019",
    "17000000027",
    "17000001000",
    "5",
    "57397011",
    "99999002664904",
    "99999002713015",
    "99999002908899",
    "99999002908905",
    "99999002908906",
    "99999003613385",
    "99999004164145"
  ],
  "flags": [
    "pdf",
    "page_scan",
    "free"
  ],
  "resource_type": "chapter",
  "book_doi": "10.3138/j.ctt1whm96v",
  "publication_year": 2017,
  "ab_segment": "0/basic_search_gsv2/control",
  "human_readable_type": "Book Chapter",
  "cty": "chapter",
  "la": [
    "eng"
  ],
[...]

Issue #4087, opened 2022-06-15
Some JSTOR article titles contain HTML tags


robertknight:
Some JSTOR article titles contain HTML tags, which get escaped and rendered as text:

<img width="723" alt="JSTOR article title tags" src="https://user-images.githubusercontent.com/2458/173812612-8f860c54-152c-4e4e-87bd-6ece0a852587.png">

The example article here is https://www.jstor.org/stable/10.1525/as.2003.43.3.423.

The simplest thing we could do here would be to just strip any tags on the backend so that the frontend doesn't have to worry about them.

So far the tags I have seen are: `<em>`.

robertknight:
Some other tags seen in a [big list of titles of previously annotated JSTOR documents](https://gist.github.com/robertknight/d7cb89dfff84085c2951191c7af209b7?h=1):

```
# <span> with attributes
10.1086/675345: Maryland <span class="small-caps">v</span> King: Terry <span class="small-caps">v</span> Ohio Redux

20715416: Volumes of Transnational Vengeance: Fixing Race and Feminism on the Way to <italic>Kill Bill</italic>
40664126: <strong>structural domination and structural freedom: a feminist perspective</strong>

# <sup>. Note also new lines which we should probably normalize
42708472: Usage politique des corps et rituel de l'exécution publique à Paris, XVII
          <sup>e</sup>
          -XVIII
          <sup>e</sup>
          siècles
```

Issue #4049, opened 2022-06-06
Improve presentation of JSTOR articles that are reviews of other works


robertknight:
Some JSTOR articles are reviews of other works and do not have a title of their own. eg. https://www.jstor.org/stable/331473.

For these the JSTOR assignment picker added in https://github.com/hypothesis/lms/pull/4026 does present a thumbnail but no useful information about the article:

<img width="728" alt="JSTOR picker no title" src="https://user-images.githubusercontent.com/2458/172167953-c7575a4e-1417-403f-b1da-9f651728e0ef.png">

The metadata response we receive from the JSTOR API for the above article looks like:

(Update: The metadata has changed. See comments below.)

```json
{
  "id": "6898ee6b-e080-3b80-b8a2-2fc3b71a7015",
  "doi": "10.2307/331473",
  "title": [],
  "subtitle": [],
  "author": [
    "Aurelio M. Espinosa"
  ],
  "citation_line": "<cite>Hispania</cite>, Vol. 9, No. 3 (May, 1926), pp. 191-193",
  "journal": [
    "Hispania"
  ]
}
```

Here is how the JSTOR article page presents it:

<img width="437" alt="JSTOR article no title" src="https://user-images.githubusercontent.com/2458/172167853-f792dc0f-252e-4133-8e42-6cc242a13ed9.png">

PR #4212, opened 2022-07-12
Adapt to new JSTOR metadata API schema

robertknight:
Adapt to the new JSTOR metadata API schema, as described in https://hypothes-is.slack.com/archives/C02T75RKBTK/p1655930468948019 and in the OpenAPI docs at https://labs.jstor.org/api/anno/docs#/default/get_metadata_new_metadata_new__doi__get.

 - Expect "title" and "subtitle" fields to be strings rather than
   single-item lists
 - Expect "reviewed_works" field to be an array of titles rather than
   items
 - Expect the "title" and "subtitle" fields to be used for more item
   types, reducing the need for us to check different fields depending
   on the item type.

The new API has been temporarily exposed under the `/metadata-new`
endpoint. It will be exposed under the original `/metadata` endpoint in
future.

A set of test cases have been collated from existing issues into a list at
https://gist.github.com/robertknight/a2b045b717d6c7dda18879df91c16e90.
Some of these could not be tested because the API server is returning
5xx responses. I assume these responses will be similar to the items
that do work.

**Testing:**

Try entering the article IDs from the list of test cases, you should see a sensible title displayed. If there is a thumbnail (not always available), and the title is clearly visible in it, the two should match.

* Research article: `10.2307/26202019`
* Primary source image: `10.2307/community.22669911`
* Item with subtitle: `10.3138/j.ctt1whm96v.10` (see https://github.com/hypothesis/lms/issues/4088). 
   * Displayed title should be `{title}: {subtitle}`
* Item with HTML tags in title: `10.1525/as.2003.43.3.423` (see https://github.com/hypothesis/lms/issues/4087). 
   * Displayed title should not include HTML tags.
* Review of another work: `10.2307/331473` (see https://github.com/hypothesis/lms/issues/4049). 
   * Displayed title should be Review: {title of original work}

In the case of book chapters, you will see the chapter title rather than the book title. These chapter titles can sometimes be somewhat generic ("Intro", "Appendix"). We might want to include some details from the containing book in this case. That's outside the scope of this PR.

* Book chapter: `10.7249/mg876rc.10`
* Book: `10.7249/mg876rc`
* Primary source book: `10.2307/community.30528851`

### DIFF
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -16,7 +16,7 @@ class JSTORMetadataSchema(RequestsResponseSchema):
     """
     Response schema for `/metadata/{doi}` endpoint in the JSTOR API.
 
-    See https://hypothes-is.slack.com/archives/C02T75RKBTK/p1655930468948019.
+    See https://labs.jstor.org/api/anno/docs
     """
 
     title = fields.Str()
--- a/lms/services/jstor.py
+++ b/lms/services/jstor.py
@@ -1,10 +1,9 @@
 from datetime import timedelta
 from html.parser import HTMLParser
-from typing import Optional
 from urllib.parse import quote
 
 import requests
-from marshmallow import EXCLUDE, Schema, fields
+from marshmallow import fields
 
 from lms.services.exceptions import ExternalRequestError
 from lms.services.http import HTTPService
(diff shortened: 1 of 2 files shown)


robertknight:
In testing this I found an issue with handling IDs or URLs that have leading/trailing spaces. This is handled in https://github.com/hypothesis/lms/pull/4213.

lms/services/jstor.py
```diff
@@ -168,38 +153,26 @@ def _api_request(self, path_template, doi, params=None) -> requests.Response:
...
 
             # Some articles have a subtitle which needs to be appended for the
             # title to make sense.
-            if subtitles := metadata.get("subtitle"):
-                subtitle = subtitles[0]
+            if subtitle := metadata.get("subtitle"):
+                # Some titles include a delimiter, some do not.
+                if title.endswith(":"):
+                    title = title[:-1]
+                title = f"{title}: {subtitle}"
```
jon-betts, line 172:
```suggestion
                title = f"{title.rstrip(':')}: {subtitle}"
```

jon-betts (approved):
Yep this looks good to me and a _lot_ simpler than before. I've made one suggestion for a small tidy up.

PR #4213, opened 2022-07-12
Ignore leading and trailing space in JSTOR input field

robertknight:
This avoids an error if the article ID was copied and pasted from a selection
that had some extra spaces.


lyzadanger (approved):
Straightforward! Ship it.

PR #4222, opened 2022-07-13
Add a gateway end-point which provides H API connection details

jon-betts:
For:

This adds a version of the gateway end-point which only provides information sufficient for connecting to H.

This version enforces that the GUID provided matches the credentials we expect for the application instance. I think this really should happen in our security layer but due to the way we try policies in order this might be a little tricky.

### Testing notes

 * Check out and try the tools in: https://github.com/hypothesis/lms/pull/4224

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -29,6 +29,15 @@ def includeme(config):  # pylint:disable=too-many-statements
 
     config.add_route("api.assignments.create", "/api/assignment", request_method="POST")
 
+    # The gateway end-point is a bit of an oddball as it uses LTI launch as the
+    # authentication method, and so requires the LTILaunch resource
+    config.add_route(
+        "api.gateway.h.lti",
+        "/api/gateway/h/lti",
+        request_method="POST",
+        factory="lms.resources.LTILaunchResource",
+    )
+
     config.add_route(
         "blackboard_api.oauth.authorize",
         "/api/blackboard/oauth/authorize",
--- a/lms/views/api/gateway.py
+++ b/lms/views/api/gateway.py
@@ -1,7 +1,9 @@
 from marshmallow import fields
 from marshmallow.validate import Equal
+from pyramid.exceptions import HTTPForbidden
 from pyramid.view import view_config
 
+from lms.models import ReusedConsumerKey
 from lms.security import Permissions
 from lms.validation import LTIV11CoreSchema
 
@@ -30,12 +32,23 @@ class GatewayLTISchema(LTIV11CoreSchema):
     route_name="api.gateway.h.lti",
     schema=GatewayLTISchema,
 )
-def h_lti(request):
+def h_lti(context, request):
     """
     Provide tokens and information to allow customers to query H.
 
     We expect the user to authenticate with us using an LTI launch.
     """
+
+    # Ensure no funny business is going on trying to access content out of the
+    # current application instance scope.
+    try:
+        context.application_instance.check_guid_aligns(
+            request.lti_params["tool_consumer_instance_guid"]
+        )
+    except ReusedConsumerKey as err:
+        raise HTTPForbidden(
+            "Claimed `tool_consumer_instance_guid` does not match credentials."
+        ) from err
 
     # Add API end-point details
     h_api_url = request.registry.settings["h_api_url_public"]
--- a/lms/views/api/gateway.py
+++ b/lms/views/api/gateway.py
@@ -1,13 +1,34 @@
+from marshmallow import fields
+from marshmallow.validate import Equal
 from pyramid.view import view_config
 
 from lms.security import Permissions
+from lms.validation import LTIV11CoreSchema
+
+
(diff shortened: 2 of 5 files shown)


tests/functional/oauth1.py
```diff
@@ -0,0 +1,82 @@
...
+from tests import factories
+
+__all__ = (
+    "do_lti_launch",
+    "oauth1_credentials",
+    "application_instance",
+    "oauth1_sign_data",
+)
+
+
```
jon-betts, line 19: This stack of fixtures works differently from the other ones. In this case you call the launch fixture and it signs the params for you.

In the previous set, you have to sign them first, which is awkward because you can only sign them with the URL. This makes the existing setup much more clumsy.

It's a potential next step to try and refactor the existing tests to use this sort of arrangement.

tests/functional/api/gateway_test.py
```diff
@@ -0,0 +1,89 @@
...
+            "context_id": "321",
+            "lti_version": "LTI-1p0",
+            "lti_message_type": "basic-lti-launch-request",
+        }
+
+    @pytest.fixture
+    def gateway_launch(self, do_lti_launch):
+        return partial(
+            do_lti_launch,
+            "http://localhost/api/gateway/h/lti",
```
jon-betts, line 87: In other tests this isn't fully qualified. I couldn't get a path alone to work

tests/functional/api/gateway_test.py
```diff
@@ -0,0 +1,89 @@
...
+                        "Content-Type": "application/x-www-form-urlencoded",
+                    },
+                    "method": "POST",
+                    "url": "https://example.com/api/token",
+                },
+                "list": {
+                    "headers": {"Accept": "application/vnd.hypothesis.v2+json"},
+                    "method": "GET",
+                    "url": "https://example.com/api/",
+                },
```
jon-betts, line 33: Bit of duplication here, but it's easier than lots of Any.dict() stuff. Do you think we should cut this down?

lms/routes.py
```diff
@@ -29,6 +29,15 @@ def includeme(config):  # pylint:disable=too-many-statements
 
     config.add_route("api.assignments.create", "/api/assignment", request_method="POST")
 
+    # The gateway end-point is a bit of an oddball as it uses LTI launch as the
+    # authentication method, and so requires the LTILaunch resource
+    config.add_route(
+        "api.gateway.h.lti",
+        "/api/gateway/h/lti",
```
jon-betts, line 36: I wanted to leave the possibility open of:

 * Providing access to other products
 * Providing access to `h` in ways other than `lti` launch

lms/routes.py
```diff
@@ -29,6 +29,15 @@ def includeme(config):  # pylint:disable=too-many-statements
 
     config.add_route("api.assignments.create", "/api/assignment", request_method="POST")
 
+    # The gateway end-point is a bit of an oddball as it uses LTI launch as the
+    # authentication method, and so requires the LTILaunch resource
```
marcospri, line 33: Does it require the context authentication of convince?

I reckon the security part will be `LTISecurityPolicy` jobs and having the resource here will make `contex.application_instance` and friend available.

The security stuff it's a bit convoluted so I might be getting it wrong.

lms/views/api/gateway.py
```diff
@@ -0,0 +1,82 @@
...
+            "Claimed `tool_consumer_instance_guid` does not match credentials."
+        ) from err
+
+    # Add API end-point details
+    h_api_url = request.registry.settings["h_api_url_public"]
+    return {
+        "h_api": {
+            # These sections are arranged so you can use
+            # `requests.Request.request(**data)` and make the correct request
+            "list": {
```
marcospri, line 59: Is `list` a name on the `H` side? The list API?

Or `list` here means `list stuff` in general?

lms/routes.py
jon-betts: I _think_ the context stuff happens after authentication. This makes `context.application_instance` available to us easily in the view.

lms/views/api/gateway.py
jon-betts: Yeah, this is vague. It's not a schema exactly. You can see the results here:

https://hypothes.is/api/

I think "list_endpoints" might be the sweet spot?

lms/views/api/gateway.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    # Ensure no funny business is going on trying to access content out of the
+    # current application instance scope.
+    try:
+        context.application_instance.check_guid_aligns(
+            request.lti_params["tool_consumer_instance_guid"]
+        )
+    except ReusedConsumerKey as err:
+        raise HTTPForbidden(
```
marcospri, line 49: I reckon this message is not displayed to end users. It's replaced by the generic exception view message.

marcospri (approved):

PR #4225, opened 2022-07-14
Fallback for product family detection based on the application instance

marcospri:
Guess the product family more accurately leaning on the information stored in the DB

This will work in all grading endpoints in https://github.com/hypothesis/lms/blob/b3b6170240841a3fd7fdf21bc8403fa0d878f2c6/lms/views/api/lti.py where we always guess "UNKOWN" now.

### DIFF
--- a/lms/product/factory.py
+++ b/lms/product/factory.py
@@ -1,6 +1,7 @@
 from lms.product.blackboard import Blackboard
 from lms.product.canvas import Canvas
 from lms.product.product import Product
+from lms.services.application_instance import ApplicationInstanceNotFound
 
 _PRODUCT_MAP = {product.family: product for product in (Blackboard, Canvas)}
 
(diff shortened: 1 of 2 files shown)


lms/product/factory.py
```diff
@@ -14,7 +15,7 @@ def get_product_from_request(request) -> Product:
     return product
 
 
-def _get_family(request):
+def _get_family(request):  # pylint:disable=too-many-return-statements
```
marcospri, line 18: This could be refactor to avoid pylint's complain but I think the method reads well as it is, try different methods to get the family, early return when one works.

jon-betts:
I suspect there's something we _could_ do with the context factory argument to [add_route()](https://docs.pylonsproject.org/projects/pyramid/en/2.0-branch/api/config.html#pyramid.config.Configurator.add_route) using a closure / `functools.partial` around something, but I think that would be much larger and harder to understand.

I do think we might want to do something to limit the routes this applies to though so we remember. Or at least a comment

lms/product/factory.py
```diff
@@ -34,4 +35,13 @@ def _get_family(request):
     if "custom_canvas_course_id" in request.lti_params:
         return Product.Family.CANVAS
 
-    return Product.Family.UNKNOWN
+    # Finally try to match using the stored family_code in the application instance
```
jon-betts, line 38: Ninja comment: I think we should point out this is actually only for the outcomes routes, and potentially consider limiting the times we apply this (like `if request.matched_route` above) to make it clear where this is for. We could expand it as necessary if we found we had to, which might for documentation of a sort of the places this is required.

Otherwise this would be very hard to prove backwards to only apply to that one place.

PR #4230, opened 2022-07-15
Add consumer tests to prove we meet our own schema

jon-betts:
For:

This PR includes consumer testing, which is something we don't do by name, which is to run tests which check the agreement you make with 3rd parties.

If we change our API in a way such that the documentation isn't valid any more (at least as far as the schema goes) this should result in a test failure.

This keeps us honest with the schema we provide to customers and will alert us of any changes.

### DIFF
--- a/tests/unit/lms/views/api/gateway_test.py
+++ b/tests/unit/lms/views/api/gateway_test.py
@@ -1,6 +1,9 @@
+import json
 from unittest.mock import create_autospec, sentinel
 
+import importlib_resources
 import pytest
+from jsonschema import Draft202012Validator
 from pyramid.httpexceptions import HTTPForbidden
 
 from lms.models import ReusedConsumerKey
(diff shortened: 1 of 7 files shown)


tests/unit/lms/views/api/gateway_test.py
```diff
@@ -45,12 +48,43 @@ def test_it_checks_for_guid_agreement(self, context, pyramid_request):
...
+    def test_schema_examples_are_valid(self, validator, schema):
+        for example in schema["examples"]:
+            validator.validate(example)
+
+    def test_gateway_output_matches_the_schema(
+        self, validator, context, pyramid_request
+    ):
+        response = h_lti(context, pyramid_request)
+
+        validator.validate(response)
```
marcospri, line 69: Yup, this is not a lot of work but checking we are returning something that looks like what we are saying we return will go along way to keep the docs up to date.

marcospri (approved):

PR #4231, opened 2022-07-16
Automate finding requirements file prerequisites

seanh:
Automate finding requirements file prerequisites

**Problem:** As pointed out in [a comment](https://github.com/hypothesis/lms/pull/4191#pullrequestreview-1037279596) `Makefile` currently contains a [list of the prerequisite requirements files](https://github.com/hypothesis/lms/blob/c15652ef25adfb039a54089f4d745bbabbd2c27b/Makefile#L157-L164) for each requirements file that has prerequisites. If we add a new requirements file with prerequisites or change the prerequisites of an existing requirements file we may forget to update this list in `Makefile` with confusing results.

**Solution:** This PR removes the hard-coded list of prerequisites from `Makefile` in favour of getting `make` to find the prerequisites itself by grepping for `-r` lines in the `requirements/*.in` files.

How this works
--------------

### `grep` and `sed`

It's fairly easy to find the list of prerequisites for a given requirements.txt file: you just need to grep its corresponding requirements.in file for lines like `-r foo.txt` or `-c bar.txt`:

```terminal
$ grep '^[[:blank:]]*\(-r\|-c\)[[:blank:]]*' requirements/foo.in
-r tests.txt
-rfunctests.txt
   -r   bddtests.txt
-c	constraints.txt
```

There's no need to further search those `*.txt` files to find out which files they import: we only need to inform `make` of the direct dependencies and it will build the full dependency tree itself.

You can pipe the output of this `grep` command to a `sed` command that gets rid of the whitespace and the actual `-r`'s and `-c`'s and gives you just the prerequisite file names:

```terminal
$ grep '...' requirements/foo.in | sed 's/^[[:blank:]]*\(-r\|-c\)[[:blank:]]*//'
tests.txt
functests.txt
bddtests.txt
constraints.txt
```

### `$(addprefix ...)` and `$(shell ...)`

We actually need `requirements/tests.txt` not just `tests.txt`: we need to add the `requirements/` prefix to each prerequisite. We could do that by piping to another `sed` but `make` actually has a `$(addprefix)` function (it's documented under [Functions for File Names](https://www.gnu.org/software/make/manual/make.html#File-Name-Functions) in the `make` docs). It takes a series of names separated by whitespace (newlines in our case), prepends the prefix to each of them, and returns the prefixed names separated by spaces. So this actually does two things that we need: it prepends the prefix and it joins our newline-separated strings into a single line of space-separated strings:

```make
find_prerequisites = $(addprefix requirements/,$(shell grep '...' "requirements/$*.in" | sed '...'))
```

The `$*` is a `make` [automatic variable](https://www.gnu.org/software/make/manual/make.html#Automatic-Variables) that expands to the "stem" of the target requirements.txt file that's currently being made: for `requirements/foo.txt` `$*` expands to `foo`.

`$(shell ...)` is `make`'s thing for shelling out, the `$(shell ...)` gets replaced by the output of the given shell command. See [The `shell` Function](https://www.gnu.org/software/make/manual/make.html#Shell-Function) in `make`'s docs.

### Putting it all together

So far all we've done is define a variable called `find_prerequisites` that contains our `make` code for finding the prerequisites for a given requirements.txt file. Nothing actually calls `find_prerequisites` yet. All we need to do is add it to the prerequisites list for the `requirements/%.txt` rule:

```make
.SECONDEXPANSION:
requirements/%.txt: requirements/%.in $$(find_references)
    <recipe goes here>
```

The `requirements/%.in` just tells `make` that each `requirements/%.txt` file depends on its corresponding `requirements/%.in` (so if the `%.in` has changed the `%.txt` must be re-compiled). We already had that in the `Makefile`.

The `$$(find_references)` calls our `find_references` command to add the list of `requirements.txt` files that the `%.in` imports to the `%.txt`'s prerequisites.

`.SECONDEXPANSION` is needed to make the `$*` in `find_references` work. Normally the list of prerequisites in a `Makefile` is processed very early in `make`'s processing long before `make` is actually making a particular target. `$*` isn't available in this early phase, it would not be expanded and `make` would pass a literal `*` to the shell. (See [How `make` Reads a Makefile](https://www.gnu.org/software/make/manual/make.html#Reading-Makefiles).)

The workaround is to activate `make`'s [Secondary Expansion](https://www.gnu.org/software/make/manual/make.html#Secondary-Expansion) feature which means that `make` will expand the prerequisites a second time just before it uses the rule to make a particular requirements file. During this second expansion `$*` does work.

Secondary expansion is also the reason for the `$$` in `$$(find_references)`: the first `$` gets removed by the first expansion leaving `$(find_references)` to be expanded by the secondary expansion.

Testing
-------

`--dry-run` is handy for testing because it just prints out the commands that it runs instead of actually running them. And `--assume-new` is a handy way to tell make to assume that a given file has been changed without having to actually change it.

```terminal
# If lint.in has changed it recompiles just lint.txt (nothing depends on lint.txt).
$ make --dry-run --assume-new requirements/lint.in requirements
touch -a requirements/lint.txt
tox -qe lint --run-command 'pip-compile --allow-unsafe --quiet  requirements/lint.in'

# lint.in depends on functtests.txt so if functests.in has changed then it first recompiles
# functests.txt then lint.txt.
$ make --dry-run --assume-new requirements/functests.in requirements
touch -a requirements/functests.txt
tox -qe functests --run-command 'pip-compile --allow-unsafe --quiet  requirements/functests.in'
touch -a requirements/lint.txt
tox -qe lint --run-command 'pip-compile --allow-unsafe --quiet  requirements/lint.in'

# {bddtests,dev,functests,tests}.in all depend on requirements.txt so if requirements.in has
# changed it will first recompile requirements.txt then {bddtests,dev,functests,tests}.txt.
# lint.in depends on {tests,functests,bddtests}.txt so after recompiling those it will then
# recompile lint.txt.
$ make --dry-run --assume-new requirements/requirements.in requirements
touch -a requirements/dev.txt
tox -qe dev --run-command 'pip-compile --allow-unsafe --quiet  requirements/requirements.in'
touch -a requirements/bddtests.txt
tox -qe bddtests --run-command 'pip-compile --allow-unsafe --quiet  requirements/bddtests.in'
touch -a requirements/dev.txt
tox -qe dev --run-command 'pip-compile --allow-unsafe --quiet  requirements/dev.in'
touch -a requirements/functests.txt
tox -qe functests --run-command 'pip-compile --allow-unsafe --quiet  requirements/functests.in'
touch -a requirements/tests.txt
tox -qe tests --run-command 'pip-compile --allow-unsafe --quiet  requirements/tests.in'
touch -a requirements/lint.txt
tox -qe lint --run-command 'pip-compile --allow-unsafe --quiet  requirements/lint.in'

# For good measure you could test that actually recompiling the files without --dry-run still works.
# Recompile all the files exactly the same as they are:
$ make --always-make requirements

# Recompile all the files upgrading all dependencies to their latest versions:
$ make --always-make requirments args=--upgrade

# Alternative way to upgrade all dependencies to their latest versions:
$ rm -rf .tox requirements/*.txt; make requirements
```

Taking a step back
------------------

With comments removed the entire implementation of `make requirements` is now 9 lines:

```make
find_prerequisites = $(addprefix requirements/,$(shell grep '^[[:blank:]]*\(-r\|-c\)[[:blank:]]*' "requirements/$*.in" | sed 's/^[[:blank:]]*\(-r\|-c\)[[:blank:]]*//'))

.SECONDEXPANSION:
requirements/%.txt: requirements/%.in $$(find_prerequisites)
>   @touch -a $(subst requirements.txt,dev.txt,$@)
>   @tox -qe $(subst requirements,dev,$(basename $(notdir $@))) --run-command 'pip-compile --allow-unsafe --quiet $(args) $<'

.PHONY: requirements requirements/
requirements requirements/: $(foreach file,$(wildcard requirements/*.in),$(basename $(file)).txt)
```

Admittedly 9 pretty dense lines. But for this we get quite a lot of functionality:

* The list of requirements files and their dependencies doesn't appear anywhere in `Makefile`. It figures out what `requirements/*.txt` files should exist from what `*.in`'s there are, and figures out the prerequisites of each `*.txt` from the `-r`'s in its `*.in`. If we add or remove requirements files or `-r`'s we don't have to change `Makefile`, and `Makefile` shouldn't need to differ across projects if they have different requirements files or `-r`'s.

* If you've changed one or more `*.in` files (e.g. added or removed dependencies) just run `make requirements` and it will re-compile only the `*.txt` files that need to be re-compiled, including any other `*.txt` files that depend on the `*.txt`'s of the `*.in`'s that you changed

* To force it to recompile all the files: `make --always-make requirements`

* You can also compile just one or more files (rather than all of them) with commands like `make requirements/foo.txt` or `make requirements/foo.txt requirements/bar.txt`. If you ask it to compile `foo.txt` and the `*.in`'s of any `*.txt`'s that `foo.in` depends on have changed then it will compile those `*.txt`'s first. But you don't really need this given that `make requirements` already compiles only the files that need to be

* Upgrade all dependencies in all files to their latest versions with `make --always-make requirements args=--upgrade`

* Upgrade or downgrade a dependency in all files with `make --always-make requirements args='--upgrade-package newrelic'` or `--upgrade-package newrelic==1.2.3`

* If for some reason you wanted to you could upgrade or downgrade a dependency only in one or more files not all: `make --always-make requirements/dev.txt args=--upgrade` or `args='--upgrade-package newrelic`' or `--upgrade-package newrelic==1.2.3`

* You can speed it up significantly by compiling requirements files in parallel with `make -j` (and it knows that a file must finish compiling before any files that depend on it can start compiling)


Makefile
```diff
@@ -139,6 +139,10 @@ frontend-lint: node_modules/.uptodate
 bddtests: python
 	@tox -qe bddtests
 
+# For a given requirements/*.in file find all the requirements/*.txt files that
+# it imports by grepping the *.in file for lines like '-r foo.txt' or '-c bar.txt'.
```
seanh, line 143:
```suggestion
# For a given requirements/*.txt file find all the requirements/*.txt files that
# it imports by grepping its *.in file for lines like '-r foo.txt' or '-c bar.txt'.
```

Makefile
```diff
@@ -139,6 +139,10 @@ frontend-lint: node_modules/.uptodate
...
+find_prerequisites = $(addprefix requirements/,$(shell grep '^[[:blank:]]*\(-r\|-c\)[[:blank:]]*' "requirements/$*.in" | sed 's/^[[:blank:]]*\(-r\|-c\)[[:blank:]]*//'))
```
seanh, line 144: This can be reduced to just one `sed` command instead of a `grep '...' "requirements/$*.in | sed '...'. Something like this:

```suggestion
find_prerequisites = $(addprefix requirements/,$(shell sed 's/^[[:blank:]]*\(-r\|-c\)[[:blank:]]*//p' "requirements/$*.in"))
```

Makefile
```diff
@@ -139,30 +139,41 @@ frontend-lint: node_modules/.uptodate
...
+#
+#     sed -n 's/<regex>/<replacement>/p'
+#
+# * The -n tells sed not to print the non-matching lines
+# * The /p tells sed to print the matching lines
+# * The regex part finds '-r foo.txt' or '-c bar.txt' lines
+# * The replacement part prints 'requirements/foo.txt' with the -r or -c,
+#   whitespace and comments removed
+# * The tr at the very end joins the output into one line
+find_prerequisites = $(shell sed -n 's/^[[:blank:]]*\(-r\|-c\)[[:blank:]]*\([[:alnum:]\/-_\.]\+\).*$$/requirements\/\2/p' "requirements/$*.in" | tr '\n' ' ')
```
seanh, line 154: Given this input file:

```
# These normal package names should not be printed:
pip-tools
toml # Needed for pydocstyle to support pyproject.toml.
pylint
pydocstyle
pycodestyle

# These imports should be printed with "requirements/" prepended and their
# -r's, -c's, whitespace and comments stripped:
-r tests.txt
    -r  functests.txt
-rbddtests.txt#Trailing comment
 -cconstraints.txt # Trailing comment

# These lines that look like imports but aren't should not be printed:
# -r bar.txt
"-rbar.txt"

# These -r's have no filename after them, they shouldn't be printed.
-r
-r# Trailing comment
```

The command outputs:

    requirements/tests.txt requirements/functests.txt requirements/bddtests.txt requirements/constraints.txt

jon-betts: I think this would be better as a separate script which would be:

* More familiar to Python developers
* Require much less explanation
* Be testable if we wanted to
* About the same size overall when comments are included

This could still be pretty small. You could make this script smaller, but I'm not sure it's worth it.

```python
#!/usr/bin/env python
"""Find all items in a requirements file marked with `-r` or `-c`."""

from os.path import dirname, join
from pathlib import Path
import re
import sys


def main(file_name):
    requirement_files = re.findall(
        r"^\s*-[cr]\s*([^\s#]+)", Path(file_name).read_text(), re.MULTILINE
    )

    print(" ".join(join(dirname(file_name), i) for i in requirement_files))


if __name__ == "__main__":
    main(file_name=sys.argv[1])
```

I'm not sure how this would integrate with the make variable interpolation. Part of this would make the boundary clearer.

jon-betts (commented):
I think the idea here is good, automated detection means one less thing to customise in our templates. I think we've also demonstrated something quite small is probably good enough to do that.

I think using `sed` and `bash` makes it harder to understand, than it needs to be. I've given an example below of Python script which does the same sort of thing and would be more familiar to us. I'm not sure how this would play with Make. 

This is very short without the comments, but I think we might want to balance the scales a bit more on the side of simple and easy to understand.

Makefile
seanh: > I think this would be better as a separate script which would be:
> 
> * More familiar to Python developers

I think that's very reasonable.

Not tested much but I think this is how you'd integrate a separate `bin/find_prerequisites` script into the `Makefile`:

```make
.SECONDEXPANSION:
requirements/%.txt: requirements/%.in $$(shell bin/find_prerequisites requirements/$$*.in)
	@touch -a $(subst requirements.txt,dev.txt,$@)
	@tox -qe $(subst requirements,dev,$(basename $(notdir $@))) --run-command 'pip-compile --allow-unsafe --quiet $(args) $<'
```

I wonder what it'd be like if it were a separate script file but was a bash script still using sed and tr? But I think Python is very reasonable as well.

seanh:
I'm gonna close this as I don't think it's the thing to work on right now but I've opened an issue on the cookiecutter repo for it <https://github.com/hypothesis/cookiecutters/issues/37>. We know how to get this working now, so we can do it whenever we want to

PR #4240, opened 2022-07-19
Allow LTI1.3 upgrades from the admin pages

marcospri:
Add a path to upgrade existing instances to LTI 1.3

I went through different options to do this with @chrisshaw and settled on the one implemented here, unifying the creation/upgrading flows into one. In any case we can revisit it once it get enough use.

# Testing 

- Search for issuer `https://blackboard.com` at http://localhost:8001/admin/registrations/

- Pick any of the results and click `New instance`

- Use the consumer key `Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b`

- Save, you are taken to the application instance details, which new include the registration and a deployment id.

### DIFF
--- a/lms/templates/admin/macros.jinja
+++ b/lms/templates/admin/macros.jinja
@@ -31,14 +31,14 @@
     {% endcall %}
 {% endmacro %}
 
-{% macro form_text_field(request, label, field_name, field_value=None) %}
+{% macro form_text_field(request, label, field_name, field_value=None, placeholder="") %}
     {% call field_body(label) %}
         {% if field_value %}
             <input class="input" name="{{field_name}}" value="{{ field_value }}" type="text">
         {% elif request.params.get(field_name) %}
             <input class="input" name="{{field_name}}" value="{{ request.params.get(field_name) }}" type="text">
         {% else %}
-            <input class="input" name="{{field_name}}" type="text">
+            <input class="input" name="{{field_name}}" type="text" placeholder="{{placeholder}}">
         {% endif %}
     {% endcall %}
 {% endmacro %}
--- a/lms/views/admin/__init__.py
+++ b/lms/views/admin/__init__.py
@@ -1,4 +1,5 @@
 from pyramid.httpexceptions import HTTPFound, HTTPNotFound
+from pyramid.renderers import render_to_response
 from pyramid.view import forbidden_view_config, notfound_view_config, view_config
 
 from lms.validation._exceptions import ValidationError
@@ -29,5 +30,13 @@ def flash_validation(request, schema):
     except ValidationError as err:
         request.session.flash(err.messages["form"], "validation")
         return True
-
     return False
+
+
+def error_render_to_response(
+    request, error_message, template, template_args, flash_type="errors"
+):
+    request.session.flash(error_message, flash_type)
+    response = render_to_response(template, template_args, request=request)
+    response.status = 400
+    return response
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -9,15 +9,61 @@ from lms.models import ApplicationInstance
 from lms.security import Permissions
 from lms.services import ApplicationInstanceNotFound, LTIRegistrationService
 from lms.validation._base import PyramidRequestSchema
-from lms.views.admin import flash_validation
+from lms.views.admin import error_render_to_response, flash_validation
+
+
+class EmptyStringNoneMixin:
+    """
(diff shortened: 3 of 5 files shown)


lms/templates/admin/macros.jinja
```diff
@@ -31,14 +31,14 @@
     {% endcall %}
 {% endmacro %}
 
-{% macro form_text_field(request, label, field_name, field_value=None) %}
+{% macro form_text_field(request, label, field_name, field_value=None, placeholder="") %}
```
marcospri, line 34: Allow a placeholder in the macro

lms/views/admin/__init__.py
```diff
@@ -29,5 +30,13 @@ def flash_validation(request, schema):
     except ValidationError as err:
         request.session.flash(err.messages["form"], "validation")
         return True
-
     return False
+
+
+def error_render_to_response(
```
marcospri, line 36: Helper function to a common pattern in the admin pages

lms/views/admin/__init__.py
```diff
@@ -29,5 +30,13 @@ def flash_validation(request, schema):
...
+    request, error_message, template, template_args, flash_type="errors"
+):
+    request.session.flash(error_message, flash_type)
+    response = render_to_response(template, template_args, request=request)
```
marcospri, line 40: We render to response when rendering a different template that the view's default.

lms/views/admin/application_instance.py
```diff
@@ -9,17 +9,63 @@
...
+class EmptyStringNoneMixin:
+    """
+    Allows empty string as "missing value".
+
+    Marshmallow doesn't have a clean solution yet to POSTed values
+    (that are always present  in the request as empty strings)
+
+    Here we convert them explicitly to None
+
+    https://github.com/marshmallow-code/marshmallow/issues/713
```
marcospri, line 24: I thought it was missing something but this seems like big omission, can't really have optional fields in a form without this.

In any case a custom field it's not the end of the world.

lms/views/admin/application_instance.py
```diff
@@ -9,17 +9,63 @@
...
+
     lms_url = fields.URL(required=True)
     email = fields.Email(required=True)
 
 
+class UpdateApplicationInstanceSchema(ApplicationInstanceSchema):
+    location = "form"
+
+    lms_url = EmptyStringURL(required=False, allow_none=True)
+    email = EmptyStringEmail(required=False, allow_none=True)
```
jon-betts, line 66: Note to self: this looks like `NewApplicationInstanceSchema` could be a child of this, but it can't because the email and url fields have different missing behavior.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -55,6 +56,95 @@ def test_instance_with_duplicate(self, views, application_instance_service):
...
+
+    @pytest.mark.usefixtures("with_upgrade_form")
+    def test_upgrade_non_existing_instance(self, views, application_instance_service):
+        application_instance_service.get_by_consumer_key.side_effect = (
+            ApplicationInstanceNotFound
+        )
+
+        response = views.new_instance()
+
+        assert response.status_code == 400
```
jon-betts, line 146: Nice clear test cases here.

jon-betts (approved):
I think this works and easy to use, and coming in from the registration route is fine. I'd be interested to hear the thinking behind the flow. As mentioned below I wonder if starting from the application instance might be a little less error prone?

I'm a bit concerned that a support person could make a mistake and get themselves in a situation they can't fix. I think there's nothing stopping you from upgrading the wrong application instance. On the other hand I'm not sure what we could do to prevent this, as there's not much to go on to guess that it's a mistake?

You could look at the existing GUIDs and see if we are among them, but it's possible you could be adding a new one and that's fine? So that would be a warning at best. I wonder if you can use the LMS URL somehow to try and spot mistakes?

The current flow is that you start from the registration and then add in the consumer key and then the first time you see the application instance it's already been upgraded. This could be the point you spot a mistake and there's presently no way of undo-ing it. As far as I understand a downgrade should be quite easy to add, as I think you drop some fields and as long as you have a consumer key + secret, you are a 1.1 application instance again. Maybe this is worth waiting for evidence of demand before implementing (i.e. When someone messes up). In any case a different PR.

I think there's some stuff around the use of `render_to_response` which I think could potentially be removed and streamline the code a bit. This is also something that would be a different PR I think. I'd be interested in having a go at this if you didn't mind (change is as good as a break),

PR #4243, opened 2022-07-20
Update ESLint config and simplify project config

robertknight:
Update eslint-config-hypothesis and remove local overrides that are no
longer required.

See also notes in https://github.com/hypothesis/client/pull/4657.


robertknight:
I'm going to go ahead and merge this, as it is a very simple PR that is almost identical to one already merged in the client.

Issue #4214, opened 2022-07-12
Prove we can implement Vitalsource single sign-on


jon-betts:
We want to streamline the user experience by introducing Vitalsource single sign-on, but we need to de-risk the technical aspects of the problem with a working prototype.

This prototype needs to:

* Prove each major element of the project (listed below)
* Demonstrate a working end-to-end process
* Get us into a state where we can develop this feature

It does not need:

* To resolve all UI issues to final quality
* Have tests
* Have production quality code

### Tasks

 * [x] Create a prototype which proves the point
     * https://github.com/hypothesis/lms/pull/4151
 * [x] Create any devdata and instructions required to make this easy to work on
    * https://github.com/hypothesis/devdata/pull/60
    * https://github.com/hypothesis/flimflam/pull/1
    * https://github.com/hypothesis/lms/pull/4266
 * [x] Write up the outcomes of the spike

### Tickets

 * https://github.com/hypothesis/lms/pull/4151
 * https://github.com/hypothesis/lms/pull/4247 - Refactors
 * https://github.com/hypothesis/lms/pull/4246 - Refactors
 * https://github.com/hypothesis/lms/pull/4253 - Refactors
 * https://github.com/hypothesis/lms/pull/4259 - Refactors
 * https://github.com/hypothesis/lms/pull/4260 - Refactors
 * https://github.com/hypothesis/devdata/pull/60 - Dev data

### Questions we are trying to resolve

 * Can we use existing Vitalsource APIs to automatically login users using the details we find in an LTI launch?
   * Generate an SSO link we can present in an iframe which results in the user logged into a book, at a specific location
   * Get a Vitalsource user from the LTI user
      * Required for the SSO link  
   * Check that user has permissions on the given book id (SKUs / Vitalsource Book Id (VBID)) 
       * This is to work around failures in the existing SSO
 * Can we setup a development env?
    * A hypothesis institution to represent customers
    * Can we get access to the Vitalsource LTI book reader app in one of our test LMSs (like Canvas)
    * This app can provision a user account or grant a book license automatically
    * Get credentials for the institution which will work with our test LMS installations
      * These allow you to make calls on behalf of users
      * List books that user can access etc
      * See: Vitalsource developer network
 * How can we reset a given user to start from scratch?
   * Looks like we can't as Vitalsource control half the data
   * We end up "burning" users as a result
   * Can we do better? Simulate failures?

jon-betts:
### Notes

User accounts:

 * Reference accounts
[...]

PR #4246, opened 2022-07-20
Refactor the `VitalSource` service to split out a client

jon-betts:
For:

There's a lot going on here, but basically:

* Hide away any methods which aren't public
* Split the client and the service so that:
   * The service is the public interface
   * The client makes individual calls to the API
* Improvements around the HTTPService to allow it to be used as a session
   * This allows setting headers once instead of on every call
* Split out a small helper for dealing with VS book locations (this untangled some mess)

### About the size

 * This has gotten a little bigger as a result of duplicated imports and testing interfaces that weren't there before
 * We've also added a set of proxy methods in the service
 * We've also added some doc strings
 * Overall it's not much bigger because some of the refactors reduced the size

## Testing notes

 * Get everything bang up to date and started
 * Make sure you have run `make devdata`
 * Visit: https://hypothesis.instructure.com/courses/125/assignments and create a new localhost assignment
 * Choose Vitalsource and add https://hypothesis.vitalsource.com/reader/books/9781786894700
 * You should see the cover and the name
 * Continue on and you should be able to pick a chapter
 * Save and publish and you should see the book
 * You **will not** see the client as it's not available in local dev yet


lms/services/vitalsource/service.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def get_book_info(self, book_id: str):
+        """Get details of a book."""
+
+        return self.client.get_book_info(book_id)
+
+    def get_book_toc(self, book_id: str):
+        """Get the table of contents for a book."""
+
+        return self.client.get_book_toc(book_id)
```
jon-betts, line 19: These are a bit of a pain, but pretty soon we're probably going to have two clients and it would be good to hide the details from the consumer.

This also makes the public interface very clear. At the moment there are only two methods on the client and they are both exposed, but soon there will be a lot more and they will not all be directly exposed. Instead we will have methods which use two or three of them to get something done.

lms/services/vitalsource/model.py
```diff
@@ -0,0 +1,31 @@
...
+
+@dataclass
+class VSBookLocation:
+    """A location within a VitalSource Book."""
+
+    book_id: str
+    """Id of the book"""
+
+    cfi: str
+    """Location within than book."""
```
jon-betts, line 13: This helper class extracts some functionality that's used in multiple places and solves a dependency problem with the service and client as both need access to this.

lms/services/__init__.py
```diff
@@ -72,7 +72,7 @@ def includeme(config):
         "lms.services.assignment.factory", name="assignment"
     )
     config.register_service_factory(
-        "lms.services.vitalsource.factory", name="vitalsource"
+        "lms.services.vitalsource.service_factory", name="vitalsource"
```
marcospri, line 75: Could we register by `iface` too? I mentioned that on the other PR. We could do that on a follow up PR and discuss it there.

lms/services/vitalsource/_client.py
```diff
@@ -0,0 +1,94 @@
...
+        if not api_key:
+            raise ValueError("VitalSource credentials are missing")
+
+        self._http_service = HTTPService()
+        self._http_service.session.headers = {
+            "X-VitalSource-API-Key": api_key,
+            "Accept": "application/json",
+        }
+
+    class _BookInfoSchema(RequestsResponseSchema):
```
marcospri, line 32: I personally don't like the inlined schemas. I'd prefer them to be at the top of the file, a bit in between being in another file and inlined.

For me that makes it easier to read the class but this is probably 100% subjective.

marcospri (approved):
:+1: Good refactor 

The client/service split allows makes it easier to understand what's ours vs just existing API wrapping.

lms/services/vitalsource/_client.py
```diff
@@ -0,0 +1,94 @@
...
+
+        toc = self._BookTOCSchema(response).parse()
+        for chapter in toc["table_of_contents"]:
+            chapter["url"] = VSBookLocation(book_id, chapter["cfi"]).document_url
+
+        return toc
+
+    def _get(self, endpoint):
+        return self._http_service.request(
+            method="GET", url=f"https://api.vitalsource.com/{endpoint}"
```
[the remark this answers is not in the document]
jon-betts: :thinking: I wonder if we want a shared base class and have more than one client or something?

lms/services/vitalsource/_client.py
```diff
@@ -0,0 +1,94 @@
...
+                unknown = EXCLUDE
+
+            cover_image = fields.Str(required=True)
+
+        resource_links = fields.Nested(ResourceLinks, required=True)
+
+    def get_book_info(self, book_id: str):
+        """Get details of a book."""
+
+        # See: https://developer.vitalsource.com/hc/en-us/articles/360010967153-GET-v4-products-vbid-Title-TOC-Metadata
```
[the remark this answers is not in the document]
jon-betts: I've moved these into the doc-strings

lms/services/vitalsource/_client.py
jon-betts: I've popped them at the bottom. I like the localism of it (everything you need to read is right there), but you're right it is an eyesore.

lms/services/vitalsource/_client.py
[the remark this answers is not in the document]
jon-betts: I've just stopped being so presumptuous about the URL and let the caller define it.

lms/services/vitalsource/_client.py
```diff
@@ -0,0 +1,94 @@
...
+        Initialise a client object.
+
+        :param api_key: Key for VitalSource API
+        :raises ValueError: If credentials are invalid
+        """
+        if not api_key:
+            raise ValueError("VitalSource credentials are missing")
+
+        self._http_service = HTTPService()
+        self._http_service.session.headers = {
```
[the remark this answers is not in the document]
jon-betts: I named this to `_http_session` to remind you it's a session object and added some comments

PR #4247, opened 2022-07-20
Expose the `HTTPService` session and refactor

jon-betts:
For:

 * https://github.com/hypothesis/lms/pull/4246

This exposes the session so people can modify it. At the moment this would be a bad idea (as everyone shares it), but we intent to _not_ use the singleton in VitalSource work.

This also:

 * Cuts down the documentation which restates a lot of the requests documentation
 * Removes a parameter for session which was only used for testing
 * Refactors the testing to be simpler and smaller

### DIFF
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -1,5 +1,4 @@
-import requests
-from requests import RequestException
+from requests import RequestException, Session
 
 from lms.services.exceptions import ExternalRequestError
 
@@ -7,14 +6,18 @@ from lms.services.exceptions import ExternalRequestError
 class HTTPService:
     """Send HTTP requests with `requests` and receive the responses."""
 
-    def __init__(self, _session=None):
-        # A requests session is used so that cookies are persisted across
+    # This is here mostly to let auto-spec know about it in the tests
+    session: Session = None
+    """The underlying requests Session."""
+
+    def __init__(self):
+        # A session is used so that cookies are persisted across
         # requests and urllib3 connection pooling is used (which means that
         # underlying TCP connections are re-used when making multiple requests
         # to the same host, e.g. pagination).
-        #
+
         # See https://docs.python-requests.org/en/latest/user/advanced/#session-objects
-        self._session = _session or requests.Session()
+        self.session = Session()
 
     def get(self, *args, **kwargs):
         return self.request("GET", *args, **kwargs)
@@ -75,12 +78,7 @@ class HTTPService:
         response = None
 
         try:
-            response = self._session.request(
-                method,
-                url,
-                timeout=timeout,
-                **kwargs,
-            )
+            response = self.session.request(method, url, timeout=timeout, **kwargs)
             response.raise_for_status()
         except RequestException as err:
             raise ExternalRequestError(request=err.request, response=response) from err
--- a/lms/services/http.py
+++ b/lms/services/http.py
@@ -1,4 +1,4 @@
-from requests import RequestException, Session
+from requests import RequestException, Response, Session
 
 from lms.services.exceptions import ExternalRequestError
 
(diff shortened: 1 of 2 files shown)


tests/unit/lms/services/http_test.py
```diff
@@ -1,134 +1,86 @@
...
-    @pytest.fixture
-    def url(self):
-        """Return the URL that we'll be sending test requests to."""
-        return "https://example.com/example"
-
-    @pytest.fixture(autouse=True)
-    def test_response(self, url):
-        httpretty.register_uri(
-            "GET", url, body='{"test_response_key": "test_response_value"}', priority=-1
-        )
```
jon-betts, line 121: There's an argument to say these aren't great tests, because if requests changes in such a way as this breaks, perhaps we wouldn't notice.

Some observations:

 * The mock will keep us mostly straight on the acceptable args
 * We don't really do much here, so there's not much interface to get wrong
 * If requests is broken... we'll notice, and almost certainly have bigger fish to fry

lms/services/http.py
```diff
@@ -31,62 +55,6 @@ def patch(self, *args, **kwargs):
...
-            In this case ExternalRequestError.response will be None.
-
-        :raise ExternalRequestError: If an error response (4xx or 5xx) is
-            received.
-
-            The original exception from requests will be available as
-            ExternalRequestError.__cause__.
-
-            The error response will be available as
-            ExternalRequestError.response.
```
jon-betts, line 73: I found this doc-string too long to be useful. There's some interesting and useful info here, but I don't think this is the right place for a tutorial.

The detail in the `ExternalRequestError` seems to be going into things which are true, but I'm not sure why you need to read about them here. There are many other true statements we could include here about how all these objects work, but it's 8 lines of code, so perhaps it's not proportionate?

lms/services/http.py
```diff
@@ -1,20 +1,44 @@
...
 from lms.services.exceptions import ExternalRequestError
 
 
 class HTTPService:
     """Send HTTP requests with `requests` and receive the responses."""
 
-    def __init__(self, _session=None):
-        # A requests session is used so that cookies are persisted across
+    session: Session = None
+    """The underlying requests Session."""
```
marcospri, line 10: Is this for auto specing it only? I thinks we should mention in on the doc string. Otherwise it's a bit confusing to have class and instance variables with the same name.

lms/services/http.py
```diff
@@ -1,20 +1,44 @@
...
         # requests and urllib3 connection pooling is used (which means that
         # underlying TCP connections are re-used when making multiple requests
         # to the same host, e.g. pagination).
-        #
+
         # See https://docs.python-requests.org/en/latest/user/advanced/#session-objects
-        self._session = _session or requests.Session()
+        self.session = Session()
+
+    def request(self, method, url, timeout=(10, 10), **kwargs) -> Response:
```
marcospri, line 21: Having this at the top makes it easier to understand what's going on with the other methods :+1:

marcospri (approved):
Looks good. 

Having the sorter docstring on the request method better highlights the important bits there :+1: 

Compleltly out of scope of this PR but I wonder if we could register services both by name and "interface".

```diff
diff --git a/lms/services/__init__.py b/lms/services/__init__.py
index 683f7ffe..568416dd 100644
--- a/lms/services/__init__.py
+++ b/lms/services/__init__.py
@@ -27,10 +27,13 @@ from lms.services.lti_role_service import LTIRoleService
 from lms.services.ltia_http import LTIAHTTPService
 from lms.services.rsa_key import RSAKeyService
 from lms.services.user import UserService
+from lms.services.http import HTTPService
 
 
 def includeme(config):
-    config.register_service_factory("lms.services.http.factory", name="http")
+    config.register_service_factory(
+        "lms.services.http.factory", name="http", iface=HTTPService
+    )
     config.register_service_factory(
         "lms.services.oauth_http.factory", name="oauth_http"
     )
```

This is probably the case where this is less useful (http is very sort name and the class has small API) but for example service.head  will not autocomplete (missing from our implementation). The test would catch that but it's always nice to catch a problem with your editor if you were doing `service :HTTPService = ...`

lms/services/http.py
jon-betts: Yes it was for that. I'll add a comment

PR #4250, opened 2022-07-22
Disable spell-checking for JSTOR and VitalSource input fields

robertknight:
These fields accept both URLs and item IDs. If the user enters a URL
then browsers can figure out not to apply spell checking. In case the
user enters a book/item ID however, we need to explicitly disable spell
checking.

**Testing:**

Enter an article ID like "10.7249/mg876rc.10" in the JSTOR picker. Previously you'd see a spelling error underline on part of the ID. With this change you shouldn't.


lyzadanger (approved):
This seems sensible!

Issue #4215, opened 2022-07-12
Reduce user friction with an MVP Vitalsource single sign on


jon-betts:
Depends on:

  * https://github.com/hypothesis/lms/issues/4214

Once we have a working prototype which has proven all the required technical aspects we need to get this to the point of being something we are happy to deploy. This means:

 * Code style, linting, comment, factoring and layout as we'd expect etc.
 * Test coverage
 * There are no example features which don't work
 * A fully working MVP we can deploy

This does **not** mean solving all the UI and other aspects that would result in a polished experience

## Tasks

 * https://github.com/hypothesis/lms/pull/4269
 * https://github.com/hypothesis/lms/pull/4270
 * https://github.com/hypothesis/lms/pull/4271

jon-betts:
Done

PR #4252, opened 2022-07-22
VitalSource  automatic sign-in proof of concept

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4215
 
This builds on the work in:

 * https://github.com/hypothesis/lms/pull/4151

And other refactoring work to try and make this more real. There are a couple of bigger changes to the approach taken by the previous approach, along with lots of smaller refactors:

 * There is now a division between service and client
 * The client methods all wrap a single call to the API
 * The client now handles authentication transparently (caller isn't involved)
 * ~Instead of one client with two API keys, we have the same client twice with different API keys~
    * I'm making a gamble here that we only need the one client now - All in separate PRs
 * Instead of parsing the document URL (and passing ids) when calling our own proxy API view, we now pass the whole URL and parse it inside the service

### DIFF
--- a/lms/services/vitalsource/_client.py
+++ b/lms/services/vitalsource/_client.py
@@ -1,9 +1,14 @@
-from typing import List
+from functools import lru_cache
+from typing import List, Optional
 
+import xmltodict
 from marshmallow import EXCLUDE, Schema, fields
+from requests import Request
+from requests.auth import AuthBase
 
 from lms.services.exceptions import ExternalRequestError
 from lms.services.http import HTTPService
+from lms.services.vitalsource.exceptions import VitalSourceError
 from lms.services.vitalsource.model import VSBookLocation
 from lms.validation._base import RequestsResponseSchema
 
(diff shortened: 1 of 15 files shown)


lms/services/vitalsource/_client.py
```diff
@@ -76,12 +80,112 @@ def get_table_of_contents(self, book_id: str):
...
+        # Implement requests auth interface, which is just a callable which
+        # gets the request object and modifies it.
+        def _user_auth(request):
+            request.headers["X-VitalSource-Access-Token"] = self._get_user_access_token(
+                user_reference
+            )
+
+            return request
+
+        return _user_auth
```
jon-betts, line 128: This method looks a bit strange, but has a nice effect for us, in that it doesn't involve having to merge headers anywhere.

The way this is setup it also means you can call any method in any order without having to worry about auth for yourself, and we will only ever make one call to get the access token.

lms/services/vitalsource/_client.py
```diff
@@ -103,10 +214,12 @@ class Meta:
...
+        """Title of the chapter."""
+
         cfi = fields.Str(required=True)
-        page = fields.Str(required=True)
+        """A reference to the location within the book."""
 
-        url = fields.Str(required=False)
-        """vitalsource:// like url identifying the book and chapter"""
+        page = fields.Str(required=True)
+        """The start page of the chapter."""
```
jon-betts, line 223: The URL param has gone here because it should have gone before. We have this as "not required" but in truth it never comes from the API, we add it on ourselves later, so this does nothing.

lms/resources/_js_config/file_picker_config.py
```diff
@@ -105,10 +105,10 @@ def microsoft_onedrive(cls, request, application_instance):
         }
 
     @classmethod
-    def vital_source_config(cls, _request, application_instance):
+    def vital_source_config(cls, request, _application_instance):
         """Get Vital Source config."""
-        enabled = application_instance.settings.get("vitalsource", "enabled", False)
-        return {"enabled": enabled}
+
+        return {"enabled": request.find_service(VitalSourceService).enabled}
```
jon-betts, line 111: It's not enough to know that the tick box is on. You also need to check all the required fields exist.

lms/services/vitalsource/_client.py
```diff
@@ -97,6 +179,43 @@ def _json_request(self, method, url):
...
+
+    See: https://requests.readthedocs.io/en/latest/user/advanced/#custom-authentication
+
+    Using this allows us to completely separate authentication from other
+    behaviors which leads to simpler tests, and code. Authentication is also
+    transparent to the caller and results can be cached.
+    """
+
+    def __init__(self, client: VitalSourceClient, user_reference: str):
+        self._client = client
```
marcospri, line 210: I think the dependency of client here for `get_user_credentials` is not very satisfying at fist glance but I can't see a better alternative.

lms/services/vitalsource/factory.py
```diff
@@ -5,9 +5,13 @@
...
+
     return VitalSourceService(
-        client=VitalSourceClient(
-            api_key=request.registry.settings["vitalsource_api_key"]
-        ),
+        # It's important to pass None here if there's no key, so the service
+        # knows it's been disabled. The client will raise an error anyway if
+        # you try and create one with no API key.
+        client=VitalSourceClient(customer_key) if customer_key else None,
+        user_lti_param=settings.get("vitalsource", "user_lti_param"),
```
marcospri, line 15: We could do `user_reference=request.lti_params[settings.get("vitalsource", "user_lti_param")]`

That simplifies a bit the usage but also makes the service only usable on LTILaunches so the current setup is a good choice too

lms/services/vitalsource/_client.py
jon-betts: Yeah I tried a few things. One option is to create the object within a closure in a function:

```python
def user_auth(self, user_reference):
    def _auth(request):
        request.headers['Blah'] = self.blah
    return _auth
```

As technically, any callable will do. This works, but is a bit weird, and makes testing really funky. Have a separate object:

* Allows us to use our normal patching, instead of doing `patch.object(client, _auth)`
* Allows us to follow the examples in the docs, for whatever that's worth

lms/services/vitalsource/factory.py
jon-betts: This hadn't occurred to me. I wonder if this would be confusing in methods like `get_launch_url` where you accept `user_reference` as an argument? In that case it would actually be blank when called (as it's coming from the API, not an LTI launch), but it would be reasonable to ask why it's not using `self.user_reference` if it's around.

We could make it work, by calling the API with the param listed so they always line up, but I feel like that's actually quite a weird way to follow how the data got in.

I suspect we might run into a problem with this even as is, if people actually do use really different values for this, as anything which falls outside of the LTI params we look for would require us to use `request.params`, or add them and at that point it would lose all LTI 1.3 compatibility.

jon-betts:
This has been closed in favour of:


 * https://github.com/hypothesis/lms/pull/4269
 * https://github.com/hypothesis/lms/pull/4270
 * https://github.com/hypothesis/lms/pull/4271

PR #4253, opened 2022-07-22
Clean up the VitalSource api view

jon-betts:
We were previously returning data verbatim from VitalSource and then manipulating it to be the format we want. We might as well do that all in the client directly as our abstraction over the real API.

There are some other small refactors here as well.

### DIFF
--- a/lms/services/vitalsource/_client.py
+++ b/lms/services/vitalsource/_client.py
@@ -47,7 +47,7 @@ class VitalSourceClient:
 
         return _BookInfoSchema(response).parse()
 
-    def get_book_toc(self, book_id: str):
+    def get_table_of_contents(self, book_id: str):
         """
         Get the table of contents for a book.
 
@@ -64,8 +64,8 @@ class VitalSourceClient:
 
             raise
 
-        toc = _BookTOCSchema(response).parse()
-        for chapter in toc["table_of_contents"]:
+        toc = _BookTOCSchema(response).parse()["table_of_contents"]
+        for chapter in toc:
             chapter["url"] = VSBookLocation(book_id, chapter["cfi"]).document_url
 
         return toc
--- a/lms/services/vitalsource/_client.py
+++ b/lms/services/vitalsource/_client.py
@@ -45,7 +45,13 @@ class VitalSourceClient:
 
             raise
 
-        return _BookInfoSchema(response).parse()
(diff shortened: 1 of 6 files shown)


lms/views/api/vitalsource.py
```diff
@@ -11,33 +11,23 @@
...
+    _VALID_BOOK_ID = re.compile(r"^[\dA-Z-]+$")
 
-    def _get_book_id(self):
+    @property
+    def _book_id(self):
         book_id = self.request.matchdict["book_id"]
-        if not self._is_valid_book_id(book_id):
+
+        if not self._VALID_BOOK_ID.match(book_id):
             raise HTTPBadRequest("Invalid `book_id`. It must only contain [0-9A-Z-].")
```
jon-betts, line 31: I'd like to do this in the VS client, but then I'm not sure what error we'd raise. If we had to catch it here again, it's probably a waste of time.

lms/services/vitalsource/_client.py
```diff
@@ -45,9 +45,15 @@ def get_book_info(self, book_id: str):
...
+        book_info = _BookInfoSchema(response).parse()
 
-    def get_book_toc(self, book_id: str):
+        return {
+            "id": book_info["vbid"],
+            "title": book_info["title"],
+            "cover_image": book_info["resource_links"]["cover_image"],
+        }
+
+    def get_table_of_contents(self, book_id: str):
```
jon-betts, line 56: I changed the name here to reflect the name it has all the way down the stack from API view to here.

lms/views/api/vitalsource.py
jon-betts: Turns out the answer was to apply a schema here, then once the perimeter is controlled, everything else can assume this is valid.

Not the only way to go about it, but it matches what we do elsewhere and is simple to implement.

Unfortunately this means we can't unit test this any more. I tried adding functional tests, but there's too much missing machinery on that side to do this easily. I got into messing about trying to generate valid auth tokens and eventually gave up.

It might be good to one day crack that nut so we can test these end-points without huge amounts of effort, but the first will be hard.

marcospri (approved):
Looks good.

If we pushed to many decisions to the client's method we might make them not very reusable from the service but I don't see any of that here.

PR #4260, opened 2022-07-25
Use a service method to decide if `VitalSource` is available

jon-betts:
Currently if LMS was misconfigured with:

 * Vitalsource enabled in the application instance settings
 * But the API key not setup

We would still present the VitalSource file picker, which would then fail.

This PR makes the responsibility of deciding whether the picker is available to the service. This has some follow up benefits for future PRs where the decision will become more complicated. This means we can change the service and nothing else should have to change.

### DIFF
--- a/lms/resources/_js_config/file_picker_config.py
+++ b/lms/resources/_js_config/file_picker_config.py
@@ -1,7 +1,7 @@
 from lms.product import Product
 from lms.product.blackboard import Blackboard
 from lms.product.canvas import Canvas
-from lms.services import JSTORService
+from lms.services import JSTORService, VitalSourceService
 
 
 class FilePickerConfig:
@@ -105,10 +105,10 @@ class FilePickerConfig:
         }
 
     @classmethod
-    def vital_source_config(cls, _request, application_instance):
-        """Get Vital Source config."""
-        enabled = application_instance.settings.get("vitalsource", "enabled", False)
-        return {"enabled": enabled}
+    def vitalsource_config(cls, request, _application_instance):
+        """Get VitalSource config."""
+
+        return {"enabled": request.find_service(VitalSourceService).enabled}
 
     @classmethod
     def jstor_config(cls, request, _application_instance):
--- a/lms/services/vitalsource/service.py
+++ b/lms/services/vitalsource/service.py
@@ -7,8 +7,15 @@ from lms.services.vitalsource.model import VSBookLocation
 class VitalSourceService:
     """A high-level interface for dealing with VitalSource."""
 
-    def __init__(self, client: VitalSourceClient):
+    def __init__(self, client: VitalSourceClient, enabled):
+        self._enabled = enabled
         self.client = client
+
+    @property
+    def enabled(self) -> bool:
+        """Check if the service has everything it needs to work."""
+
+        return bool(self._enabled and self.client)
 
     def get_book_info(self, book_id: str) -> dict:
         """Get details of a book."""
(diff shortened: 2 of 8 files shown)


lms/resources/_js_config/file_picker_config.py
```diff
@@ -105,10 +105,10 @@ def microsoft_onedrive(cls, request, application_instance):
         }
 
     @classmethod
-    def vital_source_config(cls, _request, application_instance):
-        """Get Vital Source config."""
-        enabled = application_instance.settings.get("vitalsource", "enabled", False)
-        return {"enabled": enabled}
+    def vitalsource_config(cls, request, _application_instance):
+        """Get VitalSource config."""
```
jon-betts, line 109: The name of this method was wrong as the product is "VitalSource" rather than "Vital Source". Thought I'd fix it while I was here.

lms/services/vitalsource/service.py
```diff
@@ -7,9 +7,16 @@
...
-    def __init__(self, client: VitalSourceClient):
+    def __init__(self, client: VitalSourceClient, enabled):
+        self._enabled = enabled
         self.client = client
 
+    @property
+    def enabled(self) -> bool:
+        """Check if the service has everything it needs to work."""
+
+        return bool(self._enabled and self.client)
```
jon-betts, line 18: This is going to get a bunch of things added to it pretty soon. This keeps all the changes here instead of distributed about.

tests/unit/lms/services/vitalsource/factory_test.py
```diff
@@ -6,11 +6,29 @@
 
 
 class TestServiceFactory:
-    def test_it(self, pyramid_request, VitalSourceService, VitalSourceClient):
+    @pytest.mark.parametrize(
+        "enabled,expected", ((sentinel.enabled, sentinel.enabled), (None, False))
+    )
```
jon-betts, line 11: Trying to catch the `False`, default here

lms/services/vitalsource/service.py
```diff
@@ -7,9 +7,16 @@
 class VitalSourceService:
     """A high-level interface for dealing with VitalSource."""
 
-    def __init__(self, client: VitalSourceClient):
+    def __init__(self, client: VitalSourceClient, enabled):
+        self._enabled = enabled
         self.client = client
 
+    @property
+    def enabled(self) -> bool:
```
marcospri, line 15: Is this `enabled` or `file_picker_enabled`?

marcospri (approved):
Looks good, this is the same patter we used for jstor, much nicer that having to check settings in a few places.

lms/services/vitalsource/service.py
jon-betts: Yeah, it's probably that in a local sense, but I think a missing piece is that this probably should be a new type of file source concept in it's entirety (rather than a service). It's not obvious to a casual observer, but this is 100% about serving up that type of content. There's no other type of VitalSourcey thing we do.

The same goes for JSTOR. In that world it's really whether "VitalSource" or "GoogleDrive" etc is enabled on this account. In that world, it's kind of about the whole concept being enabled or disabled, rather than just the picker. For example, if you pick a file, and then we break the settings, we should probably do something more intelligent to warn you we can't get the doc than we do right now, which would also use the enabled setting.

Long story short:

* One point of view is this is a generic service which does many things, one of which is turning on and off the file picker
* Another is this is _all_ about VitalSource as a source of documents and it's enabled or disabled as a whole - and _currently_ that's only used in one place (the file picker)

I think I come down on the latter side. I'd really like to have a crack at creating a new concept around document sources at some point which include things like this.

Issue #4218, opened 2022-07-12
Work out deployment plan for VS SSO to customers


jon-betts:
Requires:

 * https://github.com/hypothesis/lms/issues/4215

The new work requires extra settings and fields for customers. We should know how we intend to onboard customers.

This should ensure:

 * Continuity for any existing customers - i.e. backwards compatibility (are there any customers?)
 * How we will upgrade people (does this even matter?)
 * Who will be doing this work (support?)
 * Vitalsource are happy with what we are going to put out - Wording / sign-off?
    * Are they happy for us to expand to new customers?
    * We can't without credentials Vitalsource give us, which implies sign-off

Once this is done we can immediately start gaining some value even if we are planning to refine the solution.

### Questions

 * Do we need to keep backward compatibility?
 * Who are our current customers? Do we even have to worry about this?
     * Pilot customers? More detail here...

## Implementation idea

 * Gather VS API keys and settings from / for the customers
   * We need a VitalSource API key
   * ... and the correct LTI param to read the user reference from (e.g. `user_id`)
   * ... for each application instance we need to upgrade
 * We can deploy the settings page in order to collect the required settings for customers
 * Then we can deploy the bulk of the code once the settings are in place

*RISK!* - You aren't going to know if a users settings work until we deploy the full code. Should we have an interstitial PR which starts using the customer key right away?

## Tasks

 * [x] https://github.com/hypothesis/lms/pull/4262
 * [x] Fill out the required application instances

chrisshaw:
@jon-betts what does gather user data and verify data somehow mean?

robertknight:
Summary of discussion with Jon, Daryl and myself: We've implemented the code changes required to allow customers to use SSO for VitalSource. We need to check if the customers we planned to enable SSO for have it enabled. If they do, we can close this issue. If there are customers that have specific complications, we can open new issues to address those.

robertknight:
I can't remember exactly where we decided for all of these questions, but VitalSource SSO has been shipped for a long time now.

PR #4262, opened 2022-07-28
Add per-instance VitalSource API keys and user ID field settings

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4218

 - The per-instance VitalSource API key needs to be used to query for
   information about users or courses associated with a particular
   institution.
 - The user ID field identifies which field from the LTI launch request
   should be passed to VitalSource's `POST /v3/credentials.xml` API in
   order to obtain a token for use in subsequent user-specific requests
   (eg. to list the books a user has access to)

We need this information to be in place before we can deploy the new work, otherwise existing customers will break.

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -89,11 +89,14 @@
         {{ settings_checkbox("Groups enabled", "blackboard", "groups_enabled") }}
         {{ settings_checkbox("Files enabled", "blackboard", "files_enabled") }}
     </fieldset>
-
     <fieldset class="box">
         <legend class="label has-text-centered">Content integrations</legend>
         {{ settings_checkbox("Microsoft OneDrive enabled", "microsoft_onedrive", "files_enabled", default=True) }}
+
         {{ settings_checkbox("VitalSource enabled", "vitalsource", "enabled") }}
+        {{ settings_text_field("VitalSource user ID param", "vitalsource", "user_lti_param") }}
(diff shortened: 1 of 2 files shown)


lms/templates/admin/instance.html.jinja2
```diff
@@ -89,11 +89,14 @@
         {{ settings_checkbox("Groups enabled", "blackboard", "groups_enabled") }}
         {{ settings_checkbox("Files enabled", "blackboard", "files_enabled") }}
     </fieldset>
-
     <fieldset class="box">
         <legend class="label has-text-centered">Content integrations</legend>
         {{ settings_checkbox("Microsoft OneDrive enabled", "microsoft_onedrive", "files_enabled", default=True) }}
+
         {{ settings_checkbox("VitalSource enabled", "vitalsource", "enabled") }}
+        {{ settings_text_field("VitalSource user ID param", "vitalsource", "user_lti_param") }}
```
marcospri, line 97: At some point VS could get its own `<fieldset>` but we probably should wait until a bit of progress is made in the VS inclusive access project. 

Maybe we'll discover that some settings are for VS filepicker while other are only for VS IA.

marcospri (approved):
Looks good, we can start to fill those in with this.

lms/templates/admin/instance.html.jinja2
jon-betts: I tried laying these out in columns, but it looked really ugly

PR #4263, opened 2022-07-28
Remove the VitalSource API key env var as we no longer use it

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4218

This removes the global Vitalsource API key. This can only be deployed when the new customer specific key is in place


jon-betts:
We have gone back to this being critical, so this doesn't matter

PR #4265, opened 2022-07-29
Sync the user before we issue credentials in the Gateway

jon-betts:
It's possible for a client to launch our gateway view before a user has ever launched us. This will result in the credentials we issue belonging to a user which is not yet in H and so not valid. To get around this we insert the user.

A secondary problem is that the user will have no anntations to look at. We can get some of the way around this by allocating them to the course group. For course grouped assignments, this is actually all you will need.

For groups / sections, the user will still see nothing and will need to launch some assignments.

## What goes wrong the for customer

In all of these cases assume the call is being made for a user which has never interacted with our products.

Assuming we don't sync a user:

 * The customer calls the gateway
* **When they try to exchange our token for an OAuth token they are told the `sub` (user) is invalid**

Assuming we sync the user only:

 * The customer calls the gateway
 * The user successfully exchanges a token 
 * **The customer cannot retrieve any annotations, because the user is in no groups**

Assuming we sync the user and course (this is the solution in this ticket):

 * The customer calls the gateway
 * The user successfully exchanges a token 
 * The customer retrieves annotations in the course group
 * **The customer does not see any annotations from groups or sections**

## Review notes

This is exactly what we do for a regular LTI launch, so doesn't seem risky.

## Change in behavior

This means it's not possible for us to know that a user is "missing" in any way as we will create it from the settings given. This also means we can't use mock values as they will get updated.

As long as the caller issues the same LTI params as they got from the LMS this should work fine.

### DIFF
--- a/bin/gateway/config/gateway_qa_example.json
+++ b/bin/gateway/config/gateway_qa_example.json
@@ -13,6 +13,7 @@
             "tool_consumer_instance_guid": "GUID",
 
             "context_id": "__COURSE_ID__",
+            "context_title": "__COURSE_NAME__",
             "resource_link_id": "__ASSIGNMENT_ID__",
 
             "user_id":  "__USER_ID__",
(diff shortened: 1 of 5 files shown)


marcospri (approved):
Needs some rebasing after merging the fixture PR.

Other than that looks good, it's what we do in any other launch as you mention.

PR #4267, opened 2022-07-29
Fix functional/bdd test H-API mocks

jon-betts:
This ensures we do not cross contaminate our tests by setting up accidental global state in `httpretty` and also uses a more targeted value for `h` URLs in our tests to distinguish them from other calls we might make to `example.com`.

Prior to this change we would not release the HTTPretty mock, even when calling disable. This could cause tests which didn't want it to be subject to it.


tests/functional/conftest.py
```diff
@@ -94,21 +94,19 @@ def intercept_http_calls_to_h():
...
         raise NotImplementedError(f"Unexpected call to URL: {request.method} {uri}")
 
-    httpretty.register_uri(method=Any(), uri=re.compile(".*"), body=error_response)
+    with httpretty.enabled():
+        # Mock all calls to the H API
+        httpretty.register_uri(
+            method=Any(), uri=re.compile(r"^https://h.example.com/.*"), body=""
+        )
+
+        httpretty.register_uri(method=Any(), uri=re.compile(".*"), body=error_response)
```
jon-betts, line 110: I think the problem was registering things outside of the context.

tests/bdd/steps/h_api.py
```diff
@@ -14,10 +14,13 @@ def do_setup(self):
 
         httpretty.reset()
 
+        # Start interception
+        httpretty.enable()
+
```
jon-betts, line 19: I'm not sure if this has the same effect or not, but registering after the enable seems sensible.

tests/bdd/steps/h_api.py
```diff
@@ -14,10 +14,13 @@ def do_setup(self):
...
         # This is the URL we expect all H API calls to go to
         httpretty.register_uri(
             method=Any(),
-            uri=re.compile(r"^https://example.com/private/api/.*"),
+            uri=re.compile(r"^https://h.example.com/.*"),
```
jon-betts, line 23: Could read this from the settings too?

marcospri (approved):

PR #4268, opened 2022-07-29
Fix JKT endpoint httppretty mock

marcospri:
Same fix as https://github.com/hypothesis/lms/pull/4267 but in a different fixture.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -88,7 +88,7 @@ def do_lti_launch(app):
     return do_lti_launch
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 def intercept_http_calls_to_h():
     """
     Monkey-patch Python's socket core module to mock all HTTP responses.
--- a/tests/functional/lti_certification/v13/conftest.py
+++ b/tests/functional/lti_certification/v13/conftest.py
@@ -1,10 +1,10 @@
 import json
 from datetime import datetime
 
+import httpretty
 import importlib_resources
 import jwt
 import pytest
-from httpretty import httpretty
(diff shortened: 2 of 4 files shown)


tests/functional/conftest.py
```diff
@@ -88,7 +88,7 @@ def do_lti_launch(post_params, get_params=None, **kwargs):
     return do_lti_launch
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
```
marcospri, line 91: Default to always intercept all HTTP requests. We can revisit if we find we do want to make them for real but not sure if that is desirable ever.

tests/functional/lti_certification/v13/conftest.py
```diff
@@ -63,18 +63,18 @@ def application_instance(lti_registration):
...
     jwk = {"keys": [jwt_public_key]}
 
-    httpretty.register_uri(
-        method="GET",
-        uri=lti_registration.key_set_url,
-        body=json.dumps(jwk),
-    )
+    with httpretty.enabled():
+        httpretty.register_uri(
+            method="GET",
```
marcospri, line 72: To prove this is actually in effect now changing GET to POST or anything else will result if failing tests.

jon-betts (approved):
I've tried running this with and without the auto-use to see if it slows things down, and it doesn't seem to. So I can't see a reason not to leave this on.

There's no clash I can see with the two fixtures both doing `httpretty` things.

I also think I might be understanding more of what happened here. I suspect the setup for one `httpretty` session was getting into the other. So `httpretty` wasn't active all the time, but whem it was active the settings were a mishmash of both maybe?

PR #4269, opened 2022-07-29
Return a redirect to automatically log users in to Vitalsource

jon-betts:
For:

 * https://github.com/orgs/hypothesis/projects/71/views/1

## What's changed?

Previous to this PR we would direct the user straight to the bookshelf on the correct book. This required the user to have independently signed into the bookshelf for this to work.

Now we:

 * Check if the user has a licences
    *  If not show an error
 * Generate a redirect link that signs the user in, and then redirects to the shelf

This means the user does not always need to log in

### What's not so great about this?

If the user does not have a licence then we show them an error instead of showing them the book viewer. However users can sometimes be automatically granted a licence by the book viewer. Unfortunately the redirect mechanism doesn't take this into account and fails.

This means we are deciding between our error page and the one generated by the redirect. It was decided that our error page could contain more helpful information about what had gone wrong than the default. At the moment there's no very good way to help the user to resolve the situation.

## ~Do not merge!~

~We can't deploy this until existing customers have settings added to their application instances.~

We've merged in this PR now which should make this safe to release:

 * https://github.com/hypothesis/lms/pull/4351

## Testing notes

Checking SSO:

 * Visit: https://bookshelf.vitalsource.com/ and ensure you are not logged in
 * Start all the things
 * Make sure you've run `make devdata` recently in LMS
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/3153 as the teacher account
 * You should be redirected to the book eventually (it's a little slow for me)
 * Visit: https://bookshelf.vitalsource.com
 * You will see you are now logged in

When there's no licence:

 * Visit: https://bookshelf.vitalsource.com/ and ensure you are not logged in
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/3127
 * You should see an unhandled error "Sorry, but something went wrong."

When there's no user (faking it):

 * Edit `lms.services.vitalsource._client.py:get_user_credentials()`
 * Comment out the if above the raised error
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/3153
 * You should see an unhandled error "Sorry, but something went wrong."

### DIFF
--- a/lms/services/vitalsource/factory.py
+++ b/lms/services/vitalsource/factory.py
@@ -4,9 +4,14 @@ from lms.services.vitalsource.service import VitalSourceService
 
 def service_factory(_context, request):
     settings = request.find_service(name="application_instance").get_current().settings
-    api_key = request.registry.settings["vitalsource_api_key"]
+
+    customer_key = settings.get("vitalsource", "api_key")
 
     return VitalSourceService(
-        client=VitalSourceClient(api_key) if api_key else None,
+        # It's important to pass None here if there's no key, so the service
+        # knows it's been disabled. The client will raise an error anyway if
+        # you try and create one with no API key.
+        client=VitalSourceClient(customer_key) if customer_key else None,
+        user_lti_param=settings.get("vitalsource", "user_lti_param"),
         enabled=settings.get("vitalsource", "enabled", False),
     )
--- a/lms/services/vitalsource/factory.py
+++ b/lms/services/vitalsource/factory.py
@@ -5,13 +5,15 @@ from lms.services.vitalsource.service import VitalSourceService
 def service_factory(_context, request):
     settings = request.find_service(name="application_instance").get_current().settings
 
+    global_key = request.registry.settings["vitalsource_api_key"]
     customer_key = settings.get("vitalsource", "api_key")
 
     return VitalSourceService(
         # It's important to pass None here if there's no key, so the service
         # knows it's been disabled. The client will raise an error anyway if
         # you try and create one with no API key.
-        client=VitalSourceClient(customer_key) if customer_key else None,
-        user_lti_param=settings.get("vitalsource", "user_lti_param"),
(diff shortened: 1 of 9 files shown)


lms/services/vitalsource/service.py
```diff
@@ -1,41 +1,60 @@
...
         self._enabled = enabled
-        self.client = client
+        self.user_lti_param = user_lti_param
 
     @property
     def enabled(self) -> bool:
         """Check if the service has everything it needs to work."""
 
-        return bool(self._enabled and self.client)
+        return bool(self._enabled and self._client and self.user_lti_param)
```
marcospri, line 31: Is `user_lti_param` required for any type of VS integration?

lms/services/vitalsource/factory.py
```diff
@@ -4,9 +4,14 @@
...
+
+    customer_key = settings.get("vitalsource", "api_key")
 
     return VitalSourceService(
-        client=VitalSourceClient(api_key) if api_key else None,
+        # It's important to pass None here if there's no key, so the service
+        # knows it's been disabled. The client will raise an error anyway if
+        # you try and create one with no API key.
+        client=VitalSourceClient(customer_key) if customer_key else None,
+        user_lti_param=settings.get("vitalsource", "user_lti_param"),
```
marcospri, line 15: Do we need to default here to "user_id" or similar? Or is that the job of the admin pages when enabling VS support?

marcospri (approved):
I reckon I already looked at this before, in any case, looks good.

I have a couple of question around `user_lti_param`.

lms/services/vitalsource/service.py
jon-betts: It was, but isn't any more.

lms/services/vitalsource/factory.py
jon-betts: We don't know that there's any kind of standard yet. Until we see a few we don't really know what the fields are.

PR #4270, opened 2022-07-29
Move the call to get the SSO link to the front end as a call back

jon-betts:
For:

 * https://github.com/orgs/hypothesis/projects/71/views/1

Requires:

 * https://github.com/hypothesis/lms/pull/4269

Put the front-end in charge of getting Vitalsource links to allow more parallelism. 

## Testing notes

Nothing should have changed from: https://github.com/hypothesis/lms/pull/4269.

You can follow the instructions there and should get the same outcome, except you should see the calls being made from the front end.

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -3,7 +3,6 @@ from enum import Enum
 from typing import List, Optional
 
 from lms.models import Assignment, GroupInfo, Grouping, HUser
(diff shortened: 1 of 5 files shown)


marcospri (approved):
Yeah, this makes sense, no need the launch to wait for VS when we can offload part of that work to happen later.

Nice that this change is not that big and is a built in mechanism in LMS.

PR #4271, opened 2022-07-29
Add specific error handling and dialogs for new Vitalsource error modes

jon-betts:


marcospri (approved):
Tested both cases to get the error dialogs.

Can't comment much else, it probably needs a quick check from @lyzadanger. There's a couple TODOs here, not sure if we can address them now or they need some extra work.

I was going to say that there's no link back to VS but there's indeed one, `VitalSource book reader` is a link but couldn't tell it was in a different color. I'm guessing that that color been there for a while, am I more color blind today?

robertknight:
> I was going to say that there's no link back to VS but there's indeed one, VitalSource book reader is a link but couldn't tell it was in a different color.

That's a mistake (on my part IIRC). Links in the middle of text should always have a non-color hint, eg. by adding the `underline` attribute to the `Link`.

I wonder if perhaps we should make the underline opt-out rather than opt-in so the component is accessible by default. Thoughts @lyzadanger?

PR #4272, opened 2022-07-29
Return assignment info in the gateway data

jon-betts:
For:

https://github.com/hypothesis/lms/issues/4197

This PR adds dumping of assignments alone. This isn't enough to use the API in all circumstances, but is the first meaningful piece which allows the user to test.

I'm planning to release this piece by piece, partially because that's how I'm working on it, but also because it reduces the size of each PR. I suspect we may find that certain features aren't needed immediately  / ever, so this should get something the customer can use to them quicker.

## Testing notes

 * Start `h` and `lms`
 * Ensure you have visited some assignments
 * Use `bin/gateway/oauth_call.py` to issue a request and view the results
    * If you are missing a config file for this please ask
 * You should see assignment data dumped in the output

### DIFF
--- a/docs/gateway/schema.json
+++ b/docs/gateway/schema.json
@@ -9,7 +9,9 @@
                     "list_endpoints": {
                         "method": "GET",
                         "url": "https://h.example.com/api",
-                        "headers": {"Accept": "application/vnd.hypothesis.v2+json"}
+                        "headers": {
+                            "Accept": "application/vnd.hypothesis.v2+json"
+                        }
                     },
                     "exchange_grant_token": {
                         "method": "POST",
@@ -24,6 +26,20 @@
                         }
                     }
                 }
+            },
+            "data": {
+                "assignments": [
+                    {
+                        "lms": {
+                            "document_url": "https://example.com"
+                        },
+                        "lti": {
+                            "resource_link_id": "23987234",
+                            "resource_link_title": "Assignment name",
+                            "resource_link_description": "Assignment description"
+                        }
+                    }
+                ]
             }
         }
     ],
@@ -50,12 +66,52 @@
                 }
             },
             "required": ["h"]
+        },
+        "data": {
+            "type": "object",
+            "properties": {
+                "assignments": {
+                    "type": "array",
+                    "items": {"$ref": "#/$defs/assignment"}
+                }
+            },
+            "required": ["assignments"],
+            "additionalProperties": false
         }
     },
-    "required": ["api"],
+    "required": ["api", "data"],
     "additionalProperties": false,
 
     "$defs": {
+        "assignment": {
+            "type": "object",
+            "properties": {
+                "lms": {
+                    "type": "object",
+                    "properties": {
+                        "document_url": {"type":  "string"}
+                    },
+                    "additionalProperties": false,
+                    "required": ["document_url"]
+                },
+                "lti": {
+                    "type": "object",
+                    "properties": {
+                        "resource_link_id": {"type":  "string"},
+                        "resource_link_title": {"type":  "string"},
+                        "resource_link_description": {"type":  "string"}
+                    },
+                    "additionalProperties": false,
+                    "required": [
+                        "resource_link_id",
+                        "resource_link_title",
+                        "resource_link_description"
+                    ]
+                }
+            },
+            "additionalProperties": false,
+            "required": ["lms", "lti"]
+        },
         "api_call": {
             "title": "Pre-defined API call",
             "type": "object",
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -20,7 +20,9 @@ class AssignmentService:
     def __init__(self, db: Session):
         self._db = db
 
-    def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
+    def get_assignment(
+        self, tool_consumer_instance_guid, resource_link_id
+    ) -> Assignment:
         """Get an assignment by resource_link_id."""
 
         return (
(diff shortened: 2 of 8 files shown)


docs/gateway/schema.json
```diff
@@ -24,6 +26,20 @@
...
+                "assignments": [
+                    {
+                        "lms": {
+                            "document_url": "https://example.com"
+                        },
+                        "lti": {
+                            "resource_link_id": "23987234",
+                            "resource_link_title": "Assignment name",
+                            "resource_link_description": "Assignment description"
+                        }
```
jon-betts, line 40: This layout looks a bit weird in isolation, but the thinking in general is something along the lines of:

 * The root looks like the object in `h` speak (in this case there is no equivalent)
 * The `lti` section has things which are expressible in LTI params
 * The `lms` has things which aren't `h` and aren't directly `lti` params either. For example with groups the parent id etc.

lms/services/assignment.py
```diff
@@ -32,6 +34,16 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
...
 
+    def get_assignments_for_grouping(self, grouping_id) -> List[Assignment]:
+        """Get all assignments within a specified grouping."""
+
+        return (
+            self._db.query(Assignment)
+            .join(AssignmentGrouping)
+            .join(Grouping, Grouping.id == grouping_id)
+            .all()
+        )
```
jon-betts, line 45: As we aren't iterating over groups yet we don't need the join and eager loading stuff. As it turns out this was kind of a bless in disguise as how I'd written it didn't work.

The "noise" fixtures in this test really saved my bacon here by showing me that the eager join condition _does not_ apply to the query, so you end-up getting all the the assignments without this explicit join.

I'm not sure now how the two will play together.

lms/views/api/gateway.py
```diff
@@ -58,35 +59,80 @@ def h_lti(context, request):
...
-                    # List the API end-points
-                    "method": "GET",
-                    "url": h_api_url,
-                    "headers": {"Accept": "application/vnd.hypothesis.v2+json"},
+        "api": {"h": _GatewayService.render_h_connection_info(request)},
+        "data": _GatewayService.render_lti_context(request),
+    }
+
+
+class _GatewayService:
```
jon-betts, line 68: This is heavily hinting this might want to be a service without actually doing it yet. We can choose later to either:

 * Make it a close collaborator (and maybe pick another name)
 * Make it a full service

In the event we do either having the test split up like this makes it easy and also make the main testing of the view simple.

tests/functional/api/gateway_test.py
```diff
@@ -35,7 +35,8 @@ def test_minimum_viable_login(self, gateway_launch, required_params):
                         "url": TEST_SETTINGS["h_api_url_public"] + "token",
                     },
                 }
-            }
+            },
+            "data": {"assignments": Any.list()},
```
jon-betts, line 39: We check this in detail in the back-end, no need to duplicate it all again here.

tests/unit/lms/views/api/gateway_test.py
```diff
@@ -85,6 +180,17 @@ def validator(self, schema):
...
 
+@pytest.fixture
+def assignment():
+    # Use baked values, so we can have static comparisons in the tests
+    return factories.Assignment.create(
+        document_url="document_url",
+        resource_link_id="resource_link_id",
+        title="title",
+        description="description",
+    )
```
jon-betts, line 191: I'm planning to keep adding more and more joined things and relationships to this assignment as we get more data to dump.

lms/services/assignment.py
```diff
@@ -32,6 +34,16 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
             .one_or_none()
         )
 
+    def get_assignments_for_grouping(self, grouping_id) -> List[Assignment]:
```
marcospri, line 37: I reckon we might have to revisit this at some point. I still think we'll get  more flexible and easiert to compose  services if we remove the `.all` here and return a query instead.

The joins here will be about what data you get, any eager-loading applied outside would be about how you get it.

That's in any case a bigger discussion for another moment.

marcospri (approved):

lms/services/assignment.py
jon-betts: I thought having about having a lazy load spec object with a bunch of booleans on it so you could say what you wanted but:

* The way SQLAlchemy makes you do it, it's a real pain to layout and would be a nest of ifs
* I think we'd only ever call it in one configuration, so all that conditionality would have to be tested, but would only be in one configuration in actual usage

It's a tough one. I can see why giving control to the caller would be good from a full control point of view, but I'm not sure you have much of a service boundary in that world. The application of the correct eager loading etc. is quite difficult to get right (harder than the query here). I'm not sure what kind of encapsulation the service would be providing then? It becomes pretty close to just letting everyone do their own ORM queries.

One possible more radical concept is we might have a missing layer? I kind of hate to suggest this, but pre-baked queries as their own thing, without specifying how they are used? This isn't a fleshed out idea, but maybe there's something between services and models? Dunno.

In the next few PR's we'll be starting to bring eager loading in. My plan to start with was you'd ask `eager=True` and have the full shebang applied. I'm not sure we'll ever have a second caller for most of these things? So maybe if we ever do, and that second caller want's a different set of stuff we tackle it then?

marcospri: Yeah I don't have a concrete proposal and I agree  with this point 

> I'm not sure what kind of encapsulation the service would be providing then

But I think we are already in a world were that encapsulation is already very thin, we return ORM objects from the service and the views/other services are free to have DB side effects with them.

SQLA is huge and I think trying to build a layer around it would be mistake. Is the alternative to have all the views using it freely? Probably not.

My preferred solution in other projects has been to have other plain python objects (data-classes?) as the return values when the service knows how many objects it's returning (typically, one) and return queries when the number is unknown and >1. 

This brings a lot of duplication with those objects that are very similar to the models but helps avoiding a lot of problems with tracking where the queries happens and composes very well in the query side, for example implementing pagination can be done without shoving all that logic into the services.

jon-betts:
I think we're not going to return assignment data at all, because there's no sensible way to use it.

PR #4273, opened 2022-07-30
Add hashes to requirements files

seanh:
Now that we have [make requirements](https://github.com/hypothesis/lms/pull/4167) and have [removed `tox-pip-sync`](https://github.com/hypothesis/lms/pull/4191) (which chokes on hashes in requirements files) we can add hashes to LMS's requirements files. Hashes are a security feature to protect against supply chain attacks, see <https://pip-tools.readthedocs.io/en/latest/#using-hashes> and <https://pip.pypa.io/en/stable/topics/repeatable-installs/#hash-checking>. This PR is the result of adding `--generate-hashes` to `Makefile`'s `pip-compile` command and then running `make --always-make requirements` which force-recompiles all the requirements files but without changing the version numbers of any of the dependencies, so it should only add the hashes to the files. Dependabot should see that our requirements files now have hashes and work with them.


seanh:
Closing in favour of https://github.com/hypothesis/lms/pull/4285

PR #4274, opened 2022-08-01
Allow editing of some AI settings in the admin pages

marcospri:
Mistakes while we input these couldn't be solved, allow editing them.

This is more relevant now with the LTI1.3 upgrade flow were these might need updating anyway.

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -42,8 +42,8 @@
 
     <fieldset class="box">
         {{ macros.disabled_text_field("Consumer key", instance.consumer_key) }}
-        {{ macros.disabled_text_field("Deployment ID", instance.deployment_id) }}
-        {{ macros.disabled_text_field("LMS URL", instance.lms_url) }}
+        {{ macros.form_text_field(request, "Deployment ID", "deployment_id", field_value=instance.deployment_id) }}
+        {{ macros.form_text_field(request, "LMS URL", "lms_url", field_value=instance.lms_url) }}
     </fieldset>
 
     <fieldset class="box">
@@ -78,7 +78,7 @@
 
     <fieldset class="box">
         <legend class="label has-text-centered">Canvas settings</legend>
-        {{ macros.disabled_text_field("API domain", instance.custom_canvas_api_domain) }}
+        {{ macros.form_text_field(request, "API domain", "custom_canvas_api_domain", field_value=instance.custom_canvas_api_domain) }}
         {{ macros.disabled_text_field("Developer key", instance.developer_key) }}
         {{ settings_checkbox("Sections enabled", "canvas", "sections_enabled") }}
         {{ settings_checkbox("Groups enabled", "canvas", "groups_enabled") }}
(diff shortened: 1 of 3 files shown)


lms/views/admin/application_instance.py
```diff
@@ -59,7 +59,7 @@ class NewApplicationInstanceSchema(PyramidRequestSchema):
     email = fields.Email(required=True)
 
 
-class UpdateApplicationInstanceSchema(ApplicationInstanceSchema):
+class UpgradeApplicationInstanceSchema(ApplicationInstanceSchema):
```
marcospri, line 62: This schema is not for updating the AI but for upgrading it to LTI1.3.

lms/views/admin/application_instance.py
```diff
@@ -265,6 +265,10 @@ def show_instance(self):
     def update_instance(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
+        for ai_field in ["lms_url", "custom_canvas_api_domain", "deployment_id"]:
+            if value := self.request.params.get(ai_field):
+                setattr(ai, ai_field, value.strip())
```
jon-betts, line 270:
```suggestion
            if value := self.request.params.get(ai_field, "").strip():
                setattr(ai, ai_field, value)
```

Might be a better way than this, but a string containing only spaces will be truthy until stripped.

lms/templates/admin/instance.html.jinja2
```diff
@@ -42,8 +42,8 @@
 
     <fieldset class="box">
         {{ macros.disabled_text_field("Consumer key", instance.consumer_key) }}
-        {{ macros.disabled_text_field("Deployment ID", instance.deployment_id) }}
-        {{ macros.disabled_text_field("LMS URL", instance.lms_url) }}
+        {{ macros.form_text_field(request, "Deployment ID", "deployment_id", field_value=instance.deployment_id) }}
+        {{ macros.form_text_field(request, "LMS URL", "lms_url", field_value=instance.lms_url) }}
```
jon-betts, line 46: I thought this was one of the fields we automatically updated on application instances, but it's not. I wonder if we could / should, as this field is probably very wrong in most cases.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -269,6 +269,33 @@ def test_update_instance(
...
         )
 
+    @pytest.mark.parametrize(
+        "setting,value,expected",
+        (
+            ("lms_url", "http://some-url.com", "http://some-url.com"),
+            ("custom_canvas_api_domain", "some.domain.com", "some.domain.com"),
+            ("deployment_id", "DEPLOYMENT_ID", "DEPLOYMENT_ID"),
+        ),
+    )
```
jon-betts, line 279: Could do with a few examples to prove what happens when the values have space or are blank etc.

lms/views/admin/application_instance.py
jon-betts: Is there any uniqueness check or anything else for `deployment_id`?

jon-betts (approved):
The editing work and seems like an obvious improvement. I wonder if we should enable this for other fields too?

One thing that occurred to me is, should there be any kind of check around the deployment id?

lms/templates/admin/instance.html.jinja2
marcospri: I removed the option to edit the canvas custom domain, this is overridden in every request from the request parameters.

PR #4282, opened 2022-08-01
Migration for the event_table

marcospri:
Auto generated from the models at https://github.com/hypothesis/lms/pull/4283

This can be seen in action in https://github.com/hypothesis/lms/pull/4144


jon-betts (approved):
This is auto generated, but I gave it a manual test anyway. It upgrades and removes. There's nothing controversial here.

PR #4283, opened 2022-08-01
Models for the new event tables

marcospri:
Migration at https://github.com/hypothesis/lms/pull/4282

This can be seen in action in https://github.com/hypothesis/lms/pull/4144

### DIFF
--- a/lms/db/_columns.py
+++ b/lms/db/_columns.py
@@ -1,0 +1,23 @@
+import sqlalchemy as sa
+
+
+def varchar_enum(enum, max_length=64, nullable=False, unique=False):
+    """Return a SA column type to store the python enum.Enum as a varchar in a table."""
+    return sa.Column(
+        sa.Enum(
+            enum,
+            # In order to maintain maximum flexibility we will only enforce the
+            # type on the Python side, and leave the Postgres side open as a plain
+            # VARCHAR
+            native_enum=False,
+            create_constraint=False,
+            validate_strings=True,
+            # Without a length SQLAlchemy will constrain it to the longest value
+            # we happen to have right now, which could limit us in future
+            length=max_length,
+            # Use the string values, not the keys to persist the values
+            values_callable=lambda obj: [item.value for item in obj],
+        ),
+        nullable=nullable,
+        unique=unique,
+    )
--- a/lms/models/event.py
+++ b/lms/models/event.py
@@ -1,0 +1,127 @@
+from enum import Enum
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.ext.mutable import MutableDict
+
+from lms.db import BASE, varchar_enum
+
+
+class EventType(BASE):
+    __tablename__ = "event_type"
+
+    class Type(str, Enum):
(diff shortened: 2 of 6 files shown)


lms/db/_columns.py
```diff
@@ -0,0 +1,23 @@
+import sqlalchemy as sa
```
jon-betts, line 1: If this is for public export then I think our standard is to have this file be called `columns.py` to show it contains public things (even if they are not directly exported from here).

lms/db/_columns.py
```diff
@@ -0,0 +1,23 @@
...
+            validate_strings=True,
+            # Without a length SQLAlchemy will constrain it to the longest value
+            # we happen to have right now, which could limit us in future
+            length=max_length,
+            # Use the string values, not the keys to persist the values
+            values_callable=lambda obj: [item.value for item in obj],
+        ),
+        nullable=nullable,
+        unique=unique,
+    )
```
jon-betts, line 23: :+1: This is a mouthful, much better as a method if we are going to re-use it.

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+        sa.DateTime(), server_default=sa.func.now(), nullable=False, index=True
+    )
+    """Time the event occurred, defaults to now() if not specified"""
+
+    type_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("event_type.id", ondelete="cascade")
+    )
+    type = sa.orm.relationship("EventType")
+
+    """One of EventType"""
```
jon-betts, line 38: ?

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+    """Which assignment this event relates to"""
+
+    grouping_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("grouping.id", ondelete="cascade"), nullable=True
+    )
+    grouping = sa.orm.relationship("Grouping", foreign_keys=[grouping_id])
+    """If this event belongs to one particular grouping, which one"""
+
+
+class EventUser(BASE):
```
jon-betts, line 67:
```suggestion
class EventUser(BASE):
    """Relationship between events, users and their roles."""
```

lms/models/event.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+    assignment = sa.orm.relationship("Assignment")
+    """Which assignment this event relates to"""
+
+    grouping_id = sa.Column(
+        sa.Integer(), sa.ForeignKey("grouping.id", ondelete="cascade"), nullable=True
+    )
+    grouping = sa.orm.relationship("Grouping", foreign_keys=[grouping_id])
+    """If this event belongs to one particular grouping, which one"""
+
```
jon-betts, line 65: A relationship to the data and users would be very helpful here. Although as we discussed, hard to get right without any calling code, so perhaps now isn't the time.

jon-betts (approved):
Looks good. A bit hard to evaluate in the abstract, but we've discussed this layout a number of times and I think we're pretty happy with it.

It might help to have a few back relationships from the event to other items, particularly the data, but that can come at a later date.

PR #4289, opened 2022-08-02
Add basic test for LMS frontend entry point

robertknight:
Add some minimal tests for the LMS app entry point, to check it runs the right
app depending on the config.

To allow the entry point to only run when explicitly called in the test
environment, define an `LMS_FRONTEND_TESTS` global in the Rollup bundle, which
code can check for.


lyzadanger (approved):
LGTM.

PR #4290, opened 2022-08-02
Add a downgrade to LTI 1.1 option in the admin pages

marcospri:
This is something that *in theory* shouldn't be necessary. In practice we have found the need for it 3 times already over the last few days as we are finding bugs, strange behaviors from the LMS, etc.

### Testing

- Click new instance

```
Consumer key: Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b
Deployment id: DEPLOYMENT_ID
```

Leave the rest blank

- Check that the AI now has a registration and deployment IDs attached

- Scroll to the bottom of the application instance, click downgrade.
- The application instance now longer has the LTI 1.3 fields.

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -101,8 +101,22 @@
         {{ settings_text_field("JSTOR site code", "jstor", "site_code") }}
     </fieldset>
 
-    {% call macros.field_body(label="") %}
+    <div class="has-text-right mb-6">
         <input type="submit" class="button is-primary" value="Save" />
-    {% endcall %}
+    </div>
+
+
+    <fieldset class="box has-background-danger-light">
+        <legend class="label has-text-centered has-text-danger">Danger zone</legend>
+
+        {% call macros.field_body(label="Downgrade to LTI 1.1") %}
+            <input type="submit" class="button mb-2" formaction="{{ request.route_url("admin.instance.downgrade", id_=instance.id) }}" value="Downgrade">
+            <p>Downgrade this instance to LTI 1.1 removing its association with a registration and its deployment ID.</p>
+            <p>This action will <b>break any existing LTI 1.3 assignments</b>. To undo it, upgrade it back again to LTI 1.3 using the same registration and deployment ID.</p>
+        {% endcall %}
+
+    </fieldset>
+
+
 </form>
 {% endblock %}
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -193,6 +193,35 @@ class AdminApplicationInstanceViews:
             )
         )
 
+    @view_config(route_name="admin.instance.downgrade", request_method="POST")
+    def downgrade_instance(self):
+        ai = self._get_ai_or_404(**self.request.matchdict)
+
+        if ai.lti_version != "1.3.0":
+            self.request.session.flash(
+                f"Application instance: '{ai.id}' is not on LTI 1.3.", "errors"
+            )
+            return HTTPFound(
+                location=self.request.route_url("admin.instance.id", id_=ai.id)
+            )
+
+        if not ai.consumer_key:
+            self.request.session.flash(
+                f"Application instance: '{ai.id}' doesn't have a consumer key to fallback to.",
+                "errors",
+            )
+            return HTTPFound(
+                location=self.request.route_url("admin.instance.id", id_=ai.id)
+            )
+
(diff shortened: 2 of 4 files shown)


lms/templates/admin/instance.html.jinja2
```diff
@@ -101,8 +101,22 @@
...
 
-    {% call macros.field_body(label="") %}
+    <div class="has-text-right mb-6">
         <input type="submit" class="button is-primary" value="Save" />
-    {% endcall %}
+    </div>
+
+
+    <fieldset class="box has-background-danger-light">
+        <legend class="label has-text-centered has-text-danger">Danger zone</legend>
```
marcospri, line 110: This both how github does it and an Archer reference.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -71,10 +125,9 @@ def test_upgrade_instance(
 
         response = views.new_instance()
 
-        application_instance_service.get_by_consumer_key(
+        application_instance_service.get_by_consumer_key.assert_called_once_with(
```
marcospri, line 128: Actually assert something on this old test.

lms/templates/admin/instance.html.jinja2
```diff
@@ -101,8 +101,22 @@
...
+
+        {% call macros.field_body(label="Downgrade to LTI 1.1") %}
+            <input type="submit" class="button mb-2" formaction="{{ request.route_url("admin.instance.downgrade", id_=instance.id) }}" value="Downgrade">
+            <p>Downgrade this instance to LTI 1.1 removing its association with a registration and its deployment ID.</p>
+            <p>This action will <b>break any existing LTI 1.3 assignments</b>. To undo it, upgrade it back again to LTI 1.3 using the same registration and deployment ID.</p>
+        {% endcall %}
```
jon-betts, line 116: It might be nice to have a cheesy javascript "Are you sure?" confirmation.

lms/views/admin/application_instance.py
```diff
@@ -193,6 +193,35 @@ def upgrade_instance(self, lti_registration, consumer_key):
...
+    def downgrade_instance(self):
+        ai = self._get_ai_or_404(**self.request.matchdict)
+
+        if ai.lti_version != "1.3.0":
+            self.request.session.flash(
+                f"Application instance: '{ai.id}' is not on LTI 1.3.", "errors"
+            )
+            return HTTPFound(
+                location=self.request.route_url("admin.instance.id", id_=ai.id)
+            )
```
jon-betts, line 206: Maybe it would be better to not show the button if this isn't applicable?

lms/views/admin/application_instance.py
```diff
@@ -193,6 +193,35 @@ def upgrade_instance(self, lti_registration, consumer_key):
...
+                "errors",
+            )
+            return HTTPFound(
+                location=self.request.route_url("admin.instance.id", id_=ai.id)
+            )
+
+        ai.lti_registration_id = None
+        ai.deployment_id = None
+
+        self.request.session.flash("Downgraded LTI 1.1 successful", "messages")
```
jon-betts, line 220: The instructions in the page say to undo this you can re-add the same registration and deployment id. We have an opportunity to tell you what they are in this message.

We could add a "To undo this re-add the following..."

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -96,7 +149,7 @@ def test_upgrade_instance_allows_empty(
     ):
         lti_registration_service.get_by_id.return_value = factories.LTIRegistration()
 
-        # Replicate real's pyramid_request behavkour
+        # Replicate real's pyramid_request behaviour
```
jon-betts, line 152: Spelling just wants to be free man! Lighten up

jon-betts (approved):
I've given this a shot and it works as advertised. It's better than we have now so we should get it out.

I've added some ideas about how this could be a nicer experience without a too much effort I think:

 * Maybe include undo instructions in the success message
 * Have an "are you sure?" in JS on the button
 * We could hide the button when you are in 1.1 mode instead of (as well as?) checking for it

marcospri:
> Maybe include undo instructions in the success message
> Have an "are you sure?" in JS on the button
> We could hide the button when you are in 1.1 mode instead of (as well as?) checking for it

This are good suggestions, I'll add them to a ticker were I'm tracking all admin pages improvements so I  can tackle them all in one go.

PR #4296, opened 2022-08-03
Split the JSTOR service up into a directory

jon-betts:
The JSTOR service is getting bigger and quite a lot of it is all to do with parsing article metadata.

This splits up the service into a few parts to make each test shorter, and to make testing the metadata stuff easier.


marcospri (approved):
Yeah, this split makes sense :+1:

PR #4298, opened 2022-08-04
Inject None as a client to the VS service when the api key is missing

marcospri:
This fixes an issue were the api key is required (ValueError is thrown)
for a VitalSourceSerivce.enabled check.

Not sure if this is the right approach, the enable check in the service does check for the presence of the client so this works.

https://github.com/hypothesis/lms/blob/907c512c3929add67987f703d50985f18d068c1b/lms/services/vitalsource/service.py#L18

## Testing

- On `main` remove `VITALSOURCE_API_KEY` from `.devdata.env` (and restart the server is running)

- Try to configure an assignment, you'll get an error page.

- Same thing on this branch should work.

### DIFF
--- a/lms/services/vitalsource/factory.py
+++ b/lms/services/vitalsource/factory.py
@@ -4,10 +4,9 @@ from lms.services.vitalsource.service import VitalSourceService
 
 def service_factory(_context, request):
     settings = request.find_service(name="application_instance").get_current().settings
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
:+1: This is pretty much what we do in the backed up VS PRs as well.

PR #4300, opened 2022-08-04
Return JSTOR titles and sub-titles with less interpretation

jon-betts:
This changes the data structure we return from the JSTOR metadata proxy API from:

```json
{
    "title": "Title: Subtitle"
}
```

To be:

```json
{
    "item": {
        "title": "Title",
        "subtitle": "Subtitle"
    }
}
```

## Testing notes

 * Start LMS and friends
 * Create a new assignment using the localhost LMS instance
 * Open the F12 dev tools to the network tab and filter by "jstor"
 * Select JSTOR and enter: https://www.jstor.org/stable/3314735
 * You should see the data correctly displayed
 * You should see the new details in the network output

### DIFF
--- a/lms/services/jstor/_article_metadata.py
+++ b/lms/services/jstor/_article_metadata.py
@@ -47,7 +47,7 @@ class ArticleMetadata:
         return cls(_ArticleMetadataSchema(response).parse())
 
     def as_dict(self):
-        return {"title": self.title, "content_status": self.content_status}
+        return {"item": self.titles, "content_status": self.content_status}
 
     @property
     def content_status(self) -> str:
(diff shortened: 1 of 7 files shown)


lyzadanger:
A nice thing is that the client's failure mode on this is gentle:

image

lms/services/jstor/_article_metadata.py
```diff
@@ -64,28 +64,31 @@ def content_status(self) -> str:
...
         # field with a single entry.
-        elif title := self._data.get("title"):
+        if title := self._data.get(title_key):
+            titles["title"] = self._strip_html_tags(title)
 
             # Some articles have a subtitle which needs to be appended for the
             # title to make sense.
-            if subtitle := self._data.get("subtitle"):
-                # Some titles include a trailing ':' delimiter, some do not.
-                title = f"{title.rstrip(':')}: {subtitle}"
```
lyzadanger, line 81: How much grief would it be to add back the colon-stripping off of the end of `title` here? That's messy data and my light preference would be to have the backend take care of it, but it _can_ be done on the frontend if needed.

jon-betts: Not a problem

jon-betts: We should now strip this again.

lyzadanger:
Front-end changes are pushed as a separate commit. To test this out:

* Create a new assignment or edit an existing one in https://hypothesis.instructure.com/courses/125 to use JSTOR content. Use the JSTOR content ID `26893773 ` in the picker (yep, that interface will accept URLs or IDs).
* In the picker, you should see the full title and subtitle of the article. (This article's title and subtitle are in all caps. That's expected.)
* Launch the assignment. When the JSTOR banner eventually populates the title, you should see the title but _not_ the subtitle. That is expected for the time being.

marcospri (approved):
Not sure what motivates this but seems more sensible in any case to have access to both independently.

PR #4301, opened 2022-08-04
Catch 404's from JSTOR and try and distinguish between no item and wrong URL

jon-betts:
The JSTOR API raises 404s when:

 * You have a completely broken URL
 * You have the right URL but the particular thing you are looking for doesn't exist

There's quite a lot of variation, but by experimenting with various calls I managed to get a set of indicators that narrowed down which type of 404 we have.

## Replicating my findings

Correctly formatted by missing items:

 * https://labs.jstor.org/api/anno/metadata/10.2307/9999999999
    * `Content-Type: application/json;charset=UTF-8`
    * Content: `null`
 * https://labs.jstor.org/api/anno/thumbnail/10.2307/999999999?offset=1&width=280
   * `Content-Type: text/plain;charset=UTF-8`

Broken URLs:

 * https://labs.jstor.org/api/anno/BROKEN
    * `Content-Type: application/json;charset=UTF-8`
    * Content: `{	"detail": "Not Found"}`
 * https://labs.jstor.org/api/BROKEN
    * `Content-Type: text/html;charset=UTF-8`
    * Content: A big HTML page

## Testing notes

 * Start everything
 * Create a new assignment in https://hypothesis.instructure.com/courses/125
 * Open the JSTOR option in the file picker
 * Try very long numbers
 * You should see "Book 345345345 not found" or something similar
 * In the network tab you should see errors with 400 statuses for the metadata and thumbnail

### DIFF
--- a/lms/services/exceptions.py
+++ b/lms/services/exceptions.py
@@ -258,6 +258,29 @@ class BlackboardFileNotFoundInCourse(Exception):
         super().__init__(self.details)
 
 
+class SerializableError(Exception):
+    """An exception compatible with our default error handling for APIs."""
+
+    def __init__(
+        self,
+        message: str = None,
+        error_code: Optional[str] = None,
+        details: Optional[dict] = None,
+    ):
+        """
+        Initialise the error.
+
+        :param message: Message to display to the user (if any)
+        :param error_code: Error code to allow client side code to detect
+            this error
+        :param details: A JSON serializable payload of extra info
+        """
+        super().__init__(message)
+        self.error_code = error_code
+        self.message = message
+        self.details = details
+
+
 def _repr_external_request_exception(exception):
     # Include the details of the request and response for debugging. This
     # appears in the logs and in tools like Sentry and Papertrail.
--- a/lms/services/jstor/service.py
+++ b/lms/services/jstor/service.py
@@ -3,11 +3,16 @@ from urllib.parse import quote
 
 import requests
 
-from lms.services.exceptions import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError, SerializableError
 from lms.services.http import HTTPService
 from lms.services.jstor._article_metadata import ArticleMetadata
 from lms.services.jwt import JWTService
 from lms.views.helpers import via_url
+
+
+class ArticleNotFound(SerializableError):
+    def __init__(self, article_id):
+        super().__init__(message=f"Article '{article_id}' not found")
 
 
 class JSTORService:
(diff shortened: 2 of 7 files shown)


lms/services/jstor/service.py
```diff
@@ -124,7 +161,7 @@ def _api_request(self, path_template, doi, params=None) -> requests.Response:
         if "/" not in doi:
             doi = f"{self.DEFAULT_DOI_PREFIX}/{doi}"
 
-        url = self._api_url + path_template.format(doi=quote(doi, safe="/"))
+        url = self._api_url.rstrip("/") + path_template.format(doi=quote(doi, safe="/"))
```
jon-betts, line 164: Kind of incidental, but our URLs look broken in live as they have double slashes in them. It works fine, but looks like it could be a source of errors (especially when dealing with 404s).

lms/services/exceptions.py
```diff
@@ -258,6 +258,29 @@ def __init__(self, file_id):
...
+
+        :param message: Message to display to the user (if any)
+        :param error_code: Error code to allow client side code to detect
+            this error
+        :param details: A JSON serializable payload of extra info
+        """
+        super().__init__(message)
+        self.error_code = error_code
+        self.message = message
+        self.details = details
```
jon-betts, line 281: We have some default error handling for our API end-points but:

 * The mechanism is duck typed and kind of non-obvious
 * You have to create compatible classes on your own
 * It doesn't support a "message" to show the user

This class codifies the things you have to do to get this to work

lms/views/api/exceptions.py
```diff
@@ -170,6 +171,7 @@ def http_bad_request(self):
         context=CanvasAPIPermissionError
     )
     @exception_view_config(context=Exception)
+    @exception_view_config(context=SerializableError)
```
marcospri, line 174: I wonder if we could use 

`exclog.ignore` https://docs.pylonsproject.org/projects/pyramid-exclog/en/latest/#settings

to also don't show these on the logs? They are bit noisy. 

I don't think we need to do that right now.

lms/views/api/exceptions.py
```diff
@@ -178,23 +180,23 @@ def api_error(self):
...
+        this conveniently.
 
+        If the exception does not have an `error_code` attribute then:
         1. The response's HTTP status will be 500
-
         2. A fixed string (see below) will be used as the "message" field in
            the response's JSON body
         """
 
         if hasattr(self.context, "error_code"):
```
marcospri, line 196: This contract is a bit finicky/thin but this PR makes it a bit  better.

lms/views/api/exceptions.py
```diff
@@ -178,23 +180,23 @@ def api_error(self):
...
             return ErrorBody(
                 error_code=self.context.error_code,
+                message=getattr(self.context, "message", None),
```
marcospri, line 199: Who's consuming this message? Is it mostly for being more human-readable on the let's say chrome devtools?

marcospri (approved):
Tested the 404 case and the happy path and it's working as expected.

I think this solution is better than keep adding things to ExternalRequestError. :+1:

lms/views/api/exceptions.py
jon-betts: This is what you see when you are in the file picker.

PR #4303, opened 2022-08-05
Handle duplicated `deployment_id` while updating application instances

marcospri:
Something that came up in the review where editing of `deployment_id` was added.

## Testing

- New instance, with any "deployment_id" value
- In the newly created instance try updating deployment_id to any other value. That should update.
- Now try to update it to "1f5a3b8f-1c11-4837-9ea2-f3a1596382f2" which should trigger the new code.

### DIFF
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -294,9 +294,35 @@ class AdminApplicationInstanceViews:
     def update_instance(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
-        for ai_field in ["lms_url", "deployment_id"]:
-            if value := self.request.params.get(ai_field, "").strip():
-                setattr(ai, ai_field, value)
+        if lms_url := self.request.params.get("lms_url", "").strip():
+            ai.lms_url = lms_url
+
+        if deployment_id := self.request.params.get("deployment_id", "").strip():
+            if ai.lti_version == "1.3.0":
+                try:
+                    duplicated_ai = (
+                        self.application_instance_service.get_by_deployment_id(
+                            issuer=ai.lti_registration.issuer,
+                            client_id=ai.lti_registration.client_id,
+                            deployment_id=deployment_id,
+                        )
+                    )
+                except ApplicationInstanceNotFound:
+                    # No duplicated AI, we can update the current one.
+                    pass
+                else:
+                    if duplicated_ai.id != ai.id:
+                        self.request.session.flash(
+                            f"Can't set deployment_id to '{deployment_id}'. Already in use on: {self._get_ai_link(duplicated_ai)}",
+                            "errors",
+                        )
+
+                        return HTTPFound(
+                            location=self.request.route_url(
+                                "admin.instance.id", id_=ai.id
+                            )
+                        )
+            ai.deployment_id = deployment_id
 
         for setting, sub_setting, setting_type in (
             ("canvas", "sections_enabled", bool),
@@ -336,3 +362,6 @@ class AdminApplicationInstanceViews:
 
         except ApplicationInstanceNotFound as err:
             raise HTTPNotFound() from err
+
+    def _get_ai_link(self, ai) -> str:
+        return f"""<a href="{self.request.route_url('admin.instance.id', id_=ai.id)}">Application instance {ai.id}</a>"""
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -300,7 +300,7 @@ class AdminApplicationInstanceViews:
         if deployment_id := self.request.params.get("deployment_id", "").strip():
             if ai.lti_version == "1.3.0":
                 try:
-                    duplicated_ai = (
+                    existing_ai = (
                         self.application_instance_service.get_by_deployment_id(
                             issuer=ai.lti_registration.issuer,
                             client_id=ai.lti_registration.client_id,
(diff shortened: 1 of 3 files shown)


lms/views/admin/application_instance.py
```diff
@@ -294,9 +294,35 @@ def show_instance(self):
     def update_instance(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
-        for ai_field in ["lms_url", "deployment_id"]:
-            if value := self.request.params.get(ai_field, "").strip():
-                setattr(ai, ai_field, value)
+        if lms_url := self.request.params.get("lms_url", "").strip():
+            ai.lms_url = lms_url
+
+        if deployment_id := self.request.params.get("deployment_id", "").strip():
```
marcospri, line 300: This is a mouthful made a little bit longer by using the exception on, the service (rather than `one_or_none`).

lms/views/admin/application_instance.py
```diff
@@ -294,9 +294,35 @@ def show_instance(self):
...
+            if ai.lti_version == "1.3.0":
+                try:
+                    duplicated_ai = (
```
jon-betts, line 303: Is this the `existing_ai` rather than the `duplicated_ai`. It sounds like our `ai` is the one which would be duplicating?

This is also the language used in the tests below.

lms/views/admin/application_instance.py
```diff
@@ -336,3 +362,6 @@ def _get_ai_or_404(self, consumer_key=None, id_=None) -> ApplicationInstance:
 
         except ApplicationInstanceNotFound as err:
             raise HTTPNotFound() from err
+
+    def _get_ai_link(self, ai) -> str:
+        return f"""<a href="{self.request.route_url('admin.instance.id', id_=ai.id)}">Application instance {ai.id}</a>"""
```
jon-betts, line 367: A bit sad to have HTML in the backend, but without inventing something new with the flash system it seems like it's probably the best of a bad bunch.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -358,6 +358,73 @@ def test_update_instance_save_ai_fields_keeps_existing(
...
 
+    def test_update_instance_deployment_id(
+        self, pyramid_request, application_instance_lti_13
+    ):
+        pyramid_request.matchdict["consumer_key"] = sentinel.consumer_key
+        pyramid_request.params["deployment_id"] = "NEW_DEPLOYMENT_ID"
+
+        AdminApplicationInstanceViews(pyramid_request).update_instance()
+
+        assert pyramid_request.session.peek_flash("errors")
```
jon-betts, line 369: Why is this an error?

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -358,6 +358,73 @@ def test_update_instance_save_ai_fields_keeps_existing(
...
+        )
+        application_instance_service.get_by_deployment_id.side_effect = (
+            ApplicationInstanceNotFound
+        )
+        pyramid_request.params["deployment_id"] = "NEW_DEPLOYMENT_ID"
+
+        AdminApplicationInstanceViews(pyramid_request).update_instance()
+
+        assert not pyramid_request.session.peek_flash("errors")
+        application_instance_lti_13.deployment_id = "NEW_DEPLOYMENT_ID"
```
jon-betts, line 401: I'm not sure I understand the distinctions between these three cases properly? Is it this?

 * The first one is a new application instance?
 * The second an update, but where you find the existing one (so it looks like a duplicate, but it's not really)
 * The third one is an update, but there's no existing match?

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -358,6 +358,73 @@ def test_update_instance_save_ai_fields_keeps_existing(
...
+        self,
+        db_session,
+        pyramid_request,
+        application_instance_lti_13,
+        application_instance_service,
+    ):
+        existing_ai = factories.ApplicationInstance(
+            lti_registration_id=application_instance_lti_13.lti_registration_id,
+            deployment_id="NEW_DEPLOYMENT_ID",
+        )
```
jon-betts, line 413: Above we use `existing_ai.deployment_id`. Setting it to `NEW_DEPLOYMENT_ID` is a bit confusing, as it's the old one?

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -358,6 +358,73 @@ def test_update_instance_save_ai_fields_keeps_existing(
...
+    def test_update_instance_deployment_id_no_duplicate(
+        self, pyramid_request, application_instance_service, application_instance_lti_13
+    ):
+        application_instance_service.get_by_id.return_value = (
+            application_instance_lti_13
+        )
+        application_instance_service.get_by_deployment_id.side_effect = (
+            ApplicationInstanceNotFound
+        )
+        pyramid_request.params["deployment_id"] = "NEW_DEPLOYMENT_ID"
```
jon-betts, line 396: Why do we set the consumer key in the previous two tests, but not here?

jon-betts (commented):
I've tested this out and it works, but I can't really work out what the tests are doing any why?

Perhaps I'm just being a bit dense here, but it seems like the cases aren't clear, or maybe are changing more than one thing at once? I have a suspicion these are written to get into all the nooks and crannies of the conditionals, but as a set of testing situations I'm not really sure what they represent.

This might just be a case of naming? In any case, if I'm just being dumb then this does work and can be merged. The admin area doesn't matter as much as code end users interact with.

It is also getting pretty huge now. We might want to consider a split at some point.

PR #4306, opened 2022-08-05
Rollback failed transaction in admin pages

marcospri:
This is broken in `main` currently:

## Testing 

- Go to http://localhost:8001/admin/registration/id/2/new/instance
```
consumer key: Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b
deployment_id: 0f29b2ec-444d-462b-9c5b-fcc0803ab2d2
```

In main it will fail, it will hos a message in this branch.

### DIFF
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -177,9 +177,15 @@ class AdminApplicationInstanceViews:
             application_instance.lms_url = lms_url
 
         try:
-            # Flush here to find if we are making a duplicate in the process of upgrading
+            # Flush here to find if we are making a duplicate in the process of
+            # upgrading
             self.request.db.flush()
         except IntegrityError:
+            # Leave a clean transaction, otherwise  we get a:
+            #   "PendingRollbackError: This Session's transaction has been
+            #   rolled back due to a previous exception during flush."
+            self.request.db.rollback()
+
             return error_render_to_response(
                 self.request,
                 f"Application instance with deployment_id: {self.request.params['deployment_id']} already exists",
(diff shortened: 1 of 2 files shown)


lms/views/admin/application_instance.py
```diff
@@ -180,6 +180,9 @@ def upgrade_instance(self, lti_registration, consumer_key):
             # Flush here to find if we are making a duplicate in the process of upgrading
             self.request.db.flush()
         except IntegrityError:
+            # Leave a clean transaction, otherwise  we get a:
+            #   "This Session's transaction has been rolled back due to a previous exception during flush."
+            self.request.db.rollback()
```
marcospri, line 185: Not sure how to assert this call in the tests. 

Mocking request.db for this one test?

jon-betts: You could purposefully break the DB handle somehow?

jon-betts: Turns out this is difficult, but we can check that the session is still in working order after. I've made this change and will release this as Marcos is away.

I figure having two of us co-operate kind of makes up a review?

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -188,6 +189,11 @@ def test_upgrade_duplicate(self, views, lti_registration_service):
 
         assert response.status_code == 400
 
+        # Show that the DB connection has not been permanently broken. This
+        # would cause us to fail completely when trying to present the error.
+        # We are checking we do _not_ raise `PendingRollbackError` here.
+        db_session.query(ApplicationInstance).all()
```
jon-betts, line 195: You can verify this is doing what it should by removing the rollback above and seeing the test fail.

PR #4307, opened 2022-08-05
Return details of the current user's groups in the Gateway data

jon-betts:
For:

This PR adds group return values for the gateway end-point. The groups returned are:

* Groups in the current course
* That the user is in as well

## Testing notes

 * Start `h` and `lms`
 * Ensure you have visited some assignments
 * Use `bin/gateway/oauth_call.py` to issue a request and view the results
    * If you are missing a config file for this please ask
 * You should see profile data dumped in the output

### DIFF
--- a/lms/services/grouping/service.py
+++ b/lms/services/grouping/service.py
@@ -205,10 +205,11 @@ class GroupingService:
 
     def get_known_groupings(self, user, course):
         """
-        Get the groupings we already know about for the provided user.
+        Get the groupings from the DB for the provided user.
 
         This will include the course, and any sections / groups we have seen
-        before. This does not attempt to retrieve the groups from the LMS.
+        before. This does not attempt to retrieve the groups from the tool
+        consumer using their API (for example by calling Canvas API).
 
         This only works for direct children of the course, not children of
         children.
--- a/lms/services/grouping/service.py
+++ b/lms/services/grouping/service.py
@@ -1,7 +1,7 @@
 from typing import List, Optional, Union
 
 from sqlalchemy import func
-from sqlalchemy.orm import aliased
+from sqlalchemy.orm import aliased, subqueryload
 
 from lms.models import Course, Grouping, GroupingMembership, LTIUser, User
 from lms.models._hashed_id import hashed_id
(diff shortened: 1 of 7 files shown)


lms/services/grouping/service.py
```diff
@@ -203,6 +203,31 @@ def get_groups(
 
         return self._to_groupings(user, groupings, course, self.plugin.group_type)
 
+    def get_known_groupings(self, user, course):
```
marcospri, line 206: What do you mean by "known" here or "retrieve from the LMS" below? 

That we won't go and query the LMS API  to get the for example sections of a course?

lms/services/grouping/service.py
```diff
@@ -203,6 +203,31 @@ def get_groups(
...
+            self._db.query(Grouping)
+            .filter_by(parent_id=course.id)
+            # Which the user is a part of
+            .join(GroupingMembership)
+            .filter(GroupingMembership.user == user)
+            # Eager load the parent groups
+            .options(subqueryload(parent_groups, Grouping.parent))
+        ).all()
+
+        return [course] + groupings
```
marcospri, line 229: Not sure if returning the course + the children is the clearest from the API perspective.

In our database they are the same concept but I could see how that could be confusing for the clients. In `render_lti_context` where we could include "course" as top level concept separate from groups.

And argument against that would be later when actually using the data to query against H where queries would have to put course + groups together to get all relevant annotations.

In Another example the client might need to iterate over groups to find the course using `group.type` to do something particular there. That might be less common that just querying for all the annotations?

lms/services/grouping/service.py
```diff
@@ -203,6 +203,31 @@ def get_groups(
...
         return self._to_groupings(user, groupings, course, self.plugin.group_type)
 
+    def get_known_groupings(self, user, course):
+        """
+        Get the groupings we already know about for the provided user.
+
+        This will include the course, and any sections / groups we have seen
+        before. This does not attempt to retrieve the groups from the LMS.
+
+        This only works for direct children of the course, not children of
```
marcospri, line 213: Currently there's no more nesting as far as I know (course -> sections/groups).

marcospri (approved):
This does the trick and is not making any unnecessary DB queries as far as I can see.

I have a comment about having `course` as its own concept on the returned data (instead of being part of `groups`) but I reckon the location makes sense for how we know the data is going to be used right now and also you are calling us with the course info in the first place so are probably not going to have to search for the course inside `groups` very often.

lms/services/grouping/service.py
jon-betts: No, but I wanted to make clear that if we ever do have an additional level of nesting that this method isn't smart enough to work with it.

lms/services/grouping/service.py
jon-betts: Yes that's right. I've updated the comment to hopefully be clearer.

lms/services/grouping/service.py
jon-betts: I think I'm generally in favour of not treating the course as something special. For example, I don't really like the custom model sub-classes and separate service for it, although I know others do. On the whole I think it causes artificial divisions between examples of the same thing.

As for how the service will be used: We have very few service methods that are actually generally used around the app for different purposes. They are almost always used in one specific way. So I think it's probably a good idea to cater to the caller you have rather than a hypothetical caller you imagine you might have in future. If we get a case where we need to filter the two apart it's probably a good time to think again?

PR #4320, opened 2022-08-10
Newer Canvas instances will now return a nice error for unavailable files

jon-betts:
For: 

 * https://github.com/hypothesis/product-backlog/issues/1368

It seems Canvas have changed the output we get from a 401 to a 403 when the user is not authorized. As there are many different Canvas deploys in the wild we now catch both.

## Testing notes

 * Login to Canvas as a student
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/1371
 * On main you see a generic error
 * On this branch you should see "Couldn't get the file from Canvas"

### DIFF
--- a/tests/unit/lms/services/exceptions_test.py
+++ b/tests/unit/lms/services/exceptions_test.py
@@ -208,7 +208,20 @@ class TestCanvasAPIError:
                 OAuth2TokenError,
             ),
             # A permissions error from Canvas, because the Canvas user doesn't
-            # have permission to make the API call.
+            # have permission to make the API call. Updated Canvas instances
+            # will send 403, rather than 401.
+            (
+                403,
+                json.dumps(
+                    {
+                        "status": "unauthorized",
+                        "errors": [
+                            {"message": "user not authorized to perform that action"}
+                        ],
+                    }
+                ),
+                CanvasAPIPermissionError,
+            ),
             (
                 401,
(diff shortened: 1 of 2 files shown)


tests/unit/lms/services/exceptions_test.py
```diff
@@ -208,7 +208,20 @@ class TestCanvasAPIError:
...
+            (
+                403,
+                json.dumps(
+                    {
+                        "status": "unauthorized",
+                        "errors": [
+                            {"message": "user not authorized to perform that action"}
+                        ],
+                    }
+                ),
```
jon-betts, line 222: The repeated dumps here are a bit annoying, but it's because there's one "raw" string at the end.

tests/unit/lms/services/exceptions_test.py
```diff
@@ -223,11 +236,7 @@ class TestCanvasAPIError:
             ),
             # A 400 Bad Request response from Canvas, because we sent an invalid
             # parameter or something.
-            (
-                400,
-                json.dumps({"test": "body"}),
-                CanvasAPIServerError,
-            ),
+            (400, json.dumps({"test": "body"}), CanvasAPIServerError),
```
jon-betts, line 239: Removed the trailing comma to let black reformat this

marcospri (approved):

Issue #4325, opened 2022-08-12
Translate custom LTI 1.3 params to allow us to capture irregular user references


jon-betts:
Users can customise where we read the VS user id from. We believe customers might use [custom LTI 1.3 params](https://www.imsglobal.org/spec/lti/v1p3/#custom-properties-and-variable-substitution) to do this, and currently we have no way of reading them as they aren't translated in our 1.3 -> 1.1 conversion.

To get ahead of this we need to translated these.

## Tasks

 * https://github.com/hypothesis/lms/pull/4326

PR #4326, opened 2022-08-12
Process all custom 1.3 parameters instead of a few Canvas ones

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4325

This processes all custom LTI parameters into 1.1 ones with a `custom_` prefix. I'm not sure why we didn't do this before.

This might also untie the knot around using LTI params to determine the product type?

## Testing notes

None really. There is no user facing change, but it allows us to process all custom params instead of just some.


marcospri (approved):
This works nicely for these three.

Not sure if this can be generlized across other LMS and/or variables we don't control.

This map between the names works because we chosen this naming convention in canvas (`custom_` in LTI1.1 , bare name in LTI1.3, as is already nested in the custom claim).

Issue #4327, opened 2022-08-12
Handle teachers and students who can't use SSO


jon-betts:
Per https://hypothes-is.slack.com/archives/C4K6M7P5E/p1660070326654589:

We're finding some teachers are using VitalSource books outside their institutional relationship with VitalSource, or even when their institution doesn't have a relationship at all. So:

- In all of these cases, the students purchase the books directly from VitalSource. No bookstore relationship, so no VitalSource "company" to attach API keys to.
- In some cases, the instructors may not install the VitalSource Launch LTI app, so no LMS reference user.

So we'll need to be able to let students access VitalSource assignments, even if we have no API keys for the institution or they have no reference user.

I believe Rob had this work already, with a fallback to the current direct login if we were missing key information. It seems like we will need to support this.

chrisshaw:
Ideas
- Implement a fallback to a global key (enabled at least through end of month)
  - This assumes we will actually have a fallback. It's not yet clear that's true, although VitalSource looks open to figuring something out.
- Explicitly turn the new SSO on or off by institution
  - This seems to assume there is a fallback as well, but could coincide with...
- Don't show metadata

chrisshaw:
Conversation with @marcospri and @robertknight: 

- The simplest thing to do is to bring back in the fallback to a global key as we already had it
- This is beneficial for VitalSource because
  - We can use a setup _like_ this to drive sales
  - This simplifies onboarding of new institutions for the foreseeable future
- The plan
  - Handle weird user params
    - Add a regex field to LMS admin (blank if none required? Anything sensible is good @marcospri, I'll leave to you)
    - Parse user param based on regex
  - Add fallback: if no key, use global

robertknight:
Notes from a call with VitalSource about this on 2022-08-25: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1661447101462489. In summary, they will be OK with us having a global/fallback API key for the purposes of fetching metadata, as long as the permissions on that key can be limited to just metadata access. They are discussing options for that internally.

PR #4328, opened 2022-08-12
Allow wild usage of Vitalsource outside of an institutional relationship

jon-betts:
For:

*  https://github.com/hypothesis/lms/issues/4327

This is on hold for now as we don't know if this makes sense.


jon-betts:
I don't think we are doing this now

Issue #4200, opened 2022-07-08
Add organizations to the DB model


jon-betts:
In order to track organisations we need to agree and create a DB model for them.

## Tasks

 * [x] https://github.com/hypothesis/lms/pull/4329
 * [x] https://github.com/hypothesis/lms/pull/4330
 * [x] https://github.com/hypothesis/lms/pull/4412

### Add `organization` table

| Field | Type | Notes |
|--------|---------|---------|
| `id` | int -  Primary key |
| `name` | string | For humans
| `parent_id` | FK to `organization.id` | _(optional)_ Not for now, but we'll want it soon
| `public_id` | string | A UUID for putting in external systems
| `settings` | JSON | The usual settings object
| `created` | datetime
| `updated` | datetime

### Modify `application_instances`

Add a FK to `organization.id`

   * This can be nullable, but nullable in this context means "not launched yet" rather than being a valid long term state
   * **Alternative:** Create shadow orgs and destroy them on merge? This way the FK is always populated?

See:

 * https://github.com/hypothesis/lms/issues/4202

### Modify `registrations`

Anything to do here? Should we prevent anything or have constraints around application instances and registrations belonging to the same org?

What if you want to move org, what then?

jon-betts:
Done! This doesn't do anything by itself, but it's a building block and it's deployed.

PR #4329, opened 2022-08-12
Add organization model and DB migration

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4200

We'll probably want to split this up in the end


lms/models/organization.py
```diff
@@ -0,0 +1,42 @@
...
+from lms.models._mixins import CreatedUpdatedMixin
+from lms.models.application_settings import ApplicationSettings
+
+
+class Organization(CreatedUpdatedMixin, BASE):
+    """Model for Organizations comprised of application instances."""
+
+    __tablename__ = "organization"
+
+    id = sa.Column(sa.Integer(), autoincrement=True, primary_key=True)
```
marcospri, line 14: We talked about having a public id here.

The most common solution it's probably having a `uuid = ` with a generated UUID here.

The `uuid` name is probably a bad idea to avoid confusion with GUIDs. 

The value could still be a UUID, maybe b64'ed so it looks different?

```b"US-" + base64.urlsafe_b64encode(uuid.uuid4().bytes).replace(b"=", b"")```

File: lms/models/organization.py [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Get the globally unique id which also indicates the region.
+
+        This returns vales with prefixes like 'us:' or 'ca:'.
+
+        This is useful if you only have the id, but don't know which region you
+        should be looking for it in and is the only id suitable for sharing
+        outside a single region LMS context.
+        """
+        return region.code + ":" + self._vanity_id
```
jon-betts, line 47: This all looks like a bit of a palava, but I don't want to have any possibility of mixing the region code and the body of the id. This isn't something you can change as a user, so we don't want to get into trying to detect and split and join the values all the time.

I think this makes sense as this way you can have a normal form, and then write the value to the DB without much fuss. If it's unique and fits any other requirements you have (do we want to limit to certain chars?) then it's all golden.

Also, not sure we ever would, but this allows us to change the region code without doing much.

lms/models/region.py
```diff
@@ -0,0 +1,44 @@
...
+        :raises ValueError: If no valid region can be found.
+        """
+        authority = request.registry.settings["h_authority"]
+
+        try:
+            return cls._AUTHORITY_MAP[authority]
+        except KeyError as err:
+            raise ValueError(
+                f"Cannot find a region for the authority '{authority}'"
+            ) from err
```
jon-betts, line 44: This part isn't in use yet, so perhaps it shouldn't be in this PR? But I wanted to show how we might go about getting this.

This is setting up for either:

 * Being called from a service
 * Being added to the request as a request method (`request.region`)
 
The second is sort of "nice" in that it's very short, but it sucks because it pollutes the global namespace for a specific purpose. We don't care about the region in a large number of places (any yet).

tests/unit/lms/models/organization_test.py
```diff
@@ -0,0 +1,69 @@
...
+            [
+                Any.instance_of(ApplicationInstance).with_attrs(
+                    {"id": ai_id, "organization": None}
+                )
+                for ai_id in ai_ids
+            ]
+        )
+
+    def test_vanity_id_generation(self, organization, db_session):
+        db_session.flush()
```
jon-betts, line 52: A flush is required here to make the auto generate kick in. Otherwise you get None.

lms/models/organization.py
jon-betts: I went with something very similar after _many_ attempts at seeing if I could do better. Just so it doesn't feel like a total waste:

 * There's Base86 in the same module, but the output is too wild and would probably cause problems in URLs
 * I tried rolling my own Base-Many, but there aren't that many more printable characters you'd want to use, so it doesn't end up any shorter
 * The other types of UUID aren't of any use
 * There are various "short-code" generators, but they probably aren't going to justify their existence with regards to a new dep
 
We might want to consider "slugifying" vanity ids if/when we allow people to change them manually. We already use "python-slugify" in other places. We can't "slugify" these UUIDs as it will modify them allowing for clashes.

lms/models/organization.py
```diff
@@ -0,0 +1,80 @@
...
+        unique=True,
+        # We don't use a standard UUID-4 format here as they are common in Tool
+        # Consumer Instance GUIDs, and might be confused for them. These also
+        # happen to be shorter and guaranteed URL safe.
+        default=lambda: urlsafe_b64encode(uuid4().bytes).decode("ascii").rstrip("="),
+    )
+    """
+    A human readable URL safe vanity id.
+
+    The vanity id can be customised by the user and therefore can end-up not
```
marcospri, line 32: > The vanity id can be customized by the user

Is this true? I thought of this ID as set in stone as the primary key.

Would you then expect the value to be for example editable on the admin pages?

lms/models/organization.py
```diff
@@ -0,0 +1,80 @@
...
+        This is useful if you only have the id, but don't know which region you
+        should be looking for it in and is the only id suitable for sharing
+        outside a single region LMS context.
+        """
+
+        # We use '.' as the separator here because it's not in base64, but it
+        # is URL safe. The other option is '~'.
+        # See: https://www.ietf.org/rfc/rfc3986.txt (2.3)
+        # >    unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+        return region.code + "." + self._vanity_id
```
marcospri, line 52: Not having the region in the DB is fine locally (as we have a DB per region) but not sure if it might cause issues (or an extra step of regionifying, that's a word, the values) 

If we'd ever have to make cross region queries or aggregate some data in the same storage  this could be relevant.

We are very far away from that world so it probably doesn't matter and might make things more difficult than needed right now.

marcospri, line 52: We might want to use the same thing for other entities.

- Should we move this to `_columns.py`? I reckon not until we have the second usage.

- Should we include the "what" in the public ID? `what:where:id`

`org:us:AAAAAAAA`

Then we could use the exact same setup in other objects, get a:

`user:us:BBBBBBB`

I'm not thinking about collision here, just about making them more "human friendly". It won't make difference in most cases but it might be useful when seeing IDs in the wild and difficult to do after we ship the first of these public_ids

lms/models/region.py
marcospri: Not sure if the h_authority 1 to 1 relationship to region is the way to go (will that always be true?)

We could add a new `region_code` setting but it's true that new variables have a bit of operational complexity and if now 1 authority 1 region, why not use it :shrug:

marcospri (approved):

Issue #4201, opened 2022-07-08
Back-fill organization data


jon-betts:
Depends on:

 * https://github.com/hypothesis/lms/issues/4200

Once we have a data model in the DB, we can fill out historical records. This should:

 * We will probably want to generate a name if possible for the org
 * Replicate the effect of the [method used in Sparky](https://github.com/hypothesis/sparky/blob/main/sparky/etl/domain_model/_schools.py): 
 * This won't be Spark code
 * This doesn't need to be in the main app, but we should be either able to do it on demand or it in a migration
 * We should be mindful of sub-problems which might make useful parts of:
   * https://github.com/hypothesis/lms/issues/4202

This process should probably truncate the table before it works (this means it can be run before or after the new application handling logic is added, and decouples the work).

## Technical questions

 * How will the migration work?
 * Run this as a script manually as a one off in CA and US?

## Issues

* https://github.com/hypothesis/lms/pull/4331
* https://github.com/hypothesis/lms/pull/4421

jon-betts:
Done

PR #4331, opened 2022-08-12
Add backfill for organization data

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4201

## Release notes

This should be released after the automatic association of new accounts with orgs in:

 * https://github.com/hypothesis/lms/pull/4332
 
This will ensure there are no gaps in between. This migration will wipe out those records, but that's fine.

## Testing notes

* Reset your DB:

```shell
git checkout orgs-db-model
make services args=down
make services db devdata
git checkout orgs-backfill
make dev
```
*  Start all the things
* Before the migration will work you will need some application instances with GUIDs, e.g.:
  * https://hypothesis.instructure.com/courses/125/assignments/873
  * https://aunltd-test.blackboard.com/ultra/courses/_19_1/cl/outline
* Run the migration:
```shell
hdev alembic upgrade head
```
* You should see output about created orgs
* Check the results

```shell
make sql
```
```sql
SELECT * FROM organization;
```
```sql
SELECT id, organization_id, tool_consumer_instance_guid FROM application_instances;
```
 * You should see organizations for any `application_instances` with guids
 * Downgrade the DB and check the results:
```shell
hdev alembic downgrade -1
```

*  Everything should be gone

### DIFF
--- a/lms/migrations/versions/52755322151e_db_migration_to_back_fill_organizations.py
+++ b/lms/migrations/versions/52755322151e_db_migration_to_back_fill_organizations.py
@@ -1,0 +1,194 @@
+"""
+DB migration to back-fill organizations.
+
+Revision ID: 52755322151e
+Revises: 79eda94de79f
+Create Date: 2022-09-07 18:11:19.706885
+
+"""
+import itertools
+from base64 import urlsafe_b64encode
+from collections import Counter
+from uuid import uuid4
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.ext.declarative import declarative_base
+
+# revision identifiers, used by Alembic.
+revision = "52755322151e"
+down_revision = "79eda94de79f"
+
+Base = declarative_base()
+
+
+class ApplicationInstance(Base):
+    """Class to represent a single lms install."""
+
+    __tablename__ = "application_instances"
+
+    id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
+
+    #: A unique identifier for the LMS instance.
+    tool_consumer_instance_guid = sa.Column(sa.UnicodeText, nullable=True)
+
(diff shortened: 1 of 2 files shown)


lms/migrations/versions/52755322151e_db_migration_to_back_fill_organizations.py
```diff
@@ -0,0 +1,194 @@
...
+
+    return longest_name
+
+
+def upgrade():
+    db_session = sa.orm.Session(bind=op.get_bind())
+
+    print("Clearing old organizations...")
+    op.execute("UPDATE application_instances SET organization_id = NULL")
+    op.execute("DELETE FROM organization")
```
jon-betts, line 154: Doing this allows this to run after auto assigned orgs have been created, which would probably be wrong

lms/migrations/versions/52755322151e_db_migration_to_back_fill_organizations.py
```diff
@@ -0,0 +1,194 @@
...
+        if not ais:
+            continue
+
+        organization = Organization(
+            name=pick_name([ai.tool_consumer_instance_name for ai in ais])
+        )
+        total += 1
+        print(
+            f"\tCreating '{organization.name}' for {len(ais)} application instance(s)"
+        )
```
jon-betts, line 183: This looks like it might print out a lot, but it should only create about 1200 max given the number of GUIDs we have

marcospri (approved):
- Tested the local side of things, all my AIs with GUIDs got a reasonable named organization.

- Run the .sql file on metabase prod database. I got a lot of numbers back. As the code is ported over sparky I have to assume those are correct.

Not much else to comment on the migration code itself, LGTM :+1:

Issue #4202, opened 2022-07-08
Auto assign new application instances to organizations


jon-betts:
* On creating a new application instance by default it won't be associated with any org
   * **Alternative**: New AI's start with an org in a non-launched state which is potentially destroyed later on
* When it's launched we try and match by GUID:
  * 0 matches - Create their own org
  * 1 match - Join the matched org
  * 2+ matches
* Create their own org? And also alert or something? Mark as orphaned?
   *Alternative: Arbitrarily associate with the first hit and fix it later?
* If the org has no name, we should try and fill it out

### Technical questions

 * Resolve all the decisions above
 * Does this have any knock on for the DB design in https://github.com/hypothesis/lms/issues/4200

### Decisions taken

 * AI's start off not associated with orgs at all
 * If there's two or more matches, we'll pick the first arbitrarily
 * We will log a warning though, so we can keep an eye on this if we are curious

## Tasks

 * https://github.com/hypothesis/lms/pull/4332

jon-betts:
Done!

PR #4332, opened 2022-08-12
Automatically assign application instance to organizations

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4202

This looks at the status of the application instances org and attempts to match it with the following behavior:

 * If there is no org
    * Search for matching orgs by guid either from the application instances or group info table
    * If one or more matches are found, use the first
    * If no matches are found, create a new org
    * Set the org to the application instance
  * If the org has no name, and the application instance `tool_consumer_instance_name` is present, set the name from it

## Review notes

This includes a refactor to move the updating of application instances into the service because:

 * That's where I expected to find it and I got confused
 * We are not expanding it beyond what's ok to have on a model

This could be a separate PR if you'd prefer it

## Testing notes

Mostly just noodle about in LMS and watch the table fill up

### DIFF
--- a/lms/db/__init__.py
+++ b/lms/db/__init__.py
@@ -1,5 +1,4 @@
 import logging
-from functools import cached_property
 
 import alembic.command
 import alembic.config
@@ -7,7 +6,7 @@ import sqlalchemy
 import zope.sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.orm import sessionmaker
 from sqlalchemy.orm.properties import ColumnProperty
 
 from lms.db._columns import varchar_enum
@@ -24,6 +23,7 @@ class BaseClass:
     @classmethod
     def columns(cls):
         """Return a list of all declared SQLAlchemy column names."""
+
         return [
             property.key
             for property in inspect(cls).iterate_properties
@@ -59,13 +59,10 @@ class BaseClass:
                 setattr(self, key, data[key])
 
     def __repr__(self):
-        return "{class_}({kwargs})".format(  # pylint: disable=consider-using-f-string
-            class_=self.__class__.__name__,
-            kwargs=", ".join(
-                f"{kwarg}={repr(getattr(self, kwarg))}"
-                for kwarg in self.__table__.columns.keys()  # pylint:disable=no-member
-            ),
+        kwargs = ", ".join(
+            f"{column}={repr(getattr(self, column))}" for column in self.columns()
         )
+        return f"{self.__class__.__name__}({kwargs})"
 
 
 BASE = declarative_base(
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -5,7 +5,7 @@ from typing import List
 
 from sqlalchemy.exc import NoResultFound
 
-from lms.models import ApplicationInstance, LTIRegistration
+from lms.models import ApplicationInstance, LTIParams, LTIRegistration
 from lms.services.aes import AESService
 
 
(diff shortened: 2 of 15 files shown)


lms/db/__init__.py
```diff
@@ -1,13 +1,12 @@
 import logging
-from functools import cached_property
```
jon-betts, line 2: How as this not picked up? Do we have a linting problem?

lms/db/__init__.py
```diff
@@ -1,13 +1,12 @@
...
 
 import alembic.command
 import alembic.config
 import sqlalchemy
 import zope.sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.orm import sessionmaker
```
jon-betts, line 9: Or this?

lms/db/__init__.py
```diff
@@ -59,13 +59,10 @@ def update_from_dict(self, data, skip_keys=None):
                 setattr(self, key, data[key])
 
     def __repr__(self):
-        return "{class_}({kwargs})".format(  # pylint: disable=consider-using-f-string
-            class_=self.__class__.__name__,
-            kwargs=", ".join(
-                f"{kwarg}={repr(getattr(self, kwarg))}"
-                for kwarg in self.__table__.columns.keys()  # pylint:disable=no-member
-            ),
```
jon-betts, line 67: Our roll your own here didn't account for cases where the column name did not match the attribute.

This caused problems with `_public_id` which was worse than just not representing it correctly, as we have a method called `public_id` which when stringified, stringifies the object, which cases infinite recursion.

lms/services/application_instance.py
```diff
@@ -157,10 +165,42 @@ def create_application_instance(  # pylint:disable=too-many-arguments
...
+        tool_consumer_instance_guid = params.get("tool_consumer_instance_guid")
+        if not tool_consumer_instance_guid:
+            # guid identifies the rest of the LMS data, if not there skip any
+            # updates
+            return
+
+        application_instance.check_guid_aligns(tool_consumer_instance_guid)
+
+        # Make sure this application instance is associated with an org.
+        self._organization_service.auto_assign_organization(application_instance)
```
jon-betts, line 185: This addition here is why this had to move, as we need more info that we can get from looking at the instance itself

lms/services/organization.py
```diff
@@ -0,0 +1,78 @@
...
+        Automatically associate an application instance with an org by GUID.
+
+        If there is no GUID matching is skipped, and we return `None`. If no
+        match can be found, then a new organization is created.
+        """
+
+        # We can't match by GUID if there isn't one
+        guid = application_instance.tool_consumer_instance_guid
+        if not guid:
+            return None
```
marcospri, line 51: This should not happen on the context of a launch but it's good the function guards against it anyway.

lms/services/organization.py
```diff
@@ -0,0 +1,78 @@
...
+                    "Multiple organization matches found for application instance %s",
+                    application_instance.id,
+                )
+            org = orgs[0]
+
+        else:
+            org = Organization()
+
+        # Fill out missing names
+        if not org.name and (name := application_instance.tool_consumer_instance_name):
```
marcospri, line 68: Not sure about filling the name here, it might make the ORG <=> GUID association stronger.

It could also dis-encourage setting proper names if the previous value  was `xxxx--xxxx-xxxx-xxxxxxx` changing it to "University of XX" feels wrong.

On the other hand having 100% of empty names is not ideal either.

lms/db/__init__.py
```diff
@@ -59,13 +59,10 @@ def update_from_dict(self, data, skip_keys=None):
...
+        kwargs = ", ".join(
+            f"{column}={repr(getattr(self, column))}" for column in self.columns()
```
marcospri, line 63: To be honest I've never seen this top level `__repr__` before. This works better and it's sorter.

lms/services/application_instance.py
marcospri: Yeah I think having it here makes sense and we don't have to think that much about where to put things in the future for any model.

lms/db/__init__.py
marcospri: I can't spot the problem with a short look around .pylintrc

lms/services/organization.py
```diff
@@ -0,0 +1,78 @@
...
+            self._db_session.query(Organization)
+            .join(ApplicationInstance)
+            .outerjoin(GroupInfo)
+            .filter(
+                sa.or_(
+                    ApplicationInstance.tool_consumer_instance_guid == guid,
+                    GroupInfo.tool_consumer_instance_guid == guid,
+                )
+            )
+            .all()
```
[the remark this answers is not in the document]
jon-betts: We could do this by latest updated date? We don't get pinged on every launch, but this still seems decent.

marcospri (approved):
We talked about this one in slack and meet, all looking good :+1:

PR #4333, opened 2022-08-13
Use a separate venv to run `pip-compile`

seanh:
Use a separate `pipcompile` venv to run `pip-compile` in, instead of running it in each of the individual venvs (`dev`, `tests`, etc).

This makes the `Makefile` much simpler while retaining all of the functionality that we had before:

- `pip-compile` is still run with the right version of Python. Since all an app's tox envs use the same version of Python `pip-compile` just needs to be run in **a** tox environment, it doesn't need to be run in the particular tox environment for the requirements file that it's compiling. (Really `pip-compile` doesn't need to be run in tox at all, it just needs to be run with the right version of Python, but `tox` is just the thing that we have at hand for installing a tool and keeping it uptodate and is what we already do for other tools e.g. Docker Compose, coverage.py).

- Each project still has a pinned version of `pip-tools` that it uses to run `pip-compile`. That pinned version is now in `requirements/pipcompile.txt`

  Note that all the other `requirements/*.txt` files will still include `pip-tools` as well because they all depend on `pip-sync-faster` for syncing the environment and `pip-sync-faster` depends on `pip-tools` (for `pip-sync`). But the environments no longer have a direct dependency on `pip-tools` only an indirect one via `pip-sync-faster`. The new `pipcompile` environment is now the only environment that has a direct dependency on `pip-tools`.

- It can still remake all the `.txt` files from scratch. This still works, which the cookiecutter depends on:

  ```terminal
  $ rm -rf requirements/*.txt .tox
  $ make requirements
  ```

The environment name is `pipcompile` not `pip-compile` because `-` has a special meaning in tox environment names (this matches the naming style of existing environments like `dockercompose` for the same reason).

The environment name is `pipcompile` not `piptools` because the environment is only used to run `pip-compile`. `pip-tools` just happens to be the name of the package that it has to install in order to get `pip-compile`.


seanh:
Something is slightly wrong with this: it's using the wrong copy of `pip-compile` (one from outside of `tox`) to compile the first requirements file but not the others.

PR #4343, opened 2022-08-16
Make mocked async method, in SubmitGradeForm tests, reject rather than throw

robertknight:
Fix a mock for an async method to reject rather than throw synchronously.
This required changes to how the test waits for changes in the rendered
output.

The test passed by accident previously because all the state updates within the
component happened to be triggered synchronously.


lyzadanger (approved):
Thank you!

PR #4346, opened 2022-08-16
New mechanism to display values from the backend in the browser's console

marcospri:
Discussion: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1660658193402049

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -167,6 +167,9 @@ class JSConfig:
             "allowedOrigins": self._request.registry.settings["rpc_allowed_origins"]
         }
 
+        ai = self._context.application_instance
+        self._config["debug"]["values"]["LTI version"] = ai.lti_version
+
     def enable_file_picker_mode(self, form_action, form_fields):
         """
         Put the JavaScript code into "file picker" mode.
@@ -368,8 +371,12 @@ class JSConfig:
                 "authToken": self._auth_token()
             },
             "canvas": {},
-            # Some debug information, currently used in the Gherkin tests.
-            "debug": {"tags": []},
+            "debug": {
+                # Some debug information, currently used in the Gherkin tests.
+                "tags": [],
+                # Info dumped to the console to help support.
+                "values": {},
+            },
             # Tell the JavaScript code whether we're in "dev" mode.
             "dev": self._request.registry.settings["dev"],
             # What "mode" to put the JavaScript code in.
--- a/lms/static/scripts/frontend_apps/config.js
+++ b/lms/static/scripts/frontend_apps/config.js
@@ -156,6 +156,14 @@ import { createContext } from 'preact';
  */
 
 /**
+ * Debug information sent by the backend.
+ *
+ * @typedef DebugInfo
+ * @prop {string[]} [tags]
+ * @prop {Record<string,any>} [values]
+ */
+
+/**
  * Data/configuration needed for frontend applications in the LMS app.
  * The `mode` property specifies which frontend application should load and
  * the available configuration/data depends on the mode and LTI user role.
@@ -181,6 +189,7 @@ import { createContext } from 'preact';
  * @prop {string} viaUrl
  * @prop {OAuthErrorConfig} OAuth2RedirectError
  * @prop {ErrorDialogConfig} errorDialog
+ * @prop {DebugInfo} [debug]
  */
 
 /**
--- a/lms/static/scripts/frontend_apps/index.js
+++ b/lms/static/scripts/frontend_apps/index.js
@@ -35,6 +35,16 @@ export function init() {
    */
   const services = new Map();
 
+  if (config.debug?.values && Object.keys(config.debug.values).length) {
+    /* eslint-disable no-console */
+    console.groupCollapsed('Hypothesis debug info');
+    Object.entries(config.debug.values).forEach(([key, value]) =>
+      console.log(key + ': ' + value)
+    );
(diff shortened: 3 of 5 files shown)


lms/static/scripts/frontend_apps/index.js
```diff
@@ -35,6 +35,16 @@ export function init() {
    */
   const services = new Map();
 
+  if (config?.debug?.values && Object.keys(config.debug.values).length) {
```
robertknight, line 38:
```suggestion
  if (config.debug?.values && Object.keys(config.debug.values).length) {
```

The `readConfig` call above will either return a `config` object or throw.

lms/static/scripts/frontend_apps/test/index-test.js
```diff
@@ -89,6 +89,20 @@ describe('LMS frontend entry', () => {
     });
   });
 
+  it('console logs debug values', () => {
+    const log = sinon.spy(console, 'log');
```
robertknight, line 93: This console method won't be restored if the test fails. You can fix this with a finally block:

```js
const log = sinon.stub(console, 'log');

try {
  // Test setup, asserts etc.
} finally {
  log.restore();
}
```

Also, you'll want to use `sinon.stub` here rather than `spy`, to prevent the spy from executing the real `console.log` and producing logging noise from the test runner.

lms/static/scripts/frontend_apps/index.js
```diff
@@ -35,6 +35,16 @@ export function init() {
    */
   const services = new Map();
 
+  /* eslint-disable no-console */
+  if (config?.debug?.values) {
+    console.groupCollapsed('Hypothesis debug');
+    Object.entries(config.debug.values).forEach(([key, value]) =>
+      console.log(key + ': ' + value)
```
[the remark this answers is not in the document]
robertknight: > We want to eventually move this to an HTML representation on the client and having a nested structure there might be more difficult to display.

Indeed. Flat key/value lists are simpler to deal with at the UI level.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -155,6 +155,14 @@ import { createContext } from 'preact';
  * @prop {string} itemId
  */
 
+/**
+ * Debug information sent by the backend.
+ *
+ * @typedef DebugInfo
+ * @prop {string[]} [tags]
+ * @prop {Record<string,any>} [values]
```
robertknight, line 163: Should these properties be optional, or just the presence of the `debug` field in the config?

lms/static/scripts/frontend_apps/index.js
```diff
@@ -35,6 +35,16 @@ export function init() {
    */
   const services = new Map();
 
+  if (config?.debug?.values && Object.keys(config.debug.values).length) {
+    /* eslint-disable no-console */
+    console.groupCollapsed('Hypothesis debug');
```
robertknight, line 40:
```suggestion
    console.groupCollapsed('Hypothesis debug info');
```

Sounds better as a heading with " info" on the end.

robertknight (changes requested):
Hi @marcospri - I added some comments with minor corrections and suggestions. Otherwise this looks good.

lms/static/scripts/frontend_apps/config.js
marcospri: I reckon combining different types of app it should be optional in all cases.

lms/static/scripts/frontend_apps/test/index-test.js
```diff
@@ -89,6 +89,23 @@ describe('LMS frontend entry', () => {
     });
   });
 
+  it('console logs debug values', () => {
+    const log = sinon.stub(console, 'log');
+
+    try {
+      fakeReadConfig.returns({
+        ...minimalConfig,
+        ...{ debug: { values: { key: 'value' } } },
```
robertknight, line 98:
```suggestion
        debug: { values: { key: 'value' } },
```

lms/static/scripts/frontend_apps/test/index-test.js
```diff
@@ -89,6 +89,23 @@ describe('LMS frontend entry', () => {
...
+      });
+
+      init();
+
+      assert(log.calledWith('key: value'));
```
robertknight, line 103:
```suggestion
      assert.calledWith(log, 'key: value');
```

You get more useful assert output in the event of failure if using `assert.someCondition(<args>)` instead of `assert(<some boolean condition>)`. The `assert` function on its own is much dumber than how it behaves in Python/pytest.

robertknight (approved):

jon-betts (approved):
This works and gives us a good place to dump out content.

Supporting nested structures would allow us to present significantly more data without it being totally impossible to navigate, but it's not essential.

I'd be strongly tempted to dump all of the translated LTI params, as we can't easily read them for 1.3 installs without decoding the JWT.

lms/resources/_js_config/__init__.py
```diff
@@ -167,6 +167,9 @@ def enable_lti_launch_mode(self, assignment: Assignment):
             "allowedOrigins": self._request.registry.settings["rpc_allowed_origins"]
         }
 
+        ai = self._context.application_instance
+        self._config["debug"]["values"]["LTI version"] = ai.lti_version
```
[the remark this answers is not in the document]
marcospri: We might end up there. But for example dumping the whole LTI Params might introduce confusion on what we got on the JWT vs how we interpreted it.

PR #4351, opened 2022-08-17
Re-instate the non-SSO VitalSource path as a fallback

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4349

This allows the original, non SSO, non institutional key solution for VitalSource to continue to work. This will allow us to release this to our customers who have their own API keys without disrupting those that don't.

### Behavior

 * If we have a customer key we will use it in preference to the global key for metadata lookup
 * If we have a customer key **and** a user id field, we will enable SSO

This leads to a few different modes of operation:

 * Global key lookup, no SSO
 * User key lookup, no SSO
 * User key lookup + SSO

## Merging notes

This needs to be merged into `vs-sso` not `main` to enable `vs-sso` to be safe to release without having all the customer details updated.

## Review notes

You may wish to review this as a [diff against `main`](https://github.com/hypothesis/lms/compare/main...vs-sso-fallback) instead of `vs-sso` as this is what we will be merging in the end.

## Testing notes

We are going to setup the three modes described above. In each case picking files should be the same, although a different key is in operation. It's not really currently present to tell the difference between whether the global key is working or the built in one as they are both the same key. So to prove this we are going to break the local one.

#### Proving the SSO flow

 * `make devdata`
 * Edit: `.devdata.env` and write nonsense in the `VITALSOURCE_API_KEY` field
 * Start all the services
 * Visit: https://bookshelf.vitalsource.com/#/ and logout
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/3154 and re-configure it
 * Pick the localhost optional and add a Vitalsource documemt
 * Enter URL: https://bookshelf.vitalsource.com/reader/books/L-999-70867/pages/recent
 * Save and proceed to the assignment
 * It should work
 * Visit: https://bookshelf.vitalsource.com/#/, you should be logged back in

#### Proving the global key, no SSO

 * http://localhost:8001/admin/instance/id/102/
 * Check that Vitalsource is enabled by the key is blank
 * Visit: https://bookshelf.vitalsource.com/#/ and logout
 * Visit: https://hypothesis.instructure.com/courses/319/assignments/3310 and re-configure it
 * Pick the localhost optional and add a Vitalsource documemt
 * Enter URL: https://bookshelf.vitalsource.com/reader/books/L-999-70867/pages/recent
 * You should see "External request failed"
    * This is because we broke the key
 * `make devdata` and restart `make dev`
 * Try this again
 * Picking should now work, but viewing the assignment should show you a login screen
    * This proves SSO is disabled

#### Proving a local key with no SSO

 * Visit: http://localhost:8001/admin/instance/id/102/
 * Add the `VITALSOURCE_API_KEY` from `.devdata.env` as an API key
 * Edit: `.devdata.env` and write nonsense in the `VITALSOURCE_API_KEY` field
 * Follow the instructions for "Proving the global key, no SSO"
 * Everything should be the same, except the initial failure should not happen because we will use the customer API key

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -68,14 +68,22 @@ class JSConfig:
                 ),
             }
         elif document_url.startswith("vitalsource://"):
-            svc = self._request.find_service(VitalSourceService)
+            svc: VitalSourceService = self._request.find_service(VitalSourceService)
 
-            # nb. VitalSource doesn't use Via, but is otherwise handled exactly
-            # the same way by the frontend.
-            self._config["viaUrl"] = svc.get_launch_url(
-                user_reference=self._request.lti_params[svc.user_lti_param],
-                document_url=document_url,
-            )
+            if svc.sso_enabled:
+                # nb. VitalSource doesn't use Via, but is otherwise handled
+                # exactly the same way by the frontend.
+                self._config["viaUrl"] = svc.get_sso_redirect(
+                    user_reference=self._request.lti_params[svc.user_lti_param],
+                    document_url=document_url,
+                )
+            else:
+                # This looks a bit silly, but pretty soon the above will
+                # be setting `api.viaURL` not `viaURL`
+                self._config["viaUrl"] = svc.get_book_reader_url(
+                    document_url=document_url
+                )
+
         elif jstor_service.enabled and document_url.startswith("jstor://"):
             self._config["viaUrl"] = jstor_service.via_url(self._request, document_url)
             self._config["contentBanner"] = {
--- a/lms/services/vitalsource/service.py
+++ b/lms/services/vitalsource/service.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Optional
 
 from lms.services.vitalsource._client import VitalSourceClient
 from lms.services.vitalsource.exceptions import VitalSourceError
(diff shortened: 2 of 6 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -68,14 +68,22 @@ def add_document_url(self, document_url):
...
+                self._config["viaUrl"] = svc.get_sso_redirect(
+                    user_reference=self._request.lti_params[svc.user_lti_param],
+                    document_url=document_url,
+                )
+            else:
+                # This looks a bit silly, but pretty soon the above will
+                # be setting `api.viaURL` not `viaURL`
+                self._config["viaUrl"] = svc.get_book_reader_url(
+                    document_url=document_url
+                )
```
jon-betts, line 85: In the next PR (once rebased) this will be using different config locations as one will be setting a callback

tests/unit/lms/resources/_js_config/__init___test.py
```diff
@@ -121,21 +121,36 @@ def test_it_adds_the_viaUrl_api_config_for_Blackboard_documents(self, js_config)
...
+
+        js_config.add_document_url(document_url)
+
+        vitalsource_service.get_book_reader_url.assert_called_with(
+            document_url=document_url
+        )
+        assert (
+            js_config.asdict()["viaUrl"]
+            == vitalsource_service.get_book_reader_url.return_value
         )
```
jon-betts, line 154: I couldn't find a nice way to combine these tests that wasn't worse

tests/unit/lms/services/vitalsource/factory_test.py
```diff
@@ -6,11 +6,12 @@
 from lms.services.vitalsource.factory import service_factory
 
 
+@pytest.mark.usefixtures("application_instance_service")
 class TestServiceFactory:
```
jon-betts, line 10: I really hate these tests. I'm coming to the conclusion that it's easier to test 100 lines of code with no-ifs than 5 lines of code with 5 ifs.

marcospri (commented):
This works and will help with deploying the rest :+1: 

It ended up having more cases that we hope for but that's usually how integrations go :shrug:

lms/services/vitalsource/service.py
```diff
@@ -11,36 +11,61 @@ class VitalSourceService:
...
-            reference
+        :param enabled: Is VitalSource enabled for the customer?
+        :param global_client: Client for making generic API calls
+        :param customer_client: Client for making customer specific API calls
+        :param user_lti_param: Field to lookup user details for SSO
         """
-        self._client = client
+
         self._enabled = enabled
+        self._metadata_client = customer_client or global_client
```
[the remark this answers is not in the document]
jon-betts: I've added some extra commentary here.

marcospri (approved):

PR #4352, opened 2022-08-17
Add a new field for VitalSource user field regexes

jon-betts:
Also change the names a bit to make it clear the user is about SSO.

## Review notes

This PR is a bit longer than it really should have to have been as we didn't have a way of raising a new exception without:

 * Writing a custom handler for it in the `views.exceptions.py`
 * Writing a custom entry in the Javscript for it

This builds on the `SerializableError` work for the API's which introduced an exception you can raise and expect sensible behavior from the front end.

This really shouldn't be in the location it is in `services.exceptions`, but we don't have a generic `exceptions.py` and I don't want to get into that right now.

## Testing notes

 * `make devdata` and start everything
 * Visit a VS assignment (e.g. https://hypothesis.instructure.com/courses/125/assignments/3158)
 * This should work as normal

Error case: too many or too few capture groups

 * Visit: http://localhost:8001/admin/instance/id/8/
 * Edit the regex to contain "no_group_here" or "too (many) capture (groups)"
 * Refresh the assignment
 * You should see an error about the regex

Error case: malformed regex

 * As above with `[` as the regex

With a valid regex:

 * Open the dev tools in the assignment
 * Edit the regex to contain `^(.{4})`
 * Refresh the assignment
 * It should fail saying the user is not found
 * Filter the network tab by `user_reference`
 * You should see a call with a user reference only 4 chars long

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -78,9 +78,9 @@ class JSConfig:
                     "path": self._request.route_url(
                         "vitalsource_api.launch_url",
                         _query={
-                            "user_reference": self._request.lti_params[
-                                svc.user_lti_param
-                            ],
+                            "user_reference": svc.get_user_reference(
+                                self._request.lti_params
+                            ),
                             "document_url": document_url,
                         },
                     )
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -2,6 +2,8 @@ import functools
 from enum import Enum
 from typing import List, Optional
 
+from pyramid.httpexceptions import HTTPClientError
+
 from lms.models import Assignment, GroupInfo, Grouping, HUser
 from lms.product.blackboard import Blackboard
 from lms.product.canvas import Canvas
@@ -154,7 +156,7 @@ class JSConfig:
         if error_details:
             self._config["OAuth2RedirectError"]["errorDetails"] = error_details
 
-    def enable_error_dialog_mode(self, error_code, error_details=None):
+    def enable_error_dialog_mode(self, error_code, error_details=None, message=None):
         self._config.update(
             {
                 "mode": JSConfig.Mode.ERROR_DIALOG,
@@ -164,6 +166,9 @@ class JSConfig:
 
         if error_details:
             self._config["errorDialog"]["errorDetails"] = error_details
+
+        if message:
+            self._config["errorDialog"]["errorMessage"] = message
 
     def enable_lti_launch_mode(self, assignment: Assignment):
         """
--- a/lms/services/vitalsource/service.py
+++ b/lms/services/vitalsource/service.py
@@ -1,15 +1,17 @@
+import re
 from typing import List, Optional
 
+from lms.models import LTIParams
 from lms.services.vitalsource._client import VitalSourceClient
-from lms.services.vitalsource.exceptions import VitalSourceError
+from lms.services.vitalsource.exceptions import (
+    VitalSourceError,
+    VitalSourceMalformedRegex,
+)
 from lms.services.vitalsource.model import VSBookLocation
 
 
 class VitalSourceService:
     """A high-level interface for dealing with VitalSource."""
-
-    user_lti_param = None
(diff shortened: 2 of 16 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -3,7 +3,6 @@
 from typing import List, Optional
 
 from lms.models import Assignment, GroupInfo, Grouping, HUser
-from lms.product import Product
```
jon-betts, line 6: Wat? How was this not a problem before?

lms/services/vitalsource/service.py
```diff
@@ -65,22 +69,65 @@ def get_book_reader_url(cls, document_url) -> str:
...
+
         if not self._sso_client.get_user_book_license(user_reference, loc.book_id):
             raise VitalSourceError("vitalsource_no_book_license")
 
         return self._sso_client.get_sso_redirect(
             user_reference, self.get_book_reader_url(document_url)
         )
+
+    @staticmethod
+    def compile_user_lti_pattern(pattern: str) -> Optional[re.Pattern]:
```
jon-betts, line 96: I've made this public to make it easy to test without getting told off, but also because I thought it might be of interest to the admin pages if we end up checking things more thoroughly there.

lms/resources/_js_config/__init__.py
jon-betts, line 6: How is this possible?

tests/unit/lms/views/exceptions_test.py
```diff
@@ -6,102 +6,108 @@
...
 class TestExceptionViews:
-    def test_notfound(self, assert_response, pyramid_request):
+    def test_notfound(self, pyramid_request):
         exception = httpexceptions.HTTPNotFound()
 
         template_data = ExceptionViews(exception, pyramid_request).notfound()
 
-        assert_response(template_data, 404, message="Page not found")
+        assert pyramid_request.response.status_int == 404
+        assert template_data == {"message": "Page not found"}
```
jon-betts, line 21: I've changed this because:

* Using the helper was more code than not using it
* It made reading tests harder
* It was causing all sorts of coverage issues with the conditionality in it

lms/views/exceptions.py
```diff
@@ -60,6 +60,20 @@ def reused_consumer_key(self):
...
+    def serializable_error(self):
+        self.request.response.status_int = 400
+
+        self.request.context.js_config.enable_error_dialog_mode(
+            error_code=self.exception.error_code,
+            message=self.exception.message,
+            error_details=self.exception.details,
+        )
+
+        return {}
```
jon-betts, line 75: If you look at the `ReusedConsumerKey` case here it could be completely covered by turning that into a child of `SerializableError`.

marcospri (approved):
I have a couple of questions around the error handling but this works and the error handling is probably the best we can do without a big refactor.

lms/views/admin/application_instance.py
```diff
@@ -312,6 +312,7 @@ def update_instance(self):
             ("microsoft_onedrive", "files_enabled", bool),
             ("vitalsource", "enabled", bool),
             ("vitalsource", "user_lti_param", str),
+            ("vitalsource", "user_lti_pattern", str),
```
marcospri, line 315: I think we should try `compile_user_lti_pattern` at some point in this view in a follow up PR.

jon-betts: Yeah, that was the idea in making it public.

lms/views/exceptions.py
```diff
@@ -60,6 +60,20 @@ def reused_consumer_key(self):
 
         return {}
 
+    @exception_view_config(
+        context=SerializableError, renderer="lms:templates/error_dialog.html.jinja2"
+    )
+    def serializable_error(self):
```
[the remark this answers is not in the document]
jon-betts: It kind of doesn't, as that one catches things in the API space specifically, and this one catches all other view exceptions. 

I suppose I could bump the `SerializableError` down into `views.exceptions` instead of `views.api.exceptions`?

PR #4357, opened 2022-08-18
Use our own PyJWKClient instead of the one provided by PyJWT

marcospri:
Our version uses requests instead of urllib which provides an user agent
by default but also allows for further customizations down the line.

This is motivated by school having problems with this. I reckon it makes sense to have control over this in any case.

# Testing

- On `main` test opening an LTI1.3 assignment with a modified JWK enpoint

```tox -qe dockercompose -- exec postgres psql -U postgres -c "update lti_registration set key_set_url = 'https://elearning.aua.am/mod/lti/certs.php' where id = 3"```

- Open https://hypothesis.instructure.com/courses/319/assignments/3308

Fails with

```
web (stderr)         |   File "/home/marcos/.pyenv/versions/3.8.12/lib/python3.8/urllib/request.py", line 649, in http_error_default
web (stderr)         |     raise HTTPError(req.full_url, code, msg, hdrs, fp)
web (stderr)         | urllib.error.HTTPError: HTTP Error 403: Forbidden
```

- Open https://hypothesis.instructure.com/courses/319/assignments/3308 again on `jwt-requests`

Now it fails with:

```2022-08-19 11:26:58,809 DEBUG [lms.services.jwt:98][MainThread] Invalid JWT. Unable to find a signing key that matches: "2022-07-01T00:00:03Z_0bbb730e-14aa-4420-9efa-f03583c8b777"```

which makes sense because we are using the JWKs from another LMS but the JWK was read succesful.

- Check a successful launch in canvas, go back to the canvas JWKs

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "update lti_registration set key_set_url = 'https://canvas.instructure.com/api/lti/security/jwks' where id = 3"
```

An launch https://hypothesis.instructure.com/courses/319/assignments/3308

which should succeed now.

### DIFF
--- a/lms/services/jwt.py
+++ b/lms/services/jwt.py
@@ -4,6 +4,7 @@ import logging
 from functools import lru_cache
 
 import jwt
+import requests
 from jwt.exceptions import ExpiredSignatureError, InvalidTokenError, PyJWTError
 
 from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
@@ -115,7 +116,21 @@ class JWTService:
         PyJWKClient maintains a cache of keys it has seen we want to keep
         the clients around with `lru_cache` in case we can reuse that internal cache
         """
-        return jwt.PyJWKClient(jwk_url)
+        return _RequestsPyJWKClient(jwk_url)
+
+
+class _RequestsPyJWKClient(jwt.PyJWKClient):
+    """
+    Version of PyJWKClient which uses requests to gather JWKs.
+
+    Having our own class and using request allows for easier customization.
+    """
+
+    def fetch_data(self):
+        # We found that some Moodle instances return 403s
+        # on request without an User-Agent.
+        with requests.get(self.uri, headers={"User-Agent": "requests"}) as response:
+            return response.json()
 
 
 def factory(_context, request):
--- a/tests/unit/lms/services/jwt_test.py
+++ b/tests/unit/lms/services/jwt_test.py
@@ -12,7 +12,13 @@ from pyramid.config import Configurator
 from pytest import param
 
 from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
-from lms.services.jwt import JWTService, _get_lti_jwt, factory, includeme
+from lms.services.jwt import (
+    JWTService,
+    _get_lti_jwt,
+    _RequestsPyJWKClient,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lms/services/jwt.py [hunk 1]
```diff
@@ -115,7 +116,21 @@ def _get_jwk_client(jwk_url: str):
...
+
+
+class _RequestsPyJWKClient(jwt.PyJWKClient):
+    """
+    Version of PyJWKClient which uses requests to gather JWKs.
+
+    Having our own class and using request allows for easier customization.
+    """
+
+    def fetch_data(self):
```
marcospri, line 129: Original version at:

https://github.com/jpadilla/pyjwt/blob/a988e1a11e5abb5869dd641f3f4f6a5bb4e70fdf/jwt/jwks_client.py#L20

We could have just add the headers and keep using urllib but using requests feels more "idiomatic" in our codebase.

tests/unit/lms/services/jwt_test.py
```diff
@@ -152,6 +164,22 @@ def svc(self, lti_registration_service, rsa_key_service):
         return svc
 
 
+class Test_RequestsPyJWKClient:
```
marcospri, line 167: I think `_RequestsPyJWKClient` should be private but also tested independently.

Here we test our changes to the class and the test that use the mocked version tests how it all plays together.

jon-betts (approved):
Yep this works as described. I like splitting out and testing a sub-chunk of the problem. I think this is a good enough reason to split off a helper class sometimes and is something we should consider more for big and complicated classes with repeated chunks of behavior.

lms/services/jwt.py
```diff
@@ -115,7 +116,21 @@ def _get_jwk_client(jwk_url: str):
...
+        # We found that some Moodle instances return 403s
+        # on request without an User-Agent.
+        with requests.get(self.uri, headers={"User-Agent": "requests"}) as response:
+            return response.json()
```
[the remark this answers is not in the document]
marcospri: > I think requests sends it's own user agent by default?

It does this is a bit more explicit about what's it's doing. I wrote a version of that and tried to assert that there was indeed an user agent and it become a bit too complex on the testing side (`request.utils.default_heaaders()`)

PR #4366, opened 2022-08-23
Modify backfilling migration to work on batches.

marcospri:
Original migration timed out on production:

https://jenkins.hypothes.is/job/lms-tasks/174/console

Running the query side on metabase seem to finish in a reasonable time so the problem might be on inserting side. 

This, the whole table query, finish in metabase:

```
        SELECT
                lti_launches.created,
                1,
                application_instances.id,
                MAX(grouping.id)
        FROM lti_launches
        LEFT OUTER JOIN application_instances
            ON lti_launches.lti_key = application_instances.consumer_key
        LEFT OUTER JOIN grouping
            ON context_id = grouping.lms_id
                AND grouping.application_instance_id = application_instances.id
                AND grouping.type = 'course'
       -- Only backfill rows created before we also start tracking them directly on this table
        WHERE lti_launches.created < '2022-08-17T13:00:08.241'
        -- We need to group by to pick one (MAX grouping.id) above of the possible
        -- multiple courses on the grouping table
        GROUP BY lti_launches.id, lti_launches.created, application_instances.id
        -- Ordering will keep new rows physically sorted on disk which should later help while querying
        ORDER BY lti_launches.created ASC
```

The  batch on this PR covers 1,123,235 rows (out of 13M) 

# Testing

(You might have to `hdev alembic downgrade -1` first depending on the state of your DB)

- Start with empty tables

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate event,lti_launches cascade"
```

- Insert some test rows (2 in the range of the batch, one outside)

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "insert into lti_launches (created, context_id, lti_key) values ('2001-08-17T13:00:08.241', 'CONTEXT', 'LTI'), ('2017-08-17T13:00:08.241', 'CONTEXT', 'LTI'), ('2021-08-17T13:00:08.241', 'CONTEXT', 'LTI');"
```

- Run the migration

`hdev alembic upgrade head`

Two rows are inserted

- Run it backwards `hdev alembic downgrade -1`

Two rows are deleted

### DIFF
--- a/lms/migrations/versions/396f46318023_events_backfill.py
+++ b/lms/migrations/versions/396f46318023_events_backfill.py
@@ -1,5 +1,5 @@
 """
-Backfill "events" from lti_launches rows.
+Backfill "events" from lti_launches rows (2020).
 
 Revision ID: 396f46318023
 Revises: 42b39684836f
@@ -13,11 +13,18 @@ from alembic import op
 revision = "396f46318023"
 down_revision = "42b39684836f"
 
+BATCH_START_DATE = "2000-08-17T13:00:08.241"
+BATCH_END_DATE = "2020-08-17T13:00:08.241"
+"""
+We'll do this in batches each bach between these two dates
+"""
 
 CUT_OFF_DATE = "2022-08-17T13:00:08.241"
 """
 Date when we started inserting launches directly on events.
 Rows in lti_launches would already have corresponding event after this date.
+
+While doing this in batches this will the END_DATE of the last batch.
 """
 
 
@@ -27,9 +34,7 @@ def get_configured_launch_type_pk(conn):
     ).fetchone()[0]
 
 
-def upgrade():
-    conn = op.get_bind()
-
+def migrate_lti_launches(conn, start, end):
     launch_event_type_id = get_configured_launch_type_pk(conn)
 
     result = conn.execute(
@@ -54,7 +59,7 @@ def upgrade():
                 AND grouping.application_instance_id = application_instances.id
                 AND grouping.type = 'course'
        -- Only backfill rows created before we also start tracking them directly on this table
-        WHERE lti_launches.created < '{CUT_OFF_DATE}'
+        WHERE lti_launches.created >= '{start}' AND lti_launches.created < '{end}'
         -- We need to group by to pick one (MAX grouping.id) above of the possible
         -- multiple courses on the grouping table
         GROUP BY lti_launches.id, lti_launches.created, application_instances.id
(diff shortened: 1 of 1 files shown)


lms/migrations/versions/396f46318023_events_backfill.py
```diff
@@ -27,9 +34,7 @@ def get_configured_launch_type_pk(conn):
     ).fetchone()[0]
 
 
-def upgrade():
-    conn = op.get_bind()
-
+def migrate_lti_launches(conn, start, end):
```
marcospri, line 37: Making these methods to import them in the next batches

lms/migrations/versions/396f46318023_events_backfill.py
```diff
@@ -65,17 +70,29 @@ def upgrade():
...
         DELETE FROM event
         WHERE type_id = {launch_event_type_id}
             -- All events inserted by the forward version of this migration
             -- will have empty assignments.
             AND assignment_id is null
-            AND timestamp < '{CUT_OFF_DATE}'
+            AND timestamp BETWEEN '{start}' AND '{end}'
     """
     )
+    print("\tDeleted events rows:", result.rowcount)
```
marcospri, line 86: Printing also on the downgrade, makes testing easier.

lms/migrations/versions/396f46318023_events_backfill.py
```diff
@@ -54,7 +59,7 @@ def upgrade():
                 AND grouping.application_instance_id = application_instances.id
                 AND grouping.type = 'course'
        -- Only backfill rows created before we also start tracking them directly on this table
-        WHERE lti_launches.created < '{CUT_OFF_DATE}'
+        WHERE lti_launches.created BETWEEN '{start}' and '{end}'
```
jon-betts, line 62: Between is inclusive of both extents (https://www.postgresql.org/docs/9.1/functions-comparison.html), so I think this is the same as:

```sql
lti_launches.created >= {start} AND lti_launches.created <= {end}
```

So unless we do something funky with the dates we might end up double counting any records on the boundary.

I think what we want to match what we had before is:

```suggestion
lti_launches.created >= {start} AND lti_launches.created < {end}
```

jon-betts (commented):
I've tried the instructions and they work as described.

I think there might be a potential issue with events on the exact boundary getting inserted twice when batches are run with different values. But it might not matter if there are no events exactly on the boundary.

lms/migrations/versions/396f46318023_events_backfill.py
jon-betts: Can you import directly, or will we be copy pasting?

lms/migrations/versions/396f46318023_events_backfill.py
marcospri: Yup, it might have not affected that many rows over the batches but better be on the safe side with migrations :+1:

lms/migrations/versions/396f46318023_events_backfill.py
marcospri: Not sure, I was thinking importing but we'd have to rename this migration file to be a valid python name.

Alembic doesn't care about the file names so we could do that on the second PR after this.

jon-betts (approved):
Some of the chat here was in Slack, but the summary is:

 * I've tested this and it works
 * There was some discussion about items on the boundaries of dates
 * This has now been fixed

PR #4367, opened 2022-08-23
Refactor admin pages to make the new vs upgraded instance actions more clear

marcospri:
Refactor the views/forms for new registrations.

The current form is not very clear on the intent as is using the same form/view two do two different things:

- Add a new AI to a registration
- Upgrade an existing AI while adding it to the registration.

This PR makes the split more explicit. 

There's a behavior change, in the version in `main` you can edit `lms url` and `email` (but not canvas key/secret) while upgrading an existing AI. Removing that option here, if you want to change those while upgrading you'll have to first upgrade and then update them on the AI details page.

### Old version
Screenshot from 2022-08-29 11-11-34

### New version
Screenshot from 2022-08-29 11-14-07

# Testing 

### New AI
- Click "Add instance"
- Complete the first form, the one labeled "Create new Application instance"
- You'll be taken to newly created AI.

### Update AI

- Check http://localhost:8001/admin/instance/Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b/. It should be marked as 1.1 at the top.
- Go back to http://localhost:8001/admin/registration/id/1/new/instance
- Use the second form with the consumer of the existing AI, `Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b` and any deployment_id
- You are taken the the same AI, now marked as 1.3

### DIFF
--- a/lms/templates/admin/instance.new.html.jinja2
+++ b/lms/templates/admin/instance.new.html.jinja2
@@ -19,18 +19,28 @@ New application instance
 
 
     <fieldset class="box mt-6">
-        <legend class="label has-text-centered">Application instance</legend>
-        {{ macros.form_text_field(request, "Consumer key", "consumer_key",
-            placeholder="Optional. Existing application instance's consumer key. It will be upgraded to LTI 1.3 using this registration") }}
+        <legend class="label has-text-centered">Create new Application instance</legend>
         {{ macros.form_text_field(request, "Deployment ID", "deployment_id") }}
         {{ macros.form_text_field(request, "LMS URL", "lms_url") }}
         {{ macros.form_text_field(request, "Email", "email") }}
-
         {{ macros.form_text_field(request, "Canvas developer key", "developer_key") }}
         {{ macros.form_text_field(request, "Canvas developer secret", "developer_secret") }}
     </fieldset>
 
     <input type="submit" class="button is-info" value="New"/>
+</form>
+
+<form method="POST" action="{{ request.route_url("admin.registration.upgrade.instance", id_=lti_registration.id) }}">
+    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+
+    <fieldset class="box mt-6">
+        <legend class="label has-text-centered">Upgrade exsiting Application instance</legend>
+        {{ macros.form_text_field(request, "Consumer key", "consumer_key",
+            placeholder="Existing application instance's consumer key. It will be upgraded to LTI 1.3 using this registration") }}
+        {{ macros.form_text_field(request, "Deployment ID", "deployment_id") }}
+    </fieldset>
+
+    <input type="submit" class="button is-info" value="Upgrade"/>
 </form>
 
 {% endblock %}
--- a/lms/templates/admin/registration.html.jinja2
+++ b/lms/templates/admin/registration.html.jinja2
@@ -11,7 +11,7 @@
     <fieldset class="box">
     <legend class="label has-text-centered">Registration application instances</legend>
         <div class="block has-text-right">
-            <a class="button is-primary" href="{{ request.route_url("admin.registration.new.instance", id_=registration.id) }}">New instance</a>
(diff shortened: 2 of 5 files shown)


lms/templates/admin/instance.new.html.jinja2
```diff
@@ -19,18 +19,28 @@ New application instance
...
         {{ macros.form_text_field(request, "Email", "email") }}
-
         {{ macros.form_text_field(request, "Canvas developer key", "developer_key") }}
         {{ macros.form_text_field(request, "Canvas developer secret", "developer_secret") }}
     </fieldset>
 
     <input type="submit" class="button is-info" value="New"/>
 </form>
 
+<form method="POST" action="{{ request.route_url("admin.registration.upgrade.instance", id_=lti_registration.id) }}">
```
marcospri, line 33: We have now two forms with only the fields that are relevant.

lms/templates/admin/registration.html.jinja2
```diff
@@ -11,7 +11,7 @@
     <fieldset class="box">
     <legend class="label has-text-centered">Registration application instances</legend>
         <div class="block has-text-right">
-            <a class="button is-primary" href="{{ request.route_url("admin.registration.new.instance", id_=registration.id) }}">New instance</a>
+            <a class="button is-primary" href="{{ request.route_url("admin.registration.new.instance", id_=registration.id) }}">Add instance</a>
```
marcospri, line 14: At this point we are adding a new instance to this registration. It could be a new one or upgrading an existing one.

Making the language a bit clearer.

lms/views/admin/application_instance.py
```diff
@@ -12,41 +12,6 @@
 from lms.views.admin import error_render_to_response, flash_validation
 
 
-class EmptyStringNoneMixin:
```
marcospri, line 15: Having two view makes this hack not needed anymore. It might be handy at some point. We can bring it back later.

lms/views/admin/application_instance.py
```diff
@@ -165,17 +139,7 @@ def upgrade_instance(self, lti_registration, consumer_key):
 
         # Set the LTI1.3 values
         application_instance.lti_registration = lti_registration
-        application_instance.deployment_id = self.request.params[
-            "deployment_id"
-        ].strip()
-
-        # And upgrade any other values present in the request
-        if email := self.request.params["email"].strip():
```
marcospri, line 173: We focus on upgrading. If you want to change any of this value you can do it directly on the application instance detail view.

lms/views/admin/application_instance.py
```diff
@@ -165,17 +139,7 @@ def upgrade_instance(self, lti_registration, consumer_key):
...
-            application_instance.requesters_email = email
-
-        if lms_url := self.request.params["lms_url"].strip():
-            application_instance.lms_url = lms_url
-
+        application_instance.deployment_id = deployment_id
         try:
             # Flush here to find if we are making a duplicate in the process of
             # upgrading
```
jon-betts, line 145: The layout of this class is a little chaotic. It seems `upgrade_instance` should be next to all these `<verb>_instance` methods?

jon-betts (approved):
I think this removes a potentially confusing interaction where you upgrade something rather than creating something new. I suspect this might be possible if you were copy pasting from things in a spreadsheet or something.

It's a nice bonus that it removes complication on the backend, which suggests this is going in the right direction.

As a review for this it's basically a thumbs up, but I wanted to outline some thoughts which don't have to be part of this PR but you might want to consider?

### Ideas I had

When following the instructions the "Danger zone" downgrade from 1.3 is present on 1.1 instances, and then raises an error. It seems an `if` around that would tidy this up.

### Bad ideas I had

I was thinking it's weird you downgrade in one place, but upgrade in another, but the registration has to be found somehow. I was thinking you could jump from the registration to the application instance then upgrade it, but this doesn't make sense a 1.1 application instances won't be associated with the registration.

PR #4370, opened 2022-08-23
Handle course copy across applications instance with the same GUID

marcospri:
## Requires

*  https://github.com/hypothesis/devdata/pull/62 needs to be merged

Course copy doesn't work across different application instances that belong to the same school (GUID now).

We look up in the DB for matching `File`s using the current application instance and won't find the original ones under a different AI.

This breaks course copy in the case of an Hypothesis install moving from course level to account level, some 1.1 -> 1.3 upgrades that create a new AI on our database, etc.

## Testing setup

The testing setup in canvas involves:

- An original course https://hypothesis.instructure.com/courses/319
which contains an Hypothesis course level install and a canvas file assignment.

- Copied course from 319 https://hypothesis.instructure.com/courses/376

In the course copy process the new course got a new assignment using the same canvas ID as the original but also a new LTI 1.3 install (with a new deployment ID) which is not configured a a new AI on `make devdata`.

The creation of the new LTI1.3 Canvas install (via course copy) is not the only way to replicate the issue, it would also happen in any scenario involving a new application instance (v1.1 or 1.3, moving to course level to instance/account level) / re-installing by the LMS admin...).

Using the LTI1.3 course level install duplication here because it's easier to setup and keep it isolated from other courses.

## Testing notes

* On `main`
* Launch the original assignment at least once as a teacher (to get the Files)
* https://hypothesis.instructure.com/courses/319/assignments/3347
* `make devdata`
* Launch https://hypothesis.instructure.com/courses/376/assignments/3357
    * Both as `eng+coursecopystudent` which doesn't have access to the original course's files 
    * ... and `Hypothesis 101 Teacher` who does.
* In both cases you see an error.

- Now switch to this branch `course-copy-across-ai`
- Relaunch as the student. Same error. This is normal as the student doesn't have access to the original file.
- Relaunch as the teacher. It works now.
- A file id mapping is now stored on the DB

```shell
tox -qe dockercompose -- exec postgres psql -U postgres -c "select extra from assignment order by id desc limit 1"
```

```
                   extra                    
--------------------------------------------
 {"canvas_file_mappings": {"1245": "1442"}}
(1 row)
```

- Relaunch as a student. It works now, using the mapping on the DB.

### DIFF
--- a/lms/services/canvas.py
+++ b/lms/services/canvas.py
@@ -93,7 +93,11 @@ class CanvasFileFinder:
         Return None if no matching file is found.
         """
         for file_id in file_ids:
-            file = self._file_service.get(file_id, type_="canvas_file")
+            # Use both `get` (the most common scenario, files in th same application instance)
+            # and `get_with_guid` as a fallback for installs in the same LMS with multiple application instances.
+            file = self._file_service.get(
+                file_id, type_="canvas_file"
+            ) or self._file_service.get_with_guid(file_id, type_="canvas_file")
 
             if not file:
                 continue
--- a/lms/services/file.py
+++ b/lms/services/file.py
@@ -1,6 +1,6 @@
 from sqlalchemy import func
 
-from lms.models import File
+from lms.models import ApplicationInstance, File
 from lms.services.upsert import bulk_upsert
 
 
@@ -19,6 +19,31 @@ class FileService:
                 type=type_,
             )
             .one_or_none()
+        )
+
+    def get_with_guid(self, lms_id, type_):
+        """
+        Return with the given lms_id and type_ that belongs to the same GUID as the current ApplicationInstance.
+
+        Once we have "Organization" or a similar entity that groups together multiple
+        ApplicationInstance we should use that concept instead of GUID but for now
+        this allows us to find files across multiple installs in for example
+        migrations from course/account/global level install or in some LTI1.3 upgrade
+        scenarios.
+        """
+        return (
+            self._db.query(File)
+            .join(
+                ApplicationInstance,
+                ApplicationInstance.tool_consumer_instance_guid
+                == self._application_instance.tool_consumer_instance_guid,
+            )
+            .filter(
+                File.lms_id == lms_id,
+                File.type == type_,
+            )
+            .order_by(File.id.desc())
+            .first()
         )
 
     def upsert(self, file_dicts):
--- a/tests/unit/lms/services/canvas_test.py
+++ b/tests/unit/lms/services/canvas_test.py
@@ -213,6 +213,16 @@ class TestCanvasFileFinder:
         canvas_api_client.list_files.assert_called_once_with(sentinel.course_id)
         assert matching_file_id == str(sentinel.matching_file_id)
 
+    def test_fallback_to_guid_matching(self, finder, file_service):
+        file_service.get.return_value = None
+
+        finder.find_matching_file_in_course(sentinel.course_id, [sentinel.file_id])
+
+        file_service.get.assert_called_once_with(sentinel.file_id, type_="canvas_file")
+        file_service.get_with_guid.assert_called_once_with(
+            sentinel.file_id, type_="canvas_file"
+        )
+
     def test_find_matching_file_in_course_with_multiple_file_ids(
         self, finder, canvas_api_client, file_service
     ):
@@ -225,6 +235,9 @@ class TestCanvasFileFinder:
             # The third file_id *will* be found in the course.
             matching_file,
         ]
+        # The with GUID fallback doesn't returning anything else
+        file_service.get_with_guid.return_value = None
+
         canvas_api_client.list_files.return_value = [
             {
                 "id": sentinel.matching_file_id,
--- a/tests/factories/application_instance.py
+++ b/tests/factories/application_instance.py
@@ -1,4 +1,4 @@
-from factory import Faker, make_factory
+from factory import Faker, Sequence, make_factory
 from factory.alchemy import SQLAlchemyModelFactory
 
 from lms import models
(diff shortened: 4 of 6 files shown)


lms/services/file.py
```diff
@@ -21,6 +21,31 @@ def get(self, lms_id, type_):
             .one_or_none()
         )
 
+    def get_with_guid(self, lms_id, type_):
+        """
+        Return with the given lms_id and type_ that belongs to the same GUID as the current ApplicationInstance.
```
marcospri, line 26: We could argue that the real fix for this is then doing "Organization" or, before that, re-arrange the File table to have the "guid" column itself and probably having another `file_application_instance` n to m table.

I think those are valid points but given that organization are arriving soon. I rather get the bugfix out and deal with rearrange the DB in the context of the new organizations work.

tests/unit/lms/services/canvas_test.py
```diff
@@ -213,6 +213,16 @@ def test_find_matching_file_in_course_returns_the_matching_file_id(
         canvas_api_client.list_files.assert_called_once_with(sentinel.course_id)
         assert matching_file_id == str(sentinel.matching_file_id)
 
+    def test_fallback_to_guid_matching(self, finder, file_service):
```
marcospri, line 216: This test don't use the DB so setting up a `.return_value` with the GUID scenario is not testing much more.

Just asserting here that we are making the second type of query.

lms/services/canvas.py
```diff
@@ -93,7 +93,11 @@ def find_matching_file_in_course(self, course_id, file_ids):
         Return None if no matching file is found.
         """
         for file_id in file_ids:
-            file = self._file_service.get(file_id, type_="canvas_file")
+            # Use both `get` (the most common scenario, files in th same application instance)
+            # and `get_with_guid` as a fallback for installs in the same LMS with multiple application instances.
+            file = self._file_service.get(
+                file_id, type_="canvas_file"
+            ) or self._file_service.get_with_guid(file_id, type_="canvas_file")
```
marcospri, line 100: Using only the `with_guid` version should do the right thing but being extra careful here and keeping the original version for most use cases.

tests/unit/lms/services/canvas_test.py
```diff
@@ -225,6 +235,9 @@ def test_find_matching_file_in_course_with_multiple_file_ids(
             # The third file_id *will* be found in the course.
             matching_file,
         ]
+        # The with GUID fallback doesn't returning anything else
+        file_service.get_with_guid.return_value = None
```
marcospri, line 239: I could probably re-write this test to parameterize the return values of both `file_service` methods and remove the new test.

I rather do that in a follow up PR.

lms/services/canvas.py
```diff
@@ -93,7 +93,11 @@ def find_matching_file_in_course(self, course_id, file_ids):
         Return None if no matching file is found.
         """
         for file_id in file_ids:
-            file = self._file_service.get(file_id, type_="canvas_file")
+            # Use both `get` (the most common scenario, files in th same application instance)
+            # and `get_with_guid` as a fallback for installs in the same LMS with multiple application instances.
```
jon-betts, line 97:
```suggestion
            # Use both `get` (the most common scenario, files in th same 
            # application instance) and `get_with_guid` as a fallback for 
            # installs in the same LMS with multiple application instances.
```

lms/services/canvas.py
jon-betts: I feel we should probably do the work now to work out if there's a difference. If not now then when? Otherwise I suspect we'll be carrying this complexity for a while.

These two methods are only called once here, so the "or" here could be a private matter inside the service.

lms/services/file.py
```diff
@@ -21,6 +21,31 @@ def get(self, lms_id, type_):
...
+
+        Once we have "Organization" or a similar entity that groups together multiple
+        ApplicationInstance we should use that concept instead of GUID but for now
+        this allows us to find files across multiple installs in for example
+        migrations from course/account/global level install or in some LTI1.3 upgrade
+        scenarios.
```
jon-betts, line 32:
```suggestion
        Get matching files linked to this application instance by GUID.

        Once we have "Organization" or a similar entity that groups together
        multiple ApplicationInstance we should use that concept instead of GUID
        but for now this allows us to find files across multiple installs in,
        for example, migrations from course/account/global level install or in
        some LTI1.3 upgrade scenarios.
```

Some of the organisation info here seems more of a comment for future work than a description of what this method does?

lms/services/file.py
```diff
@@ -21,6 +21,31 @@ def get(self, lms_id, type_):
...
+            self._db.query(File)
+            .join(
+                ApplicationInstance,
+                ApplicationInstance.tool_consumer_instance_guid
+                == self._application_instance.tool_consumer_instance_guid,
+            )
+            .filter(
+                File.lms_id == lms_id,
+                File.type == type_,
+            )
```
jon-betts, line 44:
```suggestion
            .filter(File.lms_id == lms_id, File.type == type_)
```

jon-betts (commented):
I'm kind of uncomfortable having two methods doing the same thing which are only called from one place. I think there are a few different levels to tackle this require progressively more work / bravery.

### Let the service encapsulate the details

As these two methods have the same arguments, do the same thing, and are only called from one place, I feel like the details could easily be pushed into the methods in the service.

This would have the added benefit of not altering the caller and it's tests.

### Have a single query that does both

I've made some suggestions for how this could work, but if we are worried that the old query might be doing some valuable work, I think we can combine the two into one query which will replicate both.

This should cut down on the code but not require us to do the leg work of thinking about if the old method is required or not.

### Do the work to convince ourselves the new method is all we need

... or finally we could work out if we need the first one and delete if if so. If not we could fall back to one of the approaches above.

~It might be possible to craft a query which could determine if there are any matches which would be found by one method and not the other?~ I'm not sure if this is possible as any case where this might happen is likely to be because the GUID has changed over time, rather than because of the current state of the data. We might be able to get something from `group_info` here?

Issue #4369, opened 2022-08-23
Course copy issues after LTI1.3 upgrade


marcospri:
In canvas upgraded assignments (ie existing ones in 1.1 that launch as 1.3 when a 1.3 app is installed) have two different assignment IDs the 1.1 and the 1.3 one. The launch payload contains both on 1.3 launches.

We do prefer the 1.1 ID  when available but picking the wrong one would cause problems around course copy.

https://github.com/hypothesis/lms/blob/d72be6c4fe36fce33b3b0a00177bf79806ac6c87/lms/models/lti_params.py#L119

If we did store both version of the assignment by mistake (one with the LTI1.1 ID, another with the 1.3 one) the course copy file mappings might be stored in only one of the rows but not the other.

Examples on the DB of potential cases of this happening:

```
SELECT assignment.created, resource_link_id, document_url, application_instances.lms_url from assignment 
JOIN assignment_grouping ON assignment.id = assignment_id
JOIN grouping ON grouping_id = grouping.id
JOIN application_instances ON application_instance_id = application_instances.id
WHERE application_instances.tool_consumer_info_product_family_code = 'canvas'
-- Using the LTI1.3 resource_link_id style
AND assignment.resource_link_id like '%\-%'
-- For AI that have being upgraded
AND application_instances.consumer_key is not null
-- Only relevant for canvas files
AND document_url like 'canvas://%'
ORDER by assignment.created desc
```

PR #4373, opened 2022-08-24
Fix roles reported for LTI1.3 launches

marcospri:
For https://github.com/hypothesis/lms/issues/4369

LTI1.3 reports the roles for the context (same as LTI1.1, the roles for
the current launch) but also any other roles across the institution in
different namespaces.

For the  current layer of compatibility with LTI1.1 we'll take only the
relevant roles for the current course.

https://www.imsglobal.org/spec/lti/v1p3#role-vocabularies

# Testing 

## Role changes

- Clear event_user 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate event_user"```

- Switch  to this branch an launch  

https://hypothesis.instructure.com/courses/319/assignments/3308

as `Hypothesis 101 Teacher`

check the JWT coming from canvas, it contains:

```
"https://purl.imsglobal.org/spec/lti/claim/roles": [
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor",
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student",
    "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor",
    "http://purl.imsglobal.org/vocab/lis/v2/system/person#User"
]
```

note the different namespaces here, `institution`, `system`, `membership`.

- Check the roles reported on the DB

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from  event_user join lti_role on lti_role_id = lti_role.id"

 id | event_id | user_id | lti_role_id | id |                            value                             |    type    
----+----------+---------+-------------+----+--------------------------------------------------------------+------------
 95 |       44 |       2 |           4 |  4 | http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor | instructor
(1 row)
```

## Effects on course copy

This has been the cause of most of the issues we've seen around course copy I'm going to focus the testing on that here

### On `main`

To get the course copy setup working

- Launch https://hypothesis.instructure.com/courses/319/assignments/3347 once as `Hypothesis 101 Teacher` (the orignal course)

- Launch https://hypothesis.instructure.com/courses/376/assignments/3357 with the same teacher user. (the copied course)

- Now, as a `eng+coursecopystudent@hypothes.is`

launch again https://hypothesis.instructure.com/courses/376/assignments/3357

You'll get  the course copy error dialog.

`eng+coursecopystudent@hypothes.is` is now a teacher in another course but only student in this one. When the course copy logic kicks in we pick up the teacher role and try to fetch the files, but as a student in the current course we don't have access and fail.

### On branch `fix-lti13-roles`

- Relaunch https://hypothesis.instructure.com/courses/376/assignments/3357 as `eng+coursecopystudent@hypothes.is` it now should work. We correctly identify the student role only and go grab the file without trying to list all the files first.

### DIFF
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -138,8 +138,29 @@ def _to_lti_v11(v13_params):
         v11_params.update({f"custom_{key}": value for key, value in custom.items()})
 
     if "roles" in v11_params:
-        # We need to squish together the roles for v1.1
-        v11_params["roles"] = ",".join(v11_params["roles"])
+        context_roles = []
+        for role in v11_params["roles"]:
+            # From: https://www.imsglobal.org/spec/lti/v1p3#role-vocabularies
+            #
+            # Conforming implementations MAY recognize the simple
+            # names for context roles; thus, for example, vendors can use
+            # the following roles interchangeably:
+            #   http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor
+            #   Instructor
+            if "http://purl.imsglobal.org/vocab/lis/v2" not in role:
+                # If using the simple name, we'll take it
+                context_roles.append(role)
+
+            if role.startswith("http://purl.imsglobal.org/vocab/lis/v2/membership"):
+                # For roles that have the whole LIS 2.0 name, take only the one
+                # relevant for the current context. We'd need to expose the
+                # rest of the roles somewhere else if they become necessary /
+                # interesting, but for LTI 1.1 compatibility we only expose the
+                # roles of the current context (course) here.
+                context_roles.append(role)
+
+        # We need to squish together the roles for v1.1 compatibility
+        v11_params["roles"] = ",".join(context_roles)
 
     return v11_params
 
--- a/tests/unit/lms/models/lti_params_test.py
+++ b/tests/unit/lms/models/lti_params_test.py
@@ -74,6 +74,26 @@ class TestLTIParams:
         assert params["user_id"] == "user_id-v11"
         assert params["resource_link_id"] == "resource_link_id-v11"
 
+    def test_ignores_non_context_roles(self, pyramid_request):
+        pyramid_request.lti_jwt = {
+            f"{CLAIM_PREFIX}/roles": [
+                # These roles do not relate to the current context (course) and
+                # should be ignored
+                "http://purl.imsglobal.org/vocab/lis/v2/system/person#User",
+                "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor",
+                # We should accept either of these formats
+                "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner",
+                "Mentor",
+            ],
+        }
+
+        params = LTIParams.from_request(pyramid_request)
+
+        assert (
+            params["roles"]
+            == "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner,Mentor"
+        )
+
 
 class TestCanvasQuirks:
     @pytest.mark.parametrize(

### CODE REVIEW

File: lms/models/lti_params.py [hunk 1]
```diff
@@ -138,8 +138,28 @@ def _to_lti_v11(v13_params):
...
+            # the following roles interchangeably:
+            #   http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor
+            #   Instructor
+            if "http://purl.imsglobal.org/vocab/lis/v2" not in role:
+                # If using the simple name, we'll take it
+                context_roles.append(role)
+
+            if role.startswith("http://purl.imsglobal.org/vocab/lis/v2/membership"):
+                # For roles that have the whole LIS 2.0 name, take only the one relevant for the current context.
+                # We'd need to expose the rest of the roles somewhere else if
```
marcospri, line 156: How we do this, exposing the other type of riles, is to be discussed.

We could add extra fields on lti_user for example but I have the impression that when we need to know "is this user a teacher anywhere in the institution" we could reach directly into the LTI1.3 while keeping the rest of the roles notion on the app as the more straightforward "role in the current launch/course".

Another alternative could be to change this at the `is_learner` `is_teacher` level but from a bug fix point of view I reckon that more complex and more error prone (we only do `"instructor" in roles.lower()` now.

Not sure what the implications of this change are for the LTIRole model. I reckon the mapping of any of the institutions roles to eg just "instructor" are wrong, when looking at the DB and seeing instructor we have a clear understanding of what that means, we shouldn't have to check the original LTI string to filter further. 

We could do a DB migration cleanup after this if necessary.

lms/models/lti_params.py
```diff
@@ -138,8 +138,28 @@ def _to_lti_v11(v13_params):
         v11_params.update({f"custom_{key}": value for key, value in custom.items()})
 
     if "roles" in v11_params:
-        # We need to squish together the roles for v1.1
-        v11_params["roles"] = ",".join(v11_params["roles"])
+        context_roles = []
+        for role in v11_params["roles"]:
+            # From: https://www.imsglobal.org/spec/lti/v1p3#role-vocabularies
```
marcospri, line 143: I read this, probably more than once, and always understood `context` as a meaningless word until just now made the context -> course association that is of course used in the rest of the document and I'm very familiar with :facepalm:

lms/models/lti_params.py
```diff
@@ -138,8 +138,28 @@ def _to_lti_v11(v13_params):
         v11_params.update({f"custom_{key}": value for key, value in custom.items()})
 
     if "roles" in v11_params:
-        # We need to squish together the roles for v1.1
-        v11_params["roles"] = ",".join(v11_params["roles"])
+        context_roles = []
+        for role in v11_params["roles"]:
```
jon-betts, line 142: I think this might work nicely as a list comprehension along the lines of:

```python
role for role in v11_params["roles"]
if "http://purl.imsglobal.org/vocab/lis/v2" not in role
or role.startswith("http://purl.imsglobal.org/vocab/lis/v2/membership")
```

jon-betts (approved):
This seems to work as described. Some of the interactions here are quite complex and I'm trusting the user roles etc in the testing instructions are excercising the right things to some extent.

But this all seems logical and should work fine.

I think we have a similar problem elsewhere with our capturing of user roles in assignments.

lms/models/lti_params.py
marcospri: I had it originally as this but when adding the comments I found it more difficult to read.

PR #4376, opened 2022-08-26
Allow canvas API details editing in the admin pages

marcospri:
Allow editing AIs Canvas API details

#  Testing

- Go over to http://localhost:8001/admin/instance/id/100/
- Click the new `Update API details` 
- Enter any values
- They are saved on the AI 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "select developer_key, developer_secret, aes_cipher_iv from application_instances where id = 100"```

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -116,6 +116,19 @@ class ApplicationInstanceService:
 
         return query.all()
 
+    def set_canvas_api_details(
+        self, application_instance, developer_key, developer_secret
+    ):
+
+        # If either one of developer_key or developer_secret is missing, then we won't set either
+        if developer_secret and developer_key:
+            aes_iv = self._aes_service.build_iv()
+            encrypted_secret = self._aes_service.encrypt(aes_iv, developer_secret)
+
+            application_instance.developer_key = developer_key
+            application_instance.aes_cipher_iv = aes_iv
+            application_instance.developer_secret = encrypted_secret
+
     def create_application_instance(  # pylint:disable=too-many-arguments
         self,
         lms_url,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lms/services/application_instance.py [hunk 1]
```diff
@@ -131,27 +144,19 @@ def create_application_instance(  # pylint:disable=too-many-arguments
             "Hypothesis" + secrets.token_hex(16) if not deployment_id else None
         )
```
marcospri, line 146: Now doing this on `set_canvas_api_details`

lms/services/application_instance.py
```diff
@@ -116,6 +116,19 @@ def search(
 
         return query.all()
 
+    def set_canvas_api_details(
+        self, application_instance, developer_key, developer_secret
+    ):
+
+        # If either one of developer_key or developer_secret is missing, then we won't set either
+        if developer_secret and developer_key:
```
marcospri, line 124: No need for the else side of this. If both are not present we just leave the AI untouched.

For new ones this would be `None`.

lms/templates/admin/instance.canvas_api_details.html.jinja2
```diff
@@ -0,0 +1,18 @@
+{% import "macros.jinja" as macros %}
```
marcospri, line 1: Having an extra template for this form is IMO the right SRE way of doing this. 

It adds a bit more boilerplate but trying to have anything resembling a dialog would just make it more complex UI and code wise.

lms/templates/admin/instance.html.jinja2
```diff
@@ -79,7 +79,12 @@
     <fieldset class="box">
         <legend class="label has-text-centered">Canvas settings</legend>
         {{ macros.disabled_text_field("API domain", instance.custom_canvas_api_domain) }}
+
         {{ macros.disabled_text_field("Developer key", instance.developer_key) }}
+        <div class="has-text-right mb-6">
```
marcospri, line 84: Don't try to upgrade this in-place, we avoid having to worry about the secret value.

PR #4379, opened 2022-08-26
JWT better error handling

marcospri:
Current behavior of the JWT service is to fail silently and return an empty dictionary for invalid JWTs.

Raising `lms.ValidationError` instead which produces a meaningful error message.

## Testing 

- Simulate the three scenarios (not all at the same time):

```diff
--git a/lms/services/jwt.py b/lms/services/jwt.py
index 8beca8fb..a9f9e322 100644
--- a/lms/services/jwt.py
+++ b/lms/services/jwt.py
@@ -73,7 +73,7 @@ class JWTService:
         if not id_token:
             return {}
 
-        if not jwt.get_unverified_header(id_token).get("kid"):
+        if jwt.get_unverified_header(id_token).get("kid"):
             LOG.debug("Missing 'kid' value in JWT header")
             return {}
 
@@ -82,7 +82,7 @@ class JWTService:
 
         # Find the registration based on the token's claimed issuer & audience
         registration = self._registration_service.get(iss, aud)
-        if not registration:
+        if registration:
             LOG.debug("Unknown registration for lti_token. iss:%s aud:%s.", iss, aud)
             return {}
 
@@ -92,7 +92,7 @@ class JWTService:
             ).get_signing_key_from_jwt(id_token)
 
             return jwt.decode(
-                id_token, key=signing_key.key, audience=aud, algorithms=["RS256"]
+                "NOPE", key=signing_key.key, audience=aud, algorithms=["RS256"]
             )
         except PyJWTError as err:
             LOG.debug("Invalid JWT for: %s, %s. %s", iss, aud, str(err))
```

- On `main` this results in either a 

```You're not authorized to view this page```

in https://hypothesis.instructure.com/courses/319/assignments/3308

or a

```Sorry, but something went wrong. The issue has been reported and we'll try to fix it.```

on https://hypothesis.instructure.com/courses/319/assignments/3347 due to a ```KeyError: 'custom_canvas_course_id'``` on the logs. That was caused by the empty dictionary returned by the JWT service being interpreted as "This is an LTI1.1 launch".

- On this branch `jwt-error-handling` the same scenarios produce the following error messages in both assignment:

```

jwt:
    Missing 'kid' value in JWT header
```

```

jwt:
   Unknown registration for JWT. iss:https://canvas.instructure.com aud:93820000000000200.
```

```

jwt:
    Invalid JWT for: https://canvas.instructure.com, 93820000000000200. Not enough segments
```

### DIFF
--- a/lms/services/jwt.py
+++ b/lms/services/jwt.py
@@ -10,6 +10,7 @@ from jwt.exceptions import ExpiredSignatureError, InvalidTokenError, PyJWTError
 from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
 from lms.services.lti_registration import LTIRegistrationService
 from lms.services.rsa_key import RSAKeyService
+from lms.validation import ValidationError
 
 LOG = logging.getLogger(__name__)
 
@@ -75,7 +76,9 @@ class JWTService:
 
         if not jwt.get_unverified_header(id_token).get("kid"):
             LOG.debug("Missing 'kid' value in JWT header")
-            return {}
+            raise ValidationError(
+                messages={"jwt": ["Missing 'kid' value in JWT header"]}
+            )
 
         unverified_payload = jwt.decode(id_token, options={"verify_signature": False})
         iss, aud = unverified_payload.get("iss"), unverified_payload.get("aud")
@@ -84,7 +87,11 @@ class JWTService:
         registration = self._registration_service.get(iss, aud)
         if not registration:
             LOG.debug("Unknown registration for lti_token. iss:%s aud:%s.", iss, aud)
-            return {}
+            raise ValidationError(
+                messages={
+                    "jwt": [f"Unknown registration for JWT. iss:{iss} aud:{aud}."]
+                }
+            )
 
         try:
             signing_key = self._get_jwk_client(
@@ -96,7 +103,9 @@ class JWTService:
             )
         except PyJWTError as err:
             LOG.debug("Invalid JWT for: %s, %s. %s", iss, aud, str(err))
-            return {}
+            raise ValidationError(
+                messages={"jwt": [f"Invalid JWT for: {iss}, {aud}. {err}"]}
+            ) from err
 
     def encode_with_private_key(self, payload: dict):
         key = self._rsa_key_service.get_random_key()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lms/services/jwt.py [hunk 1]
```diff
@@ -75,7 +76,9 @@ def decode_lti_token(self, id_token: str) -> dict:
 
         if not jwt.get_unverified_header(id_token).get("kid"):
             LOG.debug("Missing 'kid' value in JWT header")
```
seanh, line 78: Double-check this but I think you might be able to remove these `LOG.debug()` calls now because raised exceptions get logged automatically

tests/unit/lms/security_test.py
```diff
@@ -1,9 +1,8 @@
-from unittest.mock import PropertyMock, call, create_autospec, patch, sentinel
```
seanh, line 1: Some unrelated changes here?

tests/unit/lms/services/jwt_test.py
```diff
@@ -111,20 +112,23 @@ def test_decode_lti_token_with_empty_token(self, svc):
     def test_decode_lti_token_with_no_kid(self, svc):
         jwt_str = self.encode_jwt({"key": "value"}, headers={"key": "value"})
 
-        assert not svc.decode_lti_token(jwt_str)
+        with pytest.raises(ValidationError):
+            svc.decode_lti_token(jwt_str)
```
seanh, line 116: You might also want to test that it raises the right error message? This can catch issues with either the code or the test set up (when a test isn't testing the situation that it thinks it is, but is actually triggering a different `ValidationError`). It frequently catches a mistake.

```python
with pytest.raises(ValidationError) as exc_info:
    svc.decode_lti_token(jwt_str)

assert exc_info.err.message == { ... }
```

seanh (approved):
:+1: Looks good, thanks to the previous work this is a very nice and simple change to make now.

It looks like there might be some changes from other PRs mixed in with this one?

One important question: are you sure these error messages aren't going to display any sensitive information from JWTs to the user?

lms/services/jwt.py
marcospri: The exception doesn't show up on the logs locally. I'll check in paper-trail when deployed

tests/unit/lms/security_test.py
marcospri: Not rebased, fixed now.

marcospri:
> One important question: are you sure these error messages aren't going to display any sensitive information from JWTs to the user?

We are only logging the `aud` and `issuer` which are not that interesting. 

In any case nothing on the JWT is sensitive, it can be base64 decoded and all the information is there. It's only secure because we check it against the schools private keys.

Issue #4304, opened 2022-08-05
LMS admin pages fixes and improvements


marcospri:
- [x] Add an option to downgrade LTI1.3 AI to 1.1 in the admin pages
- [ ] Handle (show an error message) when editing an AI conflicts with another one (ie duplicated deployment_id)
- [x] Strip trailing `/` in the issuer. The issuer/lms url split makes this very confusing. https://github.com/hypothesis/lms/pull/4380
- [x] Allow editing of Canvas API keys in the application instance. https://github.com/hypothesis/lms/pull/4376
- [x] Allow editing of registration URLs. https://github.com/hypothesis/lms/pull/4388

mkdir-washington-edu:
I verified today in our Canvas instance that including a Canvas Developer API ID/Key in the new deployment, and **also** including an LTI 1.1 Consumer key that does **not** have a Canvas Dev API ID/Key, makes it so the saved deployment does not have the Canvas API ID/Key.

Subaccount install: https://hypothesis.instructure.com/accounts/111/settings/configurations#tab-tools

https://lms.hypothes.is/admin/instance/id/5536/

mattdricker:
Rendering issue when viewing list of app instances from a given reg page. The ai box doesn't dynamically adjust to the browser width, and the browser doesn't know to provide a horizontal scroll bar to get to the **View** button at the right of the ai item.
![Screen Shot 2022-08-18 at 2 02 18 PM](https://user-images.githubusercontent.com/71603629/185464289-5cba526f-378b-4d88-9126-3dffd6c813e5.png)

marcospri:
> I verified today in our Canvas instance that including a Canvas Developer API ID/Key in the new deployment, and also including an LTI 1.1 Consumer key that does not have a Canvas Dev API ID/Key, makes it so the saved deployment does not have the Canvas API ID/Key.


This hasn't been "fixed" as now those fields are not present in the upgrade path. For updating them you can directly update them on he AI details page.

PR #4380, opened 2022-08-26
Don't allow `/` on issuer

marcospri:
For: https://github.com/hypothesis/lms/issues/4304

This is a common error while pasting the values for a new registration as we have this value (the issuer) but also the "LMS URL" which might look very similar depending of the LMS.

We could just remove the slash if present but I reckon this just makes it more explicit and might be a good prompt to double check things.

# Testing

- Go over http://localhost:8001/admin/registration try an new issuer ending with `/`

### DIFF
--- a/lms/views/admin/lti_registration.py
+++ b/lms/views/admin/lti_registration.py
@@ -1,6 +1,6 @@
 from urllib.parse import urlparse
 
-from marshmallow import validate
+from marshmallow import ValidationError, validate, validates
 from pyramid.httpexceptions import HTTPFound
 from pyramid.renderers import render_to_response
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
Yep this works.

Are we sure this is always this way? Are we setting ourselves to find out that one LMS has been setup to _only_ accept a trailing slash?

marcospri:
> Yep this works.
> 
> Are we sure this is always this way? Are we setting ourselves to find out that one LMS has been setup to _only_ accept a trailing slash?

We don't know for sure but that's been the case so far. It's an easy fix in case we find the odd one in the wild.

PR #4387, opened 2022-08-29
[WIP] Handle validation errors of LTI1.3 authentication parameters

marcospri:
This allows to better feedback about what's missing on an LTI1.3 lti launch.

In summary:

- In security.py we try a few methods of authentication for every request and ignore all validation errors of the data we expect for them.
- Not having a valid `lti_user` doesn't assign the right permissions for the view and we end up with a forbidden view response.
- Here we return an "empty" `lti_user` wrapping the validation exceptions for the only method of authentication we are expecting.
- On the forbidden view we display a more informative error.

Note that:

- It's not a full re-write of security.py
- Doesn't handle LTI1.1 (but it's should be easy to extend to support it)

---

This returns a "sad H" page with for example the text:

```

form:
tool_consumer_instance_guid
```

instead of 

```
You're not authorized to view this page
```

### DIFF
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,6 +1,7 @@
 from typing import NamedTuple
 
 from lms.models.h_user import HUser
+from lms.validation import ValidationError
 
 
 class LTIUser(NamedTuple):
@@ -23,6 +24,24 @@ class LTIUser(NamedTuple):
 
     email: str = ""
     """The user's email address."""
+
+    validation_error: ValidationError = None
+    """Exception found while trying to construct a LTIUser from the request"""
+
+    def __bool__(self):
+        """
+        Override the bool behaviour of LTIUser.
+
+        In case when we want to return an "empty" LTIUser behave like false so checks like:
+
+            `if request.lti_user:`
+
+        produce the expected result.
+        """
+        # Behave as "falsy" if neither the user_id or application_instance are present
+        return bool(
+            self.user_id and self.application_instance_id and not self.validation_error
+        )
 
     @property
     def h_user(self):
@@ -56,6 +75,18 @@ class LTIUser(NamedTuple):
                 lti_core_schema["lis_person_name_full"],
             ),
             email=lti_core_schema["lis_person_contact_email_primary"],
+        )
+
+    @staticmethod
+    def from_validation_error(validation_error):
+        return LTIUser(
+            user_id=None,
+            application_instance_id=None,
+            roles=None,
+            tool_consumer_instance_guid=None,
+            display_name=None,
+            email=None,
+            validation_error=validation_error,
         )
 
 
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,7 +1,6 @@
 from typing import NamedTuple
 
 from lms.models.h_user import HUser
-from lms.validation import ValidationError
 
 
 class LTIUser(NamedTuple):
@@ -24,24 +23,6 @@ class LTIUser(NamedTuple):
 
     email: str = ""
     """The user's email address."""
-
-    validation_error: ValidationError = None
-    """Exception found while trying to construct a LTIUser from the request"""
-
-    def __bool__(self):
-        """
-        Override the bool behaviour of LTIUser.
-
-        In case when we want to return an "empty" LTIUser behave like false so checks like:
-
-            `if request.lti_user:`
-
-        produce the expected result.
-        """
-        # Behave as "falsy" if neither the user_id or application_instance are present
-        return bool(
-            self.user_id and self.application_instance_id and not self.validation_error
-        )
 
     @property
     def h_user(self):
@@ -75,18 +56,6 @@ class LTIUser(NamedTuple):
                 lti_core_schema["lis_person_name_full"],
             ),
             email=lti_core_schema["lis_person_contact_email_primary"],
-        )
-
-    @staticmethod
-    def from_validation_error(validation_error):
-        return LTIUser(
-            user_id=None,
-            application_instance_id=None,
-            roles=None,
-            tool_consumer_instance_guid=None,
-            display_name=None,
-            email=None,
-            validation_error=validation_error,
         )
 
 
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,6 +1,7 @@
-from typing import NamedTuple
+from typing import NamedTuple, Optional
 
 from lms.models.h_user import HUser
+from lms.validation import ValidationError
 
 
 class LTIUser(NamedTuple):
@@ -23,6 +24,24 @@ class LTIUser(NamedTuple):
 
     email: str = ""
     """The user's email address."""
+
+    validation_error: Optional[ValidationError] = None
+    """Exception found while trying to construct a LTIUser from the request"""
+
+    def __bool__(self):
+        """
+        Override the bool behaviour of LTIUser.
+
+        In case when we want to return an "empty" LTIUser behave like false so checks like:
+
+            `if request.lti_user:`
+
+        produce the expected result.
+        """
+        # Behave as "falsy" if neither the user_id or application_instance are present
+        return bool(
+            self.user_id and self.application_instance_id and not self.validation_error
+        )
 
     @property
     def h_user(self):
@@ -56,6 +75,18 @@ class LTIUser(NamedTuple):
                 lti_core_schema["lis_person_name_full"],
             ),
             email=lti_core_schema["lis_person_contact_email_primary"],
+        )
+
+    @staticmethod
+    def from_validation_error(validation_error):
+        return LTIUser(
+            user_id=None,
+            application_instance_id=None,
+            roles=None,
+            tool_consumer_instance_guid=None,
+            display_name=None,
+            email=None,
+            validation_error=validation_error,
         )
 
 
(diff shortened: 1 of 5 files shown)


lms/security.py
```diff
@@ -61,7 +61,7 @@ def _policy(self, request):
 class LTISecurityPolicy:
     @classmethod
     def authenticated_userid(cls, request):
-        if request.lti_user is None:
+        if not request.lti_user:
```
marcospri, line 64: I reckon this could always been written like this.

lms/security.py
```diff
@@ -143,19 +143,29 @@ def _get_lti_user(request):
...
-    # Avoid checking for the LTI1.3 JWT if not there
-    if "id_token" in request.params:
-        # Don't replace all auth methods in case somehow we have a bogus token
-        # but try this one first.
-        schemas.insert(0, LTI13AuthSchema(request).lti_user)
 
     lti_user = None
-    for schema in schemas:
+    # Avoid checking for the LTI1.3 JWT if not there
+    if "id_token" in request.params:
```
marcospri, line 149: In the previous version if for some reason a valid request for any other method had, for some reason, also an "id_token" we'd first validate that and then continue with the rest.

Here we just look at the id_token if present. 

I don't think we ever expected the old behavior to happen.

lms/security.py
```diff
@@ -143,19 +143,29 @@ def _get_lti_user(request):
...
-            continue
+            lti_user = LTI13AuthSchema(request).lti_user()
+        except ValidationError as err:
+            from lms.models import LTIUser  # pylint:disable=import-outside-toplevel
+
+            # If we haven't been able to get an LTIUser still return
+            # an empty one including the ValidationError.
+            # We'll be able to use the exception later in the exception view.
+            #
+            # Directly raising here won't be picked by the exception views.
```
marcospri, line 159: > Directly raising here won't be picked by the exception views

I can't point to the right place in the docs for this, maybe here: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/router.html

I reckon that makes sense, here we are worry about security, not building a response.

It's our design on top, trying many methods at once, which make giving feedback tricky.

lms/security.py
marcospri, line 149: This could be extended to:

- Only check LTI authentication in the view that are going to use them (basic launch and deep linking?
- In those, decide between 1.1 vs 1.3 based on the presence of `id_token`.

lms/views/exceptions.py
```diff
@@ -27,6 +27,17 @@ def notfound(self):
...
     def forbidden(self):
+        if self.request.lti_user is not None and self.request.lti_user.validation_error:
+            # We haven't found a view with the right permissions and we have
+            # found an error while trying to get an LTIUser.
+            # Treat this forbidden_view as a direct cause of
+            # that error and display information on the resulting page
+            self.request.override_renderer = (
+                "lms:templates/validation_error.html.jinja2"
+            )
+            self.request.response.status_int = 403
```
marcospri, line 38: I reckon this should still be a 403 as before.

lms/security.py
```diff
@@ -143,19 +143,29 @@ def _get_lti_user(request):
...
+            return LTIUser.from_validation_error(err)
+
+    else:
+        for schema in schemas:
```
marcospri, line 163: We still try the "see what sticks" method for the rest of options

lms/views/exceptions.py
```diff
@@ -27,6 +27,17 @@ def notfound(self):
...
+            return {"error": self.request.lti_user.validation_error}
```
marcospri, line 39: This is a long winded way of raising an exception.

lms/security.py
seanh: Yep, it doesn't surprise me that an exception raised by authentication code wouldn't be picked up by the exception views: the exception views catch exceptions raised by _the views_, they don't catch exceptions raised by other parts of the request processing pipeline (like authentication). I don't think authentication code is supposed to raise exceptions up to Pyramid

lms/models/lti_user.py
```diff
@@ -24,6 +25,24 @@ class LTIUser(NamedTuple):
...
+        In case when we want to return an "empty" LTIUser behave like false so checks like:
+
+            `if request.lti_user:`
+
+        produce the expected result.
+        """
+        # Behave as "falsy" if neither the user_id or application_instance are present
+        return bool(
+            self.user_id and self.application_instance_id and not self.validation_error
+        )
```
seanh, line 44: Hmmmmm. The complex definition of falsey feels wrong to me. It seems like an `LTIUser` could have certain attributes (like an email) and still be falsey. Plus I think Pyramid already has a way of saying "no user" or "falsey user": `request.lti_user` is `None`

lms/security.py
seanh: Yeah, I think it makes sense to only use the right types of authentication in the right views, rather than always trying all of them

seanh (changes requested):
I think you're on the right path with trying to change the forbidden view so that it shows a more informative validation error rather than just the simple "you're not authorized" error.

But I don't think we should try to attach the validation error to the `request.lti_user` object.

If a view raises a `ValidationError` (or any type of exception) then the appropriate exception view is called and `self.exception` is the exception that the view raised. This is how `ValidationError`'s get passed from the views (or from the validation schemas attached to the views) through to the exception views that render those `ValidationError`'s in HTML error pages.

On the other hand if **authentication** fails (the security policy returns a `Denied` object indicating that the request was denied) then the **forbidden view** gets called and `self.exception` is an `HTTPForbidden` object representing the denial of access.

I think you should try to attach the validation error to that `self.exception` object somehow, rather than attaching it to `request.lti_user`.

Unfortunately I don't immediately know **how** you should do that and I've run out of time for today, but I can look further into it tomorrow. At first glance it looks to me like the security policy could:

* Attach a custom attribute to the `Denied` object before returning it. As in:

  ```python
  denied = Denied("...")
  denied.validation_errors = <whatever>

  return denied
  ```

* Or return a custom `Denied` subclass. As in:

  ```python
  class CustomDenied(Denied):
    def __init__(self, validation_errors):
      self.validation_errors = validation_errors
      super().__init__("...")
  ```

  It looks like Pyramid's built-in `ACL` stuff does this:

  https://github.com/Pylons/pyramid/blob/1e439035e41da3767acbcae79689fb05dfdeb317/src/pyramid/security.py#L443-L454

lms/security.py
```diff
@@ -101,6 +142,37 @@ def forget(self, request):
...
+        try:
+            lti_user = self._lti_user(request)
+        except ValidationError as err:
+            return Identity("", [], err)
+
+        return get_identity_from_lti_user(lti_user)
+
+    @staticmethod
+    @lru_cache(maxsize=1)
+    def _lti_user(request):
```
marcospri, line 167: Need to rethink this method, both 

authenticated_userid
identity

need to execute this at the moment when validation fails.

lms/security.py
```diff
@@ -14,11 +14,23 @@
...
+
+
+from lms.models import LTIUser
 
 
 class Identity(NamedTuple):
     userid: str
     permissions: List[str]
+    denied_reason: Optional[ValidationError] = None
+    lti_user: Optional[LTIUser] = None
```
marcospri, line 27: If we got a `lti_user` in a security policy, cache it here.

lms/security.py
```diff
@@ -119,10 +191,13 @@ def _permits(policy, request, _context, permission):
     if identity and permission in identity.permissions:
         return Allowed("allowed")
 
+    if identity and identity.denied_reason:
```
marcospri, line 194: As far as I can tell only permits can return one of these so we have to include it on identity and then expose it here.

lms/security.py
```diff
@@ -101,6 +142,37 @@ def forget(self, request):
         pass
 
 
+class LTILaunchesSecurityPolicy(LTIUserSecurityPolicy):
+    """
+    Security policy for LTI launches
```
marcospri, line 147: This security policy doesn't use `get_lti_user` and instead just looks at the schemas relevant for LTI authentication (as in LTI spec, not as in our LTIUser) and handles the exceptions.

lms/security.py
```diff
@@ -132,8 +216,17 @@ def _get_lti_user(request):
     If the request doesn't contain either valid LTI launch params or a valid
     bearer token then return ``None``.
 
+    :param: from_identity In case we are using this function directly on a security policy
```
marcospri, line 219: This allows all the code that does `request.lti_user` to use the LTIUser created by `LTILaunchesSecurityPolicy`

lms/security.py
seanh: I wonder if maybe the security policy needs to save the `denied_reason` on its own `self`? That would avoid the need to put a `denied_reason` on the `Identity` class and would allow `identity` to be `None` when no user can be identified.

But we may just need to refactor our code a bit first to enable it.

If we split it up so that there's one security policy for each way of authenticating the user (LTI 1.1 launch params, LTI 1.3 params, bearer token, OAuth state param, Google, ...) and so that every request is always mapped to exactly one security policy according to the request path.

```python
class LTI1SecurityPolicy:
    def identity(self):
        try:
            return self._get_identity()
            self.reason = err
            return None

    def permits(self):
        if identity is None:
            return DeniedWithReason(self.reason)
        ...

    def _get_identity(self):
        # Return an Identity from LTI 1.1 launch params or raise ValidationError.
        ...

    ...
```

marcospri:
This as been split into different PRs

PR #4393, opened 2022-08-31
Refactor SecurityPolicy to map routes to policies

marcospri:
Instead of trying policies until one of them works explicitly declare
the policy to be used based on the current URL.

This only tackles the "map routes to policies" issue and only applies it to the admin pages which use Google's Oauth. 

## Testing

- Head over http://localhost:8001/admin login, logout.
- Launch an assignment as a sanity check for the rest of the auth methods: https://hypothesis.instructure.com/courses/125/assignments/875

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -28,54 +28,34 @@ class Permissions(Enum):
     ADMIN = "admin"
 
 
-class PathPolicyGetter:
-    """Callable that returns different policies based on the request path."""
+@lru_cache(maxsize=1)
+def get_policy(request):
+    """Pick the right policy based the request's path."""
+    path = request.path
 
-    def __init__(self, path_policy_mapping):
-        """
-        Create a new PathPolicyGetter.
+    if path.startswith("/admin") or path.startswith("/googleauth"):
+        return LMSGoogleSecurityPolicy()
 
-        :param path_policy_mapping: Map between route paths and policies
-        """
-        self.path_policy_mapping = path_policy_mapping
-
-    @lru_cache(maxsize=1)
-    def __call__(self, request):
-        """Get the corresponding policy based on the request path."""
-        for path, policy in self.path_policy_mapping.items():
-            # We rely on dict's preserved insertion order
-            # and expect that more specific routes are found first.
-            if request.path.startswith(path):
-                return policy()
-
-        raise ValueError(f"No policy found for {request.path}")
+    return LTISecurityPolicy()
 
 
 class SecurityPolicy:
     """Top-level authentication policy that delegates to sub-policies."""
 
-    def __init__(self, policy_getter):
-        """
-        Create a new SecurityPolicy.
-
-        :param policy_getter: Callable that will pick the right policy based on the request object
-        """
-        self._policy_getter = policy_getter
-
     def authenticated_userid(self, request):
-        return self._policy_getter(request).authenticated_userid(request)
+        return get_policy(request).authenticated_userid(request)
 
     def identity(self, request):
-        return self._policy_getter(request).identity(request)
+        return get_policy(request).identity(request)
 
     def permits(self, request, context, permission):
-        return self._policy_getter(request).permits(request, context, permission)
+        return get_policy(request).permits(request, context, permission)
 
     def remember(self, request, userid, **kw):
-        return self._policy_getter(request).remember(request, userid, **kw)
+        return get_policy(request).remember(request, userid, **kw)
 
     def forget(self, request):
-        return self._policy_getter(request).forget(request)
+        return get_policy(request).forget(request)
 
 
 class LTISecurityPolicy:
@@ -193,20 +173,6 @@ def _get_user(request):
 
 
 def includeme(config):
-    config.set_security_policy(
-        SecurityPolicy(
-            PathPolicyGetter(
-                path_policy_mapping={
-                    # Paths are tested in sequence so the order here
-                    # is relevant.
-                    # More specific routes should be higher than shorter ones.
-                    "/admin": LMSGoogleSecurityPolicy,
-                    "/googleauth": LMSGoogleSecurityPolicy,
-                    # Fallback for the rest
-                    "/": LTISecurityPolicy,
-                }
-            )
-        )
-    )
+    config.set_security_policy(SecurityPolicy())
     config.add_request_method(_get_lti_user, name="lti_user", property=True, reify=True)
     config.add_request_method(_get_user, name="user", property=True, reify=True)
--- a/lms/security.py
+++ b/lms/security.py
@@ -28,34 +28,54 @@ class Permissions(Enum):
     ADMIN = "admin"
 
 
+class PathPolicyGetter:
+    """Callable that returns different policies based on the request path."""
+
+    def __init__(self, path_policy_mapping):
+        """
+        Create a new PathPolicyGetter.
+
+        :param path_policy_mapping: Map between route paths and policies
+        """
+        self.path_policy_mapping = path_policy_mapping
+
+    @lru_cache(maxsize=1)
+    def __call__(self, request):
+        """Get the corresponding policy based on the request path."""
+        for path, policy in self.path_policy_mapping.items():
+            # We rely on dict's preserved insertion order
+            # and expect that more specific routes are found first.
+            if request.path.startswith(path):
+                return policy()
+
+        raise ValueError(f"No policy found for {request.path}")
+
+
 class SecurityPolicy:
     """Top-level authentication policy that delegates to sub-policies."""
 
-    def __init__(self):
-        self._subpolicies = [LTISecurityPolicy(), LMSGoogleSecurityPolicy()]
+    def __init__(self, policy_getter):
+        """
+        Create a new SecurityPolicy.
+
+        :param policy_getter: Callable that will pick the right policy based on the request object
+        """
+        self._policy_getter = policy_getter
 
     def authenticated_userid(self, request):
-        return self._policy(request).authenticated_userid(request)
+        return self._policy_getter(request).authenticated_userid(request)
 
     def identity(self, request):
-        return self._policy(request).identity(request)
+        return self._policy_getter(request).identity(request)
 
     def permits(self, request, context, permission):
-        return self._policy(request).permits(request, context, permission)
+        return self._policy_getter(request).permits(request, context, permission)
 
     def remember(self, request, userid, **kw):
-        return self._policy(request).remember(request, userid, **kw)
+        return self._policy_getter(request).remember(request, userid, **kw)
 
     def forget(self, request):
-        return self._policy(request).forget(request)
-
-    @lru_cache(maxsize=1)
-    def _policy(self, request):
-        for policy in self._subpolicies:
-            if policy.authenticated_userid(request):
-                return policy
-
-        return self._subpolicies[-1]
+        return self._policy_getter(request).forget(request)
 
 
 class LTISecurityPolicy:
@@ -173,6 +193,20 @@ def _get_user(request):
 
 
 def includeme(config):
-    config.set_security_policy(SecurityPolicy())
+    config.set_security_policy(
+        SecurityPolicy(
+            PathPolicyGetter(
+                path_policy_mapping={
+                    # Paths are tested in sequence so the order here
+                    # is relevant.
+                    # More specific routes should be higher than shorter ones.
+                    "/admin": LMSGoogleSecurityPolicy,
+                    "/googleauth": LMSGoogleSecurityPolicy,
+                    # Fallback for the rest
+                    "/": LTISecurityPolicy,
+                }
+            )
+        )
+    )
     config.add_request_method(_get_lti_user, name="lti_user", property=True, reify=True)
     config.add_request_method(_get_user, name="user", property=True, reify=True)
--- a/tests/functional/views/admin_test.py
+++ b/tests/functional/views/admin_test.py
@@ -1,0 +1,5 @@
+def test_admin_authentication_redirects_to_google(app):
+    response = app.get("/admin/instances/")
+
+    assert response.status_code == 302
+    assert response.location.startswith("http://localhost/googleauth/login")
--- a/tests/unit/lms/security_test.py
+++ b/tests/unit/lms/security_test.py
@@ -1,4 +1,3 @@
-from unittest import mock
 from unittest.mock import call, sentinel
 
 import pytest
@@ -9,11 +8,11 @@ from lms.security import (
     Identity,
     LMSGoogleSecurityPolicy,
     LTISecurityPolicy,
-    PathPolicyGetter,
     Permissions,
     SecurityPolicy,
     _get_lti_user,
     _get_user,
+    get_policy,
     includeme,
 )
 from lms.validation import ValidationError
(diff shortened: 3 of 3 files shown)


lms/security.py
```diff
@@ -31,31 +31,35 @@ class Permissions(Enum):
 class SecurityPolicy:
     """Top-level authentication policy that delegates to sub-policies."""
 
-    def __init__(self):
-        self._subpolicies = [LTISecurityPolicy(), LMSGoogleSecurityPolicy()]
+    def __init__(self, path_policy_mapping, fallback_policy):
+        # Map between route paths and policies
+        self.path_policy_mapping = path_policy_mapping
```
marcospri, line 41: Injecting the configuration instead of hard-coding it. I reckon this makes testing easiser.

lms/security.py
```diff
@@ -31,31 +31,35 @@ class Permissions(Enum):
...
     def forget(self, request):
-        return self._policy(request).forget(request)
+        return self.get_route_policy(request).forget(request)
 
     @lru_cache(maxsize=1)
-    def _policy(self, request):
-        for policy in self._subpolicies:
-            if policy.authenticated_userid(request):
-                return policy
+    def get_route_policy(self, request):
```
marcospri, line 60: This class is not expected to be sub-classed so we are not defining an interface here. Using a public, longer name.

lms/security.py
```diff
@@ -31,31 +31,35 @@ class Permissions(Enum):
...
+        """Get the corresponding policy for this request's route."""
+        for path, policy in self.path_policy_mapping.items():
+            if request.path.startswith(path):
+                return policy()
 
-        return self._subpolicies[-1]
+        return self.fallback_policy()
```
marcospri, line 66: We are taking classes instead of instances now. `()` instantiates it.

lms/security.py
```diff
@@ -173,6 +177,14 @@ def _get_user(request):
 
 
 def includeme(config):
-    config.set_security_policy(SecurityPolicy())
+    config.set_security_policy(
+        SecurityPolicy(
+            path_policy_mapping={
```
marcospri, line 186: The actual mapping of routes / policies is configuration down here.

tests/functional/views/admin_test.py
```diff
@@ -0,0 +1,5 @@
+def test_admin_authentication_redirects_to_google(app):
```
marcospri, line 1: Veryyyy light touch functional testing for this. Pretty much a sanity check.

I don't think this warrants the effort of simulating the whole oauth flow with google but this is better than nothing.

tests/unit/lms/security_test.py
```diff
@@ -23,7 +23,13 @@ class TestIncludeMe:
     def test_it_sets_security_policy(self, pyramid_config, SecurityPolicy):
         includeme(pyramid_config)
 
-        SecurityPolicy.assert_called_once_with()
+        SecurityPolicy.assert_called_once_with(
+            path_policy_mapping={
```
marcospri, line 27: Asserting the exact configuration.

tests/unit/lms/security_test.py
```diff
@@ -161,102 +167,82 @@ def test_authenticated_userid(self, lti_user, expected_userid, pyramid_request):
 
 
 class TestSecurityPolicy:
-    def test_it_returns_the_userid_from_LTISecurityPolicy(
-        self, policy, LTISecurityPolicy, lti_security_policy
-    ):
-        lti_security_policy.authenticated_userid.return_value = "user"
-        user_id = policy.authenticated_userid(mock.sentinel.request)
+    def test_authenticated_userid(self, policy, get_route_policy):
```
marcospri, line 170: For the methods that just call the right method on the chosen policy I'm just asserting that. Not any of the specifics of the configuration.

tests/unit/lms/security_test.py
```diff
@@ -161,102 +167,82 @@ def test_authenticated_userid(self, lti_user, expected_userid, pyramid_request):
...
-        lti_security_policy.identity.assert_called_once_with(mock.sentinel.request)
+    @pytest.mark.parametrize(
+        "route,expected",
+        [
+            ("/match/specific", sentinel.match_specific),
+            ("/match", sentinel.match),
+            ("/no_match", sentinel.fallback),
+        ],
+    )
+    def test_get_route_policy(self, policy, pyramid_request, route, expected):
```
marcospri, line 225: Testing `get_route_policy` individually. The mechanism is quite simple.

lms/security.py
```diff
@@ -31,31 +31,35 @@ class Permissions(Enum):
...
 
     @lru_cache(maxsize=1)
-    def _policy(self, request):
-        for policy in self._subpolicies:
-            if policy.authenticated_userid(request):
-                return policy
+    def get_route_policy(self, request):
+        """Get the corresponding policy for this request's route."""
+        for path, policy in self.path_policy_mapping.items():
+            if request.path.startswith(path):
```
marcospri, line 63: `.startswith` is flexible enough for now. Maybe regexp are needed in the future? It should  be simple change.

lms/security.py
```diff
@@ -31,31 +31,35 @@ class Permissions(Enum):
 class SecurityPolicy:
     """Top-level authentication policy that delegates to sub-policies."""
 
-    def __init__(self):
-        self._subpolicies = [LTISecurityPolicy(), LMSGoogleSecurityPolicy()]
```
marcospri, line 35: I think starting with `LMSGoogleSecurityPolicy` is the right move because:

- It's only used on the admin pages. It's easy to test.
- It was the only other policy that was using the older delegation mechanism in this policy (`LTISecurityPolicy` having built-int delegation itself)

lms/security.py
```diff
@@ -31,31 +31,39 @@ class Permissions(Enum):
...
-            if policy.authenticated_userid(request):
-                return policy
+    def get_route_policy(self, request):
+        """Get the corresponding policy for this request's route."""
+        for path, policy in self.path_policy_mapping.items():
+            if request.path.startswith(path):
+                return policy()
 
-        return self._subpolicies[-1]
+        return self.fallback_policy()
```
marcospri, line 66: Fallback might not be required in the future, we might raise here.

lms/security.py
seanh: I think "public" just refers to the parts of the class that are meant to be called by other code outside of this class (i.e. Pyramid) from the parts that're only meant to be called internally by this class's other methods. By that definition `get_route_policy()` probably should be private.

Having said that I don't think it's strictly necessary to distinguish public from private in a class like this. It might have some readability benefit as it indicates that `_get_route_policy()` is just something internal rather than something that Pyramid is going to call directly. But I think that'd be the only reason for it

lms/security.py
seanh: Yeah I was thinking that. If you added a `"/": LTISecurityPolicy` to the end of `path_policy_mapping` then I think you could get rid of `fallback_policy` now.

lms/security.py
```diff
@@ -173,6 +181,14 @@ def _get_user(request):
 
 
 def includeme(config):
-    config.set_security_policy(SecurityPolicy())
+    config.set_security_policy(
+        SecurityPolicy(
+            path_policy_mapping={
+                "/admin": LMSGoogleSecurityPolicy,
+                "/googleauth": LMSGoogleSecurityPolicy,
```
seanh, line 188: I wondered with the non-ordered nature of dicts could cause the order of the mappings here to get randomised and potentially cause the results of security policy lookups to be random. But since Python 3.7 [dicts are ordered now](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) (it says so in the official Python docs as well)

tests/unit/lms/security_test.py
seanh: I think it's a bit of a code smell for the tests for a class to mock out part of that class. We want cleanly defined public interfaces between the different units of code that we're testing. Treating `SecurityPolicy.get_route_policy()` as one unit and the rest of the `SecurityPolicy` class as a separate unit doesn't seem very clearly defined to me since they're two parts of the same class. What other parts of the `SecurityPolicy` class are used by `get_route_policy()`? And which by the other methods? And which by both? In fact `SecurityPolicy.path_policy_mapping` and `SecurityPolicy.fallback_policy` are only used by `SecurityPolicy.get_route_policy()`, and the rest of `SecurityPolicy`'s methods only use `get_route_policy()`, but you have to read the code of the whole `SecurityPolicy` class to spot that division.

I wonder if it might be clearer if that division were made explicit by splitting it into two separate classes?

Here's a rough sketch:

```python
class PathPolicyGetter:
    """Callable that returns different policies based on the request path."""

    def __init__(self, path_policy_mapping, fallback_policy):
        ...

    @lru_cache
    def __call__(self, request):
        # Return the matching policy from path_policy_mapping or fallback_policy.

class SecurityPolicy:
    def __init__(self, policy_getter):
        self._policy_getter = policy_getter

    def authenticated_userid(self, request):
        return self._policy_getter(request).authenticated_userid(request)

    ...

def includeme(config):
    policy_getter = PathPolicyGetter(
        path_policy_mapping={...},
        fallback_policy=LTISecurityPolicy
    )
    config.set_security_policy(SecurityPolicy(PathPolicyGetter))
```

Then you can unit test `PathPolicyGetter` which has no dependencies (other than the injected mapping dict). And you can separately test `SecurityPolicy` by passing it in a mock `policy_getter`. You won't need to patch anything

seanh (commented):
Posting a partial review of this so you'll have some suggestions to respond to in the morning. I haven't finished reading the code yet but I need to finish for today

lms/security.py
marcospri: Yeah ordered since 3.6 as an implementation detail, guaranteed as part of python in 3.7.

Added a couple of comments about it.

tests/unit/lms/security_test.py
marcospri: Went with pretty much that solution verbatim.

I started with something very similar but with an extra layer of subclassing  (eg `PolicyGetter -> `PathPolicyGetter`) but I was adding to much "generalization" and scratched it. Might have gone to far in the other direction. 

I don't think we ever use more than one `PolicyGetter` at the same time. This makes the testing more convenient as you point out.

lms/security.py
marcospri: :+1: A bit less explicit but adding a comment for that effect on the configuration side.

lms/security.py
seanh: I think you probably want to hard-code the mapping here because the mapping itself is a likely source of bugs and it isn't covered by the tests currently (I don't think the tests for `includeme()` are likely to catch bugs).

For example if I created a mapping like this:

```python
path_policy_mapping={
    "/admin": LMSGoogleSecurityPolicy,
    "/admin/foo": FooPolicy,
    "/googleauth": LMSGoogleSecurityPolicy,
    "/": LTISecurityPolicy,
    "/bar": BarPolicy,
}
```

`FooPolicy` and `BarPolicy` will never be called because of the ordering, but the unit tests won't catch this since they don't test with the actual mapping.

If the mapping were hard-coded then I'd find the mistake when I tried to add a `PathPolicyGetter` unit test like this:

```python
class TestPathPolicyGetter:
    ...

    # This test would fail.
    def test_it_returns_the_foo_policy_for_foo_pages(self, request):
        request.path = "/admin/foo/baz"

        assert isinstance(PathPolicyGetter()(request), FooPolicy)
```

`TestPathPolicyGetter` could mock the sub-policies with `patch` fixtures. That would probably be ideal as it means the real policies aren't used at all in `PathPolicyGetter`'s unit tests so bugs in the policies (for example in their `__init__()` methods) can't cause `PathPolicyGetter`'s tests to fail. With patching the assertions would look something like:

```python
assert PathPolicyGetter()(request) == FooPolicy.return_value
```

Normally I would say that unit tests should always mock out any other units _from our own code_ that the unit-under-test uses. But in this case `PathPolicyGetter` and the security policies are close-collaborators (they're in the same file) so I don't think the tests _have to_ mock. But I think the tests might actually be better if they _do_ mock.

lms/security.py
```diff
@@ -28,34 +28,54 @@ class Permissions(Enum):
...
+    @lru_cache(maxsize=1)
+    def __call__(self, request):
+        """Get the corresponding policy based on the request path."""
+        for path, policy in self.path_policy_mapping.items():
+            # We rely on dict's preserved insertion order
+            # and expect that more specific routes are found first.
+            if request.path.startswith(path):
+                return policy()
+
+        raise ValueError(f"No policy found for {request.path}")
```
seanh, line 51: This is great :+1: very easy to understand.

If `path_policy_mapping` were hard-coded rather than injectable then it could be hard-coded right here in this method rather than an `__init__()` and then everything would be easier to see all in one place. In fact the `PathPolicyGetter` class could possibly be replaced with just a function?

lms/security.py
seanh: Yeah :+1: Incremental is good. Do you think we should break up `LTISecurityPolicy` into separate policies in future PRs?

lms/security.py
```diff
@@ -173,6 +193,20 @@ def _get_user(request):
...
+        SecurityPolicy(
+            PathPolicyGetter(
+                path_policy_mapping={
+                    # Paths are tested in sequence so the order here
+                    # is relevant.
+                    # More specific routes should be higher than shorter ones.
+                    "/admin": LMSGoogleSecurityPolicy,
+                    "/googleauth": LMSGoogleSecurityPolicy,
+                    # Fallback for the rest
+                    "/": LTISecurityPolicy,
```
seanh, line 206: I think you probably want the values here to be instances not classes. A security policy class may want to use `self` to store state and may get caught out when it doesn't realise that there are multiple instances of itself. It would also mean that the `__init__()`'s of security policies could have different arguments and that could be handled here in the `includeme()`, rather than them all having to have the same interface.

tests/functional/views/admin_test.py
seanh: Yeah, makes sense. You've got to be careful not to create more problems than you're solving when writing tests. Test code maintenance is a real issue. And functional tests are particularly prone to being really hard to understand and maintain because so much code is being executed under them. Best to keep it simple :+1:

tests/functional/views/admin_test.py
```diff
@@ -0,0 +1,5 @@
...
+    response = app.get("/admin/instances/")
+
+    assert response.status_code == 302
+    assert response.location.startswith("http://localhost/googleauth/login")
```
seanh, line 5: I think you might be able to use `h-matchers` to do this in a single `assert`: https://github.com/hypothesis/h-matchers/blob/main/docs/matching-web.md

tests/unit/lms/security_test.py
seanh: You could avoid having to do this if `path_policy_mapping` were hard-coded

tests/unit/lms/security_test.py
```diff
@@ -20,19 +21,33 @@
...
+            }
+        )
+        SecurityPolicy.assert_called_once_with(PathPolicyGetter.return_value)
         assert (
             pyramid_config.registry.queryUtility(ISecurityPolicy)
             == SecurityPolicy.return_value
         )
 
-    @pytest.fixture(autouse=True)
+    @pytest.fixture()
```
seanh, line 43: What's the reason for removing `autouse=True` here? I'd guess that we never want any of `includeme()`'s unit tests to accidentally be using the real `SecurityPolicy` to `autouse` probably makes sense

seanh (changes requested):
This is looking good. I think the idea of path-based mapping is a clear and simple-to-understand system for selecting a security policy. It's nice and flat: we have the one top-level `SecurityPolicy` that selects a single sub-policy based on the path and that's all there is to it (or it will be once `LTISecurityPolicy` is split up into simple policies).

Because this is `security.py` I think we want to make sure that the unit tests do a good job of covering the actual behaviour of the code. I think they'll do that better if `path_policy_mapping` is hard-coded rather than the tests injecting a different mapping than is actually used in production. Currently the mapping dict itself isn't really tested. I think hard-coding it may simplify the code as well: `PathPolicyGetter` may even be able to be changed into a function.

seanh:
One more observation: what if `path_policy_mapping` wasn't even a dict?

```python
@cache
def __call__(self, request):
    path = request.path

    if path.startswith("/admin") or path.startswith("/googleauth"):

    return LTISecurityPolicy()
```

I think this may have one significant advantage: if there's a missing unit test then coverage will point that out. For example if I didn't actually have a test for the `"/admin"` or `"/googleauth"` case then there'd be missing test coverage for the `return LMSGoogleSecurityPolicy()` line.

seanh (approved):
:+1: This is looking good, I think it's a nice and clear setup

PR #4394, opened 2022-08-31
Security LTISecurityPolicy refactor

marcospri:
Small refactor in security.py to accommodate further changes and new policies. 

Diff might be easier to read if done by commit.

## Testing 

There should not be any functional changes here, as a sanity check launch an assignment:

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -6,6 +6,7 @@ from typing import List, NamedTuple
 from pyramid.security import Allowed, Denied
 from pyramid_googleauth import GoogleSecurityPolicy
 
+from lms.models import LTIUser
 from lms.services import UserService
 from lms.validation import ValidationError
 from lms.validation.authentication import (
@@ -58,38 +59,17 @@ class SecurityPolicy:
         return get_policy(request).forget(request)
 
 
-class LTISecurityPolicy:
+class _LTIUserBasedSecurityPolicy:
+    """Base class for `LTIUser` based security policies."""
+
     @classmethod
-    def authenticated_userid(cls, request):
-        if request.lti_user is None:
-            return None
+    def get_lti_user(cls, request) -> LTIUser:  # pragma: no cover
+        """
+        Get and LTIUser to use as the authenticated user in this policy.
 
-        # urlsafe_b64encode() requires bytes, so encode the userid to bytes.
-        user_id_bytes = request.lti_user.user_id.encode()
-
-        safe_user_id_bytes = base64.urlsafe_b64encode(user_id_bytes)
-
-        # urlsafe_b64encode() returns ASCII bytes but we need unicode, so
-        # decode it.
-        safe_user_id = safe_user_id_bytes.decode("ascii")
-
-        return ":".join([safe_user_id, str(request.lti_user.application_instance_id)])
-
-    def identity(self, request):
-        userid = self.authenticated_userid(request)
-
-        if userid:
-            permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
-
-            if any(
-                role in request.lti_user.roles.lower()
-                for role in ["administrator", "instructor", "teachingassistant"]
-            ):
-                permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
-
-            return Identity(userid, permissions)
-
-        return Identity("", [])
+        How this LTIUser is constructed is up to the subclasses.
+        """
+        raise NotImplementedError()
 
     def permits(self, request, context, permission):
         return _permits(self, request, context, permission)
@@ -99,6 +79,45 @@ class LTISecurityPolicy:
 
     def forget(self, request):
         pass
+
+    @classmethod
+    def authenticated_userid(cls, request):
+        lti_user = cls.get_lti_user(request)
+        if lti_user is None:
+            return None
+
+        # urlsafe_b64encode() requires bytes, so encode the userid to bytes.
+        user_id_bytes = lti_user.user_id.encode()
+
+        safe_user_id_bytes = base64.urlsafe_b64encode(user_id_bytes)
+
+        # urlsafe_b64encode() returns ASCII bytes but we need unicode, so
+        # decode it.
+        safe_user_id = safe_user_id_bytes.decode("ascii")
+
+        return ":".join([safe_user_id, str(lti_user.application_instance_id)])
+
+    def identity(self, request):
+        userid = self.authenticated_userid(request)
+
+        if userid:
+            permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
+
+            if any(
+                role in self.get_lti_user(request).roles.lower()
+                for role in ["administrator", "instructor", "teachingassistant"]
+            ):
+                permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
+
+            return Identity(userid, permissions)
+
+        return Identity("", [])
+
+
+class LTISecurityPolicy(_LTIUserBasedSecurityPolicy):
+    @classmethod
+    def get_lti_user(cls, request):
+        return request.lti_user
 
 
 class LMSGoogleSecurityPolicy(GoogleSecurityPolicy):
--- a/lms/security.py
+++ b/lms/security.py
@@ -1,8 +1,9 @@
 import base64
 from enum import Enum
 from functools import lru_cache, partial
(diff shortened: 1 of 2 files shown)


lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
         return self.fallback_policy()
 
 
-class LTISecurityPolicy:
+class _LTIUserBasedSecurityPolicy:
```
marcospri, line 70: Base class for all policies that rely on `LTIUser` (which in this PR is just one).

lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
...
     @classmethod
-    def authenticated_userid(cls, request):
-        if request.lti_user is None:
-            return None
+    def get_lti_user(cls, request) -> LTIUser:
```
marcospri, line 74: This is required to implement by the subclasses.

lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
...
+        raise NotImplemented()
+
+    def permits(self, request, context, permission):
+        return _permits(self, request, context, permission)
+
+    def remember(self, request, userid, **kw):
```
marcospri, line 85: This is the default implementation (as opposed to declaring a method in an "abstract" class).

lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
...
+        pass
+
+    def forget(self, request):
+        pass
 
+    @staticmethod
+    def get_userid(lti_user):
```
marcospri, line 92: This method deals exclusively with these bytes/string transformation from and `lti_user`

lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
...
         # urlsafe_b64encode() returns ASCII bytes but we need unicode, so
         # decode it.
         safe_user_id = safe_user_id_bytes.decode("ascii")
 
-        return ":".join([safe_user_id, str(request.lti_user.application_instance_id)])
+        return ":".join([safe_user_id, str(lti_user.application_instance_id)])
 
-    def identity(self, request):
-        userid = self.authenticated_userid(request)
+    def authenticated_userid(self, request):
```
marcospri, line 104: Now the value it's taken from identity directly. This simplifies things moving forward as only of of `authenticated_userid` / `identity` needs to build `LTIUser`.

lms/security.py
```diff
@@ -66,47 +67,60 @@ def get_route_policy(self, request):
...
-    def permits(self, request, context, permission):
-        return _permits(self, request, context, permission)
+        return Identity(self.get_userid(lti_user), permissions)
 
-    def remember(self, request, userid, **kw):
-        pass
 
-    def forget(self, request):
-        pass
+class LTISecurityPolicy(_LTIUserBasedSecurityPolicy):
```
marcospri, line 127: The existing Policy just defines where it takes the `LTIUser` from.

lms/security.py
```diff
@@ -66,47 +67,67 @@ def get_route_policy(self, request):
...
-            permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]
+        return identity.userid
 
-            if any(
-                role in request.lti_user.roles.lower()
-                for role in ["administrator", "instructor", "teachingassistant"]
-            ):
-                permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
+    def identity(self, request):
+        lti_user = self.get_lti_user(request)
```
marcospri, line 112: The benefits are not obvious here but now this is the only place where we need to call `get_lti_user` so doing error handling would only happen once here.

lms/security.py
seanh: Is `get_userid()` only used internally by this base class? If so it should maybe be `_get_userid()`.

seanh (commented):
What do you think about following the lead of Pyramid's security helpers and using composition rather than inheritance? Details below...

lms/security.py
```diff
@@ -78,47 +79,67 @@ def forget(self, request):
...
-    def authenticated_userid(cls, request):
-        if request.lti_user is None:
-            return None
+    def get_lti_user(cls, request) -> LTIUser:  # pragma: no cover
+        """
+        Get and LTIUser to use as the authenticated user in this policy.
+
+        How this LTIUser is constructed is up to the subclasses.
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
marcospri: I reckon inheritance is the right tool here but it might not be evident because there's only one class in this PR, in a follow if using composition we'd need:

```python
class LTILaunchesPolicy: # New class
## This is all the same to LTISecurityPolicy
  def __init__(self):
        self._helper = LTIUserAuthenticationHelper()

    def permits(self, *args, **kwargs):
        # _permits() is already a compositional helper, just a function not a class.
        return _permits(*args, **kwargs)

    # There's probably no need to inherit or use a helper for remember() and 
    # forget() if they don't actually do anything, and it's nice to be able to see
    # easily right here in LTISecurityPolicy that it doesn't support remember() and
    # forget(). You will have to duplicate these lines in each security policy but :shrug:
    def remember(self, *args, **kwargs):
         pass

    def forget(self, *args, **kwargs):
        pass

    # authenticated_userid() and identity() seem to be where LTIUserAuthenticationHelper
    # provides its real value.
    def authenticated_userid(self, request):
        lti_user = self._get_lti_user(request)
        return self._helper.authenticated_userid(lti_user, request)

    def identity(self, request):
        lti_user = self._get_lti_user(request)
        return self._helper.identity(lti_user, request)

# This is different
     def _get_lti_user(self, request):
        ...
```

We could probably still trim some of that repetition but again, I think here inheritance is the answer, we know now that all the classes we are going to implement are not going to extend any of the other methods (we are refactoring the existing ones into this shape)  .

If keep adding policies and having to modify any other methods we can always refactor again to use composition instead.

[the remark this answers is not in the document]
marcospri: > It would probably make unit tests easier to write as you can unit test helper functions or helper classes and then separately unit test the security policies that use those helpers, patching or passing in mocks in place of the helpers. With subclassing it's not possible for tests to patch or inject the base class in order to test the subclasses so you can't cleanly break things into separate units for testing separately.

I disagree with this, we only need to test the method that the sub classes define



We could go the composing route but based on an `lti_getter` and use the same class for all of them


```python
def LTIUserSecurityPolicy:
  def __init__(self, get_lti_user):
    self._get_lti_user = get_lti_user
    ...
   def identity(self, request):
     lti_user = self._get_lti_user(request)
     return self._helper.identity(lti_user, request)
  
```

and on our SecurityPolicy do something like


```python
if path.starswith("/lti_launches`): 
  return LTIUserSecurityPolicy(launches_lti_user)
```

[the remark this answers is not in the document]
marcospri: Added a version of the "injected _get_lti_user" version in this commit:

https://github.com/hypothesis/lms/pull/4394/commits/a6faf7cab3f6bd7dcc51a5f325d55a2667b392b8


I prefer this approach but it means that instead of having a subclass (or a separate class with composition) per policy we'll have a function/callable for each and that's configured on the top level "get_policy".

tests/unit/lms/security_test.py
```diff
@@ -82,10 +82,9 @@ def policy(self):
 
 
 @pytest.mark.usefixtures("pyramid_config")
-class TestLTISecurityPolicy:
+class TestLTIUserSecurityPolicy:
     def test_it_returns_empty_identity_if_theres_no_lti_user(self, pyramid_request):
-        pyramid_request.lti_user = None
-        policy = LTISecurityPolicy()
+        policy = LTIUserSecurityPolicy(Mock(return_value=None))
```
seanh, line 87: Can we autospec this so that it ensures that `_get_lti_user()` is called with the right number of args? 

```suggestion
        policy = LTIUserSecurityPolicy(create_autospec(_get_lti_user, return_value=None))
```

Or if you don't want to use the `_get_lti_user()` function specifically (since in future that'll be split into more than one function) then you can do `create_autospace(lambda request: None, return_value=None)` or `security.py` could define an explicit interface for the getter functions and the tests could autospec the interface.

And the same for the other calls to `Mock()` in this PR. In general we should always avoid creating unspecced mocks because you can call any method on them with any arguments and it'll always succeed.

seanh (approved):
:+1: Yep, I think the injection is a really neat approach. No code duplication, and cleanly separates the units for testing

PR #4395, opened 2022-08-31
Created workflow: tasks.yml

indigobravo:
A workflow for running hypothesis commands inside an Elastic Beanstalk
Docker container

Part of the work towards: https://github.com/hypothesis/playbook/issues/957

---

This is a very similar workflow to what is being used by H. A PR with a better functional description can be found: https://github.com/hypothesis/h/pull/7580

The difference with the H workflow is we are running, `hypothesis` commands. In the LMS workflow we are running `alembic` commands.

### DIFF
--- a/.github/workflows/tasks.yml
+++ b/.github/workflows/tasks.yml
@@ -1,0 +1,113 @@
+# Run database migration commands inside the Docker container running
+# the Elastic Beanstalk application
+
+name: Tasks
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


robertknight (approved):
Related discussion - https://hypothes-is.slack.com/archives/CR3E3S7K8/p1662020601283479

marcospri (approved):
I can't spot any issues here. We probably should run over all of this on QA as sanity check after merging.

PR #4401, opened 2022-09-01
Move LTI authentication to LTIUserSecurityPolicy

marcospri:
Use `LTIUserSecurityPolicy` introduced in https://github.com/hypothesis/lms/pull/4394 for LTIAuthentication and expose any validation error in them on `forbidden_view`.

# Testing

- Launch an assignment should work as before in this branch: https://hypothesis.instructure.com/courses/125/assignments/873

- Same for deep linking https://hypothesis.instructure.com/courses/125/assignments/3118/edit?name=Testing+upgrade+-+marcos&due_at=null&points_possible=10

- Switch to `main` a make change that would result in a validation error

```
diff 
--git a/lms/validation/_lti_launch_params.py b/lms/validation/_lti_launch_params.py
index 0ad79a84..e27a7cf8 100644
--- a/lms/validation/_lti_launch_params.py
+++ b/lms/validation/_lti_launch_params.py
@@ -25,7 +25,7 @@ class LTIV11CoreSchema(PyramidRequestSchema):
 
     user_id = fields.Str(required=True)
     roles = fields.Str(required=True)
-    tool_consumer_instance_guid = fields.Str(required=True)
+    tool_consumer_instance_guid = fields.Float(required=True)
     lis_person_name_given = fields.Str(load_default="")
     lis_person_name_family = fields.Str(load_default="")
     lis_person_name_full = fields.Str(load_default="")
~
```

- Launching  https://hypothesis.instructure.com/courses/125/assignments/873 fails with 

 
`You're not authorized to view this page`

- Switch to this branch, `security-refactor-launches`, re-launch and the error message will point to the offending field:

```

form:
tool_consumer_instance_guid
````

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -101,8 +101,8 @@ class LTIUserSecurityPolicy:
 
         return Identity(self._get_userid(lti_user), permissions)
 
-    def permits(self, request, context, permission):
-        return _permits(self, request, context, permission)
+    def permits(self, request, _context, permission):
+        return _permits(self.identity(request), permission)
 
     def remember(self, request, userid, **kw):
         pass
@@ -120,12 +120,11 @@ class LMSGoogleSecurityPolicy(GoogleSecurityPolicy):
 
         return Identity("", [])
 
-    def permits(self, request, context, permission):
-        return _permits(self, request, context, permission)
+    def permits(self, request, _context, permission):
+        return _permits(self.identity(request), permission)
 
 
-def _permits(policy, request, _context, permission):
-    identity = policy.identity(request)
+def _permits(identity, permission):
     if identity and permission in identity.permissions:
         return Allowed("allowed")
 
--- a/lms/security.py
+++ b/lms/security.py
@@ -1,7 +1,7 @@
 import base64
 from enum import Enum
 from functools import lru_cache, partial
-from typing import Callable, List, NamedTuple
+from typing import Callable, List, NamedTuple, Optional
 
 from pyramid.request import Request
 from pyramid.security import Allowed, Denied
@@ -29,6 +29,7 @@ class DeniedWithValidationError(Denied):
 class Identity(NamedTuple):
     userid: str
     permissions: List[str]
+    lti_user: Optional[LTIUser] = None
 
 
 class Permissions(Enum):
@@ -120,7 +121,7 @@ class LTIUserSecurityPolicy:
         ):
             permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
 
-        return Identity(self._get_userid(lti_user), permissions)
+        return Identity(self._get_userid(lti_user), permissions, lti_user)
 
     def permits(self, request, _context, permission):
         identity = self.identity(request)
@@ -163,7 +164,7 @@ def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
     return schema()
 
 
-def _get_lti_user(request) -> LTIUser:
+def _get_lti_user(request, from_identity=False) -> LTIUser:
     """
     Return a models.LTIUser for the authenticated LTI user.
 
@@ -175,8 +176,10 @@ def _get_lti_user(request) -> LTIUser:
 
     :rtype: models.LTIUser
     """
-    bearer_token_schema = BearerTokenSchema(request)
+    if from_identity and request.identity and request.identity.lti_user:
+        return request.identity.lti_user
 
+    bearer_token_schema = BearerTokenSchema(request)
     schemas = [
         LTI11AuthSchema(request).lti_user,
         partial(bearer_token_schema.lti_user, location="headers"),
@@ -215,5 +218,10 @@ def _get_user(request):
 
 def includeme(config):
     config.set_security_policy(SecurityPolicy())
-    config.add_request_method(_get_lti_user, name="lti_user", property=True, reify=True)
+    config.add_request_method(
+        partial(_get_lti_user, from_identity=True),
+        name="lti_user",
+        property=True,
+        reify=True,
+    )
     config.add_request_method(_get_user, name="user", property=True, reify=True)
--- a/lms/security.py
+++ b/lms/security.py
@@ -46,6 +46,9 @@ def get_policy(request):
     if path.startswith("/admin") or path.startswith("/googleauth"):
         return LMSGoogleSecurityPolicy()
 
+    if path in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"}:
+        return LTIUserSecurityPolicy(_get_lti_user_from_lti_launch_params)
+
     return LTIUserSecurityPolicy(_get_lti_user)
 
 
@@ -151,6 +154,13 @@ def _permits(identity, permission):
         return Allowed("allowed")
 
     return Denied("denied")
+
+
+def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
+    schema = LTI11AuthSchema(request).lti_user
+    if "id_token" in request.params:
+        schema = LTI13AuthSchema(request).lti_user
+    return schema()
 
 
 def _get_lti_user(request) -> LTIUser:
--- a/lms/security.py
+++ b/lms/security.py
@@ -18,6 +18,14 @@ from lms.validation.authentication import (
 )
 
 
+class DeniedWithValidationError(Denied):
+    """A denial of an action due to a validation error."""
+
+    def __init__(self, validation_error: ValidationError):
+        super().__init__()
+        self.validation_error = validation_error
+
+
 class Identity(NamedTuple):
     userid: str
     permissions: List[str]
@@ -65,6 +73,7 @@ class LTIUserSecurityPolicy:
 
     def __init__(self, get_lti_user: Callable[[Request], LTIUser]):
         self._get_lti_user = get_lti_user
+        self._validation_error = None
 
     @staticmethod
     def _get_userid(lti_user):
@@ -87,7 +96,16 @@ class LTIUserSecurityPolicy:
         return identity.userid
 
     def identity(self, request):
-        lti_user = self._get_lti_user(request)
+        lti_user = None
+        try:
+            lti_user = self._get_lti_user(request)
+        except ValidationError as err:
+            # Keep a reference to the exception
+            # to later return a DeniedWithValidationError in `permits`
+            self._validation_error = err
+        else:
+            self._validation_error = None
+
         if lti_user is None:
             return Identity("", [])
 
@@ -102,7 +120,11 @@ class LTIUserSecurityPolicy:
         return Identity(self._get_userid(lti_user), permissions)
 
     def permits(self, request, _context, permission):
-        return _permits(self.identity(request), permission)
+        identity = self.identity(request)
+        if self._validation_error:
+            return DeniedWithValidationError(self._validation_error)
+
+        return _permits(identity, permission)
 
     def remember(self, request, userid, **kw):
         pass
--- a/lms/security.py
+++ b/lms/security.py
@@ -48,9 +48,9 @@ def get_policy(request):
         return LMSGoogleSecurityPolicy()
 
     if path in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"}:
-        return LTIUserSecurityPolicy(_get_lti_user_from_lti_launch_params)
+        return LTIUserSecurityPolicy(get_lti_user_from_launch_params)
 
-    return LTIUserSecurityPolicy(_get_lti_user)
+    return LTIUserSecurityPolicy(get_lti_user)
 
 
 class SecurityPolicy:
@@ -75,9 +75,8 @@ class SecurityPolicy:
 class LTIUserSecurityPolicy:
     """Security policy based on the information of an LTIUser."""
 
-    def __init__(self, get_lti_user: Callable[[Request], LTIUser]):
-        self._get_lti_user = get_lti_user
-        self._validation_error = None
+    def __init__(self, get_lti_user_: Callable[[Request], LTIUser]):
+        self._get_lti_user = get_lti_user_
 
     @staticmethod
     def _get_userid(lti_user):
@@ -100,15 +99,10 @@ class LTIUserSecurityPolicy:
         return identity.userid
 
     def identity(self, request):
-        lti_user = None
         try:
             lti_user = self._get_lti_user(request)
-        except ValidationError as err:
-            # Keep a reference to the exception
-            # to later return a DeniedWithValidationError in `permits`
-            self._validation_error = err
-        else:
-            self._validation_error = None
+        except ValidationError:
+            return Identity("", [])
 
         if lti_user is None:
             return Identity("", [])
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lms/security.py [hunk 1]
```diff
@@ -134,28 +172,24 @@ def _get_lti_user(request):
...
     lti_user = None
-    for schema in schemas:
-        try:
-            lti_user = schema()
-            break
-        except ValidationError:
-            continue
+    if from_identity and request.identity and request.identity.lti_user:
+        lti_user = request.identity.lti_user
+    else:
```
marcospri, line 186: Adding this `else` to still do the user_service call at the bottom in both cases.

lms/security.py
```diff
@@ -110,19 +142,25 @@ def identity(self, request):
...
 
-    def permits(self, request, context, permission):
-        return _permits(self, request, context, permission)
+    def permits(self, request, _context, permission):
+        return _permits(self.identity(request), permission)
 
 
-def _permits(policy, request, _context, permission):
-    identity = policy.identity(request)
+def _permits(identity, permission):
```
marcospri, line 153: `permits` only takes the arguments it needs

lms/security.py
```diff
@@ -38,6 +47,9 @@ def get_policy(request):
     if path.startswith("/admin") or path.startswith("/googleauth"):
         return LMSGoogleSecurityPolicy()
 
+    if path in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"}:
```
marcospri, line 50: We can use exact matches for these.

lms/security.py
```diff
@@ -87,7 +100,16 @@ def authenticated_userid(self, request):
...
-        lti_user = self._get_lti_user(request)
+        lti_user = None
+        try:
+            lti_user = self._get_lti_user(request)
+        except ValidationError as err:
+            # Keep a reference to the exception
+            # to later return a DeniedWithReason in `permits`
+            self._validation_error = err
+        else:
+            self._validation_error = None
```
marcospri, line 111: This branch is not necessary as this is instantiated per request but I reckon that not evident if reading this method alone.

lms/security.py
```diff
@@ -120,19 +146,25 @@ def identity(self, request):
...
+def _permits(identity, permission):
     if identity and permission in identity.permissions:
         return Allowed("allowed")
 
     return Denied("denied")
 
 
-def _get_lti_user(request) -> LTIUser:
+def _get_launches_lti_user(request) -> LTIUser:
+    schema = LTI11AuthSchema(request).lti_user
```
marcospri, line 161: Same URLs would work with both. We only try the 1.3 if the required parameter is there.

lms/security.py
```diff
@@ -142,30 +174,30 @@ def _get_lti_user(request) -> LTIUser:
...
+    if from_identity and request.identity and request.identity.lti_user:
+        lti_user = request.identity.lti_user
+    else:
+
+        bearer_token_schema = BearerTokenSchema(request)
+        schemas = [
+            partial(bearer_token_schema.lti_user, location="headers"),
+            partial(bearer_token_schema.lti_user, location="querystring"),
+            partial(bearer_token_schema.lti_user, location="form"),
+            OAuthCallbackSchema(request).lti_user,
```
marcospri, line 193: Plan now is to move this out one by one and then this method becomes just: `return request.identity.lti_user`

(Or get's removed and we remove every `request.lti_user` usage by request.identity.lti_user)

tests/unit/lms/security_test.py
```diff
@@ -327,63 +365,60 @@ def test_it_returns_LTIUsers_from_authorization_form_fields(
...
+    @pytest.mark.usefixtures("pyramid_request_with_identity")
+    def test_it_picks_lit_user_from_identity(self, pyramid_request):
+        lti_user = _get_lti_user(pyramid_request, from_identity=True)
+
+        assert lti_user == sentinel.lti_user
+
+    @pytest.fixture
+    def pyramid_request_with_identity(self):
+
+        with patch.object(
```
marcospri, line 413: This is a bit too much but haven't find a better way

lms/security.py
```diff
@@ -120,19 +146,25 @@ def identity(self, request):
...
-    identity = policy.identity(request)
+def _permits(identity, permission):
     if identity and permission in identity.permissions:
         return Allowed("allowed")
 
     return Denied("denied")
 
 
-def _get_lti_user(request) -> LTIUser:
+def _get_launches_lti_user(request) -> LTIUser:
```
seanh, line 160: Maybe:

```suggestion
def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
```

?

lms/security.py
seanh: :+1:

The two cases in one method is a bit ugly right now, it'll be good once these have been split out

lms/views/exceptions.py
```diff
@@ -27,6 +27,13 @@ def notfound(self):
 
     @forbidden_view_config()
     def forbidden(self):
+        if reason := getattr(self.exception.result, "reason", None):
+            self.request.override_renderer = (
+                "lms:templates/validation_error.html.jinja2"
+            )
+            self.request.response.status_int = 403
+            return {"error": reason}
```
seanh, line 35: This definitely looks like `reason` should be called `validation_error`

seanh (commented):
I wasn't able to finish reviewing this before going on holiday, just posting as far as I've got.

tests/unit/lms/security_test.py
marcospri: TODO: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.testing_securitypolicy

lms/security.py
```diff
@@ -18,9 +18,18 @@
 )
 
 
+class DeniedWithValidationError(Denied):
+    """A denial of an action due to a validation error."""
+
+    def __init__(self, validation_error: ValidationError):
+        super().__init__()
+        self.validation_error = validation_error
```
seanh, line 26: :+1:

lms/security.py
```diff
@@ -120,19 +146,25 @@ def identity(self, request):
...
-    identity = policy.identity(request)
+def _permits(identity, permission):
     if identity and permission in identity.permissions:
         return Allowed("allowed")
 
     return Denied("denied")
 
 
-def _get_lti_user(request) -> LTIUser:
+def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
```
seanh, line 160: Apologies because I know this name was my suggestion but we can probably get away with avoiding the repetition of `lti` and it'll still be clear enough:

```suggestion
def _get_lti_user_from_launch_params(request) -> LTIUser:
```

lms/security.py
```diff
@@ -99,10 +121,14 @@ def identity(self, request):
...
             permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)
 
-        return Identity(self._get_userid(lti_user), permissions)
+        return Identity(self._get_userid(lti_user), permissions, lti_user)
 
-    def permits(self, request, context, permission):
-        return _permits(self, request, context, permission)
+    def permits(self, request, _context, permission):
+        identity = self.identity(request)
+        if self._validation_error:
```
seanh, line 128:
```suggestion
        if self._validation_error is not None:
```

You might want to be a bit careful here because exceptions are sometimes falsey. IIRC this is true of Pyramid's built-in HTTP exception classes -- the ones for `{2,3}xx`'s are truthy but the ones for `{4,5}xx`'s are falsey -- and this has bitten us a couple of times in the past. I believe you're dealing with a custom validation error class here but still.

lms/security.py
seanh: I wonder if the code would be simpler if `self._validation_error` could be avoided? Mutable state often makes if difficult to understand the flow of code. Could `permits()` below call `self._get_lti_user()` itself and catch the `ValidationError`,  instead of using `self._validation_error`?

```python
def permits(self, request, _context, permission):
    try:
        self._get_lti_user(request)
        return DeniedWithValidationError(err)

    return _permits(self.identity(request), permission)
```

Then `identity()` would also do something similar:

```python
def identity(self, request):
    try:
        lti_user = self._get_lti_user(request)
    except ValidationError:
        return Identity("", [])

    permissions = [Permissions.LTI_LAUNCH_ASSIGNMENT, Permissions.API]

    if any(
        role in lti_user.roles.lower()
        for role in ["administrator", "instructor", "teachingassistant"]
    ):
        permissions.append(Permissions.LTI_CONFIGURE_ASSIGNMENT)

    return Identity(self._get_userid(lti_user), permissions, lti_user)
```

And maybe an `@cache` on the `_get_lti_user*()` functions could save any unnecessary work?

lms/security.py
seanh: :+1:

lms/security.py
```diff
@@ -120,19 +146,25 @@ def identity(self, request):
...
 
     return Denied("denied")
 
 
-def _get_lti_user(request) -> LTIUser:
+def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
+    schema = LTI11AuthSchema(request).lti_user
+    if "id_token" in request.params:
+        schema = LTI13AuthSchema(request).lti_user
+    return schema()
```
seanh, line 164: Would early-return make this function a little simpler? I think the below works the same? But no need to keep track of a local variable `schema`.

```python
def _get_lti_user_from_lti_launch_params(request) -> LTIUser:
    if "id_token" in request.params:
        return LTI13AuthSchema(request).lti_user()

    return LTI11AuthSchema(request).lti_user()
```

lms/security.py
```diff
@@ -120,19 +146,25 @@ def identity(self, request):
...
+
+
+def _get_lti_user(request, from_identity=False) -> LTIUser:
```
seanh, line 167: I think the docstring of this function is now out of date?

lms/views/exceptions.py
```diff
@@ -27,6 +27,13 @@ def notfound(self):
 
     @forbidden_view_config()
     def forbidden(self):
+        if validation_error := getattr(self.exception.result, "validation_error", None):
```
seanh, line 30: Does `exception.result` always exist or might this crash in some circumstances?

tests/unit/lms/security_test.py
```diff
@@ -147,6 +149,16 @@ def test_permits_denied(self, pyramid_request):
...
 
+    def test_permits_denied_with_validation_error(self, pyramid_request):
+        validation_error = ValidationError(sentinel.message)
+        policy = LTIUserSecurityPolicy(
+            create_autospec(_get_lti_user, side_effect=validation_error)
+        )
+
+        is_allowed = policy.permits(pyramid_request, None, "some-permission")
+
+        assert is_allowed.validation_error == validation_error
```
seanh, line 160: Do we also need to assert that it returned a denied rather than an allowed?

seanh (changes requested):
I wonder if `_get_lti_user()` and `_get_lti_user_from_launch_params()` should actually be public functions (no underscore) since we want to unit test them directly?

Otherwise, just a handful of code suggestions

lms/views/exceptions.py
marcospri: Seems to be part of the exception object in all cases: 

https://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html#pyramid.httpexceptions.HTTPForbidden

tests/unit/lms/security_test.py
marcospri: Uses testing_securitypolicy, much cleaner.

tests/functional/lti_certification/v13/conftest.py
```diff
@@ -59,6 +59,7 @@ def application_instance(lti_registration):
         tool_consumer_instance_guid="TEST_CONSUMER_INSTANCE_GUID",
         lti_registration=lti_registration,
         deployment_id="testdeploy",
+        organization=factories.Organization(),
```
marcospri, line 62: These two seem to be required but I can't point exactly what change in this PR makes it necessary.

I think it working before was an accident of how / when we flush to the DB.

lms/security.py
```diff
@@ -87,7 +99,12 @@ def authenticated_userid(self, request):
         return identity.userid
 
     def identity(self, request):
-        lti_user = self._get_lti_user(request)
+        lti_user = None
```
seanh, line 102: I think this line is unnecessary now? `self._get_lti_user()` will either return an `LTIUser`, return `None`, or raise `ValidationError` (in which case we early return).

lms/security.py
```diff
@@ -63,8 +75,8 @@ def forget(self, request):
 class LTIUserSecurityPolicy:
     """Security policy based on the information of an LTIUser."""
 
-    def __init__(self, get_lti_user: Callable[[Request], LTIUser]):
-        self._get_lti_user = get_lti_user
+    def __init__(self, get_lti_user_: Callable[[Request], LTIUser]):
```
seanh, line 78: I'm not a Python typing expert but can `get_lti_user()` return `None`? Should this be   `Callable[[Request], Optional[LTIUser]]`?

lms/security.py
```diff
@@ -120,52 +144,58 @@ def identity(self, request):
...
     return Denied("denied")
 
 
-def _get_lti_user(request) -> LTIUser:
+@lru_cache(maxsize=1)
+def get_lti_user_from_launch_params(request) -> LTIUser:
+    if "id_token" in request.params:
+        return LTI13AuthSchema(request).lti_user()
+
+    return LTI11AuthSchema(request).lti_user()
```
seanh, line 163: Do these schemas ever return `None`? Should it be `Optional[LTIUser]`?

lms/security.py
```diff
@@ -120,52 +144,58 @@ def identity(self, request):
...
+
+
+@lru_cache(maxsize=1)
+def get_lti_user(request, from_identity=False) -> Optional[LTIUser]:
     """
     Return a models.LTIUser for the authenticated LTI user.
 
-    Get the authenticated user from the validated LTI launch params or, failing
-    that, from one of our LTI bearer tokens (also validated).
+    Get the authenticated from from one of our LTI bearer tokens.
```
seanh, line 171:
```suggestion
    Get the authenticated user from one of our LTI bearer tokens.
```

lms/security.py
```diff
@@ -120,52 +144,58 @@ def identity(self, request):
...
 
-    If the request doesn't contain either valid LTI launch params or a valid
-    bearer token then return ``None``.
+    If the request doesn't contain a valid bearer token then it returns ``None``.
```
seanh, line 173:
```suggestion
    If the request doesn't contain a valid bearer token then return None.
```

lms/security.py
```diff
@@ -120,52 +144,58 @@ def identity(self, request):
...
+
+    :param: from_identity Use the lt_user from request.identity.
```
seanh, line 175:
```suggestion
    :param from_identity: Use the lti_user from request.identity.
```

lms/security.py
```diff
@@ -120,52 +144,58 @@ def identity(self, request):
...
+        Defaults to false to avoid a circular dependency and allow
```
seanh, line 176:
```suggestion
        Defaults to False to avoid a circular dependency and allow
```

seanh (approved):

lms/security.py
marcospri: No, they either raise or return the LTIUser.

lms/security.py
marcospri: The aim is for it to be the non-optional one.  But because we still rely on the "legacy" `get_lti_user` that one can in theory return `None` but we assume it won't.

We can lie a bit on the typing annotation.

PR #4409, opened 2022-09-07
Allow editing canvas API details in the admin pages

marcospri:
Allow editing AIs Canvas API details

Simpler alternative than https://github.com/hypothesis/lms/pull/4376

For #4304 

#  Testing

- Go over to http://localhost:8001/admin/instance/id/100/
- Enter any values for key/secret
- They are saved on the AI 

```tox -qe dockercompose -- exec postgres psql -U postgres -c "select developer_key, developer_secret, aes_cipher_iv from application_instances where id = 100"```

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -116,6 +116,29 @@ class ApplicationInstanceService:
 
         return query.all()
 
+    def update_application_instance(  # pylint:disable=too-many-arguments
+        self,
+        application_instance,
+        lms_url=None,
+        deployment_id=None,
+        developer_key=None,
+        developer_secret=None,
+    ):
+        if lms_url:
+            application_instance.lms_url = lms_url
+
+        if deployment_id:
+            application_instance.deployment_id = deployment_id
+
+        if developer_key:
+            application_instance.developer_key = developer_key
+
+        if developer_secret:
+            aes_iv = self._aes_service.build_iv()
+            encrypted_secret = self._aes_service.encrypt(aes_iv, developer_secret)
+            application_instance.aes_cipher_iv = aes_iv
+            application_instance.developer_secret = encrypted_secret
+
     def create_application_instance(  # pylint:disable=too-many-arguments
         self,
         lms_url,
@@ -131,26 +154,25 @@ class ApplicationInstanceService:
             "Hypothesis" + secrets.token_hex(16) if not deployment_id else None
         )
 
-        if developer_secret and developer_key:
-            aes_iv = self._aes_service.build_iv()
-            encrypted_secret = self._aes_service.encrypt(aes_iv, developer_secret)
-        else:
-            # If either one of developer_key or developer_secret is missing, then we
-            # don't save the other one either.
-            developer_key = encrypted_secret = developer_secret = aes_iv = None
-
         application_instance = ApplicationInstance(
             consumer_key=consumer_key,
             shared_secret=secrets.token_hex(32),
             lms_url=lms_url,
             requesters_email=email,
-            developer_key=developer_key,
-            developer_secret=encrypted_secret,
-            aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
             settings=settings or {},
             deployment_id=deployment_id,
             lti_registration_id=lti_registration_id,
+        )
+        # If either one of developer_key or developer_secret is missing,
+        # then we won't set either on creation
+        if not all([developer_secret, developer_key]):
+            developer_key = developer_secret = None
+
+        self.update_application_instance(
+            application_instance,
+            developer_key=developer_key,
+            developer_secret=developer_secret,
         )
         self._db.add(application_instance)
         self._db.flush()  # Force the returned AI to have an ID
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -79,7 +79,13 @@
     <fieldset class="box">
         <legend class="label has-text-centered">Canvas settings</legend>
         {{ macros.disabled_text_field("API domain", instance.custom_canvas_api_domain) }}
-        {{ macros.disabled_text_field("Developer key", instance.developer_key) }}
+        {{ macros.form_text_field(request, "Developer key", "developer_key", field_value=instance.developer_key) }}
+        {{ macros.form_text_field(
+            request,
+            "Developer secret",
+            "developer_secret",
+            placeholder="*" * 25 if instance.developer_secret else "")
+        }}
         {{ settings_checkbox("Sections enabled", "canvas", "sections_enabled") }}
         {{ settings_checkbox("Groups enabled", "canvas", "groups_enabled") }}
     </fieldset>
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lms/templates/admin/instance.html.jinja2 [hunk 1]
```diff
@@ -80,6 +80,13 @@
         <legend class="label has-text-centered">Canvas settings</legend>
         {{ macros.disabled_text_field("API domain", instance.custom_canvas_api_domain) }}
         {{ macros.disabled_text_field("Developer key", instance.developer_key) }}
+        {{ macros.form_text_field(request, "Developer key", "developer_key", field_value=instance.developer_key) }}
+        {{ macros.form_text_field(
+            request,
+            "Developer secret",
+            "developer_secret",
+            placeholder="*" * 25 if instance.developer_secret else "")
```
marcospri, line 88: I think this is good compromise between having to decrypt the actual key and not having any hint about the key having a value currently or not.

lms/services/application_instance.py
```diff
@@ -131,27 +144,19 @@ def create_application_instance(  # pylint:disable=too-many-arguments
             "Hypothesis" + secrets.token_hex(16) if not deployment_id else None
         )
 
-        if developer_secret and developer_key:
```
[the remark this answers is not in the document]
jon-betts: This had to move out because previously this was only done on create, now you can do it on update as well.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -381,6 +381,30 @@ def test_update_instance_saves_ai_settings(
...
+        views.update_instance()
+
+        if update:
+            application_instance_service.set_canvas_api_details.assert_called_once_with(
+                application_instance_service.get_by_id.return_value,
+                key,
+                secret,
+            )
+        else:
+            application_instance_service.set_canvas_api_details.assert_not_called()
```
jon-betts, line 406:
```suggestion
    @pytest.mark.parametrize("key,", (" ", " key "))
    @pytest.mark.parametrize("secret", (" ", " secret "))
    def test_update_instance_canvas_api_details(
        self, pyramid_request, application_instance_service, views, key, secret
    ):
        pyramid_request.params = {"developer_key": key, "developer_secret": secret}

        views.update_instance()

        application_instance_service.set_canvas_api_details.assert_called_once_with(
            application_instance_service.get_by_id.return_value,
            developer_key=key.strip() if key else '',
            developer_secret=secret.strip() if secret else '',
        )
```

This isn't great, but it's the best I could come up with.

tests/unit/lms/services/application_instance_test.py
```diff
@@ -162,6 +182,13 @@ def service(self, db_session, pyramid_request, aes_service):
             db=db_session, request=pyramid_request, aes_service=aes_service
         )
 
+    @pytest.fixture
+    def set_canvas_api_details(self, service):
+        with mock.patch.object(
+            service, "set_canvas_api_details"
+        ) as set_canvas_api_details:
+            yield set_canvas_api_details
```
jon-betts, line 190: :+1: We've tested this separately. No need to duplicate the tests for the create.

tests/unit/lms/services/application_instance_test.py
```diff
@@ -84,17 +84,42 @@ def test_get_by_deployment_id_raises_on_missing(
             (None, None),
         ],
     )
```
jon-betts, line 86: Hard to leave a suggestion here as it wasn't in this PR, but we can let `pytest` do the combinations for us:

```suggestion
    @pytest.mark.parametrize("developer_secret", (None, "SECRET"))
    @pytest.mark.parametrize("developer_key", (None, "KEY"))
```

lms/services/application_instance.py
```diff
@@ -116,6 +116,19 @@ def search(
 
         return query.all()
 
+    def set_canvas_api_details(
+        self, application_instance, developer_key, developer_secret
+    ):
```
jon-betts, line 121: Although this happens to only deal with Canvas stuff, is this actually the start of an `update_application_instance()`?

jon-betts (changes requested):
I don't think this is quite right at the moment. Try the following:

 * Set the key and secret on an application instance
 * Save
 * Now change the key only
 * Save
 * Changes to the key are not saved

This is because they secret is blank and saving requires both key and secret.

lms/services/application_instance.py
marcospri: Done this but felt like I should include other fields we do upgrade instead of using two different method to update AI fields.

lms/services/application_instance.py
```diff
@@ -131,27 +154,26 @@ def create_application_instance(  # pylint:disable=too-many-arguments
...
-            developer_secret=encrypted_secret,
-            aes_cipher_iv=aes_iv,
             created=datetime.utcnow(),
             settings=settings or {},
             deployment_id=deployment_id,
             lti_registration_id=lti_registration_id,
         )
+        # If either one of developer_key or developer_secret is missing,
+        # then we won't set either on creation
+        if not all([developer_secret, developer_key]):
```
marcospri, line 169: On creation it makes sense to keep the original behavior, only set one if both are present.

While upgrading update/set either.

lms/views/admin/application_instance.py
```diff
@@ -264,9 +264,13 @@ def show_instance(self):
     def update_instance(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
-        for ai_field in ["lms_url", "deployment_id"]:
```
marcospri, line 267: Keeping this while having `update_application_instance` was a bit odd. Doing it all in one place now.

tests/unit/lms/services/application_instance_test.py
```diff
@@ -75,20 +75,60 @@ def test_get_by_deployment_id_raises_on_missing(
...
+        if developer_secret:
+            aes_service.build_iv.assert_called_once()
+            aes_service.encrypt.assert_called_once_with(
+                aes_service.build_iv.return_value, developer_secret
+            )
+
+            assert application_instance.developer_secret
+
+        if developer_key:
+            assert application_instance.developer_key == developer_key
```
jon-betts, line 110: These tests don't make any assertions when these values are not truthy, so that side of the test doesn't test anything. We should probably make assertions about what should happen in both cases.

lms/services/application_instance.py
jon-betts: One way to express the difference between being passed None and not being passed something is to use something like `...`. This is pretty much good enough:

```python
def update_application_instance(self, lms_url=...):
    if lms_url is not ...:
        # Do whatever
```

In the out-there world where you might actually want to store `...` you can create an alternative singleton like `MISSING = object()` and then use that instead.

jon-betts (approved):
This works and is an improvement on what came before (where you couldn't change this stuff). I think there is still some UI weirdness, as there is no way to remove a key or secret, even though the column is nullable.

I think there might also be some optimisation we could do to the new test by maybe _not_ parameterising it?

tests/unit/lms/services/application_instance_test.py
marcospri: > These tests don't make any assertions when these values are not truthy, so that side of the test doesn't test anything. We should probably make assertions about what should happen in both cases.

Not ignoring this feedback, handling it in a different PR, want to get merge all the admin PRs at the same time.

Issue #4413, opened 2022-09-08
Honour the "enabled" flag on organisations


jon-betts:
If a customer is persistently naughty and won't pay their bills. It would be a nice to be able to disable their account in some way.

We could currently break their application instance, but there's nothing stopping them from creating a new one.

By having a switch at the organization level we would have greater control as:

 * All new application instances with the same GUID will automatically be attached to the organization
 * This means creating a new application instance will link back to the disabled org
 * If you create an application instance and change the GUID on your LMS you would get allocated to a new enabled org but... you will lose all your current content, which might defeat the purpose

### For this ticket

 * When the "enabled" flag is set to False, we will prevent all LMS operations
 * We will log that access was prevented (so we can look for this in papertrail)
 * We will present an error explaining to the user what's happening **(with what wording?)**
 * This error would not distinguish between users, and so would be shown to students and faculty alike

### Not for this ticket

 * This does not cover any UI to perform this work - at the time of writing this would be enabled with Metabase or equivalent
 * Minimal UI for this can come later
 
This also paves the way to greater control in future:

 * We may chose to make [free range orgs created by auto allocation](https://github.com/hypothesis/lms/pull/4332) disabled by default, so that customers have to contact sales
 * In the future we may also make adding new application instances invite only based on an organization, which would largely close the opportunity to dodge this - You could either assign to an org in good standing (and therefore bump their numbers) or not

# Tasks

 * https://github.com/hypothesis/lms/pull/4419

jon-betts:
This has been deployed out to live, but I don't want to test it there (for fairly obvious reasons) until we have a nice safe admin interface for turning it on and off.

PR #4419, opened 2022-09-08
Raise an exception for disabled accounts

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4413

This is a pretty blunt instrument, but any action that attempts to get the "current" application instance will fail with an "AccountDisabled" error.

As this is a serialisable error, and serialisable errors are handled in both the UI and API, this should result in some sensible output in all cases.

## Implementation notes

This might be improved by using the public id. We might want to merge the helper ticket first.

I tried this in the `check_guid_matches()`, but that isn't on the service, so would require moving to the service first. This has the "advantage" of _really_ breaking absolutely everything.

This might not kick in instantly, as a the application instance might be allocated to a new org when it's launched, but that can only happen once... so I don't think it really matters.

## Testing notes

 * Start everything
 * `make sql`
 * `update organization set enabled=False`
 * The result should look like this:

image

 * You should also see log output like (before the exception):
 
> Account access was blocked for application_instance=8 org=1

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -1,17 +1,36 @@
 import secrets
 from datetime import datetime
 from functools import lru_cache
+from logging import getLogger
 from typing import List
 
 from sqlalchemy.exc import NoResultFound
 
 from lms.models import ApplicationInstance, LTIParams, LTIRegistration
+from lms.models.region import Region
 from lms.services.aes import AESService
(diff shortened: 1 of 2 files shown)


marcospri (approved):
Looks good :+1: 

Tested locally and also tripled checked that the we do default to `True` for enabled.

PR #4422, opened 2022-09-12
Enable caching for ESLint and Prettier

robertknight:
See notes in https://github.com/hypothesis/client/pull/4781, though the performance improvement is less for this smaller codebase.


lyzadanger (approved):
Most definitely yes.

Issue #4416, opened 2022-09-08
Add name changing to org admin pages


jon-betts:
You should be able to update the name of an organisation in the admin pages.

Issue #4203, opened 2022-07-08
Add basic read only admin interface for organizations


jon-betts:
We need to expand our admin pages to let people interact with organisations. There are a few different directions we might go, but all require there to be basic finding and viewing, before any other interactions can take place.

### Required features

 * [x] A page which will display organisations containing:
     * Name + internal id (as per other admin pages)
     * Public id (this should be displayed in the public id format)
     * Enabled status
     * Created / updated
     * A table of linked application instances and link to their pages
     * _We don't need:_ parent links, settings - As these have no effect yet
 * [x] Application instances should have a link to their organisation, or indicate they don't have one

Issue #4410, opened 2022-09-07
Rendering issue when viewing list of app instances from a given reg page. The ai box doesn't dynamically adjust to the browser width, and the browser doesn't know to provide a horizontal scroll bar to get to the **View** button at the right of the ai item


marcospri:
Rendering issue when viewing list of app instances from a given reg page. The ai box doesn't dynamically adjust to the browser width, and the browser doesn't know to provide a horizontal scroll bar to get to the **View** button at the right of the ai item.
![Screen Shot 2022-08-18 at 2 02 18 PM](https://user-images.githubusercontent.com/71603629/185464289-5cba526f-378b-4d88-9126-3dffd6c813e5.png)

_Originally posted by @mattdricker in https://github.com/hypothesis/lms/issues/4304#issuecomment-1219785464_

PR #4423, opened 2022-09-12
Organization detail on the admin pages

marcospri:
For:

- https://github.com/hypothesis/lms/issues/4203
- https://github.com/hypothesis/lms/issues/4416

# Testing 

Once you have a AI with an org naviate to the URL for example something like

 * http://localhost:8001/admin/org/us.lms.org.gkPHJPSRSHC7YWAYJ7s1LA/

- Try the URL with wrong format and non existing ID
- Try navigation to the AI from the list at the bottom
- Navigate back from the AI
- Update the org's name


lms/services/organization.py
```diff
@@ -36,6 +38,41 @@ def get_by_linked_guid(self, guid) -> List[Organization]:
             .all()
         )
 
+    def get_by_public_id(self, public_id: str):
```
marcospri, line 41: This is a long and boring method that could be moved elsewhere in the future (PublicIDService?) but I  reckon it's fine here for now.

lms/services/organization.py
```diff
@@ -36,6 +38,41 @@ def get_by_linked_guid(self, guid) -> List[Organization]:
...
+        try:
+            region_code, app, type_, public_id = public_id.split(".")
+        except ValueError:
+            raise ValidationError(
```
marcospri, line 45: Giving some feedback on each case.

lms/templates/admin/macros.jinja
```diff
@@ -63,6 +63,16 @@
     {{ form_text_field(request, "Token URL", "token_url", lti_registration.token_url) }}
 {% endmacro %}
 
+
+{% macro created_updated_fields(model) %}
+    <fieldset class="box">
+    {{ disabled_text_field("Created", model.created) }}
+    {% if model.updated %}
```
marcospri, line 70: Some older models don't have `updated`

lms/templates/admin/macros.jinja
```diff
@@ -131,4 +140,28 @@
             {{ disabled_text_field("Client ID", lti_registration.client_id) }}
         </div>
     </div>
+    {% else %}
+    <div class="is-size-5 has-text-centered">No LTI registration</div>
```
marcospri, line 144: Adding the message instead of having blank fields on the existing LTIRegistraion marco

lms/templates/admin/organization.html.jinja2
```diff
@@ -0,0 +1,34 @@
...
+{% endblock %}
+
+{% block content %}
+    <fieldset class="box">
+    <legend class="label has-text-centered">Organization</legend>
+    <form method="POST" action="{{ request.route_url("admin.organization", public_id=public_id) }}">
+        <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+        {{ macros.disabled_text_field("ID", public_id) }}
+        {{ macros.form_text_field(request, "Name", "name", org.name) }}
+        {% call macros.field_body("Enabled") %}
```
marcospri, line 16: This is for now just disabled checkbox.

lms/routes.py
```diff
@@ -129,6 +129,11 @@ def includeme(config):  # pylint:disable=too-many-statements
         "admin.instance.new.registration", "/admin/instance/new/registration"
     )
 
+    org_public_id_pat = r"\w{2}\.lms\.org\.\w+"
+    config.add_route(
```
marcospri, line 133: Using the public_id on the URL.

In application instance we use either ID or consumer key that's a bit of an accident.

jon-betts: Should we be this picky about the format? If we change it this would break. Are you imagining we'll arrive here with different ids and need to route?

lms/services/organization.py
```diff
@@ -78,4 +115,4 @@ def auto_assign_organization(
 def service_factory(_context, request) -> OrganizationService:
     """Get a new instance of OrganizationService."""
 
-    return OrganizationService(db_session=request.db)
+    return OrganizationService(db_session=request.db, region=request.region)
```
jon-betts, line 118: It's the way we tend to do things, but it seems a shame that the region is only available on the request rather than in some more global way. We know this before any request is made. The downside of this is, a service which is otherwise a global instance might be forced to become a request instance just to get this.

I wonder if there is any other peg to hang this off?

lms/routes.py
jon-betts: I don't think this regex is right as it doesn't allow dashes as in: `18RYrpLBR1egA-NE4f7A_g`. I'd probably be in favour of dropping it.

lms/templates/admin/macros.jinja
jon-betts: Is the `fieldset` here creating the secondary inner box around this? I think it might be better to let it blend in with the other fields, rather than a box in a box etc.

lms/templates/admin/organization.html.jinja2
```diff
@@ -0,0 +1,34 @@
...
+        {% endcall %}
+        {{ macros.created_updated_fields(org) }}
+        <div class="has-text-right mb-6">
+            <input type="submit" class="button is-primary" value="Save" />
+        </div>
+    </form>
+
+    <legend class="label has-text-centered">Organization application instances</legend>
+        {% if org.application_instances %}
+            {{ macros.instances_table(request, org.application_instances) }}
```
jon-betts, line 30: It would be very good to include some human readable fields for the org, like:

 * tool_consumer_instance_name
 * tool_consumer_instance_contact_email
 * lms_url
 
As these are the fields that would let you know as a human which is which. Currently it's all opaque ids.

I think this is particularly important in the case of orgs, as we are looking for ones which don't belong, and without clicking through to each, there's not much to go on.

If there's not enough room, I think we could afford to drop a few of the opaque fields to make room.

lms/views/admin/organization.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @view_config(
+        route_name="admin.organization",
+        request_method="POST",
+        renderer="lms:templates/admin/organization.html.jinja2",
+    )
+    def update_organization(self):
+        org = self._get_org_or_404(self.request.matchdict["public_id"])
+
+        org.name = self.request.params.get("name", "").strip() or None
```
jon-betts, line 34: I think I can just about cope with the additional complexity this has introduced in this PR. :)

lms/services/organization.py
jon-betts: Maybe pull this out as a separate `_parse_public_id()` method to more cleanly separate the org'y stuff from the id related stuff?

I'm imagining it would be something like:

```python
public_id = self._parse_public_id(public_id, type_="org")
```

So you could assert what you were expecting.

jon-betts: Looking ahead one step, could this be the first arg in a `search()` method? Otherwise I suspect we'll end-up having some duplication.

I'm thinking something like:

```python
def search(self, public_id=None):
    query =  self._db_session.query(Organization)

    if public_id:
        query.filter_by(_public_id=self._parse_public_id(public_id))

    return query.all()
```

lms/templates/admin/organization.html.jinja2
```diff
@@ -0,0 +1,34 @@
+{% import "macros.jinja" as macros %}
+
+{% extends "lms:templates/admin/base.html.jinja2" %}
+{% set public_id = org.public_id(request.region) %}
+{% block header %}
+    Organization {{org.name}} / {{ public_id }}
```
jon-betts, line 6: This is a bit eyeboggling in practice I think as it ends up with things like:

> Organization Blackboard, Inc! / us.lms.org.fY9f8UQORZCPopiRNGkK1w

As most of this detail is replicated below I wonder if we want to use this to display something distinct. For example we don't expose the primary key anywhere, and this is what we might use for most queries. This would like up with other pages, so the title would be:

> Organization 123

lms/routes.py
jon-betts: I'm coming to the conclusion we might be making things overly difficult by using the public id here:

 * Currently this is the only way to access the page, but it won't be soon
 * Keying by primary key would avoid any complication around parsing the id

jon-betts (commented):
I think it's probably worth having a think about whether we need to use the public key here for all id related duties. On other pages we use the primary key, and I think that's complicating things here.

I think we need to be careful about presenting the public id outside of LMS, but I think we can use it internally.

lms/services/organization.py
marcospri: Having to access global configuration through the request is a bit strange but I reckon finding an alternative would be too much pyramid fighting for this PR.

lms/templates/admin/organization.html.jinja2
marcospri: Added name and url, removed GUID.

lms/templates/admin/macros.jinja
```diff
@@ -63,57 +63,51 @@
...
-{% endfor %}
+
+{% macro created_updated_fields(model) %}
+    {{ disabled_text_field("Created", model.created) }}
+    {% if model.updated %}
+        {{ disabled_text_field("Updated", model.updated) }}
+    {% endif %}
+{% endmacro %}
+
+{% macro object_list_table(request, route, objects, headers, fields) %}
```
marcospri, line 74: Generic table macro

lms/templates/admin/macros.jinja
```diff
@@ -63,57 +63,51 @@
...
-    <tr>
-      <td>{{registration.issuer}}</th>
-      <td>{{registration.client_id}}</td>
-      <td>
-        <a class="button" href="{{ request.route_url("admin.registration.id", id_=registration.id) }}">View</a>
-      </td>
-    </tr>
-{% endfor %}
-{% endif %}
+{% macro instances_table(request,  instances) %}
```
marcospri, line 104: Special case usage of the macro for instances and registrations

lms/templates/admin/macros.jinja
```diff
@@ -63,57 +63,51 @@
...
+{% macro created_updated_fields(model) %}
+    {{ disabled_text_field("Created", model.created) }}
+    {% if model.updated %}
+        {{ disabled_text_field("Updated", model.updated) }}
+    {% endif %}
+{% endmacro %}
+
+{% macro object_list_table(request, route, objects, headers, fields) %}
+<div class="container">
+    <div class="table-container">
```
marcospri, line 76: This new `div` allow for the table to overflow

https://bulma.io/documentation/elements/table/#table-container
https://github.com/hypothesis/lms/issues/4410

lms/templates/admin/macros.jinja
```diff
@@ -63,57 +63,51 @@
...
+                <tr>
+                  <td>
+                    <a class="button" href="{{ request.route_url(route, id_=object.id) }}">View</a>
+                  </td>
+                  {% for field in fields %}
+                  <td>{{object[field]}}</th>
+                  {% endfor %}
+               </tr>
+            {% endfor %}
+          </tbody>
```
marcospri, line 97: `tbody` was not closed

jon-betts: The view button on the left is a bit out of step with all the other buttons we have. I can see that when making the page narrower it means the button can still be reached.

I don't think this is quite right yet, as making page narrower has horizontal scroll, but only up to the min width of the container, which doesn't collapse.

So we have full width:

image

Within the stretch of the container:

image

But beyond this it doesn't get any smaller:

image

I don't think we need to fix this in this PR, but we might want to log something about it.

lms/templates/admin/macros.jinja
```diff
@@ -63,57 +63,51 @@
     {{ form_text_field(request, "Token URL", "token_url", lti_registration.token_url) }}
```
jon-betts, line 63: Should this file be `.html.jina2`? It's nuking my highlighting!

lms/templates/admin/macros.jinja
jon-betts: This looks nicer now

lms/templates/admin/macros.jinja
jon-betts: Right, so this allows us to have different headers on each page. Nice.

lms/templates/admin/organization.html.jinja2
```diff
@@ -0,0 +1,38 @@
...
+                    name="enabled" disabled>
+            </label>
+        {% endcall %}
+        {{ macros.created_updated_fields(org) }}
+        <div class="has-text-right mb-6">
+            <input type="submit" class="button is-primary" value="Save" />
+        </div>
+    </form>
+
+    <legend class="label has-text-centered">Organization application instances</legend>
```
jon-betts, line 27:
```suggestion
    <legend class="label has-text-centered">Application instances</legend>
```
Is this implicit from context?

lms/templates/admin/organization.html.jinja2
```diff
@@ -0,0 +1,38 @@
...
+        {% if org.application_instances %}
+            {{ macros.object_list_table(
+                    request, 'admin.instance.id', org.application_instances,
+                    ['Consumer key', 'Deployment ID', 'Tool name', 'URL'],
+                    ['consumer_key', 'deployment_id', 'tool_consumer_instance_name', 'lms_url']
```
jon-betts, line 32: I think `requesters_email` would be very good here, as it's never blank, and contains a domain.

If we can fix the scrolling issues we can probably have them all. If not, try and work out which of `consumer_key` or `deployment_id` is less useful in this context?

lms/services/organization.py
```diff
@@ -15,6 +15,9 @@ class OrganizationService:
     def __init__(self, db_session: Session):
         self._db_session = db_session
 
+    def get_by_id(self, id_) -> Optional[Organization]:
```
jon-betts, line 18:
```suggestion
    def get_by_id(self, id_) -> Optional[Organization]:
        """Get an organization by it's primary key."""
```

lms/templates/admin/macros.jinja
jon-betts: As mentioned elsewhere, I don't think this is working because the container isn't flexible.

jon-betts (approved):
I think this is looking pretty good, but I do have some feedback.

Not for this PR?

 * The horizontal scrolling isn't working right yet, but it's a pre-existing problem
 * Similarly should `macros.jinja` be `macros.html.jinja2`? This helps my highlighting kick in
 
Things about this PR:

 * I think `requesters_email` is very handy, as it's never blank, and shows the domain name
 * Maybe have "Application instances" instead of "Organization application instances"?

Issue #4417, opened 2022-09-08
Add the ability to (re-)allocate an application instance to an organisation


jon-betts:
In order to allow users to arrange application instances into organisations correctly we need to let them change values.

We need to support:

 * [x] Moving an application instance from one organisation to another
 * [ ] Adding an application instance which has no organisation to one

These could be achieved with a text field which accepted and verified an org public id.

marcospri:
Tacking: 

`Adding an application instance which has no organisation to one`

with the option to create new orgs from scratch

PR #4434, opened 2022-09-12
Allow moving AI to different organizations

marcospri:
For https://github.com/hypothesis/lms/issues/4417

this doesn't yet tackle:

- Creating a new organisation from an application instance

# Testing 

- Go to an AI that has an organization
- Save without filling the new field with any value. Nothing will  update
- Save with an invalid/non existing ID. You'll get feedback on the error.
- Save with an existing valid ID. The organization wolud be updated.

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -170,6 +170,7 @@ class ApplicationInstanceService:
         deployment_id=None,
         developer_key=None,
         developer_secret=None,
+        organization_id=None,
     ):
         if lms_url:
             application_instance.lms_url = lms_url
@@ -185,6 +186,8 @@ class ApplicationInstanceService:
             encrypted_secret = self._aes_service.encrypt(aes_iv, developer_secret)
             application_instance.aes_cipher_iv = aes_iv
             application_instance.developer_secret = encrypted_secret
+
+        application_instance.organization_id = organization_id
 
     def create_application_instance(  # pylint:disable=too-many-arguments
         self,
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -11,6 +11,7 @@ from lms.models.region import Region
 from lms.services.aes import AESService
 from lms.services.exceptions import SerializableError
 from lms.services.organization import OrganizationService
+from lms.validation import ValidationError
 
 LOG = getLogger(__name__)
 
@@ -170,7 +171,7 @@ class ApplicationInstanceService:
         deployment_id=None,
         developer_key=None,
         developer_secret=None,
-        organization_id=None,
+        organization_public_id=None,
     ):
         if lms_url:
             application_instance.lms_url = lms_url
@@ -187,7 +188,18 @@ class ApplicationInstanceService:
             application_instance.aes_cipher_iv = aes_iv
             application_instance.developer_secret = encrypted_secret
 
-        application_instance.organization_id = organization_id
+        if organization_public_id:
+            org = self._organization_service.get_by_public_id(organization_public_id)
+            if org:
+                application_instance.organization = org
+            else:
+                raise ValidationError(
+                    messages={
+                        "organization_public_id": [
+                            f"Organization {organization_public_id} not found"
+                        ]
+                    }
+                )
 
     def create_application_instance(  # pylint:disable=too-many-arguments
         self,
--- a/lms/services/public_id.py
+++ b/lms/services/public_id.py
@@ -1,0 +1,50 @@
+from typing import Optional, TypeVar
+
+from lms.models import Region
+from lms.validation import ValidationError
+
+Model = TypeVar("Model")
+
+
+def get_by_public_id(  # pylint:disable=too-many-arguments
+    db, model: Model, public_id: str, region: Region, app="lms", type_="org"
+) -> Optional[Model]:
+    """
+    Get a `model` by their public_id.
+
+    Public IDs have a format like:
+        region.app.type.id
+
+    eg:
+        us.lms.org.gkPHJPSRSHC7YWAYJ7s1LA
+
+    :raises ValidationError: If the given ID doesn't have the right format
+        or doesn't contain the expected constraints.
+    """
+    try:
+        id_region, id_app, id_type, id_public_id = public_id.split(".")
+    except ValueError as err:
+        raise ValidationError(
+            messages={"public_id": [f"{public_id} doesn't have the right format"]}
+        ) from err
+
+    if id_region != region.code:
+        raise ValidationError(
+            messages={
+                "public_id": [
+                    f"{id_region} doesn't match current region: {region.code}"
+                ]
+            }
+        )
+
+    if id_app != app:
+        raise ValidationError(
+            messages={"public_id": [f"{id_app} doesn't match app region: {app}"]}
+        )
+
+    if id_type != type_:
+        raise ValidationError(
+            messages={"public_id": [f"{id_type} doesn't match type: '{type_}'"]}
+        )
+
+    return db.query(model).filter_by(_public_id=id_public_id).one_or_none()
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -38,7 +38,6 @@
     <fieldset class="box mt-6">
         <legend class="label has-text-centered">Organization</legend>
         {{ macros.organization_preview(request, instance.organization) }}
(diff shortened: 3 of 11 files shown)


lms/services/application_instance.py
```diff
@@ -186,6 +187,8 @@ def update_application_instance(  # pylint:disable=too-many-arguments
             application_instance.aes_cipher_iv = aes_iv
             application_instance.developer_secret = encrypted_secret
 
+        application_instance.organization_id = organization_id
```
marcospri, line 190: The service allows setting `None` for organization_id

lms/views/admin/application_instance.py
```diff
@@ -264,12 +271,33 @@ def show_instance(self):
     def update_instance(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
+        # Keep the same org if no public_id is sent
+        organization_id = ai.organization_id
+        if org_public_id := self.request.params.get("org_public_id", "").strip():
+            try:
+                if org := self.organization_service.get_by_public_id(org_public_id):
```
marcospri, line 278: The admin pages don't allow setting `None` for the organization_id.

I rather tackle that as a general update for all optional values.

lms/views/admin/application_instance.py
```diff
@@ -264,12 +271,33 @@ def show_instance(self):
...
+                    location=self.request.route_url("admin.instance.id", id_=ai.id)
+                )
+
         self.application_instance_service.update_application_instance(
             ai,
             lms_url=self.request.params.get("lms_url", "").strip(),
             deployment_id=self.request.params.get("deployment_id", "").strip(),
             developer_key=self.request.params.get("developer_key", "").strip(),
             developer_secret=self.request.params.get("developer_secret", "").strip(),
+            organization_id=organization_id,
```
marcospri, line 300: It could be argued that the AI service itself could take the public_id and deal with finding it itself.

I reckon for the admin pages this makes sense. If we see this being done more often we could look into refactoring it into the service.

lms/views/admin/application_instance.py
jon-betts: Should any of this checking and safety be in the service instead?

jon-betts (commented):
Some general comments rather than a deep review:

 * The application instance page is quite intimidating now
 * I'm not sure if this is a "Danger" action or if it's just a reaction to the above
 * One option is to move this to the org, which has less going on
 * I can't think of a reason to de-allocate an application instance from an org without moving it to a new one, so I think some of the verification happening in the admin pages should be in the service?

I did notice when reviewing this that if the ai summary in the org page gets too wide, there is no horizontal scrolling, so you can't use it at all.

lms/views/admin/application_instance.py
marcospri: Moved to the service

lms/services/application_instance.py
```diff
@@ -186,6 +188,19 @@ def update_application_instance(  # pylint:disable=too-many-arguments
             application_instance.aes_cipher_iv = aes_iv
             application_instance.developer_secret = encrypted_secret
 
+        if organization_public_id:
+            org = self._organization_service.get_by_public_id(organization_public_id)
+            if org:
+                application_instance.organization = org
+            else:
+                raise ValidationError(
```
marcospri, line 196: Now setting an empty ValidationError

lms/services/public_id.py
```diff
@@ -0,0 +1,50 @@
+from typing import Optional, TypeVar
+
+from lms.models import Region
+from lms.validation import ValidationError
+
+Model = TypeVar("Model")
+
+
+def get_by_public_id(  # pylint:disable=too-many-arguments
```
marcospri, line 9: Moved a generic version of this here.

lms/services/application_instance.py
```diff
@@ -186,6 +188,19 @@ def update_application_instance(  # pylint:disable=too-many-arguments
             application_instance.aes_cipher_iv = aes_iv
             application_instance.developer_secret = encrypted_secret
 
+        if organization_public_id:
```
marcospri, line 191: This is now checked on the service and not the view.

lms/services/public_id.py
```diff
@@ -0,0 +1,50 @@
+from typing import Optional, TypeVar
+
+from lms.models import Region
+from lms.validation import ValidationError
+
+Model = TypeVar("Model")
```
marcospri, line 6: I'm drinking all the typing kool aid

lms/templates/admin/instance.html.jinja2
```diff
@@ -38,6 +38,7 @@
     <fieldset class="box mt-6">
         <legend class="label has-text-centered">Organization</legend>
         {{ macros.organization_preview(request, instance.organization) }}
+        {{ macros.form_text_field(request, "Move to organization ID", "org_public_id") }}
```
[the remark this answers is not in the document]
marcospri: I think there are issues around to much white space and general organization but I rathe tackle those as a follow up to all org PRs.

This operation has some risk (maybe more than enable/disable), moved to the danger zone.

lms/services/application_instance.py
```diff
@@ -170,6 +171,7 @@ def update_application_instance(  # pylint:disable=too-many-arguments
         deployment_id=None,
         developer_key=None,
         developer_secret=None,
+        organization_public_id=None,
     ):
```
jon-betts, line 175: Docstring?

lms/services/application_instance.py
```diff
@@ -186,6 +188,19 @@ def update_application_instance(  # pylint:disable=too-many-arguments
...
+            if org:
+                application_instance.organization = org
+            else:
+                raise ValidationError(
+                    messages={
+                        "organization_public_id": [
+                            f"Organization {organization_public_id} not found"
+                        ]
+                    }
+                )
```
jon-betts, line 202: I'm not sure if it's weird to raise a `ValidationError` here? How do we know this came from a field called `organization_public_id` in this context?

lms/services/public_id.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def get_by_public_id(  # pylint:disable=too-many-arguments
+    db, model: Model, public_id: str, region: Region, app="lms", type_="org"
```
jon-betts, line 10:
```suggestion
    db, model: Model, public_id: str, region: Region,  type_, app="lms"
```
If this is intended to be generic, it probably shouldn't specify type

lms/services/public_id.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if id_app != app:
+        raise ValidationError(
+            messages={"public_id": [f"{id_app} doesn't match app region: {app}"]}
+        )
+
+    if id_type != type_:
+        raise ValidationError(
+            messages={"public_id": [f"{id_type} doesn't match type: '{type_}'"]}
+        )
```
jon-betts, line 48: Again, it seems a bit sus to be raising `ValidationError` outside of a schema / view context

lms/views/admin/application_instance.py
```diff
@@ -264,13 +264,24 @@ def show_instance(self):
...
+                lms_url=self.request.params.get("lms_url", "").strip(),
+                deployment_id=self.request.params.get("deployment_id", "").strip(),
+                developer_key=self.request.params.get("developer_key", "").strip(),
+                developer_secret=self.request.params.get(
+                    "developer_secret", ""
+                ).strip(),
+                organization_public_id=self.request.params.get(
+                    "org_public_id", ""
+                ).strip(),
+            )
```
jon-betts, line 279: Note to self: It looks like we have all the handling to be able to update the org settings at once, but we actually use it two separate ways.

One for normal fields, and another in the _danger zone_ which only sends the single field `org_public_id`.

jon-betts (approved):
I've checked this out and I think it works well.

Having this in the danger zone makes sense to be, but it does make it awkward. Moving loads of these at once would be a chore. But hopefully it's not something we'll do too often.

Also we can't support much in the way of rich interaction with server side pages alone.

## Some rando comments

 * I dislike the "View" button on the left, it feels very un-natural, but works around our horizontal scrolling problems
 * We've talked about different ways of wrapping up some of the re-used stuff here, and I think I'd prefer to see non-SQLA based parsing. Then it could be used in different contexts, but that can be tackled elsewhere.
 * I'm pretty sure raising `ValidationError`s is kind of cheeky - but if we tackle the work above we'd have to deal with that there, so it's probably fine for now

PR #4435, opened 2022-09-12
Audit changes made to `organization.enabled`

marcospri:
The audit mechanism is more general than strictly needed for keeping changes for `enabled`.

My plan is to apply this to other type of objects in the admin pages. Ideally with one liner PRs after some testing.

## Testing

- Go over the admin pages, to one particular organization 
- Toggle the enabled status at the bottom
- Check the audit trail on the database with a query like:

```
select * from event 
join event_data on event.id = event_data.event_id 
join event_type on event.type_id = event_type.id
where type ='audit';
```

### DIFF
--- a/lms/events/event.py
+++ b/lms/events/event.py
@@ -1,6 +1,6 @@
 import json
-from dataclasses import dataclass, field, fields
-from typing import List, Optional
+from dataclasses import asdict, dataclass, field, fields
+from typing import Dict, List, Optional
 
 from pyramid.request import Request
 from sqlalchemy import inspect
@@ -107,16 +107,39 @@ class LTIEvent(BaseEvent):
 
 @dataclass
 class AuditTrailEvent(BaseEvent):
+    @dataclass
+    class ModelChange:
+        action: str
+        model: str
+        id: int  # pylint: disable=invalid-name
+        source: str
+        userid: str
+        changes: Dict[str, tuple]
+
+        @classmethod
+        def from_instance(cls, instance, **kwargs):
+            changes = {}
+            instance_details = inspect(instance)
+            for attr in instance_details.attrs:
+                history = instance_details.get_history(attr.key, True)
+
+                if not history.has_changes():
+                    continue
+
+                changes[attr.key] = (
+                    history.deleted[0] if history.deleted else None,
+                    history.added[0] if history.added else None,
+                )
+
+            return cls(
+                model=instance.__class__.__name__,
+                id=instance.id,
+                changes=changes,
+                **kwargs,
+            )
+
     type: EventType.Type = EventType.Type.AUDIT_TRAIL
-
-    instance: BASE = None
-    """Object for which we are tracking changes"""
-
-    action: str = None
-    """What happen to the object: crated,updated,deleted..."""
-
-    source: str = None
-    """In which context the change happen"""
+    change: ModelChange = None
 
     def _get_data(self):
         """
@@ -124,30 +147,7 @@ class AuditTrailEvent(BaseEvent):
 
         This is called on BaseEvent.__post_init__
         """
-        changes = {}
-        instance_details = inspect(self.instance)
-        for attr in instance_details.attrs:
-            history = instance_details.get_history(attr.key, True)
-
-            if not history.has_changes():
-                continue
-
-            changes[attr.key] = (
-                history.deleted[0] if history.deleted else None,
-                history.added[0] if history.added else None,
-            )
-
-        return {
-            "model": self.instance.__class__.__name__,
-            "id": self.instance.id,
-            "action": self.action,
-            "source": self.source,
-            # LTI users will have a FK to the User table.
-            # userid is useful for other authentication methods.
-            # For example this will be the user's email while using google oauth.
-            "userid": self.request.identity.userid,
-            "changes": changes,
-        }
+        return asdict(self.change)
 
     @staticmethod
     def notify(request: Request, instance: BASE, source="admin_pages"):
@@ -155,8 +155,14 @@ class AuditTrailEvent(BaseEvent):
             request.registry.notify(
                 AuditTrailEvent(
                     request=request,
-                    instance=instance,
-                    action="update",
-                    source=source,
+                    change=AuditTrailEvent.ModelChange.from_instance(
+                        instance,
+                        action="update",
+                        source=source,
+                        # LTI users will have a FK to the User table.
+                        # userid is useful for other authentication methods.
+                        # For example this will be the user's email while using google oauth.
+                        userid=request.identity.userid,
+                    ),
                 )
             )
--- a/lms/events/event.py
+++ b/lms/events/event.py
@@ -3,7 +3,9 @@ from dataclasses import dataclass, field, fields
 from typing import List, Optional
 
 from pyramid.request import Request
+from sqlalchemy import inspect
 
+from lms.db import BASE
 from lms.models import EventType
 from lms.services.lti_role_service import LTIRoleService
 
@@ -11,6 +13,9 @@ from lms.services.lti_role_service import LTIRoleService
 @dataclass
 class BaseEvent:  # pylint:disable=too-many-instance-attributes
     """Base class for generic events."""
+
+    Type = EventType.Type
+    """Expose the type here for the callers convenience"""
 
     request: Request
     """Reference to the current request"""
@@ -52,9 +57,6 @@ class LTIEvent(BaseEvent):
     """
 
     # pylint:disable=no-member
-
-    Type = EventType.Type
-    """Expose the type here for the callers convenience"""
 
     def _get_user_id(self):
         return self.request.user.id
@@ -101,3 +103,60 @@ class LTIEvent(BaseEvent):
             return assignment.id
 
         return None
+
+
+@dataclass
+class AuditTrailEvent(BaseEvent):
+    type: EventType.Type = EventType.Type.AUDIT_TRAIL
+
+    instance: BASE = None
+    """Object for which we are tracking changes"""
+
+    action: str = None
+    """What happen to the object: crated,updated,deleted..."""
+
+    source: str = None
+    """In which context the change happen"""
+
+    def _get_data(self):
+        """
+        Fill the event's data value.
+
+        This is called on BaseEvent.__post_init__
+        """
+        changes = {}
+        instance_details = inspect(self.instance)
+        for attr in instance_details.attrs:
+            history = instance_details.get_history(attr.key, True)
+
+            if not history.has_changes():
+                continue
+
+            changes[attr.key] = (
+                history.deleted[0] if history.deleted else None,
+                history.added[0] if history.added else None,
+            )
+
+        return {
+            "model": self.instance.__class__.__name__,
+            "id": self.instance.id,
+            "action": self.action,
+            "source": self.source,
+            # LTI users will have a FK to the User table.
+            # userid is useful for other authentication methods.
+            # For example this will be the user's email while using google oauth.
+            "userid": self.request.identity.userid,
+            "changes": changes,
+        }
+
+    @staticmethod
+    def notify(request: Request, instance: BASE, source="admin_pages"):
+        if request.db.is_modified(instance):
+            request.registry.notify(
+                AuditTrailEvent(
+                    request=request,
+                    instance=instance,
+                    action="update",
+                    source=source,
+                )
+            )
--- a/lms/models/event.py
+++ b/lms/models/event.py
@@ -13,6 +13,7 @@ class EventType(BASE):
     class Type(str, Enum):
         CONFIGURED_LAUNCH = "configured_launch"
         DEEP_LINKING = "deep_linking"
+        AUDIT_TRAIL = "audit"
 
     id = sa.Column(sa.Integer, autoincrement=True, primary_key=True)
     type = varchar_enum(Type)
--- a/lms/views/admin/organization.py
+++ b/lms/views/admin/organization.py
@@ -3,6 +3,7 @@ from pyramid.httpexceptions import HTTPFound, HTTPNotFound
 from pyramid.view import view_config, view_defaults
 from webargs import fields
 
+from lms.events import AuditTrailEvent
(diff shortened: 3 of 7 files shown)


lms/db/__init__.py
```diff
@@ -85,7 +85,7 @@ def __repr__(self):
 )
 
 
-def init(engine, drop=False, stamp=True, data=True):  # pragma: nocover
```
marcospri, line 88: No need for this if we autocreate events when we see them.

lms/events/event.py
```diff
@@ -12,6 +13,9 @@
 class BaseEvent:  # pylint:disable=too-many-instance-attributes
     """Base class for generic events."""
 
+    Type = EventType.Type
```
marcospri, line 16: Expose this on the base class so it's available for all subclasses.

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
             return assignment.id
 
         return None
+
+
+class AdminChangeEvent(BaseEvent):
```
marcospri, line 107: New event type.

It's on the generic side but making it only for `OrganizationEnabledChangedEvent` would be very similar code overall.

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+    def __init__(self, instance, *args, **kwargs):
+        self._instance = instance
+        super().__init__(*args, **kwargs)
+
+    def _get_data(self):
+        changes = {}
```
marcospri, line 113: TODO this is a bit opaque. Add some comments here.

lms/services/event.py
```diff
@@ -41,7 +41,13 @@ def insert_event(self, event: BaseEvent):
     @lru_cache(maxsize=10)
     def _get_type_pk(self, type_: EventType.Type) -> int:
         """Cache the PK of the event_type table to avoid an extra query while inserting events."""
-        return self._db.query(EventType).filter_by(type=type_).one().id
+        event_type = self._db.query(EventType).filter_by(type=type_).one_or_none()
```
marcospri, line 44: Creating new event types as we see them.

The alternative is doing a migration but I reckon as this is all based on the `Type` enum it's fine for python to control the DB state as well.

lms/services/organization.py
```diff
@@ -73,6 +73,19 @@ def get_by_public_id(self, public_id: str):
             .one_or_none()
         )
 
+    def update_organization(self, organization: Organization, **kwargs):
```
marcospri, line 76: Update the org thru this method instead of using SQLA outside.

lms/views/admin/organization.py
```diff
@@ -31,8 +32,13 @@ def show_organization(self):
...
         org = self._get_org_or_404(self.request.matchdict["public_id"])
 
-        org.name = self.request.params.get("name", "").strip() or None
+        self.organization_service.update_organization(
+            org,
+            name=self.request.params.get("name", "").strip(),
+            enabled=self.request.params.get("enabled", "") == "on",
+        )
 
+        AdminChangeEvent.notify_changes(self.request, org)
```
marcospri, line 41: `maybe_notify_changes`? I think the name is it's fine as it is. 

The method will not create the event if no changes are found.

lms/events/event.py
jon-betts: I remember, this works because all `_get_<field>` methods cause something to happen right?

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+        instance_details = inspect(self._instance)
+        for attr in instance_details.attrs:
+            history = instance_details.get_history(attr.key, True)
```
jon-betts, line 116: I don't know what this magic is, but it looks cool!

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+    @staticmethod
+    def notify_changes(request, instance):
+        if request.db.is_modified(instance):
+            request.registry.notify(
+                AdminChangeEvent(
+                    instance=instance,
+                    request=request,
+                    type=AdminChangeEvent.Type.ADMIN_PAGES_UPDATE,
+                )
+            )
```
jon-betts, line 134: I think it's worth thinking about whether we can work this out another way. I like being able to do:

```python
request.registry.notify(
    AdminChangeEvent(
        instance=instance,
        request=request,
        type=AdminChangeEvent.Type.ADMIN_PAGES_UPDATE,
    )
)
```

In the code without having to worry about it

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+        return {"user": self.request.identity.userid, "changes": changes}
+
+    @staticmethod
+    def notify_changes(request, instance):
+        if request.db.is_modified(instance):
+            request.registry.notify(
+                AdminChangeEvent(
+                    instance=instance,
+                    request=request,
+                    type=AdminChangeEvent.Type.ADMIN_PAGES_UPDATE,
```
jon-betts, line 132: I think the type is implicit from raising an `AdminChangeEvent`? Or is the idea that this is like my idea for the separation between the audit event and the source?

jon-betts (commented):
I like the auto storing of ids rather than setting it in the `db/__init__.py`. That stuck me as fiddly and far away. This sort of distance between parts of the same system is one of our big maintenance problems I think.

On the event itself, I think we are missing some crucial pieces of information we will need to make any sense of it:

 * The object type being modified
 * The id of the type being modified

After that, there's a question of how many other types of operations you want to be able to support:

 * Storing before, as well as after data, would make for a nice visual comparison, and also allow you to "undo" certain actions
 * An "action" like "create / update / delete" would let you support all crud operations, not just update
 * A "source" like "admin_pages" would open the door to changes from elsewhere, like user settings changes

There are various other thoughts below of a more stylistic nature.

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+    def _get_data(self):
+        changes = {}
+        instance_details = inspect(self._instance)
+        for attr in instance_details.attrs:
+            history = instance_details.get_history(attr.key, True)
+
+            if not history.has_changes():
+                continue
+
+            changes[attr.key] = (1, history.added[0])
```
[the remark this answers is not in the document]
marcospri: Went for `"field": [old_value, new_value]`

lms/events/event.py
```diff
@@ -101,3 +102,33 @@ def _get_assignment_id(self):
...
+
+        return {"user": self.request.identity.userid, "changes": changes}
```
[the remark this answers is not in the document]
marcospri: Added a comment around this. Its `userid` and that depends on the auth method.

For LTIUser this is irrelevant because the event itself (rather than event_data) would relate to one "User".

For the  immediate use case of this `userid` would be the email used in google's oauth.

lms/models/event.py
```diff
@@ -13,6 +13,7 @@ class EventType(BASE):
     class Type(str, Enum):
         CONFIGURED_LAUNCH = "configured_launch"
         DEEP_LINKING = "deep_linking"
+        ADMIN_PAGES_UPDATE = "admin_update"
```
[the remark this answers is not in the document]
marcospri: Went pretty much for that structure except:

` "type": "audit_trail",` which is already part of the main row in `event` (this json goes on a separate `event_data` table).

`source` and `action` are mostly hard-coded for now but makes sense to have them in the json already for when we have different values.

lms/events/event.py
marcospri: This allows to `notify` without having to worry if there are changes or not (when we don't want a new row in the DB.

Later it can be extended to handle new objects, deletions etc with the same `notify(object)`

lms/events/event.py
```diff
@@ -101,3 +103,60 @@ def _get_assignment_id(self):
...
+            "source": self.source,
+            # LTI users will have a FK to the User table.
+            # userid is useful for other authentication methods.
+            # For example this will be the user's email while using google oauth.
+            "userid": self.request.identity.userid,
+            "changes": changes,
+        }
+
+    @staticmethod
+    def notify(request: Request, instance: BASE, source="admin_pages"):
```
jon-betts, line 153:
```suggestion
    def notify(request: Request, instance: BASE, action, source):
```

Should we be accepting the `action` here? Then this wouldn't need to change the tests or defaults etc on new things?

lms/events/event.py
jon-betts: I was thinking we could have a "should_notify: bool" which could be set from `request.db.is_modified(instance)`. Then the subscriber would just "Nope" out if it's false.

But this is fine. There's no need to have a "system" yet (despite everything else we are doing!) as there's only one caller.

lms/events/event.py
```diff
@@ -101,3 +103,60 @@ def _get_assignment_id(self):
...
 
         return None
+
+
+@dataclass
+class AuditTrailEvent(BaseEvent):
+    type: EventType.Type = EventType.Type.AUDIT_TRAIL
+
+    instance: BASE = None
+    """Object for which we are tracking changes"""
```
jon-betts, line 113: I don't think the instance is safe in general in events as we can't handle these out of band of the transaction without weird things happening. Keeping `model` and `id` would probably be safer?

tests/unit/lms/events/test_event.py
```diff
@@ -121,3 +123,51 @@ def pyramid_request(self, pyramid_request):
...
+            source=sentinel.source,
+        )
+
+        assert event.data == {
+            "model": "Organization",
+            "id": Any.int(),
+            "userid": pyramid_request.identity.userid,
+            "source": sentinel.source,
+            "action": sentinel.action,
+            "changes": {"name": ("OLD_NAME", "NEW_NAME"), "enabled": (True, False)},
```
jon-betts, line 148: :+1:

lms/events/event.py
```diff
@@ -101,3 +103,60 @@ def _get_assignment_id(self):
...
+            "model": self.instance.__class__.__name__,
+            "id": self.instance.id,
+            "action": self.action,
+            "source": self.source,
+            # LTI users will have a FK to the User table.
+            # userid is useful for other authentication methods.
+            # For example this will be the user's email while using google oauth.
+            "userid": self.request.identity.userid,
+            "changes": changes,
+        }
```
jon-betts, line 150: I'm not suggesting this as a change here, as it's quite disruptive, but I think we might be fighting with two levels of nesting being flat here. It looks like we want a lot of attributes to be inside the data, but have attached them to the parent object.

If we have a separate object just for this we can make use of `asdict` on the object. That sub-object might look like this:

```python
@dataclass
class ModelChange:
    action: str
    model: str
    id: str
    source: str
    userid: str
    changes: Dict[str, tuple]

    @classmethod
    def from_instance(cls, instance, **kwargs):
        changes = {}
        instance_details = inspect(instance)
        for attr in instance_details.attrs:
            history = instance_details.get_history(attr.key, True)

            if not history.has_changes():
                continue

            changes[attr.key] = (
                history.deleted[0] if history.deleted else None,
                history.added[0] if history.added else None,
            )

        return cls(
            id=instance.id, model=instance.__class__.name, changes=changes, **kwargs
        )
```

This would save us the job of making the dict ourself and would clean up the parent object a lot:

```python
@dataclass
class AuditTrailEvent(BaseEvent):
    type: EventType.Type = EventType.Type.AUDIT_TRAIL

    change_data: ModelChange = None

    def _get_data(self):
        return asdict(self.change_data)
```

jon-betts (commented):
It's a bit late and my head is fried, so I'll leave this as a comment rather than a proper review.

This looks good to me, but I wonder if we are tripping over ourselves with the structure of the event object. I feel like we are fighting the dataclass rather than getting a benefit from it.

I wonder if there's a way around?

lms/events/event.py
marcospri: The way things are strucutred now notify decides the action. The idea is you could for example do:

```
obj = get_or_create_whatever()
...
AuditTrail.notify(obj)
```
And notify would pick the right `action`

lms/events/event.py
marcospri: I think this will also require the `"should_notify: bool"`

and then emit the event with `request.notify(AuditTrailEvent(request, ModelChange.from_instance(obj))`

I don't see any immediate benefits

tests/unit/lms/events/test_event.py
```diff
@@ -121,3 +123,72 @@ def pyramid_request(self, pyramid_request):
...
+        org = factories.Organization(name="OLD_NAME", enabled=True)
+        db_session.flush()
+        org.name = "NEW_NAME"
+        org.enabled = False
+
+        assert AuditTrailEvent.ModelChange.from_instance(
+            org, action=sentinel.action, source=sentinel.source, userid=sentinel.userid
+        ) == AuditTrailEvent.ModelChange(
+            model="Organization",
+            id=Any.int(),
```
jon-betts, line 139:
```suggestion
            id=org.id,
```

tests/unit/lms/events/test_event.py
```diff
@@ -121,3 +123,72 @@ def pyramid_request(self, pyramid_request):
...
+        event = AuditTrailEvent(request=pyramid_request, change=change)
+
+        assert event.data == {
+            "id": sentinel.id,
+            "model": sentinel.model,
+            "source": sentinel.source,
+            "action": sentinel.action,
+            "userid": sentinel.userid,
+            "changes": sentinel.changes,
+        }
```
jon-betts, line 165: How about something like this?

```suggestion
        kwargs = {
            "id": sentinel.id,
            "model": sentinel.model,
            "source": sentinel.source,
            "action": sentinel.action,
            "userid": sentinel.userid,
            "changes": sentinel.changes,
        }

        event = AuditTrailEvent(
            request=pyramid_request, change=AuditTrailEvent.ModelChange(**kwargs)
        )

        assert event.data == kwargs
```

jon-betts (approved):
Yep this is looking good.

It only triggers on updates to the enabled at the moment (not the name etc).

PR #4439, opened 2022-09-13
Fix deeplinking requests missing `deep_linking_settings`

marcospri:
For https://github.com/hypothesis/lms/issues/4438

It looks like most LMS don't send a `data` field but there at least a few instances where the whole `deep_linking_settings` is not present.

### DIFF
--- a/lms/views/lti/deep_linking.py
+++ b/lms/views/lti/deep_linking.py
@@ -136,7 +136,8 @@ class DeepLinkingFieldsViews:
         #
         # This claim is required if present in the `LtiDeepLinkingRequest`
         # message.
-        if data := self.request.parsed_params["deep_linking_settings"].get("data"):
+        settings = self.request.parsed_params["deep_linking_settings"] or {}
+        if data := settings.get("data"):
             message["https://purl.imsglobal.org/spec/lti-dl/claim/data"] = data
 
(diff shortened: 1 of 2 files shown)


lms/views/lti/deep_linking.py
```diff
@@ -136,7 +136,8 @@ def file_picker_to_form_fields_v13(self):
         #
         # This claim is required if present in the `LtiDeepLinkingRequest`
         # message.
-        if data := self.request.parsed_params["deep_linking_settings"].get("data"):
+        settings = self.request.parsed_params["deep_linking_settings"] or {}
```
marcospri, line 139: `deep_linking_settings` is present but the value is None in the exceptions in sentry.

jon-betts: Another option here is a double get play:

```python
self.request.parsed_params.get("deep_linking_settings", {}).get("data")
```

jon-betts (approved):
Yep, this looks good.

lms/views/lti/deep_linking.py
marcospri: I tried that first but I always forget `get` doesn't do a falsy test, just presence. So `{"deep_linking_settings":  None}.get` fails.

PR #4444, opened 2022-09-13
Fix result retrieval for cases for `None` values

marcospri:
For:  https://github.com/hypothesis/lms/issues/4443 
Fixes: https://sentry.io/organizations/hypothesis/issues/3486870164/?project=259908

Until now we only seen LMS not sending a value at all when there's no grade but some LMS seem to send a `None` value instead.

https://hypothes-is.slack.com/archives/C2BLQDKHA/p1663045160950229

### DIFF
--- a/lms/services/lti_grading/_v13.py
+++ b/lms/services/lti_grading/_v13.py
@@ -39,7 +39,7 @@ class LTI13GradingService(LTIGradingService):
 
         try:
             return results[-1]["resultScore"] / results[-1]["resultMaximum"]
-        except (ZeroDivisionError, KeyError, IndexError):
+        except (TypeError, ZeroDivisionError, KeyError, IndexError):
             return None
 
     def record_result(self, grading_id, score=None, pre_record_hook=None):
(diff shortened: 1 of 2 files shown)


lms/services/lti_grading/_v13.py
```diff
@@ -39,7 +39,8 @@ def read_result(self, grading_id):
 
         try:
             return results[-1]["resultScore"] / results[-1]["resultMaximum"]
-        except (ZeroDivisionError, KeyError, IndexError):
```
marcospri, line 42: Added TypeError for the `None / 100` case.

lms/services/lti_grading/_v13.py
```diff
@@ -39,7 +39,8 @@ def read_result(self, grading_id):
...
+        except (TypeError, ZeroDivisionError, KeyError, IndexError) as exc:
+            print(type(exc))
```
jon-betts, line 43:
```suggestion
```

Debug line?

jon-betts (approved):
This looks good, but I think there's a debug print line left in?

PR #4446, opened 2022-09-15
Insert new event types when first seen.

marcospri:
Instead of keeping them in sync with migrations and other methods for testing.

### DIFF
--- a/tests/unit/lms/services/event_test.py
+++ b/tests/unit/lms/services/event_test.py
@@ -60,6 +60,27 @@ class TestEventService:
         )
         assert not db_session.query(EventData).one_or_none()
 
+    def test_insert_event_type(self, svc, db_session):
+        type_query = db_session.query(EventType).filter_by(
+            type=EventType.Type.CONFIGURED_LAUNCH
+        )
+
+        # Insert the event type for the first time
+        svc.insert_event(
+            BaseEvent(request=sentinel.request, type=EventType.Type.CONFIGURED_LAUNCH)
+        )
+
+        # It will create a new type
+        event_type = type_query.one()
+        # Clear the @lru_cache of the private method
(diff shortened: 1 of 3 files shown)


tests/unit/lms/services/event_test.py
```diff
@@ -60,6 +60,26 @@ def test_insert_event_no_data(self, svc, db_session):
...
+        )
+        # It will create a new type
+        event_type = type_query.one()
+        # Clear the @lru_cache of the private method
+        svc._get_type_pk.cache_clear()  # pylint:disable=protected-access
+        # Insert the event type again
+        svc.insert_event(
+            BaseEvent(request=sentinel.request, type=EventType.Type.CONFIGURED_LAUNCH)
+        )
+        # It will create a new type
```
jon-betts, line 80: This comment doesn't seem quite right? Are we checking we get the same one back?

The arrange, act, assert here is a bit unclear

tests/unit/lms/services/event_test.py
```diff
@@ -60,6 +60,26 @@ def test_insert_event_no_data(self, svc, db_session):
...
+        assert event_type == type_query.one()
```
jon-betts, line 81: I'm not clear on what running a query twice does? When does a cursor get bound? Only when calling `one()` etc?

jon-betts (approved):
I think this is sensible and keeps the management of this all in one place.

tests/unit/lms/services/event_test.py
marcospri: Updated the comment and added a bit of white space.

tests/unit/lms/services/event_test.py
marcospri: It just runs it twice, no more magic involve.

Issue #4340, opened 2022-08-16
Add tracking user ID and user agent to JSTOR PDF URL requests


robertknight:
JSTOR have requested that we add a tracking user ID to requests to their API for usage accounting purposes.

_Update 2022-09-07: JSTOR have informed us that forwarding the User-Agent to them is more important than we initially thought._

When a user accesses JSTOR PDFs directly via their website, they would normally do this with a cookie. In our application however, we are fetching the content on behalf of the user, so they have no ability to set a cookie.

Some notes:

- The user ID is mainly needed for when we fetch a PDF URL via the `/pdf/{doi}` endpoint in the API. See Slack thread for confirmation of which API calls need to include this
- The user ID format they are used to dealing with is a UUID, although they could convert it to another format themselves if needed
- The longevity and origin of the user ID do not need to be strongly guaranteed, since it is a substitute for a cookie which could be cleared at any time
- The location of the user ID in API requests is TBD, I suggested a `Tracking-UserID` header

See discussion at https://hypothes-is.slack.com/archives/C02T75RKBTK/p1660577524314679.

robertknight:
See https://hypothes-is.slack.com/archives/C4K6M7P5E/p1661267330080879 for discussion around the details of how we'll generate a user ID. Current thought is `hash(LMS guid + LMS user ID)`.

robertknight:
A related request from JSTOR is to also forward the `User-Agent` from the browser, so they can do some user segmentation (eg. by device) on their side. From https://hypothes-is.slack.com/archives/C02T75RKBTK/p1660657856860449?thread_ts=1660577524.314679&cid=C02T75RKBTK:

> See Also, and if it's easy, passing the User-Agent would be useful as our logging/reporting tools are able to do some level of user segmentation reporting..

rsnyder:
I've come to understand that the User Agent is an important data item used in COUNTER usage reporting/aggregation.  In light of this I'd put the passing of a User Agent header in the request as a "must have" rather than "nice to have" item.

PR #4447, opened 2022-09-15
Add user ID and browser ID info to JSTOR API requests

robertknight:
Add a unique user ID and user agent (browser) information to JSTOR API requests as `Tracking-User-ID` and `Tracking-User-Agent` headers respectively. These are needed by JSTOR for usage counting purposes. See https://github.com/hypothesis/lms/issues/4340 for background.

Example of the headers that are now sent with each API request:

```py
{'Authorization': 'Bearer ...', 'Tracking-User-ID': '3a022b6c146dfd9df4ea8662178eac', 'Tracking-User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36'
```

Fixes https://github.com/hypothesis/lms/issues/4340

### DIFF
--- a/lms/services/jstor/service.py
+++ b/lms/services/jstor/service.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from typing import Optional
 from urllib.parse import quote
 
 import requests
@@ -22,7 +23,16 @@ class JSTORService:
     """Used when no DOI prefix can be found."""
 
     # pylint: disable=too-many-arguments
-    def __init__(self, api_url, secret, enabled, site_code, http_service: HTTPService):
+    def __init__(
+        self,
+        api_url,
+        secret,
+        enabled,
+        site_code,
+        http_service: HTTPService,
+        tracking_user_id: Optional[str],
+        tracking_user_agent: Optional[str],
+    ):
         """
         Initialise the JSTOR service.
 
@@ -31,6 +41,8 @@ class JSTORService:
         :param enabled: Whether JSTOR is enabled on this instance
         :param site_code: The site code to use to identify the organization
         :param http_service: HTTPService instance
+        :param tracking_user_id: Unique user ID, passed to JSTOR for analytics purposes
+        :param tracking_user_agent: User-Agent header from user's browser, passed to JSTOR for analytics purposes
         """
         self._api_url = api_url
         self._secret = secret
@@ -38,6 +50,8 @@ class JSTORService:
         self._site_code = site_code
 
         self._http = http_service
+        self._tracking_user_id = tracking_user_id
+        self._tracking_user_agent = tracking_user_agent
 
     @property
     def enabled(self) -> bool:
@@ -169,6 +183,15 @@ class JSTORService:
             secret=self._secret,
             lifetime=timedelta(hours=1),
         )
+        headers = {"Authorization": f"Bearer {token}"}
+
+        if self._tracking_user_id:
+            headers["Tracking-User-ID"] = self._tracking_user_id
+        if self._tracking_user_agent:
+            headers["Tracking-User-Agent"] = self._tracking_user_agent
+
         return self._http.get(
-            url=url, headers={"Authorization": f"Bearer {token}"}, params=params
+            url=url,
+            headers=headers,
+            params=params,
         )
--- a/lms/services/jstor/service.py
+++ b/lms/services/jstor/service.py
@@ -1,5 +1,4 @@
 from datetime import timedelta
-from typing import Optional
 from urllib.parse import quote
 
 import requests
@@ -23,16 +22,7 @@ class JSTORService:
     """Used when no DOI prefix can be found."""
 
     # pylint: disable=too-many-arguments
-    def __init__(
-        self,
-        api_url,
-        secret,
-        enabled,
-        site_code,
-        http_service: HTTPService,
-        tracking_user_id: Optional[str],
-        tracking_user_agent: Optional[str],
-    ):
+    def __init__(self, api_url, secret, enabled, site_code, headers=None):
         """
         Initialise the JSTOR service.
 
@@ -40,18 +30,16 @@ class JSTORService:
         :param secret: Secret for authenticating with JSTOR
         :param enabled: Whether JSTOR is enabled on this instance
         :param site_code: The site code to use to identify the organization
-        :param http_service: HTTPService instance
-        :param tracking_user_id: Unique user ID, passed to JSTOR for analytics purposes
-        :param tracking_user_agent: User-Agent header from user's browser, passed to JSTOR for analytics purposes
+        :param headers: Additional headers to pass onto JSTOR when making
+            requests
         """
         self._api_url = api_url
         self._secret = secret
         self._enabled = enabled
         self._site_code = site_code
 
-        self._http = http_service
-        self._tracking_user_id = tracking_user_id
-        self._tracking_user_agent = tracking_user_agent
+        self._http = HTTPService()
+        self._http.session.headers = headers
 
     @property
     def enabled(self) -> bool:
@@ -183,15 +171,6 @@ class JSTORService:
             secret=self._secret,
             lifetime=timedelta(hours=1),
         )
-        headers = {"Authorization": f"Bearer {token}"}
-
-        if self._tracking_user_id:
-            headers["Tracking-User-ID"] = self._tracking_user_id
-        if self._tracking_user_agent:
-            headers["Tracking-User-Agent"] = self._tracking_user_agent
-
         return self._http.get(
-            url=url,
-            headers=headers,
-            params=params,
+            url=url, headers={"Authorization": f"Bearer {token}"}, params=params
         )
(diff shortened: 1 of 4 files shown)


tests/unit/lms/services/jstor/factory_test.py
```diff
@@ -35,9 +23,54 @@ def test_it(
...
             http_service=http_service,
+            tracking_user_agent="Chrome 100",
+            tracking_user_id=pyramid_request.lti_user.h_user.username,
         )
         assert svc == JSTORService.return_value
 
+    def test_it_omits_tracking_info_if_not_available(
+        self, pyramid_request, http_service, JSTORService
+    ):
+        pyramid_request.registry.settings.update(
```
robertknight, line 34: I tried to put this statement in the shared `pyramid_request` fixture but ran into an issue where the `settings` object was not available at that time:

```
    @pytest.fixture(autouse=True)
>       pyramid_request.registry.settings.update(
            {
                "jstor_api_url": sentinel.jstor_api_url,
                "jstor_api_secret": sentinel.jstor_api_secret,
            }
        )
E       AttributeError: 'NoneType' object has no attribute 'update'
```

I guess this may be an issue with the order in which overridden fixtures are run. Any suggestions on how best to fix it?

tests/unit/lms/services/jstor/factory_test.py
```diff
@@ -7,9 +7,44 @@
...
+            secret=sentinel.jstor_api_secret,
+            enabled=sentinel.jstor_enabled,
+            site_code=sentinel.jstor_site_code,
+            http_service=http_service,
+            tracking_user_agent="Chrome 100",
+            tracking_user_id=pyramid_request.lti_user.h_user.username,
+        )
+        assert svc == JSTORService.return_value
+
+    def test_it_omits_tracking_info_if_not_available(
```
robertknight, line 24: The main `test_it` test has tracking info enabled because that's the normal case. This is for the exceptional scenario that the JSTOR service is used in some context where there is no LTI user or User-Agent header.

lms/services/jstor/service.py
```diff
@@ -22,7 +23,16 @@ class JSTORService:
...
-    def __init__(self, api_url, secret, enabled, site_code, http_service: HTTPService):
+    def __init__(
+        self,
+        api_url,
+        secret,
+        enabled,
+        site_code,
+        http_service: HTTPService,
+        tracking_user_id: Optional[str],
+        tracking_user_agent: Optional[str],
```
jon-betts, line 34: If these are optional, maybe they should be defaulted to `None`?

jon-betts: The number of arguments here is getting a bit intense. I have a suggestion for cutting this down which is to use a similar approach to the `VitalSourceClient`. We could accept a dict of additional headers to pass to JSTOR without having each one specified as a new variable. This might look something like this:

```python
def service_factory(_context, request):
    ...
    return JSTORService(
        ...,
        headers={
            "Tracking-User-ID": request.lti_user.h_user.username
            if request.lti_user
            else None,
            "Tracking-User-Agent": request.headers.get("User-Agent", None),
        }
    )

...

class JSTORService:
    def __init__(self, api_url, secret, enabled, site_code, headers):
        ...
        self._http = HTTPService()
        self._http.session.headers = headers
```

By setting these headers in the session, they will be sent on every request. This would mean less code was required overall, and we keep he key and value of the headers next to each other rather than separated between two files.

This will require a change in the tests as we aren't passing the session any more.

jon-betts (commented):
The number of arguments is giving me the non-specific heebie-jeebies with this service. I think I'd also prefer to see the keys and values we send together in one place, as piecing together what we send requires you to look across two files to work out what header names are associated with which.

I've made a suggestion for how we could fix this by borrowing the approach in `VitalSourceClient` to set the headers in the session.

If we want to go this way it will change the PR significantly so I won't review the rest just now.

lms/services/jstor/service.py
robertknight: > If these are optional, maybe they should be defaulted to None?

Makes sense.

> The number of arguments here is getting a bit intense.

I see having many optional arguments as mainly a problem when they are all positional, leading to readability/refactoring hazards if some call site passes them positionally. Keyword-only args is one way to avoid that hazard, and similar to what one would do in other languages with an options hash/dict/struct at the end.

> We could accept a dict of additional headers to pass to JSTOR without having each one specified as a new variable

This is fine, as long as the logic remains strictly pass-through (eg. we don't decide to, say, limit tracking info to certain API calls) and the mechanism for passing them to the API doesn't change (eg. from headers to query params). I suspect that both these conditions will hold in this case.

> By setting these headers in the session, they will be sent on every request. 

Is there a downside to creating an extra `HTTPService` in the service? Is there anything from the shared service that would be lost in the process or some efficiency cost to be aware of?

jon-betts: >Is there a downside to creating an extra HTTPService in the service? Is there anything from the shared service that would be lost in the process or some efficiency cost to be aware of?

No not really. The only efficiency it brings is if more than one service were to talk to the same site, then the SSL handshake would be saved. As we should be the only one talking to JSTOR, this is fine.

We can't use the global HTTP session, as putting headers in there would mess with everything that does use the global one.

It's probably a moot point anyway, as the "global" session is request scoped, so I doubt we get much from it anyway unless making more than one call during the same request.

tests/unit/lms/services/jstor/service_test.py
```diff
@@ -1,17 +1,23 @@
...
+from unittest.mock import patch, sentinel
 
 import pytest
 
 from lms.services import ExternalRequestError
 from lms.services.jstor.service import ArticleNotFound, JSTORService
 from tests import factories
 
-API_URL = "http://api.jstor.org"
+API_URL = "http://jstore_api.example.com"
```
jon-betts, line 11: Just some belt and braces against making real calls from the tests. Shouldn't happen, but doesn't cost us anything

tests/unit/lms/services/jstor/service_test.py
```diff
@@ -1,17 +1,23 @@
...
 
 
 class TestJSTORService:
+    def test_it_sets_tracking_headers(self, get_service):
+        svc = get_service(headers={"X-Header": "value"})
+
+        # pylint: disable=protected-access
+        assert svc._http.session.headers == {"X-Header": "value"}
```
robertknight, line 19: Patching private properties is unfortunate as an innocuous refactoring such as renaming the field can break the tests. Other fixtures in this module rely instead on patching the imports. I think that would look something like this:

```diff
diff --git a/tests/unit/lms/services/jstor/service_test.py b/tests/unit/lms/services/jstor/service_test.py
index 9ed4c6df..782ac25c 100644
--- a/tests/unit/lms/services/jstor/service_test.py
+++ b/tests/unit/lms/services/jstor/service_test.py
@@ -12,11 +12,10 @@ API_URL = "http://jstore_api.example.com"
 
 
 class TestJSTORService:
-    def test_it_sets_tracking_headers(self, get_service):
+    def test_it_sets_tracking_headers(self, get_service, http_service):
         svc = get_service(headers={"X-Header": "value"})
 
-        # pylint: disable=protected-access
-        assert svc._http.session.headers == {"X-Header": "value"}
+        assert http_service.session.headers == {"X-Header": "value"}
 
     @pytest.mark.parametrize("enabled", (True, False, None))
     @pytest.mark.parametrize("site_code", ("code", None, ""))
@@ -187,10 +186,13 @@ class TestJSTORService:
     def svc(self, get_service):
         return get_service()
 
+    @pytest.fixture
+    def HTTPService(self, patch):
+        return patch("lms.services.jstor.service.HTTPService")
+
     @pytest.fixture(autouse=True)
-    def http_service(self, svc):
-        with patch.object(svc, "_http") as http_service:
-            yield http_service
+    def http_service(self, HTTPService):
+        return HTTPService.return_value
 
     @pytest.fixture
     def ArticleMetadata(self, patch):
```

tests/unit/lms/services/jstor/service_test.py
```diff
@@ -168,28 +174,24 @@ def test_thumbnail_handles_404s(self, svc, http_service, response, exception):
             svc.thumbnail("article_id")
 
     @pytest.fixture
-    def http_service(self, http_service):
-        http_service.get.return_value = factories.requests.Response(
-            raw="https://s3.example.com/pdf"
```
robertknight, line 173: It is interesting that eliminating this fake response does not cause any tests to fail. Presumably that is a mistake elsewhere.

robertknight (commented):
I can't approve this PR in GitHub because I'm the original author, but the changes look reasonable to me. I had a comment regarding `_http` patching, I leave it up to you what to do with that. These changes do embed the assumption that we are going to pass the same tracking information to all requests, which I think will be the case.

tests/unit/lms/services/jstor/service_test.py
jon-betts: Done!

tests/unit/lms/services/jstor/service_test.py
jon-betts: Yeah. Some of our fixtures are also weird and don't return mocks. Not sure what was happening here.

PR #4449, opened 2022-09-16
Make sure newly created Organization have IDs

marcospri:
For: https://github.com/hypothesis/lms/issues/4448

Both PK and a `_public_id`. In some cases this might work as:

- The new org is added to the session implicitly on the AI relationship.

- Any query or the final commit will have the same effect as the flush.

but at least in the "unconfigured_launch" case an implicit flush doesn't happen before we try to use `public_id` on the organization.

# Testing

- Set un-org your application instances:

```update application_instances set organization_id = null;```

- On `main`, go to blackboard, create a new assignment and launch it, it will fail.

- Test again on this branch, it will succeed.

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -65,10 +65,14 @@ class OrganizationService:
                     "Multiple organization matches found for application instance %s",
                     application_instance.id,
                 )
+
             org = orgs[0]
 
         else:
             org = Organization()
+            self._db_session.add(org)
+            # Ensure we have ids
+            self._db_session.flush()
 
         # Fill out missing names
         if not org.name and (name := application_instance.tool_consumer_instance_name):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lms/services/organization.py [hunk 1]
```diff
@@ -65,10 +65,14 @@ def auto_assign_organization(
...
                     application_instance.id,
                 )
+
             org = orgs[0]
 
         else:
             org = Organization()
+            self._db_session.add(org)
+            # Get org.id and org._public_id from the DB
+            self._db_session.flush()
```
marcospri, line 75: This doesn't seem to be a problem on other code path and only affecting the non-canvas configuration. 

Some other implicit flush might be hiding the problem in other cases.

jon-betts: The comment here is a little misleading, as we get the id from the DB, but Python is writing the `_public_id` to the DB.

Maybe `Ensure we have ids`?

jon-betts (approved):

Issue #4462, opened 2022-09-20
Handle unbuilt catalog errors


darylhedley:
When making calls to VitalSource we can get errors like this back:

```json
{
	"errors": [
		"Catalog not found"
	]
}
```

This indicates that VitalSource have not "built the catalog" which is an action they need to perform for the key. There isn't anything we can do about this, but we can catch the error and present some useful info.

PR #4465, opened 2022-09-20
Handle catalog errors and group our error handling

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4462

This pulls all the error handling for the two book methods into a single chain of methods and unifies the testing.

## Testing notes (or lack thereof)

There's not much we can do to test this really, as it requires a situation on VitalSource's side. If you want to see the exception you can hack the code to raise the exception.

 * Hack `lms.services.vitalsource._client` to add `raise VitalSourceConfigurationError("test")` in `get_book_info()`
 * Start all the things
 * Visit: http://localhost:8001/admin/instance/id/8/
 * Enable Vitalsource
 * Visit: https://hypothesis.instructure.com/courses/125/
 * Create your own localhost assignment with VitalSource
 * Paste in any book, e.g. https://bookshelf.vitalsource.com/reader/books/9781786894700
 * You should see a message indicating VitalSource is misconfigured

### DIFF
--- a/lms/services/vitalsource/_client.py
+++ b/lms/services/vitalsource/_client.py
@@ -18,6 +18,11 @@ class BookNotFound(SerializableError):
         super().__init__(message=f"Book {book_id} not found")
 
 
+class VitalSourceConfigurationError(SerializableError):
+    def __init__(self, message):
+        super().__init__(message=f"VitalSource is misconfigured: {message}")
+
+
 class VitalSourceClient:
     """
     A client for making individual calls to VitalSource API.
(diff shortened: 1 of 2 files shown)


lms/services/vitalsource/_client.py
```diff
@@ -161,21 +160,30 @@ def get_user_credentials(self, user_reference: str) -> dict:
 
         raise VitalSourceError(error_code="vitalsource_user_not_found")
 
-    @staticmethod
-    def _is_resource_not_found(err: ExternalRequestError) -> bool:
-        """Get if an exception is due to a missing resource or a bad URL."""
+    @classmethod
+    def _handle_book_errors(cls, book_id: str, err: ExternalRequestError):
+        if json_errors := cls._get_json_errors(err):
+            if "Catalog not found" in json_errors:
```
marcospri, line 166: Do we know what this means? What's the solution? Needs VS intervention?

We could expand with a comment here if that's the case.

marcospri (approved):
This looks like a good solution to unify the error handling.

Not very clear when the new catalog error could happen or what it means.

lms/services/vitalsource/_client.py
jon-betts: Eh... not really. Apparently the "catalog was not intialised" the solution is to "initialise the catalog". This isn't something which really means much to us. I'll try and add a bit more info.

Issue #4463, opened 2022-09-20
Remove book licence checks


darylhedley:
When we call `v3/licenses.xml` we don't get all licences, we search for a specific book by adding `?sku=<vbid>` to the URL as per the documentation:

* https://developer.vitalsource.com/hc/en-us/articles/204332688-GET-v3-licenses-Read

But it's not a VBID, it's an SKU which is a different thing. To confuse matters one of the SKUs happens to be the same value as the VBID, but others don't. 

If the student has a licences for the version where the SKU and VBID match, it will work, if they don't they will see an error.

We should either:

 * Change our get book info method to return SKU information
 * Change our get licences method to no longer filter by SKU
 * Compare the two on the Python side to work out if a licences is present

OR...

 * Just remove the whole thing entirely and let the SSO do whatever it's going to do

Then...

 * We should remove the option to disable licence checking (once we are sure this is fine).

## Before and after

When the licence checking is working, it looks like this:

![Image](https://user-images.githubusercontent.com/7131143/191336134-536ae85f-fbe2-4373-8032-6c87144bbaf4.png)

After this has been removed it looks like this:

![Image](https://user-images.githubusercontent.com/7131143/191336270-16e0fede-13c9-4873-b213-9c308997a439.png)

# Tasks

 * [ ] https://github.com/hypothesis/lms/pull/4466

darylhedley:
After chatting with Jon, Rob and Daryl we have decided to move on this option:
"Just remove the whole thing entirely and let the SSO do whatever it's going to do".

(Add images of what the user will see if they don't have the book)

PR #4466, opened 2022-09-20
Remove licence checks and the disable licence check option

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4463

Our current licence checking method doesn't work. We're removing it instead of adding more.

## Testing notes

* Start all the things
* Add a VitalSource assignment
* e.g. https://bookshelf.vitalsource.com/reader/books/L-999-70867
* It should work!

### DIFF
--- a/lms/services/vitalsource/_client.py
+++ b/lms/services/vitalsource/_client.py
@@ -1,5 +1,5 @@
 from functools import lru_cache
-from typing import List, Optional
+from typing import List
 
 import xmltodict
 from marshmallow import EXCLUDE, Schema, fields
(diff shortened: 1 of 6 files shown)


lms/services/vitalsource/_client.py
```diff
@@ -213,17 +190,17 @@ def _json_request(self, method, url):
...
-        kwargs["headers"] = {"Accept": "application/xml; charset=utf-8"}
-
-        if data:
-            kwargs["data"] = xmltodict.unparse(data)
-            kwargs["headers"]["Content-Type"] = "application/xml; charset=utf-8"
+        kwargs["headers"] = {
+            "Accept": "application/xml; charset=utf-8",
+            "Content-Type": "application/xml; charset=utf-8",
+        }
+        kwargs["data"] = xmltodict.unparse(data)
```
jon-betts, line 203: This `if` no longer ever happens, so we end up with a coverage gap.

tests/unit/lms/services/vitalsource/_client_test.py
```diff
@@ -123,63 +123,6 @@ def test_get_table_of_contents_error(self, client, http_service, status_code):
...
-                    <license imprint="IMPRINT" name="NAME"/>
-                </licenses>
-            """,
-            """<?xml version="1.0" encoding="UTF-8"?>
-                <licenses>
-                    <license imprint="IMPRINT" name="NAME"/>
-                    <license imprint="IMPRINT2" name="NAME2"/>
-                </licenses>
-            """,
-        ),
```
jon-betts, line 142: This area here exercised a piece of code which checks for single or multiple items. Removing this caused a coverage gap.

marcospri (approved):
Looks good. Removes the license check and the confusing flag to make it optional.

PR #4470, opened 2022-09-21
Allow organizations search in the admin pages

marcospri:
Search organizations by name or public id.

# Testing 

- Go over http://localhost:8001/admin/orgs
- Search by public id
- Search by name

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -47,6 +47,24 @@ class OrganizationService:
             self._db_session, Organization, public_id, region=self._region
         )
 
+    def search(self, name, limit=100):
+        """
+        Search organizations.
+
+        :param name: Match organization by name. Case insensitive.
+        :param limit: Limit the number of results
+        """
+        return (
+            self._db_session.query(Organization)
+            .filter(
+                sa.func.to_tsvector("english", Organization.name).op("@@")(
+                    sa.func.websearch_to_tsquery(name, postgresql_regconfig="english")
+                )
+            )
+            .limit(limit)
+            .all()
+        )
+
     def auto_assign_organization(
         self, application_instance: ApplicationInstance
     ) -> Optional[Organization]:
--- a/lms/templates/admin/organizations.html.jinja2
+++ b/lms/templates/admin/organizations.html.jinja2
@@ -6,13 +6,54 @@ Organizations
 {% endblock %}
 
 {% block content %}
+<div class="block has-text-right">
+    <a class="button is-primary" 
+        href="{{ request.route_url("admin.organization.new") }}">
+        New organization
+    </a>
+</div>
 
 <fieldset class="box mt-6">
-    <div class="block has-text-right">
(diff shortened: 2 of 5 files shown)


lms/services/organization.py
```diff
@@ -47,6 +47,18 @@ def get_by_public_id(self, public_id: str) -> Optional[List]:
...
 
+    def search(self, name):
+        """
+        Search organizations.
+
+        :param name: Match organization by name. Case insensitive.
+        """
+        return (
+            self._db_session.query(Organization)
+            .filter(Organization.name.ilike(f"%{name}%"))
```
marcospri, line 58: This is not a very sophisticated text search but I reckon it would go a long way.

lms/templates/admin/organizations.html.jinja2
```diff
@@ -14,5 +14,46 @@ Organizations
             New organization
         </a>
     </div>
+    <legend class="label has-text-centered">Find organization</legend>
+    <form method="POST" action="{{ request.route_url("admin.organizations") }}">
+        <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+
+        {{ macros.form_text_field(request, "ID", "public_id") }}
```
marcospri, line 21: I prefer the explicit field per search criteria approach.

lms/views/admin/organization.py
```diff
@@ -96,6 +95,27 @@ def toggle_organization_enabled(self):
...
+    )
+    def search(self):
+        if not any((self.request.params.get(param) for param in ["public_id", "name"])):
+            self.request.session.flash(
+                "Need to pass at least one search criteria", "errors"
+            )
+            return HTTPFound(location=self.request.route_url("admin.organizations"))
+
+        if public_id := self.request.params.get("public_id"):
+            orgs = [self.organization_service.get_by_public_id(public_id)]
```
marcospri, line 111: Technically you are not "searching" for a public_id, if you find something that's it.

But I prefer the behavior to get the result back in the same page for both name and ID.

lms/templates/admin/organizations.html.jinja2
```diff
@@ -14,5 +14,46 @@ Organizations
             New organization
         </a>
     </div>
```
jon-betts, line 16: The "New organization" button shows up in the "Find organization" area, but doesn't use any of the fields. This should probably be outside of this box?

jon-betts (commented):
There are lots of suggestions here, but the two I think are probably worth considering are:
 
* Moving the "New organization" button outside of the find org area, where it looks like would listen to the boxes, but doesn't
* We can add fancy text searching really cheaply. I recon we should try

lms/services/organization.py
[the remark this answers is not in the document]
marcospri: Yeah I've used this before. It's probably overkill for a name but on the other hand, it doesn't cost us much, good idea :+1:

Issue #4467, opened 2022-09-20
Separate public id functionality into separate mixins


jon-betts:
The idea here so to pull out the public id functionality into two parts:

 * An abstract class for parsing and combining public ids
 * An `SQLAlchemy` mixin

This should lay the ground work to break up the searching functionality, and eventually remove the public id service.

## Tasks

 * https://github.com/hypothesis/lms/pull/4472
 * ~https://github.com/hypothesis/lms/pull/4473~
 * https://github.com/hypothesis/lms/pull/4502
    * Alternative to the above
    * From: https://github.com/hypothesis/lms/issues/4500

PR #4472, opened 2022-09-22
Split existing public id functionality into a model and mixin

jon-betts:
For:

This takes the existing functionality (not related to searching) and splits it out into separate classes.

This includes:

 * A `PublicIdMixin` which adds functionality to an SQLAlchemy model
 * A `PublicId` model which contains abstract functionality for generating, representing and parsing public id strings.

I think there's a bug fix in here too which catches creating public ids when the internal `_public_id` is None.


lms/models/public_id.py
```diff
@@ -0,0 +1,52 @@
...
+
+@dataclass
+class PublicId:
+    """
+    Get a globally unique value with prefixes like 'us.' or 'ca.'.
+
+    This is useful if you only have the id, but don't know which region you
+    should be looking for it in and is the only id suitable for sharing
+    outside a single region LMS context.
+    """
```
jon-betts, line 16: Splitting this out into it's own object separate from the mixin seems a bit much, but allows us to reason about public ids outside of an SQLAlchemy context.

For example, if you wanted to accept a public id string, and _then_ determine which model to get, you could use this class.

It also allows for some splitting up of the tests. Unfortunately the way these things compose together means we don't get an overall reduction in testing, as we tend to be forced by practicality to test the composed objects in a slightly black box way rather than our usual mocking.

lms/models/_mixins/public_id.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    public_id_model_code = None
+    """The short code which identifies this type of model."""
+
+    def public_id(self, region: Region) -> str:
+        """Get the globally unique id which also indicates the region."""
+
+        if not self._public_id:
+            return None
```
jon-betts, line 30: I think this was a bug before and would have resulted in nonsense like `us.lms.org.None`

jon-betts: Eugh... should probably be `-> Optional[str]`

marcospri: Yeah, mypy and similar tools would yell at you for this.

marcospri (approved):
Yeah this is tidier and pretty much orthogonal with the discussion we had about storing it on the DB.

PR #4473, opened 2022-09-22
Add a public id comparator to the `PublicIDMixin` and replace the public id service

jon-betts:
For:

Based on:

 * https://github.com/hypothesis/lms/pull/4472

This adds:

 * A `Regions.from_code()` method for lookup by code (like "us")
 * A `PublicId.parse()` method for parsing region strings with optional enforcement of expected values
 * A `PublicIdMixin.public_id_eq()` method as a convenience for comparing with public ids
 
This is then used do:

 * Remove the publid id service
 * Replace the query as a direct lookup in the organization service

This also enhances the experience by:

 * Catching errors and presenting them in-page, rather than as a separate page


lms/models/_mixins/public_id.py
```diff
@@ -36,3 +36,26 @@ def public_id(self, region: Region) -> str:
                 instance_id=self._public_id,
             )
         )
+
+    @classmethod
+    def public_id_eq(cls, public_id, region: Region):
```
marcospri, line 41: I find the comparator to be a bit confusing, I tend to prefer to leave the query side of SQLA as vanilla as possible, column on the left, value on the right.

I rather expose the parsed PublicID here and then just do `_public_id==model.parsed_public_id.instance_id` on the service side.

We'd get `model.parsed_public_id.region|app_code|...` for free which could be useful and more convenient than having another usage of `PublicId.parse`

`parsed_public_id` is not the best name ever.

jon-betts:
This is being closed in favour of:

 * https://github.com/hypothesis/lms/pull/4502

PR #4474, opened 2022-09-23
Ensure `course_groups()` returns users when we ask for them

jon-betts:
No matter what Canvas chooses to do. 

For:

 * https://sentry.io/organizations/hypothesis/issues/3604779293/?project=259908&referrer=slack

It seems that despite asking for users, Canvas sometimes doesn't send them. We will tack in empty lists so callers don't crash when trying to access the key.

### DIFF
--- a/lms/services/canvas_api/client.py
+++ b/lms/services/canvas_api/client.py
@@ -327,13 +327,21 @@ class CanvasAPIClient:
             # requests.
             send_kwargs["timeout"] = (31, 31)
 
-        return self._client.send(
+        groups = self._client.send(
             "GET",
             f"courses/{course_id}/groups",
             params=params,
             schema=self._ListGroups,
             **send_kwargs,
         )
+
+        # Sometimes Canvas doesn't send us users, even if we ask for them, so
+        # to ensure nobody crashes down-stream, we'll pop back in blank lists
+        if include_users:
+            for group in groups:
+                group.setdefault("users", [])
+
+        return groups
 
     def current_user_groups(self, course_id, group_category_id=None):
         """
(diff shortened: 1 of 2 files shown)


marcospri (commented):
With only this change this result in a constant spinner on the client which might be just hidding the error.


We could do something along this lines on top of the change in the PR:

```diff
diff --git a/lms/product/canvas/_plugin/grouping.py b/lms/product/canvas/_plugin/grouping.py
index a1306997..32c73be2 100644
--- a/lms/product/canvas/_plugin/grouping.py
+++ b/lms/product/canvas/_plugin/grouping.py
@@ -74,9 +74,12 @@ class CanvasGroupingPlugin(GroupingServicePlugin):
         self, _svc, course, group_set_id, grading_student_id=None
     ):
         # SpeedGrader requests are made by the teacher, get the student we are grading
-        return self._canvas_api.user_groups(
+        if groups := self._canvas_api.user_groups(
             self._custom_course_id(course), grading_student_id, group_set_id
-        )
+        ):
+            return groups
+
+        raise GroupError(ErrorCodes.GROUP_SET_EMPTY, group_set=group_set_id)
```

which then results in


```
The group set for this Hypothesis assignment is empty.

To fix this problem, add groups to the group set or use a different group set for this assignment.
```

Not sure if this is 100% correct. Are we not getting the users in the groups because there's none? Or there's some other canvas setting preventing the API to send the users along?


The groups returned in the sentry error look like:


```
...	
{
description: None, 
group_category_id: 25471, 
id: 359526, 
name: 'Friday 2pm Online Tutorial - Ben'
}, 
{
description: None, 
group_category_id: 25471, 
id: 355699, 
name: "Robin's tute, Tuesday 5pm"
}, 
...
```

which look very student specific so maybe the groups are not correctly set yet?


I reckon if we got feedback with that error we would know where too look next.

PR #4477, opened 2022-09-23
A more powerful unified search function for the organization service

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4480

This merges public id and name searching into one function under the hood. This allows a few improvements:

 * This allows an OR search across both
 * You can also hit search with no values and get a list of orgs. This could be handy for devs

## Notes

This also fixes text searching (hopefully) in QA and Live

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -44,14 +44,17 @@ class OrganizationService:
 
         return self._search_query(public_id=public_id).one_or_none()
 
-    def search(self, name, limit=100):
+    def search(self, public_id=None, name=None, limit=100) -> List[Organization]:
         """
-        Search organizations.
+        Search for organizations.
 
+        The results are returned as an OR of the specified filters.
+
+        :param public_id: Match on public id
         :param name: Match organization by name. Case-insensitive.
         :param limit: Limit the number of results
         """
-        return self._search_query(name=name).limit(limit).all()
+        return self._search_query(public_id=public_id, name=name).limit(limit).all()
 
     def _search_query(self, public_id=None, name=None):
         clauses = []
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -42,11 +42,7 @@ class OrganizationService:
     def get_by_public_id(self, public_id: str) -> Optional[List]:
         """Get an organization by its public_id."""
 
-        return (
-            self._db_session.query(Organization)
-            .filter(Organization.public_id == public_id)
-            .one_or_none()
-        )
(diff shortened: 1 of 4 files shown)


lms/views/admin/organization.py
```diff
@@ -105,17 +105,14 @@ def toggle_organization_enabled(self):
...
-                orgs = [self.organization_service.get_by_public_id(public_id)]
-            except InvalidPublicId as err:
-                self.request.session.flash(str(err), "errors")
-                orgs = []
-
-        else:
+        try:
             orgs = self.organization_service.search(
-                name=self.request.params.get("name", "").strip()
+                name=self.request.params.get("name"),
```
jon-betts, line 110: We don't need the strip any more as the new full text search processing stuff takes care of it

lms/services/organization.py
```diff
@@ -43,29 +43,38 @@ def get_by_linked_guid(self, guid) -> List[Organization]:
...
                     sa.func.websearch_to_tsquery(name, postgresql_regconfig="english")
                 )
             )
-            .limit(limit)
-            .all()
-        )
+
+        query = self._db_session.query(Organization)
+        if clauses:
+            query = query.filter(sa.or_(*clauses))
```
jon-betts, line 75: It would be very easy to parameterise the `sa.or_` here to an `and` if there was ever a need.

lms/services/organization.py
```diff
@@ -43,29 +43,38 @@ def get_by_linked_guid(self, guid) -> List[Organization]:
...
+        :param public_id: Match on public id
+        :param name: Match organization by name. Case-insensitive.
         :param limit: Limit the number of results
         """
-        return (
-            self._db_session.query(Organization)
-            .filter(
+        return self._search_query(public_id=public_id, name=name).limit(limit).all()
+
+    def _search_query(self, public_id=None, name=None):
```
marcospri, line 60: I was going to propose to include ID here but I've seen you done it in a follow up PR.

We could rename this to `organization_query` or something like that? I guess we are "searching" for them in all cases so not a strong suggestion on the name change.

marcospri (approved):

lms/services/organization.py
jon-betts: My thinking with the name was:

 * You get the noun (organization) from context
 * The verb is "searching" - I know "query" is a pseudomyn of search, but in an SQL context a "query" could write instead of read
 
How about a best of both with `_organization_search_query`?

PR #4479, opened 2022-09-23
Allow searching for organizations by GUID

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4480

Depends on:

 * https://github.com/hypothesis/lms/issues/4480
 * https://github.com/hypothesis/lms/pull/4478

Now we have a unified search, it's quite easy to add more variations to it. This PR:

 * Inlines the GUID search into the unified search function
 * Adds it to the admin pages
 * Removes the old function which is completely superseded by search now

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -23,15 +23,7 @@ class OrganizationService:
             return []
 
         return (
-            self._db_session.query(Organization)
-            .join(ApplicationInstance)
-            .outerjoin(GroupInfo)
-            .filter(
-                sa.or_(
-                    ApplicationInstance.tool_consumer_instance_guid == guid,
-                    GroupInfo.tool_consumer_instance_guid == guid,
-                )
-            )
+            self._organization_search_query(guid=guid)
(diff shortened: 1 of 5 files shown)


lms/services/organization.py
```diff
@@ -82,7 +64,16 @@ def _search_query(self, id_=None, public_id=None, name=None):
...
 
-        query = self._db_session.query(Organization)
+        if guid:
+            query = query.join(ApplicationInstance).outerjoin(GroupInfo)
+
+            clauses.extend(
+                sa.or_(
+                    ApplicationInstance.tool_consumer_instance_guid == guid,
+                    GroupInfo.tool_consumer_instance_guid == guid,
+                )
```
jon-betts, line 74: We _could_ replace this with two separate additions to `clauses` as they are `or`-ed later anyway. But if we want to be able to change the later `or` to an `and` this would still remain an `or`. So I think this is the right formulation.

marcospri (approved):
Yup, this will be useful for a while :+1:

PR #4481, opened 2022-09-23
POC - Example search engine style search for organizations

jon-betts:
This is a proof of concept to show using the unified search function to provide a search engine like experience with a single input query field.

The layout here is not what you would do for this kind of interface, but it demonstrates how it might work.


jon-betts:
I think this is cool, but it's not essential and it's time to move on. We always have this in the back pocket if we want it.

Issue #4500, opened 2022-10-04
Investigate global region object


jon-betts:
Currently a lot of the code around public ids is quite irritating due to the need to inject the region from the request. This means providing the request (which is at the outer layer of the app) all the way down to the DB models at the center, and is quite disruptive.

It would be ideal if:

 * The correct region was available to the models directly
 * The region was configured during the normal app configuration
 * Used the value for authority after any merging and parsing of config values and env vars is done

This is the same situation as we have with injecting authority for usernames etc, and we might have something here we can re-use.

## Tasks

 * https://github.com/hypothesis/lms/pull/4501
 * https://github.com/hypothesis/lms/pull/4503

PR #4501, opened 2022-10-04
Using a global singleton instead of a request method for region

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4500

This keeps the request method for now, but makes it optional. This is just to demonstrate the way of setting this up.

### DIFF
--- a/lms/models/region.py
+++ b/lms/models/region.py
@@ -1,7 +1,5 @@
 from dataclasses import dataclass
 
-from pyramid.request import Request
-
 
 @dataclass(frozen=True)
 class Region:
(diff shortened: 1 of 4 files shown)


lms/models/region.py
```diff
@@ -46,5 +63,8 @@ def from_request(cls, request: Request) -> Region:
 
 def includeme(config):
     config.add_request_method(
-        Regions.from_request, name="region", property=True, reify=True
+        lambda _request: Regions.get_region(),
+        name="region",
+        property=True,
+        reify=True,
```
jon-betts, line 69: When the dust settles, we shouldn't need this request method

lms/models/region.py
```diff
@@ -27,17 +25,36 @@ class Regions:
...
         """
-        Get a region object based on a request.
+        Get the region this app is running in.
 
-        :raises ValueError: If no valid region can be found.
+        :raises ValueError: If the active region has not been set
         """
-        authority = request.registry.settings["h_authority"]
+        if cls._current_region is None:
+            raise ValueError(
```
marcospri, line 38: I reckon this is enough to guard against any usage we might do that for some strange reason happen before `create_app`. Pretty unlikely.

tests/unit/lms/models/region_test.py
```diff
@@ -23,17 +27,34 @@ def test_from_request_raises_ValueError_for_bad_authorities(
...
 
         with pytest.raises(ValueError):
-            Regions.from_request(pyramid_request)
+            Regions.set_region(pyramid_request)
+
+    @pytest.fixture
+    def current_region(self):
+        with mock.patch.object(Regions, "_current_region") as _current_region:
+            _current_region.__get__ = mock.Mock(return_value=None)
+            yield _current_region.__get__
```
jon-betts, line 36: The global nature of this has the potential to cause some havok in testing, which is one down side. Here I've avoided the problem by ensuring these tests are isolated from the global object.

marcospri (approved):

PR #4502, opened 2022-10-04
Replace the public id service with a hybrid comparator on `PublicIDMixin` 

jon-betts:
For:

Alternative to:

 * https://github.com/hypothesis/lms/pull/4473


lms/services/organization.py
```diff
@@ -42,7 +41,12 @@ def get_by_linked_guid(self, guid) -> List[Organization]:
 
     def get_by_public_id(self, public_id: str) -> Optional[List]:
         """Get an organization by its public_id."""
-        return get_by_public_id(self._db_session, Organization, public_id)
+
+        return (
```
marcospri, line 45: Still not a huge fan of the comparator, now this query could raise a `InvalidPublicId` and that's not entirely clear.

I'd prefer to parse explicitly here and then query with _public_id but that's probably a matter of taste.

marcospri (approved):

lms/services/organization.py
jon-betts: Yeah, it's a little surprising it can raise, but it does mean you can't work around it.

The upside of this approach is nothing but the model+mixin needs to know about the model code ("org") which you would need to leak out here in order to parse it.

If it's a matter of taste I'll probably leave it as is. Sorry!

PR #4503, opened 2022-10-04
Use a hybrid property on `PublicIDMixin` to remove the request method

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4500


marcospri (approved):
This is much nicer, go have now committed the sin of having a global singleton but it's worth the effort.

If we move to storing the full public_id in the DB I reckon we'd only need to change the model now.

PR #4505, opened 2022-10-05
Created CodeQL GitHub Action

indigobravo:
The default workflow with a custom configuration file to enable us to exclude paths as required.

Part of the work towards: https://github.com/hypothesis/playbook/issues/968


indigobravo:
Tests have now been filtered out, but we are still seeing alerts against what looks like frontend code. Going to run this passed the team.

PR #4506, opened 2022-10-05
Update concurrency settings

indigobravo:
Deploys were taking too long. This update moves concurrency settings inside the individual jobs.

- continuous integration (allow in parallel)
- create-image (allow in parallel)
- qa-us-west-1 (queue, don't cancel)
- prod-us-west-1 (cancel others)
- prod-ca-central-1 (cancel others)

For more information see: https://github.com/hypothesis/playbook/issues/987

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,5 @@
 name: Deploy
 
-concurrency:
-  group: ${{ github.event.repository.name }}-deploy
-  cancel-in-progress: true
-
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


indigobravo:
Setting this to a draft status for now. I have an issue going to track work regarding these changes: https://github.com/hypothesis/playbook/issues/987

.github/workflows/deploy.yml
```diff
@@ -23,6 +19,9 @@ jobs:
     secrets: inherit
```
jon-betts, line 19: There is some concern that deploying concurrently to Dockerhub could cause concerns. I'd be in favour of adding:

```yaml
 concurrency:
      group: ${{ github.event.repository.name }}-deploy-dockerhub
      cancel-in-progress: false
```

To eliminate this worry and then trying this out. It should be miles better than what we have.

indigobravo: Update has been made.

jon-betts (approved):
I think this looks good, and we should try it and see how it works out.

If it works well this could really speed us up.

PR #4508, opened 2022-10-05
PoC Vitalsource inclusive access

marcospri:


lms/services/vitalsource/service.py
```diff
@@ -148,3 +160,54 @@ def compile_user_lti_pattern(pattern: str) -> Optional[re.Pattern]:
...
+        if vitalsource_svc.get_opt_out_for_user(vs_course_id, user_reference):
+            raise VitalSourceOptedOut()
+
+    def get_vs_course_id(self, course: Course):
+        if course_id := course.extra.get("vs", {}).get("course_id"):
+            return course_id
+
+        for vs_course in self._sso_client.get_courses():
+            if vs_course["lms_context_id"] == course.lms_id:
+                course.extra.setdefault("vs", {})
```
marcospri, line 185: I had this caching bit here, just on course but this would make sense probably also on "grouping_membership" for the student opt out deccison.

We'll have to decide
- Is this too slow to do in every launch?
- If it's indeed to slow we could:


    1. Follow the caching approach to speed thing up.
    2. Defer all of this to a celery task. A bit of delay of getting the data has the effect of allowing the student in instead of blocking.

lms/services/vitalsource/_client.py
```diff
@@ -96,6 +97,22 @@ def get_table_of_contents(self, book_id: str) -> List[dict]:
 
         return toc
 
+    def get_courses(self):
```
marcospri, line 100: If I remember correctly of off this course and roster fetching was needed to get the ID of the student in the VS system.

lms/services/vitalsource/service.py
```diff
@@ -22,6 +26,8 @@ def __init__(
         user_lti_param: Optional[str] = None,
         user_lti_pattern: Optional[str] = None,
         enable_licence_check: bool = True,
+        enable_inclusive_access: bool = False,
+        inclusive_access_vbid: Optional[str] = None,
```
marcospri, line 30: I don't remember the details of VBID vs SKU but given this is about just one Hypothesis value it shouldn't be a problem to use the right value here.

lms/views/lti/basic_launch.py
```diff
@@ -239,9 +240,26 @@ def _record_launch(self):
...
             self.context.application_instance, self.request.lti_params
         )
-
         if not self.request.lti_user.is_instructor and not self.context.is_canvas:
             # Create or update a record of LIS result data for a student launch
             self.request.find_service(name="grading_info").upsert_from_request(
                 self.request
             )
+
+    def _check_inclusive_access(self):
```
marcospri, line 249: Maybe this should be eventually `check_license` or similar of which VS inclusive access is just one of the possibilities.

PR #4511, opened 2022-10-05
Present a message on the organization search instead of a blank table when there are no results

jon-betts:
Just a small quality of life improvement. The page looks kind of goofy with an empty table.


marcospri (approved):
Too many cats, but looks good :dog:

Issue #4507, opened 2022-10-05
CodeQL errors


indigobravo:
2 errors were found when enabling CodeQL scanning for LMS. Details can be found here: https://github.com/hypothesis/lms/pull/4505/checks?check_run_id=8721551036

<img width="1071" alt="Screenshot 2022-10-05 at 14 37 11" src="https://user-images.githubusercontent.com/18003117/194074100-846775f2-aafc-4f54-96c1-1511340e886a.png">

Please advise once a fix has been provided so we can start using the CodeQL tool.

https://hypothes-is.slack.com/archives/C1M8NH76X/p1664973964524899

PR #4522, opened 2022-10-06
Escape "." in DOI regex

robertknight:
This fixes a mistake found by CodeQL. See https://github.com/hypothesis/lms/issues/4507.


lyzadanger (approved):
Yay for when tools find actual things!

PR #4523, opened 2022-10-06
Refine check for whether a sharing URL refers to SharePoint

robertknight:
Make sure we check the top-level domain as a whole. The check should match hostnames like "companyname.sharepoint.com" but not "notsharepoint.com". As far as I know, the hostname should always have a subdomain (ie. will never be just "sharepoint.com").

This issue was found by CodeQL. See https://github.com/hypothesis/lms/issues/4507

### DIFF
--- a/lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
+++ b/lms/static/scripts/frontend_apps/utils/onedrive-picker-client.js
@@ -121,7 +121,7 @@ export class OneDrivePickerClient {
    */
   static downloadURL(sharingURL) {
     const url = new URL(sharingURL);
-    if (url.hostname.endsWith('sharepoint.com')) {
+    if (url.hostname.endsWith('.sharepoint.com')) {
       url.search = 'download=1';
       return url.href;
     } else {


robertknight:
>  As far as I know, the hostname should always have a subdomain (ie. will never be just "sharepoint.com").

@marcospri, can you confirm if I'm correct about this?

marcospri:
> > As far as I know, the hostname should always have a subdomain (ie. will never be just "sharepoint.com").
> 
> @marcospri, can you confirm if I'm correct about this?

I think so but not 100% sure. 

Looks like for example this exists https://microsoft.sharepoint.com/ and points to a login page. The bare https://sharepoint.com points to a marking page.

Also, while is true, as CodeQL suggest, that we could tag a `notsharepoint.com` URL as sharepoint the missing context is that that URL is coming from the MS api in the first place.

I'd say this is safe to merge but can't guarantee it.

marcospri (approved):
As I said in the comment on the PR

`I'd say this is safe to merge but can't guarantee it.`

robertknight:
> I'd say this is safe to merge but can't guarantee it.

OK. I'll merge it. If we run into a problem we can revisit.

PR #4525, opened 2022-10-06
[PoC] Rollup tables 

marcospri:
Based on https://www.citusdata.com/blog/2018/10/31/materialized-views-vs-rollup-tables/

without any extra functions, relying only on the date of the original table. It needs to recalculate the last "bucket" every time but that should be fine for our scale.

My idea is that this approach could be complementary to postgres `MATERILIZED VIEWS` for tables that produce aggregates.


lms/models/reporting/org_events.py
```diff
@@ -0,0 +1,115 @@
...
+class RollUpMixin:
+    rollup_index_elements = []
+    """Which columns are unique in the rollup table to find conflicts"""
+    rollup_query = None
+    """Query that produces the full row set of the rollup table"""
+    rollup_condition = None
+    """Condition to add to `rollup_query` when running this periodically"""
+
+    @classmethod
+    def rollup(cls, db_session, regenerate=False):
```
marcospri, line 23: This probably should take two dates in case we need to do the initial fill by batches.

lms/models/reporting/org_events.py
```diff
@@ -0,0 +1,115 @@
...
+        query = cls.rollup_query()
+        if not regenerate:
+            # Filter by the current day when updating periodically
+            query = query.where(cls.rollup_condition)
```
marcospri, line 27: The main assumption is that when using this, it much faster than a full materialized view refresh.

lms/models/reporting/org_events.py
```diff
@@ -0,0 +1,115 @@
...
+from sqlalchemy import PrimaryKeyConstraint
+
+
+from lms.db import BASE
+from lms.models import Organization, Event, ApplicationInstance, EventUser, LTIRole
+
+LOG = logging.getLogger(__name__)
+
+
+class RollUpMixin:
```
marcospri, line 14: This is a bit extra for a PoC but :shrug:

lms/models/reporting/org_events.py
```diff
@@ -0,0 +1,115 @@
...
+                func.count(Event.id).label("count"),
+            )
+            .select_from(Event)
+            .join(ApplicationInstance)
+            .join(Organization)
+            .group_by(Organization.id, Event.type_id, column("day"))
+        )
+
+
+class RollUpUsersByOrganization(BASE, RollUpMixin):
```
marcospri, line 81: Looks like 

```
organization_id |    role    |      timestamp      | count 
-----------------+------------+---------------------+-------
               4 | instructor | 2022-09-14 00:00:00 |     2
               4 | instructor | 2022-09-15 00:00:00 |     1
               4 | learner    | 2022-09-16 00:00:00 |     1
               4 | instructor | 2022-09-16 00:00:00 |     1
               3 | instructor | 2022-09-16 00:00:00 |     1
               4 | instructor | 2022-09-19 00:00:00 |     2
               4 | instructor | 2022-09-23 00:00:00 |     1
               4 | learner    | 2022-10-03 00:00:00 |     1
               4 | instructor | 2022-10-03 00:00:00 |     1
               4 | admin      | 2022-10-03 00:00:00 |     1
               4 | instructor | 2022-10-05 00:00:00 |     1
               4 | learner    | 2022-10-06 00:00:00 |     1
               4 | instructor | 2022-10-06 00:00:00 |     2
               4 | admin      | 2022-10-06 00:00:00 |     1
```

PR #4533, opened 2022-10-07
Ensure we ask for permission before moving onto deploy jobs

jon-betts:
For:

 * https://github.com/hypothesis/playbook/issues/987
 
A follow up to:

 * https://github.com/hypothesis/lms/pull/4506

When GHA moves from one job to another in the workflow it appears to:

 1. Check the concurrency rules
 2. Cancel the job if required by those rules
 3. Ask for approval if required

This makes sense, as otherwise you would potentially have lots of jobs waiting on manual approval, which would then immediately get cancelled when you said yes.

Adding an extra step with the approval inside prevents GHA from cancelling the jobs before we've even had a chance to approve them.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,11 +36,26 @@ jobs:
       Version: latest
     secrets: inherit
 
+  approval-step:
+    # This step is here just to get approval, without it GHA will proceed onto
+    # the next steps, check the concurrency rules and cancel them in certain
+    # cases. The resolution order appears to be:
+    # 1. Move to next job
+    # 2. Check concurrency (cancelling here is not what we want)
+    # 3. Ask for approval
+    needs: qa-us-west-1
+    with:
+      Environment: prod
+    steps:
+      # We need some dummy content for this job to be valid in GHAs eyes
+      - name: Log approval
+        run: echo "Approval granted"
+
   prod-us-west-1:
     concurrency:
       group: ${{ github.event.repository.name }}-deploy-us
       cancel-in-progress: true
-    needs: qa-us-west-1
+    needs: approval-step
     name: ${{ github.event.repository.name }}
     uses: hypothesis/workflows/.github/workflows/eb-update.yml@main
     with:
@@ -55,7 +70,7 @@ jobs:
     concurrency:
       group: ${{ github.event.repository.name }}-deploy-ca
       cancel-in-progress: true
-    needs: qa-us-west-1
+    needs: approval-step
     name: ${{ github.event.repository.name }}
     uses: hypothesis/workflows/.github/workflows/eb-update.yml@main
     with:


.github/workflows/deploy.yml
```diff
@@ -36,11 +36,22 @@ jobs:
...
+  approval-step:
+    # This step is here just to get approval, without it GHA will proceed onto
+    # the next steps, check the concurrency rules and cancel them in certain
+    # cases. The resolution order appears to be:
+    # 1. Move to next job
+    # 2. Check concurrency (cancelling here is not what we want)
+    # 3. Ask for approval
+    needs: qa-us-west-1
+    with:
+      Environment: prod
```
jon-betts, line 48: Should this be:

```suggestion
      Environment: approval
```

And move the requirement for approval to this environment, not prod?

indigobravo, line 48: I think we need to test this to find out for sure, but I have a feeling we will hit 2 approval stages.

- One for the new approval-step job being added
- One for the prod deploys

However, I could be wrong.

indigobravo: And your thinking is we remove the protection rules against the `prod` environment and add them to `approval`?

robertknight: If this is what you're thinking of doing then it should work, but would be "unconventional" usage of environments. As long as we're documenting why we're doing this I think we can go with it. It might be a good idea to ping GitHub support and tell them we're having to use this workaround in case they have other ideas.

jon-betts: I suspect you are right, but should we try this first as it involves less shenanigans? And then try the "approval" environment as a fallback?

robertknight: If we test this out at least we'll verify our understanding of how this GHA behavior works.

indigobravo: I think the `approval` environment will work around the problem and means we don't have to modify the function being called. We can just remove the protection rules.

I save give the `approval` route a test.

indigobravo (approved):

seanh (changes requested):
Could the PR title and description be updated to clarify what problem this PR fixes? It's unclear to me currently. It says <q>Adding an extra step with the approval inside prevents GHA from cancelling the jobs before we've even had a chance to approve them</q>. So is GHA cancelling jobs before we get a chance to approve them, in some circumstances? Could we have an example of what is currently happening (that we don't like) and what will happen if this PR is merged (that we like better)?

seanh (commented):
I think I like the idea of a "manual approval" step:

If I understand it correctly the problem on `main` is that as soon as GitHub Actions enqueues the jobs belonging to the `lms-deploy-us` and `lms-deploy-qa` concurrency groups those jobs will go into a pending state (because they're waiting on manual approval of the `prod` environment) but they will immediately cancel any other `lms-deploy-us` and `lms-deploy-qa` jobs from previous workflow runs that were already pending. This is different from what used to happen on Jenkins and it's undesirable: we don't want those previous workflow runs to get cancelled.

If you merge your PR and it's deployed to QA and waiting for approval, and then I merge my PR and it gets deployed to QA and waits for approval, we don't want my workflow run to cancel your workflow run because you'd still like to be able to approve your workflow run and deploy your changes to prod without also deploying my changes at the same time. And then I can come along later and deploy my changes separately.

On this PR by adding a manual approval step we prevent the jobs using the `lms-deploy-us` and `lms-deploy-qa` concurrency groups from getting enqueued at all (even in a pending state) and cancelling other pending jobs because the jobs won't get enqueued until the `prod` environment is approved and releases the approval step.

But...

### Is this further misusing GHA environments?

Is there any way to make a job wait for approval _without using an environment?_

I think we might be mis-using environments in a couple of ways here:

1. (This is already on main): in some cases we use a single `qa` GHA environment to deploy to multiple actual environments (for example `qa` and `qa-websocket` in h) and/or we use a single `prod` GHA environment to deploy to multiple actual environments (h and websocket, US and Canada). Reading the docs I think each actual deployment environment is supposed to be a separate GHA environment. You're supposed to be able to approve all the deploys at once or approve each deploy separately (e.g. deploy to US right now but not to Canada yet). Each environment is also supposed to be able to have a URL which GitHub displays in various places in the UI, handy for being able to click on the link to open the QA or prod environment that you just deployed to test it, but each GHA environment can only have one URL not multiple.

2. I think the changes on this branch are based on the above mis-use of environments. Since all our actual production environments use a single GHA environment named `prod` it's possible to add an approval step that also references this `prod` environment. But this only works because we're mis-using environments, and it'd prevent us from being able to fix the workflow to use environments properly.

3. The alternative suggestion on this PR is to add a fake `approval` environment. As mentioned in the comments I don't think that's gonna do what we want: it's gonna make you approve each workflow twice (first for `approval` and then for `prod`). The fake `approval` environment will also show up in places in the GitHub UI like on the sidebar of the repo's front page, because GitHub thinks it's an actual environment where the app can be seen running.

seanh:
Manual approval without using environments: https://github.com/marketplace/actions/manual-workflow-approval

.github/workflows/deploy.yml
seanh: Hmm. The fake `approval` environment would show up in various places in the GitHub UI. In the app's list of environments in the sidebar, on the app's deployment history page, etc. It's gonna say things like "**Deployed to approval**" in the UI.

A fake `approval` environment may potentially be the lesser of two evils compared to using the `prod` environment for the approval job though because I don't think there should _be_ a single `prod` environment in the first place (see my notes about this below)

seanh:
I'm not sure that we should do this, Slack post: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1666287339445079?thread_ts=1666277181.188469&cid=C4K6M7P5E

seanh:
This is no longer applicable because it turns out that we can't use job-level concurrency controls at all for things like deployments and releases ([details](https://hypothes-is.slack.com/archives/CR3E3S7K8/p1666949909080719?thread_ts=1666891634.395559&cid=CR3E3S7K8)). LMS's `deploy.yml` was [reverted to workflow-level concurrency](https://github.com/hypothesis/lms/pull/4595). Even if job-level concurrency were usable I wouldn't want to misuse GitHub environments by creating a fake approval job and/or a fake approval environment, as both of those things would have various undesirable consequences

PR #4551, opened 2022-10-17
Fix AI moving organization overriding all settings

marcospri:
Fixes https://github.com/hypothesis/product-backlog/issues/1398

# Testing notes

- Go over an AI in the admin pages http://localhost:8001/admin/instance/id/101/
- Make sure it has some settings set
- Move it to a an org 
- All the settings are preserved
- In `main`, repeat org moving operation.
- All settings are set to false/null

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -136,7 +136,7 @@
 
 
     {% call macros.field_body(label="Move to organization") %}
-        <form method="POST" action="{{ request.route_url("admin.instance.id", id_=instance.id) }}">
+        <form method="POST" action="{{ request.route_url("admin.instance.move_org", id_=instance.id) }}">
           <input type="hidden" name="csrf_token" value="{{ get_csrf_token() }}">
           <input class="input" type="text" name="org_public_id">
           <input type="submit" class="button mb-2" value="Move">
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -252,6 +252,34 @@ class AdminApplicationInstanceViews:
         return {"instance": ai}
 
     @view_config(
+        route_name="admin.instance.move_org",
+        request_method="POST",
+        require_csrf=True,
+    )
+    def move_application_instance_org(self):
+        ai = self._get_ai_or_404(**self.request.matchdict)
+
+        try:
+            self.application_instance_service.update_application_instance(
+                ai,
+                organization_public_id=self.request.params.get(
+                    "org_public_id", ""
+                ).strip(),
+            )
+            self.request.session.flash(
+                f"Updated application instance {ai.id}", "messages"
+            )
+        except ValidationError as err:
+            self.request.session.flash(err.messages, "validation")
+            return HTTPFound(
+                location=self.request.route_url("admin.instance.id", id_=ai.id)
+            )
+
+        return HTTPFound(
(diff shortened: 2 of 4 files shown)


lms/templates/admin/instance.html.jinja2
```diff
@@ -136,7 +136,7 @@
 
 
     {% call macros.field_body(label="Move to organization") %}
-        <form method="POST" action="{{ request.route_url("admin.instance.id", id_=instance.id) }}">
+        <form method="POST" action="{{ request.route_url("admin.instance.move_org", id_=instance.id) }}">
```
marcospri, line 139: There are many alternatives to this fix that still allow to use the same endpoint but I reckon this is the easiest to implement and reason about.

lms/views/admin/application_instance.py
```diff
@@ -252,37 +252,54 @@ def show_instance(self):
...
         require_csrf=True,
     )
-    def update_instance(self):
+    def move_application_instance_org(self):
         ai = self._get_ai_or_404(**self.request.matchdict)
 
         try:
             self.application_instance_service.update_application_instance(
                 ai,
-                lms_url=self.request.params.get("lms_url", "").strip(),
```
marcospri, line 270: Diff it's a bit strange here.

- No more try/except on `update_instance`
- New view for move_application_instance_org, with the try.

lms/views/admin/application_instance.py
```diff
@@ -252,37 +252,54 @@ def show_instance(self):
...
+            self.request.session.flash(
+                f"Updated application instance {ai.id}", "messages"
+            )
         except ValidationError as err:
             self.request.session.flash(err.messages, "validation")
             return HTTPFound(
                 location=self.request.route_url("admin.instance.id", id_=ai.id)
             )
 
+        return HTTPFound(
```
marcospri, line 278: This view is a good candidate to have the audit trail call. Not in this PR as it has some extra complexity.

lms/views/admin/application_instance.py
jon-betts: It's a bit weird for the service to be raising ValidationErrors which have specific params specified in them. Seems like a mix of view and service concerns.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -61,6 +61,44 @@ def test_instance_with_duplicate(self, views, application_instance_service):
...
+    def test_move_application_instance_org(
+        self,
+        views,
+        pyramid_request,
+        application_instance,
+        application_instance_service,
+    ):
+        application_instance_service.get_by_id.return_value = application_instance
+        pyramid_request.params["org_public_id"] = "PUBLIC_ID"
+
```
jon-betts, line 73: Might this benefit from an explicit test case that says we don't lose the settings? Or is that an example of testing one of the infinite number of things we don't expect to happen.

I'm more inclined towards it as it's a bug that actually happened.

jon-betts (approved):
Yep this looks good. I wonder if this would benefit from a test to prevent regression on this?

tests/unit/lms/views/admin/application_instance_test.py
marcospri: Not sure how the test could look:

- The update AI view still has the same issue, if you don't pass the settings it will update them all.

- The new view doesn't get near the settings. We could test that it doesn't indeed change them.

I'll have a think and maybe propose an alternative fix and test. I'll merge this now to avoid further updating data in production.

PR #4566, opened 2022-10-18
D2L api integration PoC

marcospri:
## General API

- We need to use https endpoints to use the API. The approach taken here is enough.
- The authentication (OAuth2) is similar to other LMSs

## Groups 

- Groups match pretty much 1 to 1 with other implementations

## Files

- The API around files seems a bit incomplete.
- The last endpoint, to get the file contents, return the PDF bytes directly and needs to be authenticated. This is not how it usually works with via (not 100% sure, maybe we can reuse some other source's existing flow)

- "public files"

Not linked to a course. No API integration with those.

- There seems to be a dropbox integration accessible from the API :thinking: 

https://docs.valence.desire2learn.com/res/dropbox.html

- Files can be added directly to the course here 
https://aunltd.brightspacedemo.com/d2l/lp/manageFiles/main.d2l?ou=6782

I can't find a way to list those through the API

- Or added directly as "topics" (the same as assignment)

This also end up in "Manage files" but they can be accessed. This is what the PR works on but it seems and odd choice to organize files.

## Product questions

- Are there multiple UI flavors like in blackboard we don't know about?

- Could we offer this as an LTI1.3 only feature. It does not need LTI1.3 in particular but we could avoid some extra compatibility work.

- Can we confirm how schools use groups. The assumption here is "like in canvas or blackboard".

- Can we confirm how schools use files. Is not very clear if they'd use them like I assume here.
- If we phased this in 1/ groups and 2/files. Would adding extra scopes for files later on be a problem.
- Should we tackle "course copy for groups" attached to this.

### DIFF
--- a/conf/development_https.ini
+++ b/conf/development_https.ini
@@ -1,0 +1,1 @@
+development.ini
--- a/conf/supervisord-dev.conf
+++ b/conf/supervisord-dev.conf
@@ -9,6 +9,15 @@ stderr_events_enabled=true
 stopsignal = KILL
 stopasgroup = true
 
+[program:web-https]
+command=gunicorn --paste conf/development_https.ini --certfile=/home/marcos/localhost.crt --keyfile=/home/marcos/localhost.key  -b :48001
+
+stdout_events_enabled=true
+stderr_events_enabled=true
+stopsignal = KILL
+stopasgroup = true
+
+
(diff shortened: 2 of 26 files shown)


marcospri:
PoC not needed as an open PR. Bits of this area already on `main`.

PR #4587, opened 2022-10-25
Refactor groups config

marcospri:
While doing the D2L groups PoC there was lots of repetition around the config (is the feature enabled? what are the endpoint details?)

This aims to reduce the duplication exposing the details once, not once per product.

## Testing 

To check nothing's been broken in the process:

### Canvas

- Reconfigure https://hypothesis.instructure.com/courses/125/assignments/3118/edit?name=Testing+upgrade+-+marcos&due_at=null&points_possible=10

Set it to any document source and pick groups and a groupset. Launch it.

### Blackboard

- Same testing with https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_183_1&course_id=_19_1

### DIFF
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -28,11 +28,14 @@ class Family(str, Enum):
 class Routes:
     """A collection of Pyramid route names for various functions."""
 
-    oauth2_authorize: str = None
+    oauth2_authorize: Optional[str] = None
     """Authorizing with OAuth 2."""
 
-    oauth2_refresh: str = None
+    oauth2_refresh: Optional[str] = None
     """Refreshing OAuth 2 tokens."""
+
+    list_group_sets: Optional[str] = None
+    """List available group sets. Takes a course_id parameter"""
 
 
 @dataclass
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -64,10 +64,9 @@ class Product:
     Family = Family
 
     @classmethod
-    def from_request(cls, request):
+    def from_request(cls, request, ai_settings: Dict):
         """Create a populated product object from the provided request."""
-        ai = request.find_service(name="application_instance").get_current()
-        product_settings = ai.settings.get(cls.settings_key, {})
+        product_settings = ai_settings.get(cls.settings_key, {})
 
         return cls(
             plugin=Plugins(request, cls.plugin_config),
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -40,7 +40,10 @@ class Routes:
 
 @dataclass
 class Settings:
+    """Product specific settings."""
+
     groups_enabled: bool = False
+    """Is the course groups feature enabled"""
 
     product_settings: InitVar[Dict] = field(default_factory=dict)
 
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -42,10 +42,10 @@ class Routes:
 class Settings:
     """Product specific settings."""
 
+    product_settings: InitVar[Dict]
+
     groups_enabled: bool = False
     """Is the course groups feature enabled"""
-
-    product_settings: InitVar[Dict] = field(default_factory=dict)
 
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -1,7 +1,8 @@
 """Core models of the product."""
 
-from dataclasses import dataclass
+from dataclasses import InitVar, dataclass
 from enum import Enum
+from typing import Dict, Optional
 
 from lms.product.plugin import PluginConfig, Plugins
 
@@ -35,13 +36,26 @@ class Routes:
 
 
 @dataclass
+class Settings:
+    groups_enabled: bool = False
+
+    product_settings: InitVar[Dict] = field(default_factory=dict)
+
+    def __post_init__(self, product_settings):
+        self.groups_enabled = product_settings.get("groups_enabled", False)
+
+
+@dataclass
 class Product:
     """The main product object which is passed around the app."""
 
     plugin: Plugins
+    settings: Settings
     plugin_config: PluginConfig = PluginConfig()
     route: Routes = Routes()
     family: Family = Family.UNKNOWN
+    settings_key: Optional[str] = None
+    """Key in the ai.settings dictionary that holds the product specific settings"""
 
     # Accessor for external consumption
     Family = Family
@@ -49,5 +63,10 @@ class Product:
     @classmethod
     def from_request(cls, request):
         """Create a populated product object from the provided request."""
+        ai = request.find_service(name="application_instance").get_current()
+        product_settings = ai.settings.get(cls.settings_key, {})
 
-        return cls(plugin=Plugins(request, cls.plugin_config))
+        return cls(
+            plugin=Plugins(request, cls.plugin_config),
+            settings=Settings(product_settings=product_settings),
+        )
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -208,6 +208,7 @@ class JSConfig:
         self._config.update(
             {
                 "mode": JSConfig.Mode.FILE_PICKER,
+                # Info about the product we are currently running in
                 "product": self._get_product_info(),
                 "filePicker": {
                     "formAction": form_action,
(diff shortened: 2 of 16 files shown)


lms/resources/_js_config/__init__.py
```diff
@@ -418,6 +419,35 @@ def _config(self):
...
+            "features": {
+                "groups": self._request.product.settings.groups_enabled,
+            },
+            # List of API endpoints we proxy for this product
+            "api": {},
+        }
+
+        # This should be abstracted as "lms_course_id" or similar in either
+        # LTIParams or as part of product.
+        api_course_id = self._request.lti_params.get(
```
marcospri, line 436: This is what I want to tackle in a second refactor.

We have this wart here but D2L will also introduce "custom_d2l_user_id".

For D2L I'd like that to be named "lms_user_id" in the configuration and unify a way of accessing:

```
lms_user_id 
lms_course_id
...
```

that's LMS agnostic.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -164,6 +160,35 @@ import { createContext } from 'preact';
  * @prop {Record<string,any>} [values]
  */
 
+/**
+ * Each product type known
+ * @typedef {'BlackbaudK12'|'BlackboardLearn'|'canvas'|'desire2learn'|'moodle'|'sakai'| 'schoology' | 'unknown'} ProductFamily
```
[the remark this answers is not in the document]
jon-betts:
```suggestion
 * @typedef {'BlackbaudK12' | 'BlackboardLearn' | 'canvas' | 'desire2learn' | 'moodle' | 'sakai' | 'schoology' | 'unknown'} ProductFamily
```

Do we want to be fussy on this side about this? It seems like another thing to forget / have to maintain as we add new products.

lms/resources/_js_config/__init__.py
```diff
@@ -208,6 +208,7 @@ def enable_file_picker_mode(self, form_action, form_fields):
         self._config.update(
             {
                 "mode": JSConfig.Mode.FILE_PICKER,
+                "product": self._get_product_info(),
```
jon-betts, line 211: It would be nice if we explained what this is for? I can't tell looking at this what it does or why we need it.

lms/resources/_js_config/__init__.py
jon-betts: Yeah, the params are a bit tricky as you need the params to get the product, but you also need the product to get the params. I wonder if we need a two phase thing where we:

 * Get basic params
 * Get the product
 * Add more product params?
 
 Unfortunately we use the presence of things like the canvas custom fields to tip us off it's canvas. I'm sure theres a way around it.

lms/resources/_js_config/file_picker_config.py
```diff
@@ -31,14 +25,6 @@ def blackboard_config(cls, request, application_instance):
                 ),
             }
 
-        if groups_enabled:
-            config["listGroupSets"] = {
-                "authUrl": auth_url,
-                "path": request.route_path(
-                    "blackboard_api.courses.group_sets.list", course_id=course_id
-                ),
-            }
```
jon-betts, line 40: Was this only used for file picker config or was this a mistake?

I'm wondering if something has been lost if this is file picker config to remove it from the file picker config file and put it unlabeled as such in the main one? Could we do with an extra `common_config()` or something?

lms/resources/_js_config/__init__.py
```diff
@@ -418,6 +419,35 @@ def _config(self):
...
+            },
+            # List of API endpoints we proxy for this product
+            "api": {},
+        }
+
+        # This should be abstracted as "lms_course_id" or similar in either
+        # LTIParams or as part of product.
+        api_course_id = self._request.lti_params.get(
+            "custom_canvas_course_id"
+        ) or self._request.lti_params.get("context_id")
```
jon-betts, line 438: ^ This section should be nested inside the if below. We don't use the value outside the if

lms/resources/_js_config/file_picker_config.py
jon-betts: If this isn't picker stuff, that's great, because having two definitions of "enabled" here isn't very nice.

lms/static/scripts/frontend_apps/config.js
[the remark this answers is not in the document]
marcospri: This is not used right now so maybe it's probably best to leave it out.

We could revisit it and then expose more sane values. These happen to be tool_consumer_version value verbatim from the LMS.

lms/resources/_js_config/__init__.py
```diff
@@ -418,6 +419,35 @@ def _config(self):
...
         return config
 
+    def _get_product_info(self):
+        product = self._request.product
+
+        product_info = {
+            "family": product.family,
+            "features": {
+                "groups": self._request.product.settings.groups_enabled,
+            },
```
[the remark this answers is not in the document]
marcospri: Yeah, much better :+1:

lms/resources/_js_config/file_picker_config.py
marcospri: I reckon this is not "filepicker" related in the sense used by the backend in this file.

lms/static/scripts/frontend_apps/components/FilePickerApp.js
```diff
@@ -90,12 +90,13 @@ export default function FilePickerApp({ onSubmit }) {
   const submitButton = /** @type {{ current: HTMLInputElement }} */ (useRef());
   const {
     api: { authToken },
-    filePicker: { deepLinkingAPI, formAction, formFields, blackboard, canvas },
+    product: { settings },
```
marcospri, line 93: Is it possible to destructure `settings.groupsEnable` directly here as 
`enableGroupConfig`?

robertknight: It is possible with `product: { settings: { groupsEnable: enableGroupConfig } }`. You might find it more readable to do the destructuring in multiple stages though, ie do one or two levels in the first statement, then the rest in a separate statement.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -142,12 +142,12 @@ export default function GroupConfigSelector({
 
   const {
     api: { authToken },
-    filePicker: { blackboard, canvas },
+    product: { api },
```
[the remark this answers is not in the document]
robertknight: Yes. This is not something we ever do much, but the general syntax is `const { level1: { level2: { innerLevel: localAlias }}}` where `localAlias` is the name of the binding that is then available locally.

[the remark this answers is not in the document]
robertknight: Note that ES imports, although they look similar to destructuring, don't support nesting and use `as` rather than `:` for aliasing.

lms/product/product.py
```diff
@@ -27,27 +28,50 @@ def _missing_(cls, _value):
...
 
+    list_group_sets: Optional[str] = None
+    """List available group sets. Takes a course_id parameter"""
+
+
+@dataclass
+class Settings:
+    """Product specific settings."""
+
+    product_settings: InitVar[Dict]
```
marcospri, line 45: `InitVar` can't have a `default_factory`

No need for a default in any case, just moved it to the top.

lms/resources/_js_config/file_picker_config.py
```diff
@@ -11,17 +11,11 @@ class FilePickerConfig:
...
-        )
 
         auth_url = request.route_url(Blackboard.route.oauth2_authorize)
         course_id = request.lti_params.get("context_id")
 
-        config = {
-            "enabled": files_enabled,
-            "groupsEnabled": groups_enabled,
-        }
+        config = {"enabled": files_enabled}
```
marcospri, line 18: `enabled` inside `settings.canvas` means "filesEnabled". Will get to that at some point.

jon-betts (approved):
Yeah I think this is looking good.

I think the way I've come to peace with it is, groups being enabled is something that informs the file picker, but isn't specific to it. So it makes sense to do this outside.

Groups are, or at least could be, used for more than the file picker.

lms/static/scripts/frontend_apps/config.js
```diff
@@ -164,6 +160,29 @@ import { createContext } from 'preact';
  * @prop {Record<string,any>} [values]
  */
 
+/**
+ * Settings or the current product
```
robertknight, line 164:
```suggestion
 * Settings of the current product
```

lms/static/scripts/frontend_apps/config.js
```diff
@@ -164,6 +160,29 @@ import { createContext } from 'preact';
...
+
+/**
+ * API endpoints for the product
+ *
+ * @typedef ProductAPI
+ * @prop {APICallInfo} listGroupSets
+ */
+
+/**
+ * Information for the current product
```
robertknight, line 178:
```suggestion
 * Information about the current product
```

robertknight (approved):
LGTM. I do have some misgivings about how generic the term "product" is, but the change in principle is good, and we can bikeshed that another time.

PR #4593, opened 2022-10-27
Always deploy the correct Docker image version

seanh:
Fixes https://github.com/hypothesis/lms/issues/4588. For details see: https://github.com/hypothesis/via/pull/814

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,11 +31,11 @@ jobs:
       Environment: qa
       Region: us-west-1
       Operation: deploy
-      Version: latest
+      Version: ${{ needs.create-image.outputs.docker_tag }}
     secrets: inherit
 
   prod-us-west-1:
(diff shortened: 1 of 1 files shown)


indigobravo:
Heads up on this one. Conflicts will need to be fixed.


I removed the job level concurrency options from deploy.yml for LMS. It has been reverted back to a known working state. This was to help fix the problem Lyza was running into. https://github.com/hypothesis/lms/pull/4595

indigobravo (changes requested):
Heads up on this one. Conflicts will need to be fixed.


I removed the job level concurrency options from deploy.yml for LMS. It has been reverted back to a known working state. This was to help fix the problem Lyza was running into. https://github.com/hypothesis/lms/pull/4595

seanh:
Good catch and thanks for that revert, it was the right decision.  I'll rebase this next week...

seanh:
Rebased

PR #4598, opened 2022-10-28
Test FDW creation and GHA password masking

marcospri:
Main objective here is see how GHA handles secrets 

Once that's checked this needs:

- Integration with the functest
- Readme for the new tasks
- Proper handling of the new variables, env var -> pyramid_setting -> jinja variable.

### DIFF
--- a/lms/sql_tasks/tasks/fdw/create_from_scratch/00_create.jinja2.sql
+++ b/lms/sql_tasks/tasks/fdw/create_from_scratch/00_create.jinja2.sql
@@ -1,0 +1,15 @@
+DROP SERVER IF EXISTS {{h_fdw_server_name}} CASCADE;
+CREATE SERVER {{h_fdw_server_name}} FOREIGN DATA WRAPPER postgres_fdw 
+    OPTIONS (host '{{h_fdw_host}}', port '{{h_fdw_port}}', dbname '{{h_fdw_dbname}}');
+
+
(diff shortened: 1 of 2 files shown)


lms/sql_tasks/tasks/fdw/create_from_scratch/00_create.jinja2.sql
```diff
@@ -0,0 +1,17 @@
...
+    SERVER {{h_fdw_server_name}}
+    OPTIONS (user '{{h_fdw_user}}', password '{{h_fdw_password}}');
+
+DROP SCHEMA IF EXISTS h CASCADE;
+-- Keep foreign tables on its own schema
+CREATE SCHEMA h;
+
+{% for schema_name, table_name in h_fdw_tables %}
+IMPORT FOREIGN SCHEMA {{schema_name}} LIMIT TO ({{table_name}}) FROM SERVER {{h_fdw_server_name}} INTO h;
+{% endfor %}
```
jon-betts, line 17: I think it might be nice to have the tables here. They don't seem like a variable in the same way as other things which change from environment to environment.

If I tweak a view to need a new table, I'd probably come here first, and then have to bounce out to the runner binary, which seems an unnatural leap.

marcospri: Yeah, good point, it's something very local to all this sql business. Makes sense to keep it here :+1:

jon-betts (approved):
This is all WIP stuff. We're running this expecting to break to test what GHA does with passwords.

PR #4600, opened 2022-10-28
Quote variables to allow any valid characters

marcospri:


jon-betts (approved):
This could with some tweaks as described here: https://github.com/hypothesis/lms/pull/4601/files

But with those it works

PR #4603, opened 2022-10-28
Hide lines marked as secret in SQL tasks

marcospri:
This is a pretty silly method but I'm convinced the real solution should be done at the variable level (which contains the secrets) rather then at the script/line level.

I added a note to revisit this for this project but with this I'm sure if we remember something is a secret will stick a `-- secret` to it without worrying to much about messing up the formatting.

### DIFF
--- a/lms/sql_tasks/sql_query.py
+++ b/lms/sql_tasks/sql_query.py
@@ -1,3 +1,4 @@
+import re
 import textwrap
 from dataclasses import dataclass
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


lms/sql_tasks/sql_query.py
```diff
@@ -58,3 +59,14 @@ def dump(self, indent=""):
...
+    @staticmethod
+    def _clean_query(query: str) -> str:
+        """Clean any of the query's lines marked as sensitive."""
+        lines = query.split("\n")
+        for i, line in enumerate(lines):
+            # Heavy handed here, we much rather have false positives than negatives
+            if "secret" in line.lower():
+                lines[i] = re.sub(r"[^\s]", "*", line)
+
+        return "\n".join(lines)
```
jon-betts, line 72:
```suggestion
        return "\n".join([
            re.sub(r"[^\s]", "*", line) if "secret" in line.lower() else line
            for line in query.split("\n")
        ])
```

Generally enumerate and index on a list suggest a missing list comprehension?

jon-betts (approved):
Looks good to me. I think we can make this a bit neater with a list comprehension though.

PR #4613, opened 2022-11-01
Fix a bug where names with '-' in them result in invalid SQL

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/3

This doesn't happen locally because our users don't have any fancy chars.


marcospri (approved):
"LGTM"

PR #4620, opened 2022-11-03
Add user activity and calculate teacher annotations from it

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/4

This adds a count of annotations by users to the organization activity report. As a side effect it also adds role based activity for users in general, which we will use in reports coming soon.


lms/sql_tasks/tasks/report/create_from_scratch/04_activity_counts/03_organization_activity/01_create_view.sql
```diff
@@ -119,16 +141,23 @@ CREATE MATERIALIZED VIEW report.organization_activity AS (
...
+
+            UNION ALL
 
             SELECT
                 period, timescale, organization_id, role,
                 0, active, 0
             FROM active
 
-            UNION
+            UNION ALL
```
jon-betts, line 160: We can use UNION ALL's here, which prevents Postgres from having to check each row for uniqueness. By the way we've constructed this here we know no rows can be duplicates, as each has something different in it.

marcospri (approved):

PR #4639, opened 2022-11-11
D2L API authentication

marcospri:
Adds the required service and views for D2L oauth2 authentication.

On top of that adds a method and view to get the list of group categories so we can test the whole oauth2 process using the UI.

## Testing 

- `make devdata` to start with.
- Go to https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2120/View as a teacher
- Select any type of document and check the group assignment
- When prompted to authorize, deny the request (to test https://github.com/hypothesis/lms/pull/4639/files#diff-acd99c9193ec519d3248994780ce552fef977e8915329f06f5d300e18aee5a5eR73)
- Try again and authorize the request this time
- You'll get a group category in the drop down coming from the D2L API.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -63,6 +63,8 @@ good-names=
 
     # ApplicationInstance
     ai,
+    # Organization unit (D2L)
+    ou,
 
 [REPORTS]
 output-format=colorized
--- a/.pylintrc
+++ b/.pylintrc
@@ -63,8 +63,6 @@ good-names=
 
     # ApplicationInstance
     ai,
-    # Organization unit (D2L)
-    ou,
 
 [REPORTS]
 output-format=colorized
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -36,11 +36,11 @@ class D2LAPIClient:
         """
         self._api.refresh_access_token()
 
-    def course_group_sets(self, ou):
+    def course_group_sets(self, org_unit):
         """
-        Get the group categories of an org unit (OU).
+        Get the group categories of an org unit.
 
         https://docs.valence.desire2learn.com/res/groups.html#get--d2l-api-lp-(version)-(orgUnitId)-groupcategories-
         """
-        response = self._api.request("GET", f"/{ou}/groupcategories/")
+        response = self._api.request("GET", f"/{org_unit}/groupcategories/")
         return D2LGroupSetsSchema(response).parse()
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -1,0 +1,46 @@
+from marshmallow import EXCLUDE, fields
+
+from lms.validation._base import RequestsResponseSchema
+
+
+class D2LGroupSetsSchema(RequestsResponseSchema):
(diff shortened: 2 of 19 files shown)


lms/services/d2l_api/client.py
```diff
@@ -0,0 +1,46 @@
...
+    def refresh_access_token(self):
+        """
+        Refresh the current user's access token in the DB.
+
+        :raise services.ExternalRequestError: if something goes wrong with the
+            refresh token request
+        """
+        self._api.refresh_access_token()
+
+    def course_group_sets(self, ou):
```
marcospri, line 39: Added this one so it's easier to test the whole authentication flow.

lms/views/api/refresh.py
```diff
@@ -23,3 +24,7 @@ def get_refreshed_token_from_canvas(self):
     def get_refreshed_token_from_blackboard(self):
         blackboard_api_client = self.request.find_service(name="blackboard_api_client")
         blackboard_api_client.refresh_access_token()
+
+    @view_config(route_name="d2l_api.oauth.refresh")
+    def get_refreshed_token_from_d2l(self):
```
marcospri, line 29: Same here about generalizing. We'll have to at least give it a go.

lms/services/d2l_api/client.py
```diff
@@ -0,0 +1,46 @@
...
+    class Meta:
+        unknown = EXCLUDE
+
+    id = fields.Int(required=True, data_key="GroupCategoryId")
+    name = fields.Str(required=True, data_key="Name")
+
+
+class D2LAPIClient:
+    def __init__(self, basic_client, request):
+        self._api = basic_client
```
jon-betts, line 18: We could expose this directly and let external callers use it directly? This would avoid having to duplicate the `get_token` and `refresh_access_token` methods. As currently they don't change it in anyway.

I know this breaks the one level rule or whatever it's called, but I'm a mad rebel.

jon-betts (approved):
Yep this works. I've made some mostly stylistic suggestions throughout.

I don't think it's a massive hassle to type `organization_unit` so I think an exception for that isn't the way to go in our linting setup. I suspect what we actually want is to translate it into non-D2L speak at the API boundary anyway and leave it as `course_id` or something similar.

In more substantial feedback the refreshing stuff doesn't feel right. That seems like something that should be abstracted into the product object. This would also remove the need to have an `outh2_authorize` route, because it would always be the same, and do something different based on product. We might need a `auth_method=oauth2` or something? In any case, that can be a follow up.

.pylintrc
```diff
@@ -63,6 +63,8 @@ good-names=
 
     # ApplicationInstance
     ai,
+    # Organization unit (D2L)
+    ou,
```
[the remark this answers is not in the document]
marcospri: I went for `org_unit` which the D2L docs use :+1: 

I kept that name in the client and used `course_id` elsewhere.

marcospri:
Added the suggestions about auth here to the project board. I rather tackle those as part of the refactor that does the same changes for Canvas and Blackboard.

PR #4640, opened 2022-11-11
D2L details in the admin pages

marcospri:
Make room for D2L setting on the admin pages.

Took the opportunity I was around this code to tackle: https://github.com/hypothesis/product-backlog/issues/1399 

# Testing

(requires https://github.com/hypothesis/devdata/pull/65)

- Ahead to http://localhost:8001/admin/instance/id/106/
- Play with the settings for D2L

### DIFF
--- a/lms/models/application_settings.py
+++ b/lms/models/application_settings.py
@@ -1,3 +1,6 @@
+import base64
+from typing import Optional
+
 from sqlalchemy.ext.mutable import MutableDict
 
 
@@ -58,6 +61,44 @@ class ApplicationSettings(MutableDict):
         :param value: The value to set
         """
         super().setdefault(group, {})[key] = value
+
+    def set_secret(self, aes_service, group, key, value: str):
+        """
+        Store a setting as a secret.
+
+        Store the setting AES encrypted.
+
+        :param aes_service: AESService to encrypt the string
+        :param group: The name of the group of settings
+        :param key: The key in that group
+        :param value: The value to set
+        """
+        aes_iv = aes_service.build_iv()
+        value = aes_service.encrypt(aes_iv, value)
+
+        # Store both the setting and the IV
+        # We can't store the bytes directly in JSON so we store it as base64
+        super().setdefault(group, {})[key] = base64.b64encode(value).decode("utf-8")
+        super().setdefault(group, {})[f"{key}_aes_iv"] = base64.b64encode(
+            aes_iv
+        ).decode("utf-8")
+
+    def get_secret(self, aes_service, group, key) -> Optional[str]:
+        """
+        Get a secret setting.
+
+        :param aes_service: AESService to decrypt the string
+        :param group: The name of the group of settings
+        :param key: The key in that group
+        """
+        value = super().get(group, {}).get(key)
+        if not value:
+            return None
+
+        aes_value = base64.b64decode(value)
+        aes_iv = base64.b64decode(super().get(group, {}).get(f"{key}_aes_iv"))
+
+        return aes_service.decrypt(aes_iv, aes_value).decode("utf-8")
 
     def __repr__(self):
         return f"{self.__class__.__name__}({super().__repr__()})"
--- a/lms/product/d2l/product.py
+++ b/lms/product/d2l/product.py
@@ -8,3 +8,5 @@ class D2L(Product):
     """A product for D2L specific settings and tweaks."""
 
     family: Product.Family = Product.Family.D2L
+
+    settings_key = "desire2learn"
--- a/lms/services/aes.py
+++ b/lms/services/aes.py
@@ -38,6 +38,9 @@ class AESService:
         # Using the cryptographic secure Cryptodome.Random instead of the stlib random package.
         return Random.new().read(AES.block_size)
 
+    class AESSecret:
+        """Dummy type to identify strings that must be encrypted."""
+
 
 def factory(_context, request):
     return AESService(secret=request.registry.settings["aes_secret"])
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -106,14 +106,9 @@
 
     <fieldset class="box">
         <legend class="label has-text-centered">D2L settings (EXPERIMENTAL)</legend>
-        {{ macros.form_text_field(request, "Developer key", "developer_key", field_value=instance.developer_key) }}
-        {{ macros.form_text_field(
-            request,
-            "Developer secret",
-            "developer_secret",
-            placeholder="*" * 25 if instance.developer_secret else "")
-        }}
-        {{ settings_checkbox("Groups enabled", "d2l", "groups_enabled") }}
+        {{ settings_text_field("API Client ID", "desire2learn", "client_id") }}
+        {{ settings_text_field("API Client secret", "desire2learn", "client_secret") }}
+        {{ settings_checkbox("Groups enabled", "desire2learn", "groups_enabled") }}
     </fieldset>
 
     <fieldset class="box">
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -100,6 +100,19 @@
         {{ settings_checkbox("Groups enabled", "blackboard", "groups_enabled") }}
         {{ settings_checkbox("Files enabled", "blackboard", "files_enabled") }}
     </fieldset>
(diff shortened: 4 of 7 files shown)


lms/templates/admin/instance.html.jinja2
```diff
@@ -84,10 +84,13 @@
     <fieldset class="box">
         <legend class="label has-text-centered">Canvas settings</legend>
         {{ macros.disabled_text_field("API domain", instance.custom_canvas_api_domain) }}
-        {{ macros.form_text_field(request, "Developer key", "developer_key", field_value=instance.developer_key) }}
+        {# Note the mismatch betwween canvas nomenclature and ours #}
+        {# developer id -> developer key #}
+        {# developer key -> developer secret #}
+        {{ macros.form_text_field(request, "Developer ID", "developer_key", field_value=instance.developer_key) }}
```
marcospri, line 90: I can see how these are confusing

lms/models/application_settings.py
```diff
@@ -59,6 +62,44 @@ def set(self, group, key, value):
         """
         super().setdefault(group, {})[key] = value
 
+    def set_secret(self, aes_service, group, key, value: str):
```
marcospri, line 65: I reckon this belongs here, not in the admin pages around the code that deals with settings.

lms/product/d2l/product.py
```diff
@@ -8,3 +8,5 @@ class D2L(Product):
     """A product for D2L specific settings and tweaks."""
 
     family: Product.Family = Product.Family.D2L
+
+    settings_key = "desire2learn"
```
marcospri, line 12: This is the string we have not for the family, we have to do the mapping anyway so we could go for "d2l" instead. 

We are not typing this very often in any case.

lms/services/aes.py
```diff
@@ -38,6 +38,9 @@ def build_iv():
         # Using the cryptographic secure Cryptodome.Random instead of the stlib random package.
         return Random.new().read(AES.block_size)
 
+    class AESSecret:
```
marcospri, line 41: I'm using this in the settings definition. 

Arguably this could be just "secret" over there but this is cheap as well.

lms/templates/admin/instance.html.jinja2
```diff
@@ -31,6 +31,17 @@
 {% endmacro %}
 
 
+{% macro settings_secret_field(label, setting, sub_setting, field_name, default='') %}
+    {% call macros.field_body(label) %}
+        <input
+            value="{{ "*" * 25 if instance.settings.get(setting, sub_setting, default) else '' }}"
```
marcospri, line 37: This is how we are presenting other secrets.

lms/templates/admin/instance.html.jinja2
```diff
@@ -100,6 +114,14 @@
         {{ settings_checkbox("Groups enabled", "blackboard", "groups_enabled") }}
         {{ settings_checkbox("Files enabled", "blackboard", "files_enabled") }}
     </fieldset>
+
+    <fieldset class="box">
+        <legend class="label has-text-centered">D2L settings (EXPERIMENTAL)</legend>
```
marcospri, line 119: I have to remember to ping #support about this being there now.

tests/unit/lms/models/application_settings_test.py
```diff
@@ -43,6 +44,17 @@ def test_set(self, application_settings, group, key, value, expected_value):
 
         assert application_settings[group][key] == expected_value
 
+    def test_secrets(self, application_settings, aes_service):
```
marcospri, line 47: I reckon for this type of functionality that deals with values (that need to be base64 -> encoded) this type of test is preferable to mocking aes_service, base64, etc.

Other tests in the module are similar too.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -419,6 +422,31 @@ def test_update_instance_saves_ai_settings(
...
+        pyramid_request.params[f"{setting}.{sub_setting}"] = "SECRET"
+        # This fixture returns a real AI, let's use a mock for this test
+        application_instance_service.get_by_consumer_key.return_value = create_autospec(
+            ApplicationInstance
+        )
+
+        AdminApplicationInstanceViews(pyramid_request).update_instance()
+
+        ai = application_instance_service.get_by_consumer_key.return_value
+        ai.settings.set_secret.assert_called_once_with(
```
marcospri, line 446: Here actually asserting what we are expecting gets called.

lms/models/application_settings.py
```diff
@@ -59,6 +62,44 @@ def set(self, group, key, value):
...
+        value = aes_service.encrypt(aes_iv, value)
+
+        # Store both the setting and the IV
+        # We can't store the bytes directly in JSON so we store it as base64
+        super().setdefault(group, {})[key] = base64.b64encode(value).decode("utf-8")
+        super().setdefault(group, {})[f"{key}_aes_iv"] = base64.b64encode(
+            aes_iv
+        ).decode("utf-8")
+
+    def get_secret(self, aes_service, group, key) -> Optional[str]:
```
marcospri, line 86: This is not used but  I reckon the PR would be incomplete without the get part.

jon-betts: Yep. It would be very strange to be able to encrypt, but not decrypt the values. Also round trip testing would be impossible.

tests/unit/lms/models/application_settings_test.py
```diff
@@ -43,6 +44,17 @@ def test_set(self, application_settings, group, key, value, expected_value):
...
+        application_settings.set_secret(aes_service, "GROUP", "KEY", "VERY SECRET")
+
+        assert (
+            application_settings.get_secret(aes_service, "GROUP", "KEY")
+            == "VERY SECRET"
+        )
```
jon-betts, line 53: Do we want to make any assertions to ensure we called the AES service and didn't store the content in the clear? I think this would pass if it didn't encrypt anything?

I think a decent pattern for this is to have three tests:

 * One which sets a value and compares the results with baked values (you might have to do something to fix the IV?)
 * One with baked values to see it decodes
 * One round trip like this

Of these the last two probably prove most of what we want and are easy.

This has the nice effect of protecting you from arbitrary changes which still round trip, but make stored values unable to be decrypted. On the other hand perhaps we've proved the AES service works and we need to ensure we call it correctly?

So I guess, either we should lean into mockist tests and mock the service, or do more integrated tests and use real values.

lms/services/aes.py
```diff
@@ -38,6 +38,9 @@ def build_iv():
...
+        """Dummy type to identify strings that must be encrypted."""
```
jon-betts, line 42: This is really far from where it's used. I'm not sure this makes sense on the service. 

I think this might be better defined where it's used.

lms/views/admin/application_instance.py
```diff
@@ -301,10 +303,13 @@ def update_instance(self):
...
 
         for setting, sub_setting, setting_type in (
-            ("canvas", "sections_enabled", bool),
-            ("canvas", "groups_enabled", bool),
             ("blackboard", "files_enabled", bool),
             ("blackboard", "groups_enabled", bool),
+            ("canvas", "sections_enabled", bool),
+            ("canvas", "groups_enabled", bool),
+            ("desire2learn", "client_id", str),
+            ("desire2learn", "client_secret", AESService.AESSecret),
```
jon-betts, line 311: I think we can define a local value here for this? Something like `AES_SECRET = object()`?

lms/views/admin/application_instance.py
```diff
@@ -317,11 +322,18 @@ def update_instance(self):
...
+                    continue
+
+                ai.settings.set_secret(self._aes_service, setting, sub_setting, value)
+
             else:
                 assert setting_type == str
                 value = value.strip() if value else None
-
-            ai.settings.set(setting, sub_setting, value)
+                ai.settings.set(setting, sub_setting, value)
```
jon-betts, line 336: Maybe group the `settings.set` type things together with the secrets after?

jon-betts (approved):
Yep, I've given this a shot locally and it seems to work.

I think there's an opportunity to tighten up the tests a bit. The way they are written at the moment, we don't actually check we encrypt anything. We could store it in the clear and have the tests still pass.

I think we should remove the constant from the AES service. It's only used in the admin pages code, and should probably be defined there. It's very mysterious as it is right now.

PR #4659, opened 2022-11-17
D2L grouping plugin

marcospri:
Needs

- Migration on https://github.com/hypothesis/lms/pull/4660
- devdata from https://github.com/hypothesis/devdata/pull/69
- and... this fix in H https://github.com/hypothesis/h/pull/7712

---

Implements the grouping plugin for D2L. We can get away with just two API method endpoints and a query to our grouping service (ie getting them from the DB).

### Testing 

Using the asssigments in https://aunltd.brightspacedemo.com/d2l/le/content/6782/Home?itemIdentifier=D2L.LE.Content.ContentObject.ModuleCO-2122

#### Happy path

I created assignment for LTI 1.1 and 1.3 for completeness but there's not differences in the code of this PR about any differences between them.

- Launch as an instructor
- You might need to visit https://localhost:48001/lti/1.3/oidc and approve a fake SSL certificate

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2132/View
https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2128/View

you should get the groups selector.

- Same assignments as a student, should only get one group. 
- Make an annotation as the student.
- Refresh the assignment as a teacher, the student who annotated will appear in the student selector and switching to him will set the "groups selector" focus on the right group.

#### Failures modes

There are different assignments here https://aunltd.brightspacedemo.com/d2l/le/content/6782/Home?itemIdentifier=D2L.LE.Content.ContentObject.ModuleCO-2122

to test failures modes. In this PR you can test that errors are handled in the backend, they don't "explode" but just produce a generic error message.

There's separate PR for the error dialogs here: https://github.com/hypothesis/lms/pull/4662

## ~~TODO~~

- [x] Split migration to another PR
- [x] Split error codes to another PR
- [x] Test in LTI1.3
- [x] Test in LTI1.1
- [x] Test error codes

### DIFF
--- a/lms/product/d2l/_plugin/grouping.py
+++ b/lms/product/d2l/_plugin/grouping.py
@@ -1,0 +1,80 @@
+from enum import Enum
+
+from lms.models import Grouping
+from lms.product.plugin.grouping_service import GroupError, GroupingServicePlugin
+from lms.services import D2LAPIClient
+from lms.services.exceptions import ExternalRequestError
+
+
+class ErrorCodes(str, Enum):
+    """Error codes that the FE is going to check for."""
+
+    GROUP_SET_NOT_FOUND = "d2l_group_set_not_found"
+    GROUP_SET_EMPTY = "d2l_group_set_empty"
+    STUDENT_NOT_IN_GROUP = "d2l_student_not_in_group"
+
+
+class D2LGroupingPlugin(GroupingServicePlugin):
+    """A plugin which implements D2L specific grouping functions."""
+
+    group_type = Grouping.Type.D2L_GROUP
+    sections_type = None  # We don't support sections in D2L
+
+    def __init__(self, d2l_api, api_user_id):
+        self._d2l_api = d2l_api
+        self._api_user_id = api_user_id
+
+    def get_groups_for_learner(self, _svc, course, group_set_id):
+        if learner_groups := self._d2l_api.group_set_groups(
+            course.lms_id, group_set_id, self._api_user_id
+        ):
+            return learner_groups
+
+        raise GroupError(ErrorCodes.STUDENT_NOT_IN_GROUP, group_set=group_set_id)
+
+    def get_groups_for_grading(
+        self, svc, course, group_set_id, grading_student_id=None
+    ):
+        return svc.get_course_groupings_for_user(
+            course,
+            grading_student_id,
+            type_=self.group_type,
+            group_set_id=int(group_set_id),
+        )
+
+    def get_groups_for_instructor(self, _svc, course, group_set_id):
+        try:
+            groups = self._d2l_api.group_set_groups(course.lms_id, group_set_id)
+        except ExternalRequestError as exc:
+            if exc.status_code == 404:
+                raise GroupError(
+                    ErrorCodes.GROUP_SET_NOT_FOUND, group_set=group_set_id
+                ) from exc
+
+            raise
+
+        if not groups:
+            raise GroupError(ErrorCodes.GROUP_SET_EMPTY, group_set=group_set_id)
+
+        return groups
+
+    @staticmethod
+    def _get_api_user_id(user_id: str):
+        """Get the user id to use with the API from the LTI user_id."""
+        id_parts = user_id.split("_")
+        if len(id_parts) > 1:
+            # D2L seems to send the LTI1.3 user id (sub) as the expected
+            # UUID format *except* they attach the API user id at the end separated by "_"
+            api_user_id = id_parts[1]
+        else:
+            # In LTI1.1 they just sent the whole user_id
+            api_user_id = user_id
+
+        return api_user_id
+
+    @classmethod
+    def factory(cls, _context, request):
+        return cls(
+            d2l_api=request.find_service(D2LAPIClient),
+            api_user_id=cls._get_api_user_id(request.lti_user.user_id),
+        )
--- a/lms/product/d2l/_plugin/grouping.py
+++ b/lms/product/d2l/_plugin/grouping.py
@@ -60,17 +60,18 @@ class D2LGroupingPlugin(GroupingServicePlugin):
 
     @staticmethod
     def _get_api_user_id(user_id: str):
-        """Get the user id to use with the API from the LTI user_id."""
-        id_parts = user_id.split("_")
-        if len(id_parts) > 1:
-            # D2L seems to send the LTI1.3 user id (sub) as the expected
-            # UUID format *except* they attach the API user id at the end separated by "_"
-            api_user_id = id_parts[1]
-        else:
-            # In LTI1.1 they just sent the whole user_id
-            api_user_id = user_id
+        """
+        Get the user id to use with the API from the LTI user_id.
 
-        return api_user_id
+        D2L user_id seem to follow two schemas:
+
+        LONG-UUID-HEX_ID
+        shorthumanreadablename_ID
+
+        Note that in the second type we've seen names that include "_"
+        so we don't take the second part of the string but the last.
+        """
+        return user_id.split("_")[-1]
 
     @classmethod
     def factory(cls, _context, request):
--- a/lms/product/d2l/_plugin/grouping.py
+++ b/lms/product/d2l/_plugin/grouping.py
@@ -58,24 +58,10 @@ class D2LGroupingPlugin(GroupingServicePlugin):
 
         return groups
 
-    @staticmethod
-    def _get_api_user_id(user_id: str):
-        """
-        Get the user id to use with the API from the LTI user_id.
-
-        D2L user_id seem to follow two schemas:
-
-        LONG-UUID-HEX_ID
-        shorthumanreadablename_ID
-
-        Note that in the second type we've seen names that include "_"
-        so we don't take the second part of the string but the last.
-        """
-        return user_id.split("_")[-1]
-
     @classmethod
     def factory(cls, _context, request):
+        d2l_api = request.find_service(D2LAPIClient)
         return cls(
-            d2l_api=request.find_service(D2LAPIClient),
-            api_user_id=cls._get_api_user_id(request.lti_user.user_id),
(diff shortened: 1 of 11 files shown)


lms/product/d2l/_plugin/grouping.py
```diff
@@ -0,0 +1,80 @@
...
+
+            raise
+
+        if not groups:
+            raise GroupError(ErrorCodes.GROUP_SET_EMPTY, group_set=group_set_id)
+
+        return groups
+
+    @staticmethod
+    def _get_api_user_id(user_id: str):
```
marcospri, line 62: Here's the biggest LMS specific wart in all of this.

I reckon the right solution is to abstract this inside LTIUser as `api_lti_user` or similar and handle it as a plugin or similar for Canvas at the same time.

Added a note to the board about that.

On this PR this is relatively hidden for the rest of the code. I have it here but you could argue it belongs closer to the API client.

In the canvas case one ID can't be derived from the other so we can't have a translation function in that case.

lms/product/d2l/_plugin/grouping.py
```diff
@@ -0,0 +1,80 @@
...
+            # In LTI1.1 they just sent the whole user_id
+            api_user_id = user_id
+
+        return api_user_id
+
+    @classmethod
+    def factory(cls, _context, request):
+        return cls(
+            d2l_api=request.find_service(D2LAPIClient),
+            api_user_id=cls._get_api_user_id(request.lti_user.user_id),
```
marcospri, line 79: Note we take this from LTIUser, not `lti_params` as this has to work in the context of an API request (not necessary in the LTI launch one).

marcospri, line 79: This being injected in the plugin avoids the need to change the interface.

lms/product/d2l/_plugin/grouping.py
```diff
@@ -0,0 +1,83 @@
...
+    def _get_api_user_id(user_id: str):
+        """
+        Get the user id to use with the API from the LTI user_id.
+
+        D2L user_id seem to follow two schemas:
+
+        LONG-UUID-HEX_ID
+        shorthumanreadablename_ID
+
+        Note that in the second type we've seen names that include "_"
```
marcospri, line 71: Found this out while testing in values from the DB.

jon-betts (commented):
I've only given this a once over, but it looks very neat and easy to follow. I wanted to put a comment on here as I think we can parallelise the rest of the review and maybe answering some questions / ideas.

I think we should add the `d2l_group_group` to the list of translations for the reporting stuff. This is a nice outcome of keeping the two together. I don't think anything would break as it stands, but we might end up with blank values. These don't power anything right now, but I think it's probably the way we want to go in future. Keep the changes and reporting in sync.

There was a question around where the user id translation code should go. Some candidates seem to be:

 * In the plugin where it is right now
 * In the D2L service
 * In a new thing somewhere close to the user object?
 
I think I'm missing some context to be able to give good input on that. My initial instinct is that putting it inside the service seems nice, because it means the caller doesn't have to worry about this quirk. As our API wrapping services job is to translate away all sorts of API specific quirks that seems like a good place. Although I think I was picking up on the idea this might be bigger than that? Would be good to think.

In any case, that shouldn't hold up the PR, because we can change it any time, but if it's something we can thrash out quickly and include here, that's a bonus.

jon-betts:
I've just run through the testing instructions with the following results:

**As admin (1.3)**

Visit: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2132/View
Get two group sets
The assignment works

**As student (1.3)**

Visit: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2132/View
It works!
I can see "Test Group 1"

**As admin (1.1)**

Visit: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2128/View

> Hypothesis received an invalid request. There were problems with these request parameters:
>form:
>   _schema

**Testing error scenarios**

1.3 scenarios get me "Hypothesis assignment isn't configured" - I'm probably missing the devdata here is all
1.1 scenarios get me the `form: _schema` error from above

So I suspect I'm missing the devdata, but the 1.1 seems like there might be a bug?

jon-betts:
I think both of these are a dev data issue, but we are failing non-gracefully in the 1.1 cases? So even though this is me not having the right setup. I think there might be a real issue, but I'm having trouble replicating it?

I tried executing: `DELETE FROM assignment WHERE id = 26;` and visiting: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2128/View, but now I'm seeing sensible stuff again (not configured)?

jon-betts:
**Error conditions**

I've tested the error conditions and I see:

 * Missing group (1.1 or 1.3): External request failed
 * Empty group set (1.1 or 1.3): There was a problem fetching this Hypothesis assignment
 
Which counts as handled, but no super helpful messages. I wonder if this is because `GroupError` is not a child of `SerializableError`? If it was I suspect we might get something more helpful out? Or at least give a space we could put nice messages. This is pre-existing, so not something which has to be in this PR.

**Student selector**

Annotate as student
Shows up in a group
Manually toggle as admin
See the annotation as you'd expect
Use the student selector at the top - **no change of group!**

Do we have a bug here?

jon-betts:
I missed the update in H. This fixes the group changing.

lms/product/d2l/_plugin/grouping.py
```diff
@@ -0,0 +1,81 @@
...
+                raise GroupError(
+                    ErrorCodes.GROUP_SET_NOT_FOUND, group_set=group_set_id
+                ) from exc
+
+            raise
+
+        if not groups:
+            raise GroupError(ErrorCodes.GROUP_SET_EMPTY, group_set=group_set_id)
+
+        return groups
```
jon-betts, line 59: My IDE is giving me a whinge because this is a "duplicated code fragment". It's actually not as the objects in play are different, but it's nearly exactly the same code as in Blackboard which is interesting.

tests/unit/lms/product/d2l/_plugin/grouping_test.py
```diff
@@ -0,0 +1,123 @@
...
+        # pylint: disable=protected-access
+        assert plugin._d2l_api == d2l_api_client
+
+    @pytest.fixture
+    def plugin(self, d2l_api_client):
+        return D2LGroupingPlugin(d2l_api_client, sentinel.api_user_id)
+
+    @pytest.fixture
+    def course(self):
+        return factories.Course()
```
jon-betts, line 123: :+1: These tests are very boring in the best possible way. Very clear and 1:1 with the code they test. No frills or confusing parts.

lms/services/d2l_api/client.py
```diff
@@ -44,3 +56,23 @@ def course_group_sets(self, org_unit):
...
+        https://docs.valence.desire2learn.com/res/groups.html#post--d2l-api-lp-(version)-(orgUnitId)-groupcategories-(groupCategoryId)-groups-
+        """
+        response = self._api.request(
+            "GET",
+            f"/{org_unit}/groupcategories/{group_category_id}/groups/",
+        )
+        groups = D2LGroupsSchema(response).parse()
+        # D2L doesn't return the group_set_id of the listed groups as other LMS
+        # but we know which one it is because we queried for it, inject it:
+        groups = [dict(group, group_set_id=group_category_id) for group in groups]
```
jon-betts, line 73: It's a little strange to redefine the object type here. I suspect they aren't different types though. Are we using a list comprehension to replace a loop here?

i.e. Could / should this be:

```suggestion
        for group in groups:
            group["group_set_id"] = group_category_id
```

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -28,6 +28,37 @@ def test_course_group_sets(
...
+    ):
+        d2l_groups_schema.parse.return_value = groups
+
+        response = svc.group_set_groups("COURSE_ID", "GROUP_SET")
+
+        basic_client.request.assert_called_once_with(
+            "GET", "/COURSE_ID/groupcategories/GROUP_SET/groups/"
+        )
+        D2LGroupsSchema.assert_called_once_with(basic_client.request.return_value)
+        assert response == [dict(values, group_set_id="GROUP_SET") for values in groups]
```
jon-betts, line 42: And having put a note about a list comprehension as a way of modifying a list above, I think it's good here.

I think brevity is probably better than speed in the tests.

The alternative mentioned below is some canned values.

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -28,6 +28,37 @@ def test_course_group_sets(
...
+    def test_group_set_groups_with_user_id(self, svc, d2l_groups_schema, groups):
+        d2l_groups_schema.parse.return_value = groups
+
+        response = svc.group_set_groups("COURSE_ID", "GROUP_SET", user_id=100)
+
+        assert response == [
+            dict(values, group_set_id="GROUP_SET")
+            for values in groups
+            if 100 in values["enrollments"]
+        ]
```
jon-betts, line 53: I wonder if canned values might be better here?

jon-betts (approved):
Yeah I've given this a good looking over and I think it looks good. Some boo-boos from me following the instructions, but other than that it's all very clean and easy to follow. I think the plugin structure helped with this.

There are lots of detailed testing and notes, but some highlights of the feedback:

* **Should**: Update SQL tasks to include the new group type
* Could: Move the user id extraction to the API client / have a chat about the right place
* Could: Try GroupError as SerializableError (if it doesn't immediately work give up?)
* _Dealers choice_: Some suggestions for style changes here and there

marcospri:
- Moved the helper `parser` function into the service
- Tracked down the `form: _schema` error to missing LTI1.1 AI. The error is the same in other LMS. Created https://github.com/hypothesis/lms/pull/4672 to address the bad error message.

PR #4673, opened 2022-11-21
Add a passing test to the SQL runner functests

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/84

This expands the number of failures the tests can catch (like missing environment variables).

This also fixes a bug where the report schema can prevent the functests from running. I think this can only happen in dev where you have successfully run the script against `h`. In CI it should never run and create the schema.

### DIFF
--- a/tests/functional/bin/run_sql_task_test.py
+++ b/tests/functional/bin/run_sql_task_test.py
@@ -10,33 +10,39 @@ from tests.functional.conftest import TEST_ENVIRONMENT
 
 
 class TestRunSQLTask:
+    def test_sanity(self, environ):
+        self.run_task(environ, "hello_world")
+
     # We use "db_engine" here to ensure the schema is created
     @pytest.mark.usefixtures("db_engine")
     # If you want to run these tests, you first need to start services in H
     # and run the `report/create_from_scratch` task there.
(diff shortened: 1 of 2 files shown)


tests/functional/bin/run_sql_task_test.py
```diff
@@ -10,33 +10,39 @@
...
+        self.run_task(environ, "hello_world")
+
     # We use "db_engine" here to ensure the schema is created
     @pytest.mark.usefixtures("db_engine")
     # If you want to run these tests, you first need to start services in H
     # and run the `report/create_from_scratch` task there.
     @pytest.mark.xfail(reason="Requires reporting tasks to be run in H")
     def test_reporting_tasks(self, environ):
-        for report_path in (
+        for task_name in (
```
marcospri, line 22: I vaguely remember writing these as a `pytest.mark.parametrize` once but here you'll have to mark all of them as xfail except "hello_world" which might be a bit less clear.

marcospri (approved):

tests/functional/bin/run_sql_task_test.py
jon-betts: This is a bit different as we don't want the DB cleared between runs. I'm not sure it actually is, but we explicitly want to run them back to back.

PR #4677, opened 2022-11-23
Create grade items on assignment configuration in D2L

marcospri:
For 

In summary:

- In other LMS the moment you create a gradable assignment in the LMS the container to hold the grades it's also created by the LMS itself.
- We get a reference (a URL) to that container on launch and we decide if the assignment is gradable based on its presence.
- This is also true in D2L for LTI 1.1
- In D2L 1.3 we found that the behavior cahnged and the container is not created. 
- As a work around the container (lineitem in 1.3 speak, grade item in D2Ls UI) can be created directly in the UI. This add an extra step to use hypothesis.
- A new version of D2L doesn't allow  creating this in the UI.

The proposed solution is to mimic  the LTI 1.1 behavior creating the lineitem on assignment configuration using LTIA.

We'll use a AI setting as a feature flag to make it easier to roll-out (and reverting) and just in case some schools prefer using the UI.

### DIFF
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -47,8 +47,12 @@ class Settings:
     groups_enabled: bool = False
     """Is the course groups feature enabled"""
 
+    auto_create_lineitem: bool = False
+    """Create lineitem on assignment configuration"""
+
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
+        self.groups_enabled = product_settings.get("auto_create_lineitem", False)
 
 
 @dataclass
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -73,6 +73,10 @@ class Product:
     def configure_assigment(self):
         pass
 
+    def ltia_aud_claim(self, lti_registration):
+        """Return the value of the `aud` claim used in LTI advantage requests."""
+        return lti_registration.client_id
+
     @classmethod
     def from_request(cls, request, ai_settings: Dict):
         """Create a populated product object from the provided request."""
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -66,6 +66,10 @@ class Product:
     # Accessor for external consumption
     Family = Family
 
+    def is_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        return bool(lti_params.get("lis_outcome_service_url"))
+
     @classmethod
     def from_request(cls, request, ai_settings: Dict):
         """Create a populated product object from the provided request."""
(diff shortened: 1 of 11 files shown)


lms/product/product.py
```diff
@@ -66,6 +70,17 @@ class Product:
     # Accessor for external consumption
     Family = Family
 
+    def is_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        return bool(lti_params.get("lis_outcome_service_url"))
+
+    def configure_assigment(self):
```
marcospri, line 77: TODO typo here.

lms/product/product.py
```diff
@@ -66,6 +70,17 @@ class Product:
     # Accessor for external consumption
     Family = Family
 
+    def is_gradable(self, lti_params):
```
marcospri, line 73: Adding three new methods here.

I'm adding them at the top level class as there's not much else to generalize a new smaller one. It should be easy to clean if the interface grows to much.


Adding:

- `is_gradable`

Used in: 
https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/views/lti/basic_launch.py#L156

  Special D2L case in: 
  https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/product/d2l/product.py#L26


I reckon this ones is fine. Maybe in other LMS we'll need to take request instead or the params? We'll see when we get there.

- `ltia_aud_claim`


D2L case

https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/product/d2l/product.py#L35

Used in https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/services/ltia_http.py#L72


We might find out that we need more changes for other LMS. Again we can change it later and I like how you can go to the LTIA service see the common parameters there except for the special one, which is clearly marked.


- `configure_assignment`

Used in https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/views/lti/basic_launch.py#L132


D2L change https://github.com/hypothesis/lms/blob/fe9bac5786a15f8eb937bba296343526c7ba1117/lms/product/d2l/product.py#L41 

(which the main focus of this PR)

This one might be a bit to broad. It's a big hook to do "whatever" while configuring the assignment.

What we are doing here is at the moment something that only makes sense in D2L (ie, not something we are deciding not to to elsewhere, but something that doesn't make sense to do). 

We could, in the the main `configure_assignment` in the view use a setting to conditionally do it there, based on the setting, not the product.

I reckon in that scenario we'd need a comment explaining the situation in D2L so that suggests it belongs to the product.

marcospri:
Superseded by https://github.com/hypothesis/lms/pull/4726

PR #4679, opened 2022-11-23
Include events which don't have user records

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/92

We didn't need this before (as we were only interested in organizations), but without rows which have no user, we miss a massive number of launches from the before times.

### DIFF
--- a/lms/sql_tasks/tasks/report/create_from_scratch/02_entities/05_events/01_create_view.sql
+++ b/lms/sql_tasks/tasks/report/create_from_scratch/02_entities/05_events/01_create_view.sql
@@ -1,7 +1,7 @@
 DROP TYPE IF EXISTS report.event_type CASCADE;
 
 CREATE TYPE report.event_type AS ENUM (
-    'configured_launch', 'deep_linking'
+    'configured_launch', 'deep_linking', 'audit'
 );
 
 DROP MATERIALIZED VIEW IF EXISTS report.events CASCADE;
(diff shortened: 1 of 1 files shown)


lms/sql_tasks/tasks/report/create_from_scratch/02_entities/05_events/01_create_view.sql
```diff
@@ -10,32 +10,32 @@ DROP MATERIALIZED VIEW IF EXISTS report.events CASCADE;
...
                 event.type_id = event_type.id
-            JOIN report.user_map ON
-                event_user.user_id = user_map.lms_user_id
-            GROUP BY
-                timestamp, event_id, event_type.type, user_map.user_id, user_map.organization_id
+            -- Lots of events don't have users
+            LEFT OUTER JOIN unique_event_users ON
+                unique_event_users.event_id = event.id
+            LEFT OUTER JOIN report.user_map ON
+                unique_event_users.user_id = user_map.lms_user_id
```
jon-betts, line 31: This is the meat of the change here, making these left outers, so that rows without users still show up

lms/sql_tasks/tasks/report/create_from_scratch/02_entities/05_events/01_create_view.sql
```diff
@@ -1,7 +1,7 @@
 DROP TYPE IF EXISTS report.event_type CASCADE;
 
 CREATE TYPE report.event_type AS ENUM (
-    'configured_launch', 'deep_linking'
+    'configured_launch', 'deep_linking', 'audit'
```
jon-betts, line 4: Might as well fix this while we are here

lms/sql_tasks/tasks/report/create_from_scratch/02_entities/05_events/01_create_view.sql
marcospri: Yeah, missed this too, for launches missing users it's about historical data being incomplete but for other types of events, it might not make sense to link them to a particular user.

marcospri (approved):

PR #4681, opened 2022-11-24
Use data-task library

marcospri:


jon-betts (approved):
Looks good.

This does make me wonder if there's a way to have the script as part of the library. It would be a bit odd because we'd need that script to load something from us to work? Maybe there's a way around it?

PR #4683, opened 2022-11-24
Admin pages tweaks

marcospri:
For both registrations and application instances:

- Search by ID.
- Searching with no criteria returns some results.
- Don't redirect to the result when there only one. It's a bit confusing, you just might want to copy something from the results table for example.
- Correctly 404 when registration doesn't exist.

## Testing

- http://localhost:8001/admin/registrations/

  - Search directly there, you'll get results
  - Search using ID `1`
  - http://localhost:8001/admin/registration/id/78789789/ is a 404 not an empty page

- http://localhost:8001/admin/instances/

  - Search directly there, you'll get results
  - Search using ID `100`

### DIFF
--- a/lms/views/admin/__init__.py
+++ b/lms/views/admin/__init__.py
@@ -1,8 +1,32 @@
+from marshmallow import fields, missing
 from pyramid.httpexceptions import HTTPFound, HTTPNotFound
 from pyramid.renderers import render_to_response
 from pyramid.view import forbidden_view_config, notfound_view_config, view_config
 
 from lms.validation._exceptions import ValidationError
+
+
+class EmptyStringNoneMixin:
+    """
+    Allows empty string as "missing value".
+
+    Marshmallow doesn't have a clean solution yet to POSTed values
+    (that are always present in the request as empty strings)
+
+    Here we convert them explicitly to None
+
+    https://github.com/marshmallow-code/marshmallow/issues/713
+    """
+
+    def deserialize(self, value, attr, data, **kwargs):
+        # pylint:disable=compare-to-empty-string
+        if value == missing or value.strip() == "":
+            return None
+        return super().deserialize(value, attr, data, **kwargs)
+
+
+class EmptyStringInt(EmptyStringNoneMixin, fields.Int):
+    """Allow empty string as "missing value" instead of failing integer validation."""
 
 
 @forbidden_view_config(path_info="/admin/*")
--- a/lms/views/admin/lti_registration.py
+++ b/lms/views/admin/lti_registration.py
@@ -168,8 +168,8 @@ class AdminLTIRegistrationViews:
 
         registrations = self.lti_registration_service.search_registrations(
             id_=self.request.params.get("id"),
-            issuer=self.request.params.get("issuer"),
-            client_id=self.request.params.get("client_id"),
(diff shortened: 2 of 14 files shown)


tests/unit/lms/services/application_instance_test.py
```diff
@@ -63,7 +63,7 @@ def test_get_by_id(self, service, application_instance):
 
     def test_get_by_not_found(self, service):
         with pytest.raises(ApplicationInstanceNotFound):
-            service.get_by_id(0)
+            service.get_by_id(100_000_000)
```
marcospri, line 66: `0` not a valid ID, use a real one.

tests/unit/lms/services/lti_registration_test.py
```diff
@@ -57,6 +61,7 @@ def test_search_registrations(self, svc):
 
     @pytest.fixture
     def registration(self):
+        factories.LTIRegistration.create_batch(4)
```
marcospri, line 64: Add some more regs here.

jon-betts:
Testing results:

 * Visit: http://localhost:8001/admin/registrations
 * Enter 1
 * Search works and shows a value at the bottom
 * Enter 87465873465
 * :lady_beetle:  **Search crashes with "Sorry, but something went wrong"** !
 * Visit: http://localhost:8001/admin/registration/id/78789789/
 * Shows a plain 404
 
 * Visit: http://localhost:8001/admin/instances
 * Enter 100
 * Search works and shows a value at the bottom
 * Enter 87465873465
 * :lady_beetle:  **Search crashes with "Sorry, but something went wrong"** !
 * Visit: http://localhost:8001/admin/instance/id/23453454/
 * Shows a plain 404

marcospri:
> ## Testing results
> Registrations
> 
> * Visit: http://localhost:8001/admin/registrations
> * Enter 1
> * Search works and shows a value at the bottom
> * Enter 87465873465
> * lady_beetle  **Search crashes with "Sorry, but something went wrong"** !


Fixed this in a new commit for registrations, application instances and organizations (which already had the search by ID).

lms/views/admin/__init__.py
```diff
@@ -1,10 +1,34 @@
+from marshmallow import fields, missing
 from pyramid.httpexceptions import HTTPFound, HTTPNotFound
 from pyramid.renderers import render_to_response
 from pyramid.view import forbidden_view_config, notfound_view_config, view_config
 
 from lms.validation._exceptions import ValidationError
 
 
+class EmptyStringNoneMixin:
```
marcospri, line 9: This is very verbose but I reckon we should handle any of this validations with Marshmallow.

lms/views/admin/lti_registration.py
```diff
@@ -240,3 +235,9 @@ def _get_url_suggestions(issuer, client_id):
                 }
 
         return {"auth_login_url": None, "key_set_url": None, "token_url": None}
+
+    def _get_registration_or_404(self, id_) -> LTIRegistration:
+        if org := self.lti_registration_service.get_by_id(id_):
+            return org
+
+        raise HTTPNotFound()
```
jon-betts, line 243: This isn't very fancy, but users shouldn't be typing in URLs by hand, so this seems fine.

Issue #4678, opened 2022-11-23
Use D2L nomenclature for group sets


marcospri:
We use the name "group sets" in various parts of the UI but we should match the nomenclature 
of the LMS we are running in.

For D2L that's "group categories".

For instances of this in error modals that are based on a error code we can create a version of dialog with the right name.

When this is part of the general UI for example in:

https://github.com/hypothesis/lms/blob/lms/static/scripts/frontend_apps/components/GroupConfigSelector.js#L44

we could rely on the new product config option:

https://github.com/hypothesis/lms/blob/main/lms/static/scripts/frontend_apps/config.js#L181

and either:

- Use the family property and a few conditionals in the FE to pick the right value
- Send a new "group_set_naming" or similar from the BE and just use that
- Trying to use a more generic name in all LMS
  Not a fan of this one.
- ???

I reckon I'd prefer the first option.

robertknight:
> I reckon I'd prefer the first option.

I agree. The risk with trying to send a value from the backend is that you run into a wall as soon as you need just a bit more flexibility. Also when we come to translate to other languages, there may be issues with other parts of a sentence around the changed terminology needing to change.

> Trying to use a more generic name in all LMS. Not a fan of this one.

I agree here too. It seems like it will be important to use the same terminology as the LMS the user is in.

PR #4684, opened 2022-11-24
WIP use the right D2L terminology

marcospri:
For:

- https://github.com/hypothesis/lms/issues/4678

This is a half done PR, the aim is to use the term "group category/ies" throughout the application when using D2L instead of "group set/s" nomenclature used in other LMS.

This is a very small thing to do but there's no existing examples of it on the code base so I'm struggling to find the right "design" so I'm happy to:

- Take some notes on the PR and continue to write a merge-able one
- Handing it over to you to finish it

Note that there's two commits here, the first one around error dialog, seems more obvious to me, we already have a mechanism there to display different message based on the error code. 

Is in the second one where strings appear in different ways in the  code and I'm not sure what the right approach is.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
+++ b/lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
@@ -33,28 +33,35 @@ import ErrorModal from './ErrorModal';
  * @prop {GroupSet[]|null} groupSets
  * @prop {(id: string|null) => void} onInput
  * @prop {string|null} selectedGroupSetId
+ * @prop {string} productFamily - Which product (canvas, blackboard...) we are running in.
  *
  * @param {GroupSelectProps} props
  */
-function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
+function GroupSelect({
+  busy,
+  groupSets,
+  onInput,
+  selectedGroupSetId,
+  productFamily,
+}) {
   const selectId = useUniqueId('GroupSetSelector__select');
 
   return (
     <>
-      <label htmlFor={selectId}>Group set: </label>
-      <select
+      <GroupSelectHeader htmlFor={selectId} productFamily={productFamily} /> <select
         disabled={busy}
         id={selectId}
         onInput={e =>
           onInput(/** @type {HTMLInputElement} */ (e.target).value || null)
         }
       >
-        {busy && <option>Fetching group sets…</option>}
+        {busy && <GroupSelectLoadingOption productFamily={productFamily} />}
         {groupSets && (
           <>
-            <option disabled selected={selectedGroupSetId === null}>
-              Select group set
-            </option>
+            <GroupSelectPromptOption
+              productFamily={productFamily}
+              groupSelected={selectedGroupSetId === null}
+            />
             <hr />
             {groupSets.map(gs => (
               <option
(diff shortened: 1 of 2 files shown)


lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -33,28 +33,35 @@ import ErrorModal from './ErrorModal';
  * @prop {GroupSet[]|null} groupSets
  * @prop {(id: string|null) => void} onInput
  * @prop {string|null} selectedGroupSetId
+ * @prop {string} productFamily - Which product (canvas, blackboard...) we are running in.
```
marcospri, line 36: This is a string but it could be declared a type/enum in config.js.

The values there are currently all over the place:

https://github.com/hypothesis/lms/blob/main/lms/product/product.py#L10

we could send over the names of that enum instead of the values to be more consistent between them.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+    </option>
+  );
+}
+
 /**
  * ErrorModal shown when the fetched list of course group sets is empty
  *
  * @param {object} props
  *   @param {() => void} props.onCancel
+ *   @param {string} props.productFamily
```
marcospri, line 140: Oh so I can declare the props type without giving the type a name, learner that too late in the file.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+ *
+ *  @typedef GroupSelectPromptOptionProps
+ * @prop {string} productFamily
+ * @prop {boolean} groupSelected
+ *
+ * @param {GroupSelectPromptOptionProps} props
+ */
+
+function GroupSelectPromptOption({ productFamily, groupSelected }) {
+  const groupSetNaming =
```
marcospri, line 125: Using a string value here

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+ *
+ *  @typedef GroupSelectLoadingOptionProps
+ * @prop {string} productFamily
+ *
+ * @param {GroupSelectLoadingOptionProps} props
+ */
+
+function GroupSelectLoadingOption({ productFamily }) {
+  if (productFamily === 'desire2learn') {
+    return <option>Fetching group categories…</option>;
```
marcospri, line 109: Just one return per family over here

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+          <p>
+            Hypothesis relies on canvas group sets to place students into
+            groups, but we could not find any available group sets in this
+            course.
+          </p>
+          <p>
+            Please add one or more group sets to your course and try again. We
+            also have a{' '}
+            <Link
+              href="https://web.hypothes.is/help/using-canvas-groups-to-create-hypothesis-reading-groups/"
```
marcospri, line 179: We could take the opportunity to link to the right article directly

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+          </p>
+        </>
+      </ErrorModal>
+    );
+  } else if (productFamily === 'canvas') {
+    return (
+      <ErrorModal onCancel={onCancel} title="No group sets found">
+        <>
+          <p>
+            Hypothesis relies on canvas group sets to place students into
```
marcospri, line 171: added `canvas` here too

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -211,8 +321,11 @@ export default function GroupConfigSelector({
 
   if (fetchError) {
     if (fetchError instanceof GroupListEmptyError) {
-      return <NoGroupsError onCancel={onErrorCancel} />;
+      return (
+        <NoGroupsError onCancel={onErrorCancel} productFamily={productFamily} />
+      );
     } else if (isAuthorizationError(fetchError)) {
+      // TODO
```
marcospri, line 328: I few places with "group set" I haven't tackled

lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -161,21 +160,52 @@ export default function LaunchErrorDialog({
           </p>
         </ErrorModal>
       );
+    case 'd2l_group_set_not_found':
```
marcospri, line 163: For these everything is in place to just create a new branch and do the full dialog.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
robertknight: > This is a string but it could be declared a type/enum in config.js.

I think it would be wise to indicate at least where the source of truth is for possible values.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -33,28 +33,35 @@ import ErrorModal from './ErrorModal';
...
+  selectedGroupSetId,
+  productFamily,
+}) {
   const selectId = useUniqueId('GroupSetSelector__select');
 
   return (
     <>
-      <label htmlFor={selectId}>Group set: </label>
-      <select
+      <GroupSelectHeader htmlFor={selectId} productFamily={productFamily} /> <select
```
robertknight, line 51: Do you foresee anything other than the text changing for the components other than the error dialog. If not, perhaps it would make sense to build a simple message ID / message text lookup, with a mechanism to support adjusting the language depending on the product.

```js
const messages = {
  selectLabel: 'Group set: ', // Default
  selectLabel_desire2learn: 'Group categories: ' // Customized for d2l
};
const message = id => messages[id + "_" + productFamily] ?? messages[id];
...
<label htmlFor={selectId}>{message("selectLabel")}</label>
```

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
   );
 }
 
+/**
+ *
+ *  @typedef GroupSelectHeaderProps
```
robertknight, line 85:
```suggestion
/**
 * @typedef GroupSelectHeaderProps
```

Small formatting correction here and for several other cases. Our formatter unfortunately doesn't touch JSDoc. When we switch this over to TypeScript in future, that problem will go away.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+  if (productFamily === 'desire2learn') {
+    return <label htmlFor={htmlFor}>Group category: </label>;
+  } else {
+    return <label htmlFor={htmlFor}>Group set: </label>;
+  }
+}
+
+/**
+ *
+ *  @typedef GroupSelectLoadingOptionProps
```
robertknight, line 101:
```suggestion
/**
 * @typedef GroupSelectLoadingOptionProps
```

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+  if (productFamily === 'desire2learn') {
+    return <option>Fetching group categories…</option>;
+  } else {
+    return <option>Fetching group sets…</option>;
+  }
+}
+
+/**
+ *
+ *  @typedef GroupSelectPromptOptionProps
```
robertknight, line 117:
```suggestion
/**
 * @typedef GroupSelectPromptOptionProps
```

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -73,13 +80,115 @@ function GroupSelect({ busy, groupSets, onInput, selectedGroupSetId }) {
...
+            >
+              help article about using groups in Canvas
+            </Link>
+            .
+          </p>
+        </>
+      </ErrorModal>
+    );
+  }
+  // TODO BLACKBOARD
```
robertknight, line 190: Are you planning to address this as part of this PR?

robertknight (commented):
Hi @marcospri - Sorry to keep you waiting on this. Please do ping me more aggressively about PRs that need my input. I suggest we get this finished up using the current pattern. In cases where only the text needs to change, the problem of tweaking the text seems to me somewhat like translating strings, so we could probably create a generic message ID => message text lookup, where the lookup result can be adjusted depending on the current product family.

marcospri:
Continues over https://github.com/hypothesis/lms/pull/4803

PR #4686, opened 2022-11-25
Move tasks to data_tasks directory

marcospri:
Adapt executable and github actions naming as well

## Testing 

- `tox -e dev --run-command "python bin/run_data_task.py --task report/create_from_scratch --config conf/development.ini"`


jon-betts (approved):
This looks good an works, but I think we've left behind an `sql_tasks/__init__.py`?

PR #4708, opened 2022-11-30
Create new LTILaunchService and launch plug

marcospri:
Moves bits of the views into a new service. 

With the extra boilerplate of the new service and plugin this looks bigger than the original for not much benefit.

This PR doesn't:

- Make any changes around JSConfig (which could belong to the new service). That would be its own separate refactor.
- Change the behavior of the moved code, except for the bits moved the plugin. Again that would make the size of this PR much bigger.

That being said this PR allows to move some product specific behavior into plugins, allows re-use of some of the launch logic and moves us close to remove the last `is_canvas` in the codebase.

## Testing

Sanity check with some launches:

- https://hypothesis.instructure.com/courses/125/assignments/873
- https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_398_1&course_id=_19_1
- And re-configure:  https://hypothesis.instructure.com/courses/125/assignments/3118/edit?return_to=https%3A%2F%2Fhypothesis.instructure.com%2Fcourses%2F125%2Fassignments%2F3118

### DIFF
--- a/lms/product/canvas/_plugin/launch.py
+++ b/lms/product/canvas/_plugin/launch.py
@@ -1,0 +1,10 @@
+from lms.product.plugin.launch import LaunchPlugin
+
+
+class CanvasLaunchPlugin(LaunchPlugin):
+    def course_extra(self, lti_params):
+        return {
+            "canvas": {
+                "custom_canvas_course_id": lti_params.get("custom_canvas_course_id")
+            }
+        }
--- a/lms/product/canvas/product.py
+++ b/lms/product/canvas/product.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 
 from lms.product.canvas._plugin.grouping import CanvasGroupingPlugin
+from lms.product.canvas._plugin.launch import CanvasLaunchPlugin
 from lms.product.product import PluginConfig, Product, Routes
 
 
@@ -16,7 +17,9 @@ class Canvas(Product):
         list_group_sets="canvas_api.courses.group_sets.list",
     )
 
-    plugin_config: PluginConfig = PluginConfig(grouping=CanvasGroupingPlugin)
+    plugin_config: PluginConfig = PluginConfig(
+        grouping=CanvasGroupingPlugin, launch=CanvasLaunchPlugin
+    )
 
     settings_key = "canvas"
 
--- a/lms/product/canvas/product.py
+++ b/lms/product/canvas/product.py
@@ -19,3 +19,6 @@ class Canvas(Product):
     plugin_config: PluginConfig = PluginConfig(grouping=CanvasGroupingPlugin)
 
     settings_key = "canvas"
+
+    use_grading_bar = False
+    """We integrate with SpeedGrader and don't need to use the grading bar"""
--- a/lms/product/plugin/launch.py
+++ b/lms/product/plugin/launch.py
@@ -1,0 +1,9 @@
+# pylint: disable=unused-argument
+class LaunchPlugin:
+    def is_assignment_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        return bool(lti_params.get("lis_outcome_service_url"))
+
+    def course_extra(self, lti_params):
+        """Extra information to store for courses."""
+        return {}
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -56,8 +56,7 @@ class LTILaunchResource:
             self._request.lti_params["tool_consumer_instance_guid"],
             self._request.lti_params.get("resource_link_id"),
         )
-
-        return self._request.product.plugin.grouping_service.launch_grouping_type(
+        return self._request.find_service(name="grouping").get_launch_grouping_type(
             self._request, self.application_instance, self.course, assignment
         )
 
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -51,72 +51,15 @@ class LTILaunchResource:
         return JSConfig(self, self._request)
 
     @property
-    def sections_enabled(self):
-        """Return if sections are enabled for this request."""
-
-        if not self.is_canvas:
-            # Sections are only implemented in Canvas
-            return False
-
-        params = self._request.params
-        if "focused_user" in params and "learner_canvas_user_id" not in params:
-            # This is a legacy SpeedGrader URL, submitted to Canvas before our
-            # Canvas course sections feature was released.
-            return False
-
-        if not bool(self.application_instance.developer_key):
-            # We need a developer key to talk to the API
-            return False
-
-        return self.course.settings.get("canvas", "sections_enabled")
-
-    @property
-    def group_set_id(self):
-        """
-        Get the group set ID for group launches.
-
-        A course can be divided in multiple "small groups" but it's possible to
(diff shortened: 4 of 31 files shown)


lms/product/canvas/_plugin/launch.py
```diff
@@ -0,0 +1,10 @@
+from lms.product.plugin.launch import LaunchPlugin
+
+
+class CanvasLaunchPlugin(LaunchPlugin):
+    def course_extra(self, lti_params):
```
marcospri, line 5: We could first super() and append this to the result but currently this is a canvas only thing, not sure how (if) the next one will look.

lms/product/canvas/product.py
```diff
@@ -19,3 +19,6 @@ class Canvas(Product):
     plugin_config: PluginConfig = PluginConfig(grouping=CanvasGroupingPlugin)
 
     settings_key = "canvas"
+
+    use_grading_bar = False
```
marcospri, line 23: This is a product wide decision for now. I reckon it belongs here at the top level.

lms/product/plugin/launch.py
```diff
@@ -0,0 +1,9 @@
+# pylint: disable=unused-argument
+class LaunchPlugin:
+    def is_assignment_gradable(self, lti_params):
```
marcospri, line 3: Only implementation for now, it will be different for D2L.

lms/resources/lti_launch.py
```diff
@@ -29,12 +30,7 @@ def __init__(self, request):
     @cached_property
     def course(self):
         """Get the course this LTI launch based on the request's params."""
-
-        return self._request.find_service(name="course").upsert_course(
-            context_id=self._request.parsed_params["context_id"],
-            name=self._request.parsed_params["context_title"],
-            extra=self._course_extra(),
-        )
+        return self._request.find_service(LTILaunchService).record_course()
```
marcospri, line 33: There's still a couple of places where this is used in JSConfig. I'd aim to remove it but I want to do that in a different refactor otherwise this one will get huge.

lms/resources/lti_launch.py
```diff
@@ -59,18 +55,3 @@ def grouping_type(self) -> Grouping.Type:
         return self._request.find_service(name="grouping").get_launch_grouping_type(
             self._request, self.course, assignment
         )
-
-    def _course_extra(self):
```
marcospri, line 63: Happy to remove this wart from this file.

lms/services/lti_launch.py
```diff
@@ -0,0 +1,102 @@
...
+from lms.services.application_instance import ApplicationInstanceService
+from lms.services.assignment import AssignmentService
+from lms.services.grading_info import GradingInfoService
+from lms.services.grouping import GroupingService
+from lms.services.lti_role_service import LTIRoleService
+
+
+class LTILaunchService:
+    # pylint:disable=too-many-instance-attributes,too-many-arguments
+    def __init__(
```
marcospri, line 11: This list is enormous. 

The services/dependencies are the same as in the view but here they are listed explicitly.

We could try to get "parent"  things from the "children"· 

Assignment->Course->Application Instance.


I reckon that will be easier to refactor with this service in place.

lms/views/api/gateway.py
```diff
@@ -33,34 +34,34 @@ class GatewayLTISchema(LTIV11CoreSchema):
...
-    # Ensure no funny business is going on trying to access content out of the
-    # current application instance scope.
+    lti_launch_service = request.find_service(LTILaunchService)
     try:
-        context.application_instance.check_guid_aligns(
-            request.lti_params["tool_consumer_instance_guid"]
-        )
+        # Ensure no funny business is going on trying to access content out of the
+        # current application instance scope
+        lti_launch_service.validate_launch()
```
marcospri, line 48: We handle the exception differently here.

lms/views/lti/deep_linking.py
```diff
@@ -62,11 +62,13 @@
 )
 def deep_linking_launch(context, request):
     """Handle deep linking launches."""
+    lti_launch_service: LTILaunchService = request.find_service(LTILaunchService)
 
-    request.find_service(name="application_instance").update_from_lti_params(
-        context.application_instance, request.lti_params
-    )
-    request.find_service(name="lti_h").sync([context.course], request.params)
+    lti_launch_service.validate_launch()
```
marcospri, line 67: This was a bug, we were not checking this for this type of launch.

tests/unit/conftest.py
```diff
@@ -124,7 +124,7 @@ def pyramid_request(db_session, application_instance, lti_v11_params):
     )
 
     pyramid_request.lti_jwt = {}
-    pyramid_request.lti_params = {}
+    pyramid_request.lti_params = LTIParams(lti_v11_params)
```
marcospri, line 127: params/lti_params were out of sync if the fixture.

jon-betts (commented):
This is almost certainly worth talking about, but I think anything that cuts the launch view down more so we can think about it is a good thing, but I'm a little fuzzy on the unique responsibilities of the new objects.

I think a lot of this would be helped with docstrings which explain what the objects are intended to be, rather than what they are at the moment. So the object might have X methods, but when it comes to adding X+1 what should I go by to decide if the object is the right place?

### About the service

I think the motivators for this are:

 * We want plugin points, and this matches how we approached the group service
 * We want the same functionality in basic and deep linked launches
 * The launch view is very long and busy still

For me, the service doesn't seem to have a strong identity at the moment. It's about launch stuff, but there are plenty of things left in the view that could have come here. Why aren't those things here as well? There's not much to go on by the name and docs to know.

Looking at this I can't shake the feeling that this is a "LTIPersistenceService" which saves various LTI related objects. If we removed the `validate_launch` then every method in the service is `record_<object>`. That sounds like a definable role and identity to me.

### About `views/deep_linking.py`

One of the interesting effect of all our refactors here is they've often revealed what felt like the truth we couldn't see before, and I wonder if that's happened here.

It really looks to me like the `deep_linking_launch()` part of that file should be in with the rest of the launches. The start of the method is now identical to the init and the rest looks a lot like an `unconfigured_launch` which it is.

I can't remember if there is a more prosaic Pyramid routing reason these are in different places, but I wonder if it's just as simple as deciding the "deep-linkiness" of it was the organizing principal, rather than the "launchiness" of it?

If we had done this before the bug spotted in this ticket would not have been possible. This would also remove one of the motivations for the service mentioned above.

### About the plugin

I think the plugin is sort of a slave to random requirements of the thing it's plugged into, so the name, interface etc are all dictated by that thing. This one looks really weird, but I think that's fine.

Again, I think we really want to spell out what the plugin points are intended to do, and be used for, and what the plugin as a whole is for in docstrings.

## Ok... so what does this all mean?

* We should state our intention of what objects should continue to be as well as what they are now in docstrings so our design isn't forgotten and corrupted
* We should think about putting the deep linked launch next to the unconfigured launch
* **_If we do that, we can probably introduce the plugin without the service at first_**
   * This is because some of the reasons to do the service are removed
   * This might mean a much shorter PR for that part?
* We can then also probably take a breather to think about the services as tech-debt, rather than a feature blocker
* We should think about whether this is a `LTIPersistenceService` or something similar

PR #4710, opened 2022-12-01
Include jstor in Speedgrader URL bug workaround

marcospri:
For 

- https://github.com/hypothesis/product-backlog/issues/1408

Speedgrader double quotes chars in the URL. We need to unquote those to correctly identify the schema.

Add `jstor://` to the whitelisted schemas.

## Ideas for a more general fix

This PR addresses the bug without making any bigger changes, we could potentially:

- Make the regexp more general, if it starts with some chars followed by `%3a` just unquote it. (See https://github.com/hypothesis/lms/pull/4711)

- Report unparseable URLs in (lms) via to sentry. We might get a few false positives there (typos) but it might worth a try. We could start with a regular (papertrail) log to see how much volume it generates.

## Testing

- Launch https://hypothesis.instructure.com/courses/125/assignments/3849 in Speedgrader (you might need to first make an annotation to get the document to show)

- In main via will error. It should work in this branch

### DIFF
--- a/lms/services/document_url.py
+++ b/lms/services/document_url.py
@@ -32,7 +32,7 @@ class DocumentURLService:
 
         return None
 
-    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource)%3a", re.IGNORECASE)
+    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource|jstor)%3a", re.IGNORECASE)
 
     @classmethod
     def _from_deep_linking_provided_url(cls, request):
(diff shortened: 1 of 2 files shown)


robertknight:
> Make the regexp more general, if it starts with some chars followed by %3a just unquote it. (See https://github.com/hypothesis/lms/pull/4711)

The use of the regex is to detect whether a URL has been double percent-encoded by Canvas, so for this it is sufficient to check for `[scheme_chars]%3a` at the start of the URL where `scheme_chars` are characters allowed in URL schemes. The reasons to match specific schemes would be if there was a security risk with unknown schemes (doesn't apply here) or if the Canvas bug that causes double encoding (https://github.com/instructure/canvas-lms/issues/1486) somehow necessitated handling certain schemes differently.

marcospri:
Superseded by https://github.com/hypothesis/lms/pull/4711

PR #4711, opened 2022-12-01
Use a more general regexp to handle double quoting

marcospri:
For 

- https://github.com/hypothesis/product-backlog/issues/1408

Speedgrader double-quotes URLs so we need to fix our `document_url` url param before anything else.

Change the regexp from a whitelisted set of schemas to a more general one.

See https://github.com/hypothesis/lms/pull/4710

### DIFF
--- a/lms/services/document_url.py
+++ b/lms/services/document_url.py
@@ -32,7 +32,9 @@ class DocumentURLService:
 
         return None
 
-    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource|jstor)%3a", re.IGNORECASE)
+    # If the URL looks like "schema%3a" (schema:) it's double-encoded
+    # https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+    _ENCODED_URL = re.compile("^[a-z][a-z0-9+.-]*%3a", re.IGNORECASE)
 
     @classmethod
     def _from_deep_linking_provided_url(cls, request):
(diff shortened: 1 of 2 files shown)


lms/services/document_url.py
```diff
@@ -32,7 +32,8 @@ def get_document_url(self, request) -> Optional[str]:
 
         return None
 
-    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource|jstor)%3a", re.IGNORECASE)
+    # If the URL looks like "schema%3a" (schema:) it's double-encoded
+    _ENCODED_URL = re.compile("^(?:[a-z]+)%3a", re.IGNORECASE)
```
robertknight, line 36: We currently only use letters in schemas, but it is conceivable that someone might try to use other schema-allowed characters in future. I'm mainly thinking of digits and "-", but "." and "+" are also technically allowed.

lms/services/document_url.py
```diff
@@ -32,7 +32,9 @@ def get_document_url(self, request) -> Optional[str]:
 
         return None
 
-    _ENCODED_URL = re.compile("^(?:https?|canvas|vitalsource|jstor)%3a", re.IGNORECASE)
+    # If the URL looks like "schema%3a" (schema:) it's double-encoded
+    # https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+    _ENCODED_URL = re.compile("^[a-z][a-z0-9+0.-]+%3a", re.IGNORECASE)
```
robertknight, line 37:
```suggestion
    _ENCODED_URL = re.compile("^[a-z][a-z0-9+.-]*%3a", re.IGNORECASE)
```

Removed a repeated "0". The change from "+" to "*" at the end is not practically important, but just for spec consistency.

robertknight (approved):
LGTM with one minor correction.

Issue #4700, opened 2022-11-03
Add note field to LMS application instances


mattdricker:

chrisdaaz:
This is now available in LMS admin.

PR #4712, opened 2022-12-01
Notes fields on application instances admin pages

marcospri:
For:

- https://github.com/hypothesis/lms/issues/4700

Add a note field in the admin pages for Application instances.

Also took the opportunity to move the admin schema to their own file in a separate commit

## Testing

- Go to http://localhost:8001/admin/instance/id/2/
- Write some (funny) note
- Save
- Refresh the page, it should appear again.

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -32,7 +32,8 @@
 
 {% macro settings_textarea(label, setting, sub_setting, field_name, default='') %}
     {% call macros.field_body(label) %}
-        <textarea class="textarea" name="{{ setting }}.{{ sub_setting }}">{{ instance.settings.get(setting, sub_setting, default) or '' }}</textarea>
+        {% set text = instance.settings.get(setting, sub_setting, default) or ''  %}
+        <textarea class="textarea" name="{{ setting }}.{{ sub_setting }}">{{ text }}</textarea>
     {% endcall %}
 {% endmacro %}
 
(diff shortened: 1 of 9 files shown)


lms/views/admin/application_instance.py
```diff
@@ -309,6 +310,7 @@ def update_instance(self):
             ("vitalsource", "disable_licence_check", bool),
             ("jstor", "enabled", bool),
             ("jstor", "site_code", str),
+            ("hypothesis", "notes", str),
```
marcospri, line 313: Using a nested value inside `ai.settings`, this is both:

- The path of least resistance. Most of our settings setup expect this nested structure.
- Tidier to keep our own (as opposed to canvas/vitalsoruce/...) values in one place.

lms/templates/admin/instance.html.jinja2
```diff
@@ -136,6 +141,7 @@
         {{ settings_text_field("JSTOR site code", "jstor", "site_code") }}
     </fieldset>
 
+    {{ settings_textarea("Notes", "hypothesis", "notes") }}
```
marcospri, line 144: I reckon the notes should go at the very top or bottom (as they are easier to spot/go to).

Here at the bottom I reckon they fit better with some other "meta" fields like the dates.

tests/unit/lms/views/admin/__init___test.py
```diff
@@ -29,27 +26,3 @@ def test_index(pyramid_request):
     assert response == temporary_redirect_to(
         pyramid_request.route_url("admin.instances")
     )
-
-
-class TestEmptyStringInt:
```
marcospri, line 34: Ha! I removed this but nod added the new file.

Fixing

PR #4715, opened 2022-12-02
[WIP] Change admin layout to include tabs

marcospri:


jon-betts:
Some random thoughts. We've not really paid attention to this layout, and I think if we are opening the "make it nicer box", there are some other things we might try to do on the way.

## The theme is wasting a lot of vertical space

tabbed_ui

It's partly the theme, but it's also how we are using it. Some ideas for trying to get the "fold" where we start seeing info about the application instance higher up the page (although it applies to others too)

* Can we move the title into the menu bar? Or a sub-menu bar (would be a good place for high level buttons like "New organization")
* Can we cut down the white space below the title?
* Can we move the org and registration into a more tablular structure?
* Ideas include, moving the relationship type to the left and the view button to the right instead of vertical

## Some suggestions for locations / names

### Application Instance

 * "Notes" and "Created" are in the "Tool Consumer" box but they aren't related to the Tool Consumer
 * I'm assuming the save buttons will be fixed later
 * Info / settings is a strange divide. You can change things in info, so it's not read-only (which this might imply). What about "Settings / Content Integrations"?
 * VitalSource / JSTOR looks like it could do with it's own box?
 * Maybe try two columns on the content integration page? Might mean you can see them all at once?

### Organization

 * I really like having the Danger Zone in it's own tab, might be worth it just for that
 * If not, probably put the Danger Zone as the last item in the page?

### Application instances / LTI Registrations / Organizations search pages

 * Most menu systems let you highlight an item
 * Could we use this instead of repeating it as a title?
 * This wouldn't apply if we used a sub-menu, to accomodate "New Registration" etc.
 * If you click through the three pages, the placements of the buttons are inconsistent

marcospri:
Now tracked in https://github.com/hypothesis/lms/issues/4724

PR #4722, opened 2022-12-05
Generalize access to the current LMS files

marcospri:
Refactor before D2L files

---

Create a new plugin to generalize the behaviour of "integrate with the LMS files".

This is similar across different LMS, canvas needing a few exceptions, the future D2L one being very similar to blackboard (not in this PR).

Close to this code we have third party file pickers (OneDrive, Google...) but those are a difference concept both on the backend and the effect in the UI, not tackled in this PR.

### DIFF
--- a/lms/product/blackboard/_plugin/files.py
+++ b/lms/product/blackboard/_plugin/files.py
@@ -1,0 +1,5 @@
+from lms.product.plugin.files import FilesPlugin
+
+
+class BlackboardFilesPlugin(FilesPlugin):
+    missing_files_help_link = "https://web.hypothes.is/help/bb-files"
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -18,7 +18,7 @@ import URLPicker from './URLPicker';
  *
  * @typedef {import('./FilePickerApp').ErrorInfo} ErrorInfo
  *
- * @typedef {'blackboardFile'|'canvasFile'|'jstor'|'url'|'vitalSourceBook'|null} DialogType
+ * @typedef {'lmsFile'|'jstor'|'url'|'vitalSourceBook'|null} DialogType
  *
  * @typedef {import('../utils/content-item').Content} Content
  */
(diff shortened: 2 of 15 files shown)


lms/product/blackboard/_plugin/files.py
```diff
@@ -0,0 +1,5 @@
+from lms.product.plugin.files import FilesPlugin
+
+
+class BlackboardFilesPlugin(FilesPlugin):
```
marcospri, line 4: BB (and soon D2L) behave like the general case. Only need to change this URL.

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -147,28 +147,14 @@ export default function ContentSelector({
     case 'url':
       dialog = <URLPicker onCancel={cancelDialog} onSelectURL={selectURL} />;
       break;
-    case 'canvasFile':
+    case 'lmsFile':
```
jon-betts, line 150: Can you get most of what you want by changing this to a default case or a list of values?

lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -239,18 +225,18 @@ export default function ContentSelector({
...
               data-testid="canvas-file-button"
             >
               Select PDF from Canvas
             </LabeledButton>
           )}
-          {blackboardFilesEnabled && (
+          {lmsFilesEnabled && family === 'BlackboardLearn' && (
             <LabeledButton
-              onClick={() => selectDialog('blackboardFile')}
+              onClick={() => selectDialog('lmsFile')}
```
jon-betts, line 239: This feels like we are having to work around the fact we've collapsed different picker types into the same type?

jon-betts (commented):
I'm not opposed to starting with one set of files or another completely, but I wonder if it's the best way to go, as we have a list of things we want to unify. By not including all the examples we have it feels like we're more likely to make avoidable errors in our abstraction because we're not using the info we have.

I think assuming there is a single way to get files for an LMS has a few effects:

 * I don't think it's true - Canvas has a few different ways of storing files, we just happen to use one. This could change or not apply to a future
 * I think it sets us up badly for generalization

I think I'd be a lot happier if this was the start of a content integration setup which would apply to all of them equally, rather than caving out "LMS file" as a unique concept. At a minimum I think it should be an an array of available integrations.

I wonder if we don't have time to do the whole thing, if biting off half is actually worse? A bit more copy paste isn't actually the worst thing in the world.

## tl;dr

 * I think the "LMS file" is a specific sub-case of the general content provider
 * By starting with that specific sub-case I worry we're baking in all sorts of specifics to that situation

PR #4723, opened 2022-12-05
Admin pages layout changes

marcospri:
Low hanging fruit for:

- https://github.com/hypothesis/lms/issues/4724

## Changes summary

- Navbar has a different color than the rest of the page (me being colorblind that color is not very flashy)
- Less (but probably not yet done here) white space around the page title.
- Consistency of the buttons on the search pages (both the ones at the top and at the bottom).
- Slightly smaller text size for empty organizations/registrations.

### DIFF
--- a/lms/templates/admin/base.html.jinja2
+++ b/lms/templates/admin/base.html.jinja2
@@ -9,7 +9,7 @@
   <body>
 
   {% block base_navbar %}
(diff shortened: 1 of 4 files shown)


jon-betts (approved):
A nice collection of small improvements. Each might not do a lot, but when you've added enough of these it can make a really significant difference.

Issue #4699, opened 2022-09-22
New grade item is not created in D2L when LTI 1.3 LMS assignment is launched


mattdricker:

marcospri:
Further discussion here

https://hypothes-is.slack.com/archives/C04BXQNU145/p1670491621331899

mattdricker:
In the time since this issue was first written up, D2L has now added to the New Content Experience the ability for instructors to add their own grading items. This upends the decision making here about whether to enable to enable automatic grading item creation or not.

We will need to have more discussion with Product, Success, and Support teams about the best way forward here.

marcospri:
This had a new feature release for it but it seems like not the right one judging for the discussion https://hypothes-is.slack.com/archives/C04BXQNU145/p1670491621331899

In needs product discussion on how to proceed.

PR #4725, opened 2022-12-05
New misc plugin

marcospri:
For:

Placeholder location for product differences until a better abstraction
is found.

The exact methods included in this first version correspond to the ones
that will be overridden by a D2L feature coming next.

## Testing

No functional changes, sanity check the changes with:

- Open https://hypothesis.instructure.com/courses/319/assignments/3308 as a student to test the aud claim change.

- Create a new assignment in blackboard (`Hypothesis (localhost, make devdata)`) to exercise the post configure hook and the is_gradable method.

### DIFF
--- a/lms/product/plugin/misc.py
+++ b/lms/product/plugin/misc.py
@@ -1,0 +1,42 @@
+from pyramid.request import Request
+
+from lms.models import LTIParams, LTIRegistration
+
+
+class MiscPlugin:
+    """
+    Plugin for product differences that don't have a better abstractions yet.
+
+    Works as a placeholder location to centralize LMS differences here without
+    the need to create bigger abstractions before it's clear what that
+    abstraction should look like.
+
+    The aim also includes to remove any checks around `product.family` around
+    the code base.
+
+    New methods here should not try to get a very tight API as easier to
+    refactor once multiple MS have the same issue vs getting the right
+    parameters in the first occurrence.
+
+    Once any of these is implemented by more than one product or a group of
+    methods looks like it could belong to their own plugin it's time to
+    refactor them out.
+    """
+
+    def post_configure_assignment(self, request: Request):  # pragma: nocover
+        """
+        Run any actions needed after configuring an assignment.
+
+        This doesn't apply on deep linked setups where:
+            - The exact moment of configuration is not know.
+            - The deep linking message could include details about grading.
+        """
+
+    def is_assignment_gradable(self, lti_params: LTIParams) -> bool:
+        """Check if the assignment of the current launch is gradable."""
+
+        return bool(lti_params.get("lis_outcome_service_url"))
+
+    def get_ltia_aud_claim(self, lti_registration: LTIRegistration) -> str:
+        """Get the value of the `aud` claim used in LTI advantage requests."""
+        return lti_registration.token_url
--- a/lms/product/plugin/misc.py
+++ b/lms/product/plugin/misc.py
@@ -34,7 +34,6 @@ class MiscPlugin:
 
     def is_assignment_gradable(self, lti_params: LTIParams) -> bool:
         """Check if the assignment of the current launch is gradable."""
-
         return bool(lti_params.get("lis_outcome_service_url"))
 
     def get_ltia_aud_claim(self, lti_registration: LTIRegistration) -> str:
--- a/lms/services/ltia_http.py
+++ b/lms/services/ltia_http.py
@@ -9,11 +9,16 @@ class LTIAHTTPService:
     """Send LTI Advantage requests and return the responses."""
 
     def __init__(
-        self, lti_registration: LTIRegistration, jwt_service: JWTService, http
+        self,
+        lti_registration: LTIRegistration,
+        aud_claim: str,
+        jwt_service: JWTService,
+        http,
     ):
         self._lti_registration = lti_registration
         self._jwt_service = jwt_service
         self._http = http
+        self._aud_claim = aud_claim
 
     def request(self, method, url, scopes, headers=None, **kwargs):
         headers = headers or {}
@@ -39,7 +44,7 @@ class LTIAHTTPService:
                 "iat": now,
                 "iss": self._lti_registration.client_id,
                 "sub": self._lti_registration.client_id,
-                "aud": self._lti_registration.token_url,
+                "aud": self._aud_claim,
                 "jti": uuid.uuid4().hex,
             }
         )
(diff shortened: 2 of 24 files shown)


lms/product/plugin/misc.py
```diff
@@ -0,0 +1,26 @@
...
+    The aim also includes to remove any checks around product.family around the code base.
+
+    New methods here should not try to get a very tight API as easier to refactor once multiple
+    LMS have the same issue vs getting the right parameters in the first occurrence.
+
+    Once any of these is implemented by more than one product or a group of
+    methods looks like it could belong to their own plugin it's time to refactor them out.
+    """
+
+    def post_configure_assignment(self):  # pragma: nocover
```
marcospri, line 17: No cover as the default implementation is "nothing".

lms/services/ltia_http.py
```diff
@@ -58,10 +63,13 @@ def _get_access_token(self, scopes):
...
+    lti_registration = (
+        request.find_service(name="application_instance").get_current().lti_registration
+    )
+
     return LTIAHTTPService(
-        request.find_service(name="application_instance")
-        .get_current()
-        .lti_registration,
+        lti_registration,
+        request.product.plugin.misc.get_ltia_aud_claim(lti_registration),
```
marcospri, line 72: Injecting the value directly rather than using the plugin in the service.

lms/product/plugin/misc.py
marcospri, line 17: This could grow to have pre/post hooks for show document/configure etc.

Only needs this one now so stopping here.

tests/unit/lms/product/plugin/grouping_test.py
```diff
@@ -7,7 +7,7 @@
 from tests import factories
 
 
-class TestGroupingServicePlugin:
+class TestGroupingPlugin:
```
marcospri, line 10: Leftover for previous renaming.

lms/product/plugin/misc.py
```diff
@@ -0,0 +1,26 @@
...
+    bigger abstractions before it's clear what that abstraction should look like.
+
+    The aim also includes to remove any checks around product.family around the code base.
+
+    New methods here should not try to get a very tight API as easier to refactor once multiple
+    LMS have the same issue vs getting the right parameters in the first occurrence.
+
+    Once any of these is implemented by more than one product or a group of
+    methods looks like it could belong to their own plugin it's time to refactor them out.
+    """
```
jon-betts, line 15:
```suggestion
    """
    Plugin for product differences that don't have a better abstractions yet.

    Works as a placeholder location to centralize LMS differences here without
    the need to create bigger abstractions before it's clear what that
    abstraction should look like.

    The aim also includes to remove any checks around `product.family` around
    the code base.

    New methods here should not try to get a very tight API as easier to
    refactor once multiple MS have the same issue vs getting the right
    parameters in the first occurrence.

    Once any of these is implemented by more than one product or a group of
    methods looks like it could belong to their own plugin it's time to
    refactor them out.
    """
```

Nit-picky 79 char pep8 stuff. I have a grey line in my IDE that shows this limit which is quite helpful.

Apart from anything else it ensures it will show up nicely in Github without line wrapping as you can see above.

lms/product/plugin/misc.py
```diff
@@ -0,0 +1,26 @@
...
+    def post_configure_assignment(self, request):  # pragma: nocover
+        """Run any actions needed after configuring an assignment."""
+
+    def is_assignment_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        return bool(lti_params.get("lis_outcome_service_url"))
+
+    def get_ltia_aud_claim(self, lti_registration):
+        """Return the value of the `aud` claim used in LTI advantage requests."""
+        return lti_registration.token_url
```
jon-betts, line 26:
```suggestion
    def post_configure_assignment(self, request: Request):  # pragma: nocover
        """Run any actions needed after configuring an assignment."""

    def is_assignment_gradable(self, lti_params: LTIParams) -> bool:
        """Check if the assignment of the current launch is gradable."""

        return bool(lti_params.get("lis_outcome_service_url"))

    def get_ltia_aud_claim(self, lti_registration: LTIRegistration) -> str:
        """Get the value of the `aud` claim used in LTI advantage requests."""

        return lti_registration.token_url
```
Worth adding types?

lms/views/lti/basic_launch.py
```diff
@@ -128,6 +128,9 @@ def configure_assignment(self):
         if group_set := self.request.parsed_params.get("group_set"):
             extra["group_set_id"] = group_set
 
+        # Make any product-specific actions after configuring the assignment
+        self.request.product.plugin.misc.post_configure_assignment(self.request)
```
jon-betts, line 132: Should this also be in the equivalent `deep_linked` method?

lms/product/plugin/misc.py
```diff
@@ -0,0 +1,26 @@
+class MiscPlugin:
+    """
+    Plugin for product differences that don't have a better abstractions around them.
+
+    Works as a placeholder location to centralize LMS differences here without the need to create
+    bigger abstractions before it's clear what that abstraction should look like.
```
jon-betts, line 6: :+1: I think this is a really important thing to note. Otherwise we'll end up backronyming this into the "Multi Instructure Service Configurator" or something one day.

jon-betts (commented):
This looks like a reasonable approach to me. We know we need some entry points, but we don't want to get overly help up on worrying about where to put them just now.

We can let them accumulate a bit and then see what we've got and try and organise it a bit better. This runs a bit of a risk of becoming a dumping ground, so we need to commit to sorting it out one day. This is the definition of consciously taking on technical debt at this point.

I have one question which is we add a `post_configure_assignment`, but it doesn't seem to apply in the `deep_linked` situation. Should it?

lms/product/plugin/misc.py
```diff
@@ -0,0 +1,26 @@
...
+
+    New methods here should not try to get a very tight API as easier to refactor once multiple
+    LMS have the same issue vs getting the right parameters in the first occurrence.
+
+    Once any of these is implemented by more than one product or a group of
+    methods looks like it could belong to their own plugin it's time to refactor them out.
+    """
+
+    def post_configure_assignment(self, request):  # pragma: nocover
+        """Run any actions needed after configuring an assignment."""
```
jon-betts, line 18: Worth putting any advice about what _not_ to do in here? We have policies about not making loads of synchronous calls in general.

Or should we just trust our own good judgement?

marcospri: Not sure, this is mostly about what you'll do for an specific product, I reckon the same judgement should apply here than in the rest of the code base.

lms/views/lti/basic_launch.py
marcospri: Added a note about deep linking in the plugin.

lms/services/ltia_http.py
marcospri: You are right, it doesn't matter right now but the usual pattern it's to inject the service.

I was a bit hesitant to depend on the "Misc" plugin but if this becomes "LTIAPlugin" it just a change on the interface.

jon-betts (approved):
I think this looks good. I'm not sure exactly what I was looking for in the testing instructions, but nothing broke, so I assume that's good!

marcospri:
Merged https://github.com/hypothesis/lms/pull/4726 here

PR #4726, opened 2022-12-06
Auto create D2L grade items

marcospri:
For:

Workaround for  a behavior difference in D2L in LTI 1.1 vs 1.3.

In LTI 1.1 creating an assignment also creates the "container" to hold grades for it. In LTI1.3 the container is not created.

The fix is to create it from our end as soon as the assignment is configured. We use a setting/feature flag to control the roll out of the fix and in case D2L changes the behavior again.

## LTI 1.1 behavior

- Create a new assignment in https://aunltd.brightspacedemo.com/d2l/le/content/6782/Home
- "Existing activites" -> "External Learning Tools" -> "Create new LTI link " at the bottom 

Use http://localhost:8001/lti_launches for LTI1.1

Finish the setup and launch the assignment. Note the grading bar at the top.

## LTI1.3 (without enabling the new setting)

- Same as in LTI1.1 but use `https://localhost:48001/lti_launches` as the LTI URL and make sure `Hypothesis (https localhost)` is picked as "Tool"
- Open https://localhost:48001/lti/1.3/oidc in a new window and ok the security exception
- Pick a pdf for easier testing https://h.readthedocs.io/_/downloads/client/en/latest/pdf/
- Note that there's no grading bar here.

Now we'll check the manual work around we asked schools to do:

 * Click "Add a grade item..." at the bottom
 * Pick something?
 * Reload the assignment, now you get the grading bar.

## LTI1.3 (with the new setting)

- Enable the new setting (`Create grade items`) over http://localhost:8001/admin/instance/id/106/
- Repeat the assignment creation for LTI1.3 , now the grading bar will show without creating a grade item manually.
- The grade item will show up in the "Assessment" section after a refresh.

### DIFF
--- a/lms/models/lti_params.py
+++ b/lms/models/lti_params.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import List, Optional
 
 CLAIM_PREFIX = "https://purl.imsglobal.org/spec/lti/claim"
 
@@ -18,7 +18,7 @@ class LTIParams(dict):
     v11 and the object's dict interface.
     """
 
-    def __init__(self, v11: dict, v13: dict = None):
+    def __init__(self, v11: dict, v13: Optional[dict] = None):
         super().__init__(v11)
         self.v13 = v13
 
@@ -122,6 +122,10 @@ _V11_TO_V13 = (
     (
         "lis_outcome_service_url",
         ["https://purl.imsglobal.org/spec/lti-ags/claim/endpoint", "lineitem"],
+    ),
+    (
+        "lineitems",
+        ["https://purl.imsglobal.org/spec/lti-ags/claim/endpoint", "lineitems"],
     ),
     ("lis_result_sourcedid", ["sub"]),
     (
--- a/lms/product/d2l/_plugin/misc.py
+++ b/lms/product/d2l/_plugin/misc.py
@@ -27,9 +27,9 @@ class D2LMiscPlugin(MiscPlugin):
         resource_link_id = lti_params.get("resource_link_id")
         resource_link_title = lti_params.get("resource_link_title")
 
-        if not lti_grading_service.read_lineitems(
-            lti_params.get("lineitems"), resource_link_id
-        ):
+        # If we already identified this assignment as gradable no need to
+        # create anything
+        if not super().is_assignment_gradable(lti_params):
             # If there are no existing lineitems, create one.
             lti_grading_service.create_lineitem(
                 lti_params.get("lineitems"),
--- a/lms/product/d2l/_plugin/misc.py
+++ b/lms/product/d2l/_plugin/misc.py
@@ -3,8 +3,8 @@ from lms.services.lti_grading.interface import LTIGradingService
 
 
 class D2LMiscPlugin(MiscPlugin):
-    def __init__(self, settings: Settings):
-        self._settings = settings
+    def __init__(self, create_lineitem: bool):
+        self._create_lineitem = create_lineitem
 
     def post_configure_assignment(self, request):
         """
@@ -20,7 +20,7 @@ class D2LMiscPlugin(MiscPlugin):
         lti_params = request.lti_params
         lti_grading_service = request.find_service(LTIGradingService)
 
-        if not self._settings.create_lineitem:
+        if not self._create_lineitem:
             # No need to do anything if this option is off.
             return
 
@@ -54,4 +54,4 @@ class D2LMiscPlugin(MiscPlugin):
 
     @classmethod
     def factory(cls, _context, request):
-        return cls(request.product.settings)
+        return cls(request.product.settings.custom.get("create_lineitem", False))
--- a/lms/product/d2l/_plugin/misc.py
+++ b/lms/product/d2l/_plugin/misc.py
@@ -3,24 +3,24 @@ from lms.services.lti_grading.interface import LTIGradingService
 
 
 class D2LMiscPlugin(MiscPlugin):
-    def __init__(self, create_lineitem: bool):
-        self._create_lineitem = create_lineitem
+    def __init__(self, create_line_item: bool):
+        self._create_line_item = create_line_item
 
     def post_configure_assignment(self, request):
         """
         Run any actions needed after configuring an assignment.
 
-        D2L doesn't create the container that holds grades (lineitems) LTI 1.3
+        D2L doesn't create the container that holds grades (line item) LTI 1.3
         assignments.
 
         As a work-around, as soon as we configure the assignment on our end
-        we'll use the LTIA grading API to create a new lineitem we can use
+        we'll use the LTIA grading API to create a new line item we can use
         later to record the grade.
         """
         lti_params = request.lti_params
         lti_grading_service = request.find_service(LTIGradingService)
 
-        if not self._create_lineitem:
+        if not self._create_line_item:
             # No need to do anything if this option is off.
             return
 
@@ -30,12 +30,7 @@ class D2LMiscPlugin(MiscPlugin):
         # If we already identified this assignment as gradable no need to
         # create anything
         if not super().is_assignment_gradable(lti_params):
-            # If there are no existing lineitems, create one.
-            lti_grading_service.create_lineitem(
-                lti_params.get("lineitems"),
-                resource_link_id,
-                resource_link_title,
-            )
+            lti_grading_service.create_line_item(resource_link_id, resource_link_title)
 
     def is_assignment_gradable(self, lti_params):
         """Check if the assignment of the current launch is gradable."""
@@ -54,4 +49,4 @@ class D2LMiscPlugin(MiscPlugin):
 
     @classmethod
     def factory(cls, _context, request):
-        return cls(request.product.settings.custom.get("create_lineitem", False))
+        return cls(request.product.settings.custom.get("create_line_item", False))
--- a/lms/product/d2l/_plugin/misc.py
+++ b/lms/product/d2l/_plugin/misc.py
@@ -17,20 +17,16 @@ class D2LMiscPlugin(MiscPlugin):
         we'll use the LTIA grading API to create a new line item we can use
         later to record the grade.
         """
-        lti_params = request.lti_params
-        lti_grading_service = request.find_service(LTIGradingService)
-
-        if not self._create_line_item:
-            # No need to do anything if this option is off.
+        # Nothing to do if line item creation is off, or it's already there
+        if not self._create_line_item or super().is_assignment_gradable(
+            request.lti_params
+        ):
             return
 
-        resource_link_id = lti_params.get("resource_link_id")
-        resource_link_title = lti_params.get("resource_link_title")
-
-        # If we already identified this assignment as gradable no need to
-        # create anything
-        if not super().is_assignment_gradable(lti_params):
-            lti_grading_service.create_line_item(resource_link_id, resource_link_title)
+        request.find_service(LTIGradingService).create_line_item(
+            resource_link_id=request.lti_params.get("resource_link_id"),
+            label=request.lti_params.get("resource_link_title"),
+        )
 
     def is_assignment_gradable(self, lti_params):
         """Check if the assignment of the current launch is gradable."""
--- a/lms/product/d2l/_plugin/misc.py
+++ b/lms/product/d2l/_plugin/misc.py
@@ -1,0 +1,57 @@
+from lms.product.plugin.misc import MiscPlugin
+from lms.services.lti_grading.interface import LTIGradingService
+
+
+class D2LMiscPlugin(MiscPlugin):
+    def __init__(self, settings: Settings):
+        self._settings = settings
+
+    def post_configure_assignment(self, request):
+        """
+        Run any actions needed after configuring an assignment.
+
+        D2L doesn't create the container that holds grades (lineitems) LTI 1.3
+        assignments.
+
+        As a work-around, as soon as we configure the assignment on our end
+        we'll use the LTIA grading API to create a new lineitem we can use
+        later to record the grade.
+        """
+        lti_params = request.lti_params
+        lti_grading_service = request.find_service(LTIGradingService)
+
+        if not self._settings.create_lineitem:
+            # No need to do anything if this option is off.
+            return
+
+        resource_link_id = lti_params.get("resource_link_id")
+        resource_link_title = lti_params.get("resource_link_title")
+
+        if not lti_grading_service.read_lineitems(
+            lti_params.get("lineitems"), resource_link_id
+        ):
+            # If there are no existing lineitems, create one.
+            lti_grading_service.create_lineitem(
+                lti_params.get("lineitems"),
+                resource_link_id,
+                resource_link_title,
+            )
+
+    def is_assignment_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        if self._create_line_item and lti_params["lti_version"] == "1.3.0":
+            # D2L doesn't automatically create a line item for assignments by
+            # default like it does for 1.1. If we are creating them
+            # automatically all assignments will be gradable.
+            return True
+
+        return super().is_assignment_gradable(lti_params)
+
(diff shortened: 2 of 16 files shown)


lms/product/d2l/_plugin/misc.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def post_configure_assignment(self, request):
+        """
+        Run any actions needed after configuring an assignment.
+
+        D2L doesn't create the container that holds grades (lineitems) LTI 1.3 assignments.
+
+        As a work around, as soon as we configure the assignment on our end
+        we'll use the LTIA grading API to create a new lineitem we can use later to record the grade.
+        """
```
jon-betts, line 18:
```suggestion
        """
        Run any actions needed after configuring an assignment.

        D2L doesn't create the container that holds grades (lineitems) LTI 1.3
        assignments.

        As a work-around, as soon as we configure the assignment on our end
        we'll use the LTIA grading API to create a new lineitem we can use
        later to record the grade.
        """
```

lms/product/d2l/_plugin/misc.py
```diff
@@ -0,0 +1,54 @@
...
+                lti_params.get("lineitems"),
+                resource_link_id,
+                resource_link_title,
+            )
+
+    def is_assignment_gradable(self, lti_params):
+        """Check if the assignment of the current launch is gradable."""
+        if self._settings.create_lineitem and lti_params["lti_version"] == "1.3.0":
+            # D2L doesn't automatically create a line item for assignments by default like it does for 1.1.
+            # If we are creating them automatically all assignments will be gradable.
```
jon-betts, line 43:
```suggestion
        """Check if the assignment of the current launch is gradable."""

        if self._settings.create_lineitem and lti_params["lti_version"] == "1.3.0":
            # D2L doesn't automatically create a line item for assignments by
            # default like it does for 1.1. If we are creating them
            # automatically all assignments will be gradable.
            return True
```

lms/product/d2l/_plugin/misc.py
```diff
@@ -0,0 +1,54 @@
...
+            return True
+
+        return super().is_assignment_gradable(lti_params)
+
+    def get_ltia_aud_claim(self, lti_registration):
+        # In D2L this value is always the same and different from lti_registration.token url as in other LMS
```
jon-betts, line 49:
```suggestion
        # In D2L this value is always the same and different from
        # `lti_registration.token_url` as in other LMS
```

lms/product/product.py
```diff
@@ -47,8 +47,12 @@ class Settings:
     groups_enabled: bool = False
     """Is the course groups feature enabled"""
 
+    create_lineitem: bool = False
+    """Create lineitem on assignment configuration"""
```
jon-betts, line 51: I don't get what this setting is for? This looks specific to the implementation of the misc plugin for D2L, rather than a general thing.

Why would we ever turn this off for D2L or on for anyone else?

lms/product/product.py
```diff
@@ -47,8 +47,12 @@ class Settings:
...
+
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
+        self.create_lineitem = product_settings.get("create_lineitem", False)
```
jon-betts, line 55: Are we doing this here because it's the only place to do it? This might suggest a flaw in the design, because we will accumulate every possible setting for every possible product no matter which plugin it relates to here.

jon-betts (commented):
I think I'm missing a bit of context here without which it's hard to review this.

I don't understand the distinction between grades / scores / lineitems etc. Are these all separate things? I suspect we might be sticking too closely to terms used by LTI in some places and mixing them with our own definitions elsewhere, but I'm not sure.

My preference would be to work out what a normal human being who works for us would call one of these things. I doubt teachers consider themselves to be manipulating `lineitems` on `resource_link_ids` for example, and talking in those terms kind of ensures we'll be misunderstood by people inside and outside of the business.

I'm also not sure why you appear to be able to turn this function off? Do you ever _not_ want this in D2L?

jon-betts:
https://www.imsglobal.org/spec/lti-ags/v2p0#nomenclature

Line item container
 * Belongs in an "LTI context" - course?
 * Has many "line items"

Line item
 * Associated with a "specific activity" - assignment
 * For a set of users
    
Result
 * Belongs to a specific "line item"
 * Associates to a specific user
 * A dict of stuff like description, numeric score, comments etc. 

Score:
 * Part of a "result"?
 * Seems like this relates to altering one part of a result 

### Some observations

* The docs don't refer to these as "lineitems" but "line items", so even if we are going to follow the spec, that's slightly less horrible.
* We don't deal with "scores" separately from "results" from what I can see, so we don't need to worry about that distinction
* I think in slightly more human comprehensible terms these might be:
   * Line item container - Course grade container
   * Line item - Assignment grade container
   * Result - Grade
 * "Container" is a lame name - we can probably do better

lms/services/lti_grading/_v13.py
```diff
@@ -62,6 +62,54 @@ def record_result(self, grading_id, score=None, pre_record_hook=None):
...
+        if resource_link_id:
+            params["resource_link_id"] = resource_link_id
+
+        return self._ltia_service.request(
+            "GET",
+            lineitems_url,
+            scopes=self.LTIA_SCOPES,
+            params=params,
+            headers={"Accept": "application/vnd.ims.lis.v2.lineitemcontainer+json"},
+        ).json()
```
jon-betts, line 85: I think the only reason we read the line items is to indirectly check if the line item exists for a specific assignment before we create it.

So I think a few things are true:

 * We never want to call this without a `resource_link_id` - So it probably shouldn't be optional
 * We don't actually care about the response, although we aren't doing much conversion here
 * i.e. This could be `has_line_items` if all we care about is existence
 * I think this could also be a private method called from `create_lineitem` so that `create_lineitem` becomes `upsert_lineitem`. 

Also, what happens if you create a line item for an assignment without checking if one is there already. Can you end up with two, overwrite it, or does it ignore you?

jon-betts:
I'm piecing this together bit by bit I think. I think the URL provided in the claim: "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint" - "lineitems"

Is like the item we currently call `lis_outcome_service_url`, which is injected into the `LTIGradingService`, except it's scoped to the course, rather than the assignment. I think this suggests we should be injecting the `lineitems` URL into the service the same way. We might also consider calling these something like "course_grading_url" and "assignment_grading_url"?

We should also probably check if it's present as the docs say:

>lineitems: the endpoint URL for accessing the line item container for the current context. **May be omitted if the tool has no permissions to access this endpoint.**

Which would indicate a possible permissions issue?

lms/models/lti_params.py
```diff
@@ -123,6 +123,10 @@ def _apply_canvas_quirks(lti_params, request):
         "lis_outcome_service_url",
         ["https://purl.imsglobal.org/spec/lti-ags/claim/endpoint", "lineitem"],
     ),
+    (
+        "lineitems",
+        ["https://purl.imsglobal.org/spec/lti-ags/claim/endpoint", "lineitems"],
+    ),
```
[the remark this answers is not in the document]
marcospri: This is the lti_params object I reckon we should follow the naming of the original parameter.

lms/services/lti_grading/_v13.py
marcospri: POST'ing multiple does create one per call. 

Removed the read_line_item method in any case in favor of just checking the presence of `lineitem` (via is_gradable_assignment)

marcospri:
Added a few changes:

- Settled on the LTI1.3 naming convention. So far the existing service has a 1 to 1 spec endpoint to method relationship so that works nicely IMO. 

  Once we add methods that combine various API endpoints or use them in a more sophisticated way we could use other names. This is similar to the approach for canvas/blackboard APIs.

  For the current usage I find being explicit about "line items" being a collection of line item is less (barely) confusing than trying to come up with a course/assignment distinction.



- Avoiding the call to "read_line_items". This is a very good improvement, saves an API call.
  
  "PUT" doesn't upsert.
  Multiple "POST"s do create multiple line item.

  But this doesn't matter as much, I was being overly conservative calling the read_items because:

     - We only call this on configure assignment. This is supposed to be only happens once per assignment and is backed by the presence of the assignment row in the DB
     - In case we did hit some race condition using the `is_gradable_assignment` of the base class as you suggest is good enough as it would be provided by the LMS after the creation of line item. If two or the same teacher configuring the same assignment at the same time becomes a real problem we should deal with it earlier.


> lineitems: the endpoint URL for accessing the line item container for the current context. May be omitted if the tool has no permissions to access this endpoint.

We are assuming we'd get this on gradable assigmietn for teachers (in LTI1.1 and 1.3) I reckon we can keep that assumption un-handled and get a loud sentry report when we hit an edge case.

lms/models/lti_params.py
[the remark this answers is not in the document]
jon-betts: Yep, that's fair enough.

In a more brave future we might decide on a name we like which is neither the 1.1 or the 1.3 name.

lms/product/d2l/_plugin/misc.py
```diff
@@ -0,0 +1,52 @@
...
+            # No need to do anything if this option is off.
+            return
+
+        resource_link_id = lti_params.get("resource_link_id")
+        resource_link_title = lti_params.get("resource_link_title")
+
+        # If we already identified this assignment as gradable no need to
+        # create anything
+        if not super().is_assignment_gradable(lti_params):
+            lti_grading_service.create_line_item(resource_link_id, resource_link_title)
```
jon-betts, line 33:
```suggestion
        # Nothing to do if line item creation is off, or it's already there
        if not self._create_line_item or super().is_assignment_gradable(
            request.lti_params
        ):
            return

        request.find_service(LTIGradingService).create_line_item(
            resource_link_id=request.lti_params.get("resource_link_id"),
            label=request.lti_params.get("resource_link_title"),
        )
```

I think we might have some left overs from refactors here?

lms/product/product.py
```diff
@@ -47,8 +47,12 @@ class Settings:
     groups_enabled: bool = False
     """Is the course groups feature enabled"""
 
+    custom: Optional[dict] = None
+    """Other non-standard settings."""
+
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
```
jon-betts, line 54: Do we want to `pop` here to ensure `custom` is _only_ non-standard things? This would prevent us from accessing `groups_enabled` directly and making different assumptions about the default.

lms/services/lti_grading/interface.py
```diff
@@ -1,8 +1,37 @@
...
+    Result
+        A result is usually a cell in the tool platform's gradebook; it is unique for a specific line item and user.
+
+    Score
+        A score represents the last score obtained by the student for the tool's activity. It also exposes the current status of the activity (like completed or in progress), and status of the grade.
+
+
+    These are the LTI1.3 concepts but with the exception of line item container they have a counter part in LTI 1.1:
+
+    https://www.imsglobal.org/spec/lti-ags/v2p0#migrating-from-basic-outcomes-service
```
jon-betts, line 22:
```suggestion
    Line item:
        A line item is usually a column in the tool platform's gradebook; it is
        able to hold the results associated with a specific activity 
        (assignment) for a set of users.

    Line item container:
        A line item container has an array of line items (for a course).

    Result:
        A result is usually a cell in the tool platform's gradebook; it is
        unique for a specific line item and user.

    Score:
        A score represents the last score obtained by the student for the
        tool's activity. It also exposes the current status of the activity
        (like completed or in progress), and status of the grade.

    These are the LTI1.3 concepts, but they have a counter-part in LTI 1.1 
    (except for the line item container):

    https://www.imsglobal.org/spec/lti-ags/v2p0#migrating-from-basic-outcomes-service
```

I'll never stop.

This is very good context for an LTI noob such as myself.

lms/services/lti_grading/interface.py
```diff
@@ -1,8 +1,37 @@
...
+    def __init__(self, line_item_url: str, line_item_container_url: str):
+        self.line_item_url = line_item_url
+        """
+        Identifies one line item to read/write grades to.
+
+        In LTI 1.1 this maps to the equivalent `lis_outcome_service_url` parameter.
+        """
+
+        self.line_item_container_url = line_item_container_url
+        """Identifies a container that might hold many line items."""
```
jon-betts, line 34:
```suggestion
        """
        Initialize the service.
        
        :param line_item_url: Identifies one line item to read/write grades 
            to. In LTI 1.1 this maps to the equivalent 
            `lis_outcome_service_url` parameter.
        :param line_item_container_url: Identifies a container that might hold 
            many line items
        """
        self.line_item_url = line_item_url
        self.line_item_container_url = line_item_container_url
```

This looks a bit weird? Not sure that it's wrong, but it looks like the way we document class level attributes, but in a method.

I think it would be good to pick one style or the other? If we added class level attributes we could document like this, or in a method I would expect to see `:param line_item_url:` etc.

lms/services/lti_grading/interface.py
```diff
@@ -31,3 +31,19 @@ def record_result(self, grading_id, score=None, pre_record_hook=None):
...
+        """
+        raise NotImplementedError()
+
+    def create_lineitem(self, lineitems_url, resource_link_id, label):
+        """
+        Create a new lineitem associated to one resource_link_id.
+
+        https://www.imsglobal.org/spec/lti-ags/v2p0#container-request-filters
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
jon-betts: I think we've decided to stick with the LTI 1.3 names for now.

lms/views/admin/application_instance.py
```diff
@@ -302,6 +302,7 @@ def update_instance(self):
             ("desire2learn", "client_id", str),
             ("desire2learn", "client_secret", aes_secret),
             ("desire2learn", "groups_enabled", bool),
+            ("desire2learn", "create_line_item", bool),
```
jon-betts, line 305: This would be a sweet thing to somehow bundle up into the plugins. But this would require some way of abstracting the form you'd like to see to be useful. Form abstractions inevitably lead to madness, so maybe that's not a good idea?

In any case, not an idea for now.

lms/services/lti_grading/_v13.py
```diff
@@ -56,12 +58,35 @@ def record_result(self, grading_id, score=None, pre_record_hook=None):
...
             headers={"Content-Type": "application/vnd.ims.lis.v1.score+json"},
         )
 
+    def create_line_item(self, resource_link_id, label, score_maximum=100):
+        """
+        Create a new line item associated to one resource_link_id.
+
+        https://www.imsglobal.org/spec/lti-ags/v2p0#container-request-filters
+
+        :param resource_link_id: ID of the assignment this line item will belong to.
```
jon-betts, line 73:
```suggestion
        :param resource_link_id: ID of the assignment this line item will 
            belong to.
```

tests/unit/lms/services/lti_grading/_v13_test.py
```diff
@@ -84,6 +84,26 @@ def test_record_result(self, svc, ltia_http_service):
...
+            svc.line_item_container_url,
+            scopes=svc.LTIA_SCOPES,
+            json={
+                "scoreMaximum": sentinel.score_maximum,
+                "label": sentinel.label,
+                "resourceLinkId": sentinel.resource_link_id,
+            },
+            headers={"Content-Type": "application/vnd.ims.lis.v2.lineitem+json"},
+        )
+        assert response == ltia_http_service.request.return_value.json.return_value
```
jon-betts, line 105: Yup :+1: . So straight forward!

jon-betts (approved):
I've given this a test and it seems to work as advertised. I think this is ready to go.

I've made a few stylistic-ish suggestions, but nothing major:

* Suggested a shorter layout for `D2LMiscPlugin.post_configure_assignment()`
* Use a `pop` in product settings before allocating custom?
* Various docstring suggestions

PR #4728, opened 2022-12-06
Fix product guessing for json request

marcospri:
Not all JSON request include the product hint we rely on. If not present use the next fallback methods.

This is not a problem in production right now but as we include more product/plugins (eg D2L in follow up PRs) we'll reach for the value of `product.family` in more contexts which might not currently send the hint.

This allows using the next fallback method in the chain. Ideally we'll settle just in one or two methods but that would be bigger PR/refactor.

### DIFF
--- a/lms/product/factory.py
+++ b/lms/product/factory.py
@@ -35,7 +35,7 @@ def _get_family(
         return Product.Family.BLACKBOARD
 
     # If we are in an API request, where we are forwarding the product type ourselves
-    if request.content_type == "application/json":
+    if request.content_type == "application/json" and "lms" in request.json:
         return Product.Family(request.json["lms"]["product"])
 
     # In an LTI launch we'll use the parameters available to guess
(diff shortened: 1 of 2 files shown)


lms/product/factory.py
```diff
@@ -35,7 +35,7 @@ def _get_family(
         return Product.Family.BLACKBOARD
 
     # If we are in an API request, where we are forwarding the product type ourselves
-    if request.content_type == "application/json":
+    if request.content_type == "application/json" and "lms" in request.json:
         return Product.Family(request.json["lms"]["product"])
 
     # In an LTI launch we'll use the parameters available to guess
```
marcospri, line 41: I can't comment on that line, but below in line 50 we mention `LTIOutcomesViews` that's a view that uses json but doesn't include the product hint.

jon-betts (approved):
This looks fair enough!

Do we actually want to check for "lms" and "product"? Or are we going to assume that if we get as far as "lms" being there the rest will be correct?

marcospri:
> Do we actually want to check for "lms" and "product"? Or are we going to assume that if we get as far as "lms" being there the rest will be correct?

Yeah I think having `lms` looks like we planned to do get it from there, if it fails probably worth checking in sentry.

PR #4734, opened 2022-12-08
Pin to tox<4

marcospri:
The release of tox 4.0.0 to PyPI has broken all our tox commands.

The reason for the breakage is that tox 3.x installs the new tox 4.0.0 in the `.tox/.tox` venv. Breaking changes in tox 4.0.0 then cause problems. In particular tox 4.0.0 breaks compatibility with all tox plugins. Fix this by telling it to install `tox<4`.


marcospri:
From https://github.com/hypothesis/cookiecutter-pyapp-test/pull/6

PR #4737, opened 2022-12-12
Migration for lti_role.scope

marcospri:
For:

Based on the model changes in https://github.com/hypothesis/lms/pull/4738

And split from https://github.com/hypothesis/lms/pull/4641

Adds a new, nullable, column to hold the scope of each role.


jon-betts (approved):
I've checked this out and it adds and removes the column as you'd expect.

Issue #4749, opened 2022-12-12
Refactor LTI role handling 


marcospri:
TLDR, with the following plan the role handling in LMS should be saner:

- [x] Add scope as a concept in LTIRole
- [x] Unify `LTIUser.roles` to use `LTIRole`
- [x] Take scope into account when checking for `is_instructor/learner`

### How roles currently work on LMS

We use  roles from the LTI launch in two contexts:

#### Authorization

Based on the roles from the launch we decide authorization (can you edit assignments? can you grade them?)

In LTI1.1 we look at the raw string from the launch and do very basic checking for instructor/learner:

https://github.com/hypothesis/lms/blob/b32b97a308e0e7cb8a6127fd80dde705683183f6/lms/models/lti_user.py#L33

In LTI1.3, where we are very often get sent roles in institution and system scopes we first process the roles. We exclude non-context roles as those are not relevant in the course.

https://github.com/hypothesis/lms/blob/b32b97a308e0e7cb8a6127fd80dde705683183f6/lms/models/lti_params.py

Once that filtering is done they arrive at the same `is_instructor`|`is_learner` functions that only look at strings.

#### Data gathering

We often want to store the relationship between an user and "something" (a course, assignment...) in those cases we created a new LTIRole model that stores the raw role from the launch and a mapping to a known set of roles (admin, instructor and learner).

https://github.com/hypothesis/lms/blob/b32b97a308e0e7cb8a6127fd80dde705683183f6/lms/models/lti_role.py

We then use that model in `_membership` tables like:

https://github.com/hypothesis/lms/blob/b32b97a308e0e7cb8a6127fd80dde705683183f6/lms/models/assignment_membership.py#L29

### Current issues

In LTI roles are scoped to one of system, institution or context (course).

In 1.3 roles are almost always scoped while in 1.1 only sometimes.

When we first released 1.3 I was not aware of this feature and we continue parsing the roles as usual causing issues because for example an "instructor" role with the institution scope meant that the user was a teacher in some course but not in the current one (the only one our app cares about).

The fix for this was to ignore institution roles in LTI1.3 

- https://github.com/hypothesis/lms/pull/4373
 
We never tackled this issue in 1.1 so we probably have some weird edge cases bugs due to this since the beginning.

- https://github.com/hypothesis/lms/issues/4411

Now we have reports of admin users not being able to create courses when they are only admins (but not instructors in the course) 

- https://github.com/hypothesis/lms/issues/4702

[...]

PR #4750, opened 2022-12-12
Roles scope parsing

marcospri:
For:

Introduces the `scope` concept in the LTIRole table and simplifies the parsing.

Note that we don't yet based any authorization decisions on this table but having this here will make that possible.

The plan is to, after we have this out for a little bit,  connect LTIUser with LTIRole and have only one role parsing/checking location in the code base.

## Testing

- Launch any assignment, maybe with both a student and teacher:

- Check scopes are now in the DB

```tox -qe dockercompose -- exec postgres psql -U postgres -c "select * from lti_role where scope is not null"```

```
 id |   value    |    type    | scope  
----+------------+------------+--------
  7 | Learner    | learner    | course
  4 | Instructor | instructor | course
(2 rows)

```

### DIFF
--- a/lms/models/lti_role.py
+++ b/lms/models/lti_role.py
@@ -54,8 +54,11 @@ class LTIRole(BASE):
     @value.setter
     def value(self, value):
         self._value = value
-        # pylint: disable=protected-access
-        self.scope, self.type = _RoleParser._parse_role(value)
+        self.update_from_value()
+
+    def update_from_value(self):
+        """Set scope and type based on `_value`."""
+        self.scope, self.type = _RoleParser.parse_role(self._value)
 
 
 class _RoleParser:
@@ -126,7 +129,7 @@ class _RoleParser:
     }
 
     @classmethod
-    def _parse_role(cls, role) -> Tuple[RoleScope, RoleType]:
+    def parse_role(cls, role) -> Tuple[RoleScope, RoleType]:
         """Parse roles according to the expected values from the specs."""
         role_parts = {}
         for regex in cls._ROLE_REGEXP:
--- a/lms/models/lti_role.py
+++ b/lms/models/lti_role.py
@@ -99,98 +99,61 @@ class _RoleParser:
 
     @classmethod
     def parse_role(cls, role) -> RoleType:
-        if role := cls._parse_v13_role(role) or cls._parse_v11_role(role):
+        if role := cls._parse_v13_role(role):
             return role
 
         # If we can't match any, pick the most restrictive
         LOG.debug("Can't find role type for %s", role)
         return RoleType.LEARNER
 
-    _V11_INSTRUCTOR_STRINGS = (
-        "instructor",
-        "teachingassistant",
-        "contentdeveloper",
-        "faculty",
-        "mentor",
-        "staff",
-    )
-
-    @classmethod
-    def _parse_v11_role(cls, role) -> RoleType:
-        role = role.lower()
-
-        for string in cls._V11_INSTRUCTOR_STRINGS:
-            if string in role:
-                return RoleType.INSTRUCTOR
-
-        if "learner" in role:
-            return RoleType.LEARNER
-
-        if "admin" in role:
-            return RoleType.ADMIN
-
-        return None
-
-    _V13_PREFIX = "http://purl.imsglobal.org/vocab/lis/v2"
     _V13_ROLE_MAPPINGS = {
-        # https://www.imsglobal.org/spec/lti/v1p3/#lis-vocabulary-for-context-roles
-        # A.2.1 LIS vocabulary for system roles
-        # Core system roles
-        f"{_V13_PREFIX}/system/person#Administrator": RoleType.ADMIN,
-        f"{_V13_PREFIX}/system/person#None": RoleType.LEARNER,
-        # Non‑core system roles
-        f"{_V13_PREFIX}/system/person#AccountAdmin": RoleType.ADMIN,
-        f"{_V13_PREFIX}/system/person#Creator": RoleType.INSTRUCTOR,
-        f"{_V13_PREFIX}/system/person#SysAdmin": RoleType.ADMIN,
-        f"{_V13_PREFIX}/system/person#SysSupport": RoleType.ADMIN,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lms/models/lti_role.py [hunk 1]
```diff
@@ -55,103 +54,95 @@ def value(self):
...
+        re.compile(
+            r"http://purl.imsglobal.org/vocab/lti/(?P<scope>system)/(?P<type>\w+)\#(?P<sub_type>\w+)",
+        ),
+        # Non scoped roles are deprecated but allowed in LTI 1.3
+        # Conforming implementations MAY recognize the simple names for context roles;
+        # thus, for example, vendors can use the following roles interchangeably:
+        #   http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor
+        #   Instructor
+        # However, support for simple names in this manner for context roles is deprecated;
+        # by best practice, vendors should use the full URIs for all roles (context roles included).
```
jon-betts, line 85:
```suggestion
        # Non scoped roles are deprecated but allowed in LTI 1.3
        # Conforming implementations MAY recognize the simple names for context 
        # roles; thus, for example, vendors can use the following roles 
        # interchangeably:
        #   http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor
        #   Instructor
        # However, support for simple names in this manner for context roles is 
        # deprecated; by best practice, vendors should use the full URIs for 
        # all roles (context roles included).
```

tests/unit/lms/models/lti_role_test.py
```diff
@@ -81,20 +226,12 @@ def test_parse_lti_role_v13_exact_matches(self, value, role_type):
...
-    )
-    def test_it_sets_the_type_from_the_value(self, value, role_type):
-        lti_role = LTIRole(value=value)
-
-        assert lti_role.type == role_type
+        # We are ignoring the sub type, the spec says:
+        #
+        # LTI does not classify any of the sub-roles as a core role.
+        # Whenever a platform specifies a sub-role,
+        # by best practice it should also include the associated principal role;
```
jon-betts, line 233:
```suggestion
        # We are ignoring the subtype, the spec says:
        
        # LTI does not classify any of the sub-roles as a core role. Whenever a 
        # platform specifies a sub-role, by best practice it should also 
        # include the associated principal role.
```

jon-betts (changes requested):
This all looks good, and more robust than before. I've made a number of stylistic suggestions, but that's not why I'm marking this as requesting changes.

The actual change is that we have regexes with unescaped `.` in the regex which are wild. This means we will accept the following:

 * `"http://purlXimsglobalXorg/vocab/lis/v2/institution/person#Administrator"`

Where we should reject it. It's not a practical problem I suspect, but it's definitely unintended behavior.

tests/unit/lms/models/lti_role_test.py
```diff
@@ -71,7 +216,7 @@ def test_parse_lti_role_v13_exact_matches(self, value, role_type):
             ),
             (
                 f"{_V13_PREFIX}membership/Learner#Instructor",
-                RoleType.INSTRUCTOR,
+                RoleType.LEARNER,
```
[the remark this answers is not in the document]
marcospri: I rather leave this out for now until we have to deal with it for a reason

lms/services/lti_role_service.py
```diff
@@ -23,6 +23,13 @@ def get_roles(self, role_description: str) -> List[LTIRole]:
         # Pylint is confused about the `in_` for some reason
         roles = self._db.query(LTIRole).filter(LTIRole.value.in_(role_strings)).all()
 
+        for role in roles:
+            # LTIRole.value is a setter so this has the side effect changing
+            # both role.type and role.scope.
+            # This is useful when the logic for those fields change, updating the
+            # values in the DB and also exposing the right values in the rest to the application.
+            role.value = role.value
```
marcospri, line 31: This seems a bit silly without the comment

lms/services/lti_role_service.py
```diff
@@ -23,6 +23,13 @@ def get_roles(self, role_description: str) -> List[LTIRole]:
         # Pylint is confused about the `in_` for some reason
         roles = self._db.query(LTIRole).filter(LTIRole.value.in_(role_strings)).all()
 
+        for role in roles:
```
marcospri, line 26: In some instances we'd might want to do a DB migration and update all roles in one go but I don't see any downside to doing it as we see the roles from the launches.

jon-betts (approved):

tests/unit/lms/services/lti_role_service_test.py
```diff
@@ -35,6 +35,17 @@ def test_get_roles_with_existing_only(self, svc, existing_roles):
 
         assert roles == existing_roles
 
+    def test_get_roles_updates_value(self, svc):
```
jon-betts, line 38:
```suggestion
    def test_get_roles_updates_value(self, svc):
    # Create a role where the type and scope don't match the value
```

lms/services/lti_role_service.py
```diff
@@ -23,6 +23,13 @@ def get_roles(self, role_description: str) -> List[LTIRole]:
...
+            # LTIRole.value is a setter so this has the side effect changing
+            # both role.type and role.scope.
+            # This is useful when the logic for those fields change, updating the
+            # values in the DB and also exposing the right values in the rest to the application.
```
jon-betts, line 30:
```suggestion
            # LTIRole.value is a setter so this has the side effect changing
            # both role.type and role.scope.
            # This is useful when the logic for those fields change, updating 
            # the values in the DB and also exposing the right values in the 
            # rest to the application.
```

If we didn't want to use a side effect like this we could change the model object to be like this:

```python
    @value.setter
    def value(self, value):
        self._value = value

        self.update_from_value()

    def update_from_value(self):
        self.scope, self.type = _RoleParser._parse_role(self._value)
```

Then we could call the `update_from_value()` public method, rather than rely on it as a side effect.

Also looking at this now, having a private method on a private object doesn't seem right? Should `_parse_role` be public?

jon-betts (approved):
I still think this is fine, but I've gone over the specific commit in more detail.

I think we can avoid relying on a side effect, by making that side effect the main effect of a public method. Also `_parse_role` should be part of the public interface of the private object I think?

PR #4751, opened 2022-12-12
Add grouping parent information

jon-betts:
This adds parent info to groups, which I'd like to be in place so I can try some different queries out against the real data.


jon-betts:
This has been replaced

PR #4753, opened 2022-12-13
Simplify `BearerTokenSchema`

marcospri:
LTIUser, security policies and marshmallow schemas are all part of various authentication schemes which are all interconnected in sometimes, surprisingly ways.

This PR tries to untangle one of the pieces and simplify one of the marshmallow schemas to be more explicit about the data flow around LTIUser.

LTIUser is now in charge of it's own serialization in both BearerTokenSchema  and OAuthCallbackSchema.

## Testing 

There's no new behavior in this branch but to sanity check the changes make any launch that will use the sync API (which needs to get LTIUser using BearerTokenSchema)

- Launch https://hypothesis.instructure.com/courses/125/assignments/1833 successfully

### DIFF
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,9 +1,10 @@
-from typing import NamedTuple
+from dataclasses import dataclass
 
 from lms.models.h_user import HUser
 
 
-class LTIUser(NamedTuple):
+@dataclass
+class LTIUser:
     """An LTI user."""
 
     user_id: str
@@ -57,6 +58,21 @@ class LTIUser(NamedTuple):
             ),
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
+
+    def _asdict(self) -> dict:
+        """
+        Return a dict representing the LTIUser.
+
+        LTIUser is often serialized. We can pick here the exact representation.
+        """
+        return {
+            "user_id": self.user_id,
+            "roles": self.roles,
+            "tool_consumer_instance_guid": self.tool_consumer_instance_guid,
+            "display_name": self.display_name,
+            "application_instance_id": self.application_instance_id,
+            "email": self.email,
+        }
 
 
 def display_name(given_name, family_name, full_name):
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -59,7 +59,7 @@ class LTIUser:
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
 
-    def _asdict(self) -> dict:
+    def serialize(self) -> dict:
         """
         Return a dict representing the LTIUser.
 
--- a/lms/validation/authentication/_bearer_token.py
+++ b/lms/validation/authentication/_bearer_token.py
@@ -70,7 +70,7 @@ class BearerTokenSchema(PyramidRequestSchema):
         self._jwt_service = request.find_service(iface=JWTService)
         self._secret = request.registry.settings["jwt_secret"]
 
-    def authorization_param(self, lti_user):
+    def authorization_param(self, lti_user: LTIUser):
         """
         Return ``lti_user`` serialized into an authorization param.
 
@@ -84,7 +84,7 @@ class BearerTokenSchema(PyramidRequestSchema):
         """
         return self.dump(lti_user)["authorization"]
 
-    def lti_user(self, location):
+    def lti_user(self, location) -> LTIUser:
         """
         Return an models.LTIUser from the request's authorization param.
 
@@ -109,8 +109,6 @@ class BearerTokenSchema(PyramidRequestSchema):
           ``"Bearer <ENCODED_JWT>"`` format.
         :raise ValidationError: if the JWT's payload is invalid, for example if
           it's missing a required parameter
-
-        :rtype: LTIUser
         """
         try:
             return self.parse(location=location)
@@ -172,6 +170,6 @@ class BearerTokenSchema(PyramidRequestSchema):
             ) from err
 
     @marshmallow.post_load
-    def _make_user(self, data, **_kwargs):
+    def _make_user(self, data, **_kwargs) -> LTIUser:
         # See https://marshmallow.readthedocs.io/en/2.x-line/quickstart.html#deserializing-to-objects
         return LTIUser(**data)
--- a/lms/validation/authentication/_bearer_token.py
+++ b/lms/validation/authentication/_bearer_token.py
@@ -8,11 +8,6 @@ from lms.services import JWTService
 from lms.services.exceptions import ExpiredJWTError, InvalidJWTError
 from lms.validation import ValidationError
 from lms.validation._base import PyramidRequestSchema
-from lms.validation.authentication._exceptions import (
-    ExpiredSessionTokenError,
-    InvalidSessionTokenError,
-    MissingSessionTokenError,
-)
 
 __all__ = ("BearerTokenSchema",)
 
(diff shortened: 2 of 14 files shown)


lms/models/lti_user.py
```diff
@@ -58,6 +59,21 @@ def from_auth_params(application_instance, lti_core_schema):
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
 
+    def serialize(self) -> dict:
```
marcospri, line 62: Replaces `namedtuple`._asdict.

We don't want to have any general dataclass->dict conversion (dataclass.asdict for example) but rather have control of what gets seralized to be sent over the wire.

lms/validation/authentication/_bearer_token.py
```diff
@@ -109,69 +84,32 @@ def lti_user(self, location):
...
-        :rtype: LTIUser
         """
-        try:
-            return self.parse(location=location)
-        except ValidationError as error:
-            try:
-                authorization_error_message = error.messages[location]["authorization"][
-                    0
-                ]
-            except KeyError:
```
marcospri, line 122: Lots of guessing what exception to raise, but the right type is just below in a different method.

lms/validation/authentication/_exceptions.py
```diff
@@ -4,27 +4,12 @@
...
-    "ExpiredSessionTokenError",
-    "MissingSessionTokenError",
-    "InvalidSessionTokenError",
     "MissingStateParamError",
     "ExpiredStateParamError",
     "InvalidStateParamError",
 ]
 
 
-class ExpiredSessionTokenError(ValidationError):
```
marcospri, line 16: I don't think these types add much information. They are all `ValidationError` and just instantiated and raised once. 

Just raising `ValidationError` seems more straightforward and a new type can be created when different behavior is needed.

lms/validation/authentication/_bearer_token.py
```diff
@@ -36,55 +31,35 @@ class BearerTokenSchema(PyramidRequestSchema):
...
         LTIUser(user_id='...', application_instance_id='...', ...)
     """
 
-    user_id = marshmallow.fields.Str(required=True)
-    application_instance_id = marshmallow.fields.Int(required=True)
-    roles = marshmallow.fields.Str(required=True)
-    tool_consumer_instance_guid = marshmallow.fields.Str(required=True)
-    display_name = marshmallow.fields.Str(required=True)
-    email = marshmallow.fields.Str()
+    authorization = marshmallow.fields.Str(required=True)
```
marcospri, line 38: Changed this schema to behave like:

https://github.com/hypothesis/lms/blob/2fff474202d162dc3787354bea0425bb7b0f3cba/lms/validation/authentication/_oauth.py

I does have less features, it does not validate the actual fields of LTIUser but this is data we send to ourselves rather than user input. 

IMO Missing/wrong data here should raise and exception in sentry (like it will in this version) instead of being masked as 403/401 and lost in the monitoring.

lms/validation/authentication/_bearer_token.py
```diff
@@ -109,69 +84,32 @@ def lti_user(self, location):
           ``"Bearer <ENCODED_JWT>"`` format.
         :raise ValidationError: if the JWT's payload is invalid, for example if
           it's missing a required parameter
-
-        :rtype: LTIUser
         """
-        try:
-            return self.parse(location=location)
```
marcospri, line 116: I found this a bit difficult to follow, it does what's expected in marshmallow but I reckon this version (copied from the Oauth equivalent) is more explicit.

In the older version this would rely on `@marshmallow.pre_load` to get the authorization header and then marshmallow will iterate over every field declared in the schema to get the value as part of `data` and finally ```@marshmallow.post_load``` would create the LTIUser.

lms/models/lti_user.py
```diff
@@ -1,9 +1,10 @@
-from typing import NamedTuple
+from dataclasses import dataclass
 
 from lms.models.h_user import HUser
 
 
-class LTIUser(NamedTuple):
+@dataclass
```
marcospri, line 6: This being a dataclass doesn't seem like a huge improvement in this PR but it would make any further changes much easier.

lms/models/lti_user.py
```diff
@@ -58,6 +59,21 @@ def from_auth_params(application_instance, lti_core_schema):
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
 
+    def serialize(self) -> dict:
+        """
+        Return a dict representing the LTIUser.
+
+        LTIUser is often serialized. We can pick here the exact representation.
```
marcospri, line 66: With the change to `BearerTokenSchema` this is now the only place that controls how LTIUser is serialize.

lms/validation/authentication/_bearer_token.py
```diff
@@ -36,55 +31,35 @@ class BearerTokenSchema(PyramidRequestSchema):
...
-
-        >>> schema.load({'authorization': 'Bearer eyJ...YoM'}).data
-        LTIUser(user_id='...', application_instance_id='...', ...)
-
-    Or to parse an models.LTIUser out of a Pyramid
-    request's authorization param using webargs::
-
-        >>> from webargs.pyramidparser import parser
-        >>> parser.parse(s, request)
+        >>> schema.lti_user(location='headers')
```
marcospri, line 34: missing required location parameter.

lms/validation/authentication/_bearer_token.py
```diff
@@ -36,55 +31,35 @@ class BearerTokenSchema(PyramidRequestSchema):
         'Bearer eyJ...YoM'
 
         >>> # Deserialize the request's authorization param into an LTI user.
-        >>> schema.lti_user()
-        LTIUser(user_id='...', application_instance_id='...', ...)
-
-    The above are convenience methods that wrap webargs and marshmallow. But
```
marcospri, line 42: This seems like marshmallow docs, we don't use this schema in that way.

lms/models/lti_user.py
jon-betts: I'd like to hear more about this. `dataclass.asdict` would seem to do exactly what we want for now. If we had a test which locked this down, is there a risk I'm missing?

lms/validation/authentication/_exceptions.py
jon-betts: Yep. This all seems very roundabout.

tests/unit/conftest.py
```diff
@@ -138,12 +138,12 @@ def pyramid_request(db_session, application_instance, lti_v11_params):
 
 @pytest.fixture
 def user_is_learner(pyramid_request):
-    pyramid_request.lti_user = pyramid_request.lti_user._replace(roles="Learner")
+    pyramid_request.lti_user.roles = "Learner"
```
jon-betts, line 141: This is so much better.

lms/models/lti_user.py
jon-betts: I think this is definitely better. I'm not sure if `NamedTuples` were picked because they were immutable, or whether they were picked for their `dataclass` like behaviours. I suspect the immutable aspect might be a retro-justification?

We certainly want to mutate them regularly (in tests at a minimum).

lms/models/lti_user.py
jon-betts: Centralising this seems sensible.

lms/validation/authentication/_bearer_token.py
jon-betts: As a summary for myself, the change seems to be:

Before we used lots of marshmallow magic to move directly from the bearer token to the decomposed elements packed inside the JWT as attributes on schema. These attributes aren't actually used outside of this class (despite being public) and are then used to reasonably trivially create an `LTIUser` object.

Now we go part of the way by storing the token as a single attribute `authorization`, and then perform the JWT decomposition on demand when accessing the `LTIUser` object.

As a change in behavior we now no longer validate the contents of the JWT to ensure they form a valid user. We hope that any problems caused by missing elements will show up elsewhere in the code.

jon-betts:
```suggestion
    _jwt_token = marshmallow.fields.Str(required=True, data_key="authorization")
```

Is this a private attribute? I don't think we actually do / should access it from outside this class? The fact we are a schema is actually a private matter I think?

jon-betts (approved):
The auth stuff in LMS is a real stinker. This doesn't fix it, but it does make this chunk considerably easier to follow along with and is much smaller to boot. The process if fixing this in the end will probably be many changes like this which reduce the size until we can see what's going on better to fix it.

The patterns of the auth "schema" used here is really weird. The "schema" part of them is really a minor detail of what they do, they are really half of a security policy (which might, or might not employ schema within it). Hopefully one day we can get closer to something like we have in `h`.

Unfortunately being a schema and the way we then go about using it makes this a difficult knot to unpick.

In summary:

 * Anything to cut down on how confusing this is, is a bonus
 * Cutting the doc-strings down to relevant information, rather than abstract tutorials about related libraries or possible usage patterns is really important too
 * Moving to a mutable data structure for the user makes sense to me (we mutate them frequently)

## Possible improvements

 * `BearerTokenSchema._decode_jwt` doesn't look like it's serving much purpose and should probably be inlined into the parent where it is called from
 * Some of the old doc-string formatting is strange
 * Should we start making the fact this isn't really a schema more explicit by making the `authorization` param private?

## I spotted this while I was looking

Not super related to this PR, but I'll not spot it again, but `lms/resources/_js_config/__init__.py` has some duplication in `enable_oauth2_redirect_error_mode()`:

```python
    bearer_token = BearerTokenSchema(self._request).authorization_param(
        self._lti_user
    )
    auth_url = self._request.route_url(
        auth_route,
        _query=[("authorization", bearer_token)],
    )
```

The bearer token part there replicates a private function we already have and could be:

```python
    auth_url = self._request.route_url(
        auth_route,
        _query=[("authorization", self._auth_token())],
    )
```

It might be nice to call it `_bearer_token` as well?

lms/models/lti_user.py
marcospri: This started with me trying to have a "lti_roles: LTIRoles" property here which was complicated with namedtupled and the BearerToken mess.

With the dataclass we can add new fields no problem, but if we use `asdict` it's tricker to control what get serialized and how. A dumb method seems simpler.

marcospri:
This doesn't arrive to a very satisfactory state but it's in a better place to do another round.

PR #4754, opened 2022-12-13
Make the security policy/path relationship explicit

marcospri:
Instead of trying different locations in succession explicitly have a one to one relationship between security policy and the current path.

#### path.startswith("/googleauth") LMSGoogleSecurityPolicy 

Google oauth routes

Go to http://localhost:8001/admin and logout (or follow the auth flow if not already logged in)

#### path.startswith("/admin") LMSGoogleSecurityPolicy 

Auth in all the admin pages

Navigate around http://localhost:8001/admin

#### /content_item_selection LTIUserSecurityPolicy(get_lti_user_from_launch_params) 

Display the configure assignment screen in non-deeplinking LMS's. 

Reconfigure https://hypothesis.instructure.com/courses/125/assignments/3930

#### /lti_launches LTIUserSecurityPolicy(get_lti_user_from_launch_params) 

Regular launch in all LMS's

Launch https://hypothesis.instructure.com/courses/125/assignments/3930

#### /canvas_oauth_callback LTIUserSecurityPolicy(get_lti_user_from_oauth_callback)

Oauth flow in canvas

`make sql` and then `truncate oauth2_token ;`

then re-launch Launch https://hypothesis.instructure.com/courses/125/assignments/3930
 and complete the oauth flow.

#### /api/blackboard/oauth/callback LTIUserSecurityPolicy(get_lti_user_from_oauth_callback)

Oauth flow in blackboard

Launch 

and pick "Blackboard files", complete the oauth flow.

#### /api/d2l/oauth/callback  LTIUserSecurityPolicy(get_lti_user_from_oauth_callback)

Oauth flow in D2L

Launch https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View

pick D2L files and complete the oauth flow.

#### /api/* LTIUserSecurityPolicy ("headers")

Auth between the LMS backend and frontend for our internal API

Launch https://hypothesis.instructure.com/courses/125/assignments/873

The /api/sync request contains the auth header.

#### /lti/1.3/deep_linking/form_fields LTIUserSecurityPolicy ("headers")

Finishing configuration in deep linking LMS's (1.3)

Reconfigure https://hypothesis.instructure.com/courses/319/assignments/2988/edit?name=No+grade+LTI1.3&due_at=null&points_possible=10

#### /lti/1.1/deep_linking/form_fields LTIUserSecurityPolicy ("headers")

Finishing configuration in deep linking LMS's (1.1)

Reconfigure https://hypothesis.instructure.com/courses/125/assignments/3930/edit?name=Marcos+Assignment&due_at=null&points_possible=0

#### /assignment LTIUserSecurityPolicy ("form")

Configuration in non-deeplinking LMS's

Reconfigure https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_183_1&course_id=_19_1

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -42,15 +42,65 @@ class Permissions(Enum):
 @lru_cache(maxsize=1)
 def get_policy(request):
     """Pick the right policy based the request's path."""
+    # pylint:disable=too-many-return-statements
     path = request.path
 
     if path.startswith("/admin") or path.startswith("/googleauth"):
         return LMSGoogleSecurityPolicy()
 
     if path in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"}:
+        # Actual LTI backed authentication
         return LTIUserSecurityPolicy(get_lti_user_from_launch_params)
 
-    return LTIUserSecurityPolicy(get_lti_user)
+    if path in {
+        "/canvas_oauth_callback",
+        "/api/blackboard/oauth/callback",
+        "/api/d2l/oauth/callback",
+    }:
+        # LTIUser serialized in the state param for the oauth flow
+        return LTIUserSecurityPolicy(get_lti_user_from_oauth_callback)
+
+    if path.startswith("/api") and path.endswith("authorize"):
+        # LTUser serialized as query param for authorization failures
+        return LTIUserSecurityPolicy(
+            partial(get_lti_user_from_bearer_token, location="querystring")
+        )
+
+    if path.startswith("/api") or path in {
+        "/lti/1.3/deep_linking/form_fields",
+        "/lti/1.1/deep_linking/form_fields",
+    }:
+        # LTUser serialized in the headers for API calls from the frontend
+        return LTIUserSecurityPolicy(
+            partial(get_lti_user_from_bearer_token, location="headers")
+        )
+
+    if path in {"/assignment"}:
+        # LTUser serialized in a from for non deep-linked assignment configuration
+        return LTIUserSecurityPolicy(
+            partial(get_lti_user_from_bearer_token, location="form")
+        )
+
+    return UnAutheticatedSecurityPolicy()
+
+
+class UnAutheticatedSecurityPolicy:  # pragma: no cover
+    """Security policy that always returns an unauthenticated user."""
+
+    def authenticated_userid(self, _request):
+        return None
+
+    def identity(self, _request):
+        return None
+
+    def permits(self, _request, _context, _permission):
+        return Denied("")
+
+    def remember(self, _request, _userid, **_kw):
+        return []
+
+    def forget(self, _request):
+        return []
 
 
 class SecurityPolicy:
@@ -163,39 +213,27 @@ def get_lti_user_from_launch_params(request) -> LTIUser:
 
 
 @lru_cache(maxsize=1)
-def get_lti_user(request, from_identity=False) -> Optional[LTIUser]:
+def get_lti_user_from_bearer_token(request, location) -> LTIUser:
+    return BearerTokenSchema(request).lti_user(location=location)
+
+
+@lru_cache(maxsize=1)
+def get_lti_user_from_oauth_callback(request) -> LTIUser:
+    return OAuthCallbackSchema(request).lti_user()
+
+
+@lru_cache(maxsize=1)
+def get_lti_user(request) -> Optional[LTIUser]:
     """
     Return a models.LTIUser for the authenticated LTI user.
-
-    Get the authenticated user from one of our LTI bearer tokens.
 
     If the request doesn't contain a valid bearer token then return None.
 
     :param request: Current request.
-    :param from_identity: Use the lti_user from request.identity.
-        Defaults to False to avoid a circular dependency and allow
-        security policies to use this function.
-
-    :rtype: models.LTIUser
     """
     lti_user = None
-    if from_identity and request.identity and request.identity.lti_user:
+    if request.identity and request.identity.lti_user:
         lti_user = request.identity.lti_user
-    else:
-
-        bearer_token_schema = BearerTokenSchema(request)
-        schemas = [
-            partial(bearer_token_schema.lti_user, location="headers"),
-            partial(bearer_token_schema.lti_user, location="querystring"),
-            partial(bearer_token_schema.lti_user, location="form"),
-            OAuthCallbackSchema(request).lti_user,
-        ]
-        for schema in schemas:
-            try:
-                lti_user = schema()
-                break
-            except ValidationError:
-                continue
 
     if lti_user:
         # Make a record of the user for analytics so we can map from the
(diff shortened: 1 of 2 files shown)


lms/security.py
```diff
@@ -48,9 +48,51 @@ def get_policy(request):
...
+
+    if path.startswith("/api"):
+        # LTUser serialized in the headers for API calls from the frontend
+        return LTIUserSecurityPolicy(
+            partial(get_lti_user_from_bearer_token, location="headers")
+        )
+
+    if path in {"/assignment"}:
+        # LTUser serialized in a from for non deep-linked assignment configuration
+        return LTIUserSecurityPolicy(
```
marcospri, line 72: We add that to the form we send to ourselves (via frontend) on configuration here:

https://github.com/hypothesis/lms/blob/main/lms/views/lti/basic_launch.py#L103

lms/security.py
```diff
@@ -48,9 +48,51 @@ def get_policy(request):
         return LMSGoogleSecurityPolicy()
 
     if path in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"}:
+        # Actual LTI backed authentication
         return LTIUserSecurityPolicy(get_lti_user_from_launch_params)
 
-    return LTIUserSecurityPolicy(get_lti_user)
+    if path.startswith("/api") and path.endswith("callback"):
+        # LTIUser serialized in the state param for the oauth flow
+        return LTIUserSecurityPolicy(get_lti_user_from_oauth_callback)
```
marcospri, line 56: Sent as the state param here:

https://github.com/hypothesis/lms/blob/main/lms/views/api/d2l/authorize.py#L20 and similar routes.

lms/security.py
```diff
@@ -42,15 +42,58 @@ class Permissions(Enum):
 @lru_cache(maxsize=1)
 def get_policy(request):
     """Pick the right policy based the request's path."""
+    # pylint:disable=too-many-return-statements
```
marcospri, line 45: This method is not very sophisticated. The aim here is to get the explicitlness in place first, we can refine later.

We could move this direclty to the `add_route` side of things with maybe a route predicate https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_route_predicate.

seanh: :+1: I think it's fine to disable the Pylint warning here.

If you wanted to get rid of the warning, I think what Pylint is trying to tell you to do is to separate configuration from logic. What you have here is some logic that takes as its input a list of `(boolean_function, security_policy_factory)` 2-tuples. It iterates over the list calling each boolean function and if the boolean function returns `True` then it uses that security policy factory.

You probably wouldn't want to do it quite like this (with a bunch of tuples and lambdas) but here's an example implementation that has no Pylint warnings and still passes all the tests.

I think a better version of this would probably recognise that all of the boolean functions are pattern-matching against `path` (they're all `path in <list_of_paths>` or `path.startswith(prefix)` or `path.endswith(suffix)`). This can be implemented by associating a regular expression or list of regular expressions with each security policy factory, rather than a boolean function.

```python
@lru_cache(maxsize=1)
def get_policy(request):
    """Pick the right policy based the request's path."""
    path = request.path

    policies = [
        (
            lambda: path.startswith("/admin") or path.startswith("/googleauth"),
            LMSGoogleSecurityPolicy,
        ),
        (
            lambda: path
            in {"/lti_launches", "/content_item_selection", "/api/gateway/h/lti"},
            lambda: LTIUserSecurityPolicy(get_lti_user_from_launch_params),
        ),
        (
            lambda: path
            in {
                "/canvas_oauth_callback",
                "/api/blackboard/oauth/callback",
                "/api/d2l/oauth/callback",
            },
            lambda: LTIUserSecurityPolicy(get_lti_user_from_oauth_callback),
        ),
        (
            lambda: path.startswith("/api") and path.endswith("authorize"),
            lambda: LTIUserSecurityPolicy(
                partial(get_lti_user_from_bearer_token, location="querystring")
            ),
        ),
        (
            lambda: path.startswith("/api")
            or path
            in {
                "/lti/1.3/deep_linking/form_fields",
                "/lti/1.1/deep_linking/form_fields",
            },
            lambda: LTIUserSecurityPolicy(
                partial(get_lti_user_from_bearer_token, location="headers")
            ),
        ),
        (
            lambda: path in {"/assignment"},
            lambda: LTIUserSecurityPolicy(
                partial(get_lti_user_from_bearer_token, location="form")
            ),
        ),
    ]

    for boolean_func, securitypolicy_factory in policies:
        if boolean_func():
            return securitypolicy_factory()

    return UnAutheticatedSecurityPolicy()
```

seanh: Might wanna just take `request.path` as the argument in the first place (`get_policy(path)`) since `request.path` is the only thing that `get_policy()` uses `request` for

lms/security.py
```diff
@@ -48,9 +48,51 @@ def get_policy(request):
...
+    if path in {"/assignment"}:
+        # LTUser serialized in a from for non deep-linked assignment configuration
+        return LTIUserSecurityPolicy(
+            partial(get_lti_user_from_bearer_token, location="form")
+        )
+
+    return UnAutheticatedSecurityPolicy()
+
+
+class UnAutheticatedSecurityPolicy:  # pragma: no cover
```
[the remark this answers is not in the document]
seanh:
```suggestion
class UnautheticatedSecurityPolicy:  # pragma: no cover
```

"Unauthenticated" is one word

lms/security.py
seanh: `get_policy()` could probably be a `@staticmethod` within `SecurityPolicy` since that's the only class that uses it

lms/security.py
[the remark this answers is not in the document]
seanh: > As an alternative we could have to have a few `if not policy` on `LTIUserSecurityPolicy`

Yeah :man_shrugging: I think either way is fine

seanh (approved):
<q>Make the security policy/path relationship explicit</q> &larr; good PR title

marcospri:
I went thought every policy/path pair and did at least a sanity check in different LMS.

I reckon this is good to go.

lms/security.py
marcospri: Kept the method silly for now. I'd like to do another pass that just focuses in making it nicer but once that the policy <-> path is already merged.

lms/security.py
```diff
@@ -39,37 +39,85 @@ class Permissions(Enum):
     ADMIN = "admin"
 
 
-@lru_cache(maxsize=1)
-def get_policy(request):
-    """Pick the right policy based the request's path."""
-    path = request.path
+class UnautheticatedSecurityPolicy:  # pragma: no cover
```
seanh, line 42: Why not just add unit tests? They should be pretty simple

seanh (approved):

PR #4760, opened 2022-12-15
D2L files: Authorization and API endpoints

marcospri:
This PRs add some of the bolilerplate needed for authorization, basic fetching of the list of file from D2L and the bits needed on the frontend to be able to test it all together.

This PR doesn't however support:

- https://github.com/hypothesis/lms/issues/4769
- https://github.com/hypothesis/lms/issues/4768
- https://github.com/hypothesis/lms/issues/4771

It also doesn't try to find a better abstraction for files/file pickers. We can tackle that  once this lands and we have canvas, blackboard and D2L examples to build that abstraction.

## Testing

- Lunch https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2184/View

There should not be a D2L files button.

- Enable the files setting on http://localhost:8001/admin/instance/id/106/

Now you'll get a D2L files button.

- Click the "D2L files button. After you authorize you'll should see files in the dialog.

- You can try updating new files at the top level menu:

Screenshot from 2022-12-19 15-24-20

- Configuring and launching that file will fail.

### DIFF
--- a/lms/product/product.py
+++ b/lms/product/product.py
@@ -47,11 +47,15 @@ class Settings:
     groups_enabled: bool = False
     """Is the course groups feature enabled"""
 
+    files_enabled: bool = False
+    """Is this product files feature enabled"""
+
     custom: Optional[dict] = None
     """Other non-standard settings."""
 
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
+        self.files_enabled = product_settings.get("files_enabled", False)
         self.custom = product_settings
 
 
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -1,4 +1,6 @@
-from marshmallow import EXCLUDE, fields
+from typing import List
+
+from marshmallow import EXCLUDE, Schema, fields
 
 from lms.validation._base import RequestsResponseSchema
 
@@ -23,6 +25,32 @@ class D2LGroupsSchema(RequestsResponseSchema):
     name = fields.Str(required=True, data_key="Name")
     group_set_id = fields.Int(required=False)
     enrollments = fields.List(fields.Int(), required=True, data_key="Enrollments")
+
+
+class _D2LTopic(Schema):
+    class Meta:
+        unknown = EXCLUDE
+
+    id = fields.Str(required=True, data_key="Identifier")
+    name = fields.Str(required=True, data_key="Title")
+    updated_at = fields.Str(required=True, data_key="LastModifiedDate")
+    type = fields.Str(required=True, data_key="TypeIdentifier")
+
+
+class _D2LModuleSchema(Schema):
+    class Meta:
+        unknown = EXCLUDE
+
+    modules = fields.List(
+        fields.Nested(lambda: _D2LModuleSchema), required=False, data_key="Modules"
+    )
+    topics = fields.List(fields.Nested(_D2LTopic), required=False, data_key="Topics")
+
+
+class D2LTableOfContentsSchema(RequestsResponseSchema):
+    modules = fields.List(
+        fields.Nested(_D2LModuleSchema), required=True, data_key="Modules"
+    )
 
 
 class D2LAPIClient:
@@ -77,6 +105,17 @@ class D2LAPIClient:
             groups = [group for group in groups if int(user_id) in group["enrollments"]]
 
         return groups
+
+    def course_table_of_contents(self, org_unit) -> List[dict]:
+        """
+        Get a list of modules in the given course.
+
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-toc
+        """
+        response = self._api.request(
+            "GET", self._api.api_url(f"/{org_unit}/content/toc", product="le")
+        )
+        return D2LTableOfContentsSchema(response).parse().get("modules", [])
 
     @staticmethod
     def get_api_user_id(user_id: str):
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.js
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.js
@@ -18,7 +18,7 @@ import URLPicker from './URLPicker';
  *
  * @typedef {import('./FilePickerApp').ErrorInfo} ErrorInfo
  *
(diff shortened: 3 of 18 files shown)


lms/static/scripts/frontend_apps/components/ContentSelector.js
```diff
@@ -172,6 +180,19 @@ export default function ContentSelector({
...
       );
       break;
+    case 'd2lFile':
+      dialog = (
+        <LMSFilePicker
+          authToken={authToken}
+          listFilesApi={d2lListFilesApi}
+          onCancel={cancelDialog}
+          onSelectFile={selectD2LFile}
+          missingFilesHelpLink={'https://web.hypothes.is/help-categories/d2l/'}
```
marcospri, line 190: Actual link pending: https://github.com/hypothesis/lms/issues/4770

lms/services/d2l_api/client.py
```diff
@@ -78,6 +106,17 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
 
         return groups
 
+    def course_table_of_contents(self, org_unit) -> List[dict]:
```
marcospri, line 109: Keeping this very close to the actual D2L API as we plan to move things around to support folders.

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -3,7 +3,73 @@
...
+                                    "name": "Title 1",
+                                    "updated_at": "Date 1",
+                                    "type": "File",
+                                }
+                            ],
+                        }
+                    ],
+                }
+            ]
+        }
```
jon-betts, line 43: :shrug: This is kind of painful, but with something this nested, and with every key different I'm not sure what you can do?

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -3,7 +3,73 @@
...
+                        }
+                    ],
+                    "Modules": [
+                        {
+                            "Topics": [
+                                {
+                                    "Identifier": "ID 1",
+                                    "Title": "Title 1",
+                                    "LastModifiedDate": "Date 1",
+                                    "TypeIdentifier": "File",
```
jon-betts, line 65:
```suggestion
                                    "Identifier": "ID 2",
                                    "Title": "Title 2",
                                    "LastModifiedDate": "Date 2",
                                    "TypeIdentifier": "File",
```

Might as well take advantage to test we don't scramble the results?

lms/views/api/d2l/files.py
```diff
@@ -0,0 +1,40 @@
+from pyramid.view import view_config
+
+from lms.security import Permissions
+from lms.services.d2l_api import D2LAPIClient
+
+
+def _find_files(modules):
```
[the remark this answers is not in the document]
jon-betts: Yeah, I think this would address my comments about the `course_table_of_contents` method. I think this would be good to push down into the service.

lms/views/api/d2l/files.py
```diff
@@ -0,0 +1,40 @@
...
+    for file in _find_files(modules):
+        response_results.append(
+            {
+                "id": f"d2l://content-resource/{file['id']}/",
+                "display_name": file["name"],
+                "updated_at": file["updated_at"],
+                "type": "File",
+            }
+        )
+    return response_results
```
jon-betts, line 40:
```suggestion
    return [
        {
            "id": f"d2l://content-resource/{file['id']}/",
            "display_name": file["name"],
            "updated_at": file["updated_at"],
            "type": "File",
        }
        for file in _find_files(modules)
    ]
```

?

jon-betts: If this is the mapping we want, we should push this down into the service, rather than spit out a mapping we don't want.

This would probably cut down on the testing mass a lot, as we could simply test that this method returns the value it gets passed from the service without testing the additional mapping.

tests/unit/lms/resources/_js_config/file_picker_config_test.py
```diff
@@ -30,6 +30,26 @@ def test_blackboard_config(
...
+            pyramid_request, sentinel.application_instance
+        )
+
+        expected_config = {"enabled": files_enabled}
+
+        if files_enabled:
+            expected_config["listFiles"] = {
+                "authUrl": "http://example.com/api/d2l/oauth/authorize",
+                "path": "/api/d2l/courses/COURSE_ID/files",
+            }
```
jon-betts, line 49: `expected_config` might be a good fixture.

jon-betts (approved):
I've followed through with the testing notes and everything works as described.

I think this PR is in a bit of flux as there are more changes planned to follow up, but I think we would benefit from pushing the output we want in the view right down into the service.

I think the service can be a translation layer between the D2L service and the format we want, rather than faithfully return the shap the API returns. We've seen this in a few integrations before and I think it results in less testing when we map directly. Currently we have:

* Raw-ish output from D2l api (service output)
* Remapping and filtering into a flat structure (private method in view)
* Remapping into a new dict (inside view)

In order to test this we have to inspect the data at each step. If we output the format we want right away, the following steps either don't exist at all, or can do a much simpler `fixture.return_value` type of test.

I think we might also be able to leverage some stuff in Marshmallow to rename the dicts into the fields we want in the first step?

robertknight:
From a quick skim over the frontend changes it looks like this is a straightforward extension of the existing code. Per Marcos's comments on the PR description, we can refactor this to reduce duplication once this lands.

lms/product/product.py
```diff
@@ -47,11 +47,15 @@ class Settings:
...
 
+    files_enabled: bool = False
+    """Is this product files feature enabled"""
+
     custom: Optional[dict] = None
     """Other non-standard settings."""
 
     def __post_init__(self, product_settings):
         self.groups_enabled = product_settings.get("groups_enabled", False)
+        self.files_enabled = product_settings.get("files_enabled", False)
```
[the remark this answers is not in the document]
marcospri: This is Product.settings.files_enabled so I reckon it makes sense it always about the product's own files feature.

Canvas/Blackboard/D2L files are very similar and I reckon a refactor there would cut down a lot of repetition.

One Drive/JSTOR... they are files but they have more in common between them than to the product ones.

marcospri:
>I think the service can be a translation layer between the D2L service and the format we want, rather than faithfully return the shap the API returns

Agree with this but I reckon, as we plan to return files and folders before this going into production I rather tackle that in a separate PR

PR #4761, opened 2022-12-15
Notes field on the admin pages

marcospri:
Removed by mistake in instances, added also to organizations.

## Testing:

Play with the notes in orgs and AIs:

- http://localhost:8001/admin/org/1
- http://localhost:8001/admin/instance/id/4/

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -135,13 +135,16 @@ class OrganizationService:
 
         return org
 
-    def update_organization(self, organization, name=None, enabled=None):
+    def update_organization(self, organization, name=None, enabled=None, notes=None):
         """Update an existing organization."""
         if name:
             organization.name = name
 
         if enabled is not None:
             organization.enabled = enabled
+
+        if notes is not None:
+            organization.settings.set("hypothesis", "notes", notes)
 
         return organization
 
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -46,6 +46,13 @@
 <form method="POST" action="{{ request.route_url("admin.instance.id", id_=instance.id) }}">
     <input type="hidden" name="csrf_token" value="{{ get_csrf_token() }}">
 
+    <fieldset class="box">
+        {{ macros.settings_textarea(instance, "Notes", "hypothesis", "notes") }}
+        {{ macros.disabled_text_field("Consumer key", instance.consumer_key) }}
+        {{ macros.form_text_field(request, "Deployment ID", "deployment_id", field_value=instance.deployment_id) }}
+        {{ macros.form_text_field(request, "LMS URL", "lms_url", field_value=instance.lms_url) }}
+    </fieldset>
+
     <fieldset class="box mt-6">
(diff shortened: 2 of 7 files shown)


lms/services/organization.py
```diff
@@ -135,14 +135,17 @@ def create_organization(self, name=None) -> Organization:
 
         return org
 
-    def update_organization(self, organization, name=None, enabled=None):
+    def update_organization(self, organization, name=None, enabled=None, notes=None):
```
marcospri, line 138: `notes` could become a `settings` parameter here or similar, we only have `notes` right now so that can wait.

lms/templates/admin/instance.html.jinja2
```diff
@@ -46,6 +46,13 @@
 <form method="POST" action="{{ request.route_url("admin.instance.id", id_=instance.id) }}">
     <input type="hidden" name="csrf_token" value="{{ get_csrf_token() }}">
 
+    <fieldset class="box">
+        {{ macros.settings_textarea(instance, "Notes", "hypothesis", "notes") }}
```
marcospri, line 50: This was embarrassingly removed here https://github.com/hypothesis/lms/pull/4726/files#diff-fd5bcca6d2ee9b73ec82594547d82c2d0682f22cc2fedbb9f8ee5aac9e355015

lms/views/admin/organization.py
```diff
@@ -88,8 +88,11 @@ def update_organization(self):
         org = self._get_org_or_404(self.request.matchdict["id_"])
 
         self.organization_service.update_organization(
-            org, name=self.request.params.get("name", "").strip()
+            org,
+            name=self.request.params.get("name", "").strip(),
+            notes=self.request.params.get("hypothesis.notes", "").strip(),
         )
+        self.request.session.flash("Updated organization", "messages")
```
marcospri, line 95: This was missing and there was no indication that updates were successful.

lms/templates/admin/macros.html.jinja2
```diff
@@ -44,6 +44,14 @@
 {% endmacro %}
 
 
+{% macro settings_textarea(object, label, setting, sub_setting, field_name, default='') %}
+    {% call field_body(label) %}
+        {% set text = object.settings.get(setting, sub_setting, default) or ''  %}
+        <textarea class="textarea" name="{{ setting }}.{{ sub_setting }}">{{ text }}</textarea>
```
jon-betts, line 50: It might be nicer to call this "settings.{{ setting }}.{{ sub_setting }}". That way you could deal with all possible at once maybe?

jon-betts (approved):
Yep this works although it's not great that we missed these going away. I guess when you are focusing on adding one thing it's more likely to miss removing another?

There are two suggestions I'd make:

 * Probably put the org name above the notes field?
 * If we namespace the settings we might be able to deal with them all at once some day

lms/templates/admin/macros.html.jinja2
marcospri: Makes sense but it will mean changing the settings over the application instance template, probably for a follow up PR.

PR #4762, opened 2022-12-15
Fix secret field value in application instance settings

marcospri:
Setting the value would override the setting in the DB.

Using placeholder as a hint of the presence of the value.


jon-betts (approved):
Whoops!

PR #4788, opened 2022-12-19
Store and update email and name in "User"

marcospri:
For:

 * https://github.com/hypothesis/lms/issues/4730

### Testing

- Remove any that from other testing:

```tox -qe dockercompose -- exec postgres psql -U postgres -c "update \"user\" set display_name=null, email=null;"```

- Launch https://hypothesis.instructure.com/courses/125/assignments/873 both as an student and instructor.

- Check what's store for each user

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select roles,email,display_name from \"user\""

   roles    |             email             |      display_name      
------------+-------------------------------+------------------------
 Instructor | eng+canvasteacher@hypothes.is | Hypothesis 101 Teacher
 Learner    |                               | 
(2 rows)
```

### DIFF
--- a/lms/services/user.py
+++ b/lms/services/user.py
@@ -23,16 +23,27 @@ class UserService:
 
     def upsert_user(self, lti_user: LTIUser) -> User:
         """Store a record of having seen a particular user."""
-        new_user = self._from_lti_user(lti_user)
+        user = User(
+            application_instance_id=lti_user.application_instance_id,
+            user_id=lti_user.user_id,
+            roles=lti_user.roles,
+            h_userid=lti_user.h_user.userid(self._h_authority),
+        )
 
-        if existing_user := self._find_existing_user(model_user=new_user):
+        if existing_user := self._find_existing_user(model_user=user):
             # Update the existing user from the fields which can change on a
             # new one
-            existing_user.roles = new_user.roles
-            return existing_user
+            existing_user.roles = user.roles
+            user = existing_user
+        else:
+            self._db.add(user)
 
-        self._db.add(new_user)
-        return new_user
+        if lti_user.is_instructor:
+            # We are only storing these personal details for teachers now.
+            user.email = lti_user.email
+            user.display_name = lti_user.display_name
+
+        return user
 
     @lru_cache(maxsize=128)
     def get(self, application_instance, user_id: str) -> User:
@@ -65,14 +76,6 @@ class UserService:
                 user_id=model_user.user_id,
             )
             .one_or_none()
-        )
-
-    def _from_lti_user(self, lti_user: LTIUser) -> User:
-        return User(
-            application_instance_id=lti_user.application_instance_id,
-            user_id=lti_user.user_id,
-            roles=lti_user.roles,
-            h_userid=lti_user.h_user.userid(self._h_authority),
         )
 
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lms/services/user.py [hunk 1]
```diff
@@ -29,6 +29,7 @@ def upsert_user(self, lti_user: LTIUser) -> User:
             # Update the existing user from the fields which can change on a
             # new one
             existing_user.roles = new_user.roles
+            self._update_user(existing_user, lti_user)
             return existing_user
```
jon-betts, line 33: Something seems a bit iffy here? I think we end up calling `_update_user` twice in the case of an existing user, which doesn't seem necessary.

This is quite hard to see as we use many out of method helpers. If you inline it all, then I think the structure is something like this:

```python
    def upsert_user(self, lti_user: LTIUser) -> User:
        """Store a record of having seen a particular user."""
        new_user =  User(
            application_instance_id=lti_user.application_instance_id,
            user_id=lti_user.user_id,
            roles=lti_user.roles,
            h_userid=lti_user.h_user.userid(self._h_authority),
        )
        self._update_user(new_user, lti_user)

        if existing_user := self._find_existing_user(model_user=new_user):
            # Update the existing user from the fields which can change on a
            # new one
            existing_user.roles = new_user.roles
            self._update_user(existing_user, lti_user)
            return existing_user

        self._db.add(new_user)
        return new_user
```

Whereas I think we could do something like:

```python
    def upsert_user(self, lti_user: LTIUser) -> User:
        """Store a record of having seen a particular user."""
        user = User(
            application_instance_id=lti_user.application_instance_id,
            user_id=lti_user.user_id,
            roles=lti_user.roles,
            h_userid=lti_user.h_user.userid(self._h_authority),
        )

        if existing_user := self._find_existing_user(model_user=user):
            # Update the existing user from the fields which can change on a
            # new one
            existing_user.roles = user.roles
            user = existing_user
        else:
            self._db.add(user)

        self._update_user(user, lti_user)

        return user
```

As the functions are only called once in-lining might be the way to go?

tests/unit/lms/services/user_test.py
```diff
@@ -38,6 +40,18 @@ def test_upsert_user_with_an_existing_user(
         assert saved_user.roles == lti_user.roles
         assert user == saved_user
```
jon-betts, line 41: I think there is some missing coverage here. If you comment out the `self._update_user` inside the existing user block the tests still pass.

jon-betts (changes requested):
I think there are two issues here, where resolving one would resolve the other, but you could resolve both.

The bigger (although not big) of the two is that there's missing coverage. We have two paths which call the new updating code: one for new users and another for existing ones. In the current code if you comment out the existing users update the tests still pass.

The second issue is I think as a result of the use of small helper functions we've kind of obscured the fact this is true. I think we might be better of inlining, and then refactoring the various helper functions we have. They are all used by `upsert` and I think they might be doing more to harm readability than help it.

The second is more a matter of taste, but if you fixed the second the first might not be as important, as there is only one path, and it would be covered by tests.

lms/services/user.py
marcospri: Yeah that's nicer without the helpers :+1:

jon-betts (approved):

PR #4789, opened 2022-12-19
Backfill user emails and names from group_info

marcospri:
For:

 * https://github.com/hypothesis/lms/issues/4730

### Testing

- Launch https://hypothesis.instructure.com/courses/125/assignments/873 as different users to fill up `group_info`

- Clear the "user" table  ```update "user" set display_name=null, email=null;```

- Run the migration

```hdev alembic upgrade head```

- Check the data copied over to "user":

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select roles,email,display_name from \"user\""

   roles    |             email             |      display_name      
------------+-------------------------------+------------------------
 Learner    |                               | 
 Instructor | eng+canvasteacher@hypothes.is | Hypothesis 101 Teacher
(2 rows)
```

### DIFF
--- a/lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py
+++ b/lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py
@@ -1,0 +1,86 @@
+"""
+Backfill email and name from group_info.
+
+Revision ID: 0b7391f5b1e3
+Revises: e32d2cf33591
+Create Date: 2022-12-19 10:42:00.577163
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0b7391f5b1e3"
+down_revision = "e32d2cf33591"
+
+
+def upgrade():
+    conn = op.get_bind()
+
+    # Flatten group_info.info.instructors
+    conn.execute(
+        """CREATE TEMPORARY VIEW group_info_instructors AS (
+            SELECT
+                group_info.application_instance_id,
+                instructors.email,
+                instructors.display_name,
+                instructors.provider_unique_id,
+                -- Rank emails by latest group_info.id
+                row_number() over (partition by (instructors.provider_unique_id,  group_info.application_instance_id) order by group_info.id desc) as rank
+            FROM
+                group_info,
+                jsonb_to_recordset(group_info.info->'instructors') as instructors(email text, username text, display_name text,provider_unique_id text )
+        );"""
+    )
+
+    # Migrate emails
+    result_emails = conn.execute(
+        """
+        UPDATE "user"
+            SET email=group_info_emails.email
+            FROM (
+                SELECT "user".id, group_info_instructors.email
+                FROM group_info_instructors
+                LEFT OUTER JOIN "user" ON group_info_instructors.application_instance_id = "user".application_instance_id AND provider_unique_id = user_id
+                WHERE
+                    -- Only take the latest info from group_info for duplicates
+                    rank = 1
+                    -- Only take instructors that have an email in group_info
+                    AND group_info_instructors.email IS NOT NULL AND group_info_instructors.email <> ''
+                    --  And that we can match to a user in "user"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+                    --  And that we can match to a user in "user"
+                    AND "user".id IS NOT NULL
+                    -- Don't override any names we already had
+                    AND "user".display_name IS NULL
+            ) as group_info_names
+        WHERE "user".id = group_info_names.id
+    """
+    )
+    print("\tUpdated user rows with email from group_info:", result_emails.rowcount)
+    print("\tUpdated user rows with names from group_info:", result_names.rowcount)
```
jon-betts, line 82: I don't think this matters, but I think you can do the same thing all in one go with something along the lines of:

```sql
WITH 
    user_details AS (
        SELECT
            "user".id,
            -- Pick the last non-empty value by group info id
            (
                ARRAY_AGG(instructors.email ORDER BY group_info.id DESC) 
                FILTER (WHERE instructors.email IS NOT NULL AND instructors.email <> '')
            )[1] AS email,
            (
                ARRAY_AGG(instructors.display_name ORDER BY group_info.id DESC) 
                FILTER (WHERE instructors.display_name IS NOT NULL AND instructors.display_name <> '')
            )[1] AS display_name
        FROM
            group_info
        CROSS JOIN LATERAL
            jsonb_to_recordset(group_info.info->'instructors') AS instructors(email text, display_name text, provider_unique_id text)
        JOIN "user" ON
            "user".application_instance_id = group_info.application_instance_id
            AND "user".user_id = instructors.provider_unique_id
        GROUP BY "user".id
    )

UPDATE "user"
SET
    -- Only overwrite nulls
    email=COALESCE(users.email, details.email),
    display_name=COALESCE(users.display_name, details.display_name)
FROM user_details
WHERE "user".id = details.id
```

Which is still frankly quite horrible SQL, but a bit easier to understand as a whole I think?

lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def downgrade():
+    pass
```
jon-betts, line 86: I take it we don't want to destroy info?

lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py
```diff
@@ -0,0 +1,86 @@
...
+            SET display_name=group_info_names.display_name
+            FROM (
+                SELECT "user".id, group_info_instructors.display_name
+                FROM group_info_instructors
+                LEFT OUTER JOIN "user" ON group_info_instructors.application_instance_id = "user".application_instance_id AND provider_unique_id = user_id
+                WHERE
+                    -- Only take the latest info from group_info for duplicates
+                    rank = 1
+                    -- Only take instructors that have a name in group_info
+                    AND group_info_instructors.display_name IS NOT NULL AND group_info_instructors.display_name <> ''
```
jon-betts, line 72: Hangs head in shame. I had to look up `<>`.

jon-betts (approved):
I've checked this out and it works (I think), but the SQL here is quite complicated. I've suggested a shorter alternative which I think also works?

I assume not removing the data in the downgrade is intentional so we don't destroy data that we might have collected by now, but it seems a bit strange as if we roll back one more we remove the columns? Doesn't seem important anyway.

lms/migrations/versions/0b7391f5b1e3_backfill_user_email.py
marcospri: Yeah, for adding new columns something could go wrong (I guess) and we might want to roll it back. If we got this one wrong I reckon we'd write a new migration to fix it.

PR #4792, opened 2022-12-21
Suppport for D2L files proxing using Via

marcospri:
#### Needs

- https://github.com/hypothesis/h-vialib/pull/42 
- https://github.com/hypothesis/via/pull/845

### Testing

- Checkout https://github.com/hypothesis/h-vialib/pull/42  h_vialib
- Checkout https://github.com/hypothesis/via/pull/845 in Via
- Install the local version of h-vialib in both Via and LMS

I've been using 

```.tox/dev/bin/pip install -e /home/marcos/hypo/h-vialib``` for this.

- Enable D2L files in http://localhost:8001/admin/instance/id/106/
- Launch a D2L assignment as a teacher and configure it as D2L assignment:

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2184/View

- You should be able to see the contents of the file a follow the trip of URL
  - via_url request to LMS to get the file which contains the `via.secret.headers`
  - Via request to get the pdf, again with the header

### DIFF
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -7,6 +7,7 @@ from pyramid.httpexceptions import HTTPClientError
 from lms.models import Assignment, GroupInfo, Grouping, HUser
 from lms.product.blackboard import Blackboard
 from lms.product.canvas import Canvas
+from lms.product.d2l import D2L
 from lms.resources._js_config.file_picker_config import FilePickerConfig
 from lms.services import HAPIError, JSTORService, VitalSourceService
 from lms.validation.authentication import BearerTokenSchema
@@ -68,6 +69,16 @@ class JSConfig:
                     resource_link_id=self._request.lti_params["resource_link_id"],
                 ),
             }
+        elif document_url.startswith("d2l://"):
+            self._config["api"]["viaUrl"] = {
+                "authUrl": self._request.route_url(D2L.route.oauth2_authorize),
+                "path": self._request.route_path(
+                    "d2l_api.courses.files.via_url",
+                    course_id=self._request.lti_params["context_id"],
+                    _query={"document_url": document_url},
+                ),
+            }
+
         elif document_url.startswith("vitalsource://"):
             svc: VitalSourceService = self._request.find_service(VitalSourceService)
 
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -53,6 +53,9 @@ def includeme(config):  # pylint:disable=too-many-statements
         "d2l_api.courses.group_sets.list", "/api/d2l/courses/{course_id}/group_sets"
     )
     config.add_route("d2l_api.courses.files.list", "/api/d2l/courses/{course_id}/files")
+    config.add_route(
+        "d2l_api.courses.files.via_url", "/api/d2l/courses/{course_id}/via_url"
+    )
 
     config.add_route(
         "blackboard_api.oauth.authorize",
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -117,7 +117,33 @@ class D2LAPIClient:
         """Get a nested list of files and folders for the given `org_unit`."""
 
         modules = self._get_course_modules(org_unit)
-        return list(self._find_files(modules))
+        return list(self._find_files(org_unit, modules))
+
+    def public_url(self, org_unit, file_id) -> str:
+        """
+        Return the URL to download the given file.
+
+        As opposed to other LMS's that return a one time signed URL that we can then pass along to Via
+        D2L requires us to send the API authentication header to get the file contents.
+
+        To make sure the API authentication header is not expired we'll make a API
+        request here so if it needs to be refreshed we follow the standard token refresh procedure.
+
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-topics-(topicId)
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: lms/services/d2l_api/client.py [hunk 1]
```diff
@@ -117,6 +117,32 @@ def course_table_of_contents(self, org_unit) -> List[dict]:
...
+        request here so if it needs to be refreshed we follow the standard token refresh procedure.
+
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-topics-(topicId)
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-topics-(topicId)-file
+        """
+
+        # We don't need the data from this call.
+        # We are only interested on the potential side effect of needing
+        # a new access token and/or refreshing an existing one.
+        _ = self._api.request(
```
marcospri, line 137: This is a bit of a hack but the whole setup is (sending the header etc).

lms/views/api/d2l/files.py
```diff
@@ -23,15 +30,36 @@ def _find_files(modules):
...
+    permission=Permissions.API,
+)
+def via_url(_context, request):
+    course_id = request.matchdict["course_id"]
+    document_url = request.params["document_url"]
+    file_id = DOCUMENT_URL_REGEX.search(document_url)["file_id"]
+
+    public_url = request.find_service(D2LAPIClient).public_url(course_id, file_id)
+
+    access_token = request.find_service(name="oauth2_token").get().access_token
```
marcospri, line 61: I thought I might need to get this from a private method in the client but we can just use the oauth2 service :+1:

lms/resources/_js_config/__init__.py
```diff
@@ -68,6 +69,16 @@ def add_document_url(self, document_url):
...
             }
+        elif document_url.startswith("d2l://"):
+            self._config["api"]["viaUrl"] = {
+                "authUrl": self._request.route_url(D2L.route.oauth2_authorize),
+                "path": self._request.route_path(
+                    "d2l_api.courses.files.via_url",
+                    course_id=self._request.lti_params["context_id"],
+                    _query={"document_url": document_url},
+                ),
+            }
```
seanh, line 80: Ah ok, this is the config telling the frontend to make the API call to get the Via URL? :+1:

lms/routes.py
```diff
@@ -53,6 +53,9 @@ def includeme(config):  # pylint:disable=too-many-statements
         "d2l_api.courses.group_sets.list", "/api/d2l/courses/{course_id}/group_sets"
     )
     config.add_route("d2l_api.courses.files.list", "/api/d2l/courses/{course_id}/files")
+    config.add_route(
+        "d2l_api.courses.files.via_url", "/api/d2l/courses/{course_id}/via_url"
+    )
```
seanh, line 58: And the new Via URL API for the frontend to call go get D2L files Via URLs :+1:

lms/services/d2l_api/client.py
```diff
@@ -117,6 +117,32 @@ def course_table_of_contents(self, org_unit) -> List[dict]:
         )
         return D2LTableOfContentsSchema(response).parse().get("modules", [])
 
+    def public_url(self, org_unit, file_id) -> str:
```
seanh, line 120: Service method for getting a public URL for a D2L file. So far this is all very familiar :smile:

lms/services/d2l_api/client.py
```diff
@@ -117,6 +117,32 @@ def course_table_of_contents(self, org_unit) -> List[dict]:
...
+        """
+        Return the URL to download the given file.
+
+        As opposed to other LMS that return a one time signed URL that we can then pass along to via
+        D2L requires us to send the API authentication header to get the file contents.
```
seanh, line 125:
```suggestion
        As opposed to other LMS's that return a one time signed URL that we can then pass along to Via
        D2L requires us to send the API authentication header to get the file contents.
```

lms/services/d2l_api/client.py
```diff
@@ -117,6 +117,32 @@ def course_table_of_contents(self, org_unit) -> List[dict]:
...
+
+        To make sure the API authentication header is not expired we'll make a API
+        request here so if it needs to be refreshed we follow the standard token refresh procedure.
+
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-topics-(topicId)
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-topics-(topicId)-file
```
seanh, line 131: I think this is probably better in code comments next to the code that makes these requests. Otherwise it seems likely that if the code changes this docstring will get out of date.

lms/services/d2l_api/client.py
seanh: Is this because we're going to be sending our access token to Via (in the new secret headers), and we therefore need to make sure that the access token is still valid first?

lms/views/api/d2l/authorize.py
```diff
@@ -8,7 +8,7 @@
 from lms.validation.authentication import OAuthCallbackSchema
 
 GROUPS_SCOPES = ("groups:*:*",)
-FILES_SCOPES = ("content:toc:read",)
+FILES_SCOPES = ("content:toc:read", "content:topics:read")
```
seanh, line 11: I wonder what will happen to people who've already granted us access tokens without the topics scope? Will they get prompted to re-authorize us? Maybe no real users are using this feature yet anyway?

seanh (approved):
I haven't tested it myself, but this all looks totally sound to me.

Issue #4769, opened 2022-12-16
Support for folder structure in D2L files


marcospri:
The endpoint we are using to fetch the files in a course in D2L returns all the tree containing modules/sub modules and files.

Expose this structure to the file picker.

PR #4794, opened 2022-12-22
Expose D2L folder structure for the file picker

marcospri:
For:

- #4769 

## Testing

* See: https://github.com/hypothesis/lms/pull/4795
* Check out the branch from that ticket: `d2l-file-folders-frontend` which has a prototype implementation of the front end

### DIFF
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -28,6 +28,8 @@ class D2LGroupsSchema(RequestsResponseSchema):
 
 
 class _D2LTopic(Schema):
+    """Files and assigmetns are topics within a module."""
+
     class Meta:
         unknown = EXCLUDE
 
@@ -38,6 +40,8 @@ class _D2LTopic(Schema):
 
 
 class _D2LModuleSchema(Schema):
+    """D2L course sections are called "modules". They can contain nested sub-modules."""
+
     class Meta:
         unknown = EXCLUDE
 
@@ -109,16 +113,11 @@ class D2LAPIClient:
 
         return groups
 
-    def course_table_of_contents(self, org_unit) -> List[dict]:
-        """
-        Get a list of modules in the given course.
+    def list_files(self, org_unit) -> List[dict]:
+        """Get a nested list of files and folders for the given `org_unit`."""
 
-        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-toc
-        """
-        response = self._api.request(
-            "GET", self._api.api_url(f"/{org_unit}/content/toc", product="le")
-        )
-        return D2LTableOfContentsSchema(response).parse().get("modules", [])
+        modules = self._get_course_modules(org_unit)
+        return list(self._find_files(modules))
 
     @staticmethod
     def get_api_user_id(user_id: str):
@@ -134,3 +133,37 @@ class D2LAPIClient:
         so we don't take the second part of the string but the last.
         """
         return user_id.split("_")[-1]
+
+    def _get_course_modules(self, org_unit) -> List[dict]:
+        """
+        Get a list of modules in the given course.
+
+        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-toc
+        """
+        response = self._api.request(
+            "GET", self._api.api_url(f"/{org_unit}/content/toc", product="le")
+        )
+        return D2LTableOfContentsSchema(response).parse().get("modules", [])
+
+    def _find_files(self, modules):
+        """Recursively find files in modules."""
+        for module in modules:
+            module_files = [
+                {
+                    "type": "File",
+                    "display_name": topic["display_name"],
+                    "id": topic["id"],
+                    "updated_at": topic["updated_at"],
+                }
+                for topic in module.get("topics", [])
+                if topic.get("type") == "File"
+            ]
+
+            module_children = self._find_files(module.get("modules", []))
+            yield {
+                "type": "Folder",
+                "display_name": module["display_name"],
+                "id": module["id"],
+                "updated_at": module["updated_at"],
+                "children": module_files + list(module_children),
+            }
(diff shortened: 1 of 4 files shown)


lms/services/d2l_api/client.py
```diff
@@ -41,6 +41,9 @@ class _D2LModuleSchema(Schema):
     class Meta:
         unknown = EXCLUDE
 
+    id = fields.Int(required=True, data_key="ModuleId")
```
marcospri, line 44: Now that we are exposing folder we need these.

lms/services/d2l_api/client.py
```diff
@@ -131,3 +153,14 @@ def get_api_user_id(user_id: str):
         so we don't take the second part of the string but the last.
         """
         return user_id.split("_")[-1]
+
+    def _get_course_modules(self, org_unit) -> List[dict]:
```
marcospri, line 157: I like to keep this method with the 1-1 relation to the API.

Keeping it private and not testing it independently in any case.

lms/services/d2l_api/client.py
```diff
@@ -106,16 +109,35 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
...
-        https://docs.valence.desire2learn.com/res/content.html#get--d2l-api-le-(version)-(orgUnitId)-content-toc
-        """
-        response = self._api.request(
-            "GET", self._api.api_url(f"/{org_unit}/content/toc", product="le")
-        )
-        return D2LTableOfContentsSchema(response).parse().get("modules", [])
+    def list_files(self, org_unit) -> List[dict]:
+        """Get a list of files and folders for the given `org_unit`."""
+
+        def _find_files(modules):
```
marcospri, line 115: Not sure why but this seem right as nested function.

lms/services/d2l_api/client.py
```diff
@@ -106,16 +109,35 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
...
+            """Recursively find files in modules."""
+            for module in modules:
+                module_files = []
+                for topic in module.get("topics", []):
+                    if topic.get("type") == "File":
+                        module_files.append(
+                            {
+                                "type": "File",
```
marcospri, line 123: Spelling every field here as we don't want to expose everything (ie topics/modules).

We could use another marshmallow schema for the output side of this but it seems overkill.

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -118,23 +130,135 @@ def test_group_set_groups_with_user_id(self, svc, d2l_groups_schema, groups):
...
+    def test_list_files(
         self,
         svc,
-        D2LTableOfContentsSchema,
         basic_client,
+        modules,
+        files,
     ):
-        svc.course_table_of_contents("COURSE_ID")
+        basic_client.request.return_value.json.return_value = modules
```
marcospri, line 172: This test is not mocking the schema.

I find this a more useful test as we can test what we'll return given a response from the API direclty.

lms/services/d2l_api/client.py
```diff
@@ -32,7 +32,7 @@ class Meta:
         unknown = EXCLUDE
 
     id = fields.Str(required=True, data_key="Identifier")
-    name = fields.Str(required=True, data_key="Title")
+    display_name = fields.Str(required=True, data_key="Title")
```
marcospri, line 35: Use the names we'll use in the rest of the app.

lms/views/api/d2l/files.py
```diff
@@ -4,17 +4,6 @@
 from lms.services.d2l_api import D2LAPIClient
 
 
-def _find_files(modules):
```
marcospri, line 7: Moved to the service

lms/views/api/d2l/files.py
```diff
@@ -23,15 +12,6 @@ def _find_files(modules):
 )
 def list_files(_context, request):
     """Return the list of files in the given course."""
-    modules = request.find_service(D2LAPIClient).course_table_of_contents(
+    return request.find_service(D2LAPIClient).list_files(
```
marcospri, line 15: Not much going on here now.

jon-betts (approved):
I've tested this in the alternate branch, and it appears to work as we'd expect. I've not tried picking a file as that was listed as not working.

I like the removal of layers of indirection we had before. Previously the service returned a format we didn't want, which was then mapped again in a few places before being returned to the user. This forced lots of testing of these conversions. Now the service is in charge of all of it, and returns the format we want.

This means although this PR adds functionality, it's actually removing code. I have some suggestions for removing some more here as well:

##  I think we should

 * Convert the nested function to a normal private function. We don't make use of the closure and I think this would reduce indentation and increase readability
 * I think the two test cases could be combined into one without losing anything. This would hopefully make it quite a bit shorter

## I think we could do

 * Some list comprehensions might help with the recursive method
 * Consider in-lining the main content of the method from the existing private method as the existing method doesn't have much functionality of it's own

## I don't know if we should

Can D2L folders contain things which aren't PDFs? I can't see anything which limits the files you can pick to PDFs and a user might pick another document type expecting it to work?

Should we display folder if they don't contain any files? We could prune directories which don't have a least one eligible file descendent.

Pros:

 * Doesn't waste a users time clicking into nested empty folders
 * Gets the user where they want to go in fewer clicks

Cons:

 * Real users probably don't have lots of empty folders (although they might have folders which contain no PDFs?)
 * Users might be confused if the structure doesn't mirror 1:1 the structure they expect?

lms/services/d2l_api/client.py
```diff
@@ -106,16 +109,35 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
...
+                for topic in module.get("topics", []):
+                    if topic.get("type") == "File":
+                        module_files.append(
+                            {
+                                "type": "File",
+                                "display_name": topic["display_name"],
+                                "id": topic["id"],
+                                "updated_at": topic["updated_at"],
+                            }
+                        )
```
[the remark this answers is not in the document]
marcospri: Went for only one level of list comprehensions, the nested one was a bit trickier to read.

marcospri:
Merging this one to get the nested folder support in and tackling filtering out PDFs in a separate PR.

Issue #4773, opened 2022-12-16
User varchar_enum in grouping.type


marcospri:
https://github.com/hypothesis/lms/pull/4660/files#r1026234684

PR #4804, opened 2022-12-27
Remove check constraint in Grouping.type

marcospri:
For:

- https://github.com/hypothesis/lms/issues/4773

## Testing for the migration

```
hdev alembic upgrade head
dev run-test-pre: PYTHONHASHSEED='551793646'
dev run-test: commands[0] | alembic -c /home/marcos/hypo/lms/conf/alembic.ini upgrade head
INFO  [alembic.runtime.migration] Running upgrade 0b7391f5b1e3 -> debaebd4a95c, Remove enum DB constraint.
```

```
dev alembic downgrade -1
dev run-test-pre: PYTHONHASHSEED='3452916469'
dev run-test: commands[0] | alembic -c /home/marcos/hypo/lms/conf/alembic.ini downgrade -1
INFO  [alembic.runtime.migration] Running downgrade debaebd4a95c -> 0b7391f5b1e3, Remove enum DB constraint.
```

### DIFF
--- a/lms/migrations/versions/debaebd4a95c_remove_enum_constraint.py
+++ b/lms/migrations/versions/debaebd4a95c_remove_enum_constraint.py
@@ -1,0 +1,29 @@
+"""
+Remove enum DB constraint.
+
+Revision ID: debaebd4a95c
+Revises: 0b7391f5b1e3
+Create Date: 2022-12-27 14:19:35.278944
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "debaebd4a95c"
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
Yep this looks good. I've tested the upgrade and downgrade. The app appears (after very light testing) to work with this code in either situation as well, so a separate PR for the code is not necessary, but some testing will be needed when deployed (unlike a normal DB patch).

I think this is the direction we want to go in, as it leaves the Python side in the driving seat for the enumeration. I'd probably prefer to see something like this on the front end eventually, although I think there's a stronger argument for a check there as the front end currently needs explicit support for each type.

Issue #4771, opened 2022-12-16
Store all files in the DB as we do with other products


marcospri:
In blackboard and canvas we use the file service to store all files as we see them (ie not only the ones selected as assignment documents).

This is useful for course copy and general bookkeeping.

PR #4811, opened 2022-12-30
Store D2L files into the DB

marcospri:
For:

- https://github.com/hypothesis/lms/issues/4771

## Testing

- `make sql` 
- `truncate "file";` to get a blank slate.
- Reconfigure https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View
- Pick the D2L files option (no need to pick any),
- Check the new data on the DB with `select * from file;`

### DIFF
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -61,9 +61,10 @@ class D2LTableOfContentsSchema(RequestsResponseSchema):
 
 
 class D2LAPIClient:
-    def __init__(self, basic_client, request):
+    def __init__(self, basic_client, request, file_service):
         self._api = basic_client
         self._request = request
+        self._file_service = file_service
 
     def get_token(self, authorization_code):
         """
@@ -115,9 +116,11 @@ class D2LAPIClient:
 
     def list_files(self, org_unit) -> List[dict]:
         """Get a nested list of files and folders for the given `org_unit`."""
-
         modules = self._get_course_modules(org_unit)
-        return list(self._find_files(org_unit, modules))
+        files = list(self._find_files(org_unit, modules))
+        self._file_service.upsert(list(self._files_for_storage(org_unit, files)))
+
+        return files
 
     def public_url(self, org_unit, file_id) -> str:
         """
@@ -193,3 +196,17 @@ class D2LAPIClient:
                 "updated_at": module["updated_at"],
                 "children": module_files + list(module_children),
             }
+
+    def _files_for_storage(self, course_id, files, parent_id=None):
+        for file in files:
+            yield {
+                "type": "d2l_file" if file["type"] == "File" else "d2l_folder",
+                "course_id": course_id,
+                "lms_id": file["id"],
+                "name": file["display_name"],
+                "parent_lms_id": parent_id,
+            }
+
+            yield from self._files_for_storage(
+                course_id, file.get("children", []), file["id"]
+            )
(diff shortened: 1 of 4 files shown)


lms/services/d2l_api/client.py
```diff
@@ -115,9 +116,11 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
 
     def list_files(self, org_unit) -> List[dict]:
         """Get a nested list of files and folders for the given `org_unit`."""
-
         modules = self._get_course_modules(org_unit)
```
marcospri, line 119: Get the files from the API

lms/services/d2l_api/client.py
```diff
@@ -115,9 +116,11 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
...
-        return list(self._find_files(org_unit, modules))
+        files = list(self._find_files(org_unit, modules))
```
marcospri, line 120: Translate them to the format expected by the frontend (nesting them)

lms/services/d2l_api/client.py
```diff
@@ -115,9 +116,11 @@ def group_set_groups(self, org_unit, group_category_id, user_id=None):
...
+        self._file_service.upsert(list(self._files_for_storage(org_unit, files)))
```
marcospri, line 121: Now translating them again to the format expected by `_files_for_storage`.

`_files_for_storage` is picky about what it takes, any dictionary won't do.

lms/services/d2l_api/client.py
```diff
@@ -193,3 +196,17 @@ def _find_files(self, course_id, modules):
                 "updated_at": module["updated_at"],
                 "children": module_files + list(module_children),
             }
+
+    def _files_for_storage(self, course_id, files, parent_id=None):
+        for file in files:
+            yield {
+                "type": "d2l_file" if file["type"] == "File" else "d2l_folder",
```
marcospri, line 203: Diffent type for the DB

lms/services/d2l_api/client.py
```diff
@@ -193,3 +196,17 @@ def _find_files(self, course_id, modules):
...
+                "course_id": course_id,
+                "lms_id": file["id"],
+                "name": file["display_name"],
+                "parent_lms_id": parent_id,
+            }
+
+            yield from self._files_for_storage(
```
marcospri, line 210: For the DB this needs to be a flat list

lms/services/d2l_api/client.py
```diff
@@ -193,3 +196,17 @@ def _find_files(self, course_id, modules):
...
+                course_id, file.get("children", []), file["id"]
+            )
```
jon-betts, line 212: Presumably the two formats here are:

 * The format expected by the front end
 * The format expected by the DB / file service
 
It's a shame they are different, but they are. 

I wonder if it's worth extracting this conversion between formats and flattening behavior into something re-usable?  Then you could emit the tree with children in one function, and then have it flatten and reformat itself as required? This could presumably be used elsewhere (Canvas / Blackboard?), where we really should be doing the same thing.

This might also nudge us towards emitting the correct things at the service level, rather than our usual of emitting something in an LMS specific format and then having to convert it a lot.

lms/services/d2l_api/client.py
jon-betts, line 121: I was checking, but both listifications here are required, as these inputs get iterated multiple times, and would otherwise get consumed as a generator.

jon-betts (approved):
This seems pretty straight forward and works as described.

I can't remember why we store folders, but it seems like a better to have it and not need it situation.

I've put an idea about pulling out the format shifting / flattening behavior. One or both of those might be a good idea (I think the flattening is the most useful), but it's not a recommendation for this PR.

PR #4821, opened 2023-01-09
Fix D2L file enable toggle

marcospri:
Just checking "files_enabled" will be true in both d2l and blackboard at the same time.

Added a family check.

### In main

Launch as teacher:

You'll see the D2L button in the file picker.

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2128/View?ou=6782
(with the toggle enabled at http://localhost:8001/admin/instance/id/107/)

the D2L files button will appear.

### Switch to `fix-d2l-files-toggle`

doesn't show the toggle

While in D2L https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2128/View?ou=6782

it does appaear.

### DIFF
--- a/lms/resources/_js_config/file_picker_config.py
+++ b/lms/resources/_js_config/file_picker_config.py
@@ -11,8 +11,10 @@ class FilePickerConfig:
     @classmethod
     def d2l_config(cls, request, _application_instance):
         """Get D2L files config."""
-        files_enabled = request.product.settings.files_enabled
-
+        files_enabled = (
+            request.product.family == D2L.family
+            and request.product.settings.files_enabled
+        )
         config = {"enabled": files_enabled}
         if files_enabled:
             config["listFiles"] = {
--- a/tests/unit/lms/resources/_js_config/file_picker_config_test.py
+++ b/tests/unit/lms/resources/_js_config/file_picker_config_test.py
@@ -30,19 +30,27 @@ class TestFilePickerConfig:
 
         assert config == expected_config
 
-    @pytest.mark.parametrize("files_enabled", [False, True])
-    def test_d2l_config(self, pyramid_request, files_enabled):
-
+    @pytest.mark.parametrize(
+        "family,enabled,expected",
+        [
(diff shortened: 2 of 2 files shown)


lms/resources/_js_config/file_picker_config.py
```diff
@@ -11,8 +11,10 @@ class FilePickerConfig:
     @classmethod
     def d2l_config(cls, request, _application_instance):
         """Get D2L files config."""
-        files_enabled = request.product.settings.files_enabled
-
+        files_enabled = (
```
marcospri, line 14: Yes, files are enabled for "this" LMS but "this" might not be D2L right now.

We could send the title for the button here ("D2L files" | "Blackboard files") or the frontend could do that itself.

For now this is a straightforward fix.

jon-betts: I think this comment caused a bit of confusion, so I'll try and explain. 

Our setting structure is a two level deep naming scheme where we have an outer and inner name like `canvas.sections_enabled`, `canvas.files`, or `microsoft_onedrive.files_enabled`.

I think the issue we are having is that in the `<a>.<b>` scheme the type of thing that `<a>` is changes. In some cases it's a document source (`microsoft_onedrive`) in others it's a product, and in an increasing number it's both at the same time. This confusion is extended into the product system which uses these keys now. This last part is the confusion I would be keen to eliminate.

This is just an example to show the idea, not an actual suggestion to fix it, but if the concepts were differentiated we might have:

* `PRODUCT:canvas.sections_enabled`
* `DOCUMENT_SOURCE:canvas_files.enabled`
* `DOCUMENT_SOURCE:microsoft_onedrive.enabled`

Of course there is a link between the document source "Canvas files" and the product "Canvas" we would still want and need to represent and use, but this would make a distinction between the two.

This also naturally leaves space for a second type of file integration for a given product: Canvas for example has different ways of storing files of which we only integrate with one.

jon-betts (approved):

PR #4831, opened 2023-01-10
Remove duplication to get auth token in JSConfig.

marcospri:
From https://github.com/hypothesis/lms/pull/4753#pullrequestreview-1240691743

In a separate PR to better keep track of it.


jon-betts (approved):
Finally, I can sleep again.

PR #4832, opened 2023-01-10
Limit the scopes for groups to the minimum necessary

marcospri:
No need to require more than we need.

The two endpoints we use:

https://docs.valence.desire2learn.com/res/groups.html#get--d2l-api-lp-(version)-(orgUnitId)-groupcategories-

https://docs.valence.desire2learn.com/res/groups.html#get--d2l-api-lp-(version)-(orgUnitId)-groupcategories-(groupCategoryId)-groups-

need the same scope.

We might be able to also remove the `core:*` one but I'll do more testing and a separate PR

## Testing

Test the assignment in https://aunltd.brightspacedemo.com/d2l/le/content/6782/Home?itemIdentifier=D2L.LE.Content.ContentObject.ModuleCO-2122

### DIFF
--- a/lms/views/api/d2l/authorize.py
+++ b/lms/views/api/d2l/authorize.py
@@ -7,7 +7,7 @@ from lms.security import Permissions
 from lms.services.d2l_api import D2LAPIClient
 from lms.validation.authentication import OAuthCallbackSchema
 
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
Groups still show up in the testing assignment, so I think this is fine?

PR #4834, opened 2023-01-10
Include LTIRoles in LTIUser

marcospri:
For:

Include LTIRoles in LTIUser.

The main motivation will be to later use this objects to determine the role of the current user, leveraging LTIRole instead of just matching strings:

https://github.com/hypothesis/lms/blob/668361e73615f9efc6b351bb2533433cc926d983/lms/models/lti_user.py#L39

That would centralize all role decision in LTIRole.

See a draft of that here:

- https://github.com/hypothesis/lms/pull/5123

### DIFF
--- a/lms/events/event.py
+++ b/lms/events/event.py
@@ -7,7 +7,6 @@ from sqlalchemy import inspect
 
 from lms.db import BASE
 from lms.models import EventType
-from lms.services.lti_role_service import LTIRoleService
 
 
 @dataclass
@@ -64,12 +63,7 @@ class LTIEvent(BaseEvent):
     # These methods provide defaults for each field and are used in
     # `BaseEvent.__post_init__` above.
     def _get_role_ids(self):
-        return [
-            role.id
-            for role in self.request.find_service(LTIRoleService).get_roles(
-                self.request.lti_user.roles
-            )
-        ]
+        return [role.id for role in self.request.lti_user.lti_roles]
 
     def _get_application_instance_id(self):
         if application_instance := self.request.find_service(
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
+from typing import List
 
 from lms.models.h_user import HUser
+from lms.models.lti_role import LTIRole
 
 
 @dataclass
@@ -11,6 +13,9 @@ class LTIUser:
     """The user_id LTI launch parameter."""
 
     roles: str
+    """The user's raw LTI roles string."""
+
+    lti_roles: List[LTIRole]
     """The user's LTI roles."""
 
     tool_consumer_instance_guid: str
@@ -44,9 +49,11 @@ class LTIUser:
         return "learner" in self.roles.lower()
 
     @staticmethod
-    def from_auth_params(application_instance, lti_core_schema):
+    def from_auth_params(request, application_instance, lti_core_schema):
         """Create an LTIUser from a LTIV11CoreSchema like dict."""
-        return LTIUser(
+
+        return LTIUser.unserialize(
+            request,
             user_id=lti_core_schema["user_id"],
             application_instance_id=application_instance.id,
             roles=lti_core_schema["roles"],
@@ -58,6 +65,15 @@ class LTIUser:
             ),
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
+
+    @staticmethod
+    def unserialize(request, **kwargs: dict):
+        # pylint: disable= import-outside-toplevel
+        from lms.services.lti_role_service import LTIRoleService
+
+        lti_roles = request.find_service(LTIRoleService).get_roles(kwargs["roles"])
+
+        return LTIUser(lti_roles=lti_roles, **kwargs)
 
     def serialize(self) -> dict:
         """
--- a/lms/models/lti_user.py
+++ b/lms/models/lti_user.py
@@ -48,48 +48,6 @@ class LTIUser:
         """Whether this user is a learner."""
         return "learner" in self.roles.lower()
 
-    @staticmethod
-    def from_auth_params(request, application_instance, lti_core_schema):
-        """Create an LTIUser from a LTIV11CoreSchema like dict."""
-
-        return LTIUser.unserialize(
-            request,
-            user_id=lti_core_schema["user_id"],
-            application_instance_id=application_instance.id,
-            roles=lti_core_schema["roles"],
-            tool_consumer_instance_guid=lti_core_schema["tool_consumer_instance_guid"],
-            display_name=display_name(
-                lti_core_schema["lis_person_name_given"],
-                lti_core_schema["lis_person_name_family"],
-                lti_core_schema["lis_person_name_full"],
-            ),
-            email=lti_core_schema["lis_person_contact_email_primary"],
-        )
-
-    @staticmethod
-    def unserialize(request, **kwargs: dict):
-        # pylint: disable= import-outside-toplevel
-        from lms.services.lti_role_service import LTIRoleService
-
-        lti_roles = request.find_service(LTIRoleService).get_roles(kwargs["roles"])
-
-        return LTIUser(lti_roles=lti_roles, **kwargs)
-
-    def serialize(self) -> dict:
-        """
-        Return a dict representing the LTIUser.
-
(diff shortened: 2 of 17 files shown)


lms/models/lti_user.py
```diff
@@ -11,6 +13,9 @@ class LTIUser:
     """The user_id LTI launch parameter."""
 
     roles: str
```
marcospri, line 15: Renaming this will take a bit of extra effort as there are in flight tokens with the current name.

We could tackled that in a follow up.

lms/validation/authentication/_bearer_token.py
```diff
@@ -101,7 +101,7 @@ def lti_user(self, location) -> LTIUser:
                 messages={"authorization": ["Invalid session token"]}
             ) from err
 
-        return LTIUser(**jwt_data)
+        return LTIUser.unserialize(self.context["request"], **jwt_data)
```
marcospri, line 104: Calling a method now instead of just using the default constructor.

lms/validation/authentication/_oauth.py
```diff
@@ -112,7 +112,7 @@ def lti_user(self):
             raise MissingStateParamError() from err
 
         decoded_user = self._decode_state(state)["user"]
-        return LTIUser(**decoded_user)
+        return LTIUser.unserialize(self.context["request"], **decoded_user)
```
marcospri, line 115: Same, call the explicit method.

lms/events/event.py
```diff
@@ -64,12 +63,7 @@ def _get_user_id(self):
     # These methods provide defaults for each field and are used in
     # `BaseEvent.__post_init__` above.
     def _get_role_ids(self):
-        return [
-            role.id
-            for role in self.request.find_service(LTIRoleService).get_roles(
-                self.request.lti_user.roles
-            )
-        ]
+        return [role.id for role in self.request.lti_user.lti_roles]
```
marcospri, line 66: LTIRoles are now part of the object, no need to go from str -> LTIRoles

lms/models/lti_user.py
```diff
@@ -59,6 +66,15 @@ def from_auth_params(application_instance, lti_core_schema):
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
 
+    @staticmethod
+    def unserialize(request, **kwargs: dict):
```
marcospri, line 70: LTIUser gets serialized/unserialized as part of different authentication schemas.

I reckon it makes sense to make that explicit.

Not sure about the "unserialize" name thought, we do have `serialize` below

lms/models/lti_user.py
```diff
@@ -44,9 +49,11 @@ def is_learner(self):
         return "learner" in self.roles.lower()
 
     @staticmethod
-    def from_auth_params(application_instance, lti_core_schema):
+    def from_auth_params(request, application_instance, lti_core_schema):
```
marcospri, line 52: This method itself could become "unserlize" but it will take more work for clients that already use it. I reckon we should keep it for now.

lms/models/lti_user.py
jon-betts: I think `deserialize` would be the norm?

lms/models/lti_user.py
```diff
@@ -59,6 +66,15 @@ def from_auth_params(application_instance, lti_core_schema):
             email=lti_core_schema["lis_person_contact_email_primary"],
         )
 
+    @staticmethod
+    def unserialize(request, **kwargs: dict):
+        # pylint: disable= import-outside-toplevel
+        from lms.services.lti_role_service import LTIRoleService
+
+        lti_roles = request.find_service(LTIRoleService).get_roles(kwargs["roles"])
```
jon-betts, line 74: This really screams to me that something here isn't quite right. I don't think our models should be calling services. The import here is something that can probably be worked around, but I suspect is indicating the same.

I like keeping the serialization and deserializaton next to each other, but this seems to lead to either both going to a service, or both coming here (but pulling in a service). Both bad in their own way.

The minimal tweak would be to inject the service as an argument. We already inject the request, and we only use it to get the service, but I do think it's telling us something bigger isn't right.

jon-betts (commented):
I think that having to inject the service into the model for it to call heavily suggests something isn't quite right here. I'm not 100% on the solution, but I think it could do with some more thought. This feels very much like service code getting into the models.

One option that occurs to me is to store the `lti_roles` on the user object, but not make it responsible for generating them on instantiation.

So you could have an accessor that raises an exception like "You need to call `populate_roles` from the LTIRole service before you can access lti_roles". This means the object can be instantiated outside of the request / service eco-system.

Alternatively you make it an accessor/helper on the object like:

```python
def parse_roles(self, lti_role_service):
    self._lti_roles = lti_role_service.get_roles(self.roles)

@property
def lti_roles(self):
    if not self._lti_roles:
        raise ValueError("You must call `parse_roles()` before accessing lti_roles`)
    return self._lti_roles
```

lms/validation/authentication/_bearer_token.py
```diff
@@ -101,7 +102,7 @@ def lti_user(self, location) -> LTIUser:
                 messages={"authorization": ["Invalid session token"]}
             ) from err
 
-        return LTIUser(**jwt_data)
+        return self._lti_user_service.deserialize(**jwt_data)
```
marcospri, line 105: Call the new deserialize method here instead of the raw constructor.

lms/validation/authentication/_oauth.py
```diff
@@ -112,7 +116,7 @@ def lti_user(self):
             raise MissingStateParamError() from err
 
         decoded_user = self._decode_state(state)["user"]
-        return LTIUser(**decoded_user)
+        return self._lti_user_service.deserialize(**decoded_user)
```
marcospri, line 119: Call the new deserialize method here instead of the raw constructor here too.

lms/services/lti_user.py
```diff
@@ -0,0 +1,56 @@
+from lms.models.lti_user import LTIUser, display_name
+from lms.services import LTIRoleService
+
+
+class LTIUserService:
```
marcospri, line 5: This service could also contain this:

https://github.com/hypothesis/lms/blob/09c3da9cc73c2a030f093f2719567e0a80438d70/lms/security.py#L225


as `from_request` or similar.


I'd save that for a follow up PR if we go with the service approach.

lms/services/lti_user.py
```diff
@@ -0,0 +1,56 @@
...
+
+    LTIUser is the center of many of our authentication schemas.
+    This service handles de/serialization of LTIUsers.
+    """
+
+    def __init__(self, lti_role_service):
+        self._lti_roles_service = lti_role_service
+
+    def from_auth_params(self, application_instance, lti_core_schema) -> LTIUser:
+        """Create an LTIUser from a LTIV11CoreSchema like dict."""
```
jon-betts, line 17: This description is a bit weird. Do we just mean that the dict contains LTI 1.1 parameters?

lms/validation/authentication/_bearer_token.py
```diff
@@ -52,7 +53,7 @@ def authorization_param(self, lti_user: LTIUser) -> str:
         :arg lti_user: the LTI user to return an auth param for
         """
         token = self._jwt_service.encode_with_secret(
-            lti_user.serialize() if lti_user else {},
+            self._lti_user_service.serialize(lti_user) if lti_user else {},
```
jon-betts, line 56: This if could go inside the serialization method?

lms/services/lti_user.py
```diff
@@ -0,0 +1,56 @@
...
+        return {
+            "user_id": lti_user.user_id,
+            "roles": lti_user.roles,
+            "tool_consumer_instance_guid": lti_user.tool_consumer_instance_guid,
+            "display_name": lti_user.display_name,
+            "application_instance_id": lti_user.application_instance_id,
+            "email": lti_user.email,
+        }
+
+    def deserialize(self, **kwargs: dict) -> LTIUser:
```
jon-betts, line 48: I wonder if accepting a dict here would be better than kwargs? It would leave space for us to pass other modifiers in future.

jon-betts (approved):
I think this works quite well. It's not massively disruptive and being a service allows us to do much more fancy stuff than the model.

This also leaves the model more in-active, which is probably good.

PR #4835, opened 2023-01-10
Add views for sensitive user information in LMS

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/121


lms/data_tasks/report/create_from_scratch/02_entities/01_users/03_users_sensitive/01_create_view.sql
```diff
@@ -0,0 +1,30 @@
...
+
+        -- The order here should ensure empty strings and nulls are sorted to
+        -- the back, so the "first" value should be non-null if possible
+        FIRST_VALUE(users.email) OVER (
+            PARTITION BY user_map.user_id
+            ORDER BY users.email
+            RANGE BETWEEN
+                UNBOUNDED PRECEDING AND
+                UNBOUNDED FOLLOWING
+        ) AS email,
```
jon-betts, line 17: This is kind of a lot, but does seem to work. I don't know why sorting ASC (which is the default) doesn't result in NULL and empty string coming first, but it doesn't seem to.

lms/data_tasks/report/create_from_scratch/02_entities/01_users/03_users_sensitive/01_create_view.sql
```diff
@@ -0,0 +1,30 @@
+DROP MATERIALIZED VIEW IF EXISTS report.users_sensitive CASCADE;
```
jon-betts, line 1: The thinking with the name here is that we have have `<x>` and `<x>_sensitive` to remind us to treat these carefully.

marcospri (approved):
Had to read the query a few times but this makes sense :+1:

PR #4838, opened 2023-01-12
Fix course copy assignment URL for BB and D2L in 1.3

marcospri:
In blackboard and D2L we rely on a parameter to point to the old id of an assignment and get its URL from the old one in the DB.

For LTI1.3 this parameter is in a different location.

# Testing

- Configure the original assignment 

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782

for with example with: https://buildmedia.readthedocs.org/media/pdf/docs/latest/docs.pdf

You'll need something like 

`make sql` 
`truncate assignment cascade;`

if you already configured it before.

- Launch the equivalent assignment in the copied course. 

https://aunltd.brightspacedemo.com/d2l/le/content/6845/viewContent/2310/View
 
 
In `main` it will show the `configure assignment` selector.

In `course-copy-1.3` it will point already to the same PDF as the orignial.

### DIFF
--- a/lms/services/document_url.py
+++ b/lms/services/document_url.py
@@ -93,17 +93,18 @@ class DocumentURLService:
         for param in (
             "resource_link_id",  # A normal LTI (non-deep linked) launch
             "resource_link_id_history",  # A Blackboard course we can copy
-            "ext_d2l_resource_link_id_history",  # Ditto for Brightspace
+            "ext_d2l_resource_link_id_history",  # Ditto for D2L
+            "custom_ResourceLink.id.history",  # LTI 1.3 location for D2L and blackboard
         ):
             if (resource_link_id := request.lti_params.get(param)) and (
-                assigment := self._assignment_service.get_assignment(
+                assignment := self._assignment_service.get_assignment(
                     tool_consumer_instance_guid=request.lti_params.get(
                         "tool_consumer_instance_guid"
                     ),
                     resource_link_id=resource_link_id,
                 )
             ):
-                return assigment.document_url
+                return assignment.document_url
 
         return None
 
--- a/tests/unit/lms/services/document_url_test.py
+++ b/tests/unit/lms/services/document_url_test.py
@@ -108,6 +108,7 @@ class TestDocumentURLService:
             "resource_link_id",
             "ext_d2l_resource_link_id_history",
             "resource_link_id_history",
+            "custom_ResourceLink.id.history",
         ),
     )
     def test_get_document_url_with_assignment_in_db(


lms/services/document_url.py
```diff
@@ -93,17 +93,18 @@ def _from_assignment_in_db(self, request):
         for param in (
             "resource_link_id",  # A normal LTI (non-deep linked) launch
             "resource_link_id_history",  # A Blackboard course we can copy
-            "ext_d2l_resource_link_id_history",  # Ditto for Brightspace
+            "ext_d2l_resource_link_id_history",  # Ditto for D2L
```
marcospri, line 96: D2L it's just sorter for the next line

lms/services/document_url.py
```diff
@@ -93,17 +93,18 @@ def _from_assignment_in_db(self, request):
...
+            "custom_ResourceLink.id.history",  # LTI 1.3 location for D2L and blackboard
```
marcospri, line 97: Our lti_parms will prefix all custom parameters with `custom_`

Note that:

- In D2L we were already including this but not using it
- In Blackboard the custom parameter was not configured.

seanh: Ok, so this is when we're looking for the document URL in an `Assignment` stored in our DB, which we do if there's no deep linking-provided URL, Canvas files URL, or legacy VitalSource book URL. And this change means that if there's no `"resource_link_id"`, `"resource_link_id_history"` or `"ext_d2l_resource_link_id_history"` query param then we'll now look for a `"custom_ResourceLink.id.history"` query param as well, which we weren't looking for before. And in the case of D2L launches it is already sending us this `"custom_ResourceLink.id.history"` param, in the case of Blackboard admins will have to do an update to make it send us that param.

One thing I notice is that I think this means we'll potentially look for the `"custom_ResourceLink.id.history"` for *any* LTI launch (as long as all the earlier places where we look for a document URL don't find one). It is not restricted to only D2L and Blackboard LTI 1.3 launches, at least not explicitly. But that already seems to be the case with everything that `DocumentURLService` does, and I'm not sure that's a bad thing.

lms/services/document_url.py
seanh: Is it worth expanding the `"resource_link_id_history"` and `"ext_d2l_resource_link_id_history"` params to say that these are used by Blackboard and D2L _in LTI 1.1_?

lms/services/document_url.py
```diff
@@ -93,17 +93,18 @@ def _from_assignment_in_db(self, request):
...
         ):
             if (resource_link_id := request.lti_params.get(param)) and (
-                assigment := self._assignment_service.get_assignment(
+                assignment := self._assignment_service.get_assignment(
```
seanh, line 100: Typo fix :+1:

seanh (commented):
The code change looks good to me :+1: I can help test once there are testing instructions

marcospri:
Added some sanity check testing notes to the description.

seanh (approved):
:+1: Works!

PR #4850, opened 2023-01-16
Migrate LMSFilePicker to TS

acelaya:
While working on #4769, I took the opportunity to migrate the `LMSFilePicker` component to TS, as the majority of changes are happening there and it's more convenient to not have to deal with JSDocs.

### DIFF
--- a/lms/static/scripts/frontend_apps/api-types.js
+++ b/lms/static/scripts/frontend_apps/api-types.js
@@ -13,7 +13,7 @@
  * @typedef File
  * @prop {string} id - Identifier for the resource within the LMS's file storage
  * @prop {string} display_name - Name of the resource to present in the file picker
- * @prop {string} updated_at - An ISO 8601 date string
+ * @prop {string} [updated_at] - An ISO 8601 date string
  * @prop {'File'|'Folder'} [type]
  * @prop {APICallInfo} [contents] - APICallInfo for fetching a folders's
  *   content. Only present if `type` is 'Folder'.
--- a/lms/static/scripts/frontend_apps/components/LMSFilePicker.tsx
+++ b/lms/static/scripts/frontend_apps/components/LMSFilePicker.tsx
@@ -7,6 +7,8 @@ import classnames from 'classnames';
 
 import { useCallback, useEffect, useState } from 'preact/hooks';
 
+import type { File } from '../api-types';
+import type { APICallInfo } from '../config';
 import { isAuthorizationError } from '../errors';
 import { apiCall } from '../utils/api';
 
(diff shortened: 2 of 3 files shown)


lms/static/scripts/frontend_apps/api-types.js
```diff
@@ -13,7 +13,7 @@
  * @typedef File
  * @prop {string} id - Identifier for the resource within the LMS's file storage
  * @prop {string} display_name - Name of the resource to present in the file picker
- * @prop {string} updated_at - An ISO 8601 date string
+ * @prop {string} [updated_at] - An ISO 8601 date string
```
acelaya, line 16: This was already being treated as optional, and it was only explicitly highlighted by TS once `LMSFilePicker` was migrated.

robertknight: Does this definition agree with the backend code? ie. Are there cases where `updated_at` is not set by the backend, or was the frontend code that treats this as optional incorrect?

lms/static/scripts/frontend_apps/components/LMSFilePicker.tsx
```diff
@@ -45,66 +41,90 @@ function NoFilesMessage({ href, inSubfolder }) {
...
+
+  /**
+   * Render path breadcrumbs and allow sub-folder navigation?
+   */
+  withBreadcrumbs?: boolean;
+};
+
+type FetchingState = {
+  state: 'fetching';
+  // Flag indicating that files are being re-fetched after clicking "Reload"
```
robertknight, line 79: Same comment about `/** ... */` vs `//` comments as above.

lms/static/scripts/frontend_apps/components/LMSFilePicker.tsx
```diff
@@ -259,7 +270,7 @@ export default function LMSFilePicker({
     case 'authorize':
       continueButton = (
         <AuthButton
-          authURL={/** @type {string} */ (listFilesApi.authUrl)}
+          authURL={listFilesApi.authUrl ?? ''}
```
robertknight, line 273: This should use `listFilesApi.authUrl!` to be equivalent to the previous code. If we get to this point and no `authUrl` has been provided, that is a mistake. Using a cast or `!` makes it clear what we are expecting, whereas adding a runtime `?? <fallback>` suggests there are valid cases where we expect no auth URL has been provided.

robertknight (approved):
This looks good except for a few small details. In summary:

1. JSDoc / TSDoc-style comments (`/** ... */`) should generally be used when documenting the purpose of top-level types / functions etc. in a module.
2. If there is a situation where you expect some nullable value/property should not be null for reasons that are not visible to TS, use an assertion `value!` rather `?? <fallback>` to make expectations clear to other readers.

lms/static/scripts/frontend_apps/api-types.js
acelaya: It's optional in the BE as well, because there was one LMS where this is not available (I would need to check which one)

lms/static/scripts/frontend_apps/components/LMSFilePicker.tsx
acelaya: we seem to have a linting rule forbidding this. Should I suppress it?

image

robertknight: Ah, yes. That rule should be turned off in all projects. It is part of @typescript-eslint's default ruleset. See also the client's ESLint config for some other typescript-eslint rules that we turned off.

lms/static/scripts/frontend_apps/api-types.js
acelaya: Ok, my mistake. Looks like it's mandatory in the BE, so we may need to check this

PR #4852, opened 2023-01-16
Filter out non-PDFs for D2L files

marcospri:
~~This needs #4769 before it can be tested without messing around with different branches.~~

We need to filter out non-PDF files as those are the only ones we can proxy / annotate.

In other LMS's this is done by either the API exposing a content-type query parameter (the LMS does the filtering for us) or a content type field we can filter ourselves by.

D2L offers neither of those we we'll have to rely on the extension of the file.

## Testing

- Compare the files in `Scratch area" while configuring: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782 and selecting D2L files

In main you'll get a `this_is_not_fine` file (which is an image).
In this branch you'll only get the PDF.

### DIFF
--- a/lms/services/d2l_api/client.py
+++ b/lms/services/d2l_api/client.py
@@ -37,6 +37,8 @@ class _D2LTopic(Schema):
     display_name = fields.Str(required=True, data_key="Title")
     updated_at = fields.Str(required=True, data_key="LastModifiedDate")
     type = fields.Str(required=True, data_key="TypeIdentifier")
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
Yep pretty much what it says on the tin. I had a quick squiz at the API and I couldn't see anything like a MIME type which could be useful for catching PDFs which don't have a PDF extension should such a thing exist.

Issue #4856, opened 2023-01-16
Handle new scopes for existing tokens


marcospri:
Handle inclusion of new scopes of existing tokens after enabling new features.

## Testing 

While on `main`

- tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token"

- Disable files for http://localhost:8001/admin/instance/id/106/
- Configure https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782 to use groups

This will get a token with only the scopes needed for groups

- Re-enable files https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782
- Reset assignments https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782
- Configure it again, now try ot use D2L files, you'll get an error:

```
lms.services.exceptions.ExternalRequestError: ExternalRequestError(message=None, request=Request(method='GET', url='https://aunltd.brightspacedemo.com/d2l/api/le/1.31/6782/content/toc', body=None), response=Response(status_code=403, reason='Forbidden', body='{ Errors: [ {Message: "Insufficient scope to call API.Required: content:toc:read"} ] }'), validation_errors=None)
```

marcospri:
This is different than:


```{"error":"invalid_scope","error_description":"Client is not authorized to request the scope content:toc:read"}```

that we get when the scopes are not declared correctly on the D2L install.

It seems that in that case we never get the callback result so we can't show a different error.

PR #4857, opened 2023-01-17
Handle tokens with not all the required scopes

marcospri:
For:

- #4856 

Raise OAuth2TokenError to force a new token generation with the necessary tokens.

From the PoV of the user this just changes the name of the button where they have to click ("Try again" vs "Authorize"). It's a slightly better experience but it's also more explicit in the code around what's going on in this case. It's not a lot of code to manage it in any case.

## Testing 

While on `main`:

- Remove existing tokens
`tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token"`
- Disable files for http://localhost:8001/admin/instance/id/106/
- Configure https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View?ou=6782 to use groups
This will get a token with only the scopes needed for groups.
- Re-enable files: http://localhost:8001/admin/instance/id/106/
- Reset the assignment 
`tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment cascade"`
- Configure it again, now try to use D2L files, you'll get an error:
```
lms.services.exceptions.ExternalRequestError: ExternalRequestError(message=None, request=Request(method='GET', url='https://aunltd.brightspacedemo.com/d2l/api/le/1.31/6782/content/toc', body=None), response=Response(status_code=403, reason='Forbidden', body='{ Errors: [ {Message: "Insufficient scope to call API.Required: content:toc:read"} ] }'), validation_errors=None)
```
- Don't click "Try again"
- Now, switch to this branch, refresh the assignment and repeat the process. You'll see the authorize screen instead of the error one this time.

### DIFF
--- a/lms/services/d2l_api/_basic.py
+++ b/lms/services/d2l_api/_basic.py
@@ -1,4 +1,4 @@
-from lms.services.exceptions import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError, OAuth2TokenError
 
 TOKEN_URL = "https://auth.brightspace.com/core/connect/token"
 """This is constant for all D2L instances"""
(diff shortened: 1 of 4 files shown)


jon-betts (approved):
Nice testing instructions on this one. This all seems to work as described. It's good to have one less situational dead end we can get stuck in for historical reasons.

I've put some suggestions on the specific code to do this, as I think there are some refactoring opportunities there which make the logic clearer.

Issue #4860, opened 2023-01-17
Add UI to allow modifying organization parents


jon-betts:
Our current flat structure for organizations is not enough to represent the complexity of our customers. We have a built in parent id, but there is no way to change it.

The organization page in the admin interface should allow you to:

* Move an organization from one parent to another
* Remove a parent from an organization

This should be marked as a dangerous action.

## Tasks

 * https://github.com/hypothesis/lms/pull/4866

jon-betts:
This is done now!

PR #4866, opened 2023-01-17
Allow editing of organization parents in the admin pages

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4860

### Adding a valid parent

 * Visit: http://localhost:8001/admin/org/3
 * The "Move (or remove) parent organization" in the danger zone should be empty
 * Add "us.lms.org.MUfTxdQOQDqdWlqHM2cVFw"
 * Press "Move"
 * You should be on the org page, and the parent value is "us.lms.org.MUfTxdQOQDqdWlqHM2cVFw"

### Removing a parent

 * Do the test above
 * Clear the value in the box
 * Press "Move"
 * You should be on the org page, and the parent value is blank

### Various types of invalid entries

Try the following bad inputs:

 * `us.lms.org.FqoiV2OyQ9y-gdWjGHx0xQ` - The orgs own public id
 * Delete or modify any parts that make a valid public id like "org" etc - Should get an invalid id message
 * `us.lms.org.not-a-real-org` - Should get a not found error

In all cases the previous parent should remain


lms/services/organization.py
```diff
@@ -146,8 +169,73 @@ def update_organization(self, organization, name=None, enabled=None, notes=None)
...
+        # CTE
+        cols = [Organization.id, Organization.parent_id]
+        base_case = (
+            self._db_session.query(*cols).filter(Organization.id == id_)
+            # The name of the CTE is arbitrary, but must be present
+            .cte("organizations", recursive=True)
+        )
+
+        # We are going to self join onto the above with anything that is a
+        # child of the objects we've only seen
```
marcospri, line 225: Is this sentence missing something at the end?

lms/services/organization.py
```diff
@@ -146,8 +169,73 @@ def update_organization(self, organization, name=None, enabled=None, notes=None)
...
+        recursive_case = self._db_session.query(*cols).join(base_case, join_condition)
+
+        # This will recurse until no new rows are added
+        rows = self._db_session.query(base_case.union(recursive_case)).all()
+
+        # Convert the Row objects into ORM objects with another query
+        return (
+            self._db_session.query(Organization)
+            .filter(Organization.id.in_([row[0] for row in rows]))
+            .all()
```
marcospri, line 236: We could use this query to force some kind of order. 

by parent_id? That might simplify `get_hierarchy_root` later

lms/services/organization.py
```diff
@@ -146,8 +169,73 @@ def update_organization(self, organization, name=None, enabled=None, notes=None)
...
+            public_id=parent_public_id
+        ).one_or_none()
+        if not parent:
+            raise InvalidOrganizationParent(
+                f"Could not find parent organization: '{parent_public_id}'"
+            )
+
+        organization.parent = parent
+
+    def _get_hierarchy(self, id_) -> List[Organization]:
```
marcospri, line 209: We are not in the "save a query" mindset at all in for the admin pages but if we split this in two `_get_hierarchy_pk` and `_get_hierarchy` we could do a simpler `parent_public_id in _get_hierarchy_pk()` in `_move_organization_parent`

marcospri (approved):

lms/services/organization.py
```diff
@@ -146,8 +169,73 @@ def update_organization(self, organization, name=None, enabled=None, notes=None)
...
+            .cte("organizations", recursive=True)
+        )
+
+        # We are going to self join onto the above with anything that is a
+        # child of the objects we've only seen
+        join_condition = Organization.parent_id == base_case.c.id
+        recursive_case = self._db_session.query(*cols).join(base_case, join_condition)
+
+        # This will recurse until no new rows are added
+        rows = self._db_session.query(base_case.union(recursive_case)).all()
```
marcospri, line 230: It might be the type of comment you get asked to remove (by me?) in another review but a

```
this generates a query like:

WITH RECURSIVE organizations(id, parent_id) AS 
 (SELECT organization.id AS id, organization.parent_id AS parent_id 
 FROM organization 
 WHERE organization.id = %(id_1)s 

UNION SELECT organization.id AS organization_id, organization.parent_id AS organization_parent_id 
 FROM organization JOIN organizations ON organization.parent_id = organizations.id OR organization.id = organizations.parent_id)
  SELECT organizations.id AS organizations_id, organizations.parent_id AS organizations_parent_id 
 FROM organizations
```

or maybe that will be just noise?

lms/services/organization.py
jon-betts: I don't think there's a sensible order without having calculated the "depth" from the root. The parent won't necessarily have an id which came before it's children, as these hierarchies will be cobbled together adhoc by people.

lms/services/organization.py
jon-betts: Yes, this makes sense. We don't need to query here for the actual items. We need that in the next PR but we can add it on.

lms/services/organization.py
jon-betts: I might point to an article about recursive queries in general, but I'm a bit allergic to queries which need a lot of maintenance to be correct. If the query changed, and we didn't update it, I'm worried it would be misleading.

lms/services/organization.py
jon-betts: Yeah, I was going to go on to let you know about

lms/services/organization.py
jon-betts: Hang on I think see what you mean, if we sort by `parent_id` those without one should come first and be the root?

jon-betts: This actually doesn't make any difference when returning ids as we use an `in` query which does not respect order. I will use this in the next PR though.

marcospri (approved):

Issue #4861, opened 2023-01-17
Add a view of the organization hierarchy in the admin pages


jon-betts:
By now we should be able to allocate organizations as parents of other organizations. 

In order to help visualise this layout, when viewing a specific organization you should be able to see:

 * The whole hierarchy it is a part of
 * Where it is in that hierarchy
 * Each entry should include the name and organization id

This doesn't need to be anything fancy, it could be:

 * A nested unordered list
 * It might be nice if you could click through to go to the org

## Tasks

 * https://github.com/hypothesis/lms/pull/4867
 * https://github.com/hypothesis/lms/pull/4874

PR #4867, opened 2023-01-17
Display organization hierarchy in the admin pages

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4861
 
Requires: 

 * https://github.com/hypothesis/lms/pull/4866

## Testing notes

 * Create a hierarchy of orgs
 * Look at them
 * You should be able to navigate between and view the full hierarchy
 * The organization you are on is not navigable and should be highlighted

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -19,15 +19,45 @@ class OrganizationService:
     def __init__(self, db_session: Session):
         self._db_session = db_session
 
-    def get_by_id(self, id_) -> Optional[Organization]:
-        """Get an organization by its private primary key."""
+    def get_by_id(self, id_: int) -> Optional[Organization]:
+        """
+        Get an organization by its private primary key.
+
+        :param id_: Primary key of the organization
+        """
 
         return self._organization_search_query(id_=id_).one_or_none()
 
     def get_by_public_id(self, public_id: str) -> Optional[List]:
-        """Get an organization by its public_id."""
+        """
+        Get an organization by its public_id.
+
+        :param public_id: Fully qualified public id
+        """
 
         return self._organization_search_query(public_id=public_id).one_or_none()
+
+    def get_hierarchy_root(self, id_: int) -> Organization:
+        """
+        Get the root of the hierarchy from the given organization.
+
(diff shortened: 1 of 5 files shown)


lms/templates/admin/organization.html.jinja2
```diff
@@ -1,5 +1,24 @@
...
+            <b>(You are here)</b> {{ node.name }} - <code>{{ node.public_id }}</code>
+        {% else %}
+            <a href="{{ request.route_url('admin.organization', id_=node.id) }}">{{ node.name }}</a>
+            - <code>{{ node.public_id }}</code>
+        {% endif %}
+
+        {% if node.children %}
+            <ul>
+            {% for child in node.children %}
+                {{ org_hierarchy(child, focus_org) }}
```
marcospri, line 15: Recursive jinja2 macro, nice :+1:

lms/services/organization.py
```diff
@@ -29,6 +29,16 @@ def get_by_public_id(self, public_id: str) -> Optional[List]:
 
         return self._organization_search_query(public_id=public_id).one_or_none()
 
+    def get_hierarchy_root(self, id_) -> Organization:
```
marcospri, line 32: Maybe the type annotation here for `id_` will be helpful to distinguish between the pk and public id (although we usually hint that also with the name, but the more the merrier)

lms/templates/admin/organization.html.jinja2
```diff
@@ -1,5 +1,24 @@
 {% import "macros.html.jinja2" as macros %}
 
+{% macro org_hierarchy(node, focus_org) %}
+    <li>
+        {% if node.id == focus_org.id %}
+            <b>(You are here)</b> {{ node.name }} - <code>{{ node.public_id }}</code>
+        {% else %}
+            <a href="{{ request.route_url('admin.organization', id_=node.id) }}">{{ node.name }}</a>
```
marcospri, line 8: We display the `node.name` for "you are here" case but not the others. It might be more valuable the other way around.

marcospri (approved):

Issue #4862, opened 2023-01-17
Show the root status of an organization on the search page


jon-betts:
When searching for organizations you should be able to filter by root level elements or not (those with no parents).

_Question:_ Can we think of a time this would be useful? Generally you do not browse organizations in this interface, you search for a specific thing. Can we get an example where this would help?

 * For now we are going to show the status
 * Filtering can wait

## Tasks

 * https://github.com/hypothesis/lms/pull/4868

PR #4868, opened 2023-01-17
Add an indicator to show if an organization is top level or not

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4862
 
Requires:

 * https://github.com/hypothesis/lms/pull/4866

## Implementation notes

It's not super clear what use this information is and whether we need this, don't need it, or only care about filtering. For now seeing the status seems like a reasonable thing to do. Hopefully we'll get some clarity soon.

## Testing notes

* It's not strictly required for this PR to based on #4866, but it makes it much easier to set up some parents and children.
* Do this
* Search for them in the interface, and you should see a checkbox indicating whether they are root or not


lms/templates/admin/organizations.html.jinja2
```diff
@@ -35,27 +35,35 @@ Organizations
...
-                       </tr>
-                    {% endfor %}
-                  </tbody>
+                          <th></th>
+                          <th>Public ID</th>
+                          <th>ID</th>
+                          <th>Name</th>
+                          <th>Root</th>
+                        </tr>
+                    </thead>
```
jon-betts, line 46: The indentation in this file was mixed up. There was both 2 and 4 space. I've moved it to 4 space as my IDE was doing my head in.

lms/templates/admin/organizations.html.jinja2
```diff
@@ -35,27 +35,35 @@ Organizations
...
+                    <tbody>
+                        {% for org in organizations %}
+                            <tr>
+                                <td>
+                                    <a class="button" href="{{ request.route_url('admin.organization', id_=org.id) }}">View</a>
+                                </td>
+                                <td>{{org.public_id}}</td>
+                                <td>{{org.id}}</td>
+                                <td>{{org.name}}</td>
+                                <td>
```
marcospri, line 57: We could link/display the parent instead,  we still have horizontal space here.

marcospri (approved):
We could add  a "parent" column on top/instead of this one but this works nicely in any case.

Issue #4863, opened 2023-01-17
Make application instance creation an admin page (not user facing)


jon-betts:

jon-betts:
Done

PR #4869, opened 2023-01-17
Refactor the new application instance to allow the creation of LTI 1.1

jon-betts:
For:

This also disentangles the registration from the application instance and makes it optional.

## Implementation notes

 * `/admin/instance/new/registration` has become `/admin/instance/new`
    * Some of the methods have moved over to the application instance view as a result
 * We now read all of the values from params instead of any path params
 * This makes is possible to optionally accept a registration id (which enables 1.3 mode)
 * If it's missing we use a different schema and only expect 1.1 values
 * As the `POST` methods all return redirects we don't need `error_render_to_response` any more

This mixes in a refactor for a `_redirect()` which is probably better as a separate PR / commit.

## Testing notes

The LTI 1.3 path should not be effected and should keep working:

 * Visit: http://localhost:8001/admin/registration/id/2/
 * Click "Add instance"
 * This page should work as before, specifically check that:
   * Deployment ID is required
   * You can save a filled out application instance
   * You can upgrade an exising LTI1.1 (try "Hypothesis37c8a8ac186ad5d7e6da83c060c3b18b")

The new LTI 1.1 path should also work:

 * Visit: http://localhost:8001/admin/instances/
 * Click: "New LTI 1.1 application instance"
 * Notice the following are not visible:
   * Upgrade an instnace
   * Deployment Id
   * Should we hide the "No LTI registration"?
 * Try saving without filling anything out, notice that "Deployment id" is not required
 * Try filling out the required fields, it should work and redirect you to view the application instance

### DIFF
--- a/lms/templates/admin/instance.html.jinja2
+++ b/lms/templates/admin/instance.html.jinja2
@@ -49,6 +49,7 @@
     <fieldset class="box">
         {{ macros.settings_textarea(instance, "Notes", "hypothesis", "notes") }}
         {{ macros.disabled_text_field("Consumer key", instance.consumer_key) }}
+        {{ macros.disabled_text_field("Shared secret", instance.shared_secret) }}
         {{ macros.form_text_field(request, "Deployment ID", "deployment_id", field_value=instance.deployment_id) }}
         {{ macros.form_text_field(request, "LMS URL", "lms_url", field_value=instance.lms_url) }}
     </fieldset>
--- a/lms/templates/admin/instance.new.html.jinja2
+++ b/lms/templates/admin/instance.new.html.jinja2
@@ -2,25 +2,29 @@
 
 {% extends "lms:templates/admin/base.html.jinja2" %}
 
-
 {% block header %}
 New application instance
 {% endblock %}
 
 {% block content %}
 
-<form method="POST" action="{{ request.route_url("admin.registration.new.instance", id_=lti_registration.id) }}">
-    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+{% if lti_registration %}
     <fieldset class="box mt-6">
         <legend class="label has-text-centered">Registration</legend>
 
         {{ macros.registration_preview(request, lti_registration) }}
     </fieldset>
+{% endif %}
 
+<form method="POST" action="{{ request.route_url("admin.instance.new") }}">
+    <input type="hidden" name="csrf_token" value="{{ get_csrf_token() }}">
+    <input type="hidden" name="lti_registration_id" value="{{ lti_registration.id }}">
 
     <fieldset class="box mt-6">
         <legend class="label has-text-centered">Create new Application instance</legend>
-        {{ macros.form_text_field(request, "Deployment ID", "deployment_id") }}
+        {% if lti_registration %}
+            {{ macros.form_text_field(request, "Deployment ID", "deployment_id") }}
+        {% endif %}
         {{ macros.form_text_field(request, "LMS URL", "lms_url") }}
         {{ macros.form_text_field(request, "Email", "email") }}
         {{ macros.form_text_field(request, "Canvas developer key", "developer_key") }}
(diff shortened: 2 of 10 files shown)


lms/templates/admin/instance.new.html.jinja2
```diff
@@ -30,17 +32,19 @@ New application instance
     <input type="submit" class="button is-info" value="New"/>
 </form>
 
-<form method="POST" action="{{ request.route_url("admin.registration.upgrade.instance", id_=lti_registration.id) }}">
-    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
+{% if lti_registration %}
```
jon-betts, line 35: The change here is mostly a result of indentation changes from the `if` here

lms/templates/admin/instance.new.html.jinja2
```diff
@@ -2,25 +2,27 @@
...
+    <legend class="label has-text-centered">Registration</legend>
 
-        {{ macros.registration_preview(request, lti_registration) }}
-    </fieldset>
+    {{ macros.registration_preview(request, lti_registration) }}
+</fieldset>
 
+<form method="POST" action="{{ request.route_url("admin.instance.new") }}">
+    <input type="hidden" name="csrf_token" value="{{ get_csrf_token() }}">
+    <input type="hidden" name="lti_registration_id" value="{{ lti_registration.id }}">
```
jon-betts, line 19: We now persist the `lti_registration_id` as a param. This make it possible for it to be optional, and also makes it really easy to pass the params back to ourselves, by passing them all.

lms/templates/admin/instance.new.html.jinja2
```diff
@@ -2,25 +2,27 @@
...
-    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
-    <fieldset class="box mt-6">
-        <legend class="label has-text-centered">Registration</legend>
+<fieldset class="box mt-6">
+    <legend class="label has-text-centered">Registration</legend>
 
-        {{ macros.registration_preview(request, lti_registration) }}
-    </fieldset>
+    {{ macros.registration_preview(request, lti_registration) }}
+</fieldset>
```
jon-betts, line 15: I've moved this above the form, because it's not part of it, and it more closely maps to the visual layout.

lms/templates/admin/instances.html.jinja2
```diff
@@ -7,6 +7,12 @@ Application instances
 
 {% block content %}
 
+<div class="block has-text-right">
+    <a class="button is-primary" href="{{request.route_url("admin.instance.new")}}">
+        New LTI 1.1 application instance
+    </a>
+</div>
```
jon-betts, line 14: Is this OTT or is it good to be explicit? The 1.3 journey from the moment is from the registration first. Maybe one day we can have a nicer single journey?

marcospri: Makes sense to be explicit, this is a bit confusing, AI, registrations, LTI versions and orgs...

Alternatively we could just take `deployment_id` in a single journey, maybe simple to understand at the expense to have to copy the deployment_id around. 

Probably something we can leave for later, sprinkle some JS and have a nice "find and select" component for getting the deployment.

marcospri (approved):

lms/templates/admin/instance.new.html.jinja2
```diff
@@ -2,25 +2,27 @@
...
 {% endblock %}
 
 {% block content %}
 
-<form method="POST" action="{{ request.route_url("admin.registration.new.instance", id_=lti_registration.id) }}">
-    <input type="hidden" name="csrf_token" value="{{get_csrf_token()}}">
-    <fieldset class="box mt-6">
-        <legend class="label has-text-centered">Registration</legend>
+<fieldset class="box mt-6">
+    <legend class="label has-text-centered">Registration</legend>
```
marcospri, line 12: Yeah probably an `if lti_registration` around this could help

lms/templates/admin/instances.html.jinja2
marcospri: Maybe even the button for LTI 1.3 that now says "Add instance" could change to say "Add LTI 1.3 application instance".

It clear after doing it a few times but text is cheap.

jon-betts: I was thinking we might want to split the "upgrade an instance" and "create a new instance" and have them as different buttons.

lms/templates/admin/instance.html.jinja2
```diff
@@ -49,6 +49,7 @@
     <fieldset class="box">
         {{ macros.settings_textarea(instance, "Notes", "hypothesis", "notes") }}
         {{ macros.disabled_text_field("Consumer key", instance.consumer_key) }}
+        {{ macros.disabled_text_field("Shared secret", instance.shared_secret) }}
```
jon-betts, line 52: Without this the journey doesn't make a huge amount as you can't complete the LTI 1.1 journey

marcospri (approved):

PR #4872, opened 2023-01-19
Don't request unnecessary scopes for D2L

marcospri:
I added this initially as it was part of D2L examples in the documentation I thought it was a requirement but it's not needed at all.

---

`core:*` scopes don't seem to be a requirement for any of the endpoints we use.

See the full list at:

* https://docs.valence.desire2learn.com/http-scopestable.html

## Testing

- Remove existing tokens:
```shell
tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token"
```
- Remove assignments to allow re-configuring:
```shell
 tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment cascade"
```
- Configure: https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View
with either files or groups or both. 
- It will succeed.

### DIFF
--- a/lms/views/api/d2l/authorize.py
+++ b/lms/views/api/d2l/authorize.py
@@ -17,18 +17,21 @@ FILES_SCOPES = ("content:toc:read", "content:topics:read")
     permission=Permissions.API,
 )
 def authorize(request):
-    application_instance = request.find_service(
-        name="application_instance"
-    ).get_current()
-    state = OAuthCallbackSchema(request).state_param()
-
-    scopes = ["core:*:*"]
+    scopes = []
 
     if request.product.settings.files_enabled:
         scopes += FILES_SCOPES
 
     if request.product.settings.groups_enabled:
         scopes += GROUPS_SCOPES
+
+    if not scopes:
+        raise NotImplementedError("Trying to get a D2L without any scopes")
+
+    application_instance = request.find_service(
+        name="application_instance"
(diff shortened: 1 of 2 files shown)


lms/views/api/d2l/authorize.py
```diff
@@ -17,19 +17,22 @@
...
-    scopes = ["core:*:*"]
+    scopes = []
 
     if request.product.settings.files_enabled:
         scopes += FILES_SCOPES
 
     if request.product.settings.groups_enabled:
         scopes += GROUPS_SCOPES
 
+    if not scopes:
```
marcospri, line 28: We don't want to request a token with no scopes, if we are doing that  something went wrong.

lms/views/api/d2l/authorize.py
```diff
@@ -17,19 +17,22 @@
     permission=Permissions.API,
 )
 def authorize(request):
-    application_instance = request.find_service(
```
marcospri, line 20: Moving this below to exit early when no scopes are required.

jon-betts (approved):
I've tried this with a regular PDF and groups and it works. I've also checked we can launch the D2L file picker, but at the moment there are no files you can pick as a result of missing recursion support: https://github.com/hypothesis/lms/pull/4853

So with those caveats, it looks good.

PR #4874, opened 2023-01-19
Fix crashes in the organization hierarchy view caused by null names

jon-betts:
In Python 3 None has no natural sort, so the built in sort operation in Jinja2 isn't very useful over objects with attributes which can be None. To fix that we inject our own sorting method here.

## The issue

In order to trigger this issue you need two organizations at the same level, where one has a string name and another has a null name. You cannot set a null name through the interface (it interprets an empty string as "don't set a name").

You can see a failing example in live here:

  * https://lms.hypothes.is/admin/org/378

## Testing notes

To set this up:

 * Pick an organization to be the parent and copy the public id
 * Pick two other orgs and set them to be children of that org
 * Pick the id of one of the child orgs
 * `UPDATE organization SET name = NULL where id=<ID>`
 * On `main`: This should fail with an exception like:
```python
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```
* In this branch it should sort ahead of other items in the list

### DIFF
--- a/lms/templates/admin/organization.html.jinja2
+++ b/lms/templates/admin/organization.html.jinja2
@@ -11,7 +11,7 @@
 
         {% if node.children %}
             <ul>
-            {% for child in node.children | sort(attribute='name') %}
+            {% for child in sort_by_name(node.children) %}
                 {{ org_hierarchy(child, focus_org) }}
             {% endfor %}
             </ul>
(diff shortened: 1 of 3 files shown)


lms/templates/admin/organization.html.jinja2
```diff
@@ -11,7 +11,7 @@
 
         {% if node.children %}
             <ul>
-            {% for child in node.children | sort(attribute='name') %}
```
marcospri, line 14: Oh for sure can use `key: lambda` directly here, nope...

marcospri (approved):

PR #4879, opened 2023-01-20
Adjust some LMSFilePicker tests to await on components rendering when appropriate

acelaya:
This does not have user-facing or behavior changes, it just refactors a bit the `LMSFilePicker` tests so that they await on the "proper thing" (so to speak).

Some of those tests were "accidentally" passing just because there was an `await`, forcing the next tick, and consequently making some children components to be there.

However, as part of the changes in  #4853, this is not enough anymore.

Since the same pattern is present in a handful of tests, I have extracted the refactoring to a separated PR, to avoid polluting #4853, making it harder to review.

I have locally rebased that branch and it now works with these changes.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
+++ b/lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
@@ -184,16 +184,10 @@ describe('LMSFilePicker', () => {
 
     // Wait for the initial file fetch to fail.
     const wrapper = renderFilePicker();
-    try {
-      await fakeApiCall;
-    } catch {
-      /* unused */
-    }
-    wrapper.update();
     assert.called(fakeApiCall);
 
     // Check that the "Authorize" button is shown.
-    const authButton = wrapper.find('AuthButton');
+    const authButton = await waitForElement(wrapper, 'AuthButton');
     assert.isTrue(authButton.exists());
 
     // Click the "Authorize" button and check that files are re-fetched.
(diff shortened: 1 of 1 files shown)


lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -184,16 +184,11 @@ describe('LMSFilePicker', () => {
 
     // Wait for the initial file fetch to fail.
     const wrapper = renderFilePicker();
-    try {
-      await fakeApiCall;
-    } catch {
-      /* unused */
-    }
     wrapper.update();
```
robertknight, line 187: This `wrapper.update()` should now be redundant, assuming `renderFilePicker` returns a "current" wrapper at the end, which is what I would expect as a reader.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
```diff
@@ -249,7 +238,6 @@ describe('LMSFilePicker', () => {
     // Make successful authorization attempt and wait for the auth window to close.
     fakeApiCall.reset();
     fakeApiCall.resolves([0]);
-    await fakeApiCall;
     wrapper.update();
```
robertknight, line 241: This `wrapper.update()` should also be redundant.

robertknight (approved):
Thanks for this! There are a couple of remaining `wrapper.update()` calls that I believe should be redundant. Otherwise looks good.

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
acelaya: Fixed!

lms/static/scripts/frontend_apps/components/test/LMSFilePicker-test.js
acelaya: Fixed!

Issue #4877, opened 2023-01-20
SPIKE: How does course copy work currently on Canvas


marcospri:
### How the course copy files fix works in canvas

- The moment when an instructor configures an assignment with Canvas files, we store all the files metadata (for every file in the course) in our DB.
- Students might have access to the `list_files` API endpoint or not. One way to simulate this is to limit visibility of  the files section in Course -> Settings -> Navigation.

The process is then:

- An instructor configures the original assignment to use canvas files. In that moment we store all the files in our DB. The assignment points to file ID `ORIGINAL` 

- The course gets copied, the new assignment still points to `ORIGINAL`.

#### If an instructor launches the new assignment first:

1. Check if we have a mapping stored that points `ORIGINAL` to another file

2. We check that the file `ORIGINAL` belongs to the new course

3. That check fails

4. We query all files from the new course (and store them in the DB)

5. For each file in the new course we check if we find a match in the DB, if we do we point `ORIGINAL` -> `NEW` and store that in the DB.

6. If we don't find a match we show the Error dialog about not having access to the file.

#### Any subsequent launches of the same assignment, by any user:

1. Check if we have a mapping stored that points `ORIGINAL` to another file
2. Return the mapped file, `NEW`

#### If a **student** launches an assignment before the teacher and the student **has access to list the files** in the new course:

Same process as for the teacher, except we don't check that the file `ORIGINAL` belongs to the new course.

#### If a **student** launches an assignment before the teacher and the student **doesn't have access to list the files** in the new course:

We show an error dialog

seanh:
@marcospri This is super helpful. I think we can close the issue now as the spike is completed? The closed issue will still be around for reference

seanh:
Do we only store file metadata about the course when an instructor _configures_ an assignment? Or we also do it when they launch an assignment?

When do we _update_ our file metadata about a course? Since feels can be added, updated, renamed, deleted in Canvas

marcospri:
> Do we only store file metadata about the course when an instructor configures an assignment? Or we also do it when they launch an assignment?

Only when configuring.


> When do we update our file metadata about a course? 

Just when configuring, we don't handle deletions, just new and updated files.

mattdricker:
What happens if there is more that one file match in the new course?

[...]

PR #4880, opened 2023-01-23
Files course copy fix for Blackboard

marcospri:
This PR adds the `CourseCopyPlugin`  implementation for Blackboard.

After a better understanding of #4877 I reckon this is very similar but better approach. 

### As a teacher

Login into BB as `blackboardteacher`, `make devdata`

- Launch the original `localhost (make devdata) Blackboard Files Assignment`

It should work as before

- Launch the original `localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted`

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_80_1&course_id=_19_1

It should display the corresponding error dialog.

- Before heading to the copied assignments, reset any state in the DB courses `tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate grouping cascade;"`

- Launch the copied `localhost (make devdata) Blackboard Files Assignment`

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1371_1&course_id=_139_1

It should be successful. Now the DB should have the mapping between files:

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select extra from grouping where type ='course';"                        
                         extra                          
--------------------------------------------------------
 {"course_copy_file_mappings": {"_7851_1": "_57000_1"}}
(1 row)
```

- Launch the copied `localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted`

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1379_1&course_id=_139_1

You should get the error dialog.

### As a student

Login into BB as `blackboardstudent`

- Launch the copied `localhost (make devdata) Blackboard Files Assignment`

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1371_1&course_id=_139_1

All good

- Launch the copied `localhost (make devdata) Blackboard Files Assignment Whose File Has Been Deleted`

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_80_1&course_id=_19_1

It should display the corresponding error dialog.

### DIFF
--- a/lms/product/blackboard/_plugin/course_copy.py
+++ b/lms/product/blackboard/_plugin/course_copy.py
@@ -1,0 +1,43 @@
+from lms.product.plugin.course_copy import CourseCopyFilesHelper
+from lms.services.file import FileService
+
+
+class BlackboardCourseCopyPlugin:
+    """Handle course copy for Blackboard."""
+
+    file_type = "blackboard_file"
+
+    def __init__(
+        self, api, file_service: FileService, files_helper: CourseCopyFilesHelper
+    ):
+        self._api = api
+        self._file_service = file_service
+        self._files_helper = files_helper
+
+    def is_file_in_course(self, course_id, file_id):
+        return self._files_helper.is_file_in_course(
+            self._file_service, course_id, file_id, self.file_type
+        )
+
+    def find_matching_file_in_course(self, original_file_id, new_course_id):
+        return self._files_helper.find_matching_file_in_course(
+            self._api.list_all_files,
+            self._file_service,
+            self.file_type,
+            original_file_id,
+            new_course_id,
+        )
+
+    def get_mapped_file_id(self, course, file_id):
+        return self._files_helper.get_mapped_file_id(course, file_id)
+
+    def set_mapped_file_id(self, course, old_file_id, new_file_id):
+        self._files_helper.set_mapped_file_id(course, old_file_id, new_file_id)
+
+    @classmethod
+    def factory(cls, _context, request):
+        return cls(
+            request.find_service(name="blackboard_api_client"),
+            file_service=request.find_service(name="file"),
+            files_helper=CourseCopyFilesHelper(),
+        )
--- a/lms/product/blackboard/product.py
+++ b/lms/product/blackboard/product.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 
+from lms.product.blackboard._plugin.course_copy import BlackboardCourseCopyPlugin
 from lms.product.blackboard._plugin.grouping import BlackboardGroupingPlugin
 from lms.product.product import PluginConfig, Product, Routes
 
@@ -16,5 +17,7 @@ class Blackboard(Product):
         list_group_sets="blackboard_api.courses.group_sets.list",
     )
 
-    plugin_config: PluginConfig = PluginConfig(grouping=BlackboardGroupingPlugin)
+    plugin_config: PluginConfig = PluginConfig(
+        grouping=BlackboardGroupingPlugin, course_copy=BlackboardCourseCopyPlugin
+    )
     settings_key = "blackboard"
--- a/lms/product/plugin/course_copy.py
+++ b/lms/product/plugin/course_copy.py
@@ -58,8 +58,12 @@ class CourseCopyFilesHelper:
 
     @staticmethod
     def get_mapped_file_id(course, file_id):
-        """Get a previously mapped file id in course."""
-        return course.extra.get("course_copy_file_mappings", {}).get(file_id, None)
+        """
+        Get a previously mapped file id in course.
+
+        Returns the original `file_id` if no mapped one can be found.
+        """
+        return course.extra.get("course_copy_file_mappings", {}).get(file_id, file_id)
 
     @staticmethod
     def set_mapped_file_id(course, old_file_id, new_file_id):
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -5,6 +5,7 @@ from pyramid.view import view_config, view_defaults
 
 from lms.product.blackboard import Blackboard
 from lms.security import Permissions
+from lms.services.exceptions import BlackboardFileNotFoundInCourse
 from lms.views import helpers
 
 #: A regex for parsing just the file_id part out of one of our custom
@@ -19,6 +20,7 @@ class BlackboardFilesAPIViews:
     def __init__(self, request):
         self.request = request
         self.blackboard_api_client = request.find_service(name="blackboard_api_client")
+        self.course_copy_plugin = request.product.plugin.course_copy
 
     @view_config(request_method="GET", route_name="blackboard_api.courses.files.list")
     @view_config(
@@ -66,11 +68,35 @@ class BlackboardFilesAPIViews:
         """Return the Via URL for annotating the given Blackboard file."""
 
         course_id = self.request.matchdict["course_id"]
-        document_url = self.request.params["document_url"]
-        file_id = DOCUMENT_URL_REGEX.search(document_url)["file_id"]
+        course = self.request.find_service(name="course").get_by_context_id(course_id)
 
-        public_url = self.blackboard_api_client.public_url(course_id, file_id)
+        document_url = self.request.params["document_url"]
+        file_id = self.course_copy_plugin.get_mapped_file_id(
+            course, DOCUMENT_URL_REGEX.search(document_url)["file_id"]
+        )
+        try:
+            if self.request.lti_user.is_instructor:
+                if not self.course_copy_plugin.is_file_in_course(course_id, file_id):
+                    raise BlackboardFileNotFoundInCourse(file_id)
+
+            public_url = self.blackboard_api_client.public_url(course_id, file_id)
+
+        except BlackboardFileNotFoundInCourse:
+            found_file = self.course_copy_plugin.find_matching_file_in_course(
+                file_id, course_id
+            )
+            if not found_file:
+                raise
+
+            # Try again to return a public URL, this time using found_file_id.
+            public_url = self.blackboard_api_client.public_url(
+                course_id, found_file.lms_id
+            )
+
+            # Store a mapping so we don't have to re-search next time.
+            self.course_copy_plugin.set_mapped_file_id(
+                course, file_id, found_file.lms_id
+            )
 
         via_url = helpers.via_url(self.request, public_url, content_type="pdf")
-
         return {"via_url": via_url}
--- a/tests/unit/lms/product/blackboard/_plugin/course_copy_test.py
+++ b/tests/unit/lms/product/blackboard/_plugin/course_copy_test.py
@@ -1,0 +1,69 @@
+from unittest.mock import create_autospec, sentinel
+
+import pytest
+
+from lms.product.blackboard import BlackboardCourseCopyPlugin
+from lms.product.plugin.course_copy import CourseCopyFilesHelper
+
+
+class TestBlackboardCourseCopyPlugin:
+    def test_is_file_in_course(self, plugin, helper, file_service):
+
+        result = plugin.is_file_in_course(sentinel.course_id, sentinel.file_id)
+
+        helper.is_file_in_course.assert_called_once_with(
+            file_service, sentinel.course_id, sentinel.file_id, "blackboard_file"
+        )
+
+        assert result == helper.is_file_in_course.return_value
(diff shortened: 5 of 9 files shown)


lms/product/blackboard/_plugin/course_copy.py
```diff
@@ -0,0 +1,25 @@
+from lms.product.plugin.course_copy import CourseCopyPlugin
+
+
+class BlackboardCourseCopyPlugin(CourseCopyPlugin):
```
marcospri, line 4: Blackboard plugin just override the BB specific bits, getting the files from the API.

lms/product/blackboard/product.py
```diff
@@ -16,5 +17,7 @@ class Blackboard(Product):
         list_group_sets="blackboard_api.courses.group_sets.list",
     )
 
-    plugin_config: PluginConfig = PluginConfig(grouping=BlackboardGroupingPlugin)
+    plugin_config: PluginConfig = PluginConfig(
```
marcospri, line 20: Sets the blackboard plugin in the blackboard product.

lms/product/plugin/__init__.py
```diff
@@ -8,3 +9,4 @@ def includeme(config):  # pragma: nocover
 
     config.register_service(GroupingPlugin(), iface=GroupingPlugin)
     config.register_service(MiscPlugin(), iface=MiscPlugin)
+    config.register_service(CourseCopyPlugin(), iface=CourseCopyPlugin)
```
marcospri, line 12: Register the base plugin

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,74 @@
+from lms.models import File
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CourseCopyPlugin:
```
marcospri, line 5: Base CourseCopy plugin

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,74 @@
...
+    def find_matching_file_in_course(self, original_file_id, new_course_id):
+        try:
+            # Get the current (copied) courses files, that will have the side effect of storing files in the DB
+            _ = self._store_new_course_files(new_course_id)
+        except ExternalRequestError:
+            # We might not have access to use the API for that endpoint.
+            # That will depend on our role and the course's permissions settings.
+            # We will continue anyway, maybe the files of the new course are already in the DB
+            # after an instructor launched corrected the issue.
+            pass
```
marcospri, line 21: This is an improvement over the Canvas solution, store the files in the DB and use those to do the course copy matching.

That means that any launch after the "storing" takes place can in theory fix course copy itself based on the info in the DB.

In canvas we rely on the API call to do so (even if we also store he files in the DB).

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,74 @@
...
+        file = self._file_service.get(original_file_id, type_=self.file_type)
+        if not file:
+            # That file must have been recorded when the original assignment was configured.
+            # If we can't find that one something odd is going on, stop here.
+            return
+
+        # Now we'll try to find a matching file in the DB in the new course
+        # We might have a record of this because we just called `_store_new_course_files` as the current user
+        # or another user might have done it before for us.
+        new_file = (
```
marcospri, line 33: This query won't be just pasted here in the final version, it will over over `file_service`

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # No match for the file.
+        # This could be an issue with our heuristic to find the new file
+        # or other edge case, for example a file was deleted after course copy or similar.
+        return None
+
+    @staticmethod
+    def get_mapped_file_id(course, file_id):
+        """Get a previously mapped file id in course."""
+        return course.extra.get("course_copy_file_mappings", {}).get(file_id, file_id)
```
marcospri, line 63: Another change over the canvas solution, we store the mapping in the course not the assignment.

All of this course copy business is a problem because files belong to the course, we should map them the same way in our model.

This also has a theoretical benefit of allowing reusing the same file mapping for multiple assignments that use the same file.

lms/product/plugin/plugin.py
```diff
@@ -10,6 +11,7 @@ class PluginConfig:
 
     # These also provide the default implementations
     grouping: type = GroupingPlugin
+    course_copy: type = CourseCopyPlugin
```
marcospri, line 14: Some bolilerplate to setup a new product plugin.

lms/services/blackboard_api/client.py
```diff
@@ -88,10 +78,74 @@ def list_files(self, course_id, folder_id=None):
...
+                    break
+                response = self._api.request("GET", path)
+                folder_results.extend(BlackboardListFilesSchema(response).parse())
+
+            # For each folder, get it children too
+            results.extend(self._get_all_folders_files(course_id, folder_results))
+
+        return results
+
+    def list_all_files(self, course_id):
```
marcospri, line 131: This has a lot of copy pasted code but the problem is trying to solve is:


- We need all the files in the new course to be able to fix course copy.
- In BB we only get files as the user navigates over the folders.


We'd use this method only during course copy to get the list of files of the new course.

We could try to use the stored info for the original file to only query the folder structure of that tree. 

We'll save some API queries but:

- That code will be more complex, we need to match the tree structure of both courses
- It will need to be executed once per assignment, as opposed to once per course of this version.


Some light checking of the info we have in the DB suggest that the worst offenders are folders with many pages not very deep folder structures:

https://hypothes-is.slack.com/archives/C04JN7PMG9X/p1674234224864519

lms/services/blackboard_api/client.py
```diff
@@ -88,10 +78,74 @@ def list_files(self, course_id, folder_id=None):
...
+        # Get this level all pages
+        path = self._list_files_url(course_id)
+        for _ in range(PAGINATION_MAX_REQUESTS):
+            response = self._api.request("GET", path)
+            results.extend(BlackboardListFilesSchema(response).parse())
+            path = response.json().get("paging", {}).get("nextPage")
+            if not path:
+                break
+
+        results += self._get_all_folders_files(course_id, results)
```
marcospri, line 143: Disregard this code in particular but I reckon our options are:

- Naively get all pages and children sequentially
- Get all pages sequentially but all children at once with async
- Get all children at once and all the "next pages" we can for each of them using async.

lms/views/api/blackboard/files.py
```diff
@@ -66,11 +67,30 @@ def via_url(self):
         """Return the Via URL for annotating the given Blackboard file."""
 
         course_id = self.request.matchdict["course_id"]
+        course = self.request.find_service(name="course").get_by_context_id(course_id)
```
marcospri, line 70: Now on the view use the plugin to handle course copy.

This code will be very similar in BB/Canvas and D2L, we  can find a common abstraction later. Maybe even a common `list_files` view for all of them.

I rather tackle that after the course copy issues are solved.

lms/product/plugin/course_copy.py
marcospri: Plugins are services that are implemented once per product.

I like the idea of a CourseCopy oriented one because it give more visibility to that problem/functionality. Otherwise is just a implantation detail in each of the product's APIs.

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,74 @@
+from lms.models import File
+from lms.services.exceptions import CanvasFileNotFoundInCourse
+
+
+class CourseCopyPlugin:
+    file_type = None
+
+    def assert_file_in_course(self, course_id, file_id):
+        """Raise if the current user can't see file_id in course_id."""
```
seanh, line 9: What should this raise? `AssertionError`?

seanh (commented):
I think the "course copy plugin" pattern. It seems like a nice way to group all the course copy-related code together.

I'm not sure whether there really is a generic "course copy plugin" here, rather than just a specific "Blackboard course copy plugin"? Since it's only the Blackboard API views that actually call the course copy plugin.

I think it'd be nice to avoid inheritance-based code design, for the usual reasons:

* I can't understand the `BlackboardCourseCopyPlugin` class without understanding the whole implementation of the `CourseCopyPlugin` class as well (what methods and attributes it defines; exactly when it calls `_store_new_course_files()`, etc)
* `CourseCopyPlugin` will be tightly coupled to all its subclasses. For example it calls a private method of `BlackboardCourseCopyPlugin`
* You can't unit test `BlackboardCourseCopyPlugin` in isolation from `CourseCopyPlugin` (you can't easily mock out the base class)

Rather than an abstract `CourseCopyPlugin` base class, a typical composition-over-inheritance design might have a `CourseCopyHelper` class for different `FooCourseCopyPlugin` classes to use, something like this:

```python
class BlackboardCourseCopyPlugin:
    def __init__(self, helper, ...):
        self._helper = helper
        ...

    def assert_file_in_course(self, *args, **kwargs):
        return self._helper.assert_file_in_course(*args, **kwargs)
        
    def find_matching_file_in_course(self, *args, **kwargs):
        return self._helper.find_matching_file_in_course(self._store_new_course_files, *args, **kwargs)

    def get_mapped_file_id(self, *args, **kwargs):
        return self._helper(*args, **kwargs)
```

As is always the case with composition-over-inheritance this is a little more wordy: every `FooCourseCopyPlugin` class has to define each of the methods `assert_file_in_course()`, `find_matching_file_in_course()`, etc even though they mostly just forward calls to `CourseCopyHelper`. But composition has more important advantages:

* I can understand the whole interface of the `BlackboardCourseCopyPlugin` class just by reading that class
* The coupling between `BlackboardCourseCopyPlugin` and `CourseCopyHelper` is loose, they integrate only on `CourseCopyHelper`'s well-defined interface
* You can easily unit test `BlackboardCourseCopyPlugin` and `CourseCopyHelper` in isolation from each other
* It's more flexible

It would be typical in Python to leave the interface that the different `FooCourseCopyPlugin` classes must provide just informal (`assert_file_in_course()`, `find_matching_file_in_course()`, etc). But if you wanted to explicitly define the interface I think you'd use [`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol).

lms/product/blackboard/_plugin/course_copy.py
```diff
@@ -0,0 +1,43 @@
+from lms.product.plugin.course_copy import CourseCopyFilesHelper
+from lms.services.file import FileService
+
+
+class BlackboardCourseCopyPlugin:
```
seanh, line 5: Note that if `CourseCopyPlugin` were an `ABC` or a `Protocol` then `BlackboardCourseCopyPlugin` could optionally inherit from it just to explicitly document that it implements that interface (not to inherit any code)

lms/product/plugin/course_copy.py
seanh: I think there was a reason why these mappings were stored in assignments rather than courses, but I can't remember.

tests/unit/lms/product/blackboard/_plugin/course_copy_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+from lms.product.blackboard import BlackboardCourseCopyPlugin
+from lms.product.plugin.course_copy import CourseCopyFilesHelper
+
+
+class TestBlackboardCourseCopyPlugin:
+    def test_is_file_in_course(self, plugin, helper, file_service):
+
+        result = plugin.is_file_in_course(sentinel.course_id, sentinel.file_id)
+
```
seanh, line 13:
```suggestion
```

lms/views/api/blackboard/files.py
```diff
@@ -66,11 +68,35 @@ def via_url(self):
...
+
+            # Try again to return a public URL, this time using found_file_id.
+            public_url = self.blackboard_api_client.public_url(
+                course_id, found_file.lms_id
+            )
+
+            # Store a mapping so we don't have to re-search next time.
+            self.course_copy_plugin.set_mapped_file_id(
+                course, file_id, found_file.lms_id
+            )
```
seanh, line 99: :+1:

seanh (approved):

PR #4885, opened 2023-01-23
Get existing course instead of trying to recreate it

marcospri:
While working on course copy I found that we never persist the extra portion of course other than the one we set on creation.

Arguably this is a bug that we are not yet hitting but worth fixing in preparation for course copy and not mixed with the rest of the changes.

# Testing

- Remove courses

`tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate grouping cascade;"`

- Check the course's extra

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "select extra from grouping where type = 'course'" 
                     extra                                                           
------------------------------------------------                                     
 {"canvas": {"custom_canvas_course_id": "125"}}                                      
(1 row)                   
```

- Update the course extra directly on the DB:

make sql
```
postgres=# update grouping set extra=jsonb_set(extra, '{other}', '"value"') where type = 'course';
UPDATE 1
postgres=# select extra from grouping where type = 'course';
                              extra                               
------------------------------------------------------------------
 {"other": "value", "canvas": {"custom_canvas_course_id": "125"}}
(1 row)
```

- Launch the assignment again

On `main` extra gets reset back to the original, in this branch the changes are preserved.

### DIFF
--- a/lms/resources/lti_launch.py
+++ b/lms/resources/lti_launch.py
@@ -29,11 +29,20 @@ class LTILaunchResource:
     @cached_property
     def course(self):
         """Get the course this LTI launch based on the request's params."""
+        course_service = self._request.find_service(name="course")
 
-        return self._request.find_service(name="course").upsert_course(
+        if existing_course := course_service.get_by_context_id(
+            self._request.parsed_params["context_id"]
+        ):
+            # Keep existing `extra` instead of replacing it with the default
+            extra = existing_course.extra
+        else:
+            extra = self._new_course_extra()
+
+        return course_service.upsert_course(
             context_id=self._request.parsed_params["context_id"],
             name=self._request.parsed_params["context_title"],
-            extra=self._course_extra(),
+            extra=extra,
         )
 
     @property
@@ -60,8 +69,8 @@ class LTILaunchResource:
             self._request, self.course, assignment
         )
 
-    def _course_extra(self):
-        """Extra information to store for courses."""
+    def _new_course_extra(self):
+        """Extra information to store for new courses."""
         extra = {}
 
         if self.is_canvas:
--- a/tests/unit/lms/resources/lti_launch_test.py
+++ b/tests/unit/lms/resources/lti_launch_test.py
@@ -1,4 +1,4 @@
-from unittest.mock import patch, sentinel
+from unittest.mock import sentinel
 
 import pytest
 
@@ -13,15 +13,13 @@ pytestmark = pytest.mark.usefixtures(
 )
 
 
-class TestApplicationInstance:
-    def test_it(self, lti_launch, application_instance_service):
+class TestLTILaunchResource:
+    def test_application_instance(self, lti_launch, application_instance_service):
         assert (
             lti_launch.application_instance
             == application_instance_service.get_current.return_value
         )
 
-
-class TestIsCanvas:
     @pytest.mark.parametrize(
         "product,expected",
         [
@@ -30,14 +28,12 @@ class TestIsCanvas:
             (Product.Family.UNKNOWN, False),
         ],
     )
-    def test_it(self, pyramid_request, product, expected):
+    def test_is_canvas(self, pyramid_request, product, expected):
         pyramid_request.product.family = product
 
         assert LTILaunchResource(pyramid_request).is_canvas == expected
 
-
-class TestJSConfig:
-    def test_it_returns_the_js_config(self, pyramid_request, JSConfig):
+    def test_js_config(self, pyramid_request, JSConfig):
         lti_launch = LTILaunchResource(pyramid_request)
 
         js_config = lti_launch.js_config
@@ -45,33 +41,7 @@ class TestJSConfig:
         JSConfig.assert_called_once_with(lti_launch, pyramid_request)
         assert js_config == JSConfig.return_value
 
-
-class TestNewCourseExtra:
-    # pylint: disable=protected-access
-    def test_empty_in_non_canvas(self, pyramid_request):
-        parsed_params = {}
-        pyramid_request.parsed_params = parsed_params
-
-        assert not LTILaunchResource(pyramid_request)._new_course_extra()
-
-    @pytest.mark.usefixtures("with_canvas")
-    def test_includes_course_id(self, pyramid_request):
-        parsed_params = {
-            "custom_canvas_course_id": "ID",
-        }
-        pyramid_request.parsed_params = parsed_params
-
-        assert LTILaunchResource(pyramid_request)._new_course_extra() == {
-            "canvas": {"custom_canvas_course_id": "ID"}
-        }
-
-    @pytest.fixture
-    def with_canvas(self, pyramid_request):
-        pyramid_request.product.family = Product.Family.CANVAS
-
-
-class TestCourse:
-    def test_it_when_existing(self, course_service, lti_launch):
+    def test_course_when_existing(self, course_service, lti_launch):
         course_service.get_by_context_id.return_value = factories.Course(
             extra={"existing": "extra"}
         )
@@ -88,7 +58,19 @@ class TestCourse:
         )
         assert course == course_service.upsert_course.return_value
 
-    def test_it_when_new(self, course_service, lti_launch, _new_course_extra):
+    def test_course_when_new(self, course_service, lti_launch):
+        course_service.get_by_context_id.return_value = None
+
+        course = lti_launch.course
+
+        course_service.get_by_context_id.assert_called_once_with(sentinel.context_id)
+        course_service.upsert_course.assert_called_once_with(
+            context_id=sentinel.context_id, name=sentinel.context_title, extra={}
+        )
+        assert course == course_service.upsert_course.return_value
+
+    @pytest.mark.usefixtures("with_canvas")
+    def test_course_when_new_and_canvas(self, course_service, lti_launch):
         course_service.get_by_context_id.return_value = None
 
         course = lti_launch.course
@@ -97,24 +79,12 @@ class TestCourse:
         course_service.upsert_course.assert_called_once_with(
             context_id=sentinel.context_id,
             name=sentinel.context_title,
-            extra=_new_course_extra.return_value,
+            extra={"canvas": {"custom_canvas_course_id": "ID"}},
         )
         assert course == course_service.upsert_course.return_value
 
-    @pytest.fixture
-    def _new_course_extra(self, lti_launch):
-        with patch.object(lti_launch, "_new_course_extra", autospec=True) as patched:
-            yield patched
-
-
-class TestGroupingType:
-    def test_it(
-        self,
-        grouping_service,
-        lti_launch,
-        assignment_service,
-        pyramid_request,
-        course,
+    def test_grouping_type(
+        self, grouping_service, lti_launch, assignment_service, pyramid_request
     ):
         assert (
             lti_launch.grouping_type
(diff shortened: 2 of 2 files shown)


lms/resources/lti_launch.py
```diff
@@ -29,11 +29,15 @@ def __init__(self, request):
     @cached_property
     def course(self):
         """Get the course this LTI launch based on the request's params."""
+        if existing_course := self._request.find_service(
+            name="course"
+        ).get_by_context_id(self._request.parsed_params["context_id"]):
+            return existing_course
```
marcospri, line 35: Get the course if we find it by ID

lms/resources/lti_launch.py
```diff
@@ -29,11 +29,15 @@ def __init__(self, request):
...
 
         return self._request.find_service(name="course").upsert_course(
```
marcospri, line 37: If that fails, create it with the initial extra.

This call here could be to an hypothetical `create` but `upsert` is the method we have.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -68,28 +69,61 @@ def with_canvas(self, pyramid_request):
         pyramid_request.product.family = Product.Family.CANVAS
 
 
+class TestCourse:
```
marcospri, line 72: This was not explicitly tested

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -68,28 +69,61 @@ def with_canvas(self, pyramid_request):
...
         )
         grouping_service.get_launch_grouping_type.assert_called_once_with(
             pyramid_request,
-            course_service.upsert_course.return_value,
+            course,
             assignment_service.get_assignment.return_value,
         )
 
+    @pytest.fixture
+    def course(self, lti_launch):
```
marcospri, line 123: I reckon despite having this `patch` here this is a more clear test.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -68,28 +70,70 @@ def with_canvas(self, pyramid_request):
         pyramid_request.product.family = Product.Family.CANVAS
 
 
+class TestCourse:
+    def test_it_when_existing(self, course_service, lti_launch):
+        course_service.get_by_context_id.return_value = factories.Course(
+            extra={"existing": "extra"}
```
marcospri, line 76: `sentinel` upsets SQLA here

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -44,13 +46,13 @@ def test_it_returns_the_js_config(self, pyramid_request, JSConfig):
         assert js_config == JSConfig.return_value
 
 
-class TestCourseExtra:
+class TestNewCourseExtra:
```
jon-betts, line 49: I'm not a big fan of the split up style of testing (a class per method) as it tends to string the test file out more. It's "good" when tests get very long, but I'd say that's a good time to think about refactoring the code under test. I don't think we're at that level with this code, but I suspect we used to be, which is why it's this way (we've removed a lot of this file over time).

I realise it's more of a refactor, but I'm not sure this is complicated enough to make testing it using a patch over the private method is worth it? I think it might be easier to grok (and refactor) if the testing was all done in `TestCourse`.

This is as you found it, but arguably this should always have been `TestCourse` rather than `TestCourseExtra`?

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -68,28 +70,70 @@ def with_canvas(self, pyramid_request):
...
             == grouping_service.get_launch_grouping_type.return_value
         )
         assignment_service.get_assignment.assert_called_once_with(
-            mock.sentinel.tool_guid, mock.sentinel.resource_link_id
+            sentinel.tool_guid, sentinel.resource_link_id
         )
         grouping_service.get_launch_grouping_type.assert_called_once_with(
             pyramid_request,
-            course_service.upsert_course.return_value,
+            course,
```
jon-betts, line 128:
```suggestion
            lti_launch.course,
```

Alternatively we could use `lti_launch.course` and let it work out as it will? This is effectively the same command and would very closely follow the layout in the code without having to use a separate fixture.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -104,9 +148,9 @@ def JSConfig(patch):
...
 def pyramid_request(pyramid_request):
     pyramid_request.parsed_params = pyramid_request.lti_params = {
-        "tool_consumer_instance_guid": mock.sentinel.tool_guid,
-        "resource_link_id": mock.sentinel.resource_link_id,
-        "context_id": mock.sentinel.context_id,
-        "context_title": mock.sentinel.context_title,
+        "tool_consumer_instance_guid": sentinel.tool_guid,
+        "resource_link_id": sentinel.resource_link_id,
+        "context_id": sentinel.context_id,
+        "context_title": sentinel.context_title,
```
jon-betts, line 154: :+1: It's ok to import things. Let's treat ourselves.

Seriously though, we have enough eye clutter in general, changes like this lower the cognitive load to read the file and make this easier to maintain. We use `sentinel` everywhere, it's not ambiguous what we mean.

lms/resources/lti_launch.py
```diff
@@ -30,10 +30,17 @@ def __init__(self, request):
...
+            name="course"
+        ).get_by_context_id(self._request.parsed_params["context_id"]):
+            # Keep existing `extra` instead of replacing it with the default
+            extra = existing_course.extra
+
         return self._request.find_service(name="course").upsert_course(
             context_id=self._request.parsed_params["context_id"],
             name=self._request.parsed_params["context_title"],
-            extra=self._course_extra(),
+            extra=extra,
```
jon-betts, line 43: We are using `self._request.find_service(name="course")` twice here, we should probably keep a reference.

On the other hand, this is the only time we ever call this method in the codebase. If we need to make another call to the service immediately before, this strongly suggests that this should get pushed down into the service.

lms/resources/lti_launch.py
```diff
@@ -30,10 +30,17 @@ def __init__(self, request):
     def course(self):
         """Get the course this LTI launch based on the request's params."""
 
+        extra = self._new_course_extra()
+        if existing_course := self._request.find_service(
+            name="course"
+        ).get_by_context_id(self._request.parsed_params["context_id"]):
+            # Keep existing `extra` instead of replacing it with the default
+            extra = existing_course.extra
```
jon-betts, line 38:
```suggestion
        if existing_course := self._request.find_service(
            name="course"
        ).get_by_context_id(self._request.parsed_params["context_id"]):
            # Keep existing `extra` instead of replacing it with the default
            extra = existing_course.extra
        else:
            extra = self._new_course_extra()
```

Might just be my brain, but it's nice to not have to follow off into the code unless the effect takes place. This is true for computation too (but nothing very expensive happens in there). It's more of a "go read this, not realise it's getting thrown away" kind of thing.

jon-betts (commented):
There's nothing wrong with the refactor here and it solves a real issue (it's also an object lesson in the difference between line coverage and behavior coverage), but I think there are some missed opportunities to make the code here a bit nicer still.

These are all "coulds" rather than "musts".

 * Either:
    * Keep a reference to the course service
    * Or... consider pushing the behavior down into the service
  * Test `_new_course_extra()` as a part of `course()` instead of a separate unit
  * Some suggestions for extra tidies here and there
  * I'd be interested to see a single `TestLTILaunchResource` test class

tests/unit/lms/resources/lti_launch_test.py
jon-betts: Random observation, this is specific to the `TestGroupingType` test and should probably be inlined there. Try commenting it out and running the tests again.

lms/resources/lti_launch.py
marcospri: > this strongly suggests that this should get pushed down into the service.

Yeah, that would be much nicer, but `new_course_extra` needs a `is_canvas`. I have another refactor going on to disentangle  LTIParams and Product and that will go nicely with this.

It would also move more stuff out of `resource/lit_launch.py` which an odd spot in our code base now.


I'll leave as is now so it's more obvious what need to be done later. Made a note of this :+1:

tests/unit/lms/resources/lti_launch_test.py
marcospri: Moving to one class with multiple methods was pretty trivial to change, done that.

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -28,85 +28,96 @@ class TestIsCanvas:
...
+            sentinel.context_id,
+        )
+        course_service.upsert_course.assert_called_once_with(
+            context_id=sentinel.context_id,
+            name=sentinel.context_title,
+            extra={"existing": "extra"},
+        )
+        assert course == course_service.upsert_course.return_value
 
-    @pytest.mark.usefixtures("with_canvas")
```
marcospri, line 55: Part of testing `course` now

lms/resources/lti_launch.py
```diff
@@ -29,11 +29,20 @@ def __init__(self, request):
...
+        course_service = self._request.find_service(name="course")
 
-        return self._request.find_service(name="course").upsert_course(
+        if existing_course := course_service.get_by_context_id(
+            self._request.parsed_params["context_id"]
+        ):
+            # Keep existing `extra` instead of replacing it with the default
+            extra = existing_course.extra
+        else:
+            extra = self._new_course_extra()
```
marcospri, line 40: `new_course_extra` could be inlined here but reckon with this is more obvious which part can be "productized"

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -1,25 +1,25 @@
...
 
 from lms.product import Product
 from lms.resources import LTILaunchResource
+from tests import factories
 
 pytestmark = pytest.mark.usefixtures(
     "application_instance_service",
     "assignment_service",
+    "course_service",
 )
```
jon-betts, line 15:
```suggestion
@pytest.mark.usefixtures(
    "application_instance_service", "assignment_service", "course_service"
)
```

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -28,85 +28,96 @@ class TestIsCanvas:
...
+    def test_is_canvas(self, pyramid_request, product, expected):
         pyramid_request.product.family = product
 
         assert LTILaunchResource(pyramid_request).is_canvas == expected
 
-
-class TestJSConfig:
-    def test_it_returns_the_js_config(self, pyramid_request, JSConfig):
+    def test_js_config(self, pyramid_request, JSConfig):
         lti_launch = LTILaunchResource(pyramid_request)
```
jon-betts, line 38:
```suggestion
    def test_js_config(self, lti_launch, pyramid_request, JSConfig):
```

tests/unit/lms/resources/lti_launch_test.py
```diff
@@ -28,85 +28,96 @@ class TestIsCanvas:
             (Product.Family.UNKNOWN, False),
         ],
     )
-    def test_it(self, pyramid_request, product, expected):
+    def test_is_canvas(self, pyramid_request, product, expected):
         pyramid_request.product.family = product
 
         assert LTILaunchResource(pyramid_request).is_canvas == expected
```
jon-betts, line 34:
```suggestion
    def test_is_canvas(self, lti_launch, pyramid_request, product, expected):
        pyramid_request.product.family = product

        assert lti_launch.is_canvas == expected
```

jon-betts (approved):
I think this is looking good. I made a couple of extra suggestions to re-use an existing fixture and to tack the `usefixtures` onto the class as a decorator.

PR #4886, opened 2023-01-23
Make organization a required field in admin pages application instance create

jon-betts:
For:

## Implementation notes

This PR:

 * Moves to kwargs for calls to `create_application_instance` to make it easier to change them
 * Fixes some oddities in the test for `create_application_instance`
 * Makes organization id mandatory on create
 * Adds a path to jump from org to the creation of a 1.1 application instance (but not 1.3)

## Testing notes

 * Visit an org: http://localhost:8001/admin/org/1
 * Press "Add New LTI 1.1 Instance"
 * Notice the organization id is passed through
 * Delete the id
 * Try to save
 * Notice you get an error about organization id
 * Fill out all the required fields, but put a mangled value in organization id
 * Try to save
 * Notice you get a malformed public id error
 * Add the original public id back again from  http://localhost:8001/admin/org/1
 * Save
 * It should work and redirect to the application instance page
 * Notice the linked organization is the correct one

### DIFF
--- a/lms/views/admin/application_instance.py
+++ b/lms/views/admin/application_instance.py
@@ -5,6 +5,7 @@ from sqlalchemy.exc import IntegrityError
 from webargs import fields
 
 from lms.models import ApplicationInstance
+from lms.models.public_id import InvalidPublicId
 from lms.security import Permissions
 from lms.services import ApplicationInstanceNotFound, LTIRegistrationService
 from lms.services.aes import AESService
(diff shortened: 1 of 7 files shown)


lms/views/admin/application_instance.py
```diff
@@ -23,6 +24,7 @@ class NewAppInstanceSchema(PyramidRequestSchema):
 
     lms_url = fields.URL(required=True)
     email = fields.Email(required=True)
+    organization_public_id = fields.Str(required=True, validate=validate.Length(min=1))
```
jon-betts, line 27: I investigated making a custom validator / field type based on `PublicId`, but it's a bit of a non-starter as we need the region passed in based on the `request` which means we need to create the schema on the fly.

We could do that, but also we could just catch the error we will get anyway.

The advantage of looking before we leap is the user is presented with all the errors they need to fix in one go, rather than having to work through them one by one, as they do when it's a result of exceptions being raised.

PR #4887, opened 2023-01-23
Move settings defaults for application instances into the service

jon-betts:
For:

Requires:

  * https://github.com/hypothesis/lms/pull/4886

Our internal and external application instance creation pages apply different settings as defaults. This moves those settings inside the service and ensures both will set the same value.

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -230,7 +230,6 @@ class ApplicationInstanceService:
         developer_key,
         developer_secret,
         organization_public_id=None,  # We want to make this mandatory
-        settings=None,
         deployment_id=None,
         lti_registration_id=None,
     ):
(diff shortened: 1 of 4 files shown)


lms/services/application_instance.py
```diff
@@ -245,7 +244,13 @@ def create_application_instance(  # pylint:disable=too-many-arguments
             lms_url=lms_url,
             requesters_email=email,
             created=datetime.utcnow(),
-            settings=settings or {},
+            # Some helpful defaults for settings
+            settings={
```
marcospri, line 248: This is nicer than the alternative but I guess at some point we'd want to be more clever on the UI side.

For example, you could start an AI creation wizard where you explicitly choose the product family and the get asked for the relevant settings there (for D2L the client id/secret, based on which you can enable things like we do here for canvas).

Not for today.

marcospri (approved):

lms/services/application_instance.py
jon-betts: I was thinking about "productification" and I couldn't see a way right now. One idea I had was every product gets to set it's defaults. As long as they never clash it's not actually a problem, and would pull this out into the separate products.

If we could pick a product up front that would work too.

PR #4876, opened 2023-01-19
Remove public sign-up page

jon-betts:
For:

  * https://github.com/hypothesis/lms/issues/4864

Requires:

 * https://github.com/hypothesis/lms/pull/4887

This removes the public `/welcome` page and replaces it with the same blank page as `/`. When we come to release this we may want to release this and the redirect at the same time:

 * https://github.com/hypothesis/lms/pull/4888

Issue #4864, opened 2023-01-17
Make the sign-up page private and redirect to a form


jon-betts:
We need to remove the publicly available sign-up page for application instances. This is to:

 * Ensure people who want to sign-up end up in our sales funnel
 * To ensure assignments are attached to organizations correctly
 * Cover some issues which arise from the details not being verified by us

## Tasks

* [x] https://github.com/hypothesis/lms/pull/4876
* [x] https://github.com/hypothesis/lms/pull/4888
   * The page is: https://web.hypothes.is/help/installing-the-hypothesis-lms-app/
* [x] https://github.com/hypothesis/lms/pull/4892
   * It is in the UI, but it can be in the service once the public view is gone
* [x] https://github.com/hypothesis/lms/pull/4890
   * Existing application instances in the wild which have no organizations and get launched still fall outside of our ability control
   * We can at least flag them so someone can take action

jon-betts:
I think this is blocked on a discussion with Joe about policy here?

jon-betts:
We have the go ahead to deploy

PR #4888, opened 2023-01-23
Redirect users who visit us directly to help pages

jon-betts:
For:

Requires:

 * https://github.com/hypothesis/lms/pull/4876

We are removing the welcome page, but curious users might still visit. We help them better by redirecting them to some relevant help.

## Testing notes

 * Visit: http://localhost:8001/
 * Visit: http://localhost:8001/welcome
 * Both should redirect you to: https://web.hypothes.is/help/installing-the-hypothesis-lms-app/

### DIFF
--- a/lms/views/index.py
+++ b/lms/views/index.py
@@ -1,8 +1,12 @@
+from pyramid.httpexceptions import HTTPFound
 from pyramid.view import view_config
 
 
-@view_config(route_name="welcome", renderer="lms:templates/base.html.jinja2")
-@view_config(route_name="index", renderer="lms:templates/base.html.jinja2")
-def index(request):  # pylint: disable=unused-argument
-    """Render an empty page that contains a needed Google verification meta tag."""
-    return {}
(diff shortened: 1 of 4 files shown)


lms/views/index.py
```diff
@@ -1,8 +1,12 @@
...
-    """Render an empty page that contains a needed Google verification meta tag."""
-    return {}
+@view_config(route_name="welcome")
+@view_config(route_name="index")
+def index(_):
+    """Redirect curious users to some help."""
+
+    return HTTPFound(
+        location="https://web.hypothes.is/help/installing-the-hypothesis-lms-app/"
+    )
```
jon-betts, line 12: Hard coding this seems like the easiest thing to do. I don't think it's going to change, and if it does it's dirt simple to change.

If we find we are changing it every week we would think about adding an environment variable.

marcospri: This link also includes the Canadian version so we don't have to worry about two different links.

PR #4890, opened 2023-01-23
Flag auto assigned application instances

jon-betts:
For:

This does two things:

 * Point out when an application instance was automatically added to an organization
 * Point out when an organization was created for this process

This allows us to vet them at a later date if we want to. Currently it has no effect, but if we capture the information we can chose to act on it.

### DIFF
--- a/lms/services/organization.py
+++ b/lms/services/organization.py
@@ -133,28 +133,38 @@ class OrganizationService:
 
         if application_instance.organization:
             org = application_instance.organization
-
-        elif (
-            orgs := self._organization_search_query(guid=guid)
-            .order_by(Organization.updated.desc())
-            .all()
-        ):
-            if len(orgs) > 1:
-                LOG.warning(
-                    "Multiple organization matches found for application instance %s",
-                    application_instance.id,
-                )
-
-            org = orgs[0]
-
         else:
-            org = self.create_organization()
+            # Add a note to indicate the application instance was automatically
+            # allocated to an organization
+            application_instance.settings.set(
+                "hypothesis", "auto_assigned_to_org", True
+            )
+
+            if (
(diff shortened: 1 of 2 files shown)


lms/services/organization.py
```diff
@@ -136,28 +136,36 @@ def auto_assign_organization(
...
-                )
-
-            org = orgs[0]
-
         else:
-            org = self.create_organization()
+            # Add a note to indicate the application instance was automatically
+            # allocated to an organization
+            application_instance.settings.set("flag", "auto_assigned_to_org", True)
+
```
jon-betts, line 143: The change here is we've fully indented this whole chunk. This is because we want the "else" here to note the assignment. Previously we didn't care where the org came from.

lms/services/organization.py
```diff
@@ -136,28 +136,36 @@ def auto_assign_organization(
...
-                    application_instance.id,
-                )
-
-            org = orgs[0]
-
         else:
-            org = self.create_organization()
+            # Add a note to indicate the application instance was automatically
+            # allocated to an organization
+            application_instance.settings.set("flag", "auto_assigned_to_org", True)
```
jon-betts, line 142: Not super pretty, but it's easy and doesn't involve a migration. Open to suggestions for a better name if you have one?

marcospri: We have notes in "hypothesis" -> "notes"

Not sure if `hypothesis" -> "auto_assigned_to_org" is better :shrug:

marcospri (approved):
Nice, quite curious now to know how often this happens :+1:

lms/services/organization.py
jon-betts: Yep, if we have a place already we might as well re-use it

PR #4891, opened 2023-01-23
Tidy up the comment formatting in the application instance model

jon-betts:
This isn't really about anything, I just spotted it when working on something else and it annoyed me.

### DIFF
--- a/lms/models/application_instance.py
+++ b/lms/models/application_instance.py
@@ -17,15 +17,18 @@ class ApplicationInstance(BASE):
 
     __tablename__ = "application_instances"
     __table_args__ = (
-        # For LTI1.3 instances we allow consumer_key to be null as long as we have a registration and deployment_id.
-        # Note that when consumer_key is present we don't require lti_registration_id and deployment_id to be null
-        # it could be an instance that has been upgraded from LTI1.1 to LTI1.3 having values for all three fields.
+        # For LTI1.3 instances we allow consumer_key to be null as long as we
+        # have a registration and deployment_id. Note that when consumer_key is
+        # present we don't require lti_registration_id and deployment_id to be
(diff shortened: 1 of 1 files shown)


lms/models/application_instance.py
```diff
@@ -61,58 +64,58 @@ class ApplicationInstance(BASE):
         nullable=False,
     )
 
-    #: A unique identifier for the LMS instance.
+    # A unique identifier for the LMS instance
```
jon-betts, line 67: It was these guys. What is the `#:` all about?

lms/models/application_instance.py
```diff
@@ -61,58 +64,58 @@ class ApplicationInstance(BASE):
...
-    #: This Canvas custom variable substitution $Canvas.api.domain.
-    #: We request this in our config.xml file and name it "custom_canvas_api_domain":
-    #:
-    #: https://github.com/hypothesis/lms/blob/5394cf2bfb92cb219e177f3c0a7991add024f242/lms/templates/config.xml.jinja2#L20
-    #:
-    #: See https://canvas.instructure.com/doc/api/file.tools_variable_substitutions.html
+    # This Canvas custom variable substitution $Canvas.api.domain. We request
+    # this in our config.xml file and name it "custom_canvas_api_domain":
+    #
+    # https://github.com/hypothesis/lms/blob/5394cf2b/lms/templates/config.xml.jinja2#L20
```
marcospri, line 91: The comment was guarding against a collision

marcospri (approved):

lms/models/application_instance.py
jon-betts: I'm not entirely sure if you are joking or not, but with 8 chars we should have a collision chance of 1 in 4 billion or so, so we're probably fine?

I can put it back?

PR #4892, opened 2023-01-23
Make organization id a required param on application instance creation

jon-betts:
For:

Requires:

 * https://github.com/hypothesis/lms/pull/4886
 * https://github.com/hypothesis/lms/pull/4887
 * https://github.com/hypothesis/lms/pull/4876

## Review notes

This has a pile of merging and cherry picking and will need rebasing when the other PRs are all merged through. The only relevant commit to review is the last one.

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -242,12 +242,12 @@ class ApplicationInstanceService:
         email,
         developer_key,
         developer_secret,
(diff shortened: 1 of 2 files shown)


marcospri (approved):
Looks like the only place where this is used is the admin pages now and in both LTI version of that we require the org id in the schema :+1:

PR #4893, opened 2023-01-24
Split FileService.get into query and get methods.

marcospri:
In the context of course copy will need a more complex query (that will be different potentially across different LMS's)

https://github.com/hypothesis/lms/pull/4880/files#diff-ceb59d3a160f3c225cc276ebb1b2d1c2620e50bc476182059ffc9c8251848526R33

---

This PR prepares the design for that, splitting a new `_search_query` that can be then used by the service.

### DIFF
--- a/lms/services/file.py
+++ b/lms/services/file.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from sqlalchemy import func
 
 from lms.models import File
@@ -9,17 +11,13 @@ class FileService:
         self._application_instance = application_instance
         self._db = db
 
-    def get(self, lms_id, type_):
-        """Return the file with the given lms_id and type_ or None."""
-        return (
-            self._db.query(File)
-            .filter_by(
-                application_instance=self._application_instance,
-                lms_id=lms_id,
-                type=type_,
-            )
-            .one_or_none()
-        )
+    def get(self, lms_id, type_) -> Optional[File]:
+        """Return the file with the given lms_id and type_."""
+        return self._search_query(
+            application_instance=self._application_instance,
+            lms_id=lms_id,
+            type_=type_,
+        ).one_or_none()
 
     def upsert(self, file_dicts):
         """Insert or update a batch of files."""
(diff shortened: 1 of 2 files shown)


lms/services/file.py
```diff
@@ -35,6 +33,25 @@ def upsert(self, file_dicts):
             update_columns=["name", "size", "updated"],
         )
 
+    def _search_query(self, application_instance=None, lms_id=None, type_=None):
```
marcospri, line 36: The query over https://github.com/hypothesis/lms/pull/4880/files#diff-ceb59d3a160f3c225cc276ebb1b2d1c2620e50bc476182059ffc9c8251848526R33


has a `lms_id != id` type of query.

Not sure how to fit that into this design:

- Take a `lms_id_not_equal` parameter and build the query.
- Take a `*args` of the form `File.whatever == value` and push them directly into `filter` here.
- Build the query without that and add it outside the service:
```
file = filer_service.search(some=value, and=another).filter(File.lms_id != id).first()
```


```

File: tests/unit/lms/services/file_test.py [hunk 2]
```diff
@@ -76,6 +76,24 @@ def test_upsert(self, db_session, svc, application_instance):
...
 
+    @pytest.mark.parametrize(
+        "param,field",
+        (
+            ("application_instance", "application_instance"),
+            ("lms_id", "lms_id"),
+            ("type_", "type"),
+        ),
+    )
+    def test__search_query(self, svc, param, field):
```
marcospri, line 87: `_search_query` is interesting enough to get its own test

lms/services/file.py
jon-betts: I like this search method  / accessor pattern in general. So as far as adding it here, I'm in favour.

As for including the negative case, I think my instincts are:

 * I don't like splitting the query over the service boundary - this seems like what the service is for
 * Can we avoid the question? Is this something which is generalisable enough that the entire query moves into the service, instead of calling a complicated search?
* If not I'd probably prefer taking *args, or a `filter` arg because:
   * Testing this becomes harder as there is no boundary you can mock easily
   * Encapsulation around the service is broken
   * This is somewhat unavoidable, as you probably want to blackbox test the outer service anyway, as the filter might be passed as you expect, but not do what you want
   * ^ This is the reason I'd prefer skipping it if possible and keeping everything inside the service
   
So in answer to the question, I'd prefer to be a coward and not answer it by having the whole thing inside the service, instead of the plugin if possible.

marcospri:
Closing this now and revisiting it after course copy is finished.

Issue #4889, opened 2023-01-23
Add names to application instances


jon-betts:
Currently application instances have no name. This will make it hard to tell one from another when designing customer layouts involving application instances.

We should:

 * [x] Add a nullable "name" field to application instances
    * [x] https://github.com/hypothesis/lms/pull/4912
    * [x] https://github.com/hypothesis/lms/pull/4914
    * [x] https://github.com/hypothesis/lms/pull/4913
 * [x] Make the name searchable in the admin pages like the organization name
    * [x] Show this field in preference to other more LMS (tool consumer) specific fields in the search page
    * [x] https://github.com/hypothesis/lms/pull/4917
 * ~Force users to add a name when creating or updating an application instance (both in the admin pages and the service)~
   * Split off into: https://github.com/hypothesis/lms/issues/4945

## Notes

There's nothing much we can do about existing application instances, so the vast majority will have no name.

jon-betts:
I'm going to split off the mandatory part from this ticket to let the rest move over.

PR #4912, opened 2023-01-24
Refactor application instance service and admin view tests

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4889

This is largely around reducing the amount of conditionality we have in the tests.

There are a couple of places where our effective tests are to call with all arguments full and then again with all of them empty. But we do this in every binary combination of the tests. As we add more and more arguments this leads to an explosion of tests runs.

My hope is this will make adding name a bit easier and not double the number of tests run in the relevant places.

### DIFF
--- a/tests/unit/lms/views/admin/application_instance_test.py
+++ b/tests/unit/lms/views/admin/application_instance_test.py
@@ -492,19 +492,13 @@ class TestAdminApplicationInstanceViews:
 
     @pytest.fixture
     def with_upgrade_form(self, pyramid_request, application_instance):
-        application_instance_data = {
+        pyramid_request.POST = pyramid_request.params = {
             "lms_url": "http://lms-url.com",
             "email": "test@email.com",
             "deployment_id": "DEPLOYMENT_ID",
+            "consumer_key": application_instance.consumer_key,
         }
-        # Real pyramid request have the same params available
-        # via POST and params
-        pyramid_request.POST.update(application_instance_data)
-        pyramid_request.params.update(application_instance_data)
 
-        pyramid_request.params["consumer_key"] = pyramid_request.POST[
-            "consumer_key"
-        ] = application_instance.consumer_key
         return pyramid_request
 
     @pytest.fixture


tests/unit/lms/services/application_instance_test.py
```diff
@@ -92,44 +92,39 @@ def test_get_by_deployment_id_raises_on_missing(
...
-            lms_url=lms_url,
-            deployment_id=deployment_id,
-            developer_key=developer_key,
-            developer_secret=developer_secret,
+            lms_url="http://example.com",
+            deployment_id="DEPLOYMENT_ID",
+            developer_key="DEVELOPER_KEY",
+            developer_secret="DEVELOPER_SECRET",
+            organization_public_id=mock.sentinel.org_id,
         )
```
jon-betts, line 108: Test all of this in one go, instead of in a binary combination of every possibility.

tests/unit/lms/services/application_instance_test.py
```diff
@@ -92,44 +92,39 @@ def test_get_by_deployment_id_raises_on_missing(
...
-            assert application_instance.lms_url == lms_url
+        assert application_instance.lms_url == "http://example.com"
+        assert application_instance.deployment_id == "DEPLOYMENT_ID"
+        assert application_instance.developer_key == "DEVELOPER_KEY"
+        aes_service.build_iv.assert_called_once()
+        aes_service.encrypt.assert_called_once_with(
+            aes_service.build_iv.return_value, "DEVELOPER_SECRET"
+        )
+        assert application_instance.aes_cipher_iv == aes_service.build_iv.return_value
+        assert application_instance.developer_secret == aes_service.encrypt.return_value
```
jon-betts, line 118: We now assert the actual value, rather than just any value

tests/unit/lms/services/application_instance_test.py
```diff
@@ -92,44 +92,39 @@ def test_get_by_deployment_id_raises_on_missing(
...
+        organization_service.get_by_public_id.assert_called_once_with(
+            mock.sentinel.org_id
+        )
+        assert application_instance.organization == org
```
jon-betts, line 122: Rolled in the org tests too

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -349,35 +349,37 @@ def test_update_instance(
...
         pyramid_request.params = {
-            "developer_key": key,
-            "developer_secret": secret,
-            "lms_url": lms_url,
-            "deployment_id": deployment_id,
+            "lms_url": "   http://example.com    ",
+            "deployment_id": " DEPLOYMENT_ID  ",
+            "developer_key": "  DEVELOPER KEY  ",
+            "developer_secret": " DEVELOPER SECRET  ",
         }
```
jon-betts, line 360: Similar happy "full-on" path testing here

tests/unit/lms/services/application_instance_test.py
```diff
@@ -92,44 +92,39 @@ def test_get_by_deployment_id_raises_on_missing(
...
 
-        if deployment_id:
-            assert application_instance.deployment_id == deployment_id
+    def test_update_application_instance_with_no_arguments(
+        self, service, application_instance
+    ):
+        service.update_application_instance(application_instance)
```
jon-betts, line 127: We don't make any assertions here. It's not a very good test, but it was hidden in the previous formulation.

tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -490,19 +492,13 @@ def application_instance_lti_13(self, application_instance, db_session):
...
-        # Real pyramid request have the same params available
-        # via POST and params
-        pyramid_request.POST.update(application_instance_data)
-        pyramid_request.params.update(application_instance_data)
-
-        pyramid_request.params["consumer_key"] = pyramid_request.POST[
-            "consumer_key"
-        ] = application_instance.consumer_key
+
         return pyramid_request
```
jon-betts, line 502: Just a little tidy up to make it more obvious which things are in the params

marcospri (approved):
Yeah this makes sense and it's easier to follow. I reckon that if there was a dependency between some parameters will test that in isolation.

PR #4919, opened 2023-01-24
Make application instance name mandatory on update

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4889

Requires:

 * https://github.com/hypothesis/lms/pull/4918

This makes application instance name mandatory when updating an existing application instance. The idea is to encourage better data quality over time.

## Testing notes

 * `make dev`
 * Visit: http://localhost:8001/admin/instance/id/5/
   * If that has a name search for one without
 * Try saving
 * You should get a name validation error
 * Fill out a name
 * It should save now

### DIFF
--- a/tests/unit/lms/views/admin/application_instance_test.py
+++ b/tests/unit/lms/views/admin/application_instance_test.py
@@ -334,9 +334,9 @@ class TestAdminApplicationInstanceViews:
     def test_update_application_instance(
         self, views, pyramid_request, ai_from_matchdict, application_instance_service
     ):
-        pyramid_request.params = {
+        pyramid_request.params = pyramid_request.POST = {
             "name": "  NAME  ",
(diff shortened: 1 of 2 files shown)


tests/unit/lms/views/admin/application_instance_test.py
```diff
@@ -354,11 +354,14 @@ def test_update_instance(
...
+        self,
+        views,
+        pyramid_request,
+        application_instance_service,
     ):
-        pyramid_request.params = {
+        pyramid_request.params = pyramid_request.POST = {
             "name": "  NAME  ",
-            "lms_url": "   http://example.com    ",
+            "lms_url": "http://example.com",
```
jon-betts, line 364: By using the URL pattern we no longer accept white space (it will say it's not a valid URL)

marcospri (approved):

PR #4920, opened 2023-01-24
Split out upgrading application instances as a separate activity

jon-betts:
This PR splits up the journey for application instance upgrade and new 1.3 instance so they have different starting and templates. 

This makes each journey and template simpler for the user and in the code, at the cost of making the overall size bigger.

## Testing notes

Testing new

 * `make dev`
 * Visit: http://localhost:8001/admin/registration/id/2/
 * Click "Add LTI 1.3 instance"
 * The registration you started with should appear
 * The section at the bottom of the page for upgrading should not appear
 * Everything here should work as normal

Testing upgrade

 * Visit: http://localhost:8001/admin/registration/id/2/
 * Click "Upgrade LTI 1.3 instance"
 * The registration you started with should appear
 * The regular "new" fields should not appear
 * Press "Save"
 * You should get validation errors for both "deployment_id" and "consumer_key"
 * Fill out "consumer_key": `Hypothesis4685565a4f5cf3c9cc3b199b0bd9abf4`
 * Fill out "deployment_id": `98477234`
 * Press "Save"
 * You should end up on an application instance now linked to this registration

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -146,6 +146,7 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route("admin.instance.consumer_key", "/admin/instance/{consumer_key}/")
     config.add_route("admin.instance.id", "/admin/instance/id/{id_}/")
     config.add_route("admin.instance.new", "/admin/instance/new")
+    config.add_route("admin.instance.upgrade", "/admin/instance/upgrade")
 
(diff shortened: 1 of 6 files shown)


marcospri (approved):
Nice, slightly longer but much more clear code and UI wise :+1:

PR #4929, opened 2023-01-26
New `list_all_files` method in BB API client

marcospri:
To populate the file picker in blackboard we make one API request per folder as the user navigates the tree. This is sensible as it keeps the API requests themselves small and fast.

Unfortunately for course copy we need the full tree of files to be able to:

- Find the current file that triggered the course copy process
- Guarantee that subsequent launches of other assignments in the course will also find their files without instructor intervention.

We'll use the new method only while doing course copy and keep the original `list_files` as the driver of the file picker.

## Testing 

We need this method for course copy but it's not used in the code base yet, one way to quickly check it is to apply a diff like:

```diff
diff --git a/lms/views/api/blackboard/files.py b/lms/views/api/blackboard/files.py
index cf6aa9f22..0a9df83ea 100644
--- a/lms/views/api/blackboard/files.py
+++ b/lms/views/api/blackboard/files.py
@@ -30,7 +30,7 @@ class BlackboardFilesAPIViews:
         course_id = self.request.matchdict["course_id"]
         folder_id = self.request.matchdict.get("folder_id")
 
-        results = self.blackboard_api_client.list_files(course_id, folder_id)
+        results = self.blackboard_api_client.list_all_files(course_id)
 
         response_results = []
```

and then head over the scratch assignment in blackboard and pick blackboard files. 

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1380_1&course_id=_139_1

That will make the file picker behave strangely as it will list all the tree directly on the root level but it will exercise the new code path.

### DIFF
--- a/lms/services/blackboard_api/client.py
+++ b/lms/services/blackboard_api/client.py
@@ -1,3 +1,4 @@
+from typing import List
 from urllib.parse import urlencode
 
 from lms.services.blackboard_api._schemas import (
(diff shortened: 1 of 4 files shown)


lms/services/blackboard_api/client.py
```diff
@@ -186,3 +160,84 @@ def course_groups(
...
+        results = schema(response).parse()
+        for _ in range(PAGINATION_MAX_REQUESTS - 1):
+            path = response.json().get("paging", {}).get("nextPage")
+            if not path:
+                break
+
+            response = self._api.request("GET", path)
+            results.extend(schema(response).parse())
+
+        return results
```
seanh, line 217: Ok, so this is pagination code that was already here, but it's just been moved out into a separate method?

lms/services/blackboard_api/client.py
```diff
@@ -186,3 +160,84 @@ def course_groups(
...
+
+        for response in responses:
+            # Get all the pages for this folder
+            folder_files = self._get_all_following_pages(
+                response, BlackboardListFilesSchema
+            )
+            # For each folder, get it children too
+            results.extend(self._get_all_folders_files(course_id, folder_files))
+
+        return results
```
seanh, line 243: But this is new I think: getting all the files from all the folders in a course. This looks like it could be a lot of requests!

lms/services/blackboard_api/client.py
```diff
@@ -186,3 +160,84 @@ def course_groups(
...
+                    else "blackboard_folder",
+                    "course_id": course_id,
+                    "lms_id": file["id"],
+                    "name": file["name"],
+                    "size": file["size"],
+                    "parent_lms_id": file["parentId"],
+                }
+                for file in files
+            ]
+        )
```
seanh, line 200: This is just a call that was already here that's been moved out into a helper method

lms/services/blackboard_api/client.py
```diff
@@ -186,3 +160,84 @@ def course_groups(
...
+        return (
+            path
+            + "?"
+            + urlencode(
+                {
+                    "limit": PAGINATION_LIMIT,
+                    "fields": "id,name,type,modified,mimeType,size,parentId",
+                }
+            )
+        )
```
seanh, line 182: The URL-generation that was already here, moved into a separate method

seanh (approved):

PR #4930, opened 2023-01-26
Fix the definition of billable to include at least one annotation in the group

jon-betts:
It appears we inadvertently changed the definition of billable users during the creation of report by missing out the requirement that a billable user be in a group with at least one annotation.

For more details see:

 * https://github.com/hypothesis/report/pull/167

This leads to a larger amount of users being counted as billable as a teacher can end up including users into other periods of time simply by visiting groups.


seanh (approved):

jon-betts:
There is on-going discussion about this metric change which ranges outside the technical. So when that's resolved we can remove the "do not merge".

PR #4935, opened 2023-01-27
Course copy plugin base implementation

marcospri:
See https://github.com/hypothesis/lms/pull/4880 for the blackboard implementation.

### DIFF
--- a/lms/product/plugin/course_copy.py
+++ b/lms/product/plugin/course_copy.py
@@ -58,8 +58,12 @@ class CourseCopyFilesHelper:
 
     @staticmethod
     def get_mapped_file_id(course, file_id):
-        """Get a previously mapped file id in course."""
-        return course.extra.get("course_copy_file_mappings", {}).get(file_id, None)
+        """
+        Get a previously mapped file id in course.
+
+        Returns the original `file_id` if no mapped one can be found.
+        """
+        return course.extra.get("course_copy_file_mappings", {}).get(file_id, file_id)
 
     @staticmethod
     def set_mapped_file_id(course, old_file_id, new_file_id):
--- a/lms/product/plugin/course_copy.py
+++ b/lms/product/plugin/course_copy.py
@@ -1,0 +1,89 @@
+from typing import Callable, Optional
+
+from lms.models import File
+from lms.services.exceptions import ExternalRequestError
+from lms.services.file import FileService
+
+
+class CourseCopyFilesHelper:
+    """Helper class to abstract common behaviour around LMS file / course copy."""
+
+    @staticmethod
+    def is_file_in_course(file_service: FileService, course_id, file_id, type_) -> bool:
+        """Check if `file_id` belongs to `course_id`."""
+
+        file = file_service.get(lms_id=file_id, type_=type_)
+        if not file or file.course_id != course_id:
+            return False
+
+        return True
+
+    @staticmethod
+    def find_matching_file_in_course(
+        store_new_course_files: Callable[[str], None],
+        file_service: FileService,
+        file_type: str,
+        original_file_id,
+        new_course_id,
+    ) -> Optional[File]:
+        try:
+            # Get the current (copied) courses files, that will have the side effect of storing files in the DB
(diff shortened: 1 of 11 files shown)


lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,93 @@
...
+from lms.models import File
+from lms.services.exceptions import ExternalRequestError
+from lms.services.file import FileService
+
+
+class FileNotFoundInCourse(Exception):
+    pass
+
+
+class CourseCopyFilesHelper:
```
marcospri, line 8: Collaborator centered around files. These are all staticfunctions.

I rekcon we'll have soon a `CourseCopyGroupsHelper`

lms/services/file.py
```diff
@@ -21,6 +21,26 @@ def get(self, lms_id, type_):
             .one_or_none()
         )
 
+    def find_copied_file(self, new_course_id, original_file: File):
```
marcospri, line 24: Not part of the plugin but is used by it. Small and straightforward enough to include here.

tests/unit/lms/product/plugin/course_copy_test.py
```diff
@@ -0,0 +1,120 @@
...
+            file_service, sentinel.course_id, sentinel.file_id, sentinel.type
+        )
+
+        file_service.get.assert_called_once_with(sentinel.file_id, sentinel.type)
+        assert is_in_course == expected
+
+    @pytest.mark.parametrize("raising", [True, False])
+    def test_find_matching_file_in_course(self, helper, file_service, raising):
+        store_new_course_files = Mock()
+        if raising:
```
marcospri, line 32: Raising or not should have the same effect. Tackling it in the same tests.

tests/unit/lms/services/file_test.py
```diff
@@ -7,24 +7,39 @@
 from tests import factories
 
 
-class TestGet:
-    def test_it_returns_the_matching_file_if_there_is_one(
+class TestFileService:
```
marcospri, line 10: Named after a method but testing the whole class. Updated the naming of the class and methods.

tests/unit/lms/product/plugin/course_copy_test.py
```diff
@@ -0,0 +1,120 @@
...
+        is_in_course = helper.is_file_in_course(
+            file_service, sentinel.course_id, sentinel.file_id, sentinel.type
+        )
+
+        file_service.get.assert_called_once_with(sentinel.file_id, sentinel.type)
+        assert is_in_course == expected
+
+    @pytest.mark.parametrize("raising", [True, False])
+    def test_find_matching_file_in_course(self, helper, file_service, raising):
+        store_new_course_files = Mock()
```
seanh, line 31:
```suggestion
        store_new_course_files = create_autospec(lambda new_course_id: None)
```

Best to create a specced mock, otherwise the code could pass any number of arguments to `store_new_course_files()` and the test would still pass.

tests/unit/lms/services/file_test.py
seanh: :+1: Good catch

seanh (approved):

lms/product/plugin/course_copy.py
```diff
@@ -0,0 +1,89 @@
...
+        raise NotImplementedError()
+
+    def find_matching_file_in_course(self, *args, **kwargs):
+        raise NotImplementedError()
+
+    def get_mapped_file_id(self, *args, **kwargs):
+        raise NotImplementedError()
+
+    def set_mapped_file_id(self, course, old_file_id, new_file_id):
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
marcospri: We instantiate it as the default implementation of the plugin, so for example in moodle `product.plugin.course_copy.XXXX` would raise NotImplementedError()

I reckon you are right about the `args` `kwargs` here and the general weirdness of this class. I'd like to revisit it after D2L uses it, it will be easier to spot the actual protocol then. Created an issue to track that.

Issue #4946, opened 2023-01-31
Final wording for the "Can't find file" modal in D2L


marcospri:
Slack thread: https://hypothes-is.slack.com/archives/C04JRP516GG/p1678373845421789?thread_ts=1678361298.854569&cid=C04JRP516GG

Current placeholder looks like:

![Image](https://user-images.githubusercontent.com/1433832/215786255-c1daa094-5cc4-4434-9eb9-b2f9da3c0ba1.png)

The final version might have to:

- Include links to KB articles / D2L docs
- Mention course copy (and maybe also link to articles around that)

marcospri:
Placeholder copy for the student and instructor versions:

Note that `If the problem persists, you can open a support ticket or visit our help documents.` is part of both versions. If we want to change that it a bit more complex so I rather get a version out and tackle that separately if we wan to do so.


## Instructor

```
Title: Hypothesis couldn't find the file in this course

This might have happened because:
- The file has been deleted from D2L
- The course was copied and the selected one is not in the new course.

To fix the issue recreate the assignment with a different file. Also check our document
Using Hypothesis With D2L Course Content Files(https://web.hypothes.is/help/using-hypothesis-with-d2l-course-content-files/)

If the problem persists, you can open a support ticket or visit our help documents.
```

## Student
```
Title: Hypothesis couldn't find the file in the course

This might have happened because:
- The file has been deleted from D2L
- The course was copied and the selected is not in the new course.

Please constant the course instructor pointing to this assignment. 

If the problem persists, you can open a support ticket or visit our help documents.
```

nairiboo:
Instructor one looks good to me.

For the student one a couple of points:

Title: Hypothesis couldn't find the file in the course

This might have happened because:
- The file has been deleted from D2L
- The course was copied and the selected **one** is not in the new course.

Please **contact** the course instructor pointing **them** to this assignment. 

If the problem persists, you can open a support ticket or visit our help documents.

PR #4948, opened 2023-01-31
Handle missing D2L files

marcospri:
The cause for them to be missing could be them being removed, course copy or something else.

This brings D2L inline with other LMS and we can use this as the basics to handle course copy.

## Testing

In main, launch:

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2376/View

you'll get a generic error message.

Launch again in this branch and you'll get slightly more specific error modal this time.


lms/static/scripts/frontend_apps/components/LaunchErrorDialog.js
```diff
@@ -85,6 +85,22 @@ export default function LaunchErrorDialog({
...
+
+    case 'd2l_file_not_found_in_course':
+      return (
+        <ErrorModal
+          busy={busy}
+          error={error}
+          onRetry={onRetry}
+          title="Hypothesis couldn't find the file in the course"
+        >
+          <p>
```
marcospri, line 97: D2L files is not yet released so the actual wording here is not that important.

Tracking getting the final message (once course copy is done for D2L) here https://github.com/hypothesis/lms/issues/4946

lms/services/exceptions.py
```diff
@@ -238,24 +238,28 @@ class CanvasAPIServerError(CanvasAPIError):
     """
 
 
-class CanvasFileNotFoundInCourse(Exception):
-    """A Canvas file ID wasn't found in the current course."""
-
-    error_code = "canvas_file_not_found_in_course"
-
+class FileNotFoundInCourse(Exception):
```
marcospri, line 241: I just realized that having this type here could be useful for course copy (detect a problem in any LMS).

It also avoid a little amount of repetition.

jon-betts: Do you think we should be making use of the `SerializableError` here for even less duplication? Something like:

```python
class FileNotFoundInCourse(SerializableError):
    def __init__(self, product_code, file_id):
        super().__init__(
            error_code=f"{{product}}_file_not_found_in_course", 
            details={"file_id": file_id}
        )
```

If we had that, I don't think I'd bother creating a D2L specific sub-class, it could be raised directly with:

```python
raise FileNotFoundInCourse("d2l", file_id)
```

If all the different exceptions are supposed to have the same interface, do we need to have all the different kinds?

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -203,6 +205,26 @@ def test_public_url(self, svc, basic_client):
...
+
+    def test_public_url_raises_ExternalRequestError_if_the_api_fails_in_any_other_way(
+        self, svc, basic_client
+    ):
+        basic_client.request.side_effect = ExternalRequestError(
+            response=factories.requests.Response(status_code=400)
+        )
+
+        with pytest.raises(ExternalRequestError):
+            svc.public_url("COURSE_ID", "FILE_ID")
```
jon-betts, line 226: This looks like a good spot to parameterize?

jon-betts (dismissed):
I couldn't follow the instructions as written, as I think it relies on some setup I don't have? Instead I selected the file and then edited the URL to be broken, which seemed to do the trick.

I think this follows the patterns set down before and is totally good to go as is. 

If you wanted to I have some other ideas for possible improvements:

 * I think we should use the `SerializableError` as the base for the new exception
 * We could think about whether we need all the different product specific sub-classes, where we could instantiate the missing file error with all the things it needs
 * Do we need?
    * `blackboard_file_not_found_in_course`
    * `canvas_file_not_found_in_course`
    * `d2l_file_not_found_in_course`
  * Could we perhaps get away with:
     * `file_not_found_in_course`
     * `deep_linked_file_not_found_in_course`
  * I think the guidance for `blackboard_file_not_found_in_course` and `d2l_file_not_found_in_course` is basically the same, but one is better written?

lms/services/exceptions.py
marcospri: :+1: Went for basically this approach

lms/product/__init__.py
```diff
@@ -13,6 +12,9 @@ def includeme(config):
     config.include("lms.product.canvas")
     config.include("lms.product.d2l")
 
+    # Lazy import to avoid circular dependencies
+    from lms.product.factory import get_product_from_request
```
marcospri, line 16: This triggers an import of each product and their plugins so its quite heavy.

I'm sure shuffling things around might be a better solution but it will make this quite a bigger PR.

lms/services/exceptions.py
```diff
@@ -238,26 +240,6 @@ class CanvasAPIServerError(CanvasAPIError):
     """
 
 
-class CanvasFileNotFoundInCourse(Exception):
```
marcospri, line 241: These were all the same, except for the error code :+1:

lms/services/exceptions.py
```diff
@@ -281,6 +263,23 @@ def __init__(
         self.details = details
 
 
+class FileNotFoundInCourse(SerializableError):
+    """A file wasn't found in the current course."""
+
+    def __init__(self, family, file_id):
+
+        ERROR_CODE_MAP = {
+            Family.D2L: "d2l_file_not_found_in_course",
```
marcospri, line 272: I like to have the map here instead of doing something like `{family}_file_not_found`

I reckon I'll forget about this in a week, with this approach the error code is more easily grepeable (and a comment would go out of date).

tests/unit/lms/services/blackboard_api/client_test.py
```diff
@@ -319,16 +319,18 @@ def test_it(
...
     ):
         basic_client.request.side_effect = ExternalRequestError(
             response=factories.requests.Response(status_code=404)
         )
 
-        with pytest.raises(BlackboardFileNotFoundInCourse):
+        with pytest.raises(FileNotFoundInCourse) as excinfo:
             svc.public_url("COURSE_ID", "FILE_ID")
 
+        assert excinfo.value.error_code == "blackboard_file_not_found_in_course"
```
marcospri, line 332: Checking the error code in all of these to make the change safer.

tests/unit/lms/services/d2l_api/client_test.py
```diff
@@ -192,6 +194,26 @@ def test_public_url(self, svc, basic_client):
...
+        self, svc, basic_client, status_code, exception, error_code
+    ):
+        basic_client.request.side_effect = ExternalRequestError(
+            response=factories.requests.Response(status_code=status_code)
+        )
+
+        with pytest.raises(exception) as excinfo:
+            svc.public_url("COURSE_ID", "FILE_ID")
+
+        if error_code:
```
marcospri, line 214: Parametrized this, it does have this conditional wrinkle in it.

marcospri:
> Could we perhaps get away with:
    file_not_found_in_course
    deep_linked_file_not_found_in_course

Nah, they are all the same, we just need the LMS name there to link to different KB articles etc. Ideally we could move to a world with  an error code meaning just what went wrong and a separate product to indicate where.

I reckon that will simplify the FE code as well but that's out of scope for today.

>I think the guidance for blackboard_file_not_found_in_course and d2l_file_not_found_in_course is basically the same, but one is better written?

Yeah, D2L one it's just a placeholder for now.

marcospri:
> I couldn't follow the instructions as written, as I think it relies on some setup I don't have? Instead I selected the file and then edited the URL to be broken, which seemed to do the trick.

I merged the devdata PR, I reckon that should have been fixed.

lms/models/family.py
```diff
@@ -0,0 +1,18 @@
+from enum import Enum
+
+
+class Family(str, Enum):
```
marcospri, line 4: Moved to `models` easier to import it from here.

tests/unit/lms/services/blackboard_api/client_test.py
jon-betts: Yeah, I'd say this is a feature worth testing. I'm not a fan of testing human readable strings, but codes are here exactly here for this reason.

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -174,12 +171,12 @@ def test_it_when_file_not_in_course(
         self, view, blackboard_api_client, course_service, course_copy_plugin
     ):
         course_copy_plugin.is_file_in_course.return_value = False
-        blackboard_api_client.public_url.side_effect = BlackboardFileNotFoundInCourse(
-            file_id="FILE_ID"
+        blackboard_api_client.public_url.side_effect = FileNotFoundInCourse(
+            "blackboard_file_not_found_in_course", file_id="FILE_ID"
         )
```
jon-betts: Something is up here. This test still passes if you remove this line?

I was investigating having to instantiate the error correctly, but it turns out it doesn't matter if you set it at all in this test?

tests/unit/lms/services/canvas_test.py
```diff
@@ -55,8 +52,8 @@ def test_if_the_file_isnt_in_the_course_it_finds_a_matching_file_instead(
         assignment,
         public_url_for_file,
     ):
-        canvas_file_finder.assert_file_in_course.side_effect = (
-            CanvasFileNotFoundInCourse(sentinel.file_id)
+        canvas_file_finder.assert_file_in_course.side_effect = FileNotFoundInCourse(
+            "canvas_file_not_found_in_course", sentinel.file_id
         )
```
jon-betts, line 57: Having to instantiate a full exception here kind of sucks. You have to do it because we have non-nullable inputs to the exception. I think it's not nice for a couple of reasons

#### It's not as neat
 
This is partially just general neat freakery on my part, but it also harms readability.
    
#### It includes spuriously accurate information

There are values in the test here which actually don't make any difference to the test. I think we should take a kind of "Chekhovs value" approach, that:

 * If it's here we should use it
 * If we don't use it it shouldn't be here

This is because it loads the test with lots of details which _might_ be relevant, but actually aren't. For example we use `sentinel.file_id` later on. Is that read from the exception? It's not, and if it wasn't here we can't make that incorrect leap. This also narrows the focus of what this test is, and isn't about. 

I think there's a competing school of thought that says values should be "realistic". I think the idea is to highlight how the code operates by reading the tests. 

#### What can we do about it?

a) We could **change the exception to have default values**, but in this case there aren't any defaults, and not providing inputs would be an error, so that's not great.

b) We could **choose names to convey meaning** and explain they don't matter.

```python
FileNotFoundInCourse("arbitrary_error_code", file_id="not_important")
```

c) We could **add a factory class** for these? These could re-used between tests, and a set of exception factories seems reasonable enough. Factories allow us to have randomised values, which positively states we don't care about the value. This is the opposite of having an irrelevant but likely looking string:

 * A fixed string: Implies the value is important
 * No value: Implies the value isn't important (or maybe the default is?)
 * A randomized value: States the value isn't important

d) **Do nothing**, as it's not super important, I mostly wanted to have a chat about testing methodology.

tests/unit/lms/services/d2l_api/client_test.py
jon-betts: Ack! Can't win 'em all.

tests/unit/lms/views/api/blackboard/files_test.py
```diff
@@ -170,18 +167,16 @@ def test_it_when_file_not_in_course_fixed_by_course_copy(
...
         assert response == {"via_url": helpers.via_url.return_value}
 
-    def test_it_when_file_not_in_course(
-        self, view, blackboard_api_client, course_service, course_copy_plugin
-    ):
+    @pytest.mark.usefixtures("user_is_instructor")
```
jon-betts, line 170:
```suggestion
    # Instructor is the default, but this test shouldn't pass for students
    @pytest.mark.usefixtures("user_is_instructor")
```

jon-betts (approved):
I've already tested that the PR works, the remaining discussion has been about factoring. I think this is looking good now, and clears up some unnecessary differentiation in our exception hierarchy.

I've left a comment about not ways to avoid using fixed values to tests where those values don't actually matter in the test. Partially because I think it's a good idea, but partially because I find it interesting.

It's not a big deal, and shouldn't hold up the PR, but the potential fixes are also quite easy (factory, or different values probably?), so dealers choice.

tests/unit/lms/services/canvas_test.py
marcospri: Created an issue off this comment. I do revisit these as I add them to the project itself.

PR #4957, opened 2023-02-01
Generic `course_group_sets` view for every LMS

marcospri:
Taking advantage of the existing `GroupingPlugin` scaffolding to settle on a generic view instead of one per product.

This is done in the context of course copy, some changes might be required here, with this approach they are easier to generalize.

This is a similar approach to the current `/sync` API for which we only have one version that uses the plugins instead of multiples views.

# Testing

- There should not be behavior changes here but as sanity check we'll see if retrieving the group sets is still working on all LMS's

### Canvas

https://hypothesis.instructure.com/courses/125/assignments/3930/edit?name=Marcos+Assignment&due_at=null&points_possible=0

(pick the right LTI tool, configure with any document and check the `This is a group assignment`

### Blackboard 

Same process, configure and check the groups checkbox.

### D2L

Same story again: 

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2177/View


lms/product/blackboard/_plugin/grouping.py
```diff
@@ -22,6 +22,9 @@ class BlackboardGroupingPlugin(GroupingPlugin):
     def __init__(self, blackboard_api):
         self._blackboard_api = blackboard_api
 
+    def get_group_sets(self, course_id):
```
marcospri, line 25: Each plugin know how to get the list of groups sets.

lms/product/blackboard/product.py
```diff
@@ -14,7 +14,6 @@ class Blackboard(Product):
     route: Routes = Routes(
         oauth2_authorize="blackboard_api.oauth.authorize",
         oauth2_refresh="blackboard_api.oauth.refresh",
-        list_group_sets="blackboard_api.courses.group_sets.list",
```
marcospri, line 17: No need to parametrize the name of this view. There's only one now.

lms/product/product.py
```diff
@@ -34,9 +34,6 @@ class Routes:
     oauth2_refresh: Optional[str] = None
     """Refreshing OAuth 2 tokens."""
 
-    list_group_sets: Optional[str] = None
```
marcospri, line 37: Again, this is not part of the product, all product will use the same view/route.

lms/resources/_js_config/__init__.py
```diff
@@ -443,7 +443,7 @@ def _get_product_info(self):
             "api": {},
         }
 
-        if product.route.list_group_sets:
+        if self._request.product.settings.groups_enabled:
```
marcospri, line 446: Not sure about this conditional.

I reckon this version is better, before it was:

- Return the `group_set_api` if the product is configured for it

While this version is:

- Return the `group_set_api` if groups are enabled.


We could return the endpoint in all cases and the client well decide to call it based on the value of `groupsEnabled` above. :shrug:

lms/resources/_js_config/__init__.py
```diff
@@ -453,8 +453,7 @@ def _get_product_info(self):
             product_info["api"]["listGroupSets"] = {
                 "authUrl": self._request.route_url(product.route.oauth2_authorize),
                 "path": self._request.route_path(
-                    product.route.list_group_sets,
-                    course_id=api_course_id,
+                    "api.courses.group_sets.list", course_id=api_course_id
```
marcospri, line 456: It's fine to hard code the endpoint, there's only one for all products.

lms/views/api/groups.py
```diff
@@ -7,9 +7,9 @@
     request_method="GET",
     permission=Permissions.API,
     renderer="json",
-    route_name="blackboard_api.courses.group_sets.list",
+    route_name="api.courses.group_sets.list",
```
marcospri, line 10: The diff is finding the blackboard version is now the "generic" one.

They are very similar indeed.

lms/views/api/groups.py
```diff
@@ -7,9 +7,9 @@
...
 )
 def course_group_sets(_context, request):
-    return request.find_service(name="blackboard_api_client").course_group_sets(
+    return request.product.plugin.grouping.get_group_sets(
```
marcospri, line 13: Go from product -> plugin -> method instead of `lms-api` -> method.

lms/static/scripts/frontend_apps/components/GroupConfigSelector.js
```diff
@@ -166,6 +166,7 @@ export default function GroupConfigSelector({
         await apiCall({
           authToken,
           path: listGroupSetsAPI.path,
+          data: listGroupSetsAPI.data,
```
marcospri, line 169: We were not forwarding the data property here.

lms/resources/_js_config/__init__.py
```diff
@@ -453,9 +453,13 @@ def _get_product_info(self):
...
                 ),
+                "data": {
```
marcospri, line 458: We need to explicitly pass the product family back to ourselves.

In the context of the API it might be impossible to know in which LMS we operate.

But we do know that in the launch, so just send that back to us.

This mechanism is not new, just new for this view now.

lms/product/plugin/grouping.py
```diff
@@ -45,6 +45,9 @@ def get_sections_for_grading(
 
         return None
 
+    def get_group_sets(self, course_id):  # prama: nocover
```
seanh, line 48:
```suggestion
    def get_group_sets(self, course_id):  # pragma: nocover
```

But I think you might want an ABC or a protocol, so that coverage knows that these aren't real methods and won't expect them to be covered in the first place.

seanh: Do we want type annotations on this method? The rest of the methods in this class seem to have them.

lms/product/plugin/grouping.py
```diff
@@ -45,6 +45,9 @@ def get_sections_for_grading(
...
+        """Get the group sets available for this course. Used while configuring the assignments."""
```
seanh, line 49:
```suggestion
        """Return the list of group sets for the given course."""
```

I think the docstring should stick to describing the contract of the method rather than when the method is used, otherwise it's likely to become out of date. I think just grepping for the method is going to be the only reliable way to find out when it's used

lms/resources/_js_config/__init__.py
seanh: Can't we tell from the authentication of the API request which LMS we're operating in? The authenticated user should be associated with an application instance

seanh (approved):
:+1: This looks sound to me from a backend point of view

lms/resources/_js_config/__init__.py
marcospri: No, unfortunately we can tell 100% of the time which product a given application instance belongs to. It's probably very close to that 100% thought.

We probably should start logging the edge cases and get a sense of the actual numbers.

robertknight (approved):
The frontend changes look straightforward to me.

PR #4958, opened 2023-02-01
Rename LMSGrader => GradingToolbar and convert to TS

robertknight:
The "LMS" part of the component name was redundant, and the "Toolbar" suffix makes it more obvious what kind of control this is.

This might not be the final name, as we're considering adding an "edit assignment" button to this toolbar in future and displaying it even when grading is not active.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/GradingToolbar.tsx
+++ b/lms/static/scripts/frontend_apps/components/GradingToolbar.tsx
@@ -1,3 +1,5 @@
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
 import {
   useCallback,
   useContext,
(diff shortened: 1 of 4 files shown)


lms/static/scripts/frontend_apps/components/GradingToolbar.tsx
```diff
@@ -71,8 +69,7 @@ export default function LMSGrader({
    * Makes an RPC call to the sidebar to change to the focused user.
    */
   const changeFocusedUser = useCallback(
-    /** @param {StudentInfo|null} user - The user to focus on in the sidebar */
-    async user => {
+    async (user: StudentInfo | null) => {
```
robertknight, line 72: I deleted a couple of comments which, IMO, did not add value. eg. Here there is a comment above which says the same thing.

lyzadanger (approved):
I appreciate the direction the naming is going! I wonder if we might end up with a `Toolbar, which might contain a grading-control component but might also contain other things...

PR #4959, opened 2023-02-01
Handle OAuth2TokenError during course copy

marcospri:
OAuth2TokenError is a ExternalRequestError and the more general except could in some cases shallow the need for a new token or refreshing the existing one before getting a relevant answer from the API.

Catch OAuth2TokenError independently and raise it to get a new token.

I'm struggling to reproduce the exact same issue but I reckon this makes sense in isolation.

If we look at:

https://github.com/hypothesis/lms/blob/d42e45267255cc0efdab30fd4b716038deaf120b/lms/views/api/blackboard/files.py#L77-L99

If self.request.lti_user.is_instructor is True
and `self.course_copy_plugin.is_file_in_course(course_id, file_id)` is False.

The first API request might be made inside `find_matching_file_in_course`.If that returns false will raise `BlackboardFileNotFoundInCourse`. If find_matching_file_in_course shallowed 
OAuth2TokenError we'll display the "we can't fin the file" modal instead of the "authorize" one.

With this change, if `find_matching_file_in_course` raises OAuth2TokenError that will make all the way to the top.

### Testing

I'm not able to reliably reproduce this, as a sanity check launch:

- The original course file assignment

- The copied course file assignment

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1371_1&course_id=_139_1

### DIFF
--- a/lms/product/plugin/course_copy.py
+++ b/lms/product/plugin/course_copy.py
@@ -1,7 +1,7 @@
 from typing import Callable, Optional
 
 from lms.models import File
-from lms.services.exceptions import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError, OAuth2TokenError
 from lms.services.file import FileService
 
 
@@ -29,6 +29,10 @@ class CourseCopyFilesHelper:
         try:
             # Get the current (copied) courses files, that will have the side effect of storing files in the DB
             _ = store_new_course_files(new_course_id)
+        except OAuth2TokenError:
+            # If the issue while trying to talk to the API is OAuth related
+            # let that bubble up and be dealt with.
+            raise
         except ExternalRequestError:
             # We might not have access to use the API for that endpoint.
             # That will depend on our role and the course's permissions settings.
--- a/tests/unit/lms/product/plugin/course_copy_test.py
+++ b/tests/unit/lms/product/plugin/course_copy_test.py
@@ -3,7 +3,7 @@ from unittest.mock import Mock, create_autospec, sentinel
 import pytest
 
 from lms.product.plugin.course_copy import CourseCopyFilesHelper
-from lms.services.exceptions import ExternalRequestError
+from lms.services.exceptions import ExternalRequestError, OAuth2TokenError
 from tests import factories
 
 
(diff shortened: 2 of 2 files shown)


tests/unit/lms/product/plugin/course_copy_test.py
```diff
@@ -118,3 +133,7 @@ def test_set_mapped_file_id(self, helper):
     @pytest.fixture
     def helper(self):
         return CourseCopyFilesHelper()
+
+    @pytest.fixture
+    def store_new_course_files(self):
+        return create_autospec(lambda _: None)  # pragma: nocover
```
marcospri, line 139: No cover, the lambada only exists to create the spec, but it never gets called itself.

lms/product/plugin/course_copy.py
```diff
@@ -29,6 +29,10 @@ def find_matching_file_in_course(
         try:
             # Get the current (copied) courses files, that will have the side effect of storing files in the DB
             _ = store_new_course_files(new_course_id)
+        except OAuth2TokenError:
+            # If the issue while trying to talk to the API is OAuth related
+            # let that bubble up and be dealt with.
+            raise
```
seanh, line 35: This makes sense to me: we need `OAuth2TokenError` to be handled differently from other errors

tests/unit/lms/product/plugin/course_copy_test.py
```diff
@@ -27,8 +27,9 @@ def test_is_file_in_course(self, file_service, helper, file, expected):
         assert is_in_course == expected
 
     @pytest.mark.parametrize("raising", [True, False])
-    def test_find_matching_file_in_course(self, helper, file_service, raising):
-        store_new_course_files = create_autospec(lambda _: None)  # pragma: nocover
+    def test_find_matching_file_in_course(
+        self, helper, file_service, raising, store_new_course_files
+    ):
```
seanh, line 32: `store_new_course_files()` mock moved out into a shared fixture :+1:

seanh (approved):
:+1: I can see how this would be difficult to reproduce but the change makes total sense to me

PR #4962, opened 2023-02-02
Reduce console noise for expected OAuth2TokenError

marcospri:
## Testing

Compare `main` and thins branch's console output while doing:

- tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate oauth2_token" to force re-auth

- Lauch anything requiring a token 

https://hypothesis.instructure.com/courses/125/assignments/1833

### DIFF
--- a/lms/app.py
+++ b/lms/app.py
@@ -25,7 +25,12 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.add_tween(
         "pyramid_exclog.exclog_tween_factory", under="pyramid_tm.tm_tween_factory"
     )
-    config.add_settings({"exclog.extra_info": True})
+    config.add_settings(
+        {
+            "exclog.extra_info": True,
+            "exclog.ignore": ["lms.services.exceptions.OAuth2TokenError"],
+        }
+    )
     config.include("pyramid_exclog")
 
     config.include("pyramid_jinja2")
--- a/lms/views/api/exceptions.py
+++ b/lms/views/api/exceptions.py
@@ -1,4 +1,5 @@
 """Error views for the API."""
+import logging
 from dataclasses import asdict, dataclass
 from urllib.parse import urlparse, urlunparse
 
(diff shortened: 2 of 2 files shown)


lms/app.py
```diff
@@ -25,7 +25,12 @@ def create_app(global_config, **settings):  # pylint: disable=unused-argument
     config.add_tween(
         "pyramid_exclog.exclog_tween_factory", under="pyramid_tm.tm_tween_factory"
     )
-    config.add_settings({"exclog.extra_info": True})
+    config.add_settings(
+        {
+            "exclog.extra_info": True,
+            "exclog.ignore": ["lms.services.exceptions.OAuth2TokenError"],
+        }
+    )
```
seanh, line 33: :+1: I think this makes sense -- `OAuth2TokenError`'s are a normal and expected (and frequent) part of the app's operation. We don't need a bunch of info logged about them for debugging purposes.

I believe this will affect both production and development the same, right? :+1:

seanh: What about Sentry? I can't remember whether `OAuth2TokenError`'s are going there currently or are they already filtered out?

lms/views/api/exceptions.py
```diff
@@ -157,6 +161,7 @@ def external_request_error(self):
 
     @exception_view_config(context=OAuth2TokenError)
     def oauth2_token_error(self):
+        LOG.debug("Exception view for: OAuth2TokenError")
```
seanh, line 164: I'm not sure that I'd bother with this, to be honest

seanh (approved):
Yeah this is annoying, and this seems like a sensible change to me.

Docs for pyramid_exclog: <https://docs.pylonsproject.org/projects/pyramid-exclog/en/latest/>. It's generally quite a noisy extension. It is useful though: it prevents us from ever having to log exceptions ourselves (something that is frequently done wrong: failing to log them or not logging enough info about them, and that can often complicate code as well). But it maybe makes the looks a bit too noisy with our current settings

seanh (approved):

lms/app.py
marcospri: I don't see them in sentry but I can't find how they are being filtered.

lms/views/api/exceptions.py
marcospri: I reckon if we logged every request in the local server (and probably the status code) this won't be needed but without it now the server logs would be silent for this case which can be confusing.

marcospri:
I reckon we can look at the sentry side of things independently

PR #4963, opened 2023-02-02
Convert frontend config types to TS

robertknight:
This is a direct conversion except for changing `any` to `unknown` in a couple of places.


lyzadanger (approved):
I have a few questions about some of the types here and their purpose but I recognize that that is outside the scope of these changes and I don't want to slow you down! Looks straightforward.

PR #4967, opened 2023-02-02
Remove application instance admin pages consumer key route

jon-betts:
A lot of our logic is a bit confused by having two paths to get to the same place. We often test the consumer key route, but it's not used in practice. This removes the route and makes id the default.

There is definitely some repeated setup and testing with the id stuff now, but I think this is enough for one PR to try and get right?

Some other changes I've made in the tests:

 * Replaced `AdminApplicationInstanceViews(pyramid_request)` with `views` fixture
 * Tried to get the argument order a bit more sane


marcospri (approved):
Nice, one less thing to keep track on the admin pages :+1:

Issue #4965, opened 2023-02-02
Add last updated to application instances and show it in the UI


jon-betts:
Some feedback from support people was that it's very good information to have some indication of when an application instance was last active.

This could take the form of the last launch associated with it for example. Another option is to have an updated field.

Both of these suffer in varying degrees by not existing. We have a create date field, but no update field, so if we added one it would only take effect from now. Last launch is a bit better in that it will cover last August to now.

Application instances are associated with groups however, which do have updated dates, so what we could do is create a conventional updated date and then back-fill it with the `MAX(created)` from associated groups. This would probably be a decent approximation. This also has the benefit of being simple (it's a field on the object you are looking at), rather than requiring any combining with other tables to work out.

## Proposed solution

 * Use the `CreatedUpdatedMixin` to add an updated field (as a separate book keeping measure)
 * Add a new `last_launched` field
 * Back-fill the values based on a query using last launch date

## Tasks

 * [x] https://github.com/hypothesis/lms/pull/4969
 * [x] https://github.com/hypothesis/lms/pull/4968
 * [x] https://github.com/hypothesis/lms/pull/4970
 * [x] https://github.com/hypothesis/lms/pull/5030
 * [x] https://github.com/hypothesis/lms/pull/5032

jon-betts:
Done!

PR #4968, opened 2023-02-02
Track last launched date against application instances

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4965

This PR adds tracking of last launch date to the application instance service. There is a slight fudge as we count a deep link as a launch because the follow the same code path.

This is probably technically wrong, practically right. People want to know the most recent and "in-play" application instances. Something you just created would meet this brief.

Requires:

 * https://github.com/hypothesis/lms/pull/4969

## Testing notes

 * Follow the notes in: https://github.com/hypothesis/lms/pull/4969
    * `make sql`
    * `SELECT created, updated, last_launched FROM application_instances WHERE id=8`
    * Keep the values on screen to compare
 * Relaunch: https://hypothesis.instructure.com/courses/125/assignments/873
    * `SELECT created, updated, last_launched FROM application_instances WHERE id=8`
    * The updated and last launched values should have changed
 * Visit: http://localhost:8001/admin/instance/8/
    * Press "Save"
    * `SELECT created, updated, last_launched FROM application_instances WHERE id=8`
   * The update date should change, but last launched should not

### DIFF
--- a/lms/services/application_instance.py
+++ b/lms/services/application_instance.py
@@ -316,6 +316,10 @@ class ApplicationInstanceService:
         ]:
             setattr(application_instance, attr, params.get(attr))
 
+        # This is a potentially misleading, as we can get here from deep-linked
+        # "launches". Depending on whether you count that as a launch or not.
+        application_instance.last_launched = datetime.now()
+
 
 def factory(_context, request):
     return ApplicationInstanceService(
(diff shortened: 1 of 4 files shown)


lms/services/application_instance.py
```diff
@@ -316,6 +316,10 @@ def update_from_lti_params(self, application_instance, params: LTIParams):
         ]:
             setattr(application_instance, attr, params.get(attr))
 
+        # This is a potentially misleading, as we can get here from deep-linked
+        # "launches". Depending on whether you count that as a launch or not.
+        application_instance.last_launched = datetime.now()
```
marcospri, line 321: I think counting the deep linking is correct, is "technically" also a launch and it shows activity on the AI.

In the events table we have the raw data if that distinction mattered for some troubleshooting.

marcospri (approved):

PR #4970, opened 2023-02-02
Display the last launched dates on application instances admin pages

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4965

Requires:

 * https://github.com/hypothesis/lms/pull/4968
 * https://github.com/hypothesis/lms/pull/4969

## Implementation notes

There are some big rough edges on this one. 

The first is date formatting. We haven't tried harder anywhere else by we are showing dates to the micro-second, which is long and unhelpful.

~The second is that we are triggering horizontal scrolling in the application instance tables. I've asked support if we can drop / re-order / truncate any of the fields (https://hypothes-is.slack.com/archives/C2BLQDKHA/p1675706515773809)~. There are some extensive notes at the bottom about new more useful (and compact) columns.

image

These new columns have more likely break points for inserting softbreaks, so they tend to wrap and don't trigger horizontal scroll. A very long URL could probably still do it, but I've put it as the last column.

## Testing notes

 * Follow the steps in: https://github.com/hypothesis/lms/pull/4968
     * Visit: http://localhost:8001/admin/instance/8/
     * The last launch date should be visible
  * Visit: http://localhost:8001/admin/instances/
     * Press "Search"
     * You should see the last launch dates
     * You may have to horizontally scroll to see them

## Feedback on useful columns

The columns we show at the moment aren't very useful for telling application instances apart. They tend to fall into two (overlapping) categores:

 * Long inscrutable strings
 * The thing you searched for, and so don't need to be told

Current practice in the support team is to search in Metabase to find the correct record using other fields, and then locate it in the admin pages. Ideally we'd present the most relevant details in the admin pages so this wasn't required.

### Column summary

* LMS URL - useful to see
* Requesters email - useful to see 
* Last launched - useful
* Created - useful
* Name - useful some day
* Settings - useful to see (but huge!)
* GUID - not so useful
* Consumer key - not so useful
* Deployment ID - not useful?

### Current columns

The current columns heavily cluster around not-useful settings:

 * Name 
 * Consumer key
 * GUID
 * Deployment ID 

### Proposed columns

* Name 
* LMS URL
* Requesters Email
* Created - As `YYYY-MM-DD HH:MM`
* Last Launched - As `YYYY-MM-DD HH:MM`
* `tool_consumer_info_product_family_code` - Family

### DIFF
--- a/lms/templates/admin/macros.html.jinja2
+++ b/lms/templates/admin/macros.html.jinja2
@@ -79,6 +79,7 @@
     {% endif %}
 {% endmacro %}
 
+
 {% macro object_list_table(request, route, objects, fields) %}
 <div class="container">
     <div class="table-container">
@@ -120,10 +121,11 @@
     request, 'admin.instance', instances,
     fields=[
         {"label": "Name", "name": "name"},
-        {"label": "Consumer key", "name": "consumer_key"},
-        {"label": "GUID", "name": "tool_consumer_instance_guid"},
-        {"label": "Deployment ID", "name": "deployment_id"},
+        {"label": "Created", "name": "created"},
         {"label": "Last Launched", "name": "last_launched"},
+        {"label": "Product Family", "name": "tool_consumer_info_product_family_code"},
+        {"label": "Email", "name": "requesters_email"},
+        {"label": "LMS URL", "name": "lms_url"},
     ]
 ) }}
 {% endmacro %}
--- a/lms/templates/admin/macros.html.jinja2
+++ b/lms/templates/admin/macros.html.jinja2
@@ -118,8 +118,8 @@
 {% macro instances_table(request,  instances) %}
 {{ object_list_table(
     request, 'admin.instance', instances,
-    ['Name', 'Consumer key', 'GUID', 'Deployment ID'],
-    ['name', 'consumer_key', 'tool_consumer_instance_guid', 'deployment_id']
+    ['Name', 'Consumer key', 'GUID', 'Deployment ID', 'Last Launched'],
+    ['name', 'consumer_key', 'tool_consumer_instance_guid', 'deployment_id', 'last_launched']
 ) }}
 {% endmacro %}
 
--- a/lms/templates/admin/macros.html.jinja2
+++ b/lms/templates/admin/macros.html.jinja2
@@ -79,15 +79,15 @@
     {% endif %}
 {% endmacro %}
 
-{% macro object_list_table(request, route, objects, headers, fields) %}
+{% macro object_list_table(request, route, objects, fields) %}
 <div class="container">
     <div class="table-container">
         <table class="table is-fullwidth">
           <thead>
             <tr>
               <th></th>
-              {% for heading in headers %}
-              <th>{{heading}}</th>
+              {% for field in fields %}
+              <th>{{field.label}}</th>
               {% endfor %}
             </tr>
           </thead>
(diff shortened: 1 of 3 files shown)


lms/templates/admin/macros.html.jinja2
```diff
@@ -118,8 +118,8 @@
 {% macro instances_table(request,  instances) %}
 {{ object_list_table(
     request, 'admin.instance', instances,
-    ['Name', 'Consumer key', 'GUID', 'Deployment ID'],
-    ['name', 'consumer_key', 'tool_consumer_instance_guid', 'deployment_id']
+    ['Name', 'Consumer key', 'GUID', 'Deployment ID', 'Last Launched'],
+    ['name', 'consumer_key', 'tool_consumer_instance_guid', 'deployment_id', 'last_launched']
```
marcospri, line 122: I'd say that in the context of listing multiple AI the least differentiating attribute would be the GUID.

marcospri (approved):
:+1: 

We could wait for the answer around the table view of AIs. Or not, that would be a small PR to follow up.

lms/templates/admin/macros.html.jinja2
```diff
@@ -115,17 +116,29 @@
...
-    ['Name', 'Consumer key', 'GUID', 'Deployment ID'],
-    ['name', 'consumer_key', 'tool_consumer_instance_guid', 'deployment_id']
+    fields=[
+        {"label": "Name", "name": "name"},
+        {"label": "Created", "name": "created"},
+        {"label": "Last Launched", "name": "last_launched"},
+        {"label": "Product Family", "name": "tool_consumer_info_product_family_code"},
+        {"label": "Email", "name": "requesters_email"},
+        {"label": "LMS URL", "name": "lms_url"},
+    ]
```
jon-betts, line 129: Originally I made this change with the idea of adding an optional "format" argument on the end with a function and passing in a function or something similar. We might yet do that, but we don't in this PR.

I actually think this way of doing things has some advantages. It's easier to see which fields and names are correlated and it's much easier to re-order the columns if you want to.

lms/templates/admin/organization.html.jinja2
```diff
@@ -48,12 +48,7 @@
         </div>
 
         {% if org.application_instances %}
-            {{ macros.object_list_table(
-                    request, 'admin.instance', org.application_instances,
-                    ['Consumer key', 'Deployment ID', 'Tool name', 'URL', "Email"],
-                    ['consumer_key', 'deployment_id', 'tool_consumer_instance_name', 'lms_url', 'requesters_email']
-               )
-            }}
+            {{ macros.instances_table(request, org.application_instances) }}
```
jon-betts, line 51: I don't think these were different on purpose?

marcospri: I reckon they were different on purpose, not sure if the selection of columns made much sense.

The idea was that while you were in an organization you already have some context about things that are potentially the same across all the AIs in it.

GUID is likely to be the same for example. 

It will probably make more sense to revisit the common columns like you are doing instead of trying to fine-tune them for every view.

marcospri (approved):

lms/templates/admin/organization.html.jinja2
jon-betts: Oh ok! Well hopefully this is better for now?

PR #4971, opened 2023-02-03
Explicitly tag sentry reports with AI.id

marcospri:
Tag sentry reports with the ApplicationInstance ID

Useful to find specific reports while investigating issues at a particular institution.

See a tagged event on the testing project:

https://hypothesis.sentry.io/issues/?project=5952630&query=application_instance_id%3A8&referrer=issue-list&sort=freq&statsPeriod=30d

### DIFF
--- a/lms/security.py
+++ b/lms/security.py
@@ -3,6 +3,7 @@ from enum import Enum
 from functools import lru_cache, partial
 from typing import Callable, List, NamedTuple, Optional
 
+import sentry_sdk
 from pyramid.request import Request
 from pyramid.security import Allowed, Denied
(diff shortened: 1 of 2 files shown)


lms/security.py
```diff
@@ -238,6 +239,9 @@ def get_lti_user(request) -> Optional[LTIUser]:
         # LTI users and the corresponding user in H
         request.find_service(UserService).upsert_user(lti_user)
 
+        # Attach useful information to sentry in case we get an exception further down the line
+        set_tag("application_instance_id", lti_user.application_instance_id)
```
marcospri, line 243: We could also send the organization ID but that's more awkward to do right now. Let's start with just the AI.id

PR #4974, opened 2023-02-03
Switch type-checking from opt-in to opt-out

robertknight:
This will catch mistakes where code is unintentionally omitted from checking. It also makes the lms TS setup more similar to the client's.

### DIFF
--- a/lms/static/scripts/browser_check/.babelrc
+++ b/lms/static/scripts/browser_check/.babelrc
@@ -4,6 +4,7 @@
 // This baseline here is based on browsers that support `<script type="module">`.
 {
   "presets": [
+    "@babel/preset-typescript",
     ["@babel/preset-env", {
        "targets": {
          "chrome": "61",
--- a/lms/static/scripts/browser_check/index.ts
+++ b/lms/static/scripts/browser_check/index.ts
@@ -12,7 +12,7 @@ if (!isBrowserSupported()) {
   <button class="Button">Dismiss</button>
 `;
 
-  const dismissButton = browserWarning.querySelector('button');
+  const dismissButton = browserWarning.querySelector('button')!;
   dismissButton.onclick = () => {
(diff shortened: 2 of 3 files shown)


lms/static/scripts/browser_check/index.ts
```diff
@@ -12,7 +12,7 @@ if (!isBrowserSupported()) {
   <button class="Button">Dismiss</button>
 `;
 
-  const dismissButton = browserWarning.querySelector('button');
+  const dismissButton = browserWarning.querySelector('button')!;
```
robertknight, line 15: This file had previously been excluded by mistake.

lms/static/scripts/browser_check/.babelrc
```diff
@@ -4,6 +4,7 @@
 // This baseline here is based on browsers that support `<script type="module">`.
 {
   "presets": [
+    "@babel/preset-typescript",
```
robertknight, line 7: `index.ts` in this directory was being parsed as JavaScript. That happened to work because, until now, it didn't actually contain any TS-specific syntax.

lms/static/scripts/tsconfig.json
```diff
@@ -12,6 +12,15 @@
     "target": "ES2020",
     "useUnknownInCatchVariables": false
   },
-  "include": ["frontend_apps/**/*.js", "frontend_apps/**/*d.ts"],
-  "exclude": ["frontend_apps/**/test/*.js"]
+  "include": ["**/*.js", "**/*.ts", "**/*.tsx"],
+  "exclude": [
+    "new-application-instance.js",
```
robertknight, line 17: `new-application-instance.js` is going away very soon. See https://github.com/hypothesis/lms/pull/4975.

lyzadanger (approved):
LGTM

PR #4975, opened 2023-02-03
Remove new-application-instance.js

robertknight:
The templates that use this JS were removed in https://github.com/hypothesis/lms/pull/4876.


jon-betts (approved):
Good catch! We don't have any JS in the new pages, so it didn't occur to me to look for it in the old.

PR #4976, opened 2023-02-03
Mark package.json as private

robertknight:
This indicates to tools and readers that we don't publish this package, it just exists to track dependencies and declare metadata used by various tools. Also remove the obsolete description field and correct the incorrect license ID.

h's package.json is similar.


lyzadanger (approved):
👍🏻

PR #4978, opened 2023-02-03
D2l course copy files

marcospri:


lms/services/d2l_api/client.py
```diff
@@ -193,6 +193,7 @@ def _find_files(self, course_id, modules):
                 {
                     "type": "File",
                     "display_name": topic["display_name"],
+                    "lms_id": topic["id"],
```
marcospri, line 196: Keep the ID, not only the "d2l://" format.

lms/views/api/d2l/files.py
```diff
@@ -29,15 +30,37 @@ def list_files(_context, request):
     permission=Permissions.API,
 )
 def via_url(_context, request):
+    course_copy_plugin = request.product.plugin.course_copy
```
marcospri, line 33: This is very similar to the Blackboard version. I think we'll have an opportunity to refactor this when we refactor the Canvas version.

lms/product/d2l/_plugin/course_copy.py
```diff
@@ -0,0 +1,47 @@
+from lms.product.plugin.course_copy import CourseCopyFilesHelper
+from lms.services.d2l_api import D2LAPIClient
+from lms.services.file import FileService
+
+
+class D2LCourseCopyPlugin:
```
marcospri, line 6: Very similar, almost identical to the blackboard version.

seanh (approved):
Nothing to say, looks good

PR #4984, opened 2023-02-06
Add AssignmentService.get_copied_from_assignment

marcospri:
This makes the behavior of DocumentUrlService more explicit around course copy.

Ideally the final state of this would be to have each of the possible parameter locations be part of the individual product/product plugins instead of trying them in succession. Unfortunately the exact behaviour we rely on doesn't seem to be documented in D2L / Blackboard and we might be relying in it for other smaller LMS's even if it's not documented here.

Keeping the same behaviour for now in all LMS's while making the course copy aspect of it more apparent.

## Testing:

These all should point to the same document after:

```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment cascade"```

```make devdata```

### Blackboard

- Original

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_36_1&course_id=_19_1

- Copy

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1367_1&course_id=_139_1

D2L 

- Original

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2132/View

- Copy

https://aunltd.brightspacedemo.com/d2l/le/content/6855/viewContent/2409/View?ou=6855

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Optional
 
 from sqlalchemy.orm import Session
 
@@ -31,6 +31,27 @@ class AssignmentService:
             )
             .one_or_none()
         )
+
+    def get_copied_from_assignment(self, lti_params) -> Optional[Assignment]:
+        """Return the assignment that the current assignment was copied from."""
+
+        resource_link_history_params = [
+            "resource_link_id_history",  # Blackboard, LTI 1.1
+            "ext_d2l_resource_link_id_history",  # D2L, LTI 1.1
+            "custom_ResourceLink.id.history",  # Blackboard and D2L, LTI 1.3
+        ]
+
+        for param in resource_link_history_params:
+            if historical_resource_link_id := lti_params.get(param):
+                if historical_assignment := self.get_assignment(
+                    tool_consumer_instance_guid=lti_params.get(
+                        "tool_consumer_instance_guid"
+                    ),
+                    resource_link_id=historical_resource_link_id,
+                ):
+                    return historical_assignment
+
+        return None
 
     # pylint: disable=too-many-arguments
     def upsert_assignment(
--- a/lms/services/document_url.py
+++ b/lms/services/document_url.py
@@ -90,23 +90,20 @@ class DocumentURLService:
     def _from_assignment_in_db(self, request):
         """Get a document URL from an assignment in the DB matching a param."""
 
-        for param in (
-            "resource_link_id",  # A normal LTI (non-deep linked) launch
-            "resource_link_id_history",  # A Blackboard course we can copy
-            "ext_d2l_resource_link_id_history",  # Ditto for D2L
-            "custom_ResourceLink.id.history",  # LTI 1.3 location for D2L and blackboard
-        ):
-            if (resource_link_id := request.lti_params.get(param)) and (
-                assignment := self._assignment_service.get_assignment(
-                    tool_consumer_instance_guid=request.lti_params.get(
-                        "tool_consumer_instance_guid"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lms/services/assignment.py [hunk 1]
```diff
@@ -32,6 +32,27 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
             .one_or_none()
         )
 
+    def get_copied_from_assignment(self, lti_params) -> Optional[Assignment]:
```
marcospri, line 35: Not sure about this name

seanh: My first thought was `get_original_assignment()` but "original" might imply the very first original which I'm not sure this necessarily is.

Alternative: `get_historical_assignment()` to match the the wording used in the `resource_link_id_history`, `ext_d2l_resource_link_id_history` and `custom_ResourceLink.id.history` param names.

lms/services/assignment.py
```diff
@@ -32,6 +32,27 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
...
+        """Get the assignment the current launch was copied from."""
+
+        resource_link_history_params = [
+            "resource_link_id_history",  # Blackboard  LTI1.1
+            "ext_d2l_resource_link_id_history",  # Ditto for D2L
+            "custom_ResourceLink.id.history",  # LTI 1.3 location for D2L, blackboard
```
seanh, line 41:
```suggestion
            "resource_link_id_history",  # Blackboard, LTI 1.1
            "ext_d2l_resource_link_id_history",  # D2L, LTI 1.1
            "custom_ResourceLink.id.history",  # Blackboard and D2L, LTI 1.3
```

lms/services/assignment.py
```diff
@@ -32,6 +32,27 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
...
+        ]
+
+        for param in resource_link_history_params:
+            if copied_from_resource_link_id := lti_params.get(param):
```
seanh, line 45:
```suggestion
            if historical_resource_link_id := lti_params.get(param):
```

lms/services/assignment.py
```diff
@@ -32,6 +32,27 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
             .one_or_none()
         )
 
+    def get_copied_from_assignment(self, lti_params) -> Optional[Assignment]:
+        """Get the assignment the current launch was copied from."""
```
seanh, line 36:
```suggestion
        """Return the assignment that the current assignment was copied from."""
```

lms/services/assignment.py
```diff
@@ -32,6 +32,27 @@ def get_assignment(self, tool_consumer_instance_guid, resource_link_id):
...
+
+        resource_link_history_params = [
+            "resource_link_id_history",  # Blackboard  LTI1.1
+            "ext_d2l_resource_link_id_history",  # Ditto for D2L
+            "custom_ResourceLink.id.history",  # LTI 1.3 location for D2L, blackboard
+        ]
+
+        for param in resource_link_history_params:
+            if copied_from_resource_link_id := lti_params.get(param):
+                if assignment := self.get_assignment(
```
seanh, line 46:
```suggestion
                if historical_assignment := self.get_assignment(
```

seanh (approved):

lms/services/assignment.py
marcospri: I like the `copied` part of the name, not 100% convinced though.

PR #4988, opened 2023-02-07
Show error for copied Blackboard and D2L groups assignments.

marcospri:
upsert_assignment already takes lti_params as one of the mandatory parameters.

Use LTIParams in combination with the new get_copied_from_assignment to, in new assignments to copy some of the configuration of the old assignment over to he new one.

We only copy the value of `group_set_id` for now. As a result copied assignments will still be group ones after copy. They will point to an unavailable group set for now.

We don't look at document_url for now as that handled separately at the moment over in DocumentService.

####  Testing

- Reset assignment (once per branch):

 ```tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment cascade"; make devdata```

- Now both in `main` and this branch (`course-copy-group-set`) try:

#### D2L

https://aunltd.brightspacedemo.com/d2l/le/content/6855/viewContent/2409/View?ou=6855

#### Blackboard 

https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_1373_1&course_id=_139_1

On main, both will launch successfully but they'll become "course assignments", ie the group drop-down just shows the course group.

On this branch the "can't find group set" will appear.

(Canvas is not affected by this as the group set is part of the URL parameters sent by canvas itself)

### DIFF
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Optional
 
 from sqlalchemy.orm import Session
 
@@ -31,6 +31,27 @@ class AssignmentService:
             )
             .one_or_none()
         )
+
+    def get_copied_from_assignment(self, lti_params) -> Optional[Assignment]:
+        """Return the assignment that the current assignment was copied from."""
+
+        resource_link_history_params = [
+            "resource_link_id_history",  # Blackboard, LTI 1.1
+            "ext_d2l_resource_link_id_history",  # D2L, LTI 1.1
+            "custom_ResourceLink.id.history",  # Blackboard and D2L, LTI 1.3
+        ]
+
+        for param in resource_link_history_params:
+            if historical_resource_link_id := lti_params.get(param):
+                if historical_assignment := self.get_assignment(
+                    tool_consumer_instance_guid=lti_params.get(
+                        "tool_consumer_instance_guid"
+                    ),
+                    resource_link_id=historical_resource_link_id,
+                ):
+                    return historical_assignment
+
+        return None
 
     # pylint: disable=too-many-arguments
     def upsert_assignment(
--- a/lms/services/assignment.py
+++ b/lms/services/assignment.py
@@ -62,7 +62,7 @@ class AssignmentService:
         lti_params: LTIParams,
         is_gradable=False,
         extra=None,
-    ):
+    ) -> Assignment:
         """
         Update or create an assignment with the given document_url.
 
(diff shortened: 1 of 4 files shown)


lms/services/assignment.py
```diff
@@ -71,7 +71,12 @@ def upsert_assignment(
         or create a new one if none exist on the DB.
 
         Any existing document_url for this assignment will be overwritten.
+
+        When a new assignment is created values of `extra`
```
marcospri, line 75: This is a weird sentence, could use some help with rewording.

lms/services/assignment.py
```diff
@@ -82,11 +87,21 @@ def upsert_assignment(
             )
             self._db.add(assignment)
 
+            # For new assignments check if we are copying
+            # from an existing one on the current launch
+            historical_assignment = self.get_copied_from_assignment(lti_params)
+            if (
+                historical_assignment
+                and historical_assignment.extra.get("group_set_id")
+                and not extra.get("group_set_id")
```
marcospri, line 96: If we get passed one value directly, use that one. I reckon that the least surprising behavior.

lms/services/assignment.py
seanh: > If we detect that the new assignment in the LMS has been copied from a historical assignment (perhaps by using the LMS's "course copy" feature) then some of the new assignment's `extra` values might be taken from the historical assignment.

lms/services/assignment.py
```diff
@@ -82,11 +87,21 @@ def upsert_assignment(
...
+            ):
+                extra["group_set_id"] = historical_assignment.extra.get("group_set_id")
+
         assignment.document_url = document_url
         assignment.title = lti_params.get("resource_link_title")
         assignment.description = lti_params.get("resource_link_description")
         assignment.is_gradable = is_gradable
-        assignment.extra = extra if extra else assignment.extra or {}
+        assignment.extra = extra if extra else assignment.extra
```
seanh, line 104: This might be a bit misleading because there's an unconditional assignment statement here but it actually doesn't always change the attribute. Might be clearer if just done the "normal" way:

```suggestion
        if extra:
            assignment.extra = extra
```

seanh (approved):
:+1: Nice and small PR, the changes look reasonable to me. Testing instructions worked as expected for me.

I was a little confused at first by what the purpose of this PR is, since the PR title describes _what_ the PR does (or how it does it) not _why_: _Make `AssignmentService.upsert_assignment()` aware of course copy_. It wasn't until I got to the bottom of the testing instructions that I saw that this fixes copied Blackboard and D2L groups assignments to at least show an error rather than silently changing into course assignments. The PR could have been titled something like _Show error for copied Blackboard and D2L groups assignments_.

PR #4993, opened 2023-02-07
Record courses in the DB on configuration

marcospri:
Create a course in the DB as soon as we start configuring an assignment in it.

This brings the behavior in line with the deeplinking counterpart where this was already the case, the deeplinking request creates the course.

The new behavior allows APIs that are called during configuration (files, groups) to rely on the presence of the course.

### Testing 

- On `main`, reset the DB state:

```
tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment,grouping cascade"; make devdata
```

- Launch https://aunltd-test.blackboard.com/webapps/blackboard/execute/blti/launchPlacement?blti_placement_id=_20_1&content_id=_183_1&course_id=_19_1

do not configure it.

- Check the DB for groupings:

```
select * from grouping;
(0 rows)
```

- Repeat the process in this branch `store-course`

```
select * from grouping;
-[ RECORD 1 ]-----------+----------------------------------------------------------------
created                 | 2023-02-07 13:44:05.447744
updated                 | 2023-02-07 13:44:05.447744
id                      | 426
application_instance_id | 110
authority_provided_id   | 026ffd48cdfbe1c6f32a4f434390cbc12c8b4292
parent_id               | 
lms_id                  | c494dc384def49bfb66b9dd356010a63
lms_name                | COPIED Test Course with Original Course View
settings                | {"blackboard": {"files_enabled": true, "groups_enabled": true}}
extra                   | {}
```

now the course is created.

### DIFF
--- a/lms/views/lti/basic_launch.py
+++ b/lms/views/lti/basic_launch.py
@@ -50,7 +50,7 @@ class BasicLaunchViews:
         self.context.application_instance.check_guid_aligns(
             self.request.lti_params.get("tool_consumer_instance_guid")
         )
-
+        self._record_course()
         self._record_launch()
 
     @view_config(
@@ -108,7 +108,6 @@ class BasicLaunchViews:
             form_action=self.request.route_url("configure_assignment"),
             form_fields=form_fields,
         )
-
         return {}
 
     @view_config(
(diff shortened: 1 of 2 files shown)


jon-betts (approved):
This is good. It's another small way in which we are finding all of these journeys really are variations of the same thing.

PR #4994, opened 2023-02-07
Use a single name for searching for instances which matches the pattern

jon-betts:
For:

 * https://github.com/hypothesis/lms/issues/4998

Previously we had `GET:admin.instances` to start the search and `POST:admin.instances.search` for the callback. In all other cases we have the same route and distinguish by HTTP method. This keeps the two halves of the search more obviously intertwined. I've also named them similarly and put them next to each other.

We have `admin.instance.blah` for pretty much everything else, so I've made this `admin.instance.search` and renamed the template accordingly.

### DIFF
--- a/lms/routes.py
+++ b/lms/routes.py
@@ -129,8 +129,8 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route("vitalsource_api.launch_url", "/api/vitalsource/launch_url")
 
     config.add_route("admin.index", "/admin/")
-    config.add_route("admin.instances", "/admin/instances/")
-    config.add_route("admin.instances.search", "/admin/instances/search")
+
+    config.add_route("admin.instance.search", "/admin/instances/")
     config.add_route("admin.instance.new", "/admin/instance/new")
     config.add_route("admin.instance.upgrade", "/admin/instance/upgrade")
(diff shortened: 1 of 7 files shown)


lms/routes.py
```diff
@@ -129,8 +129,8 @@ def includeme(config):  # pylint:disable=too-many-statements
     config.add_route("vitalsource_api.launch_url", "/api/vitalsource/launch_url")
 
     config.add_route("admin.index", "/admin/")
-    config.add_route("admin.instances", "/admin/instances/")
-    config.add_route("admin.instances.search", "/admin/instances/search")
+
+    config.add_route("admin.instance.search", "/admin/instances/")
```
marcospri, line 133: We seem to use the plural for this in the registrations and organizations.

Not that important but we could settle in one for all of them.

marcospri (approved):
Yeah this seems to make more sense. AIs were the first entity here so the might be the one with more strange choices.

Did some light testing on the search, all looks good.

lms/routes.py
jon-betts: Yes, I just didn't want to go on a rampage. The next thing I wanted to do was to potentially put these into a directory structure with the same name.

This would prevent us from having to have `/instance` and `/instances` with a single thing inside.

PR #4999, opened 2023-02-08
Add `useConfig` hook to read frontend configuration from UI components

robertknight:
This PR adds a `useConfig` hook to simplify reading configuration embedded into the page by the backend, from within UI components. It is a simple wrapper around the previous `useContext(Config)` pattern, which checks that certain app/route-specific keys are provided and throws a sensible error if missing.

I have converted the existing uses of `useContext(Config)` to use this new hook.

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
+++ b/lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
@@ -1,8 +1,6 @@
 import { Link } from '@hypothesis/frontend-shared';
-import { useContext } from 'preact/hooks';
 
-import { Config } from '../config';
-
+import { useConfig } from '../config';
 import ErrorModal from './ErrorModal';
 
 /**
@@ -14,12 +12,12 @@ import ErrorModal from './ErrorModal';
  * (e.g. OAuth2RedirectErrorApp, LaunchErrorDialog).
  */
 export default function ErrorDialogApp() {
-  const { errorDialog } = useContext(Config);
+  const { errorDialog } = useConfig(['errorDialog']);
 
   const error = {
-    errorCode: errorDialog?.errorCode,
-    details: errorDialog?.errorDetails ?? '',
-    message: errorDialog?.errorMessage ?? '',
+    errorCode: errorDialog.errorCode,
+    details: errorDialog.errorDetails ?? '',
+    message: errorDialog.errorMessage ?? '',
   };
 
   let description;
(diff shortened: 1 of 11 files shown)


lms/static/scripts/frontend_apps/components/ErrorDialogApp.js
```diff
@@ -14,12 +12,12 @@ import ErrorModal from './ErrorModal';
  * (e.g. OAuth2RedirectErrorApp, LaunchErrorDialog).
  */
 export default function ErrorDialogApp() {
-  const { errorDialog } = useContext(Config);
+  const { errorDialog } = useConfig(['errorDialog']);
```
robertknight, line 15: The `errorDialog` field should always be set by the backend when this app mode is used. See `enable_error_dialog_mode`.

lms/static/scripts/frontend_apps/components/OAuth2RedirectErrorApp.js
```diff
@@ -21,15 +19,14 @@ import ErrorModal from './ErrorModal';
...
-    errorDetails = '',
-    canvasScopes = /** @type {string[]} */ ([]),
-  } = OAuth2RedirectError;
+    OAuth2RedirectError: {
+      authUrl = null,
+      errorCode,
+      errorDetails = '',
+      canvasScopes = /** @type {string[]} */ ([]),
+    },
+  } = useConfig(['OAuth2RedirectError']);
```
robertknight, line 29: The `OAuth2RedirectError` key should always be set by the backend when the corresponding app mode is set. See `enable_oauth2_redirect_error_mode` in the backend.

lms/static/scripts/frontend_apps/config.ts
```diff
@@ -249,6 +250,30 @@ const defaultConfig = { api: {} } as unknown;
...
+ * is omitted, otherwise it would change all optional keys to required.
+ */
+export function useConfig<R extends keyof ConfigObject = never>(
+  requiredKeys: R[] = []
+): Ensure<ConfigObject, R> {
+  const config = useContext(Config);
+  for (const key of requiredKeys) {
+    if (!(key in config)) {
+      throw new Error(`Required configuration key "${key}" not set`);
+    }
```
acelaya, line 272: I guess it's mostly an edge case, but this condition won't trigger if the key is actually there but it has value `undefined`.

image

Should we maybe check `!config[key]` instead?

acelaya (approved):
Looks good. Just noticed a possible edge case that we might want to handle.

lms/static/scripts/frontend_apps/config.ts
robertknight: Settings might legitimately have falsey values (eg. `false` for a disabled flag or `0` for a zero count). I chose to define "field is set" to mean that the object has a property here, though I think comparison to `undefined` would work just as well.  We _might_ find that it will be more convenient to treat all nullish values as "not set" in future. That would allow Python code to do something like `config["foo"] = some_value if condition else None`.

PR #5004, opened 2023-02-09
Add a server side (postgres) default for application instance create

jon-betts:
I think this was missed as we introduced the `CreatedUpdateMixin` which uses server side defaults over an existing column with Python defaults. I used an automatic alembic upgrade which didn't seem to spot the change.

This should fix problems with the create date with `make devdata`.


marcospri (approved):
Yeah, alembic probably was not smart enough to make the update.

This works :+1:

PR #5005, opened 2023-02-09
Take only `context_id` in `api.courses.group_sets.list`

marcospri:
Based on `context_id` we can get the course from the DB get the relevant data from there.

`course_id` was a different ID only in canvas and the grouping plugin was already aware of that difference.

# Testing

- There should not be behavior changes here but as sanity check we'll see if retrieving the group sets is still working on all LMS's

- Clear any local state in the DB: `tox -qe dockercompose -- exec postgres psql -U postgres -c "truncate assignment,grouping cascade"; make devdata`

### Canvas

https://hypothesis.instructure.com/courses/125/assignments/3930/edit?name=Marcos+Assignment&due_at=null&points_possible=0

(pick the right LTI tool, configure with any document and check the `This is a group assignment`)

Getting the drop-down to populate with a list of group sets is what we are after.

### Blackboard 

Same process, configure and check the groups checkbox.

### D2L

Same story again: 

https://aunltd.brightspacedemo.com/d2l/le/content/6782/viewContent/2431/View

(there are files in "scratch" area)

### DIFF
--- a/lms/product/blackboard/_plugin/grouping.py
+++ b/lms/product/blackboard/_plugin/grouping.py
@@ -22,8 +22,8 @@ class BlackboardGroupingPlugin(GroupingPlugin):
     def __init__(self, blackboard_api):
         self._blackboard_api = blackboard_api
 
-    def get_group_sets(self, course_id):
-        return self._blackboard_api.course_group_sets(course_id)
+    def get_group_sets(self, course):
+        return self._blackboard_api.course_group_sets(course.lms_id)
 
     def get_groups_for_learner(self, _svc, course, group_set_id):
         if learner_groups := self._blackboard_api.course_groups(
--- a/lms/product/canvas/_plugin/grouping.py
+++ b/lms/product/canvas/_plugin/grouping.py
@@ -47,8 +47,8 @@ class CanvasGroupingPlugin(GroupingPlugin):
 
         return [sec for sec in course_sections if sec["id"] in learner_section_ids]
 
-    def get_group_sets(self, course_id):
-        return self._canvas_api.course_group_categories(course_id)
+    def get_group_sets(self, course):
+        return self._canvas_api.course_group_categories(self._custom_course_id(course))
 
     def get_groups_for_learner(self, _svc, course, group_set_id):
         # For learners, the groups they belong within the course
--- a/lms/resources/_js_config/__init__.py
+++ b/lms/resources/_js_config/__init__.py
@@ -444,16 +444,11 @@ class JSConfig:
         }
 
         if self._request.product.settings.groups_enabled:
-            # This should be abstracted as "lms_course_id" or similar in either
-            # LTIParams or as part of product.
-            api_course_id = self._request.lti_params.get(
-                "custom_canvas_course_id"
-            ) or self._request.lti_params.get("context_id")
-
             product_info["api"]["listGroupSets"] = {
                 "authUrl": self._request.route_url(product.route.oauth2_authorize),
                 "path": self._request.route_path(
-                    "api.courses.group_sets.list", course_id=api_course_id
+                    "api.courses.group_sets.list",
+                    course_id=self._request.lti_params["context_id"],
                 ),
                 "data": {
                     "lms": {
--- a/lms/views/api/groups.py
+++ b/lms/views/api/groups.py
@@ -11,6 +11,6 @@ from lms.security import Permissions
 )
 def course_group_sets(_context, request):
     course = request.find_service(name="course").get_by_context_id(
-        request.matchdict["course_id"]
+        request.matchdict["course_id"], raise_on_missing=True
     )
     return request.product.plugin.grouping.get_group_sets(course)
--- a/lms/views/api/groups.py
+++ b/lms/views/api/groups.py
@@ -10,6 +10,7 @@ from lms.security import Permissions
     route_name="api.courses.group_sets.list",
 )
 def course_group_sets(_context, request):
-    return request.product.plugin.grouping.get_group_sets(
+    course = request.find_service(name="course").get_by_context_id(
         request.matchdict["course_id"]
     )
+    return request.product.plugin.grouping.get_group_sets(course)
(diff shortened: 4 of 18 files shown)


lms/views/api/groups.py
```diff
@@ -1,15 +1,27 @@
...
 from pyramid.view import view_config
+from webargs import fields
 
 from lms.security import Permissions
+from lms.validation._base import PyramidRequestSchema
+
+
+class APIGroupSetsSchema(PyramidRequestSchema):
+    class LMS(Schema):
+        product = fields.Str(required=True)
```
marcospri, line 11: This was already being sent to allow request.product to correctly identify the right one.

lms/views/api/groups.py
```diff
@@ -1,15 +1,27 @@
...
+
+    context_id = fields.Str(required=True)
```
marcospri, line 13: New parameter.

lms/views/api/groups.py
```diff
@@ -1,15 +1,27 @@
...
     permission=Permissions.API,
     renderer="json",
     route_name="api.courses.group_sets.list",
+    schema=APIGroupSetsSchema,
 )
 def course_group_sets(_context, request):
-    return request.product.plugin.grouping.get_group_sets(
-        request.matchdict["course_id"]
+    course = request.find_service(name="course").get_by_context_id(
+        context_id=request.parsed_params["context_id"]
```
marcospri, line 25: Nothing now on request.matchdict

lms/product/canvas/_plugin/grouping.py
```diff
@@ -47,8 +47,8 @@ def get_sections_for_grading(self, _svc, course, grading_student_id):
 
         return [sec for sec in course_sections if sec["id"] in learner_section_ids]
 
-    def get_group_sets(self, course_id):
-        return self._canvas_api.course_group_categories(course_id)
+    def get_group_sets(self, course):
+        return self._canvas_api.course_group_categories(self._custom_course_id(course))
```
marcospri, line 51: We already have the self._custom_course_id helper function here to deal with the difference in Canvas.

lms/resources/_js_config/__init__.py
```diff
@@ -444,21 +444,14 @@ def _get_product_info(self):
         }
 
         if self._request.product.settings.groups_enabled:
-            # This should be abstracted as "lms_course_id" or similar in either
-            # LTIParams or as part of product.
-            api_course_id = self._request.lti_params.get(
-                "custom_canvas_course_id"
-            ) or self._request.lti_params.get("context_id")
```
marcospri, line 451: It's nice to remove this wart here.

lms/product/blackboard/_plugin/grouping.py
```diff
@@ -22,8 +22,8 @@ class BlackboardGroupingPlugin(GroupingPlugin):
     def __init__(self, blackboard_api):
         self._blackboard_api = blackboard_api
 
-    def get_group_sets(self, course_id):
-        return self._blackboard_api.course_group_sets(course_id)
+    def get_group_sets(self, course):
```
marcospri, line 25: We area after context_id in every LMS but canvas. That's course.lms_id

lms/views/api/groups.py
jon-betts: Why do we look-up the course here? Most plugins so far don't want the object, just the raw `context_id`. This seems like we've added a query in most places that we don't have to?

marcospri: We'll soon will use `course` in the plugin itself so I wanted to tackled both the change in the parameter to change the interface.

In retrospect his might not be as clear as the intent is not present in this PR.

lms/resources/_js_config/__init__.py
```diff
@@ -444,21 +444,14 @@ def _get_product_info(self):
...
-
             product_info["api"]["listGroupSets"] = {
                 "authUrl": self._request.route_url(product.route.oauth2_authorize),
-                "path": self._request.route_path(
-                    "api.courses.group_sets.list", course_id=api_course_id
-                ),
+                "path": self._request.route_path("api.courses.group_sets.list"),
```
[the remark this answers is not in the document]
marcospri: Kept it on the URL for now :+1:

lms/resources/_js_config/__init__.py
```diff
@@ -444,21 +444,14 @@ def _get_product_info(self):
...
                 "data": {
                     "lms": {
                         "product": self._request.product.family,
-                    }
+                    },
+                    "context_id": self._request.lti_params["context_id"],
```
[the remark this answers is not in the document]
marcospri: Keep it in the URL for now.

We probably should build a common set of parameters for all/most of our internal API endpoints, we might not need `context_id` everywhere but why not just send it along product, resource_link_id and some others. With a documented schema etc it will be much clearer what that we have available.

I'll leave that for other PR in any case.

lms/views/api/groups.py
```diff
@@ -10,6 +10,7 @@
     route_name="api.courses.group_sets.list",
 )
 def course_group_sets(_context, request):
-    return request.product.plugin.grouping.get_group_sets(
+    course = request.find_service(name="course").get_by_context_id(
         request.matchdict["course_id"]
     )
```
jon-betts, line 15: Course can be `None` here, which I think is an error in all cases. If it is, then the result is a `None` has no attribute `lms_id` type of error, which is unhandled.

I think we have to deal with this here?

jon-betts (changes requested):
I've tried this out and the happy path works. I got stuck in some weird situation we can't replicate. I'm not exactly happy about not knowing how it happened, but I think we can't do much about it.

What it did point out is that `get_by_context_id` is a `one_or_none()` type of method, and for us the course is mandatory at this point. I think we should be treating getting `None` as an exception instead of letting the error bubble up elsewhere in the plugins.

marcospri:
Deleted an empty test file (/api vs root level)

jon-betts (approved):
Ooo. Looks like we wanted to raise this in quite a few places. 

This all looks good, but I had no idea we wanted it this way around most of the time. This PR should go through, but a quick scan to see if the default might be better the other way around might be good?

PR #5028, opened 2023-02-13
Update `ContentSelector` to use non-deprecated components from pattern library

acelaya:
Refactor `ContentSelector` to replace deprecated components from the pattern library by their "next" equivalents.

This PR is part of #5013

### DIFF
--- a/lms/static/scripts/frontend_apps/components/ContentSelector.tsx
+++ b/lms/static/scripts/frontend_apps/components/ContentSelector.tsx
@@ -1,4 +1,4 @@
-import { FullScreenSpinner, LabeledButton } from '@hypothesis/frontend-shared';
+import { Button, SpinnerOverlay } from '@hypothesis/frontend-shared/lib/next';
 import { useMemo, useState } from 'preact/hooks';
 
 import type { Book, File, Chapter } from '../api-types';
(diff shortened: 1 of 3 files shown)


lms/static/scripts/frontend_apps/components/ContentSelector.tsx
```diff
@@ -237,79 +237,78 @@ export default function ContentSelector({
...
   return (
     <>
-      {isLoadingIndicatorVisible && <FullScreenSpinner />}
+      {isLoadingIndicatorVisible && <SpinnerOverlay />}
       <div className="flex flex-row p-y-2">
         <div className="flex flex-col space-y-1.5">
-          <LabeledButton
+          <Button
             onClick={() => selectDialog('url')}
-            type="button"
```
acelaya, line 245: This is now the default value.

yarn.lock
```diff
@@ -1127,9 +1127,9 @@
     glob "^7.2.0"
 
 "@hypothesis/frontend-shared@^5.7.0":
-  version "5.9.0"
-  resolved "https://registry.yarnpkg.com/@hypothesis/frontend-shared/-/frontend-shared-5.9.0.tgz#1ccd6007e237c90fee77689417bfd19a0fde0498"
-  integrity sha512-FyeTvDLJ55Ma676f3pJoy1Ybst++eiNnDAuXgIH0tlYBE+OTg2paQBnC7evb19E9O1m0m9zpTZV0TKjjQTF7SA==
+  version "5.9.1"
```
acelaya, line 1130: Upgrading to latest version of the pattern library, which makes sure the default value for `type` attribute in buttons is `button`.

acelaya:
> `frontend-shared` is at `5.9.1` on `main` now, so you'll want to remove the commit that bumps it from this branch

Done!

lyzadanger:
We've got a bit of a problem here, one we knew was coming.

Before these changes in Canvas, with a bunch of features enabled:

image

After:

image

I see an immediate workaround, but I'll need to follow up on this overall situation ASAP.

Suggested workaround: edit `_FilePickerApp.scss` (one of the very last remaining bits of non-Tailwind CSS in our apps), change `.FilePickerApp__form` to set `margin-top` to `1rem` instead of `10vh`. This results in this layout:

image

I think that is acceptable for the moment, but please test through the whole assignment-configuration workflow in both Canvas and another LMS (BlackBoard is an option) and in a couple of browsers.

I'll bring this situation up in the sync today and ask if I can remediate it more better.

lyzadanger (changes requested):
This particular interface is rickety and right on the edge of manageable and needs a redesign (this has been a known issue, but now it's becoming urgent, see my longer comment).

We'll need to reduce the top margin of the overall layout before merging this.

acelaya:
@lyzadanger I have pushed the styles change.

I have also tested it in D2L and Blackboard, and those have more space for the component, so there should be no problem.

lyzadanger (approved):
OK, let's get this landed. I will be following up soon with changes to `FilePickerApp` and this layout.

PR #5029, opened 2023-02-13
Migrate ErrorDialogApp to non-deprecated components from the pattern library

acelaya:
Migrating to TypeScript at the same time, as the component does not really need any changes related to that.

This PR is part of #5013


lyzadanger (approved):
Well, this one is easy!

